PR #9, opened 2020-01-10
fix(client): skip body if method is `head`

dnlup:
In case of HEAD method `client.parser.[HTTPParser.kOnHeadersComplete]`
should tell the parser to skip the body by returning 2.

See https://github.com/creationix/http-parser-js/blob/master/http-parser.js#L357

I ran tests multiple times and on my machine the `lib/pool.js` tests failed once in the git precommit hook. 

I'll look into adding tests also.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -217,10 +217,13 @@ class Client extends EventEmitter {
       // TODO move this[kCallbacks] from being an array. The array allocation
       // is showing up in the flamegraph.
       const cb = this[kCallbacks].shift()
-      this._needHeaders--
-      this._lastBody = new Readable({ read: this[kRead].bind(this) })
-      this._lastBody.push = this[kRequests].shift().wrapSimple(this._lastBody, this._lastBody.push)
+      const request = this[kRequests].shift()
+      const skipBody = request.method === 'HEAD'
 
+      if (!skipBody) {
+        this._lastBody = new Readable({ read: this[kRead].bind(this) })
+        this._lastBody.push = request.wrapSimple(this._lastBody, this._lastBody.push)
+      }
       cb(null, {
         statusCode,
         headers: parseHeaders(headers),
@@ -230,6 +233,8 @@ class Client extends EventEmitter {
       if (this.closed && this[kQueue].length() === 0) {
         this.destroy()
       }
+
+      return skipBody
     }
 
     this.parser[HTTPParser.kOnBody] = (chunk, offset, length) => {
@@ -239,7 +244,9 @@ class Client extends EventEmitter {
     this.parser[HTTPParser.kOnMessageComplete] = () => {
       const body = this._lastBody
       this._lastBody = null
-      body.push(null)
+      if (body !== null) {
+        body.push(null)
+      }
     }
 
     this[kReadCb] = () => {
--- a/test/client.js
+++ b/test/client.js
@@ -36,6 +36,31 @@ test('basic get', (t) => {
   })
 })
 
+test('basic head', (t) => {
+  t.plan(7)
+
+  const server = createServer((req, res) => {
+    t.strictEqual('/', req.url)
+    t.strictEqual('HEAD', req.method)
+    t.strictEqual('localhost', req.headers.host)
+    res.setHeader('content-type', 'text/plain')
+    res.end('hello')
+  })
+  t.tearDown(server.close.bind(server))
+
+  server.listen(0, () => {
+    const client = new Client(`http://localhost:${server.address().port}`)
+    t.tearDown(client.close.bind(client))
+
+    client.request({ path: '/', method: 'HEAD' }, (err, { statusCode, headers, body }) => {
+      t.error(err)
+      t.strictEqual(statusCode, 200)
+      t.strictEqual(headers['content-type'], 'text/plain')
+      t.strictEqual(body, null)
+    })
+  })
+})
+
 test('get with host header', (t) => {
   t.plan(7)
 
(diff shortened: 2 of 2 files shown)


test/client.js
```diff
@@ -348,6 +443,64 @@ function makeRequestAndExpectUrl (client, i, t, cb) {
   })
 }
 
+test('20 times HEAD with pipelining 10', (t) => {
```
dnlup: This test fails and I still can't figure out why. If I run manually a client with pipelining = 10 I have no issues.

For example:
```js
const { Client } = require('/path/to/undici')

for (let i = 0; i < 100; i++) {
    client.request({
        path: '/',
        method: 'HEAD',
        pipelining: 10
    }, (error, { statusCode, body }) => {
        console.error(error)
        console.log(statusCode)
    })
}
```
works fine.

dnlup: I used the server in `benchmarks/origins.js` to test.

test/client.js
```diff
@@ -348,6 +443,64 @@ function makeRequestAndExpectUrl (client, i, t, cb) {
...
+  t.plan(3 * num + 1)
+
+  let count = 0
+  let countGreaterThanOne = false
+  const server = createServer((req, res) => {
+    count++
+    setTimeout(function () {
+      countGreaterThanOne = countGreaterThanOne || count > 1
+      res.end(req.url)
+    }, 10)
```
mcollina: I think the problem is related to the timeout, which might happen before receiving more requests.

I would change this test to _not_ end the request until you have seen more than one parallel requests.

dnlup: You mean something like:
```js
    const interval = setInterval(() => {
      countGreaterThanOne = countGreaterThanOne || count > 1
      if (countGreaterThanOne) {
        clearInterval(interval)
        res.end(req.url)
      }
    }, 50)
```
Because neither this is working. It actually fails even if I disable all the checks and the count logic. The error is always 
Screen Shot 2020-01-13 at 18 38 07

I need to investigate further... Now I have no clue 😢

mcollina: I mean to _not rely on timers, intervals, etc_ at all. Rely on the sequence of events. You want to receive _10_ requests in parallel... wait to receive 10 requests before responding to them all.

dnlup: I investigated a bit further and I don't think it is something correlated to use of timers.
First, [the example](https://github.com/mcollina/undici/pull/9#discussion_r365336897) I posted before was wrong. Sorry about that. The client was not being initialized with pipelining.
I am much more inclined to think I am causing an issue with pipelining because using the server in `benchmark/origin.js` to test even something simple like
```js
const client = new Client('http://localhost:3000', {
    pipelining: 2
})

for (let i = 0; i < 4; i++) {
    client.request({
        path: '/',
        method: 'HEAD',
    }, (error, data) => {
        console.log(i)
        console.error(error)
        console.log(data)
        console.log()
    })
}
```
causes the last `n` requests to fail with the error `Error: other side closed - ended` from `undici/lib/client.js:47:23`. In this case is `n = 1` (the last request fails). If I try `100` requests with a pipelining factor of `10` the last `9` requests fail with the same error and so on...

This is the output of the server in debug mode responding to  4 HEAD requests with pipelining 2

Screen Shot 2020-01-16 at 17 47 37

While this is 4 GET requests with pipelining 2

Screen Shot 2020-01-16 at 17 47 01

I did try to remove the use of timers in the test, but I think the problem relies in my changes.

lib/client.js
```diff
@@ -217,9 +217,10 @@ class Client extends EventEmitter {
       // TODO move this[kCallbacks] from being an array. The array allocation
       // is showing up in the flamegraph.
       const cb = this[kCallbacks].shift()
+      const request = this[kRequests].shift()
       this._needHeaders--
       this._lastBody = new Readable({ read: this[kRead].bind(this) })
-      this._lastBody.push = this[kRequests].shift().wrapSimple(this._lastBody, this._lastBody.push)
+      this._lastBody.push = request.wrapSimple(this._lastBody, this._lastBody.push)
```
mcollina: assigning `_lastBody` should be skipped if it's a `HEAD` because it has no body. I _think_ you might need to call `this[kRead]` anyway to kickstart reading the next response.

lib/client.js
```diff
@@ -217,10 +217,15 @@ class Client extends EventEmitter {
...
       // is showing up in the flamegraph.
       const cb = this[kCallbacks].shift()
+      const request = this[kRequests].shift()
+      let skipBody
       this._needHeaders--
-      this._lastBody = new Readable({ read: this[kRead].bind(this) })
-      this._lastBody.push = this[kRequests].shift().wrapSimple(this._lastBody, this._lastBody.push)
-
+      if (request.method === 'HEAD') {
+        skipBody = 1
```
dnlup: So by returning `1` or `true` the parser finishes all the requests

lib/client.js
dnlup: I did skip the assignment, however the problem could have been the parser.
It wasn't calling kOnMessageComplete and kOnHeadersComplete. By changing the `skipBody` value to `1` or `true` it seems fixed.

See https://github.com/mcollina/undici/pull/9#discussion_r367902840

lib/client.js
dnlup: Could this be node version dependent?
https://github.com/creationix/http-parser-js/blob/8a81c92201e49fda4668386263fdce284789ed73/http-parser.js#L362

lib/client.js
```diff
@@ -217,10 +217,15 @@ class Client extends EventEmitter {
       // TODO move this[kCallbacks] from being an array. The array allocation
       // is showing up in the flamegraph.
       const cb = this[kCallbacks].shift()
+      const request = this[kRequests].shift()
+      let skipBody
```
dnlup: maybe this part would be better written like this
```js
const skipBody = request.method === 'HEAD'
if (!skipBody) {
// assign lastBody
}
// logic...
return skipBody
```

lib/client.js
```diff
@@ -239,7 +246,7 @@ class Client extends EventEmitter {
     this.parser[HTTPParser.kOnMessageComplete] = () => {
       const body = this._lastBody
       this._lastBody = null
-      body.push(null)
+      body && body.push(null)
```
mcollina: can you please make it

if (body !== null) ..

lib/client.js
mcollina: I don’t know. why do you ask?

lib/client.js
mcollina: yes!

lib/client.js
dnlup: Because of this comment in the snippet I posted
```js
      // For older versions of node (v6.x and older?), that return skipBody=1 or skipBody=true,
      //   need this "return true;" if it's an upgrade request.
```

test/client.js
mcollina: this should be resolved now?

test/client.js
mcollina: is this still true? CI seems to be passing

test/client.js
mcollina: this should be resolved now?

mcollina:
Is this still a wip or can it land?

dnlup:
WIP removed

PR #12, opened 2020-01-22
Pass callback directly to connect, reconnect

doesdev:
This is to ensure that initial socket connection errors are caught.
Alternatively, we could push the callback into the queue before calling
`connect`. However, that has potential side-effects. I went for the
solution with the least chance of unintended consequences.


doesdev:
Nevermind, there are some issues with this. That is, it does not resolve the issue save for the very first call. Will work on a better solution.

PR #13, opened 2020-05-02
fix: remove syncthrough

ronag:
Inlining is both simpler and faster.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -9,7 +9,6 @@ const Q = require('fastq')
 const { HTTPParser } = require('http-parser-js')
 const { Readable } = require('readable-stream')
 const eos = require('end-of-stream')
-const syncthrough = require('syncthrough')
 const retimer = require('retimer')
 const { EventEmitter } = require('events')
 const Request = require('./request')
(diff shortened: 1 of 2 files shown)


lib/client.js
```diff
@@ -183,11 +182,24 @@ class Client extends EventEmitter {
...
+          const onDrain = () => {
+            body.resume()
+          }
+
+          body.on('data', onData)
+          this.socket.on('drain', onDrain)
+
           this.socket.uncork()
           eos(body, (err) => {
+            body.destroy(err)
```
ronag, line 201: Should probably also destroy the body here.

mcollina: body is already finished here, so this is superfluous I think.
However, if the socket errors we should destroy the body. This is not done here.

ronag: makes sense, I'll remove it.

ronag:
We could do the same thing in the non chunked case and avoid `pipe()` with a possibly small memory & performance improvement, i.e. `pipe` does some not strictly necessary stuff.

mcollina:
Send a PR, I’m happy to experiment with this module.

PR #19, opened 2020-05-02
fix: error request body if socket fails

ronag:


ronag:
This pool error seems flaky, I get it sometimes locally as well, not always.

lib/client.js
```diff
@@ -164,18 +165,29 @@ class Client extends EventEmitter {
         }
         this.socket.write(body)
       } else if (body && typeof body.pipe === 'function') {
+        const cleanup = eos(this.socket, err => {
```
ronag, line 168: I'm a little unsure here what happens if the socket unexpectedly emits `'end'`. Does `Client` somewhere detect this and call `Client.destroy(err)`?

ronag: Created issue https://github.com/mcollina/undici/issues/21

mcollina:
could you fix the conflicts?

ronag:
fixed

PR #25, opened 2020-05-02
fix: let inflight complete and add optional callback to close

ronag:


ronag:
There is some more to be done in regards to teardown. I've added TODO's which will have follow up PRs.

One pending question though, shouldn't close also wait for queued requests, not just inflight?

mcollina:
I would say so, but it’s probably something worth an option somewhere.

PR #37, opened 2020-05-03
fix: close should gracefully wait for all queued requests

ronag:
Fixes: https://github.com/mcollina/undici/issues/20

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -354,38 +354,33 @@ class Client extends EventEmitter {
   }
 
   close (cb) {
-    if (this.closed) {
-      return
-    }
-    this.closed = true
-
-    // TODO: in-flight should still be able to timeout?
-    if (this[kTimer]) {
-      this[kTimer].clear()
-      this[kTimer] = null
-    }
-
-    if (cb) {
-      eos(this.socket, (err) => {
-        if (err && err.message !== 'premature close') {
-          cb(err, null)
-        } else {
-          cb(null, null)
-        }
+    if (cb === undefined) {
+      return new Promise((resolve, reject) => {
+        this.close((err, data) => {
+          return err ? reject(err) : resolve(data)
+        })
       })
     }
 
-    // TODO wait for queued requests as well, not just in-flight?
+    this.closed = true
 
-    // TODO test this
-    if (
-      this[kQueue].length() === 0 &&
-      this[kCallbacks].length === 0 &&
-      this.socket
-    ) {
-      this.socket.end()
-      // TODO wait util socket is closed before setting to null?
-      this.socket = null
+    destroyMaybe(this)
+
+    if (!cb) {
+      return
+    }
+
+    if (this.socket) {
+      eos(this.socket, (err) => {
+        if (err && err.message !== 'premature close') {
+          cb(err)
+        } else {
+          cb(null)
+        }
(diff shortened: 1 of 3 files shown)


lib/client.js
```diff
@@ -402,8 +386,6 @@ class Client extends EventEmitter {
     }
 
     if (this.socket) {
-      // TODO make sure we error everything that
-      // is in flight
```
ronag, line 406: This was fixed in one of the other PR's. Just removing the TODO.

lib/client.js
```diff
@@ -375,18 +370,7 @@ class Client extends EventEmitter {
...
-    // TODO test this
-    if (
-      this[kQueue].length() === 0 &&
-      this[kCallbacks].length === 0 &&
-      this.socket
-    ) {
-      this.socket.end()
-      // TODO wait util socket is closed before setting to null?
-      this.socket = null
-    }
```
ronag, line 389: Don't end the socket if _lastBody exists. Could use allowHalfOpen but there is not much advantage to that, just destroy the socket once we are done.

lib/client.js
```diff
@@ -359,13 +359,8 @@ class Client extends EventEmitter {
     }
     this.closed = true
 
-    // TODO: in-flight should still be able to timeout?
-    if (this[kTimer]) {
-      this[kTimer].clear()
-      this[kTimer] = null
-    }
```
ronag, line 366: close waits for requests so timeout is still relevant

ronag:
flaky test

test/client-errors.js
```diff
@@ -27,7 +27,7 @@ test('GET errors and reconnect with pipelining 1', (t) => {
     const client = new Client(`http://localhost:${server.address().port}`, {
       pipelining: 1
     })
-    t.tearDown(client.close.bind(client))
+    t.tearDown(client.destroy.bind(client))
```
ronag, line 30: This tests don't actually finish the requests which means if we do close we have to wait for a timeout... just do destroy instead

lib/client.js
```diff
@@ -354,38 +354,33 @@ class Client extends EventEmitter {
...
+        })
+      })
     }
+
     this.closed = true
 
-    // TODO: in-flight should still be able to timeout?
-    if (this[kTimer]) {
-      this[kTimer].clear()
-      this[kTimer] = null
```
ronag, line 365: close waits for requests so timeout is still relevant

lib/client.js
```diff
@@ -354,38 +354,33 @@ class Client extends EventEmitter {
...
   close (cb) {
-    if (this.closed) {
-      return
+    if (cb === undefined) {
+      return new Promise((resolve, reject) => {
+        this.close((err, data) => {
+          return err ? reject(err) : resolve(data)
+        })
+      })
     }
```
ronag, line 363: promisify

lib/client.js
```diff
@@ -354,38 +354,33 @@ class Client extends EventEmitter {
   }
 
   close (cb) {
-    if (this.closed) {
-      return
```
mcollina, line 358: I think those two lines might still be needed somewhere?

ronag: No, not required. If it's already closed it doesn't do anything.

PR #43, opened 2020-05-03
fix: throw on GET & HEAD request body

ronag:


lib/request.js
```diff
@@ -70,13 +70,14 @@ class Request extends AsyncResource {
     }
     this.path = opts.path
 
-    // TODO we should validate that the http method accepts a body or not
-    if (!isValidBody(opts.body)) {
+    if (opts.body && (this.method === 'GET' || this.method === 'HEAD')) {
```
ronag, line 73: I think HEAD and GET are the only methods where a body is not allowed?

ronag:
Close due to discussion in https://github.com/mcollina/undici/issues/35

PR #45, opened 2020-05-03
fix: closed client should be able to reconnect

ronag:


ronag:
added test

ronag:
I've put a few commits into this PR to avoid the whole endless conflicts thing. The commits should be somewhat related. Hope you don't mind.

test/client-errors.js
```diff
@@ -265,7 +265,7 @@ test('invalid URL throws', (t) => {
 })
 
 test('POST which fails should error response', (t) => {
-  t.plan(2)
+  t.plan(4)
```
mcollina, line 268: why this change?

test/client.js
```diff
@@ -586,3 +586,35 @@ test('close should call callback once finished', (t) => {
...
   })
 })
+
+test('close should still reconnect', (t) => {
+  t.plan(6)
+
+  const server = createServer((req, res) => {
+    setTimeout(function () {
+      res.end(req.url)
+    }, 10)
```
mcollina, line 596: please avoid the timeout, it would make the tests even more brittle

test/client-errors.js
ronag: It didn’t call all the calllbacks before

test/client.js
ronag: removed, was not needed, came with copy paste of another test

PR #48, opened 2020-05-03
fix: reschedule task while socket is destroyed

ronag:


lib/client.js
```diff
@@ -143,11 +143,19 @@ class Client extends EventEmitter {
...
+    const dispatch = (request, cb) => {
       if (this.destroyed) {
         return cb(new Error('The client is destroyed'))
       }
 
+      if (this.socket.destroyed) {
+        // TODO: This doesn't work
+        // this[kQueue].unshift(request, cb)
+        // this[kQueue].pause()
+        // cb()
```
ronag, line 155: @mcolline: I need help here... fastq is doing some magic here where it's not entirely clear how to reschedule a task.

ronag:
@mcollina: This isn't pretty but solves the problem... do you have any better ideas?

lib/client.js
```diff
@@ -143,11 +143,18 @@ class Client extends EventEmitter {
...
-    this[kQueue] = Q((request, cb) => {
+    const dispatch = (request, cb) => {
       if (this.destroyed) {
         return cb(new Error('The client is destroyed'))
       }
 
+      if (this.socket.destroyed) {
+        this[kQueue].unshift(request, cb)
+        this[kQueue].pause()
+        cb()
```
ronag, line 154: The problem here is that we just want to reschedule the task without invoking the outer callback...

ronag: Maybe something can be added to the fastq API in this regard?

ronag:
This is kind of an edge case that might not be worth fixing...

lib/client.js
mcollina: take a look and see if it’s feasible

ronag:
Won't be looking into this now, would be nice but not necessary

PR #51, opened 2020-05-03
fix: simplify reading

ronag:
I believe this is a simpler, faster and more correct way of doing this, maybe?

Fixes: https://github.com/mcollina/undici/issues/50

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -108,22 +108,18 @@ class Client extends EventEmitter {
     }
 
     this.url = url
-
-    this[kStream] = opts.stream || require('readable-stream')
-
-    // state machine, might need more states
     this.closed = false
     this.destroyed = false
+    this.timeout = opts.timeout || 30000 // 30 seconds
 
+    this[kStream] = opts.stream || require('readable-stream')
     this[kTLSOpts] = opts.tls || opts.https
+    this[kInflight] = []
 
     const endRequest = () => {
       this.socket.write('\r\n', 'ascii')
       this.socket.uncork()
     }
-
-    this.timeout = opts.timeout || 30000 // 30 seconds
-    this[kInflight] = []
 
     this[kQueue] = Q((request, cb) => {
       if (this.destroyed) {
--- a/lib/client.js
+++ b/lib/client.js
@@ -11,14 +11,10 @@ const { EventEmitter } = require('events')
 const Request = require('./request')
 const assert = require('assert')
 
-const kRead = Symbol('read')
-const kReadCb = Symbol('readCallback')
-const kIsWaiting = Symbol('isWaiting')
 const kQueue = Symbol('queue')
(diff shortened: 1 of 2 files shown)


ronag:
> My empiric testing showed that .read() is generally faster than pause()/resume().

I'm sceptical whether that difference is significant enough to warrant the complexity here. Not sure how to benchmark it though.

ronag:
I would like request we do this, at least to reduce the cognitive load until we've ironed out the rest of the issues/bugs.

ronag:
The next problem to solve here will be that the client should keep reading data from the connection if the body is `destroyed`. Which in this implementation would be:

```js
// only pause if we have a valid body
if (body && !body.destroyed && body.push(...)) {
  socket.pause()
}
```

Not sure how to resolve with the existing one.

mcollina:
this now conflicts.

ronag:
I encountered some timing bugs while playing around with this, will try to sort those out together with this.

ronag:
@mcollina: PTAL

test/client-errors.js
```diff
@@ -268,7 +268,7 @@ test('POST which fails should error response', (t) => {
   t.plan(4)
 
   const server = createServer()
-  server.once('request', (req, res) => {
+  server.on('request', (req, res) => {
```
ronag, line 271: This test shouldn't have passed before... some other bug that got resolved with this PR.

lib/client.js
```diff
@@ -394,7 +370,13 @@ function reset (client, err) {
     const skipBody = request.method === 'HEAD'
 
     if (!skipBody) {
-      client[kLastBody] = new client[kStream].Readable({ read: client[kRead] })
+      client[kLastBody] = new client[kStream].Readable({
+        read () {
+          if (client.socket) {
+            client.socket.resume()
+          }
+        }
```
ronag, line 378: Assuming turbofan can extract and optimize this.

ronag:
flaky test again

lib/client.js
```diff
@@ -204,9 +202,9 @@ class Client extends EventEmitter {
             // TODO we might want to wait before previous in-flight
             // requests are finished before destroying
             if (this.socket) {
-              destroySocket(this, err, cb)
+              destroySocket(this, err, callback)
             } else {
-              cb(err, null)
+              callback(err, null)
```
ronag, line 207: This was using a non-wrapped callback before

mcollina:
this conflicts now

ronag:
fixed

mcollina:
before lgtm this, can you please check if the benchmarks degrades? https://github.com/mcollina/undici/tree/master/benchmarks

ronag:
master:
benchmarks$ node undici.js 
requests: 21.902s

#51:
benchmarks$ node undici.js 
requests: 22.202s

PR #55, opened 2020-05-04
refactor: extract reset parser

ronag:
Not sure what you prefer, but resetParser doesn't need to live on the object and take up an unnecessary property.


lib/client.js
```diff
@@ -36,8 +35,6 @@ function connect (client) {
 
   client.socket = socket
 
-  client[kResetParser]()
```
ronag, line 39: This call was unnecssary.

PR #65, opened 2020-05-05
fix: handle parser error

ronag:
Fixes: https://github.com/mcollina/undici/issues/62


ronag:
I though `instanceof Error` would have significant performance implications but doesn't seem so based on the benchmark.

PR #67, opened 2020-05-05
fix: request body edge cases

ronag:
Readable can emit 'end' after .destroy() as well
as legacy streams might not implement the whole
Readable contract properly. Try to handle these
edge cases better.

Fixes: https://github.com/mcollina/undici/issues/66

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -18,6 +18,8 @@ const kLastBody = Symbol('lastBody')
 const kStream = Symbol('kStream')
 const kClosed = Symbol('kClosed')
 
+function nop () {}
+
 function connect (client) {
   var socket = null
   var url = client.url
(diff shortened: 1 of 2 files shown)


lib/client.js
```diff
@@ -181,39 +178,56 @@ class Client extends EventEmitter {
...
           }
 
-          assert(this.socket)
+          body
+            .removeListener('data', onData)
+            .removeListener('end', onEnd)
 
-          this.socket.removeListener('drain', onDrain)
+          if (typeof body.destroy === 'function') {
+            body.destroy(err)
```
ronag, line 207: I'm a little unsure whether we should forward the error to the request body here or not.

ronag:
valid errors on 10 & 12

lib/client.js
ronag: @szmarczak you got any thoughts? i.e.

```js
const myBody = ...
await client.request({ ...options, body: myBody })
```

If the request fails would you expect:

- `myBody` to be destroyed and emit the request error.
- `myBody` to be destroyed.
- Do nothing. Caller owns `myBody` life time.

szmarczak: Number 1. IMO it should be as close to the pipeline behavior as possible.

PR #68, opened 2020-05-05
fix: add simple reconnect backoff

ronag:
Fixes: https://github.com/mcollina/undici/issues/11

This is a rather simple backoff implementaiton. We could make it more advanced with more options. However, I think this should be good enough and adding more options will just confuse and make things more complicated.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -15,12 +15,14 @@ const kQueue = Symbol('queue')
 const kInflight = Symbol('inflight')
 const kTLSOpts = Symbol('TLS Options')
 const kLastBody = Symbol('lastBody')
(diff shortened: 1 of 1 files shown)


mcollina:
Could you add a test? would it be needed?

ronag:
> Could you add a test? would it be needed?

I was thinking about this. I wouldn't mind adding one. But it would make the tests run slower.

mcollina:
Should be ok anyway. You could also use sinonjs fake-timers to speed up time.

ronag:
> You could also use sinonjs fake-timers to speed up time.

You mind if I merge this as is and make a note in the Test todo issue?

PR #70, opened 2020-05-05
fix: wrap destroyed callback

ronag:


ronag:
I'm not quite sure I understand this async wrap stuff so right now I'm just trying to follow it logically. I'll see if I can try to understand it from the existing test.

mcollina:
You'll need to expand https://github.com/mcollina/undici/blob/master/test/async_hooks.js for the error case.

ronag:
@mcollina: added test

PR #72, opened 2020-05-06
perf: prefer connected clients in pool

ronag:
- Prefer connected clients in pool.
- Round robin connected clients to try and keep the maximum number of inflight request for a connected client as low as possible, i.e. saturating all active connections.

This will require some work on the tests. Would appreciate some initial feedback before I start sorting those out.

Fixes: https://github.com/mcollina/undici/issues/69
Fixes: https://github.com/mcollina/undici/issues/73

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -234,6 +234,10 @@ class Client extends EventEmitter {
     reset(this)
   }
 
+  get connected () {
+    return this.socket && !this.socket.destroyed
+  }
+
   get pipelining () {
     return this[kQueue].concurrency
   }
--- a/lib/pool.js
+++ b/lib/pool.js
@@ -1,7 +1,8 @@
 'use strict'
 
 const Client = require('./client')
-const current = Symbol('current')
+const kIndex = Symbol('index')
+const kDrained = Symbol('drained')
 
 class Pool {
   constructor (url, opts = {}) {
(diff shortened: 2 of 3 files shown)


ronag:
This has the advantage of skipping clients that disconnects due to errors or abortions.

ronag:
Question here, should we try and maximize the number of connections first, and then start pipelining, or vice versa?

mcollina:
I strongly disagree that round robin is better than LIFO :(.

LIFO ensure that

1. the minimum amount of sockets are used (these costs on both sides)
2. connections works at maximum pipelining

The result is fewer sockets with very large TCP windows and bandwidth, especially when client and server are “close”.

I think different use cases require different trade offs, so possibly we should make this pluggable.

ronag:
> 1. the minimum amount of sockets are used (these costs on both sides)

What I'm suggesting here is round robin of *connected* clients, i.e. the number of connections is the same as with LIFO.

> 2. connections works at maximum pipelining

Again, round robin of *connected* clients will ensure that a new connection is only created if all existing connections are at maximum pipelining.

ronag:
I think making it pluggable is reasonable.

ronag:
I think it is better to focus on sorting out client issues for now. Will revisit this sometime in the future.

PR #76, opened 2020-05-06
fix: pause while piping

ronag:
Fixes: https://github.com/mcollina/undici/issues/52

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -60,17 +60,20 @@ function _connect (client) {
   client[kQueue].pause()
   client[kLastBody] = null
 
-  let body = null
+  let _body = null
+  let _reset = null
 
   parser[HTTPParser.kOnHeaders] = () => {}
   parser[HTTPParser.kOnHeadersComplete] = ({ statusCode, headers }) => {
     // TODO move client[kInflight] from being an array. The array allocation
     // is showing up in the flamegraph.
-    const { request, callback } = client[kInflight].shift()
+    const { request, callback, reset } = client[kInflight].shift()
     const skipBody = request.method === 'HEAD'
 
+    _reset = reset
+
     if (!skipBody) {
-      body = client[kLastBody] = new client[kStream].Readable({
+      _body = client[kLastBody] = new client[kStream].Readable({
         autoDestroy: true,
         read () {
           socket.resume()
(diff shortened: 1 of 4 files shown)


ronag:
Missing tests

lib/client.js
```diff
@@ -309,6 +309,10 @@ class Client extends EventEmitter {
             if (chunked) {
               socket.cork()
               socket.write('\r\n0\r\n', 'ascii')
+              // Don't resume queue. Server should destroy connection once
+              // completed. Queue will automatically resume once reconnected.
```
ronag, line 313: This is the behavior I'm seeing. Though I'm unsure whether the standard enforces this?

ronag: The standard seems to enforce the opposite, i.e. non chunked MUST close connection.

https://www.w3.org/Protocols/rfc2616/rfc2616-sec8.html

> 8.2.2 Monitoring Connections for Error Status Messages
> An HTTP/1.1 (or later) client sending a message-body SHOULD monitor the network connection for an error status while it is transmitting the request. If the client sees an error status, it SHOULD immediately cease transmitting the body. If the body is being sent using a "chunked" encoding (section 3.6), a zero length chunk and empty trailer MAY be used to prematurely mark the end of the message. If the body was preceded by a Content-Length header, the client MUST close the connection.

ronag: @mcollina: Do you know anything about this or do you maybe know someone that might?

ronag:
This should not land until the strange behavior is either understood or a better implementation is found.

lib/client.js
ronag: Futher strange behavior, if the next request is sent immediate the server will close the connection quickly. However, if the next request is not sent, then the server will wait a few second before closing the connection.

mcollina: I don’t understand the question, can you please clarify?

ronag: So if I send a request with a chunked body. The server always ends the connection once the request has completed. Is this intendent behavior or a possible bug (server or client)?

mcollina: > If I send a request with a chunked body / no content-length. The server always ends the connection once the request has completed. Is this intendent behavior or a possible bug (server or client)?

It does not seem to covered by the spec. I would close the connection anyway.

ronag: Then this PR should ok, I should probably add so that the client explicitly closes the connection and doesn't assume the server is going to do it.

ronag:
Don't merge yet. Need to fix so that client closed connection and doesn't assume server does it.

ronag:
@mcollina: PTAL. I've made this very strict right now just to be sure. It will reset the connection for any request that contains a body. Please let me know if you think this is too strict.

PR #80, opened 2020-05-07
fix: pipelining corresponds to number of active requests

ronag:
Fixes: https://github.com/mcollina/undici/issues/77
Fixes: #81


ronag:
work in progress

PR #83, opened 2020-05-08
fix: requeue if socket is destroying

ronag:


test/client-errors.js
```diff
@@ -549,7 +549,9 @@ test('socket fail while ending request body', (t) => {
 
     const _err = new Error('kaboom')
     client.on('connect', () => {
-      client.socket.destroy(_err)
+      process.nextTick(() => {
```
mcollina, line 552: why this nextTick?

ronag: Because otherwise the socket will be destroyed before the queue has been resumed, which kind of invalidates the test.

ronag:
I've missed something. Test failures are valid.

test/client-errors.js
ronag: changed to resume before emit.

ronag:
valid failure on node < 13

ronag:
Found some timing bugs that got activated with this changed. Was different between different version of node as well. Fixed.

lib/client.js
```diff
@@ -440,24 +442,25 @@ class Client extends EventEmitter {
     this[kRetryTimeout] = null
 
     const onDestroyed = (err) => {
-      for (const callback of this[kDestroyed]) {
+      const callbacks = this[kDestroyed]
+      this[kDestroyed] = null
```
ronag, line 446: There was a re-entrat bug here

lib/client.js
```diff
@@ -440,24 +442,25 @@ class Client extends EventEmitter {
...
-          _err = err
-        })
         .on('close', () => {
-          onDestroyed(_err)
+          onDestroyed(this.socket[kError])
         })
         .destroy(err)
-    } else {
-      process.nextTick(onDestroyed, null, null)
     }
```
ronag, line 464: Was a problem here when `client.close` was invoked between `'error'` and `'close'` being emitted on the socket.

test/client-errors.js
```diff
@@ -519,7 +517,9 @@ test('socket fail while writing request body', (t) => {
     body.push('asd')
 
     client.on('connect', () => {
-      client.socket.destroy('kaboom')
+      process.nextTick(() => {
+        client.socket.destroy('kaboom')
+      })
```
ronag, line 522: We resume the queue *after* emitting connect, thus we need to defer the destroy to keep the test valid.

PR #90, opened 2020-05-08
fix: ignore request header mutations

ronag:


lib/request.js
```diff
@@ -63,22 +63,37 @@ class Request extends AsyncResource {
...
+    this.method = opts.method
+
+    this.path = opts.path
+
     this.body = opts.body
 
-    this.headers = opts.headers
+    this.host = opts.headers && (opts.headers.host || opts.headers.Host)
+
+    this.chunked = !opts.headers || opts.headers['content-length'] === undefined
```
ronag, line 83: linter complained about `hasOwnProperty`, I think `=== undefined` should be good enough.

PR #92, opened 2020-05-09
fix: response handling …

ronag:
- Don't emit 'end' if aborted.
    - Don't start the next request until aborted res
      is read or connection is reset.
    - Reset max payload counter after each request.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -64,8 +64,9 @@ function _connect (client) {
 
   client[kQueue].pause()
 
-  let body = null
-  let read = 0
+  let _body = null
+  let _read = 0
+  let _callback = null
 
   parser[HTTPParser.kOnHeaders] = () => {}
   parser[HTTPParser.kOnHeadersComplete] = ({ statusCode, headers }) => {
@@ -75,7 +76,7 @@ function _connect (client) {
     const skipBody = request.method === 'HEAD'
 
     if (!skipBody) {
-      body = new client[kStream].Readable({
+      _body = new client[kStream].Readable({
         autoDestroy: true,
         read () {
           socket.resume()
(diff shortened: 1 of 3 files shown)


lib/client.js
```diff
@@ -91,43 +92,52 @@ function _connect (client) {
...
 
   parser[HTTPParser.kOnMessageComplete] = () => {
-    if (body) {
-      body.push(null)
-      body = null
+    if (_body) {
+      // Stop Readable from emitting 'end' when destroyed.
+      if (!_body.destroyed) {
+        _body.push(null)
+      }
```
ronag, line 133: Without this we can get an `'end'` after having aborted the response which is surprising

lib/client.js
```diff
@@ -64,8 +64,9 @@ function _connect (client) {
 
   client[kQueue].pause()
 
-  let body = null
-  let read = 0
+  let _body = null
+  let _read = 0
+  let _callback = null
```
ronag, line 69: using _ to indicate state vars to avoid shadowing local vars

lib/client.js
```diff
@@ -91,43 +92,52 @@ function _connect (client) {
...
-    if (body) {
-      body.push(null)
-      body = null
+    if (_body) {
+      // Stop Readable from emitting 'end' when destroyed.
+      if (!_body.destroyed) {
+        _body.push(null)
+      }
+      _body = null
+      _read = 0
```
ronag, line 135: We must reset _read after response is done so that it doesn't accumulate between requests

lib/client.js
```diff
@@ -91,43 +92,52 @@ function _connect (client) {
             err = new Error('aborted')
           }
 
-          callback()
```
ronag, line 94: Don't allow more requests until whole response has been processed

mcollina: This was done to increase throughput and utilization. Essentially if we are starting to receive data, we can send more as most of the response has been computed on the server.

lib/client.js
mcollina: I'm not really a fan. I think we might look into rewriting/simplifying this whole method. Maybe we can use a state object somewhere and avoid keeping the context in the closures.

Anyway, this is ok for now.

lib/client.js
ronag: > This was done to increase throughput and utilization. Essentially if we are starting to receive data, we can send more as most of the response has been computed on the server.

That's still achievable by using a higher pipelining factor? This goes back to, https://github.com/mcollina/undici/issues/77.

lib/client.js
ronag: I'm happy to revisit this later.

lib/client.js
ronag: note, that this PR only changes the case where the response is aborted.

PR #95, opened 2020-05-09
fix: wait for inflight to complete before reset

ronag:


ronag:
Just a test for now... this one is quite tricky to sort in an elegant way using the current implementation. Might need a little rework in term of queue and infligth array.

ronag:
I'm not entirely happy with this solution in terms of it being a bit too strict. Not sure if relevant in practice. 

It does the job for now...

ronag:
valid failure on node 12 & 13

PR #97, opened 2020-05-10
refactor: avoid closure state

ronag:


ronag:
@mcollina: Another try, hope you like this better

PR #100, opened 2020-05-11
fix: broken assertion

ronag:
finished will be invoked on 'end' while destroy(err) will be delayed.
This can break the assertion in theory on edge cases.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -197,10 +197,10 @@ function _connect (client) {
   client[kQueue].pause()
 
   client[kStream].finished(socket, (err) => {
-    assert(err)
+    err = err || new Error('other side closed')
 
-    client._parser.destroy(err)
-    client._socket.destroy(err)
+    parser.destroy(err)
+    socket.destroy(err)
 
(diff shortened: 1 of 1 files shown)


ronag:
Haven't been able to trigger this in practice, but in theory it was a problem.

lib/client.js
```diff
@@ -187,20 +187,17 @@ function _connect (client) {
     .on('error', function (err) {
       this[kError] = err
     })
-    .on('end', function () {
-      this.destroy(new Error('other side closed'))
```
ronag, line 191: This won't cause `'error'` until nextTick, which would be after `finished(...)` which could invoke already at `'end'`.

lib/client.js
```diff
@@ -197,10 +197,10 @@ function _connect (client) {
   client[kQueue].pause()
 
   client[kStream].finished(socket, (err) => {
```
ronag, line 199: The actual behavior of how and when this is invoked, with or without error is a slightly dependent on node/readable-stream version. This is a safer way to implement.

PR #102, opened 2020-05-11
retry indirect failures

ronag:


lib/client.js
```diff
@@ -335,22 +338,7 @@ class Client extends EventEmitter {
...
-        // successful completion. Assume this always happens and
-        // explicitly reset the connection once response has been
-        // received. Queue will automatically resume once reconnected.
-
-        // TODO: This is too strict. Try to figure out when/how/why some
-        // servers do this.
-
-        reset = true
-        this[kQueue].pause()
-      }
```
ronag, line 351: This is actually no needed. The problem we were seeing was because of message interleaving since we weren't pausing the queue when piping request body.

ronag:
fails on travis but not locally for some reason, trying to revert one commit

PR #104, opened 2020-05-12
perf: optimize queue

ronag:
This replaces the fastq + inflight array with a single array + 2 indices.

Using the pattern moving things around is simply a matter of updating the indices to a buffer, with an amortized max limit on the buffer size.

undici benchmark

This PR: 2908.378ms
master: 6487.863ms

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -5,7 +5,6 @@
 const { URL } = require('url')
 const net = require('net')
 const tls = require('tls')
-const Q = require('fastq')
 const { HTTPParser } = require('http-parser-js')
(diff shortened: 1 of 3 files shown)


ronag:
commits and messages need cleanup

ronag:
Another optimization would be to re-use the requests objects in the queue.

ronag:
@mcollina: Is it possible to reuse `AsyncResource` objects? i.e. re-using the `Request` objects?

mcollina:
I would not reuse them.

PR #105, opened 2020-05-12
fix: don't resume while piping request

ronag:


ronag:
Fixed

test/client-pipelining.js
```diff
@@ -372,3 +372,54 @@ test('pipelining non-idempotent', (t) => {
...
 })
+
+test('pipelining non-idempotent w body', (t) => {
+  t.plan(4)
+
+  const server = createServer()
+  server.on('request', (req, res) => {
+    setTimeout(() => {
+      res.end('asd')
+    }, 10)
```
mcollina, line 383: there another `setTimeout` here

ronag: Sorry. Fixed.

PR #107, opened 2020-05-12
avoid stream.finished 

ronag:


ronag:
passes locally on node v10, fails on travis

PR #109, opened 2020-05-12
feat: request.consume

ronag:
By providing a `resume` property on the callback response and making `body` a lazy factory getter; it's possible to avoid the intermediate `Readable`. This would speed up e.g. creating a proxy using undici.

e.g.

Pre this PS
```js
function proxy (req, res) {
  client.request(fromReq(req), ({ statusCode, headers, body }) => {
    res.writeHead(statusCode, headers)
    // Body is an intermediate Readable
    body
      .on('error', (err) => res.destroy(err))
      .pipe(res)
  })
}
```

With this PR:
```js
function proxy (req, res) {
  client.request(fromReq(req), ({ statusCode, headers, resume }) => {
    res.writeHead(statusCode, headers)
    return pipe(res, resume)
  })
}

function pipe(dst, resume) {
  dst.on('drain', resume)
  return (err, chunk) => {
    if (err) {
      dst.destroy(err)
    } else if (chunk == null) {
      dst.end()
    } else {
      return dst.write(chunk)
    }
  }
}
```

Or just allow returning a `Writable` in the callback which does the above implicitly.

```js
function proxy (req, res) {
  client.request(fromReq(req), ({ statusCode, headers }) => {
    res.writeHead(statusCode, headers)
    return res
  })
}
```

Doesn't change anything in the existing API. Is purely opt-in.

Benchmarks:

body (pre PR): requests: 10346.729ms
resume (post PR): requests: 8902.416ms

### DIFF
--- a/README.md
+++ b/README.md
@@ -153,6 +153,51 @@ they fail due to indirect failure from the request
 at the head of the pipeline. This does not apply to
 idempotent requests with a stream request body.
 
+<a name='stream'></a>
+#### `client.stream(opts, cb(err, data))`
+
+A faster version of `'request'`. Not promisified.
+
+Unlike `request` this method expects `callback`
+to return a `Writable` which the response will be
+written to. This improves performance by avoiding
+creating an intermediate `Readable` when the user
+expects to directly pipe the response to a 
+`Writable`.
+
+```js
+const { Client } = require('undici')
+const client = new Client(`http://localhost:3000`)
+const fs = require('fs')
+const { finished } = require('stream')
+
+client.stream({
+  path: '/',
+  method: 'GET'
+}, function (err, data) {
+  if (err) {
+    console.error('failure', err)
+    return
+  }
+  const {
+    statusCode,
+    headers
+  } = data
+
+  console.log('response received', statusCode)
+  console.log('headers', headers)
+  const dst = fs.createWriteStream('destination.raw')
+  finished(dst, (err) => {
+    if (err) {
+      console.error('failure', err)
+    } else {
+      console.log('success')
+    }
+  })
+  return dst
+})
+````
+
 #### `client.pipelining`
 
 Property to get and set the pipelining factor.
--- a/lib/client.js
+++ b/lib/client.js
@@ -15,6 +15,7 @@ const {
   kQueue,
   kTimeout,
   kTLSOpts,
+  kWrite,
   kClosed,
   kDestroyed,
   kInflight,
@@ -37,20 +38,9 @@ class Parser extends HTTPParser {
 
     this.client = client
     this.socket = socket
+    this.resume = () => socket.resume()
     this.read = 0
     this.body = null
-    this.bodyRead = function () {
-      socket.resume()
-    }
-    this.bodyDestroy = function (err, cb) {
-      socket.resume()
-
-      if (!err && !this._readableState.endEmitted) {
-        err = new Error('aborted')
-      }
-
-      cb(err, null)
-    }
   }
 
   [HTTPParser.kOnHeaders] () {
@@ -62,24 +52,42 @@ class Parser extends HTTPParser {
     const request = client[kQueue][client[kComplete]]
     const skipBody = request.method === 'HEAD'
 
-    const body = new Readable({
-      autoDestroy: true,
-      read: this.bodyRead,
-      destroy: this.bodyDestroy
-    })
-    body.push = request.wrapSimple(body, body.push)
-
-    this.body = body
     this.read = 0
-
-    request.callback(null, {
+    const body = request.callback(null, {
       statusCode,
-      headers: parseHeaders(headers),
-      body
+      headers: parseHeaders(headers)
     })
 
-    if (skipBody) {
-      this[HTTPParser.kOnMessageComplete]()
+    if (body && typeof body.write === 'function') {
+      if (skipBody) {
+        body.end()
+      } else {
+        body
+          .on('drain', this.resume)
+          // TODO: 'error' and 'close' should only resume if waiting for 'drain'.
+          .on('error', this.resume)
+          .on('close', this.resume)
+        body[kWrite] = request.wrapSimple(body, function (chunk) {
+          if (chunk == null) {
+            this.end()
+          } else {
+            this.write(chunk)
+          }
+        })
+        this.body = body
+      }
+    } else if (body && typeof body.read === 'function') {
+      if (skipBody) {
+        body.push(null)
+      } else {
+        body._read = this.resume
+        body[kWrite] = request.wrapSimple(body, body.push)
+        this.body = body
+      }
+    }
+
+    if (!this.body) {
+      this.next()
     }
 
     return skipBody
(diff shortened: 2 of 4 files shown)


ronag:
@szmarczak

szmarczak:
Can't we just pass the server response as an e.g. `output` option? Then make the `body` return an empty readable.

We can have the `consume` option in case people want to implement advanced logic.

ronag:
> Can't we just pass the server response as an e.g. output option? Then make the body return an empty readable.

That would work. We could implement that on top of `consume` in the `Request` constructor, i.e. another `defaultConsumeWithWritable`.

ronag:
@szmarczak Updated API to be seamless with current API.

szmarczak:
I would just do

```js
function proxy (req, res) {
  client.request(fromReq(req), { output: res });
  // The `res.writeHead` would be handled automatically.
  // Alternatively the promise would be an event emitter,
  // so people could do `undiciRequest.on('headers', ...)`.
  //
  // All errors would be handled automatically, and in such case,
  // `output.destroy(error)` would be called.
  //
  // When the `output` option is present,
  // the `body` resulting from the promise would be either an empty readable,
  // or `undefined` / `null`.
}
```

ronag:
> A slightly better API would be to support to return a Writable/Transform somewhere so we can avoid allocating a Readable.

This would probably cover 90% of the cases, yes.

So something like:

```js
function proxy (req, res) {
  client.request(fromReq(req), ({ statusCode, headers }) => {
    res.writeHead(statusCode, headers)
    return res
  })
}
```

ronag:
> I would just do

That feels a little too high level to me.

ronag:
btw, this is not that important. The reason I bring it up is that there is a minor performance improvement opportunity here. Unfortunately it might affect the public API so we should probably consider this before cutting a release.

ronag:
```js
import { pipeline, PassThrough } from "stream"

// Callback
function proxy (req, res, cb) {
  client.request(fromReq(req), (err, val) => {
    if (err) {
      res.destroy(err)
    } else {
      res.writeHead(val.statusCode, val.headers)
      stream.finished(res, cb)
      return res
    }
  })
}

// Promise
async function proxy (req, res) {
  await new Promise((resolve, reject) => {
    client.request(fromReq(req), (err, val) => {
      if (err) {
        res.destroy(err)
      } else {
        res.writeHead(val.statusCode, val.headers)
        stream.finished(res, err => err ? reject(err) : resolve())
        return res
      }
    })
  })
}

// Pipeline
async function proxy(req, res) {
  await pipeline(
    req,
    async function * (source) {
      const pt = new PassThrough()
  
      client.request(fromReq(req, Readable.from(source)), (err, val) => {
        if (err) {
          pt.destroy(err)
        } else if (val.statusCode !== 200) {
          pt.destroy(new Error('invalid response'))
        } else {
          return pt
        }
      })
  
      for await (const chunk of pt) {
        yield chunk
      }
    },
    dst
  )
}
```

ronag:
@mcollina @szmarczak PTAL. Maintains current API but adds a other method `request2` (better name to be decided).

szmarczak:
I'm sticking to the `output` option proposal. The options doesn't have to include header proxying. It can be done via e.g. another option `onHeaders`.

ronag:
> I'm sticking to the output option proposal. The options doesn't have to include header proxying. It can be done via e.g. another option onHeaders.

Sorry, but I personally don't find it very elegant...

```js
await client.request({ ...options, output, onHeaders })
```

Though I can't argue `request2` is much better... especially if we can't find a better name.

@mcollina can cast the deciding vote if we want to include this 😄

lib/client.js
```diff
@@ -62,21 +51,18 @@ class Parser extends HTTPParser {
...
-      body
+      headers: parseHeaders(headers)
     })
+    // TODO: Should only resume if body.write paused.
+    this.body
+      .on('drain', this.resume)
+      .on('error', this.resume)
+      .on('close', this.resume)
+    this.body.write = request.wrapSimple(this.body, this.body.write)
+    this.body.end = request.wrapSimple(this.body, this.body.end)
```
ronag, line 65: I'm not exactly sure of the purpose of wrapSimple... should `destroy` also be wrapped?

mcollina: I'm really -1 on this wrapping. Why is this needed? This looks less maintainable.

README.md
```diff
@@ -153,6 +153,52 @@ they fail due to indirect failure from the request
...
+const fs = require('fs')
+
+const dst = fs
+  .createWriteStream('destination.raw')
+  .on('finish', function () {
+    console.log('file success')
+  })  
+  .on('error', function (err) {
+    console.log('file error', err)
+  })
```
mcollina, line 180: Can you move this inside the callback?

README.md
```diff
@@ -153,6 +153,52 @@ they fail due to indirect failure from the request
...
+  }
+  const {
+    statusCode,
+    headers
+  } = data
+
+  console.log('response received', statusCode)
+  console.log('headers', headers)
+
+  return dst
```
mcollina, line 198: Most developers would not know how to check when the file has been fully stored on disk. Can you expand this example using `finished`?

test/client-stream.js
```diff
@@ -0,0 +1,42 @@
...
+      const bufs = []
+      pt.on('data', (buf) => {
+        bufs.push(buf)
+      })
+      pt.on('end', () => {
+        t.strictEqual('hello', Buffer.concat(bufs).toString('utf8'))
+      })
+      return pt
+    })
+  })
```
mcollina, line 41: Can you please add a few tests for the errors?

lib/client.js
ronag: You mean the `wrapSimple`? I don't know. Just following the old code:

```js
body.push = request.wrapSimple(body, body.push)
```

I'm unsure what exactly this does.

README.md
ronag: Updated. PTAL.

lib/client.js
szmarczak: https://github.com/mcollina/undici/blob/269811b01db873f49c4d8172e95e8e82701d1f1d/lib/request.js#L118-L130

README.md
```diff
@@ -153,6 +153,51 @@ they fail due to indirect failure from the request
...
+written to. This improves performance by avoiding
+creating an intermediate `Readable` when the user
+expects to directly pipe the response to a 
+`Writable`.
+
+```js
+const { Client } = require('undici')
+const client = new Client(`http://localhost:3000`)
+const fs = require('fs')
+const stream = require('stream')
```
mcollina, line 172:
```suggestion
const { finished } = require('stream')
```

lib/client.js
mcollina: Ok, maybe it's ok :).

lib/client.js
```diff
@@ -7,7 +7,7 @@ const { HTTPParser } = require('http-parser-js')
 const { EventEmitter } = require('events')
 const Request = require('./request')
 const assert = require('assert')
-const { Readable } = require('stream')
+const { PassThrough } = require('stream')
```
mcollina, line 10: I think this will slow down the normal case, can you check?

lib/client.js
```diff
@@ -62,24 +51,29 @@ class Parser extends HTTPParser {
...
+        this.body.end()
+        this.body = null
+      } else {
+        this.body
+          .on('drain', this.resume)
+          // TODO: 'error' and 'close' should only resume if waiting for 'drain'.
+          .on('error', this.resume)
+          .on('close', this.resume)
+        this.body.write = request.wrapSimple(this.body, this.body.write)
+        this.body.end = request.wrapSimple(this.body, this.body.end)
```
ronag, line 71: Both wraps are needed otherwise async_hooks tests fail

mcollina: perfect, let's go with the wraps.

lib/client.js
szmarczak: It does. About 500 req/s less.

mcollina: I think we should have a slightly complicated approach that keeps using the readable with the previous API.

ronag: I will fix this. Shouldn’t be too bad.

ronag: PTAL. I've changed it so that you can return either a `Writable` or a `Readable`.

test/client-stream.js
ronag: added more tests

PR #111, opened 2020-05-13
fix: pool should be depth first connected

ronag:
This does differ a bit from the previous implementation as it does not take `'drain'` into account. I will probably have to change it but wanted to see whether this makes sense.

Basically this will always priotize the first connection in the list assuming that's the one least likely to timeout and the one that is fully saturated.

I hope this will also sort out the flakiness.

Also prioritizes clients which are connected.

### DIFF
--- a/lib/pool.js
+++ b/lib/pool.js
@@ -1,7 +1,6 @@
 'use strict'
 
 const Client = require('./client')
-const current = Symbol('current')
 
 class Pool {
   constructor (url, opts = {}) {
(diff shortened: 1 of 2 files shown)


mcollina:
It seems travis is being difficult, maybe we should just move everything to github actions.

mcollina:
Can you possibly benchmark this? If it does not decrease, I'm good to go.

ronag:
PRE: 11.892s
POST: 10173.528ms

mcollina:
That's unfortunate. I would prefer to have it retain the perf :/.

ronag:
> That's unfortunate. I would prefer to have it retain the perf :/.

It's faster? 11.89s vs 10.17s

mcollina:
Ahhh sorry, I thought those where req/s :D.

PR #113, opened 2020-05-13
fix: pool should avoid opening connections

ronag:
Noticed while testing that Pool would open lots of connections in the edge case where no client is connected.

If there is no connected client, pick the first one that is not full, instead of a random one.


ronag:
Needs a regression test.

PR #118, opened 2020-05-13
fix: don't swallow error

ronag:


ronag:
Hm... `errorMonitor` is not backported to v12 & v10. Any ideas to resolve this in a nice way? Add a version check on the test? I would like to avoid the usual hack for this where you remove the handler and re-emit...

PR #119, opened 2020-05-13
refactor: improve client.stream API

ronag:
This improves the `client.stream` API at the cost of a slightly (negligible?) slower request dispatch.

```js
await client.stream({
  path: '/',
  method: 'GET'
}, ({ statusCode, headers }) => {
  return fs.createWriteStream('destination.raw')
})
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,7 +45,7 @@ Options:
   [RFC7230](https://tools.ietf.org/html/rfc7230#section-6.3.2). Default: `1`.
 
 <a name='request'></a>
-#### `client.request(opts, cb(err, data))`
+#### `client.request(opts, callback(err, data))`
 
 Performs an HTTP request.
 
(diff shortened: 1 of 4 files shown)


ronag:
@szmarczak

mcollina:
What’s the impact on the normal usage?

ronag:
PRE: 2.679s
POST: 2.683s

ronag:
Further thoughts. It might make sense to return the stream, i.e.

```js
await pipeline(
	await client.stream({
	  path: '/',
	  method: 'GET',
      body: src
	}, ({ statusCode, headers }) => {
	  return new PassThrough()
	}).
    dst
)
```

Might alos make sense to support something like:

```js
await pipeline(
    src,
    client.pipeline({
	  path: '/',
	  method: 'GET'
	}, ({ statusCode, headers }) => {
	  return fs.createWriteStream('destination.raw')
	}).
    dst
)
```

I believe I can implement the second option with close to no overhead. Should I maybe try that first?

PR #120, opened 2020-05-14
fix: request.stream edge cases …

ronag:
Fixes some edge case TODOs

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -57,36 +57,22 @@ class Parser extends HTTPParser {
     let body = request.callback(null, {
       statusCode,
       headers: parseHeaders(headers)
-    })
+    }, this.resume)
     request.callback = null
 
     if (body) {
       if (typeof body.write === 'function') {
-        // TODO: 'error' and 'close' should only resume if waiting for 'drain'.
-        // TODO: Don't assume 'error' and 'close' will set destroyed to ´true`.
-        body
-          .once('drain', this.resume)
-          .once('close', this.resume)
-        if (EventEmitter.errorMonitor) {
-          body.once(EventEmitter.errorMonitor, this.resume)
-        } else {
-          body.once('error', (err) => {
-            this.resume()
-            if (body.listenerCount('error') === 0) {
-              body.emit('error', err)
-            }
-          })
-        }
         body[kWrite] = request.wrapSimple(body, function (chunk) {
           if (chunk == null) {
             this.end()
           } else {
-            this.write(chunk)
+            return this.write(chunk)
           }
         })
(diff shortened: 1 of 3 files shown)


test/client-stream.js
```diff
@@ -184,29 +184,3 @@ test('stream GET remote destroy', (t) => {
...
-    t.expectUncaughtException()
-    client.stream({
-      path: '/',
-      method: 'GET'
-    }, (err, { statusCode, headers }) => {
-      t.error(err)
-      return new PassThrough()
-    })
-  })
-})
```
ronag, line 212: This test should have been removed with the API change.

ronag:
Node v10 failure... Node 10 is end of life. Could we remove it?

mcollina:
This probably needs tests

lib/client.js
```diff
@@ -62,21 +62,9 @@ class Parser extends HTTPParser {
...
-        if (EventEmitter.errorMonitor) {
-          body.once(EventEmitter.errorMonitor, this.resume)
-        } else {
-          body.once('error', (err) => {
-            this.resume()
-            if (body.listenerCount('error') === 0) {
-              body.emit('error', err)
-            }
-          })
-        }
```
ronag, line 79: The `error` stuff no longer do anything so there is nothing to test.

lib/request.js
```diff
@@ -138,20 +138,17 @@ function dispatch (err, data) {
...
-        const callback = this.callback
-        this.callback = null
-        callback(null, null)
       }
     } catch (err) {
-      if (!this.callback) {
-        throw err
-      } else {
-        this.callback(err, null)
-      }
```
ronag, line 153: The `this.callback` stuff just moves things around a bit, nothing to test.

lib/client.js
```diff
@@ -62,21 +62,9 @@ class Parser extends HTTPParser {
...
+          .once('error', this.resume)
```
ronag, line 67: We will always either emit `'error'` unless gracefully ended. That's the only case we need resume.

lib/request.js
```diff
@@ -138,20 +138,17 @@ function dispatch (err, data) {
     try {
       const stream = this.factory(data)
       if (stream) {
-        finished(stream, this.callback)
+        finished(stream, (err) => {
+          stream.destroy(err)
```
ronag, line 142: This is a bit of an edge case I'm unsure how to trigger exactly in a test... will need to give it more thought.

ronag:
Found some other issues while trying to figure out some tests. Wip

ronag:
@mcollina PTAL, added test and found another issue which is resolved

PR #121, opened 2020-05-14
fix: add node 14 to travis

ronag:
Node 10 is EOL. Node 14 is released.


ronag:
> Node 10 is not EOL.

Sorry, I misread. I'll change to add 14 instead.

PR #122, opened 2020-05-15
pool stream + forward client opts

ronag:

PR #123, opened 2020-05-15
Update benchmarks

szmarczak:
#122 needs to be merged first


szmarczak:
@ronag Should I update the other benchmarks too?

ronag:
Yea please

ronag:
I think you should also write the response to a pass through. So we can show the stream API advantage.

benchmarks/index.js
```diff
@@ -0,0 +1,70 @@
...
+const http = require('http')
+const Benchmark = require('benchmark')
+const undici = require('..')
+
+const httpOptions = {
+  protocol: 'http:',
+  hostname: 'localhost',
+  method: 'GET',
+  path: '/',
+  port: 8080,
```
szmarczak, line 12:
```suggestion
  port: 3000,
```

benchmarks/index.js
```diff
@@ -0,0 +1,70 @@
...
+  fn: deferred => {
+    pool.request(undiciOptions, (error, { body }) => {
+      if (error) {
+        throw error
+      }
+
+      body.resume()
+      body.once('end', () => {
+        deferred.resolve()
+      })
```
ronag, line 49: Can you pipe this to a pass through? Same as stream version

benchmarks/index.js
```diff
@@ -0,0 +1,70 @@
...
+const suite = new Benchmark.Suite()
+
+suite.add('http - keepalive', {
+  defer: true,
+  fn: deferred => {
+    http.get(httpOptions, response => {
+      response.resume()
+      response.once('end', () => {
+        deferred.resolve()
+      })
```
ronag, line 35: Passthrough

benchmarks/index.js
szmarczak: You sure? If we do `pool.stream` we must provide a writable stream, right? But if you use the Node.js API then it automatically produces a stream. I mean to avoid unnecessary overhead.

szmarczak:
Also I wanted to note that if you use [h2o](https://github.com/h2o/h2o) server instead, the results are much higher (+2k req/s difference for me).

benchmarks/index.js
ronag: Otherwise the comparison with stream is unfair.

szmarczak: But `undici` directly calls `body.write` (on the user provided stream) whereas Node.js has to generate the body by itself...

ronag: So the point of `request.stream` is that `request.stream` can directly write to the final destination, while `stream.request` has to `pipe` it. However, if you don't `pipe` to something in the `stream.request` case then this advantage is not shown in the benchmark.

szmarczak: I see your point now. I will update the tests.

ronag:
If you could add a section in the README with your final results that would be great.

szmarczak:
Sure, will do, but I'd rather wait until #122 is merged :)

ronag:
@szmarczak #122 has landed 🎆

README.md
```diff
@@ -12,12 +12,28 @@ It is also a Stranger Things reference.
...
+
+```
+http - keepalive - read x 17,074 ops/sec ±6.99% (79 runs sampled)
+undici - request - read x 25,428 ops/sec ±1.73% (83 runs sampled)
+
+http - keepalive - pipe x 15,504 ops/sec ±7.32% (79 runs sampled)
+undici - stream  - pipe x 24,073 ops/sec ±1.86% (88 runs sampled)
+```
+
+This is a basic `hello world` [example](benchmarks/index.js).
```
szmarczak, line 35:
```suggestion
The benchmark is a basic `hello world` [example](benchmarks/index.js).
```

szmarczak:
`undici - stream - pipe` is almost 10k req/s faster than the Node.js native :fire:

ronag:
I think this needs a rebase?

README.md
```diff
@@ -12,12 +12,28 @@ It is also a Stranger Things reference.
 Picture of Eleven
 -->
 
-## Install
+### Install
```
ronag, line 15: Why change title level?

README.md
```diff
@@ -12,12 +12,28 @@ It is also a Stranger Things reference.
...
+Server: [H2O](https://github.com/h2o/h2o) 2.2.5\
+Configuration: HTTP/1.1 without TLS
+
+```
+http - keepalive - read x 17,074 ops/sec ±6.99% (79 runs sampled)
+undici - request - read x 25,428 ops/sec ±1.73% (83 runs sampled)
+
+http - keepalive - pipe x 15,504 ops/sec ±7.32% (79 runs sampled)
+undici - stream  - pipe x 24,073 ops/sec ±1.86% (88 runs sampled)
+```
```
ronag, line 33: Why 4 and not 3? It's a bit confusing to have http twice but with different results. :/

ronag: Looks like stream is slower than request

README.md
szmarczak: The same is done in `fastify`. But I'll revert that. Should benchmarks stay here or should I move them down?

README.md
ronag: Ah, `pipe` vs `read`.

ronag: I would just do the pipe version for all 3

szmarczak: It's different usage. `pipe` does `stream.pipe(anotherStream)`, while `read` does `stream.resume(); stream.once('end', () => deffered.resolve())`.

szmarczak: Ah, for some reason I had to refresh to get your comments. Will take another look at this.

szmarczak: > Looks like stream is slower than request

That's just an error. `±1.73%` vs `±1.86%`. It's basically the same.

ronag: But is read relevant in the real world? Seems a bit synthetic. I’m fine with it. Just thinking out loud.

README.md
ronag: I’ll leave it at your discretion.

README.md
szmarczak: Hmm... Honestly I'd prefer all four, but I don't mind having only three. @mcollina Any thoughts?

szmarczak: > But is read relevant in the real world? Seems a bit synthetic.

Nah, you're right. It **is** synthetic. Let me fix that.

szmarczak:
Welp, the README wasn't added somehow...

szmarczak:
@ronag PTAL :)

README.md
```diff
@@ -18,6 +18,20 @@ Picture of Eleven
...
 
+## Benchmarks
+
+Machine: Intel i7-7700k, 4.5GHz, 4C/8T, 32GB RAM, NVMe SSD
+Server: [H2O](https://github.com/h2o/h2o) 2.2.5
+Configuration: HTTP/1.1 without TLS
+
+```
+undici - request - pipe x 23,060 ops/sec ±1.11% (88 runs sampled)
+undici - stream  - pipe x 24,409 ops/sec ±1.70% (85 runs sampled)
```
ronag, line 29: Put the fastest on top and then we are good! Thanks for the help!

szmarczak: Done! :D

PR #134, opened 2020-05-16
Some more cleanup

ronag:
Please note `fix: remove https option`. I see future problems with having two names for the same option when useland modules start forwarding options upstream to undici and all of a sudden they have overlap of tls and https and unexpectedly lose options. Removes this ambiguity by just having one of the options.


mcollina:
(lgtm)

PR #138, opened 2020-05-16
refactor: remove EventEmitter

ronag:
Reduce API surface by removing event emitter and events. Only used for tests and there are otherwise to test the corresponding funcitonality.


ronag:
I agree. I think it’s probably best to leave as is.

PR #141, opened 2020-05-16
feat: add opaque to request.stream

ronag:
Fixes: https://github.com/mcollina/undici/issues/140

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,10 +181,17 @@ creating an intermediate `Readable` when the user
 expects to directly pipe the response body to a
 `Writable`.
 
+Options:
+
+* ... same as `request`.
+* `opaque`, passed as `opaque` to `factory`. Used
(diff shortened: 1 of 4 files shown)


mcollina:
the docs is not clear on how to use this

ronag:
> the docs is not clear on how to use this

I’m not sure how to make it more clear. Suggestion?

mcollina:
what closure does this avoid?

ronag:
Consider:

```js
function (req, res) {
   client.stream(opts, (data) => {
     proxy({ ...data, opaque: res })
   }
}
```

vs


```js
function (req, res) {
   client.stream({ ...opts, opaque: res }, proxy)
}
```

mcollina:
Please add the example, so it’s clearer in the docs

ronag:
Updated docs

PR #147, opened 2020-05-17
feat: stream.pipeline

ronag:
Fixes: https://github.com/mcollina/undici/issues/106

```js
stream.pipeline(
  fs.createReadStream('source.raw'),
  client.pipeline({
    path: '/',
    method: 'PUT',
  }, ({ statusCode }) => {
    if (statusCode !== 201) {
      throw new Error('invalid response')
    }
  }),
  fs.createReadStream('response.raw'),
  (err) => {
    if (err) {
      console.error('failed')
    } else {
      console.log('succeeded')
    }
  }
)
```

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -7,7 +7,12 @@ const { HTTPParser } = require('http-parser-js')
 const EventEmitter = require('events')
 const Request = require('./request')
 const assert = require('assert')
-const { Readable, finished } = require('stream')
+const {
+  Readable,
+  Duplex,
+  PassThrough,
+  finished
+} = require('stream')
 
 const {
   kUrl,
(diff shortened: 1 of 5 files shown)


ronag:
This could use some optimization. But I'm putting in place for now so that people use it instead of rolling their own. That way once we optimize this everyone will reap the benefits.

ronag:
@szmarczak

ronag:
timing bug somewhere, fails randomly

lib/client.js
```diff
@@ -327,6 +334,65 @@ class Client extends EventEmitter {
...
+    const req = new PassThrough({ autoDestroy: true })
+    const res = new PassThrough({ autoDestroy: true })
+    const ret = new Duplex({
+      autoDestroy: true,
+      read () {
+        res.resume()
+
+        // TODO: Workaround for possible Node bug?
+        // Doesn't always emit 'end' if resumed after
+        // ended.
```
ronag, line 351: @mcollina: FYI. I haven't had time to dig into this yet but it seems like a Node core bug affecting all versions of Node I've tested (12-14).

ronag: I can reproduce on Node master.

ronag: Found it. Not a core bug.

ronag:
This needs https://github.com/mcollina/undici/pull/150

lib/client.js
ronag: https://github.com/mcollina/undici/pull/150

PR #155, opened 2020-05-18
pipeline Duplex:like

ronag:
This is an experiment. By returning a `Duplex` like object we can slightly increase performance of `client.pipeline`. However, at the cost of full `streams` compatibility (which probably can be fixed). Basically the returned object should only be used with `stream.pipeline`.

Basically this avoids 2 levels of buffering:

- `req` becomes a `Readable` instead of `PassThrough`.
- `ret` becomes a `Readable` instead of `Duplex`.

Based on https://github.com/mcollina/undici/pull/153

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,10 +24,10 @@ Machine: 2.7 GHz Quad-Core Intel Core i7
 Configuration: Node v14.2, HTTP/1.1 without TLS, 100 connections
 
 ```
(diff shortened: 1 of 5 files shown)


ronag:
Wait for https://github.com/mcollina/undici/pull/153 to be merged before reviewing the actual code.

ronag:
Do this later.

PR #162, opened 2020-05-19
Introduce custom errors classes

delvedor:
As titled.

Related: https://github.com/mcollina/undici/issues/160

### DIFF
--- a/README.md
+++ b/README.md
@@ -405,9 +405,9 @@ const { errors } = require('undici')
 
 | Error                  | Error Codes              | Description                                            |
 | ---------------------- |--------------------------|--------------------------------------------------------|
-| `ConfigurationError`   |  `UND_ERR_CONFIGURATION` | Generated when in case of a bad configuration.         |
+| `InvalidArgumentError` |  `UND_ERR_INVALID_ARG`   | Generated when in case of a bad configuration.         |
 | `TimeoutError`         |  `UND_ERR_TIMEOUT`       | Generated when a request exceeds the `timeout` option. |
-| `RequestAbortedError`  |  `UND_ERR_ABORTED`       | Generated in case of an aborted request.               |
+| `RequestAbortedError`  |  `UND_ERR_ABORTED`       | Generated if the request has been aborted by the user  |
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -393,6 +393,25 @@ Calls [`client.close(callback)`](#close) on all the clients.
...
+Undici exposes a variety of error objects that you can use to enhance your error handling.
+You can find all the error objects inside the `errors` key.
+
+```js
+const { errors } = require('undici')
+```
+
+| Error                  | Error Codes              | Description                                            |
+| ---------------------- |--------------------------|--------------------------------------------------------|
+| `ConfigurationError`   |  `UND_ERR_CONFIGURATION` | Generated when in case of a bad configuration.         |
```
ronag, line 408: Would prefer invalid argument error?

lib/errors.js
```diff
@@ -0,0 +1,79 @@
...
+    super(message)
+    this.name = 'UndiciError'
+    this.code = 'UND_ERR'
+  }
+}
+
+class TimeoutError extends UndiciError {
+  constructor (message) {
+    super(message)
+    Error.captureStackTrace(this, TimeoutError)
```
ronag, line 14: I’m sure there is a reason for this. I just don’t know it. 😅

README.md
```diff
@@ -393,6 +393,25 @@ Calls [`client.close(callback)`](#close) on all the clients.
...
+| `TimeoutError`         |  `UND_ERR_TIMEOUT`       | Generated when a request exceeds the `timeout` option. |
+| `RequestAbortedError`  |  `UND_ERR_ABORTED`       | Generated in case of an aborted request.               |
```
ronag, line 410: I guess this should only be if the user aborted the event?

Server aborts I guess falls under socket error?

lib/errors.js
delvedor: https://nodejs.org/api/errors.html#errors_error_capturestacktrace_targetobject_constructoropt :)

ronag:
Landed in f679a0655d7e2e2153e0e556b738faf9f02f3850

mcollina:
I likely prefer to use the "squash" button on Github vs the node core way. It's way simpler to maintain.

PR #166, opened 2020-05-19
feat: abort signal

ronag:


ronag:
@delvedor if you can pick up from here and add docs + tests that would be great and I will continue with the timeout stuff.

ronag:
Right now it only works with `EventEmitter` might want to add support for https://www.npmjs.com/package/abort-controller as well.

delvedor:
@ronag given that the project is small and we are just three persons, I think we can all create branches here instead of forking, which causes a lot of overhead if you want to collaborate on the same code :D
What do you think?

ronag:
> What do you think?

Let's do it. Feel free to create a branch from this PR and close it.

PR #169, opened 2020-05-20
Abort signal

delvedor:
I've added some initial testing, the first one never ends, while the fourth fails.

Closes: https://github.com/mcollina/undici/issues/164


ronag:
> the first one never ends, while the fourth fails.

I will take a look. Can you also add docs?

We currently don't support https://www.npmjs.com/package/abort-controller. Do we want to?

delvedor:
> I will take a look. Can you also add docs?

Yes, I finish working on the tests :)

> We currently don't support https://www.npmjs.com/package/abort-controller. Do we want to?

My plan is to support it and I already have some tests for it, but I would like to fix these ones before continuing.

test/abort.js
```diff
@@ -0,0 +1,106 @@
...
+  })
+  t.teardown(server.close.bind(server))
+
+  server.listen(0, () => {
+    const client = new Client(`http://localhost:${server.address().port}`)
+    const ee = new EventEmitter()
+    t.teardown(client.close.bind(client))
+
+    client.request({ path: '/', method: 'GET', signal: ee }, (err, response) => {
+      t.ok(err instanceof errors.RequestAbortedError)
```
ronag, line 99: I think you want the error on the `response.body` here. It won't fail here since it gets aborted after receiving headers.

test/abort.js
```diff
@@ -0,0 +1,106 @@
...
+    })
+
+    setTimeout(() => {
+      ee.emit('abort')
+    }, 500)
```
ronag, line 104: Try to avoid timeouts if/when possible?

test/abort.js
delvedor: In the test above it already got the test (as far as I know `res.writeHead` sends directly the headers.
Here is not failing (I guess) because it already started receiving the body?

test/abort.js
delvedor: I agree that it should be desirable, but I don't know how could I call the abort method in specific moments without it.
Do we have some events I could use that I'm not aware of?

test/abort.js
delvedor: I've updated the test, you are right, in this case the error will appear on the body error event :)

test/abort.js
ronag: abort inside the request callback?

delvedor: That would work only for the test where we already start receiving the body, isn't it?

delvedor:
@ronag @mcollina I've added the support for `abort-controller`, but the tests are failing, what am I missing? 🤔

ronag:
I guess it's just the abort-controller tests? Have you confirmed `addEventListener` works as you expect?

delvedor:
@ronag found it, apparently using the options `{ once: true }` never triggers the event 🤷‍♂️

Now all the tests are passing except for the first test of each file.

test/abort-controller.js
```diff
@@ -0,0 +1,210 @@
...
+  server.listen(0, () => {
+    const client = new Client(`http://localhost:${server.address().port}`)
+    const abortController = new AbortController()
+    t.teardown(client.close.bind(client))
+
+    client.request({ path: '/', method: 'GET', signal: abortController.signal }, (err, response) => {
+      t.ok(err instanceof errors.RequestAbortedError)
+    })
+
+    abortController.abort()
```
ronag, line 26: since you don't have a body, the request gets dispached straight away... you can't abort it like this.

ronag: maybe we should dispatch in next tick?

ronag: I can change this

delvedor: Gotcha! Test updated :)

delvedor: > maybe we should dispatch in next tick?

I'm not sure it is worth, also, in Fastify we had some issue in the past by overusing next tick in hot paths. (cc @mcollina)

ronag: We can make it so we only use it if you send a signal. But I don't think this is such a hot path.

ronag: I've added the nextTick btw, without a fast path for `!signal`.

README.md
```diff
@@ -129,45 +130,59 @@ client.request({
...
-
-Non-idempotent requests will not be pipelined in order
-to avoid indirect failures.
 
-Idempotent requests will be automatically retried if
-they fail due to indirect failure from the request
-at the head of the pipeline. This does not apply to
-idempotent requests with a stream request body.
+ee.emit('abort')
+```
```
ronag, line 185: I think you should mention that aborting the request or response body will also have the same effect.

delvedor:
I think this pr is now ready, @ronag @mcollina do you think we are still missing something?

lib/client.js
```diff
@@ -767,19 +795,30 @@ function resume (client) {
     const onDrain = () => {
       body.resume()
     }
-    const onClose = () => {
+    const onAbort = () => {
       onFinished(new RequestAbortedError())
     }
+    const onSocketClose = () => {
+      onFinished(new SocketError('other side close'))
```
ronag, line 802: nit, other side closed

ronag:
Some of those `setTimeout`s seem unnecessary but I'm fine with landing it with those. @mcollina ?

test/abort-event-emitter.js
```diff
@@ -0,0 +1,227 @@
...
+
+function waitingWithBody (body, type) {
+  test(`Abort while waiting response (with body ${type})`, (t) => {
+    t.plan(1)
+
+    const server = createServer((req, res) => {
+      setTimeout(() => {
+        res.setHeader('content-type', 'text/plain')
+        res.end('hello world')
+      }, 200)
```
mcollina, line 137: Please remove all setTimeouts.

mcollina:
Agreed, no setTimeout

test/abort-event-emitter.js
delvedor: I've removed most of the timeouts, the ones left are needed to simulate a slow response.

test/abort-controller.js
```diff
@@ -0,0 +1,204 @@
...
+    t.plan(2)
+
+    const abortController = new AbortController()
+    const server = createServer((req, res) => {
+      res.writeHead(200, { 'content-type': 'text/plain' })
+      res.write('hello')
+      setTimeout(() => {
+        abortController.abort()
+        res.end('world')
+      }, 100)
```
ronag, line 184: You could move this to `response.body.once('data', ...)`

test/abort-event-emitter.js
```diff
@@ -0,0 +1,205 @@
...
+test('Abort while waiting response (write headers started) (no body)', (t) => {
+  t.plan(1)
+
+  const ee = new EventEmitter()
+  const server = createServer((req, res) => {
+    res.writeHead(200, { 'content-type': 'text/plain' })
+    setTimeout(() => {
+      ee.emit('abort')
+      res.end('hello world')
+    }, 100)
```
ronag, line 77: ditto

test/abort-controller.js
```diff
@@ -0,0 +1,204 @@
...
+
+  server.listen(0, () => {
+    const client = new Client(`http://localhost:${server.address().port}`)
+    t.teardown(client.destroy.bind(client))
+
+    client.request({ path: '/', method: 'GET', signal: abortController.signal }, (err, response) => {
+      t.ok(err instanceof errors.RequestAbortedError)
+    })
+  })
+})
```
ronag, line 88: This test is a little weird? Shouldn't the callback be invoked with `writeHead`? i.e. this test should fail with `t.ok(err)`.

test/abort-controller.js
```diff
@@ -0,0 +1,204 @@
...
+      }, 100)
+    })
+    t.teardown(server.close.bind(server))
+
+    server.listen(0, () => {
+      const client = new Client(`http://localhost:${server.address().port}`)
+      t.teardown(client.destroy.bind(client))
+
+      client.request({ path: '/', method: 'POST', body, signal: abortController.signal }, (err, response) => {
+        t.ok(err instanceof errors.RequestAbortedError)
```
ronag, line 163: Same thing here

test/abort-event-emitter.js
ronag, line 77: same thing here

test/abort-event-emitter.js
```diff
@@ -0,0 +1,205 @@
...
+  test(`Abort while waiting response (write headers started) (with body ${type})`, (t) => {
+    t.plan(1)
+
+    const ee = new EventEmitter()
+    const server = createServer((req, res) => {
+      res.writeHead(200, { 'content-type': 'text/plain' })
+      setTimeout(() => {
+        ee.emit('abort')
+        res.end('hello world')
+      }, 100)
```
ronag, line 155: same thing here

ronag:
Seems like a minor thing with http-parser. It won't invoke `kOnHeadersComplete` unless we start the body, i.e. the following test fails:

```js
test('writeHead invokes callback', (t) => {
  t.plan(1)

    res.writeHead(200)
  })
  t.tearDown(server.close.bind(server))

  server.listen(0, () => {
    const client = new Client({
      hostname: 'localhost',
      port: server.address().port,
      protocol: 'http'
    })
    t.tearDown(client.destroy.bind(client))

    client.request({ path: '/', method: 'GET' }, (err, data) => {
      // never called
      t.ok(err)
    })
  })
})
```

@mcollina: I don't see this as a huge issue.

@delvedor: I don't think the tests should depend on this behavior as we might want to fix it at some point.

test/abort-controller.js
ronag: `writeHead` doesn't flush.

test/abort-controller.js
```diff
@@ -0,0 +1,204 @@
...
+  test(`Abort while waiting response (write headers started) (with body ${type})`, (t) => {
+    t.plan(1)
+
+    const abortController = new AbortController()
+    const server = createServer((req, res) => {
+      res.writeHead(200, { 'content-type': 'text/plain' })
+      setTimeout(() => {
+        abortController.abort()
+        res.end('hello world')
+      }, 100)
```
ronag, line 154: I don't think this timeout is needed?

test/abort-controller.js
```diff
@@ -0,0 +1,204 @@
...
+waitingWithBody(createReadStream(__filename), 'stream')
+waitingWithBody(new Uint8Array([42]), 'Uint8Array')
+
+function writeHeadersStartedWithBody (body, type) {
+  test(`Abort while waiting response (write headers started) (with body ${type})`, (t) => {
+    t.plan(1)
+
+    const abortController = new AbortController()
+    const server = createServer((req, res) => {
+      res.writeHead(200, { 'content-type': 'text/plain' })
```
ronag, line 150: This does nothing until `res.end` is called. It doesn't flush

ronag: You probablye want `res.flushHeaders()` directly afterwards.

PR #170, opened 2020-05-20
feat: request timeout

ronag:


ronag:
@delvedor PR for timeout based on the abort PR.

ronag:
> Out of curiosity, why not implement the entire timeout logic inside the request method?

Because we want it to work for `pipeline` and `stream` as well?

> Also, would make sense to use retimer? (cc @mcollina )

Note required since we are not refreshing the timer. Also `retimer` was for before node implemented `Timer.refresh()`.

delvedor:
> Because we want it to work for pipeline and stream as well?

Sorry, I mean where to write the implementation, but after re-reading the code, it makes absolute sense the way you did implement it :)

Can you add a few tests, also one that uses the `abort-controller`?
Also, with this implementation the timer starts when the user schedules the request or when it is actually sent by undici?

ronag:
> > Because we want it to work for pipeline and stream as well?
> 
> Sorry, I mean where to write the implementation, but after re-reading the code, it makes absolute sense the way you did implement it :)

🖖

> Can you add a few tests, also one that uses the `abort-controller`?

I was hoping you could help out? 😅 Also needs docs.

> Also, with this implementation the timer starts when the user schedules the request or when it is actually sent by undici?

When the user schedules the request.

delvedor:
> I was hoping you could help out? 😅 Also needs docs.

Sure thing! But this time I'll need to use `setTimeout` 🤣
I'll push some code tomorrow morning (CEST)!

> When the user schedules the request.

👍

ronag:
> But this time I'll need to use setTimeout 🤣

yea, probably. For bonus points you might be able to use https://sinonjs.org/releases/v9.0.2/fake-timers/ :smile:. I have no experience with it myself.

delvedor:
To avoid confusion for the users, I was thinking to rename the "top level" timeout to `socketTimeout`, while the one introduced in this pr `requestTimeout`, what do you think?

mcollina:
+1

delvedor:
I did some changes :)

```
- timeout (client opt) -> socketTimeout
- timeout (request opt) -> requestTimeout
- TimeoutError -> SocketTimeoutError
- Added RequestTimeoutError
```

Do you think there is still some test that needs to be done?

delvedor:
I had to comment the following line, otherwise many tests will fail, I'm not sure why it's happening, would you mind giving it a look?

https://github.com/mcollina/undici/blob/56792165d3dd253cf23013f339348ffe22062c07/lib/client.js#L323

delvedor:
Oks, I've found the source of the bug, apparently, if you have a request timeout running and call `close` or `destroy`, the timeout will not be cleared.

test/request-timeout.js
```diff
@@ -342,3 +342,41 @@ test('Request options overrides global option', { skip: 'buggy' }, (t) => {
...
+
+    client.request({ path: '/', method: 'GET', requestTimeout: 100 }, (err, response) => {
+      t.ok(err instanceof errors.ClientDestroyedError)
+    })
+
+    client.destroy(err => t.error(err))
+  })
+})
+
+test('client.close should cancel the timeout', (t) => {
```
ronag, line 365: close should not cancel timeout. It should gracefully wait for all enqueued requests to finish and not allow any new ones.

README.md
```diff
@@ -44,7 +44,11 @@ address. The path is discarded.
 
 Options:
 
-- `timeout`, the timeout after which a request will time out, in
+- `socketTimeout`, the timeout after which a socket will time out, in
+  milliseconds.
+  Default: `30e3` milliseconds (30s).
+
+- `requestTimeout`, the timeout after which a request will time out, in
   milliseconds.
```
ronag, line 52: I don't think we should have this enabled by default. Also I think we need to be a bit more explicit about what exactly a request timeout is.

test/request-timeout.js
delvedor: I think it should (if the request has not been sent yet, like in this test), otherwise the callback will be called again with `RequestTimeoutError`.

README.md
delvedor: I agree that the description should be improved, while regarding the timeout I disagree. From my experience with Fastify, it is always better to have a sane default (30s) rather than not having one at all.
We had a similar issue with the Plugin timeout where we were waiting indefinitely.

ronag: We do have a timeout on the socket. Shouldn't that be enough?

ronag: I'm a little worried with the request timeout thing in the cases where the user spams with requests, due to pipelining they might never finish within the request timeout, and then the user keeps retrying them.

test/request-timeout.js
ronag: >  otherwise the callback will be called again with RequestTimeoutError,

There is some other bug here then. `close` should wait for everything finish normally, i.e. no errors unless something unexpected happened.

delvedor: Do you have time to track it down?

ronag: Comment out this test and finish up the rest and I will sort it out later.

mcollina: `close()` should finish things normally (or have them timeout).  `destroy()` should cancel it.

test/request-timeout.js
```diff
@@ -0,0 +1,391 @@
...
+  server.listen(0, () => {
+    const client = new Client(`http://localhost:${server.address().port}`)
+    t.teardown(client.destroy.bind(client))
+
+    client.request({ path: '/', method: 'GET', requestTimeout: 100 }, (err, response) => {
+      t.ok(err instanceof errors.RequestTimeoutError)
+    })
+
+    client.close((err) => {
+      t.error(err)
```
ronag, line 386: @mcollina @delvedor One problem... `RequestTimeoutError`s do not propagate to `close`. There is no elegant way to make it so with the current code... is this a problem?

Only "hard" errors that destroy the socket will propagate to `close` at the moment.

test/request-timeout.js
ronag: I have fixed this. However, please see the commit below in regard to error propagation to close.

README.md
ronag: @mcollina: opinion, `requestTimeout` enabled by default or not?

lib/client.js
```diff
@@ -317,6 +321,8 @@ class Client extends EventEmitter {
       return
     }
 
+    opts.requestTimeout = opts.requestTimeout || this[kRequestTimeout]
```
ronag, line 324: I think this should be `timeout` here. `requestTimeout` is kind of double in this context, i.e.

```client.request({ timeout: 10e3, ...opts })```

File: test/request-timeout.js (hunk 3 above)
Comment (ronag) [reply test/request-timeout.js:386]: The error modes of `close` are a bit ambiguous at the moment.

File: README.md (hunk 2 above)
Comment (delvedor) [reply README.md:52]: I would prefer to have it because in the end every use will configure it at some point.
But if you think that it's overkill, I'm also fine not having it.
Still, there should be a way to configure it both globally and for a single request.

File: lib/client.js (hunk 4 above)
Comment (delvedor) [reply lib/client.js:324]: Not sure I'm following 🤔

Comment (ronag) [reply lib/client.js:324]: The option to `request` should be `timeout` instead of `requestTimeout`. Otherwise it's duplicate. In my opinion.

File: test/request-timeout.js (hunk 3 above)
Comment (delvedor) [reply test/request-timeout.js:386]: I like the current solution, it makes sense that `destroy` closes everything as soon as you call it, while `close` uses a graceful shutdown :)

File: README.md (hunk 2 above)
Comment (ronag) [reply README.md:52]: > Still, there should be a way to configure it both globally and for a single request.

Agreed. I'm mostly concerned with the default.

File: test/request-timeout.js (hunk 3 above)
Comment (ronag) [reply test/request-timeout.js:386]: With the current solution requests can still timeout after close and those errors are not propagated to close... which is the problem, or not? needs to be documented at least

**Comment by delvedor** [comment]
I've updated the test scripts, now `npm test` no longer checks the coverage and as result the tests are way faster.
I've added `npm run coverage-ui` for checking the actual code coverage :)

File: lib/client.js (hunk 4 above)
Comment (delvedor) [reply lib/client.js:324]: I see your point!
I agree that it's a "useless" duplication, but on the other hand by keeping the same name as the global option it is more intuitive when you think about overriding the globals. What do you think?

File: test/request-timeout.js (hunk 3 above)
Comment (delvedor) [reply test/request-timeout.js:386]: I don't see it as a problem. @mcollina?

File: package.json [hunk 5]
```diff
@@ -4,7 +4,8 @@
   "description": "An HTTP/1.1 client, written from scratch for Node.js",
   "main": "index.js",
   "scripts": {
-    "test": "standard | snazzy && tap test/*.js"
+    "test": "standard | snazzy && tap test/*.js --no-coverage",
+    "coverage-ui": "tap test/*.js --coverage-report=html"
```
mcollina, line 8: Can you please do this in another PR? I think we should ignore ts, flow, esm, and jsx first. Once those are done we can keep coverage in as it would be faster anyway.

README.md
mcollina: > opinion, requestTimeout enabled by default or not?

As long as we support `requestTimeout: 0` to disable it, I do not see a difference. Probably it's safer to keep a default so a long running request on the server will not have a pipeline starve.

test/request-timeout.js
mcollina: I don't understand this. Can you please recap it for me?

lib/client.js
mcollina: `request({ timeout })` is likely nicer. However there might be some confusion with the socket timeout, which has a completely different meaning.

README.md
delvedor: Agreed! I’ll work on this :)

test/request-timeout.js
ronag: So `close(err)` will be invoked with an error if any request fails... unless the request errors due to an error from abort or timeout.

README.md
delvedor: Done 🎉

README.md
```diff
@@ -78,7 +78,7 @@ Options:
 * `headers`, an object with header-value pairs.
 * `signal`, either an `AbortController` or an `EventEmitter`.
 * `requestTimeout`, the timeout after which a request will time out, in
-  milliseconds.
+  milliseconds.  Use `0` to disable it entirely.
```
ronag, line 81: Probably the same on socketTimeout?

lib/client.js
```diff
@@ -242,7 +242,7 @@ class Client extends EventEmitter {
     this[kPipelining] = pipelining || 1
     this[kUrl] = url
     this[kSocketTimeout] = socketTimeout || 30e3
-    this[kRequestTimeout] = requestTimeout || 30e3
+    this[kRequestTimeout] = requestTimeout === undefined ? 30e3 : requestTimeout
```
ronag, line 245: == null

lib/client.js
```diff
@@ -321,7 +321,9 @@ class Client extends EventEmitter {
       return
     }
 
-    opts.requestTimeout = opts.requestTimeout || this[kRequestTimeout]
+    opts.requestTimeout = opts.requestTimeout !== undefined
```
ronag, line 324: != null

test/request-timeout.js
ronag: Because of this I'm kind of considering removing `close` until we figure out what the exact semantics of it is...

One definition could be that `client.close()` should be equal to `Promise.all(client.allRequestsAsPromises)`? Would require a little work to make that happen.

ronag: I will fix this as part of https://github.com/mcollina/undici/issues/175 after this PR has landed. So this is no longer blocking.

README.md
```diff
@@ -73,6 +77,9 @@ Options:
 * `body`, it can be a `String`, a `Buffer`, `Uint8Array` or a `stream.Readable`.
 * `headers`, an object with header-value pairs.
 * `signal`, either an `AbortController` or an `EventEmitter`.
+* `requestTimeout`, the timeout after which a request will time out, in
+  milliseconds.  Use `0` to disable it entirely.
```
ronag, line 81: I think `null` for disabling is more intuitive and normal.

README.md
```diff
@@ -44,8 +44,12 @@ address. The path is discarded.
...
 Options:
 
-- `timeout`, the timeout after which a request will time out, in
-  milliseconds.
+- `socketTimeout`, the timeout after which a socket will time out, in
+  milliseconds. Use `0` to disable it entirely.
+  Default: `30e3` milliseconds (30s).
+
+- `requestTimeout`, the timeout after which a request will time out, in
+  milliseconds. Use `0` to disable it entirely.
```
ronag, line 52: ditto

README.md
```diff
@@ -44,8 +44,12 @@ address. The path is discarded.
 
 Options:
 
-- `timeout`, the timeout after which a request will time out, in
-  milliseconds.
+- `socketTimeout`, the timeout after which a socket will time out, in
+  milliseconds. Use `0` to disable it entirely.
```
ronag, line 48: ditto

README.md
delvedor: I see your point, but using `0` to disable a numeric configuration is a well-established standard in the community IMHO.

ronag: Yep, looking through the node docs you are right.

ronag:
@delvedor is there anything left on this PR (since it's still draft)?.

delvedor:
@ronag I think it should be good to go!

ronag:
@delvedor Do you think you could fix the conflicts and cleanup the commits?

delvedor:
I don't see any conflict 🤔 
Regarding the commits, once we merge, we can use the "Squash and merge" feature from Github 😄 

@mcollina can you make it the default?

ronag:
Aah! Sorry. I was using the rebase and merge which was complaining about conflicts. Squash and merge is fine.

PR #171, opened 2020-05-20
fix: InvalidReturnValueError

ronag:


lib/client.js
```diff
@@ -422,10 +426,9 @@ class Client extends EventEmitter {
         return
       }
 
-      if (!body) {
-        // TODO: InvalidReturnValueError?
+      if (!body || typeof body.pipe !== 'function') {
         if (!ret.destroyed) {
-          ret.destroy(new InvalidArgumentError('invalid body'))
+          ret.destroy(new InvalidReturnValueError('invalid body'))
```
delvedor, line 431: I would write a more descriptive error message, as the user might not understand what's going on here.

ronag: How about now?

PR #174, opened 2020-05-20
perf: optimize stream.pipeline

ronag:
- Make req Readable instead of PassThrough.
- Bypass Duplex writable buffering.

Brings stream.pipeline to parity with stream.requests.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -27,6 +27,7 @@ const {
   kUrl,
   kWriting,
   kQueue,
+  kResume,
   kTimeout,
   kTLSOpts,
   kClosed,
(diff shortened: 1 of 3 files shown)


lib/client.js
```diff
@@ -360,22 +361,40 @@ class Client extends EventEmitter {
...
+        this._read()
+        callback(err)
+      }
+    })
     let res
     let body
 
     const ret = new Duplex({
       autoDestroy: true,
+      writableHighWaterMark: 1,
```
ronag, line 383: This will disable buffering in `Writable` and always pick the fast path.

PR #176, opened 2020-05-23
fix: close semantics

ronag:
https://github.com/mcollina/undici/issues/175

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -637,8 +637,8 @@ class Client extends EventEmitter {
     }
   }
 
-  close (cb) {
-    if (cb === undefined) {
+  close (callback) {
+    if (callback === undefined) {
       return new Promise((resolve, reject) => {
         this.close((err, data) => {
           return err ? reject(err) : resolve(data)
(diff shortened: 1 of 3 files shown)


lib/client.js
```diff
@@ -621,17 +644,31 @@ class Client extends EventEmitter {
...
+        callback(err, null)
+      }
+    }
+
     this[kClosed] = true
+    this[kOnDestroyed].push(onDestroyed)
+
+    for (const request of this[kQueue].slice(this[kInflight])) {
+      if (request.signal) {
+        request.signal.on('error', onDestroyed)
```
ronag, line 664: should this still wait for the socket to be destroyed before invoking the callback?

ronag:
I'm struggling with this. It's rather complicated.

I don't think the semantics of:

```js
const requests = makeLotsOfRequests()
await Promise.all(requests)
```

Is correct either since it doesn't actually wait for the responses.

We have to wait for the response (if exists) to complete as well.

ronag:
@mcollina: Much of the difficult here lies with propagating the errors... would it be an option that `close(callback)` does not propagate errors? i.e. it just closes the client for further requests and waits for all requests to complete, whether they error or not.

Essentially it would be:

```js
Promise.allSettled(...)
  .then(() => callback(null))
  .catch((err) => callback(err))
```

mcollina:
ok for me

ronag:
this is ready for review

PR #182, opened 2020-05-24
doc: add ronag as collaborator

ronag:


ronag:
Not sure exactly what the preferred format of this is.

PR #189, opened 2020-05-25
fix: lowercase response header keys

ronag:
Fixes: https://github.com/mcollina/undici/issues/188

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,6 +107,8 @@ The `data` parameter in `callback` is defined as follow:
   either fully consume or destroy the body unless there is an error, or no further requests
   will be processed.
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -107,6 +107,8 @@ The `data` parameter in `callback` is defined as follow:
   either fully consume or destroy the body unless there is an error, or no further requests
   will be processed.
 
+`headers` is an object where all keys has been lowercased.
```
delvedor, line 110: Nit :)

```suggestion
`headers` is an object where all keys have been lowercased.
```

ronag: You did not see this. It never happened. :smile:

PR #191, opened 2020-05-25
fix: URL default port validation

delvedor:
The current port validation will fail if you use the default http/https port or not use a port at all, for example:
```js
undici('http://localhost')
undici('http://localhost:80')
undici('https://localhost')
undici('https://localhost:443')
```
All the examples above will throw an exception, even if the URL is valid.

Given that internally we are using the `URL` constructor, I think we should leave to it the bad port validation, for example, the following will cause an exception:
```js
undici('http://localhost:foobar')
```

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -201,6 +201,10 @@ class Client extends EventEmitter {
       throw new InvalidArgumentError('invalid url')
     }
 
+    if (url.port != null && url.port !== '' && !Number.isFinite(parseInt(url.port))) {
+      throw new InvalidArgumentError('invalid port')
+    }
+
     if (url.hostname != null && typeof url.hostname !== 'string') {
       throw new InvalidArgumentError('invalid hostname')
     }
--- a/lib/client.js
+++ b/lib/client.js
@@ -201,10 +201,6 @@ class Client extends EventEmitter {
       throw new InvalidArgumentError('invalid url')
     }
 
-    if (url.port != null && !Number.isFinite(parseInt(url.port))) {
-      throw new InvalidArgumentError('invalid port')
-    }
-
     if (url.hostname != null && typeof url.hostname !== 'string') {
       throw new InvalidArgumentError('invalid hostname')
     }
--- a/test/client-errors.js
+++ b/test/client-errors.js
@@ -241,7 +241,14 @@ test('POST with chunked encoding that errors and pipelining 1 should reconnect',
 })
 
 test('invalid options throws', (t) => {
-  t.plan(18)
+  t.plan(20)
+
+  try {
+    new Client({ port: 'foobar' }) // eslint-disable-line
+  } catch (err) {
+    t.ok(err instanceof errors.InvalidArgumentError)
+    t.strictEqual(err.message, 'invalid port')
+  }
 
   try {
     new Client(new URL('http://asd:200/somepath')) // eslint-disable-line
--- a/test/client-errors.js
+++ b/test/client-errors.js
@@ -241,14 +241,7 @@ test('POST with chunked encoding that errors and pipelining 1 should reconnect',
 })
 
 test('invalid options throws', (t) => {
-  t.plan(20)
-
-  try {
-    new Client(new URL('asd://asd')) // eslint-disable-line
-  } catch (err) {
-    t.ok(err instanceof errors.InvalidArgumentError)
-    t.strictEqual(err.message, 'invalid port')
-  }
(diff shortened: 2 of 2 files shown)


delvedor:
The reason why the validation is failing, it's due to how `URL` works:
```
> new URL('http://localhost:80')
URL {
  href: 'http://localhost/',
  origin: 'http://localhost',
  protocol: 'http:',
  username: '',
  password: '',
  host: 'localhost',
  hostname: 'localhost',
  port: '',
  pathname: '/',
  search: '',
  hash: ''
}
> new URL('http://localhost')
URL {
  href: 'http://localhost/',
  origin: 'http://localhost',
  protocol: 'http:',
  username: '',
  password: '',
  host: 'localhost',
  hostname: 'localhost',
  port: '',
  pathname: '/',
  search: '',
  hash: ''
}
```

lib/client.js
```diff
@@ -201,10 +201,6 @@ class Client extends EventEmitter {
       throw new InvalidArgumentError('invalid url')
     }
```
ronag, line 203: I think I would prefer something like:

```js
if (url.port == null) {
  if (protocol === 'http:') {
    url = { ...url, port: 80 }
  } else if (protocol === 'https:') {
    url = { ...url, port: 443 }
  }
}
```

delvedor: That would work only if the url would be a plain object, but if we use the `URL` constructor (as we are, ad we should probably enforce), restructuring it will result in a not really nice object:
```
> var url = new URL('http://localhost')
undefined
> url
URL {
  href: 'http://localhost/',
  origin: 'http://localhost',
  protocol: 'http:',
  username: '',
  password: '',
  host: 'localhost',
  hostname: 'localhost',
  port: '',
  pathname: '/',
  search: '',
  hash: ''
}
> { ...url, port: 80 }
{
  port: 80,
  [Symbol(context)]: URLContext {
    flags: 400,
    scheme: 'http:',
    username: '',
    password: '',
    host: 'localhost',
    port: null,
    path: [ '' ],
    query: null,
    fragment: null
  },
  [Symbol(query)]: URLSearchParams {}
}
```

delvedor: Maybe we should add a new validation like the following:
```js
if (typeof url !== 'string' && url instanceof URL !== true) {
  throw new InvalidArgumentError('invalid url')
}
```

ronag: No, we should allow url:like objects.

lib/client.js
```diff
@@ -201,10 +201,6 @@ class Client extends EventEmitter {
...
 
-    if (url.port != null && !Number.isFinite(parseInt(url.port))) {
-      throw new InvalidArgumentError('invalid port')
-    }
```
ronag, line 206: what about:

```js    
if (url.port != null && url.port !== '' && !Number.isFinite(parseInt(url.port))) {
  throw new InvalidArgumentError('invalid port')
}
```

lib/client.js
delvedor: Out of curiosity, why? URL is the whatwg standard, and given that it has some subtle behavior, like the handling of `80` and `443`, supporting both the URL and a url-like object looks like a bug-prone thing to support.

Furthermore, but this is my personal opinion, configuration is the root of all bugs :D

ronag: I have use cases where we have an options object with `port, hostname, path` it seems silly and error prone to have to do:

`${protocol}//${hostname}${port ? `:${port}`: ''}${path}`

ronag: I personally don't like the `URL` way because a valid URL can contain stuff that we don't want, e.g. path and well strings are unstructured.

delvedor:
I've added back the port validation with your suggestion.
I still think that supporting url like objects will bite us in the future :S

ronag:
> I still think that supporting url like objects will bite us in the future :S

Care to expand?

delvedor:
Sure! :)
I had some issues caused by the same thing we are discussing now, where a user could pass an url in different ways, and it's easy to miss something in the validation code. As a result, you get an issue that might be straightforward to fix, but it will cause either an increased code complexity, or a breaking change.
While if you enforce a clear contract, for example, string or URL, your internal code will be easier in many ways! Both validation wise and code wise (eg: if you know it is an URL object, you also know what to find and where).

But as said, this is my opinion and I don't want to enforce it. Since you said that there are cases where the url like object is useful, we could work on improving our validation and/or try to convert everything to an URL internally.

PR #192, opened 2020-05-25
fix: retry delay

ronag:
Avoid delay during first retry. Also add test for multiple retries.

### DIFF
--- a/test/client-reconnect.js
+++ b/test/client-reconnect.js
@@ -1,0 +1,35 @@
+'use strict'
+
+const { test } = require('tap')
+const { Client } = require('..')
+const { createServer } = require('http')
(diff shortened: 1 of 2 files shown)


test/client-reconnect.js
```diff
@@ -0,0 +1,35 @@
...
+
+  const clock = FakeTimers.install()
+  t.teardown(clock.uninstall.bind(clock))
+
+  const server = createServer((req, res) => {
+    res.end()
+  })
+  t.tearDown(server.close.bind(server))
+
+  const client = new Client('http://localhost:5555')
```
ronag, line 19: unsure what works well here... need a port before the server will start accepting connections

PR #194, opened 2020-05-25
perf: faster test run

ronag:
Makes `npm run test` more than twice as fast.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,7 @@
   "description": "An HTTP/1.1 client, written from scratch for Node.js",
   "main": "index.js",
   "scripts": {
-    "test": "standard | snazzy && tap test/*.js --100"
+    "test": "standard | snazzy && tap test/*.js --100 --no-ts --no-jsx --no-flow --no-esm"
   },
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -39,5 +39,7 @@
   "dependencies": {
     "http-parser-js": "^0.5.2"
   },
-  "pre-commit": ["coverage"]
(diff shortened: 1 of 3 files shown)


ronag:
> if you disable typescript, flow, esm, and jsx you can achieve the same speed without removing functionality.

Changed. It's faster... but not as far. 22s vs 14s

mcollina:
running code coverage for every run locally is important, otherwise prs will fail while it would pass locally.

ronag:
> running code coverage for every run locally is important, otherwise prs will fail while it would pass locally.

I'm original suggestion was to always run it through the pre-commit hook, but not when manually running `npm test`.

mcollina:
Ok, let's do that.

ronag:
PTAL

package.json
```diff
@@ -4,7 +4,9 @@
   "description": "An HTTP/1.1 client, written from scratch for Node.js",
   "main": "index.js",
   "scripts": {
-    "test": "standard | snazzy && tap test/*.js --100 --no-ts --no-jsx --no-flow --no-esm"
+    "lint": "standard | snazzy",
+    "test": "tap test/*.js --no-coverage",
+    "coverage": "standard | snazzy && tap test/*.js --100 --no-ts --no-jsx --no-flow --no-esm"
```
mcollina, line 9: please move the `--no-ts --to-jsx --no-flow --no-esm` to a `.taprc` file. See https://github.com/fastify/fastify/blob/master/.taprc

ronag: I guess coverage: true is the same as --100?

PR #205, opened 2020-05-27
doc: document spec non-compliance

ronag:
Fixes: https://github.com/mcollina/undici/issues/200
Fixes: https://github.com/mcollina/undici/issues/79


ronag:
@szmarczak

ronag:
It is also an option to actually fix any of this items of we consider them important before a next release.

PR #207, opened 2020-05-30
fix: request.pipeline edgecases

ronag:
The optimizated verison of request.pipeline had some hard
to determine edge cases. Reverting back to non-optimized
version until a better solution can be implemented.

Fixes: https://github.com/mcollina/undici/issues/206

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -9,10 +9,8 @@ const {
   InvalidReturnValueError,
   RequestAbortedError
 } = require('./errors')
-const {
-  kResume
-} = require('./symbols')
 const ClientBase = require('./client-base')
+const assert = require('assert')
 
 class Client extends ClientBase {
   request (opts, callback) {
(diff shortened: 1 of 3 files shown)


ronag:
Will follow up at some point with a version where the `Duplex` writes directly to the socket.

lib/client.js
```diff
@@ -107,37 +103,38 @@ class Client extends ClientBase {
...
+        // to a successful write to the socket.
+        req.write(chunk, encoding, callback)
       },
       final (callback) {
-        req.push(null)
-        callback()
+        // TODO: The callback might not be invoked
+        // on error in pre Node 14.
+        // TODO: The callback might not correspond
+        // to a successful flush of the socket.
```
ronag, line 123: These notes probably don't have much if any impact in practical usage. But it might not strictly follow the expected API contract.

ronag: Will require some further work (#208) to fully resolve but I don't think this is a blocker for release.

lib/client.js
```diff
@@ -88,17 +86,15 @@ class Client extends ClientBase {
       return new PassThrough().destroy(new InvalidArgumentError('invalid handler'))
     }
 
-    let req = new Readable({
+    // TODO: Let ret write directly to socket without intermediate.
+    const req = new PassThrough({
```
mcollina, line 90: This should really be a Readable. Otherwise it’s costing us dearly.

ronag: Ok. I'll give it another try.

ronag:
Fixed through https://github.com/mcollina/undici/commit/c553ba19c71bb4d00e0699f43c8bdd8c91d96d42 instead.

PR #211, opened 2020-06-04
max connection duration proof-of-concept

omsmith:
Been watching some of the recent development here. A more performant http client for use in a reverse proxy (especially with the `.stream` method) is potentially exciting and something I'll investigate using in time.

Something we do by extending the core node Http(s)Agent is to destroy the kept-alive sockets after n minutes. This is to deal with unfortunate behaviour of some servers and cloud environments that don't necessarily drop connections when "moving"/coming out of load. AWS' ElasticBeanstalk has the option for a "CNAME swap" deployment model for instance, where both environments stay running and DNS is just updated to switch between A and B.

Put together a PoC how supporting that behaviour could be implemented in undici. The real goal is just "after some timeout has passed, re-resolve the hostname"; theoretically dropping the connection if it stays the same is unnecessary (although "same" is variable for load-balanced DNS responses).

Anyway, curious if you'd be interested in supporting such a case at the client and/or pool level, and if so I can bring the PR the rest of the way - or take it in a different direction.

### DIFF
--- a/lib/client-base.js
+++ b/lib/client-base.js
@@ -37,7 +37,10 @@ const {
   kParser,
   kSocket,
   kEnqueue,
-  kClient
+  kClient,
+  kMaxConnectionDuration,
+  kConnectionDurationTimer,
+  kNeedsNewConnection
 } = require('./symbols')
 
 const CRLF = Buffer.from('\r\n', 'ascii')
@@ -52,6 +55,7 @@ class ClientBase extends EventEmitter {
     maxAbortedPayload,
     socketTimeout,
     requestTimeout,
+    maxConnectionDuration,
     pipelining,
     tls
   } = {}) {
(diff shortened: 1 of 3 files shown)


ronag:
This could be implemented in user space or Pool as well, possibly a bit simpler:

```js
let client
function connect (opts) {
  let timeout
  client.on('connect', () => {
    timeout setTimeout(() => {
      client.close() // gracefully finishes all running requwests
      connect(opts
    }, maxDuration)
  }).on('disconnect', () => {
    clearTimeout(timeout)
  )}
}

connect(opts)

// use client
```

omsmith:
> and when the pipeline becomes empty reset the connection, i.e. avoid unnecessarily killing any running requests.

oh that was my intention by the `running` check in resume, but I may have misunderstood. haven't built a test-suite yet

omsmith:
> This could be implemented in user space or Pool as well, possibly a bit simpler

Yeah, I guess given userspace would already be handling the hostname -> client/pool mapping it gives a natural spot for it.

ronag:
>oh that was my intention by the running check in resume, but I may have misunderstood. haven't built a test-suite yet

Ah, nice. Sorry I missed that.

omsmith:
Thanks for taking the time to have a look at this.

Other things have come up and I don't think I'll be completing this any time soon so I'll close it up.

Might open something in the future, pethaps at the pool level so pending requests don't get blocked. Like the current implementation.

ronag:
Sure! Feel free to re-open if you get time and we can continue discussing.

PR #213, opened 2020-06-05
fix: avoid unhandled exception on early abort

ronag:
- Early abort of body could cause unhandled exception.
- Abort could deadlock close.

### DIFF
--- a/lib/client-base.js
+++ b/lib/client-base.js
@@ -546,14 +546,6 @@ function resume (client) {
       return
     }
 
-    if (!client.connected) {
-      return
-    }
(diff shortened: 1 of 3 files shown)


lib/client-base.js
```diff
@@ -563,6 +555,14 @@ function resume (client) {
       continue
     }
 
+    if (!client.connected) {
+      return
+    }
```
ronag, line 560: We won't connect if there is no pending request, therefore it would never splice away cancelled requests, therefore it wouldn't close the client.

PR #219, opened 2020-06-11
fix: only retry recoverable socket errors

ronag:
If the socket fails while there is no active request then it might not be recoverable and we should error and destroy.

### DIFF
--- a/lib/client-base.js
+++ b/lib/client-base.js
@@ -477,6 +477,19 @@ function connect (client) {
         while (client.pending && client[kQueue][client[kPendingIdx]].servername === servername) {
           client[kQueue][client[kPendingIdx]++].invoke(err, null)
         }
+      } else if (
+        !client.running &&
+        err.code !== 'ECONNRESET' &&
+        err.code !== 'ECONNREFUSED' &&
+        err.code !== 'EHOSTUNREACH' &&
+        err.code !== 'EHOSTDOWN' &&
+        err.code !== 'UND_ERR_SOCKET'
+      ) {
+        // Error is not caused by running request and not a recoverable
+        // socket error.
+        for (const request of client[kQueue].splice(client[kPendingIdx])) {
+          request.invoke(err, null)
+        }
       }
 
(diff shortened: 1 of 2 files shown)


ronag:
Alternatively emit error on the `Client`?

lib/client-base.js
```diff
@@ -474,7 +484,11 @@ function connect (client) {
 
       const err = socket[kError]
 
-      parser.destroy(err)
+      if (!client.running && !isConnectionError(err)) {
```
ronag, line 487: client.running is in order to only apply this if no request is currently running which might have caused some other error, e.g. aborted.

ronag: We could also turn this around and error on errors we know are unrecoverable... not sure which approach is best... unrecoverable might be a long list...

ronag:
@mcollina: Should we maybe emit `'error'` or some other event on the `Client` in these cases? I'm unsure...

Another alternative is to not do this PR and instead document that user should handle the `'disconnect'` event and determine whether it's "fatal" or not.

@delvedor any opinion?

mcollina:
>  some other event on the Client in these cases? 

Definitely not `error`. Didn't we have `reconnect`? I think that's enough.

ronag:
> Didn't we have reconnect? I think that's enough.

We have `'disconnect'`. But that's no indication to the user whether it will recover or not. Usually `'disconnect'` means that it will retry.

mcollina:
I think having a `retry` would be nice.

ronag:
@mcollina: Another solution, which I feel is less controversial would be:

- If socket errors without any request being dispatched/running, error all pending requests with this error.

Basically what we are doing here https://github.com/mcollina/undici/commit/e1f176c9e91fd82aaa4516e43c42b69a1cea3e86#diff-d4588047630c1180b519ff7191a2cf29R475 but with a "if (!running)" condition.

mcollina:
go for it

ronag:
@mcollina PTAL

PR #226, opened 2020-06-22
Add support for unix:// connections

yoshuawuyts:
Adds support for HTTP over `AF_UNIX` sockets (UDS) by enabling the `unix://` scheme. Most of the functionality for this exists as part of Node's `net` module, and all this patch does is ensure the connection is created the right way.

This is useful for projects that want to use HTTP as an RPC protocol, but don't want to risk exposing a TCP socket to the public. 
If this direction seems good I'd be happy to author tests for this as well.

This patch also fixes a bug where in some cases an empty `Host` header was sent which didn't seem intentional.

Overall I hope this patch is useful. Thanks heaps!

cc/ @mcollina 

## Example

```js
const { Client } = require('undici')
const client = new Client(`unix:///tmp/hello.sock`)

client.request({
    path: '/',
    method: 'GET'
}, function (err, data) {
    if (err) throw err

    const {
        statusCode,
        headers,
        body
    } = data

    body.setEncoding('utf8')
    body.on('data', console.log)

    client.close()
})
```

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -104,7 +104,7 @@ class Request extends AsyncResource {
       }
 
       if (!hostHeader) {
-        header += `host: ${hostname}\r\n`
+        header += `host: ${hostname || 'localhost'}\r\n`
       }
 
       header += 'connection: keep-alive\r\n'
--- a/lib/client-base.js
+++ b/lib/client-base.js
@@ -72,11 +72,11 @@ class ClientBase extends EventEmitter {
       throw new InvalidArgumentError('invalid hostname')
     }
 
-    if (!/https?/.test(url.protocol)) {
+    if (!/[https?|unix]/.test(url.protocol)) {
       throw new InvalidArgumentError('invalid protocol')
     }
 
-    if (/\/.+/.test(url.pathname) || url.search || url.hash) {
+    if ((url.protocol != 'unix:' && /\/.+/.test(url.pathname)) || url.search || url.hash) {
       throw new InvalidArgumentError('invalid url')
     }
 
@@ -136,7 +136,7 @@ class ClientBase extends EventEmitter {
       this[kSocket].connecting !== true &&
       // Older versions of Node don't set secureConnecting to false.
       (this[kSocket].authorized !== false ||
-       this[kSocket].authorizationError
+        this[kSocket].authorizationError
       ) &&
       !this[kSocket].destroyed
     )
@@ -436,14 +436,22 @@ function connect (client) {
   assert(!client[kSocket])
   assert(!client[kRetryTimeout])
 
-  const { protocol, port, hostname } = client[kUrl]
+  const { protocol, port, hostname, pathname } = client[kUrl]
   const servername = client[kServerName] || (client[kTLSOpts] && client[kTLSOpts].servername)
-  const socket = protocol === 'https:'
-    ? tls.connect(port || /* istanbul ignore next */ 443, hostname, {
+
+  let socket;
+  if (protocol == 'https:') {
+    socket = tls.connect(port || /* istanbul ignore next */ 443, hostname, {
       ...client[kTLSOpts],
       servername
     })
-    : net.connect(port || /* istanbul ignore next */ 80, hostname)
+  } else if (protocol == 'http:') {
+    socket = net.connect(port || /* istanbul ignore next */ 80, hostname)
+  } else if (protocol == 'unix:') {
+    socket = net.connect(pathname)
+  } else {
+    throw new InvalidArgumentError('invalid protocol')
+  }
 
   client[kSocket] = socket
 
@@ -768,3 +776,4 @@ function parseHeaders (headers) {
 }
 
 module.exports = ClientBase
+
--- a/lib/client-base.js
+++ b/lib/client-base.js
@@ -72,7 +72,7 @@ class ClientBase extends EventEmitter {
       throw new InvalidArgumentError('invalid hostname')
     }
 
-    if (!/^(http|https|unix):$/.test(url.protocol)) {
+    if (!/^(http|https|unix)/.test(url.protocol)) {
       throw new InvalidArgumentError('invalid protocol')
     }
 
--- a/lib/client-base.js
+++ b/lib/client-base.js
@@ -72,7 +72,7 @@ class ClientBase extends EventEmitter {
       throw new InvalidArgumentError('invalid hostname')
     }
 
-    if (!/[https?|unix]/.test(url.protocol)) {
+    if (!/^(http|https|unix):$/.test(url.protocol)) {
       throw new InvalidArgumentError('invalid protocol')
     }
 
@@ -445,12 +445,10 @@ function connect (client) {
       ...client[kTLSOpts],
       servername
     })
-  } else if (protocol === 'http:') {
-    socket = net.connect(port || /* istanbul ignore next */ 80, hostname)
   } else if (protocol === 'unix:') {
     socket = net.connect(pathname)
   } else {
-    throw new InvalidArgumentError('invalid protocol')
+    socket = net.connect(port || /* istanbul ignore next */ 80, hostname)
   }
 
   client[kSocket] = socket
--- a/lib/client-base.js
+++ b/lib/client-base.js
@@ -76,7 +76,7 @@ class ClientBase extends EventEmitter {
       throw new InvalidArgumentError('invalid protocol')
     }
 
-    if ((url.protocol != 'unix:' && /\/.+/.test(url.pathname)) || url.search || url.hash) {
+    if ((url.protocol !== 'unix:' && /\/.+/.test(url.pathname)) || url.search || url.hash) {
       throw new InvalidArgumentError('invalid url')
     }
 
@@ -439,15 +439,15 @@ function connect (client) {
   const { protocol, port, hostname, pathname } = client[kUrl]
   const servername = client[kServerName] || (client[kTLSOpts] && client[kTLSOpts].servername)
 
-  let socket;
-  if (protocol == 'https:') {
+  let socket
+  if (protocol === 'https:') {
     socket = tls.connect(port || /* istanbul ignore next */ 443, hostname, {
       ...client[kTLSOpts],
       servername
     })
-  } else if (protocol == 'http:') {
+  } else if (protocol === 'http:') {
     socket = net.connect(port || /* istanbul ignore next */ 80, hostname)
-  } else if (protocol == 'unix:') {
+  } else if (protocol === 'unix:') {
     socket = net.connect(pathname)
   } else {
     throw new InvalidArgumentError('invalid protocol')
@@ -776,4 +776,3 @@ function parseHeaders (headers) {
 }
 
 module.exports = ClientBase
-


ronag:
How does a unix:// url look like? Is there a spec? I'm not familiar with it. Maybe some tests would help.

lib/request.js
```diff
@@ -104,7 +104,7 @@ class Request extends AsyncResource {
       }
 
       if (!hostHeader) {
-        header += `host: ${hostname}\r\n`
+        header += `host: ${hostname || 'localhost'}\r\n`
```
ronag, line 107: This should never happen? hostname should always have a value?

ronag: I see `unix:///`

ronag: Is `localhost` correct here?

ronag:
https://github.com/msabramo/requests-unixsocket

ronag:
curl choose to not support this: https://github.com/curl/curl/issues/1338

lib/request.js
yoshuawuyts: I was hitting this case -- believe if no host name is passed this value is initialized as `''`. I didn't validate we were only hitting it through the UDS path, but it seemed likely the TCP path hits this as well.

I only noticed this because my server was returning a 400 complaining `Host` hadn't been set.

ronag:
I'm a little unsure what would make sense here in terms of url scheme... maybe we should do with an options objects like Node `{ socketPath }`.

ronag:
@delvedor @mcollina @kmamal any thoughts?

yoshuawuyts:
> I'm a little unsure what would make sense here in terms of url scheme...

Yeah, I wasn't too sure myself either. I picked `unix://` because it seemed like a reasonable starting point -- but something like `http+unix://` indeed makes more sense to me. It captures it's still `http`, but over a different transport than expected. One could imagine this would be enable extensions such as `http+pipe://` or `https+unix://` to work as well.

---

Regarding escaping: `http+unix:///tmp/foo.sock` is a valid WhatWG URL. [I just ran it through Firefox's URL parser](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=8eb9cdfa966fd063b7e088f4665e13b5) and it seems to parse it as expected:

__code__
```rust
use url::Url;

fn main () {
    let url = Url::parse("http+unix:///tmp/foo.sock").unwrap();
    dbg!(url.scheme());
    dbg!(url.path());
    dbg!(url.domain());
}
```

__output__
```
   Compiling playground v0.0.1 (/playground)
    Finished dev [unoptimized + debuginfo] target(s) in 0.67s
     Running `target/debug/playground`
[src/main.rs:5] url.scheme() = "http+unix"
[src/main.rs:6] url.path() = "/tmp/foo.sock"
[src/main.rs:7] url.domain() = None
```

ronag:
> seems to parse it as expected:

Well, `/tmp/foo.sock` isn't really a path in the http sense of path. So I'm not sure whether it makes sense. It should be the hostname, no?

ronag:
I'm a little surprised `URL` doesn't decode `%2F`. Probably correct.

mcollina:
I'm good with whatever scheme works.

yoshuawuyts:
> Well, /tmp/foo.sock isn't really a path in the http sense of path. 

Yeah, there's definitely a distinction with what a url (or path or uri) is according to [RFC7230](https://tools.ietf.org/html/rfc7230#section-2.7) and according to the [WhatWG URL spec](https://url.spec.whatwg.org/). Over the years we've seen Node.js transition from the RFC7230 interpretation of URLs to the more strict but backwards-compatible WhatWG URLs, culminating in [the deprecation of the older URL APIs](https://nodejs.org/api/url.html#url_legacy_url_api).

I think the right way for us to reason about this is to follow the WhatWG URL spec. That's how both browsers and Node reason about URLs, and I think is the right way for Undici to follow as well. As I've shown in https://github.com/mcollina/undici/pull/226#issuecomment-647559681 `http+unix:///tmp/foo.sock` is parsed by WhatWG URL parsers the way we'd hope it would. And because WhatWG URLs are valid RFC7230 URLs as well, I think it's a safe choice for us to use here.

__tldr:__ The proposed URL schema is correctly parsed by a WhatWG URL parser which means we're good.

ronag:
> is parsed by WhatWG URL parsers the way we'd hope it would

It is? Looks to me like it parses it as pathname instead of hostname? Which is not what I hoped.

i.e. if you want to make a request against a http end-point `/api/hello` the url would look somehting like:

`http+unix:///tmp/foo.sock/api/hello` which would parse everything as a pathname and we have no way to differentiate the host `/tmp/foo.sock` from the path `/api/hello`.

yoshuawuyts:
@ronag Oh I now realize I probably misinterpreted what you said.

```rust
let url = url::Url::parse("http+unix:///tmp/foo.sock/hello").unwrap();
dbg!(url.scheme());
dbg!(url.path());
```
```text
   Compiling playground v0.0.1 (/playground)
    Finished dev [unoptimized + debuginfo] target(s) in 0.63s
     Running `target/debug/playground`
[src/main.rs:5] url.scheme() = "http+unix"
[src/main.rs:6] url.path() = "/tmp/foo.sock/hello"
```

Is what you're saying that you would expect "path" to be `/hello` instead of `/tmp/foo.sock/hello`?

ronag:
maybe we could just keep it simple and not support it through a URL scheme?, i.e.

```js
const client = new Client({
  protocol: 'unix:',
  hostname: '/tmp/foo.sock'
})

```

yoshuawuyts:
@ronag yeah that's not a bad option. One open question tho is what we'd use to populate the `Host` header. E.g. if we do `Host: /tmp/foo.sock` then on the server side this may cause bugs. It's not uncommon for Servers to reconstruct URLs using logic like:

https://github.com/http-rs/async-h1/blob/581540a9b983d8a3fc97456d321b5bcc4eff10b7/src/server/decode.rs#L122

```rust
        Ok(Url::parse(&format!("http://{}/", host))?.join(path)?)
```

This means it'd cause the same problems on the server since URL parsing there may no longer work as expected. Should we consider escaping the hostname before we encode it as the `Host` header? Do we know of a better solution?

ronag:
> escaping the hostname before we encode it as the Host header

I think escaping sounds like a good idea.

ronag:
@yoshuawuyts Do you want to finish this? It's pretty close.

yoshuawuyts:
Hey, yeah I'm planning to continue work on it this week. We need this at work, so I can spend some time on it to see it through to completion (:

szmarczak:
What about going UNIX with TLS?

yoshuawuyts:
> What about going UNIX with TLS?

I would suggest we scope that for a future patch. But populating the `tls` value of the options passed to [`new Client`](https://github.com/mcollina/undici#new-undiciclienturl-opts) would be the obvious approach here.

ronag:
> I suggest we scope that for a future patch.

I think we should consider the API for that. Even if we don't implement it. How would that look? `protocol: 'unix+https:'`?

yoshuawuyts:
@ronag I was thinking:

```js
// http+unix
const client = new Client({
  protocol: 'unix:',
  hostname: '/tmp/foo.sock'
})

// https+unix
const client = new Client({
  protocol: 'unix:',
  hostname: '/tmp/foo.sock',
  tls: tlsOpts
})
```

yoshuawuyts:
Oh but I guess you're asking about the Host url we set? `https+unix://<escaped path>/my/sub/route` is what I was thinking.

ronag:
@ronag I was thinking:

Hm, but that's a bit inconsistent, consider the following:

```js
const client = new Client({
  protocol: 'http:',
  hostname: '/tmp/foo.sock',
  tls: tlsOpts
})
```

Should the above error?

yoshuawuyts:
@ronag what is the current behavior? I would assume it would error even if `hostname` wasn't set to a path.

ronag:
> @ronag what is the current behavior? 

I think tls ignored if it's not relevant for the provided protocol.

ronag:
@szmarczak @delvedor Do you have any good idea? This one is a little tricky.

szmarczak:
In Got we just go UNIX if the hostname equals `unix`. E.g. `https://unix/path/to/socket` is perfectly valid.

yoshuawuyts:
## The TLS option

> I think tls ignored if it's not relevant for the provided protocol.

That seems unexpected -- if a user went through the trouble of initializing TLS config and passing it to the constructer they would likely want to be made of the aware that the connection is in fact not secure. Forgetting to set `https` instead of `http` seems like an easy thing to get wrong.

But that's on current behavior of undici. Regarding this patch we could make it so that `protocol: "unix:"` needs to be passed in order for `hostname` to be able to be processed as a filesystem path.


## The got package

> In Got we just go UNIX if the hostname equals unix. E.g. `https://unix/path/to/socket` is perfectly valid.

In the [`got` UDS docs](https://github.com/sindresorhus/got#unix-domain-sockets) the following example is provided:

```js
const got = require('got');

got('http://unix:/var/run/docker.sock:/containers/json');

// Or without protocol (HTTP by default)
got('unix:/var/run/docker.sock:/containers/json');
```
This looks a bit differently than what was described.

szmarczak:
> This looks a bit differently than what was described.

Can you be more precise? I cannot find a difference.

> My understanding is that this is passed through a custom parser (not WHATWG URL).

Your understanding is wrong. Got doesn't use any custom parsers.

szmarczak:
> splits it into http:// for the protocol, unix: for the transport, /var/run/docker.sock for the host, and /containers/json for the path.

`/var/run/docker.sock` is NOT the host. It's the socket path. The `/containers/json` is a HTTP path. As you can see

https://github.com/sindresorhus/got/blob/7582d916db72f7bab2b2c162d37cfcfde08bbba7/source/core/index.ts#L1379-L1391

The `host` is empty.

Note there is NOT [yet] a spec that defines WHATWG URL schema for UNIX socket paths.

ronag:
`'http://unix:/var/run/docker.sock:/containers/json'` is strange for me. `unix` has nothing to do with the host, it just specifies the transport protocol, while `/var/run/docker.sock` is what I would consider the host in this context, and '/containers/json' as the path.

If we can't agree on the URL version, I would suggest to go with an options version for this use case? And leave the URL parsing to higher level libraries, like got to do in whatever way they deem is best.

The following would make sense to me if we disregard URL parsing and what not:

```js
{
   protocol: 'https:',
   hostname: '/var/run/docker.sock',
   pathname: '/containers/json'
}
```

ronag:
I wouldn't mind even removing the URL parsing from this library just to avoid these ambiguities. @mcollina

ronag:
@yoshuawuyts Given the reaction I think we can all compromise on my solution. Would you mind updating the PR accordingly. Leave the existing URL parsing as is for now.

mcollina:
> I wouldn't mind even removing the URL parsing from this library just to avoid these ambiguities.

That would make things simpler however push the problem down. If we could accept both an object and a string, that would be handy.

ronag:
> That would make things simpler however push the problem down

I don't follow that.

> If we could accept both an object and a string, that would be handy.

That already works. Though for unix sockets it would only work with an object since it's unclear how to encode the socket path in a regular URL.

ronag:
Please se https://github.com/mcollina/undici/pull/238

mcollina:
Removing URL parsing from this library will push the problem down to our users. I'm not sure that's a good idea.

ronag:
closing in favor of the other one.

PR #229, opened 2020-06-26
fix: use node native parser

ronag:
Using llhttp through node binding. llhttp is more maintained, secure and "correct" than http-parser-js.

@dnlup: Thanks for the hints!

### DIFF
--- a/lib/client-base.js
+++ b/lib/client-base.js
@@ -3,7 +3,8 @@
 const { URL } = require('url')
 const net = require('net')
 const tls = require('tls')
-const { HTTPParser } = require('http-parser-js')
+// TODO: This is not really allowed by Node but it works for now.
+const { HTTPParser } = process.binding('http_parser') // eslint-disable-line
 const EventEmitter = require('events')
 const Request = require('./request')
 const assert = require('assert')
@@ -37,7 +38,8 @@ const {
   kParser,
   kSocket,
   kEnqueue,
-  kClient
+  kClient,
+  kMaxHeadersSize
 } = require('./symbols')
 
 const CRLF = Buffer.from('\r\n', 'ascii')
@@ -46,9 +48,12 @@ const TE_CHUNKED_EOF = Buffer.from('\r\n0\r\n\r\n', 'ascii')
 
 function nop () {}
 
+const NODE_MAJOR_VERSION = parseInt(process.version.split('.')[0].slice(1))
+
 class ClientBase extends EventEmitter {
   constructor (url, {
     maxAbortedPayload,
+    maxHeaderSize,
     socketTimeout,
     requestTimeout,
     pipelining,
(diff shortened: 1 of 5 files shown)


ronag:
My benchmarks are all over the place... could someone compare results with and without this PR? @szmarczak @delvedor @dnlup

ronag:
Needs some work to support pre 14 node

dnlup:
A quick bench.

##### Node 12.18.1

**http-parser-js**

Screenshot from 2020-06-26 21-20-34

**native parser**

Screenshot from 2020-06-26 21-17-54

##### Node 12.4.0

**http-parser-js**

Screenshot from 2020-06-26 21-26-53

**native parser***

Screenshot from 2020-06-26 21-29-25

ronag:
slightly faster, and probably more secure

ronag:
@dnlup are you sure you ran on v12? I can only run the test on v14+.

ronag:
This needs help with getting it working on v10 & v12

dnlup:
> @dnlup are you sure you ran on v12? I can only run the test on v14+.

Yes, but I only ran the benchmarks. I am on Linux, maybe this has some influence?

ronag:
test pass! except for v10 which has a different API... we would have to drop v10 support... I'm kind of excited about this (though would have preferred a less hacky way).

mcollina:
Something that would be really nice is to use `llhttp` if it is available and `http-parser-js` if it is not. A flag would also be ok.

ronag:
@mcillina: the way it’s implemented now I think it’s always available? I.e. for old version it will fallback to the old node parser.

lib/client-base.js
```diff
@@ -287,7 +297,19 @@ class ClientBase extends EventEmitter {
...
+    /* istanbul ignore next */
+    if (NODE_MAJOR_VERSION >= 12) {
+      super()
+      this.initialize(
+        HTTPParser.RESPONSE,
+        {},
+        client[kMaxHeadersSize],
+        false,
+        0
+      )
```
mcollina, line 309: In node 12 you can still use the old parser using a command line flag.

ronag: I think that was in order to avoid breaking. I don't think that's relevant here. Also I'm not sure if/how a library can inspect command line options like an internal module.

mcollina: I know, but this will fail if that option is specified :/.

ronag: > I know, but this will fail if that option is specified :/.

Oh, any ideas? try/catch with a retry with option enabled?

dnlup: Can a check be implemented using:

https://nodejs.org/dist/latest-v12.x/docs/api/process.html#process_process_execargv

?

PR #230, opened 2020-06-27
feat: headers and trailers

ronag:
Makes it possible to implement 1xx responses and trailers using the internal `kEnqueue` API (which all the public API methods are based upon). I'm considering making this API public (possibly with a better name).

Basically the internal request callback is a "stream/callbag" which is invoked multiple times:

```
callback({ type: 'headers', statusCode, headers }) // 1xx
callback({ type: 'body', statusCode, headers, body }) // body
callback({ type: 'trailers', headers }) // trailers
callback(null) // eof
```

Missing tests and actually making the API public.

### DIFF
--- a/lib/client-base.js
+++ b/lib/client-base.js
@@ -13,8 +13,7 @@ const {
   RequestAbortedError,
   ClientDestroyedError,
   ClientClosedError,
-  SocketError,
-  NotSupportedError
+  SocketError
 } = require('./errors')
 const {
   kUrl,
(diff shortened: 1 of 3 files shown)


lib/client-base.js
```diff
@@ -310,17 +316,18 @@ class Parser extends HTTPParser {
...
-    }
-
     if (statusCode < 200) {
-      // TODO: Informational response.
+      request.invoke(null, {
+        type: 'headers',
+        statusCode,
+        headers: parseHeaders(headers),
+        opaque
+      })
```
ronag, line 325: What exactly happens with 101 responses?

mcollina:
I don't think invoking a callback multiple times is a good thing to do. We should pass in a `onTrailers` option that is called if there are trailers. Or we can emit a `trailers` event on the body.

ronag:
> This lacks tests, it's a bit hard to review right now.

Yep, just brainstorming at the moment.

> I don't think invoking a callback multiple times is a good thing to do. We should pass in a onTrailers option that is called if there are trailers. Or we can emit a trailers event on the body.

Hm, I'll give this some more thought.

PR #233, opened 2020-06-28
feat: headersTimeout

ronag:
Unfortunately haven't been able to make it actually trigger the timeout on a test.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,6 +69,10 @@ Options:
 - `maxHeaderSize`, the maximum length of request headers in bytes. 
   Default: `16384` (16KiB)
 
+- `headersTimeout`, the amount of time the parser will wait to receive the complete
+  HTTP headers (Node 14 and above only).
+  Default: `30e3` milliseconds (30s).
+
 <a name='request'></a>
 #### `client.request(opts, callback(err, data))`
 
--- a/lib/client-base.js
+++ b/lib/client-base.js
@@ -527,6 +527,9 @@ function connect (client) {
       client.emit('connect')
       resume(client)
     })
+    .on('data', function () {
+      assert(false)
+    })
     .on('error', function (err) {
       if (err.code === 'ERR_TLS_CERT_ALTNAME_INVALID') {
         assert(!client.running)
--- a/lib/client-base.js
+++ b/lib/client-base.js
@@ -339,6 +339,7 @@ class Parser extends HTTPParser {
 
   [HTTPParser.kOnTimeout] () {
     const { socket } = this
+
     socket.destroy(new HeadersTimeoutError())
   }
 
@@ -348,12 +349,14 @@ class Parser extends HTTPParser {
   }
 
   [HTTPParser.kOnExecute] (ret) {
-    if (ret instanceof Error && !this.destroyed) {
+    const { socket } = this
+
+    if (ret instanceof Error && !socket.destroyed) {
       const err = ret
       if (typeof err.reason === 'string') {
         err.message = `Parse Error: ${err.reason}`
       }
-      this.destroy(err)
+      socket.destroy(err)
     }
   }
 
--- a/lib/client-base.js
+++ b/lib/client-base.js
@@ -420,6 +420,10 @@ class Parser extends HTTPParser {
 
     const { client, socket, body, read } = this
 
+    if (socket.destroyed) {
+      return
+    }
+
     const ret = body
       ? body(null, chunk.slice(offset, offset + length))
       : null
@@ -432,10 +436,14 @@ class Parser extends HTTPParser {
   }
 
   [HTTPParser.kOnMessageComplete] () {
-    const { body } = this
+    const { body, socket } = this
 
     this.read = 0
     this.body = null
+
+    if (socket.destroyed) {
+      return
+    }
 
     if (body) {
       body(null, null)
--- a/lib/client-base.js
+++ b/lib/client-base.js
@@ -351,6 +351,9 @@ class Parser extends HTTPParser {
   [HTTPParser.kOnExecute] (ret) {
     const { socket } = this
 
+    // TODO: Node does this. Unclear why.
+    socket._unrefTimer()
+
     if (ret instanceof Error && !socket.destroyed) {
       const err = ret
       if (typeof err.reason === 'string') {
--- a/lib/client-base.js
+++ b/lib/client-base.js
@@ -527,18 +527,6 @@ function connect (client) {
       client.emit('connect')
       resume(client)
     })
-    .on('resume', function () {
-      if (this._handle && !this._handle.reading) {
-        this._handle.reading = true
-        this._handle.readStart()
-      }
-    })
-    .on('pause', function () {
-      if (this._handle && this._handle.reading) {
-        this._handle.reading = false
-        this._handle.readStop()
-      }
-    })
     .on('error', function (err) {
       if (err.code === 'ERR_TLS_CERT_ALTNAME_INVALID') {
         assert(!client.running)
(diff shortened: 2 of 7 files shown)


ronag:
@dnlup Maybe something you'd like to give a try?

dnlup:
Yup, I will!

README.md
```diff
@@ -69,6 +69,10 @@ Options:
 - `maxHeaderSize`, the maximum length of request headers in bytes. 
   Default: `16384` (16KiB)
 
+- `headersTimeout`, the amount of time the parser will wait to receive the complete
+  HTTP headers (Node 12 and above only).
```
ronag, line 73: Node 14

test/headers-timeout.js
```diff
@@ -0,0 +1,32 @@
...
+    t.tearDown(server.close.bind(server))
+
+    server.listen(0, () => {
+      const client = new Client(`http://localhost:${server.address().port}`, {
+        headersTimeout: 1e3
+      })
+      t.tearDown(client.close.bind(client))
+
+      client.request({ path: '/', method: 'GET' }, (err, data) => {
+        t.error(err instanceof errors.HeadersTimeoutError)
```
ronag, line 28: Never runs :(

lib/client-base.js
```diff
@@ -319,6 +327,11 @@ class Parser extends HTTPParser {
     this.body = null
   }
 
+  [HTTPParser.kOnTimeout] () {
```
ronag, line 330: What happens if `kOnTimeout` is undefined, i.e. node < 14?

README.md
```diff
@@ -69,6 +69,10 @@ Options:
...
+  Default: `60e3` milliseconds (30s).
```
ronag, line 74: Maybe change to 30s to be consistent with our other timeouts?

README.md
dnlup: I think it might be because the use the new parser, so the different behavior would still be there when the old-parser option is used?

README.md
dnlup: I agree

README.md
wa-Nadoo: The approach to headersTimeout measurement was changed in Node 13.13 (PR: https://github.com/nodejs/node/pull/32329). I think this is the root cause of the difference between Node 12 and 14.

lib/client-base.js
dnlup: I see from [here](https://github.com/nodejs/node/blob/v12.18.1/lib/_http_common.js#L47) that a default is always assigned to the key.

On 14 the same is done for the `kOnTimeout` (https://github.com/nodejs/node/blob/v14.4.0/lib/_http_common.js#L50). Maybe we could use the same approach here?

test/headers-timeout.js
```diff
@@ -0,0 +1,32 @@
...
+const { Client, errors } = require('..')
+const { createServer } = require('net')
+
+const nodeMajorVersion = parseInt(process.version.split('.')[0].slice(1))
+
+if (nodeMajorVersion >= 14) {
+  test('headers timeout', (t) => {
+    t.plan(1)
+
+    const server = createServer((socket) => {
```
dnlup, line 13: I don't see much difference here from what is implemented  in the Node tests you posted

ronag:
@dnlup Yea, the way headersTimeout is implemented in `node_http_parser` is a bit unfortunate and only works with the `OnStreamRead` API. It's a bit weird... not sure how to resolve. Also I don't quite understand why it's implemented this way. Maybe we should just implement it in JS ourselves instead?

mcollina:
> Also I don't quite understand why it's implemented this way. Maybe we should just implement it in JS ourselves instead?'

Performance.

dnlup:
> @dnlup Yea, the way headersTimeout is implemented in `node_http_parser` is a bit unfortunate and only works with the `OnStreamRead` API. It's a bit weird... not sure how to resolve. Also I don't quite understand why it's implemented this way. Maybe we should just implement it in JS ourselves instead?'
> 
> It's basically a timeout between the first `parser.execute` for a new request and  `kOnHeadersComplete`. It's not entirely trivial. Let me know if you still want to have a try at it or if I should take look.

I am still wrapping my head around this because I am not very familiar with the details. The explanation helps a lot, thanks.
The `kOnTimeout` handler does not get called at all in the test case and I was trying to debug.
 I still want to try, but on the other hand, I don't want to slow you down. Realistically I would need this week to try some things. If you already have something on your mind I understand. Besides performance (potentially) I don't see a problem in implementing it JS. Also, I wanted to see if I could find the reason for the handler not being called.

Unfortunately, my time is a bit limited and doesn't match my will to contribute.

ronag:
> Unfortunately, my time is a bit limited and doesn't match my will to contribute.

No worries. I think I'll take another look at this. Thanks for the help so far!

ronag:
@szmarczak @dnlup Changed to the consume/unconsume API. In theory this should give a performance boost as we skip one c++->js->c++ roundtrip. Would either of you mind doing a benchmark compare against master? (again my computer for some reason provides results all over the place)

ronag:
@szmarczak btw, the `examples/h2o/h2o.conf` file seems to be missing (mentioned in benchmarks/index).

ronag:
@delvedor I think you are the one that setup the CI runs... do you have any tip on how to find what failed in the logs? https://github.com/mcollina/undici/pull/233/checks?check_run_id=823183208 I can't find the failure :/

dnlup:
Node `14.4.0`

image

image

A quick bench, I am sure is not on the latest changes. I'll try to run them later again.

ronag:
Hm, that looks like no difference. INteresting.

ronag:
I think in `readable.pipe` the option is called `end` to not `end()`. Maybe we should have `{ end, destroy }` as options?

ronag:
Test pass 🥂

szmarczak:
> @szmarczak btw, the `examples/h2o/h2o.conf` file seems to be missing (mentioned in benchmarks/index).

Sorry, I meant https://github.com/h2o/h2o/blob/master/examples/h2o/h2o.conf

szmarczak:
> @szmarczak @dnlup Changed to the consume/unconsume API. In theory this should give a performance boost as we skip one c++->js->c++ roundtrip. Would either of you mind doing a benchmark compare against master? (again my computer for some reason provides results all over the place)

I'll do tomorrow. I've been busy IRL, sorry. I've just moved today.

dnlup:
Node `14.4.0`

image

image

ronag:
That looks like a 10% improvement. I guess that's good.

lib/client-base.js
```diff
@@ -329,11 +337,38 @@ class Parser extends HTTPParser {
...
   /* istanbul ignore next: we don't support trailers yet */
   [HTTPParser.kOnHeaders] () {
     // TODO: Handle trailers.
   }
 
+  [HTTPParser.kOnExecute] (ret) {
+    const { socket } = this
+
+    // TODO: Node does this. Unclear why.
+    socket._unrefTimer()
```
ronag, line 355: https://github.com/nodejs/node/commit/ec2822adaad76b126b5cccdeaa1addf2376c9aa6

ronag:
Node does something related which I am not entirely sure about and whether is relevant here:

https://github.com/nodejs/node/blob/master/lib/_http_server.js#L754-L763

ronag:
This PR seems to make the `async hooks client is destroyed` test flaky

ronag:
I'm a little unsure about this direct with using consume/unconsume as it makes things more fragile. Not sure if the extra performance is worth it.

ronag:
Spamming CI jobs a bit to try to confirm whether flakiness is fixed.

ronag:
flakiness seems to have been resolved.

PR #238, opened 2020-07-03
feat: add support for unix socket

ronag:
This is the least controversial solution I could come up with. Basically just do it like Node and curl already does.

For the host header it uses the passed hostname in the url. curl does the same.

### DIFF
--- a/lib/client-base.js
+++ b/lib/client-base.js
@@ -75,6 +75,10 @@ class ClientBase extends EventEmitter {
       throw new InvalidArgumentError('invalid port')
     }
 
+    if (url.socketPath != null && typeof url.socketPath !== 'string') {
+      throw new InvalidArgumentError('invalid socketPath')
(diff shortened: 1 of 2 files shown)


ronag:
@yoshuawuyts @szmarczak

mcollina:
this might require some docs and maybe the tests should be skipped for Windows.

ronag:
@yoshuawuyts would you mind taking a look at the docs?

ronag:
@szmarczak @mcollina I might have jumped the gun here, should this live on the url or the options object?

i.e.

```js
new Client({
  ...url,
  socketPath
}, {
  socketTimeout: 1e3
})
```

vs


```js
new Client(url, {
  socketPath,
  socketTimeout: 1e3
})
```

szmarczak:
I'm voting for the URL approach.

PR #243, opened 2020-07-05
1.1.0

ronag:
- Uses Node native HTTP parser. Improves performance and security. Removes dependency on `http-parser-js`.
- Added `headersTimeout` option.
- Fixed an assertion when using `client.pipeline` and the server does not consume the entire request body.
- Replace `client.full` with `client.busy` property.
- Improved support for legacy streams.
- Added support for IPC (e.g. unix sockets) through `socketPath` option.
- Refactoring and simplification.
- Throw not supported error for `CONNECT` method.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "undici",
-  "version": "1.0.3",
+  "version": "1.1.0",
   "description": "An HTTP/1.1 client, written from scratch for Node.js",
   "main": "index.js",
   "scripts": {


ronag:
Also waiting for decision on https://github.com/mcollina/undici/pull/241 before release.

ronag:
> Replace client.full with client.busy property.

Kind of makes this semver major. WDYT?

ronag:
The switch to Node native parser has been a bit risky and I have found some misc bugs post the initial PR. But it feels pretty good now.

mcollina:
Ok for me!

ronag:
published

PR #250, opened 2020-07-11
fix: reset socket on GET & HEAD with body

ronag:
Fixes: https://github.com/mcollina/undici/issues/246

### DIFF
--- a/lib/client-base.js
+++ b/lib/client-base.js
@@ -460,7 +460,12 @@ class Parser extends HTTPParser {
     if (statusCode >= 200) {
       this.read = 0
       client[kQueue][client[kRunningIdx]++] = null
-      resume(client)
+      if (request.reset) {
+        // TODO: Provide a descriptive error?
+        socket.destroy()
+      } else {
+        resume(client)
+      }
     }
 
     socket.resume()
@@ -711,6 +716,7 @@ function resume (client) {
 
         request.body = null
         request.streaming = false
+        request.reset = false
       }
     }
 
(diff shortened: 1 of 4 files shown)


ronag:
@mcollina How strongly do you feel in regards to "semversiness" on this?

mcollina:
If we were crashing with GET and HEAD with a body before, it's a patch.

ronag:
Added a "notes" section in the docs, documenting this.

ronag:
@szmarczak

szmarczak:
I suggest adding an option to disable this. At Got we had bunch of issues complaining about this behavior.

Otherwise looks good :)

szmarczak:
I agree.

ronag:
> Is this happening only if pipelining is higher than 1?

It happens the connection is re-used. Should probably clarify.

ronag:
An alternative solution is to only send the request when nothing else is running, not send any further requests while it's running, and reconnect once the first response has been received.

ronag:
@delvedor @mcollina @szmarczak PTAL. I think this might be a better solution.

szmarczak:
Just a part of my comment from https://github.com/mcollina/undici/issues/258#issuecomment-658433206

> IMO, if it looks like a proper response, it should be parsed, and if the path doesn't match with the pipelined request (or there isn't any) then ~throw~ reconnect.

What about this?

ronag:
> if the path doesn't match with the pipelined request 

I don't think responses contain the path?

szmarczak:
If the data seems legit (looks like a HTTP/1.1 request), then you can check the path (`url`) here:

https://github.com/mcollina/undici/blob/e1990ce23c51a8b3fa4d4e18570f47467672ee28/lib/client-base.js#L370-L371

And if it doesn't match the pipelined request or there is none just simply reconnect.

ronag:
> If the data seems legit (looks like a HTTP/1.1 request), then you can check the path (url) here:

url is only set for requests, not responses

szmarczak:
Oooooh I was so confused for a while 😂 You're totally right. Pardon me.

PR #254, opened 2020-07-12
fix: fail on content-length mismatch

ronag:
Fixes: https://github.com/mcollina/undici/issues/253

### DIFF
--- a/lib/client-base.js
+++ b/lib/client-base.js
@@ -747,9 +747,12 @@ function resume (client) {
       }
     }
 
-    client[kPendingIdx]++
-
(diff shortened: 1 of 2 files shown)


lib/client-base.js
```diff
@@ -785,7 +788,7 @@ function write (client, request) {
   }
 
   if (request.contentLength !== undefined && request.contentLength !== contentLength) {
-    // TODO: What to do?
+    throw new InvalidArgumentError('invalid content-length header')
```
ronag, line 791: Not sure if this is the most appropriate error type.

PR #260, opened 2020-07-17
pipeline assume GET,HEAD,DELETE no body

ronag:
I'm struggling a little bit with finding the right balance between performance and code complexity when it comes to `pipeline`.

The main problem is that `pipeline` returns a `Duplex` which the user is expected to `.end()` after the request has been scheduled in undici. Depending on how this is implemented we end up in the range of `6000-9000` ops/sec.

The implementation would be simplest **and** fastest if we could do the following assumptions:

If the method is `HEAD`, `GET` or `DELETE` assume that there is no payload for the request and don't initialize the writable side of the returned stream. Also provide a `writable` option to `pipeline` to allow overriding this behavior in special cases e.g. payload with `GET`.

Does this sound like a reasonable compromise?

This also has the advantage of making calling `end()` optional (noop) in the case of these method. It's 
my impression it is a common user error to miss this for these methods.

### DIFF
--- a/lib/client-base.js
+++ b/lib/client-base.js
@@ -26,7 +26,6 @@ const {
   kReset,
   kResume,
   kConnect,
-  kEnd,
   kWriting,
   kQueue,
   kServerName,
@@ -132,8 +131,7 @@ class ClientBase extends EventEmitter {
     this[kRetryDelay] = 0
     this[kRetryTimeout] = null
     this[kOnDestroyed] = []
-    this[kWriting] = null
-    this[kEnd] = null
+    this[kWriting] = false
     this[kMaxAbortedPayload] = maxAbortedPayload || 1048576
 
     // kQueue is built up of 3 sections separated by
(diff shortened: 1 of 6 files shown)


ronag:
@szmarczak @delvedor

lib/client-base.js
```diff
@@ -699,13 +689,9 @@ function resume (client) {
...
 
-    if (
-      client.running &&
-      util.isStream(request.body) &&
-      util.bodyLength(request.body) !== 0
-    ) {
+    if (client.running && util.isStream(request.body)) {
       // Request with stream body can error while other requests
-      // are inflight and indirectly error those as well.
+      // are inflight and indirec:wqtly error those as well.
```
ronag, line 694:
```suggestion
      // are inflight and indirectly error those as well.
```

ronag:
I think I managed to find a sufficiently clean solution while maintain performance. Though I believe making this consideration does open up for further improvement.

I'm closing this as less important at the moment. Sorry to bother.

PR #261, opened 2020-07-19
refactor: split impl into apis

ronag:
This refactors the implementation a bit by moving logic out from `Request` and into the specific API implementations. This will allow future improvements to the API and extend it to support trailers, informational headers, upgrades and connect.

It does introduce some code duplication that did not exist before but I think it's worth it.

An API (`client.request`, `client.pipeline`, `client.stream`) is now expected to extend `Request` and implement the following methods:

```js
class MyApi extends Request {
  onHeaders(statusCode, headers, resume),
  onBody(chunk, offset, length),
  onComplete(trailers),
  onError(err)
}
```

Plan to add `onUpgrade` for websocket upgrades and `CONNECT` later, e.g. for `client.upgrade` and `client.connect` or something...

### DIFF
--- a/lib/client-base.js
+++ b/lib/client-base.js
@@ -666,7 +666,7 @@ function resume (client) {
     }
 
     const request = client[kQueue][client[kPendingIdx]]
-    if (request.finished) {
+    if (request.aborted) {
       // Request was aborted.
       // TODO: Avoid splice one by one.
       client[kQueue].splice(client[kPendingIdx], 1)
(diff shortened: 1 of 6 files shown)


ronag:
This turned out to be a huge performance regression. Not sure why yet.

Just wanted to some feedback on the general approach and then I'll see if I can achieve this without the performance regression.

ronag:
Fixed performance regression.

PR #266, opened 2020-07-24
ci: add benchmarks jobs

### LINKED ISSUE #256: Benchmark CI Job
Would it be possible to add a CI jobs that runs the benchmarks. So it's easier to track any performance regressions?

dnlup:
Add benchmarks jobs in CI workflow. It was taken from https://github.com/mcollina/undici/issues/256#issuecomment-657255734 .

Did you have some customization in mind specific for `undici`?

TODO:

- [x] add CI job
- [x] add npm job

Fix #256

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -1,0 +1,37 @@
+name: Benchmarks
+on:
+  - push
+  - pull_request
+
+jobs:
+  benchmark_current:
+    name: benchmark current
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Code
+        uses: actions/checkout@v2
+        with:
+          ref: ${{ github.base_ref }}
+      - name: Setup Node
+        uses: actions/setup-node@v1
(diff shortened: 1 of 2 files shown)


ronag:
Nice! Though how does this work? I can see 2 more actions but how do one inspect the results?

dnlup:
The results are saved in 2 files, `current.txt` and `branch.txt` which are the benchmarks for the default (stable) branch and the working branch (the one where you're working on, being a PR or anything else), respectively. These files are uploaded as artifacts in the job and you can download them for inspection.

image

dnlup:
We could also just use `stdout` if you prefer.

ronag:
> We could also just use stdout if you prefer.

I think that would be easier.

dnlup:
Ok, so now the results are available in the job stdout. The only thing is that atm the bench on the default branch is failing because it's missing the script.

ronag:
Awesome! Thanks.

PR #270, opened 2020-07-29
Compatibility with IncomingMessage

mcollina:
In one of the previous changes we did lost compatibility with `IncomingMessage`, which is a feature needed to avoid parsing the bodies of requests in proxies.

I think this is the simplest fix. I know it is using `IncomingMessage` directly, but I think it's pretty safe.
We might need a similar fix for http2 compat.


mcollina:
Yes, I'll land and release later on if you don't beat me to it.

PR #271, opened 2020-07-29
Throw if connection header is specified

mcollina:
As titled. Without this patch we had a double `connection` header which was confusing some server.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -99,7 +99,7 @@ class Request extends AsyncResource {
         for (const [key, val] of Object.entries(headers)) {
           if (typeof val === 'object') {
             throw new InvalidArgumentError(`invalid ${key} header`)
-          } else if (val === undefined) {
+          } else if (val === undefined || key === 'connection') {
             continue
           }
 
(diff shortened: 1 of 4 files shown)


ronag:
I guess this is for proxying? Isn't it up to the proxy code to remove hop-by-hop headers?

lib/request.js
```diff
@@ -99,7 +99,7 @@ class Request extends AsyncResource {
         for (const [key, val] of Object.entries(headers)) {
           if (typeof val === 'object') {
             throw new InvalidArgumentError(`invalid ${key} header`)
-          } else if (val === undefined) {
+          } else if (val === undefined || key === 'connection') {
```
ronag, line 102: see how it's done below, I think this is more correct:

```js
if (key.length === 10 && key.toLowerCase() === 'connection') {
  continue
}

```

mcollina:
I'm also ok in throwing! The key point is that the double connection header should not be sent down the line.

ronag:
> I'm also ok in throwing!

I would suggest we handle it the same as transfer-encoding then.

mcollina:
Done

PR #273, opened 2020-07-30
fix: throw on keep-alive header

ronag:
Provide a hint for the server to keep connection alive until it is
considered no longer re-usable by client. This reduces the risk of
a race where the client sends a request and the server closes it
before receiving it.

I don't think Node core uses this... but I guess it still could be useful.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -412,6 +412,7 @@ class Parser extends HTTPParser {
     const request = client[kQueue][client[kRunningIdx]]
 
     // TODO: Check for content-length mismatch?
+    // TODO: keep-alive timeout & max?
 
     assert(this.statusCode < 200)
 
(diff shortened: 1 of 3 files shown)


ronag:
@szmarczak

ronag:
> Is this header understood by servers at all?

I have no idea. I might be interested in implementing it for Node core.

> Maybe behind an option?

Unsure if the perf impact is even measurable. Wouldn't mind to put it behind an option if this is a concern.

ronag:
Removing the send of header, keeping throw, opening issue in Node core to get feedback.

mcollina:
> I have no idea. I might be interested in implementing it for Node core.

From looking at nginx docs, it seems an header that is send from the server to the client.

From http://nginx.org/en/docs/http/ngx_http_core_module.html

> The first parameter sets a timeout during which a keep-alive client connection will stay open on the server side. The zero value disables keep-alive client connections. The optional second parameter sets a value in the “Keep-Alive: timeout=time” response header field. Two parameters may differ.

> The “Keep-Alive: timeout=time” header field is recognized by Mozilla and Konqueror. MSIE closes keep-alive connections by itself in about 60 seconds.

ronag:
@mcollina Thanks! Let's implement it on our side then and adjust idleTimeout accordingly.

szmarczak:
:+1:

PR #275, opened 2020-07-30
add support for keep-alive timeout server hint

ronag:


ronag:
@szmarczak

ronag:
https://github.com/nodejs/node/issues/34560

lib/client.js
```diff
@@ -435,12 +439,23 @@ class Parser extends HTTPParser {
...
 
+    if (headers['keep-alive']) {
+      const m = headers['keep-alive'].match(/timeout=(\d+)/)
+      if (m) {
+        const keepAliveTimeout = parseInt(m[1]) * 1000
+        // Set timeout to half of hint to account for timing inaccuracies.
+        client[kKeepAliveTimeout] = Math.min(keepAliveTimeout / 2, client[kIdleTimeout])
+      }
+    } else {
+      client[kKeepAliveTimeout] = client[kIdleTimeout]
```
szmarczak, line 450: Is this line necessary?

lib/client.js
```diff
@@ -435,12 +439,23 @@ class Parser extends HTTPParser {
       return true
     }
 
+    if (headers['keep-alive']) {
+      const m = headers['keep-alive'].match(/timeout=(\d+)/)
```
szmarczak, line 443: Maybe let's define the regex at the top of the file, so we don't have to generate it every time?

lib/client.js
```diff
@@ -435,12 +439,23 @@ class Parser extends HTTPParser {
...
+      if (m) {
+        const keepAliveTimeout = parseInt(m[1]) * 1000
+        // Set timeout to half of hint to account for timing inaccuracies.
+        client[kKeepAliveTimeout] = Math.min(keepAliveTimeout / 2, client[kIdleTimeout])
```
szmarczak, line 447: `/ 2` seems to be too much. Why not `keepAliveTimeout - 100`? That should be pretty accurate even on slow machines.

lib/client.js
```diff
@@ -435,12 +439,23 @@ class Parser extends HTTPParser {
       return true
     }
 
+    if (headers['keep-alive']) {
+      const m = headers['keep-alive'].match(/timeout=(\d+)/)
+      if (m) {
+        const keepAliveTimeout = parseInt(m[1]) * 1000
```
szmarczak, line 445: What about `Number(m[1])`?

lib/client.js
ronag: > so we don't have to generate it every time?

I think v8 optimizes it out as a constant?

lib/client.js
ronag: I think so. If the server does not provide a hint for subsequent requests I think we should fallback to the idle timeout option.

lib/client.js
szmarczak: I have no deep idea about V8 optimizations.

The only thing I know is just that lambdas should be faster :P

lib/client.js
szmarczak: But you already set this at line 137, no?

lib/client.js
ronag: lambdas?

szmarczak: In general. I don't know anything 'bout V8 optimizes variables.

lib/client.js
ronag: Not sure I follow. Line 447 will override line 137.

szmarczak: I mean line 450 is the same as 137.

ronag: no, 137 is in the constructor, the value might have been changed due to a response, however if a following response does not have a keep-alive hint it should go back.

szmarczak: Ah, I get it now :D

PR #279, opened 2020-07-30
fix: set default idle timeout to 4 seconds

ronag:
Node HTTP server seem to have a default of 5 seconds and
we should be below that.

Fixes: https://github.com/mcollina/undici/issues/276

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -133,7 +133,7 @@ class Client extends EventEmitter {
     this[kUrl] = url
     this[kSocketPath] = socketPath
     this[kSocketTimeout] = socketTimeout == null ? 30e3 : socketTimeout
-    this[kIdleTimeout] = idleTimeout == null ? 30e3 : idleTimeout
+    this[kIdleTimeout] = idleTimeout == null ? 4e3 : idleTimeout
     this[kKeepAliveTimeout] = this[kIdleTimeout]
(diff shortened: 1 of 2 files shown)


ronag:
@szmarczak

lib/client.js
```diff
@@ -442,9 +442,12 @@ class Parser extends HTTPParser {
     if (headers['keep-alive']) {
       const m = headers['keep-alive'].match(/timeout=(\d+)/)
       if (m) {
-        const keepAliveTimeout = Number(m[1]) * 1000
-        // Set timeout to half of hint to account for timing inaccuracies.
-        client[kKeepAliveTimeout] = Math.min(keepAliveTimeout - 500, client[kIdleTimeout])
+        let keepAliveTimeout = Number(m[1]) * 1000
+        keepAliveTimeout = keepAliveTimeout > 2000 ? keepAliveTimeout - 1000 : keepAliveTimeout / 2
```
ronag, line 446: We do 1s as to take into account bad rounding from ms to s on server side.

ronag:
> lgtm. I'm not 100% conviced about this, as keeping it alive for a long time is typically worthwhile.

Not convinced either... but not sure what the option is... both Node and Cloudfront seem to have quite short keep alive times.

mcollina:
Let's go for it

PR #282, opened 2020-07-31
fix: pass socketPath in pool

### LINKED ISSUE #281: `socketPath` not being passed into `Client` from `Pool`
As in the [unix domain sockets test](https://github.com/mcollina/undici/blob/25ce5c740cc4845ccaf392af72ee2b79d429a77b/test/unix.js#L23), the `Client` gets passed in the `socketPath` like so:
```ts
const client = new Client({
  hostname: 'localhost',
  protocol: 'http:'
}, {
  socketPath: '/var/tmp/test3.sock'
})
```

However, in the `Pool` implementation, that arg is not passed in:
https://github.com/mcollina/undici/blob/master/lib/pool.js#L27

It looks like this would be a small patch, making sure, that the `socketPath` argument is being passed from `Pool` down to `Client`.

timsuchanek:
Closes #281

### DIFF
--- a/lib/pool.js
+++ b/lib/pool.js
@@ -16,7 +16,8 @@ class Pool {
     socketTimeout,
     requestTimeout,
     pipelining,
-    tls
+    tls,
+    socketPath
   } = {}) {
     if (connections != null && (!Number.isFinite(connections) || connections <= 0)) {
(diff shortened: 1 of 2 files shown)


ronag:
Thanks!

PR #285, opened 2020-07-31
feat: upgrade support

ronag:
```js
client.upgrade({
  path: '/',
  method: 'GET',
  protocol: 'Websocket'
}, (err, data) => {
  const { headers, socket } = data
})
```

```js
client.connect({
  path: '/'
}, (err, data) => {
  const { statusCode, headers, socket } = data
})
```

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -430,7 +430,7 @@ class Parser extends HTTPParser {
 
     assert(Number.isFinite(ret))
 
-    if (this.upgrade) {
+    if (this.upgrade && !this.socket.destroyed) {
       const { socket, client, headers, statusCode } = this
       const request = client[kQueue][client[kRunningIdx]]
 
--- a/lib/client.js
+++ b/lib/client.js
@@ -474,16 +474,12 @@ class Parser extends HTTPParser {
     this.headers = util.parseHeaders(rawHeaders, this.headers)
     this.statusCode = statusCode
 
-    if (upgrade) {
-      if (!request.upgrade && request.method !== 'CONNECT') {
-        util.destroy(socket, new SocketError('unexpected upgrade'))
-      } else {
-        this.upgrade = true
-      }
-      return 2
+    if (upgrade !== Boolean(request.upgrade)) {
+      util.destroy(socket, new SocketError('bad upgrade'))
+      return 1
     }
 
-    if (request.method === 'CONNECT') {
+    if (upgrade || request.method === 'CONNECT') {
       this.upgrade = true
       return 2
     }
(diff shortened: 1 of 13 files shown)


ronag:
@mcollina Is this something we want to have? Feedback on API?

ronag:
@szmarczak @delvedor

ronag:
Was thinking to do the same for connect and/or include connect here as well.

mcollina:
Or maybe it should throw. What about calling request if the connection has been upgraded?

ronag:
> What happens if there are request in flight? Or if there are a few pipelined? I don't think upgrade should throw.

No more requests should be dispatched after an upgrade until it has completed. Not implemented in this PR yet.

> I don't think upgrade should throw.

Not following. We already throw on unexpected upgrade?

ronag:
Any feedback on the `protocol` argument? Seems ok?

ronag:
> What about calling request if the connection has been upgraded?

What do you mean?

lib/client.js
```diff
@@ -437,7 +446,26 @@ class Parser extends HTTPParser {
...
+      detachSocket(socket)
+
+      this.unconsume()
+      setImmediate(() => this.close())
+
+      this.socket = null
+
+      // this.currentBufferOffset is updated in nextTick
+      process.nextTick(() => {
+        const head = this.getCurrentBuffer().slice(this.currentBufferOffset)
```
ronag, line 458: `socket.unshift()` should also work?

lib/client.js
```diff
@@ -427,20 +441,36 @@ class Parser extends HTTPParser {
...
-      util.destroy(socket, new NotSupportedError('upgrade not supported'))
+      this.upgrade = true
+
+      detachSocket(socket)
+
+      this.unconsume()
+      setImmediate(() => this.close())
+
+      // this.currentBufferOffset is updated in nextTick
+      process.nextTick(() => {
```
ronag, line 459: I don't like this nextTick  but can't think of a better way...

ronag:
I think this is somewhat ready. Lacking some tests.

ronag:
Notice a significant improvement over the node core API is that the `head` argument is not necessary.

mcollina:
I mean what happens if we start an upgrade to websocket or http2 and there are request in the pipeline.

ronag:
> I mean what happens if we start an upgrade to websocket or http2 and there are request in the pipeline.

Before or after the request?

If it's before, I would think the server doesn't start processing the upgrade until all other requests has completed. Though I'm not sure. Maybe we should change so that an upgrade is not dispatched while there are other requests at head of line.

If it's after they will stay queued until the upgrade is completed and a new socket has been assigned.

PR #287, opened 2020-08-01
feat: onInfo & onTrailers options

ronag:


ronag:
@szmarczak

ronag:
Will add tests after API feedback.

ronag:
I'm not sure `onInfo` makes sense for anything else but https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/103. Rename to `onHints`?

README.md
```diff
@@ -102,6 +102,9 @@ Options:
   If `false` the request won't be sent until all preceeding
   requests in the pipeline has completed.
   Default: `true` if `method` is `HEAD` or `GET`.
+* `onInfo(statusCode, headers)`, function invoked for informational
+  1xx responses. `100 Continue` is ignored and will not invoke `onInfo`.
```
ronag, line 106: I wonder if we only should support known 1xx responses with defined semantics?

i.e.

101
102
103

e.g. 100 is actually not useful for the user (I think?) and is something that would have to be implemented internally.

README.md
```diff
@@ -220,13 +223,17 @@ Destroying the request or response body will have the same effect.
...
+This method expects `factory` to return a 
+[`Writable`](https://nodejs.org/api/stream.html#stream_class_stream_writable) 
+which the response will be written to. This improves performance by avoiding
+an intermediate 
+[`Readable`](https://nodejs.org/api/stream.html#stream_readable_streams) 
+when the user expects to directly pipe the response body to a
 [`Writable`](https://nodejs.org/api/stream.html#stream_class_stream_writable).
 
+`factory` may return `null` in which case the response body is skipped and no
+trailers will be consumed.
```
ronag, line 235: Note that not trailers if return `null`.

ronag:
There is also the option of having `client.request` return a `trailer: Promise`?

ronag:
@szmarczak

README.md
szmarczak: IMO let's go for all 1XX. Custom informational status codes may be implemented e.g. by CloudFlare in the future like they did with 5XX.

szmarczak:
> There is also the option of having client.request return a trailers: Promise?

Sounds great!

ronag:
> There is also the option of having client.request return a trailers: Promise?

@mcollina WDYT? There will be an extra overhead of always creating the trailer promise.

mcollina:
> @mcollina WDYT? There will be an extra overhead of always creating the trailer promise.

That could work. We need to be careful of resolving it anyway even if we don't get trailers.

Generating this could be behind an option if it's too much overhead.

ronag:
I don’t like it. Feels weird when we have a callback api returning a promise. I’m sure it doesn’t snob just doesn’t feel elegant.

PR #300, opened 2020-08-04
feat: dispatch

ronag:
Makes our internal API public. This is the API which our public API is implemented on top.

There is need for some internal cleanup and refactor post this PR. I've tried to keep the changes as small as possible for easy of review.

I am not 100% satisfied with the API. However, I believe it does fullfil functional and performance requirements. Would appreciate some feedback and maybe ideas. I suspect this API might evolve/improve over time and be less stable than the other ones.

Some problems that might or might not need to be resolved:

- ~~There is now two `AsyncResource`s, the `handler` object and `Request` object. Depending on whether the implementation needs to asynchronously propagate async scope.~~

- ~~It's a bit weird that `_onConnect` takes a `resume` function.~~
- ~~Not sure about naming for `_onConnect`.~~
- It's a bit weird that `_onData` should return `null` to indicate an aborted message.

Refs: https://github.com/mcollina/undici/issues/293

```js
_onUpgrade(statusCode, headers, socker): Void
_onHeaders(statusCode, headers, resume): Void
_onData(chunk): Boolean|Null
_onComplete(trailers): Void
_onErrror(err): Void
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -23,10 +23,11 @@ Machine: 2.8GHz AMD EPYC 7402P<br/>
 Configuration: Node v14.4, HTTP/1.1 without TLS, 100 connections, Linux 5.4.12-1-lts
 
 ```
-http - keepalive x 5,521 ops/sec ±3.37% (73 runs sampled)
-undici - pipeline x 9,292 ops/sec ±4.28% (79 runs sampled)
-undici - request x 11,949 ops/sec ±0.99% (85 runs sampled)
-undici - stream x 12,223 ops/sec ±0.76% (85 runs sampled)
+http - keepalive x 5,826 ops/sec ±1.45% (275 runs sampled)
+undici - pipeline x 7,281 ops/sec ±1.68% (273 runs sampled)
+undici - request x 11,700 ops/sec ±0.56% (278 runs sampled)
+undici - stream x 12,684 ops/sec ±0.72% (280 runs sampled)
+undici - dispatch x 13,446 ops/sec ±0.37% (276 runs sampled)
 ```
 
 The benchmark is a simple `hello world` [example](benchmarks/index.js).
@@ -451,6 +452,34 @@ there might still be some progress on dispatched requests.
 
 Returns a promise if no callback is provided.
 
+
+<a name='dispatch'></a>
+#### `client.dispatch(opts, handler): Promise|Void`
+
+This is the low level API which all the preceeding API's are implemented on top of.
+
+Options:
+
+* ... same as [`client.request(opts[, callback])`][request].
+
+The `handler` parameter is defined as follow:
+
+* `onUpgrade(statusCode, headers, socket): Void`, invoked when request is upgraded  either due to a `Upgrade` header or `CONNECT` method.
+  * `statusCode: Number`
+  * `headers: Object`
+  * `socket: Duplex`
+* `onHeaders(statusCode, headers, resume): Void`, invoked when statusCode and headers have been received. 
+  May be invoked multiple times due to 1xx informational headers.
+  * `statusCode: Number`
+  * `headers: Object`
+  * `resume(): Void`, resume `onData` after returning `false`.
+* `onData(chunk): Null|Boolean`, invoked when response payload data is received.
+  * `chunk: Buffer`
+* `onComplete(trailers): Void`, invoked when response payload and trailers have been received and the request has completed.
+  * `trailers: Object`
+* `onError(err): Void`, invoked when an error has occured.
+  * `err: Error`
+
 #### `client.pipelining: Number`
 
 Property to get and set the pipelining factor.
@@ -528,6 +557,10 @@ Calls [`client.upgrade(opts, callback)`][upgrade] on one of the clients.
 
 Calls [`client.connect(opts, callback)`][connect] on one of the clients.
 
+#### `pool.dispatch(opts, handler): Void`
+
+Calls [`client.dispatch(opts, handler)`][dispatch] on one of the clients.
+
 #### `pool.close([callback]): Promise|Void`
 
 Calls [`client.close(callback)`](#close) on all the clients.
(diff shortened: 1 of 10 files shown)


ronag:
@szmarczak

lib/client-pipeline.js
```diff
@@ -10,17 +10,16 @@ const {
...
 const util = require('./util')
-const { kEnqueue } = require('./symbols')
+const { AsyncResource } = require('async_hooks')
 
 // TODO: Refactor
 
 const kResume = Symbol('resume')
 
-class PipelineRequest extends Request {
+class PipelineRequest extends AsyncResource {
```
ronag, line 20: Needs to propagate async scope by overriding `body.destroy`

lib/client-request.js
```diff
@@ -32,8 +31,10 @@ class RequestResponse extends Readable {
   }
 }
 
-class RequestRequest extends Request {
+class RequestRequest extends AsyncResource {
```
ronag, line 34: Needs to propagate async scope by overriding `body.destroy`

lib/client-stream.js
```diff
@@ -1,16 +1,17 @@
...
 const {
   InvalidArgumentError,
   InvalidReturnValueError
 } = require('./errors')
 const util = require('./util')
-const { kEnqueue } = require('./symbols')
+const { AsyncResource } = require('async_hooks')
 
-class StreamRequest extends Request {
+class StreamRequest extends AsyncResource {
```
ronag, line 11: Needs to propagate async scope by inside `body._destroy`

README.md
```diff
@@ -23,10 +23,11 @@ Machine: 2.8GHz AMD EPYC 7402P<br/>
 Configuration: Node v14.4, HTTP/1.1 without TLS, 100 connections, Linux 5.4.12-1-lts
 
 ```
-http - keepalive x 5,521 ops/sec ±3.37% (73 runs sampled)
-undici - pipeline x 9,292 ops/sec ±4.28% (79 runs sampled)
-undici - request x 11,949 ops/sec ±0.99% (85 runs sampled)
-undici - stream x 12,223 ops/sec ±0.76% (85 runs sampled)
+http - keepalive x 5,392 ops/sec ±6.66% (68 runs sampled)
+undici - pipeline x 7,055 ops/sec ±3.54% (79 runs sampled)
```
ronag, line 27: pipeline performance regression is from a previous commit. Haven't bisected yet.

szmarczak:
The new API looks good IMO.

Instead of returning `true` / `false` / `null` I think it would be clearer to return a constant e.g. `undici.RESUME` `undici.PAUSE` `undici.ABORT`. Just an idea.

ronag:
> Just an idea.

That's kind of nice. Less magical.

ronag:
Another alternative is something like:

```js
_onConnect(controller)
```

Where controller has the following methods:

```
pause()
resume()
abort()
```

README.md
```diff
@@ -23,10 +23,11 @@ Machine: 2.8GHz AMD EPYC 7402P<br/>
 Configuration: Node v14.4, HTTP/1.1 without TLS, 100 connections, Linux 5.4.12-1-lts
 
 ```
-http - keepalive x 5,521 ops/sec ±3.37% (73 runs sampled)
-undici - pipeline x 9,292 ops/sec ±4.28% (79 runs sampled)
-undici - request x 11,949 ops/sec ±0.99% (85 runs sampled)
-undici - stream x 12,223 ops/sec ±0.76% (85 runs sampled)
+http - keepalive x 5,595 ops/sec ±3.33% (69 runs sampled)
+undici - pipeline x 7,055 ops/sec ±3.54% (79 runs sampled)
+undici - request x 11,551 ops/sec ±1.67% (81 runs sampled)
```
mcollina, line 28: It seems request is also regressing a bit?

lib/client-connect.js
```diff
@@ -66,7 +62,10 @@ module.exports = function connect (client, opts, callback) {
   }
 
   try {
-    client[kEnqueue](new ConnectRequest(client, opts, callback))
+    client.dispatch({
+      ...opts,
```
mcollina, line 66: Can you please avoid `...`? It's performance has got slower in recent Node.js.

lib/client-upgrade.js
```diff
@@ -65,7 +57,11 @@ module.exports = function upgrade (client, opts, callback) {
   }
 
   try {
-    client[kEnqueue](new UpgradeRequest(client, opts, callback))
+    client.dispatch({
+      ...opts,
```
mcollina, line 61: avoid `...`

README.md
ronag: I will dig into this.

lib/client-connect.js
ronag: Is object assign better? Also do you have a reference for the perf regression? Not important. Just curious.

ronag: Fixed

lib/client-upgrade.js
ronag: Fixed

README.md
mcollina: any updates?

ronag: not yet, benchmarks has a lot of variance between runs so it's a bit tricky. Still working on it.

ronag:
@mcollina I think perf diff is within margin of error now...

README.md
ronag: I think this is fixed sufficiently now.

ronag:
master:
```js
undici - request x 11,675 ops/sec ±0.86% (278 runs sampled)
undici - stream x 12,722 ops/sec ±0.73% (279 runs sampled)
```

dispatch2:
```js
undici - request x 11,709 ops/sec ±1.21% (86 runs sampled)
undici - stream x 12,642 ops/sec ±1.20% (81 runs sampled)
undici - dispatch x 13,277 ops/sec ±0.86% (84 runs sampled)
```

PR #301, opened 2020-08-06
don't propagate async scope across events

ronag:
Don't propagate async scope across events.

Pending recommendation/consensus in https://github.com/nodejs/node/issues/34430

Refs: https://github.com/nodejs/node/issues/34430

Small perf boost:

```
undici - stream x 12,681 ops/sec ±1.95% (83 runs sampled)
```

vs

```
undici - stream x 12,226 ops/sec ±2.12% (80 runs sampled)
```

### DIFF
--- a/lib/client-pipeline.js
+++ b/lib/client-pipeline.js
@@ -11,16 +11,13 @@ const {
   RequestAbortedError
 } = require('./errors')
 const util = require('./util')
-const { AsyncResource } = require('async_hooks')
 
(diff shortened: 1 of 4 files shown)


ronag:
@mcollina Do you have any opinion on this or should we wait and see how https://github.com/nodejs/node/issues/34430 turns out?

mcollina:
I think this is the right path tbh

PR #308, opened 2020-08-08
fix: pipeline res ret destroy

ronag:
investigating ubuntu failure

### DIFF
--- a/test/client-pipeline.js
+++ b/test/client-pipeline.js
@@ -525,7 +525,7 @@ test('pipeline abort piped res', (t) => {
         if (process.versions.node.split('.')[0] < 13) {
           t.ok(err)
         } else {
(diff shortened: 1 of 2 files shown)


test/client-pipeline.js
```diff
@@ -525,7 +525,7 @@ test('pipeline abort piped res', (t) => {
         if (process.versions.node.split('.')[0] < 13) {
           t.ok(err)
         } else {
-          t.strictEqual(err.code, 'ERR_STREAM_PREMATURE_CLOSE')
+          t.strictEqual(err.code, 'UND_ERR_ABORTED')
```
szmarczak, line 528: Why not `UNDICI_ERR_ABORTED`?

ronag: it should be UNDICI_ERR_ABORTED, that was the problem. Fixed.

PR #314, opened 2020-08-12
perf: improved pool balancing

ronag:
If all clients are busy then put requests in a queue and dispatch on first client that is no longer busy.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -54,11 +54,11 @@ const {
   kMaxKeepAliveTimeout,
   kKeepAliveTimeoutThreshold
 } = require('./symbols')
-const makeStream = require('./client-stream')
-const makeRequest = require('./client-request')
-const makePipeline = require('./client-pipeline')
-const makeUpgrade = require('./client-upgrade')
-const makeConnect = require('./client-connect')
+const makeStream = require('./client-stream').stream
+const makeRequest = require('./client-request').request
+const makePipeline = require('./client-pipeline').pipeline
+const makeUpgrade = require('./client-upgrade').upgrade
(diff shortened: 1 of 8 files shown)


lib/client.js
```diff
@@ -794,6 +798,10 @@ function resume (client) {
     }
 
     if (!client.pending) {
+      if (client.connected && !client[kWriting] && !client[kReset]) {
+        // TODO: Only emit if was busy.
+        client.emit('drain')
```
ronag, line 803: I'm a little unsure about the naming here... Also it might be emitted multiple times, though that is probably mostly harmless.

mcollina: drain is fine!

lib/pool.js
```diff
@@ -17,49 +21,104 @@ class Pool {
...
     }
 
+    this[kQueue] = []
+    this[kPendingIdx] = 0
     this[kClients] = Array.from({
       length: connections || 10
     }, () => new Client(url, options))
+
+    for (const client of this[kClients]) {
+      client.on('drain', () => {
```
mcollina, line 31: can you please create this closure in a separate function? closures and loops are.. weird.

ronag: it needs to capture the client instance though?

mcollina: you don't need it, it is `this` in the function (if you remove the arrow).

mcollina: Yes, something like `client.on('drain', onDrain(this))`

ronag: How about now?

ronag:
> It would be good to have a test to cover this.

I will revisit this later and write more tests.

PR #316, opened 2020-08-13
fix: change set-cookie header to always be an array

belochub:
For compatibility with the Node.js API `set-cookie` header, if present,
has to always be an array, even when there is only one value in it.

For reference, see https://nodejs.org/api/http.html#http_message_headers

There is also a possible compatibility problem with the way duplicate headers are handled right now since you always collect all duplicates to an array, but Node.js discards duplicates of some of the headers and joins the duplicate values with ', ' or '; ' for others. This can be fixed separately if you want full compatibility with Node.js, but for now, I only wanted to handle one issue that I discovered while using this package.

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -59,7 +59,7 @@ function parseHeaders (headers, obj) {
     var key = headers[i].toLowerCase()
     var val = obj[key]
     if (!val) {
-      obj[key] = headers[i + 1]
+      obj[key] = key === 'set-cookie' ? [headers[i + 1]] : headers[i + 1]
     } else {
       if (!Array.isArray(val)) {
         val = [val]
--- a/test/client.js
+++ b/test/client.js
@@ -570,6 +570,35 @@ test('Set-Cookie', (t) => {
   })
 })
 
+test('Set-Cookie with a single cookie', (t) => {
+  t.plan(4)
+
+  const server = createServer((req, res) => {
+    res.setHeader('content-type', 'text/plain')
+    res.setHeader('Set-Cookie', 'a single cookie')
+    res.end('hello')
+  })
+  t.tearDown(server.close.bind(server))
+
+  server.listen(0, () => {
+    const client = new Client(`http://localhost:${server.address().port}`)
+    t.tearDown(client.close.bind(client))
+
+    client.request({ path: '/', method: 'GET' }, (err, { statusCode, headers, body }) => {
+      t.error(err)
+      t.strictEqual(statusCode, 200)
+      t.strictDeepEqual(headers['set-cookie'], ['a single cookie'])
+      const bufs = []
+      body.on('data', (buf) => {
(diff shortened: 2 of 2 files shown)


ronag:
I think the original thought here was that undici considers as little user land edge cases as possible and assumes what is provided is correct (or at least correct in terms of the protocol) and leave these kind of compat logic to higher level libraries.

That being said I'm not opposed to including these kind of things. @mcollina WDYT?

ronag:
@delvedor @szmarczak

ronag:
I'm having some second thoughts on this. It feels a bit like a rabbit hole. If we do this should we implement all the other weird Node behaviors as well? Where do we stop?

Would like to think about this a bit more. Not sure adding edge cases is a good idea.

ronag:
Anyone disagreeing to my objection or can we close this?

mcollina:
no, not really

belochub:
@mcollina @ronag I would like to clarify, to explain my idea for this PR.
I discovered the problem when using `fastify-reply-from` with `unidici` enabled in TypeScript. When passing `rewriteHeaders()` option, the way it is [typed](https://github.com/fastify/fastify-reply-from/blob/0d760697ac7158329d17ca6f79391ef5fcee1a57/index.d.ts#L43-L45) right now, that function accepts Node.js [`http.IncomingHttpHeaders`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/340147c67949c525323bd5e51d5d8eda1256e65b/types/node/http.d.ts#L7) type, which specifies that the value for `set-cookie` header is an array of strings, but as of right now, when `undici` is enabled in `fastify-reply-from`, that makes the provided types incorrect due to this and other possible differences related to header collection this package has from the Node.js API.
My first thought was to open an issue in the `fastify-reply-from` repository, but after looking into the details, I noticed that the issue was actually in the way this package is handling headers, and seeing that according to https://github.com/mcollina/undici/issues/312 you want this to possibly become a replacement for `http.request`, I thought that maybe it would make sense to notify you of this issue, and maybe fix this one specific difference that stands out the most. But I now see that I misunderstood the idea and that this package does not have to have the same API and behavior as Node.js API has.
I guess, the problem I discovered should be fixed in `fastify-reply-from` instead, but in my opinion, the current behavior when this header can be either a single string or an array of strings, depending on the number of times it is present in the response, is still less convenient than always having it presented as an array.

ronag:
@belochub Thanks for the description. I think this should be resolved in `fastify-reply-from`.

> is still less convenient than always having it presented as an array.

Depends... for most headers it's more convenient to have it as a string.

belochub:
>Depends... for most headers it's more convenient to have it as a string.

@ronag I agree, that for most headers this is better, thus the current behavior of this package, combining duplicate headers into an array instead of a string, might be an inconvenience for some users.

PR #319, opened 2020-08-13
fix: buffer check for jest

### LINKED ISSUE #318: Doesn't work within jest
When running an undici client in jest, jest is doing some kind of override magic in the background, that makes [this check for a buffer in undici](https://github.com/mcollina/undici/blob/master/lib/client.js#L946) fail.

Reproduction: https://github.com/timsuchanek/undici-buffer-repro

I'm already working on a fix :) I suggest adding a ` || Buffer.isBuffer(body)` in https://github.com/mcollina/undici/blob/master/lib/client.js#L946

timsuchanek:
Closes #318

### DIFF
--- a/lib/util.js
+++ b/lib/util.js
@@ -71,11 +71,17 @@ function parseHeaders (headers, obj) {
   return obj
 }
 
+function isBuffer (buffer) {
+  // See, https://github.com/mcollina/undici/pull/319
+  return buffer instanceof Uint8Array || Buffer.isBuffer(buffer)
+}
+
 module.exports = {
   nop,
   isStream,
   isDestroyed,
   parseHeaders,
   destroy,
-  bodyLength
+  bodyLength,
+  isBuffer
 }
--- a/test/jest/test.js
+++ b/test/jest/test.js
@@ -13,7 +13,7 @@ test('should work in jest', async () => {
     res.end('hello')
   })
   await new Promise((resolve, reject) => {
-    server.listen(0, async () => {
+    server.listen(0, () => {
       const client = new Client(`http://localhost:${server.address().port}`)
       client.request({
         path: '/',
@@ -22,7 +22,7 @@ test('should work in jest', async () => {
           'Content-Type': 'application/json'
         },
         body: '{}'
-      }, async (err, result) => {
+      }, (err, result) => {
         server.close()
         client.close()
         if (err) {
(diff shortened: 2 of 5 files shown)


ronag:
Could you add a `isBuffer` in `util` instead?

```js
function isBuffer (buffer) {
  return buffer instanceof Uint8Array || buffer instanceof Buffer
}
```

and replace all occurrences of `instanceof Uint8Array` with it?

timsuchanek:
Yep will do

timsuchanek:
Done

test/jest/test.js
```diff
@@ -0,0 +1,36 @@
...
+test('should work in jest', async () => {
+  const server = createServer((req, res) => {
+    expect(req.url).toBe('/')
+    expect(req.method).toBe('POST')
+    expect(req.headers.host).toBe('localhost')
+    res.setHeader('Content-Type', 'text/plain')
+    res.end('hello')
+  })
+  await new Promise((resolve, reject) => {
+    server.listen(0, async () => {
```
mcollina, line 16: please just wrap server listen in a Promise, and do not use async functions inside new Promise.

lib/util.js
```diff
@@ -71,11 +71,17 @@ function parseHeaders (headers, obj) {
   return obj
 }
 
+function isBuffer (buffer) {
+  // See, https://github.com/mcollina/undici/pull/319
+  return buffer instanceof Uint8Array || Buffer.isBuffer(buffer)
```
mcollina, line 76: I'm totally lost on why Buffer.isBuffer would work, but hey.

ronag: Related? https://github.com/facebook/jest/issues/9983

ronag: https://github.com/facebook/jest/issues/9993

ronag: https://github.com/facebook/jest/blob/4bd3d4a05999170f423f7050d4e0537648499e88/packages/jest-environment-jsdom/src/index.ts#L49

timsuchanek: I'm also not sure, as the `isBuffer` implementation uses an `instanceof` as well: https://github.com/nodejs/node/blob/7a447bcd54f5e153d14da377963541349d7bda09/lib/buffer.js#L504
But yeah the new test fails without the fix and works with the fix.

jimmywarting: Shouldn't it be enough to do instanceof Uint8Array?
Buffer is a instanceof Uint8Array after all...

PR #327, opened 2020-08-15
v1.3.2: Improved Pool Balancing

ronag:
- perf: improved pool balancing (#314)

`Pool` will now wait for first non busy `Client` before dispatching a requests. This fixes the case of starvation where a client might have no more work to do while other clients have full queues. 

- fix: async scope upgrade & connect

`upgrade` and `connect` will now properly propagate async context to `callback`.

- fix: naming
- refactor: removed unnecessary kGetNext
- refactor: move checks into constructor
- refactor: pool doesn't need to create promise
- refactor: defaultHostname => hostname
- removed some assertions
- fix: busy if pending and not connected
- refactor: add reset prop
- refactor: make reset computed
- add assertion
- fix: use length instead of byteLength
- refactor: socket write & cork
- refactor: method => fn
- improve test coverage
- fix: Pool deadlock
- 100 coverage
- refactor: simplify
- move upgrade check earlier
- write stream headers later
- fix: reduce number of emit 'drain'
- prefer ! over === 0
- refactor: inline read(0) with comment
- test: fix flaky test
- fix: check bytesWritten on body.'end'
- refactor: move timeout logic to resume loop
- refactor: inline CRLF
- don't requeue aborted requests
- perf: Number.isFinite is faster than instanceof
- fix: buffer check for jest (#319)

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "undici",
-  "version": "1.3.1",
+  "version": "1.3.2",
   "description": "An HTTP/1.1 client, written from scratch for Node.js",
   "main": "index.js",
   "scripts": {


ronag:
Not sure about the release process but I'll later close this PR and add the above description into the GH release notes.

ronag:
FYI: I've removed from the description commits related to changes that were later reverted.

ronag:
Should this be a patch or minor?

Jolg42:
Hello there!

@mcollina @ronag 

I'm curious to know if you were planning to release 1.3.2 on npm? (maybe forgot 😄)

I can see this PR bumps the release but `npm view undici` says latest is `1.3.1`

Thanks 🙇‍♂️

ronag:
@Jolg42 found some other stuff so right now it looks like we'll have a 2.0 in a few days. Would like to get coverage to 100 and consider whether there are any other semver major changes that would be appropriate to include.

Jolg42:
Ok 💯 👍 Thanks for the update 😃

PR #329, opened 2020-08-15
perf: avoid deep call stack in sync 'drain'

ronag:
Noticed while profiling.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -163,7 +163,7 @@ class Client extends EventEmitter {
     this[kRetryTimeout] = null
     this[kOnDestroyed] = []
     this[kWriting] = false
-    this[kResuming] = false
+    this[kResuming] = 0 // 0, idle, 1, scheduled, 2 resuming
     this[kDrained] = false
     this[kMaxAbortedPayload] = maxAbortedPayload || 1048576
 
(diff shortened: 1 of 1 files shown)


lib/client.js
```diff
@@ -762,7 +762,20 @@ function connect (client) {
...
+    client[kResuming] = 2
+    _resume(client)
+    client[kResuming] = 0
+  }
+
+  if (client[kRunningIdx] > 256) {
+    client[kQueue].splice(0, client[kRunningIdx])
+    client[kPendingIdx] -= client[kRunningIdx]
+    client[kRunningIdx] = 0
+  }
```
ronag, line 775: Just moved this out here since it doesn't depend on state, it's just cleanup.

lib/client.js
```diff
@@ -772,7 +785,7 @@ function resume (client) {
       return
     }
 
-    if (!client.size && client[kClosed]) {
+    if (client[kClosed] && !client.size) {
```
ronag, line 788: unrelated minor perf improvement

lib/client.js
```diff
@@ -787,16 +802,11 @@ function resume (client) {
...
-    if (client[kRunningIdx] > 256) {
-      client[kQueue].splice(0, client[kRunningIdx])
-      client[kPendingIdx] -= client[kRunningIdx]
-      client[kRunningIdx] = 0
-    }
-
     if (!client.pending) {
       if (!client[kDrained] && !client.busy) {
         client[kDrained] = true
         client.emit('drain')
```
ronag, line 808: drain handler might call resume again through dispatch, this PR fixes that

PR #330, opened 2020-08-15
perf: try to avoid slow net.isIP

ronag:
`net.isIP` is slow. Avoid it when possible.

### DIFF
--- a/lib/request.js
+++ b/lib/request.js
@@ -29,7 +29,7 @@ class Request {
     requestTimeout
   }, {
     [kRequestTimeout]: defaultRequestTimeout,
-    [kUrl]: { hostname }
+    [kUrl]: { hostname, protocol }
   }, handler) {
     if (typeof path !== 'string' || path[0] !== '/') {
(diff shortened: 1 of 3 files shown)


lib/request.js
```diff
@@ -24,12 +24,11 @@ class Request {
     headers,
     idempotent,
     upgrade,
-    servername,
```
ronag, line 27: This option is not documented and I'm not sure how/when it would be useful?

lib/request.js
```diff
@@ -73,9 +72,14 @@ class Request {
...
     const hostHeader = headers && (headers.host || headers.Host)
 
-    this.servername = servername || hostHeader || null
-    if (net.isIP(this.servername) || /^\[/.test(this.servername)) {
-      this.servername = null
+    this.servername = null
+    if (
+      hostHeader &&
+      protocol === 'https:' &&
+      !/^[[\d]/.test(hostHeader) &&
```
ronag, line 79: starts with `[` or `\d`, a faster check before `net.isIP`

PR #331, opened 2020-08-15
perf: optimize pool queue

ronag:


ronag:
probably not worth it

PR #337, opened 2020-08-17
fix: dispatch receives raw headers

ronag:
https://github.com/mcollina/undici/pull/316 made me realize that we don't have an efficient way for library implementors to implement a different behavior than default.

This changes so that the dispatch methods receives raw headers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,10 +24,10 @@ Configuration: Node v14.4, HTTP/1.1 without TLS, 100 connections, Linux 5.4.12-1
 
 ```
 http - keepalive x 5,634 ops/sec ±2.53% (274 runs sampled)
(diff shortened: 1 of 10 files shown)


ronag:
> I think this needs some tests. Also, it's semver-major

If I land this should I create a 1.x release branch and we'll publish this at a later point? Or do you prefer if we just make a semver major release straight away before people have a chance of using the previous API?

ronag:
Added tests

mcollina:
> If we land this should I create a 1.x release branch and we'll publish this at a later point? Or do you prefer if we just make a semver major release straight away before people have a chance of using the previous API?

Just bump to 2.x

PR #343, opened 2020-08-19
fix: requestTimeout monitors from sent

ronag:
This changes so that requestTimeout measures how long a requests
has been in the pipeline. This is a more useful measure than how
long the request has been enqueued on the client, which is something
that is predictable.

Fixes: https://github.com/mcollina/undici/issues/341

If user properly uses `Client.busy` there should be no difference in observed behavior.


ronag:
Some failing tests to sort out if this is an acceptable approach.

test/request-timeout.js
```diff
@@ -405,31 +373,35 @@ test('client.destroy should cancel the timeout', (t) => {
   })
 })
 
-test('client.close should wait for the timeout', (t) => {
-  t.plan(2)
+// TODO: Caused by bug where client does not resume loop
+// if request is errored.
```
ronag, line 377: https://github.com/mcollina/undici/issues/344

ronag:
Counting from the client queue makes little sense as a higher level abstraction (like Pool) might also have a queue.

ronag:
@mcollina this was not reviewable, fixed now, sorry about that

mcollina:
Can you add a test?

ronag:
> Can you add a test?

I'll take that preliminary approval. Fixing tests.

mcollina:
The change is ok

PR #345, opened 2020-08-19
fix: close() deadlock

ronag:
When no resume is going to be scheduled, request.onError might never be detected or in the case of `Client.pipeline` the stream destroy.

See test case in first commit, https://github.com/mcollina/undici/pull/345/commits/adb4faee1f7a1dc545ea2373f39d49f30c4455e9.

Fixes: https://github.com/mcollina/undici/issues/344

----

There is a problem with the existing API where if a running request is aborted while the server is not providing any data nothing is ticking the state machine.

There are some API changes:

- A new hook is added on dispatch handler, `onConnect(abort)`. 

- `onData` now only returns a boolean (no magical `null`).

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -166,6 +166,9 @@ class Client extends EventEmitter {
     this[kResuming] = 0 // 0, idle, 1, scheduled, 2 resuming
     this[kDrained] = false
     this[kMaxAbortedPayload] = maxAbortedPayload || 1048576
+    this[kResume] = () => {
+      resume(this)
+    }
 
     // kQueue is built up of 3 sections separated by
     // the kRunningIdx and kPendingIdx indices.
@@ -932,11 +935,10 @@ function _resume (client) {
       return
     }
 
-    try {
-      write(client, request)
+    if (write(client, request)) {
       client[kPendingIdx]++
-    } catch (err) {
-      request.onError(err)
+    } else {
+      client[kQueue].splice(client[kPendingIdx], 1)
     }
   }
 }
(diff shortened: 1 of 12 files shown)


ronag:
Just a failing test for now... WIP

ronag:
Some tests remain to be sorted (related to maxAbortedPayload). I'll look into those if this gets approved.

lib/client.js
```diff
@@ -776,7 +752,14 @@ function connect (client) {
   }
 
   socket[kPause] = socket.pause.bind(socket)
-  socket[kResume] = socket.resume.bind(socket)
+  socket[kResume] = (err) => {
+    if (err) {
+      // TODO: Only destroy if head of pipeline.
```
ronag, line 757: This must be fixed before landing.

lib/client.js
```diff
@@ -776,7 +752,14 @@ function connect (client) {
   }
 
   socket[kPause] = socket.pause.bind(socket)
-  socket[kResume] = socket.resume.bind(socket)
+  socket[kResume] = (err) => {
+    if (err) {
+      // TODO: Only destroy when at head of pipeline.
```
ronag, line 757: Must be fixed before landing.

lib/client.js
```diff
@@ -980,6 +963,8 @@ function write (client, request) {
 
   const socket = client[kSocket]
 
+  request.onConnect(socket[kResume])
```
ronag, line 966: - onConnect
- onStart
- onDispatch
- ?

ronag:
Needs more work

mcollina:
Really good work @ronag!

Issue #210, opened 2020-06-02
Reuse TLS Sessions


ronag:
Cache and re-use TLS sessions. Similar to https://github.com/nodejs/node/blob/master/lib/https.js#L245.

PR #358, opened 2020-08-23
Reuse TLS sessions

delvedor:
As titled 🚀 
~~I'm not sure how to test this, any suggestion?~~

Closes: #210.


lib/core/client.js
```diff
@@ -739,6 +762,63 @@ function connect (client) {
...
+  let name = `${url.hostname}:${url.port}`
+  if (tls) {
+    name += ':'
+    if (tls.ca) { name += tls.ca }
+
+    name += ':'
+    if (tls.cert) { name += tls.cert }
+
+    name += ':'
+    if (tls.clientCertEngine) { name += tls.clientCertEngine }
```
ronag, line 775: I guess this comes from Node core? Can we keep with the undici style, i.e. not single line `{ statement }`?

lib/core/client.js
```diff
@@ -700,10 +702,31 @@ function connect (client) {
   if (protocol === 'https:') {
     servername = client[kServerName] || (client[kTLSOpts] && client[kTLSOpts].servername)
     const tlsOpts = { ...client[kTLSOpts], servername }
+    const sessionName = createName(client[kUrl], tlsOpts)
```
ronag, line 705: I think sessionname can be mostly precalculated in the constructor? except for `client[kServerName]` which we could simple append.

delvedor: Yeah, we could cache it but let's keep it as is for now, as I don't think it would save that much 🤔

lib/core/client.js
```diff
@@ -700,10 +702,31 @@ function connect (client) {
...
+      if (client[kTLSSessions].size >= 100) {
+        // `map.keys()` returns an iterator of the keys in insertion order
+        const firstInsertedKey = client[kTLSSessions].keys().next().value
+        client[kTLSSessions].delete(firstInsertedKey)
+      }
+      client[kTLSSessions].set(sessionName, newSession)
+    })
+
+    // Evict session on error
+    socket.once('close', (err) => {
```
ronag, line 725: Can we do this in `onSocketClose` handler?

ronag: Can we save `sessionName` as `kSessionName` on the socket?

lib/core/client.js
```diff
@@ -700,10 +702,31 @@ function connect (client) {
   if (protocol === 'https:') {
     servername = client[kServerName] || (client[kTLSOpts] && client[kTLSOpts].servername)
     const tlsOpts = { ...client[kTLSOpts], servername }
+    const sessionName = createName(client[kUrl], tlsOpts)
+    // Get cached session if it does exists
+    const session = client[kTLSSessions].get(sessionName)
+    if (session !== undefined) tlsOpts.session = session
```
ronag, line 708: style

ronag: move down `tlsOpts =` and initialize it directly?

delvedor:
I've pushed the revisited version of the [Node.js core test](https://github.com/nodejs/node/blob/master/test/parallel/test-https-agent-session-reuse.js).
As you can see it's failing, and I'm struggling for understanding why.
I've noticed that the internal `connect` function gets called only once, while the `.on('session')` event gets fired twice with two different sessions.
Are we already supporting tls session reuse and we didn't notice it?

lib/core/client.js
delvedor: What do you mean by "style"? We are using [standard](https://standardjs.com/) and the code above is valid.

delvedor: > move down tlsOpts = and initialize it directly? const tlsOpts = { ...client[kTLSOpts], servername, session }

I can't, as `createName` needs tls options.

ronag: > What do you mean by "style"? We are using standard and the code above is valid.

Use braces please.

lib/core/client.js
```diff
@@ -145,6 +151,8 @@ class Client extends EventEmitter {
     this[kClosed] = false
     this[kDestroyed] = false
     this[kServerName] = null
+    this[kTLSMaxCachedSessions] = (tls == null || tls.maxCachedSessions == null) ? 100 : tls.maxCachedSessions
+    if (tls && tls.maxCachedSessions) delete tls.maxCachedSessions
```
ronag, line 155: Is this the same as node core?

delvedor: Nope, in node core it's an agent option.

ronag: We should not mutate options.

delvedor: Do you prefer a `tls` and `tlsMaxCachedSessions` options?

ronag: Not sure why you do the delete?

delvedor:
The tests are still failing, but for a different reason 😄 
The reason why they were failing before was that I didn't merge https://github.com/nodejs/undici/pull/357 back here 😅

lib/core/client.js
delvedor: I can remove it, it was for not passing to the Node core tls implementation an option that should not be there :)
I can destructure the undici tls option if you would like to avoid mutating objects.

delvedor:
@ronag @mcollina I've fixed the code and now the tests are passing.
The issue was caused by the listening of the `close` event, which takes a single boolean parameter [`hadError`](https://nodejs.org/api/net.html#net_event_close_1), which identifies if the socket was closed due to an error. This is always happening in undici as we are using `InformatinalError`s. To solve it, I'm now listening to the error event. What do you think?

lib/core/client.js
```diff
@@ -738,8 +740,8 @@ function connect (client) {
     })
 
     // Evict session on error
-    socket.once('close', (err) => {
-      if (err) {
+    socket.on('error', (err) => {
+      if (!(err instanceof InformationalError)) {
```
ronag, line 744: err.code !== UND_...

delvedor:
While cleaning up the code I've noticed that we are actually doing more work than needed.
The Node.js core implementation of TLS reuse is built for the HTTP agent, which is designed to talk to anything on the internet.
Per our documentation:

> A basic HTTP/1.1 client, mapped on top a single TCP/TLS connection.

This means that we will never use more than one TLS session since we always talk to the same endpoint. I've refactored the implementation to cache just one TLS session.

delvedor:
We could probably share the session between clients by using the Pool api 🤔

ronag (approved):
Nice!

mcollina (approved):
lgtm

PR #363, opened 2020-08-27
add support for iterators

### LINKED ISSUE #198: Request Body as Iterable/Iterator
Current flow is to just use `Readable.from(i)`. Which is fine. But we could make this faster.

Ethan-Arrowood:
closes #198


lib/core/util.js
```diff
@@ -10,6 +10,11 @@ function isStream (body) {
   return !!(body && typeof body.on === 'function')
 }
 
+function isGenerator (fn) {
+  const name = fn.constructor.name
```
ronag, line 14: fn could be null and/or not a function

lib/core/request.js
```diff
@@ -74,6 +75,8 @@ class Request {
         // TODO: Ignore error if body has ended?
         this.onError(err)
       })
+    } else if (util.isGenerator(body)) {
+      this.body = Readable.from(body())
     } else if (util.isBuffer(body)) {
```
ronag, line 80: We could also add support for iterable at the same time?

ronag (commented):
Think out load.

- I wonder if whether it is iterable and/or generator function we should support.
- Should we assume a generator function is retryable?

lib/core/request.js
Ethan-Arrowood: Ahh so checking for `[Symbol.iterator]` or something like that right?

Ethan-Arrowood:
I think if I make this work for just iterators in general then it'll automatically support generators. I'll give that a try now

lib/core/request.js
ronag: check what Readable.from does

Ethan-Arrowood: yep so in that case if its an iterable object then we can pass it directly to Readable.from

if its a generator function we have to pass the return value of it to Readable.from

Ethan-Arrowood:
So I added the generator support in as a separate commit. easy to revert if we don't want that

Ethan-Arrowood:
Strange, not sure why tests failed this time. node 14 works on my mac

ronag:
> Strange, not sure why tests failed this time. node 14 works on my mac

Looks like a flaky test.

lib/core/util.js
```diff
@@ -10,6 +10,15 @@ function isStream (body) {
   return !!(body && typeof body.on === 'function')
 }
 
+function isIterator (fn) {
+  return !!(fn != null && typeof fn[Symbol.iterator] === 'function')
```
ronag, line 14: asyncIterator

ronag: `isIterable`

ronag: `isIterator` is a different check, i.e. `typeof next === 'function'`.

lib/core/request.js
```diff
@@ -78,6 +79,10 @@ class Request {
       this.body = body.length ? body : null
     } else if (typeof body === 'string') {
       this.body = body.length ? Buffer.from(body) : null
+    } else if (util.isIterator(body)) { // check this after buffer and string so it doesn't return false positives
+      this.body = Readable.from(body)
+    } else if (util.isGenerator(body)) {
+      this.body = Readable.from(body())
```
ronag, line 85: I think you might be able to skip the checks here and just pass it to `Readable.from` and catch INVALID_ARG_TYPE errors from there. That way you don't need to re-implement the type checks.

Ethan-Arrowood: Hmm okay I also found this line in the [documentaion](https://nodejs.org/dist/latest-v14.x/docs/api/stream.html#stream_stream_readable_from_iterable_options):

> Calling Readable.from(string) or Readable.from(buffer) will not have the strings or buffers be iterated to match the other streams semantics for performance reasons.

So like is it potentially safe to just always pass anything but an existing stream to Readable.from?

mcollina (changes requested):
I might be too late to the party but I do not think that using `Readable.from()` is a good way of doing this. It's the simplest way... however it would add significant overhead. If we want those to be fast we should not wrap the generator in a string.

Ethan-Arrowood:
Should I use a `for...of` loop instead?

mcollina:
I would use `pipeline` somewhere down, it handles those (or copy what they are doing).

Ethan-Arrowood:
And would `this.body` be the writestream in that case?

Ethan-Arrowood:
If the new try catch method is better I'll delete the util methods

ronag:
You basically need to replicate all the logic behind útil.isStream but with útil.isIterable

Ethan-Arrowood:
what is missing from my current implementation of `util.isIterable`? 

Also, with the try/catch do we even need those util methods?

ronag:
> what is missing from my current implementation of util.isIterable?

Nothing is missing.

> Also, with the try/catch do we even need those util methods?

No really...

But as @mcollina pointed out, this implementation probably works but is not as fast as it could be.

lib/core/request.js
```diff
@@ -79,7 +80,15 @@ class Request {
     } else if (typeof body === 'string') {
       this.body = body.length ? Buffer.from(body) : null
     } else {
-      throw new InvalidArgumentError('body must be a string, a Buffer or a Readable stream')
+      try {
+        this.body = Readable.from(typeof body === 'function' ? body() : body)
```
ronag, line 84: I think we might want to skip generator support and only to iterable. Generators have some semantics we might want to give more consideration to.

Ethan-Arrowood: Okay I'll remove generator support

Ethan-Arrowood:
Okay sounds good. I'll remove the utils and rebase for the conflict. @mcollina if you have any more pointers on making this faster let me know; i'd like to look into it and give it a shot.

ronag:
> if you have any more pointers on making this faster let me know; i'd like to look into it and give it a shot.

In order to make this faster you should avoid converting it to stream and iterate it directly in the `write(client, request)` method. Also take into account all the places where there is logic based on `util.isStream(body)` and implement accordingly.

Ethan-Arrowood:
> In order to make this faster you should avoid converting it to stream and iterate it directly in the write(client, request) method. Also take into account all the places where there is logic based on util.isStream(body) and implement accordingly.

Ahh I understand better now. Okay I'll give it a shot too

Ethan-Arrowood:
Help me think through something please:

The iterable object a user passes to the body, does it need to always yield a string or buffer? If so do I need to check for that when I'm iterating over the iterable inside of `write` and handle accordingly?

ronag:
I think you should check for string or buffer and throw if something else

Ethan-Arrowood:
And I should ditch generator support?

lib/core/client.js
```diff
@@ -1153,6 +1153,36 @@ function write (client, request) {
...
+    let chunks = stringOrBuffer(result.value) ? result.value : [result.value]
+    while (true) {
+      if (result.done) break
+      result = iterator.next()
+      if (stringOrBuffer(result.value)) {
+        chunks += result.value
+      } else {
+        chunks.push(result.value)
+      }
+    }
```
ronag, line 1177: Close, I would probably do it like this:

```js
async function write(iterable, socket) {
  try {
    if (!contentLength) {
      // chunked transfer
    }
    for await (const chunk of iterable) {
      socket.write(chunk)
    }
    socket.write('\r\n', 'ascii')

    request.body = null
  } catch (err) {
    util.destroy(socket, err)
  } 
}
```

The iterable could be huge and not appropriate to read into memory.

ronag (changes requested):

lib/core/client.js
ronag: You basically need to do the same thing as for streams.

Ethan-Arrowood: Ahh okay I'll try this out. Thank you!

Ethan-Arrowood: making some progress here, but since this is taking me quite a bit of time i don't want to impede the overall progress on the library. @ronag would you be interested in maybe doing a pair programming call and working through this together? No worries if you're too busy right now; i'll keep plugging away at it

ronag: This week is swamped for me. If you are still working on it next week we could try to do some pair programming.

Ethan-Arrowood: Okay sounds good!

mcollina (approved):
lgtm

lib/core/client.js
```diff
@@ -1153,6 +1153,29 @@ function write (client, request) {
...
+    try {
+      if (contentLength === null) {
+        socket.write(`${header}transfer-encoding: chunked\r\n`, 'ascii')
+      } else {
+        socket.write(`${header}content-length: ${contentLength}\r\n\r\n`, 'ascii')
+      }
+      for await (const chunk of body) {
+        if (typeof chunk !== 'string' && !util.isBuffer(chunk)) {
+          throw new Error('Body iterable must yield either string or buffer')
+        } else if (socket.write(chunk)) {
```
ronag, line 1166: Chunked encoding needs to write length becore each chunk

Ethan-Arrowood:
From [http watch](https://www.httpwatch.com/httpgallery/chunked/)

> However, with chunked encoding you could just write the data one row at a time and write a zero sized chunk when the end of the query was reached.

I think I'm missing the zero sized chunk at the end. Will try and add this tomorrow

Ethan-Arrowood:
So by making `write` async, i've now broken other things. Hmmmm

Ethan-Arrowood:
Hmm removing async seems to solve things but I guess we lose support for asyncIterators? Or maybe not? This is really pressing my understanding of async things

lib/core/util.js
```diff
@@ -10,12 +10,18 @@ function isStream (body) {
   return !!(body && typeof body.on === 'function')
 }
 
+function isIterable (obj) {
+  return !!(obj != null && (typeof obj[Symbol.iterator] === 'function' || typeof obj[Symbol.asyncIterator] === 'function'))
```
Ethan-Arrowood, line 14:
```suggestion
  return !!(obj != null && typeof obj[Symbol.iterator] === 'function')
```
Without async support in `write` method I don't think `asyncIterators` will work any longer

ronag: async iterators should work

lib/core/client.js
```diff
@@ -1153,6 +1153,41 @@ function write (client, request) {
...
       .on('error', onFinished)
       .on('close', onFinished)
+  } else if (util.isIterable(body)) {
+    try {
+      if (contentLength === null) {
+        socket.write(`${header}transfer-encoding: chunked\r\n`, 'ascii')
+      } else {
+        socket.write(`${header}content-length: ${contentLength}\r\n\r\n`, 'ascii')
+      }
+      for (const chunk of body) {
```
ronag, line 1163: for await

ronag: move this part out to a `async function writeIterable` and maybe refactor the above into a `function writeStream`

Ethan-Arrowood: when I make the write method async it breaks a bunch of stuff since write is called synchronously in a while loop. any ideas to change that?

ronag: Yes... what I wrote above 😊

ronag (changes requested):
You need to add a check here as well:

```js
    if (client.running && (util.isStream(request.body) || util.isIterable(request.body)) {
      // Request with stream body can error while other requests
      // are inflight and indirectly error those as well.
      // Ensure this doesn't happen by waiting for inflight
      // to complete before dispatching.

      // Request with stream body cannot be retried.
      // Ensure that no other requests are inflight and
      // could cause failure.
      return
    }
```

Also a test that fails without it.

Also please try to make sure `npm run coverage` is at `100`.

lib/core/client.js
```diff
@@ -916,6 +916,42 @@ function _resume (client) {
...
+  try {
+    if (contentLength === null) {
+      socket.write(`${header}transfer-encoding: chunked\r\n`, 'ascii')
+    } else {
+      socket.write(`${header}content-length: ${contentLength}\r\n\r\n`, 'ascii')
+    }
+    for await (const chunk of body) {
+      if (typeof chunk !== 'string' && !util.sBuffer(chunk)) {
+        throw new Error('Body iterable must yield either string or buffer')
+      }
```
ronag, line 929: I'm not sure you need this check? socket.write() will fail anyway

lib/core/client.js
```diff
@@ -916,6 +916,42 @@ function _resume (client) {
...
+      if (socket.write(chunk)) {
+        continue
+      } else if (socket.destroyed) {
+        return
+      } else {
+        await EventEmitter.once(socket, 'drain')
+      }
+    }
+    socket.write('\r\n0\r\n\r\n', 'ascii')
+    socket.cork()
```
ronag, line 949: this cork looks wrong?

Ethan-Arrowood:
Just working on unit tests now! Some branches i'm having a hard time reaching but I'm getting there 👍

lib/core/client.js
```diff
@@ -909,7 +909,10 @@ function _resume (client, sync) {
       request.body = null
     }
 
-    if (client.running && util.isStream(request.body)) {
+    if (client.running &&
```
Ethan-Arrowood, line 912: @ronag how do I get the request into a "running" state while also still having the request.body defined?

ronag: Create two request with a non empty stream body?

Ethan-Arrowood: haven't forgotten about this. I tried again but didn't make much progress. I'm going to try another way later by manually stepping through the execution and seeing what I will need to do to get the code down a certain path

Ethan-Arrowood: I think I gotta ask for help on this. Really struggling to write tests to get to some of the edge cases in the writeIterable method. @ronag would you like to pair on this? otherwise, if you want to author the necessary tests that is okay with me

ronag: I'll get back to you.

ronag: @Ethan-Arrowood We already have tests which hit this when using streams? Can't you just find the test and replace the stream part with an iterable?

lib/core/client.js
```diff
@@ -909,7 +909,10 @@ function _resume (client, sync) {
...
+      !util.isBuffer(request.body) &&
+      typeof request.body !== 'string' &&
+      (util.isStream(request.body) || util.isIterable(request.body))) {
```
ronag, line 915: This looks wrong, it should just be `if (client.running && (isStream(request.body) || isIterable(request.body)))`

lib/core/client.js
```diff
@@ -929,6 +932,41 @@ function _resume (client, sync) {
...
+  try {
+    if (contentLength === null) {
+      socket.write(`${header}transfer-encoding: chunked\r\n`, 'ascii')
+    } else {
+      socket.write(`${header}content-length: ${contentLength}\r\n\r\n`, 'ascii')
+    }
+    for await (const chunk of body) {
+      // if (typeof chunk !== 'string' && !util.sBuffer(chunk)) {
+      //   throw new Error('Body iterable must yield either string or buffer')
+      // }
```
ronag, line 945: Uncommented?

lib/core/request.js
```diff
@@ -49,8 +49,10 @@ class Request {
       this.body = body.length ? body : null
     } else if (typeof body === 'string') {
       this.body = body.length ? Buffer.from(body) : null
+    } else if (util.isIterable(body)) {
+      this.body = body
     } else {
-      throw new InvalidArgumentError('body must be a string, a Buffer or a Readable stream')
+      throw new InvalidArgumentError('body must be a string, a Buffer, a Readable stream, or an iterable')
```
ronag, line 55: `Iterable, AsyncIterable`

lib/core/client.js
Ethan-Arrowood: yeah thats what I thought i must've introduce some regressions cause the coverage reporter was saying some of these lines were not getting tested.

lib/core/client.js
```diff
@@ -929,6 +932,41 @@ function _resume (client, sync) {
...
+        return
+      } else {
+        await EventEmitter.once(socket, 'drain')
+      }
+    }
+    socket.write('\r\n0\r\n\r\n', 'ascii')
+  } catch (err) {
+    util.destroy(socket, err)
+  }
+}
```
ronag, line 968: Take  another look at the streams version. You need to e.g. set `client[kWriting]`.

ronag: Missing checks with `bytesWritten`

ronag: You basically want to almost the exact same thing but with a for await loop instead of stream events.

lib/core/client.js
Ethan-Arrowood: not sure why i commented this; will fix.

lib/core/client.js
```diff
@@ -909,7 +909,10 @@ function _resume (client, sync) {
       request.body = null
     }
 
-    if (client.running && util.isStream(request.body)) {
+    if (client.running &&
+      !util.isBuffer(request.body) &&
+      typeof request.body !== 'string' &&
```
ronag, line 914: I don't think body can be a string here?

Ethan-Arrowood: It failed without these checks. I'll have to look back on it though

lib/core/client.js
ronag: This needs significantly more tests. Preferably we should just port most of the body as stream test into a body as iterable test.

lib/core/client.js
ronag: Let me know what tests fail and I can maybe give a hint.

Ethan-Arrowood: This test fails when I remove the buff and string checks: https://github.com/nodejs/undici/blob/master/test/client-pipelining.js#L453

lib/core/client.js
Ethan-Arrowood: okay i've gone ahead and modified writeIterable to be as close as I can make it to the stream handling. I'm failing some tests now but I'd appreciate your input on my new changes.

lib/core/client.js
Ethan-Arrowood: fwiw you asked this to be removed previously so its gone now!

Ethan-Arrowood:
Since its been a while since I made any progress on this I'm going to self-mark this as **stale**. If someone else wants to take a stab at implementing iterators please do so ❤️

Linkgoron:
I've started working on this, unless there are any objections, based on @Ethan-Arrowood original work.

ronag:
This landed in https://github.com/nodejs/undici/pull/848

PR #366, opened 2020-08-28
refactor: simplify busy

ronag:
This significantly simplifies the `busy` logic. Slightly changes behavior of `busy`. However, since we now have `'drain'` it has minor impact.

Actually seems to improve performance slightly.

The difference is mostly for the `Client.pipeline` API.


ronag:
Still has some test that need to be sorted to account for changed behavior.

mcollina (approved):
lgtm

PR #367, opened 2020-08-28
Share the TLS session among all clients in the pool

delvedor:
As titled, continuation of https://github.com/nodejs/undici/pull/358.

### DIFF
--- a/lib/pool.js
+++ b/lib/pool.js
@@ -6,6 +6,7 @@ const {
 } = require('./core/errors')
 const FixedQueue = require('./core/node/fixed-queue')
 
+const { kTLSSession } = require('./core/symbols')
 const kClients = Symbol('clients')
 const kQueue = Symbol('queue')
 
@@ -24,6 +25,7 @@ class Pool {
     }, () => new Client(url, options))
 
     const queue = this[kQueue]
+    const clients = this[kClients]
     function onDrain () {
       while (!this.busy) {
         const item = queue.shift()
@@ -34,8 +36,15 @@ class Pool {
       }
     }
 
-    for (const client of this[kClients]) {
+    function onTLSSession (session) {
+      for (const client of clients) {
+        client[kTLSSession] = session
+      }
+    }
+
+    for (const client of clients) {
       client.on('drain', onDrain)
+      client.on('tlsSession', onTLSSession)
     }
   }
 
@@ -73,3 +82,4 @@ class Pool {
 }
 
 module.exports = Pool
+module.exports[Symbol.for('pool-symbols')] = { kClients, kQueue }
--- a/lib/core/client.js
+++ b/lib/core/client.js
@@ -685,6 +685,7 @@ function onSocketClose () {
       client[kQueue][client[kRunningIdx]++].onError(err)
       // Evict session on error
       client[kTLSSession] = null
+      client.emit('tlsSession', null)
     }
 
     const retryRequests = []
@@ -742,6 +743,7 @@ function connect (client) {
     // Cache new session for reuse
     socket.on('session', (newSession) => {
       client[kTLSSession] = newSession
+      client.emit('tlsSession', newSession)
     })
   } else {
     socket = client[kSocketPath]
--- a/test/tls-session-reuse.js
+++ b/test/tls-session-reuse.js
@@ -5,8 +5,9 @@ const { join } = require('path')
 const https = require('https')
 const crypto = require('crypto')
 const { test } = require('tap')
-const { Client } = require('..')
-const { kSocket, kTLSOpts } = require('../lib/core/symbols')
+const { Client, Pool } = require('..')
+const { kClients } = Pool[Symbol.for('pool-symbols')]
+const { kSocket, kTLSOpts, kTLSSession } = require('../lib/core/symbols')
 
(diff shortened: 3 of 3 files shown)


ronag (changes requested):
Pool should not need access to any internal api. Also I don’t see the advantage here?

delvedor:
> Pool should not need access to any internal api.

Do you see a different way of implementing this?

> Also I don’t see the advantage here?

The advantage is that once a client has established a tls connection, we share the tls session to every other client, so we don't need to establish one multiple times.

ronag:
> The advantage is that once a client has established a tls connection, we share the tls session to every other client, so we don't need to establish one multiple times.

Can the same session be used multiple times by different sockets?

ronag:
> Do you see a different way of implementing this?

Add a `'session'` event on `Client`?

ronag:
Ah sorry, you are not using internal API. Though not sure I like `tlsSession`. But that's mostly cosmetics.

delvedor:
> Can the same session be used multiple times by different sockets?

Isn't what the [Node core agent is doing](https://github.com/nodejs/node/blob/master/lib/https.js#L117-L126)? If you open multiple sockets with the same name, then the session will be reused.

lib/pool.js
```diff
@@ -6,6 +6,7 @@ const {
 } = require('./core/errors')
 const FixedQueue = require('./core/node/fixed-queue')
 
+const { kTLSSession } = require('./core/symbols')
```
ronag, line 9: `const kTLSSession = Symbol('tlsSession')`

ronag: I'd like to avoid confusion for people looking at `Pool` so that they don't think the implementation depends on internals.

lib/pool.js
```diff
@@ -73,3 +82,4 @@ class Pool {
 }
 
 module.exports = Pool
+module.exports[Symbol.for('pool-symbols')] = { kClients, kQueue }
```
ronag, line 85: If pool also emits `tlsSession` do you need to export the symbol like this?

lib/pool.js
delvedor: I'm not sure I follow. I **must** use the symbol present in our symbol registry, otherwise, the pool will not access the correct cache.

lib/pool.js
delvedor: Using an event will not work for how the test is designed, because the event will be fired fewer times, and I'm not able to access the client that has been used for a specific request.

delvedor:
> Though not sure I like tlsSession. But that's mostly cosmetics.

Any name is fine for me, I think it's important it contains "tls" tho.

lib/pool.js
ronag: I'm not sure I understand either. Why?

```js
function onTLSSession (session) {
  this[kTLSSession] = session
}
```

lib/pool.js
```diff
@@ -34,8 +36,15 @@ class Pool {
       }
     }
 
-    for (const client of this[kClients]) {
+    function onTLSSession (session) {
+      for (const client of clients) {
+        client[kTLSSession] = session
```
ronag, line 41: I see, you are overwriting here. This is a no no from me.

lib/pool.js
delvedor: I want to test that the correct session has been propagated to all clients.

lib/pool.js
delvedor: What would you suggest? Add a method to `Client` for overriding the value?

ronag: No. That's weird side effects. Maybe we need some kind of `agent` that you can pass to `Client` which is responsible for creating sockets and can be shared between clients.

```js
const client = new Client({  agent/connectionFactory/socketFactory })
```

ronag: That way we don't need any extra props or events in `Client`.

delvedor: Yes, an agent-like object could be a solution, but I feel that given the scope of what we need (share a tls session), it might be an overengineered solution.

ronag: I would rather not have this optimization than breaking encapsulation and/or adding hacks to the API. :/

ronag: `Pool` is for me at the moment a reference implementation and is something I'd like to encourage userland to experiment in building their own versions.

delvedor: I think that undici should ship an official pool, as in most of the cases, people don't have time or the knowledge for building one.

ronag: > I think that undici should ship an official pool, as in most of the cases, people don't have time or the knowledge for building one.

We are shipping an official pool?

delvedor: > Pool is for me at the moment a reference implementation

Maybe I misunderstood this?

mcollina (changes requested):
I don't think copying the session from one to the other would work well. Also this would not help at all with connecting those clients. The best approach for TLS would be to connect once, and then reuse that session _for all the new clientss_.

ronag:
There is also the problem where servername can change due to the host header for a specific request.

ronag:
I don't think this is the direction we want to go and the PR has been stale for a while. Closing.

PR #369, opened 2020-08-29
Add ESM wrapper for named imports

talentlessguy:
Hello, I'd like to add ESM wrapper to be able to use named imports in Node ESM, like this:

```js

const client = new Client(`http://example.com`)

client.request({
  path: '/',
  method: 'GET'
}, function (err, data) {
  if (err) {
    // handle this in some way!
    return
  }

  const {
    statusCode,
    headers,
    body
  } = data

  body.setEncoding('utf8')
  body.on('data', console.log)

  client.close()
})
```

I simply wrapped the index.js file and put a `"module"` field to package.json

Do I need to do any other work, or it is ok as it is now?

thanks in advance for PR feedback


mcollina (commented):
Can you please add a unit test for this? Thanks.

talentlessguy:
> Can you please add a unit test for this? Thanks.

pardon for forgetting to do it, going to add it soon and write back here

talentlessguy:
@mcollina I added the tests, they passed on all environments, except macOS Node 14.X

And the error seems to drop in another place: https://github.com/nodejs/undici/pull/369/checks?check_run_id=1044833326#step:5:6274

```
 not ok 8 - should be equal
          ---
          found: true
          wanted: false
          compare: ===
          at:
            line: 92
            column: 7
            file: test/tls.js
          stack: |
            test/tls.js:92:7
            lib/client-request.js:90:14
          source: |2
                t.strictEqual(didDisconnect, true)
                t.strictEqual(client[kSocket].authorized, false)
            ------^
                t.ok(err)
              })
          ...
        
        ok 9 - expect truthy value
        # failed 1 of 9 tests
```

not sure what is causing it, but is not my ESM wrapper because it is not used in that test suite that fails

What should I do in this case? :thinking:

ronag:
> not sure what is causing it, but is not my ESM wrapper because it is not used in that test suite that fails

It's a flaky test not related. I'm investigating it.

ronag (approved):

mcollina (approved):
lgtm

talentlessguy:
yaaay it got merged :DD

Issue #356, opened 2020-08-23
body is lacking `'error'` handler while queued in pool


ronag:
i.e. the following logic does not execute until the request is dispatched on the client:

https://github.com/mcollina/undici/blob/master/lib/core/request.js#L166-L172
```js
    if (util.isStream(this.body)) {
      // TODO: Cleanup listeners?
      this.body.on('error', (err) => {
        // TODO: Ignore error if body has ended?
        this.onError(err)
      })
    }
```

PR #374, opened 2020-08-29
fix: register body error handler early

ronag:
~~Fixes: https://github.com/nodejs/undici/issues/356~~


ronag:
Doesn't fix #356 but does fix it for requests in the `Client` pending queue.

PR #377, opened 2020-08-30
refactor: move logic out from Request

ronag:
A step towards trying to avoid dependency on `Client` inside `Request`.


mcollina (approved):
lgtm

PR #379, opened 2020-08-30
refactor: reduce Request dependency on client

ronag:
With a few more changes `Request` would no longer need to live in core.


mcollina (approved):
the new api needs tests.

ronag:
> the new api needs tests.

This PR does not introduce any new public API.

PR #383, opened 2020-08-30
refactor: remove Client requestTimeout option

ronag:
Fully decouples `Request` from `Client` at the cost of losing the global `requestTimeout` option.

A little bit hard to motivate but still something I would personally like to do. Eventually maybe even move `requestTimeout` out from core and into handlers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,11 +69,6 @@ Options:
   transport latency.
   Default: `1e3` milliseconds (1s).
 
-- `requestTimeout: Number`, the timeout after which a request will time out.
-  Monitors time between request is dispatched on socket and receiving
-  a response. Use `0` to disable it entirely.
-  Default: `30e3` milliseconds (30s).
-
 - `pipelining: Number`, the amount of concurrent requests to be sent over the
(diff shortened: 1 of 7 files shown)


mcollina (commented):
I think this is kind of a needed feature. I do not see many reasons to _not_ have this on all the time.

ronag:
> I think this is kind of a needed feature. I do not see many reasons to not have this on all the time.

This PR doesn't change that. It is still always on by default. However, post this PR if the user wants to override the default value (30s) they must do so for every request and can't globally change it through the "duplicate" setting in `new Client(opts)`.

i.e.

```js
const client = new Client(...)
client.request(requestTimeout: 10s, ...)
client.request(requestTimeout: 10s, ...)
```
vs:

```js
const client = new Client({ requestTimeout: 10s, ... })
client.request(...)
client.request(...)
```

ronag:
Note I do not feeling strongly about this other than trying to decouple things as much as possible to make things simpler and keep "core" as small as possible.

mcollina (approved):
lgtm

PR #392, opened 2020-08-30
fix: Pool.destroy + queued requests

ronag:
Fixes: https://github.com/nodejs/undici/issues/390
Fixes: https://github.com/nodejs/undici/issues/393


mcollina (approved):
lgtm

PR #400, opened 2020-09-05
 doc: bench node 15

ronag:
Not sure what we've done in node 15 but it's significantly faster.


mcollina:
wooow!

ronag:
Included in https://github.com/nodejs/undici/pull/406

PR #410, opened 2020-09-08
fix: catch invalid arg type body

ronag:
If body is an object stream don't crash the process. Error the request.


mcollina (approved):
lgtm

Issue #418, opened 2020-09-17
Undici tries to reconnect indefenitely on `ECONNRFUSED`


dnlup:
Sorry to bother again :pray: .

I noticed that on `ECONNREFUSED` the client keeps on trying to reconnect.

Steps to reproduce:

```js
const { Client } = require('undici')

const client = new Client('http://localhost:3000') // no server is listening on this port

client.request({ path: '/', method: 'GET' }).then(console.log).catch(console.error)
```

On version `1.3.1` the call fails with a `Socket Timeout` error.
 
Is this behavior by design on version 2?

I might be wrong, but It looks like there is a connection `retryTimeout`, but not a maximum number of attempts.

These are the parts that I have seen handling the socket connection:

* https://github.com/nodejs/undici/blob/9365968fd3cbcbf13c83c6b62e691bdd57c5e163/lib/core/client.js#L710

* https://github.com/nodejs/undici/blob/9365968fd3cbcbf13c83c6b62e691bdd57c5e163/lib/core/client.js#L851

Particularly the socket error handler

* https://github.com/nodejs/undici/blob/9365968fd3cbcbf13c83c6b62e691bdd57c5e163/lib/core/client.js#L619

ronag:
This is by design. There is no maximum retry. How would you like it to behave and why?

ronag:
Does your example no fail at all? That is something we should probably resolve.

dnlup:
> Does your example not fail at all?

It hangs, I would say indefinitely or for a very long time. Before version 2 it would have failed after the default 30 seconds of `requestTimeout` or a similar timeout.

ronag:
> It hangs, I would say indefinitely or for a very long time. 

I don't think this is right but I'm unsure what would be the appropriate behavior. Some kind of timeout somewhere should apply.

dnlup:
> This is by design. There is no maximum retry. How would you like it to behave and why?

Well, the fact that this is by design makes me think. I understand your point of view. I had a test suite that was simulating network errors and one of the codes I was using was ECONNREFUSED.

I am a bit unsure too, the only thing that I am certain of is that undici should tell something to the user. WDYT @ronag ?

> Some kind of timeout somewhere should apply.

Do you prefer this approach over the number of attempt? I don't have a particular preference, just curious.

ronag:
> Do you prefer this approach over the number of attempt? I don't have a particular preference, just curious.

Yes, I prefer a timeout.

ronag:
[...]

PR #419, opened 2020-09-17
fix: error pending on socket error

ronag:
Fixes: https://github.com/nodejs/undici/issues/418

### DIFF
--- a/lib/core/client.js
+++ b/lib/core/client.js
@@ -626,15 +626,7 @@ function onSocketError (err) {
     while (client.pending && client[kQueue][client[kPendingIdx]].host === client[kHost]) {
       client[kQueue][client[kPendingIdx]++].onError(err)
     }
-  } else if (
-    !client.running &&
-    err.code !== 'ECONNRESET' &&
-    err.code !== 'ECONNREFUSED' &&
-    err.code !== 'EHOSTUNREACH' &&
-    err.code !== 'EHOSTDOWN' &&
-    err.code !== 'UND_ERR_SOCKET' &&
-    err.code !== 'UND_ERR_INFO'
-  ) {
+  } else if (!client.running) {
     assert(client[kPendingIdx] === client[kRunningIdx])
     // Error is not caused by running request and not a recoverable
     // socket error.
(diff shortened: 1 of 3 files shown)


ronag:
@dnlup would you like to try to sort out the tests and also add additional tests (e.g. the case you showed in the [issue](https://github.com/nodejs/undici/issues/418))?

dnlup:
@ronag yes! Is it ok if I do it tomorrow?

dnlup:
After doing some digging, I think the issue is caused by moving the timeout initialization [here](https://github.com/nodejs/undici/blob/v2.0.1/lib/core/request.js#L142) from [here](https://github.com/nodejs/undici/blob/v1.3.1/lib/request.js#L181). I'll work on it.

dnlup:
I think I am going to expose a `onStart` method on the `Request` to allow passing the `resume` function from the `client`.

ronag:
I'm also fine wiht doing it like in this PR, i.e. fail all pending requests on socket error and leave it up to the user to retry. I think that's actually better in some ways. Gives more control to the user.

dnlup:
Ok, I agree. I'll work on the tests then :wink:

lib/core/client.js
```diff
@@ -626,15 +626,7 @@ function onSocketError (err) {
...
       client[kQueue][client[kPendingIdx]++].onError(err)
     }
-  } else if (
-    !client.running &&
-    err.code !== 'ECONNRESET' &&
-    err.code !== 'ECONNREFUSED' &&
-    err.code !== 'EHOSTUNREACH' &&
-    err.code !== 'EHOSTDOWN' &&
-    err.code !== 'UND_ERR_SOCKET' &&
-    err.code !== 'UND_ERR_INFO'
```
ronag, line 636: UND_ERR_INFO should probably still be here

dnlup: Should the client try to reconnect with those error codes?

dnlup:
I opened https://github.com/nodejs/undici/pull/423 trying to address your suggestions.

Issue #394, opened 2020-08-31
Throwing in `'drain'` handler doesn't always lead to unhandled exception


ronag:

zhangwinning:
could you explanation the question by example,  I'll try to polish the patch and submit a fix PR.

ronag:
So:

```js
const client = new Client(...)

client.on('drain', () => {
  throw new Error() // should always lead to unhandled exception
})
```

Depending on the call stack during `.emit('drain')` the thrown exception might or might not cause a unhandled exception.

mcollina:
It seems Zalgo is here and we have a missing nextTick somewhere.

mcollina:
In other terms, we should not support throwing in event handlers, and it should always go in uncaughtException.

zhangwinning:
> So:
> 
> ```js
> const client = new Client(...)
> 
> client.on('drain', () => {
>   throw new Error() // should always lead to unhandled exception
> })
> ```
> 
> Depending on the call stack during `.emit('drain')` the thrown exception might or might not cause a unhandled exception.
> 
> Alternatively we catch the error and emit `'error'` on the client object or something else. We might also have the same issue with the `'connect'` and `'disconnect'` events.
> 
> The important thing is that the behavior is consistent and documented.
> 
> @mcollina: Wdyt? How should we handle users throwing in event handlers?

Hi @ronag, I think the below code must thrown exception, not throw exception which it ？ 
```
const { Client } = require('.')

const { createServer } = require('http')

let server = createServer((req, res) => {
  res.end('ok')
}).listen(0)

const client = new Client(`http://localhost:${server.address().port}`)

client.on('drain', () => {
  throw new Error()
})

client.emit('drain')

```

ronag:
Sorry I don't understand the question.

PR #420, opened 2020-09-17
fix: always resume in next tick

ronag:
Fixes: https://github.com/nodejs/undici/issues/394


ronag:
Needs test and a benchmark run.

ronag:
Might be a better idea to just do `process.nextTick(emitDrain, this)`

mcollina (approved):
lgtm... are the benchmarks affected?

ronag:
@mcollina: Benchmarks were affected. Made an alternative solution.

mcollina (approved):
lgtm, does this affect benchmarks?

ronag:
> lgtm, does this affect benchmarks?

Nope

PR #423, opened 2020-09-18
test: add socket error and remove reconnect from client

dnlup:
Don't error requests in client `onSocketError` on these codes:

* `UND_ERR_INFO`
* `UND_ERR_SOCKET`

Modify `client-reconnect.js` test. Add socket error test.

### DIFF
--- a/test/client-reconnect.js
+++ b/test/client-reconnect.js
@@ -6,36 +6,42 @@ const { createServer } = require('http')
 const FakeTimers = require('@sinonjs/fake-timers')
 
 test('multiple reconnect', (t) => {
-  t.plan(3)
+  t.plan(5)
 
+  let n = 0
   const clock = FakeTimers.install()
   t.teardown(clock.uninstall.bind(clock))
 
   const server = createServer((req, res) => {
-    res.end()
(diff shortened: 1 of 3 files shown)


ronag (changes requested):
Rather than removing the reconnect test. Please modify it

dnlup:
I tried triggering a `UND_ERR_SOCKET` but I wasn't able to follow the same flow of that test. I'll try again, I am sure I missed something.

test/client-reconnect.js
```diff
@@ -6,36 +6,42 @@ const { createServer } = require('http')
 const FakeTimers = require('@sinonjs/fake-timers')
```
dnlup, line 7: I am not sure I got this right and completely understood how the reconnection works in this case.

dnlup: I have seen similar tests in `client-request.js` if I am not mistaken.

mcollina (approved):
lgtm

ronag:
Thanks!

dnlup:
Thank you @ronag for the help

PR #425, opened 2020-09-19
add proxy example

ronag:


examples/proxy/proxy.js
```diff
@@ -0,0 +1,237 @@
...
+  onHeaders (statusCode, headers, resume) {
+    if (statusCode < 200) {
+      return
+    }
+
+    this.res.on('drain', resume)
+
+    headers = getResponseHeaders(headers)
+
+    // TODO (perf): res.writeHead should support Array and/or string.
```
ronag, line 110: @mcollina: It would be nice if core `writeHead` support the key/value array directly without having to convert it into an object.

examples/proxy/index.js
```diff
@@ -0,0 +1,47 @@
...
+
+const client = new Pool('http://localhost:4001', {
+  connections: 256,
+  pipelining: 1
+})
+
+Promise.all([
+  new Promise(resolve => {
+    // Proxy
+    http.createServer(async (req, res) => {
```
mcollina, line 13: let's not use async here. `writeHead()` could potentially throw and we'll be in a unhandledRejection situation.

examples/proxy/index.js
```diff
@@ -0,0 +1,47 @@
...
+  const client = new Client('http://localhost:4000')
+  const { body } = await client.request({
+    method: 'GET',
+    path: '/'
+  })
+
+  for await (const chunk of body) {
+    console.log(String(chunk))
+  }
+})
```
mcollina, line 45: I would recommend to just avoid `.then` in this file and use a `async run() {}` function at the top.

examples/proxy/proxy.js
```diff
@@ -0,0 +1,257 @@
...
+  }
+
+  onError (err) {
+    this.callback(err)
+  }
+}
+
+const HOP_EXPR = /^(te|via|host|upgrade|trailers|forwarded|connection|keep-alive|http2-settings|transfer-encoding|proxy-connection|proxy-authenticate|proxy-authorization)$/i
+
+function getRequestHeaders (ctx) {
```
mcollina, line 136: Can you add a comment explaining this?

ronag: Which one the HOP_EXPR or getRequestHeaders function?

mcollina: likely both

examples/proxy/index.js
```diff
@@ -0,0 +1,49 @@
...
+          if (res.headersSent) {
+            res.destroy(err)
+          } else {
+            for (const name of res.getHeaderNames()) {
+              res.removeHeader(name)
+            }
+            res.writeHead(err.statusCode || 500)
+            res.end()
+          }
+        })
```
ronag, line 25: Not having async doesn't help much here though...

ronag:
@mcollina: This is starting to feel pretty solid. I think for an example it is good.

If we want to include it as a helper I believe we need to in addition to test also add some hooks for modifying req/res headers and maybe also the response body. Not sure if informational headers and trailers are a must.

examples/proxy/proxy.js
```diff
@@ -0,0 +1,279 @@
...
+    headers = getHeaders({
+      headers,
+      proxyName: this.proxyName,
+      httpVersion: this.httpVersion
+    })
+    for (let n = 0; n < headers.length; n += 2) {
+      const key = headers[n + 0]
+      const val = headers[n + 1]
+
+      if (!Array.isArray(val)) {
```
ronag, line 132: This check is unnecessary. Will never be array.

examples/proxy/proxy.js
```diff
@@ -0,0 +1,279 @@
...
+      `for=${printIp(socket.remoteAddress, socket.remotePort)}`,
+      `proto=${socket.encrypted ? 'https' : 'http'}`,
+      `host=${printIp(authority || host || '')}`
+    ].join(';'))
+  } else if (forwarded) {
+    // The forwarded header should not be included in response.
+    throw new createError.BadGateway()
+  }
+
+  if (httpVersion && proxyName) {
```
ronag, line 221: Should do via check regardless of httpVersion.

examples/proxy/index.js
ronag: As far as I can see this cannot throw inside the catch handler.

mcollina (approved):
lgtm

Issue #432, opened 2020-09-23
Do we need to verify trailers?


ronag:
>    When a message includes a message body encoded with the chunked
>    transfer coding and the sender desires to send metadata in the form
>    of trailer fields at the end of the message, the sender SHOULD
>    generate a Trailer header field before the message body to indicate
>    which fields will be present in the trailers.  This allows the
>    recipient to prepare for receipt of that metadata before it starts
>    processing the body, which is useful if the message is being streamed
>    and the recipient wishes to confirm an integrity check on the fly.

The way I read the spec the server/client may send trailers that are not included in the `Trailers` header. However, how should one handle the case where the client/server does **not** send a specified trailer? i.e. to we need a "trailer mismatch error"?

ronag:
@szmarczak @mcollina

szmarczak:
I think no. I read `SHOULD` as `is highly recommended to`.

ronag:
> I think no. I read SHOULD as is highly recommended to, which means: optional.

Actually I'm referring to something else. The spec said that the request header should be included. It does not say anything about what to do if the request header includes a key that is not part of the trailers.

ronag:
Consider:

Ok
```js
res.writeHead(200, { Trailer: 'foo' })
res.writeTrailer('foo', 'asd')
res.writeTrailer('bar', 'asd')
res.end()
```

Not ok?


```js
res.writeHead(200, { Trailer: 'foo' })
res.writeTrailer('bar', 'asd')
res.end()
```

szmarczak:
Ah. Yeah, it makes sense to throw. IMO it's invalid.

olizilla:
Just a note that this causes an Error to be thrown when using undici to fetch from an IPFS Gateway API. 

That api always sends a `trailer: X-Stream-Error` header, but will only actually send that trailer if an error occurs mid-stream. Code that used to work is now failing when we switch to using undici. Originally reported here: https://github.com/web3-storage/web3.storage/issues/1017 by @flea89

Would you consider a PR to make absent trailers not be an error?

flea89:
Thanks, @olizilla for reporting this here.

@ronag, we'd really appreciate your view on this one? 
Happy to help.

ronag:
What does the spec say?

szmarczak:
https://httpwg.org/specs/rfc7230.html#rfc.section.4.1.2

It doesn't explicitly say the headers must be present, so I guess they can be omitted.

[...]

PR #433, opened 2020-09-23
fix: verify trailers

ronag:
Fixes: https://github.com/nodejs/undici/issues/432

### DIFF
--- a/lib/core/errors.js
+++ b/lib/core/errors.js
@@ -88,6 +88,16 @@ class ContentLengthMismatchError extends UndiciError {
   }
 }
 
+class TrailerMismatchError extends UndiciError {
+  constructor (message) {
+    super(message)
+    Error.captureStackTrace(this, TrailerMismatchError)
+    this.name = 'TrailerMismatchError'
+    this.message = message || /* istanbul ignore next */ 'Trailers does not match trailer header'
+    this.code = 'UND_ERR_TRAILER_MISMATCH'
+  }
+}
+
 class ClientDestroyedError extends UndiciError {
   constructor (message) {
     super(message)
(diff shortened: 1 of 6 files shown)


lib/core/errors.js
```diff
@@ -88,6 +88,16 @@ class ContentLengthMismatchError extends UndiciError {
   }
 }
 
+class TrailerMismatchError extends UndiciError {
+  constructor (message) {
+    super(message)
+    Error.captureStackTrace(this, TrailerMismatchError)
```
szmarczak, line 94: Is this really needed? Why?

ronag: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error

ronag: https://nodejs.org/api/errors.html#errors_error_capturestacktrace_targetobject_constructoropt

mcollina (approved):
lgtm

lib/core/errors.js
szmarczak: But `UndiciError` extends `Error`, so isn't `.stack` generated automatically?

ronag: I'm actually unsure why this is necessary but given the best practices I've been able to read it seems to be the correct way...

ronag: See the example from MDN:

```js
class CustomError extends Error {
  constructor(foo = 'bar', ...params) {
    // Pass remaining arguments (including vendor specific ones) to parent constructor
    super(...params)

    // Maintains proper stack trace for where our error was thrown (only available on V8)
    if (Error.captureStackTrace) {
      Error.captureStackTrace(this, CustomError)
    }

    this.name = 'CustomError'
    // Custom debugging information
    this.foo = foo
    this.date = new Date()
  }
}
```

Issue #434, opened 2020-09-24
Leaking array buffers


ronag:
Not sure if this is a user, undici, node or V8 bug but I have case where I have lot's of ArrayBuffers allocated and never freed in our Node based proxy:

```
{rss: 4519784448, heapTotal: 25522176, heapUsed: 20927688, external: 722839852, arrayBuffers: 721332770}
```

@mcollina @addaleax any hints on how to debug this?

mcollina:
Have you got a repro? You can possibly find them using heapdump.
If you need something programmatic to run on servers, use https://github.com/nearform/heap-profiler.

szmarczak:
@ronag I remember there was a bug in Node.js 12 related to TLS sockets. What Node.js version are you running?

ronag:
14.8

Happens at multiple locations with different node versions.

I see that V8 recently did something with GC of ArrayBuffers. Maybe coincidence...

ronag:
I’ll have to continue digging on Monday.

ronag:
> Have you got a repro? You can possibly find them using heapdump.
> If you need something programmatic to run on servers, use https://github.com/nearform/heap-profiler.

Does that work with cluster?

mcollina:
It might work given some adaptation.

ronag:
Seem to have the same problem on a totally different service which recently started to use undici

ronag:
Been trying to create a repro without success 😞

szmarczak:
@ronag How does the `undici` call look like? What function is used?

ronag:
`Client.request``

I think it might be broken back pressure.

ronag:
My current guess it that it has something to do with pause/resume multiple times. This seems to be growing arrayBuffer usage infinitiely:

```js
'use strict'

const { Client } = require('..')
const { createServer } = require('http')
const { Readable } = require('stream')

const requests = 128

const server = createServer()

const limit = 1e9

server.on('request', (req, res) => {
  let bytesWritten = 0

  const interval = setInterval(() => {
    console.log(process.memoryUsage().arrayBuffers / 1e6, bytesWritten, bytesWritten / limit)
  }, 1e3)

  const buf = Buffer.allocUnsafe(16384)
  new Readable({
    read () {
      bytesWritten += buf.length
      this.push(buf)
      if (bytesWritten >= limit) {
        clearInterval(interval)
        this.push(null)
      }
    }
  }).pipe(res)
})

server.listen(0, () => {
[...]

PR #435, opened 2020-09-24
fix: socket back pressure memory leak

ronag:
Fixes: https://github.com/nodejs/undici/issues/434

### DIFF
--- a/lib/core/client.js
+++ b/lib/core/client.js
@@ -555,7 +555,7 @@ class Parser extends HTTPParser {
 
     try {
       if (request.onBody(chunk, offset, length) === false) {
-        socket.pause()
+        socket[kPause]()
       }
     } catch (err) {
       util.destroy(socket, err)
@@ -630,7 +630,7 @@ class Parser extends HTTPParser {
         util.destroy(socket, new InformationalError('reset'))
       }
     } else {
-      socket.resume()
+      socket[kResume]()
       resume(client)
     }
   }
@@ -779,8 +779,8 @@ function connect (client) {
     parser.consume(socket._handle._externalStream)
   }
 
-  socket[kPause] = socket.pause.bind(socket)
-  socket[kResume] = socket.resume.bind(socket)
+  socket[kPause] = socketPause.bind(socket)
+  socket[kResume] = socketResume.bind(socket)
   socket[kError] = null
   socket[kParser] = parser
   socket[kClient] = client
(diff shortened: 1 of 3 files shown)


ronag:
@szmarczak

ronag:
@mcollina I would very much like to make a release out of this as soon as possible.

mcollina (approved):
lgtm

mcollina:
go for it!

szmarczak:
Is `socket[kBuffer]` supposed to be `null`?

ronag:
`socket[kBuffer]`?

szmarczak:
https://github.com/nodejs/node/blob/62c44af5c5f2964849bba92ff90c32e867d63f83/lib/net.js#L599-L605

ronag:
Hm, I don't understand how that works at the moment unfortunately.

ronag:
@mcollina Do we know someone that knows these internals well and we could ask for a quick review in regards to `consume` and socket.

ronag:
Seems like whatever `kBuffer` does it's opt-in. https://github.com/nodejs/node/pull/25436

szmarczak:
I think @mcollina was right here: https://github.com/nodejs/node/pull/25436/files#diff-e6ef024c3775d787c38487a6309e491dR602 Overriding Stream API sounds like a mistake.

szmarczak:
Why does it override `.read()` at all? Shouldn't it set `._read()` instead?

ronag:
> Why does it override .read() at all? Shouldn't it set ._read() instead?

What overrides it?

ronag:
> I think @mcollina was right here: nodejs/node#25436 (files) Overriding [public] Stream API sounds like a mistake.

Where is that comment?

szmarczak:
> Where is that comment?

`lib/net.js` after line 603

If you click the `(files)` link it should scroll down to the comment. It is marked as resolved so you need to click `Show conversation`.

<details>
<summary>Screenshot</summary>

image

</details>

> What overrides it?

I'm still talking about Node.js core (lib/net.js file) :sweat_smile: 

image

Shouldn't it be:

```diff
-Socket.prototype.read = function(n) {
+Socket.prototype._read = function(n) {
  if (this[kBuffer] && !this.connecting && this._handle &&
      !this._handle.reading) {
    tryReadStart(this);
  }
- return stream.Duplex.prototype.read.call(this, n);
};
```

ronag:
@szmarczak Yea, that's a bit weird... does it affect us though?

PR #436, opened 2020-09-24
fix: handle socket handle errors

ronag:
Not sure how/when these could actually error but I guess it's appropriate to handle.


szmarczak (approved):

mcollina:
Can you add a test?

ronag:
> Can you add a test?

I’m a little unsure how to trigger these errors... will have to keep trying

mcollina (approved):
lgtm

Issue #441, opened 2020-09-25
writableNeedDrain


ronag:
`Client.stream` needs to take [`writableNeedDrain`](https://github.com/nodejs/node/pull/35348) into consideration

PR #442, opened 2020-09-25
fix: Client.stream writableNeedDrain

ronag:
Fixes: https://github.com/nodejs/undici/issues/441
Refs: https://github.com/nodejs/node/pull/35348
Refs: https://github.com/nodejs/node/issues/35341

### DIFF
--- a/lib/core/client.js
+++ b/lib/core/client.js
@@ -375,6 +375,9 @@ class Parser extends HTTPParser {
     this.shouldKeepAlive = false
     this.request = null
     this.paused = false
+
+    // Parser can't be paused from within a callback.
+    // Use a buffer in JS land in order to stop further progress while paused.
     this.resuming = false
     this.queue = []
 
--- a/lib/core/client.js
+++ b/lib/core/client.js
@@ -375,19 +375,29 @@ class Parser extends HTTPParser {
     this.shouldKeepAlive = false
     this.request = null
     this.paused = false
+    this.resuming = false
     this.queue = []
 
     this._resume = () => {
+      if (this.resuming) {
+        return
+      }
+
       this.paused = false
 
-      while (this.queue.length) {
-        const [fn, ...args] = this.queue.shift()
+      this.resuming = true
+      try {
+        while (this.queue.length) {
+          const [fn, ...args] = this.queue.shift()
 
-        fn.apply(this, args)
+          Reflect.apply(fn, this, args)
 
-        if (this.paused) {
-          return
+          if (this.paused) {
+            return
+          }
         }
+      } finally {
+        this.resuming = false
       }
 
       socketResume(socket)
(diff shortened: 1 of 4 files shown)


ronag:
Missing tests.

ronag:
@mcollina: PTAL

lib/core/client.js
```diff
@@ -376,9 +375,37 @@ class Parser extends HTTPParser {
     this.shouldKeepAlive = false
     this.read = 0
     this.request = null
+    this.paused = false
+    this.queue = []
```
ronag, line 379: We can't pause the parser from within a callback so we need to implement a buffer in JS land in order to stop further progress while paused.

ronag:
@indutny @addaleax I'm working around the fact that I can't pause the parser in a callback. Do you have any better suggestions on how to go about that?

mcollina (changes requested):
I don't understand this fix. Why is this needed at all?

(A slightly better implementation would go and replace the kOn* methods when paused to avoid the ifs on an hot path).

ronag:
> I don't understand this fix. Why is this needed at all?

Because currently when pausing we don't actually pause. It's more of a hint.

> A slightly better implementation would go and replace the kOn* methods when paused to avoid the ifs on an hot path

I think this is negligable.

ronag:
I guess it's not entirely necessary... though the current implementation isn't strictly correct. The docs would need to be updated to reflect that returning `false` from `onBody` is only a hint.

mcollina:
Have you benchmarked this?

ronag:
Yes, difference is within margin of error.

ronag:
I've cleaned this up a bit and also added more assertions to make it clearer what this fixes.

lib/core/client.js
mcollina: Can you add a comment so it's clear to future readers?

test/client-stream.js
```diff
@@ -657,3 +657,46 @@ test('stream body destroyed on invalid callback', (t) => {
...
+
+test('stream needDrain', (t) => {
+  t.plan(1)
+
+  const server = createServer((req, res) => {
+    res.end(Buffer.alloc(4096))
+  })
+  t.tearDown(server.close.bind(server))
+
+  server.listen(0, async () => {
```
mcollina, line 669: Don't pass async here, it encourages bad practices about promises.

test/client-stream.js
```diff
@@ -657,3 +657,46 @@ test('stream body destroyed on invalid callback', (t) => {
...
+    }, () => {
+      return dst
+    })
+
+    setTimeout(() => {
+      dst.write = (...args) => {
+        orgWrite.call(dst, ...args)
+      }
+      dst.resume()
+    }, 1e3)
```
mcollina, line 696: is 1s needed?

mcollina (approved):
LGTM

PR #449, opened 2020-10-07
fix: node 12.9

ronag:


mcollina (approved):
lgtm, but it would be good to have a test for this

ronag:
Existing tests fail with node 12.9

PR #453, opened 2020-10-09
fix(core/request): handle header set to null

### LINKED ISSUE #416: Header set to `null` validation
I have a use case where I get a header set to `null` and [this](https://github.com/nodejs/undici/blob/9365968fd3cbcbf13c83c6b62e691bdd57c5e163/lib/core/request.js#L159) check fails. Can `null` be considered as `undefined` so that this check can be modified, or should I handle this separately? My gut tells me probably it's better separately, maybe because of streams behavior.

dnlup:
If a header is set to null it is casted to the string 'null'.

This is the same behavior we have in Node's HTTP client if I am not mistaken.

Note: I am not sure this is a fix, though. I can amend the commit :wink: 

Fix #416

### DIFF
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -155,7 +155,7 @@ class Request {
 }
 
 function processHeader (request, key, val) {
-  if (typeof val === 'object') {
+  if (typeof val === 'object' && val !== null) {
     throw new InvalidArgumentError(`invalid ${key} header`)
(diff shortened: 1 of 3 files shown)


dnlup:
One other thing. Are benchmarks working? It looks like they exit immediately. I tried to run them locally also.

mcollina (approved):
lgtm

PR #454, opened 2020-10-09
test: support earliest versions in range in CI

bengl:
I'm attempting to use `undici` in a library that supports entire semver major ranges of Node.js. I found that on Windows with Node 12.0.0 there was an abort when used with my library's test suite.

This is an attempt to establish an explicit baseline version range to support.

**EDIT:** I tested locally and the minimum versions included now are the minimum ones that would pass on my machine.


mcollina (approved):
lgtm

PR #457, opened 2020-10-10
fix(benchmarks): fix timeout errors

### LINKED ISSUE #456: Missing results on benchmarks in Node 10 and 12
Hi :wave: :smile: 

I am getting timeout errors when running benchmarks on Node 12 and 10, resulting in a premature exit without results. This doesn't happen on Node 14. Switching to the use of the `finish` event instead of `close` fixes this issue. Is this behavior wanted?

dnlup:
The use of `close` event was causing timeouts. Switching to the `finish` event fixed it.

Fixes #456

### DIFF
--- a/benchmarks/index.js
+++ b/benchmarks/index.js
@@ -53,7 +53,7 @@ suite
                 callback()
               }
             }))
-            .on('close', resolve)
+            .on('finish', resolve)
         })
       }))).then(() => deferred.resolve())
     }
@@ -72,7 +72,7 @@ suite
               callback()
             }
           }))
-          .on('close', resolve)
+          .on('finish', resolve)
       }))).then(() => deferred.resolve())
     }
   })
@@ -89,7 +89,7 @@ suite
                   callback()
                 }
               }))
-              .on('close', resolve)
+              .on('finish', resolve)
           })
       }))).then(() => deferred.resolve())
     }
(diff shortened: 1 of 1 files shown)


benchmarks/index.js
```diff
@@ -37,6 +37,16 @@ client.on('disconnect', (err) => {
...
+// http.get(httpOptions, (res) => {
+//   res
+//     .pipe(new Writable({
+//       write (chunk, encoding, callback) {
+//         console.log(`${chunk}`)
+//         callback()
+//       }
+//     }))
+//     .on('finish', () => console.log('done'))
+// })
```
mcollina, line 49: Why this is here?

mcollina (commented):
Are these running well on all versions of Node.js now?

benchmarks/index.js
dnlup: Sorry, I removed it. It was trash I forgot to remove.

dnlup:
> Are these running well on all versions of Node.js now?

No, I think we should add the matrix. If you agree I'll work on it.

dnlup:
> Are these running well on all versions of Node.js now?

Sorry, I misunderstood. Yes, they run well on 10, 12, and 14. Do we want to add a matrix for benchmarks too?

dnlup:
Maybe the same matrix used in tests. I am not sure if it might be too much.

mcollina (approved):
lgtm

mcollina (approved):
lgtm

PR #458, opened 2020-10-23
Fix for node 12.19

mcollina:
Unfortunately the private API of the HTTP parser changed in 12.19.0. It would crash without this fix.


mcollina:
@ronag could you take a look asap? This is broken without this change.

ronag:
@mcollina can you make a release? I’m not by my computer for a while.

PR #462, opened 2020-10-26
fix: simpler and faster idle timeout

ronag:
Avoid using socket timeout. We don't need to refresh the timer on every chunk.


mcollina (approved):
lgtm

Issue #463, opened 2020-10-30
HTTP host header is missing the port


antonnyman:
Currently the host header is set to `hostname` which could cause errors if you use an endpoint that has a port, for example `http://localhost:1234`. The current workaround is to manually set the host header with a port.

PR #464, opened 2020-10-30
Set hostname to host

antonnyman:
https://github.com/nodejs/undici/issues/463

### DIFF
--- a/lib/core/client.js
+++ b/lib/core/client.js
@@ -159,7 +159,7 @@ class Client extends EventEmitter {
       }
     }
     this[kTLSSession] = null
-    this[kHostHeader] = `host: ${this[kUrl].hostname}\r\n`
+    this[kHostHeader] = `host: ${this[kUrl].host}\r\n`
 
     // kQueue is built up of 3 sections separated by
     // the kRunningIdx and kPendingIdx indices.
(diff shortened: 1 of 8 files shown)


ronag (approved):

lib/core/client.js
```diff
@@ -159,7 +159,7 @@ class Client extends EventEmitter {
       }
     }
     this[kTLSSession] = null
-    this[kHostHeader] = `host: ${this[kUrl].hostname}\r\n`
+    this[kHostHeader] = `host: ${this[kUrl].host}\r\n`
```
ronag, line 162: const host = this[kUrl].host ? this[kUrl].host : `${this[kUrl].hostname}${this[kUrl].port ? `:${this[kUrl].port}`}`

ronag (changes requested):

lib/core/client.js
ronag: Also probably an invalid arugment check in constructor to ensure that either host or hostname is set and a string

mcollina (approved):
lgtm

ronag:
Landed with some minor compat fixes in 908be7db7781579ae327a144509a483a8b28f728

PR #446, opened 2020-10-05
core/request.js: http proxies support

sergeysedoy97:
To support http proxies path must accept url.

e.g.
```
GET http://example.com/ HTTP/1.1
```

PR #468, opened 2020-11-02
fix: accept absolute urls as path

szmarczak:
Fixes #446

### DIFF
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -17,6 +17,10 @@ const kHandler = Symbol('handler')
 const kPaused = Symbol('paused')
 const kAbort = Symbol('abort')
 
+// Borrowed from https://gist.github.com/dperini/729294
+// eslint-disable-next-line no-control-regex
+const REGEXP_ABSOLUTE_URL = /^(?:(?:https?|ftp):\/\/)(?:\S+(?::\S*)?@)?(?:(?!10(?:\.\d{1,3}){3})(?!127(?:\.\d{1,3}){3})(?!169\.254(?:\.\d{1,3}){2})(?!192\.168(?:\.\d{1,3}){2})(?!172\.(?:1[6-9]|2\d|3[0-1])(?:\.\d{1,3}){2})(?:[1-9]\d?|1\d\d|2[01]\d|22[0-3])(?:\.(?:1?\d{1,2}|2[0-4]\d|25[0-5])){2}(?:\.(?:[1-9]\d?|1\d\d|2[0-4]\d|25[0-4]))|(?:(?:[a-z\x00a1-\xffff0-9]+-?)*[a-z\x00a1-\xffff0-9]+)(?:\.(?:[a-z\x00a1-\xffff0-9]+-?)*[a-z\x00a1-\xffff0-9]+)*(?:\.(?:[a-z\x00a1-\xffff]{2,})))(?::\d{2,5})?(?:\/[^\s]*)?$/ius
+
 class Request {
   constructor ({
     path,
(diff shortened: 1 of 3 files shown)


lib/core/request.js
```diff
@@ -28,8 +28,15 @@ class Request {
...
     bodyTimeout
   }, handler) {
-    if (typeof path !== 'string' || path[0] !== '/') {
-      throw new InvalidArgumentError('path must be a valid path')
+    if (typeof path !== 'string') {
+      throw new InvalidArgumentError('path must be a string')
+    } else if (path[0] !== '/') {
+      try {
+        // eslint-disable-next-line no-new
+        new URL(path)
```
szmarczak, line 36: In Got we also do `decodeURI(absoluteUrl)`.

ronag: Not sure I like the `new URL` throw check but I can't really argue that it's a bad idea either. I'm good with it if @mcollina approves.

mcollina: This is going to be terribly slow :/. What's the use case of having a full url here?

szmarczak: @mcollina https://en.wikipedia.org/wiki/Proxy_server#Web_proxy_servers

I don't like `RegExp`s either. In some cases they can be very slow as well.

szmarczak: Node.js uses `new URL`: https://github.com/nodejs/node/blob/dfc288e7fdbe888f4e134772de421684247f0bea/lib/_http_client.js#L95

szmarczak: image

It still seems that RegExp is faster but it doesn't look as nice as `new URL`... :thinking:

szmarczak: <details>
<summary>The entire RegExp here</summary>

```js
/^(?:(?:https?|ftp):\/\/)(?:\S+(?::\S*)?@)?(?:(?!10(?:\.\d{1,3}){3})(?!127(?:\.\d{1,3}){3})(?!169\.254(?:\.\d{1,3}){2})(?!192\.168(?:\.\d{1,3}){2})(?!172\.(?:1[6-9]|2\d|3[0-1])(?:\.\d{1,3}){2})(?:[1-9]\d?|1\d\d|2[01]\d|22[0-3])(?:\.(?:1?\d{1,2}|2[0-4]\d|25[0-5])){2}(?:\.(?:[1-9]\d?|1\d\d|2[0-4]\d|25[0-4]))|(?:(?:[a-z\x00a1-\xffff0-9]+-?)*[a-z\x00a1-\xffff0-9]+)(?:\.(?:[a-z\x00a1-\xffff0-9]+-?)*[a-z\x00a1-\xffff0-9]+)*(?:\.(?:[a-z\x00a1-\xffff]{2,})))(?::\d{2,5})?(?:\/[^\s]*)?$/ius
```

https://mathiasbynens.be/demo/url-regex `@diegoperini (502 chars)`

</details>

ronag: Do it need to be that complicated? Isn't it enough with:

`(/^https?:\/)?\//`

szmarczak: Performance-wise that's great. Not so complicated but allows silly mistakes like `https://space here.com`

szmarczak: I'm good with any, it's up to you :)

ronag: @mcollina wdyt?

mcollina: let's go for the mastodontic regexp

szmarczak: Eventually we can use `@^(https?|ftp)://[^\s/$.?#].[^\s]*$@iS` (`@stephenhay` version) which passes on all tested URLs but allows some mistakes like `http://1.1.1.1.1`.

ronag (approved):

mcollina (approved):
lgtm

Issue #470, opened 2020-11-02
Retry of idempotent request with body broken


ronag:
We clear the body after sending it to the socket. Hence the retry will done incorrectly without a body.

PR #471, opened 2020-11-02
fix: retry idempotent with original body

ronag:
Fixes:https://github.com/nodejs/undici/issues/470


benjamingr:
The fixes is probably wrong?

ronag:
Updated ref

mcollina (approved):
lgtm

PR #472, opened 2020-11-02
Update abort docs to native AbortController

benjamingr:
Noticed this very randomly ^^

We have native abortcontroller so it can be used.

Also - a few of the places used `AbortController` for signals incorrectly.

Cheers

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -13,7 +13,7 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [10.16.0, 10.x, 12.3.0, 12.18.0, 12.x, 14.0.0, 14.x]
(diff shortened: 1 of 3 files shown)


ronag (approved):

mcollina (commented):
Could you add a test with the native thing for Node v15?

benjamingr:
@mcollina happy to, here or in a different non-docs PR?

mcollina:
I would prefer here tbh, so we test what we write in the docs.

benjamingr:
@mcollina done, PTAL, note that I am not familiar with the CI in unidici (and whether or not it runs on Node 15 which has AbortController). I tested locally (tap) in Node 14 and Node 15.

mcollina:
You need to add Node v15 to .github/workflows/nodejs.yml

mcollina (approved):
lgtm

Issue #388, opened 2020-08-30
Arg check `handler` in `dispatch(opts, handler)`


ronag:
Currently it is assumed that all handler methods are implemented and no check is done for whether e.g. `handler.onError` is a function.

Sirius79:
Hey! I would like to take this up if no one is working on it

ronag:
Go for it!

PR #473, opened 2020-11-05
add args checks for dispatch handler

Sirius79:
Fixes: #388

### DIFF
--- a/lib/core/client.js
+++ b/lib/core/client.js
@@ -227,6 +227,10 @@ class Client extends EventEmitter {
   }
 
   dispatch (opts, handler) {
+    if (typeof handler.onError !== 'function') {
+      throw new InvalidArgumentError('invalid onError method')
+    }
+
     try {
       const request = new Request(opts, handler)
       if (this[kDestroyed]) {
--- a/lib/core/client.js
+++ b/lib/core/client.js
@@ -227,10 +227,6 @@ class Client extends EventEmitter {
   }
 
   dispatch (opts, handler) {
-    if (typeof handler.onError !== 'function') {
-      throw new InvalidArgumentError('invalid onError method')
-    }
-
     try {
       const request = new Request(opts, handler)
       if (this[kDestroyed]) {
@@ -252,6 +248,10 @@ class Client extends EventEmitter {
         resume(this, true)
       }
     } catch (err) {
+      if (typeof handler.onError !== 'function') {
+        throw new InvalidArgumentError('invalid onError method')
+      }
+
       handler.onError(err)
     }
   }
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -93,6 +93,26 @@ class Request {
       throw new InvalidArgumentError('headers must be an object or an array')
     }
 
+    if (typeof handler.onConnect !== 'function') {
+      throw new InvalidArgumentError('invalid onConnect method')
+    }
+
+    if (typeof handler.onHeaders !== 'function') {
+      throw new InvalidArgumentError('invalid onHeaders method')
+    }
+
+    if (typeof handler.onData !== 'function') {
+      throw new InvalidArgumentError('invalid onData method')
+    }
+
+    if (typeof handler.onComplete !== 'function') {
+      throw new InvalidArgumentError('invalid onComplete method')
+    }
+
+    if (typeof handler.onError !== 'function') {
+      throw new InvalidArgumentError('invalid onError method')
+    }
+
     this[kPaused] = null
     this[kHeadersTimeout] = headersTimeout != null ? headersTimeout : 30e3
     this[kBodyTimeout] = bodyTimeout != null ? bodyTimeout : 30e3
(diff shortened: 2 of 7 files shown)


Sirius79:
@ronag does this look like the right approach?

lib/core/client.js
```diff
@@ -227,6 +227,10 @@ class Client extends EventEmitter {
   }
 
   dispatch (opts, handler) {
+    if (typeof handler.onError !== 'function') {
+      throw new InvalidArgumentError('invalid onError method')
+    }
```
ronag, line 232: All handler methods are required.

lib/core/request.js
```diff
@@ -185,6 +201,10 @@ class Request {
     clearTimeout(this[kTimeout])
     this[kTimeout] = null
 
+    if (typeof this[kHandler].onComplete !== 'function') {
+      throw new InvalidArgumentError('invalid onComplete method')
+    }
+
     this[kHandler].onComplete(trailers)
   }
```
ronag, line 210: I don't think you need to check here. Just check in the request constructor.

ronag (changes requested):
Please put the check in the constructor and also try to avoid changing existing tests.

lib/core/request.js
Sirius79: Alright. I'll do the necessary changes and update the PR

Sirius79: So I have reverted the changes to the existing tests. But it seems like moving the check to the request constructor is breaking a few tests as some of the handlers don't have all the methods (like `onComplete` ).

Sirius79: @ronag I have updated the PR but I'm not sure how to proceed with the check in the request constructor. Can you please help here?

ronag: Put the checks in the constructor and fix the tests that fail due to missing handlers?

Sirius79: Okay. I was under the impression that it wasn't necessary for all the handler methods to be implemented. :sweat_smile:

mcollina (approved):
lgtm

lib/core/request.js
Sirius79: @ronag , it seems that @mcollina has already approved this. Are the changes to the constructor still required?

If yes, the failing tests would need a change in them. But as you suggested earlier, we should avoid changing existing tests. 

Also, I would like a bit more clarity. If I add it to the constructor, it would become mandatory to implement all the handler methods. Is this what is intended?

ronag: > Are the changes to the constructor still required?

Yes

> If yes, the failing tests would need a change in them. But as you suggested earlier, we should avoid changing existing tests.

Try to avoid but if it's necessary we should change them.

> Is this what is intended?

Yes

Sirius79: I have put the checks in the constructor and fixed the tests.

test/pool.js
```diff
@@ -425,6 +425,9 @@ test('pool dispatch', (t) => {
       },
       onComplete () {
         t.strictEqual(buf, 'asd')
+      },
+      onError (err) {
+        t.strictEqual(err.code, 'UND_ERR_INVALID_ARG')
```
ronag, line 430: This breaks the test. It's testing something different now.

test/pool.js
```diff
@@ -559,6 +562,9 @@ test('pool dispatch error', (t) => {
       },
       onComplete () {
         t.pass()
+      },
+      onError (err) {
+        t.strictEqual(err.code, 'UND_ERR_INVALID_ARG')
```
ronag, line 567: This breaks the test. It's testing something different now.

test/pool.js
```diff
@@ -613,6 +619,9 @@ test('pool request abort in queue', (t) => {
...
```
ronag, line 624: This breaks the test. It's testing something different now.

test/pool.js
```diff
@@ -656,6 +665,9 @@ test('pool stream abort in queue', (t) => {
...
```
ronag, line 670: This breaks the test. It's testing something different now.

test/pool.js
```diff
@@ -699,6 +711,9 @@ test('pool pipeline abort in queue', (t) => {
...
```
ronag, line 716: This breaks the test. It's testing something different now.

test/pool.js
Sirius79: Fixed this

test/pool.js
Sirius79: Fixed this

test/pool.js
Sirius79: Fixed this

test/pool.js
Sirius79: Fixed this

test/pool.js
Sirius79: Fixed this

ronag (approved):
lgtm, thanks!

PR #478, opened 2020-11-12
fix: null is not valid header value

ronag:


ronag:
strangely enough we support null header values per https://github.com/nodejs/undici/pull/453

PR #479, opened 2020-11-13
fix: check signal.aborted

ronag:


lib/abort-signal.js
```diff
@@ -24,6 +24,10 @@ function addSignal (self, signal) {
   } else {
     self[kSignal].addListener('abort', self[kListener])
   }
+
+  if (signal.aborted) {
+    self[kListener]()
```
benjamingr, line 29: You _probably_ want to not add the signal and just call the listener? Is the signal correctly removed in this case?

benjamingr (approved):

lib/abort-signal.js
ronag: Fixed

mcollina (approved):
lgtm

PR #482, opened 2020-11-17
fix AbortController types

### LINKED ISSUE #480: undici installed in a TypeScript project errors with `...types/client.d.ts:5:33 - error TS2307: Cannot find module 'abort-controller' or its corresponding type declarations.`
When updgrading from 2.1.1 to [2.2.0](https://github.com/nodejs/undici/releases/tag/v2.2.0) I got this error in TypeScript:
```
> tsc -d

../../node_modules/.pnpm/undici@2.2.0/node_modules/undici/types/client.d.ts:5:33 - error TS2307: Cannot find module 'abort-controller' or its corresponding type declarations.

5 import { AbortController } from 'abort-controller'
                                  ~~~~~~~~~~~~~~~~~~

Found 1 error.
```

The source should be this line https://github.com/nodejs/undici/blob/master/types/client.d.ts#L5

So here the `abort-controller` package is missing and because TypeScript cannot find it, it errors.
Meaning that it needs to be installed / be a required dependency.

I added the package as a dev dependency in my project and it worked 😃 

Maybe this would need to be documented or the dependency required? (Note: I am not using abort-controller functionality here).

Note: Many Thanks for the types! 💚

Ethan-Arrowood:
Closes #480 

Also adds some missing semicolons (probably should add a linter to ts definitions eventually)

### DIFF
--- a/types/client.d.ts
+++ b/types/client.d.ts
@@ -2,8 +2,17 @@ import { URL } from 'url'
 import { TlsOptions } from 'tls'
 import { Duplex, Readable, Writable } from 'stream'
 import { EventEmitter } from 'events'
-import { AbortController } from 'abort-controller'
 import { IncomingHttpHeaders } from 'http'
+
+/**
+ * Issue https://github.com/nodejs/undici/issues/480
+ * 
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

PR #490, opened 2020-11-26
fix: fallback to url as servername

ronag:
Fails on certain endpoints. Not sure why and wasn't able to create a test. The endpoint I was able to reproduce with is unfortunately private.

### DIFF
--- a/lib/core/client.js
+++ b/lib/core/client.js
@@ -57,6 +57,15 @@ const nodeMajorVersion = parseInt(nodeVersions[0].slice(1))
 const nodeMinorVersion = parseInt(nodeVersions[1])
 const insecureHTTPParser = process.execArgv.includes('--insecure-http-parser')
 
+function getServerName (client, host) {
+  return (
+    util.getServerName(host) ||
+    (client[kTLSOpts] && client[kTLSOpts].servername) ||
+    util.getServerName(client[kUrl].host || client[kUrl].hostname) ||
+    null
+  )
+}
+
 class Client extends EventEmitter {
   constructor (url, {
     maxHeaderSize,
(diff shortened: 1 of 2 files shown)


mcollina:
Can you add a test?

ronag:
I tried without success. Can try again next week. Not sure exactly by it fails on some servers while it’s fine on others.

szmarczak:
Some servers just reject the connection when there is no servername given. The only place where you set `servername` is here:

https://github.com/nodejs/undici/blob/1ba80a7a6e5e97d6d67725fa463483f09176d310/lib/core/client.js#L943-L965

which requires `client.running` to be `0`. `_resume(...)` is called by [global `resume(...)`](https://github.com/nodejs/undici/blob/1ba80a7a6e5e97d6d67725fa463483f09176d310/lib/core/client.js#L870).

Why just not set the servername here?

https://github.com/nodejs/undici/blob/1ba80a7a6e5e97d6d67725fa463483f09176d310/lib/core/client.js#L148

szmarczak:
My guess is that you don't initialize `servername` at all (please correct me if I'm wrong). Node.js does this:

https://github.com/nodejs/node/blob/6eec858f34a40ffa489c1ec54bb24da72a28c781/lib/_http_agent.js#L240-L241

ronag:
> Why just not set the servername here?

Because when `request.host` goes back to `null` it needs to be reset.

ronag:
> My guess is that you don't initialize servername at all (please correct me if I'm wrong). Node.js does this:

Yep. My problem at the moment is how to setup a test, i.e. create a server that denies connections without servername.

ronag:
@mcollina I'm not getting further on making a test for this. Are you fine with landing without?

mcollina:
Sure as long as the original test suite passes.

PR #496, opened 2020-12-15
fix: simplify timeouts

ronag:
body and headers timeouts live in the parser and not on the request. Makes the implementation cleaner and simpler.


mcollina (approved):
lgtm

PR #498, opened 2020-12-15
fix: destroy socket on active request abort

ronag:


mcollina (commented):
I would prefer a "fix" to have a test

ronag:
Giving this more though this is wrong.

PR #502, opened 2020-12-30
Test WeakRef support

### LINKED ISSUE #501: WeakRef support
We should test that a `WeakRef`:d `Pool` or `Client` does not get garbage collected while there is an active socket.

So basically if the pool/client has no pending request and idle/keepalive timeout has expired it should be garbage collected, otherwise not.

I think this already works but needs a regression test.

`Client` has a ref to `Pool` through `onDrain` listener and `Socket` has a ref to `Client` through `[kClient]`.

I think sockets are GC roots so I believe they will never be GC:d while there is still an active connection.

dnlup:
Add a test to check [`WeakRef`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WeakRef) support. 

I wasn't able to trigger the gc cycle any faster, I had to increase the default timeout of `tap`.

Fixes #501

### DIFF
--- a/test/gc.js
+++ b/test/gc.js
@@ -1,0 +1,100 @@
+'use strict'
+/* global WeakRef, FinalizationRegistry */
+
+const { test } = require('tap')
+const { createServer } = require('net')
+const { Client, Pool } = require('..')
+
+const SKIP = typeof WeakRef === 'undefined' || typeof FinalizationRegistry === 'undefined'
+
+test('gc should collect the client if, and only if, there are no active sockets', { skip: SKIP }, t => {
+  t.plan(4)
+
+  const server = createServer((socket) => {
+    socket.write('HTTP/1.1 200 OK\r\n')
+    socket.write('Content-Length: 0\r\n')
+    socket.write('Keep-Alive: timeout=1s\r\n')
+    socket.write('Connection: keep-alive\r\n')
+    socket.write('\r\n\r\n')
+  })
+  t.teardown(server.close.bind(server))
+
+  let weakRef
+  let disconnected = false
+
+  const registry = new FinalizationRegistry((data) => {
+    t.strictEqual(data, 'test')
+    t.strictEqual(disconnected, true)
+    t.strictEqual(weakRef.deref(), undefined)
+  })
+
+  server.listen(0, () => {
+    const client = new Client(`http://localhost:${server.address().port}`, {
+      keepAliveTimeoutThreshold: 100
+    })
+    client.once('disconnect', () => {
+      disconnected = true
+      setImmediate(() => {
+        global.gc()
+      })
+    })
+
+    weakRef = new WeakRef(client)
+    registry.register(client, 'test')
+
+    client.request({
+      path: '/',
+      method: 'GET'
+    }, (err, { body }) => {
+      t.error(err)
+      body.resume()
+    })
+  })
+})
+
+test('gc should collect the pool if, and only if, there are no active sockets', { skip: SKIP }, t => {
+  t.plan(4)
+
+  const server = createServer((socket) => {
+    socket.write('HTTP/1.1 200 OK\r\n')
+    socket.write('Content-Length: 0\r\n')
+    socket.write('Keep-Alive: timeout=1s\r\n')
+    socket.write('Connection: keep-alive\r\n')
+    socket.write('\r\n\r\n')
+  })
+  t.teardown(server.close.bind(server))
+
+  let weakRef
+  let disconnected = false
+
+  const registry = new FinalizationRegistry((data) => {
+    t.strictEqual(data, 'test')
+    t.strictEqual(disconnected, true)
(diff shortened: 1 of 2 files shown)


test/gc.js
```diff
@@ -0,0 +1,55 @@
...
+    wrapper.request({
+      path: '/',
+      method: 'GET'
+    }, (err, { body }) => {
+      t.error(err)
+      body.once('end', () => {
+        clock.tick(600)
+      }).resume()
+    })
+  }
```
ronag, line 47: Maybe something like:

```js

t.plan(3)

const weakRef = new WeakRef(client)
let disconnected = false

const registry = new FinalizationRegistry((name) => {
  t.strictEqual(disconnected, true)
  t.strictEqual(name, 'test')
  t.strictEqual(weakRef.deref(), undefined)
})

client.once('disconnect', () => {
  disconnected = true
})

registry.register(client, 'test')

wrapper.request({
  path: '/',
  method: 'GET'
}, (err, { body }) => {
  t.error(err)
  body.once('end', () => {
    clock.tick(600)
  }).resume()
})
```

dnlup: It looks like it could be the fake clock

dnlup: Manually spinning the fake clock seems to trigger a GC operation. The test is taking more than 10 seconds, though.

mcollina:
Have you tried by exposing the gc and calling it?

dnlup:
> Have you tried by exposing the gc and calling it?

Yes, but the thing that is making the difference seems to be the clock ticking. If I don't do this nothing happens, even if I trigger the `gc` manually.

test/gc.js
```diff
@@ -0,0 +1,58 @@
...
+    t.strictEqual(weakRef.deref(), undefined)
+  })
+
+  server.listen(0, () => {
+    const client = new Client(`http://localhost:${server.address().port}`, {
+      keepAliveTimeoutThreshold: 500
+    })
+    client.once('disconnect', () => {
+      disconnected = true
+      clock.tick(1e6)
```
ronag, line 42: Why is this tick needed?

dnlup: This is the tick that allows for the `registry` callback to be called. Could it be that the fake clock interferes with the `gc` activity?

ronag: Can you try without a fake clock?

dnlup: The test remains stuck waiting for the callback without the tick, even if I call the `gc` manually. Maybe I'm missing something.

dnlup: > Can you try without a fake clock?

Yes, let's try without the fake clock.

test/gc.js
```diff
@@ -0,0 +1,58 @@
...
+/* global WeakRef, FinalizationRegistry */
+
+const { test } = require('tap')
+const { Client } = require('..')
+const { createServer } = require('net')
+const FakeTimers = require('@sinonjs/fake-timers')
+
+const SKIP = typeof WeakRef === 'undefined' || typeof FinalizationRegistry === 'undefined'
+
+test('gc should collect the client if there are no active sockets', { skip: SKIP }, t => {
```
ronag, line 11: client if, and only if, there

test/gc.js
dnlup: Removed the fake clock and triggered `gc` manually. The timing is similar to before.

test/gc.js
```diff
@@ -0,0 +1,101 @@
...
+    t.strictEqual(weakRef.deref(), undefined)
+  })
+
+  server.listen(0, () => {
+    const pool = new Pool(`http://localhost:${server.address().port}`, {
+      connections: 1,
+      keepAliveTimeoutThreshold: 500
+    })
+
+//    pool.once('disconnect', () => {
```
dnlup, line 83: To test the same way we did for the `Client`, maybe the `Pool` should emit a `disconnect` event when all clients have disconnected, wdyt?

ronag: I think pool should emit connect/disconnect with the client as first argument, i.e. simply forwarding the client events.

dnlup: Ok. I think it would be better to open a separate issue/PR for this and then rebase the work we are doing here.

Ethan-Arrowood (approved):
lgtm - strange that its failing on one ci though

dnlup:
> lgtm - strange that its failing on one ci though

I know 😢 . I can't test macOS locally, unfortunately. I want to be sure these tests are not flaky before merging, though.

Ethan-Arrowood:
Just pulled locally - gc tests pass on my mac 😄

mcollina (approved):
lgtm

ronag:
Unfortunately we can't merge this if the CI is flaky :/

mcollina:
> Unfortunately we can't merge this if the CI is flaky :/. Could this be a GC bug?

Unfortunately CI is flaky independently of this. I would recommend landing.

Issue #504, opened 2021-01-01
Pool helper


ronag:
Not sure how to make this part of the library but this is a pattern I've started to use:

```js
const getClient = weakCache(origin => new undici.Pool(origin, {
  connections: 16,
  pipelining: 2,
  headersTimeout: 10e3,
  bodyTimeout: 10e3
})

function request(url, { method = 'GET', headers, body, signal }) {
  url = typeof url === 'string' ? new URL(url) : url
  return getClient(url.origin).request({
     method,
     path: url.path || `${url.pathname || ''}${url.search || ''}`,
     headers,
     body,
     signal
  })
}

function weakCache (fn) {
  const cache = new Map()
  const finalizationRegistry = new FinalizationRegistry(key => {
    const ref = cache.get(key)
    if (ref !== undefined && ref.deref() === undefined) {
      cache.delete(key)
    }
  })
  return (key) => {
    const ref = cache.get(key)
    if (ref !== undefined) {
      const deref = ref.deref()
      if (deref !== undefined) {
        return deref
      }
    }
    const val = fn(key)
    cache.set(key, new WeakRef(val))
    finalizationRegistry.register(val, key)
    return val
  }
}
```

ronag:
@Ethan-Arrowood maybe interested?

ronag:
Maybe `SharedPool` or something?

Ethan-Arrowood:
Yes very interested in this. I think this could help undici-fetch too. Will take a closer look soon.

ronag:
sketching...

```js
const globalAgent = new Agent({
  connections: 16,
  pipelining: 2,
  headersTimeout: 10e3,
  bodyTimeout: 10e3
})

function request(url, opts) {
  url = typeof url === 'string' ? new URL(url) : url
  const agent = opts.agent ?? globalAgent
  return agent.get(url.origin).request({
    ...opts,
     method: opts.method || 'GET',
     path: url.path || `${url.pathname || ''}${url.search || ''}`,
  })
}

module.exports = {
  request
}

class Agent {
  #factory = null

  constructor (opts) {
    this.#factory = weakCache(origin => new undici.Pool(origin, opts))
  }

  get(origin) {
    return this.$factory(origin)
  }
}

function weakCache (fn) {
  const cache = new Map()
  const finalizationRegistry = new FinalizationRegistry(key => {
    const ref = cache.get(key)
    if (ref !== undefined && ref.deref() === undefined) {
      cache.delete(key)
    }
  })
[...]

PR #508, opened 2021-01-05
init agent implementation + test

Ethan-Arrowood:
ref #504 

heres a draft pr for some of the ideas in the ref issue. going to add more tests before publishing + im having some issue getting standard to recognize WeakRef and FinalizationRegistry

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,7 +3,7 @@
 const Client = require('./lib/core/client')
 const errors = require('./lib/core/errors')
 const Pool = require('./lib/pool')
-const { Agent, request, stream, pipeline, setGlobalAgent } = require('./lib/agent')
+// const { Agent, request, stream, pipeline, setGlobalAgent } = require('./lib/agent')
 
 Client.prototype.request = require('./lib/client-request')
 Client.prototype.stream = require('./lib/client-stream')
@@ -25,10 +25,10 @@ undici.Pool = Pool
 undici.Client = Client
 undici.errors = errors
 
-undici.Agent = Agent
-undici.request = request
-undici.stream = stream
-undici.pipeline = pipeline
-undici.setGlobalAgent = setGlobalAgent
+// undici.Agent = Agent
+// undici.request = request
+// undici.stream = stream
+// undici.pipeline = pipeline
+// undici.setGlobalAgent = setGlobalAgent
 
 module.exports = undici
--- a/index.js
+++ b/index.js
@@ -3,6 +3,7 @@
 const Client = require('./lib/core/client')
 const errors = require('./lib/core/errors')
 const Pool = require('./lib/pool')
+const { Agent, request, stream, pipeline, setGlobalAgent } = require('./lib/agent')
 
 Client.prototype.request = require('./lib/client-request')
 Client.prototype.stream = require('./lib/client-stream')
@@ -23,5 +24,11 @@ function undici (url, opts) {
 undici.Pool = Pool
 undici.Client = Client
 undici.errors = errors
+
+undici.Agent = Agent
+undici.request = request
+undici.stream = stream
+undici.pipeline = pipeline
+undici.setGlobalAgent = setGlobalAgent
 
 module.exports = undici
--- a/index.js
+++ b/index.js
@@ -3,19 +3,13 @@
 const Client = require('./lib/core/client')
 const errors = require('./lib/core/errors')
 const Pool = require('./lib/pool')
-// const { Agent, request, stream, pipeline, setGlobalAgent } = require('./lib/agent')
+const { Agent, request, stream, pipeline, setGlobalAgent } = require('./lib/agent')
 
 Client.prototype.request = require('./lib/client-request')
 Client.prototype.stream = require('./lib/client-stream')
 Client.prototype.pipeline = require('./lib/client-pipeline')
 Client.prototype.upgrade = require('./lib/client-upgrade')
 Client.prototype.connect = require('./lib/client-connect')
-
-Pool.prototype.request = require('./lib/client-request')
-Pool.prototype.stream = require('./lib/client-stream')
-Pool.prototype.pipeline = require('./lib/client-pipeline')
-Pool.prototype.upgrade = require('./lib/client-upgrade')
-Pool.prototype.connect = require('./lib/client-connect')
 
 function undici (url, opts) {
   return new Pool(url, opts)
@@ -25,10 +19,10 @@ undici.Pool = Pool
 undici.Client = Client
 undici.errors = errors
 
-// undici.Agent = Agent
-// undici.request = request
-// undici.stream = stream
-// undici.pipeline = pipeline
-// undici.setGlobalAgent = setGlobalAgent
+undici.Agent = Agent
+undici.request = request
+undici.stream = stream
+undici.pipeline = pipeline
+undici.setGlobalAgent = setGlobalAgent
 
 module.exports = undici
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -1,17 +1,16 @@
 'use strict'
 
-const { InvalidArgumentError, InvalidReturnValueError, NotSupportedError } = require('./core/errors')
-const { Pool } = require('../')
-const { weakCache, nodeMajorVersionIsGreaterThanOrEqualTo } = require('./core/util')
+const { InvalidArgumentError, InvalidReturnValueError } = require('./core/errors')
+const Pool = require('./pool')
+const { weakCache, checkForWeakRefAndFinalizationRegistrySupport } = require('./core/util')
 
 const kAgentFactory = Symbol('agent factory')
 
 class Agent {
   constructor (opts) {
     /* istanbul ignore next */
-    if (!nodeMajorVersionIsGreaterThanOrEqualTo(14)) {
-      throw new NotSupportedError('Node.js version must be 14 or above')
-    }
+    checkForWeakRefAndFinalizationRegistrySupport()
+
     this[kAgentFactory] = weakCache(origin => new Pool(origin, opts))
   }
 
@@ -28,9 +27,7 @@ let globalAgent = new Agent()
 
 function setGlobalAgent (agent) {
   /* istanbul ignore next */
-  if (!nodeMajorVersionIsGreaterThanOrEqualTo(14)) {
-    throw new NotSupportedError('Node.js version must be 14 or above')
-  }
+  checkForWeakRefAndFinalizationRegistrySupport()
 
   if (typeof agent.get === 'function') {
     globalAgent = agent
@@ -46,9 +43,7 @@ function getPath (url) {
 function dispatchFromAgent (requestType) {
   return (url, { agent = globalAgent, method = 'GET', ...opts } = {}, ...additionalArgs) => {
     /* istanbul ignore next */
-    if (!nodeMajorVersionIsGreaterThanOrEqualTo(14)) {
-      throw new NotSupportedError('Node.js version must be 14 or above')
-    }
+    checkForWeakRefAndFinalizationRegistrySupport()
 
     if (url === undefined || url === null) {
       throw new InvalidArgumentError('Argument url must be defined')
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -1,46 +1,17 @@
 'use strict'
 /* global WeakRef, FinalizationRegistry */
 
-const { InvalidArgumentError } = require('./core/errors')
+const { InvalidArgumentError, InvalidReturnValueError, NotSupportedError } = require('./core/errors')
 const { Pool } = require('../')
-
-function weakCache (fn) {
-  const cache = new Map()
-  const cleanup = new FinalizationRegistry(key => {
-    // get the WeakRef from the cache
-    const ref = cache.get(key)
-    // if the WeakRef exists and the object has been reclaimed
-    if (ref !== undefined && ref.deref() === undefined) {
-      // remove the WeakRef from the cache
-      cache.delete(key)
-    }
-  })
-  return key => {
-    // check the cache for an existing WeakRef
-    const ref = cache.get(key)
-
-    // if one exists in the cache try to return the WeakRef
-    if (ref !== undefined) {
-      const cached = ref.deref()
-      if (cached !== undefined) {
-        return cached
-      }
-    }
-
-    // otherwise, if it isn't in the cache or the reference has been cleaned up, create a new one!
-    const value = fn(key)
-    // add a WeakRef of the value to the cache
-    cache.set(key, new WeakRef(value))
-    // add the value to the finalization registry
-    cleanup.register(value, key)
-    return value
-  }
-}
+const { weakCache, nodeMajorVersionIsGreaterThanOrEqualTo } = require('./core/util')
 
 const kAgentFactory = Symbol('agent factory')
 
 class Agent {
   constructor (opts) {
+    if (!nodeMajorVersionIsGreaterThanOrEqualTo(14)) {
+      throw NotSupportedError('Node.js version must be 14 or above')
+    }
     this[kAgentFactory] = weakCache(origin => new Pool(origin, opts))
   }
 
@@ -56,11 +27,36 @@ class Agent {
 let globalAgent = new Agent()
 
 function setGlobalAgent (agent) {
-  if (agent instanceof Agent) {
+  if (!nodeMajorVersionIsGreaterThanOrEqualTo(14)) {
+    throw NotSupportedError('Node.js version must be 14 or above')
+  }
+
+  if (typeof agent.get === 'function') {
     globalAgent = agent
   } else {
-    throw InvalidArgumentError('Argument agent must be instance of Agent')
+    throw InvalidArgumentError('Argument agent must implement Agent')
   }
+}
+
+function getClientFromAgent (agent, url) {
+  if (!nodeMajorVersionIsGreaterThanOrEqualTo(14)) {
+    throw NotSupportedError('Node.js version must be 14 or above')
+  }
+
+  // if url is a string transform into URL inst. otherwise use as is
+  url = typeof url === 'string' ? new URL(url) : url
+  // ensure it atleast has a non-empty string origin
+  if (typeof url.origin !== 'string' || url.origin === '') {
+    throw InvalidArgumentError('Argument url.origin must be a non-empty string')
+  }
+
+  const client = agent.get(url.origin)
+
+  if (client && typeof client.request !== 'function') {
+    throw InvalidReturnValueError('Client returned from Agent.get() does not implement method `.request()`')
+  }
+
+  return client
 }
 
 /**
@@ -68,21 +64,8 @@ function setGlobalAgent (agent) {
  * @param {string | import('url').URL} url
  * @param {*} opts
  */
-function request (url, opts = {}) {
-  // if url is a string transform into URL inst. otherwise use as is
-  url = typeof url === 'string' ? new URL(url) : url
-
-  if (typeof url.origin !== 'string' || url.origin === '') {
-    throw InvalidArgumentError('Argument url.origin must be a non-empty string')
-  }
-
-  const agent = (opts.agent === undefined || opts.agent === null) ? globalAgent : opts.agent
-
-  const client = agent.get(url.origin)
-
-  if (client && typeof client.request !== 'function') {
-    throw TypeError('Pool returned from Agent does not implement method `.request`')
-  }
+function request (url, { agent = globalAgent, ...opts } = {}) {
+  const client = getClientFromAgent(agent, url)
 
   return client.request({
     ...opts,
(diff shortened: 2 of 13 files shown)


lib/agent.js
```diff
@@ -0,0 +1,70 @@
...
+    constructor (opts) {
+        this[kAgentPoolOpts] = opts
+        this[kAgentCache] = new Map()
+        this[kAgentRegistry] = new FinalizationRegistry(key => {
+            // get the Pool WeakRef from the cache
+            const ref = this[kAgentCache].get(key)
+            // if the WeakRef exists and the object has been reclaimed
+            // according to mdn the second logic operation here is superfluous as a deref call in a FR will always return undefined
+            if (ref !== undefined && ref.deref() === undefined) {
+                // remove the WeakRef from the cache
```
ronag, line 18: Would be nice with an assertion checking that there are not pending requests in the pool

lib/agent.js
```diff
@@ -0,0 +1,70 @@
...
+                this[kAgentCache].delete(key)
+            }
+        })
+    }
+
+    get (origin) {
```
ronag, line 24: check origin is non empty string

lib/agent.js
```diff
@@ -0,0 +1,70 @@
...
+
+const globalAgent = new Agent({
+    connections: 16,
+    pipelining: 2,
+    headersTimeout: 10e3,
+    bodyTimeout: 10e3
+})
+
+function request (url, opts = {}) {
+    url = url instanceof URL ? url : new URL(url)
```
ronag, line 55: would be nice if we support url-like POJO

lib/agent.js
```diff
@@ -0,0 +1,70 @@
...
+    const agent = opts.agent == undefined ? globalAgent : opts.agent
+
+    const pool = agent.get(url.origin)
+    return pool.request({
+        ...opts,
+        method: opts.method || 'GET',
+        path: url.path || `${url.pathname || ''}${url.search || ''}`
+    })
+}
+
```
ronag, line 66: Should also export the 2 (3) other API methods

lib/agent.js
```diff
@@ -0,0 +1,70 @@
...
+        return pool
+    }
+}
+
+const globalAgent = new Agent({
+    connections: 16,
+    pipelining: 2,
+    headersTimeout: 10e3,
+    bodyTimeout: 10e3
+})
```
ronag, line 52: Should it be possible to override the global agent?

lib/agent.js
```diff
@@ -0,0 +1,70 @@
...
+        // add the Pool to the finalization registry
+        this[kAgentRegistry].register(pool, origin)
+
+        return pool
+    }
+}
+
+const globalAgent = new Agent({
+    connections: 16,
+    pipelining: 2,
```
ronag, line 49: I think we need to go without pipelining for default

lib/agent.js
```diff
@@ -0,0 +1,70 @@
...
+    headersTimeout: 10e3,
+    bodyTimeout: 10e3
```
ronag, line 51: Use default timeouts?

lib/agent.js
```diff
@@ -0,0 +1,70 @@
...
+            // according to mdn the second logic operation here is superfluous as a deref call in a FR will always return undefined
+            if (ref !== undefined && ref.deref() === undefined) {
+                // remove the WeakRef from the cache
+                this[kAgentCache].delete(key)
+            }
+        })
+    }
+
+    get (origin) {
+        // check the cache for an existing Pool
```
ronag, line 25: Later we might need to support 'origin' as an object containing more info to make the key from, in particular TLS stuff.

lib/agent.js
```diff
@@ -0,0 +1,70 @@
...
+    headersTimeout: 10e3,
+    bodyTimeout: 10e3
+})
+
+function request (url, opts = {}) {
+    url = url instanceof URL ? url : new URL(url)
+
+    const agent = opts.agent == undefined ? globalAgent : opts.agent
+
+    const pool = agent.get(url.origin)
```
ronag, line 59: since agent can be implemented in user land we should probably check the return value here.

lib/agent.js
Ethan-Arrowood: what should this callback do in the case there are requests pending or in flight? Also any ideas how to test this scenario? As far as I understand it the GC would not clean up a Pool currently being used

ronag: > what should this callback do in the case there are requests pending or in flight? 

It shouldn't be possible. If that happens it's a bug which is why I suggest an assertion.

ronag: Actually never mind. It's not possible to assert this.

lib/agent.js
Ethan-Arrowood: should this match the existing http.Agent ?

lib/agent.js
Ethan-Arrowood: What about:

```js
let globalAgent = new Agent()

function setGlobalAgent(agent) {
  if (agent instanceof Agent) {
    globalAgent = agent
  } else {
    throw TypeError('agent must be instance of Agent')
  }
}

function request () { ... }
```

lib/agent.js
Ethan-Arrowood: ill just make the default globalAgent `new Agent()` and rely on defaulting within the constructor

lib/agent.js
```diff
@@ -0,0 +1,81 @@
...
+      if (ref !== undefined && ref.deref() === undefined) {
+        // remove the WeakRef from the cache
+        this[kAgentCache].delete(key)
+      }
+    })
+  }
+
+  get (origin) {
+    if (typeof origin !== 'string' || origin === '') {
+      throw TypeError('Origin must be a non-empty string.')
```
ronag, line 27: use an error from `errors.js`.

lib/agent.js
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ *
+ * @param {string | import('url').URL} url
+ * @param {*} opts
+ */
+function request (url, opts = {}) {
+  // if url is a string transform into URL inst. otherwise use as is
+  url = typeof url === 'string' ? new URL(url) : url
+
```
ronag, line 62: ensure origin and path (or pathname) are valid strings.

lib/agent.js
```diff
@@ -0,0 +1,81 @@
...
+  const agent = (opts.agent === undefined || opts.agent === null) ? globalAgent : opts.agent
```
ronag, line 63: `opts.agent == null`

lib/agent.js
ronag: http.Agent? I mean stream and pipeline.

lib/agent.js
```diff
@@ -0,0 +1,81 @@
...
+
+  const pool = agent.get(url.origin)
```
ronag, line 65: agent could return a regular client as well... no need to assume it's a Pool. Just an object with `Client.request` trait

lib/agent.js
```diff
@@ -0,0 +1,81 @@
+'use strict'
+/* global WeakRef, FinalizationRegistry */
+
+const { Pool } = require('../')
+
+const kAgentPoolOpts = Symbol('agent pool opts')
+const kAgentCache = Symbol('agent cache')
+const kAgentRegistry = Symbol('agent registry')
```
ronag, line 8: missing new line

lib/agent.js
```diff
@@ -0,0 +1,81 @@
...
+class Agent {
+  constructor (opts) {
+    this[kAgentPoolOpts] = opts
+    this[kAgentCache] = new Map()
+    this[kAgentRegistry] = new FinalizationRegistry(key => {
+      // get the Pool WeakRef from the cache
+      const ref = this[kAgentCache].get(key)
+      // if the WeakRef exists and the object has been reclaimed
+      // according to mdn the second logic operation here is superfluous as a deref call in a FR will always return undefined
```
ronag, line 17: I actually think the second call is necessary as there is a possible ref race otherwise. See the Chrome page on WeakRefs.

ronag: https://github.com/tc39/proposal-weakrefs

lib/agent.js
Ethan-Arrowood: standard doesn't like that so I had to be more declarative

ronag: ah, intersting

lib/agent.js
Ethan-Arrowood: For validation:
origin **cannot** be empty
path and pathname **can** be empty

ronag: I think you only need to validate origin. path is validated by core.

lib/agent.js
Ethan-Arrowood: Ooh I also realized me putting all of this in the class is wrong too. V8 page on WeakRefs is very in depth

lib/agent.js
Ethan-Arrowood: Ahh I see what you mean. Adding two more functions `stream` and `pipeline` that rely on the Agent being created here.

lib/agent.js
```diff
@@ -1,56 +1,68 @@
...
-    const pool = new Pool(origin, this[kAgentPoolOpts])
-    // add a WeakRef of the Pool to the cache
-    this[kAgentCache].set(origin, new WeakRef(pool))
-    // add the Pool to the finalization registry
-    this[kAgentRegistry].register(pool, origin)
+let globalAgent = new Agent()
 
-    return pool
+function setGlobalAgent (agent) {
+  if (agent instanceof Agent) {
```
ronag, line 59: Doesn't need to be an instance of Agent. Just implement a get method.

lib/agent.js
```diff
@@ -0,0 +1,98 @@
...
+
+    // otherwise, if it isn't in the cache or the reference has been cleaned up, create a new one!
+    const value = fn(key)
+    // add a WeakRef of the value to the cache
+    cache.set(key, new WeakRef(value))
+    // add the value to the finalization registry
+    cleanup.register(value, key)
+    return value
+  }
+}
```
ronag, line 38: Move to util?

Ethan-Arrowood: this is agnostic enough to have its own module but for now i'll move it to util

test/agent.js
```diff
@@ -0,0 +1,36 @@
+const tap = require('tap')
+const http = require('http')
+const { request } = require('../lib/agent')
+
+tap.test('request a resource using globalAgent', t => {
```
ronag, line 5: Only enable for Node 14+

lib/agent.js
```diff
@@ -0,0 +1,98 @@
...
+    method: opts.method || 'GET',
+    path: url.path || `${url.pathname || ''}${url.search || ''}`
+  })
+}
+
+module.exports = {
+  request,
+  setGlobalAgent,
+  Agent
+}
```
ronag, line 98: Should we export these in index.js?

test/agent.js
Ethan-Arrowood: was going to ask about this... what should I do in the actual implementation if they aren't using Node 14+? Load a polyfill dep?

lib/agent.js
Ethan-Arrowood: I think so! Have something along the lines of:
```js
const undici = require('undici')

const agent = new undici.Agent()

const result = await undici.request('https://example.com', { agent })
```

test/agent.js
ronag: Just don't export it?

lib/agent.js
```diff
@@ -0,0 +1,98 @@
...
+    throw InvalidArgumentError('Argument agent must be instance of Agent')
+  }
+}
+
+/**
+ *
+ * @param {string | import('url').URL} url
+ * @param {*} opts
+ */
+function request (url, opts = {}) {
```
ronag, line 71:
```suggestion
function request (url, { agent = globalAgent, ...opts } = {}) {
```

test/agent.js
Ethan-Arrowood: This is going to be impossible to type. I propose we throw an error instead?

Ethan-Arrowood: Lol i'm copy/pasting the util i just wrote in case we need it later:
```js
function nodeMajorVersionIsGreaterThanOrEqualTo ({ major: expectedMajor, minor: expectedMinor, patch: expectedPatch }) {
  const [ foundMajor, foundMinor, foundPatch ] = process.versions.node.split('.').map(str => parseInt(str, 10))

  return foundMajor >= expectedMajor && foundMinor >= expectedMinor && foundPatch >= expectedPatch
}
```

Ethan-Arrowood:
Last commit has a first attempt at adding the other two methods. its probably wrong but i'll test and make it better tomorrow

ronag (commented):
LGTM, but needs more tests

lib/agent.js
```diff
@@ -0,0 +1,103 @@
...
+  }
+
+  return client
+}
+
+/**
+ *
+ * @param {string | import('url').URL} url
+ * @param {*} opts
+ */
```
ronag, line 66: Either add this to everything in file or remove it.

Ethan-Arrowood: im going to remove it for now - and soon ill revisit doing this for the whole project

ronag:
@dnlup

lib/agent.js
```diff
@@ -0,0 +1,103 @@
...
+
+function getClientFromAgent (agent, url) {
+  if (!nodeMajorVersionIsGreaterThanOrEqualTo(14)) {
+    throw NotSupportedError('Node.js version must be 14 or above')
+  }
+
+  // if url is a string transform into URL inst. otherwise use as is
+  url = typeof url === 'string' ? new URL(url) : url
+  // ensure it atleast has a non-empty string origin
+  if (typeof url.origin !== 'string' || url.origin === '') {
```
ronag, line 49: url can be nully here

Ethan-Arrowood:
Any thoughts on this part:

```js
{
  path: url.path || `${url.pathname || ''}${url.search || ''}`
}
```

Should we just stick to url.path?

index.js
```diff
@@ -3,6 +3,7 @@
 const Client = require('./lib/core/client')
 const errors = require('./lib/core/errors')
 const Pool = require('./lib/pool')
+const { Agent, request, stream, pipeline, setGlobalAgent } = require('./lib/agent')
```
Ethan-Arrowood, line 6: @ronag so this is causing a circular dependency issue. 

`lib/agent.js` requires `Pool` from this file since it needs the prototype methods on lines 14-18. How would you like me to resolve this? Can I move the prototype changes to `lib/pool.js`?

ronag: What prototype methods? I don't see how this is circular?

lib/agent.js
```diff
@@ -0,0 +1,80 @@
...
+const { InvalidArgumentError, InvalidReturnValueError, NotSupportedError } = require('./core/errors')
+const { Pool } = require('../')
+const { weakCache, nodeMajorVersionIsGreaterThanOrEqualTo } = require('./core/util')
+
+const kAgentFactory = Symbol('agent factory')
+
+class Agent {
+  constructor (opts) {
+    /* istanbul ignore next */
+    if (!nodeMajorVersionIsGreaterThanOrEqualTo(14)) {
```
dnlup, line 12: I did some checks and `FinalizationRegistry` seems supported from version `14.6.0`. Wouldn't be better to check if `FinalizationRegistry` and `WeakRef` are defined globally? Assuming users might use version `14.5.0` or something like that.

lib/agent.js
```diff
@@ -0,0 +1,80 @@
...
+    }
+    this[kAgentFactory] = weakCache(origin => new Pool(origin, opts))
+  }
+
+  get (origin) {
+    if (typeof origin !== 'string' || origin === '') {
+      throw new InvalidArgumentError('Origin must be a non-empty string.')
+    }
+
+    return this[kAgentFactory](origin)
```
dnlup, line 23: Should we support unix socket also?

dnlup:
> Any thoughts on this part:
> 
> ```js
> {
>   path: url.path || `${url.pathname || ''}${url.search || ''}`
> }
> ```
> 
> Should we just stick to url.path?

That should cover all cases, what doubts do you have about it?

lib/agent.js
ronag: nice to have

lib/agent.js
Ethan-Arrowood: yeah good point - i saw how you did it in your PR and will copy that pattern

index.js
Ethan-Arrowood: lines 14-18 are like `Pool.prototype.request = require('client-request.js')`

Agent requires access to this modded Pool class rather than the one exported from `./lib/pool`. 

So since `agent.js` requires `../index.js` and `index.js` requires `./lib/agent` its a circular ref

lib/agent.js
```diff
@@ -0,0 +1,80 @@
...
+
+  if (typeof agent.get === 'function') {
+    globalAgent = agent
+  } else {
+    throw new InvalidArgumentError('Argument agent must implement Agent')
+  }
+}
+
+function getPath (url) {
+  return url.path || `${url.pathname || ''}${url.search || ''}`
```
dnlup, line 43: On second thought, I think `pathname` should be `||` with `'/'`. If `url` is an instance of  `URL` the right side will never be reached (`pathname` always defaults to `'/'`), otherwise we should reproduce the same behavior (i.e., the user passes a plain object)

Ethan-Arrowood:
> That should cover all cases, what doubts do you have about it? 

Just haven't seen this before. I'll move it to the util file and test it

lib/agent.js
dnlup: We should also include `hash`:

```bash
URL {
  href: 'http://localhost/?test=1#readme',
  origin: 'http://localhost',
  protocol: 'http:',
  username: '',
  password: '',
  host: 'localhost',
  hostname: 'localhost',
  port: '',
  pathname: '/',
  search: '?test=1',
  searchParams: URLSearchParams { 'test' => '1' },
  hash: '#readme'
}
```

Ethan-Arrowood:
I have to be missing something obvious - it seems like the CI is blatantly ignoring the SKIP I added to the test and still running the Agent stuff 😕

lib/agent.js
Ethan-Arrowood: Is there some existing utility method for this kind of pathing? I feel weird adding this support when it feels like theres lots of edge cases

mcollina (approved):
lgtm

index.js
ronag: Implement pool in agent with require('./lib/client-request') etc...

ronag: alterantively move the pool modding to the pool file

lib/agent.js
dnlup: I don't know, unfortunately. I think this

```js
`${url.pathname || ''}${url.search || ''}${url.hash || ''}`
```

given how `hash` and `search` are parsed:

```bash
> new URL('http://localhost/?test=1#readme')
URL {
  href: 'http://localhost/?test=1#readme',
  origin: 'http://localhost',
  protocol: 'http:',
  username: '',
  password: '',
  host: 'localhost',
  hostname: 'localhost',
  port: '',
  pathname: '/',
  search: '?test=1',
  searchParams: URLSearchParams { 'test' => '1' },
  hash: '#readme'
}
```
```bash
> new URL('http://localhost/#readme?test=1')
URL {
  href: 'http://localhost/#readme?test=1',
  origin: 'http://localhost',
  protocol: 'http:',
  username: '',
  password: '',
  host: 'localhost',
  hostname: 'localhost',
  port: '',
  pathname: '/',
  search: '',
  hash: '#readme?test=1'
}
```

Do you see other corner cases?

lib/agent.js
dnlup: The `socketPath` option should be used in `origin` to identify a pool of connections. Also, it should be passed to the `kAgentFactory`. It can be done in a new PR if it's not urgent.

lib/agent.js
```diff
@@ -0,0 +1,79 @@
...
+  return (url, { agent = globalAgent, method = 'GET', ...opts } = {}, ...additionalArgs) => {
+    /* istanbul ignore next */
+    checkForWeakRefAndFinalizationRegistrySupport()
+
+    if (url === undefined || url === null) {
+      throw new InvalidArgumentError('Argument url must be defined')
+    }
+
+    // if url is a string transform into URL inst. otherwise use as is
+    url = typeof url === 'string' ? new URL(url) : url
```
dnlup, line 57: The last thing I am bringing up because I have seen it brought up in the past, using `URL` here is acceptable in terms of performance drop?

lib/agent.js
dnlup: I think this one is resolved 😄

dnlup (approved):
None of the comments I made are blocking, lgtm

lib/agent.js
Ethan-Arrowood: I'm +1 for follow up PR as this one is already quite large

lib/agent.js
Ethan-Arrowood: @ronag @mcollina thoughts on this?

mcollina (approved):
lgtm

ronag (changes requested):
LGTM, need docs

ronag:
thanks a lot!

lib/agent.js
ronag: This api is more for convenience than performance so I don't see it as a huge issue. But yea, there is room for optimization.

PR #510, opened 2021-01-06
feat(pool): forward connect and disconnect events

### LINKED ISSUE #509: Pool should forward Client(s) connect and disconnect events
Following https://github.com/nodejs/undici/pull/502#discussion_r551835540 , it could be useful if the `Pool` forwarded its `Client`s `connect`/`disconnect` events, with the `Client` instance as an argument.

dnlup:
Make the `Pool` forward the `connect` and `disconnect` events emitted by its
clients. The `Client` instance that emitted the event is passed as
the argument.

Fixes #509

### DIFF
--- a/lib/pool.js
+++ b/lib/pool.js
@@ -1,5 +1,6 @@
 'use strict'
 
+const EventEmitter = require('events')
 const Client = require('./core/client')
 const {
   ClientClosedError,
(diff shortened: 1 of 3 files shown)


lib/pool.js
```diff
@@ -52,6 +63,8 @@ class Pool {
```
ronag, line 63:
```js
const onConnect = function () {
  pool.emit('connect', this)
}
```

This way you only create one closure, instead of one for each client.

dnlup: Yes, you're right. Thank you!

ronag (approved):

mcollina (approved):
lgtm

PR #521, opened 2021-01-14
add close and destroy methods to Agent class

### LINKED ISSUE #520: Agent is missing close and destroy methods.
Agent is missing close and destroy methods.

Ethan-Arrowood:
Closes #520

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -2,14 +2,25 @@
 
 const { InvalidArgumentError, InvalidReturnValueError } = require('./core/errors')
 const Pool = require('./pool')
-const { weakCache } = require('./core/util')
+const { parseOrigin } = require('./core/util')
 
-const kAgentFactory = Symbol('agent factory')
+const kOpts = Symbol('opts')
+const kCache = Symbol('cache')
+const kCleanup = Symbol('cleanup')
 
 class Agent {
   constructor (opts) {
-    this[kAgentFactory] = weakCache(origin => new Pool(origin, opts))
-    this.origins = []
+    this[kOpts] = opts
+    this[kCache] = new Map()
+    this[kCleanup] = new FinalizationRegistry(key => {
+      // get the WeakRef from the cache
+      const ref = cache.get(key)
+      // if the WeakRef exists and the object has been reclaimed
+      if (ref !== undefined && ref.deref() === undefined) {
+        // remove the WeakRef from the cache
+        cache.delete(key)
+      }
+    })
   }
 
   get (origin) {
@@ -17,27 +28,47 @@ class Agent {
       throw new InvalidArgumentError('Origin must be a non-empty string.')
     }
 
-    if (this.origins.indexOf(origin) === -1) {
-      this.origins.push(origin)
+    // check the cache for an existing WeakRef
+    const ref = this[kCache].get(origin)
+
+    // if one exists in the cache try to return the WeakRef
+    if (ref !== undefined) {
+      const cached = ref.deref()
+      if (cached !== undefined) {
+        return cached
+      }
     }
 
-    return this[kAgentFactory](origin)
+    // otherwise, if it isn't in the cache or the reference has been cleaned up, create a new one!
+    const value = new Pool(origin, this[kOpts])
+    // add a WeakRef of the value to the cache
+    this[kCache].set(origin, new WeakRef(value))
+    // add the value to the finalization registry
+    this[kCleanup].register(value, origin)
+
+    return value
   }
 
  close () {
     const closePromises = []
-    for (const origin in origins) {
-      const client = this[kAgentFactory](origin)
-      closePromises.push(client.close.bind(client)())
+    for (const ref in this[kCache].values()) {
+      const pool = ref.deref()
+
+      if (pool) {
+        closePromises.push(pool.close())
+      }
     }
     return Promise.all(closePromises)
   }
 
- destroy () {
+  destroy () {
     const destroyPromises = []
-    for (const origin in origins) {
-      const client = this[kAgentFactory](origin)
-      destroyPromises.push(client.destroy.bind(client)())
+    for (const ref in this[kCache].values()) {
+      const pool = ref.deref()
+
+      if (pool) {
+        destroyPromises.push(pool.destroy())
+      }
     }
     return Promise.all(destroyPromises)
   }
@@ -58,24 +89,11 @@ function setGlobalAgent (agent) {
 }
 
 function dispatchFromAgent (requestType) {
-  return (url, { agent = globalAgent, method = 'GET', ...opts } = {}, ...additionalArgs) => {
-    if (url === undefined || url === null) {
-      throw new InvalidArgumentError('Argument url must be defined')
-    }
+  return (origin, { agent = globalAgent, method = 'GET', ...opts } = {}, ...additionalArgs) => {
+    const url = parseOrigin(origin)
 
     if (!agent || typeof agent.get !== 'function') {
       throw new InvalidArgumentError('Argument agent must implement Agent')
-    }
-
-    if (opts.path != null) {
-      throw new InvalidArgumentError('unsupported opts.path')
-    }
-
-    // if url is a string transform into URL inst. otherwise use as is
-    url = typeof url === 'string' ? new URL(url) : url
-    // ensure it atleast has a non-empty string origin
-    if (!url || typeof url.origin !== 'string' || url.origin === '') {
-      throw new InvalidArgumentError('Argument url.origin must be a non-empty string')
     }
 
     const client = agent.get(url.origin)
@@ -84,9 +102,7 @@ function dispatchFromAgent (requestType) {
       throw new InvalidReturnValueError(`Client returned from Agent.get() does not implement method ${requestType}`)
     }
 
-    const path = url.path || `${url.pathname || '/'}${url.search || ''}`
-
-    return client[requestType]({ ...opts, method, path }, ...additionalArgs)
+    return client[requestType]({ ...opts, method, path: url.path }, ...additionalArgs)
   }
 }
 
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -3,22 +3,19 @@
 
 const { InvalidArgumentError, InvalidReturnValueError } = require('./core/errors')
 const Pool = require('./pool')
-
-const kOpts = Symbol('opts')
-const kCache = Symbol('cache')
-const kCleanup = Symbol('cleanup')
+const { kAgentOpts, kAgentCache, kAgentCleanup } = require('./core/symbols')
 
 class Agent {
   constructor (opts) {
-    this[kOpts] = opts
-    this[kCache] = new Map()
-    this[kCleanup] = new FinalizationRegistry(key => {
+    this[kAgentOpts] = opts
+    this[kAgentCache] = new Map()
+    this[kAgentCleanup] = new FinalizationRegistry(key => {
       // get the WeakRef from the cache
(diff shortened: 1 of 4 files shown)


lib/agent.js
```diff
@@ -11,15 +11,38 @@ class Agent {
...
     }
 
+    if (this.origins.indexOf(origin) === -1) {
+      this.origins.push(origin)
+    }
+
     return this[kAgentFactory](origin)
   }
+
+  async close () {
```
mcollina, line 29:
```suggestion
 close () {
```

lib/agent.js
```diff
@@ -11,15 +11,38 @@ class Agent {
...
+    const closePromises = []
+    for (const origin in origins) {
+      const client = this[kAgentFactory](origin)
+      closePromises.push(client.close.bind(client)())
+    }
+    return Promise.all(closePromises)
+  }
+
+  async destroy () {
```
mcollina, line 38:
```suggestion
 destroy () {
```

mcollina (changes requested):
async was not needed here, as we already return a promise.

ronag:
@Ethan-Arrowood can you give edit access to this branch?

ronag:
```js

const kCache = Symbol('cache')
const kCleanup = Symbol('cleanup')
const kOpts = Symbol('opts')

class Agent {
  constructor (opts) {
    /* istanbul ignore next: */
    if (typeof WeakRef === 'undefined' || typeof FinalizationRegistry === 'undefined') {
      throw new NotSupportedError('In order to use this feature, `WeakRef` and `FinalizationRegistry` must be defined as global objects. Check your Node.js version to be sure it is v14.6.0 or greater.')
    }

    this[kOpts] = opts
    this[kCache] = new Map()
    this[kCleanup] = new FinalizationRegistry(key => {
      // get the WeakRef from the cache
      const ref = this[kCache].get(key)
      // if the WeakRef exists and the object has been reclaimed
      if (ref !== undefined && ref.deref() === undefined) {
        // remove the WeakRef from the cache
        this[kCache].delete(key)
      }
    })
  }

  get (origin) {
    if (typeof origin !== 'string' || origin === '') {
    }

    // check the cache for an existing WeakRef
    const ref = this[kCache].get(origin)

    // if one exists in the cache try to return the WeakRef
    if (ref !== undefined) {
      const cached = ref.deref()
      if (cached !== undefined) {
        return cached
      }
    }

    // otherwise, if it isn't in the cache or the reference has been cleaned up, create a new one!
    const value = new Pool(origin, this[kOpts])
    // add a WeakRef of the value to the cache
    this[kCache].set(origin, new WeakRef(value))
    // add the value to the finalization registry
    this[kCleanup].register(value, origin)

    return value
  }

  close () {
    const promises = []
    for (const ref of this[kCache].values()) {
      const pool = ref.deref()
      if (pool) {
        promises.push(pool.close())
      }
    }
    return Promise.all(promises)
  }

  destroy () {
    const promises = []
    for (const ref of this[kCache].values()) {
      const pool = ref.deref()
      if (pool) {
        promises.push(pool.destroy())
      }
    }
    return Promise.all(promises)
  }
}
```

Ethan-Arrowood:
> @Ethan-Arrowood can you give edit access to this branch?

its enabled

ronag:
Hm. Doesn't let me push... anyway you have my proposed changes above.

Ethan-Arrowood:
> Hm. Doesn't let me push... anyway you have my proposed changes above.

I've always found that feature really strange. It doesn't work how I expect it to all the time

Ethan-Arrowood:
Working on a rebase right now. Didn't realize things changed upstream, whoops!

Ethan-Arrowood:
Okay getting there, rebased with master now. Trying to use the new `parseOrigin` util. Need to work on testing next. Apologies for how long this is taking me - I'm in a wicked busy week right now 😬

lib/agent.js
```diff
@@ -35,35 +90,20 @@ function setGlobalAgent (agent) {
...
     const client = agent.get(url.origin)
 
     if (client && typeof client[requestType] !== 'function') {
       throw new InvalidReturnValueError(`Client returned from Agent.get() does not implement method ${requestType}`)
     }
 
-    const path = url.path || `${url.pathname || '/'}${url.search || ''}`
-
-    return client[requestType]({ ...opts, method, path }, ...additionalArgs)
+    return client[requestType]({ ...opts, method, path: url.path }, ...additionalArgs)
```
ronag, line 106: [`URL`](https://nodejs.org/api/url.html) doesn't have `path`.

lib/agent.js
```diff
@@ -48,22 +101,13 @@ function dispatchFromAgent (requestType) {
...
-      throw new InvalidArgumentError('Argument url.origin must be a non-empty string')
-    }
-
     const client = agent.get(url.origin)
 
     if (client && typeof client[requestType] !== 'function') {
       throw new InvalidReturnValueError(`Client returned from Agent.get() does not implement method ${requestType}`)
     }
 
-    const path = url.path || `${url.pathname || '/'}${url.search || ''}`
```
ronag, line 64: You should keep this

ronag (changes requested):

lib/agent.js
Ethan-Arrowood: I'm not super clear on parseOrigin util, but can its result have the `path` prop? If so i'll add that piece back. I've included the right hand side inline of the return statement

ronag: `${url.pathname || '/'}${url.search || ''}` is sufficient.

lib/agent.js
```diff
@@ -48,22 +101,13 @@ function dispatchFromAgent (requestType) {
...
-
-    return client[requestType]({ ...opts, method, path }, ...additionalArgs)
+    return client[requestType]({ ...opts, method, path: `${url.pathname || '/'}${url.search || ''}` }, ...additionalArgs)
```
ronag, line 110: actually url can't contain pathname nor search, it's just origin URL.

ronag: I think we're both confused :D

ronag: Maybe just skip parseOrigin here for now and do this part like you did before.

ronag:
@Ethan-Arrowood I would like to cut a new release soonish. Do you have any timeframe on this PR? i.e. should I wait a few more days before doing a new release?

Ethan-Arrowood:
> I would like to cut a new release soonish. Do you have any timeframe on this PR? i.e. is there any meaning to waiting a few more days before doing a new release?

Yeah lets push to close out this PR. I have some cycles now to wrap it up. Other than the URL piece what is missing?

lib/agent.js
Ethan-Arrowood: Okay cool

ronag:
> Other than the URL piece what is missing?

Tests

Ethan-Arrowood:
@ronag I just pushed the tests I wrote for this feature. They are both failing at the `closed` and `destroyed` checks. I'm going to come back and debug this after i wrap up some meetings, but if you spot anything obvious let me know.

test/agent.js
```diff
@@ -40,6 +41,82 @@ tap.test('Agent', { skip: SKIP }, t => {
...
+
+      await Promise.all(servers.map(server => promisifyServerListen(server)))
+      await Promise.all(servers.map(server => _request(t, agent, server)))
+      for (const server of servers) {
+        t.ok(server.listening, 'server should be listening')
+      }
+      await agent.destroy()
+      for (const ref of agent[kAgentCache].values()) {
+        const pool = ref.deref()
+        t.ok(pool.destroyed, 'pools should be destroyed')
```
ronag, line 115: You should be able to test for this through disconnect events? No need for internal symbol.

Ethan-Arrowood: The internal symbol is required because it is the only way to reference the pools inside of the agent. Unless you're expecting the agent itself to emit some disconnect event?

Ethan-Arrowood: But yes I can make use of the pool disconnect events for the test - will try that approach now.

ronag: Maybe agent should also emit disconnect?

Ethan-Arrowood:
After some debugging I figured out that they are closing, its just doing so after the test completes. I think I've hit this issue before when writing `async` function tests in tap. Will try dropping the async keyword and going back to callbacks now

test/agent.js
Ethan-Arrowood: Yeah I could try to forward that along - seems partially related to this https://github.com/nodejs/undici/issues/529 too

Ethan-Arrowood:
```js
t.test('agent should close internal pools', t => {
      t.plan(2)

      const wanted = 'payload'

        res.end(wanted)
      })

      t.tearDown(server.close.bind(server))

      server.listen(0, () => {
        const agent = new Agent()
        
        request(origin, { agent })
          .then(() => {
            t.fail()
          })
          .catch(err => {
            t.error(err instanceof errors.ClientClosedError)
          })

        const pool = agent.get(origin)
        pool.once('connect', () => {
          let closed = false
          pool.on('disconnect', () => {
            console.log('DISCONNECT')
            closed = true
          })
          agent.close().then(() => {
            console.log('CLOSE CB')
            t.ok(closed, 'pool should be disconnected')
          })
        })
      })
    })
 ```

So I'm trying this (which is based off of a test in close-and-destroy.js

Still, the `DISCONNECT` console log is coming after the `CLOSE CB` one is. Does this mean the agent.close implementation is wrong?

**Comment by Ethan-Arrowood** [comment]
Trying another way and it still fails at `t.fail()`

```js
      server.listen(0, () => {
        const agent = new Agent()

        request(origin, { agent })
          .then(() => {
            t.pass()
          })
          .catch(err => {
            t.error(err instanceof errors.ClientClosedError)
          })

        agent.close().then(() => {
          request(origin, { agent })
            .then(() => {
              t.fail()
            })
            .catch(err => {
              t.error(err instanceof errors.ClientClosedError)
            })
        })
    })
```

🤷‍♂️ not sure what im doing wrong

ronag:
I'm taking a look now

lib/agent.js
```diff
@@ -1,22 +1,73 @@
...
+    this[kAgentCache].set(origin, new WeakRef(value))
+    // add the value to the finalization registry
+    this[kAgentCleanup].register(value, origin)
+
+    return value
+  }
+
+  close () {
+    const closePromises = []
+    for (const ref in this[kAgentCache].values()) {
```
ronag, line 51:
```suggestion
    for (const ref of this[kAgentCache].values()) {
```

lib/agent.js
```diff
@@ -1,22 +1,73 @@
...
+      if (pool) {
+        closePromises.push(pool.close())
+      }
+    }
+    return Promise.all(closePromises)
+  }
+
+  destroy () {
+    const destroyPromises = []
+    for (const ref in this[kAgentCache].values()) {
```
ronag, line 63:
```suggestion
    for (const ref of this[kAgentCache].values()) {
```

ronag (changes requested):
See comments, typo, `in` should be `of`.

Ethan-Arrowood:
Maybe the issue I'm having is that the request is actually happening after I call close, but cause of the logic of Agent it is just standing up a new pool instance for that origin.

Ethan-Arrowood:
Alright these are passing locally lets see what happens.

ronag (approved):

ronag:
Thanks!

ronag:
@Ethan-Arrowood I just noticed that we are missing docs. Can you open a PR for that?

PR #533, opened 2021-02-02
agent: add support on Node 12 and lower

### LINKED ISSUE #529: Agent without WeakRef dependency
It should be possible to implement `Agent` without using weak refs by using the `'connect'`/`'disconnect'` events.

dnlup:
Use `Pool`'s `connect`/`disconnect` events to add support of the `Agent` on Node 12 and lower.

Fixes #529 

Note:

I think it's better to wait for https://github.com/nodejs/undici/pull/521 to get merged before continuing. I have to touch many of the same blocks of that PR.

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -1,24 +1,13 @@
 'use strict'
-/* global WeakRef, FinalizationRegistry */
-
 const { InvalidArgumentError, InvalidReturnValueError } = require('./core/errors')
 const Pool = require('./pool')
 const util = require('./core/util')
-const { kAgentOpts, kAgentCache, kAgentCleanup } = require('./core/symbols')
+const { kAgentOpts, kAgentCache } = require('./core/symbols')
 
 class Agent {
   constructor (opts) {
     this[kAgentOpts] = opts
     this[kAgentCache] = new Map()
-    this[kAgentCleanup] = new FinalizationRegistry(key => {
-      // get the WeakRef from the cache
-      const ref = this[kAgentCache].get(key)
-      // if the WeakRef exists and the object has been reclaimed
-      if (ref !== undefined && ref.deref() === undefined) {
-        // remove the WeakRef from the cache
-        this[kAgentCache].delete(key)
-      }
-    })
   }
 
   get (origin) {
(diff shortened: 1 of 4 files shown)


Ethan-Arrowood:
@dnlup if you have the time, I could use a hand closing my PR. I'm struggling to get the tests to behave like I want them to and I can't seem to crack it

dnlup:
> @dnlup if you have the time, I could use a hand closing my PR. I'm struggling to get the tests to behave like I want them to and I can't seem to crack it

Ok, I'll see what I can do

lib/agent.js
```diff
@@ -72,6 +83,7 @@ class Agent {
   }
 }
 
+Agent.prototype.get = util.isWeakRefsSupported ? getFromWeakCache : getFromCache
```
ronag, line 86: I think we can skip the weak cache version if this works?

dnlup: Sounds good, I'll leave the logic for now but use only the version without the weakCache and see what happens.

dnlup: It looks like it works, so I think we can switch to a version without `WeakRefs` easily.

lib/agent.js
```diff
@@ -72,13 +71,61 @@ class Agent {
...
+  constructor (opts) {
+    this[kAgentOpts] = opts
+    this[kAgentCache] = new Map()
+    if (util.isWeakRefsSupported) {
+      initWeakCache(this)
+    }
+  }
+}
+
+if (util.isWeakRefsSupported) {
```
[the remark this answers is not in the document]
dnlup: Ok, I removed the `WeakRef` logic. I am missing the reason for using the `connect` event, though. Once a `Pool` instance has no clients connected, it gets removed from the cache. When requesting a pool, a new one is initialized if it has not been created before for the given `origin`. What's the key point I am not getting here?

lib/agent.js
```diff
@@ -1,83 +1,57 @@
...
+    let pool = self[kAgentCache].get(origin)
 
-    // if one exists in the cache try to return the WeakRef
-    if (ref !== undefined) {
-      const cached = ref.deref()
-      if (cached !== undefined) {
-        return cached
+    function onDisconnect () {
+      if (this.connected === 0) {
+        this.removeListener('disconnect', onDisconnect)
```
ronag, line 23:
```suggestion
        this.off('disconnect', onDisconnect)
```

ronag:
Don't forget to update docs

lib/agent.js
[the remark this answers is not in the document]
mcollina: Nothing, using `disconnect` is just fine.

mcollina (approved):
lgtm

lib/agent.js
```diff
@@ -1,83 +1,57 @@
...
+    const self = this
+    let pool = self[kAgentCache].get(origin)
 
-    // if one exists in the cache try to return the WeakRef
-    if (ref !== undefined) {
-      const cached = ref.deref()
-      if (cached !== undefined) {
-        return cached
+    function onDisconnect () {
+      if (this.connected === 0) {
```
ronag, line 22:
```suggestion
      if (this.connected === 0 && this.pending === 0) {
```

dnlup: Thanks, great catch

ronag:
Done?

dnlup:
> Done?

Yup, unless I missed something else

ronag:
@dnlup This commit seems to cause CI failures. Could you investigate?

dnlup:
Sure, are you referring to the failures https://github.com/nodejs/undici/pull/535 ?

ronag:
Screen Shot 2021-02-05 at 14 21 19

dnlup:
Ok, thanks

dnlup:
That specific error was caused by a failure during the setup phase (it couldn't download the `checkout` action). I have seen similar errors [here](https://github.com/nodejs/undici/runs/1838763598?check_suite_focus=true). I'll dig some more to find failures related to this commit.

ronag:
Ah I see. I'll just try re-running it then.

PR #535, opened 2021-02-04
feat: Pool size & running

ronag:


ronag:
Tests needs fixing

ronag:
@dnlup Feel free to sort out the tests here if you want. But I was referring to the failures on master.

ronag:
@mcollina

mcollina (changes requested):

ronag:
merge after https://github.com/nodejs/undici/pull/541

lib/pool.js
```diff
@@ -59,17 +59,15 @@ class Pool extends EventEmitter {
...
 
-      if (queue.isEmpty()) {
-        if (pool[kNeedDrain]) {
-          pool[kNeedDrain] = false
-          pool.emit('drain')
-        }
+      if (pool[kNeedDrain] && !this.busy) {
+        pool[kNeedDrain] = false
+        pool.emit('drain')
+      }
```
mcollina, line 65: why this if was removed?

mcollina (changes requested):

lib/pool.js
ronag: I don't understand the question.

mcollina: you removed the `if (queue.isEmpty())` check. Why?

ronag: I replaced it with `this.busy` which contains the same check through `this[kPending] > 0`

ronag: the empty check is moved to just `if (pool[kClosedResolve] && queue.isEmpty()) {`

mcollina (approved):
lgtm

PR #538, opened 2021-02-05
feat: support long polling requests and event-source responses

ronag:


ronag:
Needs tests

ronag:
This logic is only relevant when pipelining > 1 and ensures that no more requests are enqueued.

MetCoder95:
This is an implementation of SSR or pure long-polling? (I'm asking because I see the `text/event-stream` header involved) :)

mcollina (commented):
wow!

PR #541, opened 2021-02-07
fix: busy if full

ronag:
busy should be true if pipeline is full. Should not have request unnecessarily queued.


mcollina (approved):
lgtm

PR #544, opened 2021-02-10
fix: unify Pool and Client APIs

ronag:
- 'connect' and 'disconnect' event should have same arguments.
- connected should be an integer

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,8 +115,6 @@ Options:
 
 Performs a HTTP request.
 
-Requests are not guaranteeed to be dispatched in order of invocation.
-
 Options:
 
 * `path: String`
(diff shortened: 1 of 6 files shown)


ronag:
This allows e.g. using `Agent` with `Client` instead of `Pool`. So if `connections === 1` we could fallback to `Client` and avoid the overhead of `Pool`.

mcollina (changes requested):
This needs tests

ronag:
test added

mcollina (approved):
lgtm

Ethan-Arrowood:
I'm going to update my docs PR to include these changes now too

PR #548, opened 2021-02-12
fix: wait for queued handlers to run before onError

ronag:
When e.g. returning a stream through resolve the user
needs to get a chance to register an 'error' handler
before the stream is destroyed. Otherwise we risk an
unhandled exception.

### DIFF
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -145,7 +145,7 @@ class Request {
     this.aborted = true
 
     // Ensure all queued handlers are invoked before calling onError.
-    queueMicrotask(() => this[kHandler].onError(err))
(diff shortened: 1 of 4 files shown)


lib/core/request.js
```diff
@@ -144,7 +144,8 @@ class Request {
     }
     this.aborted = true
 
-    return this[kHandler].onError(err)
+    // Ensure all queued handlers are invoked before calling onError.
+    Promise.resolve().then(() => this[kHandler].onError(err))
```
ronag, line 148: Has some performance implication but this should not be a hot path.

mcollina (approved):
lgtm

PR #550, opened 2021-02-16
Improve docs - Client

Ethan-Arrowood:
This PR is the start to improving the API docs. It follows a style guide that I'm in the process of formalizing. It is completely CommonMark compliant markdown and should render very nicely with documentation site generation tools. I'd love as much feedback as possible on this before I set out and write the Pool, Agent, and Errors docs like this too.

Most of these docs are copy-paste from the existing readme. The code examples have all been run locally on my machine and should work just by copy-pasting. Some examples are incomplete as I just didn't have the time to dive into them. We can keep em this way and let others fill them in or we can wait till I fill them in myself.

The types here conflict with our actual typescript types because the current types are wrong. Thats okay - i'll follow up shortly with a types fix. A big drive from writing the docs this way is to, for now, make our types easier to write and manage (since they'll closely mirror the detailed api docs), and in the future I hope to derive the types themselves from this API documentation.

Here is a quick link to view the new markdown file in my fork (so you can read it rendered by GitHub): https://github.com/Ethan-Arrowood/undici/blob/improve-docs/docs/Client.md 

Some things to consider when reviewing:

- Should we keep the Table of Contents?
- Are class method definitions clear? 
  - Can the "overloaded" syntax be improved in any way?
  - Is it necessary to call a promise/callback function overloaded? ([Client.upgrade](https://github.com/Ethan-Arrowood/undici/blob/improve-docs/docs/Client.md#clientupgradeoptions-callback) implements an alternate callback overload style)
  - Is the order of the method sub-sections clear? (i.e. Overloads, Parameters, Examples)
- How should Node.js dependencies be represented? Is it clear enough to just use the class names such as `Readable` or is `stream.Readable` preferred?


ronag:
The docs seems a little outdated.

I'm not sure I prefer this very explicit overload style over the more compact one that node uses.

Other than that good job.

Would be nice of docs were based on e.g. Doxygen (which would give some typescript like tooling).

docs/Client.md
```diff
@@ -0,0 +1,928 @@
...
+
+nodeServer.listen(0, () => {
+  console.log('Node Server listening')
+
+  const client = new undici.Client(`http://localhost:${nodeServer.address().port}`)
+
+  fastifyServer.route({
+    url: '/',
+    method: 'GET',
+    handler: async (request, response) => {
```
ronag, line 748: async is redundant here

Ethan-Arrowood:
> The docs content seems a little outdated.

Yeah I did my best to validate everything against the current API (from referencing code and tests). This is what indicated to me that the current types are so out-of-date

> I'm not sure I prefer this very explicit overload style over the more compact one that node uses.

Same here. I've tried a couple of things but I don't love any of them. I just pushed a commit that changes client.upgrade to another style that I prefer. Let me know your thoughts.

> Other than that good job.

Thank you 😁  

> Would be nice of docs were based on e.g. Doxygen (which would also give some typescript like tooling).

Yeah I've been on the hunt for something to base the docs around. Since there is no major tool adopted by JS projects some folks and myself have been brainstorming for awhile. Electron has there docs backed by a style-guide which they have then gone and built tooling around. The groups larger goal is to formalize what Electron is doing and make it usable for other Node.js projects. My undici-fetch docs were an experiment with an Electron inspired documentation style guide and this PR is another one.

package.json
```diff
@@ -36,6 +36,7 @@
     "abort-controller": "^3.0.0",
     "benchmark": "^2.1.4",
     "concurrently": "^5.2.0",
+    "fastify": "^3.11.0",
```
mcollina, line 39: is this needed?

docs/Client.md
```diff
@@ -0,0 +1,928 @@
...
+  client
+    .upgrade({ path: '/' })
+    .then(({ headers, socket }) => {
+      socket.on('end', () => {
+        console.log(`upgrade: ${headers.upgrade}`)
+        client.close()
+        server.close()
+      })
+      socket.end()
+    })
```
mcollina, line 826: what happens if this rejects? Maybe add a catch?

package.json
Ethan-Arrowood: Oh oops - its used in one of the examples and i forgot to remove it here.

docs/Client.md
Ethan-Arrowood: Good point, will do! (Best to not have unhandeledRejections laying around 😉 )

ronag:
> I just pushed a commit that changes client.upgrade to another style that I prefer. Let me know your thoughts.

Much better!

docs/Client.md
```diff
@@ -99,19 +87,15 @@ const client = Client('http://localhost:3000')
...
-
-Returns: `Promise<null>`
-
-#### (2) `Client.close(callback)`
+Closes the client and gracefully waits for enqueued requests to complete before resolving.
 
 Arguments:
 
-* **callback** `(error: Error | null, data: null) => void`
+* **callback** `(error: Error | null, data: null) => void` (optional) - If a `callback` is not provided, a `Promise` is returned
```
ronag, line 96:
```suggestion
* **callback** `(error: Error | null, data: null) => void` (optional)
```

ronag: Duplicate info. You already mention this below.

Ethan-Arrowood: Do you want that description line removed for all of the callback arguments?

docs/Client.md
```diff
@@ -0,0 +1,883 @@
...
+    server.close()
+  })
+})
+```
+
+### `Client.stream(options, factory [, callback])`
+
+A faster version of `Client.request`. This method expects the second argument `factory` to return a [`Writable`](https://nodejs.org/api/stream.html#stream_class_stream_writable) stream which the response will be written to. This improves performance by avoiding creating an intermediate [`Readable`](https://nodejs.org/api/stream.html#stream_readable_streams) stream when the user expects to directly pipe the response body to a [`Writable`](https://nodejs.org/api/stream.html#stream_class_stream_writable) stream.
+
+As demonstrated in [Example 1 - Basic GET stream request](#example-1---basic-get-stream-request), it is strongly recommended to use the `option.opaque` property to avoid creating a closure for the `factory` method. This pattern works extremely well with Node.js Web Frameworks such as [Fastify](https://fastify.io). See [Example 2 - Stream to Fastify Response](#example-2---stream-to-fastify-response) for more details.
```
ronag, line 609:
```suggestion
As demonstrated in [Example 1 - Basic GET stream request](#example-1---basic-get-stream-request), it is recommended to use the `option.opaque` property to avoid creating a closure for the `factory` method. This pattern works well with Node.js Web Frameworks such as [Fastify](https://fastify.io). See [Example 2 - Stream to Fastify Response](#example-2---stream-to-fastify-response) for more details.
```

docs/Client.md
```diff
@@ -0,0 +1,883 @@
...
+* **factory** `(data: StreamFactoryData) => stream.Writable`
+* **callback** `(error: Error | null, data: StreamData) => void` (optional) - If a `callback` is not provided, a `Promise` is returned
+
+Returns: `void | Promise<StreamData>` - Only returns a `Promise` if no `callback` argument was passed
+
+#### Parameter: `StreamFactoryData`
+
+* **statusCode** `number`
+* **headers** `http.IncomingHttpHeaders`
+* **opaque** `unknown`
```
ronag, line 623: what is the diff between unknown and any?

docs/Client.md
```diff
@@ -0,0 +1,883 @@
...
+  }).then(({ opaque: { bufs } }) => {
+
+    console.log(Buffer.concat(bufs).toString('utf8'))
+
+    client.close()
+    server.close()
+  }).catch(error => {
+    console.error(error)
+  })
+})
```
ronag, line 670: This is example is a bit weird.

docs/Client.md
```diff
@@ -0,0 +1,883 @@
...
+
+### `Client.size`
+
+* `number`
+
+Number of pending and running requests.
+
+## Instance Events
+
+### Event: `'connect'`
```
ronag, line 837: missing docs for arguments

docs/Client.md
```diff
@@ -0,0 +1,883 @@
...
+
+Emitted when a socket has been created and connected. The client will connect once `Client.size > 0`.
+
+### Event: `'disconnect'`
+
+Returns:
```
ronag, line 843: returns?

docs/Client.md
```diff
@@ -0,0 +1,883 @@
...
+
+* **error** `Error`
+
+Emitted when socket has disconnected. The first argument of the event is the error which caused the socket to disconnect. The client will reconnect if or once `Client.size > 0`.
+
+### Event: `'drain'`
+
+Emitted when pipeline is no longer fully saturated.
```
ronag, line 851: No strictly true. Emitted when no longer `busy` (needs ref to busy prop).

docs/Client.md
Ethan-Arrowood: `unknown` is a safer `any`. It requires the TS users to specify what it is before they use it (usually with a type guard or type casting).

docs/Client.md
```diff
@@ -0,0 +1,883 @@
...
+
+## Parameter: `UndiciHeaders`
+
+* `http.IncomingHttpHeaders | string[] | null`
+
+Header arguments such as `options.headers` in [`Client.dispatch`](./Client.md#client-dispatchoptions-handlers) can be specified in two forms; either as an object specified by the `http.IncomingHttpHeaders` type, or an array of strings. An array representation of a header list must have an even length or an `InvalidArgumentError` will be thrown.
+
+Keys are lowercase and values are not modified. If you don't specify a `host` header, it will be derived from the `url` of the [Client](#class-client) instance.
```
ronag, line 859: This is not true for dispatch nor request headers (only response headers)

docs/Client.md
Ethan-Arrowood: Whats a better way to demonstrate a basic GET request for the stream method? Totally open to suggestions - my only requirement is that its a completely executable code block so users can copy-paste and run it themselves directly.

ronag: just write it to a fs.createWriteStream?

ronag:
```js

({ opaque: filename, headers, statusCode }) => {
  // check status code

  // check headers for e.g. gzip compression

  return fs.createWriteStream(filename)
})
```

docs/Client.md
Ethan-Arrowood: Yeah not sure how to best denote events. They are also missing examples I will add now

docs/Client.md
Ethan-Arrowood: This line was copied from https://github.com/nodejs/undici#clientrequestopts-callbackerr-data-promisevoid - i'll remove it now

docs/Client.md
ronag: or 

```js
let str = ''
({ opaque: filename, headers, statusCode }) => {
  // check status code

  // check headers for e.g. gzip compression

  return new Writable({
    write (chunk, encoding, callback) {
       str += chunk
    }
  })
})

```

docs/Client.md
ronag: Parameters? events are basically the same as function invocations.

Ethan-Arrowood: Okay I found two disconnect event instances:

https://github.com/nodejs/undici/blob/10486556f3a2c40a7e90851da7867d1868fb0027/lib/core/client.js#L493

https://github.com/nodejs/undici/blob/10486556f3a2c40a7e90851da7867d1868fb0027/lib/core/client.js#L823

One calls with `client, error` and the other just `error`. Is this okay? Should we fix the inconsistency and also return client in the later disconnect event?

ronag: The first one is a bug. Would you mind creating a PR for it (and any other instances)?

mcollina (approved):
lgtm

docs/Client.md
```diff
@@ -0,0 +1,989 @@
...
+      const res = Buffer.concat(data).toString('utf8')
+      console.log(`Data: ${res}`)
+      client.close()
+      server.close()
+    }
+  })
+})
+```
+
+#### Example 2 - Dispatch Upgrade Request
```
Ethan-Arrowood, line 317: @ronag @mcollina can y'all help me put together this example? I'm struggling to properly demonstrate an upgrade request using a http.Server

docs/Client.md
```diff
@@ -0,0 +1,989 @@
...
+
+Returns: `Client`
+
+### Example - Basic Client instantiation
+
+```js
+'use strict'
+const { Client } = require('undici')
+
+const client = Client('http://localhost:3000')
```
delvedor, line 88:
```suggestion
const client = new Client('http://localhost:3000')
```

docs/Client.md
```diff
@@ -0,0 +1,989 @@
...
+    }).then(({ statusCode, body }) => {
+      console.log(`response received ${statusCode}`)
+      console.log('body:', body)
+
+      client.close()
+      fastifyServer.close()
+      nodeServer.close()
+    })
+  })
+})
```
delvedor, line 723: Here you are mixing callbacks and promises, which is generally not recommended.
Also, using `listen` and `inject` together is not very useful, how about using another undici instance for running the request?

delvedor (commented):
This documentation is awesome, fantastic work!

docs/Client.md
Ethan-Arrowood: Yes unfortunately, node `server.listen()` does not return a promise. But I think I'll modify the fastify piece to be completely promise based + incorporate another undici instance for querying fastify. Good idea!

Ethan-Arrowood:
https://github.com/nodejs/undici/pull/544 - need to make sure these changes have been taken into account

docs/Client.md
ronag: I think you need a websocket server. Just base example on ws?

Ethan-Arrowood:
FYI, once this lands I will begin work on Pool, Errors, and Agent files. I will also probably do an Undici file which details the default export object

docs/Client.md
Ethan-Arrowood: I'm struggling to get the node.js http.server to fire the `'upgrade'` event.

I've tried a dispatch call with `method: 'CONNECT'` and `upgrade: 'websocket'` but it errors with `"SocketError: other side closed"`.

I also tried a GET request with connection and upgrade headers but those trigger an InvalidArgumentError 

🤷‍♂️

Ethan-Arrowood: OH I need to use `method: 'GET'` and `upgrade: 'websocket'` - got it now!

Ethan-Arrowood:
Alright folks - I think this is good to go! Thank you for all the fantastic reviews 😄  I'm really happy to have so many more examples now. It will really help new users!

docs/Client.md
```diff
@@ -0,0 +1,1011 @@
...
+    - [Example 2 - Array](#example-2---array)
+
+## `new Client(url, [options])`
+
+Arguments:
+
+* **url** `URL | string` - It should only include the protocol, hostname, and port.
+* **options** `object` (optional)
+  * **bodyTimeout** `number | null` (optional) - Default: `30e3` -
+  * **headersTimeout** `number | null` (optional) - Default: `30e3` - Node.js v14+ - The amount of time the parser will wait to receive the complete HTTP headers. Defaults to 30 seconds.
```
ronag, line 71: No longer requires Node 14+

docs/Client.md
```diff
@@ -0,0 +1,1011 @@
...
+    - [Example 1 - Object](#example-1---object)
+    - [Example 2 - Array](#example-2---array)
+
+## `new Client(url, [options])`
+
+Arguments:
+
+* **url** `URL | string` - It should only include the protocol, hostname, and port.
+* **options** `object` (optional)
+  * **bodyTimeout** `number | null` (optional) - Default: `30e3` -
```
ronag, line 70: missing explanation

docs/Client.md
```diff
@@ -0,0 +1,1011 @@
...
+  * **headersTimeout** `number | null` (optional) - Default: `30e3` - Node.js v14+ - The amount of time the parser will wait to receive the complete HTTP headers. Defaults to 30 seconds.
+  * **keepAliveMaxTimeout** `number | null` (optional) - Default: `600e3` - The maximum allowed `idleTimeout` when overriden by *keep-alive* hints from the server. Defaults to 10 minutes.
```
ronag, line 72:
```suggestion
  * **keepAliveMaxTimeout** `number | null` (optional) - Default: `600e3` - The maximum allowed `keepAliveTimeout` when overriden by *keep-alive* hints from the server. Defaults to 10 minutes.
```

ronag: Add reference to https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Keep-Alive#directives

docs/Client.md
```diff
@@ -0,0 +1,1011 @@
...
+  * **keepAliveTimeout** `number | null` (optional) - Default: `4e3` - The timeout after which a socket without active requests will time out. Monitors time between activity on a connected socket. This value may be overriden by *keep-alive* hints from the server. Defaults to 4 seconds.
```
ronag, line 73: Add reference to developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Keep-Alive#directives

docs/Client.md
```diff
@@ -0,0 +1,1011 @@
...
+  * **keepAliveTimeoutThreshold** `number | null` (optional) - Default: `1e3` - A number subtracted from server *keep-alive* hints when overriding `idleTimeout` to account for timing inaccuries caused by e.g. transport latency. Defaults to 1 second.
```
ronag, line 74:
```suggestion
  * **keepAliveTimeoutThreshold** `number | null` (optional) - Default: `1e3` - A number subtracted from server *keep-alive* hints when overriding `keepAliveTimeout ` to account for timing inaccuries caused by e.g. transport latency. Defaults to 1 second.
```

docs/Client.md
```diff
@@ -0,0 +1,1011 @@
...
+      - [Example - Client drain event](#example---client-drain-event)
+  - [Parameter: `UndiciHeaders`](#parameter-undiciheaders)
+    - [Example 1 - Object](#example-1---object)
+    - [Example 2 - Array](#example-2---array)
+
+## `new Client(url, [options])`
+
+Arguments:
+
+* **url** `URL | string` - It should only include the protocol, hostname, and port.
```
ronag, line 68: The `It should only include the protocol, hostname, and port.` is often missed. Can we somehow emphasize it? Maybe also explain why and provide a link to `undici.request` etc... methods

docs/Client.md
```diff
@@ -0,0 +1,1011 @@
...
+  - [Instance Events](#instance-events)
+    - [Event: `'connect'`](#event-connect)
+      - [Example - Client connect event](#example---client-connect-event)
+    - [Event: `'disconnect'`](#event-disconnect)
+      - [Example - Client disconnect event](#example---client-disconnect-event)
+    - [Event: `'drain'`](#event-drain)
+      - [Example - Client drain event](#example---client-drain-event)
+  - [Parameter: `UndiciHeaders`](#parameter-undiciheaders)
+    - [Example 1 - Object](#example-1---object)
+    - [Example 2 - Array](#example-2---array)
```
ronag, line 62: The "static" methods are not documented?

docs/Client.md
```diff
@@ -0,0 +1,1011 @@
...
+
```
ronag, line 63: Agent is not documented?

ronag, line 63: Pool is not documented?

docs/Client.md
```diff
@@ -0,0 +1,1011 @@
...
+  * **maxHeaderSize** `number | null` (optional) - Default: `16384` - The maximum length of request headers in bytes. Defaults to 16KiB.
+  * **pipelining** `number | null` (optional) - Default: `1` - The amount of concurrent requests to be sent over the single TCP/TLS connection according to [RFC7230](https://tools.ietf.org/html/rfc7230#section-6.3.2). Carefully consider your workload and environment before enabling concurrent requests as pipelining may reduce performance if used incorrectly. Pipelining is sensitive to network stack settings as well as head of line blocking caused by e.g. long running requests. Set to `0` to disable keep-alive connections.
+  * **socketPath** `string | null` (optional) - Default: `null` - An IPC endpoint, either Unix domain socket or Windows named pipe.
+  * **tls** `TlsOptions | null` (optional) - Default: `null` - An options object which in the case of `https` will be passed to [`tls.connect`](https://nodejs.org/api/tls.html#tls_tls_connect_options_callback).
+
+Returns: `Client`
+
+### Example - Basic Client instantiation
+
```
ronag, line 83: Maybe explain that this initializes a client. However, it won't connect until something is queued.

docs/Client.md
```diff
@@ -0,0 +1,1011 @@
...
+      // This waits for the previous request to complete
+      console.log('Client closed')
+      server.close()
+    })
+
+  request.then(({ body }) => {
+    body.setEncoding('utf8')
+    body.on('data', console.log) // This logs before 'Client closed'
+  })
+})
```
ronag, line 133: Wouldn't it be nicer to implement in terms of async/await?

docs/Client.md
```diff
@@ -0,0 +1,1011 @@
...
+
+Closes the client and gracefully waits for enqueued requests to complete before resolving.
+
+Arguments:
+
+* **callback** `(error: Error | null, data: null) => void` (optional)
+
+Returns: `void | Promise<null>` - Only returns a `Promise` if no `callback` argument was passed
+
+#### Example - Request resolves before Client closes
```
ronag, line 103: It's a bit weird that this kind of edge case is the first example.

docs/Client.md
Ethan-Arrowood: Yeah this is Client only for now. Wanted to focus on one at a time

docs/Client.md
```diff
@@ -0,0 +1,1011 @@
...
+  request.then(({ body }) => {
+    body.setEncoding('utf8')
+    body.on('data', console.log) // This logs before 'Client closed'
+  })
+})
+```
+
+### `Client.connect(options [, callback])`
+
+Starts two-way communications with the requested resource.
```
ronag, line 138: Reference https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/CONNECT?

It's a bit ambigious with "Starts two-way communications with the requested resource." otherwise.

docs/Client.md
```diff
@@ -0,0 +1,1011 @@
...
+```js
+'use strict'
+const { createServer } = require('http')
+const { Client } = require('undici')
+
+const server = createServer((request, response) => {
+  throw Error('should never get here')
+})
+
+server.on('connect', (req, socket, firstBodyChunk) => {
```
ronag, line 172: I don't think we do `firstBodyChunk`. That should not be part of signature.

docs/Client.md
```diff
@@ -0,0 +1,1011 @@
...
+    })
+
+  // The request promise will reject with an Undici Client Destroyed error
+  request.catch(error => {
+    console.error(error)
+  })
+})
+```
+
+### `Client.dispatch(options, handlers)`
```
ronag, line 258: It would be nice to put this in a separate section for library developers?

docs/Client.md
```diff
@@ -0,0 +1,1011 @@
...
+* **signal** `AbortSignal | events.EventEmitter | null` (optional) - Default: `null`
+
+The `RequestOptions.method` property should not be value `'CONNECT'`.
+
+#### Parameter: `ResponseData`
+
+* **statusCode** `number`
+* **headers** `http.IncomingHttpHeaders`
+* **body** `stream.Readable`
+* **trailers** `Record<string, string>` - This object starts out
```
ronag, line 497: I think `Record` means immutable, no? In this case that is not true.

docs/Client.md
```diff
@@ -0,0 +1,1011 @@
...
+
+* **opaque** `unknown` (optional) - Default: `null` - Used for passing through context to `ResponseData`
+* **signal** `AbortSignal | events.EventEmitter | null` (optional) - Default: `null`
+
+The `RequestOptions.method` property should not be value `'CONNECT'`.
+
+#### Parameter: `ResponseData`
+
+* **statusCode** `number`
+* **headers** `http.IncomingHttpHeaders`
```
ronag, line 495: where/how is this declared?

ronag: I think these (and others such as stream.Readable etc...) should have a link.

docs/Client.md
```diff
@@ -0,0 +1,1011 @@
...
+
+  client.request({
+    path: '/',
+    method: 'GET',
+    signal: abortController.signal
+  }).catch(error => {
+    console.error(error) // should print an RequestAbortedError
+    client.close()
+    server.close()
+  })
```
ronag, line 560: I think it's better to use the callback API in the examples that don't use async/await.

docs/Client.md
```diff
@@ -0,0 +1,1011 @@
...
+})
+```
+
+### `Client.stream(options, factory [, callback])`
+
+A faster version of `Client.request`. This method expects the second argument `factory` to return a [`Writable`](https://nodejs.org/api/stream.html#stream_class_stream_writable) stream which the response will be written to. This improves performance by avoiding creating an intermediate [`Readable`](https://nodejs.org/api/stream.html#stream_readable_streams) stream when the user expects to directly pipe the response body to a [`Writable`](https://nodejs.org/api/stream.html#stream_class_stream_writable) stream.
```
ronag, line 626:
```suggestion
A faster version of `Client.request`. This method expects the second argument `factory` to return a [`Writable`](https://nodejs.org/api/stream.html#stream_class_stream_writable) stream which the response will be written to. This improves performance by avoiding creating an intermediate [`Readable`](https://nodejs.org/api/stream.html#stream_readable_streams) stream when the user expects to directly pipe the response body to a [`stream.Writable`](https://nodejs.org/api/stream.html#stream_class_stream_writable) stream.
```

docs/Client.md
```diff
@@ -0,0 +1,1011 @@
...
+
+As demonstrated in [Example 1 - Basic GET stream request](#example-1---basic-get-stream-request), it is recommended to use the `option.opaque` property to avoid creating a closure for the `factory` method. This pattern works well with Node.js Web Frameworks such as [Fastify](https://fastify.io). See [Example 2 - Stream to Fastify Response](#example-2---stream-to-fastify-response) for more details.
+
+Arguments:
+
+* **options** `RequestOptions`
+* **factory** `(data: StreamFactoryData) => stream.Writable`
```
ronag, line 633:
```suggestion
* **factory** `(data: StreamFactoryData) => [stream.Writable](https://nodejs.org/api/stream.html#stream_class_stream_writable)`
```

docs/Client.md
```diff
@@ -0,0 +1,1011 @@
...
+const fastify = require('fastify')
+
+const nodeServer = createServer((request, response) => {
+  response.end('Hello, World! From Node.js HTTP Server')
+})
+
+nodeServer.listen(() => {
+  console.log('Node Server listening')
+
+  const nodeServerUndiciClient = new undici.Client(`http://localhost:${nodeServer.address().port}`)
```
ronag, line 708: Should we use static or instance methods in examples like this? I think static makes it easier.

docs/Client.md
```diff
@@ -0,0 +1,1011 @@
...
+        fastifyServer.close()
+        nodeServer.close()
+      })
+    })
+})
+```
+
+### `Client.upgrade(options[, callback])`
+
+Upgrade the client to a different protocol.
```
ronag, line 749: Reference to https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism

docs/Client.md
```diff
@@ -0,0 +1,1011 @@
...
+
+* `number`
+
+Number of queued requests.
+
+### `Client.pipelining`
+
+* `number`
+
+Property to get and set the pipelining factor.
```
ronag, line 844: Is there a more visible way to show whether a property is both get and set?

ronag (commented):
Very promising. Left some more comments.

docs/Client.md
Ethan-Arrowood: Do you mean the ones from Agent? Those will come in a separate PR.

docs/Client.md
Ethan-Arrowood: Will contribute those docs in a follow up

docs/Client.md
Ethan-Arrowood: Yeah we can emphasize this just by bolding that text so its more obvious. For further explanations and other options I think that should belong in an example section or in another document. This page is primarily focussed on being an API doc for those who want the core details without reading the source code.

docs/Client.md
Ethan-Arrowood: Yeah I would have liked to use async/await over all of these examples, but since Node.js http is not promise based out-of-the-box I stuck to callbacks and some promise chains.

docs/Client.md
Ethan-Arrowood: So yes I agree this is not the best first example, but I wanted to keep the API docs alphabetical and then have each method's examples be specific to that method. Thus, if a user lands on this doc page and goes to this doc section for this specific method they will get the exact information they are looking for.

For a more general purpose "how to use" documentation I'd rather have that be its own file or maybe replace our current readme.

docs/Client.md
Ethan-Arrowood: Good catch - this should be called `head`

docs/Client.md
Ethan-Arrowood: Yeah I touched on this earlier in my replies, but to reiterate...

This documentation page is supposed to be a one stop place for all API information. For more generalized documentation - that should be captured in another file or in another, higher level section on this same page.

docs/Client.md
Ethan-Arrowood: I do not believe this is the case. AFAIK this is the recommended way to represent an object in TS. It declares the type of the keys and values so that users cannot do something like: `trailers[99] = ['a', 1, true]`

docs/Client.md
Ethan-Arrowood: My reasoning for using this type is that for most node.js http header objects there is atleast some verification that the user is passing in strings, thus its commonly typed as `Record<string, string>`. But, since this parameter directly corresponds to http headers, the Node.js type `http.IncomingHttpHeaders` is useful for saying `headers` is an object that has keys that match this list of http headers, and the values are also strings. Plus, it supports custom headers as well. 

TLDR: whenever I use http headers and typescript I rely on that Node.js http headers type and so I thought it would be good to include here as well since node is already a dependency. I'd say we need more TS user input to change this to something more general like `Record<string, string>` or just `object`

Ethan-Arrowood: > I think these (and others such as stream.Readable etc...) should have a link.

Yes i agree. still figuring out the best way to represent that in these sections. I'm thinking since you can't embed a markdown link within the code block itself, then I will include a `Refs: ` section which is a list of links for types mentioned in the relative definition block. Will contribute those changes shortly to get your feedback

docs/Client.md
Ethan-Arrowood: Okay good point - I'd prefer async-await, but `server.listen` only supports that callback (since I believe its actually assigned to the server as an event listener for `server.listen`. 

I don't know off the top of my head if an event listener can be an `async` function. I'm going to research that next and make changes accordingly.

docs/Client.md
Ethan-Arrowood: Unfortunately links cannot be included in code blocks https://stackoverflow.com/questions/20092433/placing-links-inside-markdown-code-blocks

docs/Client.md
Ethan-Arrowood: I think I answered this previously, but I'll reiterate anyways. I'd like to keep these embedded examples as close to the relative method so that users can see how they are used independently. For "best way to use undici" we should have a separate doc page/section.

docs/Client.md
Ethan-Arrowood: I could move these sections to be above the _Instance Methods_ so that the user sees them first when scrolling through. And then for like getter/setter indication I'm not entirely sure.

docs/Client.md
Ethan-Arrowood: https://nodejs.org/api/events.html#events_capture_rejections_of_promises

It seems like its a bit dangerous to use async with event listeners ☹️

docs/Client.md
ronag: Also I the best way to handle head is to do:

```js
if (head.length) {
  socket.unshift(head)
}
```

docs/Client.md
ronag: Ah, I might be confusing this with a ecmascript proposal which defines `Record` as immutable.

docs/Client.md
```diff
@@ -0,0 +1,1013 @@
...
+### Event: `'disconnect'`
+
+Parameters:
+
+* **client** `Client`
+* **error** `Error`
+
+Emitted when socket has disconnected. The error argument of the event is the error which caused the socket to disconnect. The client will reconnect if or once `client.size > 0`.
+
+#### Example - Client disconnect event
```
ronag, line 913:
```suggestion
#### Example - Client 'disconnect' event
```

docs/Client.md
ronag: I'm mostly concerned about the set/get. Not a blocker.

ronag (approved):

ronag:
@Ethan-Arrowood is this ready to merge?

Ethan-Arrowood:
Yes I believe so! As is the nature with this project - let's keep iterating quickly on these docs. I'll have prs up for the other pages soon!

PR #555, opened 2021-02-17
fix missing client arg

### LINKED ISSUE #554: disconnect event missing args
https://github.com/nodejs/undici/pull/550#discussion_r577715865

> Okay I found two disconnect event instances:
> 
> https://github.com/nodejs/undici/blob/10486556f3a2c40a7e90851da7867d1868fb0027/lib/core/client.js#L493
> 
> https://github.com/nodejs/undici/blob/10486556f3a2c40a7e90851da7867d1868fb0027/lib/core/client.js#L823
> 
> One calls with `client, error` and the other just `error`. Is this okay? Should we fix the inconsistency and also return client in the later disconnect event?

Ethan-Arrowood:
closes #554 

doc update included in #550

### DIFF
--- a/lib/core/client.js
+++ b/lib/core/client.js
@@ -490,7 +490,7 @@ class Parser extends HTTPParser {
           client[kSocket] = null
 
           client[kQueue][client[kRunningIdx]++] = null
-          client.emit('disconnect', new InformationalError('upgrade'))
+          client.emit('disconnect', client, new InformationalError('upgrade'))
         }
 
         resume(client)
--- a/test/client-upgrade.js
+++ b/test/client-upgrade.js
@@ -383,6 +383,36 @@ test('basic upgrade error', (t) => {
   })
 })
 
+test('upgrade disconnect', (t) => {
+  t.plan(3)
+
+  const server = net.createServer(connection => {
+    connection.destroy()
+  })
+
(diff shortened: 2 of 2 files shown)


ronag:
Would be nice with a test. Is this the only occurance?

Ethan-Arrowood:
Yeah only occurrence, most of the tests use neither arguments in the event listener, but there is one test that checks them: https://github.com/nodejs/undici/blob/master/test/pool.js#L39

ronag:
In the upgrade test can you add a check for disconnect + correct args.

mcollina (changes requested):

Ethan-Arrowood:
Also found another test for this but its not upgrade related so this new case is a good addition

https://github.com/nodejs/undici/blob/master/test/client.js#L980

mcollina (approved):
lgtm

mcollina:
I _think_ this is semver-major cc @ronag.

ronag:
> I _think_ this is semver-major cc @ronag.

This is kind of a bug of a previous semver-major. I'm fine with landing as a patch if @mcollina is.

ronag (approved):

ronag:
It is semver major. A fix for https://github.com/nodejs/undici/pull/544 which we haven't released,

PR #559, opened 2021-02-21
fix: omit 'path' in undici.request opts (ts)

apostolos:
In `undici.request|stream|pipeline` helper functions, the opts param may not contain path.

see https://github.com/nodejs/undici#undicirequesturl-opts-promise

Without this fix you get the following Typescript error:

image

But if you pass the path parameter you get a runtime error: https://github.com/nodejs/undici/blob/master/lib/agent.js#L86-L88

### DIFF
--- a/test/types/agent.test-d.ts
+++ b/test/types/agent.test-d.ts
@@ -18,7 +18,7 @@ expectAssignable<Agent>(new Agent({}))
 
 {
   expectAssignable<PromiseLike<Client.ResponseData>>(request(''))
-  expectAssignable<PromiseLike<Client.StreamData>>(stream('', { path: '', method: '' }, data => {
+  expectAssignable<PromiseLike<Client.StreamData>>(stream('', { method: '' }, data => {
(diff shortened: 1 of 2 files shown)


Ethan-Arrowood:
Unfortunately the types are quite out of date from the v3 API. I'm okay with this change though to bring it closer - I'm fairly confident not a lot of folks are using undici in TS projects just yet

mcollina (approved):
lgtm

mcollina (changes requested):
Could you fix the types tests?

apostolos:
@mcollina type tests updated

mcollina (approved):
lgtm

@Ethan-Arrowood could you take a look?

Ethan-Arrowood (approved):

thornyweb:
Hello! I've just run in to this issue in our TS project :)

I can see this code has been merged but not deployed yet, do you have a planned release date at all for this fix?

Thank you!

PR #564, opened 2021-02-24
Remove process.binding()

mcollina:
As Node.js prepares to deprecate `process.binding()`, we can still use `_http_common` to get the `HTTPParser`.

Ref: https://github.com/nodejs/node/pull/37485.


ronag:
Fails 😕

ronag (approved):

mcollina:
Landing, it seems mac os x hosts have some issues.

PR #565, opened 2021-02-24
Pool docs

Ethan-Arrowood:
I know this page might seem redundant, but I think it is the best way to document `Pool`. This clearly shows the user what methods, properties, and events are available to them, and provides easy links to the relevant part of the `Client` documentation. 

Reviewers:

If there is anything specific we should highlight for any of the properties or methods in regards to descriptions we should include that. For example, we may want to include descriptions for properties like `pending`, `busy`, etc. specifying how they are evaluated for the pool of clients.

### DIFF
--- a/docs/Pool.md
+++ b/docs/Pool.md
@@ -17,7 +17,7 @@ Arguments:
 
 Extends: [`ClientOptions`](./Client.md#parameter-clientoptions)
 
-* **connections** `number | null` (optional) - Default: `null` - The number of `Client` instances to create.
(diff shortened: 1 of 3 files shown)


docs/Pool.md
```diff
@@ -0,0 +1,107 @@
+# Class: Pool
```
mcollina, line 1: Should this be linked from the README?

Ethan-Arrowood: Yes, once I finish the Agent file I'm going to rip everything out of the current readme and replace it with links to these documents (+ include some other things I've been thinking about for the main readme file)

ronag (approved):

mcollina (approved):
lgtm

PR #566, opened 2021-02-24
fix: use queueMicrotask

ronag:
Before:

```
http - no agent  x 656 ops/sec ±1.57% (78 runs sampled)
http - keepalive x 684 ops/sec ±1.42% (77 runs sampled)
http - keepalive - multiple sockets x 4,968 ops/sec ±2.52% (78 runs sampled)
undici - pipeline x 5,839 ops/sec ±2.36% (80 runs sampled)
undici - request x 6,087 ops/sec ±1.50% (79 runs sampled)
undici - pool - request - multiple sockets x 6,126 ops/sec ±1.36% (81 runs sampled)
undici - stream x 6,457 ops/sec ±1.30% (79 runs sampled)
undici - dispatch x 6,746 ops/sec ±1.12% (82 runs sampled)
undici - noop x 6,816 ops/sec ±1.19% (82 runs sampled)
```

After:

```
http - no agent  x 670 ops/sec ±1.40% (79 runs sampled)
http - keepalive x 693 ops/sec ±1.43% (75 runs sampled)
http - keepalive - multiple sockets x 5,043 ops/sec ±1.86% (80 runs sampled)
undici - pipeline x 5,694 ops/sec ±1.87% (79 runs sampled)
undici - request x 5,947 ops/sec ±2.01% (76 runs sampled)
undici - pool - request - multiple sockets x 6,139 ops/sec ±1.13% (82 runs sampled)
undici - stream x 6,369 ops/sec ±1.33% (79 runs sampled)
undici - dispatch x 6,780 ops/sec ±1.18% (81 runs sampled)
undici - noop x 6,715 ops/sec ±1.11% (80 runs sampled)
```

### DIFF
--- a/lib/client-upgrade.js
+++ b/lib/client-upgrade.js
@@ -57,9 +57,9 @@ class UpgradeHandler extends AsyncResource {
 
     if (callback) {
       this.callback = null
-      process.nextTick((self, callback, err, opaque) => {
-        self.runInAsyncScope(callback, null, err, { opaque })
-      }, this, callback, err, opaque)
+      util.queueMicrotask(() => {
+        this.runInAsyncScope(callback, null, err, { opaque })
+      })
     }
   }
 }
@@ -94,7 +94,9 @@ function upgrade (opts, callback) {
       upgrade: protocol || 'Websocket'
     }, upgradeHandler)
   } catch (err) {
-    process.nextTick(callback, err, { opaque: opts && opts.opaque })
+    util.queueMicrotask(() => {
+      callback(err, { opaque: opts && opts.opaque })
+    })
   }
 }
 
--- a/lib/client-connect.js
+++ b/lib/client-connect.js
@@ -55,9 +55,9 @@ class ConnectHandler extends AsyncResource {
 
     if (callback) {
       this.callback = null
-      process.nextTick((self, callback, err, opaque) => {
-        self.runInAsyncScope(callback, null, err, { opaque })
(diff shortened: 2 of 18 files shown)


mcollina:
There seems to be some degradation of performance.

ronag:
Negligible?

mcollina:
> Negligible?

I don't see the benefit. Why replacing something with a slower alternative that does the same thing?

ronag:
> I don't see the benefit. Why replacing something with a slower alternative that does the same thing?

Follows best practice and removes the chance for edge cases related to timing with promises and async/await. I think it adds more maintenance overhead to have to mix the two in the code base and consider when to use which.

No strong opinion though. I'm ok with closing.

mcollina:
I would prefer to close for now.

lib/client-upgrade.js
```diff
@@ -57,9 +57,9 @@ class UpgradeHandler extends AsyncResource {
 
     if (callback) {
       this.callback = null
-      process.nextTick((self, callback, err, opaque) => {
-        self.runInAsyncScope(callback, null, err, { opaque })
-      }, this, callback, err, opaque)
+      util.queueMicrotask(() => {
```
martinheidegger, line 60: Did you try if `.bind` instead of the array function makes a performance difference?

szmarczak:
I agree with @ronag. IMO it's just a margin of error. And if it's not, I wouldn't mind losing 100 req/s just to be more stable.

PR #567, opened 2021-02-25
Client lifecycle diagram

### LINKED ISSUE #532: API Lifecycle Diagram
I think something that would be really valuable for this library is if we can provide a detailed diagram explaining the lifecycle of a request with this library. It would be really helpful for both contributors and users. The diagram should detail things like when certain events are fired as well as how it all corresponds to the Node.js event loop. We should have one for the Client class and its respective methods, another for Pool (which can reference the Client one), and then maybe another for Agent which is based on the Pool and Client ones.

I'd be happy to collaborate on this, but I'd definitely need some help from @ronag or @mcollina. I'm getting better at understanding the event loop but I'm not all the way there yet 😅

Ethan-Arrowood:
Closes #532 

Opening as a draft to start getting some feedback. I've read through the source code, tests, and existing documentation to come up with this. It may not be correct + I feel like I want to expand the "processing" state since there is a lot that happens inside of that. 

I'd love to integrate parts of the API here too like specific methods calls, emitted events, and property values


ronag (commented):
It won’t disconnect until keep alive timeout.

Ethan-Arrowood:
> It won’t disconnect until keep alive timeout.

Yes good point. I'll include some more details one that

ronag:
I thinking you should have another state which is just connected.

Idle - pending - running - keep alive - closing - destroyed

Ethan-Arrowood:
> I thinking you should have another state which is just connected.
> Idle - pending - running - keep alive - closing - destroyed

From what I can tell the Client never actually sits in a "connected" state. Once `dispatch` is called the Client immediately creates that socket connection and begins processing the request; even if it does not resolve the response back to the user immediately (as in the case of the async functions)

docs/api-lifecycle.md
```diff
@@ -0,0 +1,13 @@
...
+
+[![A state diagram representing an Undici Client instance](https://mermaid.ink/img/eyJjb2RlIjoic3RhdGVEaWFncmFtLXYyXG4gICAgWypdIC0tPiBpZGxlXG4gICAgcHJvY2Vzc2luZyAtLT4gY2xvc2luZyA6IGNsb3NlXG4gICAgaWRsZSAtLT4gY2xvc2luZyA6IGNsb3NlXG4gICAgaWRsZSAtLT4gZGVzdHJveWVkIDogZGVzdHJveVxuICAgIGlkbGUgLS0-IHByb2Nlc3NpbmcgOiBjb25uZWN0XG4gICAgcHJvY2Vzc2luZyAtLT4gaWRsZSA6IGRpc2Nvbm5lY3RcbiAgICBwcm9jZXNzaW5nIC0tPiBkZXN0cm95ZWQgOiBkZXN0cm95XG4gICAgY2xvc2luZyAtLT4gZGVzdHJveWVkIDogZGVzdHJveVxuICAgIGRlc3Ryb3llZCAtLT4gWypdIiwibWVybWFpZCI6eyJ0aGVtZSI6ImJhc2UifSwidXBkYXRlRWRpdG9yIjpmYWxzZX0)](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoic3RhdGVEaWFncmFtLXYyXG4gICAgWypdIC0tPiBpZGxlXG4gICAgcHJvY2Vzc2luZyAtLT4gY2xvc2luZyA6IGNsb3NlXG4gICAgaWRsZSAtLT4gY2xvc2luZyA6IGNsb3NlXG4gICAgaWRsZSAtLT4gZGVzdHJveWVkIDogZGVzdHJveVxuICAgIGlkbGUgLS0-IHByb2Nlc3NpbmcgOiBjb25uZWN0XG4gICAgcHJvY2Vzc2luZyAtLT4gaWRsZSA6IGRpc2Nvbm5lY3RcbiAgICBwcm9jZXNzaW5nIC0tPiBkZXN0cm95ZWQgOiBkZXN0cm95XG4gICAgY2xvc2luZyAtLT4gZGVzdHJveWVkIDogZGVzdHJveVxuICAgIGRlc3Ryb3llZCAtLT4gWypdIiwibWVybWFpZCI6eyJ0aGVtZSI6ImJhc2UifSwidXBkYXRlRWRpdG9yIjpmYWxzZX0)
```
mcollina, line 5: I would prefer if the image and source to be inside the repo

mcollina (changes requested):

Ethan-Arrowood:
[](https://mermaid-js.github.io/mermaid-live-editor/#/edit/eyJjb2RlIjoic3RhdGVEaWFncmFtLXYyXG4gICAgWypdIC0tPiBpZGxlXG4gICAgcHJvY2Vzc2luZyAtLT4gY2xvc2luZyA6IGNsb3NlXG4gICAgaWRsZSAtLT4gY2xvc2luZyA6IGNsb3NlXG4gICAgaWRsZSAtLT4gZGVzdHJveWVkIDogZGVzdHJveVxuICAgIGlkbGUgLS0-IHByb2Nlc3NpbmcgOiBjb25uZWN0XG4gICAgcHJvY2Vzc2luZyAtLT4gaWRsZSA6IGRpc2Nvbm5lY3RcbiAgICBwcm9jZXNzaW5nIC0tPiBkZXN0cm95ZWQgOiBkZXN0cm95XG4gICAgY2xvc2luZyAtLT4gZGVzdHJveWVkIDogZGVzdHJveVxuICAgIGRlc3Ryb3llZCAtLT4gWypdXG5cbiAgICBzdGF0ZSBwcm9jZXNzaW5nIHtcbiAgICAgICAgWypdIC0tPiByZXN1bWluZ1xuICAgICAgICByZXN1bWluZyAtLT4gZHJhaW5pbmcgOiBkcmFpblxuICAgICAgICBkcmFpbmluZyAtLT4gcmVzdW1pbmdcbiAgICAgICAgcmVzdW1pbmcgLS0-IHJlc3VtaW5nXG4gICAgICAgIHJlc3VtaW5nIC0tPiBbKl1cbiAgICB9IiwibWVybWFpZCI6eyJ0aGVtZSI6ImJhc2UifSwidXBkYXRlRWRpdG9yIjpmYWxzZX0)

@ronag what do you think about this breakdown of the **processing** state?

ronag:
@Ethan-Arrowood maybe we could have a voice call? How does your day look today?

Ethan-Arrowood:
Im free for the next couple hours! Send me an invite for whenever: ethan.arrowood@microsoft.com

ronag:
We had a call and discussed the life cycle and also other ways to make the code easier to understand.

Ethan-Arrowood:
@ronag when a Client is "connected" (the socket connection has been established), can requests be queued without being processed? Or is the action of dispatching a request to a client always going to immediately kick off the processing/running actions?

Ethan-Arrowood:
Updated the diagram based on our conversation @ronag - I'm working on the detailed state descriptions now so its better understood as well

mcollina (changes requested):
In the `processing` sub-diagram, you need a keep-alive state that goes back to the beginning or times out.

Ethan-Arrowood:
> In the processing sub-diagram, you need a keep-alive state that goes back to the beginning or times out.

IMO this is what the `pending` state represents, connected but not processing requests. The `keepAlive` event is what transitions the `Client` from `processing` back to `pending`. And when that socket connection timesout, the `timeOut` event transitions from `pending` to `init`.

mcollina:
I got confused by having both done and keep alive ending in the same dot.

mcollina (approved):
lgtm

docs/api-lifecycle.md
```diff
@@ -0,0 +1,62 @@
+# Client Lifecycle
+
+An Undici [Client](Client.md) can be best described as a state machine. The following list is a summary of the various state transitions the `Client` will go through in its lifecycle. This document also contains detailed breakdowns of each state.
+
+> This diagram is not a perfect representation of the undici Client. Since the Client class is not actually implemented as a state-machine, actual execution may deviate slightly from what is described below. Consider this as a general resource for understanding the inner workings of the Undici client rather than some kind of formal specification.
+
+## State Transition Overview
+
+* A `Client` begins in the **idle** state with no socket connection and no requests in queue.
+  * The *connect* event transitions the `Client` to the **pending** state where requests can be queued prior to processing
```
antsmartian, line 10: `can be queued prior to processing` should be `can be queued prior to processing.` (full stop is missing)

mcollina (approved):
still lgtm

Ethan-Arrowood:
Okay _now_ i think this is good to go. I messed things up with docsify links in the initial site build. But this should clean up everything

PR #568, opened 2021-02-26
Connecting through a proxy

### LINKED ISSUE #427: Connect through a proxy
There are some use cases where users need to connect to the upstream server via a proxy that is not handling connections transparently.

I would like to add this functionality directly into Undici, and from the user perspective, it would look like this:
```js
const client = new Client({
  url: 'http://localhost:3000',
  proxy: 'http://localhost:3030'
})
```
Supporting this feature is also useful for working with proxy for debugging purposes, such as [mitmproxy](https://mitmproxy.org/).
I've implemented a similar solution for the Node.js's core HTTP client via a [custom agent](https://github.com/delvedor/hpagent).

What do you think?

delvedor:
This pr adds a test and the documentation to show how to connect through a proxy :)
I had to remove the super complex regex as it was not validating all absolute urls (eg: `http://localhost:8080` didn't pass the check). I think that lazily checking the protocol is enough, as trying to save *all users* from themselves is impossible.

/cc @szmarczak
Closes #427

### DIFF
--- a/docs/proxy.md
+++ b/docs/proxy.md
@@ -1,8 +1,8 @@
 # Connecting through a proxy
 
-Conneting through a proxy is possible by properly configuring the client constructor and request.
+Conneting through a proxy is possible by properly configuring the `Client` or `Pool` constructor and request.
 
-The proxy url should be passed to the client constructor, while the upstream server url
+The proxy url should be passed to the `Client` or `Pool` constructor, while the upstream server url
 should be added to every request call in the `path`.
 For instance, if you need to send a request to the `/hello` route of your upstream server,
 the `path` should be `path: 'http://upstream.server:port/hello?foo=bar'`.
@@ -14,7 +14,7 @@ If you proxy requires basic authentication, you can send it via the `proxy-autho
 ```js
 import { Client } from 'undici'
 import { createServer } from 'http'
-import proxy = from 'proxy'
+import proxy from 'proxy'
 
 const server = await buildServer()
 const proxy = await buildProxy()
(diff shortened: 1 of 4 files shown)


ronag (approved):

docs/proxy.md
```diff
@@ -0,0 +1,123 @@
...
+the `path` should be `path: 'http://upstream.server:port/hello?foo=bar'`.
+
+If you proxy requires basic authentication, you can send it via the `proxy-authorization` header.
+
+### Connect without authentication
+
+```js
+import { Client } from 'undici'
+import { createServer } from 'http'
+import proxy = from 'proxy'
```
mcollina, line 17: this is not valid javascript but typescript.

mcollina (commented):
Would you mind to show how to is with `Pool`?  

It would really be nice to also include some ways of having a `ProxyAgent`.

docs/proxy.md
delvedor: That's actually a typo! :D

delvedor:
It works in the exact same way with both `Client` and `Pool`, so I've added an additional test and a note in the docs :)

mcollina:
Here is an example for the Agent API:

```js
import { request, setGlobalAgent, ProxyAgent } from 'undici'

setGlobalAgent(new ProxyAgent('https://my.proxy.agent/'))

const {
  statusCode,
  headers,
  trailers,
  body
  // send the request via the Agent
} = await request('http://localhost:3000/foo')

  console.log('data', data)
}

console.log('trailers', trailers)
```

mcollina:
FYI, you don't need to add the ProxyAgent right now. It can be another issue.

mcollina (approved):
lgtm

budarin:
is there an example with proxying a POST request?

mcollina:
No but it would be a good PR.

PR #571, opened 2021-02-26
use parser.execute

ronag:


ronag:
@dnlup I think this will help with the wasm stuff you are working on. Removes usage of `kOnExecute` in favor of `parser.execute`.

ronag:
https://github.com/nodejs/undici/issues/22

lib/client-request.js
```diff
@@ -112,13 +112,14 @@ class RequestHandler extends AsyncResource {
   onComplete (trailers) {
     const { res } = this
 
-    removeSignal(this)
-
-    if (trailers) {
-      util.parseHeaders(trailers, this.trailers)
-    }
+    util.queueMicrotask(() => {
```
mcollina, line 115: why this?

ronag: `onComplete` can be invoked before `await undici.request()` has continued execution. I think we can avoid the queueMicrotask before landing. For now this just gets tests to pass properly.

mcollina (approved):
lgtm

lib/client-request.js
ronag: fixed

dnlup:
> @dnlup I think this will help with the wasm stuff you are working on. Removes usage of `kOnExecute` in favor of `parser.execute`.

Yes, definitely ❤️

ronag:
Has been rebased into https://github.com/nodejs/undici/pull/575

PR #576, opened 2021-02-27
fix: parser callback expect sync return value

ronag:


lib/core/client.js
```diff
@@ -603,10 +587,44 @@ class Parser extends HTTPParser {
       }
     } catch (err) {
       util.destroy(socket, err)
+    }
+  }
+
+  [HTTPParser.kOnHeadersComplete] (versionMajor, versionMinor, rawHeaders, method,
```
ronag, line 593: This is the primary fix. Always needs to return an integer. I haven't actually been able to make a test case that fails in practice but in theory it was incorrect.

lib/core/client.js
```diff
@@ -603,10 +587,44 @@ class Parser extends HTTPParser {
...
+      return 1
+    }
+
+    if (this.paused) {
+      this.queue.push(['onHeadersComplete', rawHeaders, statusCode, shouldKeepAlive])
+    } else {
+      this.onHeadersComplete(rawHeaders, statusCode, shouldKeepAlive)
+    }
+
+    return request.upgrade ? 2 : request.method === 'HEAD' || statusCode < 200 ? 1 : 0
```
ronag, line 627: Always return an integer, even if we are paused.

mcollina (approved):
lgtm

PR #577, opened 2021-02-27
fix: eslint + prettier config

ronag:


ronag:
@Ethan-Arrowood @dnlup

mcollina (changes requested):
I think this change is unnecessary and it hampers delivering quality software on multiple lines. Backporting commits after this land will become more complicated.

ronag:
No strong opinion other than it would have been nice.

Ethan-Arrowood:
Id rather we stick to standard like we already are.

PR #585, opened 2021-03-03
init docs site

Ethan-Arrowood:
First stab at generating a docs site from the markdown documents. Docsify seems cool and I like the DX so far. I'll get a github actions thing working soon so it "auto deploys" on new commits and such

Check out the live preview: http://ethanarrowood.com/undici/#/


docs/README.md
```diff
@@ -0,0 +1,64 @@
+# undici
```
mcollina, line 1: This file seems a duplication of the project README. is this correct? Can we avoid that duplication?

Ethan-Arrowood: Yeah this is an unfortunate part of using these site generators. They sorta need everything to be included in the same directory. We could write some sort of copy script thats included in the build pipeline. From my cursory review of other options (docusaurus and vuepress), docsify is the simplest

mcollina: Check out https://github.com/mercurius-js/mercurius configuration. It does not duplicate the README.

Ethan-Arrowood: oh smart - but then the site file lives in repo root. If we are cool with that I'll do it here too

mcollina: Better than duplicating!

mcollina:
That would be amazing.

Ethan-Arrowood:
Yes I'd love to make that all happen together. I'll try and send Agent docs ASAP as well as wrap up this site deployment work

Ethan-Arrowood:
We could also use a logo 😉

dnlup (approved):
This looks amazing!

ronag:
Why does CI fail?

mcollina:
It passes now!

PR #586, opened 2021-03-04
feat: treat request content length mismatch as warning

### LINKED ISSUE #581: Feature: ability to ignore errors (strict: false)
We are seeing some errors from Undici which do not occur under any other client, e.g.

```
code: "UND_ERR_CONTENT_LENGTH_MISMATCH"
error: "Internal Server Error"
message: "Request body length does not match content-length header"
statusCode: 500
```

I have been told Undici is stricter with regard to the HTTP spec than other clients, and I think that's great. However, for the sake of compatibility with existing ecosystems, it would be great to be able to configure undici to ignore some (or all) such errors. This would let organisations make a gradual migration, or handle edge cases where these errors are not meaningful.

simoneb:
Closes #581

Adds ability to tread request content length mismatch as a warning instead of an error, via `strictContentLength` option.

### Checklist

- [x] tests
- [x] docs
- [x] types

### DIFF
--- a/README.md
+++ b/README.md
@@ -110,6 +110,8 @@ Options:
 - `maxHeaderSize: Number`, the maximum length of request headers in bytes.
   Default: `16384` (16KiB).
 
+- `strictContentLength: Boolean`, whether to treat request content length mismatches as errors. If true, an error is thrown when the request content-length header doesn't match the length of the request body. Default: `true`.
+
 <a name='request'></a>
 #### `client.request(opts[, callback(err, data)]): Promise|Void`
 
--- a/docs/Client.md
+++ b/docs/Client.md
@@ -83,6 +83,7 @@ Returns: `Client`
 * **pipelining** `number | null` (optional) - Default: `1` - The amount of concurrent requests to be sent over the single TCP/TLS connection according to [RFC7230](https://tools.ietf.org/html/rfc7230#section-6.3.2). Carefully consider your workload and environment before enabling concurrent requests as pipelining may reduce performance if used incorrectly. Pipelining is sensitive to network stack settings as well as head of line blocking caused by e.g. long running requests. Set to `0` to disable keep-alive connections.
 * **socketPath** `string | null` (optional) - Default: `null` - An IPC endpoint, either Unix domain socket or Windows named pipe.
 * **tls** `TlsOptions | null` (optional) - Default: `null` - An options object which in the case of `https` will be passed to [`tls.connect`](https://nodejs.org/api/tls.html#tls_tls_connect_options_callback).
+* **strictContentLength** `Boolean` (optional) - Default: `true` - Whether to treat request content length mismatches as errors. If true, an error is thrown when the request content-length header doesn't match the length of the request body.
 
 ### Example - Basic Client instantiation
 
--- a/lib/core/client.js
+++ b/lib/core/client.js
@@ -51,7 +51,8 @@ const {
   kIdleTimeout,
   kIdleTimeoutValue,
   kHeadersTimeout,
-  kBodyTimeout
+  kBodyTimeout,
+  kStrictContentLength
 } = require('./symbols')
 
 const nodeVersions = process.version.split('.')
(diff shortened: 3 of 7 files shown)


README.md
```diff
@@ -110,6 +110,8 @@ Options:
 - `maxHeaderSize: Number`, the maximum length of request headers in bytes.
   Default: `16384` (16KiB).
 
+- `strictContentLength: Boolean`, whether to treat request content length mismatches as errors. If true, an error is thrown when the request content-length header doesn't match the length of the request body. Default: `true`.
```
Ethan-Arrowood, line 113: Can you add this to the `docs/Client.md` file as well please?

simoneb: done

simoneb:
I'm having an issue I couldn't yet figure out with testing a specific scenario. I took this suite of tests from the original ones which check that an error is thrown when content length mismatches and I've narrowed it down to:

- a request sends streaming data **asynchronously** whose length doesn't match the content-length request header
- when that happens, _any_ request following the first will result in an error, which can be of two types:
  - when the content-length header is less than the body length, a native `ECONNRESET` error is thrown
  - when the content-length header is greater than the body length, a custon undici `SocketError('other side closed')` is thrown

Here's a sample test case for the first case above:

```js
    client.request({
      path: '/',
      method: 'PUT',
      headers: {
        'content-length': 2
      },
      body: new Readable({
        read () {
          // without `setImmediate`, both of the two cases work fine
          setImmediate(() => {
            this.push('abc')
            this.push(null)
          })
        }
      })
    }, (err) => {
      // this assertion will always succeed
      t.error(err)
    })

    client.request({
      path: '/',
      method: 'GET',
    }, (err) => {
      // this is where the error happens
      t.error(err)
    })
```

Apart from not having been able to figure out why this happens (especially around the asynchronous writing to the request body stream with `setImmediate`, without which everything works) and how to make this work, my main question is, is this actually supposed to work or sending a request to a server with a wrong content length (either shorter or longer than advertised) is causing the server to trigger e behavior that the client can't handle anyway?

ronag:
A server should probably fail if receiving a body which doesn't match the content length.

simoneb:
> A server should probably fail if receiving a body which doesn't match the content length. So I'm a little unsure what exactly the client should do here...

Yeah I was suspecting as much, and yet the behavior is different depending on _how_ that mismatching body is sent. As you can see in the other tests in this PR, the server never complaints except when the mismatching body is sent as a stream, asynchronously. In all other cases it works fine (or at least, no errors)

ronag:
> asynchronously. In all other cases it works fine (or at least, no errors)

Yes, when sending asynchronously it uses chunked encoding. I think some servers can make more loose assumption when sent normally.

simoneb:
> > asynchronously. In all other cases it works fine (or at least, no errors)
> 
> Yes, when sending asynchronously it uses chunked encoding. I think some servers can make more loose assumption when sent normally.

Thanks for clarifying, it's something I had not realized. For the sake of this change then, do you think that removing those tests sending chunked encoding is good enough? This new option is only to prevent the client from throwing an error when _request_ content length mismatches, but it's not intended to be able to handle server errors when that happens, of course.

simoneb:
Any preference about testing that the warnings are emitted when `useStrictContentLength` is off? I see `sinon` is not being used, if you're fine with that I can use sinon to mock process.emitWarning in tests and assert it's called

mcollina:
sinon is ok!

simoneb:
> sinon is ok!

done. this is ready for review

mcollina (approved):
lgtm

Ethan-Arrowood (approved):

delvedor (approved):
LGTM

SimenB:
I'm getting a

```
TypeError: The "warning" argument must be of type string or an instance of Error. Received an instance of RequestContentLengthMismatchError

at write (node_modules/undici/lib/client.js:1437:13)
at _resume (node_modules/undici/lib/client.js:1385:29)
at resume (node_modules/undici/lib/client.js:1244:3)
at connect (node_modules/undici/lib/client.js:1229:3)
```

when attempting to use this option.

simoneb:
Do you have a repro to share?

SimenB:
No real need, the issue isn't in this repo. Realized this was my old friend https://github.com/facebook/jest/issues/2549 after posting

simoneb:
Send our greetings to your friend

Issue #531, opened 2021-01-27
Mocking support


mcollina:
I think it would be awesome if we had mocking support, similar to nock.

adamovittorio:
@mcollina what can we do to help here, do you have an idea on how to do the magic? In my understanding undici is not using nodejs HTTP module under the hood, so we need to intercept the request in another way.

mcollina:
Send a PR :). I would recommend to _add_ a `MockedClient` and then some support for mocking inside `Pool`, or maybe a `MockingPool`, and the rest will follow.

adamovittorio:
What do you mean by adding a `MockedClient`? Just implementing a mocked version of the undici client?

In my understanding, we want something that intercepts the undici network calls and returns a mocked response.

```ts
const { MockClient } = require("undici");
const client = new MockClient("https://www.nodejs.org");

client.intercept({
  path: "/",
  method: "GET",
}).reply(200, { "foo": "bar"});

```
Wherever in you code you use undici to perform a **GET** https://www.nodejs.org/ the result is `{ "foo": "bar" } 200`.

Still, I don't know how to intercept the requests

mcollina:
The API you are sketching is correct!

> Still, I don't know how to intercept the requests

Neither would I! You might have to overload a few (or all) of the APIs in `Client`.

piranna:
> You might have to overload a few (or all) of the APIs in `Client`.

According to docs, it would be a matter of overwrite just only [client.dispatch()](https://github.com/nodejs/undici#clientdispatchopts-handler-void).

jonnydgreen:
I'm very happy to work on an implementation based on @adamovittorio's suggestion if that's okay with everyone? It'll include the `MockClient` definition, tests, docs and typings :)

mcollina:
Go for it @jonnydgreen! Open a draft PR as early as possible so when can provide iterative feedback.

jonnydgreen:
Awesome! I'll submit a draft PR for initial feedback as soon as possible! :)

Nytelife26:
If I may add, would it not be better to put the mocking in a separate package? Many users will not be using the mock functionality, and it is not integral to undici's function, so perhaps it should be put in an `undici-mocks` (or similar) package?

In regards to undici not using HTTP under the hood, I believe intercepting the requests at the `net` level is the better option since that would mock *all* Node network operations - both simultaneously removing the need for Nock and other individualized wrappers for various libraries.

piranna:
[...]

PR #587, opened 2021-03-05
Add undici mocking support

jonnydgreen:
Implementation of: https://github.com/nodejs/undici/issues/531


jonnydgreen:
Here is my initial implementation that overrides `client.dispatch` with some basic functionality. I thought I'd start off simple and add useful features as we go along!

One concern I have with this approach is test concurrency. If tests are run in parallel for exactly the same request key e.g. `GET http://localhost:9999/foo?hello=there`, it is hard to figure which response to use in the `mockDispatch` function. I believe this is fairly unlikely, but possible! As a compromise, I am assuming that the order of interception definition will match the called request, so I am taking the first response I can find and deleting it from the array of responses (I will mark as consumed in the future as this is a bit more flexible). I would love to hear your thoughts on this! :)

mcollina:
> One concern I have with this approach is test concurrency. If tests are run in parallel for exactly the same request key e.g. GET http://localhost:9999/foo?hello=there, it is hard to figure which response to use in the mockDispatch function. I believe this is fairly unlikely, but possible! As a compromise, I am assuming that the order of interception definition will match the called request, so I am taking the first response I can find and deleting it from the array of responses (I will mark as consumed in the future as this is a bit more flexible). I would love to hear your thoughts on this! :)

This approach is sound!

jonnydgreen:
Just added an update :) This includes:
 - `trailers` support
 - `body` support (as a query string dictated by the user)
 - `persist` support
 - `delay` support (by x ms)
 - I've refactored the code a bit to mark as `consumed` rather than deleting the entry - this gives us a lot more flexibility
 - Extra checks across the board within the `mockDispatch` and `MockClient`
 - Extra support for keeping `.close` separate from other MockClient instances

I've tried to keep the API for this fairly low level and in-line with undici so it's familiar to users of the `MockClient`. As well, I'm thinking we only support basic responses for now - what are your thoughts on this? Are there any other particular features you think would be useful for me to include?

Once we have all the functionality required in this PR that will go into the initial `MockClient` release, I'll get to work on:
 - Basic validation (thinking very lightweight checks on the inputs, such as type checking etc)
 - Documentation
 - Types (+ tests for types)
 - Any other tests I can think of

test/client-mock.js
```diff
@@ -0,0 +1,650 @@
...
+
+  const server = createServer((req, res) => {
+    res.setHeader('content-type', 'text/plain')
+    res.end('should not be called')
+    t.fail('should not be called')
+    t.end()
+  })
+  t.tearDown(server.close.bind(server))
+
+  server.listen(0, async () => {
```
mcollina, line 104: please use promisify on listen, let's not pass async callbacks to functions.

mcollina (commented):
Good work so far!

test/client-mock.js
jonnydgreen: sure thing, on it now!

jonnydgreen: fixed, my fault - should've sorted that straight away!

mcollina:
I would recommend also a double check against all nock functionality.

After adding the client there are a few more thing to add: MockPool and MockAgent.. however they can also land in a followup PR.

jonnydgreen:
No problem, I'll double check the nock functionality first thing tomorrow! :) I'll scope out MockPool and MockAgent to see how big these will be; as you suggested, if they're look like they'll be fairly large, I'll land them in follow-up PRs!

ryhinchey:
Thanks for putting this together @jonnydgreen!  I need this :)

jonnydgreen:
@ryhinchey no probs at all! I've got a fairly big update to this PR that increases existing functionality coming this evening - stay tuned! :)

@mcollina based on the amount of features incoming (from double checking nock functionality as you suggested), I'm thinking of adding support for MockAgent and MockPool in follow up PR's if that's okay with you?

mcollina:
I'm ok in landing this if you are confident in this API, and open a few other issues for the rest.

jonnydgreen:
> I'm ok in landing this if you are confident in this API, and open a few other issues for the rest.

Cool, thanks! I'm pretty confident about the API. I will double check this however and have a real think about the implications for `MockAgent` and `MockPool`!

In the meantime, I have added the following today:
 - Refactor mockDispatches to use `Array` instead of `Map` as this gives us a lot of flexibility for matchers :)
 - Add `RegExp` and `Function` matching support to `url`, `path`, `method`, `body`
 - Add `.times(num)` for requests occurring a set amount of times
 - Make sure `.persist()` always take precedence over `.times(num)`
 - Add a `cleanAllMocks()` function that cleans all the mocks (I have a few other things to add in relation to this for test clean functionality)
 - Add a `.replyWithError(err)` function for triggering errors

There are a few ideas I had from double checking the nock API, so I will pursue those tomorrow :)

jonnydgreen:
Hi! :) I've been reading the codebase on Agent and Pool and I think I have a fairly good grasp of them now and I think it wouldn't be too much of a stretch to include Agent and Pool (especially Pool since it implements the Client API!). I just wondered if you're able to clarify something for me if that's okay as I'm not 100% clear with what you had in mind for `MockAgent` and `MockPool`?

I was thinking a `MockPool` example could be something like this:

```js
const mockPool = new MockPool(baseUrl)
mockPool.intercept({
  path: '/foo',
  method: 'POST'
}).reply(200, { foo: 'bar' })
```

The main API would be very similar to `MockClient` as it's effectively calling `client.dispatch` under the hood. However, I was thinking we could extend this for `MockPool`, such as triggering errors such as `ClientDestroyedError` with `.clientDestroyed()` perhaps (although this would also be achieved by `.replyWithError(error)`)? Or simulating when the pool is full perhaps?

I wasn't sure about `MockAgent` however, maybe I've misunderstood it in terms of mocking! Because it calls either a Pool or a Client when `agent.get(origin)` is called, I wondered what other benefits mocking the agent would bring when we could create an interceptor for Client/Pool? Perhaps explicitly saying we are getting a client or pool perhaps? Thanks in advance!

lib/client-mock.js
```diff
@@ -0,0 +1,303 @@
+'use strict'
+
+const Client = require('./core/client')
+const { kUrl } = require('./core/symbols')
+
+const originalDispatch = Client.prototype.dispatch
+
+const mockDispatches = []
```
mcollina, line 8: Are we sure we want these to be global and not isntance-specific? I think it might be more flexible long term.

jonnydgreen: Yeah I agree! I was initially concerned about the concurrency impact of overriding Client.prototype.dispatch which I why I chose this pattern in the first place. I've done some experiments just now after your suggestion with jest and tap and confirmed that this should not be a problem as tests within the same file should be run in order and those in different files run in parallel (in different processes). I will refactor to make this instance specific on each `MockClient` :)
Screenshot 2021-03-11 at 22 54 19

jonnydgreen: Functionally, I think we're pretty much there now. I'm going to tidy up the work I've done so far, add your suggested refactors, then typings and documentation

jonnydgreen:
I've added the refactor among other changes to tidy up the implementation (including moving to a separate mock directory which I think adds a nice degree of separation) straight away to get your early perspective, let me know what you think! :) Couple of notes:
 - I added a mockClients global, which is a map of all the live MockClient instances (useful for global clean up/debug procedures)
 - The netConnect is also left as a global as I believe the settings for this should apply to all clients and not be instance specific

mcollina:
I would prefer not to have global state, but achieve global state by setting a `MockAgent` via https://github.com/nodejs/undici/blob/60c1ea0efb69007b3e6a705cdaf8a722ac84ed63/index.js#L26.

jonnydgreen:
Cool, thanks and understood, makes perfect sense and should nicely support `undici.request` etc! :) Example I was thinking of:

```js
const mockAgent = new MockAgent()
setGlobalAgent(mockAgent) // Could call this in the mock agent by default

// Could also set up the intercepts directly on the mock agent
mockAgent.origin('http://some-origin.com').intercept(...) // Same API as MockClient following intercept
...
```

Over the weekend, I was thinking about other usages such as `Client.prototype.request` which don't get dispatched from the agent. In order to achieve global state for these, I'll need to pass through an `agent` (or state) when instantiating `MockClient` (similar story for `MockPool`) like so:

```js
const agent = new MockAgent() // Only needs to be done once per test/test suite

// The agent would be optional so one wouldn't need to
// do this for a single origin where global state isn't needed
const mockClient = new MockClient('http://some-origin.com', { agent })

mockClient.intercept(...)
```

Does passing the mock agent (as the global state) through to the mock client this pollute the MockClient model in your opinion? My reason for doing this is that it should provide a clean way to achieve global state, give a lot of flexibility and make it clear to users exactly what's going on!

If you agree on the tweaked API, I'll get these suggestions pushed asap :)

mcollina:
I think MockAgent would just delegate everything to the relevant MockPool/MockClient, e.g.:

```js
mockAgent.get('http://some-origin.com').intercept({...})
```

jonnydgreen:
Gotya! So in much the same way as `Agent.get` works currently - if the origin doesn't exist, instantiate the `MockPool/Client` classes (depending on connection options) and delegate, otherwise delegate to existing instances already loaded

jonnydgreen:
Hi! Here's an update based on your suggestions :) It includes:
 - Add `MockAgent`
 - Add `MockPool`
 - Remove all mock globals and refactor logic accordingly

lib/mock/mock-agent.js
```diff
@@ -0,0 +1,185 @@
...
+        originalDispatch.call(this, opts, handler)
+      }
+    }
+    return pool
+  }
+
+  deactivate () {
+    this[kIsMockActive] = false
+    Client.prototype.dispatch = this[kOriginalClientDispatch]
+    Pool.prototype.dispatch = this[kOriginalPoolDispatch]
```
mcollina, line 82: Do not alter module-wide prototypes.

lib/mock/mock-agent.js
```diff
@@ -0,0 +1,185 @@
...
+
+  disableNetConnect () {
+    this[kNetConnect] = false
+  }
+
+  [kIsMockEnabled] () {
+    if (!this[kIsMockActive]) {
+      return false
+    }
+    return process.env.UNDICI_MOCK_OFF !== 'true'
```
mcollina, line 113: Do not use env variables, mocking is always enabled when passing through the mock.

lib/mock/mock-agent.js
```diff
@@ -0,0 +1,185 @@
...
+    if (!pool) {
+      pool = this[kCreatePool](origin)
+      this[kMockAgentSet](origin, pool)
+    }
+
+    const originalDispatch = this[kIsClient]() ? this[kOriginalClientDispatch] : this[kOriginalPoolDispatch]
+    const isMockEnabled = this[kIsMockEnabled].bind(this)
+    const getNetConnect = this[kGetNetConnect].bind(this)
+
+    pool.dispatch = function dispatch (opts, handler) {
```
mcollina, line 60: I don't think this is needed.

mcollina (changes requested):
Good work! I think we are almost there.

mcollina:
@ronag could you take a look?

lib/mock/mock-agent.js
jonnydgreen: Understood!

lib/mock/mock-agent.js
jonnydgreen: Cool, I'll take a look!

lib/mock/mock-agent.js
jonnydgreen: Are you referring to all occurrences of this or just this function? If all occurrences, I think this will affect our ability to intercept in the following way:

```js
// In test code
const mockAgent = new MockAgent({ connections: 1 })
const mockClient = mockAgent.get(baseUrl)
mockClient.intercept({
  path: '/foo',
  method: 'GET'
})

// In source code
const client = new Client(baseUrl)
const { statusCode, headers, trailers, body } = await client.request({
  path: '/foo',
  method: 'GET'
})
```

I can't think of an additional way to intercept `Client.prototype.request` (as it does not interact with an agent as far as I understand!) without explicitly overriding `client` with `mockClient` in the test? E.g. when explicitly overriding, it becomes:

```js
// In test code
const mockAgent = new MockAgent({ connections: 1 })
const mockClient = mockAgent.get(baseUrl)
mockClient.intercept({
  path: '/foo',
  method: 'GET'
})

// In source code
let client = new Client(baseUrl)

// Back in test code
client = mockClient // Through retrieving the client or intercepting it's instantiation

// Back in source code
const { statusCode, headers, trailers, body } = await client.request({
  path: '/foo',
  method: 'GET'
})
```

I don't know if you have any additional ideas/thoughts on overriding the dispatch in this situation?

mcollina: I don't think we should intercept things globally - that's exactly my reasoning from the beginning. We can achieve 

```js
import { request, setGlobalAgent, MockAgent } from 'undici'

// In the test
const mockAgent = new MockAgent({ connections: 1 })
setGlobalAgent(mockAgent)
const mockClient = mockAgent.get(baseUrl)
mockClient.intercept({
  path: '/foo',
  method: 'GET'
})

// in the actual code, this is mocked

const {
  statusCode,
  headers,
  trailers,
  body
} = await request('http://localhost:3000/foo')

  console.log('data', data)
}

console.log('trailers', trailers)
```

There is nothing shared globally minus the "global agent". If somebody wants something more granular, they can pass around the MockClient or use sinon/proxyquire to wrap.

jonnydgreen: My misunderstanding about intercepting globally, apologies - I assumed we still wanted some sort of support for `Client.prototype.request` in addition to anything passing through the agent! Makes perfect sense! I will change and remove this! :)

test/mock-agent.js
```diff
@@ -0,0 +1,1734 @@
...
+        method: 'GET'
+      })
+      t.strictEqual(statusCode, 200)
+
+      const response = await getResponse(body)
+      t.strictEqual(response, 'foo')
+    }
+  } catch (err) {
+    t.fail(err.message)
+  }
```
mcollina, line 1468: remove these try-catch blocks

test/mock-agent.js
```diff
@@ -0,0 +1,1734 @@
...
+    mockAgent.disableNetConnect()
+
+    await request(`${baseUrl}/foo`, {
+      method: 'GET'
+    })
+    t.fail('should not complete the request')
+  } catch (err) {
+    t.strictEqual(err.message, `Unable to find mock dispatch and real dispatches are disabled for http://localhost:${server.address().port}`)
+  }
+})
```
mcollina, line 1734: Can you please add a few tests that use `MockAgent` directly?

lib/mock/mock-agent.js
```diff
@@ -0,0 +1,111 @@
...
+  kIsMockActive,
+  kNetConnect,
+  kGetNetConnect
+} = require('./mock-symbols')
+const MockClient = require('./mock-client')
+const MockPool = require('./mock-pool')
+const { matchValue } = require('./mock-utils')
+const { InvalidArgumentError } = require('../core/errors')
+
+class MockAgent extends Agent {
```
mcollina, line 21: Looking at https://github.com/nodejs/undici/pull/600 it might be better that the MockAgent encapsulates an Agent rather than inherit from it.

mcollina (commented):
Good work! I really like the implementation.

test/mock-agent.js
jonnydgreen: Will do

test/mock-agent.js
jonnydgreen: Sure, I'll make sure this is also true for MockPool and MockClient where appropriate!

lib/mock/mock-agent.js
jonnydgreen: No problem, will get right on it!

ronag (commented):
I don't quite understand what this is for. Haven't worked so much with mocking in Nodejs. I defer to the other contributors.

ronag (commented):
Does this need docs?

jonnydgreen:
> I don't quite understand what this is for. Haven't worked so much with mocking in Nodejs. I defer to the other contributors.

This can be used to test functionality that performs HTTP requests. With the mocking enabled, we can intercept certain requests with user defined responses and return this data instead of making the actual request. E.g. we can set request `GET http://example.com/foo` to actually respond with statusCode `200` and data `{"some":"data"}`. In this way, we can test code that integrates with external services in isolation, as an example.

> Does this need docs?

Yep, now that the API is stable, I'll get these sorted next! :)

lib/mock/mock-agent.js
```diff
@@ -0,0 +1,112 @@
...
+const { matchValue } = require('./mock-utils')
+const { InvalidArgumentError } = require('../core/errors')
+
+class MockAgent {
+  constructor (opts) {
+    this[kNetConnect] = true
+    this[kIsMockActive] = true
+
+    // Instantiate Agent and encapsulate
+    const agent = new Agent(opts)
```
mcollina, line 26: This should be an option instead.

jonnydgreen: Sure thing, I'll include this in the next update with typings and docs! :) Just to clarify one thing, do you mean a mandatory option (e.g. `opts.agent` and fail if it's not present) or an option with a default (e.g. `opts.agent || new Agent(opts)`)?

mcollina: I'm ok with an option with a default, that'd be a nicer user experience.

jonnydgreen: Okay no probs! :)

jonnydgreen:
Quick update:
 - Have tightened a few things up
 - Added option for specifying agent in MockAgent
 - Add docs for MockAgent, MockPool, and MockClient

Adding the types and tests for them now :)

mcollina (approved):
lgtm

mcollina:
Awesome, things are really solid now!

Nytelife26:
As not everyone will use the mocking functionality and thus it adds package bloat for a large portion of the userbase, would it not be better to move mocking into a separate package, for instance, something like `undici-mocks`? That way it is available to those who need it and those who do not will not suffer any package bloat.

mcollina:
It is highly important this functionality is maintained together with the rest of the code to prevent divergence. As for package bloat, I would argue call anything bloat. 

If you are concerned about load time, it's definitely possible to load this only when needed.

Nytelife26:
> It is highly important this functionality is maintained together with the rest of the code to prevent divergence. 

Indeed it is. It would, however, be possible to maintain them side by side, with `undici-mocks` being simply an extension of undici.

> I would argue call anything bloat

That makes no sense.

> If you are concerned about load time, it's definitely possible to load this only when needed.

I am concerned about bundle size. A lot of people in the ecosystem do not use tree shaking, and as such many end users will inherit large amounts of package bloat from libraries. As such, is it important that libraries stick to their scope.

tbnritzdoge:
I echo what @Nytelife26 is saying in that it is very important to keep bundle sizes down in a world where people do not tree shake to prevent people from importing an obscene amount of things. It also should be noted that this is primarily a library to make HTTP(s) requests and mocking is not really in its scope so whilst undici-mocks is a good idea it should probably be kept out of this repository or this one should be turned into a monorepo with multiple packages (undici, undici-mocks).

mcollina:
Thanks for your opinions @Nytelife26 and @tbnritzdoge! I prefer we land this when it is ready and then discuss if we want to move undici to a mono-repo in another issue.

jonnydgreen:
> Awesome, things are really solid now!

Thanks and thank you for your advice and guidance on this! :) I have done a final pass on the repo and I think everything is there; so if everyone is happy, I will mark this ready for review

mcollina:
Good for me

ronag (approved):

docs/api/MockAgent.md
```diff
@@ -0,0 +1,388 @@
...
+#### Example - Allow all non-matching urls to be dispatched in a real HTTP request
+
+```js
+'use strict'
+const { MockAgent } = require('undici')
+
+const mockAgent = new MockAgent()
+
+mockAgent.enableNetConnect()
+
```
dnlup, line 316: Shouldn't we have `setGloblAgent` here and in the following examples?

dnlup (approved):
Great work!

docs/api/MockAgent.md
jonnydgreen: Good spot - I'll fix those now!

ronag:
Please wait with merging this. Might be conflict with https://github.com/nodejs/undici/pull/603#discussion_r598889560.

ronag:
Another one to wait for https://github.com/nodejs/undici/pull/620

@jonnydgreen sorry to be blocking this PR 😞

jonnydgreen:
@ronag No problem at all and makes perfect sense to hold this until these are merged - thanks for letting me know! I'll keep an eye on them so I am familiar with the changes required!

jonnydgreen:
Awesome, I'll refactor for the new APIs asap - thanks for letting me know @ronag !

jonnydgreen:
Hi! I have refactored the mocks to implement the new APIs - for the exposed mock API it was hardly any change! :) As well as refactoring the mocks to support the new API, I have:
  - Updated the docs
  - Added basic validation to all the public mock input fields
  - `MockAgent` now extends `Dispatcher`
  - Added extra examples for using either of `MockClient`, `MockPool`, `MockAgent` as a global (and local) dispatcher

Note, I've kept `MockAgent.get` in addition to adding the `MockAgent.dispatch` method. I think it makes sense in this case from a mocking point of view because we use `MockAgent.get` to retrieve/instantiate the MockClient, MockPool instances.

Unfortunately, I wasn't able to update the typings as these still use the old API (`setGlobalAgent` etc). I'm happy to help out with fixing this although I guess you'd want this in a separate PR to set up the Dispatcher type etc?

Let me know what you think!

lib/agent.js
```diff
@@ -16,12 +17,9 @@ const kDestroyed = Symbol('destroyed')
 const kClosed = Symbol('closed')
 const kOnConnect = Symbol('onConnect')
 const kOnDisconnect = Symbol('onDisconnect')
-const kClients = Symbol('clients')
```
jonnydgreen, line 19: Also, just drawing your attention to this as this was the only bit of actual undici code that I changed - this was so I could access the symbols I need to hook into the encapsulated agent `kClients` `Map`.

lib/mock/mock-agent.js
```diff
@@ -0,0 +1,141 @@
...
+
+    // Instantiate Agent and encapsulate
+    if ((opts && opts.agent && typeof opts.agent.dispatch !== 'function')) {
+      throw new InvalidArgumentError('Argument opts.agent must implement Agent')
+    }
+    const agent = opts && opts.agent ? opts.agent : new Agent(opts)
+    this[kAgent] = agent
+
+    this[kClients] = agent[kClients]
+    this[kOptions] = agent[kOptions]
```
ronag, line 49: This is a little dangerous. But I'm not familiar enough with this to suggest anything better. I could easily change the internals in the future and accidentally break mocking support. Hope we have test that can catch that.

jonnydgreen: Understood and yep, we have a test to catch this. I'll try and come up with sometime better that doesn't reference `kClients`, `kOptions` or `kFactory`.

mcollina: @ronag this is the exact reason why we need mocking support inside Undici.

mcollina (approved):
lgtm

lib/mock/mock-agent.js
jonnydgreen: FYI both, I have moved `kOptions` and `kFactory` back into the Agent as these were not core to the mocking functionality, so only `kClients` is referenced (and backed up with tests so if this is changed, the mocks will soon know and should be a simple fix to resolve)

jonnydgreen:
> Unfortunately, I wasn't able to update the typings as these still use the old API (setGlobalAgent etc). I'm happy to help out with fixing this although I guess you'd want this in a separate PR to set up the Dispatcher type etc?

@mcollina what are your thoughts on this in relation to this PR being good to go? Is a separate PR acceptable to sort this out, or would you prefer to include here?

mcollina:
I recommend to update the typings on a follow up PR. If you could do that for all the other changes, that would be great.

jonnydgreen:
Sure thing, no problem - I'll start that once this is merged :)

lib/mock/mock-agent.js
```diff
@@ -0,0 +1,139 @@
...
+  kOptions,
+  kFactory
+} = require('./mock-symbols')
+const MockClient = require('./mock-client')
+const MockPool = require('./mock-pool')
+const { matchValue, buildMockOptions } = require('./mock-utils')
+const { InvalidArgumentError } = require('../core/errors')
+const Dispatcher = require('../dispatcher')
+
+const WeakRef = global.WeakRef || class CompatWeakRef {
```
dnlup, line 22: Just a nit, maybe we can move this into utils in the future? Since we have the same code in `agent`. I would add a to-do comment. Not blocking over this, of course.

dnlup (approved):

lib/mock/mock-agent.js
jonnydgreen: No problem - added! Note, the original code wasn't exactly the same as in Agent, so I've made these the same and adjusted the mocks to mock the `connected` getter

lib/mock/mock-agent.js
jonnydgreen: @ronag is this solution okay with you? I've added a specific unit test to immediately catch if `kClients` changes. The main problem I found with having a separate `kClients` symbol in mock agent was that we would lose agent encapsulation and need to add additional code for close and dispatch (in other words, it's doable but with some drawbacks). Let me know what you think!

mcollina:
@ronag you ok for this to land?

ronag (approved):

PR #588, opened 2021-03-05
Reuse tls sessions accross clients of a connection pool

nherment:


lib/core/client.js
```diff
@@ -343,6 +343,15 @@ class Client extends EventEmitter {
...
   }
+
+  set tlsSession (session) {
+    this[kTLSSession] = session
+    this.emit(kTLSSession, session)
+  }
+
+  get tlsSession () {
+    return this[kTLSSession]
+  }
```
ronag, line 354: I'm not too happy about making these part of the public API.

ronag: Can't pool just pass in sessions or pass a connect method through the constructor?

nherment: Thanks for the comment @ronag and agree, this should not be part of the public API. This PR is a draft and I'm reviewing and making final tweaks, including this one.

lib/pool.js
```diff
@@ -159,11 +167,26 @@ class Pool extends EventEmitter {
 
       if (!client) {
         if (!this[kConnections] || this[kClients].length < this[kConnections]) {
-          client = new Client(this[kUrl], this[kOptions])
+          const options = { ...this[kOptions] }
+          if (!options.tls || (options.tls.reuseSessions !== false)) {
```
nherment, line 171: `options.tls.reuseSessions` is a flag to disable the session reuse in a conn. pool. I thought of it as a safe mechanism should it ever be needed. The sessions are reused within a pool by default.

nherment:
Also please let me know if some documentation should be added to the readme

lib/pool.js
```diff
@@ -44,6 +46,7 @@ class Pool extends EventEmitter {
     this[kNeedDrain] = false
     this[kPending] = 0
     this[kConnected] = 0
+    this[kTLSSessionCache] = new TLSSessionCache()
```
ronag, line 49: Why not just a `Map` here directly?

lib/pool.js
```diff
@@ -159,11 +167,26 @@ class Pool extends EventEmitter {
...
         if (!this[kConnections] || this[kClients].length < this[kConnections]) {
-          client = new Client(this[kUrl], this[kOptions])
+          const options = { ...this[kOptions] }
+          if (!options.tls || (options.tls.reuseSessions !== false)) {
+            const session = this[kTLSSessionCache].getSession()
+            if (!options.tls) {
+              options.tls = {
+                session
+              }
+            } else {
```
ronag, line 177: make sure there isn't already a session passed

ronag (commented):
I wish we could implement this without coupling Client and Pool with hidden symbols. But i guess we can improve that in the future.

ronag:
Any chance we could implement this by passing a `connect` method in the client options?

lib/pool.js
nherment: Thats a good question. It was quite practical to be able to tweak the behavior of the cache (map vs simpler or more complicated) and compare the different implementations.

This class adds very little and I don't feel strongly in keeping it or removing it. Happy to remove it if you want.

nherment:
> Any chance we could implement this by passing a `connect` method in the client options?

What would it help with? It'll definitely be more powerful but also come with some drawbacks if I understand correctly (e.g. re implementing the function in the pool).

lib/pool.js
ronag: Please remove. I'm very keen on simplicity.

ronag:
> What would it help with? It'll definitely be more powerful but also come with some drawbacks if I understand correctly (e.g. re implementing the function in the pool).

Removes ugly coupling between classes.

nherment:
@ronag I mainly bring 4 changes:

1. The Map/TLSSessionCache is gone in favor of a single session stored at the pool level. Thanks to @mcollina for this.
2. The Client now publicly exposes the 'session' event  instead of a hidden one.
3. The logic to set the TLS session on top of the `tls` options in the pool is updated and hopefully more readable
4. There is a flag to enable/disable the perf test of TLS session reuse


If I understand your suggestion of passing in a connect function, I feel it would be over-complicated as the pool would need to re-implement some logic that the client owns. 

I saw 2 options to address your following comment:

> I wish we could implement this without coupling Client and Pool with hidden symbols.

- Either implement some kind of hooks in the `Client` that would allow the Pool to override the default behaviour
- Or publicly expose the `Client` `session` event instead of using a hidden symbol.

ronag:
I think I'm happy with exposing the session event.

ronag (approved):
LGTM, good job!

nherment:
@ronag or @mcollina could you please merge this for me? I don't have the access rights to do it.

ronag:
Waiting for review from @mcollina

test/tls-session-reuse.js
```diff
@@ -130,3 +130,103 @@ test('TLS should reuse sessions', { skip: nodeMajor < 11 }, t => {
 
   t.end()
 })
+
+test('A pool should be able to reuse TLS sessions between clients', { skip: nodeMajor < 11 }, t => {
```
mcollina, line 134: What happens if this is used in node v10? Is it crashing? We should document and add a a test for it.

mcollina (commented):
Code is good to go!

mcollina:
Have you done any performance testing by chance? Did you see any improvement?

lib/pool.js
```diff
@@ -159,11 +166,18 @@ class Pool extends EventEmitter {
 
       if (!client) {
         if (!this[kConnections] || this[kClients].length < this[kConnections]) {
-          client = new Client(this[kUrl], this[kOptions])
+          const options = { ...this[kOptions] }
```
Ethan-Arrowood, line 169: is a call to `Object.assign({}, this[kOptions])` potentially faster?

lib/pool.js
```diff
@@ -249,6 +263,26 @@ class Pool extends EventEmitter {
...
 
+function addCachedSessionToTLSOptions (pool, options) {
+  let tls = options.tls
+
+  if (!tls) {
+    tls = {}
+  }
+
+  // User has chosen to opt out of TLS session reuse
+  if (tls.reuseSessions === false) {
```
Ethan-Arrowood, line 274:
```suggestion
  if (!tls.reuseSessions) {
```

lib/pool.js
```diff
@@ -159,11 +166,18 @@ class Pool extends EventEmitter {
...
-          client = new Client(this[kUrl], this[kOptions])
+          const options = { ...this[kOptions] }
+          options.tls = addCachedSessionToTLSOptions(this, options)
+
+          client = new Client(this[kUrl], options)
             .on('drain', this[kOnDrain])
             .on('connect', this[kOnConnect])
             .on('disconnect', this[kOnDisconnect])
 
+          if (!options.tls || (options.tls.reuseSessions !== false && !options.tls.session)) {
```
Ethan-Arrowood, line 177:
```suggestion
          if (!options.tls || (options.tls.reuseSessions && !options.tls.session)) {
```

test/tls-session-reuse.js
Ethan-Arrowood: i don't immediately see why this wouldn't run in node 10 but i could easily be missing something obvious

Ethan-Arrowood (changes requested):

lib/pool.js
nherment: Thanks for the suggestion @Ethan-Arrowood. As mentioned in the PR's description this strict equality check is on purpose.
Changing to your code will change the behavior of the pool to not reuse sessions by default.

lib/pool.js
nherment: The same comment as above regarding the default behavior applies here too.

test/tls-session-reuse.js
nherment: @ronag @delvedor @mcollina do you remember why that was done for the [Client's session reuse](https://github.com/nodejs/undici/pull/358)?
This code covers the same path so I decided to reuse the node version skip.

nherment:
@mcollina regarding the perf tests I saw an improvement. You can test it locally. Unfortunately the perf tests dont always pass in the test VMs.

test/tls-session-reuse.js
delvedor: The [`session` event](https://nodejs.org/api/tls.html#tls_event_session) has been added in Node v11.

mcollina (approved):
lgtm

ronag (changes requested):
I'd like to understand the node 10 thing

test/tls-session-reuse.js
nherment: Thank you!

ronag (approved):

delvedor (approved):
LGTM

lib/pool.js
Ethan-Arrowood: Oh I see now

Ethan-Arrowood (approved):
Great work!

PR #596, opened 2021-03-11
add old agent and errors docs

Ethan-Arrowood:
Hey folks, 

So i'm going to be away-from-keyboard this weekend, but I did not want to leave our docs without the Agent or Errors api. I've gone ahead and copy-pasted the old docs we had for these sections and added them to the docs folder. I also touched them up a little bit so links work and it renders nicely. I'm okay for "launching" our docs site with this content if that happens this weekend; otherwise, I'll follow up next week with documentation pages similar to `Client` and `Pool`.

🚀

### DIFF
--- a/docs/api/Agent.md
+++ b/docs/api/Agent.md
@@ -1,0 +1,69 @@
+# Agent
+
+## `new undici.Agent(opts)`
+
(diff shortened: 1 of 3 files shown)


ronag (approved):

mcollina (approved):
lgtm

Issue #553, opened 2021-02-17
Questions


jsumners:
I am evaluating using this library since `got` is planned to be ESM only by end of 2021 and I am encountering a few questions:

1. Under what conditions does `client.request` return an error? Is it only for operational errors? Or will it return an error if the remote host returns a `404` response? If the later, is it any `4xx` or `5xx`?
2. Does `client.request` follow redirects?
3. Does `client.request` automatically retry under any circumstances? For example, `got` by default will retry a `500` three times before giving up and returning an error.
4. If `undici.Client` accepts an instance of `URL`, why would it reject paths and query parameters (#552)? Since it accepts a `URL` instance, I assumed it would create a client specific to that URL and `client.request` wouldn't require any further configuration to make a request.

I'm sure I have other questions if I dig further, but these are the primary questions I have after a few minutes of trial.

ronag:
undici is a low level client not comparable to got. got could be built on top of undici.

1. No
2. No
3. No
4. Because the client is just a connection to a server, query parameters are for requests.

ShogunPanda:
What do you think about a PR for the n.2? 

I'm thinking like adding a `maxRedirects: number` option in `undici.Client`, overridable in `client.request`.

If not provided, then undici will behave as today, for backwards compatibility.

If provided, it is the number of redirects to follow, with `0` meaning `infinite` and undici will follow the redirects according to RFCs.

If any redirects is following, a `redirects` array will be provided in callback data as well.

If this spec looks fine to you I'll try to work on a PR as soon as I can.

ronag:
I'm not sure I think adding redirect support is a good idea.

ShogunPanda:
Can you elaborate? So we can find a suitable solution. I had the same issue as @jsumners so I think this might be a common concern.

ronag:
undici is supposed to be a low level client which is correct and simple to maintain. I'm concerned that adding redirect support will increase complexity. It also could easily live in a separate npm package as a decorator.

Looking at https://www.npmjs.com/package/follow-redirects I'm quite concerned about complexity.

ronag:
If you think you can implement it in a simple and elegant way that is spec compliant feel free to open a PR but I can't promise we merge it. If not then maybe we can create a separate npm package based on that PR?

ronag:
@mcollina wdyt?

ShogunPanda:
A separate package might work as well. Let me explore PR first and if doesn't feel appealing l'll go for the separate package one.

[...]

PR #597, opened 2021-03-12
feat: Added RedirectAgent.

ShogunPanda:
This PR adds the RedirectAgent as suggested by @mcollina in #553 to add redirections following to undici.

At the moment the PR works only on the top-level `request` method, but I'd like to have your opinion before extending it to `stream` and `pipeline`.

@ronag As requested, the modifications are completely isolated to the agent. I used monkey patching and thus the rest of undici is unaffected.

I have a final question: at the moment, when redirecting, I completely ignore the response body in case 3xx. In theory the server should not send a body but we know how life is :)
My question is: should I call something like `body.end` or `body.close` or can I safely ignore it?


ronag:
If you override dispatch instead of request then it will automatically work for all of the other methods.

ronag:
Also I think redirect can be quite a bit more complicated https://github.com/follow-redirects/follow-redirects/blob/main/index.js#L308

ronag:
@mcollina: I think the "correct" way to implement this is to extend from `undici.Client` and override `dispatch`, then and somehow pass a factory or prototype to `undici.Pool` which tells it to create that kind of client. Alternatively some kind of hook. wdyt? 

I don't think reimplementing all

ronag:
Actually I think it's sufficient to just create a `RedirectPool` and override `dispatch`:

```js
class RedirectPool extends Pool {
  dispatch (opts, handler) {
    super.dispatch(opts, {
      onHeaders (headers, ...) {
        if (shouldRedirect) {
          this.dispatch(...)
        },
        ...
      }
    })
  }
}
```

ronag:
Even fancier would be just to create a redirect mixin that just override dispatch of any object that implements the client api.

ShogunPanda:
Ok, I'll go for RedirectPool (with mixing) and RedirectAgent.
It will be much more clean!

From `onHeaders` if I return false it will skip the body and trailers part of redirected request, isn't it? Will undici take care of underlying socket or do I have to manage anything?

ronag:
> From `onHeaders` if I return false it will skip the body and trailers part of redirected request, isn't it? Will undici take care of underlying socket or do I have to manage anything?

false will just pause the socket. You have to manage everything unfortunately .

mcollina:
I kind of agree of using a RedirectPool being a better internal implementation rather than monkey patching.

The RedirectPool would need to have a reference to the Agent as Redirects might cross a domains.

We would need to have some logic to avoid never-ending redirect chains (or loops), either in the format of a timeout or as a limit in the number of hops.

ShogunPanda:
Ok, now I remember why I also for Agent: cross origin - Pools or Clients are single origin. 
From my understanding of the code, pool or client don't use agent. The top level method do via `dispatchRequest` internal top level method in `agent.js`. Am I wrong?

@mcollina I already added a `maxRedirects` options with defaults to 10.

@ronag I've checked `follow-redirects` link. Their implementation is weird - On a 301 or 302 you're not obligated to switch to POST to GET. The RFC 7231 says "MAY". So I haven't changed the method in that case.
And I don't even know why they drop the authorization header honestly.
But I do agree that I need to drop `Host` and `Content-*` headers when switching to GET method.

mcollina:
> Ok, now I remember why I also for Agent: cross origin - Pools or Clients are single origin. 
> 
> From my understanding of the code, pool or client don't use agent. The top level method do via `dispatchRequest` internal top level method in `agent.js`. Am I wrong?

You are correct. However monkey patching Pool during get() is not correct as it scrambles a lot of internal implementation of V8. What we are recommending is to change Agent constructor so it accept the Pool class to use. Then, in RedirectAgent, you can set that option to point to a RedirectPool, which implement the redirect logic. Somehow, you'd need to make RedirectPool aware of the agent itselt to support cross origin.

ShogunPanda:
Ok, the first part is trivial.
Let me think a little on how to implement RedirectPool using onHeaders then.

mcollina:
Why did you delete this @ShogunPanda ?

ShogunPanda:
@mcollina I messed when renaming main branch from `master` to `main` in order to be consistent with upstream.

I'm sending the new PR right now.

lib/agent.js
```diff
@@ -41,10 +46,11 @@ class Agent extends EventEmitter {
     let pool = self[kAgentCache].get(origin)
 
     if (!pool) {
-      pool = self[kAgentOpts] && self[kAgentOpts].connections === 1
-        ? new Client(origin, self[kAgentOpts])
-        : new Pool(origin, self[kAgentOpts])
+      const Klass = self[kAgentOpts] && self[kAgentOpts].connections === 1
```
ronag, line 49: Klass?

ShogunPanda: Ehm? Can you please clarify?

PR #599, opened 2021-03-15
Remove the esm wrapper, rely on commonjs autodetection

### LINKED ISSUE #598: Missing Agent in esm exports
Should we add the Agent to the `esm` wrapper or am I missing something?

Ref: https://github.com/nodejs/undici/blob/main/wrapper.mjs

mcollina:
Fixes #598


Ethan-Arrowood (approved):
lgtm

index.js
```diff
@@ -15,14 +15,14 @@ function undici (url, opts) {
...
-undici.Client = Client
-undici.errors = errors
+module.exports = undici
 
-undici.Agent = Agent
-undici.request = request
-undici.stream = stream
-undici.pipeline = pipeline
-undici.setGlobalAgent = setGlobalAgent
+module.exports.Pool = Pool
```
dnlup, line 20: This fails on Node `12.18.0`, but works on `12.21` (and later), and I could figure out a  way to fix it.

package.json
```diff
@@ -4,7 +4,6 @@
   "description": "An HTTP/1.1 client, written from scratch for Node.js",
   "main": "index.js",
   "types": "index.d.ts",
```
dnlup, line 6: Not sure if it's needed, but if we add a `"exports": "./index.js"` we can use self-resolution too.

dnlup (approved):

index.js
dnlup: From the docs [here](https://nodejs.org/dist./v12.18.0/docs/api/esm.html#esm_import_statements) seems it is not supported on that version, maybe we should skip it?

ronag (approved):

index.js
dnlup: It seems to be fixed in `12.20`:

https://nodejs.org/dist./v12.20.0/docs/api/esm.html#esm_import_statements

package.json
mcollina: I recommend against using `exports` in any way.

index.js
mcollina: Yes, this was added in v12.20

mcollina:
rebased. Lmk if you are still good to land

dnlup (approved):
LGTM
Should we squash the commits?

mcollina:
I'll squash on landing.

PR #600, opened 2021-03-17
feat: Added RedirectPool.

ShogunPanda:
This PR adds the RedirectPool as suggested by @mcollina and @ronag  in #597 to add a pool subclass which is able to follow redirected.

This PR supersedes #597 by avoiding monkey patching of the agent and clients.

At the moment has only be tested on the top-level request method, but I'd like to have your opinion before extending it to stream and pipeline.

@ronag Can you check the way I ignored the redirected request body? I realize that since HTTP 1.1 is on a single connection, I have to process the body anyway even tho I ignore it in order not to block pipelining. Thus, I just emptied the handler events.

I have two questions:

1. In https://github.com/nodejs/undici/blob/2366fa34e30ac5a8c33a3d02259f2ae0f7678c85/lib/agent.js#L43 we instantiate a Client if connections are explicitily set to `1`. Why is that? Do we still need it?
2. If the previous question answer is `yes`, does it mean I also need to create the `RedirectClient` as well?

### DIFF
--- a/docs/api/Agent.md
+++ b/docs/api/Agent.md
@@ -19,7 +19,7 @@ Extends: [`PoolOptions`](docs/api/Pool.md#parameter-pooloptions)
 ```js
 'use strict'
 const { createServer } = require('http')
-const { Agent, request, RedirectPool } = require('.')
+const { Agent, request, RedirectPool } = require('undici')
 
 const server = createServer((request, response) => {
   response.setHeader('Connection', `close`)
--- a/docs/api/Agent.md
+++ b/docs/api/Agent.md
@@ -12,7 +12,51 @@ Returns a new Agent instance for use with pool based requests or the following t
 
 Extends: [`PoolOptions`](docs/api/Pool.md#parameter-pooloptions)
 
-* **factory** `(url: string, options?: Client.Options): Pool` (optional) - A factory which returns the pool to use for the request.
+* factory `(url: string, options?: Client.Options): Pool` (optional) - A factory which returns the pool to use for the request.
+
+#### Example - Use `factory` to use a RedirectPool which follow redirects
+
+```js
+'use strict'
+const { createServer } = require('http')
+const { Agent, request, RedirectPool } = require('.')
+
+const server = createServer((request, response) => {
+  response.setHeader('Connection', `close`)
+
+  if (request.url === '/') {
+    response.statusCode = 301
+    response.setHeader('Location', `http://localhost:${server.address().port}/end`)
+    response.end('')
+    return
+  }
+
+  response.end('undici')
+})
+
+server.listen(() => {
+  request(
+    `http://localhost:${server.address().port}`,
+    {
+      agent: new Agent({
+        /*
+          Optionally, a redirectPoolFactory is also exported by undici:
+
+          factory: redirectPoolFactory
+        */
+        factory(origin, opts) {
+          return new RedirectPool(origin, opts)
+        }
+      })
+    },
+    (error, {body}) => {
+      body.setEncoding('utf8')
+      body.on('data', console.log) // This will output 'undici'
+      body.on('end', server.close.bind(server))
+    }
+  )
+})
+```
 
 ## `agent.get(origin): Pool`
 
--- a/docs/api/Agent.md
+++ b/docs/api/Agent.md
@@ -2,11 +2,18 @@
 
 ## `new undici.Agent(opts)`
 
-* opts `undici.Pool.options` - options passed through to Pool constructor
+* opts `AgentOptions` - options passed through to Pool constructor
 
 Returns: `Agent`
 
 Returns a new Agent instance for use with pool based requests or the following top-level methods `request`, `pipeline`, and `stream`.
+
+### Parameter: `AgentOptions`
+
+Extends: [`PoolOptions`](docs/api/Pool.md#parameter-pooloptions)
+
+* **clientClass** `Class extends Client` (optional) - The class to instantiate clients.
+* **poolClass** `Class extends Pool` (optional) - The class to use to instantiate pools.
 
 ## `agent.get(origin): Pool`
 
--- a/docs/api/Agent.md
+++ b/docs/api/Agent.md
@@ -12,8 +12,7 @@ Returns a new Agent instance for use with pool based requests or the following t
 
 Extends: [`PoolOptions`](docs/api/Pool.md#parameter-pooloptions)
 
-* **clientClass** `Class extends Client` (optional) - The class to instantiate clients.
-* **poolClass** `Class extends Pool` (optional) - The class to use to instantiate pools.
+* **factory** `(url: string, options?: Client.Options): Pool` (optional) - A factory which returns the pool to use for the request.
 
 ## `agent.get(origin): Pool`
 
--- a/docs/api/Pool.md
+++ b/docs/api/Pool.md
@@ -112,12 +112,22 @@ Extends: `Pool`
 
 A pool which will automatically follow redirections.
 
-When used, the option `maxRedirections` can be additionally provided to top level `request`, `stream` and `pipeline`. The option must be a boolean or a number:
+When used, the option `maxRedirections` can be additionally provided to top level `request`, `stream` and `pipeline`. The option must a positive number:
 
-* If omitted or set to `true`, up to 10 redirections are followed.
-
-* If set to `false`, redirections are not followed.
+* If omitted, up to 10 redirections are followed.
 
 * If set to a positive number, it specifies the maximum number of redirections to follow.
 
 The data returned by the top-level method (via callback or promise) will contain the additional property `redirections` that lists all the followed redirections, in order.
+
+# Function: redirectPoolFactory
+
+A factory function which returns a `RedirectPool` for a specific URL.
+
+Arguments:
+
+* **url** `URL | string` - It should only include the **protocol, hostname, and port**.
+* **options** `PoolOptions` (optional)
+
+Returns: `RedirectPool` - The pool to be used by the agent.
+
--- a/docs/api/Pool.md
+++ b/docs/api/Pool.md
@@ -105,29 +105,3 @@ Implements [Client Event: `'disconnect'`](docs/api/Client.md#event-connect)
 ### Event: `'drain'`
 
 Implements [Client Event: `'drain'`](docs/api/Client.md#event-connect)
-
-# Class: RedirectPool
-
-Extends: `Pool`
-
-A pool which will automatically follow redirections.
-
-When used, the option `maxRedirections` can be additionally provided to top level `request`, `stream` and `pipeline`. The option must a positive number:
-
-* If omitted, up to 10 redirections are followed.
-
-* If set to a positive number, it specifies the maximum number of redirections to follow.
-
-The data returned by the top-level method (via callback or promise) will contain the additional property `redirections` that lists all the followed redirections, in order.
-
-# Function: redirectPoolFactory
-
-A factory function which returns a `RedirectPool` for a specific URL.
-
-Arguments:
-
-* **url** `URL | string` - It should only include the **protocol, hostname, and port**.
-* **options** `PoolOptions` (optional)
-
-Returns: `RedirectPool` - The pool to be used by the agent.
-
--- a/docs/api/Pool.md
+++ b/docs/api/Pool.md
@@ -105,3 +105,19 @@ Implements [Client Event: `'disconnect'`](docs/api/Client.md#event-connect)
 ### Event: `'drain'`
 
 Implements [Client Event: `'drain'`](docs/api/Client.md#event-connect)
+
+# Class: RedirectPool
+
+Extends: `Pool`
+
+A pool which will automatically follow redirections.
+
+When used, the option `maxRedirections` can be additionally provided to top level `request`, `stream` and `pipeline`. The option must be a boolean or a number:
+
+* If omitted or set to `true`, up to 10 redirections are followed.
+
+* If set to `false`, redirections are not followed.
+
+* If set to a positive number, it specifies the maximum number of redirections to follow.
+
+The data returned by the top-level method (via callback or promise) will contain the additional property `redirections` that lists all the followed redirections, in order.
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -6,6 +6,8 @@ const util = require('./core/util')
 const { kAgentOpts, kAgentCache } = require('./core/symbols')
 const EventEmitter = require('events')
 
+const kClientClass = Symbol('clientClass')
+const kPoolClass = Symbol('poolClass')
 const kOnConnect = Symbol('onConnect')
 const kOnDisconnect = Symbol('onDisconnect')
 
@@ -17,6 +19,9 @@ class Agent extends EventEmitter {
     this[kAgentCache] = new Map()
 
     const agent = this
+
+    this[kClientClass] = (opts && opts.clientClass) || Client
+    this[kPoolClass] = (opts && opts.poolClass) || Pool
 
     this[kOnConnect] = function onConnect (client) {
       agent.emit('connect', client)
@@ -41,10 +46,11 @@ class Agent extends EventEmitter {
     let pool = self[kAgentCache].get(origin)
 
     if (!pool) {
-      pool = self[kAgentOpts] && self[kAgentOpts].connections === 1
-        ? new Client(origin, self[kAgentOpts])
-        : new Pool(origin, self[kAgentOpts])
+      const Klass = self[kAgentOpts] && self[kAgentOpts].connections === 1
+        ? self[kClientClass]
+        : self[kPoolClass]
 
+      pool = new Klass(origin, self[kAgentOpts])
       pool
         .on('connect', this[kOnConnect])
         .on('disconnect', this[kOnDisconnect])
@@ -97,7 +103,7 @@ function dispatchFromAgent (requestType) {
       throw new InvalidReturnValueError(`Client returned from Agent.get() does not implement method ${requestType}`)
     }
 
-    return client[requestType]({ ...opts, method, path }, ...additionalArgs)
+    return client[requestType]({ ...opts, agent, method, path }, ...additionalArgs)
   }
 }
 
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -109,7 +109,7 @@ function dispatchFromAgent (requestType) {
       throw new InvalidReturnValueError(`Client returned from Agent.get() does not implement method ${requestType}`)
     }
 
-    return client[requestType]({ ...opts, agent, requestType, method, path }, ...additionalArgs)
+    return client[requestType]({ ...opts, agent, method, path }, ...additionalArgs)
   }
 }
 
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -6,8 +6,7 @@ const util = require('./core/util')
 const { kAgentOpts, kAgentCache } = require('./core/symbols')
 const EventEmitter = require('events')
 
-const kClientClass = Symbol('clientClass')
-const kPoolClass = Symbol('poolClass')
+const kFactory = Symbol('factory')
 const kOnConnect = Symbol('onConnect')
 const kOnDisconnect = Symbol('onDisconnect')
 
@@ -20,8 +19,11 @@ class Agent extends EventEmitter {
 
     const agent = this
 
-    this[kClientClass] = (opts && opts.clientClass) || Client
-    this[kPoolClass] = (opts && opts.poolClass) || Pool
+    const factoryType = typeof opts.factory
+    if (factoryType !== 'undefined' && factoryType !== 'function') {
+      throw new InvalidArgumentError('factory must be a function.')
+    }
+    this[kFactory] = (opts && opts.factory) || defaultFactory
 
     this[kOnConnect] = function onConnect (client) {
       agent.emit('connect', client)
@@ -46,11 +48,7 @@ class Agent extends EventEmitter {
     let pool = self[kAgentCache].get(origin)
 
     if (!pool) {
-      const Klass = self[kAgentOpts] && self[kAgentOpts].connections === 1
-        ? self[kClientClass]
-        : self[kPoolClass]
-
-      pool = new Klass(origin, self[kAgentOpts])
+      pool = this[kFactory](origin, self[kAgentOpts])
       pool
         .on('connect', this[kOnConnect])
         .on('disconnect', this[kOnDisconnect])
@@ -80,6 +78,10 @@ class Agent extends EventEmitter {
 
 let globalAgent = new Agent({ connections: null })
 
+function defaultFactory(origin, opts) {
+  return opts && opts.connections === 1 ? new Client(origin, opts) : new Pool(origin, opts)
+}
+
 function setGlobalAgent (agent) {
   if (!agent || typeof agent.get !== 'function') {
     throw new InvalidArgumentError('Argument agent must implement Agent')
@@ -103,7 +105,7 @@ function dispatchFromAgent (requestType) {
       throw new InvalidReturnValueError(`Client returned from Agent.get() does not implement method ${requestType}`)
     }
 
-    return client[requestType]({ ...opts, agent, method, path }, ...additionalArgs)
+    return client[requestType]({ ...opts, agent, requestType, method, path }, ...additionalArgs)
   }
 }
 
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -19,11 +19,17 @@ class Agent extends EventEmitter {
 
     const agent = this
 
-    const factoryType = typeof opts.factory
-    if (factoryType !== 'undefined' && factoryType !== 'function') {
-      throw new InvalidArgumentError('factory must be a function.')
+    if (opts && opts.factory) {
+      const factoryType = typeof opts.factory
+
+      if (factoryType !== 'undefined' && factoryType !== 'function') {
+        throw new InvalidArgumentError('factory must be a function.')
+      }
+
+      this[kFactory] = opts.factory
+    } else {
+      this[kFactory] = defaultFactory
     }
-    this[kFactory] = (opts && opts.factory) || defaultFactory
 
     this[kOnConnect] = function onConnect (client) {
       agent.emit('connect', client)
@@ -49,9 +55,7 @@ class Agent extends EventEmitter {
 
     if (!pool) {
       pool = this[kFactory](origin, self[kAgentOpts])
-      pool
-        .on('connect', this[kOnConnect])
-        .on('disconnect', this[kOnDisconnect])
+      pool.on('connect', this[kOnConnect]).on('disconnect', this[kOnDisconnect])
 
       self[kAgentCache].set(origin, pool)
     }
@@ -78,7 +82,7 @@ class Agent extends EventEmitter {
 
 let globalAgent = new Agent({ connections: null })
 
-function defaultFactory(origin, opts) {
+function defaultFactory (origin, opts) {
   return opts && opts.connections === 1 ? new Client(origin, opts) : new Pool(origin, opts)
 }
 
--- a/lib/redirect-pool.js
+++ b/lib/redirect-pool.js
@@ -92,7 +92,7 @@ class RedirectPool extends Pool {
         Remove headers referring to the original URL.
         By default it is Host only, unless it's a 303 (see below), which removes also all Content-* headers.
       */
-      if ('headers' in opts) {
+      if (opts.headers) {
         cleanRequestHeaders(opts.headers, statusCode === 303)
       }
 
@@ -122,7 +122,7 @@ class RedirectPool extends Pool {
           stream(location, opts, handler.factory, handler.callback)
           break
         case 'pipeline':
-          pipeline(location, handler.handler)
+          pipeline(location, opts, handler.handler)
           break
       }
 
--- a/lib/redirect-pool.js
+++ b/lib/redirect-pool.js
@@ -1,14 +1,66 @@
+'use strict'
+
 const { Readable } = require('stream')
 const { pipeline, request, stream } = require('./agent')
+const { InvalidArgumentError } = require('./core/errors')
 const util = require('./core/util')
 const Pool = require('./pool')
 
-const kOrigin = Symbol('Origin')
-const kDispatchMethod = Symbol('RedirectPoolDispatchMethod')
-const kRedirectsLeft = Symbol('RedirectPoolRedirectsLeft')
+const kOrigin = Symbol('origin')
+const kLeftRedirections = Symbol('left redirections')
 
 const redirectCodes = [300, 301, 302, 303, 307, 308]
 const defaultMaxRedirections = 10
+
+function parseMaxRedirections({ maxRedirections = defaultMaxRedirections }) {
+  if (maxRedirections != null && (maxRedirections <= 0 || !Number.isInteger(maxRedirections))) {
+    throw new InvalidArgumentError('maxRedirections must be a positive number')
+  }
+
+  return maxRedirections
+}
+
+function redirectLocation(statusCode, headers, opts) {
+  if (opts[kLeftRedirections] < 0 || redirectCodes.indexOf(statusCode) === -1) {
+    return null
+  }
+
+  for (let i = 0; i < headers.length; i += 2) {
+    // Find the matching headers, then return its value
+    if (headers[i].length && headers[i].toLowerCase() === 'location') {
+      return headers[i + 1]
+    }
+  }
+
+  return null
+}
+
+// https://tools.ietf.org/html/rfc7231#section-6.4.4
+function shouldRemoveHeader(header, removeContent) {
+  const lcHeader = header.toLowerCase()
+
+  return lcHeader === 'host' || (removeContent && lcHeader.indexOf('content-') === 0)
+}
+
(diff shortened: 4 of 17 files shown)


lib/redirect-pool.js
```diff
@@ -0,0 +1,153 @@
+const { Readable } = require('stream')
```
mcollina, line 1: add 'use strict'

lib/redirect-pool.js
```diff
@@ -0,0 +1,153 @@
...
+    const pool = this
+
+    // Compute the number of left redirects, if needed
+    if (!(kRedirectsLeft in opts)) {
+      this.parseMaxRedirects(opts)
+    }
+
+    const originalOnHeaders = handler.onHeaders
+
+    handler.onHeaders = function onHeadersWithRedirect(statusCode, headers, resume) {
```
mcollina, line 54: If it's a different origin, it should go back to the Agent.

mcollina (changes requested):

ronag (changes requested):
This should be possible to implement without touching request, stream or pipeline.

lib/agent.js
```diff
@@ -41,10 +46,11 @@ class Agent extends EventEmitter {
...
 
     if (!pool) {
-      pool = self[kAgentOpts] && self[kAgentOpts].connections === 1
-        ? new Client(origin, self[kAgentOpts])
-        : new Pool(origin, self[kAgentOpts])
+      const Klass = self[kAgentOpts] && self[kAgentOpts].connections === 1
+        ? self[kClientClass]
+        : self[kPoolClass]
 
+      pool = new Klass(origin, self[kAgentOpts])
```
ronag, line 53: Use a factory function instead as argument, i.e.

```js
pool = factory(origin, self[kAgentOpts])
```

lib/redirect-pool.js
```diff
@@ -0,0 +1,153 @@
...
+      for (const rawHeaderName of Object.keys(headers)) {
+        const headerName = rawHeaderName.toLowerCase()
+
+        if (headerName === 'host' || headerName.indexOf('content-') === 0) {
+          headers[rawHeaderName] = undefined
+        }
+      }
+    }
+  }
+}
```
ronag, line 151: Move helper methods out of the class.

lib/redirect-pool.js
```diff
@@ -0,0 +1,153 @@
...
+        if (headerName === 'host' || headerName.indexOf('content-') === 0) {
+          headers.splice(i, 2)
+        }
+      }
+    } else {
+      // IncomingHttpHeaders
+      for (const rawHeaderName of Object.keys(headers)) {
+        const headerName = rawHeaderName.toLowerCase()
+
+        if (headerName === 'host' || headerName.indexOf('content-') === 0) {
```
ronag, line 145: Please reference the spec.

lib/redirect-pool.js
```diff
@@ -0,0 +1,153 @@
...
+    const headerIndex = headers.findIndex(h => h.toLowerCase() === 'location')
+    return headerIndex >= 0 ? headers[headerIndex + 1] : null
+  }
+
+  removeHostSpecificHeaders(headers) {
+    if (Array.isArray(headers)) {
+      for (let i = headers.length - 2; i >= 0; i -= 2) {
+        const headerName = headers[i].toLowerCase()
+
+        if (headerName === 'host' || headerName.indexOf('content-') === 0) {
```
ronag, line 136: Please reference the spec.

lib/redirect-pool.js
```diff
@@ -0,0 +1,153 @@
...
+    }
+  }
+
+  redirectLocation(statusCode, headers, opts) {
+    if (opts[kRedirectsLeft] < 0 || redirectCodes.indexOf(statusCode) === -1) {
+      return null
+    }
+
+    // Find the Location header and then the value
+    const headerIndex = headers.findIndex(h => h.toLowerCase() === 'location')
```
ronag, line 127: A for loop is faster and also compare length before `toLowerCase()`.

lib/redirect-pool.js
```diff
@@ -0,0 +1,153 @@
...
+
+      // Pause the current socket not to buffer the body (if any) at all
+      return true
+    }
+
+    // Call the original implementation to proceed
+    super.dispatch(opts, handler)
+  }
+
+  parseMaxRedirects(opts) {
```
ronag, line 107:
```suggestion
  parseMaxRedirects({ maxRedirections }) {
    if (maxRedirections != null && (maxRedirections < 0 || !Number.isInteger(maxRedirections)) {
       // throw invalid arg
     }
```

lib/redirect-pool.js
```diff
@@ -0,0 +1,153 @@
...
+    return super.stream(opts, factory, callback)
+  }
+
+  pipeline(opts, handler) {
+    opts[kDispatchMethod] = 'pipeline'
+
+    return super.pipeline(opts, handler)
+  }
+
+  dispatch(opts, handler) {
```
ronag, line 44: Don't mutate opts.

lib/redirect-pool.js
```diff
@@ -0,0 +1,153 @@
...
+
+    handler.onHeaders = function onHeadersWithRedirect(statusCode, headers, resume) {
+      // Check if statusCode is 3xx, if there is a location header and if the redirection can be followed
+      const location = pool.redirectLocation(statusCode, headers, opts)
+
+      if (!location) {
+        return originalOnHeaders.call(handler, statusCode, headers, resume, { redirections: opts.redirections })
+      }
+
+      // In case of HTTP 303, always replace method to be either HEAD or GET
```
ronag, line 62: Reference to spec.

lib/redirect-pool.js
ShogunPanda: I don't see your point. I do that in https://github.com/ShogunPanda/undici/blob/main/lib/redirect-pool.js#L83, which basically calls the top level method. Can you please clarify?

lib/agent.js
ShogunPanda: Nice, I really like it. It will also simplify the Typescript definitions.
I just have a question: do I also need a RedirectClient or it is perfectly fine to always instantiate a subclass of `Pool` even if `connections==1`?

lib/redirect-pool.js
ShogunPanda: Sure.

lib/redirect-pool.js
ShogunPanda: Sure.

lib/redirect-pool.js
ShogunPanda: Sure.

lib/redirect-pool.js
ShogunPanda: Bad habits never go away. Lol. Will change it soon.

lib/redirect-pool.js
ShogunPanda: Will do.

lib/redirect-pool.js
ShogunPanda: Good point. 
Coercing the argument is a bad idea. Using `true` is pointless since once can just drop the argument at all. And `false` makes no sense if your using the RedirectPool.

I can't directly accept your suggestion since the check should be `...maxRedirections <= 0 ...`, but I'll change it accordingly.

lib/redirect-pool.js
ShogunPanda: I have to think on this a little more.

While I can probably remove overloads of `request`, `stream` and `pipeline`, I still have to take track of of previous and redirects and left redirects.
And since when following a redirection I call the top level method, I'm completely in a stateless env.
Thus, adding few vars to opts it's my only choice, or not?
Moreover, I see that opts are already mutated when needed, let's say in `dispatchFromAgent` where agent is currently (in upstream) extracted from opts.

What do you think?

lib/redirect-pool.js
ShogunPanda: Will do!

lib/redirect-pool.js
ronag: You can just make a copy of the opts object the first time. The only important thing is that we don't mutate any object that the user passes.

lib/agent.js
ronag: You don't need redirect client. It's just a micro optimization.

lib/redirect-pool.js
mcollina: A `Pool` has only a single destination of protocol/domain/port. If we are changing those in the redirect, we should go back to the `Agent`. As I read the code, this is not working right now nor there is a test for it.

ShogunPanda: I do instead. When I call request in line https://github.com/ShogunPanda/undici/blob/main/lib/redirect-pool.js#L85, is the top `undici.request` which ultimately calls `dispatchFromAgent` and gets a pool for a new origin from the agent.

About the test, there is explicitly one: https://github.com/ShogunPanda/undici/blob/main/test/redirect-request.js#L339 - In this case only the port changes, but it shouldn't matter, right?

mcollina: Oh, I understand now. I think the Pool should receive an instance of its own agent and not pass through the global agent. The current agent vs the global agent could be different.

ShogunPanda: I think that's already the case. At the moment the pool calls the top-level method by sending the original agent back. 
https://github.com/ShogunPanda/undici/blob/main/lib/agent.js#L106

As you can see the agent (which might or might not be the global one) is send to the pool which in turn send it back to `dispatchFromAgent`. So only a single agent is used. Am I wrong?

ShogunPanda: Also, consider that the agent must be set explicitily configured in order to use the `RedirectPool`, so either `request` and similar are called from with a `agent`, or `setGlobalAgent` has been called.

mcollina: Amazing thanks! This work beautifully!

lib/redirect-pool.js
ShogunPanda: Done!

lib/agent.js
ShogunPanda: Done!

lib/redirect-pool.js
ShogunPanda: Done!

lib/redirect-pool.js
ShogunPanda: Done!

lib/redirect-pool.js
ShogunPanda: Done!

lib/redirect-pool.js
ShogunPanda: Done!

lib/redirect-pool.js
ShogunPanda: Done!

lib/redirect-pool.js
ShogunPanda: Done!

lib/redirect-pool.js
ShogunPanda: So, I've double checked the code. We're technically already doing that. 

https://github.com/nodejs/undici/blob/2366fa34e30ac5a8c33a3d02259f2ae0f7678c85/lib/agent.js#L85

Here's we are extracting `agent` and we are providing default values for `agent` and `method`.
In the current upstream `agent` is no longer passed to the pool, while in my PR it is (in order to allow recursion).

Now, since we use destructuring at the beginning of the function and rest assignment at the end, we're forwarding a new object to pool everytime, so we're not mutating the user provided `opts` object.

Moreover, when I follow a redirect, I only add two properties (via private symbols), so I'm not overwriting user provided values.
Am I wrong on this?

mcollina (approved):
lgtm

docs/api/Agent.md
```diff
@@ -2,12 +2,18 @@
...
 
 Returns: `Agent`
 
 Returns a new Agent instance for use with pool based requests or the following top-level methods `request`, `pipeline`, and `stream`.
 
+### Parameter: `AgentOptions`
+
+Extends: [`PoolOptions`](docs/api/Pool.md#parameter-pooloptions)
+
+* **factory** `(url: string, options?: Client.Options): Pool` (optional) - A factory which returns the pool to use for the request.
```
mcollina, line 15: Can you add an example on how to set up an agent to follow redirects? Something copy-pastable.

docs/api/Pool.md
```diff
@@ -105,3 +105,29 @@ Implements [Client Event: `'disconnect'`](docs/api/Client.md#event-connect)
 ### Event: `'drain'`
 
 Implements [Client Event: `'drain'`](docs/api/Client.md#event-connect)
+
+# Class: RedirectPool
```
mcollina, line 109: I think this should be a separate md file.

docs/api/Pool.md
```diff
@@ -105,3 +105,29 @@ Implements [Client Event: `'disconnect'`](docs/api/Client.md#event-connect)
...
+# Function: redirectPoolFactory
+
+A factory function which returns a `RedirectPool` for a specific URL.
+
+Arguments:
+
+* **url** `URL | string` - It should only include the **protocol, hostname, and port**.
+* **options** `PoolOptions` (optional)
+
+Returns: `RedirectPool` - The pool to be used by the agent.
```
mcollina, line 132: Please add a few examples

mcollina (changes requested):

docs/api/Agent.md
ShogunPanda: Sure. No problem.

docs/api/Pool.md
ShogunPanda: Good point. Moving it.

ShogunPanda:
@mcollina @ronag 

One additional comment. Even if you both approve the CI, please wait for me to provide a `redirect-request.js` similar file for `pipeline` and `stream` to make sure all top level methods are tested.

docs/api/Agent.md
ShogunPanda: Done!

docs/api/Pool.md
ShogunPanda: Done!

docs/api/Pool.md
ShogunPanda: I think that the example I added in Agent docs should suffice even for it, isn't it?

ShogunPanda:
@ronag @mcollina 

As announced, I added tests for `undici.stream` as well.
While I was working on the tests for `undici.pipeline`, I found a problem which I think applies to `undici.stream` and `undici.request`.

In my specific case I was testing using `stream.pipeline`, like this:

```js
function createDummyReadable() {
  return new Readable({
    read() {
      console.log('READ')
      this.push(Buffer.from('undici'))
      this.push(null)
    }
  })
}

function createWritable(target) {
  return new Writable({
    write(chunk, _, callback) {
      target.push(chunk.toString())
      callback()
    },
    final(callback) {
      callback()
    }
  })
}

t.test('should follow redirection after a HTTP 300', async t => {
  t.plan(4)

  const body = []
  const serverRoot = await startServer(t)

  await streamPipeline(
    createDummyReadable(),
    undiciPipeline(
      `http://${serverRoot}/300`,
      { agent: new Agent({ factory: redirectPoolFactory }) },
      ({ statusCode, headers, redirections, body }) => {
        t.strictEqual(statusCode, 200)
        t.notOk(headers.location)
        t.deepEqual(redirections, [
          `http://${serverRoot}/300`,
          `http://${serverRoot}/300/1`,
          `http://${serverRoot}/300/2`,
          `http://${serverRoot}/300/3`,
          `http://${serverRoot}/300/4`
        ])

        return body
      }
    ),
    createWritable(body)
  )

  t.strictEqual(body.join(''), `GET :: connection@keep-alive host@${serverRoot}`)
})
```

The problem that I'm noticing is that when I follow the redirect (which, to clarify, calls `undici.pipeline` again), the `Readable` has been completed consumed, so I cannot continue.
This problem should (I haven't verified explicitily but I think it's fair reasonable) arise even if `undici.stream` and `undici.request` use a `Readable` as `body`.

My first thought to solve was to copy the readable while consuming it and somehow use it when recursing. But it seems dangerous to me due to possible memory issues and leak.

The second thought would be to prevent use of `pipeline` and/or `Readable` bodies when using `RedirectPool`. But I would avoid it since the factory should be pretty transparent to the user.

Am I right? Any suggestion on how to solve it?

ronag:
You shouldn’t need to do anything with pipeline, request and stream. You should only need to override dispatch and the rest should just work.

ShogunPanda:
I'm not doing anything weird. I override dispatch which overrides `handler.onHeaders`. That override relaunches `request`, `stream` and `pipeline` internally and that's where the problem arise because at that stage the request body has been consumed.

Am I wrong?

mcollina:
I don't think a redirect is possible with a POST etc, and a body with a GET is against the spec.

ShogunPanda:
I agree that a body with a GET is against the spec. And by looking at the implementation, undici correctly does not send it to the server. I'll double check this case because probably there something wrong in my code or test.

I've checked RFC 7231 and nothing indicates that a redirection with non GET/HEAD is forbidden:
  * 303 explicitily mention doing `retrieval request`
  * 301/302 are usually changed to `GET` for historical reason, but it's not mandated from the RFC (and in RFC 2616 there is a note saying this a wrong interpretation of clients)
  * 307 (and 308, which is part of a different RFC cited within 307) explicitily forbids the client from changing the method. Which implies redirecting from non `GET/HEAD` is legal.

So, how should I deal with it?

mcollina:
The only way to deal with it is to wait for the headers before starting to send the body.

ShogunPanda:
I don't think that is a viable option, unfortunately: not all servers redirect before receiving the body. 
Actually I think it's quite the opposite.

I'm leaning towards preventing `RedirectPool` to use `Readable` request bodies.

mcollina:
Go for it.

lib/agent.js
```diff
@@ -97,7 +109,7 @@ function dispatchFromAgent (requestType) {
       throw new InvalidReturnValueError(`Client returned from Agent.get() does not implement method ${requestType}`)
     }
 
-    return client[requestType]({ ...opts, method, path }, ...additionalArgs)
+    return client[requestType]({ ...opts, agent, requestType, method, path }, ...additionalArgs)
```
ronag, line 112: I’m not sure I think this should be needed

lib/redirect-pool.js
```diff
@@ -0,0 +1,149 @@
...
+        case 'request':
+          request(location, opts, handler.callback)
+          break
+        case 'stream':
+          stream(location, opts, handler.factory, handler.callback)
+          break
+        case 'pipeline':
+          pipeline(location, opts, handler.handler)
+          break
+      }
```
ronag, line 127: I don’t think this should be needed.

ronag (changes requested):

lib/redirect-pool.js
```diff
@@ -0,0 +1,149 @@
...
+    const pool = this
+
+    // Compute the number of left redirects, if needed
+    if (!(kLeftRedirections in opts)) {
+      opts[kLeftRedirections] = parseMaxRedirections(opts)
+    }
+
+    const originalOnHeaders = handler.onHeaders
+
+    handler.onHeaders = function onHeadersWithRedirect (statusCode, headers, resume) {
```
ronag, line 80: Don’t modify user passed/owned objects.

types/agent.d.ts
```diff
@@ -13,25 +12,25 @@ export {
...
+  constructor(opts?: Pool.Options & { factory?: (url: string, opts: Pool.Options) => Pool })
   get(origin: string): Pool;
 }
 
 declare function setGlobalAgent<AgentImplementation extends Agent>(agent: AgentImplementation): void;
 
 declare function request(
   url: string | URL | UrlObject,
-  opts?: { agent?: Agent } & Omit<Client.RequestOptions, 'path'>,
+  opts?: { agent?: Agent, maxRedirections?: boolean | number } & Omit<Partial<Client.RequestOptions>, 'path'>,
```
ronag, line 23: Only number?

ronag (commented):
So instead of keeping track of the original request type just make a new dispatch call with the original handler but with new options.

lib/redirect-pool.js
```diff
@@ -0,0 +1,149 @@
...
+
+    this[kOrigin] = origin
+  }
+
+  dispatch (opts, handler) {
+    const pool = this
+
+    // Compute the number of left redirects, if needed
+    if (!(kLeftRedirections in opts)) {
+      opts[kLeftRedirections] = parseMaxRedirections(opts)
```
ronag, line 75: Don’t modify user passed/owned objects

lib/redirect-pool.js
```diff
@@ -0,0 +1,149 @@
...
+      }
+
+      /*
+        https://tools.ietf.org/html/rfc7231#section-6.4
+
+        Remove headers referring to the original URL.
+        By default it is Host only, unless it's a 303 (see below), which removes also all Content-* headers.
+      */
+      if (opts.headers) {
+        cleanRequestHeaders(opts.headers, statusCode === 303)
```
ronag, line 96: Don’t modify user passed/owned objects

lib/redirect-pool.js
```diff
@@ -0,0 +1,149 @@
...
+      }
+
+      // https://tools.ietf.org/html/rfc7231#section-6.4.4
+      // In case of HTTP 303, always replace method to be either HEAD or GET
+      if (statusCode === 303 && opts.method !== 'HEAD') {
+        opts.method = 'GET'
```
ronag, line 102: Don’t modify user passed/owned objects

types/agent.d.ts
ShogunPanda: Thanks, I forgot to change it.

lib/redirect-pool.js
ShogunPanda: I'm not sure I understand your point here. When the pool is invoked via top-level agent, the opts are already modified (since we use destructuring and then rest assignment).

But if you're worried about direct use of the class ok, I'll change it.

lib/redirect-pool.js
ShogunPanda: See my comment above.

lib/redirect-pool.js
ShogunPanda: See my comment above.

lib/agent.js
ShogunPanda: I need at least agent to being able to call dispatch in the right way. See my comment in the PR discussion.

lib/redirect-pool.js
ShogunPanda: I need this in order to support cross origin redirection.

lib/redirect-pool.js
ShogunPanda: Well, that's hard here. I'll see if I can achieve in a different way.

ShogunPanda:
@ronag I can't simply make a new dispatch call within the pool. 
This would only work in same-origin case. I could probably distinguish the two cases, but I don't think it's worth the effort. Easier to just recurse with top level method.

I'll try not to use `requestType` but I need to add agent to `dispatchFromAgent` in order to be fully able to dispatch a call to another URL with same settings.

lib/redirect-pool.js
ShogunPanda: This is solved now. I'm not mutating opts anymore.

lib/agent.js
ShogunPanda: I'm now just adding agent (which might be originally given by the user and thus keeps the promise not to mutate opts)

lib/redirect-pool.js
ShogunPanda: Removed the part about `pipeline` (since RedirectPool is not supported on pipeline).
About `request` vs `stream`, I know try to autodetect given the shape of the handler.

lib/redirect-pool.js
ShogunPanda: Done.

types/agent.d.ts
ShogunPanda: Fixed.

lib/redirect-pool.js
ShogunPanda: Done.

lib/redirect-pool.js
ShogunPanda: Fixed.

lib/redirect-pool.js
ShogunPanda: Fixed.

ShogunPanda:
@ronag 

After my latest changes, I consider this PR feature complete. `request` and `stream` have been broadly tested, `pipeline` has been restricted and no user provided objects are moved.

Let me know if there are additional modifications that needs to be done or if this finally looks good.

docs/api/Agent.md
```diff
@@ -2,12 +2,62 @@
...
+Extends: [`PoolOptions`](docs/api/Pool.md#parameter-pooloptions)
+
+* factory `(url: string, options?: Client.Options): Pool` (optional) - A factory which returns the pool to use for the request.
+
+#### Example - Use `factory` to use a RedirectPool which follow redirects
+
+```js
+'use strict'
+const { createServer } = require('http')
+const { Agent, request, RedirectPool } = require('.')
```
mcollina, line 22:
```suggestion
const { Agent, request, RedirectPool } = require('undici')
```

mcollina (commented):
A nit!

lib/redirect-pool.js
```diff
@@ -0,0 +1,187 @@
...
+          [kFollowedRedirections]: followedRedirections
+        }
+
+        // Follow the redirect using same options - Note the top-level method must be used to support cross origin
+        if (handler.factory) {
+          // This is a StreamHandler
+          stream(location, redirectedOpts, handler.factory, handler.callback)
+        } else {
+          request(location, redirectedOpts, handler.callback)
+        }
```
ronag, line 156: I still think these are not needed. I’ll take a look and try to show within a few days.

ronag (changes requested):
Looking good! I still think this can be implemented with dispatch only. I’ll clone the PR and try it myself as soon as I can and come back with my findings.

ShogunPanda:
@ronag Help yourself!
If you find that way I'll be more than happy. That part looks a little hackish, especially when detecting the current method :)

lib/redirect-pool.js
```diff
@@ -0,0 +1,187 @@
...
+        let { method, headers: requestHeaders, body } = opts
+
+        /*
+          https://tools.ietf.org/html/rfc7231#section-6.4
+
+          Remove headers referring to the original URL.
+          By default it is Host only, unless it's a 303 (see below), which removes also all Content-* headers.
+        */
+        if (requestHeaders) {
+          cleanRequestHeaders(requestHeaders, statusCode === 303)
```
ronag, line 122: This mutates user owned object

ronag:
We can make this work with pipeline.

ShogunPanda:
@ronag It's a question or an affirmation? In the latter case if you know how to deal with it let me know and I'll change the code.

ronag:
Just a note to self.

lib/redirect-pool.js
ShogunPanda: Good catch. Fixed.

mcollina: Did you push? I don't see a commit?

ShogunPanda: Apparently I forgot. :)

I pushed right now!

mcollina (approved):
lgtm

ronag:
I don't think this PR is entirely correct. This is how I would have implemented it just as a global dispatch. I haven't gotten to a point on how to inject this into `Pool` yet.

```js

const util = require('./core/util')
const { globalAgent } = require('./agent')
const { InvalidArgumentError } = require('./core/errors')

module.exports = function dispatch (url, { maxRedirections = 10, ...opts }, handler, agent = globalAgent) {
  if (maxRedirections != null && (maxRedirections < 0 || !Number.isInteger(maxRedirections))) {
    throw new InvalidArgumentError('maxRedirections must be a positive number')
  }

  const { origin, pathname, search } = util.parseURL(url)

  const path = `${pathname || '/'}${search || ''}`

  const client = agent.get(origin)

  if (!maxRedirections) {
    return client.dispatch(opts, handler)
  }

  // TODO (fix): Restartable streams?
  if (util.isStream(opts.body) && util.bodyLength(opts.body) !== 0) {
    return client.dispatch(opts, handler)
  }

  client.dispatch(opts, {
    agent,
    opts: { ...opts, path, maxRedirections: maxRedirections - 1 },
    origin,
    maxRedirections,
    handler,
    pool: this,
    location: null,

    onConnect (abort) {
      this.onConnect(abort)
    },

    onUpgrade (statusCode, headers, socket) {
      this.onUpgrade(statusCode, headers, socket)
    },

    onError (error) {
      this.onError(error)
    },

    onHeaders (statusCode, headers, resume) {
      // Check if statusCode is 3xx, if there is a location header and if the redirection can be followed
      this.location = this.maxRedirections
        ? redirectLocation(statusCode, headers)
        : null

      if (!this.location) {
        return this.handler.onHeaders(statusCode, headers, resume)
      }

      this.location = new URL(this.location, this.origin)

      // Remove headers referring to the original URL.
      // By default it is Host only, unless it's a 303 (see below), which removes also all Content-* headers.
      // https://tools.ietf.org/html/rfc7231#section-6.4
      this.opts.headers = cleanRequestHeaders(this.opts.headers, statusCode === 303)

      // https://tools.ietf.org/html/rfc7231#section-6.4.4
      // In case of HTTP 303, always replace method to be either HEAD or GET
      if (statusCode === 303) {
        this.opts.method = 'GET'
        // TOOD (fix): What if body?
      }
    },

    onData (chunk) {
      if (!this.location) {
      }
    },

    onComplete (trailers) {
      if (this.location) {
        dispatch(this.location, this.opts, this.handler, this.agent)
      } else {
        this.handler.onComplete(trailers)
      }
    }
  })
}

function redirectLocation (statusCode, headers) {
  // TODO (fix): Different redirect codes are handled differently?
  if ([300, 301, 302, 303, 307, 308].indexOf(statusCode) === -1) {
    return null
  }

    // Find the matching headers, then return its value
    if (headers[i].length && headers[i].toLowerCase() === 'location') {
      return headers[i + 1]
    }
  }

  return null
}

// https://tools.ietf.org/html/rfc7231#section-6.4.4
function shouldRemoveHeader (header, removeContent) {
  return (
    (header.length === 4 && header.toLowerCase() === 'host') ||
    (removeContent && header.toLowerCase().indexOf('content-') === 0)
  )
}

// https://tools.ietf.org/html/rfc7231#section-6.4
function cleanRequestHeaders (headers, removeContent) {
  const ret = []
  if (Array.isArray(headers)) {
      if (!shouldRemoveHeader(headers[i], removeContent)) {
        ret.push(headers[i], headers[i + 1])
      }
    }
      if (!shouldRemoveHeader(key, removeContent)) {
        ret.puhs(key, val)
      }
    }
  } else if (headers != null) {
    throw new InvalidArgumentError('headers must be an object or an array')
  }
  return ret
}
```

mcollina:
I would not have a global method for this as I would like both normal and redirect agents in the same process.

ronag:
I agree. That’s just ironing out the core logic of it.

ShogunPanda:
Your implementation is similar to mine, except that you started from a different POV.

When I started working on that, we agreed on #553 that I would have tried to modify existing undici to the minimum, in order not to affect the current implementation and performance. That's why the pool class. 

Working a that a level is a different way to solve the problem, but premises are completely different.

ronag:
Please take a look at the core logic. There are differences. It can be applied to pool as well.

ShogunPanda:
Sorry, but I don't see the difference other than a different approach due to different POV (which I couldn't follow, due to the agreement on #553). 
Can you please tell me what you want me to focus on?

Also, note that from within pool there is no agent and the origin is already fixed. So, no cross-origin possible, AFAIK.

Finally a note on one of your TODOs (within `redirectLocation`) . Only 303 requires the change to GET/HEADER. 301 and 302 were mistakenly changed from non-GET to  GET and now it is allowed from RFC but not mandated. I don't think we should follow that.
307 and 308 explicitily forbid method changing.

If we also decide to change method on 301 on 302, request bodies should be ignored.

ronag:
> Sorry, but I don't see the difference other than a different approach due to different POV (which I couldn't follow, due to the agreement on #553).
> Can you please tell me what you want me 

mill post my final suggestion and then we can discuss.

> 
> Also, note that from within pool there is no agent and the origin is already fixed. So, no cross-origin possible, AFAIK.
> 

Yes. Which makes me unsure whether w es hours implement this on client or agent level.

> Finally a note on one of your TODOs (within `redirectLocation`) . Only 303 requires the change to GET/HEADER. 301 and 302 were mistakenly changed from non-GET to GET and now it is allowed from RFC but not mandated. I don't think we should follow that.
> 307 and 308 explicitily forbid method changing.
> 

Sounds good!

> If we also decide to change method on 301 on 302, request bodies should be ignored.

ShogunPanda:
> mill post my final suggestion and then we can discuss.

I'm sorry, I don't understand the sentence at all (Italian native here). Can you please clarify?

> Yes. Which makes me unsure whether w es hours implement this on client or agent level.

Well, from what I've seen, there are pros and cons:

* Implementing at pool level like it did requires few modifications on existing undici, but implies the use of an agent and thus only `undici.request` and `undici.stream`. 
* Implementing at client level would be better if the new behavior can be injected as mixin as both top level method and the client API could benefit of it. But there is no apparent way to solve the cross-origin problem.

ronag:
closing in favor of https://github.com/nodejs/undici/pull/603

PR #603, opened 2021-03-22
feat: redirect

ronag:
Continuing on from: #600

Trying to figure out the API. Doing this through `Agent` or `Pool` doesn't quite make sense IMO.


mcollina:
We are not in agreement. Using an Agent will make adding redirect being transparent for `fetch()` and other high-level libraries. Adding a different API would provide a worse developer experience (however it would be a simpler implementation).

ronag:
Hm. I agree it would
be nicer to implement this just another client and agent. I just don’t see how to implement this in a way that makes sense in Agent or Pool. 

I have some more ideas here regarding developer experience so I’ll continue working on this and see if we can reach an agreement.

ronag:
Agent would have to return a redirect client which would have to have a reference to the agent.

Then the client needs to correctly implement all the client APIs. Many of which would not make any sense at all. Especially with cross origin in mind.

ronag:
@mcollina PTAL. Has a change to the Agent API to enable better composition.

ronag:
@ShogunPanda

lib/redirect-agent.js
```diff
@@ -0,0 +1,169 @@
...
+        this.location = parseLocation(headers)
+
+        if (!this.location) {
+          return this.handler.onHeaders(statusCode, headers, resume)
+        }
+
+        this.location = new URL(this.location, this.origin)
+
+        if (!this.opts.history) {
+          opts.history = new Set()
```
ShogunPanda, line 89: I wouldn't do this. This will make impossible for users to detect if there was any redirect cycle.
Let the users deal with it. Since we have maxRedirections we don't really care.

lib/redirect-agent.js
```diff
@@ -0,0 +1,171 @@
...
+
+      onData (chunk) {
+        if (!this.location) {
+          return this.handler.onData(chunk)
+        }
+      },
+
+      onComplete (trailers) {
+        if (this.location) {
+          this.dispatch(this.location, this.opts, this.handler)
```
ShogunPanda, line 119: Little note here.
If you trigger dispatch from here, it means you will wait for the body (which, even not common, are legit) to be transferred before following the redirect.

That's why in #600 I dispatched from `onHeaders`. Especially when going cross-origin, we don't care about the original response body, isn't it?

lib/redirect-agent.js
```diff
@@ -0,0 +1,171 @@
...
+  dispatch (origin, { maxRedirections = this[kMaxRedirections], ...opts }, handler) {
+    if (maxRedirections != null && (maxRedirections < 0 || !Number.isInteger(maxRedirections))) {
+      throw new InvalidArgumentError('maxRedirections must be a positive number')
+    }
+
+    if (!maxRedirections) {
+      return this[kAgent].dispatch(origin, opts, handler)
+    }
+
+    if (util.isStream(opts.body) && util.bodyLength(opts.body) === 0) {
```
ShogunPanda, line 35: Shouldn't it be `!== 0`? Or am I missing something?

ShogunPanda:
@ronag I like the API. Nice work.
I would suggest to copy tests from my PR, adapt them (basically replacing `RedirectPool` with `RedirectAgent` and see it they passes.

lib/redirect-agent.js
ronag: Unfortunately we need to receive the entire body or pipelining won't work.

lib/redirect-agent.js
```diff
@@ -0,0 +1,173 @@
...
+const util = require('./core/util')
+const EventEmitter = require('events')
+const { getGlobalAgent } = require('./agent')
+const assert = require('assert')
+
+const kMaxRedirections = Symbol('maxRedirections')
+const kAgent = require('agent')
+
+class RedirectAgent extends EventEmitter {
+  constructor ({ maxRedirections = 10, agent = getGlobalAgent() } = {}) {
```
ShogunPanda, line 11: Just a confirmation question here: by doing this, you allow `maxRedirections` to be specified both at agent and single request level, isn't it?

lib/redirect-agent.js
ronag: If the stream is empty (i.e. ` === 0´) we just simply destroy and remove it. So it's correct. This way we can support `pipeline`.

lib/redirect-agent.js
ronag: Correct.

lib/redirect-agent.js
ronag: How will the user deal with it though? We would need some kind of hook in that case.

lib/redirect-agent.js
ShogunPanda: Right point. 

I guess we need two approaches:

1. Same origin -> dispatch from onComplete
2. Cross origin -> dispatch from onHeaders

This way the client will experience the best performance. Am I wrong?

ronag:
> I would suggest to copy tests from my PR, adapt them (basically replacing RedirectPool with RedirectAgent and see it they passes.

Yes, I was hoping we could collaborate on the tests. I'll port them over from your PR once we know whether @mcollina agrees on the approach.

lib/redirect-agent.js
ShogunPanda: Well, we have `maxRedirections`, isn't? If they get in a cycle at some point it will stop following ending with a 3xx.
And then they'll check history in the callback data to deal with it. But at undici level we don't care. Or not?

lib/redirect-agent.js
ronag: Yes, I think we are both right actually. We must read the body, BUT you are right that we could actually dispatch the follow immediately in `onHeaders`.

ronag: The only problem with dispatching in onHeaders is that I'm unsure how to handle any error to might occur while reading the original request.

ShogunPanda: Yeah, I come to the same conclusion XD

ShogunPanda:
> > I would suggest to copy tests from my PR, adapt them (basically replacing RedirectPool with RedirectAgent and see it they passes.
> 
> Yes, I was hoping we could collaborate on the tests. I'll port them over from your PR once we know whether @mcollina agrees on the approach.

Wonderful! ^^

lib/redirect-agent.js
ronag: I'm a little unsure whether the added complexity is worth the performance gain which I'm not sure whether are relevant in practice (though I have no numbers to back my scepticism).

ronag: Usually a redirect request will receive the EOF in the same TCP packet so `onComplete` is called immediatly.

ShogunPanda: To be honest I have no idea as well. 

It's unlikely that servers will send big bodies in case of 3xx. But I don't trust any server XD

lib/redirect-agent.js
ronag: That's a valid point. Where do we put the history though. There is no generic callback here. Does the http standard provide any header (such as `"via"`) where we can put the history?

@mcollina wdyt?

lib/redirect-agent.js
ShogunPanda: So in that case I would do a noop in `data` and ignore the `error`, as we don't probably care as well.

ronag: Yea, neither of them should be ever invoke. I'm mostly guarding against broken streams in the callbacks.

lib/redirect-agent.js
ronag: You ok with leaving as is?

ShogunPanda: If body is empty yes.

But I do also see you're point about error handling.

Well, then maybe just leave a comment note that this can technically be done so we can revisit in the future.

lib/agent.js
```diff
@@ -32,7 +31,7 @@ class Agent extends EventEmitter {
     }
   }
 
-  get (origin) {
```
mcollina, line 35: Can we keep the `get`? It will be used for #587

mcollina (commented):
The approach looks good to me, I've left a comment.

lib/redirect-agent.js
ShogunPanda: Sorry, I replied before reading your last question.

Yes, I'm fine with leaving like it. Maybe, as I said above, leave a note about this possible cross-origin optimization.

lib/redirect-agent.js
ShogunPanda: As you can see in my PR, I expanded `onHeaders` in all resources so that can accept additional arguments. And I used to pass the history as it builds the callback data for all methods.

This probably is also a good idea for future expandability as new `Agent`s (either in undici, or on user-level) might benefit for it to attach custom callback data.

lib/redirect-agent.js
ShogunPanda: You're right. Better not to ignore broken stuff in the wild because they might cause leaks in users' software.

ShogunPanda: In that case check what happens when the error is thrown. I think it should propagate up to the user level.

lib/redirect-agent.js
ShogunPanda: Much better than mine then! ^^

lib/agent.js
ronag: Not really... the api becomes weird. I can try to give it more thought.

lib/redirect-agent.js
ronag: I'm not sure we can propagate it to the user. In this case the best we can do is unhandled exception. Since it should never happen I think that's maybe fine. @mcollina wdyt?

lib/agent.js
ronag: @mcollina I don't see how we can keep this... it just won't work with `RedirectAgent` and gives more api area that needs to be fulfilled...

lib/redirect-agent.js
ronag: I can maybe fix this.

ronag:
I think this is as much as I can do at the moment. @ShogunPanda do you think you could help out with test coverage and docs? @Ethan-Arrowood maybe you could also help with docs?

Basically what we know how architecturewise is:

```
class Dispatch implements EventEmitter {
  dispatch (opts, handler)
  close ()
  destroy ()
}

class Agent implements Dispatch {
}

class RedirectAgent implements Dispatch {
}

interface Client implement Dispatch {
  ... client api
}

class Client implements Client {
}

class Pool implements Client {
}
```

lib/agent.js
mcollina: This was not a problem with the other implementation. Moreover this remove observability, as there is no way to know to access the Pool object.

I really think this should be retained.

ronag: > This was not a problem with the other implementation

The other implementation is not correct IMHO.

ronag: > Moreover this remove observability, as there is no way to know to access the Pool object.

I'm not sure I follow. What does this remove? Why would you need to access the Pool object?

ronag: It makes sense to me to leave the dispatching of a request as an implementation detail of the agent.

ronag: If we drop cross origin support it could make sense to keep the existing API. But that seems like a rather large thing to not support and kind of negates the point of having it inside of undici if people have to reimplement it anyway with cross origin support.

mcollina: I don't really understand what's the problem in exposing a get method.

mcollina (changes requested):

lib/agent.js
ronag: What would `RedirectAgent.get` return?

lib/agent.js
```diff
@@ -25,48 +37,81 @@ class Agent extends EventEmitter {
...
-    const self = this
-    let pool = self[kAgentCache].get(origin)
+    let pool = this[kCache].get(origin)
 
     if (!pool) {
-      pool = self[kAgentOpts] && self[kAgentOpts].connections === 1
-        ? new Client(origin, self[kAgentOpts])
-        : new Pool(origin, self[kAgentOpts])
-
+      pool = this[kFactory](origin)
```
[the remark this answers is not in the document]
ronag: They are passed. Se the constructor. Factory is a closure.

```
this[kFactory] = (origin) => factory(origin, opts)
```

lib/agent.js
mcollina: `this[kAgent].get(origin)` would do.

dnlup (commented):
Assuming we don't have to follow redirection using a specific origin, hence `Pool` (which I think it's unlikely, but I am just thinking out loud), I like this approach. I would expose the `get` method though, at this stage we need it and I think it causes no harm. I am open to revisiting it in the future (as I understand better what high-level and low-level APIs are needed by the user).

lib/agent.js
mcollina: Note that exposing `get(origin)` enable somebody to access some stats about the `Pool` and monitor its state.

ShogunPanda:
@dnlup I think we are following cross-origin. I'll confirm when I import tests from #600 

@ronag I can help with test at the moment. One thing I noticed is that you completed remove the support for the history of redirections. May I ask you why?

ronag:
> I can help with test at the moment. How can add test to your branch? Fork under my name and then send a PR for it?

Exactly!

> One thing I noticed is that you completed remove the support for the history of redirections. May I ask you why?

I was unsure about the API for it. Do you think we could add it in a follow up PR? I think I would like to discuss and iterate over that a little more.

lib/agent.js
ronag: Hm. I guess we could do that. One important note though is that `RediretPool.get(origin).request(opts)` won't follow any redirects. Which I think is kind of weird/confusing.

@dnlup also

ShogunPanda:
> Do you think we could add it in a follow up PR?

Yes, definitely. Once we get this merge I'd like to directly switch to it once we agree on the API. 

While is not required to provide the redirect functionality, I think is really important for developer experience.

lib/agent.js
ronag: Adding @dnlup comment here so we have the whole conversation:

> Assuming we don't have to follow redirection using a specific origin, hence Pool (which I think it's unlikely, but I am just thinking out loud), I like this approach. I would expose the get method though, at this stage we need it and I think it causes no harm. I am open to revisiting it in the future (as I understand better what high-level and low-level APIs are needed by the user).

dnlup:
> @dnlup I think we are following cross-origin. I'll confirm when I import tests from #600

Yes, we are. I explained myself poorly. I meant to say that implementing this using the `Agent` is cleaner than implementing it in the `Pool`. Unless there is a specific need to have a 'low-level `Pool` instance that can follow redirection (with the caveat that are only on the same origin). For example:

```js
const pool = new RedirectPool()

pool.request(...)
```

I think we can discard this use case because if a user needs to follow redirection, they would use the `request` and other helpers directly exposed by `undici`.

lib/agent.js
ronag: I'm adding it back as @mcollina suggested. I'm not sure I think it's a good idea but it should not block this. I can open a separate PR where we can continue this discussion later.

mcollina: > Hm. I guess we could do that. One important note though is that RediretPool.get(origin).request(opts) won't follow any redirects. Which I think is kind of weird/confusing.

That would be totally ok for me as long as we document it.

dnlup: I understand your point, @ronag , and you're right. I wouldn't use a `Pool` instance for performing requests that need redirections. I agree that it wouldn't make sense. But `get(origin)` might get used for other low-level operations that do not require a request to be fired. As of now, I see more harm than good in removing it for that reason.

> Which I think is kind of weird/confusing.

On one side, it is. On the other one, if we think in therm of low-level API, it's just a small piece that does exactly one thing, and you might need it at that level. The end-user would use the high-level API and should not be concerned about it. I think we can document it well enough for someone who wants to implement a custom `Agent`. I am sure we can come up with a better approach in the future if needed.

ronag:
@dnlup I had to comment out a few of your `Agent` tests that failed and I don't understand how they should/could work. Do you want to take a look or should I remove them?

lib/agent.js
ronag: I know some of my colleagues that use `Agent.get` directly instead of the helper methods (they want to use dispatch). I could easily see that they would get confused and start using `RedirectAgent.get` thinking it does something it does not. Most people are also not very meticulous with reading the docs.

That's my main concern. It's kind of easy do things wrong which for me is a sign of a bad API. But anyways, let's discuss this in a separate PR. It's not that big of a problem.

dnlup:
> @dnlup I had to comment out a few of your `Agent` tests that failed and I don't understand how they should/could work. Do you want to take a look or should I remove them?

I'll take a look at it, I think you are talking about the tests for `connect/disconnect` events right?

lib/agent-redirect.js
```diff
@@ -0,0 +1,209 @@
...
+    // By default it is Host only, unless it's a 303 (see below), which removes also all Content-* headers.
+    // https://tools.ietf.org/html/rfc7231#section-6.4
+    this.opts.headers = cleanRequestHeaders(this.opts.headers, statusCode === 303)
+    this.opts.path = path
+    this.opts.origin = origin
+
+    // https://tools.ietf.org/html/rfc7231#section-6.4.4
+    // In case of HTTP 303, always replace method to be either HEAD or GET
+    if (statusCode === 303) {
+      this.opts.method = 'GET'
```
ShogunPanda, line 77: Little nit here. If the original method was `HEAD`, it must not be changed to get. So the line before should check for `statusCode === 303 && this.opts.method !== 'GET'`

docs/api/Agent.md
```diff
@@ -2,18 +2,21 @@
...
+Arguments:
+
+* **factory** - Default: `(origin, opts) => new Pool(origin, opts)`
+* // TODO: document rest opts?
 
 Returns: `Agent`
 
 Returns a new Agent instance used for dispatching requests.
 
 ### `Agent.get(options)`
```
ShogunPanda, line 14: Shouldn't this be `Agent.get(origin)`?

test/agent.js
```diff
@@ -1,533 +1,486 @@
...
+        .catch(err => {
+          t.fail(err)
+        })
+    }
   })
+})
 
-  t.test('Agent connect/disconnect event(s)', t => {
-    t.plan(2)
+// t.test('remove disconnect listeners when destroyed', t => {
```
dnlup, line 161: @ronag In this test what fails is the check after calling `Agent.get(origin)`. The previous implementation attached the event listeners when calling `get`:

https://github.com/nodejs/undici/blob/main/lib/agent.js#L52

ronag: Not sure I follow... nothing has changed there in this PR?

lib/redirect-agent.js
ronag: We agreed we would do this in a follow up PR.

ronag:
@Ethan-Arrowood this changes the architecture a bit and and some new nice interfaces/traits. When you have time could we have a call and discuss? I think I need your help to sort out typescript and docs accordingly.

ronag:
Question here. When the redirect cannot be followed (i.e. due to request body or too many redirects). Should we error or just continue dispatch without redirect?

ronag:
@ShogunPanda Do we need anything else here? https://github.com/follow-redirects/follow-redirects/blob/main/index.js 

e.g. https://github.com/follow-redirects/follow-redirects/blob/main/index.js#L370-L372

ShogunPanda:
No, it seems we implemented everything. Dropping `Authorization` header is not in the spec and I'm not sure why they did it.
I already (and thus you) implemented that 303 logic.

About your question above: IMHO once we exhaust maxRedirections, just continue as redirect did not exist and return the last 3xx response to the client with no errors.

lib/agent.js
```diff
@@ -47,17 +47,26 @@ class Agent extends EventEmitter {
       throw new InvalidArgumentError('Origin must be a non-empty string.')
     }
 
-    let pool = this[kCache].get(origin)
+    return this[kFactory](origin)
```
mcollina, line 50: this should return the same instance that is being cached.

mcollina (changes requested):

ronag:
> I already (and thus you) implemented that 303 logic.

Ah, I have to make sure to add you as author to commits.

lib/agent.js
ronag: Good catch!

test/agent.js
ronag: Sorted. Thanks!

mcollina (approved):
lgtm, good work

ShogunPanda:
> > I already (and thus you) implemented that 303 logic.
> 
> Ah, I have to make sure to add you as author to commits.

Lol. Thanks! :)

lib/agent-redirect.js
```diff
@@ -0,0 +1,210 @@
...
+  onUpgrade (statusCode, headers, socket) {
+    this.handler.onUpgrade(statusCode, headers, socket)
+  }
+
+  onError (error) {
+    this.handler.onError(error)
+  }
+
+  onHeaders (statusCode, headers, resume) {
+    if ([300, 301, 302, 303, 307, 308].indexOf(statusCode) === -1) {
```
Ethan-Arrowood, line 52: In node core I've seen this implemented using a Set and a `has` look up. It might be trivially faster than an array scan; however, this list only has 6 items so its probably a negligible perf difference (wouldn't be surprised if V8 optimizes this already since the list is constant)

lib/agent-redirect.js
```diff
@@ -0,0 +1,210 @@
...
+    }
+
+    if (!agent || typeof agent.dispatch !== 'function') {
+      throw new InvalidArgumentError('invalid opts.agent')
+    }
+
+    this[kMaxRedirections] = maxRedirections
+    this[kAgent] = agent
+
+    this[kAgent].on('connect', (...args) => this.emit(...args))
```
Ethan-Arrowood, line 112: Do we need to do something like:

```suggestion
    this[kAgent].on('connect', (...args) => this.emit('connect', ...args))
```

lib/client-pool.js
```diff
@@ -154,6 +154,10 @@ class Pool extends EventEmitter {
 
   dispatch (opts, handler) {
     try {
+      if (opts.origin && opts.origin !== this.url.origin) {
+        throw new InvalidArgumentError('originl')
```
Ethan-Arrowood, line 158:
```suggestion
        throw new InvalidArgumentError('origin')
```
spelling maybe

ronag:
I found some problems:

- close/destroy will do so on the passed agent (which could also be the globalAgent).

Maybe better to just inherit `Agent` and override `dispatch`?

lib/core/client.js
```diff
@@ -243,6 +243,10 @@ class Client extends EventEmitter {
 
   dispatch (opts, handler) {
     try {
+      if (opts.origin && opts.origin !== this.url.origin) {
+        throw new InvalidArgumentError('originl')
```
Ethan-Arrowood, line 247:
```suggestion
        throw new InvalidArgumentError('origin')
```

spelling

Ethan-Arrowood (changes requested):

lib/agent-redirect.js
ronag: True.

docs/api/Agent.md
```diff
@@ -2,18 +2,53 @@
 
 ## `new undici.Agent(opts)`
 
-* opts `undici.Pool.options` - options passed through to Pool constructor
+Arguments:
+
+* **factory** - Default: `(origin, opts) => new Pool(origin, opts)`
+* // TODO: document rest opts?
```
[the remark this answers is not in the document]
ronag: Not sure. Can you make a PR suggestion?

ronag:
> I found a possible proble. close/destroy will do so on the passed agent (which could also be the globalAgent).
> 
> I'm a little unsure how to resolve this? @mcollina @dnlup @ShogunPanda.
> 
> Alternatives:
> 
> * Don't allow passing custom agent and always use global or a new Agent instance.
> * Remove close/destroy from agent API and make it instance specific?
> * Make close/destroy noops?

Resolved by reduce required Agent interface.

lib/agent-redirect.js
```diff
@@ -0,0 +1,190 @@
...
+    this[kMaxRedirections] = maxRedirections
+    this[kAgent] = agent
+  }
+
+  dispatch ({ maxRedirections = this[kMaxRedirections], ...opts } = {}, handler) {
+    if (maxRedirections != null && (maxRedirections < 0 || !Number.isInteger(maxRedirections))) {
+      throw new InvalidArgumentError('maxRedirections must be a positive number')
+    }
+
+    const { [kAgent]: agent = getGlobalAgent() } = this
```
ronag, line 115: What if globalAgent is also a redirect agent or self?

lib/agent-redirect.js
```diff
@@ -0,0 +1,199 @@
...
+    const { maxRedirections = this[kMaxRedirections] } = opts
+
+    if (maxRedirections != null && (maxRedirections < 0 || !Number.isInteger(maxRedirections))) {
+      throw new InvalidArgumentError('maxRedirections must be a positive number')
+    }
+
+    opts = { ...opts }
+
+    assert(opts[kRedirectTag] !== this[kRedirectTag], 'circular dispatch')
+    opts[kRedirectTag] = this[kRedirectTag]
```
ronag, line 122: @mcollina wdyt?

ShogunPanda: Can you clarify what is this for?

ronag: It is possible to create a loop if you pass a redirect agent into a redirect agent.

ShogunPanda: Oh, I see. Cool.

ShogunPanda:
@ronag I'm writing tests but I'm bit confused on how to use RedirectAgent with the new API.
Do you mind sharing a little (like 4 lines) snippet of an example of how to use it?

ronag:
```js
const agent = new RedirectAgent()
undici.request(url, { ...opts, agent })
```

ShogunPanda:
> ```js
> const agent = new RedirectAgent()
> undici.request(url, { ...opts, agent })
> ```

That's what I thought and tried. But the test failed on the circular dispatch assert.

ronag:
maybe a bug, i'll try fix

ronag:
can you try removing the assert?

ShogunPanda:
> can you try removing the assert?

Yes, that made the trick. Don't we need the assert at all? To be honest I don't full understand how the loop is created if you pass a redirect agent to a redirect agent. 

Are you talking about passing `RedirectAgent 1` to `RedirectAgent 2`? In my case `RedirectAgent 1` was passed to `RedirectAgent 1`. Which made the assert fail. Now, since we have maxRedirections, why do we care about loops at all?

ronag:
> Which made the assert fail. Now, since we have maxRedirections, why do we care about loops at all?

This is a clear user error where I think we should be helpful and fail.

ShogunPanda:
I see. So the assert is indeed needed.
And I also think I found the problem:

```
Symbol('tag') == Symbol('tag')
> false
```

ronag:
> I see. So the assert is indeed needed.
> And I also think I found the problem:
> 
> ```
> Symbol('tag') == Symbol('tag')
> > false
> ```

Not sure that is what is happening though.

```js
const tag =  Symbol('tag')
tag === tag
```

ShogunPanda:
Nevermind. It's not about symbols. 

It's about detecting that the tag is not set (first redirect or no RedirectAgent) or set to the same value as the previous one (which means recursion, not loop):

```
assert(!opts[kRedirectTag] || (opts[kRedirectTag] === this[kRedirectTag]), 'circular dispatch')
```

What do you think?

lib/agent-redirect.js
```diff
@@ -0,0 +1,199 @@
...
+  dispatch (opts, handler) {
+    const { maxRedirections = this[kMaxRedirections] } = opts
+
+    if (!Number.isInteger(maxRedirections) || maxRedirections < 0) {
+      throw new InvalidArgumentError('maxRedirections must be a positive number')
+    }
+
+    opts = { ...opts }
+
+    assert(opts[kRedirectTag] !== this[kRedirectTag], 'circular dispatch')
```
ronag, line 121:
```suggestion
  assert(!opts[kRedirectTag] || (opts[kRedirectTag] === this[kRedirectTag]), 'circular dispatch')
```

ShogunPanda:
@ronag 

I've added tests and send a PR to this branch: https://github.com/nodejs/undici/pull/614
All my existing tests cases were passing with no issues at all. Very good job!

The idea of skipping redirects when using Readable is very good. We just need to clearly document it.

Once that PR and this are merged, we have to discuss about how to add redirect history to callback data.

lib/agent.js
```diff
@@ -60,6 +60,14 @@ class Agent extends EventEmitter {
     return pool
   }
 
+  dispatch (opts, handler) {
+    if (!opts || typeof opts.origin !== 'string' || opts.origin === '') {
+      throw new InvalidArgumentError('Origin must be a non-empty string.')
+    }
```
mcollina, line 66: I would just check `if (typeof opts !== 'object')`

test/utils/stream.js
```diff
@@ -0,0 +1,24 @@
+const { Readable, Writable } = require('stream')
```
mcollina, line 1: use strict is missing

ronag: @mcollina can we add linting for this?

ronag:
Should we merge `RedirectAgent` into `Agent`? The functionality is kind of trivial and it would have to be explicitly enabled by passing the `maxRedirects` option in the constructor.

@mcollina @dnlup @ShogunPanda  wdyt?

ShogunPanda:
I agree. 
Since it's a very common use case when dealing with HTTP agents, it's better to keep things simple for developer experience (in this case, not having to remember to use a different agent for a comon case).
And since it's protected by maxRedirects I don't see performance problems.

Also, I have to agree since it was similar to my original proposal in https://github.com/nodejs/undici/issues/553#issuecomment-781994003. Just joking. Lol.

mcollina:
Good for me!

ronag:
Done!

ronag:
This is starting to feel pretty solid. Just waiting for doc + coverage. @Ethan-Arrowood @ShogunPanda.

lib/agent/redirect.js
```diff
@@ -0,0 +1,140 @@
...
+    if (statusCode === 303 && this.opts.method !== 'HEAD') {
+      this.opts.method = 'GET'
+      this.opts.body = null
+    }
+  }
+
+  onData (chunk) {
+    if (this.location) {
+      // Ignore body.
+      // TODO (fix): Spec ref?
```
ronag, line 87: @ShogunPanda do we have a spec ref in regards to what to do with body?

ShogunPanda: So, I've read the spec again. RFC 7231, Section 6.4 
It says that redirection is used to serve the requested resource from another URL, I think it assumes that no body is generated (and thus can be ignored). Even thought generating a body is not prohibited. There is an exception to this, the status 300.

For status 301, 302, 303, 307 and 308 (the latter from RFC 7238), the spec(s) mentioned that the body usually (which means it's optional and not mandated) contain just an hyperlink to the value of the Location response header, so probably we can just ignore it.

For status 300, which is "Multiple Choices", the spec mentions both generating a Location response header AND a response body with the other possible location to follow. The interesting thing is that the spec explicitily chooses not to specify a format for such body and leave it to servers and browsers implementors. Which means, in our case, that probably we can just ignore it. (we could, in theory, let the user specify a closure and let them deal with parsing, but I think it's overly complicated).

ronag: Can we add some notes in the code on what we do and why?

ShogunPanda: Done in #621

ShogunPanda:
@ronag @mcollina 

I've sent #621 which brings coverage back to 100%. 
It also fixes couple of minor stuffs. Please let me know what you think about it.

ShogunPanda:
Also, final note. Once this get merged, I'll start working on #615.

PR #606, opened 2021-03-22
refactor: cleanup and restructure

ronag:
- Move `client-*` to api folder as `api-*`.
- Rename `pool` to `client-pool`.
- Move `api-` references to index root only.

### DIFF
--- a/lib/api/index.js
+++ b/lib/api/index.js
@@ -1,3 +1,5 @@
+'use strict'
+
 module.exports.request = require('./api-request')
 module.exports.stream = require('./api-stream')
 module.exports.pipeline = require('./api-pipeline')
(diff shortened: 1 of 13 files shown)


ronag:
Extracts some of the changes from https://github.com/nodejs/undici/pull/603 which are not necessarily related to redirect.

lib/api/index.js
```diff
@@ -0,0 +1,5 @@
+module.exports.request = require('./api-request')
```
mcollina, line 1: use strict is missing

ronag: Hm. Shouldn't the linter catch that?

mcollina (approved):
lgtm

PR #611, opened 2021-03-23
Embed llhttp

dnlup:
Follows #575


ronag (approved):

dnlup:
I tried the GitHub UI for resolving the conflicts but now we have a merge commit, sorry

mcollina (changes requested):
The `deps` must not be included in the final released module. I have a feeling that would be too big. We should just ship the wasm.

dnlup:
> The `deps` must not be included in the final released module. I have a feeling that would be too big. We should just ship the wasm.

Yes definitely, I missed it, thanks. Should we merge #601 before and then rebase the changes here?

mcollina:
> Yes definitely, I missed it, thanks. Should we merge #601 before and then rebase the changes here?

I think so.

mcollina (approved):
lgtm

PR #619, opened 2021-03-24
fix: add connect timeout

ronag:
Haven't been able to build a test case for this but in production I have case where neither `'connect'` not `'error'` is invoked on a socket after `socket.connect()`.


lib/core/errors.js
```diff
@@ -8,6 +8,15 @@ class UndiciError extends Error {
...
 
+class ConnectTimeoutError extends UndiciError {
+  constructor (message) {
+    super(message)
+    Error.captureStackTrace(this, ConnectTimeoutError)
+    this.name = 'ConnectTimeoutError'
+    this.message = message || 'Connect Timeout Error'
+    this.code = 'UND_ERR_CONNECT_TIMEOUT'
+  }
+}
```
mcollina, line 19:
```suggestion
}

```

mcollina (commented):
I would really prefer if this had a test.

Does it worsen our benchmarks?

mcollina:
Should this have a default as it can happen in reality?

ronag:
> Should this have a default as it can happen in reality?

It hasa default. 10s.

ronag:
> Should this have a default as it can happen in reality?

Any idea on how to reproduce this scenario? When I try I always get an error.

ronag:
> Does it worsen our benchmarks?

No.

dnlup:
> > Should this have a default as it can happen in reality?
> 
> Any idea on how to reproduce this scenario? When I try I always get an error. Maybe if I monkeypatch the global socket prototype :/?

Can we tap into some `net.Server` hook and delay the `connect` event?

mcollina:
> Any idea on how to reproduce this scenario? When I try I always get an error. Maybe if I monkeypatch the global socket prototype :/?

Go for it in its own test file.

ronag:
Added test

Issue #610, opened 2021-03-23
Move undici global methods out of Agent docs


ronag:
They don't really have anything to do with Agent per say and it would be nice if this is the page that most users start with and don't have to worry about the other  details.

i.e. a page with:

```
# undici.request
# undici.stream
# undici.pipeline
# undici.upgrade
# undici.connect
```

is the only thing I think 90% of users need. Maybe should be the index page?

mcollina:
I agree.

Ethan-Arrowood:
Good idea!

ShogunPanda:
I can confirm it was mess for me to follow docs when starting using undici.

Ethan-Arrowood:
I noticed these got added to the main readme - closing this issue now

Issue #616, opened 2021-03-24
Do we need `Agent.get`?


ronag:
I think `Agent.get` is a bit weird now that we have `Agent.dispatch` and adds unnecessary API surface.

I know some of my colleagues use `Agent.get` directly instead of the helper methods (they want to use dispatch). I could easily see that they would get confused and start using `RedirectAgent.get` (assuming that they can just replace `Agent` with `RedirectAgent` in existing code) thinking it does something it does not. Most people are not very meticulous with reading the docs.

That's my main concern. It's kind of easy do things wrong which for me is a sign of a bad API. But anyways, let's discuss this in a separate PR. It's not that big of a problem.

Refs: https://github.com/nodejs/undici/pull/603#discussion_r598889560

ronag:
@mcollina @dnlup

dnlup:
In a way I think we do, otherwise, we would have to reimplement it each time we extend from `Agent`. I am pretty sure it would look the same in all custom Agents. It makes sense to me as a low-level API. But I also understand your concern, so maybe we should expose this to the end-user in a different way.

For example:
https://github.com/nodejs/undici/pull/603/files#diff-0c8b0ffe565a54bbb79a3b35000d152df16910fe193065de4e2f10f3ede8ec42R84

We would have to reimplement the [`get`](https://github.com/nodejs/undici/blob/main/lib/agent.js#L45) logic in `dispatch` there.

dnlup:
Maybe, when we extend/create a custom Agent,  we could let the user provide a `_dispatch` method (like `_write` in streams) which has a `Client`/`Pool` instance passed as an argument. This way, `get` could be internal, but it would part of the default prototype of the `Agent` so that we don't have to re-implement it each time we extend from it.

ronag:
I’m just concerned with the public API for end users.

mcollina:
The reason why we need `get()` is to provide a way for end users to reach into the `Pool` instance that is being used for a given host. From `Pool`, they should be able to reach for `Client`.

Those things are deeply needed for observability and introspection. Removing access is not the solution.
As a theme, we need more introspection.

---

Overall I do not see the problem you are mentioning @ronag. If they use `.get()` their system will crash or not redirect.
I also do not see a problem in having the `RedirectAgent` forward the calls to the encapsulated agent, e.g. `close()`  and others. For all intent and purposes a RedirectAgent should behave like an Agent minus the behavior in `dispatch()` that is overridden.

[...]

Issue #612, opened 2021-03-23
Agent connect/disconnect ambiguous


ronag:
The connect and disconnect  events are a little weird at the moment.

They both pass a "client" argument but it's unclear what exactly this is, e.g. in the case of `Agent` is it a `Pool` instance of `Client` instance? i.e. inner or outer? 

Also makes it a little hard to reason about. 

I think the current behavior is not correct since it passes a client instance instead of pool.

ronag:
@nodejs/undici

mcollina:
> I think the current behavior is not correct since it passes a client instance instead of pool.

The behavior is somewhat correct because in `Agent.prototype.get()` we can return either a Client or a Pool depending on `connections`. The signature of the event is the same because they try to be identical.

---

These events are useful for debugging and diagnosing what's happening. The do not have a real use case in mind so it's a bit hard to refactor them.

ronag:
> The behavior is somewhat correct because in Agent.prototype.get() we can return either a Client or a Pool depending on connections. The signature of the event is the same because they try to be identical.

But in the current case the connect/disconnect events will always emit the inner most client, i.e. not the Pool.

dnlup:
I think the problem is that the `Pool` is just forwarding `Clients` events:

* https://github.com/nodejs/undici/blob/c604e172fda9f11d1908e0cd6ebfad43c889f3e8/lib/client-pool.js#L82

Maybe we should emit a `disconnect` event only when all clients have disconnected? Or use a custom event?

ronag:
> I think the problem is that the Pool is just forwarding Clients events:

Whether it's a problem or not depends a little on how we expect it to work. Which is what I'm unsure about.

dnlup:
> > I think the problem is that the Pool is just forwarding Clients events:
> 
> Whether it's a problem or not depends a little on how we expect it to work. Which is what I'm unsure about.

How are you using this event now? Why is it confusing for you? Do you have do access the `Pool` instance when this event is emitted?

ronag:
> How are you using this event now? Why is it confusing for you? Do you have do access the Pool instance when this event is emitted?

I'm using it to log connections/disconnections (per socket/Client). However, @mcollina seems to expect it to emit `Pool` instances.

ronag:
If it's to ambiguous/confusing I would even prefer removing it.

mcollina:
[...]

PR #620, opened 2021-03-25
refactor: Unify Client, Pool & Agent

ronag:
- Unifies `Client`, `Pool` and `Agent` under a single interface.
- Removes `Agent.get` (in favor of extra args to client connect/disconnect).
- Removes `Client/Pool.busy` in favor of return value on `dispatch`. (I'm not 100% sure this is possible, but I think so).
- Makes `connect/disconnect` emit an array of entire connect/disconnect chain for introspection.
- Makes `'drain'` emit origin as first argument.
- Makes `Agent` a client with all the api methods (request, stream etc...) on the prototype and same events as `Client`.

So `Agent`, `Pool` and `Client` all implement the same interface and could be used interchangeably. 

Fixes:

https://github.com/nodejs/undici/issues/610
https://github.com/nodejs/undici/issues/616
https://github.com/nodejs/undici/issues/612

### DIFF
--- a/README.md
+++ b/README.md
@@ -55,6 +55,110 @@ for await (const data of body) {
 console.log('trailers', trailers)
 ```
 
+## `undici.request(url[, options]): Promise`
+
+Arguments:
+
+* **url** `string | URL | object`
+* **options** [`RequestOptions`]
+  * **dispatcher** `Dispatcher` - Default: [getGlobalDispatcher]
+  * **method** `String` - Default: `GET`
+* **maxRedirections** `Integer` - Default: `0`
+
+Returns a promise with the result of the `Dispatcher.request` method.
+
+`url` may contain pathname. `options` may not contain path.
+
+Calls `options.dispatcher.request(options)`.
+
+See [Dispatcher.request] for more details.
+
+## `undici.stream(url, options, factory): Promise`
+
+Arguments:
+
+* **url** `string | URL | object`
+* **options** [`StreamOptions`]
+  * **dispatcher** `Dispatcher` - Default: [getGlobalDispatcher]
+  * **method** `String` - Default: `GET`
+* **factory** `Dispatcher.stream.factory`
+
+Returns a promise with the result of the `Dispatcher.stream` method.
+
+`url` may contain pathname. `options` may not contain path.
+
+Calls `options.dispatcher.stream(options, factory)`.
+
+See [Dispatcher.stream](docs/api/Dispatcher.md#dispatcherstream) for more details.
+
+## `undici.pipeline(url, options, handler): Duplex`
+
+Arguments:
+
+* **url** `string | URL | object`
+* **options** [`PipelineOptions`]
+  * **dispatcher** `Dispatcher` - Default: [getGlobalDispatcher]
+  * **method** `String` - Default: `GET`
+* **handler** `Dispatcher.pipeline.handler`
+
+Returns: `stream.Duplex`
+
+`url` may contain pathname. `options` may not contain path.
+
+Calls `options.dispatch.pipeline(options, handler)`.
+
+See [Dispatcher.pipeline](docs/api/Dispatcher.md#dispatcherpipeline) for more details.
+
+### `undici.connect(options[, callback])`
+
+Starts two-way communications with the requested resource using [HTTP CONNECT](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/CONNECT).
+
+Arguments:
+
+* **options** [`ConnectOptions`]
+  * **dispatcher** `Dispatcher` - Default: [getGlobalDispatcher]
+  * **method** `String` - Default: `GET`
+* **callback** `(err: Error | null, data: ConnectData | null) => void` (optional)
+
+Returns a promise with the result of the `Dispatcher.connect` method.
+
+`url` may contain pathname. `options` may not contain path.
+
+Calls `options.dispatch.connect(options)`.
+
+See [Dispatcher.connect](docs/api/Dispatcher.md#dispatcherconnect) for more details.
+
+### `undici.upgrade(options[, callback])`
+
+Upgrade to a different protocol. See [MDN - HTTP - Protocol upgrade mechanism](https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism) for more details.
+
+Arguments:
+
+* **options** [`UpgradeOptions`]
+  * **dispatcher** `Dispatcher` - Default: [getGlobalDispatcher]
+  * **method** `String` - Default: `GET`
+* **callback** `(error: Error | null, data: UpgradeData) => void` (optional)
+
+Returns a promise with the result of the `Dispatcher.upgrade` method.
+
+`url` may contain pathname. `options` may not contain path.
+
+Calls `options.dispatcher.upgrade(options)`.
+
+See [Dispatcher.upgrade](docs/api/Dispatcher.md#clientpipelining) for more details.
+
+## `undici.setGlobalDispatcher(dispatcher)`
+
+* dispatcher `Dispatcher`
+
+Sets the global dispatcher used by global API methods.
+
+## `undici.getGlobalDispatcher()`
+
+Gets the global dispatcher used by global API methods.
+
+Returns: `Dispatcher`
+
 ## Specification Compliance
 
 This section documents parts of the HTTP/1.1 specification which Undici does
--- a/docs/api/Agent.md
+++ b/docs/api/Agent.md
@@ -1,119 +1,95 @@
 # Agent
 
-## `new undici.Agent(opts)`
+Extends: `undici.Dispatcher`
+
+Agent allow dispatching requests against multiple different origins.
+
+Requests are not guaranteed to be dispatched in order of invocation.
+
+## `new undici.Agent([options])`
 
 Arguments:
 
-* **factory** - Default: `(origin, opts) => new Pool(origin, opts)`
-* // TODO: document rest opts?
+* **options** `AgentOptions` (optional)
 
 Returns: `Agent`
 
-Returns a new Agent instance used for dispatching requests.
+### Parameter: `AgentOptions`
 
-### `Agent.get(origin)`
+Extends: [`ClientOptions`](docs/api/Pool.md#parameter-pooloptions)
 
-* origin `string` - A origin to be retrieved from the Agent.
+* **factory** `(origin: URL, opts: Object) => Dispatcher` - Default: `(origin, opts) => new Pool(origin, opts)` 
+* **maxRedirections** `Integer` - Default: `0`.
 
-This method retrieves Client instances from the Agent. If the client does not exist it is automatically added by calling
-the `factory` method passed through the `Agent` constructor.
+## Instance Properties
+
+### `Agent.closed`
+
+Implements [Client.closed](docs/api/Client.md#clientclosed)
+
+### `Agent.connected`
+
+Implements [Client.connected](docs/api/Client.md#clientconnected)
+
+### `Agent.destroyed`
+
+Implements [Client.destroyed](docs/api/Client.md#clientdestroyed)
+
+### `Agent.pending`
+
+Implements [Client.pending](docs/api/Client.md#clientpending)
+
+### `Agent.running`
+
+Implements [Client.running](docs/api/Client.md#clientrunning)
+
+### `Agent.size`
+
+Implements [Client.size](docs/api/Client.md#clientsize)
+
+## Instance Methods
+
+### `Agent.close([callback])`
+
+Implements [`Dispatcher.close([callback])`](docs/api/Dispatcher.md#clientclose-callback-).
+
+### `Agent.destroy([error, callback])`
+
+Implements [`Dispatcher.destroy([error, callback])`](docs/api/Dispatcher.md#dispatcher-callback-).
+
+### `Agent.dispatch(options, handlers: AgentDispatchOptions)`
+
+Implements [`Dispatcher.dispatch(options, handlers)`](docs/api/Dispatcher.md#clientdispatchoptions-handlers).
+
+#### Parameter: `AgentDispatchOptions`
+
+Extends: [`DispatchOptions``](docs/api/Dispatcher.md#parameter-dispatchoptions)
+
+* **origin** `string | URL`
+* **maxRedirections** `Integer`.
+
+Implements [`Dispatcher.destroy([error, callback])`](docs/api/Dispatcher.md#dispatcher-callback-).
+
+### `Agent.connect(options[, callback])`
+
+See [`Dispatcher.connect(options[, callback])`](docs/api/Dispatcher.md#clientconnectoptions--callback).
 
 ### `Agent.dispatch(options, handlers)`
 
-Dispatches a request.
+Implements [`Dispatcher.dispatch(options, handlers)`](docs/api/Dispatcher.md#clientdispatchoptions-handlers).
 
-This API is expected to evolve through semver-major versions and is less stable than the preceding higher level APIs. It is primarily intended for library developers who implement higher level APIs on top of this.
+### `Agent.pipeline(options, handler)`
 
-Arguments:
+See [`Dispatcher.pipeline(options, handler)`](docs/api/Dispatcher.md#clientpipelineoptions-handler).
 
-* **options** `DispatchOptions`
-* **handlers** `DispatchHandlers`
+### `Agent.request(options[, callback])`
 
-Returns: `void`
+See [`Dispatcher.request(options [, callback])`](docs/api/Dispatcher.md#clientrequestoptions--callback).
 
-#### Parameter: `DispatchOptions`
+### `Agent.stream(options, factory[, callback])`
 
-* **origin** `string | URL`
-* **path** `string`
-* **method** `string`
-* **body** `string | Buffer | Uint8Array | stream.Readable | null` (optional) - Default: `null`
-* **headers** `UndiciHeaders` (optional) - Default: `null`
-* **idempotent** `boolean` (optional) - Default: `true` if `method` is `'HEAD'` or `'GET'` - Whether the requests can be safely retried or not. If `false` the request won't be sent until all preceeding requests in the pipeline has completed.
-* **upgrade** `string | null` (optional) - Default: `method === 'CONNECT' || null` - Upgrade the request. Should be used to specify the kind of upgrade i.e. `'Websocket'`.
+See [`Dispatcher.stream(options, factory[, callback])`](docs/api/Dispatcher.md#clientstreamoptions-factory--callback).
 
-#### Parameter: `DispatchHandlers`
+### `Agent.upgrade(options[, callback])`
 
-* **onConnect** `(abort: () => void) => void` - Invoked before request is dispatched on socket. May be invoked multiple times when a request is retried when the request at the head of the pipeline fails.
-* **onError** `(error: Error) => void` - Invoked when an error has occurred.
-* **onUpgrade** `(statusCode: number, headers: string[] | null, socket: Duplex) => void` (optional) - Invoked when request is upgraded. Required if `DispatchOptions.upgrade` is defined or `DispatchOptions.method === 'CONNECT'`.
-* **onHeaders** `(statusCode: number, headers: string[] | null, resume: () => void) => boolean` - Invoked when statusCode and headers have been received. May be invoked multiple times due to 1xx informational headers. Not required for `upgrade` requests.
-* **onData** `(chunk: Buffer) => boolean` - Invoked when response payload data is received. Not required for `upgrade` requests.
-* **onComplete** `(trailers: string[] | null) => void` - Invoked when response payload and trailers have been received and the request has completed. Not required for `upgrade` requests.
-
-## `agent.close(): Promise`
-
-Returns a `Promise.all` operation closing all of the pool instances in the Agent instance. This calls `pool.close` under the hood.
-
-## `agent.destroy(): Promise`
-
-Returns a `Promise.all` operation destroying all of the pool instances in the Agent instance. This calls `pool.destroy` under the hood.
-
-## `undici.setGlobalAgent(agent)`
-
-* agent `Agent`
-
-Sets the global agent used by `request`, `pipeline`, and `stream` methods.
-The default global agent creates `undici.Pool`s with no max number of
-connections.
-
-The agent must only **implement** the `Agent` API; not necessary extend from it.
-
-## `undici.getGlobalAgent(agent)`
-
-TODO: document
-
-## `undici.request(url[, opts]): Promise`
-
-* url `string | URL | object`
-* opts `{ agent: Agent } & client.request.opts`
-* // TODO: document maxRedirections?
-
-`url` may contain path. `opts` may not contain path. `opts.method` is `GET` by default.
-Calls `pool.request(opts)` on the pool returned from either the globalAgent (see [setGlobalAgent](#undicisetglobalagentagent)) or the agent passed to the `opts` argument.
-
-Returns a promise with the result of the `request` method.
-
-## `undici.stream(url, opts, factory): Promise`
-
-* url `string | URL | object`
-* opts `{ agent: Agent } & client.stream.opts`
-* factory `client.stream.factory`
-* // TODO: document maxRedirections?
-
-`url` may contain path. `opts` may not contain path.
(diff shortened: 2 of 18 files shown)


ronag:
@mcollina this removes `Agent.get` and instead adds `Agent.clients` and `Pool.clients` for introspection.

ronag:
Another alternative to `Agent.clients` or `Agent.get` would be that the `connect/disconnect` events emit an array of clients (according to depth). That is even more generic IMHO.

ronag:
This needs more work but I'd like to have some initial feedback before spending more time.

ronag:
@ShogunPanda

ronag:
The Client interface basically becomes:

```js
interface Client implements EventEmitter {
  connected: Integer
  running: Integer
  size: Integer
  pending: Integer,
  closed: Boolean
  destroyed: Boolean
  dispatch(opts, handler): Boolean
  close(callback)
  destroy(err, callback)
}
```

lib/agent.js
```diff
@@ -24,53 +25,99 @@ class Agent extends EventEmitter {
...
+    if (typeof opts.origin !== 'string' || opts.origin === '') {
       throw new InvalidArgumentError('Origin must be a non-empty string.')
     }
 
-    let pool = this[kCache].get(origin)
+    let pool = this[kClients].get(opts.origin)
 
     if (!pool) {
-      pool = this[kFactory](origin)
+      pool = this[kFactory](opts.origin)
```
dnlup, line 97: Not sure if this comment it's too premature, If we were building another Agent extending from this one, we would have to always add this manual bootstrap logic, which I think would be always the same for every custom Agent. That is what `get` provided, can we centralize it in some other `Agent.prototype` method?

mcollina:
Huge +1 to this.

lib/pool.js
```diff
@@ -72,14 +77,14 @@ class Pool extends EventEmitter {
       }
     }
 
-    this[kOnConnect] = function onConnect (client) {
+    this[kOnConnect] = function onConnect (origin, clients) {
       pool[kConnected]++
-      pool.emit('connect', client)
+      pool.emit('connect', origin, [...clients, pool])
```
dnlup, line 82: Does this array always contain one `Client` and the `Pool`?

lib/agent.js
```diff
@@ -24,53 +25,99 @@ class Agent extends EventEmitter {
...
+    let ret = 0
+    for (const { connected } of this[kClients].values()) {
+      ret += connected
     }
+    return ret
   }
 
-  get (origin) {
-    if (typeof origin !== 'string' || origin === '') {
+  get size () {
```
dnlup, line 58: Ok, so the stats needed are directly exposed from the Agent, am I getting this right?

lib/pool.js
ronag: Right now yes. But I envision that we could add a factory argument to the pool (similar to what we do for agent) and then it wouldn't be so sure.

lib/agent.js
ronag: Reimplementing what agent does is rather trivial though?

lib/agent.js
ronag: Not sure I understand the question?

lib/agent.js
ronag: I would be happy to add some form of "protected" get later if the need ever arises. I'm a little skeptical whether it's actually necessary in practice.

ShogunPanda:
+1 for me as well. I'd just wait for #603 (with my improved coverage which is coming up) to land so we can rely on tests.

ronag:
@jonnydgreen FYI, WIP.

jonnydgreen:
Looks awesome, thanks for keeping me in the loop :)

lib/agent.js
dnlup: Regarding the fact that `get` allowed to gather stats about a `Client`/`Pool` (https://github.com/nodejs/undici/issues/616#issuecomment-806504163), this new strategy exposes the stats directly on the `Agent`.

ronag:
Some tests are still failing and some minor API TODOs, but I think this is somewhat ready for review.

ronag:
Some thoughts:

- Should we remove/hide not strictly necessary API`s? I would like to do this to avoid users making incorrect assumptions on what methods and properties are available. This is relevant for the following properties: `url`, `busy`, `pipelining`.
- What should the signature for connect/disconnect be?
  - `connect`(origin:URL, clients:Array) + `disconnect` (origin:URL, clients:Array, err:Error)
  -  `connect`(origin:URL, ...clients) + `disconnect` (origin:URL, err: Error, ...clients:Array)
  
Some remaining tasks:

- Fix remaining tests.
- Update docs.

ronag:
@szmarczak @ShogunPanda

ronag:
I need some help with the ts definitions. @Ethan-Arrowood would you be ok with having a look at those after this has been merged?

ronag:
@dnlup If you have time to help then please have a look at the `'remove disconnect listeners when destroyed'` test.

szmarczak:
> What should the signature for connect/disconnect be?

I prefer the first one. 

> Is Agent confusing now?

I'll take a guess now. Client points to a single socket, Pool manages sockets (Client instances) within the same server name, Agent manages Pools (various server names).

szmarczak:
> Should we remove/hide not strictly necessary APIs? I would like to do this to avoid users making incorrect assumptions on what methods and properties are available.

:+1:

ronag:
> I'll take a guess now. Client points to a single socket, Pool manages sockets (Client instances) within the same server name, Agent manages Pools (various server names).

Correct!

dnlup:
> @dnlup If you have time to help then please have a look at the `'remove disconnect listeners when destroyed'` test.

Ok @ronag , I'll take a look monday

ronag:
The `clients` array in `connect`/`disconnect`. What order should the clients be? Inner first or outer first?

mcollina:
> The `clients` array in `connect`/`disconnect`. What order should the clients be? Inner first or outer first?

whatever is faster, it should not matter

lib/agent.js
```diff
@@ -0,0 +1,212 @@
...
+
+    this[kOnConnect] = function onConnect (origin, clients) {
+      agent.emit('connect', origin, [...clients, agent])
+    }
+
+    this[kOnDrain] = function onDrain (origin, clients) {
+      agent.emit('drain', origin, [...clients, agent])
+    }
+
+    this[kOnDisconnect] = function onDestroy (origin, clients, err) {
```
ronag, line 78:
```suggestion
    this[kOnDisconnect] = function onDiaconnect (origin, clients, err) {
```

test/agent.js
```diff
@@ -172,18 +166,18 @@ test('remove disconnect listeners when destroyed', t => {
...
-    t.tearDown(agent.close.bind(agent))
+    t.tearDown(dispatcher.close.bind(dispatcher))
 
-    const pool = agent.get(origin)
+    const pool = dispatcher.get(origin)
     t.true(pool.listeners('disconnect').length === 1)
 
-    agent.on('disconnect', () => {
+    dispatcher.on('disconnect', () => {
       t.true(pool.listeners('disconnect').length === 0)
```
dnlup, line 177: This test should be refactored, but I think it doesn't make sense to keep it since we returned to the `FinalizationRegistry` implementation. Or at least the `Agent` should expose some internal hook to allow testing that when the garbage collection triggers. The check on the `Pool` listeners on `disconnect` will always be `false` with this logic.

ronag: Let's remove it then.

ronag:
@mcollina: all tests should be ok now, ready for review

README.md
```diff
@@ -55,6 +55,110 @@ for await (const data of body) {
...
+
+`url` may contain pathname. `options` may not contain path.
+
+Calls `options.dispatch.connect(options)`.
+
+See [Dispatcher.connect](docs/api/Dispatcher.md#dispatcherconnect) for more details.
+
+### `undici.upgrade(options[, callback])`
+
+Upgrade to a different protocol. Visit [MDN - HTTP - Protocol upgrade mechanism](https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism) for more details.
```
ronag, line 133:
```suggestion
Upgrade to a different protocol. See [MDN - HTTP - Protocol upgrade mechanism](https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism) for more details.
```

ronag:
Would appreciate review from @szmarczak and @ShogunPanda as well.

dnlup (approved):
I like these changes a lot! The only thing that is not 100% clear to me is the behavior on Node 12, which doesn't support `FinalizationRegistry`. See my comment above. Beside that, this is great work 💪🏻

lib/agent.js
```diff
@@ -0,0 +1,212 @@
...
+      const ref = this[kClients].get(opts.origin)
+
+      let client = ref ? ref.deref() : null
+      if (!client) {
+        client = this[kFactory](opts.origin)
+          .on('connect', this[kOnConnect])
+          .on('disconnect', this[kOnDisconnect])
+          .on('drain', this[kOnDrain])
+
+        this[kClients].set(opts.origin, new WeakRef(client))
```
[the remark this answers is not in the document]
ronag: We don’t cleanup. It’s a minor mem leak. I personally think it’s fine but would like some more opinions.

README.md
```diff
@@ -55,6 +55,110 @@ for await (const data of body) {
...
+
+* **url** `string | URL | object`
+* **options** [`PipelineOptions`]
+  * **dispatcher** `Dispatcher` - Default: [getGlobalDispatcher]
+  * **method** `String` - Default: `GET`
+* **handler** `Dispatcher.pipeline.handler`
+
+Returns: `stream.Duplex`
+
+`url` may contain pathname. `options` may not contain path.
```
szmarczak, line 106: Is `url` always normalized into a URL instance?

szmarczak: `may` or `must`?

docs/api/Agent.md
```diff
@@ -1,119 +1,95 @@
 # Agent
 
-## `new undici.Agent(opts)`
+Extends: `undici.Dispatcher`
```
szmarczak, line 3: What's the difference between a Dispatcher and a Pool? Is Dispatcher some sort of an abstraction?

ronag: Dispatcher is the base interface client, pool and agent implement.

ronag: May contain pathname

index.js
```diff
@@ -1,45 +1,58 @@
...
 module.exports.errors = errors
 
-module.exports.Agent = Agent
-module.exports.setGlobalAgent = setGlobalAgent
-module.exports.getGlobalAgent = getGlobalAgent
+let globalClient = new Agent()
 
-function dispatchFromAgent (fn) {
-  return (url, { agent = getGlobalAgent(), method = 'GET', ...opts } = {}, ...additionalArgs) => {
+function setGlobalDispatcher (agent) {
```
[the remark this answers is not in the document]
ronag: Corresponds to global agent in node core

README.md
ronag: May

szmarczak: > Is url always normalized into a URL instance?

Awesome. What about calling `decodeURI(string)` in order to make sure things like `'https://example.com/%AA'` will throw? This might have a performance impact.

index.js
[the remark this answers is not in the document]
szmarczak: So shouldn't this be named `setGlobalAgent` instead?

lib/pool.js
```diff
@@ -82,14 +92,12 @@ class Pool extends EventEmitter {
       }
     }
 
-    this[kOnConnect] = function onConnect (client) {
-      pool[kConnected]++
-      pool.emit('connect', client)
+    this[kOnConnect] = function onConnect (origin, targets) {
```
szmarczak, line 95: Wouldn't it be better to pass `onConnect` as an option? That way there would be no need for an `EventEmitter`.

index.js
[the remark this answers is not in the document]
szmarczak: I think I get it - it doesn't have to be an Agent, right?

szmarczak (approved):

README.md
ronag: Im not sure. @mcollina @dnlup wdyt?

ronag: @szmarczak I think that's a separate issue though. Want to open a PR or issue for that?

ShogunPanda:
+100 from me. I really like the new API! Go for it!

lib/agent.js
```diff
@@ -0,0 +1,215 @@
...
+    this.value = value
+  }
+
+  deref () {
+    return this.value
+  }
+}
+const FinalizationRegistry = global.FinalizationRegistry || class DummyFinalizer {
+  register () {}
+}
```
mcollina, line 43: Can you please add a comment explaining why these are ok and do not leak memory?

lib/agent.js
```diff
@@ -0,0 +1,215 @@
...
+      agent.emit('drain', origin, [agent, ...targets])
+    }
+
+    this[kOnConnect] = function onConnect (origin, targets) {
+      agent.emit('connect', origin, [agent, ...targets])
+    }
+
+    this[kOnDisconnect] = function onDisconnect (origin, targets, err) {
+      agent.emit('disconnect', origin, [agent, ...targets], err)
+    }
```
mcollina, line 82: can we just use arrow functions?

lib/agent.js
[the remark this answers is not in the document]
mcollina: > We don’t cleanup. It’s a minor mem leak. I personally think it’s fine but would like some more opinions.

Is this what is described in #631? It should only be a temporary leak from my understanding of it.

test/agent.js
```diff
@@ -158,73 +152,6 @@ test('multiple connections', t => {
...
-
-  const origin = `http://localhost:${server.address().port}`
-  const agent = new Agent({ connections: 1 })
-
-  t.tearDown(agent.close.bind(agent))
-
-  const pool = agent.get(origin)
-  t.true(pool instanceof Client)
-})
-
```
mcollina, line 227: why were this removed?

lib/agent.js
```diff
@@ -0,0 +1,215 @@
...
+
+      if (this[kClosed]) {
+        throw new ClientClosedError()
+      }
+
+      const ref = this[kClients].get(opts.origin)
+
+      let client = ref ? ref.deref() : null
+      if (!client) {
+        client = this[kFactory](opts.origin, this[kOptions])
```
mcollina, line 143: is this a pool?

lib/dispatcher.js
```diff
@@ -0,0 +1,19 @@
...
+  }
+
+  close () {
+    throw new Error('not implemented')
+  }
+
+  destroy () {
+    throw new Error('not implemented')
+  }
+}
```
mcollina, line 17: Why is this needed? Those are just stubs.

test/agent.js
```diff
@@ -50,27 +46,26 @@ test('agent should close internal pools', t => {
...
       .catch(err => {
         t.fail(err)
       })
 
-    const pool = agent.get(origin)
-    pool.once('connect', () => {
-      agent.close().then(() => {
-        request(origin, { agent })
+    dispatcher.once('connect', () => {
+      dispatcher.close().then(() => {
```
mcollina, line 62: This then is breaking the promise chain. Keep a consistent promise chain.

test/agent.js
```diff
@@ -90,22 +85,21 @@ test('agent should destroy internal pools', t => {
...
       .catch(err => {
         t.ok(err instanceof errors.ClientDestroyedError)
       })
 
-    const pool = agent.get(origin)
-    pool.once('connect', () => {
-      agent.destroy().then(() => {
-        request(origin, { agent })
+    dispatcher.once('connect', () => {
+      dispatcher.destroy().then(() => {
```
mcollina, line 101: This then is breaking the promise chain. Keep a consistent promise chain.

mcollina (changes requested):
Are benchmarks impacted by this change?

test/agent.js
ronag: We don't have `agent.get` anymore.

lib/agent.js
ronag: Can be.

lib/dispatcher.js
ronag: not really needed

lib/agent.js
mcollina: Can we call the variable dispatcher? so it's clear it's not a client.

lib/agent.js
ronag: It does leak a little memory.  See, https://github.com/nodejs/undici/pull/620#discussion_r603319160.

If that's unacceptable I will have to try and find more complicate solutions.

mcollina:
@jonnydgreen do you think this API would be ok for you to support in #587

lib/dispatcher.js
mcollina: Let's keep it, it would help TS and friends.

lib/agent.js
ronag: Done

lib/agent.js
ronag: Done

lib/agent.js
ronag: Basically it will keep a few object refs for every origin it sees. Which shouldn't be that many in practice.

jonnydgreen:
@mcollina looking at the code, I don't see any major problems. If anything, I think it will make everything a lot cleaner to implement and opens up a lot of additional functionality in the mocking support as I can use `setGlobalDispatcher` for `MockClient` and `MockPool` as well :)

ronag:
> Are benchmarks impacted by this change?

We have the benchmark CI 😄 .
branch:

```
[1] http - no agent  x 653 ops/sec ±2.08% (78 runs sampled)
[1] http - keepalive x 746 ops/sec ±1.86% (72 runs sampled)
[1] http - keepalive - multiple sockets x 4,953 ops/sec ±2.91% (76 runs sampled)
[1] undici - pipeline x 5,311 ops/sec ±4.38% (74 runs sampled)
[1] undici - request x 6,034 ops/sec ±2.48% (75 runs sampled)
[1] undici - pool - request - multiple sockets x 6,324 ops/sec ±2.70% (78 runs sampled)
[1] undici - stream x 6,667 ops/sec ±2.34% (80 runs sampled)
[1] undici - dispatch x 6,732 ops/sec ±2.31% (78 runs sampled)
[1] undici - noop x 6,367 ops/sec ±1.61% (76 runs sampled)
[1] node -e 'setTimeout(() => {}, 1000)' && node benchmarks exited with code 0
--> Sending SIGTERM to other processes..
[0] node benchmarks/server.js exited with code SIGTERM
```

master:

```
[1] http - no agent  x 679 ops/sec ±1.50% (79 runs sampled)
[1] http - keepalive x 727 ops/sec ±1.34% (73 runs sampled)
[1] http - keepalive - multiple sockets x 4,236 ops/sec ±3.03% (80 runs sampled)
[1] undici - pipeline x 5,056 ops/sec ±3.70% (81 runs sampled)
[1] undici - request x 5,846 ops/sec ±1.78% (80 runs sampled)
[1] undici - pool - request - multiple sockets x 5,882 ops/sec ±1.39% (82 runs sampled)
[1] undici - stream x 6,087 ops/sec ±1.69% (79 runs sampled)
[1] undici - dispatch x 6,266 ops/sec ±1.48% (81 runs sampled)
[1] undici - noop x 6,459 ops/sec ±0.91% (81 runs sampled)
[1] node -e 'setTimeout(() => {}, 1000)' && node benchmarks exited with code 0
```

ronag:
I will run more in depth benchmarks on the same EPYC machine the README results come from before release.

lib/agent.js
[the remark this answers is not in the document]
ronag: This version is a "permanent" leak. It becomes a problem if the user is going to work against thousands of different origins. Which I think is unlikely.

lib/dispatcher.js
Ethan-Arrowood: Since this is only at the type definition level then I think we could even get away with something like this in the `.d.ts` files:

```ts
interface Dispatcher extends EventEmitter {
  dispatch: () => void;
  close: () => void;
  destroy: () => void;
}

class Client extends Dispatcher {}
class Agent extends Dispatcher {}
```

Ethan-Arrowood: but then again, I do enjoy the abstract class pattern to this. I'm +1 for keeping it since it doesn't have any perf impact

lib/pool.js
```diff
@@ -99,29 +105,22 @@ class Pool extends EventEmitter {
...
     return this[kUrl]
   }
 
   get connected () {
-    return this[kConnected]
+    let ret = 0
+    for (const { connected } of this[kClients]) {
+      ret += connected
+    }
+    return ret
```
Ethan-Arrowood, line 118: this could be simplified to `this[kClients].some(({ connected }) => !!connected)`

Ethan-Arrowood (approved):
this looks great. really happy to see the core API refactor for easier client instantiations

lib/pool.js
ronag: connected is an integer

Ethan-Arrowood: Oh yes; sorry I keep thinking this property is a boolean thing when its a number _of_ connected clients.

mcollina:
What would entail in solving the memory leak? Can't we do the connect/disconnect solution that we used in main?

ronag:
> What would entail in solving the memory leak? Can't we do the connect/disconnect solution that we used in main?

I don’t like/trust that solution but I can look into it again for node 12.

dnlup:
> > What would entail in solving the memory leak? Can't we do the connect/disconnect solution that we used in main?
> 
> I don’t like/trust that solution but I can look into it again for node 12.

Not sure if feasible, but we could keep the `FinlizationRegistry` interface implemented here, and use the `connect`/`disconnect` events to trigger the `kFinalizer`.

mcollina:
> > > What would entail in solving the memory leak? Can't we do the connect/disconnect solution that we used in main?
> > 
> > 
> > I don’t like/trust that solution but I can look into it again for node 12.
> 
> Not sure if feasible, but we could keep the `FinlizationRegistry` interface implemented here, and use the `connect`/`disconnect` events to trigger the `kFinalizer`.

Only on Node v12. It's definitely better than leaking.

ronag:
@mcollina @dnlup fixed, PTAL

@dnlup your idea was very elegant, I like it!

dnlup (approved):

lib/agent.js
```diff
@@ -0,0 +1,227 @@
...
+    : new Pool(origin, opts)
+}
+
+const WeakRef = global.WeakRef || class CompatWeakRef {
+  constructor (value) {
+    this.value = value
+  }
+
+  deref () {
+    return this.value.connected === 0 && this.value.size === 0
```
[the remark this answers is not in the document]
ronag: I think you are missing https://github.com/nodejs/undici/pull/620/files#diff-d8bb2c7932a22406705bd0c42d2cb3d720574c245ba2cc81993a7d8d563b0e3eR50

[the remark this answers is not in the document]
ronag: I don't see a problem here?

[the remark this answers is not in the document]
dnlup: I was thinking that the finalizer is called when `connected` and `size` are `0`. In the finalized then the `WeakRef` in unrefed, and this operation will return `undefined` because it performs the same check, resulting in not removing the `dispatcher` from  the map.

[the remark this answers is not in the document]
dnlup: I am referring to this `if` branch:

https://github.com/nodejs/undici/pull/620/files#diff-d8bb2c7932a22406705bd0c42d2cb3d720574c245ba2cc81993a7d8d563b0e3eR76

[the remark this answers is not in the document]
ronag: >will return undefined because it performs the same check, resulting in not removing the dispatcher from the map.

But it does remove it when and only when it is undefined.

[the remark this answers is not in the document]
dnlup: 🤦 Sorry. I don't know what I was thinking. You're right!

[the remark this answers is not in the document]
ronag:
```js
ref.deref() === undefined
```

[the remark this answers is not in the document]
dnlup: Thanks

Ethan-Arrowood (approved):

mcollina (approved):
lgtm

PR #621, opened 2021-03-25
Bring test coverage above 95%

### LINKED ISSUE #438: 100 coverage
We need 100 coverage.

- [x] `maxHeadersSize` option is missing tests.
- [x] `Agent.closed`, `Agent.destroyed` is missing tests.
- [ ] redirect to `onUpgrade`

```
npm run coverage
```

ShogunPanda:
This PR provides mainly two things:

1. Make sure it is possible to connect to IPv6 IPs. @mcollina @ronag Please double check this. I randomly found this while bringing coverage up.
2. Bring coverage above 95% (now it is around 98%).

### DIFF
--- a/index.js
+++ b/index.js
@@ -32,7 +32,7 @@ function dispatchFromAgent (fn) {
     }
 
     const { origin, pathname, search } = util.parseURL(url)
-    const path = `${pathname || '/'}${search || ''}`
+    const path = `${pathname /* istanbul ignore next */ || '/'}${search || ''}`
 
     return fn.call(agent, { ...opts, origin, method, path }, ...additionalArgs)
   }
--- a/index.js
+++ b/index.js
@@ -32,7 +32,7 @@ function dispatchFromAgent (fn) {
     }
 
     const { origin, pathname, search } = util.parseURL(url)
-    const path = `${pathname /* istanbul ignore next */ || '/'}${search || ''}`
+    const path = `${pathname}${search || ''}`
 
     return fn.call(agent, { ...opts, origin, method, path }, ...additionalArgs)
   }
--- a/lib/agent/index.js
+++ b/lib/agent/index.js
@@ -72,7 +72,7 @@ class Agent extends EventEmitter {
 
   dispatch (opts, handler) {
     if (!opts || typeof opts !== 'object') {
-      throw new InvalidArgumentError('Origin must be a non-empty string.')
+      throw new InvalidArgumentError('opts must be a object.')
     }
 
     const { maxRedirections = this[kMaxRedirections] } = opts
@@ -94,10 +94,10 @@ class Agent extends EventEmitter {
       return client.dispatch(opts, handler)
     }
 
+    /* istanbul ignore next */
     if (util.isStream(opts.body)) {
       opts.body
-        .on('data', /* istanbul ignore next */ function () {
-          /* istanbul ignore next */
+        .on('data', function () {
           assert(false)
         })
     }
--- a/lib/agent/redirect.js
+++ b/lib/agent/redirect.js
@@ -2,9 +2,6 @@
 
 const { InvalidArgumentError } = require('../core/errors')
 const util = require('../core/util')
-const assert = require('assert')
-
-const kAgent = Symbol('agent')
 
 class RedirectHandler {
   constructor (agent, opts, handler) {
@@ -16,33 +13,9 @@ class RedirectHandler {
     this.destroyed = false
   }
 
-  destroy (err) {
-    this.destroyed = true
-
-    if (this.abort) {
-      this.abort()
-    }
-
-    this.handler.onError(err)
-  }
-
-  get (origin) {
-    return this[kAgent].get(origin)
-  }
-
   onConnect (abort) {
-    if (this.destroyed) {
-      abort()
-    } else {
-      this.abort = abort
-      this.handler.onConnect(abort)
-    }
-  }
-
-  onUpgrade (statusCode, headers, socket) {
-    assert(!this.location, 'redirect should not upgrade')
-
-    this.handler.onUpgrade(statusCode, headers, socket)
+    this.abort = abort
+    this.handler.onConnect(abort)
   }
 
   onError (error) {
@@ -61,7 +34,7 @@ class RedirectHandler {
     }
 
     const { origin, pathname, search } = util.parseURL(new URL(this.location, this.opts.origin))
-    const path = search ? `${pathname || '/'}${search || ''}` : pathname
+    const path = search ? `${pathname /* istanbul ignore next */ || '/'}${search}` : pathname
 
     this.opts = { ...this.opts }
 
@@ -83,8 +56,23 @@ class RedirectHandler {
 
   onData (chunk) {
     if (this.location) {
-      // Ignore body.
-      // TODO (fix): Spec ref?
+      /*
+        https://tools.ietf.org/html/rfc7231#section-6.4
+
+        TLDR: undici always ignores 3xx response bodies.
+
+        Redirection is used to serve the requested resource from another URL, so it is assumes that
+        no body is generated (and thus can be ignored). Even thought generating a body is not prohibited.
+
+        For status 301, 302, 303, 307 and 308 (the latter from RFC 7238), the specs mention that the body usually
+        (which means it's optional and not mandated) contain just an hyperlink to the value of
+        the Location response header, so the body can be ignored safely.
+
+        For status 300, which is "Multiple Choices", the spec mentions both generating a Location
+        response header AND a response body with the other possible location to follow.
+        Since the spec explicitily chooses not to specify a format for such body and leave it to
+        servers and browsers implementors, we ignore the body as there is no specified way to eventually parse it.
+      */
     } else {
       return this.handler.onData(chunk)
     }
@@ -92,7 +80,14 @@ class RedirectHandler {
 
   onComplete (trailers) {
     if (this.location) {
-      assert(!trailers || trailers.length === 0, 'redirect should not have trailers')
+      /*
+        https://tools.ietf.org/html/rfc7231#section-6.4
+
+        TLDR: undici always ignores 3xx response trailers.
+
+        See onData method comment for more informations.
+      */
+
       this.agent.dispatch(this.opts, this.handler)
     } else {
       this.handler.onComplete(trailers)
--- a/lib/agent/redirect.js
+++ b/lib/agent/redirect.js
@@ -89,7 +89,7 @@ class RedirectHandler {
         TLDR: undici always ignores 3xx response bodies.
 
         Redirection is used to serve the requested resource from another URL, so it is assumes that
-        no body is generated (and thus can be ignored). Even thought generating a body is not prohibited.
+        no body is generated (and thus can be ignored). Even though generating a body is not prohibited.
 
         For status 301, 302, 303, 307 and 308 (the latter from RFC 7238), the specs mention that the body usually
         (which means it's optional and not mandated) contain just an hyperlink to the value of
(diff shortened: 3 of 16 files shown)


index.js
```diff
@@ -32,7 +32,7 @@ function dispatchFromAgent (fn) {
     }
 
     const { origin, pathname, search } = util.parseURL(url)
-    const path = `${pathname || '/'}${search || ''}`
+    const path = `${pathname /* istanbul ignore next */ || '/'}${search || ''}`
```
mcollina, line 35: why ignoring? Could this even happen?

mcollina (commented):
Can you please add the -100 check of tap so we do not regress?

index.js
ronag: This can't happen. Just remove the `||`.

lib/agent/redirect.js
```diff
@@ -16,33 +13,9 @@ class RedirectHandler {
...
-    }
-  }
-
-  onUpgrade (statusCode, headers, socket) {
-    assert(!this.location, 'redirect should not upgrade')
-
-    this.handler.onUpgrade(statusCode, headers, socket)
+    this.abort = abort
+    this.handler.onConnect(abort)
   }
```
ronag, line 19: All if these can happen. Please put them back.

lib/agent/redirect.js
```diff
@@ -61,7 +34,7 @@ class RedirectHandler {
     }
 
     const { origin, pathname, search } = util.parseURL(new URL(this.location, this.opts.origin))
-    const path = search ? `${pathname || '/'}${search || ''}` : pathname
+    const path = search ? `${pathname /* istanbul ignore next */ || '/'}${search}` : pathname
```
ronag, line 37: Just remove the `||`.

lib/agent/redirect.js
```diff
@@ -83,16 +56,38 @@ class RedirectHandler {
...
+        servers and browsers implementors, we ignore the body as there is no specified way to eventually parse it.
+      */
     } else {
       return this.handler.onData(chunk)
     }
   }
 
   onComplete (trailers) {
     if (this.location) {
-      assert(!trailers || trailers.length === 0, 'redirect should not have trailers')
```
ronag, line 95: This can happen. Please put back.

lib/client-pool.js
```diff
@@ -104,6 +104,7 @@ class Pool extends EventEmitter {
 
     if (this[kConnections] && this[kClients].length === this[kConnections]) {
       for (const { busy } of this[kClients]) {
+        /* istanbul ignore next */
```
ronag, line 107: This should be possible to happen. Please remove ignore.

lib/core/client.js
```diff
@@ -65,7 +65,7 @@ function getServerName (client, host) {
   return (
     util.getServerName(host) ||
     (client[kTLSOpts] && client[kTLSOpts].servername) ||
-    util.getServerName(client[kUrl].host || client[kUrl].hostname) ||
+    util.getServerName(client[kUrl].host /* istanbul ignore next */ || client[kUrl].hostname) ||
```
ronag, line 68: This can happen. Please remove ignore.

ronag (changes requested):
Excellent work. Please fix the comments and remove all the istanbul ignores I think some of them we can make tests for. I'd like to have a shot at them as well later.

lib/agent/redirect.js
ShogunPanda: I'm not sure I understand.
In the comment I'm not saying it cannot happen, I'm saying that we don't handle them, like we do with the body.

I just forgot to remove the `this.agent.dispatch` line below. Do you agree or do you explicitily want to handle trailers?

index.js
ShogunPanda: Done!

lib/agent/redirect.js
ShogunPanda: Done. I was not able to find in the code how to reach those hooks so I leave it up to you.

lib/agent/redirect.js
ShogunPanda: Done!

lib/client-pool.js
ShogunPanda: Done. I leave it to you to add a test for it. I've tried with no luck.

lib/core/client.js
ShogunPanda: How comes? Since we are on Node LTS (so > 10 at the moment), `client[kUrl]` is a instance of `URL` (the WHATWG one), so if `host` is not defined, neither is `hostname`.

Am I wrong?

ShogunPanda:
@mcollina I'm not sure I can at the moment. Since Robert has to provide missing tests for the cases I unignored, if I do that now the CI will fail. Unless we can agree for a lower threshold, like 95%.

@ronag I've made most of the fixes you asked for, except for two I left comment for. Do you want me to remove ALL remaining istanbul ignores? Just double checking.

lib/core/client.js
ronag: you are right, then please remove the `||` part.

lib/agent/redirect.js
ronag: I would add a comment so that it's explicit that we ignore them on purpose

lib/agent/index.js
```diff
@@ -94,10 +94,10 @@ class Agent extends EventEmitter {
       return client.dispatch(opts, handler)
     }
 
+    /* istanbul ignore next */
```
ronag, line 97: This can happen? It's just the data event that we don't bother testing for.

lib/agent/redirect.js
ShogunPanda: Ok, and then I'll redirect to the full explanation above. Understood. Updated coming soon.

lib/client-pool.js
ronag: Please remove the ignore

ronag (changes requested):
Please remove all the ignores. I'll have a go at it later.

lib/core/client.js
```diff
@@ -863,7 +870,18 @@ function onSocketSession (session) {
 function connect (client) {
   assert(!client[kSocket])
 
-  const { protocol, port, hostname } = client[kUrl]
+  let { protocol, port, hostname } = client[kUrl]
+
+  // Resolve ipv6
+  if (hostname.startsWith('[')) {
+    const idx = hostname.indexOf(']')
+    const ip = idx !== -1 ? hostname.substr(1, idx - 1) /* istanbul ignore next */ : hostname
```
ronag, line 878: remove ternary and just make an `assert(idx !== -1)`

lib/core/client.js
```diff
@@ -863,7 +870,18 @@ function onSocketSession (session) {
...
+
+    /* istanbul ignore else */
+    if (net.isIP(ip)) {
+      hostname = ip
+    }
```
ronag, line 883: is the else possible here? what happens then?

lib/core/util.js
```diff
@@ -167,7 +167,7 @@ module.exports = {
   destroy,
   bodyLength,
   isBuffer,
-  queueMicrotask: global.queueMicrotask || (cb => Promise.resolve()
+  queueMicrotask: global.queueMicrotask /* istanbul ignore next */ || (cb => Promise.resolve()
```
ronag, line 170: Could we just do `global.queueMicrotask = null` in a test?

test/agent.js
```diff
@@ -484,3 +484,16 @@ test('fails with invalid URL', t => {
   t.throw(() => pipeline({}), InvalidArgumentError, 'throws on missing url.origin argument')
   t.throw(() => pipeline({ origin: '' }), InvalidArgumentError, 'throws on invalid url.origin argument')
 })
+
+test('constructor validations', t => {
+  t.plan(2)
+  t.throw(() => new Agent({ factory: 'ASD' }), InvalidArgumentError, 'throws on invalid opts argument')
+  t.throw(() => new Agent({ maxRedirections: 'ASD' }), InvalidArgumentError, 'throws on invalid opts argument')
```
ronag, line 491: also try negative integer

lib/agent/index.js
ShogunPanda: If you look at line 90, there is an entire check about body  empty stream. So, the if on 98 is only satisfied by 
I was not able to simulate a empty stream which emits so I skipped the if check as well.

ronag: I guess it doesn't matter so much.

lib/core/client.js
ShogunPanda: It should not because URL parsing would fail.
Ok, will do that!

lib/core/client.js
ShogunPanda: Same thing. I think I will add an assert here as well. Else should not be possible.

lib/agent/redirect.js
```diff
@@ -83,16 +83,38 @@ class RedirectHandler {
...
     if (this.location) {
-      // Ignore body.
-      // TODO (fix): Spec ref?
+      /*
+        https://tools.ietf.org/html/rfc7231#section-6.4
+
+        TLDR: undici always ignores 3xx response bodies.
+
+        Redirection is used to serve the requested resource from another URL, so it is assumes that
+        no body is generated (and thus can be ignored). Even thought generating a body is not prohibited.
```
ronag, line 92:
```suggestion
        no body is generated (and thus can be ignored). Even though generating a body is not prohibited.
```

lib/client-pool.js
ShogunPanda: Update coming soon!

lib/core/client.js
ShogunPanda: Update coming soon!

lib/core/util.js
ShogunPanda: Mh...we could but I definitely don't like playing with global.
Also, since we only support Node and we are on >10, do we really need this?

test/agent.js
ShogunPanda: Good point. On my way!

lib/core/util.js
mcollina: I would remove this completely. `queueMicrotask` should be available in 12+, and we should stop support v10 in the next major anyway.

lib/core/client.js
ronag: I'm fine with doing istanbul ignore or assertions.

lib/core/util.js
ronag: I'm good with that

ronag: Removed in separate PR. Ignore it here.

lib/agent/redirect.js
ShogunPanda: Done!

test/agent.js
ShogunPanda: Done!

lib/client-pool.js
ShogunPanda: Done!

lib/core/client.js
ShogunPanda: Done!

lib/core/client.js
ShogunPanda: Done!

lib/core/client.js
ShogunPanda: Done!

lib/agent/redirect.js
ShogunPanda: Thanks for noticing. Fixed.

ShogunPanda:
@mcollina `-100` was not possible. For now I've added `--lines 95 --branches 95 --functions 95 --statements 95` so we have a good starting point

@ronag I've removed all the unwanted ignores and update this PR description accordingly. I'll leave it to you to write test for the missing cases (or reinstate `ignores` ;)). 

Once this is merged I think also #603 can be merged, isn't it?

PR #628, opened 2021-03-27
fix: add more strict check for streams in `util.isStream()'

### LINKED ISSUE #625: Don't mistake buffers as streams
See https://github.com/nodejs/node/commit/9b87ab407146e17452db0441e81197b5a03a0c69.

We have the same problem in undici.

jarrodconnolly:
Fixes #625 - Added a similar check that the Node.js project uses.

The check used in `undici` was the reverse of the problem in the Node.js project. The commit mentioned in #625 shows they were missing the check for `.on()` where `undici` was missing the checks for `pipe()` or `write()'.

In its current state the `util.isStream()` allows `Stream`, correctly returns false on `Buffer`, but does allow `EventEmitter` since it also has `.on()`.

I changed the `util.isStream()` method to match what the Node.js project uses.

In https://github.com/nodejs/undici/blob/main/lib/core/request.js#L39 I switched to using the newly exposed `util.isReadable()` since the error message there indicates that it should be a Readable stream. There are likely other places that should use the more specific `util.isReadable()`.

I had to change the test `stream body without destroy` in `test/stream-compat.js` that was passing an `EventEmitter` as the request `body` since that case now fails differently than the test was looking for. I believe I kept the spirit of the test intact.

Hopefully, this is the direction intended for the issue.

### DIFF
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -36,7 +36,7 @@ class Request {
 
     if (body == null) {
       this.body = null
-    } else if (util.isStream(body)) {
+    } else if (util.isReadable(body)) {
       this.body = body
     } else if (util.isBuffer(body)) {
       this.body = body.length ? body : null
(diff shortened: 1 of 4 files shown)


mcollina (approved):
lgtm

Issue #638, opened 2021-03-29
Don't copy entire root when building Docker


ronag:
Instead of

```
COPY . .
```

Could we just copy the relevant files & folders?

ronag:
The dockerignore is probably sufficient if this is complicated.

ronag:
What exactly does the Docker file do and what does it need?

dnlup:
The `Dockerfile` uses a `Debian` image to install in a consistent way `wasm` dependencies (the SDK, so you don't have to set up anything on your machine if you don't want to) and to compile the `wasm` binary.

[`npm ci`](https://docs.npmjs.com/cli/v7/commands/npm-ci) is a command that performs a clean install by:

* removing `node_modules` if present
* installing dependencies using the `package-lock.json`

This method is faster than `npm i`, provided that a lock file is present. Otherwise, the installation would fail. I thought it was a good assumption because we are running the build script locally, in an environment where the lock file should be present.

mcollina:
Let's just move it to `npm i`

dnlup:
I opened a PR  to switch to `npm i`. Do we want to remove the `.dockerignore` file then? Do you think it's better to specify the assets to copy in the `Dockerfile` directly?

PR #639, opened 2021-03-29
fix: dockerignore

ronag:
Refs: https://github.com/nodejs/undici/issues/638

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -4,3 +4,12 @@ bench
 test
 images
 lib
+.clinic
+.nyc_output
+coverage
+docs
+docsify
+types
+.husky
+examples
+benchmarks


.dockerignore
```diff
@@ -4,3 +4,12 @@ bench
...
 lib
+.clinic
+.nyc_output
+coverage
+docs
+docsify
+types
+.husky
+examples
+benchmarks
```
dnlup, line 15: Oh, I missed a lot of things 😄 . We could also use a negative approach.

For example:
```
*
!index.js
```

will ignore everything bu `index.js`, like `.gitignore`.

dnlup (approved):

mcollina (approved):
lgtm

Issue #645, opened 2021-03-30
headersTimeout on uploads


ronag:
If there is a long running upload then the `headerTimeout` will fire. I think we should start the `headersTimeout` **after** the request body has been uploaded.

ronag:
@mcollina wdyt?

mcollina:
I agree!

PR #646, opened 2021-03-30
fix: don't headersTimeout while writing request body

ronag:
Fixes: https://github.com/nodejs/undici/issues/645

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -440,14 +440,18 @@ class Parser extends HTTPParser {
   }
 
   [HTTPParser.kOnHeaders] (rawHeaders) {
+    const { client, socket, headers } = this
+
+    assert(!client[kHeadersTimeout] || socket[kWriting || this.timeout])
+
     /* istanbul ignore next: difficult to make a test case for */
     if (this.paused) {
       this.queue.push([HTTPParser.kOnHeaders, rawHeaders])
       return
     }
 
-    if (this.headers) {
-      Array.prototype.push.apply(this.headers, rawHeaders)
+    if (headers) {
+      Array.prototype.push.apply(headers, rawHeaders)
     } else {
       this.headers = rawHeaders
     }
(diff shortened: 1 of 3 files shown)


ronag:
I couldn't get this to work with fake timers. Some issue with `.refresh()`.

lib/core/client.js
```diff
@@ -516,6 +516,9 @@ class Parser extends HTTPParser {
 
     const request = client[kQueue][client[kRunningIdx]]
 
+    // TODO (fix): Why does this assertion fail?
+    // assert(!client[kHeadersTimeout] || this.timeout)
```
dnlup, line 520: `this.timeout` must be getting set to `null` before getting here. Yeah, it's puzzling

lib/core/client.js
```diff
@@ -972,6 +973,15 @@ function _resume (client, sync) {
...
       }
+
+      if (
+        client[kHeadersTimeout] &&
+        socket &&
+        !socket[kWriting] &&
+        !socket[kParser].timeout
+      ) {
+        socket[kParser].timeout = setTimeout(onHeadersTimeout, client[kHeadersTimeout], socket[kParser])
+      }
```
dnlup, line 984: I don't fully understand the move of this check right here (It is more a question than a review comment). I understand that by checking if the `socket` is writing (the body), we can skip the timeout setup.

dnlup (approved):

ronag:
@dnlup something is strange here and seems related to the parser. Download this branch and run `tap test/http-100.js`. You will see:

```js
test/http-100.js 2> kOnHeadersComplete
test/http-100.js 2> onHeadersComplete
test/http-100.js 2> onHeadersComplete 2
test/http-100.js 2> kOnMessageComplete
test/http-100.js 2> onMessageComplete
test/http-100.js 2> kOnHeadersComplete
test/http-100.js 2> onHeadersComplete
```

Notice that `kOnHeadersComplete` is called once **after** `kOnMessageComplete`. That should not be possible?

lib/core/client.js
ronag: We have to move it before `pending === 0` check.

dnlup: Got it, thanks.

dnlup (approved):
lgtm

mcollina (approved):
lgtm

PR #648, opened 2021-03-31
Docker build cleanup

### LINKED ISSUE #638: Don't copy entire root when building Docker

dnlup:
Remove `.dokerignore` file and use consistent naming in npm scripts.

Fixes #638 .


dnlup:
I think the main drawback we have here js that now we are sending the whole root dir as a context before building the image:

* https://docs.docker.com/engine/reference/builder/#dockerignore-file

ronag:
> I think the main drawback we have here js that now we are sending the whole root dir as a context before building the image:
> 
> * https://docs.docker.com/engine/reference/builder/#dockerignore-file

Good point. Maybe do both then? COPY and dockerignore?

dnlup:
> > I think the main drawback we have here js that now we are sending the whole root dir as a context before building the image:
> > 
> > * https://docs.docker.com/engine/reference/builder/#dockerignore-file
> 
> Good point. Maybe do both then? COPY and dockerignore?

Yes. Just to be sure, we still won't `COPY . .` but each asset needed individually, right?

.dockerignore
```diff
@@ -1,15 +1,8 @@
...
-.clinic
-.nyc_output
-coverage
-docs
-docsify
-types
-.husky
-examples
-benchmarks
+# Ignore everything but the stuff following the `*` with the `!`
```
dnlup, line 1: I opted for a "exclude everything" by default. It keeps the list shorter.

Ethan-Arrowood (approved):

mcollina (approved):
lgtm

Issue #660, opened 2021-04-01
Emit `'disconnect'` after `.destroy()`


ronag:
Currently we mute events after `destroy` but that causes problems for introspection tools. We should always emit disconnect when disconnecting the socket.

i.e. no further events **after** `.destroy()` has completed, i.e. not **while** it is destroying.

Ethan-Arrowood:
@ronag is this missing from all `Client`, `Pool`, and `Agent` ?

PR #665, opened 2021-04-02
fix: emit disconnect after destroy() 

ronag:
Fixes: https://github.com/nodejs/undici/issues/660

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -853,9 +853,9 @@ function onSocketClose () {
 
     // Retry remaining requests.
     client[kPendingIdx] = client[kRunningIdx]
-
-    client.emit('disconnect', client[kUrl], [client], err)
   }
+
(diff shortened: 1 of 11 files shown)


dnlup (approved):
I agree with the changes. I think the test failures are expected because this is kind of a breaking change from the previous behavior. I see a lot of checks on the `disconnect` event failing.

Ethan-Arrowood (approved):
This all looks good. I have one non-blocking doc update request:

Could you add "If a client hasn't already disconnected, when it is destroyed, it will emit the ['disconnect'](#event-disconnect) event." to the [Client.destroy()](https://undici.nodejs.org/#/docs/api/Client?id=clientdestroyerror-callback) section of the docs?

dnlup (approved):

PR #668, opened 2021-04-02
test: change `http - no agent` benchmark to use the correct options

jarrodconnolly:
Based on the naming and the look of the options there, it appears the `http - no agent` benchmark was using the incorrect options object.


mcollina (approved):
lgtm

PR #675, opened 2021-04-03
fix: update deps

ronag:


ronag:
@dnlup I'm having trouble making tests pass. Would you mind have a look? Maybe you have better luck.

mcollina (approved):
lgtm

dnlup:
No problem, I'll take a look at it @ronag

dnlup:
The `fake-timers` update is causing the failure. Apparently, the context passed to the callbacks is `undefined`. Not sure if this is something we should address in `undici` or not yet.

ronag:
Sounds like a bug in fake timers

dnlup:
> Sounds like a bug in fake timers

Could be, I don't understand why only the `onBodyTimeout` has this problem though. I would pin `fake-timers` to the previous version, or do we need the latest one now?

As a side note, should we enable depandabot updates?

ronag:
> Could be, I don't understand why only the onBodyTimeout has this problem though. 

Yea. That's weird.

> I would pin fake-timers to the previous version, or do we need the latest one now?

We can pin. I'll do so.

> As a side note, should we enable depandabot updates?

I'm neutral. @mcollina wdyt?

PR #676, opened 2021-04-03
fix: named event args

ronag:
Use named arguments (object) for events.

### DIFF
--- a/docs/api/Dispatcher.md
+++ b/docs/api/Dispatcher.md
@@ -745,7 +745,7 @@ Parameters:
 * `Object`
   * **url** `URL`
   * **targets** `Array<Dispatcher>`
-  * **error** `Error`
+  * **err** `Error`
 
 ### Event: `'drain'`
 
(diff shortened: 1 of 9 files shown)


docs/api/Dispatcher.md
```diff
@@ -734,21 +734,25 @@ Returns: `Boolean`, `false` if user should wait for `'drain'` event before calli
...
 
 Parameters:
 
-* **origin** `URL`
-* **targets** `Array<Dispatcher>`
-* **error** `Error`
+* `Object`
+  * **url** `URL`
+  * **targets** `Array<Dispatcher>`
+  * **error** `Error`
```
ronag, line 748:
```suggestion
  * **err** `Error`
```

mcollina (changes requested):
I do not understand this change. Multiple event arguments are customary in Node.js

ronag:
> I do not understand this change. Multiple event arguments are customary in Node.js

I got tired trying to figure out what order would make most sense and think it's nice if users don't need to remember the order.

Don't care that much just thought this would be nice. Close?

mcollina:
I would recommend closing.

Issue #666, opened 2021-04-02
Allow overriding `headersTimeout` and `bodyTimeout` on request basis


ronag:
Right now these are always determined by the client constructor argument.

artur-ma:
Just a question, to understand how do u see it

allowing it per request means there should be validation on each request I guess (e.g check if integer)
does it affect performance? 

another one, here on http2 thread ( https://github.com/nodejs/undici/issues/547)  u said that that new http2Client should be implemented, does it mean each client should validate its own options, even if there is some code duplication (since constructor options are validated on client level), or should it move to another place?

PR #679, opened 2021-04-05
Override timeouts per request

artur-ma:
Trying to close https://github.com/nodejs/undici/issues/666

Not sure I'm on the right way,  still missing tests 

I'm curious about the HttpParser, I don't have a context there, so I rely on `client[kRunningIdx]` inside `onHeaders`
as it is done in `onHeadersComplete`

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -271,6 +271,7 @@ class Client extends Dispatcher {
       }
 
       const request = new Request(opts, handler)
+
       if (this[kDestroyed]) {
         throw new ClientDestroyedError()
       }
@@ -609,17 +610,25 @@ class Parser {
       return -1
     }
 
+    const headersTimeout = request.headersTimeout !== undefined
+      ? request.headersTimeout
+      : client[kHeadersTimeout]
+
     assert(
       timeout || // have timeout
-      !client[kHeadersTimeout] || // no timeout
+      !headersTimeout || // no timeout
       socket[kWriting], // still writing
       'invalid headers timeout state'
     )
 
+    const bodyTimeout = request.bodyTimeout !== undefined
+      ? request.bodyTimeout
+      : client[kBodyTimeout]
+
     if (statusCode >= 200) {
       clearTimeout(timeout)
-      this.timeout = client[kBodyTimeout]
-        ? setTimeout(onBodyTimeout, client[kBodyTimeout], this.socket)
+      this.timeout = bodyTimeout
+        ? setTimeout(onBodyTimeout, bodyTimeout, this.socket)
         : null
     } else if (timeout && timeout.refresh) {
       timeout.refresh()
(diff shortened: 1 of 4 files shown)


artur-ma:
@ronag can u take a look please if it's the way you want it to be implemented, if so, I will do the same for bodyTimeout and will add tests

lib/client.js
```diff
@@ -534,10 +536,11 @@ class Parser extends HTTPParser {
     const { client, socket, timeout } = this
 
     const request = client[kQueue][client[kRunningIdx]]
+    const headersTimeout = getRunningRequestOpts(client, kHeadersTimeout)
```
ronag, line 539:
```suggestion
    const headersTimeout = request.headersTimeout != null ? request.headersTimeout  : client[kHeadersTimeout]
```

lib/client.js
```diff
@@ -1013,13 +1022,15 @@ function _resume (client, sync) {
         return
       }
 
+      const headersTimeout = getRunningRequestOpts(client, kHeadersTimeout)
```
ronag, line 1025:
```suggestion
    const headersTimeout = request.headersTimeout != null ? request.headersTimeout  : client[kHeadersTimeout]
```

lib/core/request.js
```diff
@@ -9,14 +9,17 @@ const assert = require('assert')
...
 class Request {
   constructor ({
     path,
     method,
     body,
     headers,
     idempotent,
-    upgrade
+    upgrade,
+    timeouts
```
ronag, line 22:
```suggestion
    headerTimeout,
    bodyTimeout
```

lib/core/request.js
```diff
@@ -32,6 +35,20 @@ class Request {
...
+        throw new InvalidArgumentError('timeouts must be an objects')
+      }
+
+      if (timeouts.headersTimeout) {
+        this[kHeadersTimeout] = timeouts.headersTimeout
+      }
+    }
+
+    if (typeof this[kHeadersTimeout] === 'undefined') {
+      this[kHeadersTimeout] = null
```
ronag, line 49: You don't need to use symbols here. The request class i private to the module.

lib/client.js
artur-ma: I thought symbols are used for optimization since better to have a reference always to the same "symbol" rather than creating a new "string" object every time, am I wrong?

lib/client.js
artur-ma: The same will be for body timeout, on every line that we need a timeout setting (its about ~4 places)

```
const headersTimeout = request.headersTimeout != null ? request.headersTimeout  : client[kHeadersTimeout]
const bodyTimeout = request.bodyTimeout != null ? request.bodyTimeout  : client[kBodyTimeout]
```

also when someone else will have to get timeout setting in another place in the future, need to remember to add this condition, is this the right thing to do?

lib/client.js
ronag: Symbols are just to hide things from the user, i.e. instead of `private` which exists in many other languages. It's not for optimization.

ronag:
Do a rebase against main before doing too much more work. We've done some more refactoring.

.vscode/settings.json
```diff
@@ -0,0 +1,6 @@
+{
+  "node-tap.testFilePatterns": [
+    "test/*",
+    "*.test.js"
+  ]
+}
```
mcollina, line 6: can you please add this folder to .gitignore?

mcollina (changes requested):

.vscode/settings.json
artur-ma: Sorry missed that, done

mcollina (approved):
lgtm

ronag:
Thanks!

Issue #659, opened 2021-04-01
HTTPParserError: Expected HTTP


ronag:
Got this error while running master on our test setup.

```js
"HTTPParserError: Expected HTTP/\n    at Parser.[kMakeError] (/usr/src/app/node_modules/undici/lib/llhttp/parser.js:264:12)\n    at Parser.execute (/usr/src/app/node_modules/undici/lib/llhttp/parser.js:246:31)\n    at Socket.onSocketData (/usr/src/app/node_modules/undici/lib/client.js:793:17)\n    at Socket.emit (node:events:369:20)\n    at Socket.Readable.read (node:internal/streams/readable:524:10)\n    at Socket.read (node:net:622:10)\n    at flow (node:internal/streams/readable:1006:34)\n    at resume_ (node:internal/streams/readable:987:3)\n    at processTicksAndRejections (node:internal/process/task_queues:81:21)"
```

Source is a Node 15 server. So it's either an undici or node bug.

ronag:
I suspect the WASM update. @dnlup any ideas?

ronag:
pipelining is set to 1

dnlup:
> I suspect the WASM update. @dnlup any ideas?

Not atm, I'll dig into it. Do you have some payload examples?

ronag:
> Do you have some payload examples?

Sorry, not at the moment. I will also dig further as soon as I have time.

dnlup:
Ok 😉 . The parser was starting to parse what it thought was a new message, and it didn't find `HTTP`. So it considered it malformed. Is there something particular about these requests? I did not implement any pause logic, which was present in the native Node parser. Could that mean something?

ronag:
> Is there something particular about these requests? 

Not that I can think of... the error is gone now that I've reverted back to 3.x.

> I did not implement any pause logic, which was present in the native Node parser. Could that mean something?

Maybe?

Sorry I can't be more helpful atm.

dnlup:
>  the error is gone now that I've reverted back to 3.x.

It must be the parser then

dnlup:
> >  the error is gone now that I've reverted back to 3.x.
> 
> It must be the parser then

If you could checkout the commit that introduced the parser it would be a 100% proof

ronag:
I need to try to reproduce it somewhere else. It wasn't the test setup it was another setup used in production and I can't do debugging there. So it might be a bit tricky / take a while. I'll also try to help reviewing the code and see if I can find anything obvious.

[...]

PR #680, opened 2021-04-05
parser improvements

ronag:
Refactors and simplifies parser.


lib/llhttp/parser.js
```diff
@@ -57,11 +53,6 @@ const wasm_on_url = (p, at, length) => {
 }
 
 const wasm_on_status = (p, at, length) => {
-  const ret = currentParser[kTrackHeader](length)
-  if (ret !== 0) {
-    return ret
-  }
-  currentParser[kStatusMessage] = cstr(at, length)
   return 0
```
mcollina, line 65: if this function is empty, is it still needed?

mcollina (commented):
the changes looks good but CI is failing

lib/llhttp/parser.js
ronag: Yes, otherwise the WASM fails

mcollina: @dnlup we might want to make this optional there, it will save a few cycle.

dnlup: Yes, all the functions exported in https://github.com/nodejs/undici/blob/main/deps/llhttp/src/api.c#L29 must be provided in the initialization.

lib/llhttp/parser.js
```diff
@@ -149,127 +134,89 @@ class HTTPParser {
...
-    // Fill the view with `data`
-    u8.set(data)
+
+    this[kBufferView].set(data)
+
     // Call `execute` on the wasm parser.
     // We pass the `llhttp_parser` pointer address, the pointer address of buffer view data,
     // and finally the length of bytes to parse.
     // The return value is an error code or `constants.ERROR.OK`.
     // See https://github.com/dnlup/llhttp/blob/undici_wasm/src/native/api.c#L106
```
dnlup, line 175: I forgot to update this comments, now we have the src embedded, so we can point directly there. I'll do it in a PR.

lib/llhttp/parser.js
```diff
@@ -149,127 +130,89 @@ class HTTPParser {
...
-    const ptr = inst.exports.llhttp_get_error_reason(this[kPtr])
-    const u8 = new Uint8Array(inst.exports.memory.buffer)
-    const len = u8.indexOf(0, ptr) - ptr
-    const message = cstr(ptr, len)
-    const code = constants.ERROR[number]
-    return new HTTPParserError(message, code)
+    if (err === constants.ERROR.PAUSED_UPGRADE) {
+      const offset = inst.exports.llhttp_get_error_pos(this[kPtr]) - this[kBufferPtr]
+      this[kOnUpgrade](data.slice(offset))
+      this.close()
```
dnlup, line 183: Ah, wasn't the parser getting closed before? Could that be the cause of #659 ?

ronag: I don’t think so. It’s just closed earlier now.

lib/client.js
```diff
@@ -816,7 +798,11 @@ function onSocketError (err) {
 }
 
 function onSocketData (data) {
-  this[kParser].execute(data)
+  try {
```
dnlup, line 801: I see a small drop in the benchmarks, I'll run them again to be sure the use of the `try`/`catch` is having an impact.

https://github.com/nodejs/undici/runs/2275815074?check_suite_focus=true

ronag: Removing the try catch should be easy. I’ll fix it.

dnlup: On a second run, benchmarks look good. I would not change it yet.

mcollina (approved):
lgtm

Does it impact our benchmarks?

dnlup (approved):

dnlup:
> lgtm
> 
> Does it impact our benchmarks?

Benchmarks look good

Issue #682, opened 2021-04-06
flaky tests


ronag:
```
FAIL  test/client.js
 ✖ (unnamed test)

  test/client.js                                          
  517 |       body                                        
  518 |     }, (err, { statusCode, headers, body }) => {  
> 519 |       t.error(err)                                
      | --------^                                         
  520 |       body                                        
  521 |         .on('data', () => {                       
  522 |           t.fail()                                

  test: test/client.js basic POST with empty stream
  stack: |
    test/client.js:519:9
    lib/api/api-request.js:132:14

  Error Origin:

  lib/llhttp/parser.js                                                                      
  187 |       const message = Buffer.from(inst.exports.memory.buffer, ptr, len).toString()  
  188 |       const code = constants.ERROR[err]                                             
> 189 |       return new HTTPParserError(message, code)                                     
      | -------------^                                                                      
  190 |     }                                                                               
  191 |   }                                                                                 
  192 | }                                                                                   

  stack: |
    Parser.execute (lib/llhttp/parser.js:189:14)
    Socket.onSocketData (lib/client.js:807:29)
  type: HTTPParserError
  code: HPE_INVALID_CONSTANT
  message: "HTTPParserError: "

 FAIL  test/client.js
 ✖ Cannot read property 'on' of undefined

  test/client.js                     
  519 |       t.error(err)           
  520 |       body                   
> 521 |         .on('data', () => {  
      | ---------^                   
  522 |           t.fail()           
  523 |         })                   
  524 |         .on('end', () => {   

  test: basic POST with empty stream
  stack: |
    test/client.js:521:10
    lib/api/api-request.js:132:14
  type: TypeError
  tapCaught: uncaughtException

```

ronag:
Can't repro anymore.

ronag:
Another flaky test:

```js
 FAIL  test/redirect-request.js
 ✖ should be equal


  t.equal(statusCode, 200)
----^
  t.notOk(headers.location)
  /*

  --- expected    
  +++ actual      
  @@ -1,1 +1,1 @@ 
  -200            
  +426            

  test: test/redirect-request.js should follow redirections when going cross origin
  at:
    line: 293
    column: 5
    file: test/redirect-request.js
    type: Test
  stack: |
    Test.<anonymous> (test/redirect-request.js:293:5)

 FAIL  test/redirect-request.js
 ✖ should be equal

  */
  t.equal(body, 'POST')
----^
})

  --- expected      
  +++ actual        
[...]

PR #687, opened 2021-04-06
fix: use separate web assembly instance for every Client

ronag:
Refs: https://github.com/nodejs/undici/issues/682
Refs: https://github.com/nodejs/undici/issues/659

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -379,6 +379,7 @@ class Client extends Dispatcher {
   }
 }
 
+// TODO (perf): Have a Pool of reusable HTTPParsers?
 class Parser extends HTTPParser {
   constructor (client, socket) {
     super(insecureHTTPParser === true)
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm, I think this should fix all problems

dnlup:
@ronag are you able to test this in the environment that is affected by #659 before merging?

ronag:
> @ronag are you able to test this in the environment that is affected by #659 before merging?

Not really. They were not too happy about me using that environment for testing.

I think this makes sense either way, i.e. from a security standpoint different request payloads can't corrupt eachother's memory due to various bugs.

dnlup:
> > @ronag are you able to test this in the environment that is affected by #659 before merging?
> 
> Not really. They were not too happy about me using that environment for testing.
> 
> I think this makes sense either way, i.e. from a security standpoint different request payloads can't corrupt eachother's memory due to various bugs.

Yeah, that makes sense either way. You're right.

Sorry for causing this mess. And thank you for helping in debugging this issue.

ronag:
> Sorry for causing this mess. And thank you for helping in debugging this issue.

Don't be sorry! Quite the opposite. I very much appreciate everything you have helped with.

dnlup (approved):

PR #688, opened 2021-04-06
refactor: merge Parser and HTTPParser

ronag:
HTTPParser is so simple at this point that it makes it easier
to follow to just inline


dnlup (approved):
Definitely

mcollina (approved):
lgtm

PR #691, opened 2021-04-07
Update llhttp

### LINKED ISSUE #690: Upgrade llhttp
v5.1.0 is released.

Could we also write some docs as to how we update llhttp in undici?

dnlup:
Fixes #690


CONTRIBUTING.md
```diff
@@ -1,4 +1,79 @@
...
+npm i
+```
+
+#### Run the wasm build script
+
+```bash
+npm run build-wasm
+```
+
+#### Copy the sources to `undici`
```
dnlup, line 51: All these steps could be condensed to:

```bash
cp build/llhttp/wasm/* <path-to-undici>/lib/llhttp
```

If we remove the embedding and use the `llhttp` project build. Just wanted to say that's an option if this is too clumsy.

ronag (approved):
LGTM.

CONTRIBUTING.md
dnlup: This guide should be revisited if/when we have to modify `lhttp` sources for whatever reason.

CONTRIBUTING.md
```diff
@@ -1,4 +1,81 @@
...
+cp build/wasm/*.js.map <your-path-to-undici>/lib/llhttp/
+
+cp build/wasm/*.d.ts <your-path-to-undici>/lib/llhttp/
+
+cp src/native/api.c src/native/http.c build/c/llhttp.c <your-path-to-undici>/deps/llhttp/src/
+
+cp src/native/api.h build/llhttp.h <your-path-to-undici>/deps/llhttp/include/
+```
+
+#### Build the WebAssembly module in `undici`
```
dnlup, line 67: Also this would not be needed of we didn't embed.

mcollina (approved):
lgtm

CONTRIBUTING.md
mcollina: Can we avoid embedding as of now?

dnlup: Yes, we can. We don't currently modify the source files. IMHO is a cleaner approach, as long as we don't have to customize it. One thing to consider is the direction the `llhttp` project will take towards wasm. For example, what if the only way to access the build would be by installing it as a dependency and use its `.js` API (as suggested in https://github.com/nodejs/llhttp/pull/93)? Would we be open to that?

But regarding now, we would just have to run `build-wasm` in `llhttp` and copy the `build/wasm` folder contents in `lib/llhttp`.

PR #695, opened 2021-04-07
refactor: simplify api

ronag:


ronag:
@szmarczak @ShogunPanda

ShogunPanda:
It looks fine to me. It makes sense not to expose this variables as they are used in tests only.

lib/client.js
```diff
@@ -210,43 +210,47 @@ class Client extends Dispatcher {
...
 
-  get closed () {
-    return this[kClosed]
+  get [kBusy] () {
+    const socket = this[kSocket]
+    return (
+      (socket && (socket[kReset] || socket[kWriting])) ||
+      (this[kSize] >= (this[kPipelining] || 1)) ||
+      this[kPending] > 0
+    )
```
mcollina, line 253: The reason why I was skeptical in "removing" them was because those properties are needed elsewhere.

test/unix.js
```diff
@@ -19,6 +20,12 @@ if (process.platform !== 'win32') {
     })
     t.teardown(server.close.bind(server))
 
+    try {
+      fs.unlinkSync('/var/tmp/test3.sock')
+    } catch (err) {
+
```
mcollina, line 26: why is this empty?

test/unix.js
```diff
@@ -56,6 +63,12 @@ if (process.platform !== 'win32') {
...
```
mcollina, line 69: why is this empty?

test/unix.js
```diff
@@ -92,6 +105,12 @@ if (process.platform !== 'win32') {
...
```
mcollina, line 111: why is this empty?

mcollina (changes requested):

lib/client.js
ronag: They are not needed in public api

test/unix.js
ronag: This is just in case the test crashed in earlier run and there is a orphaned file.

Ethan-Arrowood (changes requested):
This is a great step in the right direction for simplifying the API.

Missing type updates though

dnlup (approved):

mcollina (approved):
lgtm

Issue #590, opened 2021-03-08
dispatch semver major change ideas


ronag:
This is just a list of ideas for optimizing dispatch by changing the API. Mostly enabled by https://github.com/nodejs/undici/pull/575/files.

- `onBody` buffer must be handled synchronously or copied.
- `rawHeaders` is an array of buffers instead of strings.
- Add `onHeader` (s) which must be handled synchronously or copied? Remove headers argument from `onHeaders` and `onComplete`. Rename `onHeaders` to `onStatus` (or something).

mcollina:
I would go with whatever increase the throughput. The only caveat is not have a regression in `request` and the others.

PR #696, opened 2021-04-07
perf: keep headers as Buffer

ronag:
Refs: https://github.com/nodejs/undici/issues/590


Ethan-Arrowood (approved):
i like this change; and I think in the future we can align more with the headers class/utilities from undici-fetch.

Maybe an undici-headers repo is in order 😋

mcollina:
Does this improve perf? Have you got metrics?

ronag:
See the benchmark ci

mcollina:
There does not seem to be any benefit.

ronag:
It’s negligible. I have some future zero copy optimizations in mind.

Maybe also useful when proxying.

I’ll see if I can find a better motivation for this change.

mcollina:
I'm not blocking it

mcollina (approved):
lgtm

PR #698, opened 2021-04-07
feat: Add callback context and redirect history.

ShogunPanda:
This PR concludes #603 by adding support for redirections history in the response.

To guarantee extensibility and encourage other handlers, I chose to introduce in the internal API the concept of `context`. All three major API methods (`request`, `pipeline` and `stream`) now always return a `context` in the response. This context is initially just an empty object, but handlers can fill them. 

This is what the `RedirectHandler` current uses.

### DIFF
--- a/lib/handler/redirect.js
+++ b/lib/handler/redirect.js
@@ -36,6 +36,13 @@ class RedirectHandler {
       return this.handler.onHeaders(statusCode, headers, resume)
     }
 
+    // Add the URL to the list of URLs visited while following redirections
+    if (!Array.isArray(this.handler.context.history)) {
+      this.handler.context.history = []
+    }
+
+    this.handler.context.history.push(new URL(this.opts.path, this.opts.origin).toString())
+
     const { origin, pathname, search } = util.parseURL(new URL(this.location, this.opts.origin))
     const path = search ? `${pathname}${search}` : pathname
 
--- a/lib/handler/redirect.js
+++ b/lib/handler/redirect.js
@@ -3,6 +3,8 @@
 const { InvalidArgumentError } = require('../core/errors')
 const util = require('../core/util')
 
+const redirectableStatusCodes = [300, 301, 302, 303, 307, 308]
+
 class RedirectHandler {
   constructor (agent, opts, handler) {
     this.agent = agent
@@ -26,22 +28,20 @@ class RedirectHandler {
   }
 
   onHeaders (statusCode, headers, resume) {
-    if ([300, 301, 302, 303, 307, 308].indexOf(statusCode) === -1) {
-      return this.handler.onHeaders(statusCode, headers, resume)
-    }
-
-    this.location = parseLocation(headers)
+    this.location = parseLocation(statusCode, headers)
 
     if (!this.location) {
       return this.handler.onHeaders(statusCode, headers, resume)
     }
 
     // Add the URL to the list of URLs visited while following redirections
-    if (!Array.isArray(this.handler.context.history)) {
-      this.handler.context.history = []
-    }
+    if (typeof this.handler.context === 'object') {
+      if (!Array.isArray(this.handler.context.history)) {
+        this.handler.context.history = []
+      }
 
-    this.handler.context.history.push(new URL(this.opts.path, this.opts.origin).toString())
+      this.handler.context.history.push(new URL(this.opts.path, this.opts.origin).toString())
+    }
 
     const { origin, pathname, search } = util.parseURL(new URL(this.location, this.opts.origin))
     const path = search ? `${pathname}${search}` : pathname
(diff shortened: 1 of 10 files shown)


mcollina (approved):
lgtm

Ethan-Arrowood (approved):

ronag:
Very elegant solution. Good job!

ronag (approved):

lib/handler/redirect.js
```diff
@@ -54,6 +54,13 @@ class RedirectHandler {
       return this.handler.onHeaders(statusCode, headers, resume)
     }
 
+    // Add the URL to the list of URLs visited while following redirections
+    if (!Array.isArray(this.handler.context.history)) {
+      this.handler.context.history = []
+    }
+
+    this.handler.context.history.push(new URL(this.opts.path, this.opts.origin).toString())
```
ronag, line 62: What if handler doesn't have context?

ronag: Can we pass context similar to how we pass `opaque`? also make it optional.

ronag (changes requested):

lib/handler/redirect.js
ronag: I'd prefer if we don't have properties on handlers in terms of user api.

ShogunPanda: Ok, let me see what I can do. It's also a good idea because it enables custom handlers to be customizable without requiring changes in undici internals. 

I'll just clarify in docs that the context object might be modified by undici. Does it sound reasonable?

ronag: > I'll just clarify in docs that the context object might be modified by undici. 

Let's try it and see.

ronag: Another alternative could be to have another method on the handler:

```js
handler.onContext({ history })
```

ShogunPanda: I like it. It requires less modification from the current code and does not require a change in the external API. 
Also, handler implementers can better choose how to handle the context they receive (like "replacing" instead of "merging" and so on.

Stay tuned.

ShogunPanda: Nevermind. I just realized that having a method, especially in the redirect handler would be messy since it would need another method/property to get the current history.

I'll go with user provided object.

ronag: Why? The redirect handler creates the property on itself and then just calls `onContext` to give the user access to it.

ronag: > I'll go with user provided object.

Probably also fine. We can start with trying that.

ShogunPanda: > Why? The redirect handler creates the property on itself and then just calls onContext to give the user access to it.

Because the RedirectHandler is recreated every time (decreasing maxRedirections) when following the chain. I would then store the history in opts and then call `onContext`. At that point is easier to use the user object.

dnlup (approved):

lib/handler/redirect.js
ShogunPanda: So, I have analyzed both solutions.

The first one, passing the context like in opaque is not working as the opts are lost in the API handler (since we destructure there). Saving the context there is not different from what I'm doing now and it complicates the external API so I gave up on that.

The `onContext` approach instead has the issue that if the chain is interrupted (if maxRedirections reaches 0 and 3xx is still  returned) is not called at all as it is inside RedirectHandler and not inside the Agent.

I also tried to call `onContext` expecting the bottom handler to return an object to modify. But this implies saving the context in the bottom handler which is what I'm doing at the moment. 

I also tried to mix the approaches, no luck.

I therefore think we should keep the current implementation, unless you have better idea.

ronag:
I would like to give this more thought. It's easy to add API's but more difficult to remove.

ronag (changes requested):
Changes look good. I would however like to give this more thought. It's easy to add API's but more difficult to remove.

lib/handler/redirect.js
ronag: Thanks for putting the effort into considering both approaches. I'll also have a go at this.

ShogunPanda: Ok, keep me posted then. In the meanwhile I'll just send one more commit that checks that the handler has a context before attempting to set it.

ronag:
Maybe we should land this with docs and typedefs after @Ethan-Arrowood has done his iteration. I kind of told him that there would be no further API changes 😄 .

ronag:
What about an additional optional `context` arg passed to `onConnect`?

ShogunPanda:
How is it different from the `onContext` approach above? The problem is still what I highlighted above: on the last redirect the RedirectHandler is not used at all so it cannot finalize the history.

ShogunPanda:
@ronag Any update on this? How are we moving forward?

ronag:
> @ronag Any update on this? How are we moving forward?

I haven't had time yet.

ronag:
I looked into it last thursday but didn't have enough time, will try again this thursday

ronag:
Closing in favor of https://github.com/nodejs/undici/pull/769 which is a continuation.

PR #701, opened 2021-04-08
test(client): add max headers test case

dnlup:


lib/client.js
```diff
@@ -456,7 +456,7 @@ class Parser {
     this.upgrade = false
     this.headers = []
     this.headersSize = 0
-    this.headersMaxSize = client[kMaxHeadersSize] = 8 * 1024
+    this.headersMaxSize = client[kMaxHeadersSize] || 8 * 1024
```
dnlup, line 459: Otherwise it would not be possible to pass a custom size from the `Client`.

dnlup:
All the tests here are failing because of the checks on the `asserts` in the headers hooks. Should we be able to pass a custom size to the client? Maybe I took the wrong approach.

test/client-errors.js
```diff
@@ -994,3 +994,29 @@ test('socket errors', t => {
     t.end()
   })
 })
+
+test('headers overflow', { only: true }, t => {
```
ronag, line 998: only?

ronag:
Landed in https://github.com/nodejs/undici/commit/6181944109728944c3213bc1c8c7c5513e29684a

dnlup:
Thank you 😉

PR #703, opened 2021-04-08
test(client): add parser buffer reallocation case

dnlup:
Ref #438

### DIFF
--- a/test/client.js
+++ b/test/client.js
@@ -1063,3 +1063,38 @@ test('emit disconnect after destory', t => {
     })
   })
 })
+
+test('parser dinamic allocation', { only: true }, t => {
+  t.plan(3)
+  const chunksSent = []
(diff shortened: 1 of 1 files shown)


test/client.js
```diff
@@ -1063,3 +1063,38 @@ test('emit disconnect after destory', t => {
     })
   })
 })
+
+test('parser dinamic allocation', { only: true }, t => {
+  t.plan(3)
```
dnlup, line 1068: Should we go deeper and actually check the parser `bufferSize` property?

ronag: No I don't think that is necessary. But, wouldn't hurt either.

dnlup: I'll add it in https://github.com/nodejs/undici/pull/704

PR #705, opened 2021-04-08
test: Added test for RedirectHandler upgrade handling.

ShogunPanda:
This covers `undici.upgrade` use case when using `maxRedirections`, as requested in #438.

### DIFF
--- a/lib/api/api-connect.js
+++ b/lib/api/api-connect.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const { InvalidArgumentError, RequestAbortedError } = require('../core/errors')
+const { InvalidArgumentError, RequestAbortedError, SocketError } = require('../core/errors')
 const { AsyncResource } = require('async_hooks')
 const util = require('../core/util')
 const { addSignal, removeSignal } = require('./abort-signal')
@@ -32,6 +32,10 @@ class ConnectHandler extends AsyncResource {
     }
 
     this.abort = abort
+  }
+
+  onHeaders () {
+    throw new SocketError('bad upgrade')
   }
 
   onUpgrade (statusCode, headers, socket) {
--- a/lib/api/api-pipeline.js
+++ b/lib/api/api-pipeline.js
@@ -8,7 +8,8 @@ const {
 const {
   InvalidArgumentError,
   InvalidReturnValueError,
-  RequestAbortedError
+  RequestAbortedError,
+  SocketError
 } = require('../core/errors')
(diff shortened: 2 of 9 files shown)


ronag:
Do we need a test for redirecting to an upgrade? i.e. upgrade **after** redirect?

ShogunPanda:
Very good point. Let me refactor so it tests both at once!

ronag:
Maybe we should land this with docs and typedefs **after** @Ethan-Arrowood has done his iteration. I kind of told him that there would be no further API changes 😄 .

ShogunPanda:
@ronag I've update the test and found out that upgrade was not working with redirect. I was eventually able to fix it but I'd like to have your opinion.

ronag:
I think the test is flawed. The server should not upgrade the connection if the user hasn't asked for it.

What you should do here is add tests for `undici.upgrade` and `undici.connect` where the server redirects.

ronag:
Another option is that we simply don't support redirect + upgrade for now. Which IMO is unfortunate but acceptable.

ShogunPanda:
Well, the client is using upgrade. So it implicitily asked for Upgrade to undici and thus to the server. And since maxRedirections are not zeroed, we also have to handle 3xx. Am I wrong?

ronag:
Ah, sorry. My mistake. I’ll take another luck.

ronag:
@ShogunPanda  can you pull in the changes from https://github.com/nodejs/undici/tree/705?

ronag:
I'm a little unsure how `CONNECT` and redirects should behave.

ShogunPanda:
> @ShogunPanda can you pull in the changes from [51bde24](https://github.com/nodejs/undici/commit/51bde2490c054bb9f3b3041b1d4896aa69e62944)?

Sure, will do later tonight.

ShogunPanda:
> I'm a little unsure how `CONNECT` and redirects should behave.

I'm not sure as well. I will read the RFC later and give you an update here.

ShogunPanda:
@ronag I've pulled your commit as requested.

I've also checked the RFC (https://tools.ietf.org/html/rfc7231#section-4.3.6).

It says that unless the server responds with a 2xx, HTTP remains in charge and thus this implies that the server might send a 3xx and we should handle as usual.
At least that's my understanding. :)

ronag:
Landed in https://github.com/nodejs/undici/commit/aeddb6e207ecb5bce8125f0fcc820879cd029933

PR #709, opened 2021-04-08
fix: res end before req body

ronag:


dnlup (approved):

ronag:
Should we have a separate error type for when the response has ended before the request? If so, what should it be named?

ronag:
@dnlup please see https://github.com/nodejs/undici/pull/709/checks?check_run_id=2297252479#step:6:3046 dynamic allocation test seems flaky on windows

dnlup:
> @dnlup please see https://github.com/nodejs/undici/pull/709/checks?check_run_id=2297252479#step:6:3046 dynamic allocation test seems flaky on windows

Yeah, I'll make the check a little loose then.

Issue #710, opened 2021-04-08
Weird test flakiness


ronag:
I sometimes get weird errors when running tests... e.g.

```
 FAIL  test/client-upgrade.js
 ✖ should be equivalent

  test/client-upgrade.js              
  103 |     })                        
  104 |                               
> 105 |     t.same(headers, {         
      | ------^                       
  106 |       hello: 'world',         
  107 |       connection: 'upgrade',  
  108 |       upgrade: 'websocket'    

  --- expected                
  +++ actual                  
  @@ -1,5 +1,5 @@             
   Object {                   
  +  "grade": "upgra",        
     "connection": "upgrade", 
     "upgrade": "websocket",  
  -  "hello": "world",        
   }                          

  test: test/client-upgrade.js basic upgrade promise
  stack: |
    Server.<anonymous> (test/client-upgrade.js:105:7)
```

ronag:
Notice

`+"grade": "upgra"`
`-"hello": "world",``

"grade" and "upgra" are of the correct length, but either the `memory.buffer` or the `at` seems corrupted :/.

ronag:
@indutny do you have any ideas as to what could be happening here?

ronag:
I don't really see that we could be doing something here. The `wasm_on_header_field` and `wasm_on_header_value` are rather trivial.

ronag:
I've seen this kind of error several times while running tests locally. Not this exact variant but similar in the sense of weird values.

ronag:
> Notice
> 
> `+"grade": "upgra"`
> `-"hello": "world"`
> 
> `"grade"` and `"upgra"` are of the correct length, but either the `memory.buffer` or the `at` seems corrupted :/.

Consider a buffer that looks like this:

`"upgrade",\n"upgrade": "websocket"`

The relative indices corresponds with:

`"hello" : "world"`

indutny:
Could it be memory pointing to some buffer that gets overwritten?

indutny:
Yeah, looks like you don't copy the data that you get in `onHeader` and just use `Buffer.from` to create a view over that part of memory. This can go wrong when memory gets overwritten.

ronag:
Ah! Buffer.from doesn’t copy?

indutny:
Not if its argument is `ArrayBuffer`: https://nodejs.org/api/buffer.html#buffer_static_method_buffer_from_arraybuffer_byteoffset_length

PR #711, opened 2021-04-09
fix: copy buffers

ronag:
Fixes: https://github.com/nodejs/undici/issues/710

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -536,7 +536,10 @@ class Parser {
   onHeader (buf, at, len) {
     // TODO: we could optimize this further by making this part the responsibility of the user.
     // Forcing them to consume the buffer synchronously or copy it otherwise.
-    this.headers.push(Buffer.from(buf, at, len))
+    const data = Buffer.allocUnsafe(len)
+    Buffer.from(buf, at, len).copy(data)
+
+    this.headers.push(data)
 
     this.headersSize += len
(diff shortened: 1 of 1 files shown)


ronag:
@indutny

ronag:
Not sure if there is a faster way to copy which doesn't not involved creating a temporary `Buffer`.

indutny:
I think this should be about as fast as it can be, but you could experiment by creating a buffer from buffer something like wrapping your previous `Buffer.from()` calls with another `Buffer.from()`. The benefit of latter is that you don't do unsafe allocations.

ronag:
@indutny: Another question. Would it be possible for `wasm_on_header_` to instead point into the buffer passed to `execute`?  That way no copy would be required.

indutny:
@ronag that buffer is relative to the input of the `execute` already. I think the problem is that you copy it into wasm memory and so you get back a slice of the very same memory you copied it to. Not sure if there is a way to pass data to wasm without copying. Perhaps shared buffers?

mcollina (approved):
I have a feeling that this unwinds any gain we got by using buffers for Headers, would not be better to revert?

Anyway LGTM

ronag:
> I have a feeling that this unwinds any gain we got by using buffers for Headers, would not be better to revert?
> 
> Anyway LGTM

I still haven’t given up. I think I can sort this as zero copy still.

PR #712, opened 2021-04-09
perf: zero copy llhttp buffers

ronag:
Fixes: https://github.com/nodejs/llhttp/issues/100
Refs: https://github.com/nodejs/undici/pull/711


ronag:
@indutny

mcollina:
Can you add a test for that case?

ronag:
> Can you add a test for that case?

Which case?

dnlup (approved):

ronag:
nice perf boost

branch
```
[1] http - no agent  x 3,551 ops/sec ±7.68% (76 runs sampled)
[1] http - keepalive x 711 ops/sec ±1.14% (83 runs sampled)
[1] http - keepalive - multiple sockets x 5,036 ops/sec ±1.55% (80 runs sampled)
[1] undici - pipeline x 5,784 ops/sec ±3.40% (81 runs sampled)
[1] undici - request x 6,450 ops/sec ±1.89% (81 runs sampled)
[1] undici - pool - request - multiple sockets x 6,553 ops/sec ±1.17% (83 runs sampled)
[1] undici - stream x 6,735 ops/sec ±1.06% (84 runs sampled)
[1] undici - dispatch x 7,089 ops/sec ±0.71% (83 runs sampled)
[1] undici - noop x 7,120 ops/sec ±0.83% (86 runs sampled)
```

current
```
[1] http - no agent  x 3,036 ops/sec ±6.68% (76 runs sampled)
[1] http - keepalive x 665 ops/sec ±1.35% (78 runs sampled)
[1] http - keepalive - multiple sockets x 4,393 ops/sec ±1.14% (82 runs sampled)
[1] undici - pipeline x 4,861 ops/sec ±3.71% (78 runs sampled)
[1] undici - request x 5,792 ops/sec ±1.92% (80 runs sampled)
[1] undici - pool - request - multiple sockets x 5,806 ops/sec ±1.33% (82 runs sampled)
[1] undici - stream x 6,036 ops/sec ±1.46% (79 runs sampled)
[1] undici - dispatch x 6,362 ops/sec ±0.93% (85 runs sampled)
[1] undici - noop x 6,519 ops/sec ±0.91% (86 runs sampled)
```

ronag:
@indutny Just to confirm. This is safe? llhttp will always emit the entire data sent to `execute` and not keep any of the data in internal state which it later emits.

lib/client.js
```diff
@@ -499,14 +506,17 @@ class Parser {
       this.bufferView = new Uint8Array(this.inst.memory.buffer, this.bufferPtr, this.bufferSize)
     }
 
+    // TODO (perf): Can we avoid this copy somehow?
     this.bufferView.set(data)
```
ronag, line 510: @mcollina Do we know any WASM expert that might be able to tell us if there is a more efficient way to pass buffers to WASM?

dnlup: Not a wasm expert, just thinking out loud. There is an API to create a shared memory:

* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/Memory#creating_a_shared_memory

Not sure how it fits here.

ronag: Unfortunately I don't think that will help.

mcollina (approved):
lgtm

mcollina (commented):
what happens if the buffer is too small?

ronag:
> what happens if the buffer is too small?

Which buffer and where? That shouldn’t happen. Do you want an assertion?

indutny:
@ronag llhttp never stores or copies any data internally. This looks safe.

Issue #547, opened 2021-02-10
http2


ronag:
Adding http2 might be easier than I thought.

See, https://github.com/devsnek/snekfetch/blob/master/src/node/socket.js

mcollina:
I've recently been asked how I would approach adding HTTP2 support to Undici. The starting point should be to implement an HTTP2 version of https://github.com/nodejs/undici/blob/252b676756053bc8bcf70cff89cad33243a9c9e5/lib/core/client.js#L71 on top of an HTTP2 session.

Nytelife26:
I will make an attempt to work on this in a fork, as HTTP/2 support is one of my only qualms with undici (great job so far team). I will post an update here with my progress and I hope to submit a PR.

mcollina:
Go for it!

Nytelife26:
I will be waiting on the following pull requests to complete their lifecycle before I begin, to avoid risking late-stage merge conflicts:

- #606 
- #603 
- #600

mcollina:
I think this could be started right now without waiting for those. I would start writing an Http2Client that wraps a `Http2Session` object. None of those PRs touches those internals.

Nytelife26:
That's fine by me, then. I just questioned whether, given that the PRs at present change the structure and workings of some parts of the client, you'd want the HTTP/2 variants to inherit those improvements and thus need to wait for a merge, but if not I can start tonight when I get home.

mcollina:
There won't be much conflict to start with. There will be as you move up to implement Http2Pool and then a cross-protocol Agent.

Nytelife26:
It seems like the most effective solution, then, would be to start it and then follow along with the PRs when I get to those parts so I can ensure the development of the HTTP/2 variants stays up to date and won't create any delays or merge conflicts when the time comes.

mcollina:
Yes exactly!

Nytelife26:
Alright, I'll start with `Http2Session` and `Http2Client` tonight then, and from then forth engage in side-by-side development with relevant pull requests. Thank you!

mcollina:
Have you got any progress @Nytelife26 ?

ronag:
I think with https://github.com/nodejs/undici/pull/620 this should be even easier. You just need to create a new Client implementing the `Dispatcher` interface.

Nytelife26:
> How is it going?

[...]

PR #720, opened 2021-04-09
feat: add http2

Nytelife26:
## This relates to...

- #547

## Rationale

HTTP/2 support is one of the few things currently missing from Undici
that sets it behind native Node `http`.

## Changes

This pull request is a work in progress. There are no changes just
yet. The plan, however, as discussed in the original issue, is to
implement HTTP/2 support around a `Dispatcher` instance for `Pool`,
`Client` and `Agent`.

Implementation can work in one of the following ways (see RFC7540 and
RFC8740):
1. **Prior knowledge**
  - Servers using HTTP/2 can advertise the capability via methods like ALT-SVC
  - Under RFC7540 Section 3.5, the HTTP/2 connection preface must be sent
  - All future frames may immediately use the HTTP/2
2. **HTTP/1.1 Initialization** (servers where HTTP/2 support status is unknown)
  - Without prior knowledge, the client should start an HTTP/1.1 connection
  using the Upgrade header (`h2c` for http URI schema, `h2` for TLS) and
  HTTP2-Settings header
  - If the response is a HTTP 101 Switching Protocols request, the request
  will be accompanied by a response to prior requests and the connection can
  upgrade to HTTP/2. If the response is normal, exempli gratia, a HTTP 200 OK,
  the client must not switch to  HTTP/2.

As for HTTPS, TLS negotiation must occur prior to any upgrade to HTTP/2.

### Features

HTTP/2 variants of `Pool`, `Client` and `Agent`. Alternatively, an option in the
aforementioned classes that causes them to attempt a HTTP/2 upgrade.

### Bug Fixes

N/A.

N/A.

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1289,7 +1289,11 @@ function write (client, request) {
   let header
 
   if (upgrade) {
-    header = `${method} ${path} HTTP/1.1\r\nconnection: upgrade\r\nupgrade: ${upgrade}\r\n`
+    let connectionString = "upgrade";
+    if (headers.toLowerCase().includes("http2-settings")) {
+      connectionString += ", http2-settings";
+    }
+    header = `${method} ${path} HTTP/1.1\r\nconnection: ${connectionString}\r\nupgrade: ${upgrade}\r\n`
   } else if (client[kPipelining]) {
     header = `${method} ${path} HTTP/1.1\r\nconnection: keep-alive\r\n`
   } else {
--- a/lib/http2client.js
+++ b/lib/http2client.js
@@ -1,0 +1,50 @@
+const FRAME_TYPES = [
+	"DATA", // standard request / response payloads
+	"HEADERS", // starts stream, carries request headers
+	"PRIORITY", // advises the priority of a stream
+	"RST_STREAM", // immediately terminates a stream
+	"SETTINGS", // inform other end of endpoint config
+	"PUSH_PROMISE", // pre-warns peer of wanted streams
+	"PING",
+	"GOAWAY", // graceful version of RST_STREAM
+	"WINDOW_UPDATE", // defines flow-control
+	"CONTINUATION" // continue a sequence of headers
+];
+
+function parseHttp2Settings(settings) {
+	// under RFC7540 section 3.2.1, the settings must be a base64url encoded
+	// SETTINGS frame. 16 bit identifiers, 32 bit values. */
+	let parsed = "";
+	for (let i = 0; i < Object.keys(settings).length; i += 2) {
+		parsed += `${settings[i]}${settings[i + 1]}\r\n`;
+	}
+
+	// base64url polyfill, courtesy of @panva
+	let encoded;
+	if (Buffer.isEncoding("base64url")) {
+		encoded = Buffer.from(parsed)
+			.toString("base64url")
+			.replace(/[=]+$/g, "");
+	} else {
+		encoded = Buffer.from(parsed)
+			.toString("base64")
+			.replace(/\+/g, "-")
+			.replace(/\//g, "_")
+			.replace(/[=]+$/g, "");
+	}
+	return encoded;
+}
+
+function http2Connect(url, options) {
+	if (url.protocol === "https:") {
+		throw new Error("Invalid protocol - httpConnect upgrades http streams");
+	}
+	throw new Error("UNIMPLEMENTED");
+}
+
+function https2Connect(url, options) {
+	if (url.protocol === "http:") {
+		throw new Error("Invalid protocol - https2Connect upgrades https streams");
+	}
+	throw new Error("UNIMPLEMENTED");
+}
--- a/lib/http2client.js
+++ b/lib/http2client.js
@@ -43,19 +43,24 @@ function parseHttp2Settings(settings) {
 
 function http2Connect(client) {
 	assert(!client[kSocket]);
-	const { protocol, port, hostname } = client[kUrl];
+	const { protocol, port, hostname, pathname } = client[kUrl];
 	if (protocol === "https:") {
 		throw new Error("Invalid protocol - httpConnect upgrades http streams");
 	}
 	// TODO: allow ALT-SVC
 	client.upgrade({
+		path: pathname,
 		protocol: "h2c",
 		headers: {
-			"Connection": "Upgrade, HTTP2-Settings",
-			"Upgrade": "h2c",
-			"HTTP2-Settings": parseHttp2Settings(client[kHTTP2Opts])
+			"HTTP2-Settings": parseHttp2Settings(client[kHTTP2Opts] || {})
 		}
-	});
+	})
+		.then((res) => {
+			// upgrade successful, validate response and use http2
+		})
+		.catch((err) => {
+			// continue as normal
+		});
 }
 
 function https2Connect(client) {
@@ -65,3 +70,5 @@ function https2Connect(client) {
 	}
 	// TODO: tls negotiation comes FIRST
 }
+
+module.exports = http2Connect;
--- a/lib/http2client.js
+++ b/lib/http2client.js
@@ -1,3 +1,9 @@
+"use strict";
+/* eslint-disable no-unused-vars, max-len, id-length */
+
+const assert = require("assert");
+const { kUrl, kSocket, kHTTP2Opts } = require("./core/symbols");
+
 const FRAME_TYPES = [
 	"DATA", // standard request / response payloads
 	"HEADERS", // starts stream, carries request headers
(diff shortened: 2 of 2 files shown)


Nytelife26:
@mcollina as noted in the "Features" section of this pull request, we have two options.

> "HTTP/2 variants of Pool, Client and Agent. Alternatively, an option in the
aforementioned classes that causes them to attempt a HTTP/2 upgrade."

We have only discussed the prior, so I'd like to take this opportunity to ask what the desired approach is before I begin.

ronag:
I don't think you need to develop a new `Agent` or `Pool`. What we need is either modify existing `Client` to support both or create a new `Http2Client`, rename `Client` to `HttpClient` and then create another `Client` that somehow switches between the two as needed.

mcollina:
I would just focus on creating a `Http2Client` right now.

The hardest part is figuring out is the upgrade to HTTP/2

szmarczak:
@mcollina This is pretty easy. All we need to do is just create a TLS connection and check the ALPN protocol. Something like:

```
if (socket.alpnProtocol === 'h2') {
    const session = http2.connect(origin, {
        createConnection: () => socket
    });
    ...
}

const client = new HttpClient(socket);
...
```

Nytelife26:
> All we need to do is just create a TLS connection and check the ALPN protocol.

That depends - the protocol is different depending on whether the desired connection is HTTPS or HTTP. As I mentioned, if TLS is present, you must use ALPN first. If not, you have to follow the standard HTTP Upgrade procedure.

ronag:
I think http2 should always be https? I don’t think we need to support the non TLS case for http2.

Nytelife26:
> I think http2 should always be https? I don’t think we need to support the non TLS case for http2.

That doesn't sit right with me. We shouldn't just ignore usages that are part of the specification itself for our own convenience, especially when it'll likely result in someone filing an issue later on.

mcollina:
> I think http2 should always be https? I don’t think we need to support the non TLS case for http2.

It's needed and used within microservices networks.

szmarczak:
> We shouldn't just ignore usages that are part of the specification itself for our own convenience

[`undici` already does](https://github.com/nodejs/undici#specification-compliance).

> I think http2 should always be https? I don’t think we need to support the non TLS case for http2.

I thought the same - browsers don't. 

> It's needed and used within microservices networks.

I believe there are better solutions in these cases. Anyway they can just use the native `http2.connect` instead.

Nytelife26:
> It's needed and used within microservices networks.

Thank you.



> [`undici` already does](https://github.com/nodejs/undici#specification-compliance).

That's not quite the same as dropping an entire protocol. Although, I do trust and hope there are technical reasons for not following those parts too.

> I thought the same - browsers don't [support h2c].

Not yet. Most of the point of HTTP/2 was for browsers so It most likely will.

> I believe there are better solutions in these cases. Anyway they can just use the native `http2.connect` instead.

You could also argue that you could use the native `http` instead of undici altogether. But the point of undici is that it is faster. So yes, they *can* use an inferior solution, but I see no reason to not do our best just because other solutions already exist.

szmarczak:
> Most of the point of HTTP/2 was for browsers so It most likely will.

I don't think `h2c` is coming anytime soon. Even [CloudFlare doesn't support it](https://community.cloudflare.com/t/http-h2c-upgrade-requests-result-in-connection-reset-by-peer/214761).

> you could use the native http instead of undici altogether

That's not quite the same as `undici` is a complete rewrite, it doesn't use the `http` module internally. The point is that for now you have to use `http2.connect(...)` anyway. The API is very friendly so people should use it directly whenever the can.

I'd prefer if `undici` targeted HTTP/2 with TLS first, but if anybody wants to go after `h2c` as well - I don't mind.

mcollina:
FYI: Google Cloud Run supports h2c https://cloud.google.com/run/docs/configuring/http2.

My take is that h2c support might be really handy.

kanongil:
h2c is super useful for debugging tricky protocol issues using Wireshark, etc.

Nytelife26:
## Summary

> "What we need is either modify existing `Client` to support both or create a new `Http2Client`, rename `Client` to `HttpClient` and then create another `Client` that somehow switches between the two as needed."

**Decision**: To start with, I will be implementing the methods on a new `Http2Client`. It is likely that I will then merge those under an option into `Client`, given that a client with switching may be hard to maintain and create unnecessary overhead, depending on how complex the `Http2Client` will be.

> The hardest part is figuring out is the upgrade to HTTP/2

**Decision**:
- For TLS connections, the TLS side *must* be handled first. Then, we can follow standard procedure.
- The client will either initialize or use the existing HTTP/1.1 connection and send a request with the HTTP `Upgrade` (value of `h2` for HTTPS, `h2c` for HTTP) and `HTTP2-Settings` headers.
  - Implementation on capability advertisement via things like ALT-SVC is currently unclear.
- If the response is a HTTP 101, check for responses to prior requests and begin using HTTP/2. Else, do not switch to HTTP/2.

> I'd prefer if `undici` targeted HTTP/2 with TLS first, but if anybody wants to go after `h2c` as well - I don't mind.

**Decision**: As aforementioned, TLS support is part of the natural protocol of any HTTP/2 upgrade. It is an extension, not a separate method, and therefore both can be achieved simultaneously.

Nytelife26:
Does anyone here know about polyfill support for [`base64url`](https://github.com/nodejs/node/commit/f8ab632d56ea15c78b1dc5db1624109a2bf7b69b) and whether it would be possible to implement it here? The HTTP/2 specification requires that the `HTTP2-Settings` header be encoded in `base64url` (without padding), however, it seems to be a recent feature and I am uncertain of how it would affect compatibility if we were to use it here.

panva:
@mcollina you can take inspiration from https://github.com/panva/jose/blob/main/src/runtime/node/base64url.ts

Nytelife26:
What's the situation regarding `client.upgrade`? It's somewhat vital for this to work, however it doesn't seem to be implemented on `Client` or `Dispatcher`, despite what the documentation says.

image

ronag:
The api methods are assigned to the prototypes in index.js. See, https://github.com/nodejs/undici/blob/main/index.js#L15.

Nytelife26:
Interesting, thank you, that didn't display in GitHub search weirdly. Peculiar approach, but it makes life easier actually since I can avoid a circular dependency between the client and its http2 variant when they merge.

Nytelife26:
Can we safely implement a wider `Client.upgrade` return context? It only provides the headers and socket, but I need the full response body too.

Nytelife26:
> Can we safely implement a wider `Client.upgrade` return context? It only provides the headers and socket, but I need the full response body too.

cc: @ronag @mcollina 

See my above comment when you have the time, please and thank you in advance.

ronag:
> Can we safely implement a wider `Client.upgrade` return context? It only provides the headers and socket, but I need the full response body too.

No idea. As far as I know there is no body after upgrade so I'm not sure what you mean here...

Nytelife26:
> As far as I know there is no body after upgrade

According to the HTTP/2 specification, the server may bundle the
response frames to any previous requests along with its upgrade
acknowledgement. For that reason, I need to be able to access the
response body in order to check for those.

Nytelife26:
Answers, anyone?

ronag:
> Answers, anyone?

I don’t have a better answer. As far as I know there is no body with an upgrade so your question doesn’t make sense to me. Sorry. Maybe I’m missing something...

ronag:
> Right after sending the 101 status code, the server can begin speaking the new protocol, performing any additional protocol-specific handshakes as necessary. Effectively, the connection becomes a two-way pipe as soon as the upgraded response is complete, and the request that initiated the upgrade can be completed over the new protocol.

mcollina:
> Answers, anyone?

I didn't thought it was a question, more of a statement. I'd recommend to try and see if it's possible.

Nytelife26:
I can not continue to work on this, unfortunately. I have pushed my most recent work and hopefully that in combination with this discussion will serve useful to anyone willing to pick up. I apologise and wish you all well, but I no longer have the time, as I have packages to maintain and college is a heavy weight.

MetCoder95:
Hi! I would like to take it over and continue the work, but to be honest, I would need some time to sync completely with both the PR and the HTTP/2 spec, I'll do some lecture over the weekend to start putting hands-on, hope it's ok 🙂

Nytelife26:
> Hi! I would like to take it over and continue the work, but to be honest, I would need some time to sync completely with both the PR and the HTTP/2 spec, I'll do some lecture over the weekend to start putting hands-on, hope it's ok 🙂

Wonderful - I'm sure the team will appreciate your help. I documented most of the important design decisions and wrote my notes here since parsing both RFCs in their entirety is a mammoth task, so I hope they're of use to you. Good luck, and I'll be here for reviews when it's done at least.

MetCoder95:
Sadly I think I don't have write rights to the branch, any suggestion?

ronag:
Open a new PR

MetCoder95:
Closed in favor of #866

MetCoder95:
Just to double-check and give a life signal 😅 
The way to move forward will be using the native `http2` module from `node`?
And if so, any suggestion on how the abstraction/client should look like? :)

mcollina:
> Just to double-check and give a life signal 😅 
> 
> The way to move forward will be using the native `http2` module from `node`?

Yes.

> And if so, any suggestion on how the abstraction/client should look like? :)

I think this was extensively covered in this PR. To be honest, having _something_ that worked would be a good starting point. My suggestion is to start by creating a HTTP2Client.

MetCoder95:
I think @szmarczak is already working on a new PR adding HTTP/2. Happy to take a look at if so, really interested in how does it works 🙂

Nytelife26:
> I think @szmarczak is already working on a new PR adding HTTP/2. Happy
> to take a look at if so, really interested in how does it works 🙂

As far as I've seen, using node core `http2`, despite our former
discussions. Either way, hopefully it'll turn out well.

simoneb:
Is there any work being done on this?

Nytelife26:
> Is there any work being done on this?

Not to my knowledge, feel free to take over.

MetCoder95:
I'll try to revisit it as soon as I have time, but I think @szmarczak has some progress already

metcoder95:
Superseded by #2061

PR #724, opened 2021-04-10
Fix small MD issue in readme API doc

benjamingr:
:) 

(Alternatively, if it _is_ a third option and not part of `options`, the declaration needs to be updated)


mcollina (approved):
lgtm

Issue #725, opened 2021-04-12
WebAssembly instance out of memory


ronag:
```
[2021-04-12 11:08:00.287 +0000] ERROR (query): view query error
    err: {
      "type": "RangeError",
      "message": "WebAssembly.Instance(): Out of memory: wasm memory",
      "stack":
          RangeError: WebAssembly.Instance(): Out of memory: wasm memory
              at new Parser (/usr/src/app/node_modules/undici/lib/client.js:409:17)
              at connect (/usr/src/app/node_modules/undici/lib/client.js:1001:18)
              at _resume (/usr/src/app/node_modules/undici/lib/client.js:1134:7)
              at resume (/usr/src/app/node_modules/undici/lib/client.js:1035:3)
              at Client.dispatch (/usr/src/app/node_modules/undici/lib/client.js:294:9)
              at Pool.dispatch (/usr/src/app/node_modules/undici/lib/pool.js:204:30)
              at /usr/src/app/node_modules/@nxtedition/lib/couch.js:255:14
              at new Promise (<anonymous>)
              at request (/usr/src/app/node_modules/@nxtedition/lib/couch.js:254:12)
              at Object.allDocs (/usr/src/app/node_modules/@nxtedition/lib/couch.js:420:23)
    }
```

ronag:
@dnlup Thoughts? Do we need an option to set the memory limit? How do we do that?

ronag:
Started running on our test setups 😄

ronag:
I suspect it's because of large payload body. We should probably have a better error for this?

ronag:
Do we have a memory leak?

ronag:
@dnlup Thoughts? Workarounds? Pool/Re-use instances? Have single instance.

dnlup:
It seems like either:

* a memory leak
* more parsers->wasm instances are created and not cleaned up fast enough
* a V8/chromium issue?

I think it should be possible to set up a maximum limit for the memory:

* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/Memory/Memory#syntax

To understand further I would:

* measure memory consumption and which buffers are retained
* rollback to using a single web assembly instance. I suspect most of the flakiness was due to how we handled buffers (not copying them) than the fact that each parser needed its own instance

dnlup:
Good thing you have a test setup and we are in alpha.

dnlup:
[...]

PR #726, opened 2021-04-12
fix: use single wasm instance

ronag:
Fixes: https://github.com/nodejs/undici/issues/725

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -447,7 +447,6 @@ const llhttp = new WebAssembly.Instance(mod, {
   }
 }).exports
 
-// TODO (perf): Have a Pool of reusable HTTPParsers?
 class Parser {
   constructor (client, socket) {
     assert(Number.isFinite(client[kMaxHeadersSize]) && client[kMaxHeadersSize] > 0)
--- a/lib/client.js
+++ b/lib/client.js
@@ -497,6 +497,10 @@ class Parser {
   }
 
   execute (data) {
+    if (!this.ptr) {
+      return
+    }
+
     assert(this.ptr)
     assert(!currentParser)
 
--- a/lib/client.js
+++ b/lib/client.js
@@ -402,57 +402,57 @@ class HTTPParserError extends Error {
   }
 }
 
+let currentParser = null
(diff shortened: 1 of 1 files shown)


ronag:
I'll quickly publish another alpha once this has been merged.

lib/client.js
```diff
@@ -402,57 +402,57 @@ class HTTPParserError extends Error {
   }
 }
 
+let currentParser = null
+const llhttp = new WebAssembly.Instance(mod, {
```
dnlup, line 406: I think here we can specify memory limits, if we need it in the future

dnlup (approved):

lib/client.js
ronag: Do we know what the default limits are?

dnlup: I couldn't find it yet. Maybe a manual inspection could tell. But it should be easy to pass a custom memory instance, we just need to add a `memory` key to `env` if I am not mistaken:

* https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/WebAssembly/Memory/Memory#parameters
* https://ariya.io/2019/05/basics-of-memory-access-in-webassembly

Btw, I am not saying we should do it now.

mcollina:
I don't think this is correct. We have one parser per req/res in node core as well.

mcollina (approved):
lgtm anyway.

It would be better to have a regression test

dnlup:
> I don't think this is correct. We have one parser per req/res in node core as well.

I understand. Just the wasm instance would be shared (and the memory instantiated just once). The `js` parser and the `llhttp` would be one for each req/res, if I am not mistaken.

lib/client.js
```diff
@@ -831,6 +844,16 @@ class Parser {
   }
 
   destroy () {
+    if (this.ptr) {
+      llhttp.llhttp_free(this.ptr)
```
dnlup, line 848: Maybe  we were not freeing memory before?

ronag: Wasn't needed since the wasm was thrown away.

mcollina:
Good, then let's land this!

PR #731, opened 2021-04-12
docs: add an example with potentially idiomatic ways to pipe a file from one server to another

naugtur:
A PR is always better than an issue, right? This is a discussion PR. :smiley: 

I'm hopping to discuss the API surface of undici or get someone to point to one of the versions here as the right one.
I must say I don't see other ways to stream a file through using undici on both ends than these 4 and none of them feels like undici was designed to work that way. 

Which one do you like best?

BTW, I put timing in place, because in some cases I don't understand what the returned promise is waiting for. Seems like it gets resolved before the response handler is called. Should that be the case?

### DIFF
--- a/examples/transfer/index.js
+++ b/examples/transfer/index.js
@@ -1,0 +1,125 @@
+const undici = require('../../')
+const http = require('http')
+const fs = require('fs')
+const { promisify } = require('util')
+const { pipeline, Readable } = require('stream')
+const promisePipeline = promisify(pipeline)
+
+const dummyUploadConsumer = (req, res) => {
+  promisePipeline(req, fs.createWriteStream('a.jpg')).then(
+    () => {
+      console.log('✔️')
+      res.end('ok')
+    },
+    (err) => res.end(err.message)
+  )
+}
+
+async function run () {
+  const server = http.createServer(dummyUploadConsumer)
+  await promisify(server.listen.bind(server))(4000)
+
+  // one request to warm things up
+  const re = await undici.request(
+    'https://avatars.githubusercontent.com/u/509375'
+  )
+  re.body.resume()
+  await promisify(re.body.on.bind(re.body, 'end'))
+
+  // 1. Use it like any other http request lib
+  //    That doesn't seem like the way I'm supposed to use undici
+  console.time('1a')
+  const response = await undici.request(
+    'https://avatars.githubusercontent.com/u/509375'
+  )
+  console.log(response.statusCode)
+  const result = await undici.request('http://localhost:4000/', {
+    method: 'POST',
+    body: response.body
+  })
+  console.log(result.statusCode)
+  console.timeEnd('1a')
+  result.body.resume()
+
+  // 2. Attempt something more idiomatic?
+  //    I feel like I left a dangling readable end somewhere. A request-like method returning a Writable would work better than pipeline here
+  console.time('2a')
+  console.time('2b')
+  await undici.stream(
+    'https://avatars.githubusercontent.com/u/509375',
+    {},
+    ({ statusCode, headers }) => {
+      console.log(statusCode)
+      return undici.pipeline(
+        'http://localhost:4000/',
+        {
+          method: 'POST'
+        },
+        ({ statusCode, headers, body }) => {
+          console.log(statusCode)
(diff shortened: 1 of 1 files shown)


ronag:
I think either of these is fine:


```js
const response = await undici.request(
  'https://avatars.githubusercontent.com/u/509375'
)
console.log(response.statusCode)
const result = await undici.request('http://localhost:4000/', {
  method: 'POST',
  body: response.body
})
  
await undici.stream(
  'https://avatars.githubusercontent.com/u/509375',
  {},
  ({ statusCode, headers }) => {
    console.log(statusCode)
    return undici.pipeline(
      'http://localhost:4000/',
      {
        method: 'POST'
      },
      ({ statusCode, headers, body }) => {
        console.log(statusCode)
        console.timeEnd('2a')
        return body
      }
    )
  }
)

await promisePipeline(
  undici.pipeline(
    'https://avatars.githubusercontent.com/u/509375',
    {},
    ({ statusCode, headers, body }) => {
      console.log(statusCode)
      return body
    }
  ).end(),
  undici.pipeline(
    'http://localhost:4000/',
    {
      method: 'POST'
    },
    ({ statusCode, headers, body }) => {
      console.log(statusCode)
      console.timeEnd('3a')
      return body
    }
  )
)
```

Would personally probably prefer `undici.request` variant.

ronag:
We might want to consider making the `opts` argument optional so you don't have to pass `{}` which looks a little weird.

ronag:
It could be possible to achieve:

```js
await promisePipeline(
  undici.pipeline('https://avatars.githubusercontent.com/u/509375').end(),
  undici.pipeline('http://localhost:4000/', { method: 'POST' })
)
```

With some smarter defaults.

naugtur:
The stream/pipeline versions might be more performant and better at handling stream errors and cleanup than the two separate request calls. It'd be great to use stream.pipeline for the whole thing without looking like I'm misusing the library.
This came up in pair programming and we got stuck on a few variants with "It doesn't look like we're supposed to do this"

Also, a lot of "invalid protocol" errors when incorrect arguments were passed :wink:

ronag:
> The stream/pipeline versions might be more performant and better at handling stream errors and cleanup than the two separate request calls.

Actually, I think the `request` variant is most performant in this case. `pipeline` is slow.

naugtur:
IMHO the nicest version of this usecase would be if there was undici.streamTo which returns a writable that can be used within undici.stream's handler

ronag:
> IMHO the nicest version of this usecase would be if there was undici.streamTo which returns a writable that can be used within undici.stream's handler

```js
undici.stream(src, ({ statusCode }) => { 
  if (statusCode !== 200) throw new Error(statusCode)
  return undici.pipeline(dst)
})
```

Should work with latest master.

ronag:
> IMHO the nicest version of this usecase would be if there was undici.streamTo which returns a writable that can be used within undici.stream's handler

I'm not sure we should add more API's. Can you think of ways to improve the ones we currently have to make them nicer for your use case? I've recently pushed some API ergonomic improvements to master.

naugtur:
Well, all 4 examples work, so it's a matter of getting them to look more like it's the intended use.
I like the pipeline in stream handler (although pipeline is not the best name for this context) and the only concern I have is what happens to the other end of the duplex in pipeline. Would I clog my tcp queue if I ignored it entirely and just be happy about receiving 200? If I want to collect the body, where do I do that?

Also, I recall stream resolves its promise before the handler in pipeline is called. That got me scratching my head.

ronag:
@naugtur is there anything further to do here or can we close?

naugtur:
I was hoping we'd weed out the unwanted examples and keep the ones that are a good recommendation. 

Let me know which you'd like to keep soni can clean up or close if you don't want piping examples.

ronag:
> Would personally probably prefer `undici.request` variant.
> 
> ```js
> const response = await undici.request(
>   'https://avatars.githubusercontent.com/u/509375'
> )
> console.log(response.statusCode)
> const result = await undici.request('http://localhost:4000/', {
>   method: 'POST',
>   body: response.body
> })
> ```

I still prefer the above.

ronag:
closing as stale

Issue #727, opened 2021-04-12
:80 :443 origins will fail


ronag:
```
agent.dispatch({ origin: 'http://foo:80' })
```

Will fail since `parseOrigin` removes :80 which causes the `opts.origin === this[kUrl].origin` check to fail.

ronag:
@dnlup care to take a look?

PR #733, opened 2021-04-12
fix: Don't check origin for Pool & Client

ronag:
Parsed origin will remove :80 and :443 which will make
the check fail. Resolving this will incur performance loss
so let's just remove it.

Fixes: https://github.com/nodejs/undici/issues/727


mcollina (approved):
lgtm but CI is failing

Issue #429, opened 2020-09-23
response content-length mismatch waits for timeout


ronag:
I'm not sure but I think it should be possible to detect a response content-length mismatch without waiting for timeout. Unclear why the parser does not detect this.

```js
test('response invalid content length', (t) => {
  t.plan(3)

  const server = createServer((req, res) => {
    res.writeHead(200, {
      'content-length': 10
    })
    res.end('123')
  })
  t.teardown(server.close.bind(server))
  server.listen(0, () => {
    const client = new Client(`http://localhost:${server.address().port}`)
    t.teardown(client.destroy.bind(client))

    client.on('disconnect', (err) => {
      t.strictEqual(err.code, 'UND_ERR_SOCKET')
    })

    client.request({
      path: '/',
      method: 'GET'
    }, (err, data) => {
      t.error(err)
      data.body
        .on('end', () => {
          t.fail()
        })
        .on('error', (err) => {
          t.strictEqual(err.code, 'UND_ERR_SOCKET')
        })
        .resume()
    })
  })
})
```

ronag:
Would be even better if we could throw a descriptive error, e.g. "response body content-length mismatch"

simone-sanfratello:
I'll work on that

ronag:
https://github.com/nodejs/node/pull/39133

PR #742, opened 2021-04-15
fix: return content length mismatch error

simone-sanfratello:
see https://github.com/nodejs/undici/issues/429

@ronag @mcollina this fix the issue but I'm not very happy with the solution, looks a little bit "patchy" to me, please have a look, then I'll go on on the "@todo"

I also expect to send the same error if the body received is smaller then the "content-length", now it send "UND_ERR_INFO"

Also, I think we should/could refactor `client.js` code to improve readability, for example moving the `Parser` class to its own file and rename properties and methods where it is not clear they are referring to request or response phase - in another PR of course.


mcollina:
Why are you not happy with the solution?

lib/client.js
```diff
@@ -941,6 +950,12 @@ function onSocketError (err) {
 }
 
 function onSocketEnd () {
+  // @todo move to util function?
+  if (this[kClient][kBodySize] > 0 && this[kClient][kResponseContentLength] > 0 &&
+    this[kClient][kBodySize] !== this[kClient][kResponseContentLength]) {
+    util.destroy(this, new ResponseContentLengthMismatchError('end'))
+    return
+  }
```
ronag, line 958: This check should be in onBody and onMessageComplete

simone-sanfratello:
> Why are you not happy with the solution?

because I added the "this[kResponseContentLength] = 0" to client, is there a better way to do that?

mcollina:
> > Why are you not happy with the solution?
> 
> 
> 
> because I added `this[kResponseContentLength] = 0` to `client` class, is there a better way to do that?

I don't think there is any different way to do that.

lib/client.js
```diff
@@ -157,6 +160,8 @@ class Client extends Dispatcher {
     this[kSocket] = null
     this[kPipelining] = pipelining != null ? pipelining : 1
     this[kMaxHeadersSize] = maxHeaderSize || 16384
+    this[kBodySize] = 0
+    this[kResponseContentLength] = 0
```
mcollina, line 164: make sure these are reset at the beginning of dispatch.

mcollina (changes requested):

simone-sanfratello:
thanks @mcollina @ronag for the instant review :smiley: 

I'll go on

lib/client.js
ronag: These can live on the parser instance and don’t need symbols.

lib/client.js
```diff
@@ -763,6 +770,8 @@ class Parser {
   onBody (buf) {
     const { client, socket, statusCode } = this
 
+    client[kBodySize] += buf.length
```
ronag, line 773: Put after socket.destroy chech

lib/client.js
ronag: Should probably also check for the strictcontentlength flag

dnlup:
+1 to move the parser into its own file

ronag:
@simone-sanfratello What's your timeline on this? Should we aim to land it before v4?

simone-sanfratello:
I'll probably close it in the weekend

lib/client.js
simone-sanfratello: That flag refers to the sending headers, I think we should rename it, because in the current situation there are 2 cases and both return an error:

- the server content-length is greater than the actual body => connection is closed by the server, client returns "UND_ERR_SOCKET" >> fix sending "UND_ERR_RESPONSE_CONTENT_LENGTH_MISMATCH"
- the server content-length is less than the actual body => connection is closed by the client, client returns "UND_ERR_INFO" >> ? fix sending "UND_ERR_RESPONSE_CONTENT_LENGTH_MISMATCH" 

We can add another flag to suppress these errors, because I think those flags have different meanings: as client we can control what we send, so if we see an error/warning on sending headers we can fix it, but we can't control what the server is sending.

I think the flags should be something like

```
requestStrictContentLength: 'warn' (or 'error' or 'silent')
responseStrictContentLength: 'warn' (or 'error' or 'silent')
```

simone-sanfratello: The only place the check works is in `onSocketEnd`, doing in `onMessage` and/or `onBody` don't work, and also break other tests.

I think we can't use the `strictContentLength` option, or another one, because the connection always ends with an error, so we can only send the right one instead of the generic

lib/client.js
simone-sanfratello: I moved the vars to `Parser`, is the reset still needed?

docs/api/Errors.md
```diff
@@ -7,15 +7,19 @@ You can find all the error objects inside the `errors` key.
...
+| `ClientClosedError`          |  `UND_ERR_CLOSED`                          | trying to use a closed client.                        |
+| `SocketError`                |  `UND_ERR_SOCKET`                          | there is an error with the socket.                    |
+| `NotSupportedError`          |  `UND_ERR_NOT_SUPPORTED`                   | encountered unsupported functionality.                |
+| `ResponseContentLength`      |  `UND_ERR_REQUEST_CONTENT_LENGTH_MISMATCH` | body does not match content-length header in request  |
+| `RequestContentLength`       |  `UND_ERR_RESPONSE_CONTENT_LENGTH_MISMATCH`| body does not match content-length header in response * |
+| `InformationalError`         |  `UND_ERR_INFO`                            | expected error with reason                            |
+| `TrailerMismatchError`       |  `UND_ERR_TRAILER_MISMATCH`                | trailers did not match specification                  |
+
+\* The `UND_ERR_RESPONSE_CONTENT_LENGTH_MISMATCH` is returned when the received body is less than the value in the received header `Content-Length`.
```
simone-sanfratello, line 24: Please note that at https://github.com/nodejs/undici/blob/e2a9ddc07d9e0ff6be25d243523b53d1d71a3e9e/deps/llhttp/include/llhttp.h

If we want to allow body larger than content-length, we probably should expose this option

docs/api/Errors.md
```diff
@@ -7,15 +7,19 @@ You can find all the error objects inside the `errors` key.
...
+| `SocketError`                |  `UND_ERR_SOCKET`                          | there is an error with the socket.                    |
+| `NotSupportedError`          |  `UND_ERR_NOT_SUPPORTED`                   | encountered unsupported functionality.                |
+| `ResponseContentLength`      |  `UND_ERR_REQUEST_CONTENT_LENGTH_MISMATCH` | body does not match content-length header in request  |
+| `RequestContentLength`       |  `UND_ERR_RESPONSE_CONTENT_LENGTH_MISMATCH`| body does not match content-length header in response * |
+| `InformationalError`         |  `UND_ERR_INFO`                            | expected error with reason                            |
+| `TrailerMismatchError`       |  `UND_ERR_TRAILER_MISMATCH`                | trailers did not match specification                  |
+
+\* The `UND_ERR_RESPONSE_CONTENT_LENGTH_MISMATCH` is returned when the received body is less than the value in the received header `Content-Length`.  
+If the server try to send a body larger than `Content-Length`, Undici closes the connection when the length is received and return an `UND_ERR_INFO`, to prevent a cache poisoning attack.
```
simone-sanfratello, line 25: Please note that at https://github.com/nodejs/undici/blob/e2a9ddc07d9e0ff6be25d243523b53d1d71a3e9e/deps/llhttp/include/llhttp.h

If we want to allow body larger than content-length, we probably should expose this option

lib/client.js
ronag: Doing it in onSocketEnd is unfortunately just plain wrong. It should basically only be call if the server unexpectedly closes the connection.

simone-sanfratello: I don't see other solutions

ronag: > doing in onMessage and/or onBody don't work, and also break other tests.

In what way doesn't it work?

lib/core/errors.js
```diff
@@ -88,13 +88,23 @@ class InformationalError extends UndiciError {
   }
 }
 
-class ContentLengthMismatchError extends UndiciError {
+class RequestContentLength extends UndiciError {
```
ronag, line 91: All other errors end with `Error`. Also doesn't feel/sound like a descriptive error like this.

lib/client.js
ronag: > and onSocketEnd is actually called when the server close the connection unexpectedly

Yes, and what if the server doesn't close the connection when there is a content length mismatch?

lib/core/errors.js
simone-sanfratello: I think it's worth having them apart, but I'm ok to merge them

docs/api/Errors.md
```diff
@@ -7,15 +7,19 @@ You can find all the error objects inside the `errors` key.
...
+| `InvalidReturnValueError`    |  `UND_ERR_INVALID_RETURN_VALUE`            | returned an invalid value.                            |
+| `RequestAbortedError`        |  `UND_ERR_ABORTED`                         | the request has been aborted by the user              |
+| `ClientDestroyedError`       |  `UND_ERR_DESTROYED`                       | trying to use a destroyed client.                     |
+| `ClientClosedError`          |  `UND_ERR_CLOSED`                          | trying to use a closed client.                        |
+| `SocketError`                |  `UND_ERR_SOCKET`                          | there is an error with the socket.                    |
+| `NotSupportedError`          |  `UND_ERR_NOT_SUPPORTED`                   | encountered unsupported functionality.                |
+| `ResponseContentLengthError` |  `UND_ERR_REQUEST_CONTENT_LENGTH_MISMATCH` | body does not match content-length header in request  |
+| `RequestContentLengthError`  |  `UND_ERR_RESPONSE_CONTENT_LENGTH_MISMATCH`| body does not match content-length header in response * |
```
ronag, line 20: req and res are mixed up here

types/errors.d.ts
```diff
@@ -45,10 +45,16 @@ declare namespace Errors {
...
   }
 
-  /** Body does not match content-length header. */
-  export class ContentLengthMismatchError extends UndiciError {
-    name: 'ContentLengthMismatchError';
-    code: 'UND_ERR_CONTENT_LENGTH_MISMATCH';
+  /** Request body does not match content-length header. */
+  export class RequestContentLengthError extends UndiciError {
+    name: 'ResponseContentLengthError';
+    code: 'UND_ERR_REQUEST_CONTENT_LENGTH_MISMATCH';
```
ronag, line 51: the err code name and the error name doesn't quite match due to `mismatch`.

lib/core/util.js
```diff
@@ -170,6 +170,37 @@ function isBuffer (buffer) {
...
+  return bodySize > 0 && contentLength > 0 &&
+    bodySize !== contentLength
+}
+
+function extractMeaningfulHeaders (rawHeaders) {
+  const headers = {
+    keepAlive: null,
+    trailers: null,
+    contentLength: 0
+  }
```
ronag, line 183: I'm a little worried this extra object has perf implications? Can't we just send in the parser as an arg and set them direcrtly?

lib/client.js
simone-sanfratello: > > doing in onMessage and/or onBody don't work, and also break other tests.
> 
> In what way doesn't it work?

It would take me some time to answer this, like re-implementing in that way. The main issue is that `onMessage` and `onBody` are not been called when the connection is closed

test/util.js
```diff
@@ -19,3 +19,84 @@ test('isStream', (t) => {
...
+      input: [
+        Buffer.from('Date'),
+        Buffer.from('Thu, 01 Jan 1970 00:00:01 GMT'),
+        Buffer.from('Connection'),
+        Buffer.from('keep-alive'),
+        Buffer.from('Keep-Alive'),
+        Buffer.from('timeout=5'),
+        Buffer.from('Content-Length'),
+        Buffer.from('2')
+      ],
```
ronag, line 54: Maybe easier with a `map(x => Buffer.from(x))`?

lib/client.js
simone-sanfratello: > > and onSocketEnd is actually called when the server close the connection unexpectedly
> 
> Yes, and what if the server doesn't close the connection when there is a content length mismatch?

Sorry, I don't get it. As far as I've seen, the `content-length mismatch` is caused by the server when it close the connection before sending the "promised" body

ronag: Ah, now I see where the confusion is. I’m talking about the case where a misbehaving server sends a larger body than expected.

ronag: Or if the server sends a shorter body without disconnecting

ronag: The disconnecting is a different error

simone-sanfratello: I've seen body can't be larger than content-length because of https://github.com/nodejs/undici/blob/e2a9ddc07d9e0ff6be25d243523b53d1d71a3e9e/deps/llhttp/include/llhttp.h - unless we want to add this option

if the server send a shorter body without disconnecting, the timeout error is sent

is that right? Am I missing something?

ronag: With keep-alive the server can give a body that doesn't match content-length.

ronag: > I've seen body can't be larger than content-length because of https://github.com/nodejs/undici/blob/e2a9ddc07d9e0ff6be25d243523b53d1d71a3e9e/deps/llhttp/include/llhttp.h - unless we want to add this option

I think you were intending to link to a specific line?

simone-sanfratello: yes, sorry https://github.com/nodejs/undici/blob/e2a9ddc07d9e0ff6be25d243523b53d1d71a3e9e/deps/llhttp/include/llhttp.h#L472

simone-sanfratello: > With keep-alive the server can give a body that doesn't match content-length without closing the connection.

ok, I'll work on that

lib/core/util.js
simone-sanfratello: I don't see much difference between instancing 3 vars instead of one object, btw the destructuration of the result is not needed, so I removed

simone-sanfratello: I'm ok to move back without this function

lib/client.js
simone-sanfratello: @ronag to me it's working fine also with keep-alive option  

https://github.com/nodejs/undici/pull/742/files#diff-3058feaff6f3b2902370ee08a7528340aa7a98e5b76063f0176cb61972fadbb4R347

lib/core/util.js
mcollina: It would be better if we could avoid it.

test/util.js
simone-sanfratello: removed

lib/core/util.js
simone-sanfratello: removed

ronag (changes requested):
Does llhttp error or how does the keep-alive work?

lib/client.js
```diff
@@ -450,6 +451,8 @@ class Parser {
     this.headersSize = 0
     this.headersMaxSize = client[kMaxHeadersSize]
     this.shouldKeepAlive = false
+    this.bodySize = 0
+    this.contentLength = 0
```
ronag, line 455:
```suggestion
    this.contentLength = null
```

lib/client.js
```diff
@@ -689,23 +689,28 @@ class Parser {
...
 
       if (!keepAlive && key.length === 10 && key.toString().toLowerCase() === 'keep-alive') {
         keepAlive = val
-        looking = !trailers
+        looking++
       } else if (!trailers && key.length === 7 && key.toString().toLowerCase() === 'trailer') {
         trailers = val
-        looking = !keepAlive
+        looking++
+      } else if (!this.contentLength && key.length === 14 && key.toString().toLowerCase() === 'content-length') {
```
ronag, line 710:
```suggestion
      } else if (this.contentLength == null && key.length === 14 && key.toString().toLowerCase() === 'content-length') {
```

test/content-length.js
```diff
@@ -261,7 +262,123 @@ test('response invalid content length with close', (t) => {
...
+      'content-length': 10,
+      'keep-alive': 'timeout=2s',
+      connection: 'keep-alive'
+    })
+    res.end('0123')
+  })
+  t.teardown(server.close.bind(server))
+  server.listen(0, () => {
+    const client = new Client(`http://localhost:${server.address().port}`, {
+      pipelining: 0
```
ronag, line 361: you need to use `pipelining > 0` or keep-alive won't work.

test/content-length.js
```diff
@@ -261,7 +262,123 @@ test('response invalid content length with close', (t) => {
...
+  })
+})
+
+test('should get the proper error on response content-length in header is greater than actual body received (keep-alive)', (t) => {
+  t.plan(5)
+
+  const server = createServer((req, res) => {
+    res.writeHead(200, {
+      'content-length': 10,
+      'keep-alive': 'timeout=2s',
```
ronag, line 353: Node already does this so you don't need to set it.

simone-sanfratello:
> I don't think keep-alive test is working here as intended. Unless lhttp errors you need to check `bodySize` in `onBody` or `onMessageComplete`.

Already tried, is not working as expected. I'm closing the PR

Issue #741, opened 2021-04-14
Expected /HTTP


ronag:
Flaky test :/ 

```js
 FAIL  test/client-stream.js
 ✖ Expected HTTP/

  test/client-stream.js                    
  314 |       return pt                    
  315 |     }, (err) => {                  
> 316 |       t.error(err)                 
      | --------^                          
  317 |       t.equal(pt.destroyed, true)  
  318 |     })                             
  319 |   })                               

  test: test/client-stream.js stream destroy if not readable
  stack: |
    test/client-stream.js:316:9
    lib/api/api-stream.js:150:14

  Error Origin:

      }
      util.destroy(socket, new HTTPParserError(message, constants.ERROR[ret]))
---------------------------^
    }
  }

  at:
    line: 564
    column: 28
    file: lib/client.js
    function: Parser.execute
  stack: |
    Parser.execute (lib/client.js:564:28)
  type: HTTPParserError
  code: HPE_INVALID_CONSTANT
  message: Expected HTTP/

```

ronag:
Added more info to parser error to easier debug this https://github.com/nodejs/undici/commit/7a750ba998cbc4c125f747876536655c2f7b7c02.

ronag:
@dnlup FYI

ronag:
Looks like some kind of interference between `Client` instances. There is no special payload here.

dnlup:
Just out of curiosity, can we try to avoid the usage of the `currentParser`  technique and go back to the previous implementation? What do you think?

dnlup:
We are either modifying memory regions we shouldn't or maybe using the wrong parser pointer?

szmarczak:
I can have a look a this. Just give me a few minutes.

dnlup:
Thank yo @szmarczak

szmarczak:
Are you able to reproduce this or it throws randomly?

dnlup:
> Are you able to reproduce this or it throws randomly?

It is random, and apparently, it is happening only in the CI if I am not mistaken. There is work going on to introduce fuzzing, which should help. I can't find it atm 😢

ronag:
This one happened locally for me. But haven’t been able to repro since.

ronag:
> Just out of curiosity, can we try to avoid the usage of the `currentParser` technique and go back to the previous implementation? What do you think?

Let’s try to go back to the Map but check benchmarks. @dnlup can you try that?

dnlup:
[...]

PR #743, opened 2021-04-15
fix: use wasm instance map

ronag:
Refs: https://github.com/nodejs/undici/issues/741

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -395,7 +395,7 @@ class HTTPParserError extends Error {
   }
 }
 
-let currentParser = null
+const parsers = new Map()
 const llhttp = new WebAssembly.Instance(mod, {
   env: {
     /* eslint-disable camelcase */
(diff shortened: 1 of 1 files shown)


ronag:
Need to consider benchmarks.

ronag:
Base on the CI benchmarks I can't see any detectable performance diff.

dnlup (approved):

szmarczak:
I don't think this makes a difference in regards to the issue.

You could achieve the same just by storing the pointer and reassuring that `p` is the same as `ptr`.
If `llhttp` doesn't do anything asynchronously, that should be always the case.

ronag:
> I don't think this makes a difference in regards to the issue.
> 
> You could achieve the same just by storing the pointer and reassuring that `p` is the same as `parser.ptr`.
> If `llhttp` doesn't do anything asynchronously, that should be always the case.

Good point. I could just add assertions.

PR #745, opened 2021-04-15
refactor: simplify wasm api

ronag:


deps/llhttp/src/api.c
```diff
@@ -48,40 +53,16 @@ const llhttp_settings_t wasm_settings = {
 
 llhttp_t* llhttp_alloc(llhttp_type_t type) {
   llhttp_t* parser = malloc(sizeof(llhttp_t));
+  memset(parser, 0, sizeof(llhttp_t));
```
dnlup, line 56: This should fill the memory with zeroes, right? Like, cleaning the memory before using/freeing it?

dnlup (approved):

deps/llhttp/src/api.c
ronag: Yes. It just zeroes it. Should not be strictly necessary but doesn't hurt.

PR #749, opened 2021-04-18
perf: use static buffer

ronag:


mcollina:
what's the perf benefit?

ronag:
Less memory usage and allocations

mcollina:
I mean, is it measurable?

Anyway, tests are failing.

ronag:
It

> I mean, is it measurable?
> 
> Anyway, tests are failing.

Socket high watermark is 16k. So it’s 16x number of connections -1 less memory used.

In theory I guess it could be faster since the memory address stays in CPU cache. But that’s not primary point.

Also simpler code IMO.

I’ll sort the tests.

lib/client.js
```diff
@@ -882,7 +872,15 @@ function onParserTimeout (parser) {
 
 function onSocketData (data) {
   const { [kParser]: parser } = this
-  parser.execute(data)
+
+  if (data.length <= currentBufferSize) {
+    parser.execute(data)
+  } else {
+    for (let n = 0; n < data.length; n += currentBufferSize) {
+      parser.execute(data.slice(n, currentBufferSize))
```
ronag, line 880:
```suggestion
      parser.execute(data.slice(n, n + currentBufferSize))
```

lib/client.js
```diff
@@ -882,7 +872,15 @@ function onParserTimeout (parser) {
 
 function onSocketData (data) {
   const { [kParser]: parser } = this
-  parser.execute(data)
+
+  if (data.length <= currentBufferSize) {
+    parser.execute(data)
+  } else {
+    for (let n = 0; n < data.length; n += currentBufferSize) {
```
ronag, line 879:
```suggestion
    for (let n = 0; data.length; n += currentBufferSize) {
```

lib/client.js
```diff
@@ -882,7 +872,15 @@ function onParserTimeout (parser) {
 
 function onSocketData (data) {
   const { [kParser]: parser } = this
-  parser.execute(data)
+
+  if (data.length <= currentBufferSize) {
+    parser.execute(data)
+  } else {
+    for (let n = 0; data.length; n += currentBufferSize) {
+      parser.execute(data.slice(n, n + currentBufferSize))
```
ronag, line 880:
```suggestion
      parser.execute(data.slice(0, currentBufferSize))
```

lib/client.js
```diff
@@ -882,7 +872,15 @@ function onParserTimeout (parser) {
 
 function onSocketData (data) {
   const { [kParser]: parser } = this
-  parser.execute(data)
+
+  if (data.length <= currentBufferSize) {
+    parser.execute(data)
+  } else {
+    for (let n = 0; data.length; n += currentBufferSize) {
```
ronag, line 879:
```suggestion
    while (data.length) {
```

ronag:
@mcollina PTAL

dnlup (approved):

Issue #569, opened 2021-02-26
Creates a ProxyAgent to forward all HTTP request to an HTTP proxy


mcollina:
Here is some example call

```js
import { request, setGlobalAgent, ProxyAgent } from 'undici'

setGlobalAgent(new ProxyAgent('https://my.proxy.agent/'))

const {
  statusCode,
  headers,
  trailers,
  body
  // send the request via the https://my.proxy.agent/ HTTP proxy
} = await request('http://localhost:3000/foo')

console.log('response received', statusCode)
console.log('headers', headers)

for await (const data of body) {
  console.log('data', data)
}

console.log('trailers', trailers)
```

You might need to create a `ProxyPool` (that inherits from `Pool`) as well.

Follow up from #568

PR #750, opened 2021-04-20
Add support to https thru an http tunnel (proxy)

Ricardonacif:
Hey, 

I didn't know if this was supposed to be working or not, but I added a failing spec to it and I'm willing to put in the effort to make it work. Apparently Undici does not support https requests thru an http tunnel (proxy). According to [this wiki page](https://en.wikipedia.org/wiki/HTTP_tunnel) and [the CONNECT specification](https://tools.ietf.org/html/rfc7231#page-30), a `CONNECT www.domain.com:443 HTTP/1.1\r\n` should be sent after TCP handshake, and the TLS handshake should happen as usual after that.

Again, I'm willing to add support for this feature if you guys agree to. I just need some high level direction on where this logic should be (I'm kinda new in writing HTTP clients).

### DIFF
--- a/test/proxy.js
+++ b/test/proxy.js
@@ -4,6 +4,9 @@ const { test } = require('tap')
 const { Client, Pool } = require('..')
 const { createServer } = require('http')
 const proxy = require('proxy')
+const { readFileSync } = require('fs')
+const { join } = require('path')
+const https = require('https')
 
 test('connect through proxy', async (t) => {
   t.plan(3)
@@ -117,6 +120,49 @@ test('connect through proxy (with pool)', async (t) => {
   pool.close()
 })
 
+test('connect through proxy to an https server', async (t) => {
+  t.plan(3)
+
+  const server = await buildHttpsServer()
+  const proxy = await buildProxy()
+
+  const serverUrl = `https://localhost:${server.address().port}`
+  const proxyUrl = `http://localhost:${proxy.address().port}`
+
+  server.on('request', (req, res) => {
+    t.equal(req.url, '/hello?foo=bar')
+    res.setHeader('content-type', 'application/json')
+    res.end(JSON.stringify({ hello: 'world' }))
+  })
+
+  const ca = readFileSync(join(__dirname, 'fixtures', 'ca.pem'), 'utf8')
+  const client = new Client(proxyUrl, {
+    tls: {
+      ca,
+      rejectUnauthorized: false,
+      maxCachedSessions: 1,
+      servername: 'agent1'
+    }
+  })
+
+  const response = await client.request({
+    method: 'GET',
+    path: serverUrl + '/hello?foo=bar'
+  })
+
+  response.body.setEncoding('utf8')
+  let data = ''
+  for await (const chunk of response.body) {
+    data += chunk
+  }
+  t.equal(response.statusCode, 200)
+  t.same(JSON.parse(data), { hello: 'world' })
+
+  server.close()
+  proxy.close()
+  client.close()
+})
+
 function buildServer () {
   return new Promise((resolve, reject) => {
     const server = createServer()
@@ -127,6 +173,17 @@ function buildServer () {
 function buildProxy () {
   return new Promise((resolve, reject) => {
     const server = proxy(createServer())
+    server.listen(0, () => resolve(server))
+  })
+}
+
+function buildHttpsServer () {
+  return new Promise((resolve, reject) => {
+    const options = {
+      key: readFileSync(join(__dirname, 'fixtures', 'key.pem'), 'utf8'),
+      cert: readFileSync(join(__dirname, 'fixtures', 'cert.pem'), 'utf8')
+    }
+    const server = https.createServer(options)
     server.listen(0, () => resolve(server))
   })
 }


test/proxy.js
```diff
@@ -117,6 +120,49 @@ test('connect through proxy (with pool)', async (t) => {
...
+  })
+
+  const ca = readFileSync(join(__dirname, 'fixtures', 'ca.pem'), 'utf8')
+  const client = new Client(proxyUrl, {
+    tls: {
+      ca,
+      rejectUnauthorized: false,
+      maxCachedSessions: 1,
+      servername: 'agent1'
+    }
```
ronag, line 145: I think you need to do the upgrade to tls yourself also please use `client.connect` instead of `client.request`.

ronag:
I think:

> a CONNECT www.domain.com:443 HTTP/1.1\r\n should be sent after TCP handshake, and the TLS handshake should happen as usual after that.

Ricardonacif:
Thx for the replies! I found a hacky way to make it work right now, just as a POC:

```js
const proxyUrl = `http://us-30m.geosurf.io:12844`

  const ca = readFileSync(join(__dirname, 'fixtures', 'ca.pem'), 'utf8')
  const client = new Client(proxyUrl)

  client.on('connect', () => {    

    setTimeout(function() {
      
      const options = {
        socket: socket,
        rejectUnauthorized: false,
        servername: 'www.google.com'
      }
      
      var conn = tls.connect( options, function() {
        if (conn.authorized) {
          console.log("Connection authorized by a Certificate Authority.");
        } else {
          console.log("Connection not authorized: " + conn.authorizationError)
        }

        conn.on('data', (data) => {
          console.log("chunk:")
          console.log(data.toString('ascii'))
        })
        
        conn.write("GET / HTTP/1.1\r\nconnection: keep-alive\r\nhost: www.google.com\r\n\r\n", "ascii")

      });
  
    }, 2000)
    

  })
  const response = await client.connect({
    method: 'GET',
    path: serverUrl,
    headers: {
      'proxy-authorization': `Basic ${Buffer.from('proxyusername:redacted').toString('base64')}`,
      host: 'www.google.com'
    }
  })
  
  await sleeper(99999999);
```
I also had to comment this out on the beginning of the request Request constructor (lib/core/request.js:25):
```
if (typeof path !== 'string') {
      throw new InvalidArgumentError('path must be a string')
    } 
    // else if (path[0] !== '/' && !(path.startsWith('http://') || path.startsWith('https://'))) {
    //   throw new InvalidArgumentError('path must be an absolute URL or start with a slash')
    // }

```

That worked. The question is, does it make sense to add http tunnel support into the lib? If so, how do we want the API to look like? Thoughts?

ronag:
https://github.com/nodejs/undici/issues/569

ronag:
What about:

```js
const proxyUrl = `http://us-30m.geosurf.io:12844`

const ca = readFileSync(join(__dirname, 'fixtures', 'ca.pem'), 'utf8')
const client = new Client(proxyUrl)

const { socket } = await client.connect({
  method: 'GET',
  path: serverUrl,
  headers: {
    'proxy-authorization': `Basic ${Buffer.from('proxyusername:redacted').toString('base64')}`,
    host: 'www.google.com'
  }
})

const conn = tls.connect({
  socket: socket,
  rejectUnauthorized: false,
  servername: 'www.google.com'
}, () => {
  if (conn.authorized) {
    console.log("Connection authorized by a Certificate Authority.");
  } else {
    console.log("Connection not authorized: " + conn.authorizationError)
  }

  conn.on('data', (data) => {
    console.log("chunk:")
    console.log(data.toString('ascii'))
  })
  
  conn.write("GET / HTTP/1.1\r\nconnection: keep-alive\r\nhost: www.google.com\r\n\r\n", "ascii")
});
```

Ricardonacif:
Cool, there I have the TLS socket as `conn`, but how would I go about using the higher level API `client.request({ method: 'GET', path: '/'})` from this point on? Right now, the `client.connect` method is the one that creates either the TCP socket  or the TLS socket depending on the protocol. We would need it to create the TCP socket first, and `onConnect` upgrade it to the TLS socket. Would it make sense to add an option to the `connect` method to do this?

ronag:
I'm a little unsure at the moment. This needs significant more thought.

ronag:
We need to add the possibility to pass a connect factory method in the options somehow.

Ricardonacif:
Cool. Let me know if you come up with an API, I'll keep studying the codebase.

ronag:
@Ricardonacif 

What do you think about something like this:

```js
const agent = new Agent(async ({ origin, servername, session }) => {
  const { socket } = await undici.connect(`http://us-30m.geosurf.io:12844`, {
    headers: {
      'proxy-authorization': `Basic ${Buffer.from('proxyusername:redacted').toString('base64')}`,
      host: 'www.google.com'
    }
  })
  
  return await tls.connect({
    socket: socket,
    rejectUnauthorized: false,
    servername
  })
})

const { statusCode } = await agent.request('www.google.com')

console.log(statusCode)
```

Basically we would have to change so that instead of just taking an `URL` the `Client`, `Pool` and `Agent` classes can also take a connection factory.

Ricardonacif:
You are fast lol. I'll take a look at your connector branch this weekend. Thanks!

ronag:
With #763 I think something like this could work:

```js
const { Duplex, finished } = require('stream')

function connect (opts, callback) {
  return new Duplex({
    construct (callback) {
      undici.connect(`http://us-30m.geosurf.io:12844`, {
        headers: {
          'proxy-authorization': `Basic ${Buffer.from('proxyusername:redacted').toString('base64')}`,
          host: servername
        }
      }, (err, data) => {
        if (err || this.destroyed) {
          callback(err)
        }
        this.socket = tls
          .connect({
            socket: data.socket
          })
          .on('secureConnect', callback)
          .on('error', callback)
      })
    },
    write (chunk, encoding, callback) {
      return this.socket.write(chunk, encoding, callback)
    },
    final (callback) {
      return this.socket.end(callback)
    },
    read () {
      const buf = this.socket.read()
      if (buf) {
        this.push(buf)
      } else {
        this.socket.once('readable', () => {
          this._read()
        })
      }
    },
    destroy (err, callback) {
      this.socket.destroy(err)
      finished(this.socket, callback)
    }
  })
}

const client = new Client('www.google.com', { connect })
const { statusCode } = client.request({ path: '/', method: 'GET' })
```

Ricardonacif:
Great, thanks a lot for it. I'll try to use it and see if I can contribute with the specs on the proper https://github.com/nodejs/undici/pull/763 PR. Feel free to close this one.

michaelrommel:
Dear Robert,
I have now spent quite a few hours with this approach of a Connector factory and I am stuck with how the `construct` function works.

The proxy that I am using behaves like the following:

- after establishing the tcp connection, I am sending the required `CONNECT www.google.com:443 HTTP/1.1\r\n\r\n` string.
- unfortunately this proxy answers with a string: `HTTP/1.1 200 Connection Established\r\nProxy-Agent: Zscaler/6.1\r\n\r\n`
- only after reading and consuming those bytes, I can upgrade the socket to a TLS socket, which is then a second socket different from the one I gave in a s argument to the `tls.connect` call.

This means I cannot immediately in the callback of the `undici.connect()` call upgrade the  socket. So I am trying to work around the fact that the Connector factory is assumed to return a socket, but this socket is in the case of using an http proxy to connect to an https endpoint the first normal `net.connect` socket and not yet the `tls.connect` socket and I am failing at that. If you have any idea/suggestion/code snippet how this can be done I would be extremely grateful.

But maybe, given the direction that Matteo proposed in https://github.com/nodejs/undici/issues/569, I should not even go in this direction?

Sorry for this newbie question. It is not very important, I can work around this with a different library, so if you have not the time, no problem. I just wanted to try to come up with a proxy solution or good example for undici and undici-fetch, but I realize, that I am not (yet) up for that task.

Issue #736, opened 2021-04-13
Change benchmarking framework


mcollina:
As @ronag posted in https://github.com/nodejs/undici/pull/729#issuecomment-818261572, a significant part of the benchmarks is spent __inside the `benchmark` library__, making our benchmarks less useful: 

![Screenshot 2021-04-13 at 11 05 48](https://user-images.githubusercontent.com/52195/114527369-35da2800-9c48-11eb-9b94-871bd28f7fc7.png)

In the past where we had similar issues, we used https://www.npmjs.com/package/cronometro from @ShogunPanda which we developed for another _really fast_ library (https://github.com/nearform/fast-jwt).

ronag:
Also would be nice to have results relative to a baseline (e.g. http) so that the computer performance for any specific run is less of a factor when comparing results.

ronag:
@ShogunPanda Would you be interested in looking into this?

ShogunPanda:
@ronag Sure, absolutely. I'll send a PR in the next few days.

ronag:
Related https://github.com/nodejs/undici/issues/715

ShogunPanda:
Done. See #751.

PR #751, opened 2021-04-20
chore: Replace benchmark with cronometro for benchmarking.

### LINKED ISSUE #715: Make benchmarks relative to node core
[Sometimes it's difficult to compare different benchmark runs](https://github.com/nodejs/undici/pull/712#issuecomment-816484774). I think it would be easier to do comparison if we also had benchmarks that show undici results relative to a baseline (i.e. node core http client).

ShogunPanda:
This PR closes #715 and #736 by replacing `benchmark` with `cronometro`.

I also refactored the benchmark NPM scripts to use `wait-on` instead of a flat timeout for better reliability.

### DIFF
--- a/benchmarks/index.js
+++ b/benchmarks/index.js
@@ -123,14 +123,14 @@ class SimpleRequest {
   }
 }
 
-function makeParallelsRequests (cb) {
+function makeParallelRequests (cb) {
   return Promise.all(Array.from(Array(parallelRequests)).map(() => new Promise(cb)))
 }
 
 cronometro(
   {
     'http - no agent' () {
-      return makeParallelsRequests(resolve => {
+      return makeParallelRequests(resolve => {
         http.get(httpNoAgent, res => {
           res
             .pipe(
@@ -145,7 +145,7 @@ cronometro(
       })
     },
     'http - keepalive' () {
-      return makeParallelsRequests(resolve => {
+      return makeParallelRequests(resolve => {
         http.get(httpOptions, res => {
           res
             .pipe(
@@ -160,7 +160,7 @@ cronometro(
       })
     },
     'http - keepalive - multiple sockets' () {
-      return makeParallelsRequests(resolve => {
+      return makeParallelRequests(resolve => {
         http.get(httpOptionsMultiSocket, res => {
           res
             .pipe(
@@ -175,7 +175,7 @@ cronometro(
       })
     },
     'undici - pipeline' () {
-      return makeParallelsRequests(resolve => {
+      return makeParallelRequests(resolve => {
         client
           .pipeline(undiciOptions, data => {
             return data.body
@@ -192,7 +192,7 @@ cronometro(
       })
     },
     'undici - request' () {
-      return makeParallelsRequests(resolve => {
+      return makeParallelRequests(resolve => {
         client.request(undiciOptions).then(({ body }) => {
           body
             .pipe(
@@ -207,7 +207,7 @@ cronometro(
       })
     },
     'undici - pool - request - multiple sockets' () {
-      return makeParallelsRequests(resolve => {
+      return makeParallelRequests(resolve => {
         pool.request(undiciOptions).then(({ body }) => {
           body
             .pipe(
@@ -222,7 +222,7 @@ cronometro(
       })
     },
     'undici - stream' () {
-      return makeParallelsRequests(resolve => {
+      return makeParallelRequests(resolve => {
         return client
           .stream(undiciOptions, () => {
             return new Writable({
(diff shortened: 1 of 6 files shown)


ShogunPanda:
@ronag If you run the new benchmark on the reference machine I'll update the README as well.

ronag (approved):
Excellent work!

ronag:
It's a bit weird that `http - keepalive` is slowest?

ShogunPanda:
To be honest I have no idea. I've checked the code and everything seems to be fine. I wonder if the socket buffers are filled up and thus performance decrease.

ronag:
What results do you get if you run it locally? Also what order are they run in? Could it be some kind of warm up thing? Should we run each test in a separate node instance?

ronag:
Also the multiple socket results seem weird. Maybe they always have. Do we need those? @mcollina

ShogunPanda:
Cronometro runs with "warmup" mode by default. For each test, it executes twice, the first time discarding results.
Tests are run in series and each is run in a separate worker thread so no interaction should be possible.

Here's my results:

```
╔════════════════════════════════════════════╤═════════╤═══════════════╤═══════════╤═════════════════════════╗
║ Slower tests                               │ Samples │        Result │ Tolerance │ Difference with slowest ║
╟────────────────────────────────────────────┼─────────┼───────────────┼───────────┼─────────────────────────╢
║ http - keepalive                           │     100 │  61.35 op/sec │  ± 1.08 % │                         ║
║ http - no agent                            │     100 │ 201.87 op/sec │  ± 2.03 % │ + 229.03 %              ║
║ http - keepalive - multiple sockets        │     100 │ 302.28 op/sec │  ± 1.92 % │ + 392.69 %              ║
║ undici - pipeline                          │     100 │ 377.43 op/sec │  ± 2.33 % │ + 515.18 %              ║
║ undici - pool - request - multiple sockets │     100 │ 392.54 op/sec │  ± 1.99 % │ + 539.80 %              ║
║ undici - request                           │     100 │ 395.92 op/sec │  ± 1.97 % │ + 545.32 %              ║
║ undici - stream                            │     100 │ 421.28 op/sec │  ± 1.91 % │ + 586.65 %              ║
║ undici - dispatch                          │     100 │ 426.49 op/sec │  ± 1.89 % │ + 595.14 %              ║
╟────────────────────────────────────────────┼─────────┼───────────────┼───────────┼─────────────────────────╢
║ Fastest test                               │ Samples │        Result │ Tolerance │ Difference with slowest ║
╟────────────────────────────────────────────┼─────────┼───────────────┼───────────┼─────────────────────────╢
║ undici - noop                              │     100 │ 538.81 op/sec │  ± 1.81 % │ + 778.21 %              ║
╚════════════════════════════════════════════╧═════════╧═══════════════╧═══════════╧═════════════════════════╝
```

ronag:
Maybe `http - no agent` should be called `http - no agent - multiple sockts`?

dnlup:
> It's a bit weird that `http - keepalive` is slowest?

That option uses 1 socket if I am not mistaken, so I suspect that requests are run sequentially in that case.

ronag:
> > It's a bit weird that `http - keepalive` is slowest?
> 
> That option uses 1 socket if I am not mistaken, so I suspect that requests are run sequentially in that case.

Yea, I think it's the `no agent` case which is misleading.

ronag:
@ShogunPanda can we split the benchmarks into multi socket and single socket ones? That way the relative values make more sense.

dnlup:
> > > It's a bit weird that `http - keepalive` is slowest?
> > 
> > 
> > That option uses 1 socket if I am not mistaken, so I suspect that requests are run sequentially in that case.
> 
> Yea, I think it's the `no agent` case which is misleading.

I think so too because it's using the default Agent behind the scenes.

dnlup (approved):
Awesome work

benchmarks/index.js
```diff
@@ -272,3 +122,141 @@ class SimpleRequest {
...
   }
 }
+
+function makeParallelRequests (cb) {
+  return Promise.all(Array.from(Array(parallelRequests)).map(() => new Promise(cb)))
+}
+
+cronometro(
+  {
+    'http - no agent' () {
```
ronag, line 132:
```suggestion
    'http - no keepalive - multiple sockets' () {
```

benchmarks/index.js
```diff
@@ -272,3 +122,141 @@ class SimpleRequest {
...
+      return makeParallelRequests(resolve => {
+        http.get(httpNoAgent, res => {
```
ronag, line 134:
```suggestion
        http.get(httpNoKeepAlive, res => {
```

Make an agent that has the `maxSockets` options but `keepAlive: false`

ronag:
Is it possible to skip the grouping headers? The fastest test is always in the bottom so it's just noise, no?

ShogunPanda:
> Is it possible to skip the grouping headers? The fastest test is always in the bottom so it's just noise, no?

Unfortunately not, that's fixed in cronometro.

I'll update benchmarks soon.

ShogunPanda:
Benchmarks updated. We have two different suites now. Let me know if I set everything correct.

dnlup (approved):

ronag (approved):

ronag:
Not related to this to this Pr per se. It’s interesting that the multi socket version has less diff between core and undici. Maybe the server should run cluster?

PR #752, opened 2021-04-20
bench improvements

ronag:


ronag:
@ShogunPanda

README.md
```diff
@@ -17,19 +17,32 @@ npm i undici
...
+### Connections 50
+
+| Slower tests            | Samples | Result        | Tolerance | Difference with slowest |
+|-------------------------|---------|---------------|-----------|-------------------------|
+| http - no keepalive     |      99 |  31.96 op/sec |  ± 0.96 % |                         |
+| http - keepalive        |      99 |  33.74 op/sec |  ± 0.88 % | + 5.56 %                |
+| undici - pipeline       |     100 |  54.90 op/sec |  ± 1.06 % | + 71.78 %               |
+| undici - request        |     100 |  82.67 op/sec |  ± 1.69 % | + 158.65 %              |
+| undici - stream         |     100 |  94.95 op/sec |  ± 1.49 % | + 197.06 %              |
+| undici - dispatch       |     100 | 110.69 op/sec |  ± 1.57 % | + 246.33 %              |
```
mcollina, line 45: This looks odd: how come 50 connections have 1/5 op/sec than 1 connection?

There is likely something wrong in our benchmark.

ronag: It’s how the bench are built. It’s counting “ops” however 50 connections do more per ops in order to fully saturate the connections.

ronag: I think I can probably fix that

ShogunPanda: Confirmed. Those are not requests but ops so indeed is misleading. 

@ronag You can probably copy the table production code from cronometro and work with raw results to apply the multiplication that was present in the old version of the benchmarks.

Or we can merge this and I'll take care of that. Let me know.

ShogunPanda:
LGTM!

README.md
ronag: I just made sure that the number of parallel requests are the same and then it's ok.

ronag:
@mcollina PTAL

mcollina:
the benchmark workflow in CI stalled and blocked all other PRs. There is something not working right now on this PR.

ronag:
@ShogunPanda do you think you could make us a custom printer for cronometro that prints it in a way we can just copy past into README? In particular, multiply the results by `parallelRequests` and printing `reqs/s` instead of `ops/s`?

ShogunPanda:
Sure, I definitely can. Once you land this (so benchmark code is fixed) I'll work on changing the printer.

ronag:
I'm still a little confused by the results:

The difference with slowest is significantly larger for the single socket case. Also in the multi socket case `request` is faster than `stream`.

However, I think this is as a good as I can get it at the moment.

ronag:
> Sure, I definitely can. Once you land this (so benchmark code is fixed) I'll work on changing the printer.

Excellent. I'll merge once Matteo approves.

dnlup (approved):

mcollina:
ci is failing :/

ronag:
> ci is failing :/

GitHub actions seems to be having problems.

mcollina (approved):
lgtm

PR #762, opened 2021-04-21
Fix issue with util.queueMicrotask and cjs loader

pgherveou:
👋 all. Thanks for this fantastic lib.
When I require undici inside Electron
and run this code
```js
require('undici/lib/core/util.js').queueMicrotask(() => console.log('hello'))
```
I get a Uncaught TypeError: Illegal invocation
Looks like this is related to how the code is loaded via the cjs/loader

### DIFF
--- a/lib/core/util.js
+++ b/lib/core/util.js
@@ -167,7 +167,7 @@ module.exports = {
   destroy,
   bodyLength,
   isBuffer,
-  queueMicrotask: global.queueMicrotask ? global.queueMicrotask.bind(global) : (cb => Promise.resolve()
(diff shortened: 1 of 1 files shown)


ronag (approved):

ronag (changes requested):
Please fix CI and a test would be great if possible.

ronag (approved):
LGTM, a test would still be nice

mcollina (approved):
lgtm

pgherveou:
@ronag not really sure how we can test it since the bug only occurs when the module is required in electron. 🤔

ronag:
> @ronag not really sure how we can test it since the bug only occurs when the module is required in electron. 🤔

Is there a headless electron we can run in the CI?

PR #763, opened 2021-04-22
feat: connect

ronag:
Adds a connector option to Client & Pool which allows users to implement
custom connection handling for e.g. tunneling or other transport protocols
such as quic.
    
Refs: https://github.com/nodejs/undici/pull/750


ronag:
@ShogunPanda @szmarczak

ronag:
This needs more thought and probably something for v5.

szmarczak:
Would it make sense to make the connect function async?

dnlup (approved):
Looks great

ronag:
> Would it make sense to make the connect function async?

Yea. I'm experimenting with that.

ronag:
My current thinking is that we **do not** make this part of public API just yet. I think it makes sense internally as a refactoring but we need to iterate over the API more before making it public.

ronag:
> Would it make sense to make the connect function async?

It's unfortunately a bit complicated... 😞

mcollina (commented):
+1 but it needs testing

ronag:
I think I'm happy with this now. @dnlup @mcollina PTAL on the implementation and API and I'll start working on tests tonight.

ShogunPanda:
LGTM!

ronag:
This needs:

- Fix `tls-session-reuse` test. Preferably in a way that does not use internal undici symbols. 
- Add tests for `connect` constructor argument.
- Add docs.

I won't work further on this at the moment. If someone else wants to finish it up please go ahead.

ronag:
@mcollina I've removed the public API changes and fixed the existing test. I think this should be possible to land as internal refactoring. PTAL.

mcollina (approved):
lgtm

PR #769, opened 2021-04-24
feat: Add callback context and redirect history 2

ronag:


ronag:
@ShogunPanda PTAL

lib/handler/redirect.js
```diff
@@ -3,18 +3,22 @@
...
+    this.maxRedirections = maxRedirections
     this.handler = handler
+    this.history = []
   }
 
-  onConnect (abort) {
+  onConnect (abort, context) {
     this.abort = abort
-    this.handler.onConnect(abort)
+    this.handler.onConnect(abort, { ...context, history: this.history })
```
ronag, line 21: I'm thinking maybe we should put `history` as a symbol we make public as to avoid collisions in the future?

ShogunPanda: We could, but I think the external API might look a little complicated.

ShogunPanda:
LGTM!

docs/api/Dispatcher.md
```diff
@@ -201,7 +201,7 @@ Returns: `void`
 
 #### Parameter: `DispatchHandlers`
 
-* **onConnect** `(abort: () => void) => void` - Invoked before request is dispatched on socket. May be invoked multiple times when a request is retried when the request at the head of the pipeline fails.
+* **onConnect** `(abort: () => void, context: Object) => void` - Invoked before request is dispatched on socket. May be invoked multiple times when a request is retried when the request at the head of the pipeline fails.
```
ronag, line 204:
```suggestion
* **onConnect** `(abort: () => void, context: object) => void` - Invoked before request is dispatched on socket. May be invoked multiple times when a request is retried when the request at the head of the pipeline fails.
```

docs/api/Agent.md
```diff
@@ -19,7 +19,7 @@ Returns: `Agent`
 Extends: [`ClientOptions`](docs/api/Pool.md#parameter-pooloptions)
 
 * **factory** `(origin: URL, opts: Object) => Dispatcher` - Default: `(origin, opts) => new Pool(origin, opts)` 
-* **maxRedirections** `Integer` - Default: `0`.
+* **maxRedirections** `Integer` - Default: `0`. The number of HTTP redirection to follow. Redirections history will be available in `context.history` in the response callback (or the response resolved value).
```
ronag, line 22:
```suggestion
* **maxRedirections** `Integer` - Default: `0`. The number of HTTP redirection to follow unless otherwise specified in `DispatchOptions`.
```

mcollina (approved):
lgtm

PR #770, opened 2021-04-24
perf: reuse redirect handler

ronag:


dnlup (approved):
I am ok with the changes. I don't understand where the handler is reused.

lib/handler/redirect.js
```diff
@@ -92,7 +92,10 @@ class RedirectHandler {
         See comment on onData method above for more detailed informations.
       */
 
-      this.agent.dispatch(this.opts, this.handler)
+      this.location = null
+      this.abort = null
+
+      this.agent.dispatch(this.opts, this)
```
ronag, line 98: @dnlup here it's reused. Currently it would create a new RedirectHandler in Agent.dispatch.

dnlup: Ahh, I missed it. Thanks.

mcollina (approved):
lgtm

Issue #766, opened 2021-04-23
client can emit disconnect without a preceding connect


ronag:
It should not be possible to emit `'disconnect'` without `'connect'`. However, it then becomes a problem of how to make connection errors visible... do we need another event? i.e. what do we do with socket errors that occur before `'connect'`.

ronag:
Ideas? @szmarczak @dnlup @ShogunPanda

dnlup:
I think maybe we could expose an error event.

ronag:
> I think maybe we could expose an error event.

Problem with that is that forces user to listen to it.

dnlup:
> > I think maybe we could expose an error event.
> 
> Problem with that is that forces user to listen to it.

That's true. This could happen anyway If we don't use disconnect and opt for another approach (being another event or a callback passed to `connect`).

dnlup:
The big problem that I see with the `error` event is the potential uncaught exception (or maybe that's what you meant and I didn't get it)

ronag:
Another solution is to emit connect when we start connecting.

dnlup:
> Another solution is to emit connect when we start connecting.

If it doesn't create confusion on timing, sounds good to me.

szmarczak:
I'm good with anything. The current solution seems fine to me.

ronag:
It is a problem for users that use the connect/disconnect events to count the number of active connection.

ShogunPanda:
I also second @dnlup - Let's emit error, being careful to emit only if there are listeners present.

ronag:
@mcollina wdyt?

mcollina:
+1

PR #771, opened 2021-04-24
fix: only emit disconnect if connect

ronag:
Only emit disconnect if has been connected, otherwise errconnect.

Fixes: https://github.com/nodejs/undici/issues/766

### DIFF
--- a/docs/api/Dispatcher.md
+++ b/docs/api/Dispatcher.md
@@ -744,6 +744,17 @@ Parameters:
 * **targets** `Array<Dispatcher>`
 * **error** `Error`
 
+### Event: `'errconnect'`
+
+Parameters:
+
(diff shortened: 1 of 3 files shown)


ronag:
@szmarczak @ShogunPanda

ronag:
I'm not too happy with the event name...

dnlup (approved):
I don't have a strong opinion about the event name. Using the `error` event would add the complexity of checking if a handler is registered. I am ok either way.

ShogunPanda:
I also don't like the event name as well. I would go with `connectionError`, `error:connect` or `error.connect`.
The rest LGTM.

ronag:
@mcollina

mcollina (approved):
lgtm

PR #772, opened 2021-04-24
feat: unref socket while idle

ronag:
Allow process shutdown once all requests have completed without
waiting for keepAliveTimeout.

### DIFF
--- a/test/client-unref.js
+++ b/test/client-unref.js
@@ -1,31 +1,55 @@
 'use strict'
 
-const { test } = require('tap')
+const tap = require('tap')
+const { createServer } = require('http')
+const { once } = require('events')
 const { Client } = require('..')
-const http = require('http')
+const { Worker, isMainThread, parentPort } = require('worker_threads')
 
-test('end process on idle', (t) => {
-  t.plan(2)
-
-  const server = http.createServer((req, res) => {
+// On the main thread, spawn a worker that immediately launches an http server.
+if (isMainThread) {
+  tap.test('client automatically closes itself when idle', async t => {
+    const worker = new Worker(__filename)
+    // fail the test if the worker fails
+    worker.on('exit', code => {
+      if (code !== 0) {
+        t.fail(`Worker stopped with exit code ${code}`)
+      }
+    })
+    // get server url from worker
+    const [ { url } ] = await once(worker, 'message')
+    // create undici client and dispatch a request
+    const client = new Client(url)
+    const { body } = await client.request({ path: '/', method: 'GET' })
+    body.resume()
+    // once the body has been read entirely set a timeout cb that will fail the
+    // tap test if it executes. It will execute if the client does not
+    // unreference itself as well.
+    await once(body, 'end')
+    setTimeout(() => {
+      t.fail('Client did not automatically unref')
+    }, 2e3).unref()
+    // clean up the server in the worker thread so the tap test can end
+    worker.postMessage('close_server')
+    // succesfully end the tests
+    t.end()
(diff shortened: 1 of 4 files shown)


ronag:
Needs test.

ronag:
@Ethan-Arrowood ping regarding your question on Slack.

mcollina (approved):
lgtm - having a test will be nice

ronag:
Added test which unfortunately fails. @mcollina do you know how to debug/check what root ref keeps the process alive?

ronag:
Maybe @jasnell has some tip on how to debug what ref is keep event loop alive?

Ethan-Arrowood:
This looks good! I'll try and add some more test cases later today.

mcollina:
have you tried https://github.com/mafintosh/why-is-node-running?

ronag:
Yea, so the code works. The problem is the test. I think the server and client needs to run in a separate process for it to work. FYI @Ethan-Arrowood

test/client-unref.js
```diff
@@ -0,0 +1,31 @@
...
+const { Client } = require('..')
+const http = require('http')
+
+test('end process on idle', (t) => {
+  t.plan(2)
+
+  const server = http.createServer((req, res) => {
+    res.end()
+  })
+  t.teardown(server.close.bind(server))
```
mcollina, line 13: you need to remove this and unref() the server instead

mcollina (changes requested):

test/client-unref.js
ronag: Tried that as well but it didn't help.

Ethan-Arrowood:
I'm trying a multi-process test now using worker threads. will report back shortly with what I come up with

test/client-unref.js
mcollina: ☹️

Ethan-Arrowood:
This is what I have so far: https://gist.github.com/Ethan-Arrowood/30341384a724f944a334b2cae7e4f739

Details in the gist title. Still hacking at it to see if I can get tap to play nicely.

ronag:
I think you should put the server into the worker, not the other way around.

Ethan-Arrowood:
worker_thread test PR is up against this branch (from my fork)

Ethan-Arrowood:
irregardless of the worker thread test. The existing connect-timeout test is failing with this new change. Any ideas? Do we need to keep the client referenced while the `request` call fails?

test/client-unref.js
```diff
@@ -0,0 +1,55 @@
...
+    body.resume()
+    // once the body has been read entirely set a timeout cb that will fail the
+    // tap test if it executes. It will execute if the client does not
+    // unreference itself as well.
+    await once(body, 'end')
+    setTimeout(() => {
+      t.fail('Client did not automatically unref')
+    }, 2e3).unref()
+    // clean up the server in the worker thread so the tap test can end
+    worker.postMessage('close_server')
```
ronag, line 33: The problem here is that when the server closes the socket will die, thus we are not actually testing the unref case here...? Maybe I was wrong and the client should be in the worker?

Ethan-Arrowood: good point. i'll revisit in a bit and try again.

ronag:
@Ethan-Arrowood I think I managed to fix it.

mcollina (approved):
lgtm

Ethan-Arrowood:
Which release will this be published in? 3 or 4?

ronag:
4

PR #781, opened 2021-04-27
feat(Client/Pool): Same origin redirect

### LINKED ISSUE #686: Add same origin redirect support for Client and Pool
See how it's done in `Agent`. If redirect then create a `RedirectHandler` wrapper around the original dispatch handler.

MetCoder95:
**WIP**

Closes #686 
This PR introduces support for redirect by using `RedirectHandler` on `Client` for support redirection.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -7,6 +7,7 @@ const net = require('net')
 const util = require('./core/util')
 const Request = require('./core/request')
 const Dispatcher = require('./dispatcher')
+const RedirectHandler = require('./handler/redirect')
 const {
   RequestContentLengthMismatchError,
   ResponseContentLengthMismatchError,
@@ -61,7 +62,8 @@ const {
   kHeadersTimeout,
   kBodyTimeout,
   kStrictContentLength,
-  kConnector
+  kConnector,
+  kMaxRedirections
 } = require('./core/symbols')
 
 class Client extends Dispatcher {
@@ -83,7 +85,8 @@ class Client extends Dispatcher {
     tls,
     strictContentLength,
     maxCachedSessions,
-    [kConnect]: connect
+    [kConnect]: connect,
+    maxRedirections
   } = {}) {
     super()
 
@@ -145,6 +148,11 @@ class Client extends Dispatcher {
 
     this[kUrl] = util.parseOrigin(url)
     this[kConnector] = connect || makeConnect({ tls, socketPath, maxCachedSessions })
+    
+    if (maxRedirections != null && (!Number.isInteger(maxRedirections) || maxRedirections < 0)) {
+      throw new InvalidArgumentError('maxRedirections must be a positive number')
+    }
+
     this[kSocket] = null
     this[kPipelining] = pipelining != null ? pipelining : 1
     this[kMaxHeadersSize] = maxHeaderSize || 16384
@@ -163,6 +171,7 @@ class Client extends Dispatcher {
     this[kBodyTimeout] = bodyTimeout != null ? bodyTimeout : 30e3
     this[kHeadersTimeout] = headersTimeout != null ? headersTimeout : 30e3
     this[kStrictContentLength] = strictContentLength == null ? true : strictContentLength
+    this[kMaxRedirections] = maxRedirections == null ? 0 : maxRedirections
 
     // kQueue is built up of 3 sections separated by
     // the kRunningIdx and kPendingIdx indices.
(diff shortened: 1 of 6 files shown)


lib/client.js
```diff
@@ -256,7 +264,10 @@ class Client extends Dispatcher {
     }
 
     try {
-      const request = new Request(opts, handler)
+      console.log('Executing request')
+      const request = this[kMaxRedirections] > 0
+        ? new Request(opts, new RedirectHandler(this, { ...opts, origin: this[kUrl] }, handler))
+        : new Request(opts, handler)
```
MetCoder95, line 270: cc: @ronag 
I would like to know if I'm following the correct path as when trying to test it, I'm falling into the following error:

```text
error; ClientClosedError: The client is closed
    at Client.dispatch (/Users/carlosgerardofuentesrios/Documents/Workspace/MetCoder95/undici/lib/client.js:277:15)
    at RedirectHandler.onComplete (/Users/carlosgerardofuentesrios/Documents/Workspace/MetCoder95/undici/lib/handler/redirect.js:101:23)
    at Request.onComplete (/Users/carlosgerardofuentesrios/Documents/Workspace/MetCoder95/undici/lib/core/request.js:159:27)
    at Parser.onMessageComplete (/Users/carlosgerardofuentesrios/Documents/Workspace/MetCoder95/undici/lib/client.js:837:15)
    at wasm_on_message_complete (/Users/carlosgerardofuentesrios/Documents/Workspace/MetCoder95/undici/lib/client.js:430:28)
    at null.<anonymous> (wasm://wasm/0002924a:0:1998)
    at null.<anonymous> (wasm://wasm/0002924a:0:6386)
    at null.<anonymous> (wasm://wasm/0002924a:0:3006)
    at llhttp_execute (wasm://wasm/0002924a:0:1055)
    at Parser.execute (/Users/carlosgerardofuentesrios/Documents/Workspace/MetCoder95/undici/lib/client.js:528:32) {
    code: 'UND_ERR_CLOSED'
}
```

lib/client.js
```diff
@@ -256,7 +264,10 @@ class Client extends Dispatcher {
     }
 
     try {
-      const request = new Request(opts, handler)
+      console.log('Executing request')
+      const request = this[kMaxRedirections] > 0
+        ? new Request(opts, new RedirectHandler(this, { ...opts, origin: this[kUrl] }, handler))
```
MetCoder95, line 269: I also have to inject `origin` based on the sanitized `url` in order to construct the `request.history` properly.

MetCoder95:
Also, I'm not 100% sure, but I think that adding support for `maxRedirections` on the `Client` is enough to propagate the support for the `Pool` object. This due to the fact that the `Pool` creates `Client` instances when building the pool.

lib/client.js
MetCoder95: Any help? 😅 
I'm running out of options, not sure what's causing the client to be `closed` when the `RedirectHandler` tries to set dispatch the new request. Following the stack trace I only see the happy path once the request is done, but ran out of ideas what's closing the client

MetCoder95: Found that the error is mostly in the testing part, I'll continue with it soon 👍

lib/client.js
```diff
@@ -234,7 +243,10 @@ class Client extends Dispatcher {
     }
 
     try {
-      const request = new Request(opts, handler)
+      console.log('Executing request')
```
mcollina, line 246: I think this should be removed

test/client-dispatch.js
```diff
@@ -124,6 +124,69 @@ test('basic dispatch get', (t) => {
...
+        console.log(buf.toString('utf-8'))
+        bufs.push(buf)
+      },
+      onComplete (trailers) {
+        t.equal(trailers, null)
+        t.equal('hello world!', Buffer.concat(bufs).toString('utf8'))
+      },
+      onError (err) {
+        console.log('error;', err)
+        t.fail()
```
mcollina, line 184: replace this with `t.error(err)` and remove the console.log

mcollina (approved):
lgtm

lib/client.js
ronag: I'm not sure I follow this.

ronag (approved):

ronag:
@mcollina @MetCoder95 I finished up this PR. PTAL.

mcollina (approved):
lgtm

MetCoder95:
LGTM, thanks for taking care of the missing parts! 🙂

PR #784, opened 2021-04-29
fix: end message on connection close

ronag:
If response is not keep alive and socket ends then consider it as end of message.

I'm not sure whether this is spec compliant but I believe it would resolve the issue @Ethan-Arrowood encountered.

Refs: https://github.com/nodejs/undici/discussions/782

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -794,9 +794,9 @@ class Parser {
   }
 
   onMessageComplete () {
-    const { client, socket, statusCode, upgrade, trailers, headers: rawTrailers } = this
+    const { client, socket, statusCode, upgrade, trailers, headers: rawTrailers, shouldKeepAlive } = this
 
-    if (socket.destroyed) {
+    if (socket.destroyed && (!statusCode || shouldKeepAlive)) {
       return -1
     }
 
@@ -863,6 +863,10 @@ class Parser {
       resume(client)
     }
   }
+
+  finish () {
+    llhttp.exports.llhttp_finish(this.ptr)
+  }
 }
 
 function onParserTimeout (parser) {
@@ -901,6 +905,16 @@ function onSocketConnect () {
 }
 
 function onSocketError (err) {
+  const { [kParser]: parser } = this
+  if (err.code === 'ECONNRESET' && parser.statusCode && !parser.shouldKeepAlive) {
+    // Response does not contain content-length nor chunked encoding.
+    // We treat all incoming data so for as a valid response.
+    currentParser = parser
+    parser.finish()
+    currentParser = null
+    return
+  }
+
   const { [kClient]: client } = this
 
   this[kError] = err
@@ -936,8 +950,9 @@ function onSocketEnd () {
 
   if (parser.statusCode && !parser.shouldKeepAlive) {
     // Response does not contain content-length nor chunked encoding.
-    parser.onMessageComplete()
-    util.destroy()
+    currentParser = parser
+    parser.finish()
+    currentParser = null
     return
   }
 
--- a/lib/client.js
+++ b/lib/client.js
@@ -932,6 +932,13 @@ function onSocketError (err) {
 }
 
 function onSocketEnd () {
+  const { [kParser]: parser } = this
+
+  if (this.statusCode && !this.shouldKeepAlive) {
+    // Response does not contain content-length nor chunked encoding.
+    parser.onMessageComplete()
+  }
+
   util.destroy(this, new SocketError('other side closed'))
 }
 
--- a/lib/client.js
+++ b/lib/client.js
@@ -906,6 +906,9 @@ function onSocketConnect () {
 
 function onSocketError (err) {
   const { [kParser]: parser } = this
+
+  // On Mac OS, we get an ECONNRESET even if there is a full body to be forwarded
+  // to the user.
   if (err.code === 'ECONNRESET' && parser.statusCode && !parser.shouldKeepAlive) {
     // Response does not contain content-length nor chunked encoding.
     // We treat all incoming data so for as a valid response.
--- a/lib/client.js
+++ b/lib/client.js
@@ -541,6 +541,12 @@ class Parser {
     }
   }
 
+  finish () {
+    currentParser = this
+    llhttp.exports.llhttp_finish(this.ptr)
+    currentParser = null
+  }
+
   destroy () {
     assert(this.ptr != null)
     assert(currentParser == null)
@@ -863,10 +869,6 @@ class Parser {
       resume(client)
     }
   }
-
-  finish () {
-    llhttp.exports.llhttp_finish(this.ptr)
-  }
 }
 
 function onParserTimeout (parser) {
@@ -910,11 +912,8 @@ function onSocketError (err) {
   // On Mac OS, we get an ECONNRESET even if there is a full body to be forwarded
   // to the user.
   if (err.code === 'ECONNRESET' && parser.statusCode && !parser.shouldKeepAlive) {
-    // Response does not contain content-length nor chunked encoding.
     // We treat all incoming data so for as a valid response.
-    currentParser = parser
     parser.finish()
-    currentParser = null
     return
   }
 
@@ -952,10 +951,8 @@ function onSocketEnd () {
   const { [kParser]: parser } = this
 
   if (parser.statusCode && !parser.shouldKeepAlive) {
-    // Response does not contain content-length nor chunked encoding.
-    currentParser = parser
+    // We treat all incoming data so for as a valid response.
     parser.finish()
-    currentParser = null
     return
   }
 
--- a/lib/client.js
+++ b/lib/client.js
@@ -934,9 +934,11 @@ function onSocketError (err) {
 function onSocketEnd () {
   const { [kParser]: parser } = this
 
-  if (this.statusCode && !this.shouldKeepAlive) {
+  if (parser.statusCode && !parser.shouldKeepAlive) {
     // Response does not contain content-length nor chunked encoding.
     parser.onMessageComplete()
+    util.destroy()
+    return
   }
 
   util.destroy(this, new SocketError('other side closed'))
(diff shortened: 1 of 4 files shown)


ronag:
@Ethan-Arrowood if you agree with this would you mind helping with creating a test?

lib/client.js
```diff
@@ -932,6 +932,13 @@ function onSocketError (err) {
 }
 
 function onSocketEnd () {
+  const { [kParser]: parser } = this
+
+  if (this.statusCode && !this.shouldKeepAlive) {
+    // Response does not contain content-length nor chunked encoding.
+    parser.onMessageComplete()
+  }
```
ronag, line 940: This might or might not be an expected end of message. We have no way to tell at the moment?

ronag:
I actually disagree with this change myself but leaving it here for visibility and discussion.

Ethan-Arrowood:
Before I can form a real opinion here I'd like to simplify and reproduce. We should try and emulate with just a node.js http server and see what happens. I'm not well versed enough in http spec to know exactly what should happen either.

mcollina:
I think we should land this change, it's what users would expect.

There is a HTTP/1.1 spec violation that we should document as there is no `content-length` and no `transfer-encoding: chunked` specified.

ronag:
> I think we should land this change, it's what users would expect.

Just so we are all in agreement. There is actually no way to tell whether we properly received the full body. So in theory the user could receive an incomplete body without error. Are we ok with that?

ronag:
I think there **might** be a trailing `\r\n` in the last packet before the socket end. Maybe `llhttp` could consider that as eof for non persistent connections? @indutny

dnlup (commented):
Right now, I am not sure about it either. I have read the discussion, and I think this would be necessary only if `lhttp` doesn't call `onMessageComplete` in this scenario. I need to see what Node core does in this case. That might give some inspiration.

mcollina:
I stumbled upon this as well. Can we land it?

mcollina (approved):
lgtm

ronag:
> I stumbled upon this as well. Can we land it?

I'm not comfortable landing it in it's current form. Would like to see if there is any \r\n we can detect and use.

mcollina:
I'll investigate further and try to reproduce with Node.

ronag:
Also would be good to have a separate test for the case where it's keep alive and whether or not that must error.

ronag:
another thing to consider is whether to only allow this behavior for failure status codes?

indutny:
Could unidici just call `llhttp_finish()`?

mcollina:
@ronag I pushed a test and a fix.

I had to disable another test which is mutually exclusive as far as I understand. I think we should be lenient in this case.

lib/client.js
```diff
@@ -932,6 +932,15 @@ function onSocketError (err) {
 }
 
 function onSocketEnd () {
+  const { [kParser]: parser } = this
+
+  if (parser.statusCode && !parser.shouldKeepAlive) {
+    // Response does not contain content-length nor chunked encoding.
+    parser.onMessageComplete()
+    util.destroy()
```
ronag, line 940: This doesn’t do anything.

ronag (commented):
What about llhttp_finish?

mcollina:
> What about llhttp_finish?

I'm investigating it.

test/client-request.js
```diff
@@ -97,3 +99,46 @@ test('trailers', (t) => {
...
+test('destroy socket abruptly', async (t) => {
+  t.plan(2)
+
+  const server = net.createServer((socket) => {
+    const lines = [
+      'HTTP/1.1 200 OK',
+      'Date: Sat, 09 Oct 2010 14:28:02 GMT',
+      'Connection: close',
+      '',
+      'the body'
```
ronag, line 112: `\r\n`?

test/client-request.js
```diff
@@ -97,3 +99,46 @@ test('trailers', (t) => {
...
 })
+
+test('destroy socket abruptly', async (t) => {
+  t.plan(2)
+
+  const server = net.createServer((socket) => {
+    const lines = [
+      'HTTP/1.1 200 OK',
+      'Date: Sat, 09 Oct 2010 14:28:02 GMT',
+      'Connection: close',
```
ronag, line 110: Can we add another test where this is `Connection: keep-alive`?

mcollina:
Using llhtp_finish fixed most problems indeed.

lib/client.js
```diff
@@ -901,6 +905,16 @@ function onSocketConnect () {
 }
 
 function onSocketError (err) {
+  const { [kParser]: parser } = this
+  if (err.code === 'ECONNRESET' && parser.statusCode && !parser.shouldKeepAlive) {
+    // Response does not contain content-length nor chunked encoding.
+    // We treat all incoming data so for as a valid response.
+    currentParser = parser
+    parser.finish()
+    currentParser = null
```
ronag, line 914: Cna we move the currentParser stuff into `parser.finish()`?

lib/client.js
```diff
@@ -901,6 +905,16 @@ function onSocketConnect () {
 }
 
 function onSocketError (err) {
+  const { [kParser]: parser } = this
+  if (err.code === 'ECONNRESET' && parser.statusCode && !parser.shouldKeepAlive) {
```
ronag, line 909:
```suggestion
  if (err.code === 'ECONNRESET' && parser.statusCode >= 400 && !parser.shouldKeepAlive) {
```

lib/client.js
```diff
@@ -794,9 +794,9 @@ class Parser {
   }
 
   onMessageComplete () {
-    const { client, socket, statusCode, upgrade, trailers, headers: rawTrailers } = this
+    const { client, socket, statusCode, upgrade, trailers, headers: rawTrailers, shouldKeepAlive } = this
 
-    if (socket.destroyed) {
+    if (socket.destroyed && (!statusCode || shouldKeepAlive)) {
```
ronag, line 799:
```suggestion
    if (socket.destroyed && (!statusCode || statusCode < 400 || shouldKeepAlive)) {
```

lib/client.js
ronag: Not sure about the `400`. Adding here as topic for discussion.

lib/client.js
mcollina: I've followed the same pattern we are using elsewhere, we set `currentParser` outside of the parser class.

ronag: No we don't? We set it inside `execute` which is part of the parser class.

test/client-request.js
mcollina: Adding `\r\n` will show them in the end body.

lib/client.js
ronag, line 909: Do we need this here? Isn't it enough to apply this logic on "graceful" socket shutdown?

mcollina: why would the status code make any difference?

mcollina: It's really OS-dependent. We get an ECONNRESET on Mac and a normal shutdown on Linux.

test/client-request.js
```diff
@@ -97,3 +99,85 @@ test('trailers', (t) => {
...
+      'Date: Sat, 09 Oct 2010 14:28:02 GMT',
+      'Connection: keep-alive',
+      '',
+      'the body'
+    ]
+    socket.end(lines.join('\r\n'))
+
+    // Unfortunately calling destroy synchronously might get us flaky results,
+    // therefore we delay it to the next event loop run.
+    socket.destroy()
```
ronag, line 159: Missing setImmediate?

test/client-request.js
```diff
@@ -97,3 +99,85 @@ test('trailers', (t) => {
...
+  let actual = ''
+
+  for await (const chunk of body) {
+    actual += chunk
+  }
+
+  t.equal(actual, 'the body')
+})
+
+test('destroy socket abruptly with keep-alive', async (t) => {
```
ronag, line 144: Shouldn't this test error?

lib/client.js
```diff
@@ -863,6 +863,10 @@ class Parser {
       resume(client)
     }
   }
+
+  finish () {
+    llhttp.exports.llhttp_finish(this.ptr)
+  }
```
ronag, line 869: Can you put this before `destroy()`? I've tried to put all the callback methods at the end of the class and all regular methods before that.

lib/client.js
ronag, line 799: This might cause problems. I need to think a little. The current invariant is that no callbacks are invoked while `socket.destroyed`.

test/client-request.js
mcollina: We are not erroring right now because shouldKeepAlive is coming as false from llhttp.

lib/client.js
mcollina: This is needed as the socket is already destroyed due to the incoming error. I added the check on the statusCode and shouldKeepAlive instead of removing because otherwise https://github.com/nodejs/undici/blob/main/test/client-idempotent-body.js was failing.

ronag: Yea I get that. I'll just give it some thought to ensure it doens't have any unforseen side effect.

test/client-request.js
ronag: > shouldKeepAlive is coming as false from llhttp.

Why? That sounds wierd?

ronag:
LGTM, apart from the comment for the keep-alive test.

test/client-request.js
mcollina: This `shouldKeepAlive` https://github.com/nodejs/undici/blob/e61c2525a73f137a519c43cd6ada7c0334cdee04/lib/client.js#L402 is coming as `0`. Maybe @indutny can help.

mcollina: I've found the problem, content-length was missing. pushing a fix.

ronag:
LGTM, will make a rc2 release?

mcollina:
As soon as the test pass! 🎉🎉

Issue #783, opened 2021-04-29
Add maxCachedSessions to `connect`


ronag:
Currently there is no limit on the number of cached sessions. Implement a `maxCachedSessions` option similar to [node core](https://nodejs.org/api/https.html#https_new_agent_options) in [`connect`](https://github.com/nodejs/undici/blob/main/lib/core/connect.js).

PR #786, opened 2021-05-03
handle maxCachedSessions option in connect class

KiraPC:
Issue: https://github.com/nodejs/undici/issues/783

As per the issue description I've added the implementation to handle the maxCachedSessions option, inspired by the node http.Agent.

I would like to discuss about how I handled the default and the zero value.

- Is it ok to set the default to 100 as node does in its Agent?
- Is it ok to disable the session reusability if the value of this option is zero, even if you can disable it reuseSessions option?

Thanks.

### DIFF
--- a/docs/api/Client.md
+++ b/docs/api/Client.md
@@ -27,6 +27,7 @@ Returns: `Client`
 * **socketPath** `string | null` (optional) - Default: `null` - An IPC endpoint, either Unix domain socket or Windows named pipe.
 * **tls** `TlsOptions | null` (optional) - Default: `null` - An options object which in the case of `https` will be passed to [`tls.connect`](https://nodejs.org/api/tls.html#tls_tls_connect_options_callback).
 * **strictContentLength** `Boolean` (optional) - Default: `true` - Whether to treat request content length mismatches as errors. If true, an error is thrown when the request content-length header doesn't match the length of the request body.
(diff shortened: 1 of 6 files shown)


mcollina:
Can you add docs?

How does Node.js behave when the option is set to zero?

KiraPC:
@mcollina Yes, sure, I'll docs asap.

node Agent disable the caching too (I was inspired by them).

My question is, do we should handle the option reuseSessions if the behavior is the same?

ronag (changes requested):
I don’t think the extra list is necessary.

mcollina:
I think we should just remove reuseSessions.

KiraPC:
> I don’t think the extra list is necessary.

Hi, 

I introduced that list to remove the oldest session. So, if the cached sessions are more then configured, I get the first session cached and remove it from the map.

How can I do the same thing with just the Map object?

ronag:
> > I don’t think the extra list is necessary.
> 
> Hi,
> 
> I introduced that list to remove the oldest session. So, if the cached sessions are more then configured, I get the first session cached and remove it from the map.
> 
> How can I do the same thing with just the Map object?

Create an iterator and call next()

KiraPC:
@mcollina Added docs and types and removed reuseSessions (this last was never documented, so I just removed it from code).

@ronag Done with map.keys, thanks for the suggestion.

I also moved the option out from tls option object because [tls.options](https://nodejs.org/api/tls.html#tls_tls_connect_options_callback) doesn't include it.

mcollina (approved):
lgtm

PR #789, opened 2021-05-06
fix: ensure callbacks use micro task queue

ronag:
Ensure that we have proper ordering relative to promises. This only affects cold paths.


mcollina (approved):
lgtm

Issue #787, opened 2021-05-04
Not matching mocks are hard to debug


mcollina:
## Bug Description

When a mock does not match (and net.connect is disabled), we error with a generic:

```
Unable to find mock dispatch and real dispatches are disabled for XYZ
```

However, we are likely setting up a mock for this URL - so there might be something that is not matching.

We should provide better error messages.

mcollina:
cc @jonnydgreen

jonnydgreen:
Makes sense, I can get started on a solution for this this evening for improving the mock errors!

mcollina:
Thanks!

jonnydgreen:
No problem at all! :)

jonnydgreen:
Thinking of something like this which indicates exactly what the issue is and offers a solution as well:

```js
...
if (netConnect === false) {
  throw new MockNotMatchedError(`Request to ${origin} does not match any registered mock dispatches and net.connect is disabled. Please register a mock dispatch for this origin.`)
}
throw new MockNotMatchedError(`Request to ${origin} does not match any registered mock dispatches and net.connect is not enabled for this origin. Please register a mock dispatch for this origin or add this origin to your MockAgent.enableNetConnect matchers.`)
...
```

What do you reckon?

mcollina:
It would work but it needs to be more nuanced. For example, express that there was a body or header or XYZ mismatch.

jonnydgreen:
Ah okay, no problem - I'll get right on that and try and think of other situations that need more info as well!

PR #790, opened 2021-05-06
Improve mock errors

jonnydgreen:
Fixes: https://github.com/nodejs/undici/issues/787

As suggested, I've added the following improvements to help debug when a mock dispatch is not matched

For a missing path:

```txt
Mock dispatch not matched for path '/wrong': subsequent request to origin http://localhost:9999 was not allowed (net.connect is not enabled for this origin)
Mock dispatch not matched for path '/wrong': subsequent request to origin http://localhost:9999 was not allowed (net.connect is disabled)
```

For a missing method:

```txt
Mock dispatch not matched for method 'WRONG': subsequent request to origin http://localhost:9999 was not allowed (net.connect is not enabled for this origin)
Mock dispatch not matched for method 'WRONG': subsequent request to origin http://localhost:9999 was not allowed (net.connect is disabled)
```

For a missing body:

```txt
Mock dispatch not matched for body 'wrong': subsequent request to origin http://localhost:9999 was not allowed (net.connect is not enabled for this origin)
Mock dispatch not matched for body 'wrong': subsequent request to origin http://localhost:9999 was not allowed (net.connect is disabled)
```

I've also exposed the `mockErrors` as a separate object to `errors` so they are not confused, happy to refactor if you would prefer them to be all together :)

### DIFF
--- a/index.js
+++ b/index.js
@@ -11,6 +11,7 @@ const api = require('./lib/api')
 const MockClient = require('./lib/mock/mock-client')
 const MockAgent = require('./lib/mock/mock-agent')
 const MockPool = require('./lib/mock/mock-pool')
+const mockErrors = require('./lib/mock/mock-errors')
 
 Object.assign(Dispatcher.prototype, api)
 
@@ -90,3 +91,4 @@ module.exports.upgrade = makeDispatcher(api.upgrade)
 module.exports.MockClient = MockClient
 module.exports.MockPool = MockPool
 module.exports.MockAgent = MockAgent
+module.exports.mockErrors = mockErrors
--- a/docs/api/MockErrors.md
+++ b/docs/api/MockErrors.md
@@ -1,0 +1,7 @@
+# MockErrors
+
+Undici uses a variety of mock error objects.
+
+| Mock Error            | Mock Error Codes                | Description                                                |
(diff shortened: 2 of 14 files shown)


index.js
```diff
@@ -90,3 +91,4 @@ module.exports.upgrade = makeDispatcher(api.upgrade)
 module.exports.MockClient = MockClient
 module.exports.MockPool = MockPool
 module.exports.MockAgent = MockAgent
+module.exports.mockErrors = mockErrors
```
jonnydgreen, line 94: I've exposed the `mockErrors` as a separate object to `errors` so they are not confused, happy to refactor if you would prefer them to be all together :)

mcollina (approved):
Good work!

I think we should add some check/assertions on Headers as well. I just realised they are not supported.

jonnydgreen:
Thanks! Just to clarify, when you say check/assertions on Headers as well, do you mean match on mock dispatch headers? For example:

```js
mockPool.intercept({
  path: '/foo',
  method: 'GET',
  headers: ['hello', 'there'] // Only a request for this origin,path,method,headers will be intercepted
}).reply(200, 'foo')
```

If so, I can do that in a separate PR over the weekend if that works for you? :)

mcollina:
That's perfect thanks.

PR #792, opened 2021-05-08
fix: some cleanup

ronag:
- Avoid extra loop for getting keep-alive and trailer headers.
- headers in dispatch callbacks is always an array.


dnlup (commented):
I like these changes; lgtm when the ci is green.

mcollina (approved):
lgtm

Issue #791, opened 2021-05-07
Mocks should match headers


mcollina:
Currently we do not match the headers when doing intercepts:

```js
mockPool.intercept({
  path: '/foo',
  method: 'GET',
  headers: {
     'User-Agent': 'undici' // requests with different or missing user agent would not be matched.
  }
}).reply(200, 'foo')
````

I would only consider an header as a requirement if included. So passing (or not passing) an `Accept` header would not make a difference for this interceptor.

mcollina:
cc @jonnydgreen

mcollina:
@ronag this must be fixed before the v4 release.

jonnydgreen:
I'll get a PR sorted for this this weekend :)

PR #794, opened 2021-05-08
Add mock header matching support

jonnydgreen:
Closes: https://github.com/nodejs/undici/issues/791

This also adds the following features for the individual header checks:
 - Match headers by string
 - Match headers by regex
 - Match headers by function


mcollina (approved):
lgtm

PR #799, opened 2021-05-09
test: get with host header (IPv6)

trivikr:
Adds test for if block on [Line 94](https://codecov.io/gh/nodejs/undici/src/main/lib/core/util.js) of `lib/core/util.js`


mcollina (approved):
lgtm

mcollina (approved):
lgtm

Issue #803, opened 2021-05-11
Response body ends prematurely


ronag:
~~Haven't had time to check but I suspect we might have a problem with body sizes larger than 31 bit.~~

undici will complete without error before reading entire response body.

ronag:
It just works. This might be a wasm thing.

dnlup:
[Inspecting](https://github.com/nodejs/undici/blob/main/deps/llhttp/include/llhttp.h#L30) `llhttp` I see content-length is an unsigned int of 64 bits. That should exclude some issues in the parser native side.

indutny:
Isn't wasm memory space limited to 2gb? I think you need to split up the data before passing it into llhttp.

ronag:
The data is split?

indutny:
Looking. Completely unrelated, you should probably use `parseInt(..., 10)` instead of just `parseInt()` (Yes, JS is weird: `parseInt('0x1f') === 0x1f`)

indutny:
llhttp itself appears to work fine: https://gist.github.com/indutny/818e4428eeadb17bb541a668e653e2a9 . Whenever I execute it - I get back exactly the same amount of bytes that passed to execute.

I strongly suspect that it is either an issue with wasm build or the wrapping around it.

ronag:
@indutny Thanks for confirming. Much appreciated.

ronag:
@indutny I ported your test into our WebASsembly and was able to reproduce the issue: https://gist.github.com/ronag/3e316e78931f8b49f1138460a77a258a

indutny:
Added a bunch of logging and looks like `uint64_t` subtraction doesn't work:

```
parser.content_length=0x100028c68 on_body_len=65536 read_total=1605238784
parser.content_length=0x100018c68 on_body_len=65536 read_total=1605304320
parser.content_length=0x100008c68 on_body_len=65536 read_total=1605369856
parser.content_length=0x0 on_body_len=35944 read_total=1605405800
```

As you might have guessed 35944 is `0x8c68` in hex. So it subtracts 0x8c68 from 0x100008c68 and gets 0x0.

indutny:
FWIW, here's the code that I've added to `wasm_on_body` to extract this:

```js
      const _cl = new Uint32Array(llhttp.exports.memory.buffer, p + 8 * 4, 2);
      const _remaining = _cl[1] * 0x100000000 + _cl[0];
[...]

PR #805, opened 2021-05-11
fix: premature end large body

ronag:
llhttp integer truncation on ARM platform.

Refs: https://github.com/nodejs/undici/issues/803
Refs: https://github.com/nodejs/undici/issues/429


dnlup (approved):
Awesome work, lgtm.

dnlup:
I think there's just some `ts` linting that is failing

lib/client.js
```diff
@@ -834,6 +844,12 @@ class Parser {
       }
     }
 
+    /* istanbul ignore next: should be handled by llhttp? */
+    if (request.method !== 'HEAD' && contentLength && bytesRead !== parseInt(contentLength, 10)) {
+      util.destroy(socket, new ResponseContentLengthMismatchError())
+      return -1
+    }
```
mcollina, line 851: Why is this ignored by istanbul? Can't we reproduce?

mcollina (approved):
lgtm with a nit

lib/client.js
ronag: I think llhttp doesn't actually allow it, i.e. it shouldn't be possible.

ronag: I will have to dig into it further. Just too tired right now. Only slept 6 hours in total this week 😞 .

dnlup: Please take some rest 🙏🏻

Issue #756, opened 2021-04-20
Multiple requests to the same endpoint cause an error and the container is restarted


rbonillajr:
## Bug Description

With version 3.3.3 it is reporting this error when making several requests.

```
{"level":30,"time":1614213891203,"pid":28,"hostname":"host removed","msg":"Server listening at http://[::]:8080"}
^[[A^[[A^[[B^[[B{"level":30,"time":1614213926009,"pid":28,"hostname":"host removed","reqId":1,"req":{"method":"GET","url":"/playground","hostname":"10.9.0.26:8080","remoteAddress":"::ffff:10.254.252.71","remotePort":49462},"msg":"incoming request"}
{"level":30,"time":1614213926101,"pid":28,"hostname":"host removed","reqId":1,"res":{"statusCode":200},"responseTime":91.69444751739502,"msg":"request completed"}
{"level":30,"time":1614213926512,"pid":28,"hostname":"host removed","reqId":2,"req":{"method":"GET","url":"/playground/init.js","hostname":"10.9.0.26:8080","remoteAddress":"::ffff:10.254.252.71","remotePort":49462},"msg":"incoming request"}
{"level":30,"time":1614213926513,"pid":28,"hostname":"host removed","reqId":2,"res":{"statusCode":200},"responseTime":1.1579179763793945,"msg":"request completed"}
{"level":30,"time":1614213927824,"pid":28,"hostname":"host removed","reqId":3,"req":{"method":"POST","url":"/graphql","hostname":"10.9.0.26:8080","remoteAddress":"::ffff:10.254.252.71","remotePort":49467},"msg":"incoming request"}
{"level":30,"time":1614213928002,"pid":28,"hostname":"host removed","reqId":3,"res":{"statusCode":200},"responseTime":177.4089002609253,"msg":"request completed"}
{"level":30,"time":1614213945835,"pid":28,"hostname":"host removed","reqId":4,"req":{"method":"POST","url":"/graphql","hostname":"10.9.0.26:8080","remoteAddress":"::ffff:10.254.252.71","remotePort":49482},"msg":"incoming request"}
{"level":30,"time":1614213946595,"pid":28,"hostname":"host removed","reqId":4,"res":{"statusCode":200},"responseTime":760.1512966156006,"msg":"request completed"}
[...]

PR #817, opened 2021-05-30
fix: try/catch execute

ronag:
The assertion in https://github.com/nodejs/undici/issues/756 should not be possible. This is a try to make that true. I don't really know how those assertions are hit but I'm hoping this might bring some clarity.

Refs: https://github.com/nodejs/undici/issues/756


mcollina (approved):
lgtm

dnlup (approved):
Lgtm

PR #818, opened 2021-05-31
fix: guard against user throw in onError

ronag:
Consistently propagate unallowed throws in request.onError.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1175,7 +1175,6 @@ function _resume (client, sync) {
         })
         .on('error', function (err) {
           errorRequest(client, request, err)
-          assert(request.aborted)
         })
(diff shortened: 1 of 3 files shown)


lib/client.js
```diff
@@ -1180,7 +1174,7 @@ function _resume (client, sync) {
           assert(false)
         })
         .on('error', function (err) {
-          request.onError(err)
+          errorRequest(client, request, err)
           assert(request.aborted)
```
ronag, line 1178:
```suggestion
```

mcollina (commented):
good work! This needs test (I think).

ronag:
https://github.com/nodejs/undici/issues/819

Issue #677, opened 2021-04-04
Investigate if we can use fuzzying


ronag:
https://gitlab.com/gitlab-org/security-products/analyzers/fuzzers/jsfuzz

ronag:
Maybe @jonnydgreen?

jonnydgreen:
Yep, very happy to do some investigating on this - I can make a start tomorrow :)

jonnydgreen:
Just wanted to say that I haven't forgotten about this - I was on holiday for the second half of last week :) I've been running some basic investigations here and getting familiar with the jsfuzz API https://github.com/jonnydgreen/undici-fuzzing and will write up my experiences and anything I found in this issue

ronag:
@jonnydgreen this is an issue that would have been nice to catch through fuzzing https://github.com/nodejs/undici/pull/738

jonnydgreen:
Cool, understood - I will look into that as well!

ronag:
@jonnydgreen did you get anywhere with this?

jonnydgreen:
My apologies, this dropped off my radar. I will block out some time this weekend to make some significant progress but completely understand if you wanted to work on this yourself!

ronag:
I don't have any plans working on it. Was just curious if you came to any conclusion in regards to whether it's feasible or not.

jonnydgreen:
I did have some initial thoughts so far and will expand on the feasibility of these this weekend:
 - Automating in the CI would be tricky as the tool runs indefinitely, so it might need to run manually like the benchmarks. However, if can make it gracefully terminate (if there are no unexpected errors) after a short amount of time this could be run in a scheduled GH action if we wanted to automate this.
 - For constructing fuzzing scenarios, we'll probably need to provide some util functions to make constructing different fuzzing scenarios easier.

jonnydgreen:
Quick update! I've been looking into this this weekend with a main focus on creating a generic solution for different fuzzing examples as well as trying to improve the creation of different flows depending on the input buffer - I'll continue this exploration this week :) I'm pausing the CI side of things for now while I look at the raw functionality and get the key bits nailed down

jonnydgreen:
I think I have some key bits of functionality sorted with hopefully a nice way running a series of different functions within the fuzzer, I'll push up the code once I've tidied it up a bit and with my analysis to go with this!

jonnydgreen:
[...]

PR #821, opened 2021-06-01
Add fuzzing support

jonnydgreen:
Implementation of: https://github.com/nodejs/undici/issues/677

Using `jsfuzz`, I have created a basic framework for fuzzing in undici. This has the following structure:

```txt
test/fuzzing
├── client
│   ├── client-fuzz-*.js
│   └── index.js
├── fuzz.js
└── server
    ├── server-fuzz-*.js
    └── index.js
```

## `jsfuzz`

You can find details of the `jsfuzz` package [here](https://www.npmjs.com/package/jsfuzz). It has the following core features:

- Input buffer value provided as the argument to the entry function.
- Can store a corpus for replayability.
- Runs indefinitely.
- Provides metrics as it is run.
- Will quit with the appropriate information if an error is thrown.

Installation:
- `node_modules` size before installation: `239MB`
- `node_modules` size after installation: `245MB`

## Philosophy

I've tried to keep things relatively simple and flexible throughout with clear separations of concern between different parts of the fuzzing framework. `jsfuzz` only allows one function to be defined for an input, so I've tried to keep this simplicity where possible.

## Entrypoint

The entrypoint is located at `test/fuzzing/fuzz.js` and can be run with the following command:

```sh
npm run fuzz
```

This does the following:

- Creates a `net` server to be shared across fuzzing runs (see why [here](#limitations-and-things-to-improve) - I would appreciate some suggestions on this!).
- Sets up a timeout that gracefully quits the fuzzing run after a set period of time (should we want this to run in the CI).
- Retrieves the fuzz input buffer and runs against all client functions (which is turn triggers the server functions on the `net` server).
- Upon catching an error, prints useful information and throws.

## Client - Server fuzzing

We have two directories that define fuzzing behaviour for server/client side functionality: `server` and `client`. Client functions are intended for fuzzing the requests that undici makes and server functions are run (selected at random) within the receiving server. At the moment, both are pretty basic and could be extended further.

In this way, I think we enable developers to write custom fuzz functions for client/server flows should they want to extend the fuzzing framework further - we should aim to make these as generic as possible so the number of these functions is minimised. Once the function is written, developers should export in the corresponding `index.js` files and it will be called when the fuzzing is run.

## Example error

So you can see what is currently outputted, here is an example error.

```txt
#0 READ units: 44
#1 NEW     cov: 1476 corp: 45 exec/s: 0 rss: 27.39 MB
...
#29 NEW     cov: 1753 corp: 53 exec/s: 182 rss: 27.45 MB
#49 PULSE     cov: 1753 corp: 53 exec/s: 185 rss: 27.53 MB
=== Options: {"body":"��\u0000\u0000�\n","path":"��\u0000\u0000�\n","method":"��\u0000\u0000�\n"} ===
=== Failed fuzz clientFuzzOptions with input '���
 ==='
=== Raw:  <Buffer ed fa 00 00 fa 0a>  ===
=== Fuzz results start ===
{
  "clientFuzzFnName": "clientFuzzOptions",
  "options": {
    "body": "��\u0000\u0000�\n",
    "path": "��\u0000\u0000�\n",
    "method": "��\u0000\u0000�\n"
  },
  "err": {
    "name": "HTTPParserError",
    "code": "HPE_INVALID_HEADER_TOKEN",
    "data": "/%EF%BF%BD%EF%BF%BD%00%00%EF%BF%BD HTTP/1.1\r\nconnection: keep-alive\r\nhost: localhost:61337\r\ncontent-length: 12\r\n\r\n��\u0000\u0000�\n\r\n"
  }
}
=== Fuzz results end ===
=== Fuzz raw results start ===
 {
  clientFuzzFnName: 'clientFuzzOptions',
  options: {
    body: '��\x00\x00�\n',
    path: '��\x00\x00�\n',
    method: '��\x00\x00�\n'
  },
  err: HTTPParserError: Invalid header token
      ...[stacktrace_redacted] {
    code: 'HPE_INVALID_HEADER_TOKEN',
    data: '/%EF%BF%BD%EF%BF%BD%00%00%EF%BF%BD HTTP/1.1\r\n' +
      'connection: keep-alive\r\n' +
      'host: localhost:61337\r\n' +
      'content-length: 12\r\n' +
      '\r\n' +
      '��\x00\x00�\n' +
      '\r\n'
  }
}
=== Fuzz raw results end ===
=================================================================
HTTPParserError: Invalid header token
    at Parser.execute (/Users/judge/developer/github/jonnydgreen/undici/lib/client.js:1:132493)
    at Socket.onSocketData (/Users/judge/developer/github/jonnydgreen/undici/lib/client.js:1:147129)
    at Socket.emit (events.js:315:20)
    at addChunk (internal/streams/readable.js:309:12)
    at readableAddChunk (internal/streams/readable.js:284:9)
    at Socket.Readable.push (internal/streams/readable.js:223:10)
    at TCP.onStreamRead (internal/stream_base_commons.js:188:23) {
  code: 'HPE_INVALID_HEADER_TOKEN',
  data: '/%EF%BF%BD%EF%BF%BD%00%00%EF%BF%BD HTTP/1.1\r\n' +
    'connection: keep-alive\r\n' +
    'host: localhost:61337\r\n' +
    'content-length: 12\r\n' +
    '\r\n' +
    '��\x00\x00�\n' +
    '\r\n'
}
crash was written to crash-b8a6a959d038009336b7fa916807f2a1d22862f51d7dd661eb0db9982e2e29f9
crash(hex)=edfa0000fa0a
Worker exited
```

As you can see, along with the informative output, `jsfuzz` saves the run to a corpus (to replay in the future) and the crash to a file: `crash-b8a6a959d038009336b7fa916807f2a1d22862f51d7dd661eb0db9982e2e29f9`.

## Limitations and things to improve

Before making some improvements, I would love to see what you think! I've identified the following improvements to make before this PR is ready:

- I wasn't able to figure about how to isolate a receiving server within a fuzz function (kept running out of available ports as I couldn't close connections quick enough)
- It would be really cool to minimise the amount of client/server functions needed with super generic function definitions.
- Need to figure out if the current logging provides sufficient replayability if an error is found.
- I haven't actually found anything (which is both good and bad as although there are no errors found, it might mean that the fuzzing is probably a little too basic at the moment or I haven't run it for long enough! :))
- Would be cool to run within a GH action. This is the purpose of the timeout in the entrypoint file (`fuzz.js`), which will trigger after 5mins (arbitrarily chosen atm). Even further, the GH actions could even read an individual file and build a matrix to run individual fuzz functions (for each client-server function permutation). A bit of work is required to work out if this is is a good thing to do.

There is still a fair bit to work on in terms of the implementation, but let me know what you think about the design!


mcollina:
Good work! I like this very much. Could you:

1. fix/handle all the errors
2. integrate this in CI

jonnydgreen:
Sure thing, I'll get something sorted tonight! :)

dnlup (commented):
LGTM so far. Great work.

One question for the CI part: is it possible to upload the crash file as an artifact for later inspection?

jonnydgreen:
> One question for the CI part: is it possible to upload the crash file as an artifact for later inspection?

For sure, that's my plan yep! I was thinking of uploading the corpus as well for replayability :)

jonnydgreen:
Have pushed the following features, let me know what you think! :)

 - CI integration
 - Artifact results upload when an error occurs in the fuzzing: https://github.com/jonnydgreen/undici/actions/runs/901049099
 - Handling for all the undici errors:

I did find one issue where when `path` is set to `\\\\-` (from `<Buffer 5c 5c 2d>`), an `ENOTFOUND` error was thrown:

```txt
Error: getaddrinfo ENOTFOUND -
    at GetAddrInfoReqWrap.onlookup [as oncomplete] (node:dns:69:26) {
  errno: -3008,
  code: 'ENOTFOUND',
  syscall: 'getaddrinfo',
  hostname: '-'
}
```

I've added a TODO for now while I look into replicating this with a test and a potential fix if this is indeed a legit error.

For the CI integration, I took a similar philosophy to the benchmarks in that it only runs on `Node.js 16 / Ubuntu`, if you would prefer a matrix approach to cover more environments, let me know! :)

The CI workflow itself takes ~6mins which is less than the other workflows so it doesn't affect the current devx. The only risk is that a developer *could potentially have CI failures from the fuzzing workflow. With this in mind, if you would prefer to avoid this risk, then I can look into:
 - Not failing the workflow upon an error but still uploading artifacts (hard to be alerted of issues)
 - Using GitHub Action scheduled workflows - I've not used them myself, but can look into it if needs be!

jonnydgreen:
Hi! I've done a bit of experimentation and found the `ENOTFOUND` error mentioned ^ will only occur on Node.js 16. For example, given the `'\\\\-'` input path, we get the following URL:

## Node.js 16

```
$ node
Welcome to Node.js v16.2.0.
> new URL('\\\\-', 'http://asd')
URL {
  href: 'http://-/',
  origin: 'http://-',
  protocol: 'http:',
  username: '',
  password: '',
  host: '-',
  hostname: '-',
  port: '',
  pathname: '/',
  search: '',
  hash: ''
}
```

### Node.js 14

```
$ node
Welcome to Node.js v14.15.4.
> new URL('\\\\-', 'http://asd')
URL {
  href: 'http://asd///-',
  origin: 'http://asd',
  protocol: 'http:',
  username: '',
  password: '',
  host: 'asd',
  hostname: 'asd',
  port: '',
  pathname: '///-',
  search: '',
  hash: ''
}
```

My initial thoughts are that this is unexpected behaviour in Node.js 16 and potentially a security issue if one can rewrite hostnames like this, but I wondered what your opinions were on this and whether we should account for this in `undici`?

I looked at the [RFC for URI](https://datatracker.ietf.org/doc/html/rfc3986#section-3.3) and it seems like this input is an invalid character for path (`pchar = unreserved / pct-encoded / sub-delims / ":" / "@"`), so maybe `new URL(...)` or `undici` should throw an error in this case. If you agree that this is an issue and would prefer me to open an issue in the Node.js repo about this, let me know :)

mcollina:
I don't think this has any security implication. Open an issue on Node core.

jonnydgreen:
Cool, no problem - I'll open an issue later today and add a reference to it in this PR. Once that's done, I'll mark as ready for review as all the errors I've seen would be handled/fixed!

jonnydgreen:
Have now handled all errors that I've seen so far and the issue with the overriding path has been resolved (it is intended functionality since `\\hostname` is interpreted as a Windows UNC path, so it overrides the base URL - I learned something new today! :) The problem was actually with incorrect functionality with versions prior to 16 but this has recently been fixed in v14.17.0 onwards).

I upgraded `tsd` because the latest update to the `@types/estree` dependency started causing type errors. I've bumped `tsd` to the latest minor version to resolve this.

mcollina (approved):
lgtm

PR #822, opened 2021-06-06
Add tests for issue 583.

### LINKED ISSUE #583: SocketError: other side closed
Upon connecting to a proxy with user:pass it's all good, but when I try to make a request to an proxy which is IP authenticated, I get 'SocketError: other side closed'. 
I'm not sending the proxy-authorization header as it's not needed.
Blanked the proxy ip and port below.
```js
var Client = require('undici');
(async ()=>{
	const client = new Client("http://proxyip:proxyport")
	const response = await client.request({
	  method: 'GET',
	  path: 'https://api.ipify.org?format=json',
	  headers: {
	    //'proxy-authorization': `Basic ${Buffer.from(':').toString('base64')}`
	  }
	})

	response.body.setEncoding('utf8')
	let data = ''
	for await (const chunk of response.body) {
	  data += chunk
	}
	//
	console.log(response.statusCode)
	console.log(JSON.parse(data))

	client.close()
})();
```

mcollina:
Closes #583

### DIFF
--- a/test/issue-583.js
+++ b/test/issue-583.js
@@ -1,0 +1,35 @@
+'use strict'
+
+const { test } = require('tap')
+const { once } = require('events')
+const { createServer } = require('http')
+const { Client } = require('..')
+
+test('dont crash on unexpected Transfer-Encoding Header', async (t) => {
+  t.plan(1)
+
+  const server = createServer((req, res) => {
+    res.setHeader('Content-Type', 'text/plain')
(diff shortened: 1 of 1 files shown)


ronag (approved):

mcollina:
Let's wait for confirmation that the problem is fixed.

ronag:
@mcollina i think we are good here

PR #823, opened 2021-06-07
fix: relax assertion into error

ronag:
Relaxes assertion into an error which can be caught. Also adds missing return -1.

Refs: https://github.com/nodejs/undici/issues/756


ronag:
Trying to resolve https://github.com/nodejs/undici/issues/756. Currently just guesswork since we can't reproduce the issue.

deps/llhttp/src/api.c
```diff
@@ -41,6 +41,7 @@ void llhttp_init(llhttp_t* parser, llhttp_type_t type,
 
 #if defined(__wasm__)
 
+extern int wasm_on_message_begin(llhttp_t* p);
```
mcollina, line 44: have you added this or is it part of llhttp?

mcollina (commented):
good spot!

deps/llhttp/src/api.c
ronag: It's part of llhttp.

deps/llhttp/src/api.c
```diff
@@ -53,7 +54,7 @@ int on_headers_complete(llhttp_t * p)
 }
 
 const llhttp_settings_t wasm_settings = {
-  NULL,
+  wasm_on_message_begin,
```
ronag, line 57: @mcollina previously we didn't use this llhttp callback and just passed NULL.

mcollina: I mean.. this change shouldn't happen upstream?

ronag: We have other changes here that are not upstream.

ronag: I'm not sure it's worth trying to port these changes upstream as it's just C glue between undici and llhttp.

dnlup (approved):

deps/llhttp/src/api.c
mcollina: We should move them upstream as it will be simpler to bump the library.

mcollina (approved):
lgtm

deps/llhttp/src/api.c
ronag: https://github.com/nodejs/llhttp/pull/116

PR #825, opened 2021-06-09
doc: added response body handling advice

MoritzLoewenstein:
Im not really sure wether this only applies to the request method or if there is better place to put this (maybe at the top of Client.md).


mcollina (approved):
lgtm

PR #827, opened 2021-06-09
fix(request): add check for kAbort type to be a function

GO3LIN:
when running mocked objects in unit tests kAbort is not set to a function, fixes https://github.com/nodejs/undici/issues/824

### DIFF
--- a/lib/api/api-request.js
+++ b/lib/api/api-request.js
@@ -22,7 +22,7 @@ class RequestResponse extends Readable {
       err = new RequestAbortedError()
     }
 
-    if (err) {
+    if (err && typeof this[kAbort] === 'function') {
       this[kAbort]()
     }
 


mcollina (changes requested):
Could you add a unit test?

GO3LIN:
@mcollina I am trying to figure out what can be a good test scenario for that fix, maybe testing the callback is still called even if the kAbort is not a function ?

mcollina:
> @mcollina I am trying to figure out what can be a good test scenario for that fix, maybe testing the callback is still called even if the kAbort is not a function ? there may be the other side scenario too, if the kAbort is a function then it's called when the response gets destroyed.

I would start adding a test from your crash.

GO3LIN:
@mcollina I am unable to reproduce the bug in the undici environment, I tried with JavaScript and TypeScript, I can't get that error. Maybe the dependencies versions are not the same.. Here's the test I wrote:

```typescript
import { MockAgent, setGlobalDispatcher, request } from '../..';


describe("index.ts test", () => {
    let mockAgent: MockAgent;
    const origin = 'https://example.com'

    beforeEach(() => {
        mockAgent = new MockAgent({ connections: 1 });
    });

    it("Returns the Working parameter in a APIResponse type", async () => {
        
        // Given
        const expectedResponse = JSON.stringify({Working: 1})
        mockAgent.get(origin).intercept({
            path: '/tdd/everything',
            method: 'GET'
        }).reply(200, expectedResponse)
        
        // When
        const { statusCode, headers, trailers, body } = await request(`${origin}/tdd/everything`, {
            method: 'GET'
          })

        let response = undefined

            response = data
        }

        // Then
        expect(response.toString()).toBe(expectedResponse.toString())
    });


})
```

ronag (changes requested):
abort should be a function, I think the problem lies somewhere else

jaceppo (approved):

ronag:
This should not be possible and coverage is already at 100 without this. Thanks for trying. Closing.

PR #829, opened 2021-06-10
docs,test: added test for the client/pool connect option on client

Ricardonacif:
Just adding documentation and a test that covers https://github.com/nodejs/undici/pull/763.


Ricardonacif:
Wait, something is very off here. It's never reaching the proxy itself. It's late night here and I'll investigate tomorrow. Is the client option `connector` or `connect`? Maybe there's something wrong there, but I'll investigate further

Ricardonacif:
Ok, found the issue. On the current client.js, I have to import `kConnect` symbol in order to use the connect option:
```
const { kConnect, } = require('../lib/core/symbols.js')

const client = new Client(serverUrl, {
    [kConnect]: connectTlsFactory(proxyUrl, 'user:pass')
  })
```
That's probably a bug right? Or is it intentional?

mcollina:
I think `kConnect` is private.

ronag:
Connect is private due to kissing docs and tests.

ronag:
I’m fine with making it public with tests and docs.

ronag (changes requested):
- connect argument needs to be made public. Right now it's just added to docs and doesn't do anything.
- Needs tests for the above.
- Tests are currently failing

Ricardonacif:
Cool I'll see if I can tackle some of this during this weekend. Thanks!

Ricardonacif:
Guys, I'm in a critical phase of my business right now and I don't think I'll be able to contribute to this issue for at least the next 2 months. I'll close this PR for now, but I promise to be back to tackle this.

PR #834, opened 2021-06-15
fix: destroy IncomingMessage and invalid assertion

ronag:
Change so that we destroy IncomingMessage without closing the
socket. Also fixes an invalid assertion that was triggered by
the test.


mcollina (approved):
lgtm, sigh

ronag:
sigh? 😕

PR #835, opened 2021-06-16
update llhttp

ronag:
Shouldn't actually change anything just sync our code with llhttp.


deps/llhttp/src/llhttp.c
```diff
@@ -1102,7 +1102,7 @@ static llparse_state_t llhttp__internal__run(
     }
     case s_n_llhttp__internal__n_consume_content_length:
     s_n_llhttp__internal__n_consume_content_length: {
-      uint64_t avail;
+      size_t avail;
```
ronag, line 1105: @indutny won't this get us back with there previous problems we had with overflow?

deps/llhttp/src/api.c
```diff
@@ -69,16 +71,40 @@ const llhttp_settings_t wasm_settings = {
 
 llhttp_t* llhttp_alloc(llhttp_type_t type) {
   llhttp_t* parser = malloc(sizeof(llhttp_t));
-  memset(parser, 0, sizeof(llhttp_t));
```
dnlup, line 72: I think it's safe now to remove this.

dnlup (approved):
Lgtm once https://github.com/nodejs/undici/pull/835#discussion_r652589059 is settled

mcollina (approved):
lgtm

deps/llhttp/src/llhttp.c
indutny: No, this is actually good because the value of `avail` comes from pointer subtraction which is guaranteed to fit into `size_t` (technically it should be `off_t`, but they have same bit size). `uint64_t` below is sufficient to make sure that it works. Let me know if the issue resurfaces, though!

PR #836, opened 2021-06-16
fix(types): add missing string[] option for passed headers

calebboyd:
matches typescript types to [UndiciHeaders](https://github.com/nodejs/undici/blob/main/docs/api/Dispatcher.md#parameter-undiciheaders)


mcollina (approved):
lgtm

mcollina:
Thanks!

Issue #820, opened 2021-05-31
Upload progress callback


szmarczak:
## This would solve...

Currently there's no way to get upload progress other than passing a stream. What if we're passing a 10MB Buffer instead? `Readable.from(buffer)` and we're storing it twice, which is unfortunate. So for an application that handles 100MB at a time we would need 200MB. 

## The implementation should look like...

```js
  client.dispatch({
    path: '/',
    method: 'GET',
    headers: {
      'x-foo': 'bar'
    }
  }, {
...
    onUploadProgress: (bytes) => {
      console.log(`onUploadProgress : ${bytes} bytes in total`)
    },
...
  })
```

## I have also considered...

Reading the state from the underlying socket. It's not so nice.

mcollina:
This would be a nice feature to add! I think we should make the callback arrive to https://github.com/nodejs/undici/blob/b33dffa76d34d31a7ee189ac72c5ace5ef75e7f3/lib/client.js#L1423 and call it afterwards (if it's present).

gabrielcaires:
Can I get it as my first contribution?

Linkgoron:
I'll try to take a stab at this. Any opinions regarding `onBodyChunkSent` or `onDataSent` (as opposed to the current `onData`)? I'll have the callback receive two parameters, the current chunk size and the total bytes written so far.

mcollina:
I prefer `onBodyChunkSent`

mcollina:
@szmarczak I don’t understand why wrapping a buffer in a Readable would duplicate the memory used by the app.

szmarczak:
> I don’t understand why wrapping a buffer in a Readable would duplicate the memory used by the app.

In order to receive progress events we would have to disable object mode.

mcollina:
I'm questioning a bit the use case.. I would not allocate a massive buffer in the first place.

BobNobrain:
Not sure if this is the right place to ask, but is there any way to track *download* progress? Especially when using `stream` method

axetroy:
The documentation for `undici` is so bad that you can't even find any example for this.

viocha:
Has this feature been implemented or abandoned?

metcoder95:
This should be doable as an interceptor, contributions are welcomed

PR #847, opened 2021-06-20
feature: add onBodyChunkSent callback to dispatch

Linkgoron:
Add a callback that's called when a body chunk is sent to the server to `dispatch`.

The callback is invoked only after `stream.write` calls the end callback (I wasn't sure if it should be called synchronously with the call to `write` or not).
For stream bodies, the callback will get invoked after every chunk is written, and for other types (buffer etc.) it will be invoked once after the whole body is sent.

refs: https://github.com/nodejs/undici/issues/820

### DIFF
--- a/docs/api/Dispatcher.md
+++ b/docs/api/Dispatcher.md
@@ -205,6 +205,7 @@ Returns: `void`
 * **onHeaders** `(statusCode: number, headers: Buffer[], resume: () => void) => boolean` - Invoked when statusCode and headers have been received. May be invoked multiple times due to 1xx informational headers. Not required for `upgrade` requests.
 * **onData** `(chunk: Buffer) => boolean` - Invoked when response payload data is received. Not required for `upgrade` requests.
 * **onComplete** `(trailers: Buffer[]) => void` - Invoked when response payload and trailers have been received and the request has completed. Not required for `upgrade` requests.
+* **onBodyChunkSent** `(chunk: string | Buffer | Uint8Array) => void` - Invoked when a body chunk is sent to the server. Not required. For a stream or iterable body this will be invoked for every chunk. For other body types, it will be invoked once after the body is sent.
 
 #### Example 1 - Dispatch GET request
 
--- a/docs/api/Dispatcher.md
+++ b/docs/api/Dispatcher.md
@@ -205,7 +205,7 @@ Returns: `void`
 * **onHeaders** `(statusCode: number, headers: Buffer[], resume: () => void) => boolean` - Invoked when statusCode and headers have been received. May be invoked multiple times due to 1xx informational headers. Not required for `upgrade` requests.
 * **onData** `(chunk: Buffer) => boolean` - Invoked when response payload data is received. Not required for `upgrade` requests.
 * **onComplete** `(trailers: Buffer[]) => void` - Invoked when response payload and trailers have been received and the request has completed. Not required for `upgrade` requests.
-* **onBodyChunkSent** `(chunk: string | Buffer | Uint8Array) => void` - Invoked when a body chunk is sent to the server. Not required. For a stream or iterable body this will be invoked for every chunk. For other body types, it will be invoked once after the body is sent.
+* **onBodySent** `(chunk: string | Buffer | Uint8Array) => void` - Invoked when a body chunk is sent to the server. Not required. For a stream or iterable body this will be invoked for every chunk. For other body types, it will be invoked once after the body is sent.
 
 #### Example 1 - Dispatch GET request
 
--- a/lib/client.js
+++ b/lib/client.js
@@ -1430,12 +1430,9 @@ function write (client, request) {
 
     socket.cork()
     socket.write(`${header}content-length: ${contentLength}\r\n\r\n`, 'ascii')
-    const endCallback = request.onBodyChunkSent
-      ? () => request.onBodyChunkSent(body)
-      : undefined
-    socket.write(body, endCallback)
+    socket.write(body)
     socket.uncork()
-
+    request.onBodySent(body)
     if (!expectsPayload) {
       socket[kReset] = true
     }
@@ -1497,13 +1494,11 @@ function writeStream ({ client, request, socket, contentLength, header, expectsP
       }
 
       bytesWritten += len
-      const endCallback = request.onBodyChunkSent
-        ? () => request.onBodyChunkSent(chunk)
-        : undefined
 
-      if (!socket.write(chunk, endCallback) && this.pause) {
+      if (!socket.write(chunk) && this.pause) {
         this.pause()
       }
+      request.onBodySent(chunk)
     } catch (err) {
       util.destroy(this, err)
     }
@@ -1668,14 +1663,10 @@ async function writeIterable ({ client, request, socket, contentLength, header, 
       }
 
       bytesWritten += len
-
-      const endCallback = request.onBodyChunkSent
-        ? () => request.onBodyChunkSent(chunk)
-        : undefined
-
-      if (!socket.write(chunk, endCallback)) {
+      if (!socket.write(chunk)) {
         await waitForDrain()
       }
+      request.onBodySent(chunk)
 
       if (socket[kError]) {
         throw socket[kError]
--- a/lib/client.js
+++ b/lib/client.js
@@ -1430,7 +1430,10 @@ function write (client, request) {
 
     socket.cork()
     socket.write(`${header}content-length: ${contentLength}\r\n\r\n`, 'ascii')
-    socket.write(body)
+    const endCallback = request.onBodyChunkSent
+      ? () => request.onBodyChunkSent(body)
+      : undefined
+    socket.write(body, endCallback)
     socket.uncork()
 
     if (!expectsPayload) {
@@ -1494,8 +1497,11 @@ function writeStream ({ client, request, socket, contentLength, header, expectsP
       }
 
       bytesWritten += len
+      const endCallback = request.onBodyChunkSent
+        ? () => request.onBodyChunkSent(chunk)
+        : undefined
 
-      if (!socket.write(chunk) && this.pause) {
+      if (!socket.write(chunk, endCallback) && this.pause) {
         this.pause()
       }
     } catch (err) {
@@ -1663,7 +1669,11 @@ async function writeIterable ({ client, request, socket, contentLength, header, 
 
       bytesWritten += len
 
-      if (!socket.write(chunk)) {
+      const endCallback = request.onBodyChunkSent
+        ? () => request.onBodyChunkSent(chunk)
+        : undefined
+
+      if (!socket.write(chunk, endCallback)) {
         await waitForDrain()
       }
 
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -105,16 +105,8 @@ class Request {
       throw new InvalidArgumentError('invalid onError method')
     }
 
-    if (typeof handler.onBodyChunkSent !== 'function' && handler.onBodyChunkSent !== undefined) {
-      throw new InvalidArgumentError('invalid onBodyChunkSent method')
-    } else if (typeof handler.onBodyChunkSent === 'function') {
-      this.onBodyChunkSent = (chunk) => {
-        try {
-          this[kHandler].onBodyChunkSent(chunk)
-        } catch (err) {
-          this.onError(err)
-        }
-      }
+    if (typeof handler.onBodySent !== 'function' && handler.onBodySent !== undefined) {
+      throw new InvalidArgumentError('invalid onBodySent method')
     }
 
     if (this.upgrade || this.method === 'CONNECT') {
@@ -138,6 +130,16 @@ class Request {
     this.servername = util.getServerName(this.host)
 
     this[kHandler] = handler
+  }
+
+  onBodySent (chunk) {
+    if (this[kHandler].onBodySent) {
+      try {
+        this[kHandler].onBodySent(chunk)
+      } catch (err) {
+        this.onError(err)
+      }
+    }
   }
 
   onConnect (abort) {
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -105,6 +105,18 @@ class Request {
       throw new InvalidArgumentError('invalid onError method')
     }
 
+    if (typeof handler.onBodyChunkSent !== 'function' && handler.onBodyChunkSent !== undefined) {
+      throw new InvalidArgumentError('invalid onBodyChunkSent method')
+    } else if (typeof handler.onBodyChunkSent === 'function') {
+      this.onBodyChunkSent = (chunk) => {
+        try {
+          this[kHandler].onBodyChunkSent(chunk)
+        } catch (err) {
+          this.onError(err)
+        }
+      }
+    }
+
     if (this.upgrade || this.method === 'CONNECT') {
       if (typeof handler.onUpgrade !== 'function') {
         throw new InvalidArgumentError('invalid onUpgrade method')
--- a/types/dispatcher.d.ts
+++ b/types/dispatcher.d.ts
@@ -136,7 +136,7 @@ declare namespace Dispatcher {
     /** Invoked when response payload and trailers have been received and the request has completed. */
     onComplete?(trailers: string[] | null): void;
     /** Invoked when a body chunk is sent to the server. May be invoked multiple times for chunked requests */
-    onBodyChunkSent?(chunkSize: number, totalBytesSent: number): void;
+    onBodySent?(chunkSize: number, totalBytesSent: number): void;
   }
   export type PipelineHandler = (data: PipelineHandlerData) => Readable;
 }
--- a/types/dispatcher.d.ts
+++ b/types/dispatcher.d.ts
@@ -135,6 +135,8 @@ declare namespace Dispatcher {
     onData?(chunk: Buffer): boolean;
     /** Invoked when response payload and trailers have been received and the request has completed. */
     onComplete?(trailers: string[] | null): void;
+    /** Invoked when a body chunk is sent to the server. May be invoked multiple times for chunked requests */
+    onBodyChunkSent?(chunkSize: number, totalBytesSent: number): void;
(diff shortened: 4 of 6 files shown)


lib/client.js
```diff
@@ -1397,7 +1397,7 @@ function write (client, request) {
 
     socket.cork()
     socket.write(`${header}content-length: ${contentLength}\r\n\r\n`, 'ascii')
-    socket.write(body)
+    socket.write(body, () => notifyBodyChunkSent(request, contentLength, contentLength))
```
mcollina, line 1400: Could you cache this closure at a higher level? We should not allocate it every time.

Moreover, we should not allocate it at all if it's not needed by the user.

mcollina:
@szmarczak ptal, would this do?

szmarczak:
Yes.

docs/api/Dispatcher.md
```diff
@@ -207,6 +207,7 @@ Returns: `void`
 * **onHeaders** `(statusCode: number, headers: Buffer[], resume: () => void) => boolean` - Invoked when statusCode and headers have been received. May be invoked multiple times due to 1xx informational headers. Not required for `upgrade` requests.
 * **onData** `(chunk: Buffer) => boolean` - Invoked when response payload data is received. Not required for `upgrade` requests.
 * **onComplete** `(trailers: Buffer[]) => void` - Invoked when response payload and trailers have been received and the request has completed. Not required for `upgrade` requests.
+* **onBodyChunkSent** `(chunkSize: number, totalSent: number) => void` - Invoked when a body chunk is sent to the server. Not required. For a stream body this will be invoked for every chunk. For other body types, it will be invoked once after the body is sent.
```
szmarczak, line 210: > For other body types, it will be invoked once after the body is sent.

So this doesn't exactly fix the issue. It will be called only once when passing a Buffer...

mcollina: Because we pass through the buffer to the inner layers without doing any splitting ourselves.

szmarczak: Ah, right, it's okay :)

lib/client.js
```diff
@@ -1397,7 +1397,10 @@ function write (client, request) {
 
     socket.cork()
     socket.write(`${header}content-length: ${contentLength}\r\n\r\n`, 'ascii')
-    socket.write(body)
+    const endCallback = request.onBodyChunkSent
+      ? request.onBodyChunkSent.bind(request, contentLength, contentLength)
```
szmarczak, line 1401: Is `bind` really necessary in this case?

Linkgoron: I could create an arrow function instead, if it makes a difference.

Linkgoron:
Also, I think that I still need to update d.ts files as well.

mcollina (approved):
lgtm

(we shuld do a round of benchmark before/after)

mcollina:
wdyt @ronag?

lib/client.js
```diff
@@ -1397,7 +1397,10 @@ function write (client, request) {
...
+      : undefined
+    socket.write(body, endCallback)
```
ronag, line 1403: I don't like passing callback to `write` it has performance implications.

ronag (changes requested):
This doesn't actually solve https://github.com/nodejs/undici/issues/820 though... it's not really needed if you are passing a stream as body and if passing a buffer it has the same problem.

ronag:
I would prefer something like:

```js
onBodySent(chunk: Buffer)
```

Linkgoron:
> This doesn't actually solve #820 though... it's not really needed if you are passing a stream as body and if passing a buffer it has the same problem. You would have to manually split the buffer into smaller writes for this to solve the referenced issue.

Yes, it doesn't solve #820. I think that this feature would probably be more useful once iterator (async or regular) bodies are supported, and once they're supported I _assume_ that creating chunks using an iterator would be more performant than with a stream.

ronag:
> > This doesn't actually solve #820 though... it's not really needed if you are passing a stream as body and if passing a buffer it has the same problem. You would have to manually split the buffer into smaller writes for this to solve the referenced issue.
> 
> Yes, it doesn't solve #820. I think that this feature would probably be more useful once iterator (async or regular) bodies are supported, and once they're supported I _assume_ that

What does it solve?

> creating chunks using an iterator would be more performant than with a stream.

Not sure I follow.

lib/client.js
```diff
@@ -1450,8 +1453,10 @@ function write (client, request) {
         }
 
         bytesWritten += len
-
-        if (!socket.write(chunk) && this.pause) {
+        const endCallback = request.onBodyChunkSent
+          ? request.onBodyChunkSent.bind(request, len, bytesWritten)
+          : undefined
```
ronag, line 1458: Better to just capture bytesWritten, that way you don't need to create a closure for every chunk.

Linkgoron: Originally, that's what I did but It didn't pass tests, as the callback is called after the socket actually writes, but `bytesWritten` is updated when `socket.write` is called. As I said in the OP, it might be easier to call `onBodyChunkSent` synchronously with the call to write.

Linkgoron:
> > > This doesn't actually solve #820 though... it's not really needed if you are passing a stream as body and if passing a buffer it has the same problem. You would have to manually split the buffer into smaller writes for this to solve the referenced issue.
> > 
> > 
> > Yes, it doesn't solve #820. I think that this feature would probably be more useful once iterator (async or regular) bodies are supported, and once they're supported I _assume_ that
> 
> What does it solve?

It solves getting a notification when a chunk was uploaded without adding a listener to the stream, and if/when async iterators are added this would be more useful (because you can't "simply" add a listener to the iterator).

> 
> > creating chunks using an iterator would be more performant than with a stream.
> 
> Not sure I follow.

In the original issue, the main complaint was that creating chunks using a stream would cause performance/memory issues. I _assumed_ that yielding chunks of a buffer from an iterator instead of a stream would be more performant, as it doesn't have all of the stream machinery (I'm not saying that you can't do `Readable.from`, just that it would be less performant).

ronag:
> It solves getting a notification when a chunk was uploaded without adding a listener to the stream

I don't see adding a listener to a stream as a problem.

> I assumed that yielding chunks of a buffer from an iterator instead of a stream would be more performant, as it doesn't have all of the stream machiner

I don't **think** the difference is of significance.

The only think that makes sense here IMO is that if onBodyChunkSent is set then we split the buffer ourselves internally. But I don't think having 10MB+ buffers in memory is a good idea in general. @szmarczak what are we actually trying to solve here?

mcollina:
> The only think that makes sense here IMO is that if onBodyChunkSent is set then we split the buffer ourselves internally.

I don't think so. Even if we use a stream to feed data to dispatch, there is no way to know if a given chunk was written, only if it was fed to the socket. Not sure if it makes a difference.

ronag:
> I don't think so. Even if we use a stream to feed data to dispatch, there is no way to know if a given chunk was written, only if it was fed to the socket. Not sure if it makes a difference.

No, but if we wait for `'drain'` we would at least get a hint in regards to progress.

ronag:
But yea, it kind of goes back to what are we actually solving here?

szmarczak:
> @szmarczak what are we actually trying to solve here?

Sorry I haven't explained this more extensively. I though that `net`'s `.bytesWritten` is updated dynamically but it isn't. Passing a 1GB buffer will cause `.bytesWritten` to jump high, without any progress.

> But I don't think having 10MB+ buffers in memory is a good idea in general.

Correct. But some people do this.

`onBodyCunkSent` is useful here too. It includes the size of the headers etc. IIRC in Got we used to use `request._header` but that caused some problems (it was `undefined` in some cases iirc).

> creating chunks using an iterator would be more performant than with a stream.
> Not sure I follow.

I think what @Linkgoron was trying to say that you don't need to initialize the usual stream stuff when using generators.

> No, but if we wait for 'drain' we would at least get a hint in regards to progress.

Yes, although we would have to remember the size of all the queued chunks. This PR makes getting upload progress callback much easier. We don't have to manually provide the callbacks for `stream.write`.

lib/core/request.js
```diff
@@ -103,6 +103,18 @@ class Request {
...
     }
 
+    if (typeof handler.onBodyChunkSent !== 'function' && handler.onBodyChunkSent !== undefined) {
+      throw new InvalidArgumentError('invalid onBodyChunkSent method')
+    } else if (typeof handler.onBodyChunkSent === 'function') {
+      this.onBodyChunkSent = (chunkSize, totalBytesSent) => {
+        try {
+          this[kHandler].onBodyChunkSent(chunkSize, totalBytesSent)
+        } catch (err) {
+          this[kHandler].onError(err)
```
ronag, line 113:
```suggestion
          this.onError(err)
```

types/dispatcher.d.ts
```diff
@@ -130,6 +130,8 @@ declare namespace Dispatcher {
     onData?(chunk: Buffer): boolean;
     /** Invoked when response payload and trailers have been received and the request has completed. */
     onComplete?(trailers: string[] | null): void;
+    /** Invoked when a body chunk is sent to the server. May be invoked multiple times for chunked requests */
+    onBodyChunkSent?(chunkSize: number, totalBytesSent: number): void;
```
szmarczak, line 134:
```suggestion
    onBodyChunkSent?(chunk: string | Buffer | Uint8Array): void;
```

szmarczak (changes requested):

mcollina:
> > But I don't think having 10MB+ buffers in memory is a good idea in general.
> Correct. But some people do this.

There is not much we can help with :(. If they want progress updates on that, the user would need to segment the buffer and write it as a stream.

We could potentially do that ourselves but it's a different problem than the one addressed in this PR.

Linkgoron:
It's quite simple to segment the buffer without using too much memory. I assume that something like the following would work?

```js
function* chunkify(buffer, chunkSize = 10000) {
  for (let pos = 0; pos < buffer.byteLength; pos += chunkSize) {
    yield buffer.subarray(pos, pos + chunkSize)
  }
}
```

Then, either use `chunkify(buffer)` or `Readable.from(chunkify(buffer))` (if you want to add listeners to the stream) as the body

szmarczak:
I think that could work. Will check this on Sunday.

Linkgoron:
Looking at the CI benchmarks, they seem way too different between this branch and `main` (if I'm reading the results correctly, this branch is way faster, which it shouldn't be). I assume that they're not reliable/consistent enough?

mcollina:
I've checked the benchmarks on my dedicated server and there is no regressions.

@szmarczak have you had a chance to look at this?

ronag:
Sorry to be difficult but I'm still unsure how this is significantly better than just attaching a `'data'` handler on the body?

szmarczak:
I've run benchmarks on my PC and got these:

`main` branch:

```
szm@solus ~/Desktop/undici $ node benchmarks/benchmark.js 
│ Tests               │ Samples │           Result │ Tolerance │ Difference with slowest │
|─────────────────────|─────────|──────────────────|───────────|─────────────────────────|
│ http - no keepalive │      15 │  6445.03 req/sec │  ± 2.71 % │                       - │
|─────────────────────|─────────|──────────────────|───────────|─────────────────────────|
│ http - keepalive    │      25 │  8256.76 req/sec │  ± 2.50 % │               + 28.11 % │
|─────────────────────|─────────|──────────────────|───────────|─────────────────────────|
│ undici - pipeline   │      20 │ 11164.69 req/sec │  ± 2.40 % │               + 73.23 % │
|─────────────────────|─────────|──────────────────|───────────|─────────────────────────|
│ undici - request    │      20 │ 16932.07 req/sec │  ± 2.42 % │              + 162.72 % │
|─────────────────────|─────────|──────────────────|───────────|─────────────────────────|
│ undici - dispatch   │      30 │ 17782.47 req/sec │  ± 2.74 % │              + 175.91 % │
|─────────────────────|─────────|──────────────────|───────────|─────────────────────────|
│ undici - stream     │      10 │ 19110.33 req/sec │  ± 0.81 % │              + 196.51 % │
```

this branch:
```
szm@solus ~/Desktop/undici $ node benchmarks/benchmark.js 
│ Tests               │ Samples │           Result │ Tolerance │ Difference with slowest │
|─────────────────────|─────────|──────────────────|───────────|─────────────────────────|
│ http - no keepalive │      20 │  7768.10 req/sec │  ± 2.66 % │                       - │
|─────────────────────|─────────|──────────────────|───────────|─────────────────────────|
│ http - keepalive    │      30 │  8190.95 req/sec │  ± 2.90 % │                + 5.44 % │
|─────────────────────|─────────|──────────────────|───────────|─────────────────────────|
│ undici - pipeline   │      10 │ 13091.86 req/sec │  ± 0.94 % │               + 68.53 % │
|─────────────────────|─────────|──────────────────|───────────|─────────────────────────|
│ undici - stream     │      30 │ 16587.85 req/sec │  ± 2.92 % │              + 113.54 % │
|─────────────────────|─────────|──────────────────|───────────|─────────────────────────|
│ undici - request    │      15 │ 16713.28 req/sec │  ± 2.41 % │              + 115.15 % │
|─────────────────────|─────────|──────────────────|───────────|─────────────────────────|
│ undici - dispatch   │      10 │ 19770.27 req/sec │  ± 0.73 % │              + 154.51 % │
```

There is negligible difference. I'll check the `chunkify` approach now and update this comment.

szmarczak:
There's only 5% performance drop when using 10MB body and 64KB chunks:

```
│ Tests    │ Samples │        Result │ Tolerance │ Difference with slowest │
|──────────|─────────|───────────────|───────────|─────────────────────────|
│ chunkify │      60 │ 95.31 req/sec │  ± 2.96 % │                       - │
```

1MB body and 64KB chunks:

```
│ Tests    │ Samples │         Result │ Tolerance │ Difference with slowest │
|──────────|─────────|────────────────|───────────|─────────────────────────|
│ chunkify │      10 │ 840.84 req/sec │  ± 2.56 % │                       - │

│ Tests       │ Samples │         Result │ Tolerance │ Difference with slowest │
|─────────────|─────────|────────────────|───────────|─────────────────────────|
│ no chunkify │      35 │ 883.46 req/sec │  ± 2.96 % │                       - │
```

szmarczak:
Given these numbers, there's almost no difference if the chunks are 64KB or 100KB. It's still 5% performance drop.

So I think it would be best to include `chunkify` in the docs for those who want more detailed progress callbacks.

ronag:
How about just adding an example in the examples folder?

szmarczak:
Sounds good.

Linkgoron:
> Sorry to be difficult but I'm still unsure how this is significantly better than just attaching a `'data'` handler on the body?

For iterator bodies you'd have to wrap them in a stream (or build a wrapper), and this is a bit different as this PR calls you after the data was sent while 'data' calls you before the data was sent. I'm not sure if I would call that significant.

ronag:
> calls you after the chunk was sent while 'data' calls you before the chunk was sent

Yes, I just don't see how/when that would matter?

szmarczak:
I don't think this would matter. Can we get this merged? This way we don't have to wrap some bodies into streams.

lib/client.js
```diff
@@ -1490,8 +1493,11 @@ function writeStream ({ client, request, socket, contentLength, header, expectsP
       }
 
       bytesWritten += len
+      const endCallback = request.onBodyChunkSent
+        ? () => request.onBodyChunkSent(chunk)
+        : undefined
```
ronag, line 1498: It's probably better and faster to just call `onBodyChunkSent` after/before write. Using the callback doesn't really do that much, it's no guarantee that it was sent, just that it's written to the kernels buffer.

lib/core/request.js
```diff
@@ -105,6 +105,18 @@ class Request {
...
+    if (typeof handler.onBodyChunkSent !== 'function' && handler.onBodyChunkSent !== undefined) {
+      throw new InvalidArgumentError('invalid onBodyChunkSent method')
+    } else if (typeof handler.onBodyChunkSent === 'function') {
+      this.onBodyChunkSent = (chunk) => {
+        try {
+          this[kHandler].onBodyChunkSent(chunk)
+        } catch (err) {
+          this.onError(err)
+        }
+      }
```
ronag, line 117: No need to lazyily do this. Just put a `onBodySent` method on the class.

docs/api/Dispatcher.md
```diff
@@ -203,6 +203,7 @@ Returns: `void`
 * **onHeaders** `(statusCode: number, headers: Buffer[], resume: () => void) => boolean` - Invoked when statusCode and headers have been received. May be invoked multiple times due to 1xx informational headers. Not required for `upgrade` requests.
 * **onData** `(chunk: Buffer) => boolean` - Invoked when response payload data is received. Not required for `upgrade` requests.
 * **onComplete** `(trailers: Buffer[]) => void` - Invoked when response payload and trailers have been received and the request has completed. Not required for `upgrade` requests.
+* **onBodyChunkSent** `(chunk: string | Buffer | Uint8Array) => void` - Invoked when a body chunk is sent to the server. Not required. For a stream or iterable body this will be invoked for every chunk. For other body types, it will be invoked once after the body is sent.
```
ronag, line 206:
```suggestion
* **onBodySent** `(chunk: string | Buffer | Uint8Array) => void` - Invoked when a body chunk is sent to the server. Not required. For a stream or iterable body this will be invoked for every chunk. For other body types, it will be invoked once after the body is sent.
```

ronag (changes requested):
I'm -0 with this after comments are resolved.

ronag:
Should we rename `onData` to `onBodyReceived`?

szmarczak:
Eventually `onBodyChunkSent` can be just `onSent` or `onDataSent`.

szmarczak:
> Should we rename/alias onData to onBodyReceived?

I'm okay with `onData`.

Linkgoron:
I've pushed the naming changes and the behaviour changes. Note that now the callback is called after `write` is called synchronously (except in the case of async iterators, where it might happen after awaiting for `drain`)

ronag (approved):

PR #848, opened 2021-06-22
feature: add support for request iterator body

### LINKED ISSUE #198: Request Body as Iterable/Iterator

Linkgoron:
continuation of https://github.com/nodejs/undici/pull/363

closes #198

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1587,6 +1587,8 @@ async function writeIterable ({ client, request, socket, contentLength, header, 
 
   let bytesWritten = 0
   try {
+    // TODO (fix): What if socket errors while waiting for body?
+    // It's up to the user to somehow abort the async iterable.
     for await (const chunk of body) {
       if (socket[kError]) {
         throw socket[kError]
--- a/lib/client.js
+++ b/lib/client.js
@@ -1281,128 +1281,6 @@ function _resume (client, sync) {
   }
 }
 
-async function writeIterable (client, request, socket, contentLength, header, expectsPayload) {
-  const { body } = request
-  assert(contentLength !== 0 || client[kRunning] === 0, 'iterator body cannot be pipelined')
-  socket[kWriting] = true
-
-  let callback = noop
-  function onDrain (err) {
-    if (!error && err) {
-      error = err
-    }
-    const actualCallback = callback
-    callback = noop
-    actualCallback()
-  }
-
-  const waitForDrain = () => new Promise((resolve) => {
-    assert(callback === noop)
-    callback = resolve
-  })
-
-  let bytesWritten = 0
-  let error
-  try {
-    socket.on('error', onDrain)
-      .on('drain', onDrain)
-    for await (const chunk of body) {
-      if (error || socket.destroyed) {
-        return
-      }
-
-      const len = Buffer.byteLength(chunk)
-      if (!len) {
-        continue
-      }
-
-      // TODO: What if not ended and bytesWritten === contentLength?
-      // We should defer writing chunks.
-      if (contentLength !== null && bytesWritten + len > contentLength) {
-        if (client[kStrictContentLength]) {
-          throw new RequestContentLengthMismatchError()
-        }
-
-        process.emitWarning(new RequestContentLengthMismatchError())
-      }
-
-      if (bytesWritten === 0) {
-        if (!expectsPayload) {
-          socket[kReset] = true
-        }
-
-        if (contentLength === null) {
-          socket.write(`${header}transfer-encoding: chunked\r\n`, 'ascii')
-        } else {
-          socket.write(`${header}content-length: ${contentLength}\r\n\r\n`, 'ascii')
-        }
-      }
-
-      if (contentLength === null) {
-        socket.write(`\r\n${len.toString(16)}\r\n`, 'ascii')
-      }
-
-      bytesWritten += len
-
-      if (socket.write(chunk)) {
-        continue
-      } else if (socket.destroyed || error) {
-        return
-      } else {
-        await waitForDrain()
-        if (error) {
-          return
-        }
-      }
-    }
-  } catch (err) {
-    error = err
-  } finally {
-    socket[kWriting] = false
-
-    if (!error && contentLength !== null && bytesWritten !== contentLength) {
-      if (client[kStrictContentLength]) {
-        error = new RequestContentLengthMismatchError()
-      } else {
-        process.emitWarning(new RequestContentLengthMismatchError())
-      }
-    }
-
-    socket.off('error', onDrain)
-      .off('drain', onDrain)
-    if (error) {
-      assert(client[kRunning] <= 1, 'pipeline should only contain this request')
-      util.destroy(socket, error)
-    }
-
-    if (!socket.destroyed) {
-      if (bytesWritten === 0) {
-        if (expectsPayload) {
-          // https://tools.ietf.org/html/rfc7230#section-3.3.2
-          // A user agent SHOULD send a Content-Length in a request message when
-          // no Transfer-Encoding is sent and the request method defines a meaning
-          // for an enclosed payload body.
-
-          socket.write(`${header}content-length: 0\r\n\r\n\r\n`, 'ascii')
-        } else {
-          socket.write(`${header}\r\n`, 'ascii')
-        }
-      } else if (contentLength === null) {
-        socket.write('\r\n0\r\n\r\n', 'ascii')
-      }
-
-      if (socket[kParser].timeout && socket[kParser].timeoutType === TIMEOUT_HEADERS) {
-        // istanbul ignore else: only for jest
-        if (socket[kParser].timeout.refresh) {
-          socket[kParser].timeout.refresh()
-        }
-      }
-
-      resume(client)
-    }
-  }
-}
-
 function write (client, request) {
   const { body, method, path, host, upgrade, headers } = request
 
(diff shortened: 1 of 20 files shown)


lib/client.js
```diff
@@ -1278,6 +1280,132 @@ function _resume (client, sync) {
...
+
+    finished = true
+    socket[kWriting] = false
+    socket.removeListener('error', onFinished)
+    finishChunkedWrite(err, contentLength, bytesWritten, expectsPayload, header, socket, client)
+  }
+
+  let thrownError = null
+  try {
+    socket.on('error', onFinished)
```
Linkgoron, line 1303: I wasn't sure if this is strictly needed, however there's similar logic in the stream path

lib/core/util.js
```diff
@@ -107,12 +107,22 @@ function getServerName (host) {
   return servername
 }
 
+function isIterable (obj) {
+  return !!(obj != null && (typeof obj[Symbol.iterator] === 'function' || typeof obj[Symbol.asyncIterator] === 'function'))
+}
+
+function isStrictIterable (obj) {
```
Linkgoron, line 114: couldn't think of a better name for this

lib/client.js
```diff
@@ -1278,6 +1280,132 @@ function _resume (client, sync) {
...
+    for await (const chunk of body) {
+      if (finished) {
+        return
+      }
+      const chunkSize = writeChunkHeaders(finished, chunk, contentLength, bytesWritten, expectsPayload, header, socket)
+      if (!chunkSize) {
+        continue
```
Linkgoron, line 1310: wasn't sure if this should be `continue` or `return`, in the original PR this was `return`

lib/client.js
```diff
@@ -1278,6 +1280,132 @@ function _resume (client, sync) {
...
+    finished = true
+    socket[kWriting] = false
+    socket.removeListener('error', onFinished)
+    finishChunkedWrite(err, contentLength, bytesWritten, expectsPayload, header, socket, client)
+  }
+
+  let thrownError = null
+  try {
+    socket.on('error', onFinished)
+    for await (const chunk of body) {
```
Linkgoron, line 1304: should `chunk` be converted to a buffer if it's a string? Should this throw if it's not a buffer or a string?

Linkgoron:
@mcollina Does it make sense call `body.return()` on a body Iterator in places that `util.destroy(body)` is called on a stream (e.g. when throwing in the request constructor etc)?

mcollina:
> @mcollina Does it make sense call `body.return()` on a body Iterator in places that `util.destroy(body)` is called on a stream (e.g. when throwing in the request constructor etc)? I'm not sure if it makes sense to add the logic to `util.destroy`, or somewhere else.

I think so, yes. wdyt @ronag?

Linkgoron:
Should the `should not follow redirects when using Readable request bodies` behaviour (which I _assume_ exists because of "backpressure" or missing some data) also be ported to async-iterators? It's simple enough, I just wasn't sure if it's needed for async iterators as well.

Linkgoron:
> > @mcollina Does it make sense call `body.return()` on a body Iterator in places that `util.destroy(body)` is called on a stream (e.g. when throwing in the request constructor etc)? I'm not sure if it makes sense to add the logic to `util.destroy`, or somewhere else.
> 
> I think so, yes. wdyt @ronag?

After looking at this a bit more, I'm not 100% sure that it's needed. As Undici gets an async-iterable, the iterable _supposedly_ should only create the resources when iteration begins (i.e. `symbol.Iterator` or `symbol.asyncIterator`) are called. The other option is to create an iterator/asyncIterator from the body and immediately call `return` on it.

ronag (changes requested):
Can you key the existing stream code and make a separate path for iterables?

Linkgoron:
I broke up the `for await...of` to manual iteration, so that it could be closed on `socket` error (which is essentially the same behaviour as the stream body path that destroys the body on socket error). The code is a bit more complicated, as correct handling of `return` is now needed, so pointers to simplify it are welcome.

lib/client.js
```diff
@@ -1278,6 +1280,149 @@ function _resume (client, sync) {
...
+      } else {
+        await EventEmitter.once(socket, 'drain')
+      }
+    }
+  } catch (err) {
+    thrownError = err
+  } finally {
+    onFinished(thrownError)
+  }
+}
```
ronag, line 1424: Implementation is simpler if you just use a for await loop instead of manually handling the iterator.

Linkgoron: Yes of course, but then I won't be able to close the iterator on socket error as I won't have access to the iterator.

Linkgoron: Technically, now that I think about it, I may have over-complicated it as once the next iteration beginds ends the for...await will just exit by itself anyway.

Linkgoron: that's what happens when you write code in the middle of the night...

test/client-pipelining.js
```diff
@@ -479,10 +562,6 @@ test('pipelining empty pipeline before reset', (t) => {
         t.equal(ended, true)
       })
 
-      const body = new Readable({
```
Linkgoron, line 482: was unused

Linkgoron:
I think that I've essentially done what's needed with the implementation. There are some open outstanding issues that I opened that I wasn't sure how to solve, if they're correct or if they're needed (e.g. redirect), and I tried finding and porting all of the `stream` body tests that I could find and seemed relevant (I looked for `createReadStream`, `new Readable` and `createReadable`), and I also added a few extra tests.

PTAL and tell me if there are any outstanding issues, if something I did was way off in general, or if I should add benchmarks etc.

lib/client.js
```diff
@@ -247,8 +248,8 @@ class Client extends Dispatcher {
       this[kQueue].push(request)
       if (this[kResuming]) {
         // Do nothing.
-      } else if (util.isStream(request.body)) {
-        // Wait a tick in case stream is ended in the same tick.
+      } else if (util.isStream(request.body) || util.isStrictIterable(request.body)) {
```
ronag, line 251: This is not needed for iterators

lib/client.js
```diff
@@ -1278,6 +1280,120 @@ function _resume (client, sync) {
...
+      } else {
+        await EventEmitter.once(socket, 'drain')
+      }
+    }
+  } catch (err) {
+    thrownError = err
+  } finally {
+    onFinished(thrownError)
+  }
+}
```
ronag, line 1395: Look at the pump method in pipeline.js in node core

lib/client.js
Linkgoron: I thought so as well, but there's a test the fails without this. However, it an async-iterator that wraps a stream so it might be related. I'll take another look.

Linkgoron: Specifically, `pipelining idempotent busy async-iterator` (`pipeliningIdempotentBusy` in `client-pipelining.js`) doesn't pass.

ronag: I guess we need it then.

lib/client.js
Linkgoron: Thanks, that really simplified moving everything to just one flow instead of having the `finished` variable. I didn't take the `AggregateError` part, as it's only in Node 16+, and I thought that it would be weird having type-guards and two different errors for two different Node versions.

Linkgoron: Note that I didn't add a `close` listener, which exists in `pump` but doesn't exist in the `stream` version. If a premature close error makes sense, I assume that it would make sense to add it in a following PR to both stream and iterator paths.

lib/client.js
ronag: No

lib/client.js
ronag: continue is correct

lib/client.js
```diff
@@ -1278,6 +1281,128 @@ function _resume (client, sync) {
...
+    }
+
+    socket.off('error', onDrain)
+      .off('drain', onDrain)
+    if (error) {
+      assert(client[kRunning] <= 1, 'pipeline should only contain this request')
+      util.destroy(socket, error)
+    }
+
+    if (!socket.destroyed) {
```
ronag, line 1378:
```suggestion
    if (socket.destroyed) {
      return
    }
```
reduce nesting

lib/client.js
```diff
@@ -1278,6 +1281,128 @@ function _resume (client, sync) {
...
+          // no Transfer-Encoding is sent and the request method defines a meaning
+          // for an enclosed payload body.
+
+          socket.write(`${header}content-length: 0\r\n\r\n\r\n`, 'ascii')
+        } else {
+          socket.write(`${header}\r\n`, 'ascii')
+        }
+      } else if (contentLength === null) {
+        socket.write('\r\n0\r\n\r\n', 'ascii')
+      }
```
ronag, line 1392: This should only happen if socket didn't error so you can move it inside the try clause.

lib/client.js
ronag: Looks good!

lib/client.js
Linkgoron: Couldn't do it because it's in a finally, and the linter doesn't allow returns in a finally as it's unsafe. Is it OK to add a `eslint-disable-next-line`?

ronag:
@Linkgoron I helped out a little

lib/client.js
```diff
@@ -1575,6 +1575,15 @@ async function writeIterable ({ client, request, socket, contentLength, header,
     cb()
   }
 
+  function onError (err) {
+    // TODO (fix): Avoid using err.message for logic.
+    if (err && (err.code !== 'UND_ERR_INFO' || err.message !== 'reset')) {
+      util.destroy(body, err)
```
Linkgoron, line 1581: are you sure about this? body is an `iterable` and not a stream, so it doesn't need to get destroyed

ronag (approved):

lib/client.js
ronag: Good point.

lib/client.js
```diff
@@ -1587,6 +1587,8 @@ async function writeIterable ({ client, request, socket, contentLength, header,
 
   let bytesWritten = 0
   try {
+    // TODO (fix): What if socket errors while waiting for body?
```
Linkgoron, line 1590: Technically, once the next yield happens, the `for...await of` will take care of it (as it throws socket[kError])

ronag: Yes, but body could be doing a lot of work before the next yield happens.

Linkgoron: I tried doing something smarter, but calling `return` doesn't affect the iterator at all until it yields anyway, so it somehow has to be in userland.

lib/client.js
```diff
@@ -1404,150 +1407,276 @@ function write (client, request) {
...
+  socket
+    .on('close', onDrain)
+    .on('drain', onDrain)
+
+  let bytesWritten = 0
+  try {
+    // TODO (fix): What if socket errors while waiting for body?
+    // It's up to the user to somehow abort the async iterable.
+    for await (const chunk of body) {
+      if (socket[kError]) {
```
Linkgoron, line 1593: Is this enough? Could the socket somehow get closed without setting kError?

ronag: No, we ensure it always has error

ronag (approved):

Issue #858, opened 2021-06-30
Missing maxRedirections in typedef ?


fraxken:
## Bug Description

I want to use the maxRedirections argument from request method but the argument doesn't exist in the Type definition. So i don't known if this is a new feature or maybe something deprecated.

Documentation: https://undici.nodejs.org/#/?id=undicirequesturl-options-promise
Version: undici 4.1.0

## Expected Behavior

No warning from TypeScript.

## Logs & Screenshots

![image](https://user-images.githubusercontent.com/4438263/124031104-2f617e00-d9f7-11eb-8933-779db80b0cd7.png)

## Environment

Node.js v14.15.4 with Windows 10.

---

I guess the argument is missing here ?: https://github.com/nodejs/undici/blob/main/types/dispatcher.d.ts#L40

Thanks for the time.

Best Regards,
Thomas

ronag:
PR is welcome

ronag:
Merged.

PR #859, opened 2021-07-01
fix: add missing maxRedirections type definition

fraxken:
PR to fix issue reported in #858


mcollina (changes requested):

fraxken:
Oh yeah my bad didn't see that they were UT for types too.

fraxken:
Ok just made an attempt (not sure if this is ok because first time doing UT for types ^^).

mcollina (approved):
lgtm

Issue #399, opened 2020-09-04
HTTP2/3 Support


wesleytodd:
Is there any interest in adding additional api's for http2/3 support?  While I understand this is a new domain for this package to cover, if we want to get folks moving over to what will someday be our unified http server apis, it would be nice to have a corresponding client which can nicely handle all three.

I feel like the current api's are pretty close to being ready for the underlying paradigm change, and it would be an especially nice UX if the client handled the ALPN swap over via the `alt-svc` headers.  I could see this fitting pretty easily in the `.dispatch` api with just a small addition to support push streams, and then we would have to consider how to map headers which changed or are no longer allowed.  Both of those seem pretty simple to reason about.

What other things am I missing here?  There must be something.  Anyway, just wanted to have a place to have this discussion since I couldn't find an existing issue.

ronag:
> Is there any interest in adding additional api's for http2/3 support

Maybe. Depends on the scope of changes and work required. I'm not familiar with http2/3 at all and would probably need help/involvement from e.g. @jasnell to sort that out.

I'm definitely interested in trying http(s) over QUIC.

wesleytodd:
I think http1 over QUIC is the best route for the long term if it works well, [it is still up in the air if http2 over QUIC will work](https://twitter.com/jasnell/status/1301723330072817666?s=20).  That said, as we move forward with the work on the server side I think it will be difficult for us to drive adoption if we don't also have a well supported client.  I haven't read much of the internals for `undici`, so I was only thinking of the top level api with my comment above.  I will try to carve out some time to dig into the implementation details here.

ronag:
@wesleytodd: I'm a little unsure if it's even worth spending any time on http2? There is very little legacy out there (unlike http1) and no advantages over http3.

Right now it just feels like something we do to make it "complete" and we as programmers don't like to see an empty slot between 1 and 3 😄 .

What I'm asking is whether it's maybe better to focus our energy on http1 and http3 and not "waste" it on http2. Or am I to radical in this thought?

wesleytodd:
I completely agree!  My thinking is that http2 is similar enough to 3 that we might get it at least at minimal effort, but I recognize that might be a naive idea.  I don't think it should be a focus.

As I spend some time digging into the server side over the next few weeks I will make sure to also carve out some time to understand what needs a client would have and read the source here to see how much work the addition will be.  Also, happy to have anyone else take a look as well and collaborate on what a solution would look like.

szmarczak:
[...]

PR #866, opened 2021-07-06
feat: HTTP/2

metcoder95:
Inherited from #720 

This relates to...
HTTP/2-3 #399 

## Rationale
HTTP/2 support is one of the few things currently missing from Undici
that sets it behind native Node http.

## Changes
This pull request is a work in progress. There are no changes just
yet. The plan, however, as discussed in the original issue, is to
implement HTTP/2 support around a Dispatcher instance for Pool,
Client and Agent.

Implementation can work in one of the following ways (see RFC7540 and
RFC8740):

## Prior knowledge
- Servers using HTTP/2 can advertise the capability via methods like ALT-SVC
Under RFC7540 Section 3.5, the HTTP/2 connection preface must be sent
All future frames may immediately use the HTTP/2
HTTP/1.1 Initialization (servers where HTTP/2 support status is unknown)
Without prior knowledge, the client should start an HTTP/1.1 connection
using the Upgrade header (h2c for http URI schema, h2 for TLS) and
HTTP2-Settings header
- If the response is a HTTP 101 Switching Protocols request, the request
will be accompanied by a response to prior requests and the connection can
upgrade to HTTP/2. If the response is normal, exempli gratia, a HTTP 200 OK,
the client must not switch to HTTP/2.
As for HTTPS, TLS negotiation must occur prior to any upgrade to HTTP/2.

## Features
HTTP/2 variants of Pool, Client and Agent. Alternatively, an option in the
aforementioned classes that causes them to attempt a HTTP/2 upgrade.

## Bug Fixes
N/A.

N/A.

##Status

- [ ] I have read and agreed to the Developer's Certificate of Origin
- [ ] Tested
- [ ] Benchmarked (optional)
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1395,7 +1395,11 @@ function write (client, request) {
   let header
 
   if (upgrade) {
-    header = `${method} ${path} HTTP/1.1\r\nconnection: upgrade\r\nupgrade: ${upgrade}\r\n`
+    let connectionString = "upgrade";
+    if (headers.toLowerCase().includes("http2-settings")) {
+      connectionString += ", http2-settings";
+    }
+    header = `${method} ${path} HTTP/1.1\r\nconnection: ${connectionString}\r\nupgrade: ${upgrade}\r\n`
   } else if (client[kPipelining]) {
     header = `${method} ${path} HTTP/1.1\r\nconnection: keep-alive\r\n`
   } else {
--- a/lib/client.js
+++ b/lib/client.js
@@ -1395,9 +1395,9 @@ function write (client, request) {
   let header
 
   if (upgrade) {
-    let connectionString = "upgrade";
-    if (headers.toLowerCase().includes("http2-settings")) {
-      connectionString += ", http2-settings";
+    let connectionString = 'upgrade'
+    if (headers.toLowerCase().includes('http2-settings')) {
+      connectionString += ', http2-settings'
     }
     header = `${method} ${path} HTTP/1.1\r\nconnection: ${connectionString}\r\nupgrade: ${upgrade}\r\n`
   } else if (client[kPipelining]) {
--- a/lib/http2client.js
+++ b/lib/http2client.js
@@ -1,0 +1,50 @@
+const FRAME_TYPES = [
+	"DATA", // standard request / response payloads
+	"HEADERS", // starts stream, carries request headers
+	"PRIORITY", // advises the priority of a stream
+	"RST_STREAM", // immediately terminates a stream
+	"SETTINGS", // inform other end of endpoint config
+	"PUSH_PROMISE", // pre-warns peer of wanted streams
+	"PING",
+	"GOAWAY", // graceful version of RST_STREAM
+	"WINDOW_UPDATE", // defines flow-control
+	"CONTINUATION" // continue a sequence of headers
+];
+
+function parseHttp2Settings(settings) {
+	// under RFC7540 section 3.2.1, the settings must be a base64url encoded
+	// SETTINGS frame. 16 bit identifiers, 32 bit values. */
+	let parsed = "";
+	for (let i = 0; i < Object.keys(settings).length; i += 2) {
+		parsed += `${settings[i]}${settings[i + 1]}\r\n`;
+	}
+
+	// base64url polyfill, courtesy of @panva
+	let encoded;
+	if (Buffer.isEncoding("base64url")) {
+		encoded = Buffer.from(parsed)
+			.toString("base64url")
+			.replace(/[=]+$/g, "");
+	} else {
+		encoded = Buffer.from(parsed)
+			.toString("base64")
+			.replace(/\+/g, "-")
+			.replace(/\//g, "_")
+			.replace(/[=]+$/g, "");
+	}
+	return encoded;
+}
+
+function http2Connect(url, options) {
+	if (url.protocol === "https:") {
+		throw new Error("Invalid protocol - httpConnect upgrades http streams");
+	}
+	throw new Error("UNIMPLEMENTED");
+}
+
+function https2Connect(url, options) {
+	if (url.protocol === "http:") {
+		throw new Error("Invalid protocol - https2Connect upgrades https streams");
+	}
+	throw new Error("UNIMPLEMENTED");
+}
--- a/lib/http2client.js
+++ b/lib/http2client.js
@@ -43,19 +43,24 @@ function parseHttp2Settings(settings) {
 
 function http2Connect(client) {
 	assert(!client[kSocket]);
-	const { protocol, port, hostname } = client[kUrl];
+	const { protocol, port, hostname, pathname } = client[kUrl];
 	if (protocol === "https:") {
 		throw new Error("Invalid protocol - httpConnect upgrades http streams");
 	}
 	// TODO: allow ALT-SVC
 	client.upgrade({
+		path: pathname,
 		protocol: "h2c",
 		headers: {
-			"Connection": "Upgrade, HTTP2-Settings",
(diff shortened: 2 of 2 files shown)


Nytelife26:
Good luck :)

MetCoder95:
Might sound naive but does anybody know an easy way to test an upgrade from `http` to `http2`. As the last one is executed to run under the security layer is not so easy to trigger for `h2c` upgrades. Otherwise, I'll need to create a small server for it (just looking for some already made solution before making a handmade one)

cc: @ronag @mcollina

mcollina:
I don't have a clue... maybe @jasnell do.

szmarczak:
@MetCoder95 What security layer are you talking about?

MetCoder95:
@szmarczak I was referring that trying to visit an HTTP2 website forces you to go under TLS, but I'm trying to test the upgrade from HTTP/1.1 to HTTP/2 with `h2c` upgrade headers

szmarczak:
If you're running h2c then you don't need TLS

MetCoder95:
Yes sure. I'm aware of that, I was just asking if there's any open server or tool where I can do upgrades by `h2c` as HTTP/2 by default this runs under the secure version. If not, I'll need to create a small server by myself 😅

MetCoder95:
How the `http2client` should look like? Exactly equals to Client? Or we can re-use the client implementation but just changing the way we parsed the exchange of data? I think I'm missing the big picture of that

MetCoder95:
Just out of curiosity, after the `upgrade` the way how the HTTP/2 data exchange is done is quite different from an HTTP/1.1, from this point how do we proceed with the parser, is `lhttp` supporting it? I'm asking because the `upgrade` is basically done for `h2c`, the next step is to continue with the parsing of the different responses/requests (and of course `h2`).

Sorry for taking some time, I've been having rush days, I'll try to start being more concurrent next week :)

cc: @mcollina @ronag

ronag:
> from this point how do we proceed with the parser, is lhttp supporting it?

You should bypass all of that. llhttp does not support http2.

MetCoder95:
Got it, just to understand once for all, reading at the issue from where this PR comes out, I do agree with all the points stated there (mostly how HTTP/2 is quite more like a transition), then my question is, what and how do we achieve HTTP/2 support?

Besides `upgrade` which seems straightforward, we should write a parser to accept the full HTTP/2 spec or at least the most basic part?
I'm asking because of how HTTP/2 works, the different frames for headers and body, the support for duplex, headers, etc. 

Or maybe use `http2` from Node internals?

The full topic is quite interesting but I do agree with the points explained in the issue 😄

ronag:
You should use http2 from node internals.

szmarczak:
> we should write a parser to accept the full HTTP/2 spec

I don't think we need to write another parser. It's already provided by Node.js, which is a good start. The thing is that when using `h2c` the server MUST reply with the response - which is kinda hard to convince the native module to accept the HTTP/2 response with the HTTP/1 request.

If I'm missing something, then it may be easier to write another parser. /cc @ronag 

HTTP/2 is way more complicated than HTTP/1 so it will take definitely more time than implementing HTTP/1.

MetCoder95:
>I don't think we need to write another parser. It's already provided by Node.js, which is a good start. The thing is that when using h2c the server MUST reply with the response - which is kinda hard to convince the native module to accept the HTTP/2 response with the HTTP/1 request.

Actually, for those scenarios, I was asking for. This also includes, e.g. checks for prior requests if an HTTP 101 is response before the upgrade of protocols. 

>which is kinda hard to convince the native module to accept the HTTP/2 response with the HTTP/1 request.

Hmm, correct me if I'm wrong please but I think an HTTP/1.1 101 will be the response and the next exchange of data should be in the HTTP/2 protocol. Then we can "pipeline" the connection to the native module, or I'm way far?

>HTTP/2 is way more complicated than HTTP/1 so it will take definitely more time than implementing HTTP/1.

Agree

szmarczak:
> Then we can "pipeline" the connection to the native module, or I'm way far?

Yeah but you need to accept the response in HTTP/2 somehow without the request in HTTP/2. But the actual request is via HTTP/1. So we need to create a stream that has been already sent and is awaiting response.

mcollina:
I would recommend we focus on supporting http/2 -> http/2 before working on upgrade paths.

@jasnell have you done any research on implementing the upgrade in nodejs?

Nytelife26:
> have you done any research on implementing the upgrade in nodejs?

All information required to implement the upgrade has been summarized in the original PR, before I handed off the project.
Particularly, [this comment](https://github.com/nodejs/undici/pull/720#issuecomment-819474757) should be of interest.

> You should use http2 from node internals

[Didn't we already discuss that being an idea that is somewhat antithetical to undici's purpose](https://github.com/nodejs/undici/pull/720#issuecomment-817217020)?

Nytelife26:
It seems like this PR was opened solely to get around not having write permissions for the branch of the other PR. Instead of making a duplicate, if you'd like @MetCoder95 I can add you to the branch?

I'm noticing a lot of lost information and repeated discussion due to the conversations from the old PR being neglected, so that's my main concern here.

szmarczak:
> Didn't we already discuss that being an idea that is somewhat antithetical to undici's purpose?

I don't see anyone agreed to write a HTTP/2 parser from scratch.

> node's native http2 would be a bottleneck given its inferior performance

I'd rather take that 20% drop and prefer `http/1.1` over `h2` on ALPN negotiation rather to write a brand new parser for exactly this reason. I don't consider it worth the time.

Nytelife26:
> I don't see anyone agreed to write a HTTP/2 parser from scratch.

That's what was started here. It's necessary to parse HTTP/2 transport frames in order to successfully carry out the upgrade, so we'd have to parse the frames for the upgrade anyway.

> I'd rather take that 20% `h2` drop and prefer `http/1.1` over `h2` on ALPN negotiation rather to write a brand new parser for exactly this reason. I don't consider it worth the time.

Once again, we have to do most of it anyway. It wouldn't be too much extra effort to complete it for the performance boost. After all, undici's goal is to provide a fast alternative to native.

MetCoder95:
>I would recommend we focus on supporting http/2 -> http/2 before working on upgrade paths.

By using the Node.js internal module? Then a wrapper?

>All information required to implement the upgrade has been summarized in the original PR, before I handed off the project.
Particularly, this comment should be of interest.

Thanks! Actually, from the original PR, I'm basing most of the assumptions I'm doing on how the API should look, but the issue comes the way how we need to parse the frames from HTTP/2, which is where the topic gets complicated.

>It seems like this PR was opened solely to get around not having write permissions for the branch of the other PR. Instead of making a duplicate, if you'd like @MetCoder95 I can add you to the branch?

Yeah, sure, no issue. Is better to merge all the conversations into a single place 👍 

---------

If I can add my two cents here with my limited knowledge, I think that we should re-defined a little bit the scope of what we want to achieve. It is just to add support for the upgrade with a wrapper around the Node `http2` module, I do agree that is the right step to do. The complexity is quite lower (having in mind the complexity of handling two streams at the same time somehow), and we can have the feature in place in a shorter time.
On the other hand, if we want to add full support for HTTP/2, well we should consider what's the state of the protocol towards HTTP/3 as stated in the issue linked to the PR. But if we still want to go in that direction, yeah maybe a parser from scratch will be better as otherwise we'll be just making a wrapper around it `http2` (which doesn't have any issue at all but worth it to think if is really needed). 
Are just my two cents, hope I'm right in something 😅

Nytelife26:
I invited you to the repository, so you'll be able to make changes @MetCoder95.

Issue #867, opened 2021-07-08
Uncatchable error is thrown when a HTTPS post request fails


SammyWhamy:
## Bug Description
When making an HTTPS post request to a URL that doesn't respond/exist, an AssertionError is thrown that cannot be caught by the consumer.

## Reproducible By
Wrap a post request to an unreachable URL in a try...catch
Observe how the script is terminated without the error being caught

### Code example:
```js
import {request} from 'undici'
try {
  await request("https://thisis.not/avalid/url", {method: "POST");
} catch {
  console.log("Error caught, this never happens");
}
```

## Expected Behavior
The promise should be rejected so that the error can be caught and handled by the consumer

## Logs
```
AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:

  assert(this.servername)

    at TLSSocket.<anonymous> (C:\path\node_modules\undici\lib\core\connect.js:61:11)
    at TLSSocket.emit (node:events:377:35)
    at emitErrorNT (node:internal/streams/destroy:193:8)
    at emitErrorCloseNT (node:internal/streams/destroy:158:3)
    at processTicksAndRejections (node:internal/process/task_queues:83:21) {
  generatedMessage: true,
  code: 'ERR_ASSERTION',
  actual: null,
  expected: true,
  operator: '=='
}
```

## Environment
Windows 10, Node v16.2.0

vocafeuvre:
I'm getting this too, it's caused by an assert call inside the error handler of the socket created by the request. Right now I'm commenting that line as a workaround

mcollina:
@vocafeuvre would you like to send a PR?

joaopedrocampos:
Hey guys! If still up to, I would love to try this fix 🙏

mcollina:
go for it!

vocafeuvre:
@mcollina yeah i have the fix in my fork, i'm opening a PR now

ronag:
Merged fix

PR #869, opened 2021-07-11
Fix/bad assert in connect

vocafeuvre:
Linked to #867. Added a test in test/agent.js to confirm if fix was working.

### DIFF
--- a/lib/core/connect.js
+++ b/lib/core/connect.js
@@ -58,7 +58,7 @@ class Connector {
           cache.set(this.servername, session)
         })
         .on('error', function (err) {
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

lib/core/connect.js
```diff
@@ -58,7 +58,6 @@ class Connector {
           cache.set(this.servername, session)
         })
         .on('error', function (err) {
-          assert(this.servername)
           if (err.code !== 'UND_ERR_INFO') {
```
ronag, line 61:
```suggestion
          if (this.servername && err.code !== 'UND_ERR_INFO') {
```

PR #876, opened 2021-07-11
feat: onRedirect

ronag:


ronag:
@szmarczak started a bit in case you want to have a try a it. I think it's mostly the test missing.

szmarczak:
I wonder if it would be ok to include `this.history`, `statusCode` and `headers` as well.

- `this.history` can be used to detect infinite loops
- `statusCode` & `headers` - for handling [`300 Multiple Choices`](https://datatracker.ietf.org/doc/html/rfc7231#section-6.4.1)

Linkgoron:
maybe `onRedirect` should be able to "short-circuit" the redirection (e.g. by returning `false`)?

PR #878, opened 2021-07-11
fix: don't send extra \r\n

ronag:
Refs: https://github.com/nodejs/undici/discussions/858


artur-ma:
Can u take a look on this one, by any chance?
https://github.com/nodejs/undici/discussions/857

looks like its the same area, and similar issue

mcollina (approved):
lgtm

PR #881, opened 2021-07-15
fix: unify connect options

ronag:
Should be fully backwards compatible.


ronag:
I had to remove a test as it incorrectly relies on mutating options.

ronag:
@szmarczak

mcollina (approved):
lgtm

dnlup:
I am late to this, sorry. Great work!

lib/core/connect.js
```diff
@@ -12,30 +12,30 @@ const { InvalidArgumentError } = require('./errors')
...
-  constructor ({ tls, socketPath, maxCachedSessions }) {
-    this.tls = tls || {} // TODO: Make shallow copy to protect against mutations.
-
+  constructor ({ maxCachedSessions, socketPath, ...opts }) {
     if (maxCachedSessions != null && (!Number.isInteger(maxCachedSessions) || maxCachedSessions < 0)) {
       throw new InvalidArgumentError('maxCachedSessions must be a positive integer or zero')
     }
 
-    this.socketPath = socketPath
+    this.opts = { path: socketPath, ...opts }
```
szmarczak, line 20: Wouldn't it be faster to:

```js
opts.path = socketPath;
this.opts = opts;
```

szmarczak: /cc @ronag

NN-Binary:
Great job! Any manual on how to change the options?

lib/core/connect.js
ronag: We should avoid mutating user owned objects. I don’t think this is a hot path.

ronag:
> Great job! Any manual on how to change the options (I need to use a custom lookup function) ?

The `connect` option will be passed into `net.connect` so you can just do `connect: { lookup }`.

NN-Binary:
I'm not sure how to use it in Dispatch, I've just updated to the latest Undici, how do I add lookup in Dispatch?
Thank you

ronag:
You don't need to use dispatch. Just pass `connect` option to `Agent` constructor. Don't forget you need to run off master since we haven't made a release yet.

NN-Binary:
Wow, thank you so much for the quick reply.
Right now my code is quite simple:

```

                var HTTPClient = new Client('http://google.com')
                HTTPClient.dispatch({
                    path: '/',
                    method: 'GET',
                    headers: {
                        // Headers
                    },
                    lookup: CUSTOM_FUNCTION()
                }, {
                    ...
                })
```

Agent should be placed in the new Client? Sorry

ronag:
```js
const client = new Client('http://google.com', {
  connect: { lookup: CUSTOM_FUNCTION }
})

const { body, statusCode, headers } = await client.request({
                    path: '/',
                    method: 'GET',
                    headers: {
                        // Headers
                    },
})

```

NN-Binary:
Oh! You are amazing. thank you so much @ronag

NN-Binary:
EDIT: When getting the Master branch, it seems that client doesn't exist anymore.

```
import { request, Client  } from 'undici';
                  ^^^^^^
SyntaxError: Named export 'Client' not found. The requested module 'undici' is a CommonJS module, which may not support all module.exports as named exports.
```

npm install git://https://github.com/nodejs/undici/tree/master --save

This is the right branch? correct?

RuralDependencies:
> EDIT: When getting the Master branch, it seems that "Client" doesn't exist anymore.
> 
> ```
> import { request, Client  } from 'undici';
>                   ^^^^^^
> SyntaxError: Named export 'Client' not found. The requested module 'undici' is a CommonJS module, which may not support all module.exports as named exports.
> ```
> 
> npm install git://github.com/nodejs/undici/tree/master --save
> No problem here, it fetches Undici but Client Method doesn't exist.
> 
> This is the right branch? correct?
> Did anything changed between the current version and Master? In term of how I should import it?
> 
> From what I'm seeing in index.js, a lot of things changed especially in what is exported or not, is it normal?
> 
> Edit: it works if I change my imports to:
> import undici from 'undici';
> 
> Is it normal that the behavior is different or it's because I don't do it properly?
> 
> Well it seems to not work at all anymore, even changing new Client by new undici.Client, I keep getting weird errors
> 6: 0xd61c66 v8::internal::Builtin_HandleApiCall(int, unsigned long*, v8::internal::Isolate*) [/usr/bin/node]
> 
> Sorry I'm lost

The branch that works with the option you are looking for is here:
https://github.com/nodejs/undici/tree/refactor-write

ronag:
Please open a separate issue.

PR #892, opened 2021-07-19
fix: accept URL origin in Agent

### LINKED ISSUE #889: Agent.dispatch should allow URL origin
Right now only `string` works but the type definitions say that `URL` is also supported.

szmarczak:
Fixes #889

```
szm@solus ~/Desktop/undici $ node benchmarks/benchmark.js 
│ Tests               │ Samples │           Result │ Tolerance │ Difference with slowest │
|─────────────────────|─────────|──────────────────|───────────|─────────────────────────|
│ http - no keepalive │      60 │  5837.63 req/sec │  ± 2.88 % │                       - │
|─────────────────────|─────────|──────────────────|───────────|─────────────────────────|
│ http - keepalive    │      15 │  9187.24 req/sec │  ± 2.44 % │               + 57.38 % │
|─────────────────────|─────────|──────────────────|───────────|─────────────────────────|
│ undici - pipeline   │      25 │ 13515.93 req/sec │  ± 2.76 % │              + 131.53 % │
|─────────────────────|─────────|──────────────────|───────────|─────────────────────────|
│ undici - request    │      10 │ 18800.16 req/sec │  ± 2.41 % │              + 222.05 % │
|─────────────────────|─────────|──────────────────|───────────|─────────────────────────|
│ undici - stream     │      20 │ 20052.72 req/sec │  ± 2.55 % │              + 243.51 % │
|─────────────────────|─────────|──────────────────|───────────|─────────────────────────|
│ undici - dispatch   │      10 │ 22855.44 req/sec │  ± 2.85 % │              + 291.52 % │
```

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -104,7 +104,12 @@ class Agent extends Dispatcher {
         throw new InvalidArgumentError('opts must be a object.')
       }
 
-      const { origin } = util.parseOrigin(opts.origin)
+      let key
+      if (opts.origin && (typeof opts.origin === 'string' || opts.origin instanceof URL)) {
+        key = String(opts.origin)
+      } else {
+        throw new InvalidArgumentError('opts.origin must be a non-empty string or URL.')
+      }
 
       if (this[kDestroyed]) {
         throw new ClientDestroyedError()
@@ -114,18 +119,18 @@ class Agent extends Dispatcher {
         throw new ClientClosedError()
       }
 
-      const ref = this[kClients].get(origin)
+      const ref = this[kClients].get(key)
 
       let dispatcher = ref ? ref.deref() : null
       if (!dispatcher) {
-        dispatcher = this[kFactory](origin, this[kOptions])
+        dispatcher = this[kFactory](opts.origin, this[kOptions])
           .on('drain', this[kOnDrain])
           .on('connect', this[kOnConnect])
           .on('disconnect', this[kOnDisconnect])
           .on('connectionError', this[kOnConnectionError])
 
-        this[kClients].set(origin, new WeakRef(dispatcher))
-        this[kFinalizer].register(dispatcher, origin)
+        this[kClients].set(key, new WeakRef(dispatcher))
+        this[kFinalizer].register(dispatcher, key)
       }
 
       const { maxRedirections = this[kMaxRedirections] } = opts
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -104,7 +104,7 @@ class Agent extends Dispatcher {
         throw new InvalidArgumentError('opts must be a object.')
       }
 
-      const { origin } = util.parseOrigin(opts.origin);
+      const { origin } = util.parseOrigin(opts.origin)
 
       if (this[kDestroyed]) {
         throw new ClientDestroyedError()
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -104,9 +104,7 @@ class Agent extends Dispatcher {
         throw new InvalidArgumentError('opts must be a object.')
       }
 
-      if (typeof opts.origin !== 'string' || opts.origin === '') {
-        throw new InvalidArgumentError('opts.origin must be a non-empty string.')
-      }
+      const { origin } = util.parseOrigin(opts.origin);
 
       if (this[kDestroyed]) {
         throw new ClientDestroyedError()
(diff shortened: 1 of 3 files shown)


szmarczak:
I wonder if it would be better to use `options.protocol`, `options.host` and `options.port` instead of `options.origin`. That way we don't have to use `new URL(...)` again.

szmarczak:
https://github.com/nodejs/undici/blob/aebbb5df86d5584b03102cb9198aad8abae421f1/lib/client.js#L161

is unsafe because:
https://github.com/nodejs/undici/blob/aebbb5df86d5584b03102cb9198aad8abae421f1/lib/core/util.js#L75-L76

https://github.com/nodejs/undici/blob/aebbb5df86d5584b03102cb9198aad8abae421f1/lib/core/util.js#L58-L72

meaning: it does not return `new URL(...)` if the we pass a `URL` instance. So the mutations are visible while they shouldn't be. What should we do?

ronag:
> https://github.com/nodejs/undici/blob/aebbb5df86d5584b03102cb9198aad8abae421f1/lib/client.js#L161
> 
> is unsafe because:
> https://github.com/nodejs/undici/blob/aebbb5df86d5584b03102cb9198aad8abae421f1/lib/core/util.js#L75-L76
> 
> https://github.com/nodejs/undici/blob/aebbb5df86d5584b03102cb9198aad8abae421f1/lib/core/util.js#L58-L72
> 
> meaning: it does not return `new URL(...)` if the we pass a `URL` instance. So the mutations are visible while they shouldn't be. What shall we do?

Make `parseURL` always copy?

lib/agent.js
```diff
@@ -104,9 +104,7 @@ class Agent extends Dispatcher {
         throw new InvalidArgumentError('opts must be a object.')
       }
 
-      if (typeof opts.origin !== 'string' || opts.origin === '') {
-        throw new InvalidArgumentError('opts.origin must be a non-empty string.')
-      }
+      const { origin } = util.parseOrigin(opts.origin)
```
ronag, line 107: I'm a little concerned about the performance implications. Is it negligable?

ronag (approved):

ronag:
> Fixes #889
> 
> ```
> szm@solus ~/Desktop/undici $ node benchmarks/benchmark.js 
> │ Tests               │ Samples │           Result │ Tolerance │ Difference with slowest │
> |─────────────────────|─────────|──────────────────|───────────|─────────────────────────|
> │ http - no keepalive │      60 │  5837.63 req/sec │  ± 2.88 % │                       - │
> |─────────────────────|─────────|──────────────────|───────────|─────────────────────────|
> │ http - keepalive    │      15 │  9187.24 req/sec │  ± 2.44 % │               + 57.38 % │
> |─────────────────────|─────────|──────────────────|───────────|─────────────────────────|
> │ undici - pipeline   │      25 │ 13515.93 req/sec │  ± 2.76 % │              + 131.53 % │
> |─────────────────────|─────────|──────────────────|───────────|─────────────────────────|
> │ undici - request    │      10 │ 18800.16 req/sec │  ± 2.41 % │              + 222.05 % │
> |─────────────────────|─────────|──────────────────|───────────|─────────────────────────|
> │ undici - stream     │      20 │ 20052.72 req/sec │  ± 2.55 % │              + 243.51 % │
> |─────────────────────|─────────|──────────────────|───────────|─────────────────────────|
> │ undici - dispatch   │      10 │ 22855.44 req/sec │  ± 2.85 % │              + 291.52 % │
> ```

Would be great if you did before/after.

ronag:
> I wonder if it would be better to use `options.protocol`, `options.host` and `options.port` instead of `options.origin`. That way we don't have to use `new URL(...)` again.

You could let `origin` be an object with those properties...

lib/agent.js
mcollina: Parsing URLs is _slow_, we should not be creating a fresh `URL` instance every time if possible.

ronag: origin is just used as a key here so we won't actually need to validate it.

ronag:
```
let origin
if (opts.origin && typeof opts.origin === 'string') {
  origin = opts.origin
} else if (opts.origin instanceof URL) {
  origin = opts.origin.origin
} else {
  throw new InvalidArgumentError('opts.origin must be a non-empty string or URL.')
}
```

mcollina (approved):
lgtm

Issue #893, opened 2021-07-19
A few questions


szmarczak:
1. We can provide `connect` function or `connect` options but not both. Why?
2. Why `options.origin` and not `options.protocol`, `options.host` and `options.port`?
3. https://github.com/nodejs/undici/blob/aebbb5df86d5584b03102cb9198aad8abae421f1/lib/core/request.js#L100-L127 these checks are missing in [RedirectHandler](https://github.com/nodejs/undici/blob/aebbb5df86d5584b03102cb9198aad8abae421f1/lib/handler/redirect.js#L9)
4. Isn't https://github.com/nodejs/undici/blob/aebbb5df86d5584b03102cb9198aad8abae421f1/lib/core/request.js#L130 moot because https://github.com/nodejs/undici/blob/aebbb5df86d5584b03102cb9198aad8abae421f1/lib/core/request.js#L77
5. nitpick: https://github.com/nodejs/undici/blob/aebbb5df86d5584b03102cb9198aad8abae421f1/lib/client.js#L1096 is slower by ~20% than `hostname[0] === '['`
6. https://github.com/nodejs/undici/blob/aebbb5df86d5584b03102cb9198aad8abae421f1/lib/client.js#L1397-L1399 can be removed because ee64b39b97e5c39369ba66a54ff8dffaa8cd3339?
7. Can we avoid object allocation somehow? I think that should improve the performance as well. I mean https://github.com/nodejs/undici/blob/aebbb5df86d5584b03102cb9198aad8abae421f1/lib/client.js#L1439-L1449
8. https://github.com/nodejs/undici/blob/aebbb5df86d5584b03102cb9198aad8abae421f1/lib/handler/redirect.js#L28 this reassures us that there is no body sent on redirect
9. I wonder if https://github.com/nodejs/undici/blob/aebbb5df86d5584b03102cb9198aad8abae421f1/index.js#L40-L65 can be removed

mcollina:
The vast majority of those are PRs that could be sent 🙏❤️

ronag:
1. Because if you provide a method you can save all params in the closure. Thus you don't need both.
2. Simplicity I guess...
3. PR
4. `host` is set through `processHeader`
5. PR
6. PR
7. I think V8 might handle that already. Don't think it makes much of a diff.
8. PR
9. Why you think that?

joaopedrocampos:
Hey guys, decided to try the fixes that @ronag said that needed a PR, in this case, the items 3 and 5

[...]

PR #894, opened 2021-07-20
Creating util function to validate handlers

joaopedrocampos:
Creates a util function to validate handler and adds it on redirect since it was missing.

Also changes the `hostname.startsWith('[')` for `hostname[0] === '['` since it's more performing.

Standardizes invalid handler error messages

Related issue: https://github.com/nodejs/undici/issues/893

### DIFF
--- a/lib/agent.js
+++ b/lib/agent.js
@@ -96,12 +96,12 @@ class Agent extends Dispatcher {
 
   dispatch (opts, handler) {
     if (!handler || typeof handler !== 'object') {
-      throw new InvalidArgumentError('handler')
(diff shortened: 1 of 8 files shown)


joaopedrocampos:
@ronag I created the handler validation function with the same conditions that were on `request.js`, but there are other places that make a simple validation's of it, e.g:

https://github.com/nodejs/undici/blob/main/lib/client.js#L286

So I don't know if it's ok to follow this approach I did or if it's better to create many "individual validation functions" (does this makes sense?) to use them in these cases

ronag (approved):

mcollina (approved):
lgtm

PR #895, opened 2021-07-21
feat: return body as consumable body

ronag:
Change undici.request to return a Body object with
various consume helpers instead of regular node stream.

This aligns with recent work on quic and web streams in
node core.

### DIFF
--- a/lib/api/api-request.js
+++ b/lib/api/api-request.js
@@ -25,7 +25,7 @@ class RequestBody extends Readable {
     this[kReadableDidRead] = false
 
     if (typeof this.readableDidRead !== 'boolean') {
-      EE.prototype.on.call(this, 'data', function () {
+      EE.prototype.once.call(this, 'data', function () {
         this[kReadableDidRead] = true
       })
     }
@@ -63,7 +63,6 @@ class Body {
     if (this.bodyUsed) {
       throw new TypeError('disturbed')
     }
-
     return this[kBody]
   }
 
@@ -72,7 +71,10 @@ class Body {
   }
 
   get body () {
-    return this.readableWebStream()
+    if (!this[kBody].toWeb) {
+      throw new TypeError('not supported')
+    }
+    return this[kBody].toWeb()
   }
 
   async blob () {
--- a/lib/api/api-request.js
+++ b/lib/api/api-request.js
@@ -66,7 +66,7 @@ class Body {
   }
 
   get bodyUsed () {
-    return this[kUsed]
+    return this[kBody].readableDidRead
   }
 
   get body () {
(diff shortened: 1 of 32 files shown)


ronag:
Tests need to be fixed quite extensively so I ask for some feedback before continuing down this route.

ronag:
@Ethan-Arrowood @szmarczak

lib/api/api-request.js
```diff
@@ -30,6 +36,84 @@ class RequestResponse extends Readable {
...
+    return JSON.parse(ret)
+  }
+
+  async json () {
+    // TODO: Optimize.
+    // TODO: Validate content-type req & res headers?
+    return JSON.parse(await this.text())
+  }
+}
+
```
ronag, line 116: The idea here is that whatever way the user decides to consume the body it will always have the best possible performance, i.e. it shouldn't always go through `readableNodeStream()`. Instead each should have an adapter against the native API. 

I used `readableNodeStream` to start with to get the general functionality.

lib/api/api-request.js
```diff
@@ -30,6 +38,114 @@ class RequestResponse extends Readable {
...
+      sources.push(chunk)
+    }
+    return new Blob(sources)
+  }
+
+  async arrayBuffer () {
+    // TODO: Optimize.
+    const blob = await this.blob()
+    return await blob.arrayBuffer()
+  }
```
ronag, line 124: Is this needed?

mcollina:
I think this should target a `next` branch

dnlup (commented):
Love this! I think it's a great improvement in usability.

lib/api/api-request.js
```diff
@@ -30,6 +38,114 @@ class RequestResponse extends Readable {
...
+
+  async text () {
+    // TODO: Optimize.
+    // TODO: Validate content-type req & res headers?
+    let ret = ''
+    for await (const chunk of this.readableNodeStream()) {
+      // TOOD: max size?
+      ret += chunk
+    }
+    return JSON.parse(ret)
```
mcollina, line 139: This should not parse the JSON

mcollina (commented):
I really like the concept.

ronag:
Related: https://github.com/nodejs/node/pull/39483

lib/api/api-request.js
```diff
@@ -30,6 +57,91 @@ class RequestResponse extends Readable {
...
+    this[kBody].destroy(err)
+  }
+
+  get stream () {
+    if (this[kBody].readableDidRead) {
+      throw new TypeError('disturbed')
+    }
+
+    return this[kBody]
+  }
```
ronag, line 82: This is almost like `BodyMixin` from the web standard with the small exception of how to consume as Node stream.... anyone got suggestions on how that could/should look? @szmarczak @Ethan-Arrowood @mcollina @dnlup

PR #896, opened 2021-07-22
always clear the connectTimeout

mcollina:
I have seen the close of my processes being severely delayed by this.
I'm not sure how to test this though.

### DIFF
--- a/lib/core/connect.js
+++ b/lib/core/connect.js
@@ -72,28 +72,24 @@ class Connector {
       })
     }
 
-    let timeout
-
-    if (this.timeout) {
-      timeout = setTimeout(onConnectTimeout, this.timeout, socket)
-      // Needed for Jest
-      if (timeout.unref) {
-        timeout.unref()
-      }
-    }
+    const timeout = this.timeout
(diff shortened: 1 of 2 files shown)


ronag (approved):

ronag:
I'm surprised the `client-unref` test doesn't catch this? Maybe that could be improved?

mcollina:
@ronag I've updated this.

ronag (approved):

lib/core/connect.js
```diff
@@ -79,15 +79,17 @@ class Connector {
...
+        if (callback) {
           const cb = callback
           callback = null
           cb(null, this)
         }
       })
       .on('error', function (err) {
+        clearTimeout(timeout)
+
         if (callback) {
```
ronag, line 93: Probably enough to just put a `clearTimeout(timeout)` here.

mcollina: Why?

ronag: The problem is that we don't clear it in the error case. In the non error case it is already cleared.

ronag: It's fine the way you did it as well.

Issue #868, opened 2021-07-08
Small issue with the links on the docs website


pedroapfilho:
There's an error in the documentation.

Whenever it loads for the first time, it adds a `/#/` at the end of the URL, so it's like: `https://undici.nodejs.org/#/`

When you go to the homepage and click on the `Example` link, it actually goes to `https://undici.nodejs.org/#/benchmarks/benchmark.js`, which is actually a `404` page. It should go to `https://undici.nodejs.org/benchmarks/benchmark.js`

mcollina:
Would you like to send a PR?

pedroapfilho:
That's the thing! I would love to! At first, I thought it would be an easy fix, but then I saw this: https://github.com/docsifyjs/docsify/issues/235 and realized that it might not be.

I'll think about it on the weekend, run the project locally and see if I can make these links go without the `/#/`

Thanks for the awesome work @mcollina, I'll submit the PR by the end of next week!

mcollina:
Thanks!

joaopedrocampos:
@mcollina should we close this issue now the PR is merged? 🙏

PR #897, opened 2021-07-23
fix: broken link for examples on website

joaopedrocampos:
Fixes the broken `examples` link for website

Related issue: https://github.com/nodejs/undici/issues/868

**Broken link (current behaviour)**
broken_example

**Fixed link**
fixed_example

Solution using `noCompileLinks` that was created from a similar problem. More details on Docsify issue: https://github.com/docsifyjs/docsify/issues/203

### DIFF
--- a/index.html
+++ b/index.html
@@ -22,7 +22,10 @@
       auto2top: true,
       subMaxLevel: 3,
       maxLevel: 3,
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

PR #899, opened 2021-07-23
Types: improve connect options

delvedor:
It looks like the `tls` options has been removed form the types by accident, this pr adds it back.

### DIFF
--- a/docs/api/Client.md
+++ b/docs/api/Client.md
@@ -29,8 +29,13 @@ Returns: `Client`
 
 #### Parameter: `ConnectOptions`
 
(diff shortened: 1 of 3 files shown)


ronag:
It was removed on purpose. Those options should go under ‘connect’ now. I just couldn’t figure out the type defs.

delvedor:
@ronag not sure I follow. It's still an option accepted by Pool.

ronag:
It works but is deprecated. The options should be passed in the connect object parameter.

delvedor:
@ronag I’ve updated the types to be less lenient and help the user understand where to put those options, the documentation has been updated as well. Let me know if it works for you.

ronag:
Thanks!

PR #901, opened 2021-07-25
Types: changes origin to be optional on Client and Pool

joaopedrocampos:
After PR https://github.com/nodejs/undici/pull/733 `origin` was set to be optional on Client and Pool. This PR fixes the type definitions of `origin` to be optional on `dispatch` and `request` of Client and Pool

Related issue: https://github.com/nodejs/undici/issues/900

Issue #644, opened 2021-03-30
dns lookup origin before picking Pool?


ronag:
Let's say I have two different dns names pointing to the same server:

```
await undici.request('http://mydomain1.com') // 192.168.0.11
await undici.request('http://mydomain2.com') // 192.168.0.11
```

Currently this will unnecessarily create two different pools duplicating the connections.

Is this worth spending time on?

ronag:
Leave it for later.

szmarczak:
I'm not sure if this is worth doing. Most of the servers are HTTPS now which require `servername` and the server may choose to send a certificate for the subdomain only.

PR #903, opened 2021-07-25
Revert "Types: changes origin to be optional on Client and Pool"

ronag:
Reverts nodejs/undici#901

### DIFF
--- a/types/client.d.ts
+++ b/types/client.d.ts
@@ -1,6 +1,6 @@
 import { URL } from 'url'
 import { TlsOptions } from 'tls'
-import Dispatcher, { DispatchOptions, RequestOptions } from './dispatcher'
+import Dispatcher from './dispatcher'
 
 export = Client
 
@@ -13,11 +13,6 @@ declare class Client extends Dispatcher {
   closed: boolean;
   /** `true` after `client.destroyed()` has been called or `client.close()` has been called and the client shutdown has completed. */
   destroyed: boolean;
-  /** Dispatches a request. This API is expected to evolve through semver-major versions and is less stable than the preceding higher level APIs. It is primarily intended for library developers who implement higher level APIs on top of this. */
-  dispatch(options: Client.ClientDispatchOptions, handler: Dispatcher.DispatchHandlers): void;
-  /** Performs an HTTP request. */
-  request(options: Client.ClientRequestOptions): Promise<Dispatcher.ResponseData>;
-  request(options: Client.ClientRequestOptions, callback: (err: Error | null, data: Dispatcher.ResponseData) => void): void;
 }
 
 declare namespace Client {
@@ -49,13 +44,5 @@ declare namespace Client {
     socketPath?: string | null;
     timeout?: number | null;
     servername?: string | null;
-  }
-
-  export interface ClientDispatchOptions extends Partial<DispatchOptions> {
-    origin?: string | URL;
-  }
-
-  export interface ClientRequestOptions extends Partial<RequestOptions> {
-    origin?: string | URL;
   }
 }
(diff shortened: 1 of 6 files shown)


ronag:
I've changed my mind. This makes things weird it's better if origin is required (even if we currently don't enforce it in runtime, should probably fix that at some point) to keep the API clean.

delvedor (commented):
Why? Given that you are already passing the origin to the Pool constructor it feels repetitive to pass it again here.

ronag:
> Why? Given that you are already passing the origin to the Pool constructor it feels repetitive to pass it again here.

Because it makes the API sensible. Right now it becomes just so much more complicated. It's only for `dispatch` which is not intended to be an ergonomic API anyway.

types/client.d.ts
```diff
@@ -13,11 +13,6 @@ declare class Client extends Dispatcher {
   closed: boolean;
   /** `true` after `client.destroyed()` has been called or `client.close()` has been called and the client shutdown has completed. */
   destroyed: boolean;
-  /** Dispatches a request. This API is expected to evolve through semver-major versions and is less stable than the preceding higher level APIs. It is primarily intended for library developers who implement higher level APIs on top of this. */
-  dispatch(options: Client.ClientDispatchOptions, handler: Dispatcher.DispatchHandlers): void;
-  /** Performs an HTTP request. */
-  request(options: Client.ClientRequestOptions): Promise<Dispatcher.ResponseData>;
-  request(options: Client.ClientRequestOptions, callback: (err: Error | null, data: Dispatcher.ResponseData) => void): void;
```
ronag, line 20: @delvedor: There is no such thing as `request` here.

mcollina (approved):
lgtm

delvedor:
@ronag not sure what you mean.

Furthermore, I would highly discourage to have the types not to reflect the js API. The types should be changed with the js code, otherwise it's very easy to diverge and have weird types.

mcollina (changes requested):
I agree with @delvedor, types should map the JS api as close as possible, and `origin` is not needed there.

ronag:
I would personally add the runtime check back then. I'm quite keen on keeping the API clean and simple.

delvedor:
I still think that if you can configure the origin in the Pool constructor, requiring the user to input **again** the origin in the `pool.request` does not simplify the API, on the contrary, it makes it more obscure. It causes me to think "Why I'm configuring this twice?" or "Can I put two different origins?".

ronag:
> I still think that if you can configure the origin in the Pool constructor, requiring the user to input **again** the origin in the `pool.request` does not simplify the API, on the contrary, it makes it more obscure. It causes me to think "Why I'm configuring this twice?" or "Can I put two different origins?".

So the idea is that we have the `Dispatcher.dispatch` interface. This is the only thing that you need to implement and contains everything you need to implement. You don't need to know about anything else. If we move out `origin` is becomes an implementation detail that you must consider on a case by case basis.

This is all theoretical. If no one else is on the page then I guess we just close this.

ronag:
To make my point the current type defs are still incorrect. You need to add them for pipeline, stream and any other high level API.

mcollina:
`Client` and `Pool` should require an `origin` in their constructor only. They are in essence origin-specific, while the `Agent` is not. If an `origin` is specified it should equal to the one passed to the constructor. I'm -1 in making this a runtime requirement for `Client` or `Pool`.

I agree the types should be changed significantly as they do not allow enough flexibility. However I think we are better off without this revert until that piece is done.

ronag:
>  If an origin is specified it should equal to the one passed to the constructor. 

Agreed.

> I'm -1 in making this a runtime requirement for Client or Pool.

Agreed here as well.

I'm only arguing that it should be required per docs and type defs to keep complexity of those down.

ronag:
Anyway feel free to close.

szmarczak:
> I'm only arguing that it should be required per docs and type defs to keep complexity of those down. Trying to add this "loose" mode there will just make things difficult with little gain IMO.

TypeScript is known to be difficult to be hard at times. Valid usage > messy types

szmarczak:
> If an origin is specified it should equal to the one passed to the constructor.

Client should also accept different origins because of https://github.com/nodejs/undici/issues/644

ronag:
Wouldn’t that mean that origin is a valid but optional type? Can’t we just make it optional in the dispatch api?

ronag:
@delvedor @mcollina PTAL https://github.com/nodejs/undici/pull/903/commits/bc14292d8475d79b2a41daca52d55b6d3ad80228 which I think I prefer.

delvedor (approved):
LGTM

PR #905, opened 2021-07-27
Introduce onSocket hook

delvedor:
Hello! This is a proposal and I'm open to change the API to make it as much ergonomic as possible.
Undici is meant to be a low-level HTTP client, we all agree about this. One of the issues that I'm facing while using it, is that in some cases you need to access "something more", in my case the currently used socket, which is "hidden" behind a symbol.

This pr introduces a "hook" to allow the user to perform some additional operation (see https://github.com/nodejs/undici/discussions/904) before to actually use the socket.

Feedback is welcome!

### DIFF
--- a/docs/api/Client.md
+++ b/docs/api/Client.md
@@ -26,6 +26,7 @@ Returns: `Client`
 * **pipelining** `number | null` (optional) - Default: `1` - The amount of concurrent requests to be sent over the single TCP/TLS connection according to [RFC7230](https://tools.ietf.org/html/rfc7230#section-6.3.2). Carefully consider your workload and environment before enabling concurrent requests as pipelining may reduce performance if used incorrectly. Pipelining is sensitive to network stack settings as well as head of line blocking caused by e.g. long running requests. Set to `0` to disable keep-alive connections.
 * **connect** `ConnectOptions | Function | null` (optional) - Default: `null`.
 * **strictContentLength** `Boolean` (optional) - Default: `true` - Whether to treat request content length mismatches as errors. If true, an error is thrown when the request content-length header doesn't match the length of the request body.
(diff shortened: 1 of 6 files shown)


mcollina (commented):
Can you add a test for the new type?

ronag:
The problem I see with this is that it adds api surface and constraints that undici must use sockets. Preferably it should just be a `Duplex` stream. Have you considered passing a custom `connect` factory method?

ronag:
I would be more open to a more strict API that is just for certificate validation or something along those lines, i.e. as little coupling to Node core stuff as possible.

delvedor:
> Have you considered passing a custom connect factory method in options?

Yes, the current problem with the `connect` function is that is rather big, and I fear that using your own will diverge too much somewhere in the future. Unless `connect` will become the de-facto method of creating a socket/transport mechanism and the API will remain stable between majors.

> I would be more open to a more strict API that is just for certificate validation or something along those lines, i.e. as little coupling to Node core stuff as possible.

The problem here is that the certificate validation depends a lot on what transport mechanism you are using. If the idea is to abstract away net/tls sockets, then the API will become too complex, and everything will fall on the `connect` function.

ronag:
This feels like such a nishe thing that I would prefer you use connect. I'd be careful about adding these kind of things to the API unless there is more demand for it.

ronag:
> Yes, the current problem with the connect function is that is rather big, and I fear that using your own will diverge too much somewhere in the future.

What about if we make the default connector public so you can easier re-use it?

ronag:
Another option is to add the functionality you want in https://github.com/nodejs/undici/discussions/904 to our connect. Not sure how generic thing it is. Would it be something everyone wants+

delvedor:
> What about if we make the default connector public so you can easier re-use it?


I was about to suggest the same, probably this is the easiest way to achieve this without changing too much the API. If it's ok with y'all I'll open a separate pr.

mcollina:
I'm +1 on making the default connect function public.

PR #906, opened 2021-07-27
Expose Connector in the public API

delvedor:
As titled. I've also added an example to show how to validate the CA fingerprint.

Related: https://github.com/nodejs/undici/discussions/904 #905

### DIFF
--- a/docs/api/Client.md
+++ b/docs/api/Client.md
@@ -48,6 +48,31 @@ import { Client } from 'undici'
 const client = new Client('http://localhost:3000')
 ```
 
+### Example - Custom connector
+
+This will allow you to perform some additional check on the socket that will be used for the next request.
+
+```js
+'use strict'
+import { Client, buildConnector } from 'undici'
+
+const connector = buildConnector({ rejectUnauthorized: false })
+const client = new Client('https://localhost:3000', {
+  connect (opts, cb) {
+    connector(opts, (err, socket) => {
+      if (err) {
+        cb(err)
+      } else if (/* assertion */) {
+        socket.destroy()
+        cb(new Error('kaboom'))
+      } else {
+        cb(null, socket)
+      }
+    })
+  }
+})
+```
+
 ## Instance Methods
 
 ### `Client.close([callback])`
--- a/examples/ca-fingerprint/index.js
+++ b/examples/ca-fingerprint/index.js
@@ -2,7 +2,7 @@
 
 const crypto = require('crypto')
 const https = require('https')
-const { Client, Connector } = require('../..')
+const { Client, buildConnector } = require('../..')
 const pem = require('https-pem')
 
 const caFingerprint = getFingerprint(pem.cert.toString()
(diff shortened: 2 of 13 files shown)


ronag (approved):

ronag (changes requested):
Actually isn't it better to just expose the factory method instead of the class? The class is just an implementation detail.

delvedor:
> Actually isn't it better to just expose the factory method instead of the class? The class is just an implementation detail.

I like exposing the class, most of the time the user will just need to initialize it, but if needed it can be extended quite easily.

ronag:
> I like exposing the class, most of the time the user will just need to initialize it, but if needed it can be extended quite easily.

But it being a class doesn't really do anything? Nothing is exposed. It's just a silly way to implement a closure.

ronag (approved):
I'd prefer a method but I guess a class is fine...

delvedor:
Hold on before merging, there might be an issue with the types, checking...

test/ca-fingerprint.js
```diff
@@ -0,0 +1,122 @@
...
+      connect (opts, cb) {
+        connector.connect(opts, (err, socket) => {
+          if (err) {
+            return cb(err)
+          }
+          if (getIssuerCertificate(socket).fingerprint256 !== 'FO:OB:AR') {
+            socket.destroy()
+            return cb(new Error('Fingerprint does not match'))
+          }
+          return cb(null, socket)
```
ronag, line 81:
```suggestion
          if (err) {
           cb(err)
          } else if (getIssuerCertificate(socket).fingerprint256 !== 'FO:OB:AR') {
            socket.destroy()
           cb(new Error('Fingerprint does not match'))
          } else {
            cb(null, socket)
          }
```

nit

mcollina (commented):
Is there a reason why we have a class in the first place? I would just remove it and expose a function.

ronag:
> Is there a reason why we have a class in the first place? I would just remove it and expose a function.

Nothing other than convenience of implementation.

delvedor:
Oks, this pr is good to go from my point of view, the types are now correct.

delvedor:
@ronag @mcollina I've changed the connector from class to function, let me know if it's what you had in mind.

mcollina (changes requested):
Awesome work! Could you add docs?

ronag (approved):

examples/ca-fingerprint/index.js
```diff
@@ -0,0 +1,76 @@
...
+    data.body.on('end', () => {
+      console.log(Buffer.concat(bufs).toString('utf8'))
+      client.close()
+      server.close()
+    })
+  })
+})
+
+function getIssuerCertificate (socket) {
+  let certificate = socket.getPeerCertificate(true)
```
ronag, line 56: Should probably check that socket is a `tls.Socket`.

index.js
```diff
@@ -20,6 +21,7 @@ module.exports.Client = Client
 module.exports.Pool = Pool
 module.exports.Agent = Agent
 
+module.exports.buildConnector = buildConnector
```
ronag, line 24: Now I see why you preferred a class 😄 .

mcollina (approved):
lgtm

mcollina:
Docs! ;)

delvedor:
Ready to go!

docs/api/Client.md
```diff
@@ -48,6 +48,31 @@ import { Client } from 'undici'
 const client = new Client('http://localhost:3000')
 ```
 
+### Example - Custom connector
+
+This will allow you to perform some additional check on the socket that will be used for the next request.
+
+```js
+'use strict'
+import { Client, buildConnector } from 'undici'
```
mcollina, line 57: we should document `buildConnect()` somewhere else as well.

delvedor: Done!

mcollina (approved):
lgtm

PR #908, opened 2021-07-30
fix: AbortError compat

ronag:
This makes undici abort errors compatible with check for `AbortError` i.e.

instead of:

```js
catch (err) {
  if (err.code === 'UND_ERR_ABORTED' || err.name === 'AbortError') {
  }
}
```

We can now simply do:

```js
catch (err) {
  if (err.name === 'AbortError') {
  }
}
```

### DIFF
--- a/lib/core/errors.js
+++ b/lib/core/errors.js
@@ -72,7 +72,7 @@ class RequestAbortedError extends UndiciError {
   constructor (message) {
     super(message)
     Error.captureStackTrace(this, RequestAbortedError)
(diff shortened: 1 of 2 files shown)


Ethan-Arrowood:
Is AbortError available from node core yet? Maybe we could align on that. But I'm +1 for this change cause I think Undici fetch is using name = AbortError as well

ronag:
> Is AbortError available from node core yet?

Not yet unfortunately.

mcollina (approved):
lgtm

PR #911, opened 2021-08-01
add `BodyMixin` type

### LINKED ISSUE #910: BodyMixin types are missing
## Bug Description

The typings for request.body were not updated with v4.3.x so TypeScript users can't use it easily yet.

Right now `response.body` is a Readable stream type only

## Reproducible By

file.ts:
```ts

const { body } = await request('https://www.google.com')
const html = await body.text();
                        ^ Property 'text' does not exist on type 'Readable'
```

### Additional context

https://github.com/nodejs/undici/blob/main/types/dispatcher.d.ts#L101
https://github.com/nodejs/undici/blob/main/types/dispatcher.d.ts#L110

these should be the only places it needs updating although I haven't looked for other instances.

```ts
// from here: https://fetch.spec.whatwg.org/#body-mixin
interface BodyMixin {
  readonly body?: never // throws on node v16.6.0
  readonly bodyUsed: boolean
  arrayBuffer(): Promise<ArrayBuffer>
  blob(): Promise<import('buffer').Blob> // requires nodejs v15.7.0+
  formData(): Promise<never> // not implemented
  json(): Promise<any>
  text(): Promise<string>
}

// ...
{
  body: Readable & BodyMixin
}
```

KhafraDev:
Fixes #910

### DIFF
--- a/types/dispatcher.d.ts
+++ b/types/dispatcher.d.ts
@@ -149,12 +149,12 @@ declare namespace Dispatcher {
    * @link https://fetch.spec.whatwg.org/#body-mixin
    */
   interface BodyMixin {
-    readonly body?: never // throws on node v16.6.0
-    readonly bodyUsed: boolean
-    arrayBuffer(): Promise<ArrayBuffer>
-    blob(): Promise<Blob>
-    formData(): Promise<never>
-    json(): Promise<any>
-    text(): Promise<string>
+    readonly body?: never; // throws on node v16.6.0
+    readonly bodyUsed: boolean;
+    arrayBuffer(): Promise<ArrayBuffer>;
+    blob(): Promise<Blob>;
+    formData(): Promise<never>;
+    json(): Promise<any>;
+    text(): Promise<string>;
   }
 }
--- a/types/dispatcher.d.ts
+++ b/types/dispatcher.d.ts
@@ -2,6 +2,7 @@ import { URL } from 'url'
 import { Duplex, Readable, Writable } from 'stream'
 import { EventEmitter } from 'events'
 import { IncomingHttpHeaders } from 'http'
+import { Blob } from 'buffer'
 
 type AbortSignal = unknown;
 
@@ -98,7 +99,7 @@ declare namespace Dispatcher {
   export interface ResponseData {
     statusCode: number;
     headers: IncomingHttpHeaders;
-    body: Readable;
+    body: Readable & BodyMixin;
     trailers: Record<string, string>;
     opaque: unknown;
     context: object;
@@ -143,4 +144,17 @@ declare namespace Dispatcher {
     onBodySent?(chunkSize: number, totalBytesSent: number): void;
   }
   export type PipelineHandler = (data: PipelineHandlerData) => Readable;
+
+  /** 
+   * @link https://fetch.spec.whatwg.org/#body-mixin
+   */
+  interface BodyMixin {
+    readonly body?: never // throws on node v16.6.0
+    readonly bodyUsed: boolean
+    arrayBuffer(): Promise<ArrayBuffer>
+    blob(): Promise<Blob>
+    formData(): Promise<never>
+    json(): Promise<any>
+    text(): Promise<string>
+  }
 }
--- a/test/types/dispatcher.test-d.ts
+++ b/test/types/dispatcher.test-d.ts
@@ -102,9 +102,9 @@ declare const { body }: Dispatcher.ResponseData;
 
 {
   // body mixin tests
-  expectType<never | undefined>(body.body);
-  expectType<boolean>(body.bodyUsed);
-  expectType<Promise<ArrayBuffer>>(body.arrayBuffer());
+  expectType<never | undefined>(body.body)
+  expectType<boolean>(body.bodyUsed)
+  expectType<Promise<ArrayBuffer>>(body.arrayBuffer())
   expectType<Promise<Blob>>(body.blob())
   expectType<Promise<never>>(body.formData())
   expectType<Promise<string>>(body.text())
--- a/test/types/dispatcher.test-d.ts
+++ b/test/types/dispatcher.test-d.ts
@@ -1,7 +1,8 @@
 import { Duplex, Readable, Writable } from 'stream'
-import { expectAssignable } from 'tsd'
+import { expectAssignable, expectType } from 'tsd'
 import { Dispatcher } from '../..'
 import { URL } from 'url'
+import { Blob } from 'buffer'
 
 expectAssignable<Dispatcher>(new Dispatcher())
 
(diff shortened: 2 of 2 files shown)


types/dispatcher.d.ts
```diff
@@ -143,4 +144,17 @@ declare namespace Dispatcher {
     onBodySent?(chunkSize: number, totalBytesSent: number): void;
   }
   export type PipelineHandler = (data: PipelineHandlerData) => Readable;
+
+  /** 
+   * @link https://fetch.spec.whatwg.org/#body-mixin
+   */
+  interface BodyMixin {
+    readonly body?: never // throws on node v16.6.0
```
ronag: What does it throw?

ronag (approved):

types/dispatcher.d.ts
KhafraDev:
```
      this[kBody] = Readable.toWeb(this)
                             ^

TypeError: Readable.toWeb is not a function
```
and @types/node are not updated to v16 yet so they don't include a web ReadableStream type sadly.

types/dispatcher.d.ts
```diff
@@ -143,4 +144,17 @@ declare namespace Dispatcher {
...
+  /** 
+   * @link https://fetch.spec.whatwg.org/#body-mixin
+   */
+  interface BodyMixin {
+    readonly body?: never; // throws on node v16.6.0
+    readonly bodyUsed: boolean;
+    arrayBuffer(): Promise<ArrayBuffer>;
+    blob(): Promise<Blob>;
+    formData(): Promise<never>;
+    json(): Promise<any>;
```
szmarczak, line 157: Use a generic template here.

KhafraDev: I don't think it should be to keep compatibility with the native types & there is a good discussion about using generics with body.json() [here](https://github.com/DefinitelyTyped/DefinitelyTyped/pull/36868#issuecomment-512989650).

types/dispatcher.d.ts
```diff
@@ -143,4 +144,17 @@ declare namespace Dispatcher {
...
+
+  /** 
+   * @link https://fetch.spec.whatwg.org/#body-mixin
+   */
+  interface BodyMixin {
+    readonly body?: never; // throws on node v16.6.0
+    readonly bodyUsed: boolean;
+    arrayBuffer(): Promise<ArrayBuffer>;
+    blob(): Promise<Blob>;
+    formData(): Promise<never>;
```
mcollina, line 156: what is this `never`?

szmarczak: Tells us that it prevents any further code from being executed ~ throws.

types/dispatcher.d.ts
szmarczak: I strongly disagree. People shouldn't compare any other language to TypeScript. TypeScript is too different. If they compare C# generics with TypeScript, then maybe generics shouldn't exist in TypeScript at all? Nope, generics were created with a purpose.

With `any` people will have to do `(await body.json()) as Result`.

With the generics people only do `await body.json<Result>()`. Looks much cleaner IMO.

KhafraDev: There's still the issue of differing from native types which makes a future fetch implementation (or extension to undici) differ from the web types:
```ts
const res = await body.json() as Result; // web way

const res = await body.json<Result>(); // undici way
```
I would be willing to change to unknown so it forces some type check, but I'm not sure if differing from the existing types in a way that make them incompatible is the best idea.

KhafraDev: Although the original comment had a lot of unnecessary comparisons, it still mentioned that generics here would decrease type safety. I think it's more appropriate if the user explicitly casts it, or does some type check to guarantee the data is what they're expecting.

https://github.com/microsoft/TypeScript/issues/44000#issuecomment-834763741

mcollina: I think aligning us with node-fetch and others is a better choice.


cc @Ethan-Arrowood

szmarczak: > Plus, it would break any compatibility with other fetch libs (node-fetch, undici-fetch)

How so? I don't see how it would break.

szmarczak: > it still mentioned that generics here would decrease type safety.

Then people need to use it safely. Problem solved. It's like using `malloc` is bad because you can easily forget to free the memory in C.

> I think it's more appropriate if the user explicitly casts it

I don't see how `(await body.json()) as Result` is better than `await body.json<Result>()`. In both cases people cast it manually, but the latter is via a template.

KhafraDev: > 
> 
> > Plus, it would break any compatibility with other fetch libs (node-fetch, undici-fetch)
> 
> How so? I don't see how it would break.

https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEYD2A7AzgF3gIycAngFzwBKIaADqmiANwCwAUEwJYoYgwBmUYCZaAK4QsAbybxJ8QTRjFMMNgHMmAXyZNk6LACs0qAAowkAWxY14AXhx58AOj2oAPAOEYAfAAoAlLSA

szmarczak: You are comparing `undici` to `fetch`. People come from `fetch`, not the other way around. [Compare `fetch` -> `undici`](https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEYD2A7AzgF3gIycAngFzwBKIaADqmggGTwDeAsAFDzvwBWaqxAPABUAfAAoAlPAC8Q+AAUYSALYBLGoKEBuVgF8tLVspQYQMAGZQwCMmgCuELMzYcbNGMUwxDAcx2tWIqDR8FDB4cSkZRw5EaixuVCl4KAB3KGUsXAIAOniUPms7DFExPWjkdCwacuAAKR4URIDU9Jw8fBz68QlA0nJCvW0xcQ0gA).

KhafraDev: I think the comparison is fair because BodyMixin is a [fetch API](https://fetch.spec.whatwg.org/#body-mixin). Plus, keeping compatibility with other modules seems like a net positive for everyone, especially since we will likely see an actual fetch implementation in undici (#912).

szmarczak: > Plus, keeping compatibility with other modules seems like a net positive for everyone

See my link above. The compatibility is kept.

KhafraDev: Compatibility in one direction is kept (fetch -> undici), but in no other circumstance. Rather, being able to switch between browser fetch, node-fetch, whatwg-fetch, etc etc without any changes would unequivocally benefit developers.

Plus reducing type safety (links above), for the sake of cleaner code, does not seem like a beneficial change. I could get behind using `unknown` to correct past mistakes, but generics have a real reason for not being added.

szmarczak: > Plus reducing type safety (links above), for the sake of cleaner code, does not seem like a beneficial change. 

Sorry, but I still don't know how it reduces safety. The generic template safety is the same as the `as` approach.

KhafraDev: Going off of the previous conversations I linked:

> That would just decrease type safety, as the user would be mistakenly lead to believe the type will be returned. This is the opposite of what TypeScript tries to add.

...
> > however personally I'm off the opinion that type casting and any/unknown should be avoided when possible and instead design the code in such a way that casting is not required

> This is a typecast—there’s no difference aside from syntax. I do think that the generic looks and feels better, but that actually makes it worse. The better appearance is entirely an illusion, and the illusion of type safety is worse than something that is transparently unsafe.
...
```js
function isMyResource(json: unknown): json is MyResource {
  // check the shape... return true/false
}

// ...
const json = await response.json(); // unknown
if (isMyResource(json)) {
  json; // MyResource
}
```
> This would be the canonical, safe way to type such a construct, at which point the generic would be egregiously wrong by comparison.

I think at this point we're both just iterating the same points, so I'll wait for others to make a decision before changing anything. 👍

szmarczak (approved):

mcollina (approved):
lgtm

PR #914, opened 2021-08-02
fix: only redirect if undisturbed body

ronag:


lib/core/util.js
```diff
@@ -224,8 +224,22 @@ function validateHandler (handler, method, upgrade) {
   }
 }
 
+function isDisturbed (body) {
```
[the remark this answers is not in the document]
ronag: Done

ronag:
@mcollina added tests and fixed bug

lib/handler/redirect.js
```diff
@@ -1,11 +1,29 @@
...
+
+class BodyAsyncIterable {
+  constructor (body) {
+    this[kBody] = body
+    this[kBodyUsed] = false
+  }
+
+  async * [Symbol.asyncIterator] () {
+    if (this[kBodyUsed]) {
+      throw new TypeError('disturbed')
```
delvedor, line 20: I would throw a more meaningful error, this one looks a little obscure.

ronag: I'll change it to an assertion. It shouldn't happen.

test/redirect-request.js
```diff
@@ -351,6 +353,27 @@ for (const factory of [
     }
   })
 
+  t.test('should not follow redirects when using ReadableStream request bodies', { skip: nodeMajor < 16 }, async t => {
```
mcollina, line 356: What will happen if node.js < 16?

mcollina (approved):
lgtm

test/redirect-request.js
ronag: ReadableStream is only on node 16?

PR #921, opened 2021-08-03
fix: fetch response is not Readable

ronag:
It's much easier to implement spec compliant fetch API if we don't try to be node streams at the same time.

### DIFF
--- a/lib/api/api-fetch.js
+++ b/lib/api/api-fetch.js
@@ -5,7 +5,6 @@
 const Headers = require('./headers')
 const { kHeadersList } = require('../core/symbols')
 const { Readable } = require('stream')
-const { ReadableStream, TransformStream } = require('stream/web')
 const { METHODS } = require('http')
 const {
   InvalidArgumentError,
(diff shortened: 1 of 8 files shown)


lib/api/api-fetch.js
```diff
@@ -2,9 +2,10 @@
 
 'use strict'
 
-const Body = require('./readable')
 const Headers = require('./headers')
 const { kHeadersList } = require('../core/symbols')
+const { Readable } = require('stream')
+const { ReadableStream, TransformStream } = require('stream/web')
```
mcollina, line 8: How can we run these on Node < 16? Should we disable fetch?

mcollina (changes requested):

lib/api/api-fetch.js
ronag: Yea, probably.

ronag: Done

mcollina (approved):
lgtm

PR #922, opened 2021-08-04
port Header class changes from core PR#38986

Ethan-Arrowood:
This PR ports some of the changes I've made in https://github.com/nodejs/node/pull/38986 

Includes an argument length check as well that isn't in my core PR yet

### DIFF
--- a/lib/api/headers.js
+++ b/lib/api/headers.js
@@ -5,18 +5,11 @@
 const { types } = require('util')
 const { validateHeaderName, validateHeaderValue } = require('http')
 const { kHeadersList } = require('../core/symbols')
-
-const headerRegex = /^[\n\t\r\x20]+|[\n\t\r\x20]+$/g
-
-function normalizeAndValidateHeaderName (name) {
-  const normalizedHeaderName = name.toLowerCase()
-  validateHeaderName(normalizedHeaderName)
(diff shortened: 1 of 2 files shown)


ronag (approved):

mcollina (approved):
lgtm but it makes me uncomfortable in keep landing changes without tests.

ronag:
@mcollina It's only fetch related changes that we land without tests. I think this is good in order to keep velocity up and it adds a natural place for new contributors to help us. If you are still uncomfortable then I guess we can slow down a bit and focus on tests. I was hoping to try and implement functionality and then focus on tests before release.

ronag:
@Ethan-Arrowood would be great if we could port over the tests from undici-fetch.

mcollina:
> @mcollina It's only fetch related changes that we land without tests. I think this is good in order to keep velocity up and it adds a natural place for new contributors to help us. If you are still uncomfortable then I guess we can slow down a bit and focus on tests. I was hoping to try and implement functionality and then focus on tests before release.

That'd be ok if it's a feature branch and not main. Anyway, let's see what the timeline is... we can always move this into a different branch.

ronag:
> That'd be ok if it's a feature branch and not main

Yes you are totally right. Apologies.

PR #927, opened 2021-08-04
Fix `blob.type` not having the response type

### LINKED ISSUE #925: Blob should have response content type
Body mixin in `undici.request` and `undici.fetch` when returning a `Blob` should also set `Blob.type` to the corresponding `content-type` header from the response.

Search for `new Body` and you should pass in a `type` in `new Blob(chunks, { type: headers['content-type'] })`.

Don't forget tests.

MSE99:
#### Description

Closes #925 

Sets the `type` property in `Blob` objects, to the value found in `content-type` response header.

### DIFF
--- a/lib/api/api-request.js
+++ b/lib/api/api-request.js
@@ -71,14 +71,15 @@ class RequestHandler extends AsyncResource {
       return
     }
 
-    const body = new Readable(resume, abort)
+    const parsedHeaders = util.parseHeaders(headers)
+    const body = new Readable(resume, abort, parsedHeaders['content-type'])
 
     this.callback = null
     this.res = body
 
     this.runInAsyncScope(callback, null, null, {
       statusCode,
-      headers: util.parseHeaders(headers),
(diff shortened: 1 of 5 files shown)


lib/api/api-request.js
```diff
@@ -71,14 +71,15 @@ class RequestHandler extends AsyncResource {
       return
     }
 
-    const body = new Readable(resume, abort)
+    const parsedHeaders = util.parseHeaders(headers)
```
ronag, line 74: This is slow. Can you do this lazily in the blob getter?

ronag (approved):
Nice!

ronag:
Tests are failing. That needs to be fixed.

test/client-request.js
```diff
@@ -228,6 +228,31 @@ test('request text', (t) => {
   })
 })
 
+test('request blob', (t) => {
```
ronag, line 231: 'should set type of blob object to the value of the `Content-Type` header from response'

PR #928, opened 2021-08-05
fix: implement fetch per spec

ronag:


ronag:
@mcollina what's the proper procedure/text in regards to licensing?

ronag:
@Ethan-Arrowood PTAL at the disabled tests for `Headers` class.

mcollina (commented):
The node-fetch copyright should be retained somewhere.

ronag:
> The node-fetch copyright should be retained somewhere.

https://github.com/nodejs/undici/pull/928#issuecomment-893210627

Any suggestion of how to best do that?

ronag:
How about https://github.com/nodejs/undici/pull/928/commits/8c198e1491d1c3c673a7cdbcfbce2db7a385eef4?

mcollina:
ok for me.

index.js
```diff
@@ -84,7 +84,25 @@ function makeDispatcher (fn) {
 module.exports.setGlobalDispatcher = setGlobalDispatcher
 module.exports.getGlobalDispatcher = getGlobalDispatcher
 
-module.exports.fetch = makeDispatcher(api.fetch)
+if (api.fetch) {
+  const _fetch = makeDispatcher(api.fetch)
+  module.exports.fetch = async function fetch (...args) {
+    try {
+      return await _fetch(...args)
```
szmarczak, line 91: `return await` is useless if `async function`, you can just do `return _fetch(...args)` :)

ronag: You forget the try/catch

test/node-fetch/response.js
```diff
@@ -0,0 +1,215 @@
...
+    //   'status',
+    //   'ok',
+    //   'redirected',
+    //   'statusText',
+    //   'headers'
+    // ]) {
+    //   expect(() => {
+    //     res[toCheck] = 'abc'
+    //   }).to.throw()
+    // }
```
ronag, line 65: Not sure how to fix this?

test/node-fetch/headers.js
```diff
@@ -0,0 +1,281 @@
...
+    expect({ key: 'accept', value: 'application/json, text/plain', object: headers }).to.deep.equal(results[0])
+    expect({ key: 'content-type', value: 'text/html', object: headers }).to.deep.equal(results[1])
+  })
+
+  // it('should set "this" to undefined by default on forEach', () => {
+  //   const headers = new Headers({ Accept: 'application/json' })
+  //   headers.forEach(function () {
+  //     expect(this).to.be.undefined
+  //   })
+  // })
```
ronag, line 77: @Ethan-Arrowood

index.js
szmarczak: Ah indeed, you're right

test/node-fetch/main.js
```diff
@@ -0,0 +1,2310 @@
...
+
+  // it('should support redirect mode, error flag', () => {
+  //   const url = `${base}redirect/301`
+  //   const options = {
+  //     redirect: 'error'
+  //   }
+  //   return expect(fetch(url, options)).to.eventually.be.rejected
+  //     .and.be.an.instanceOf(FetchError)
+  //     .and.have.property('type', 'no-redirect')
+  // })
```
ronag, line 484: @szmarczak: Do you want to take a look at the redirect failures?

test/node-fetch/main.js
```diff
@@ -0,0 +1,2310 @@
...
+  //     'Invalid-Header-Value',
+  //     '\u0007k\r\n',
+  //     'Cookie',
+  //     '\u0007k\r\n',
+  //     'Cookie',
+  //     '\u0007kk\r\n'
+  //   ])
+  //   expect(headers).to.be.instanceOf(Headers)
+  //   expect(headers.raw()).to.deep.equal({})
+  // })
```
ronag, line 572: @Ethan-Arrowood

test/node-fetch/main.js
szmarczak: sure

test/node-fetch/main.js
```diff
@@ -0,0 +1,2310 @@
...
+
+  // it('should reject if response compression is invalid', () => {
+  //   const url = `${base}invalid-content-encoding`
+  //   return fetch(url).then(res => {
+  //     expect(res.headers.get('content-type')).to.equal('text/plain')
+  //     return expect(res.text()).to.eventually.be.rejected
+  //       .and.be.an.instanceOf(FetchError)
+  //       .and.have.property('code', 'Z_DATA_ERROR')
+  //   })
+  // })
```
ronag, line 884: Are these compression/descompression stuff part of spec or a node-fetch extension?

test/node-fetch/main.js
szmarczak: `FetchError` doesn't exist. https://fetch.spec.whatwg.org/#concept-request-redirect-mode

> "error"
> Return a network error when a request is met with a redirect.

> A response whose type is "error" is known as a network error.
> 
> A network error is a response whose status is always 0, status message is always the empty byte sequence, header list is always empty, and body is always null.

I'll look how it's done in browsers.

ronag: You can just comment out `FetchError` or replace it with `Error`. Please try all the redirect tests.

test/node-fetch/response.js
ronag: @mcollina?

test/node-fetch/main.js
szmarczak: Hmm... Google Chrome throws `TypeError: Failed to fetch`. Firefox:

```
TypeError: NetworkError when attempting to fetch resource.
```

szmarczak: I think we could just create this empty response and reject the promise with it

test/node-fetch/main.js
szmarczak: I took a quick look and couldn't find anything in this regards in the spec. Looks like Google Chrome supports compression/decompression:

```
accept-encoding: gzip, deflate, br
```

szmarczak: Firefox as well

ronag: I guess we should open an issue in regards to whether or not we should support this if it's not in spec. Would you mind doing that and write your findings?

test/node-fetch/main.js
szmarczak: > I think we could just create this empty response and reject the promise with it

Ah I see it's done already, let me fix the tests then :)

szmarczak: I'm having a hard time communicating with `chai`:

```js
  it.only('should support redirect mode, error flag', async () => {
    const url = `${base}redirect/301`
    const options = {
      redirect: 'error'
    }

    return expect(fetch(url, options)).to.be.fulfilled
      .and.be.instanceOf(Response).and.have.property('status', 0)
  })
```

should pass but gives

```js
AssertionError: expected {} to be an instance of Respons
```

btw

```js
console.log((await fetch(url, options)) instanceof Response) //=> true
```

lib/api/api-fetch/headers.js
```diff
@@ -49,21 +53,21 @@ function isHeaders (object) {
 function fill (headers, object) {
   if (isHeaders(object)) {
     // Object is instance of Headers
-    headers[kHeadersList] = Array.splice(object[kHeadersList])
+    headers[kHeadersList] = [...object[kHeadersList]]
```
ronag, line 56: @Ethan-Arrowood

ronag:
Applied the file restructure/refactor to main and rebased to make it more clear what changes this PR does.

lib/api/api-fetch/headers.js
```diff
@@ -233,8 +245,31 @@ class Headers {
       callback.call(thisArg, this[kHeadersList][index + 1], this[kHeadersList][index], this)
     }
   }
+
+  [Symbol.for('nodejs.util.inspect.custom')] () {
+    return Object.fromEntries(this.entries())
+  }
```
ronag, line 251: This is slightly different than node-fetch. What do we prefer?

lib/api/api-fetch/headers.js
```diff
@@ -94,12 +98,20 @@ class Headers {
   constructor (init = {}) {
     // validateObject allowArray = true
     if (!Array.isArray(init) && typeof init !== 'object') {
-      throw new InvalidArgumentError('The argument \'init\' must be one of type Object or Array')
+      throw new TypeError('The argument \'init\' must be one of type Object or Array')
```
ronag, line 101: @mcollina: In regards to errors? Should we do exactly what the spec says and throw generic `TypeError`s or do we want to enrich them?

lib/api/api-fetch/headers.js
ronag: @Ethan-Arrowood

ronag:
CI green ✅

lib/api/api-fetch/headers.js
ronag: It actually failed the test with `Array.splice` :/

Ethan-Arrowood: Ohhh this is an error in my code. It should be `object[kHeadersList].splice(0)`

Ethan-Arrowood: (but also spread operator is fine too)

lib/api/api-fetch/headers.js
Ethan-Arrowood: Yeah I missed this when I ported over my Node.js PR changes. In Node core we were just going to return the underlying `this[kHeadersList]`.

lib/api/api-fetch/headers.js
Ethan-Arrowood: https://github.com/nodejs/node/pull/38986/files?file-filters%5B%5D=.js&file-filters%5B%5D=.mjs#diff-3569a54af4e25669eee4cc77bd0f4e86ca5fafb0cf766759a4bd2fb482fa0984R95

lib/api/api-fetch/headers.js
Ethan-Arrowood: https://github.com/nodejs/node/pull/38986/files?file-filters%5B%5D=.js&file-filters%5B%5D=.mjs#diff-3569a54af4e25669eee4cc77bd0f4e86ca5fafb0cf766759a4bd2fb482fa0984R266-R268

lib/api/api-fetch/headers.js
Ethan-Arrowood: These changes came from Node core PR: https://github.com/nodejs/node/pull/38986/files?file-filters%5B%5D=.js&file-filters%5B%5D=.mjs#diff-3569a54af4e25669eee4cc77bd0f4e86ca5fafb0cf766759a4bd2fb482fa0984R138

I would lean towards returning TypeErrors directly as that would be better for interop

test/node-fetch/main.js
szmarczak: Ok this works (`chai` requires `.eventually` for promises):

```js
  it('should support redirect mode, error flag', async () => {
    const url = `${base}redirect/301`
    const options = {
      redirect: 'error'
    }

    return expect(fetch(url, options)).to.be.fulfilled
      .and.eventually.be.instanceOf(Response).and.have.property('status', 0)
  })
```

szmarczak:
In terms of https://github.com/nodejs/undici/blob/2452331daadafb346c48bc67b4f464726d5a7491/test/node-fetch/main.js#L466

`undici` fails because it considers `☃` an invalid header value character, which **is** valid.

[RFC2616](https://datatracker.ietf.org/doc/html/rfc2616#section-4.2):
```
       OCTET          = <any 8-bit sequence of data>
       CHAR           = <any US-ASCII character (octets 0 - 127)>
       CTL            = <any US-ASCII control character
                        (octets 0 - 31) and DEL (127)>
       CR             = <US-ASCII CR, carriage return (13)>
       LF             = <US-ASCII LF, linefeed (10)>
       SP             = <US-ASCII SP, space (32)>
       HT             = <US-ASCII HT, horizontal-tab (9)>
       CRLF           = CR LF
       LWS            = [CRLF] 1*( SP | HT )

       message-header = field-name ":" [ field-value ]
       field-name     = token
       field-value    = *( field-content | LWS )
       field-content  = <the OCTETs making up the field-value
                        and consisting of either *TEXT or combinations
                        of token, separators, and quoted-string>
```

The value can be any 8-bit data.

ronag:
> should pass but it throws.

Sounds like a node bug then. Should we report it there?

test/node-fetch/main.js
ronag: This changes the node fetch test though? It thinks we should reject promise here. Is that a spec violation?

ronag:
@Ethan-Arrowood If you looked a `Response` and `Request` you will see that I've tried to annotate the code with the steps specified in the spec. Do you think that would be possible and a good idea to do similarly in the `Headers` implementation?

ronag:
@mcollina I think this is as ready as I can make it at the moment. Can we merge and then I can add remaining issues to the fetch issue checklist?

- redirect support
- request/clone doesn't quite work for some reason. `ReadableStream.tee()` issue?
- Should `Request` take `URL`?  Where does the spec specify this is valid?
- abort signal support

test/node-fetch/main.js
ronag: I don't think redirects are working properly at all atm. Hopefully someone can take a look at that.

test/node-fetch/external-encoding.js
```diff
@@ -0,0 +1,43 @@
+// const chai = require('chai')
```
mcollina, line 1: Remove this file?

mcollina (changes requested):
There are plenty of tests that are commented out, let's skip them instead so they are reported during the run (and maybe track the fact that we want all of them to pass).

test/node-fetch/external-encoding.js
ronag: Why? We want to support this eventually.

ronag:
@mcollina: uncommented as many tests as possible and made mocha skip them.

Ethan-Arrowood:
> a good idea to do similarly in the Headers implementation?

yes great idea, ill open an issue to track

test/node-fetch/main.js
szmarczak: It does violate the spec. https://fetch.spec.whatwg.org/#fetch-method

> If response is a network error, then reject p with a TypeError and terminate these substeps.

ronag:
@mcollina: I think I've removed most uncommented tests now. PTAL

lib/api/api-fetch/body.js
```diff
@@ -1,66 +1,183 @@
...
+  const [out1, out2] = src.stream.tee()
+  src.stream = out1
+  return {
+    stream: out2,
+    length: src.length,
+    source: src.source
+  }
+}
+
+class Body {
```
Ethan-Arrowood, line 105: Not required for this PR, but Body isn't really a class per the spec. Its just a mixin and adds some things to the Request and Response class. This is slightly important to how those classes are implemented because if `Request` has to `extend Body` in order to get access to these methods and properties, then it must also call `super()` in the constructor and that breaks the specified `Request` and `Response` initialization steps. That is the main reason why i have Body implemented  as a function in undici-fetch (https://github.com/Ethan-Arrowood/undici-fetch/blob/f400f890d19034d00562763f472c0d2f5824bcb6/src/body.js#L27)

ronag: Yea I didn't want it as a class but didn't manage to get a mixin working.

lib/api/api-fetch/constants.js
```diff
@@ -0,0 +1,119 @@
...
+  'Set-Cookie',
+  'Set-Cookie2'
+].map(x => x.toLowerCase())
+
+const requestBodyHeader = [
+  'Content-Encoding',
+  'Content-Language',
+  'Content-Location',
+  'Content-Type'
+].map(x => x.toLowerCase())
```
Ethan-Arrowood, line 104:
```suggestion
const requestBodyHeader = [
  'content-encoding',
  'content-language',
  'content-location',
  'content-type'
]
```

lib/api/api-fetch/body.js
ronag: Make an issue and feel free to open a PR once this has landed!

Ethan-Arrowood (changes requested):
This PR is no longer just porting node-fetch tests. It seems like you're completely implementing the fetch spec as well. Should I continue reviewing or wait?

ronag:
> This PR is no longer just porting node-fetch tests. It seems like you're completely implementing the fetch spec as well. Should I continue reviewing or wait?

You can review. I'm done now.

lib/api/api-fetch/LICENSE
```diff
@@ -0,0 +1,21 @@
...
+The above copyright notice and this permission notice shall be included in all
+copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE
+SOFTWARE.
```
ronag, line 21: @Ethan-Arrowood I added undici-fetch license here

Ethan-Arrowood:
> > This PR is no longer just porting node-fetch tests. It seems like you're completely implementing the fetch spec as well. Should I continue reviewing or wait?
>
> You can review. I'm done now.

Okay, I'll give this a look over later today when I can focus on it. Have to start normal-work now unfortunately

lib/api/api-fetch/body.js
ronag: I fixed it based on your link. Thanks!

lib/api/api-fetch/body.js
```diff
@@ -1,66 +1,178 @@
...
-      source: body.toString()
-    }, 'application/x-www-form-urlencoded;charset=UTF-8']
-  } else if (typeof body === 'string') {
-    return [{
-      source: body
-    }, 'text/plain;charset=UTF-8']
+    source = body.toString()
+    length = Buffer.byteLength(source)
+    type = 'application/x-www-form-urlencoded;charset=UTF-8'
   } else if (body instanceof ArrayBuffer || ArrayBuffer.isView(body)) {
```
Ethan-Arrowood, line 30: Not sure if this block matches spec anymore. For BufferSource:

> Set source to a copy of the bytes held by object.

Thats it unfortunately 🤷‍♂️

lib/api/api-fetch/body.js
```diff
@@ -1,66 +1,178 @@
...
-    return [{
-      source: body
-    }, null]
+    if (body instanceof DataView) {
+      // TODO: Blob doesn't seem to work with DataView?
+      body = body.buffer
+    }
+    source = body
+    length = Buffer.byteLength(body)
   } else if (util.isBlob(body)) {
```
Ethan-Arrowood, line 37: We are only supporting fetch in 16+ right?

lib/api/api-fetch/body.js
```diff
@@ -1,66 +1,178 @@
...
-    return [{
-      source: body,
-      length: body.size
-    }, body.type || null]
-  } else if (util.isStream(body) || typeof body.pipeThrough === 'function') {
+    source = body
```
Ethan-Arrowood, line 38: I believe we have to "read" the Blob as well 

> Set action to this step: read object.

lib/api/api-fetch/body.js
```diff
@@ -1,66 +1,178 @@
...
-    let stream
-    if (util.isStream(body)) {
-      stream = Readable.toWeb(body)
-    } else {
-      if (body.locked) {
-        throw new TypeError('locked')
+    if (body.locked) {
+      throw new TypeError('locked')
+    }
+
```
Ethan-Arrowood, line 49: missing keepalive check 

> If keepalive is true, then throw a TypeError.

lib/api/api-fetch/body.js
```diff
@@ -1,66 +1,178 @@
...
+    get () {
+      return this[kState].body ? this[kState].body.stream : null
+    }
+  },
+  bodyUsed: {
+    enumerable: true,
+    get () {
+      return this[kState].body && (
+        util.isDisturbed(this[kState].body.stream) ||
+        !!this[kState].bodyUsed
```
Ethan-Arrowood, line 162: why is part necessary?

lib/api/api-fetch/headers.js
```diff
@@ -90,108 +100,193 @@ function validateArgumentLength (found, expected) {
...
+
+class Headers {
+  constructor (init = {}) {
+    // validateObject allowArray = true
+    if (!Array.isArray(init) && typeof init !== 'object') {
+      throw new TypeError('The argument \'init\' must be one of type Object or Array')
+    }
+    this[kHeadersList] = new HeadersList()
+    this[kGuard] = 'none'
+    fill(this, init)
```
Ethan-Arrowood, line 187: im not sure where I saw you ask this question but if im not mistaken you asked about why do we have the "fill" function, and it is because in the Request and Response classes when a Headers instance needs to be cloned, the algorithms do this weird thing where it will initialize an empty Headers instance and then run `fill` on it with the existing instance HeaderList. Might be avoidable if we write the Request/Response in a certain way

lib/api/api-fetch/request.js
```diff
@@ -0,0 +1,652 @@
...
+      }
+    }
+
+    // 37. Let inputOrInitBody be initBody if it is non-null; otherwise
+    // inputBody.
+    const inputOrInitBody = initBody ?? inputBody
+
+    // 38. If inputOrInitBody is non-null and inputOrInitBody’s source is
+    // null, then:
+    if (inputOrInitBody != null && inputOrInitBody.source == null) {
```
Ethan-Arrowood, line 377:
```suggestion
    if (inputOrInitBody !== null && inputOrInitBody.source === null) {
```
strict null checks

lib/api/api-fetch/request.js
```diff
@@ -0,0 +1,652 @@
...
+
+      // 2. Set this’s request’s use-CORS-preflight flag.
+      request.useCORSPreflightFlag = true
+    }
+
+    // 39. Let finalBody be inputOrInitBody.
+    let finalBody = inputOrInitBody
+
+    // 40. If initBody is null and inputBody is non-null, then:
+    if (initBody == null && inputBody != null) {
```
Ethan-Arrowood, line 392:
```suggestion
    if (initBody === null && inputBody !== null) {
```
strict null checks

lib/api/api-fetch/request.js
```diff
@@ -0,0 +1,652 @@
...
+      (request.method === 'GET' || request.method === 'HEAD')
+    ) {
+      throw new TypeError()
+    }
+
+    // 35. Let initBody be null.
+    let initBody = null
+
+    // 36. If init["body"] exists and is non-null, then:
+    if ('body' in init && init != null) {
```
Ethan-Arrowood, line 356:
```suggestion
    if ('body' in init && init !== null) {
```
strict null check

lib/api/api-fetch/request.js
```diff
@@ -0,0 +1,652 @@
...
+
+    // 33. Let inputBody be input’s request’s body if input is a Request
+    // object; otherwise null.
+    const inputBody = input instanceof Request ? input[kState].body : null
+
+    // 34. If either init["body"] exists and is non-null or inputBody is
+    // non-null, and request’s method is `GET` or `HEAD`, then throw a
+    // TypeError.
+    if (
+      (('body' in init && init != null) || inputBody != null) &&
```
Ethan-Arrowood, line 346:
```suggestion
      (('body' in init && init !== null) || inputBody !== null) &&
```
strict null checks

lib/api/api-fetch/request.js
```diff
@@ -0,0 +1,652 @@
...
+    this[kState] = request
+
+    // 28. Set this’s signal to a new AbortSignal object with this’s relevant
+    // Realm.
+    // TODO: relevant Realm?
+    const ac = new AbortController()
+    this[kSignal] = ac.signal
+
+    // 29. If signal is not null, then make this’s signal follow signal.
+    if (signal != null) {
```
Ethan-Arrowood, line 280:
```suggestion
    if (signal !== null) {
```

lib/api/api-fetch/request.js
```diff
@@ -0,0 +1,652 @@
...
+      mode = fallbackMode
+    }
+
+    // 17. If mode is "navigate", then throw a TypeError.
+    if (mode === 'navigate') {
+      throw new TypeError()
+    }
+
+    // 18. If mode is non-null, set request’s mode to mode.
+    if (mode != null) {
```
Ethan-Arrowood, line 192:
```suggestion
    if (mode !== null) {
```

lib/api/api-fetch/response.js
```diff
@@ -1,141 +1,348 @@
...
+      // TODO
+    }
+
+    // 3. Set this’s response to a new response.
+    this[kState] = makeResponse(init)
+
+    // 4. Set this’s headers to a new Headers object with this’s relevant
+    // Realm, whose header list is this’s response’s header list and guard
+    // is "response".
+    // TODO: relevant Realm?
```
Ethan-Arrowood, line 103: Realm might be browser environment only 🤷‍♂️

Ethan-Arrowood (changes requested):
I've reviewed a good chunk of what you've written so far. Basically everything except from lib/api/api-fetch/index.js and the tests. 

I'm not going to pick too much at this since I'd like to get it merged asap and then we can iterate. Great work so far and I'm impressed you were able to get this all in so quickly!

lib/api/api-fetch/body.js
ronag: IsDistrubed doesn’t work yet. Haven’t been released on 16.x yet

lib/api/api-fetch/headers.js
ronag: I think I already solved this. PTAL at the response/request classes.

lib/api/api-fetch/response.js
ronag: I think realm is going to be part of es spec. So we might get it in node as well.

ronag:
@Ethan-Arrowood Fixed your comments. PTAL at extractBody

lib/api/api-fetch/body.js
ronag: Correct,

Ethan-Arrowood (approved):
awesome. great work again!

szmarczak:
I cannot review this right now as I'm working. Feel free to skip me, I'll review this after work (after 8h) :)

ronag:
> I cannot review this right now as I'm working. Feel free to skip me, I'll review this after work (after 8h) :)

We can wait for you 😄

ronag:
@jasnell Don't know if you are maybe interested in reviewing?

jasnell:
> Don't know if you are maybe interested in reviewing?

Yes, but give me a day or two. I've been neck deep in orientation and new dev environment setup this week.

ronag:
@jasnell If you have opportunity maybe you could take a look at `should support clone() method` tests. It seems to be something with webstream tee:ing that causes it to fail :/.

mcollina (approved):
lgtm

lib/api/api-fetch/body.js
ronag: Fixed

lib/api/api-fetch/body.js
```diff
@@ -1,66 +1,270 @@
...
+}
+
+function mixinBody (prototype) {
+  Object.assign(prototype, methods)
+  Object.defineProperties(prototype, properties)
+}
+
+function isBodyReadable (body) {
+  // This is a hack!
+  return body && /state: 'readable'/.test(nodeUtil.inspect(body.stream))
```
Mesteery, line 261:
```suggestion
  return body && nodeUtil.inspect(body.stream).includes("state: 'readable'")
```

lib/api/api-fetch/response.js
```diff
@@ -1,141 +1,367 @@
...
+    responseObject[kHeaders][kGuard] = 'immutable'
+    return responseObject
+  }
+
+  // Creates a redirect Response that redirects to url with status status.
+  static redirect (url, status = 302) {
+    // 1. Let parsedURL be the result of parsing url with current settings
+    // object’s API base URL.
+    // 2. If parsedURL is failure, then throw a TypeError.
+    // TODO: base-URL?
```
szmarczak, line 37: Doesn't apply to us. Node.js doesn't run in a specific context where `baseURL` is applied. Unless this will be a feature?

lib/api/api-fetch/response.js
```diff
@@ -1,141 +1,367 @@
...
+      parsedURL = new URL(url)
+    } catch (err) {
+      const error = new TypeError()
+      error.cause = err
+      throw error
+    }
+
+    // 3. If status is not a redirect status, then throw a RangeError.
+    if (!redirectStatus.includes(status)) {
+      throw new RangeError()
```
szmarczak, line 49:
```js
Response.redirect('', 200)
```

Chrome

```
RangeError: Failed to execute 'redirect' on 'Response': Invalid status code
```

Firefox

```
RangeError: Response.redirect: Invalid redirect status code.
```

lib/api/api-fetch/response.js
```diff
@@ -1,141 +1,367 @@
...
+    return responseObject
+  }
+
+  // https://fetch.spec.whatwg.org/#dom-response
+  constructor (body = null, init = {}) {
+    // 1. If init["status"] is not in the range 200 to 599, inclusive, then
+    // throw a RangeError.
+    if ('status' in init) {
+      if (!Number.isFinite(init.status)) {
+        throw new TypeError()
```
szmarczak, line 78:
```js
new Response(null, { status: NaN })
```

Chrome

(yes, Chrome sees `NaN` as `0`)

```
RangeError: Failed to construct 'Response': The status provided (0) is outside the range [200, 599].
```

Firefox

```
RangeError: Response constructor: Invalid response status code.
```

lib/api/api-fetch/response.js
```diff
@@ -1,141 +1,367 @@
...
+      }
+
+      if (init.status < 200 || init.status > 599) {
+        throw new RangeError()
```
szmarczak, line 82:
```js
new Response(null, { status: 0 })
```

Chrome

```
RangeError: Failed to construct 'Response': The status provided (0) is outside the range [200, 599].
```

Firefox

```
RangeError: Response constructor: Invalid response status code.
```

lib/api/api-fetch/response.js
```diff
@@ -1,141 +1,367 @@
...
+      }
+    }
+
+    if ('statusText' in init) {
+      if (typeof init.statusText !== 'string') {
+        throw new TypeError()
```
szmarczak, line 88: lol this is actually the one where browsers fail:

```js
new Response(null, { statusText: 0 })
```

both Chrome and Firefox pass. A more descriptive error message would be nice 👍🏼

lib/api/api-fetch/response.js
```diff
@@ -1,141 +1,367 @@
...
+    // 7. If init["headers"] exists, then fill this’s headers with init["headers"].
+    if ('headers' in init) {
+      fill(this[kState].headersList, init.headers)
+    }
+
+    // 8. If body is non-null, then:
+    if (body !== null) {
+      // 1. If init["status"] is a null body status, then throw a TypeError.
+      if (nullBodyStatus.includes(init.status)) {
+        throw new TypeError()
```
szmarczak, line 128: Please add a more descriptive message :)

lib/api/api-fetch/response.js
```diff
@@ -1,141 +1,367 @@
...
     }
   }
 
+  get [Symbol.toStringTag] () {
+    return this.constructor.name
+  }
+
+  toString () {
+    return Object.prototype.toString.call(this)
+  }
```
szmarczak, line 152: Is this necessary as well?

lib/api/api-fetch/response.js
```diff
@@ -1,141 +1,367 @@
...
+    // given a new response, "immutable", and this’s relevant Realm.
+    // TODO: relevant Realm?
+    const responseObject = new Response()
+    responseObject[kHeaders][kGuard] = 'immutable'
+
+    // 5. Set responseObject’s response’s status to status.
+    responseObject[kState].status = status
+
+    // 6. Let value be parsedURL, serialized and isomorphic encoded.
+    // TODO: isomorphic encoded?
```
szmarczak, line 62: > To isomorphic encode a string input, run these steps:
> 
> Assert: input contains no code points greater than U+00FF.
> 
> Return a byte sequence whose length is equal to input’s code point length and whose bytes have the same values as the values of input’s code points, in the same order.

Is this basically ASCII?

szmarczak (commented):
This is an awesome work 🙌🏼 Greatly appreciated 💛 It's 2.30am here and I still got a few files left. Will do later today. Keep it up!

lib/api/api-fetch/body.js
```diff
@@ -1,66 +1,270 @@
...
+    // Set source to the result of running the application/x-www-form-urlencoded serializer with object’s list.
+    source = object.toString()
+
+    // Set Content-Type to `application/x-www-form-urlencoded;charset=UTF-8`.
+    contentType = 'application/x-www-form-urlencoded;charset=UTF-8'
+  } else if (object instanceof ArrayBuffer || ArrayBuffer.isView(object)) {
+    // BufferSource
+
+    if (object instanceof DataView) {
+      // TODO: Blob doesn't seem to work with DataView?
```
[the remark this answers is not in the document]
ronag: Blob should be able to take DataView as a BlobPart but that doesn't work @jasnell

lib/api/api-fetch/body.js
```diff
@@ -1,66 +1,270 @@
...
+    // TODO: byte sequence?
+    // TODO: FormData?
+    // TODO: else?
+    source = String(object)
+    contentType = 'text/plain;charset=UTF-8'
+  }
+
+  // 7. If source is a byte sequence, then set action to a
+  // step that returns source and length to source’s length.
+  // TODO: What is a "byte sequence?"
```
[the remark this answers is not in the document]
ronag: I guess we just do `String(object)` on that then.

lib/api/api-fetch/body.js
```diff
@@ -1,66 +1,270 @@
...
+    }
+  } else if (typeof object === 'string') {
+    // scalar value string
+    // TODO: How to check for "scalar value string"?
+    source = object
+    contentType = 'text/plain;charset=UTF-8'
+  } else {
+    // TODO: byte sequence?
+    // TODO: FormData?
+    // TODO: else?
```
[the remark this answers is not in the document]
ronag: Yes, unfortunately...https://github.com/chromium/chromium/blob/94.0.4604.1/third_party/blink/renderer/core/fetch/response.cc#L222-L230

lib/api/api-fetch/body.js
```diff
@@ -1,66 +1,270 @@
...
+  if (action !== null) {
+    // Run action.
+    action(bytes => {
+      // Whenever one or more bytes are available and stream is not errored,
+      // enqueue a Uint8Array wrapping an ArrayBuffer containing the available
+      // bytes into stream.
+      // TODO: stream is not errored?
+      controller.enqueue(new Uint8Array(bytes))
+    }, err => {
+      // TODO: Spec doesn't say anything about this?
```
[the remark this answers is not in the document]
ronag: When the blob failes... I'm not sure if that is possible or not? @jasnell

lib/api/api-fetch/body.js
```diff
@@ -1,66 +1,270 @@
...
+  }
+
+  // 8. If action is non-null, then run these steps in in parallel:
+  if (action !== null) {
+    // Run action.
+    action(bytes => {
+      // Whenever one or more bytes are available and stream is not errored,
+      // enqueue a Uint8Array wrapping an ArrayBuffer containing the available
+      // bytes into stream.
+      // TODO: stream is not errored?
```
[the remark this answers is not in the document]
ronag: Fixed

lib/api/api-fetch/response.js
ronag: Fixed

lib/api/api-fetch/response.js
ronag: Suggestion?

lib/api/api-fetch/response.js
ronag: Don't know. Grabbed it from node-fetch. wdyt?

lib/api/api-fetch/response.js
ronag: What do they do? Just stringify? Maybe that is actually more correct.

lib/api/api-fetch/response.js
ronag: https://github.com/nodejs/undici/issues/942

lib/api/api-fetch/response.js
mcollina: I would remove from now

lib/api/api-fetch/util.js
```diff
@@ -181,7 +114,74 @@ function isValidHTTPToken (characters) {
   return true
 }
 
+function toWebReadable (streamReadable) {
```
mcollina, line 117: we should really have this in core.

ronag: It is in core, just not released as far as I can tell.

ronag: It's either:

`newStreamReadableFromReadableStream` or `newReadableStreamFromStreamReadable`.

I never remember which on does what.

.github/workflows/nodejs.yml
```diff
@@ -20,25 +20,34 @@ jobs:
...
-    - name: Unit test (coverage)
+    - name: Test Jest
       run: |
-        npm run coverage:ci
+        npm run test:jest
+
+    # How to run only for Node 16+?
+    # - name: Test node-fetch
+    #   run: |
+    #     npm run test:node-fetch
```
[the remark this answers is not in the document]
mcollina: I'm not understanding why it's commented out. Are we running those tests at all in CI?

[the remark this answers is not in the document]
ronag: We are not.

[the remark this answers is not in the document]
szmarczak: Maybe it's better off as a `.js` script? `node scripts/test-node-fetch.js`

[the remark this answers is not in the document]
ronag: That's a good idea! Do you have a sample of how that could look?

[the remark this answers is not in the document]
szmarczak: #### scripts/node.js

```js
const [major, minor, patch] = process.versions.node.split('.').map(v => Number(v));
const required = process.argv.pop().split('.').map(v => Number(v));

const badMajor = major < required[0];
const badMinor = major === required[0] && minor < required[1];
const badPatch = major === required[0] && minor < required[1] && patch < required[2];

if (badMajor || badMinor || badPatch) {
    console.log(`Required Node.js >=${required.join('.')}, got ${process.versions.node}`);
    process.exit(1);
}
```

#### package.json

```
"test:node-fetch": "node scripts/node.js 16 && mocha test/node-fetch || echo Skipping",
```

scripts/test-node-fetch.js
```diff
@@ -0,0 +1,11 @@
+const [major, minor, patch] = process.versions.node.split('.').map(v => Number(v))
+const required = process.argv.pop().split('.').map(v => Number(v))
+
+const badMajor = major < required[0]
+const badMinor = major === required[0] && minor < required[1]
+const badPatch = major === required[0] && minor < required[1] && patch < required[2]
```
szmarczak, line 6: I did a little mistake here by accident, it should be `===` in the middle condition 😅

GrosSacASac:
Does it mean `fetch` will be built in to nodejs ?

szmarczak:
Just finished reviewing this. This is *really good* piece of art :)

maxloh:
> Not yet. nodejs/undici is still a separate module independent of node.js core. There is an active discussion happening about whether it will be vendored in but that decision has not been made yet.

Where do the discussion happen?

lib/fetch/body.js
```diff
@@ -0,0 +1,277 @@
+'use strict'
+
+const util = require('../core/util')
+const { toWebReadable } = require('./util')
+const { kState } = require('./symbols')
+const { Blob } = require('buffer')
+const { Readable } = require('stream')
+const { NotSupportedError } = require('../core/errors')
+const { ReadableStream } = require('stream/web')
```
ardatan, line 9: So this is not usable in older versions of Node. I think this is a breaking change because undici was supporting Node 12 before.

ronag: It still works on node 12.

ardatan: It imports `stream/web` and it fails. Are you sure because `stream/web` isn't available in <16.

ronag: This file is not imported if version <16

ronag: Our CI runs node 12,14 and 16

ardatan: Undici used to support the `request.body` from node-fetch before but now it imports something from `stream/web` and this blocks us to update the latest version. I can create a reproduction if you want.

ronag: Please open an issue and I'll have a look.

ronag: I think I see the problem. It has nothing to do with node versions. Doesn't work on any version.

ronag: https://github.com/nodejs/undici/pull/955

ardatan: Thanks! And also there is another issue with the minor versions of Node 16 that doesn't have `stream/web`.

ronag: New release published with the fixes you wanted.

ardatan: Thank you very much for quick response and fix 🙏🙏

Issue #934, opened 2021-08-07
Types: Http methods enum


zekth:
## This would solve...

When using typescript and writing dispatcher it's convinient to have access to enum values when it's possible, like http verbs. Currently `method` for dispatcher is set as `string`. Would be nice to add an interface for it. 

see: https://github.com/nodejs/undici/blob/22efd7035c24ac06ffc17ec527f6974dac6cb482/types/dispatcher.d.ts#L43

## The implementation should look like...

```ts

declare namespace Dispatcher {
  export interface DispatchOptions {
    origin?: string | URL;
    path: string;
    method: HttpMethod;
    ...
  }
  export enum HttpMethod {
    GET = 'GET',
    HEAD = 'HEAD',
    POST = 'POST',
    PUT = 'PUT',
    DELETE = 'DELETE',
    CONNECT = 'CONNECT',
    OPTIONS = 'OPTIONS',
    TRACE = 'TRACE',
    PATCH = 'PATCH'
  }
}

```

szmarczak:
```typescript
type Method = 'GET' | 'HEAD' | 'POST' | 'PUT' | 'DELETE' | 'CONNECT' | 'OPTIONS' | 'TRACE' | 'PATCH';
```

No need to use enum.

PR #935, opened 2021-08-07
types: HttpMethod type

zekth:
fix: #934

Add the easiness of use for TypeScript developers and ensure valid value to be set for `method` parameters.


mcollina (approved):
lgtm

PR #950, opened 2021-08-13
fix: terminate fetch on gc

ronag:


ronag:
@szmarczak

lib/fetch/index.js
```diff
@@ -30,6 +30,20 @@ const { ReadableStream } = require('stream/web')
...
+  const next = []
+  for (const entry of registry) {
+    if (entry.ref.deref() === undefined) {
+      entry.abort()
+    } else {
+      next.push(entry)
+    }
+  }
+  registry = next
+}, 1e3).unref()
```
ronag, line 45: We could use a finalizationRegistry as well but that would be a lot less deterministic and the user might queue up lot's of requests before GC would start up.

ronag: The number of elements in the "registry" is limited by the number of active connections so it should normally be quite small.

ronag: @mcollina wdyt? deterministic polling loop or undeterministic `FinalizationRegistry`?

mcollina: The polling is unfeasible, it will keep processes active even with no actual traffic. Let's go with FinalizationRegistry.

ronag: Fixed. PTAL.

mcollina (approved):
lgtm

szmarczak:
> The user agent can terminate the fetch because the associated body is not observable.

> `window.promise = fetch("https://www.example.com/").then(res => res.headers)`

If this isn't done already, can you patch this?

ronag:
> If this isn't done already, can you patch this?

I think that is done.

PR #954, opened 2021-08-13
feat: fetch decompression

ronag:


lib/fetch/index.js
```diff
@@ -1554,9 +1563,85 @@ function httpNetworkFetch (
...
 
+          const codings = headers.get('content-encoding')
+          if (codings === 'gzip' || codings === 'x-gzip') {
+            this.decoder = zlib.createGunzip()
+          } else if (codings === 'deflate' || codings === 'x-deflate') {
+            this.decoder = zlib.createInflate()
+          } else if (codings === 'br') {
+            this.decoder = zlib.createBrotliDecompress()
+          } else {
+            // TODO: What to do when coding is unsupported?
```
ronag, line 1574: @mcollina @szmarczak What to do when coding is unsupported?

szmarczak: Got returns raw buffer. I guess `fetch` would do this as well. Let me check

ronag: Also it's a little weird that the end user has no way of determining whether the response has been decoded or not.

ronag:
@szmarczak there is a node-fetch test that says:

> 'should make capitalised Content-Encoding lowercase'

But I can't find anything in the spec that says we should do that. wdyt?

szmarczak:
> But I can't find anything in the spec that says we should do that. 

Me neither. It needs to be case-insensitive only when creating a list from it.

szmarczak:
It should be left as it is. The only normalization going on is trimming.

ronag:
I removed that test.

lib/fetch/index.js
ronag: @szmarczak would you mind checking what chrome does?

szmarczak: Returns the value with no errors.

szmarczak: > Also it's a little weird that the end user has no way of determining whether the response has been decoded or not.

I agree. Can you ask someone from the `fetch` team about this?

mcollina (approved):
lgtm

Ethan-Arrowood (approved):
this looks great! nice work

lib/fetch/index.js
```diff
@@ -1562,78 +1556,155 @@ function httpNetworkFetch (
...
+              .map((x) => x.trim()) ?? []
+
+          const decoders = []
+
+          // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding
+          for (const coding of codings) {
+            if (/(x-)?gzip/.test(coding)) {
+              decoders.push(zlib.createGunzip())
+            } else if (/(x-)?deflate/.test(coding)) {
+              decoders.push(zlib.createInflate())
```
kanongil, line 1573: Shouldn't this be `zlib.createDeflate()`?

lib/fetch/index.js
```diff
@@ -1562,78 +1556,155 @@ function httpNetworkFetch (
...
+            } else if (coding === 'br') {
+              decoders.push(zlib.createBrotliDecompress())
+            } else {
+              // TODO: What to do when coding is invalid or unsupported?
```
kanongil, line 1577: Spec says to [just passthrough](https://fetch.spec.whatwg.org/#handle-content-codings). Similar to the `identity` encoding.

This means any unknown non-`identity` encoding should invalidate this decoder chain.

kanongil (commented):
Great, I looked at the fetch feature yesterday and wondered where the encodings where handled.

lib/fetch/index.js
ronag: Would you mind opening a PR to fix?

lib/fetch/index.js
ronag: inflate is actually correct

lib/fetch/index.js
ronag: Fixed in https://github.com/nodejs/undici/commit/963ca033466cacbf6c72e299871f1e6aa4d1ced8

PR #955, opened 2021-08-13
fix: request body.body regression

ronag:
Refs: https://github.com/nodejs/undici/pull/928#discussion_r688506456


mcollina:
seems CI is not happy

PR #957, opened 2021-08-14
Add connecting event

artur-ma:
Newrelic is using `http.Agent#createConnection` to track the time it took to create connection
https://github.com/newrelic/node-newrelic/blob/main/lib/instrumentation/core/http.js#L534-L555

Currently I dont see a way to do it in undici, correct me if im wrong


ronag:
This doesn’t actually do what I think you want it to do.  It’s better if you created your own connector. See the ‘connect’ option in the code.

artur-ma:
@ronag yes, you are right, this PR is useless :)

I can use my own connector to track it, but Im trying to thing that APMs such as NR will need to create generic instrumentation for undici.

closing

PR #961, opened 2021-08-15
feat(fetch): FormData

ronag:


lib/fetch/formdata.js
```diff
@@ -0,0 +1,176 @@
...
+  chunks.push(`--${boundary}--`)
+
+  return new Blob(chunks, {
+    type: 'multipart/form-data; boundary=' + boundary
+  })
+}
+
+function blobToFormData (blob) {
+  // TODO
+  return new FormData()
```
ronag, line 173: Still missing parsing :(

ronag: Where is the spec for this?

szmarczak: https://xhr.spec.whatwg.org/#formdata

package.json
```diff
@@ -96,5 +95,8 @@
     "compilerOptions": {
       "esModuleInterop": true
     }
+  },
+  "dependencies": {
+    "dicer": "^0.3.0"
```
mcollina, line 100: I'm -1 on using dicer.

mcollina (changes requested):

package.json
szmarczak: image

lib/fetch/formdata.js
szmarczak: https://datatracker.ietf.org/doc/html/rfc7578

package.json
ronag: It works... what would you prefer? I'd like to replace it with our own implementation. However, I do not have time to do that atm. Can we land it with dicer as experimental and then make replacing it a requirement for stable release?

ronag: Also there are more question with `FormData` that we need to sort out https://github.com/nodejs/undici/issues/964

lib/fetch/formdata.js
```diff
@@ -0,0 +1,124 @@
...
+    value: null
+  }
+
+  // Set entry’s name to name.
+  entry.name = name
+
+  // If value is a Blob object and not a File object, then set value to a new File
+  // object, representing the same bytes, whose name attribute value is "blob".
+  if (value instanceof Blob && !(value instanceof File)) {
+    // TODO
```
ronag, line 107: @jasnell How do I synchronously create a new `Blob` from an existing one?

octet-stream: Blobs can take another blob as the `blobPart`:

```
const blob1 = new Blob(["a", "b"])
const blob2 = new Blob(["c", "d"])

const blob = new Blob([blob1, blob2])
```

Same for File.

package.json
ronag: fastify-multipart uses busboy which uses dicer

ronag: It's the best option I can find on npm.

mcollina: oh I know that it's the best on npm. I'm ok in using dicer for now, but I would also like to keep the "no dependency" approach.

We might consider vendoring it and iterating.

ronag: Does that mean you are fine with landing this in it's current form for now?

lib/fetch/formdata.js
```diff
@@ -0,0 +1,131 @@
...
+
+  // If value is (now) a File object and filename is given, then set value to a
+  // new File object, representing the same bytes, whose name attribute value is
+  // filename.
+  // TODO: This is a bit weird... What if passed value is a File?
+  // Do we just override the name attribute? Since it says "if value is (now)"
+  // does that mean that this lives inside the previous condition? In which case
+  // creating one more File instance doesn't make much sense....
+  if (value instanceof File && filename) {
+    value = new File([value], filename)
```
octet-stream, line 121: I believe you can merge logic for File and Blob:

```js
if (value instanceof File) {
  // 1. If filename argument is provided then use it as the second argument
  // 2. Else try to use value.name as filename argument
  // 3. If value has no name property (value is Blob), then set filename to "blob"
  value = new File([value], filename || value.name || "blob", {
    type: value.type,
    lastModified: value.lastModified
  })
}
```

lib/fetch/formdata.js
```diff
@@ -0,0 +1,131 @@
...
+  * [Symbol.iterator] () {
+    // The value pairs to iterate over are this’s entry list’s entries with
+    // the key being the name and the value being the value.
+    for (const { name, value } of this[kState]) {
+      yield [name, value]
+    }
+  }
+}
+
+function makeEntry (name, value, filename) {
```
octet-stream, line 94: Do not forget to cast name and value (in case if it not Blob or File) to a string.

octet-stream (changes requested):

lib/fetch/formdata.js
ronag: Fixed

lib/fetch/formdata.js
ronag: I'd like to keep the exact wording of the spec.

octet-stream: This is fine. It was just a suggestion, because I just thinks it does not make any difference in terms of behaviour.

lib/fetch/file.js
```diff
@@ -0,0 +1,70 @@
+'use strict'
+
+const { Blob } = require('buffer')
+const { kState } = require('./symbols')
+
+class File extends Blob {
```
jasnell, line 6: FWIW, I'm planning to have `File` in core along side `Blob`. Was looking to add it within the next few weeks.

ronag: Excellent!

ronag: Then we can probably remove this before going out of experimental.

lib/fetch/body.js
```diff
@@ -71,17 +64,61 @@ function extractBody (object, keepalive = false) {
...
     // Set action to this step: read object.
-    action = async (onNext, onError, onComplete) => {
-      try {
-        onNext(await object.arrayBuffer())
-        onComplete()
-      } catch (err) {
-        onError(err)
-      }
+    action = async function * () {
+      yield await object.arrayBuffer() // TODO: chunkify once we have Blob.stream()
```
ronag, line 121: @jasnell: Any plans for `Blob.stream`?

jasnell: Blob.stream landed in `HEAD` last week :-)

lib/fetch/file.js
ronag: @jasnell any chance you would consider doing `FormData` as well? Feel free to copy paste from this PR.

lib/fetch/formdata.js
```diff
@@ -0,0 +1,217 @@
+'use strict'
+
+const { Blob } = require('buffer')
+const { kState } = require('./symbols')
+const { File } = require('./file')
+const { HTMLFormElement } = require('./util')
+
+class FormData {
```
jasnell, line 8: We could, and probably should, bring this into core also.

ronag: ditto with `Headers`

jasnell: @Ethan-Arrowood has already been working in that direction https://github.com/nodejs/node/pull/38986

ronag: Yes, we based our work on his. I believe he discontinued his efforts in favor of including it through undici.

lib/fetch/body.js
```diff
@@ -71,16 +74,67 @@ function extractBody (object, keepalive = false) {
 
     // Set source to a copy of the bytes held by object.
     source = new Uint8Array(object)
+  } else if (object instanceof FormData) {
+    const boundary = '----formdata-undici-' + Math.random()
+    const prefix = `--${boundary}\r\nContent-Disposition: form-data; name="`
+
+    /*! formdata-polyfill. MIT License. Jimmy Wärting <https://jimmy.warting.se/opensource> */
```
jimmywarting, line 81: Well, look at that :) 
Nice to see my code is being referred here :P

lib/fetch/file.js
jimmywarting: This is what happened to fetch-blob, We only had a Blob class first, then everyone else where creating there own File class and extended it. So that's why we added a File class to fetch-blob as well. they kind of go hand in hand

ronag:
@mcollina

lib/fetch/body.js
```diff
@@ -71,16 +74,67 @@ function extractBody (object, keepalive = false) {
...
-    action = async (onNext, onError, onComplete) => {
-      try {
-        onNext(await object.arrayBuffer())
-        onComplete()
-      } catch (err) {
-        onError(err)
+    action = async function * (object) {
+      for await (const chunk of blobGen(object)) {
+        yield chunk
       }
```
mcollina, line 138:
```suggestion
      yield * blobGen(object)
```

mcollina (changes requested):
This is extremely low on test for the amount of things it does.

ronag:
> This is extremely low on test for the amount of things it does.

I intend to go through all of the fetch code and increase coverage once most of the functionality has landed (I guess this is the last piece).

PR #962, opened 2021-08-15
Fix `toWebReadable` finishing too early

szmarczak:
Other `clone` tests error that `clone().buffer` is not a function

### DIFF
--- a/lib/fetch/util.js
+++ b/lib/fetch/util.js
@@ -180,7 +180,7 @@ function toWebReadable (streamReadable) {
     } else {
       // Must not use `process.nextTick()`,
       // see https://github.com/nodejs/node/issues/39758
-      Promise.resolve().then(() => {
+      queueMicrotask(() => {
         controller.close()
       })
     }
--- a/lib/fetch/util.js
+++ b/lib/fetch/util.js
@@ -178,7 +178,11 @@ function toWebReadable (streamReadable) {
     if (err) {
       controller.error(err)
     } else {
-      controller.close()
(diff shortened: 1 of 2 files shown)


lib/fetch/util.js
```diff
@@ -178,7 +178,11 @@ function toWebReadable (streamReadable) {
     if (err) {
       controller.error(err)
     } else {
-      controller.close()
+      // Must not use `process.nextTick()`,
+      // see https://github.com/nodejs/node/issues/39758
+      Promise.resolve().then(() => {
```
szmarczak, line 183: I chose `Promise.resolve()` because it finishes before `setImmediate` (which finishes before `queueMicrotask`). Not sure what's best.

ronag (approved):

lib/fetch/util.js
ronag: I would prefer queueMicrotask I think

szmarczak:
The spec uses `queueMicrotask` but `process.nextTick` results in race condition, so we need to `controller.close()` when all other web streams work is done - `Promise.resolve()` should be sufficient enough.

lib/fetch/util.js
szmarczak, line 183:
```suggestion
      queueMicrotask(() => {
```

ronag (approved):

ronag:
@jasnell FYI

ronag:
Thanks @szmarczak excellent work!

PR #973, opened 2021-08-17
fix: redirect handler should not modify stream state

ronag:
Use EE.prototype.on to avoid mutating stream flowing
state.

### DIFF
--- a/lib/handler/redirect.js
+++ b/lib/handler/redirect.js
@@ -91,6 +91,10 @@ class RedirectHandler {
   }
 
   onHeaders (statusCode, headers, resume, statusText) {
+    if (this.maxRedirections === 0) {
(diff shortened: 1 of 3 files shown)


lib/handler/redirect.js
```diff
@@ -51,8 +52,7 @@ class RedirectHandler {
 
       if (typeof this.opts.body.readableDidRead !== 'boolean') {
         this.opts.body[kBodyUsed] = false
-        // TODO (fix): Don't mutate readable state...
-        this.opts.body.on('data', function () {
+        EE.prototype.on.call(this.opts.body, 'data', function () {
```
szmarczak, line 55:
```suggestion
        EE.prototype.once.call(this.opts.body, 'data', function () {
```

mcollina (approved):
lgtm

PR #976, opened 2021-08-17
feat(fetch): async iterable body

ronag:


mcollina (approved):
lgtm, this is so much simpler.

Issue #481, opened 2020-11-16
Support HTTP 103


zekth:
I implemented a plugin for HTTP 103 early hints in fastify : https://github.com/zekth/fastify-early-hints
I wanted to introduce undici as part of the test stack because currently i use `curl` as a subprocress and parse the stdout for it.

The problem currently with undici are those lines : https://github.com/nodejs/undici/blob/f083cc025353c1fc0cbd81fea8797a9b033ecb5d/lib/client-request.js#L88-L90

Indeed it skips the `information` part of the raw HTTP message. I concede that HTTP 103 is experimental for now, but if it's possible to implement it i'd suggest to add those informations in the response of the request like:

```js
client.request({
  path: '/',
  method: 'GET'
}, function (err, data) {
  if (err) {
    // handle this in some way!
    return
  }

  const {
    statusCode,
    headers,
    body,
    info
  } = data
```

Because of the RFC the structure of the `info` would be something like:
```js
{
  raw: [
    [
      ['Link: </main.css>; rel=preload; as=style'],
      ['Link: </script.js>; rel=preload; as=script'],
    ],
    [
      ['Link: </style.css>; rel=preload; as=style']
    ],
  ],
  compiled: [
      ['Link: </main.css>; rel=preload; as=style'],
      ['Link: </script.js>; rel=preload; as=script'],
      ['Link: </style.css>; rel=preload; as=style']
  ]
}
```

ref:
- https://httpwg.org/specs/rfc8297.html
- https://chromium.googlesource.com/chromium/src/+/master/docs/early-hints.md
- https://www.fastly.com/blog/beyond-server-push-experimenting-with-the-103-early-hints-status-code

mcollina:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

ronag:
The point of info messages is that they arrive before the response headers. The API you suggest kind of breaks that.

We would need an api which allow listening to those events before the response headers.

I don’t know how that would look and is why we haven’t implemented it yet.

zekth:
a solution would be to emit event on Information headers like:
```js
client.on('info', handler)
```
Also do we want to have the whole http transite compiled into `request` like i proposed or not?
I surely can try to implement it in a PR.

PR #977, opened 2021-08-17
feat: onInfo

zekth:
This is a draft implementation on the onInfo feature proposal. This adds more informations on the context lifecycle of the request. Currently everything is handled in a callback but do we want to have all the infos stored in the request object to be added in the result? With a flag like `verbose` it would make sense otherwise it would fill-up the memory and impact the performances.

If this is ok to implement like so i'll port it to pipeline and stream and add tests for sure. 

## This relates to...

fix: https://github.com/nodejs/undici/issues/481

## Rationale

This adds the possibility to get HTTP info headers through a callback

## Changes

Adding a placeholder for an `onInfo` callback.

### Features

Adds an callback handler on Info Headers

### Bug Fixes

N/A

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [x] Merge ready

### DIFF
--- a/lib/api/api-request.js
+++ b/lib/api/api-request.js
@@ -49,7 +49,7 @@ class RequestHandler extends AsyncResource {
     this.body = body
     this.trailers = {}
     this.context = null
-    this.onInfo = onInfo || (() => {})
+    this.onInfo = onInfo || null
 
     if (util.isStream(body)) {
       body.on('error', (err) => {
@@ -73,7 +73,9 @@ class RequestHandler extends AsyncResource {
     const { callback, opaque, abort, context } = this
 
     if (statusCode < 200) {
-      this.onInfo({ statusCode, headers })
+      if (this.onInfo) {
+        this.onInfo({ statusCode, headers: util.parseHeaders(headers) })
+      }
       return
     }
 
--- a/lib/api/api-request.js
+++ b/lib/api/api-request.js
@@ -15,7 +15,7 @@ class RequestHandler extends AsyncResource {
       throw new InvalidArgumentError('invalid opts')
     }
 
-    const { signal, method, opaque, body } = opts
+    const { signal, method, opaque, body, onInfo } = opts
 
     try {
       if (typeof callback !== 'function') {
@@ -28,6 +28,10 @@ class RequestHandler extends AsyncResource {
 
       if (method === 'CONNECT') {
         throw new InvalidArgumentError('invalid method')
+      }
+
+      if (onInfo && typeof onInfo !== 'function') {
+        throw new InvalidArgumentError('invalid onInfo callback')
       }
 
       super('UNDICI_REQUEST')
@@ -45,7 +49,7 @@ class RequestHandler extends AsyncResource {
     this.body = body
     this.trailers = {}
     this.context = null
-    this.onInfo = opts.onInfo || (() => {})
+    this.onInfo = onInfo || (() => {})
 
     if (util.isStream(body)) {
       body.on('error', (err) => {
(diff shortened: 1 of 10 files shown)


mcollina:
Looks good to me in principle. I would like to see a full example were this is used.

zekth:
Here is a working example with earlyHints.

Client:
```js
(async () => {
  const infos = []
  await request(
    'http://localhost:3000/', {
      onInfo: (x) => { infos.push(x) }
    }
  )
  assert(!!infos.find(x=>x.statusCode===103))
})()
```
Server
```js
const Fastify = require("fastify");
const eh = require("fastify-early-hints");

const fastify = Fastify({ logger: true });
fastify.register(eh);

fastify.get("/", async (request, reply) => {
  reply.eh.add([
    "Link: </style.css>; rel=preload; as=style",
    "Link: </script.js>; rel=preload; as=script",
  ]);
  return { hello: "world" };
});

const start = async () => {
  try {
    await fastify.listen(3000);
  } catch (err) {
    fastify.log.error(err);
    process.exit(1);
  }
};
start();
```

Also if the server implements HTTP 100 : https://datatracker.ietf.org/doc/html/rfc2616/#section-8.2.3 
We can add a behavior wait for the HTTP 100 before sending the body (when expect header will be supported). So at least we can assert on it as we currently have no hook like https://nodejs.org/api/http.html#http_event_information

mcollina:
Nice! Let's get this added!

test/client-request.js
```diff
@@ -370,3 +370,27 @@ test('request post body no extra data handler', { skip: nodeMajor < 16 }, (t) =>
...
   })
 })
+
+test('request with onInfo callback', (t) => {
+  t.plan(2)
+  const infos = []
+  const server = createServer((req, res) => {
+    res.writeContinue()
+    res.setHeader('Content-Type', 'application/json')
+    res.end(JSON.stringify({ foo: 'bar' }))
```
mcollina, line 380: Can you add a test where the response is ended after the client receives onInfo?

zekth: Sure

mcollina:
The change is good!

lib/api/api-request.js
```diff
@@ -69,7 +73,7 @@ class RequestHandler extends AsyncResource {
     const { callback, opaque, abort, context } = this
 
     if (statusCode < 200) {
-        this.onInfo({statusCode,headers})
+      this.onInfo({ statusCode, headers })
```
ronag, line 76:
```suggestion
      if (this.onInfo) {
        this.onInfo({ statusCode, headers })
      }
```

ronag (commented):
I don't like the API from a cosmetic standpoint (I don't have any better suggestion) but technically it LGTM

lib/api/api-request.js
zekth: Wouldn't it be slower to add a boolean condition rather than relying on a noop function? That was the point of this design.

ronag: conditions that are predicable are basically zero overhead

zekth:
ready for review @ronag @mcollina . Happy to change the api as it's not the best cosmeticly but haven't found anything better. Maybe it can be labelled as experimental to accept further api changes?

mcollina (approved):
lgtm

ronag (approved):
I'm ok with it. Implementation is trivial so if we find a better api we can just doc deprecate it and leave the code.

ronag:
@zekth tests are failing

zekth:
> @zekth tests are failing

It seems to be related to github runner cache handling permission error. Not related to this pr in particular

Issue #926, opened 2021-08-04
Port tests from undici-fetch


Ethan-Arrowood:
Porting tests from https://github.com/ethan-arrowood/undici-fetch

Ethan-Arrowood:
going to let #928 land first before doing this since it includes some changes to the fetch api

ronag:
@Ethan-Arrowood ping, it has landed

PR #978, opened 2021-08-17
Initialize some fetch tests

Ethan-Arrowood:
Worked on #926 but with all of the spec related changes, most of my previous tests were broken. Headers class mostly worked so I included those, but request response and body were just a mess. Probably best to start over with the new implementations.

I've stubbed those out so other contributors know where the tests need to go and such


Ethan-Arrowood:
Why is tap failing on previous Node versions? Do we not export the fetch api or something?

ronag:
We don't run fetch tests on node <16

Ethan-Arrowood:
i'll add a skip to these tests then

ronag:
https://github.com/nodejs/undici/blob/main/test/client-fetch.js#L9

Ethan-Arrowood:
ahh i did it with the script thing and some npm scripts. Do you have a preference?

PR #984, opened 2021-08-18
test: fix coverage reporter

ronag:


ronag:
Doesn't feel like it gets reports from all tests... https://app.codecov.io/gh/nodejs/undici/branch/coverage-report

Running `npm run coverage` locally I get more coverage on fetch.

ronag:
How do we make sure we always pick the report from v16?

szmarczak:
codecov merges those reports

szmarczak:
See #985 - maybe this will fix it

szmarczak:
Can you merge #985 into this PR? #985 should solve a conflict between tap and nyc.

package.json
```diff
@@ -41,7 +41,7 @@
     "test:tdd": "tap test/*.js -w",
     "test:typescript": "tsd",
     "coverage": "nyc npm run test",
-    "coverage:ci": "npm run coverage -- --reporter=lcovonly",
+    "coverage:ci": "node scripts/verifyVersion.js 16 && nyc --reporter=lcov npm run test || echo Skipping",
```
szmarczak, line 44: Can you undo the version check? codecov merges these reports

ronag:
I’ll fix those when I’m back at my computer. What I still don’t like is that we run all the tests twice. Once normally and once for coverage.

should we only run coverage?l (which includes testing)?

szmarczak:
> should we only run coverage?l (which includes testing)?

Yep. :)

PR #985, opened 2021-08-18
Attempt to fix coverage output

szmarczak:


szmarczak:
what? it still doesn't see it :(

Issue #981, opened 2021-08-18
toUSVString


ronag:
I notice node core is using `toUSVString`. I believe we should do the same. However, it's unclear what it's implementation is.... https://github.com/nodejs/node/blob/master/lib/internal/url.js#L144-L152

Also a little unclear in which places exactly we need to use it. Everywhere where we do `String(value)`?

ronag:
@szmarczak

szmarczak:
`process.binding('url').toUSVString` ?

szmarczak:
https://developer.mozilla.org/en-US/docs/Web/API/USVString

ronag:
> `process.binding('url').toUSVString` ?

I think process.binding is deprecated

PR #986, opened 2021-08-18
fix(fetch): toUSVString

ronag:


lib/fetch/util.js
```diff
@@ -280,6 +280,23 @@ function tryUpgradeRequestToAPotentiallyTrustworthyURL (request) {
   // TODO
 }
 
+function _toUSVString (val) {
+  // TODO: This is internal to node core...
+  return val
```
ronag, line 285: @jasnell can we expose this from node core or is there a JS/WASM implementation with the corresponding functionality?

szmarczak: https://github.com/nodejs/undici/issues/981#issuecomment-901439727

ronag: That won't work anymore unfortunately.

mcollina:
I don't understand what this brings.

ronag:
> I don't understand what this brings.

Placeholder for spec compliance. We need to eventually either implement `_toUSVString` ourselves or have node core expose it.

ronag:
We do the same thing for the `URL` module in node core.

mcollina:
Awesome, ok. Let's expose this from core.

Emciu (commented):
Ok

Emciu (commented):
.

Emciu:
> .

M

Issue #988, opened 2021-08-19
RangeError: WebAssembly.Instance(): Out of memory: wasm memory


janpio:
## Bug Description

We have been happy users of `undici` at @prisma for quite a while and finally found the time to upgrade to 4.x. Fundamentally everything seems to be working fine (after making sure the `llhttp` binaries are copied into our builds as well), but a specific test suite is failing with this error:

```
    RangeError: WebAssembly.Instance(): Out of memory: wasm memory

      at Object.<anonymous> (../../node_modules/.pnpm/undici@4.4.5/node_modules/undici/lib/client.js:400:16)
      at Object.<anonymous> (../../node_modules/.pnpm/undici@4.4.5/node_modules/undici/index.js:3:16)
```

https://github.com/prisma/prisma/pull/8823
https://github.com/prisma/prisma/runs/3373629489#step:9:1284

It also does not fail all the time on all systems, but only reliably on GH Actions (it's fine locally or on BuildKite for example).

Any idea what we should look at here?

## Reproducible By

Our test suite on GH Actions is the only reproduction.

## Expected Behavior

No failure.

## Logs & Screenshots

See links above

## Environment

Ubuntu 20.04, Node 12.22.5 or Node 14.17.5

### Additional context

Possibly interesting: In the failing test, `undici` should not even really be used as we have two implementations: An Prisma Engine binary that starts a webserver which we talk to with `undici`, and one where we have a Node-API library which we talk to directly, no webserver and `undici` involved - and _THAT_ is the test run that is failing (which makes not sense!?).

ronag:
@janpio Do you think you could try https://github.com/nodejs/undici/pull/989?

janpio:
I did, and it indeed seems to solve the problem on our side!
https://github.com/prisma/prisma/pull/8831

Awesome!

(Although I am still confused why only the code path that does _not_ really use undici runs into this... something smells funny there.)

PR #989, opened 2021-08-19
fix: lazy load llhttp

ronag:
Fixes: https://github.com/nodejs/undici/issues/988

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -384,72 +384,78 @@ const { readFileSync } = require('fs')
 const constants = require('./llhttp/constants')
 const EMPTY_BUF = Buffer.alloc(0)
 
-let mod
-try {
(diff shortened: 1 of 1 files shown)


mcollina:
why this fixes #988?

ronag:
Because it doesn't load the binaries when they are not needed. Seems to be a memory limited environment.

mcollina:
It would load them eventually. Maybe they can try this branch before we land it?

ronag:
> It would load them eventually. 

From what I understood from the issue they don't.

ronag:
https://github.com/prisma/prisma/pull/8823

> Because we import both LibraryEngine and BinaryEngine files and undici tries to load the wasm even if not instantiated / needed

mcollina (approved):
lgtm

janpio:
We did test this branch and it indeed made our tests pass: https://github.com/nodejs/undici/issues/988#issuecomment-902189995

PR #994, opened 2021-08-21
fix: Include final URL in redirect history.

ShogunPanda:
In the current implementation, a URL is added to `context.history` only if in the response there is a redirect to follow.
Unfortunately, this results in the final URL of the redirects chain to never be present in the history and apparently there are no other ways to get such URL.

This PR fixes this problem by appending each URL visited. In case of no redirections, the `context.history` will contain just one element.

Hope this helps!


mcollina (approved):
lgtm

PR #997, opened 2021-08-24
feat: `fetch` types

### LINKED ISSUE #953: Definition of `fetch` missing from types
The fetch implementation ( #928 ) does not include pdated typescript types. So it's not possible to import fetch without getting errors when using typescript.

wojpawlik:
## This relates to...

fixes #953

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/test/types/fetch.test-d.ts
+++ b/test/types/fetch.test-d.ts
@@ -1,0 +1,144 @@
+import { URL } from 'url'
+import { Blob } from 'buffer'
+import { expectType, expectError } from 'tsd'
+import {
+  BodyInit,
+  fetch,
+  FormData,
+  Headers,
+  HeadersInit,
+  Request,
+  RequestCache,
+  RequestCredentials,
+  RequestDestination,
+  RequestInit,
+  RequestMode,
+  RequestRedirect,
+  Response,
+  ResponseInit,
+  ResponseType,
+} from '../..'
+
+const requestInit: RequestInit = {}
+const responseInit: ResponseInit = { status: 200, statusText: 'OK' }
+
+declare const request: Request
+declare const headers: Headers
+declare const response: Response
+
+expectType<string | undefined>(requestInit.method)
+expectType<boolean | undefined>(requestInit.keepalive)
+expectType<HeadersInit | undefined>(requestInit.headers)
+expectType<BodyInit | undefined>(requestInit.body)
+expectType<RequestRedirect | undefined>(requestInit.redirect)
+expectType<string | undefined>(requestInit.integrity)
+expectType<AbortSignal | undefined>(requestInit.signal)
+
+expectType<number | undefined>(responseInit.status)
+expectType<string | undefined>(responseInit.statusText)
+expectType<HeadersInit | undefined>(responseInit.headers)
+
+expectType<Headers>(new Headers())
+expectType<Headers>(new Headers({}))
+expectType<Headers>(new Headers([]))
+expectType<Headers>(new Headers(headers))
+expectType<Headers>(new Headers(undefined))
+
+expectType<Request>(new Request(request))
+expectType<Request>(new Request('https://example.com'))
+expectType<Request>(new Request(new URL('https://example.com')))
+expectType<Request>(new Request(request, requestInit))
+expectType<Request>(new Request('https://example.com', requestInit))
+expectType<Request>(new Request(new URL('https://example.com'), requestInit))
+
+expectType<Promise<Response>>(fetch(request))
+expectType<Promise<Response>>(fetch('https://example.com'))
+expectType<Promise<Response>>(fetch(new URL('https://example.com')))
+expectType<Promise<Response>>(fetch(request, requestInit))
+expectType<Promise<Response>>(fetch('https://example.com', requestInit))
+expectType<Promise<Response>>(fetch(new URL('https://example.com'), requestInit))
+
+expectType<Response>(new Response())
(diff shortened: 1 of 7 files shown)


ronag (approved):

szmarczak (changes requested):
[needs tests](https://github.com/SamVerschueren/tsd)

mcollina (changes requested):
Could you add tests?

wojpawlik:
No. I have no experience testing types.

szmarczak:
You can find the current ts tests here: https://github.com/nodejs/undici/tree/main/test/types

Ethan-Arrowood:
Since @ronag did a lot of work to make fetch spec compat here lets make sure we are not missing anything. The official TS types for `fetch` are available in lib.dom (https://github.com/microsoft/TypeScript/blob/cec2fda9a53620dc545a2c4d7b0156446ab145b4/lib/lib.dom.d.ts#L18378) here is a link to the exact fetch definition

wojpawlik:
I was improving the types by comparing them on MDN. So now I'm working on adding `FormData` types as well, because it's valid `BodyInit`.

types/fetch.d.ts
```diff
@@ -0,0 +1,132 @@
...
+  readonly statusText: string
+  readonly type: ResponseType
+  readonly url: string
+  readonly redirected: boolean
+
+  readonly body: ControlledAsyncIterable | null
+  readonly bodyUsed: boolean
+
+  readonly arrayBuffer: () => Promise<Buffer>
+  readonly json: () => Promise<unknown>
```
KhafraDev, line 125: I just want to point out that this will break type compatibility with other packages & the native types.

```ts
const r = await fetch('https://some-json-api.com/');
const b = await r.json();

// in node-fetch, native types, etc (without casts)
b.something.somethingelse.anotherthing
// with this
b.something
  ^ 'cannot read property "something" of unknown'

```

`unknown` is a much better type than `any`, it adds an extra step for those transitioning from another package. Not sure if it really matters to you guys but I just wanted to point it out. 👍

wojpawlik: `node-fetch@3.0.0-beta.10` uses `Promise<unknown>` as well, and switching to `Promise<any>` wouldn't be breaking

test/types/fetch.test-d.ts
```diff
@@ -1,7 +1,7 @@
 import { URL } from 'url'
 import { Blob } from 'buffer'
 import { expectType } from 'tsd'
-import { fetch, FormData, Headers, HeadersInit, Request, Response, RequestInit, ResponseInit } from '../..'
+import { BodyInit, fetch, FormData, Headers, HeadersInit, Request, RequestCache, RequestCredentials, RequestDestination, RequestInit, RequestMode, RequestRedirect, ResponseType, Response, ResponseInit } from '../..'
```
szmarczak, line 4: Please make this multi-line :)

test/types/fetch.test-d.ts
```diff
@@ -0,0 +1,137 @@
...
+expectType<Response>(new Response(new Uint16Array(), responseInit))
+expectType<Response>(new Response(new Int32Array(), responseInit))
+expectType<Response>(new Response(new Uint32Array(), responseInit))
+expectType<Response>(new Response(new Float32Array(), responseInit))
+expectType<Response>(new Response(new Float64Array(), responseInit))
+expectType<Response>(new Response(new BigInt64Array(), responseInit))
+expectType<Response>(new Response(new BigUint64Array(), responseInit))
+expectType<Response>(new Response(new ArrayBuffer(0), responseInit))
+expectType<Response>(Response.error())
+expectType<Response>(Response.redirect('https://example.com', NaN))
```
szmarczak, line 95: `Response.redirect('https://example.com', NaN)` so throws so I would remove `NaN` from the types

test/types/fetch.test-d.ts
```diff
@@ -0,0 +1,137 @@
...
+expectType<IterableIterator<[string, string]>>(headers.entries())
+
+expectType<RequestCache>(request.cache)
+expectType<RequestCredentials>(request.credentials)
+expectType<RequestDestination>(request.destination)
+expectType<Headers>(request.headers)
+expectType<string>(request.integrity)
+expectType<string>(request.method)
+expectType<RequestMode>(request.mode)
+expectType<string>(request.redirect)
```
szmarczak, line 113: Can you strictly type this? https://developer.mozilla.org/en-US/docs/Web/API/Request/redirect

types/formdata.d.ts
```diff
@@ -0,0 +1,97 @@
+// based on https://unpkg.com/browse/formdata-node@4.0.1/@type/FormData.d.ts (MIT)
```
szmarczak, line 1:
```suggestion
// based on https://github.com/octet-stream/form-data (MIT)
```

test/types/fetch.test-d.ts
wojpawlik: `prettier` would've done it for me, if `undici` used it :wink:

szmarczak (approved):

types/formdata.d.ts
wojpawlik: I changed it, but on second thought I think it was better before

ronag (approved):

mcollina (approved):
lgtm

Issue #998, opened 2021-08-24
Fetch and GC


ronag:
The fetch standard describes [quite aggressive garbage collection expectations](https://fetch.spec.whatwg.org/#garbage-collection) which are not possible to implement in pure JS. ~~I believe we would need to have access to the actual reference count and not simply wait for GC to occur like we do atm through `FinalizationRegistry`.~~

We should at least document this discrepancy.

I'm not even sure if this is possible to do in native code (@addaleax)?

addaleax:
@ronag What’s the “actual reference count” here?

In native code, yes, I think this is possible, but since we’re talking about a mostly-JS implementation of fetch, that feels anything but trivial to implement.

ronag:
> @ronag What’s the “actual reference count” here?

Whether or not the object is observable or not.

> In native code, yes, I think this is possible, but since we’re talking about a mostly-JS implementation of fetch, that feels anything but trivial to implement.

Yea, I was just assuming native code. We could maybe do it "properly" if/when we land fetch in core.

addaleax:
> > @ronag What’s the “actual reference count” here?
> 
> Whether or not the object is observable or not.

But that’s what GC/FinalizationRegistry tells you, no?

I think, properly done, this does work without native code – e.g. you could detect when `res.body` is GC’ed, and make the assumption that nothing will start reading from it at that point if it hasn’t already.

ronag:
> But that’s what GC/FinalizationRegistry tells you, no?

Yes, but it does it lazily. In theory it might never run or it could take a very long time before it runs. The way I read the fetch spec it's expected to release the connection pretty soon after it's no longer possible to observe it.

ronag:
e.g. you might start a request every second, not consume the response body, the gc doesn't fire because memory pressure is low and you run out out socket handles or something.

ronag:
@jasnell I think you might end up with similar requirements for `File`.

ronag:
I don't know if `WeakRef.deref` returns undefined once an object is GC:d or when it's no longer observable. I would assume the former so that doesn't help either.

addaleax:
> > But that’s what GC/FinalizationRegistry tells you, no?
> 
> Yes, but it does it lazily and in an indeterministic fashion. In theory it might never run or it could take a very long time before it runs. 

I mean, yeah, but that’s also true for any GC tracking – there’s no reason to expect that tracking on the native side would give us earlier or more reliable results.

[...]

PR #999, opened 2021-08-24
docs: add notes about gc spec compliance

ronag:
Refs: https://github.com/nodejs/undici/issues/998

### DIFF
--- a/README.md
+++ b/README.md
@@ -189,7 +189,9 @@ recommended to always either consume or cancel the response body.
 // Do
 const headers = await fetch(url)
   .then(async res => {
-    await res.arrayBuffer()
+    for await (const chunk of res) {
+      // force consumption of body
+    }
     return res.headers
   })
 
--- a/README.md
+++ b/README.md
@@ -166,6 +166,24 @@ Only supported on Node 16+.
 
 This is [experimental](https://nodejs.org/api/documentation.html#documentation_stability_index) and is not yet fully compliant with the Fetch Standard. We plan to ship breaking changes to this feature until it is out of experimental.
 
+#### Specification Compliance
+
+This section documents parts of the [fetch specification](https://fetch.spec.whatwg.org) which Undici does
+not support or does not fully implement.
+
+##### Garbage collection
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -166,6 +166,40 @@ Only supported on Node 16+.
...
+      // force consumption of body
+    }
+    return res.headers
+  })
+
+// Do not
+const headers = await fetch(url)
+  .then(res => res.headers)
+```
+
```
ronag, line 202: Should we add an alternative solution through `--expose-gc`? i.e. if `global.gc` is available use some internal undici heuristics to manually force garbage collection to release connections?

ronag: @addaleax: Thoughts?

ronag: e.g. if all `Pool` connections are allocated (i.e. we stall or deadlock) then force a gc.

ronag: When moving into core we could achieve this without the runtime flag.

ronag: @mcollina

README.md
```diff
@@ -166,6 +166,39 @@ Only supported on Node 16+.
 
 This is [experimental](https://nodejs.org/api/documentation.html#documentation_stability_index) and is not yet fully compliant with the Fetch Standard. We plan to ship breaking changes to this feature until it is out of experimental.
 
+#### Specification Compliance
+
+This section documents parts of the [fetch specification](https://fetch.spec.whatwg.org) which Undici does
+not support or does not fully implement.
+
+##### Garbage collection
```
ronag, line 174:
```suggestion
##### Garbage Collection
```

mcollina (approved):
lgtm

README.md
mcollina: I don't think adding a `--expose-gc` solution is needed.
The people that would need the automatic cleaning up of resources are the ones that will never pass `--expose-gc`.

I think we should document 1) what error they will be getting and 2) what to do to mitigate that error, e.g. consume the body.

szmarczak: > The people that would need the automatic cleaning up of resources are the ones that will never pass --expose-gc.

I second this.

PR #1003, opened 2021-08-25
fix: force gc on request pause

ronag:
Refs: #998

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -1632,7 +1632,6 @@ function httpNetworkFetch (
       },
       {
         decoder: null,
-        gc: false,
 
         onConnect (abort) {
           if (connection.destroyed) {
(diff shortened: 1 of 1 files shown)


mcollina:
I'm -0 on this - I would never rely on `global.gc()` and the people that would need this behavior won't know to pass it either.

However this needs a test.

ronag:
Maybe once/if we land in core.

ronag:
@addaleax would be nice if there was a way to access global.gc without the cli flag :)

addaleax:
@ronag There’s always been good reasons not to expose that, though.

ronag:
> @ronag There’s always been good reasons not to expose that, though.

Indeed. Probably more reasons not to expose. But this is at least one possible reason to expose. 😄

PR #1008, opened 2021-08-30
fix(fetch): avoid optional chaining for node 12

Songkeys:
I'm not so sure about this PR, but undici-fetch introduced some optional chaining syntax which breaks Node.js 12. I think we can avoid this.


ronag (changes requested):
Fetch shouldn't be loaded at all for node < 16. So this shouldn't be necessary. Do you have repro?

Songkeys:
> Fetch shouldn't be loaded at all for node < 16.

Ah. Sorry. I missed this. I only noticed the `engine` attr in undici's package.json. Thanks for pointing this out.

PR #1009, opened 2021-08-31
chore: increasing code coverage

joaopedrocampos:
Adds more unit tests for some cases.

Fixes the `badPorts` logic because the URL `port` property is a string and, without ports being strings the `badPorts.includes(url.port)` never finds any value.

Fix some `lint` issues


ronag:
Good job! Thanks!

PR #1014, opened 2021-09-01
feat: HTTP/2 support

szmarczak:
**WIP**

- [x] PoC is working
- [ ] unref session when 0 pending streams, ref on stream\
- [ ] ...

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -4,6 +4,7 @@
 
 const assert = require('assert')
 const net = require('net')
+const http2 = require('http2')
 const util = require('./core/util')
 const Request = require('./core/request')
 const Dispatcher = require('./dispatcher')
@@ -44,7 +45,7 @@ const {
   kNeedDrain,
   kNoRef,
   kKeepAliveDefaultTimeout,
-  kHostHeader,
+  kHost,
   kClosed,
   kDestroyed,
   kPendingIdx,
@@ -61,7 +62,8 @@ const {
   kBodyTimeout,
   kStrictContentLength,
   kConnector,
-  kMaxRedirections
+  kMaxRedirections,
+  kHttp2Session
 } = require('./core/symbols')
 
 class Client extends Dispatcher {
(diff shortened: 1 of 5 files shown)


lib/client.js
```diff
@@ -1326,7 +1359,147 @@ function _resume (client, sync) {
...
+    if (client[kStrictContentLength]) {
+      errorRequest(client, request, new RequestContentLengthMismatchError())
+      return false
+    }
+
+    process.emitWarning(new RequestContentLengthMismatchError())
+  }
+
+  try {
+    request.onConnect((err) => {
```
szmarczak, line 1411: Should we call `onConnect` immediately or on stream `ready` event? https://nodejs.org/api/http2.html#http2_event_ready

lib/client.js
```diff
@@ -1326,7 +1360,153 @@ function _resume (client, sync) {
...
+  } else {
+    assert(false)
+  }
+
+  console.log('http/2 request')
+  // TODO: @szmarczak: ref only if current streams count is 0
+  session.ref()
+  const stream = session.request(headers)
+  stream.on('response', headers => {
+    if (request.onHeaders(Number(headers[':status']), headers, stream.resume.bind(stream), '') === false) {
```
szmarczak, line 1470: The handler expects an array but the native `http2` module provides us with an object. What should we do?

ronag: Convert to array

lib/core/request.js
```diff
@@ -226,7 +226,7 @@ function processHeader (request, key, val) {
   ) {
     throw new NotSupportedError('expect header not supported')
   } else {
-    request.headers += `${key}: ${val}\r\n`
+    request.headers[key] = val
```
ronag, line 229: This will make http1 slower... :/

lib/core/connect.js
```diff
@@ -29,6 +29,8 @@ function buildConnector ({ maxCachedSessions, socketPath, timeout, ...opts }) {
       const session = sessionCache.get(servername) || null
 
       socket = tls.connect({
+        // TODO: @szmarczak: move this to client.js
+        ALPNProtocols: ['h2', 'http/1.1'],
```
ronag, line 33: experimental + disabled by default?

ronag:
Maybe would be better to target @jasnell's node quic branch? I'm not that comfortable with the node core http2 implementation.

espoal:
@ronag what's the issue with the core node HTTP/2?

ronag:
> @ronag what's the issue with the core node HTTP/2?

https://github.com/nodejs/node/issues?q=is%3Aopen+is%3Aissue+label%3Ahttp2

nickmccurdy:
Superseded by #2061

PR #1017, opened 2021-09-03
fix(fetch): FormData.set empties this[kState] if k exists

rayw000:
### What changed

1. `FormData.set(k, v)` will empty the entire `this[kState]` if `k` exists. This is not what it is supposed to do. Fix it.
2. Improve coverage of `lib/fetch/formdata.js`

### DIFF
--- a/lib/fetch/formdata.js
+++ b/lib/fetch/formdata.js
@@ -158,9 +158,9 @@ class FormData {
     const idx = this[kState].findIndex((entry) => entry.name === name)
     if (idx !== -1) {
       this[kState] = [].concat(
-        this[kState].slice(0, idx).filter((entry) => entry.name !== name),
-        [entry],
-        this[kState].slice(idx).filter((entry) => entry.name !== name),
+        this[kState].slice(0, idx),
+        entry,
+        this[kState].slice(idx + 1).filter((entry) => entry.name !== name),
       )
     } else {
       // 4. Otherwise, append entry to this’s entry list.
--- a/lib/fetch/formdata.js
+++ b/lib/fetch/formdata.js
@@ -157,8 +157,11 @@ class FormData {
     // replace the first such entry with entry and remove the others.
     const idx = this[kState].findIndex((entry) => entry.name === name)
     if (idx !== -1) {
-      this[kState][idx] = entry
-      this[kState] = this[kState].filter((entry) => entry.name !== name)
+      this[kState] = [].concat(
+        this[kState].slice(0, idx).filter((entry) => entry.name !== name),
+        [entry],
+        this[kState].slice(idx).filter((entry) => entry.name !== name),
+      )
     } else {
       // 4. Otherwise, append entry to this’s entry list.
       this[kState].push(entry)
--- a/lib/fetch/formdata.js
+++ b/lib/fetch/formdata.js
@@ -157,11 +157,11 @@ class FormData {
     // replace the first such entry with entry and remove the others.
     const idx = this[kState].findIndex((entry) => entry.name === name)
     if (idx !== -1) {
-      this[kState] = [].concat(
-        this[kState].slice(0, idx),
+      this[kState] = [
+        ...this[kState].slice(0, idx),
         entry,
-        this[kState].slice(idx + 1).filter((entry) => entry.name !== name),
-      )
+        ...this[kState].slice(idx + 1).filter((entry) => entry.name !== name),
+      ]
     } else {
       // 4. Otherwise, append entry to this’s entry list.
       this[kState].push(entry)
--- a/test/fetch/formdata.js
+++ b/test/fetch/formdata.js
@@ -21,7 +21,7 @@ test('arg validation', (t) => {
     form.append()
   }, TypeError)
   t.throws(() => {
-    form.append('asd', 'asd', 'asd')
+    form.append('k', 'not usv', '')
   }, TypeError)
 
   // delete
@@ -37,7 +37,7 @@ test('arg validation', (t) => {
     FormData.prototype.get.call(null)
   }, TypeError)
   t.throws(() => {
-    form.delete()
+    form.get()
   }, TypeError)
 
   // getAll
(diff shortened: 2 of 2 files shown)


lib/fetch/formdata.js
```diff
@@ -157,8 +157,11 @@ class FormData {
     // replace the first such entry with entry and remove the others.
     const idx = this[kState].findIndex((entry) => entry.name === name)
     if (idx !== -1) {
-      this[kState][idx] = entry
-      this[kState] = this[kState].filter((entry) => entry.name !== name)
+      this[kState] = [].concat(
+        this[kState].slice(0, idx).filter((entry) => entry.name !== name),
+        [entry],
+        this[kState].slice(idx).filter((entry) => entry.name !== name),
+      )
```
ronag, line 164:
```suggestion
      this[kState][idx] = entry
```

Isn't this sufficient?

ronag: i.e just remove `this[kState] = this[kState].filter((entry) => entry.name !== name)`

rayw000: > i.e just remove `this[kState] = this[kState].filter((entry) => entry.name !== name)`

But this wil not remove all other entries which names are also `<name>`, as described in the comment above.

rayw000: `this[kState]` could be something like
```javascript
[
    { name: "k1", value: "v1" },
    { name: "k2", value: "v2" },
    { name: "k1", value: "v3" },  // another entry which name is `k1`
]
```
Probably it's why `FormData.getAll()` exists?

rayw000: And also, perhaps `form.append` should insert entries at the beginning of `this[kState]` if they are dup key entries. Otherwise, this code fragment will be a little weird.
```javascript
const form = new FormData()
form.set('k1', 'v1')
form.append('k1', 'v2')
form.get('k1') // we might want it to be 'v2', but get 'v1', weird.
```
How do you think of it?

ronag: The spec says 'replace the first such entry with entry and remove the others.'

rayw000: > The spec says 'replace the first such entry with entry and remove the others.'

Indeed. That's exactly what this PR does.

Perhaps I didn't make myself clear before. Let me be more specific.

Consider this following code snippet
```javascript
const form = new FormData()
form.append('k1', 'v1')
form.append('k2', 'v2')
form.append('k1', 'v3')
form.set('k1', 'v4')
```
The spec in the method `FormData.set` says (according to my understanding) it will replace the first matched entry with the new created entry, and remove all other entries sharing the same name.

After running the code snippet above, we expect `this[kState]` to be

```javascript
[
    { name: 'k1', value: 'v4' },
    { name: 'k2', value: 'v2' },
]
```

but currently we get
```javascript
[
    { name: 'k1', value: 'v4' },
]
```
This is because line 161 empties the entire `this[kState]` array.
https://github.com/nodejs/undici/blob/1c2259cf7a89ed427cf915719ee301ef7751b4eb/lib/fetch/formdata.js#L158-L166

if we just delete `this[kState] = this[kState].filter((entry) => entry.name !== name)` as you suggested previously, we get
```javascript
[
    { name: 'k1', value: 'v4' },
    { name: 'k2', value: 'v2' },
    { name: 'k1', value: 'v3' },
]
```
according to the spec described, we need to remove `{ name: 'k1', value: 'v3' }`, but we didn't.

ronag: You are right.

ronag, line 164:
```js
const arr = this[kState]
arr[idx++] = entry
const len = arr.length
for (let n = idx; n < len; ++n) {
  if (arr[n].name === name) {
    continue
  }
  if (n !== idx) {
    arr[idx++] = arr[n]
  } else {
    idx++
  }
}
arr.length = idx

```

ronag:
```js
this[kState][idx++] = entry
this[kState] = [
  ...this[kState].slice(0, idx),
  ...this[kState].slice(idx).filter((entry) => entry.name !== name)
]
```

ronag: Alternative ways of achieving the same thing with different complexity and performance.

lib/fetch/formdata.js
```diff
@@ -157,8 +157,11 @@ class FormData {
     // replace the first such entry with entry and remove the others.
     const idx = this[kState].findIndex((entry) => entry.name === name)
     if (idx !== -1) {
-      this[kState][idx] = entry
-      this[kState] = this[kState].filter((entry) => entry.name !== name)
+      this[kState] = [].concat(
+        this[kState].slice(0, idx).filter((entry) => entry.name !== name),
+        [entry],
+        this[kState].slice(idx).filter((entry) => entry.name !== name),
```
ronag, line 163:
```suggestion
        this[kState].slice(idx + 1).filter((entry) => entry.name !== name),
```

lib/fetch/formdata.js
rayw000: Oh yes. I forgot that we do not need to search `name` in the first part cuz it is the first match.

lib/fetch/formdata.js
```diff
@@ -157,8 +157,11 @@ class FormData {
     // replace the first such entry with entry and remove the others.
     const idx = this[kState].findIndex((entry) => entry.name === name)
     if (idx !== -1) {
-      this[kState][idx] = entry
-      this[kState] = this[kState].filter((entry) => entry.name !== name)
+      this[kState] = [].concat(
+        this[kState].slice(0, idx),
+        entry,
+        this[kState].slice(idx + 1).filter((entry) => entry.name !== name),
+      )
```
ronag, line 164:
```suggestion
      this[kState] = [
        ...this[kState].slice(0, idx),
        entry,
        ...this[kState].slice(idx + 1).filter((entry) => entry.name !== name),
      ]
```

ronag (approved):

ronag:
Thanks!

szmarczak (approved):

PR #1021, opened 2021-09-04
refactor: improve methods and readability

vierofernando:
Improved the methods by changing up their loops and improved the readability by removing redundant code and with the help of loop changes.


szmarczak:
What's the performance hit here?

PR #1022, opened 2021-09-05
fix(fetch): improve coverage of file.js

rayw000:
improve coverage of `lib/fetch/file.js`

### DIFF
--- a/test/fetch/client-fetch.js
+++ b/test/fetch/client-fetch.js
@@ -19,11 +19,13 @@ test('args validation', (t) => {
   }, TypeError)
 })
 
-test('call last modified', (t) => {
-  t.plan(1)
+test('return value of File.lastModified', (t) => {
+  t.plan(2)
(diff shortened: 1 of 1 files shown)


test/fetch/client-fetch.js
```diff
@@ -8,6 +8,24 @@ const { ReadableStream } = require('stream/web')
...
+  t.throws(() => {
+    File.prototype.lastModified.call(null)
+  }, TypeError)
+})
+
+test('call last modified', (t) => {
+  t.plan(1)
+
+  const f = new File(['asd1'], 'filename123')
+  t.ok(f.lastModified)
```
ronag, line 26: Can you check that the value makes sense as well?

rayw000: @ronag Sure. Updated.

ronag:
Thanks!

PR #1026, opened 2021-09-09
test(fetch): requiring consts from constants.js instead of defining

rayw000:
Better to require constants from `lib/fetch/constants.js` in `test/fetch/headers.js` rather than to define them.


ronag:
Thanks

PR #1031, opened 2021-09-13
feat: support `max` keep-alive directive

szmarczak:


lib/client.js
```diff
@@ -859,6 +863,21 @@ class Parser {
       socket[kReset] = true
     }
 
+    for (let i = 0; i < headers.length; i+= 2) {
+      if (headers[i].toString().toLowerCase() === 'keep-alive') {
```
szmarczak, line 867: I wonder what's the performance impact here...

lib/client.js
```diff
@@ -859,6 +863,21 @@ class Parser {
...
+        const value = headers[i + 1].toString().toLowerCase();
+        const maxIndex = value.indexOf('max=');
+
+        if (maxIndex !== -1) {
+          const pipelining = parseInt(value.slice(maxIndex + 4));
+          if (Number.isFinite(pipelining)) {
+            // What if client[kPipelining] is undefined?
```
szmarczak, line 874: ?

lib/client.js
szmarczak: Does the parser already makes the headers lowercase?

lib/client.js
szmarczak: Should we make `client[kPipelining]` to always be a number?

mcollina: I think so

lib/client.js
ronag: look how we do it in other places, compare with length first

szmarczak: I finally did this like the `timeout` directive is done. PTAL :)

lib/client.js
```diff
@@ -200,7 +202,9 @@ class Client extends Dispatcher {
 
   // TODO: Make private?
```
szmarczak, line 203: I think `.pipelining` could be read-only 🤔

lib/client.js
szmarczak: @ronag wdyt?

lib/client.js
```diff
@@ -854,6 +858,12 @@ class Parser {
       } else {
         client[kKeepAliveTimeoutValue] = client[kKeepAliveDefaultTimeout]
       }
+
+      const max = this.keepAlive ? util.parseKeepAliveMax(this.keepAlive) : null
+
+      if (max != null) {
+        client[kPipelining] = Math.min(client[kMaxPipelining], max)
```
mcollina, line 865: I think we should be significantly more conservative than this, as the closer we are to the maximum number of request left the higher the risk of going over the limit. Something like the following would do:

```suggestion
        client[kPipelining] = Math.min(client[kMaxPipelining], max >= client[kMaxPipelining] ? max >= client[kMaxPipelining] : 1)
```

mcollina (changes requested):
please add a few test when this is close to the limit.

mcollina (changes requested):
According to

https://datatracker.ietf.org/doc/html/draft-thomson-hybi-http-timeout-03#section-2.2.1

>    The "max" parameter has been used to indicate the maximum number of
   requests that would be made on the connection.  This parameter is
   deprecated.  Any limit on requests can be enforced by sending
   "Connection: close" and closing the connection.

`max` is deprecated and should not be sent.

Issue #940, opened 2021-08-11
Document high level fetch API


ronag:
At least minimal high level documentation with references to MDN?

RaisinTen:
Can we close this now that https://github.com/nodejs/undici/pull/1035 has landed or does this task require some more documentation to be added?

PR #1035, opened 2021-09-15
docs: add basic example of fetch

o-t-w:
Adding a simple usage example of Fetch and a very brief explanation of Node Streams vs Web Streams. 

https://github.com/nodejs/undici/issues/940

### DIFF
--- a/README.md
+++ b/README.md
@@ -186,8 +186,8 @@ Nodejs has two kinds of streams: [web streams](https://nodejs.org/dist/latest-v1
     import {fetch} from 'undici';
     
     async function fetchStream() {
-    const response = await fetch('https://example.com')
-    const stream = response.body;
+        const response = await fetch('https://example.com')
+        const stream = response.body;
     }
 ```
 
--- a/README.md
+++ b/README.md
@@ -171,19 +171,16 @@ Basic usage example:
 ```
     import {fetch} from 'undici';
     
-    // Promises
-    fetch('https://example.com')
-    .then(res => res.json())
-    .then(json => console.log(json))
-    
-    // Async/Await
+    async function fetchJson() {
     const res = await fetch('https://example.com')
     const json = await res.json()
+    console.log(json);
+    }
 ```
 
 #### `response.body`
 
-Nodejs has two kinds of streams: [web streams](https://nodejs.org/dist/latest-v16.x/docs/api/webstreams.html) which follow the API of the WHATWG web standard found in browsers, and an older Node-specific [streams API](https://nodejs.org/api/stream.html). `response.body` returns a web stream. If you require a Node stream, it is possible to convert between the two using `.fromWeb()`. 
+Nodejs has two kinds of streams: [web streams](https://nodejs.org/dist/latest-v16.x/docs/api/webstreams.html) which follow the API of the WHATWG web standard found in browsers, and an older Node-specific [streams API](https://nodejs.org/api/stream.html). `response.body` returns a readable web stream. If you require a Node stream, it is possible to convert between the two using `.fromWeb()`. 
 
 ```
     import {fetch} from 'undici';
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

README.md
```diff
@@ -166,6 +166,31 @@ Only supported on Node 16+.
...
+Basic usage example:
+
+```
+    import {fetch} from 'undici';
+    
+    async function fetchJson() {
+    const res = await fetch('https://example.com')
+    const json = await res.json()
+    console.log(json);
+    }
```
mcollina, line 178:
```suggestion
    async function fetchJson() {
        const res = await fetch('https://example.com')
        const json = await res.json()
        console.log(json);
    }
```

README.md
```diff
@@ -166,6 +166,31 @@ Only supported on Node 16+.
...
+
+Nodejs has two kinds of streams: [web streams](https://nodejs.org/dist/latest-v16.x/docs/api/webstreams.html) which follow the API of the WHATWG web standard found in browsers, and an older Node-specific [streams API](https://nodejs.org/api/stream.html). `response.body` returns a readable web stream. If you require a Node stream, it is possible to convert between the two using `.fromWeb()`. 
+
+```
+    import {fetch} from 'undici';
+    
+    async function fetchStream() {
+    const response = await fetch('https://example.com')
+    const stream = response.body;
+    }
```
mcollina, line 191:
```suggestion
    async function fetchStream() {
        const response = await fetch('https://example.com')
        const stream = response.body;
    }
```

README.md
```diff
@@ -166,6 +166,31 @@ Only supported on Node 16+.
...
+#### `response.body`
+
+Nodejs has two kinds of streams: [web streams](https://nodejs.org/dist/latest-v16.x/docs/api/webstreams.html) which follow the API of the WHATWG web standard found in browsers, and an older Node-specific [streams API](https://nodejs.org/api/stream.html). `response.body` returns a readable web stream. If you require a Node stream, it is possible to convert between the two using `.fromWeb()`. 
+
+```
+    import {fetch} from 'undici';
+    
+    async function fetchStream() {
+        const response = await fetch('https://example.com')
+        const stream = response.body;
```
ronag, line 190: Missing fromWeb?

o-t-w: I mentioned `.fromWeb()` as a brief aside. The main purpose of the code block you highlighted is just to make it super clear that response.body is a stream.

README.md
```diff
@@ -166,6 +166,38 @@ Only supported on Node 16+.
...
+    }
+```
+
+If you require a Node stream, it is possible to convert between the two using `.fromWeb()`.
```
mcollina, line 194: Can you make that a link?

ronag: it's actually `.fromWeb()` and `.toWeb()` if you are referring to converting between the two, which implies both ways.

README.md
ronag: Isn't this example supposed to show how to do it? Otherwise what's the purpose of the example?

ronag (approved):

README.md
```diff
@@ -166,6 +166,33 @@ Only supported on Node 16+.
...
+        const json = await res.json()
+        console.log(json);
+    }
+```
+
+#### `response.body`
+
+Nodejs has two kinds of streams: [web streams](https://nodejs.org/dist/latest-v16.x/docs/api/webstreams.html) which follow the API of the WHATWG web standard found in browsers, and an older Node-specific [streams API](https://nodejs.org/api/stream.html). `response.body` returns a readable web stream. If you would prefer to work with a Node stream you can convert a web stream using `.fromWeb()`. 
+
+```
```
octet-stream, line 185: Same for this example - add `js`, please.

PR #1036, opened 2021-09-16
Add more diagnostic channels

fatal10110:
Trying to close https://github.com/nodejs/undici/issues/1032

### DIFF
--- a/docs/api/DiagnosticsChannel.md
+++ b/docs/api/DiagnosticsChannel.md
@@ -80,7 +80,7 @@ diagnosticsChannel.channel('undici:request:error').subscribe(({ request, error }
 })
 ```
 
-## `undici:client:write`
+## `undici:client:sendHeaders`
 
 This message is published right before first byte of the request is written to the socket.
 
@@ -89,13 +89,13 @@ This message is published right before first byte of the request is written to t
 ```js
 import diagnosticsChannel from 'diagnostics_channel'
 
-diagnosticsChannel.channel('undici:client:write').subscribe(({ request, headers, socket }) => {
+diagnosticsChannel.channel('undici:client:sendHeaders').subscribe(({ request, headers, socket }) => {
   // request is the same object undici:request:create
   console.log(`Full headers list ${headers.split('\r\n')}`);
 })
 ```
 
-## `undici:client:connect`
+## `undici:client:BeforeConnect`
 
 This message is published before creating a new connection for **any** request. 
 You can not assume that this event is related to any specific request.
@@ -103,7 +103,7 @@ You can not assume that this event is related to any specific request.
 ```js
 import diagnosticsChannel from 'diagnostics_channel'
 
-diagnosticsChannel.channel('undici:client:connect').subscribe((connectParams) => {
+diagnosticsChannel.channel('undici:client:beforeConnect').subscribe((connectParams) => {
   // const { host, hostname, protocol, port, servername } = connectParams
 })
 ```
@@ -120,14 +120,14 @@ diagnosticsChannel.channel('undici:client:connected').subscribe((socket, ...conn
 })
 ```
 
-## `undici:client:connect:error`
+## `undici:client:connectError`
 
 This message is published if it did not succeed to create new connection
 
 ```js
 import diagnosticsChannel from 'diagnostics_channel'
 
-diagnosticsChannel.channel('undici:client:connect:error').subscribe(({ error, socket, ...connectParams }) => {
+diagnosticsChannel.channel('undici:client:connectError').subscribe(({ error, socket, ...connectParams }) => {
   // const { host, hostname, protocol, port, servername } = connectParams
   console.log(`Connect failed with ${error.message}`)
 })
--- a/docs/api/DiagnosticsChannel.md
+++ b/docs/api/DiagnosticsChannel.md
@@ -79,3 +79,55 @@ diagnosticsChannel.channel('undici:request:error').subscribe(({ request, error }
   // request is the same object undici:request:create
 })
 ```
+
+## `undici:client:write`
+
+This message is published right before first byte of the request is written to the socket.
+
+*Note*: It exact headers that will be sent to the server in raw foramt. 
+
+```js
+import diagnosticsChannel from 'diagnostics_channel'
+
+diagnosticsChannel.channel('undici:client:write').subscribe(({ request, headers }) => {
+  // request is the same object undici:request:create
+  console.log(`Full headers list ${headers.split('\r\n')}`);
+})
+```
+
+## `undici:client:connect`
+
+This message is published before creating a new connection for **any** request. 
+You can not assume that this event is related to any specific request.
+
+```js
+import diagnosticsChannel from 'diagnostics_channel'
+
+diagnosticsChannel.channel('undici:client:write').subscribe((connectParams) => {
+  // const { host, hostname, protocol, port, servername } = connectParams
+})
+```
+
+## `undici:client:connected`
+
+This message is published after connection established.
+
+```js
+import diagnosticsChannel from 'diagnostics_channel'
+
+diagnosticsChannel.channel('undici:client:write').subscribe((connectParams) => {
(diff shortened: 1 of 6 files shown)


mcollina (commented):
Nice work! Could you add tests and docs?

mcollina (commented):
I think we should pass the socket to each event.

fatal10110:
> I think we should pass the socket to each event.

The only channels that can pass the socket is "connected" and "write"
should I add it?

mcollina:
I would say so. Do we need the write publish?

mcollina:
You might just add the client everywhere.

fatal10110:
> Do we need the write publish?

I think yes, as discussed here, but maybe i'm wrong
https://github.com/nodejs/undici/issues/1032#issuecomment-920816090

> You might just add the client everywhere.
@ronag does not want to "have a dependency on the client object."
https://github.com/nodejs/undici/issues/1032#issuecomment-918901681

fatal10110:
I've passed the socket, to "write" and "connect" channels. I can pass it to all channels except "create" it will require client to pass the socket to some callbacks, but sometimes the socket may be already destroyed or missing (on "error")

ronag (approved):

mcollina:
Can you run some benchmarks?

mcollina (approved):
lgtm

docs/api/DiagnosticsChannel.md
```diff
@@ -79,3 +79,55 @@ diagnosticsChannel.channel('undici:request:error').subscribe(({ request, error }
   // request is the same object undici:request:create
 })
 ```
+
+## `undici:client:write`
```
szmarczak, line 83: `write` -> `sendHeaders`

docs/api/DiagnosticsChannel.md
```diff
@@ -79,3 +79,55 @@ diagnosticsChannel.channel('undici:request:error').subscribe(({ request, error }
...
+```js
+import diagnosticsChannel from 'diagnostics_channel'
+
+diagnosticsChannel.channel('undici:client:write').subscribe(({ request, headers, socket }) => {
+  // request is the same object undici:request:create
+  console.log(`Full headers list ${headers.split('\r\n')}`);
+})
+```
+
+## `undici:client:connect`
```
szmarczak, line 98: `connect` -> `beforeConnect`

szmarczak (approved):

artur-ma:
This branch without subscribers

image

This branch with empty subscribers

image

mcollina (approved):
lgtm

lib/client.js
```diff
@@ -1412,6 +1460,10 @@ function write (client, request) {
     header += headers
   }
 
+  if (channels.sendHeaders.hasSubscribers) {
+    channels.sendHeaders.publish({ request, headers: header, socket })
```
bizob2828, line 1464: this def will achieve what i mentioned in https://github.com/nodejs/undici/issues/1032 however the headers are a string so it's now on the user to parse that string which to me is not ideal. Is there a reason why the headers aren't associated with the request all the time?

bizob2828: I guess you can also get what you need from the socket object so my previous comment/question may be moot

lib/client.js
```diff
@@ -1097,6 +1112,17 @@ function connect (client) {
   }
 
   client[kConnecting] = true
+
+  if (channels.beforeConnect.hasSubscribers) {
+    channels.beforeConnect.publish({
```
bizob2828, line 1117: you may need to attach the client.  i tried subscribing but I had no object to attach some context in the connected and/or connectError to properly end spans

lib/client.js
fatal10110: > I guess you can also get what you need from the socket object 

Not sure I understood what you are trying to get form the socket

> so it's now on the user to parse that string 

Undici operates headers as string, IMO the parsing should not be done by undici this extra work may affect performance, but the user may not use it at all, the headers string follows the HTTP standard (\r\n separator) should be easy to parse

bizob2828: In #1032 I requested outgoing headers, specifically the host header.  Since headers in client are just a string and not key/values pairs like in request it would've been a pain to parse a string to extract the host header.  However, since the socket is here it has the host that it is connected to which is the same so I can just rely on the socket object to get what I need instead of parsing headers string

lib/client.js
bizob2828: just to be clear, the client will need to be on all 3 connect events

lib/client.js
fatal10110: > key/values pairs like in request

If u add any headers to request, please note, on request object they are also string a string 
and should be added as follow `request.addHeader('hello', 'world')`

lib/client.js
ronag: Probably more appropriate to attach he connector?

lib/client.js
bizob2828: yep, you can see in this prototype I'm using `addHeader` https://github.com/newrelic/node-newrelic/pull/901/files#diff-9fe41f6ad1e32c80844e1948f016e7d692ea907edd2620e8b9245c1c9da67b37R70

lib/client.js
bizob2828: looks like the connector is a function, if so then I would only need to subscribe to beforeConnect and properly bind the async function

fatal10110: The only benefit from passing the client, is that both `connect` and `sendHeaders` will have the same context
If we will pass connector instead of client then I would not be able to send it to `sendHeaders` 

@bizob2828 what do u prefer?

lib/client.js
fatal10110: Here u are using `request.headers` as object, but it is a string 

```
https://github.com/newrelic/node-newrelic/pull/901/files#diff-9fe41f6ad1e32c80844e1948f016e7d692ea907edd2620e8b9245c1c9da67b37R54
```

see this test
https://github.com/nodejs/undici/blob/main/test/diagnostics-channel/get.js#L41

lib/client.js
```diff
@@ -1097,6 +1112,20 @@ function connect (client) {
...
+      connectParams: {
+        host,
+        hostname,
+        protocol,
+        port,
+        servername: client[kServerName]
+      },
+      client: this
```
bizob2828, line 1125: `this` is not the client in this context

lib/client.js
bizob2828: I don't need the same context between sendHeaders and connect, But I'm fine with passing the client.  I'm very new to this lib and I just built instrumentation. @ronag appears to be involved more heavily so I will defer to him

lib/client.js
bizob2828: yea i noticed that. Honestly this is copied code from our http-outbound and I don't see how a request header could ever be a symbol when serialized

lib/client.js
bizob2828: If we pass the connector in beforeConnect I don't see any value add with adding the connector in the connected and connectError. Typically you'd be binding the function in the beforeConnect and the other two events will be handled in the instrumentation primitives of the given vendor

mcollina:
wdyt @ronag?

lib/client.js
bizob2828: > If we pass the connector in beforeConnect I don't see any value add with adding the connector in the connected and connectError. Typically you'd be binding the function in the beforeConnect and the other two events will be handled in the instrumentation primitives of the given vendor

I see the maintainers have approved. don't sweat this. it'll still work and depending on vendor methods of attaching context this will work just fine

bizob2828:
@ronag @mcollina I'm about to open a PR for the New Relic agent that adds support for undici.  I will need this to be released so I was wondering when the next undici release was going to be published to npm?

mcollina:
We usually release once a week or similar.

bizob2828:
> We usually release once a week or similar.

Great, I'll keep an eye out.

PR #1041, opened 2021-09-21
Expose the socket in case of SocketError

delvedor:
If a `SocketError` happens, a user might want to dig deeper to understand why it has happened. For example, they might want to read the socket local/remote address of the socket.
This pr adds a `.socket` property to the `SocketError` which holds the socket that caused the error.

### DIFF
--- a/lib/core/errors.js
+++ b/lib/core/errors.js
@@ -147,12 +147,13 @@ class ClientClosedError extends UndiciError {
 }
 
 class SocketError extends UndiciError {
-  constructor (message) {
+  constructor (message, socket) {
     super(message)
     Error.captureStackTrace(this, SocketError)
     this.name = 'SocketError'
     this.message = message || 'Socket error'
     this.code = 'UND_ERR_SOCKET'
+    this.socket = socket
   }
 }
 
--- a/types/errors.d.ts
+++ b/types/errors.d.ts
@@ -70,7 +70,7 @@ declare namespace Errors {
   export class SocketError extends UndiciError {
     name: 'SocketError';
     code: 'UND_ERR_SOCKET';
-    socket: Socket | TLSSocket
+    socket: Socket | TLSSocket | null
   }
 
   /** Encountered unsupported functionality. */
--- a/types/errors.d.ts
+++ b/types/errors.d.ts
@@ -1,3 +1,6 @@
+import { Socket } from 'net'
+import { TLSSocket } from 'tls'
+
 export = Errors
 
 declare namespace Errors {
@@ -67,6 +70,7 @@ declare namespace Errors {
   export class SocketError extends UndiciError {
     name: 'SocketError';
     code: 'UND_ERR_SOCKET';
+    socket: Socket | TLSSocket
   }
 
   /** Encountered unsupported functionality. */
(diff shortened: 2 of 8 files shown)


delvedor:
I'm not sure how to add the socket info in `lib/api/api-upgrade.js` and `lib/api/api-connect.js`.

ronag:
I'm +/- 0. I would like to avoid exposing internals as much as possible so that in the future we could e.g. replace the socket implementation without breaking everything.

Rather than exposing the socket directly could you create a new object that exposes the properties that you need?

delvedor:
> replace the socket implementation without breaking everything.

I don't see the problem, the `.socket` property will be the different socket implementation that the user is aware of.

> Rather than exposing the socket directly could you create a new object that exposes the properties that you need?

I fear this approach might not be future-proof. Today I need the local/remote address, tomorrow might be something else, which will require another pr to adds the new properties (and so on...).

types/errors.d.ts
```diff
@@ -67,6 +70,7 @@ declare namespace Errors {
   export class SocketError extends UndiciError {
     name: 'SocketError';
     code: 'UND_ERR_SOCKET';
+    socket: Socket | TLSSocket
```
mcollina, line 73: Can you please add a test for this type?

lib/core/errors.js
```diff
@@ -147,12 +147,13 @@ class ClientClosedError extends UndiciError {
...
 
 class SocketError extends UndiciError {
-  constructor (message) {
+  constructor (message, socket) {
     super(message)
     Error.captureStackTrace(this, SocketError)
     this.name = 'SocketError'
     this.message = message || 'Socket error'
     this.code = 'UND_ERR_SOCKET'
+    this.socket = socket
```
mcollina, line 156: Where could we document this new property (apart from the TS change?)

mcollina (commented):
Adding a reference to the socket _might_ delay the time a socket is garbage collected. This is likely _not_ a problem here as the `fd` has already been destroyed when this error is called, but it might happen in some other cases if we keep following this pattern elsewhere.

Ragarding @ronag concern about future support for different kind of Socket object... I don't think that would come with anything less than a semver-major change.

I'm ok in shipping this.

lib/core/errors.js
delvedor: I can add a new section [here](https://github.com/nodejs/undici/blob/main/docs/api/Errors.md).

mcollina (approved):
lgtm

delvedor:
There is still https://github.com/nodejs/undici/pull/1041#issuecomment-923741896 that needs to be addressed.

mcollina:
@delvedor you mean https://github.com/nodejs/undici/blob/main/lib/api/api-connect.js#L43? I think you'd need to pass the `socket` to `onHeaders`.

types/errors.d.ts
mcollina, line 73: I think it's safer to have this typed as nullable.

mcollina (changes requested):

types/errors.d.ts
delvedor: Agreed.

mcollina (approved):
lgtm

mcollina:
@ronag any final objection

PR #1042, opened 2021-09-21
fix: Improve `File` and `FormData` types declarations.

octet-stream:
This PR brings various fixes for FormData and File types declarations, including:

1. `FormData`:
  * Declare class methods as they should be;
  * Minor fixes for methods documentation in code;
  * Remove unnecessary constructor declaration;
  * Remove unnecessary part from FormData class description;
2. `File`:
  * Replace getters with readonly properties;
  * Remove `File.stream()` method, because `File` inherits `Blob` which already has this method;
  * `FileOptions` interface inherits `BlobOptions` from `@types/node` declarations;
  * Rename `name` argument in `File` constructor to `fileName` (based on: https://github.com/microsoft/TypeScript/blob/5546f207c5d20578f1d215ccc110dd42bf8d359b/lib/lib.dom.d.ts#L5220)
  * Update `fileBits` description to reflect supported input (based on: https://developer.mozilla.org/en-US/docs/Web/API/File/File)

----

I also noticed both `File` and `Blob` can be constructed with objects that have `@@iterator`, this probably have to be reflected in declarations as well? Both in undici File and `@types/node`

See: https://github.com/web-platform-tests/wpt/blob/30e724ac90157333d34198f0c29131b2e2a6b1da/FileAPI/blob/Blob-constructor.any.js#L58-L105

### DIFF
--- a/types/file.d.ts
+++ b/types/file.d.ts
@@ -1,13 +1,9 @@
-// based on https://unpkg.com/browse/formdata-node@4.0.1/@type/File.d.ts (MIT)
+// Based on https://github.com/octet-stream/form-data/blob/2d0f0dc371517444ce1f22cdde13f51995d0953a/lib/File.ts (MIT)
 /// <reference types="node" />
 
-import { Blob } from 'buffer'
+import { Blob, BlobOptions } from 'buffer'
 
-export interface FileOptions {
-  /**
-   * Returns the media type ([`MIME`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types)) of the file represented by a `File` object.
-   */
-  type?: string
+export interface FileOptions extends BlobOptions {
   /**
    * The last modified date of the file as the number of milliseconds since the Unix epoch (January 1, 1970 at midnight). Files without a known last modified date return the current date.
    */
@@ -22,17 +18,17 @@ export declare class File extends Blob {
    * @param name The name of the file.
    * @param options An options object containing optional attributes for the file.
    */
-  constructor (fileBits: ReadonlyArray<string | NodeJS.ArrayBufferView | Blob>, name: string, options?: FileOptions)
+  constructor(fileBits: ReadonlyArray<string | NodeJS.ArrayBufferView | Blob>, name: string, options?: FileOptions)
+
   /**
    * Name of the file referenced by the File object.
    */
-  get name (): string
+  readonly name: string
+
   /**
    * The last modified date of the file as the number of milliseconds since the Unix epoch (January 1, 1970 at midnight). Files without a known last modified date return the current date.
    */
-  get lastModified (): number
-  get [Symbol.toStringTag] (): string
-  stream (): {
-    [Symbol.asyncIterator]: () => AsyncIterableIterator<Uint8Array>
-  }
+  readonly lastModified: number
+
+  readonly [Symbol.toStringTag]: string
 }
--- a/types/file.d.ts
+++ b/types/file.d.ts
@@ -14,11 +14,11 @@ export declare class File extends Blob {
   /**
    * Creates a new File instance.
    *
-   * @param fileBits An `Array` strings, or [`Buffer`](https://nodejs.org/dist/latest/docs/api/buffer.html#buffer_class_buffer), [`ArrayBuffer`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/ArrayBuffer), [`ArrayBufferView`](https://developer.mozilla.org/en-US/docs/Web/API/ArrayBufferView), [`Blob`](https://developer.mozilla.org/en-US/docs/Web/API/Blob) objects, or a mix of any of such objects, that will be put inside the [`File`](https://developer.mozilla.org/en-US/docs/Web/API/File).
-   * @param name The name of the file.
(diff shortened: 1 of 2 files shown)


ronag (approved):

types/file.d.ts
```diff
@@ -18,21 +14,21 @@ export declare class File extends Blob {
...
    * @param options An options object containing optional attributes for the file.
    */
-  constructor (fileBits: ReadonlyArray<string | NodeJS.ArrayBufferView | Blob>, name: string, options?: FileOptions)
+  constructor(fileBits: ReadonlyArray<string | NodeJS.ArrayBufferView | Blob>, fileName: string, options?: FileOptions)
+
   /**
    * Name of the file referenced by the File object.
    */
-  get name (): string
+  readonly name: string
```
szmarczak, line 26: I don't think this exists: https://w3c.github.io/FileAPI/#ref-for-dfn-FilePropertyBag

```
dictionary FilePropertyBag : BlobPropertyBag {
  long long lastModified;
};

enum EndingType { "transparent", "native" };

dictionary BlobPropertyBag {
  DOMString type = "";
  EndingType endings = "transparent";
};
```

types/file.d.ts
```diff
@@ -1,13 +1,9 @@
...
 
-import { Blob } from 'buffer'
+import { Blob, BlobOptions } from 'buffer'
 
-export interface FileOptions {
-  /**
-   * Returns the media type ([`MIME`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types)) of the file represented by a `File` object.
-   */
-  type?: string
+export interface FileOptions extends BlobOptions {
```
szmarczak, line 6: Maybe let's name this [`FilePropertyBag`](https://w3c.github.io/FileAPI/#ref-for-dfn-FilePropertyBag)? What do you think?

szmarczak, line 6: Maybe let's name this [`FilePropertyBag`](https://w3c.github.io/FileAPI/#ref-for-dfn-FilePropertyBag)? What do you think?

octet-stream: I do think it might broke someone's code if we rename it right away since it is exposed interface :)
But we can deprecate it and rename as you proposed.

types/formdata.d.ts
```diff
@@ -47,51 +45,60 @@ export declare class FormData {
...
    */
-  readonly [Symbol.iterator]: () => Generator<[string, FormDataEntryValue], any, unknown>
+  [Symbol.iterator](): Generator<[string, FormDataEntryValue], void>
+
   /**
    * Executes given callback function for each field of the FormData instance
    */
-  readonly forEach: (fn: (value: FormDataEntryValue, key: string, fd: FormData) => void, ctx?: unknown) => void
-  get [Symbol.toStringTag] (): string
+  forEach(fn: (value: FormDataEntryValue, key: string, fd: FormData) => void, ctx?: unknown): void
```
szmarczak, line 101: `fd` -> `formData`

szmarczak (approved):
lgtm

types/formdata.d.ts
octet-stream: > It would be good to rename fn to something more descriptive as well but function is a reserved keyword

Maybe to `callback`?

types/file.d.ts
szmarczak: My reasoning is to follow W3C naming, but I don't have anything against `FileOptions` as well :)

types/formdata.d.ts
szmarczak: Sounds good. This is also how MDN describes `forEach`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/forEach (for Map though, but still relevant)

octet-stream: See: https://github.com/microsoft/TypeScript/blob/5546f207c5d20578f1d215ccc110dd42bf8d359b/lib/lib.dom.d.ts#L5413

szmarczak: Can we use those interfaces instead? Or should we not?

octet-stream: Not so sure about that. At least `@types/node` has separate declarations for `Blob`.

octet-stream: > Sounds good. This is also how MDN describes forEach: developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map/forEach (for Map though, but still relevant)

Can't find documentation for this method in FormData, though it exists in browsers. The only place it appears is compatibility table on this page: https://developer.mozilla.org/en-US/docs/Web/API/FormData#browser_compatibility
Even the spec does not say anything about this method, where FormData is described: https://xhr.spec.whatwg.org/#interface-formdata

octet-stream: > Can we use those interfaces instead? Or should we not?

By the way, at least for FormData constructor Node and Browser versions aren't compatible: because browser version takes `form` argument, which is DOM element. See: https://developer.mozilla.org/en-US/docs/Web/API/FormData/FormData#parameters

szmarczak: https://github.com/nodejs/undici/blob/b0ec94d1a4b679b212e441ef332eb1337ad92dbd/lib/fetch/formdata.js#L9-L17

https://github.com/nodejs/undici/blob/b0ec94d1a4b679b212e441ef332eb1337ad92dbd/lib/fetch/util.js#L287

There might be a use case where [`jsdom`](https://github.com/jsdom/jsdom) can be used 🤔

szmarczak: @ronag Should we later change the error message to e.g.

```
Failed to constructor 'FormData': parameter 'form' is unsupported
```

PR #1045, opened 2021-09-22
Update .socket property of SocketError class

delvedor:
The previous implementation didn't work well, as some of the properties of the `Socket` instance were removed while destroying the socket, making the socket instance useless for debugging purposes.
This pr stores the properties before destroying the socket in a separate object as we previously discussed.

Follow up of https://github.com/nodejs/undici/pull/1041

### DIFF
--- a/docs/api/Errors.md
+++ b/docs/api/Errors.md
@@ -23,5 +23,19 @@ import { errors } from 'undici'
 
 ### `SocketError`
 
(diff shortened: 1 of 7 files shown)


mcollina (approved):
lgtm

szmarczak (approved):

ronag:
Perfect. This is how I wanted it.

PR #1048, opened 2021-09-23
types(fetch): make json method generic

iShibi:
This PR makes the `json` method to be generic. This will let TS users provide a return type for the method instead of it just being `unknown`

```ts
import { fetch } from 'undici';

// A shorter version of what GitHub returns
interface Repository {
    id: number;
    node_id: string;
    name: string;
}

const res = await fetch('https://api.github.com/repos/nodejs/undici');
const data = await res.json<Repository>(); // <- We know the shape of 'data' now
```

### DIFF
--- a/types/fetch.d.ts
+++ b/types/fetch.d.ts
@@ -38,7 +38,7 @@ export interface BodyMixin {
   readonly arrayBuffer: () => Promise<ArrayBuffer>
   readonly blob: () => Promise<Blob>
   readonly formData: () => Promise<FormData>
-  readonly json: () => Promise<unknown>
+  readonly json: <T>() => Promise<T>
   readonly text: () => Promise<string>
 }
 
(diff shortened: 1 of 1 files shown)


szmarczak:
This has been discussed and probably won't be implemented: https://github.com/nodejs/undici/pull/997#discussion_r695334060

iShibi:
Hey, I read the comment you linked. So, I assume you want the return type to be `unknown` for compatibility. Which is still the default behaviour here. If a user doesn't provide the type `T`, tsc will show the return type of `json` to still be `unknown`. This is an opt-in change.

```ts
const a = await res.json(); // 'a' is unknown
const b = await res.json<Repository>(); // 'a' is of type Repository
```

szmarczak:
But it works only one way. If you were to switch from `undici` to something else it would break.

Issue #1052, opened 2021-10-04
DiagnosticsChannel Request: missing the full URL


flovilmart:
## This would solve...

I'm attempting to implement tracing through the diagnostics channels, and we'd need to apply some filters so spans do not get started when a new request is created.

The problem is, that the request object, emitted through `'undici:request:create'` do not have any reference to the target host, protocol or full URL for that matter, only the `path` is passed through.

## The implementation should look like...

Not sure, it seems that the scope of the Request object is really tight, and it really looks like the `host`  / `protocol` does not belong there. However, for tracing purposes, we always want this info when the span goes out.

## I have also considered...

We could implement such filter on connect, but it would not be canonical, as filters are usually applied to the full request URL.
Alternatively, we could start tracing on body sent, and keep track of which socket / connect options was used for the request.
 additional book-keeping.

## Additional context
Any other suggestion is welcome as well!

flovilmart:
Closed by #1053

PR #1053, opened 2021-10-05
Store origin on request to facilitate instrumentation

flovilmart:
## This relates to...

Fix for issue #1052 

## Rationale
When instrumenting with APM's it is valuable to have the full origin for the request so we can perform filtering.
This is also required at the time the request is created as this is usually when tracing headers are injected.

## Changes

Adds `origin` member on the `Request` entity for propagation through the diagnostics channels

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -284,7 +284,7 @@ class Client extends Dispatcher {
 
       const origin = opts.origin || this[kUrl].origin
 
-      const request = new Request({ origin, ...opts }, handler)
+      const request = new Request(origin, opts, handler)
 
       this[kQueue].push(request)
       if (this[kResuming]) {
--- a/lib/client.js
+++ b/lib/client.js
@@ -282,7 +282,9 @@ class Client extends Dispatcher {
         handler = new RedirectHandler(this, maxRedirections, opts, handler)
       }
 
-      const request = new Request(opts, handler)
+      const origin = opts.origin || this[kUrl].origin
+
+      const request = new Request({ origin, ...opts }, handler)
 
       this[kQueue].push(request)
       if (this[kResuming]) {
--- a/docs/api/DiagnosticsChannel.md
+++ b/docs/api/DiagnosticsChannel.md
@@ -15,6 +15,7 @@ This message is published when a new outgoing request is created.
 import diagnosticsChannel from 'diagnostics_channel'
 
 diagnosticsChannel.channel('undici:request:create').subscribe(({ request }) => {
+  console.log('origin', request.origin)
   console.log('completed', request.completed)
   console.log('method', request.method)
   console.log('path', request.path)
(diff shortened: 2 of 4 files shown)


mcollina:
This will need some tests.

flovilmart:
@mcollina definitely, the main issue I have now, is with the dichotomy between `undici.request("http://host")` and `new Client().request({ path: "/" })`.

The former will have `origin` in the dispatch call but not the latter.

flovilmart:
@mcollina I think this should do, the client will inject the URL origin if not present.

lib/client.js
```diff
@@ -282,7 +282,9 @@ class Client extends Dispatcher {
         handler = new RedirectHandler(this, maxRedirections, opts, handler)
       }
 
-      const request = new Request(opts, handler)
+      const origin = opts.origin || this[kUrl].origin
+
+      const request = new Request({ origin, ...opts }, handler)
```
mcollina, line 287: I would prefer avoiding to recreating this object. This is a hot path. Maybe add a new positional parameter to request?

flovilmart: yep I was thinking about this as well, would putting it first would make sense?

ronag (approved):

flovilmart:
@ronag , will you squash / merge or do you want me to do it in the PR?

flovilmart:
@mcollina and @ronag , I have a supporting question. When implementing the tracing in datadog, they would also like to have an implementation that isn't backed by the diagnostics channels. 
One way to catch the request creation, would be to wrap [this block](https://github.com/nodejs/undici/blob/main/lib/core/request.js#L138-L140) in a function `onCreate()` that we could patch externally. This would help bring on par both implementations, from the diag channels and the monkey patch.

What do you think? Can I add it in this PR?

mcollina (approved):
lgtm

mcollina:
> What do you think? Can I add it in this PR?

Please create a fresh PR and add a test for it.

I would highly recommend you to not implement the monkey-patching solution.

flovilmart:
@mcollina , agreed!

rochdev:
> When implementing the tracing in datadog, they would also like to have an implementation that isn't backed by the diagnostics channels.

Just to clarify, we should be able to do this without any change to Undici. If we can't, then it means older versions cannot be supported anyway, so no need to add new methods, just adding the diagnostics channels would be enough.

flovilmart:
diagnostic channels are not available on all nodejs version; I think that's the main concern.
For my part, I am very happy with it now, and would not require any additional changes.

rochdev:
There is a [polyfill](https://www.npmjs.com/package/diagnostics_channel) available, and since it has the same name as the built-in module, it should automatically be loaded in older versions of Node as long as it's in the dependencies.

PR #1064, opened 2021-10-16
Implements BalancedPool

mcollina:


ronag:
> > I think destroy and close need to return promsie if not passed callback.
> 
> They already do. What's missing specifically?

You don’t return anything?

mcollina:
> There is room for improvement in regards to the algorithm but this should be good enough to be useful.

I'll open an issue afterwards, the selection algorithm can be improved significantly and it will be a good contribution!

Issue #1065, opened 2021-10-18
Improve BalancedPool balancing algorithm


mcollina:
Currently `BalancedPool` use a simple round robin algorithm if the target upstream is not available.

A more nuanced algorithm could easily produce better  performance.

This is a good paper describing a better algorithm: http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.863.3985&rep=rep1&type=pdf.

jodevsa:
@mcollina what is the reason behind introducing the balancedPool? Is it to support distributing the requests to multiple upstreams?

mcollina:
Yes exactly. There can be a few situations where deploying a separate load balancer in front of a pool of services is not an option.

jodevsa:
@delvedor 

Regarding

>     // we should be able to find the next node in 1 array scan,
>     // if we don't, it means that we are in an infinite loop

https://github.com/elastic/elastic-transport-js/blob/main/src/pool/WeightedConnectionPool.ts#L54

I believe this assumption is wrong. 

Lets say you have 3 servers with the following weights [90,10]:

> GCD =10 
> maxWeight= 90 
> currentWight = maxWeight= 90 

So it will need 8 iterations to go from currentWeight=90 to currentWeight=10. 

This is why the algorithm mentioned in the table uses an infinite loop.

jodevsa:
Hi guys, I finally have a vacation so hopefully I'll be able to wrap this this things 🙏  I was stuck at testing and currently working on rebasing after the new factory thingy changes. 


I need some help. So I've been writing tests and I noticed a behaviour in the algorithm where one server would be chosen for so many times. I'll try to explain it via an example:

Let us assume that we have 2 upstreams A, and B, with initial weights of 50 for both. Also let us assume that our dynamic weights logic would have a penalty of 19 points if the upstream request fails. if B fails the weight of B would be 31 and the weight of A would stay 50. the current weight would be around 50 and the algorithm would decrease that weight every N iterations by the GCD which is "1" in our case GCD(50,31).  So then the current weight is decreased every N iterations (2 in our case) by the GCD. This would result in the algorithm choosing server A for more than 78 times `(((50-11)/1) *2`.

 It would be very helpful if someone else can validate my findings and check if this is an acceptable behaviour. I'm also up for a call if anyone would like to discuss this further :)

mcollina:
I think 78 times is not that bad from my understanding.

[...]

PR #1069, opened 2021-10-20
Use weighted round robin in balancedPool

jodevsa:
#1065

### DIFF
--- a/lib/balanced-pool.js
+++ b/lib/balanced-pool.js
@@ -18,6 +18,17 @@ const { parseOrigin } = require('./core/util')
 const kFactory = Symbol('factory')
 
 const kOptions = Symbol('options')
+const kGreatestCommonDivisor = Symbol('kGreatestCommonDivisor')
+const kCurrentWeight = Symbol('kCurrentWeight')
+const kIndex = Symbol('kIndex')
+const kWeight = Symbol('kWeight')
+const kMaxWeightPerServer = Symbol('kMaxWeightPerServer')
+const kErrorPenalty = Symbol('kErrorPenalty')
+
+function getGreatestCommonDivisor (a, b) {
+  if (b === 0) return a
+  return getGreatestCommonDivisor(b, a % b)
+}
 
 function defaultFactory (origin, opts) {
   return new Pool(origin, opts)
@@ -28,6 +39,11 @@ class BalancedPool extends PoolBase {
     super()
 
     this[kOptions] = opts
+    this[kIndex] = -1
+    this[kCurrentWeight] = 0
+
+    this[kMaxWeightPerServer] = this[kOptions].startingWeightPerServer || 100
+    this[kErrorPenalty] = this[kOptions].errorPenalty || 15
 
     if (!Array.isArray(upstreams)) {
       upstreams = [upstreams]
@@ -42,6 +58,7 @@ class BalancedPool extends PoolBase {
     for (const upstream of upstreams) {
       this.addUpstream(upstream)
     }
+    this._updateBalancedPoolStats()
   }
 
   addUpstream (upstream) {
@@ -54,10 +71,41 @@ class BalancedPool extends PoolBase {
     ))) {
       return this
     }
+    const pool = this[kFactory](upstreamOrigin, Object.assign({}, this[kOptions]))
 
-    this[kAddClient](this[kFactory](upstreamOrigin, Object.assign({}, this[kOptions])))
+    this[kAddClient](pool)
+    pool.on('connect', () => {
+      pool[kWeight] = Math.min(this[kMaxWeightPerServer], pool[kWeight] + this[kErrorPenalty])
+    })
+
+    pool.on('connectionError', () => {
+      pool[kWeight] = Math.max(1, pool[kWeight] - this[kErrorPenalty])
+      this._updateBalancedPoolStats()
+    })
+
+    pool.on('disconnect', (...args) => {
+      const err = args[2]
+      if (err && err.name !== 'ClientDestroyedError' && err.message !== 'socket idle timeout') {
+        // decrease the weight of the pool.
+        pool[kWeight] = Math.max(1, pool[kWeight] - this[kErrorPenalty])
+      } else {
+        // Set the weight of a connection to the maximum value.
+        pool[kWeight] = this[kMaxWeightPerServer]
+      }
+      this._updateBalancedPoolStats()
+    })
+
+    for (const client of this[kClients]) {
+      client[kWeight] = this[kMaxWeightPerServer]
+    }
+
+    this._updateBalancedPoolStats()
 
     return this
+  }
+
+  _updateBalancedPoolStats () {
+    this[kGreatestCommonDivisor] = this[kClients].map(p => p[kWeight]).reduce(getGreatestCommonDivisor, 0)
   }
 
   removeUpstream (upstream) {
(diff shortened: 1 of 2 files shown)


lib/balanced-pool.js
```diff
@@ -7,23 +7,44 @@ const Pool = require('./pool')
...
+    this[kCurrentWeight] = 0
 
     if (!Array.isArray(upstreams)) {
       upstreams = [upstreams]
     }
 
+    this[kRecalculate] = () => {
+      this[kGreatestCommonDivisor] = this[kPools].map(p => p[kWeight]).reduce(getGreatestCommonDivisor, 0)
+      this[kMaxWeight] = Math.max(...this[kPools].map(p => p[kWeight]))
+    }
```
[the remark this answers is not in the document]
jodevsa: I thought this is a pattern or something. Should it be a class method?

lib/balanced-pool.js
```diff
@@ -51,6 +81,11 @@ class BalancedPool extends Dispatcher {
     })
 
     this[kPools].push(pool)
+    this[kPools].forEach(pool => {
+      pool[kWeight] = Math.round(MAX_WEIGHT / (this[kPools].length || 1))
+    })
```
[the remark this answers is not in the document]
jodevsa: sure. I'm also interested in the reason :)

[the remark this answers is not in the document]
jodevsa: `for(let i=0;...)` or `for(const pool of this[kPools])` ?

lib/balanced-pool.js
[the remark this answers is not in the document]
mcollina: I think so in this case

lib/balanced-pool.js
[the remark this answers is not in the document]
mcollina: Either, forEach is an order or magniture (or more) slower than those.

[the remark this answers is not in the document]
jodevsa: thanks :)

jodevsa:
I still have to fix the comments and make them better

lib/balanced-pool.js
```diff
@@ -62,13 +103,48 @@ class BalancedPool extends Dispatcher {
...
+    let counter = 0
+
+    while (counter++ < this[kPools].length) {
+      this[kIndex] = (this[kIndex] + 1) % this[kPools].length
+      const pool = this[kPools][this[kIndex]]
+
+      // Save the index of the pool with the next lower weight
+      // This is needed incase we couldn't find a pool that is >=currentWeight in `this[kpools].length` iterations
+      // This is an optimization. it also avoids using an infinite loop as the algorithim described in the paper.
+      if (pool[kWeight] < this[kCurrentWeight] && (!pool.busy || allPoolsBusy)) {
```
jodevsa, line 122: @ronag @mcollina  I had to use add this logic to avoid using a while(true)

What do you think?

mcollina: pool.busy is not a defined property

mcollina (commented):
Could you test this against a few scenarios:

1. an upstream is unavailable from the start, the algorithm should successfully route requests to working peers.
2. the upstream become online. Request should start be fulfilled by the new upstream.
3. the upstream go offline again. Requests should be routed to other upstreams.

lib/balanced-pool.js
```diff
@@ -41,6 +65,16 @@ class BalancedPool extends Dispatcher {
...
     pool.on('disconnect', (...args) => {
       this.emit('disconnect', ...args)
+
+      const err = args[2]
+      if (err) {
+        // decrease the weight of the pool.
+        pool[kWeight] = Math.max(1, pool[kWeight] - MAX_WEIGHT * 0.01)
+      } else {
+        // Set the weight of a connection to the maximum value.
+        pool[kWeight] = Math.round(MAX_WEIGHT / this[kPools].length)
```
mcollina, line 75: I feel this might be better done on successful connection.

ronag:
Rebase main and you will get kNeedDrain to use instead of busy.

lib/balanced-pool.js
jodevsa: > I feel this might be better done on successful connection.

What if we have a server that drops the connections as soon as we get a connection? 

This is what i'm trying to cover. I'm also simulating that in the tests.


but, I also still have to cover this case: 

* Client is unable to connect to the upstream

mcollina: > > What if the upstream drops the connections as soon as the client connects?

> This is why I preferred gluing the logic on a disconnect event.

I'm ok in decreasing the weight if a connection is closed with an error, however I'm not in resetting the weight if a connection is disconnected successfully. It should either be reset on successful connection or at the first successful request?

ronag:
@jodevsa you still intend to work on this?

jodevsa:
@ronag yeah, but no time right now :( I might be able to work on it on Saturday. Please feel free to take over if you this is blocking you :)

ronag:
> @ronag yeah, but no time right now :( I might be able to work on it on Saturday. Please feel free to take over if this thing is blocking you :)

Not blocking. Just curious.

jodevsa:
@mcollina  Can you have a look and tell me what do you think of the tests before I start cleaning?

test/balanced-pool.js
```diff
@@ -163,3 +163,204 @@ test('busy', (t) => {
     }
   })
 })
+
+const cases = [ 
+  // 0
+  { 
+    iterations: 25, 
+    servers: [{ name: 'A', disabled:[0] }, { name: 'B' }, { name: 'C' }],
```
jodevsa, line 171: @mcollina  disabled[0] means that server A will be disabled only in iteration number 0 and it will be enabled again after 0. I'll have to think of a way to make it understandable or I might add comments

jodevsa: `iterations` is the number of requests.  maybe I should rename that to requests 🤔

jodevsa: @mcollina Can you have a look and tell me what do you think of the tests before I start cleaning?

test/balanced-pool.js
```diff
@@ -163,3 +163,204 @@ test('busy', (t) => {
...
+  // 1
+  { 
+    iterations: 25, servers: [{ name: 'A', fail:[0] }, { name: 'B' }, { name: 'C' }], 
+    expected: ['A', 'B', 'C', 'B', 'C', 'B', 'C', 'B', 'C', 'B', 'C', 'B','C', 'B', 'C', 'B', 'C', 'B','C', 'B', 'C', 'A', 'B', 'C','B'],
+    expectedConnectionRefusedErrors: 0, 
+    expectedRatios:[0.08,0.48,0.44]
+  },
+  // 2
+  { 
+    servers: [{ name: 'A'}, { name: 'B', disabled:true}, { name: 'C' }],
```
jodevsa, line 185: `disabled:true` means that the server will be disabled all the time

jodevsa: I'm adding notes because I have a feeling I made it so complicated 🙈

test/balanced-pool.js
```diff
@@ -163,3 +163,204 @@ test('busy', (t) => {
...
+  { 
+    iterations: 25, 
+    servers: [{ name: 'A', disabled:[0] }, { name: 'B' }, { name: 'C' }], 
+    expected: ['B', 'C', 'B', 'C', 'B', 'C','B', 'C', 'B', 'C', 'B', 'C','B', 'C', 'B', 'C', 'B', 'C','B', 'C', 'A', 'B', 'C', 'A'], 
+    expectedConnectionRefusedErrors: 1, 
+    expectedRatios:[0.08,0.44,0.44]
+  },
+  // 1
+  { 
+    iterations: 25, servers: [{ name: 'A', fail:[0] }, { name: 'B' }, { name: 'C' }],
```
jodevsa, line 178: `fail` means that the server will drop the socket connection

jodevsa:
@mcollina  please don't review yet. I need to make the tests more understanable

lib/balanced-pool.js
```diff
@@ -39,8 +55,33 @@ class BalancedPool extends PoolBase {
...
+
+    pool.on('connectionError', () => {
+      pool[kWeight] = Math.max(1, pool[kWeight] - 10)
+
+      this._updateBalancedPoolStats()
+    })
+
+    pool.on('disconnect', (...args) => {
+      const err = args[2]
+      if (err) {
```
ronag, line 70:
```suggestion
      if (err && err.code !== 'UND_ERR_INFO') {
```

lib/balanced-pool.js
```diff
@@ -73,20 +123,42 @@ class BalancedPool extends PoolBase {
...
+    if (allClientsBusy) {
       return
     }
 
-    this[kClients].splice(this[kClients].indexOf(dispatcher), 1)
-    this[kClients].push(dispatcher)
+    let counter = 0
+
+    let clientWithMaxWeightIndex = this[kClients].findIndex(pool => !pool[kNeedDrain])
+
```
ronag, line 135: allClientsBusy = clientWithMaxWeightIndex !== -1

lib/balanced-pool.js
jodevsa: you just helped me found a weird behaviour I had in the tests :D

lib/balanced-pool.js
```diff
@@ -73,20 +123,42 @@ class BalancedPool extends PoolBase {
...
 
-    if (!dispatcher) {
+    if (allClientsBusy) {
       return
     }
 
-    this[kClients].splice(this[kClients].indexOf(dispatcher), 1)
-    this[kClients].push(dispatcher)
+    let counter = 0
+
```
ronag, line 133: I think you just want the non busy clients for the rest of the logic here, i.e:

```
const clients = this[kClients].filter(pool => !pool[kNeedDrain])
```

jodevsa: i'll loose the offset and mapping of kIndex will be wrong

kibertoad:
Can I help with this one?

mcollina:
Go for it!

kibertoad:
@mcollina are the only remaining items the unaddressed comments and conflicts, or there is something else?

mcollina:
Unsure, this PR is still a draft from the author.

jodevsa:
Hi @mcollina, @ronag, 

This PR is ready for reviewing again 🎉

mcollina (approved):
lgtm

jodevsa:
Hi guys, what are the next steps for getting this merged? are we missing more reviewers?

lib/balanced-pool.js
```diff
@@ -54,12 +71,40 @@ class BalancedPool extends PoolBase {
...
+    })
+
+    pool.on('disconnect', (...args) => {
+      const err = args[2]
+      if (err && err.code === 'UND_ERR_SOCKET') {
+        // decrease the weight of the pool.
+        pool[kWeight] = Math.max(1, pool[kWeight] - this[kErrorPenalty])
+        this._updateBalancedPoolStats()
+      }
+    })
```
ronag, line 93: This event handlers need to be removed in removeUpstream.

jodevsa: good find! I'll open another PR to address your comments

PR #1081, opened 2021-10-31
fix(headers): iterate over cloned list

lukeed:
This matches the browser behavior, example:

```js
let demo = new Headers({ foo: '123', bar: '456' });

for (let [k,v] of demo) { 
  demo.delete(k);
  demo.set(`x-${k}`, v);
}

[...demo];
//=> [
//=>   ['x-bar', '456'],
//=>   ['x-foo', '123'],
//=> ]
```

> **Note:** This is included as a test

### DIFF
--- a/lib/fetch/headers.js
+++ b/lib/fetch/headers.js
@@ -313,20 +313,23 @@ class Headers {
   }
 
   * keys () {
-    for (let index = 0; index < this[kHeadersList].length; index += 2) {
-      yield this[kHeadersList][index]
+    const clone = this[kHeadersList].slice()
+    for (let index = 0; index < clone.length; index += 2) {
+      yield clone[index]
     }
   }
 
   * values () {
-    for (let index = 1; index < this[kHeadersList].length; index += 2) {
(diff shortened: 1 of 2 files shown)


ronag (approved):

ronag:
If you could also find ref to spec that would be great.

lukeed:
From the last sentence in [this section](https://fetch.spec.whatwg.org/#headers-class)

> The value pairs to iterate over are the return value of running [sort and combine](https://fetch.spec.whatwg.org/#concept-header-list-sort-and-combine) with this’s header list.

Where **sort and combine** is an instruction-set for creating a new list.

lukeed:
Adding another test to this PR, which is also failing in `main` branch

mcollina (approved):
lgtm

Ethan-Arrowood (approved):
LGTM!

Ethan-Arrowood:
The infinite iteration test is a good check as well. Totally missed that; I wonder if something like that would be valuable for the web platform tests

Issue #992, opened 2021-08-20
Incorrect `servername` calculation


szmarczak:
#1005

mcollina:
would you send a PR?

deadbeef84:
This is broken since 94ab02df631fd97f45ea144175016684de8f4637

```js
import undici from 'undici'

try {
  await undici.request('https://1.1.1.1/')
} catch (err) {
  console.log(err)
}
```

```
AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:

  assert(this.servername)

    at TLSSocket.<anonymous> (/home/jesper/nxtedition/nxt/app/node_modules/undici/lib/core/connect.js:42:11)
    at TLSSocket.emit (node:events:394:28)
    at TLSWrap.onnewsessionclient [as onnewsession] (node:_tls_wrap:298:11) {
  generatedMessage: true,
  code: 'ERR_ASSERTION',
  actual: false,
  expected: true,
  operator: '=='
}
```

IP addresses may not be valid server names, but the certificate has SAN entries with IP adresses that matches the host IP, which should make the request/cert valid.

```
$ openssl s_client -connect 1.1.1.1:443 < /dev/null | openssl x509 -noout -text | grep -A1 'Subject Alternative Name'
...
           X509v3 Subject Alternative Name: 
                DNS:cloudflare-dns.com, DNS:*.cloudflare-dns.com, DNS:one.one.one.one, IP Address:1.1.1.1, IP Address:1.0.0.1, IP Address:162.159.36.1, IP Address:162.159.46.1, IP Address:2606:4700:4700:0:0:0:0:1111, IP Address:2606:4700:4700:0:0:0:0:1001, IP Address:2606:4700:4700:0:0:0:0:64, IP Address:2606:4700:4700:0:0:0:0:6400
```

I'm not sure how that should work with http/2.

ronag:
@szmarczak any ideas how to fix?

PR #1082, opened 2021-11-03
fix: support no servername

ronag:
Refs: https://github.com/nodejs/undici/issues/992#issuecomment-958358300

### DIFF
--- a/lib/core/connect.js
+++ b/lib/core/connect.js
@@ -24,9 +24,12 @@ function buildConnector ({ maxCachedSessions, socketPath, timeout, ...opts }) {
   return function connect ({ hostname, host, protocol, port, servername }, callback) {
     let socket
     if (protocol === 'https:') {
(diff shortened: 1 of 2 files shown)


ronag:
@deadbeef84  can you confirm whether this solves your problem?

deadbeef84:
Yes, this solves my problem :+1:

mcollina (commented):
This area is a bit sensitive so I would prefer this comes with a test

ronag:
Unsure how to test this at the moment… @deadbeef84 ideas?

mcollina (approved):
lgtm not not really happy about not having tests for these bit. Maybe using some mocks?

deadbeef84:
maybe something similar to https://github.com/nodejs/undici/blob/main/test/https.js but instead of `localhost` you use `127.0.0.1`? I just tested and that reproduces the error.

szmarczak (approved):

PR #1086, opened 2021-11-05
feat: Allow buffers as mock response data

runeh:
With this change, it's possible to return binary payloads from an intercepted call:

```
const mockImage = readFileSync('/path/to/image/png')
mockPool.intercept({ path: '/foo', method: 'GET' }).reply(200, mockImage)
```

I haven't bothered with docs yet. I'll update them if there's interest in merging this.

### DIFF
--- a/docs/api/MockPool.md
+++ b/docs/api/MockPool.md
@@ -62,11 +62,13 @@ Returns: `MockInterceptor` corresponding to the input options.
 
 We can define the behaviour of an intercepted request with the following options.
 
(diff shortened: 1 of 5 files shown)


mcollina (commented):
Amazing, could you add it to the docs?

runeh:
I added some docs. Turned out to be very little, but I think it covers the change reasonably well.

mcollina (approved):
lgtm

PR #1089, opened 2021-11-11
docs: fix links

shkaper:
- swept through local links and fixed broken anchors
- updated local links to relative
    - absolute paths break links when you're browsing docs in the repo, e.g. `https://github.com/nodejs/undici/blob/main/docs/api/docs/api/Client.md`. Absolute paths were apparently added because of docsify, which by default does not recognize relative paths. This can be fixed by setting [`relativePath`](https://docsify.js.org/#/configuration?id=relativepath) to `true` in docsify configs


RafaelGSS (approved):
LGTM.

Issue #1061, opened 2021-10-14
Test errors with Node.js 17


targos:
See CITGM result: https://ci.nodejs.org/view/Node.js-citgm/job/citgm-smoker/2778/nodes=fedora-latest-x64/testReport/junit/(root)/citgm/undici_v4_7_2/

targos:
Some are also due to the OpenSSL update:

```
 # Subtest: Client using valid client certificate
         1..5
         not ok 1 - error:0A00018E:SSL routines::ca md too weak
           ---
           origin:
             at:
               line: 65
               column: 13
               file: node:internal/tls/secure-context
```

mcollina:
I cannot reproduce that locally using -rc.1.

mcollina:
The OpenSSL error looks suspicious and to be investigated. It does not reproduce on the two machines I have available (Mac and Linux)

Issue #1085, opened 2021-11-04
Tests failing under Node.js 17


dominykas:
## Bug Description

Ran tests under Node.js 17, got the following:

```
 FAIL  test/tls-client-cert.js 3 failed of 3 100.353ms
 ✖ error:0A00018E:SSL routines::ca md too weak
 ✖ should be equal
 ✖ test unfinished
```

See also in actions: https://github.com/dominykas/undici/runs/4105824000?check_suite_focus=true#step:10:10005

## Reproducible By

`npm t`

## Expected Behavior

✅ 

## Logs & Screenshots

<details>
<summary>Full stack traces</summary>

```
 FAIL  test/tls-client-cert.js
 ✖ error:0A00018E:SSL routines::ca md too weak

  test/tls-client-cert.js                                   
  52 |       method: 'GET'                                  
  53 |     }, (err, { statusCode, body }) => {              
> 54 |       t.error(err)                                   
     | --------^                                            
  55 |       t.equal(statusCode, 200)                       
  56 |                                                      
  57 |       const authorized = client[kSocket].authorized  

  test: test/tls-client-cert.js Client using valid client certificate
  origin:
    at:
      line: 65
      column: 13
      file: node:internal/tls/secure-context
    stack: |
      Array.forEach (<anonymous>)
      Client.connect (lib/core/connect.js:31:20)
      socket (lib/client.js:1144:25)
      new Promise (<anonymous>)
      connect (lib/client.js:1143:26)
    library: SSL routines
    reason: ca md too weak
    code: ERR_SSL_CA_MD_TOO_WEAK
  found:
    !error
    name: Error
    message: error:0A00018E:SSL routines::ca md too weak
    stack: >-
      Error:
          at node:internal/tls/secure-context:65:13
          at Array.forEach (<anonymous>)
          at setCerts (node:internal/tls/secure-context:63:3)
          at configSecureContext (node:internal/tls/secure-context:152:5)
          at Object.createSecureContext (node:_tls_common:116:3)
          at Object.connect (node:_tls_wrap:1621:48)
          at Client.connect (/Users/dominykas/devel/libs/undici/lib/core/connect.js:31:20)
          at socket (/Users/dominykas/devel/libs/undici/lib/client.js:1144:25)
          at new Promise (<anonymous>)
          at connect (/Users/dominykas/devel/libs/undici/lib/client.js:1143:26)
    library: SSL routines
    reason: ca md too weak
[...]

PR #1091, opened 2021-11-11
fix: add node 17 to test matrix

ronag:
Refs: https://github.com/nodejs/undici/issues/1061


ronag:
@mcollina any ideas why it fails?

szmarczak:
Probably https://github.com/nodejs/undici/issues/1085 I guess the certificates need to use modern TLS features. No idea how to regenerate them.

PR #1092, opened 2021-11-11
Use pkgjs/action for CI matrix

### LINKED ISSUE #1071: Use pkgjs/action for CI to auto-upgrade to latest nodes
(Not following a template - this is more of a chore, not a bug or a feature)

The [CI workflow](https://github.com/nodejs/undici/blob/main/.github/workflows/nodejs.yml) used here is almost identical to a standard Node.js library testing pattern (checkout, `npm install`, `npm test`).

As part of [Package Maintenance WG](https://github.com/nodejs/package-maintenance), I've recently created [`pkgjs/action`](https://github.com/pkgjs/action) - it basically does that standard Node testing pattern, but it automatically maintains the test matrix for you, i.e. it will automatically detect the lowest supported (major) version from `engines` field and will automatically add new Node.js versions as they come out.

As the action is pretty new, I'd like to get some battle testing for it, as we're entering an LTS upgrade season.

Would people mind if I opened a PR to start using it here? I'm willing to take on the maintenance of that until we feel it's stable (although I figure it will be a straight drop-in, with a side effect of starting testing in Node.js 17).

dominykas:
Closes #1071

### DIFF
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -15,7 +15,7 @@ jobs:
       - name: Setup Node
         uses: actions/setup-node@v2
         with:
-          node-version: 16
+          node-version: lts/*
       - name: Install Modules
         run: npm i
       - name: Run Benchmark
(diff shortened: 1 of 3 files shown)


ronag (approved):

mcollina (approved):
lgtm

PR #1097, opened 2021-11-13
increase FormData compatibility

### LINKED ISSUE #1095: formData.keys, .entries is not a function
## Bug Description

When using request.formData() and obtaining a FormData object, both the "entries" and "keys" methods are not found.

## Reproducible By

```
import { FromData } from "undici";
formData.append("key", "value");
console.log("entries" in formData); // false
console.log("keys" in formData); // false
```

## Expected Behavior

console.log("entries" in formData); // true
console.log("keys" in formData); // true

## Environment

`miniflare --version // 2.0.0-next.3`

### Additional context

https://github.com/cloudflare/miniflare/issues/91

KhafraDev:
Fixes #1095 

- add FormData@@toStringTag
- add FormData.entries
- add FormData.keys
- add FormData.values

Typings were already correct & did not need to
be updated. The [FormData Spec](https://xhr.spec.whatwg.org/#interface-formdata) does not include
these methods, but [MDN](https://developer.mozilla.org/en-US/docs/Web/API/FormData#methods) documents them.

### DIFF
--- a/lib/fetch/formdata.js
+++ b/lib/fetch/formdata.js
@@ -168,6 +168,28 @@ class FormData {
     }
   }
 
+  get [Symbol.toStringTag] () {
+    return 'FormData'
+  }
+
+  * entries () {
+    for (const pair of this) {
+      yield pair
+    }
+  }
+
+  * keys () {
+    for (const [key] of this) {
+      yield key
+    }
+  }
+
+  * values () {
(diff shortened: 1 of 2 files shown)


ronag (approved):

mcollina (approved):
lgtm

PR #1099, opened 2021-11-16
feat: accept third-party `Blob`s in `FormData`

### LINKED ISSUE #930: FormData
Support for form data in fetch. 

Do we know of any spec compliant implementation that we can re-use?

wojpawlik:
Works around https://github.com/nodejs/node/issues/37340.
Advances https://github.com/telegraf/telegraf/issues/1502.
Closes #930.

### DIFF
--- a/lib/fetch/util.js
+++ b/lib/fetch/util.js
@@ -2,6 +2,7 @@
 
 const { redirectStatus } = require('./constants')
 const { performance } = require('perf_hooks')
+const { Blob } = require('buffer')
 const nodeUtil = require('util')
 
 let ReadableStream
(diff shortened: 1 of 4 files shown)


lib/fetch/util.js
```diff
@@ -68,6 +68,18 @@ function requestBadPort (request) {
   return 'allowed'
 }
 
+// based on https://github.com/node-fetch/fetch-blob/blob/8ab587d34080de94140b54f07168451e7d0b655e/index.js#L229-L241 (MIT License)
+function isBlobLike (object) {
+  return (
```
mcollina, line 73: Could you add an `instanceof Blob` as a fast path?

mcollina (changes requested):
Good work! Can you please add a unit test?

ronag (approved):

mcollina (approved):
lgtm

PR #1100, opened 2021-11-16
Handle undefined `status` and `statusText` like Web fetch

### LINKED ISSUE #1094: Inconsistency with undefined `status` (and maybe `statusText`)
## Bug Description
In the `ResponseInit` body one should be able to pass `status: undefined` to have behavior akin to the browser API, in which case the status should become 200 instead of throwing an error

## Reproducible By
```
import { Response } from "undici";
new Response("", { status: undefined }); // throws TypeError
```
## Expected Behavior
No error, and the status is 200 for the `Response`.

gzuidhof:
Fixes #1094, ~but I'm not 100% sure if deleting the statusText is the nicest way to deal with this as we're mutating the input object.~ EDIT: made simpler and better now :)

### DIFF
--- a/examples/proxy-agent.js
+++ b/examples/proxy-agent.js
@@ -4,7 +4,7 @@ const { request, setGlobalDispatcher, ProxyAgent } = require('../')
 
 setGlobalDispatcher(new ProxyAgent('http://localhost:8000/'))
 
-async function main() {
+async function main () {
   const {
     statusCode,
     headers,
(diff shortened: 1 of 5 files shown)


mcollina (commented):

gzuidhof:
Done, added tests for this behavior :)

mcollina (approved):
lgtm

PR #1109, opened 2021-11-21
fix: don't stringify third-party `Blob`s

wojpawlik:


mcollina (commented):

wojpawlik:
How? [`65997b3` (#1099)](https://github.com/nodejs/undici/pull/1099/commits/65997b31dbda92ade2ab7d553277022f97bd5670) somehow didn't catch this... Anyway, https://github.com/node-fetch/node-fetch/issues/1205#issuecomment-915500645

ronag (changes requested):
I think the fix should be in isBlobLike

lib/fetch/formdata.js
```diff
@@ -213,7 +213,7 @@ function makeEntry (name, value, filename) {
 
   // 3. If value is a Blob object and not a File object, then set value to a new File
   // object, representing the same bytes, whose name attribute value is "blob".
-  if (isBlobLike(value) && !(value instanceof File)) {
+  if (isBlobLike(value) && value[Symbol.toStringTag] !== 'File') {
```
ronag, line 216:
```suggestion
  if (isBlobLike(value)) {
```

lib/fetch/formdata.js
```diff
@@ -224,7 +224,7 @@ function makeEntry (name, value, filename) {
   // Do we just override the name attribute? Since it says "if value is (now)"
   // does that mean that this lives inside the previous condition? In which case
   // creating one more File instance doesn't make much sense....
-  if (value instanceof File && filename) {
+  if (value[Symbol.toStringTag] === 'File' && filename) {
```
ronag, line 227:
```suggestion
  if (isFileLike(value) && filename) {
```

ronag: `isFileLike´ doesn't exist yet.

ronag:
> How? [`65997b3` (#1099)](https://github.com/nodejs/undici/pull/1099/commits/65997b31dbda92ade2ab7d553277022f97bd5670) somehow didn't catch this... Anyway, [node-fetch/node-fetch#1205 (comment)](https://github.com/node-fetch/node-fetch/issues/1205#issuecomment-915500645)

Where/How did you encounter an issue that needs this fix?

wojpawlik:
https://github.com/telegraf/client/commit/6e123259585c0a5bddb04ff0a5f9bc27c6b2f989

ronag:
Create a test that uses `formdata-polyfill`

ronag:
Do you have a repro for the issue you are trying to solve?

lib/fetch/formdata.js
ronag, line 216:
```suggestion
  if (isBlobLike(value) && !isFileLike(value)) {
```

wojpawlik:
> Do you have a repro for the issue you are trying to solve?

This uses https://github.com/telegraf/client/commit/6e123259585c0a5bddb04ff0a5f9bc27c6b2f989. Sadly I cannot minimize it further, because I don't know how `fetch` serializes `FormData`.

```js
// @filename: tst.cjs
"use strict";

const { Client, StreamFile } = require("@telegraf/client");
const fs = require("fs");

const client = new Client(process.env.BOT_TOKEN);

const document = new StreamFile(
  () => fs.createReadStream("tst.cjs"),
  "tst.cjs",
);

client.call("sendDocument", { chat_id: 197767148, document });
```

image

ronag:
It's because this doesn't count as a "isBlobLike"

```js
// https://github.com/telegraf/client/blob/main/src/stream-file.ts
export class StreamFile {
  readonly size = NaN;
  constructor(
    readonly stream: () =>
      | AsyncIterable<Uint8Array>
      | ReadableStream<Uint8Array>,
    readonly name: string,
  ) {}
}

Object.defineProperty(StreamFile.prototype, Symbol.toStringTag, {
  value: "File",
});
```

ronag:
```
iff --git a/lib/fetch/body.js b/lib/fetch/body.js
index 7d8edb4..de4089d 100644
--- a/lib/fetch/body.js
+++ b/lib/fetch/body.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const util = require('../core/util')
-const { ReadableStreamFrom, toUSVString } = require('./util')
+const { ReadableStreamFrom, toUSVString, isBlobLike } = require('./util')
 const { FormData } = require('./formdata')
 const { kState } = require('./symbols')
 const { Blob } = require('buffer')
@@ -121,7 +121,7 @@ function extractBody (object, keepalive = false) {
     // followed by the multipart/form-data boundary string generated
     // by the multipart/form-data encoding algorithm.
     contentType = 'multipart/form-data; boundary=' + boundary
-  } else if (object instanceof Blob) {
+  } else if (isBlobLike(object)) {
     // Blob
 
     // Set action to this step: read object.
diff --git a/lib/fetch/util.js b/lib/fetch/util.js
index e3ba281..80d59a2 100644
--- a/lib/fetch/util.js
+++ b/lib/fetch/util.js
@@ -74,9 +74,7 @@ function isBlobLike (object) {
   return object instanceof Blob || (
     object &&
     typeof object === 'object' &&
-    typeof object.constructor === 'function' &&
-    (typeof object.stream === 'function' ||
-      typeof object.arrayBuffer === 'function') &&
+    (typeof object.stream === 'function') &&
     /^(Blob|File)$/.test(object[Symbol.toStringTag])
   )
 }
```

PR #1113, opened 2021-11-27
fix: cleanup signal listener

ronag:


mcollina (commented):

lib/fetch/request.js
```diff
@@ -328,14 +332,9 @@ class Request {
...
-        // TODO: Remove this listener on failure/success.
-        signal.addEventListener(
-          'abort',
-          function () {
-            ac.abort()
-          },
-          { once: true }
-        )
+        const abort = () => ac.abort()
+        signal.addEventListener( 'abort', abort, { once: true })
```
ronag, line 336:
```suggestion
        signal.addEventListener('abort', abort, { once: true })
```

mcollina (approved):
lgtm

ronag:
@benjamingr @dominic not the prettiest solution... any suggestions? The spec doesn't specify how and when the abort listener should be cleaned up.

benjamingr:
Remove the abort listener when you know you are not going to abort (when the request finishes, was aborted, errored etc).

FinalizationRegistry is a really heavy handed tool for this

lib/fetch/request.js
```diff
@@ -328,14 +332,9 @@ class Request {
...
-        signal.addEventListener(
-          'abort',
-          function () {
-            ac.abort()
-          },
-          { once: true }
-        )
+        const abort = () => ac.abort()
+        signal.addEventListener('abort', abort, { once: true })
+        requestFinalizer.register(this, { signal, abort })
```
mcollina, line 337: We should be unregistering this after successful completion. Always passing through the registry is expensive.

ronag:
> Remove the abort listener when you know you are not going to abort (when the request finishes, was aborted, errored etc).

It's a little more complicated 😄. The spec doesn't have any request finalisation so I would have to make up something myself.

benjamingr:
Oh that's a DOM fetch request? Yeah, I think cheating is probably fine the browser cheats by doing this internally

lib/fetch/request.js
ronag: Yea, but if we could do that we wouldn't need this hack at all.

ronag:
@mcollina PTAL 82e1a5180a8b850c5d1c11a6653278d792241584

PR #1114, opened 2021-11-27
fix: balanced pool

ronag:
Unified pool and balanced-pool to use the same scheduling implementation. There is still room for improvement but this is a good first step.


mcollina (approved):
lgtm

PR #1118, opened 2021-12-01
feat: BodyReadable.dump

ronag:


ronag:
missing test

docs/api/Dispatcher.md
```diff
@@ -436,6 +436,10 @@ The `RequestOptions.method` property should not be value `'CONNECT'`.
 - `body`
 - `bodyUsed`
 
+`body` contains the following additional extentions:
+
+- `dump({ limit: Integer })`, dump the response by reading up to `limit` bytes without killing the socket.
```
mcollina, line 441: Please include the default. I think we could define in in bytes, so use a multiple of 2

Issue #1119, opened 2021-12-02
Encoding issue


srenault:
## Bug Description

For a single request, which as always the same response, I'm having an encoding issue that appears only from time to time. 
In the response, I'm having this character � instead of special characters like `é` or `à` etc...
The encoding issue doesn't appear for all special characters in the response.
Over the time, it's not always the same characters that are impacted by the encoding issue.

## Reproducible By

I used this simple gist to reproduce the issue:
https://gist.github.com/srenault/654c0179277a29b1c798461b20d5ec44

## Logs & Screenshots

<img width="326" alt="Screenshot 2021-12-02 at 20 43 58" src="https://user-images.githubusercontent.com/1011289/144491911-e6a2186f-12d7-42ce-bab9-00663badc973.png">

## Environment

Node v14.16.0

### Additional context

If I reset the `globalDispatcher` with a `new Agent()` before each call of `request`, I don't have the issue.

koh110:
We have same issue.

I think this problem is due to the lack of consideration for multibyte characters.

This is my reproduction code.

```javascript
import { createServer } from "http";
import { Client } from "undici";
import { once } from "events";
import assert from "assert";

const data = "あ".repeat(10000000);

const server = createServer((request, response) => {
  response.end(JSON.stringify({ data }));
}).listen();

await once(server, "listening");

const client = new Client(`http://localhost:${server.address().port}`);

try {
  const { body, headers, statusCode } = await client.request({
    path: "/",
    method: "GET",
  });
  console.log(`response received ${statusCode}`);
  console.log("headers", headers);
  // body.setEncoding('utf8')
  const json = await body.json();
  assert.strictEqual(json.data, data, "multi byte");
} catch (error) {
  console.error(Object.keys(error), error.code);
} finally {
  client.close();
  server.close();
}
```

It looks like the bug will be fixed if you fix this.

```diff
diff --git a/lib/api/readable.js b/lib/api/readable.js
index 0580f42..3bb454e 100644
--- a/lib/api/readable.js
+++ b/lib/api/readable.js
@@ -219,9 +219,9 @@ function consumeEnd (consume) {

   try {
     if (type === 'text') {
-      resolve(body.join(''))
+      resolve(Buffer.concat(body))
     } else if (type === 'json') {
-      resolve(JSON.parse(body.join('')))
+      resolve(JSON.parse(Buffer.concat(body)))
     } else if (type === 'arrayBuffer') {
       const dst = new Uint8Array(length)
```

test code

[...]

PR #1120, opened 2021-12-03
fix: concat multibyte characters

koh110:
move from `Array.join()` to `Buffer.concat()` to fix bugs in multibyte characters.

Fixes: https://github.com/nodejs/undici/issues/1119


ronag:
Thanks!

koh110:
@ronag Thanks for your super quick review!

PR #1124, opened 2021-12-06
feat: body mixin request response

ronag:
Just an idea to further improve developer ergonomics.

### DIFF
--- a/lib/api/api-request.js
+++ b/lib/api/api-request.js
@@ -9,6 +9,64 @@ const util = require('../core/util')
 const { AsyncResource } = require('async_hooks')
 const { addSignal, removeSignal } = require('./abort-signal')
 
+class Response {
+  constructor({
+    statusCode,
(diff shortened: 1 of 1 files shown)


lib/api/api-request.js
```diff
@@ -9,6 +9,64 @@ const util = require('../core/util')
...
+    opaque,
+    body,
+    context
+  }) {
+    this.statusCode = statusCode
+    this.headers = headers
+    this.trailers = trailers
+    this.opaque = opaque
+    this.body = body
+    this.context = context
```
ronag, line 26: Should be under getters in server major.

ronag:
This is very breaking... but would nice... @mcollina wdyt?

ronag:
Maybe we could put this under a different method name? `client.something`. It's kind of a fetch like API but with node semantics and better perf.

PR #1127, opened 2021-12-07
fix: expose socketPath in Client.Options type

punkeel:
The `Client` constructor accepts a `socketPath` argument,
but this was not reflected in the typescript definition.


mcollina (commented):
Can you please add a unit test? We use tsd for the types

ronag (changes requested):
That option is kind of deprecated and only remain for compat reasons. You have the same option available under the `connect` option.

PR #1129, opened 2021-12-07
fix: readable.setEncoding

ronag:
Fixes: https://github.com/nodejs/undici/issues/1125#issuecomment-987724800

PR #1133, opened 2021-12-08
fix: encode

koh110:
# Contexts
https://github.com/nodejs/undici/pull/1129

### DIFF
--- a/lib/api/readable.js
+++ b/lib/api/readable.js
@@ -228,13 +228,30 @@ function consumeStart (consume) {
 }
 
 function consumeEnd (consume) {
-  const { type, body, resolve, stream, length } = consume
+  const { type, resolve, stream } = consume
+
+  let length = consume.length
+  let body = consume.body
+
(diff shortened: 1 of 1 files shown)


ronag (changes requested):
This won't work unfortunately. The stream must keep functioning as a normal stream as well.

lib/api/readable.js
```diff
@@ -258,11 +275,11 @@ function consumeEnd (consume) {
   }
 }
 
-function consumePush (consume, chunk, encoding) {
-  if (encoding) {
-    chunk = Buffer.from(chunk, encoding)
+function consumePush (consume, chunk) {
+  if (consume.stream.readableEncoding) {
+    chunk = consume.stream._readableState.decoder.write(chunk)
```
ronag, line 280: This will mutate stream state. Which we can't do.

koh110:
Thank you for your review. This problem is very complicated...

PR #1134, opened 2021-12-08
fix: errored stream is disturbed

ronag:


test/node-fetch/main.js
```diff
@@ -882,7 +882,7 @@ describe('node-fetch', () => {
           return expect(res.text())
             .to.eventually.be.rejected
             .and.be.an.instanceof(Error)
-            .and.have.property('name', 'AbortError')
+            .and.have.property('name', 'TypeError')
```
ronag, line 885: @dominic I'm actually unsure what the spec expects here. Would you mind help clarifying?

ronag:
@szmarczak This is a spec thing. Care to help clarifying what is the correct behaviour.

ronag:
Screenshot 2021-12-08 at 21 03 30

ronag:
@jimmywarting: Any thoughts? I believe node-fetch fails with an `AbortError` in this test.

jimmywarting:
Umm. is there any wpt test or something in the spec that mention what should actually be happening?

lib/fetch/index.js
```diff
@@ -1742,8 +1742,9 @@ function httpNetworkFetch (
             controller.enqueue(new Uint8Array(bytes))
 
             // 8. If stream is errored, then terminate the ongoing fetch.
-            if (stream[kError]) {
-              this.context.terminate({ reason: stream[kError] })
+            // TODO (perf): This check is slow,
+            if (/state: 'errored'/.test(nodeUtil.inspect(stream))) {
+              this.context.terminate()
```
mcollina, line 1747: Maybe create an issue for this?

ronag: Where? I think I already created one for @jasnell in node core. But I don't think this is something we can fix unless undici is in core.

ronag: https://github.com/nodejs/node/pull/41121

szmarczak:
Sorry for my inactivity, had a few tests and an exam in uni. Will take a look now.

lib/core/util.js
```diff
@@ -244,13 +244,32 @@ function validateHandler (handler, method, upgrade) {
...
+
   const state = body && body._readableState
-  return !!(body && (
-    (stream.isDisturbed && stream.isDisturbed(body)) ||
+  return !!(
     body[kBodyUsed] ||
     body.readableDidRead || (state && state.dataEmitted) ||
-    isAborted(body)
-  ))
+    isReadableAborted(body)
```
szmarczak, line 259: This could be `isDestroyed`, no? Is there a reason why we check `endEmitted`?

test/node-fetch/main.js
szmarczak: It should be `AbortError`.

https://fetch.spec.whatwg.org/#fetch-method -> https://fetch.spec.whatwg.org/#abort-fetch

szmarczak:
@jimmywarting https://github.com/whatwg/streams/issues/1025 tl;dr unfortunately it's up to Node.js, because they made the mistake to make fetch access stream internals.

mcollina (commented):

lib/core/util.js
ronag: Yes. If it is destroyed before emitting end then and only the is it aborted.

ronag:

There is already unit test?

test/node-fetch/main.js
ronag: > If response’s body is not null and is readable, then error response’s body with error.

This does not say how that error should propagate through e.g. `.text()`

ronag: erroring a stream makes it disturbed and the reading algorithm says to throw a TypeError if disturbed.

ronag: @szmarczak I think this is the relevant part: https://fetch.spec.whatwg.org/#concept-body-consume-body

ronag: Remember it's not fetch that throws here, we already have resolved the promise, it's the body mixing methods that throw here. wdyt?

mcollina:
I missed it. CI is failing.

mcollina (approved):
lgtm

test/node-fetch/main.js
szmarczak: Makes sense, I guess you're right.

```js
const start = Date.now();
const response = await fetch('https://speed.hetzner.de/100MB.bin', {signal: controller.signal});
console.log(Date.now() - start, response.bodyUsed);
(async () => {
    try {
        await response.text();
        console.log(Date.now() - start, 'read');
    } catch (error) {
        console.error(Date.now() - start, error);
    }
})();
console.log(Date.now() - start, 'aborting');
controller.abort();
```

Google Chrome: `DOMException: The user aborted a request.`
Firefox: `DOMException: The operation was aborted.`

szmarczak (approved):
Indeed, it should be `TypeError` 👍🏼

Issue #1137, opened 2021-12-10
Unexpected `ERR_INVALID_STATE` error


szmarczak:
```js
import {fetch} from 'undici';

const start = Date.now();
const controller = new AbortController();
const response = await fetch('https://speed.hetzner.de/100MB.bin', {signal: controller.signal});
console.log(Date.now() - start, response.bodyUsed);
(async () => {
    try {
        await response.text();
        console.log(Date.now() - start, 'read');
    } catch (error) {
        console.error(Date.now() - start, error.name);
    }
})();
console.log(Date.now() - start, 'aborting');
controller.abort();
```

```
szm@pop-os:~/Desktop/demo$ node asdf.mjs 
(node:11537) ExperimentalWarning: stream/web is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
293 false
295 aborting
302 AbortError
node:internal/errors:464
    ErrorCaptureStackTrace(err);
    ^

TypeError [ERR_INVALID_STATE]: Invalid state: ReadableStream is locked
    at new NodeError (node:internal/errors:371:5)
    at ReadableStream.cancel (node:internal/webstreams/readablestream:263:9)
    at cancelBody (/home/szm/Desktop/demo/node_modules/undici/lib/fetch/body.js:357:17)
    at EventEmitter.abortFetch (/home/szm/Desktop/demo/node_modules/undici/lib/fetch/index.js:291:5)
    at EventTarget.requestObject.signal.addEventListener.once (/home/szm/Desktop/demo/node_modules/undici/lib/fetch/index.js:143:18)
    at EventTarget.[nodejs.internal.kHybridDispatch] (node:internal/event_target:562:20)
    at EventTarget.dispatchEvent (node:internal/event_target:504:26)
    at abortSignal (node:internal/abort_controller:115:10)
    at AbortController.abort (node:internal/abort_controller:146:5)
    at EventTarget.Request.signal.addEventListener.once (/home/szm/Desktop/demo/node_modules/undici/lib/fetch/request.js:335:16) {
  code: 'ERR_INVALID_STATE'
}

Node.js v17.2.0
```

```
szm@pop-os:~/Desktop/demo$ npm ls undici
demo@1.0.0 /home/szm/Desktop/demo
└── undici@4.11.3
```

/cc @ronag

ronag:
checking

PR #1138, opened 2021-12-10
fix: don't cancel with invalid state

ronag:
Refs: https://github.com/nodejs/undici/issues/1137

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -1605,7 +1605,7 @@ function httpNetworkFetch (
       } else {
         // 4. Otherwise, if stream is readable, error stream with a TypeError.
         if (isReadable(stream)) {
-          this.controller.error(new TypeError('terminated'))
+          this.controller.error(this.terminated.error || new TypeError('terminated'))
         }
       }
 
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -34,7 +34,7 @@ const {
 const { kState, kHeaders, kGuard, kRealm } = require('./symbols')
 const { AbortError } = require('../core/errors')
 const assert = require('assert')
-const { safelyExtractBody, cancelBody } = require('./body')
+const { safelyExtractBody } = require('./body')
 const {
   redirectStatus,
   nullBodyStatus,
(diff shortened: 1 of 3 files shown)


lib/fetch/index.js
```diff
@@ -274,8 +274,8 @@ function abortFetch (p, request, responseObject) {
 
   // 3. If request’s body is not null and is readable, then cancel request’s
   // body with error.
-  if (request.body != null) {
-    cancelBody(request.body, error)
+  if (request.body != null && isReadable(request.body?.stream)) {
+    request.body.stream.cancel(error)
```
ronag, line 278: @szmarczak: We have another spec problem here... the stream might be locked at which point we are not allowed the call cancel... even though the stream is readable... ideas?

ronag: @jasnell I think actually there might be a bug in node's implementation. It doesn't say anything about throwing invalid state if the stream is locked when cancelling? https://streams.spec.whatwg.org/#readable-stream-cancel

ronag:
```js
✖ Invalid state: ReadableStream is locked

  test: TAP
  at:
    line: 371
    column: 5
    file: node:internal/errors
    constructor: true
    function: NodeError
  stack: >
    EventEmitter.abortFetch (lib/fetch/index.js:292:26)

    EventTarget.requestObject.signal.addEventListener.once (lib/fetch/index.js:144:18)

    EventTarget.Request.signal.addEventListener.once (lib/fetch/request.js:335:16)
  type: TypeError
  code: ERR_INVALID_STATE
  tapCaught: unhandledRejection
```

ronag: @szmarczak also... the spec doesn't say what to do if the promise returned from cancel rejects...

szmarczak: > the stream might be locked at which point we are not allowed the call cancel

Locking just means it's "piped" to a reader. Cancellation still should be possible.

```js
const response = await fetch('https://speed.hetzner.de/100MB.bin', {signal: controller.signal});
const reader = response.body.getReader();
console.log(response.body.locked);
controller.abort();
console.log(await reader.read()); // GoogleChrome - DOMException: The user aborted a request.
```

szmarczak: > the spec doesn't say what to do if the promise returned from cancel rejects

I think it never rejects.

Issue #1141, opened 2021-12-13
Spec changes


ronag:
Spec changes that we have to apply:

- [ ] https://github.com/whatwg/fetch/commit/ed6221d33338d1046574e81fb9d18f4fc2a35600
- [ ] https://github.com/whatwg/fetch/commit/8e4553bcc338fd6db98cc6ba60fc03cea48fad4c
- [ ] https://github.com/whatwg/fetch/commit/02b2e8efdebfdb2e4d34b47587bf7b4d2d45791b
- [ ] https://github.com/whatwg/fetch/commit/8659a25e2dcad911a372f74ac70b60607a264e42

ronag:
@KhafraDev which one of the above do we need to check off?

KhafraDev:
@ronag 2nd (https://github.com/whatwg/fetch/commit/8e4553bcc338fd6db98cc6ba60fc03cea48fad4c) and 4th (https://github.com/whatwg/fetch/commit/8659a25e2dcad911a372f74ac70b60607a264e42)

PR #1142, opened 2021-12-13
update request.js to match new spec changes

KhafraDev:
Implements spec changes from https://github.com/whatwg/fetch/commit/8e4553bcc338fd6db98cc6ba60fc03cea48fad4c and also fixes a few spec divergences (I'll comment on each one to make it easier)

refs: #1141

### DIFF
--- a/lib/fetch/request.js
+++ b/lib/fetch/request.js
@@ -143,6 +143,7 @@ class Request {
       // method request’s method. 
       method: request.method,
       // header list A copy of request’s header list.
+      // undici implementation note: headersList is cloned in makeRequest
       headersList: request.headersList,
       // unsafe-request flag Set. 
       unsafeRequest: request.unsafeRequest,
--- a/lib/fetch/request.js
+++ b/lib/fetch/request.js
@@ -138,8 +138,47 @@ class Request {
 
     // 12. Set request to a new request with the following properties:
     request = makeRequest({
-      ...request,
-      window
+      // URL request’s URL.
+      // undici implementation note: this is set as the first item in request's urlList in makeRequest
+      // method request’s method. 
+      method: request.method,
+      // header list A copy of request’s header list.
+      headersList: request.headersList,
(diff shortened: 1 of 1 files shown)


lib/fetch/request.js
```diff
@@ -138,8 +138,47 @@ class Request {
 
     // 12. Set request to a new request with the following properties:
     request = makeRequest({
```
KhafraDev, line 140: by listing all the properties, I assumed there are some that aren't meant to be copied to the new request. It makes sense as previously origin and urlList were not supposed to be copied, but were being done by the lib.

lib/fetch/request.js
```diff
@@ -138,8 +138,47 @@ class Request {
...
-      ...request,
-      window
+      // URL request’s URL.
```
KhafraDev, line 141: this was previously defined as the [current url](https://fetch.spec.whatwg.org/#concept-request-current-url) which is the last url in the urlList, but the recent spec changes technically fixed this on behalf of undici :)

ronag (approved):

ronag:
Good work!

lib/fetch/request.js
```diff
@@ -138,8 +138,47 @@ class Request {
...
+      // undici implementation note: this is set as the first item in request's urlList in makeRequest
+      // method request’s method. 
+      method: request.method,
+      // header list A copy of request’s header list.
+      headersList: request.headersList,
```
ronag, line 146: Is this also cloned in makeRequest?

KhafraDev: it is https://github.com/nodejs/undici/blob/main/lib/fetch/request.js#L753-L755

ronag: Maybe add a comment like you did below for urlList?

ronag:
@KhafraDev excellent work, if you see any other spec discrepancies feel free to open PR!

PR #1143, opened 2021-12-13
update timingInfo to follow new spec changes

KhafraDev:
Implements spec changes from https://github.com/whatwg/fetch/commit/8659a25e2dcad911a372f74ac70b60607a264e42 

refs: #1141 (this didn't take me very long because it was mostly just renaming the utility function)

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -24,7 +24,7 @@ const {
   requestCurrentURL,
   setRequestReferrerPolicyOnRedirect,
   tryUpgradeRequestToAPotentiallyTrustworthyURL,
-  makeTimingInfo,
+  createOpaqueTimingInfo,
   appendFetchMetadata,
   corsCheck,
   crossOriginResourcePolicyCheck,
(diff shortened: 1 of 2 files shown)


lib/fetch/util.js
```diff
@@ -253,20 +253,20 @@ function coarsenedSharedCurrentTime (crossOriginIsolatedCapability) {
...
-    postRedirectStartTime: 0,
+    postRedirectStartTime: timingInfo.startTime ?? 0,
     finalServiceWorkerStartTime: 0,
     finalNetworkResponseStartTime: 0,
     finalNetworkRequestStartTime: 0,
     endTime: 0,
     encodedBodySize: 0,
     decodedBodySize: 0,
-    finalConnectionTimingInfo: null,
-    ...init
```
KhafraDev, line 269: I removed this because the spec states to "return a new fetch timing info whose start time and post-redirect start time are timingInfo’s start time.". Since only startTime and postRedirectStartTime can be changed it seems more explicit to show that their values are the only ones being updated.

lib/fetch/index.js
```diff
@@ -339,9 +337,8 @@ function fetching ({
   // post-redirect start time are the coarsened shared current time given
   // crossOriginIsolatedCapability.
   const currenTime = coarsenedSharedCurrentTime(crossOriginIsolatedCapability)
-  const timingInfo = makeTimingInfo({
-    startTime: currenTime,
-    postRedirectStartTime: currenTime
+  const timingInfo = createOpaqueTimingInfo({
```
KhafraDev, line 340: I'm not sure if this falls under the category of an opaque timing info but it follows the guidelines of having default values and having startTime = postRedirectStartTime

mcollina:

KhafraDev:

I don't know how the following code would be triggered in this package, or at all? I looked for wpt tests and found [this](https://github.com/web-platform-tests/wpt/issues/30968) open issue that relies upon [this](https://github.com/whatwg/fetch/commit/ed6221d33338d1046574e81fb9d18f4fc2a35600) spec change to trigger. So we need that 2nd spec change (which adds timingInfo to network errors (and even then, would undici ever trigger this condition?).

mcollina:
@ronag wdyt?

ronag:
I'm ok with merging. It's literally follow the spec and for now it's not observable. I was hoping to eventually make timing info observable through e.g. diagnostics channels.

mcollina (approved):
lgtm

PR #1147, opened 2021-12-16
Create timing info for HTTP(S) network errors

KhafraDev:
refs: https://github.com/nodejs/undici/issues/1141

Implements spec changes from: https://github.com/whatwg/fetch/commit/ed6221d33338d1046574e81fb9d18f4fc2a35600 
However this should be mostly good, possibly apart from the
TransformStream parts since I never used the api before this...

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -808,9 +808,8 @@ function fetchFinale (fetchParams, response) {
     }
 
     // 2. Let processBodyError be this step: run fetchParams’s process response consume body given response and failure.
-    const processBodyError = () => {
-      // TODO (spec): what is failure?
-      fetchParams.processResponseConsumeBody(response, new TypeError())
+    const processBodyError = (failure) => {
+      fetchParams.processResponseConsumeBody(response, failure)
     }
 
     // 3. If response’s body is null, then queue a fetch task to run processBody given null, with fetchParams’s task destination.
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -50,6 +50,7 @@ const { isErrored, isReadable } = require('../core/util')
 const { kIsMockActive } = require('../mock/mock-symbols')
 
 let ReadableStream
+let TransformStream
 
 class Fetch extends EE {
   constructor (dispatcher) {
@@ -740,34 +741,87 @@ function finalizeResponse (fetchParams, response) {
 
 // https://fetch.spec.whatwg.org/#fetch-finale
 function fetchFinale (fetchParams, response) {
-  const context = this
+  // 1. If response is a network error, then:
+  if (response.type === 'error') {
+    // 1. Set response’s URL list to « fetchParams’s request’s URL list[0] ».
+    response.urlList = [fetchParams.request.urlList[0]]
 
-  // 1. If fetchParams’s process response is non-null,
-  // then queue a fetch task to run fetchParams’s process response
-  // given response, with fetchParams’s task destination.
+    // 2. Set response’s timing info to the result of creating an opaque timing info for fetchParams’s timing info.
+    response.timingInfo = createOpaqueTimingInfo({
+      startTime: fetchParams.timingInfo.startTime
+    })
+  }
+
+  // 2. Let processResponseEndOfBody be the following steps:
+  const processResponseEndOfBody = () => {
+    // 1. Set fetchParams’s request’s done flag.
+    fetchParams.request.done = true
+
+    // 2. If fetchParams’s process response end-of-body is not null, then queue a fetch task to run fetchParams’s
+    //    process response end-of-body given response with fetchParams’s task destination.
+    if (fetchParams.processResponseEndOfBody != null) {
+      // TODO (spec): what exactly are we supposed to do with fetchParams.taskDestination?
+      fetchParams.processResponseEndOfBody(response)
+    }
+  }
+
+  // 3. If fetchParams’s process response is non-null, then queue a fetch task to run fetchParams’s process response given response,
+  //    with fetchParams’s task destination.
   if (fetchParams.processResponse != null) {
     fetchParams.processResponse(response)
   }
 
-  // 2. If fetchParams’s process response consume is non-null, then:.
-  //    TODO
-  //    1. Let processBody given nullOrBytes be this step: run fetchParams’s
-  //    process response consume given response and nullOrBytes.on.
-  //    TODO
-  //    2. Let processBodyError be this step: run fetchParams’s process
-  //    response consume given response and failure.on.
-  //    TODO
-  //    3. If response’s body is null, then queue a fetch task to run
-  //    processBody given null, with fetchParams’s task destination.on.
-  //    TODO
-  //    4. Otherwise, fully read response’s body given processBody,
-  //    processBodyError, and fetchParams’s task destination.on.
-  //    TODO
+  // 4. If response’s body is null, then run processResponseEndOfBody.
+  if (response.body == null) {
+    processResponseEndOfBody()
+  }
 
-  // TODO (spec): The spec doesn't specify this but we need to
-  // terminate fetch if we have an error response.
-  if (response.type === 'error') {
-    context.terminate({ reason: response.error })
+  // 5. Otherwise:
+  else {
+    TransformStream ??= require('stream/web').TransformStream
+    // 1. Let transformStream be a new a TransformStream.
+
+    // 2. Let identityTransformAlgorithm be an algorithm which, given chunk, enqueues chunk in transformStream.
+    const identityTransformAlgorithm = (chunk, controller) => {
+      controller.enqueue(chunk)
+    }
+
+    const transformStream = new TransformStream({
+      // 3. Set up transformStream with transformAlgorithm set to identityTransformAlgorithm and
+      //    flushAlgorithm set to processResponseEndOfBody.
+      transform: identityTransformAlgorithm,
+      flush: processResponseEndOfBody
+    })
+
+    // 4. Set response’s body to the result of piping response’s body through transformStream.
+    // Khafra note: although the spec is unclear, only readable streams can be piped through
+    if (response.body != null && isReadable(response.body?.stream)) {
+      response.body = response.body.pipeThrough(transformStream)
+    }
+  }
+
+  // 6. If fetchParams’s process response consume body is non-null, then:
+  if (fetchParams.processResponseConsumeBody != null) {
+    // 1. Let processBody given nullOrBytes be this step: run fetchParams’s process response consume body given response and nullOrBytes.
+    const processBody = (nullOrBytes) => {
+      fetchParams.processResponseConsumeBody(response, nullOrBytes)
+    }
+
+    // 2. Let processBodyError be this step: run fetchParams’s process response consume body given response and failure.
+    const processBodyError = () => {
+      // TODO (spec): what is failure?
+      fetchParams.processResponseConsumeBody(response, new TypeError())
+    }
+
+    // 3. If response’s body is null, then queue a fetch task to run processBody given null, with fetchParams’s task destination.
+    if (response.body == null) {
+      processBody(null)
+    }
+
+    // 4. Otherwise, fully read response’s body given processBody, processBodyError, and fetchParams’s task destination.
+    else {
+      fullyReadBody(processBody, processBodyError, fetchParams.taskDestination)
+    }
   }
 }
 
@@ -1589,11 +1643,9 @@ function httpNetworkFetch (
     //            1. If one or more bytes have been transmitted from response’s
     //            message body, then:
     //            NOTE: See onHeaders
-    //            2. Otherwise, if the bytes transmission for response’s message
+    //            2. Otherwise, if the bytes transmission for response’s message 
     //            body is done normally and stream is readable, then close stream,
-    //            finalize response for fetchParams and response, and abort these
-    //            in-parallel steps.
-    //            NOTE: See onHeaders
+    //            and abort these in-parallel steps.
 
     //    2. If aborted, then:
     function onResponseAborted () {
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -820,7 +820,7 @@ function fetchFinale (fetchParams, response) {
 
     // 4. Otherwise, fully read response’s body given processBody, processBodyError, and fetchParams’s task destination.
     else {
-      fullyReadBody(processBody, processBodyError, fetchParams.taskDestination)
+      fullyReadBody(response.body, processBody, processBodyError, fetchParams.taskDestination)
     }
   }
 }
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -688,7 +688,7 @@ async function mainFetch (fetchParams, recursive = false) {
     // 1. Let processBodyError be this step: run fetch finale given fetchParams
     // and a network error.
     const processBodyError = (reason) =>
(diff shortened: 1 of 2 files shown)


lib/fetch/index.js
```diff
@@ -727,34 +740,87 @@ function finalizeResponse (fetchParams, response) {
...
+
+  // 6. If fetchParams’s process response consume body is non-null, then:
+  if (fetchParams.processResponseConsumeBody != null) {
+    // 1. Let processBody given nullOrBytes be this step: run fetchParams’s process response consume body given response and nullOrBytes.
+    const processBody = (nullOrBytes) => {
+      fetchParams.processResponseConsumeBody(response, nullOrBytes)
+    }
+
+    // 2. Let processBodyError be this step: run fetchParams’s process response consume body given response and failure.
+    const processBodyError = () => {
```
ronag, line 810:
```suggestion
    const processBodyError = (error) => {
```

ronag (approved):

lib/fetch/index.js
ronag: Maybe?

mcollina (changes requested):
I'm still uncomfortable in landing changes without tests. My understanding is that these are observable.

lib/fetch/util.js
```diff
@@ -297,6 +297,18 @@ function tryUpgradeRequestToAPotentiallyTrustworthyURL (request) {
...
+// https://fetch.spec.whatwg.org/#body-fully-read
+function fullyReadBody (
+  // given an algorithm processBody
+  processBody,
+  // an algorithm processBodyError,
+  processBodyError,
+  // and an optional null, parallel queue, or global object taskDestination (default null)
+  taskDestination = null
+) {
+  // TODO
```
ronag, line 309: You need to either consume the response or terminate fetch here or we will have stuck connections.

ronag (changes requested):

lib/fetch/index.js
KhafraDev: I thought of doing that but the step above said "Let processBody given nullOrBytes" and this one doesn't name the params. Should I add a failure param anyways?

KhafraDev:
> I'm still uncomfortable in landing changes without tests. My understanding is that these are observable.

That's fair, I'll convert to a draft until I can add tests or adapt from the wpt

lib/fetch/index.js
ronag: >  process response consume body **given response and failure**

lib/fetch/util.js
KhafraDev: This should be handled now in 64c9984b43dd662eb405da9f21d2e3271a7d4192 although I had to deviate from the spec slightly

lib/fetch/index.js
KhafraDev: Checking back, this seems to be a mis-wording with the spec and you are correct - "processBodyError must be an algorithm accepting an exception." 👍

KhafraDev:
Would writing tests for the utility functions that were added and tests for the new changes be good enough for the scope of this pr? I'm not exactly sure how we would check the timing info on a network error because a network error typically causes a promise rejection.

ronag:
I'm not much of a fan of testing internal utilities in a way that is not accessible to users...

ronag:
@mcollina What do we do here? I understand you want tests but I also understand the difficulty of making those.

lib/fetch/util.js
```diff
@@ -297,6 +298,101 @@ function tryUpgradeRequestToAPotentiallyTrustworthyURL (request) {
...
+    }
+
+    // 2. Append the bytes represented by chunk to bytes.
+    bytes.push(chunk)
+
+    // 3. Read-loop given reader, bytes, and promise.
+    // This recursion could potentially cause a stack overflow if implemented directly. Implementations will need to mitigate this, 
+    // e.g. by using a non-recursive variant of this algorithm, or queuing a microtask, or using a more direct method of
+    // byte-reading as noted below. 
+    await new Promise(res => queueMicrotask(res))
```
mcollina, line 345: Why this is implemented in this way? I don't think this does what's written above. This will not break the promise chain. I would just remove this statement.

lib/fetch/util.js
```diff
@@ -297,6 +298,101 @@ function tryUpgradeRequestToAPotentiallyTrustworthyURL (request) {
...
+    res = resolve
+    rej = reject
+  })
+
+  return { promise, resolve: res, reject: rej }
+}
+
+// https://streams.spec.whatwg.org/#readablestreamdefaultreader-read-all-bytes
+/** @param {ReadableStreamDefaultReader} reader */
+async function ReadableStreamDefaultReaderReadAllBytes (reader) {
```
mcollina, line 314: This utility can be easily tested as a unit test.

lib/fetch/util.js
```diff
@@ -297,6 +298,101 @@ function tryUpgradeRequestToAPotentiallyTrustworthyURL (request) {
...
+  // 1. Let reader be the result of getting a reader for body’s stream. If that threw an exception,
+  //    then return a promise rejected with that exception.
+  const reader = body.stream.getReader()
+
+  // 2. Return the result of reading all bytes from reader.
+  return await ReadableStreamDefaultReaderReadAllBytes(reader)
+}
+
+// https://fetch.spec.whatwg.org/#body-fully-read
+function fullyReadBody (
```
mcollina, line 365: This utility can be easily tested as a unit test

lib/fetch/util.js
```diff
@@ -297,6 +298,101 @@ function tryUpgradeRequestToAPotentiallyTrustworthyURL (request) {
...
+  // given an algorithm processBody
+  processBody,
+  // an algorithm processBodyError,
+  processBodyError,
+  // and an optional null, parallel queue, or global object taskDestination (default null)
+  taskDestination = null
+) {
+  // 1. If taskDestination is null, then set taskDestination to the result of starting a new parallel queue.
+  if (taskDestination == null) {
+    // TODO
```
mcollina, line 377: This must be filled

lib/fetch/util.js
```diff
@@ -297,6 +298,101 @@ function tryUpgradeRequestToAPotentiallyTrustworthyURL (request) {
...
+    // e.g. by using a non-recursive variant of this algorithm, or queuing a microtask, or using a more direct method of
+    // byte-reading as noted below. 
+    await new Promise(res => queueMicrotask(res))
+    return readLoop()
+  }
+
+  await readLoop()
+
+  return promise
+}
```
mcollina, line 352: From the look of it this function is going to be pretty inefficient as it allocates an enormous amount of promises.

lib/fetch/index.js
```diff
@@ -727,34 +741,86 @@ function finalizeResponse (fetchParams, response) {
...
+    // 3. If response’s body is null, then queue a fetch task to run processBody given null, with fetchParams’s task destination.
+    if (response.body == null) {
+      processBody(null)
+    }
+
+    // 4. Otherwise, fully read response’s body given processBody, processBodyError, and fetchParams’s task destination.
+    else {
+      fullyReadBody(response.body, processBody, processBodyError, fetchParams.taskDestination)
+    }
   }
```
mcollina, line 824: These sets of steps are complex and involve a lot of behavior, tests are needed

mcollina (commented):
I'm not asking for 100% code coverage but this PR adds quite a bit of new code, complex code that should be tested to avoid regressions.

lib/fetch/util.js
ronag: True, but I don't think in the normal case there should be any data to be read here at all.

lib/fetch/util.js
ronag: Why? It's not fully spec compliant per se but I don't think it makes much of a difference for the end user.

lib/fetch/util.js
ronag: I don't think we can get a stackoverflow here anyway since we are using an await loop. So yes, just remove this line.

ronag: You could even implement `readLoop` simply as a `for await` loop.

ronag:
@KhafraDev I landed the simple parts of your PR. We just need to sort out some of the more complicated parts and make some tests for those.

lib/fetch/util.js
mcollina: I would highly prefer it than having this recursion.

lib/fetch/index.js
```diff
@@ -740,34 +742,86 @@ function finalizeResponse (fetchParams, response) {
...
+    }
+
+    // 3. If response’s body is null, then queue a fetch task to run processBody given null, with fetchParams’s task destination.
+    if (response.body == null) {
+      processBody(null)
+    }
+
+    // 4. Otherwise, fully read response’s body given processBody, processBodyError, and fetchParams’s task destination.
+    else {
+      fullyReadBody(response.body, processBody, processBodyError, fetchParams.taskDestination)
```
ronag, line 823:
```suggestion
  await response.body.arrayBuffer().then(processBody, processBodyError)
```

I think this might be sufficient?

lib/fetch/index.js
```diff
@@ -740,34 +742,86 @@ function finalizeResponse (fetchParams, response) {
...
+
+    const transformStream = new TransformStream({
+      // 3. Set up transformStream with transformAlgorithm set to identityTransformAlgorithm and
+      //    flushAlgorithm set to processResponseEndOfBody.
+      transform: identityTransformAlgorithm,
+      flush: processResponseEndOfBody
+    })
+
+    // 4. Set response’s body to the result of piping response’s body through transformStream.
+    // Khafra note: although the spec is unclear, only readable streams can be piped through
```
ronag, line 798: Is there a case where the stream is not readable here? @dominic

lib/fetch/util.js
mcollina: Then we should just remove the if.

lib/fetch/util.js
KhafraDev: I tried implementing word-for-word of what the spec says, so not all the solutions are the best IMO. But yes, I believe a for await does follow the spec rules, so I'll get on it!

lib/fetch/util.js
KhafraDev: task destinations aren't used elsewhere in the code and I didn't quite understand the spec's wording for them and the library, being in js, isn't really needed here to my understanding.

KhafraDev: https://html.spec.whatwg.org/multipage/infrastructure.html#parallel-queue

My issue with implementing a parallel queue is that it seemed quite... useless and the wording was a little strange. If we queue algorithms (functions) then we're just running them back to back, which works regardless in this library. Plus, the spec doesn't note *when* to start dequeuing so my understanding is that it's instant.

lib/fetch/util.js
KhafraDev: I agree, but unless we deviate from the spec, there isn't much else to do :(

lib/fetch/index.js
KhafraDev: If we do this, I can remove fullyReadBody, fullyReadBodyAsPromise, and ReadableStreamDefaultReaderReadAllBytes. The down side is that we aren't following the spec verbatim.

Is this a fair trade off? @mcollina @ronag

lib/fetch/index.js
KhafraDev: I don't believe so, but when I was working on this pr I was getting a lot of test failures so I added the readable check just in case. Removing it, the tests seem to pass so I have gone ahead and removed it in def5a6be2e2b6a6d7767c260d7ceab666ccc3201

lib/fetch/util.js
KhafraDev: The if has been removed in 09e0e798547ca4760363368c10c2fcdfb8f6fc74. Parallel queues are probably better left for another pr :)

lib/fetch/util.js
KhafraDev: A better read-loop algorithm has been added in 1b73431acf0f20a35681df9e16ca6a0affefd7aa

lib/fetch/index.js
KhafraDev: Done in bd4642f2bb78131c25f5e21ec72b314faae09bf6!

ronag:
can you rebase main?

lib/fetch/index.js
```diff
@@ -740,34 +741,91 @@ function finalizeResponse (fetchParams, response) {
...
+    if (response.body != null) {
+      response.body = response.body.pipeThrough(transformStream)
+    }
+  }
+
+  // 6. If fetchParams’s process response consume body is non-null, then:
+  if (fetchParams.processResponseConsumeBody != null) {
+    // 1. Let processBody given nullOrBytes be this step: run fetchParams’s process response consume body given response and nullOrBytes.
+    const processBody = (nullOrBytes) => {
```
ronag, line 806: Should bytes be a Buffer or an array of Buffer?

lib/fetch/util.js
```diff
@@ -339,5 +339,6 @@ module.exports = {
   responseLocationURL,
   isBlobLike,
   isFileLike,
-  isValidReasonPhrase
+  isValidReasonPhrase,
+  createDeferredPromise
```
ronag, line 343: You don't need this anymore

lib/fetch/index.js
KhafraDev: `If given, processResponseConsumeBody must be an algorithm accepting a response and null, failure, or a byte sequence.`

`A byte sequence is a sequence of bytes, represented as a space-separated sequence of bytes. Byte sequences with bytes in the range 0x20 (SP) to 0x7E (~), inclusive, can alternately be written as a string, but using backticks instead of quotation marks, to avoid confusion with an actual string. `

According to the spec it should just be a string

szmarczak: Not necessarily. A byte sequence is anything that's used to store bytes. Can be a Buffer, Uint8Array as well.

KhafraDev: Good to know, gotta love how the spec words things 😄

lib/fetch/index.js
```diff
@@ -740,34 +741,91 @@ function finalizeResponse (fetchParams, response) {
...
+
+    // 4. Otherwise, fully read response’s body given processBody, processBodyError, and fetchParams’s task destination.
+    else {
+      // The discussion to use the arrayBuffer method over the spec's fully read body implementation
+      // can be found here https://github.com/nodejs/undici/pull/1147#discussion_r775847886
+      response.body.arrayBuffer().then(
+        processBody,
+        processBodyError
+      )
+    }
```
ronag, line 828: We need to terminate fetch here or we will leak the connection:

```
else {
  // TODO (spec): The spec doesn't specify this but we need to
  // terminate fetch if we have an error response.
  if (response.type === 'error') {
    context.terminate({ reason: response.error })
}
```

KhafraDev: This should be resolved now 👍

ronag: How?

KhafraDev: a4c9a1a8277041a8a6c9018b07c5fba61ecb658c

ronag: Ah. That's fine.

ronag (approved):

KhafraDev:
It's honestly fine if it doesn't land, 4 of the changes have already landed so this PR is really just a shell of its former 😆.

ronag:
@mcollina I guess we close this then since we can't land it without test and no one wants to build them.

PR #1149, opened 2021-12-17
fix(fetch): integration with MockAgent

gdorsi:
fixes #1121

PR #1150, opened 2021-12-20
fix: corrects a syntax error in node 12

gdorsi:


mcollina (commented):
I'm not sure how you can use this with Node v12, fetch should be available only on Node v16.

gdorsi:
Ah ok, so that's my fault.

I shouldn't have required the fetch code [here](https://github.com/nodejs/undici/pull/1149/files#diff-c4f217e7477ce1df4b03f4b49686fed580dd178618caba17e97bb75cbdff4ecaR4)

I'll close this

PR #1158, opened 2022-01-03
fetch: fix small spec inconsistency

KhafraDev:
This condition is not *yet* possible to meet so I couldn't add a test
(sorry!). 

- `response` is always null https://github.com/nodejs/undici/blob/09059fb491b4158a25981eb5598262b43a18c6ae/lib/fetch/index.js#L475
- which means it is set here https://github.com/nodejs/undici/blob/09059fb491b4158a25981eb5598262b43a18c6ae/lib/fetch/index.js#L529 which always either returns a network error or sets the `responseTainting` to either `opaque` or `cors`.
- Then this condition runs if there wasn't a network error (`response.status === 0` would be a network error) https://github.com/nodejs/undici/blob/09059fb491b4158a25981eb5598262b43a18c6ae/lib/fetch/index.js#L605
- Therefore responseTainting is never basic as of now.

The spec says that "[a] basic filtered response is a filtered response
whose type is "basic" and header list excludes any headers in internal
response’s header list whose name is a forbidden response-header name."
The library was incorrectly excluding valid headers.

If `data:`, `blob:`, `about:`, or `file:` URIs are ever supported, this
change will be needed.

### DIFF
--- a/lib/fetch/constants.js
+++ b/lib/fetch/constants.js
@@ -58,6 +58,7 @@ const requestCache = [
   'only-if-cached'
 ]
 
+// https://fetch.spec.whatwg.org/#forbidden-response-header-name
 const forbiddenResponseHeaderNames = ['set-cookie', 'set-cookie2']
 
 const requestBodyHeader = [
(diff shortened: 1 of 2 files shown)


ronag:
Thanks!

PR #1165, opened 2022-01-11
fetch: add data: & about: uri support

KhafraDev:
Closes #1164

PR #1166, opened 2022-01-12
fetch: multiple small fixes

KhafraDev:
This PR contains 3 fetch bug fixes:

- In `mainFetch`, `response` was not being set in 2 places ("then set response to a network error.").
- In [`httpRedirectFetch`](https://fetch.spec.whatwg.org/#http-redirect-fetch), the old step 12 has been removed
- The terms "same origin" actually refers to a [series of steps](https://html.spec.whatwg.org/multipage/origin.html#same-origin) that need to be taken, not just comparing the URL's origin. ([see here](https://github.com/nodejs/undici/pull/1165#issuecomment-1011317842))

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -965,17 +965,7 @@ async function httpRedirectFetch (fetchParams, response) {
     return makeNetworkError()
   }
 
-  // 12. If locationURL’s origin is not same origin with request’s current URL’s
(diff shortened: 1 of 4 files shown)


ronag (approved):
Excellent work!

mcollina (approved):
lgtm

PR #1175, opened 2022-01-25
 fix: method is not required for request api

snyamathi:
DispatchOptions passed to the request api doesn't **NEED** to have a `method` param since it defaults to `'GET'` (or `'PUT'` if there is an optional body).

https://github.com/nodejs/undici/blob/6c2c6addcced5284144e876bce1b827717a03e47/index.js#L83-L88

```ts

// Doesn't have method param
const options = {};

// method defaults to "GET"
request('http://example.com', options).then(console.log)
```

Why is this important? If `method` is required it makes it awkward to use a default object that does't specify the method rather than letting it get computed automatically later on.

> Property 'method' is missing in type '{}' but required in type

```ts
const foo = (url, options = {}) => {
    // do stuff with options
    return request(url, options);
};
```

### DIFF
--- a/test/types/api.test-d.ts
+++ b/test/types/api.test-d.ts
@@ -4,6 +4,7 @@ import { Dispatcher, request, stream, pipeline, connect, upgrade } from '../..'
 
 // request
 expectAssignable<Promise<Dispatcher.ResponseData>>(request(''))
+expectAssignable<Promise<Dispatcher.ResponseData>>(request('', { }))
 expectAssignable<Promise<Dispatcher.ResponseData>>(request('', { method: 'GET' }))
 
 // stream
--- a/types/api.d.ts
+++ b/types/api.d.ts
@@ -13,7 +13,7 @@ export {
 /** Performs an HTTP request. */
 declare function request(
   url: string | URL | UrlObject,
-  options?: { dispatcher?: Dispatcher } & Omit<Dispatcher.RequestOptions, 'origin' | 'path'>,
+  options?: { dispatcher?: Dispatcher } & Omit<Dispatcher.RequestOptions, 'origin' | 'path' | 'method'> & Partial<Pick<Dispatcher.RequestOptions, 'method'>>,
 ): Promise<Dispatcher.ResponseData>;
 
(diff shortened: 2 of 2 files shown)


types/dispatcher.d.ts
```diff
@@ -40,7 +40,8 @@ declare namespace Dispatcher {
   export interface DispatchOptions {
     origin?: string | URL;
     path: string;
-    method: HttpMethod;
+    /** Default: `'GET'` */
```
snyamathi, line 43:
```suggestion
    /** Default: `body ? 'PUT' : 'GET'` */
```

Alternate option

ronag:
Actually it is required. It’s the global methods that don’t require it.

snyamathi:
@ronag Got it, yeah I do see in the documentation [here](https://github.com/nodejs/undici/blob/3b702b3f6d39b439491df8b949858051bb732a1f/docs/api/Dispatcher.md#parameter-dispatchoptions) that method is required. I suppose it's just the [makeDispatcher](https://github.com/nodejs/undici/blob/5d524231ff0fcb8dac761fd710fa08d7d61b1c26/index.js#L48) helper which is providing a default for us with the exported functions, `request`, `stream`, etc.

We could potentially make `method` optional for those methods:

```diff
 /** Performs an HTTP request. */
 declare function request(
   url: string | URL | UrlObject,
-  options?: { dispatcher?: Dispatcher } & Omit<Dispatcher.RequestOptions, 'origin' | 'path'>,
+  options?: { dispatcher?: Dispatcher } & Omit<Dispatcher.RequestOptions, 'origin' | 'path' | 'method'> & Partial<Pick<Dispatcher.RequestOptions, 'method'>>,
 ): Promise<Dispatcher.ResponseData>;
```

Though the typescript gymnastics required to do so is slightly awkward

mcollina (commented):
Can you please add a unit test? We use tsd.

snyamathi:
Updated with a test (which previously fails and passes w/ the change).

ronag (approved):

mcollina (approved):
lgtm

PR #1177, opened 2022-01-25
fix: fetch - don't coerce null or undefined integrity to a string

### LINKED ISSUE #1176: Fetch's request.integrity is coerced to the string "undefined" when undefined is explicitly passed
## Bug Description

The following code in `makeRequest` coerces `undefined` to the string `"undefined"` when it's explicitly passed to the fetch options:
```
    // 23. If init["integrity"] exists, then set request’s integrity metadata to it.
    if ('integrity' in init) {
      request.integrity = String(init.integrity)
    }
```

## Reproducible By

```
fetch('https://github.com', { integrity: undefined })
```

## Expected Behavior

Undefined is not coerced to a string

## Environment

Arch Linux, Node 16, server-side call (not coming from a browser)

### Additional context

I ran into this while implementing a custom `fetcher` function for Apollo Client. Because the typings for browser/native fetch don't match unidci's fetch, I had to manually map some parameters over. It's pretty easy to workaround this by testing for a falsy value and passing an empty string if needed (`{ integrity: req.integrity || ''}`), but it did result in a pretty cryptic error message:
```
Unknown server error    Error: fetch failed
```

In my case, I believe my need to map stuff manually between the two implementations is actually related to https://github.com/nodejs/undici/issues/1172.

jimsimon:
Fixes #1176

### DIFF
--- a/lib/fetch/request.js
+++ b/lib/fetch/request.js
@@ -140,35 +140,35 @@ class Request {
     request = makeRequest({
       // URL request’s URL.
       // undici implementation note: this is set as the first item in request's urlList in makeRequest
-      // method request’s method. 
+      // method request’s method.
       method: request.method,
       // header list A copy of request’s header list.
       // undici implementation note: headersList is cloned in makeRequest
       headersList: request.headersList,
-      // unsafe-request flag Set. 
+      // unsafe-request flag Set.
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

ronag (approved):

PR #1183, opened 2022-01-29
feat: add build for porting fetch to Node.js

targos:
In the discussion we had during the [mini-summit](https://github.com/nodejs/next-10/issues/108), it was decided that we would try to include `undici`'s `fetch` implementation in Node.js 18.

I just started experimenting with it and I figured it would be easier to discuss in a pull request, where we can directly try some things.

The `esbuild` output seems fine (I can push it here, but wanted to avoid too big of a diff), but we obviously have a problem with llhttp/WebAssembly.
Before I go too far in the wrong direction, what should we do about it? I can think of two solutions:
- Adapt the part of `undici` that uses the llhttp parser to conditionally use the [internal node binding](https://github.com/nodejs/node/blob/master/src/node_http_parser.cc). This is not trivial, as currently the native binding doesn't expose the exact same API that `undici` uses.
- Inline the WASM code, like it was done for [cjs-module-lexer](https://github.com/nodejs/node/tree/master/deps/cjs-module-lexer). This seems risky because WebAssembly is not guaranteed to be available in a Node.js build.

/cc @nodejs/undici


targos:
> Adapt the part of undici that uses the llhttp parser to conditionally use the internal node binding

If we do that, it is possible to test it in userland. Just run `node --expose-internals -r internal/test/binding` and `internalBinding('http_parser')` is available.

mcollina:
Why are we bundling it? Wouldn't it be simpler to just add it to `deps/` in core?

The best approach would be to use the node core bindings for http_parser. However I'm 100% happy to inline the wasm for now. I would require we use the internal binding for `fetch` to leave experimental.

targos:
> Why are we bundling it? Wouldn't it be simpler to just add it to deps/ in core?

core doesn't support relative `require` calls. For example, `const Client = require('./lib/client')` would have to be replaced with `const Client = require('deps/undici/lib/client')`.
So we need so kind of source transformation anyway, and bundling was just easier for me to experiment.

mcollina:
Ah, that's true. Bundling is perfect.

targos:
> However I'm 100% happy to inline the wasm for now.

I see that there are two different wasm builds. Do we need to bundle both?

ronag:
I think the wasm version might be faster though? Can’t we look into getting that working somehow?

targos:
Should I look into inlining the WASM code for everyone or only for node core and keep the current behavior for the npm module
?

ronag:
I think future wise, it would be nice if Node core used a WASM build of llhttp instead of the native one.

It would be nice to keep the current behavior for the npm module.

mcollina:
> > However I'm 100% happy to inline the wasm for now.
> 
> 
> 
> I see that there are two different wasm builds. Do we need to bundle both?

I think you can just bundle the SIMD version. However we might need both for some architecture or in case of a backport. I'm not sure if the wasm simd is available everywhere.

ronag:
@targos thank you for your effort on this!

targos:
Here's a Node PR based on this: https://github.com/nodejs/node/pull/41749

targos:
So, the current state is that both WASM builds are serialized to base64 strings which are exported from two CJS modules (58kb each). I did not keep the current behavior for the npm module because I'm not sure yet how to do it.

ronag:
I’m fine with this

mcollina (approved):
lgtm

mcollina (approved):
lgtm

merceyz:
> Should I look into inlining the WASM code for everyone or only for node core and keep the current behavior for the npm module?

Inlining it for everyone would make it easier for other projects that needs to bundle their dependencies.
I experimented with using Undici in Yarn and had to bundle it using [babel-plugin-static-fs](https://github.com/Jam3/babel-plugin-static-fs) which stopped working with the latest version of Undici so I'll have to manually inline it and check the bundle into our repo.

ronag:
Is this ready to merge? @targos

package.json
```diff
@@ -40,6 +40,7 @@
     "docs"
   ],
   "scripts": {
+    "build:node": "esbuild index.js --bundle --platform=node --outfile=out.js",
```
targos, line 43:
```suggestion
    "build:node": "esbuild index.js --bundle --platform=node --outfile=undici.js",
```

targos:
If it's fine for you, yes!

gr2m:
Thank you. This is amazing. Thanks to everyone who made this happen 💖

miladfarca:
Hello,
I just wanted to bring up a few items regarding Wasm and Simd execution.

> I think you can just bundle the SIMD version.

Simd support in V8 has just been recently added to some architectures such as PPC and S390 and even then they are only supported on new hardware (i.e. PPC 9 and above). Running Simd code on older node versions or older hardware will just crash. It would be better to include both versions and fall back to regular Wasm code if Simd isn't supported.

I can see this module is is compiled with wasi-sdk: https://github.com/nodejs/llhttp/pull/93
wasi-sdk doesn't seem to generate its own Javascript files to glue/call Wasm binaries and instead Js files need to be hand written per module. This could potentially cause endianness issues on big endian platforms like S390/AIX if `Typed Arrays` get used.

emscripten on the other hand does generate its own Js files and it's patched to handle endianness problems when dealing with Typed arrays: https://github.com/emscripten-core/emscripten/pull/13413

We are still working around some remaining problems related to its usage on BE machines but in general emscripten might be safer to use.

ronag:
>  It would be better to include both versions and fall back to regular Wasm code if Simd isn't supported.

We already do this?

> This could potentially cause endianness issues on big endian platforms like S390/AIX if Typed Arrays get used.

Do you have a concrete example?

miladfarca:
> We already do this?

Thanks for confirming.

> Do you have a concrete example?

Wasm memory is LE even on BE machines. Typed Arrays however follow the machine byte order. 
If a Typed Array uses Wasm memory buffer it could potently read garbage data or corrupt the data on BE machines as it doesn't enforce LE ordering:
https://github.com/nodejs/llhttp/blob/main/examples/wasm.ts#L197

emscripten has been patched (link in my previous comment) to use `DataViews` instead of TypedArrays which have an option to force LE ordering.

If emscripten is used then all the generated Js files should also be safe to run on BE. If Js files are being written manually then the developer should make sure Typed Arrays wont cause an endianness problem.

ronag:
Wouldn't it be easier to just fix the js code in the few places where the endianess is relevant?

ronag:
In the case of undici I don't think it's that complicated?

miladfarca:
Doesn't seems like its too complex. If Typed Arrays such as this: https://github.com/nodejs/llhttp/blob/main/examples/wasm.ts#L197
Are only accessing byte sized data then it should be safe to use.

Kikobeats:
@mcollina yay https://github.com/nodejs/undici/pull/1219

PR #1186, opened 2022-01-30
lazy-load tls

benjamingr:


mcollina (approved):
lgtm

lib/core/connect.js
```diff
@@ -24,6 +24,9 @@ function buildConnector ({ maxCachedSessions, socketPath, timeout, ...opts }) {
   return function connect ({ hostname, host, protocol, port, servername }, callback) {
     let socket
     if (protocol === 'https:') {
+      if (!tls) {
+        tls = require('tls')
```
ronag, line 28: Do we need a try/catch with `callback(err)`? @benjamingr

benjamingr: Probably not? I think "tls is missing" is an OK synchronous error in case someone tries to make an HTTPS request with tls missing?

However you are _way_ more familiar with undici than me so I'll defer

mcollina: I think this is ✔️

PR #1187, opened 2022-01-30
tree-shake node bundle

Mesteery:
Since Node.js only exposes these, we can make tree-sharking more efficient by exporting only what Node.js exposes. This currently saves about 54kb (381kb vs 327.4kb).
I don't know if it's worth it, but we have nothing to lose.

cc @targos


mcollina:
Wow, good finding! I think this is premature optimization now.

ronag (approved):

Mesteery:
> I think this is premature optimization now.

I totally agree. But I think the difference will never be significantly greater. So we might as well do it now and scrape off a few kb and a few milliseconds for the startup, anyway we have nothing to lose IMO.

ronag:
If I would guess. Everything under `lib/api`.

index-fetch.js
```diff
@@ -0,0 +1,16 @@
...
+
+const globalDispatcher = new Agent()
+
+let fetchImpl = null
+module.exports.fetch = async function fetch (resource, init) {
+  if (!fetchImpl) {
+    fetchImpl = require('./lib/fetch')
+  }
+  return fetchImpl.call(globalDispatcher, resource, init)
+}
```
targos, line 13: Is it necessary to lazy-load `fetchImpl` ?

Also, the small change I did in https://github.com/nodejs/undici/commit/3e267ece5f3a6b9b4cd6562f885a6884e3926937#diff-e727e4bdf3657fd1d798edcd6b099d6e092f8573cba266154583a746bba0f346 should be ported.

mcollina (approved):
lgtm

However I think we'll soon have to ship full undici to core (for the mocking capabilities).

PR #1189, opened 2022-01-30
fetch: allow file:// URL & relative paths

KhafraDev:
This PR enables fetching relative URLs and file:// URLs. This hopefully fixes issues with WPTs!

Relative URLs are typically handled in browsers because of the environment setting object's base URL. For examples:
```js
// base = https://fetch.spec.whatwg.org/some/path/ [notice trailing /]
fetch('./test'); // fetches https://fetch.spec.whatwg.org/some/path/test

// base = https://fetch.spec.whatwg.org/some/path [no trailing /]
fetch('./test'); // fetches https://fetch.spec.whatwg.org/some/test
```
However, in node, there isn't really a good alternative to base URL as there is no concept of location/origin. Therefore, the 'best' solution (as far as I could think of) was to just use `file://` uris instead and having the "base" be the current working directory.

cc. @targos, hi 😄 

- No tests yet because it needs some discussion.
- Currently allows some "unintended" behavior such as being able to fetch any file, even outside of the cwd.
- There is no spec for `file://` urls so once it is standardized, this will need to be updated.

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -52,10 +52,14 @@ const { PassThrough, pipeline } = require('stream')
 const { isErrored, isReadable } = require('../core/util')
 const { kIsMockActive } = require('../mock/mock-symbols')
 const { dataURLProcessor } = require('./dataURL')
+const { readFile, stat } = require('fs/promises')
+const { emitWarning } = require('process')
 
 /** @type {import('buffer').resolveObjectURL} */
 let resolveObjectURL
 let ReadableStream
+
+let fileURLExperimentalWarning = false
 
 class Fetch extends EE {
   constructor (dispatcher) {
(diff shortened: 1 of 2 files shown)


mcollina (commented):
As for origin, I would prefer we add some way to set the base origin. Using the current directory seems a huge security risk in case somebody passes non sanitized data into fetch().

KhafraDev:
> As for origin, I would prefer we add some way to set the base origin. 

A base URL is defined as "A URL used by APIs called by scripts that use this environment settings object to parse URLs." in https://html.spec.whatwg.org/multipage/webappapis.html#api-base-url, which has no way of being set. Being able to set it would go against the spec, but maybe that's the direction you'd like to go?

> Using the current directory seems a huge security risk in case somebody passes non sanitized data into fetch().

It is but I couldn't think of a better alternative. I could probably sanitize the input however.

I'd like to know how the lib should handle this before adding tests, but I will do so! 😃

targos:
At a first glance, I'm not sure we should do it, for two reasons:
- it would be a relatively big spec divergence
- I was thinking about adding an option to Node.js to let the user set an origin for the process. This would apply to fetch calls and solve part of the issues with wpt

PR #1070, opened 2021-10-21
feat: add proxy-agent

RafaelGSS:
Address: #569.

There is a remaining point to be discussed. I've implemented it and tested it against other HTTP libraries ([request](https://github.com/request/request), [axios](https://github.com/axios/axios)) and here are a quick HTTP response summary:

<details>
<summary> Undici </summary>
<br>

image

</details>

<details>
<summary> Axios </summary>
<br>

image

</details>

<details>
<summary> Request </summary>
<br>

image

</details>

You will notice that other HTTP libraries answer the HTTP request with `Connection: close` through the proxy. Nonetheless, `undici` keeps the connection open with `keep-alive`. I'm supposing that it's was designed by default, I'm sharing it just to align expectations.

---

TODO:

- [x] Add ProxyAgent test 
- [x] Https test
- [x] Add ProxyAgent types test
- [x] Docs

PR #1194, opened 2022-02-01
Fix import of `ProxyAgent`

RA80533:
ProxyAgent is imported with the assumption that `allowSyntheticDefaultImports` is enabled. The aforementioned option is disabled by default. As a result, users are required to enable `allowSyntheticDefaultImports` in order to have an error-free TypeScript experience.

https://github.com/nodejs/undici/pull/1070#discussion_r796794771


RafaelGSS (approved):
LGTM.

mcollina (approved):
lgtm

PR #1199, opened 2022-02-02
types: align RequestInit types with lib.dom.ts

### LINKED ISSUE #1172: RequestInit interface does not support all of the props in dom RequestInit interface
## Bug Description

The RequestInit interface does not have many of the props in the dom version of RequestInit.

undici RequestInit

https://github.com/nodejs/undici/blob/5d524231ff0fcb8dac761fd710fa08d7d61b1c26/types/fetch.d.ts#L95-L103

dom RequestInit
```ts
interface RequestInit {
    /** A BodyInit object or null to set request's body. */
    body?: BodyInit | null;
    /** A string indicating how the request will interact with the browser's cache to set request's cache. */
    cache?: RequestCache;
    /** A string indicating whether credentials will be sent with the request always, never, or only when sent to a same-origin URL. Sets request's credentials. */
    credentials?: RequestCredentials;
    /** A Headers object, an object literal, or an array of two-item arrays to set request's headers. */
    headers?: HeadersInit;
    /** A cryptographic hash of the resource to be fetched by request. Sets request's integrity. */
    integrity?: string;
    /** A boolean to set request's keepalive. */
    keepalive?: boolean;
    /** A string to set request's method. */
    method?: string;
    /** A string to indicate whether the request will use CORS, or will be restricted to same-origin URLs. Sets request's mode. */
    mode?: RequestMode;
    /** A string indicating whether request follows redirects, results in an error upon encountering a redirect, or returns the redirect (in an opaque fashion). Sets request's redirect. */
    redirect?: RequestRedirect;
    /** A string whose value is a same-origin URL, "about:client", or the empty string, to set request's referrer. */
    referrer?: string;
    /** A referrer policy to set request's referrerPolicy. */
    referrerPolicy?: ReferrerPolicy;
    /** An AbortSignal to set request's signal. */
    signal?: AbortSignal | null;
    /** Can only be null. Used to disassociate request from any Window. */
    window?: null;
}
```

## Reproducible By

```ts
fetch('https://example.com', { mode: 'cors' })
```

## Expected Behavior

No type errors

## Logs & Screenshots

```
TS2345: Argument of type '{ mode: string; }' is not assignable to parameter of type 'RequestInit'.
  Object literal may only specify known properties, and 'mode' does not exist in type 'RequestInit'.
```

## Environment

Node v17.3.1

### Additional context

Even if the props are not implemented, it is still useful to have them to properly support isomorphic fetch.

suneettipirneni:
## This relates to...

#1172

## Rationale

Types for `RequestInit` should be fully compatible with typescript dom types.

## Changes

Adds all properties to `RequestInit` that were missing based on the typescript dom types. This adjusts the `Headers` subtype to also follow the typescript definitions.

### Features

N/A

### Bug Fixes

closes #1172

N/A


types/fetch.d.ts
```diff
@@ -96,12 +96,29 @@ export interface RequestInit {
   readonly method?: string
   readonly keepalive?: boolean
   readonly headers?: HeadersInit
-  readonly body?: BodyInit
+  readonly body?: BodyInit;
   readonly redirect?: RequestRedirect
   readonly integrity?: string
   readonly signal?: AbortSignal
+  readonly cache?: RequestCache
```
mcollina, line 103: Are we sure we want to add types for properties that will never be supported?

ronag (approved):

mcollina:
Could you add a few assertions for these changes in https://github.com/nodejs/undici/blob/main/test/types/fetch.test-d.ts?

Ethan-Arrowood (approved):
approving but also like @mcollina comment, please update the assertions in the test file before merging

lib/core/util.js
```diff
@@ -1,276 +1,294 @@
-'use strict'
+"use strict";
```
mcollina, line 1: I'm not sure how this file passes linting but it's not ok. Can you please restore it to its previous state?

Ethan-Arrowood:
I assume you have prettier/eslint auto enabled in your editor. please revert all of the changes in the .js file as it does not apply to this PR's intended changes. Thank you!

mcollina (approved):
lgtm

PR #1200, opened 2022-02-03
feat(MockInterceptor): add support for callbacks on `reply`

suneettipirneni:
This adds the ability to have callbacks when replying to a mocked requests, similar to how nock enables callbacks for responses. This is especially helpful for situations where the response depends on certain request options.

Example of how it can be used:
```ts
mockPool
  .intercept({
    path: '/test',
    method: 'GET',
  })
  .reply(200, ({ headers }) => ({
    myHeader: headers.get('foo'),
  }));


const res = await fetch('http://localhost:3000/test', {
    headers: {
      foo: 'bar',
    },
  }).then((res) => res.json());

console.log(res); // Prints { myHeader: 'bar' }

```

### DIFF
--- a/docs/api/MockPool.md
+++ b/docs/api/MockPool.md
@@ -62,7 +62,7 @@ Returns: `MockInterceptor` corresponding to the input options.
 
 We can define the behaviour of an intercepted request with the following options.
 
-* **reply** `(statusCode: number, replyData: string | Buffer | object, responseOptions?: MockResponseOptions) => MockScope` - define a reply for a matching request. Default for `responseOptions` is `{}`.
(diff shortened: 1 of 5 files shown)


mcollina (changes requested):

mcollina (changes requested):
Good work!

Could you:

1. add a test for the types? we use tsd!
2. update the documentation

docs/api/MockPool.md
```diff
@@ -163,7 +196,6 @@ const mockPool = mockAgent.get('http://localhost:3000')
 mockPool.intercept({
   path: '/foo?hello=there&see=ya',
   method: 'POST',
-  body: 'form1=data1&form2=data2',
```
mcollina, line 166: why was this line removed? Could you restore it?

mcollina (commented):
Good work! one final question

mcollina (approved):
lgtm

PR #1204, opened 2022-02-04
Fix order of WASM loading

mcfedr:
Recent changes to how the wasm is loaded in #1183 have accidentally changed the order of simd/other builds.

You can see in the diff, originally the 

```diff
  try {
-    mod = await WebAssembly.compile(await readFile(resolve(__dirname, './llhttp/llhttp_simd.wasm')))
+    mod = await WebAssembly.compile(Buffer.from(llhttpWasmData, 'base64'))
  } catch (e) {
...
-    mod = await WebAssembly.compile(await readFile(resolve(__dirname, './llhttp/llhttp.wasm')))
+    mod = await WebAssembly.compile(Buffer.from(llhttpSimdWasmData, 'base64'))
  }
```

This makes sense because the simd build should throw when it isnt supported, and the catch loads the basic version.

Currently no one will be using the simd version.

This PR changes the order back, so that simd is loaded first, and on error we can the basic version.

I've also inlined the `require` call to the base64 JS module, this way most of the time the text of the basic version doesnt need to be loaded.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -24,9 +24,6 @@ const {
   HTTPParserError
 } = require('./core/errors')
 const buildConnector = require('./core/connect')
-const llhttpWasmData = require('./llhttp/llhttp.wasm.js')
-const llhttpSimdWasmData = require('./llhttp/llhttp_simd.wasm.js')
-
 const {
   kUrl,
   kReset,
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

PR #1110, opened 2021-11-22
fix: improve support for Blob and File like objects

ronag:
Refs: https://github.com/nodejs/undici/pull/1109

PR #1207, opened 2022-02-06
fix(fetch): prevent `Illegal invocations` of the `FileLike` class

RaisinTen:

### DIFF
--- a/lib/fetch/file.js
+++ b/lib/fetch/file.js
@@ -124,26 +124,50 @@ class FileLike {
   }
 
   stream (...args) {
+    if (!(this instanceof FileLike)) {
+      throw new TypeError('Illegal invocation')
+    }
+
     return this[kState].blobLike.stream(...args)
   }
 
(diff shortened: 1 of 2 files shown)


ronag (approved):

mcollina (commented):
Why is this needed? Is it part of the spec?

RaisinTen:
> Why is this needed?

I added the validators because the rest of the methods both in `File` (even the ones inherited from `require('buffer').Blob`) and `FileLike` had them and I couldn't think of a good reason why we shouldn't validate `this`.

> Is it part of the spec?

Is `FileLike` a part of the spec? It was added in https://github.com/nodejs/undici/pull/1110 without any spec link. I was looking for it in https://w3c.github.io/FileAPI/#dfn-file but I couldn't find it. Do you have a link to the `FileLike` spec?

mcollina (approved):
lgtm

Issue #1193, opened 2022-02-01
Fetch: Allow manual redirect handling


adamburgess:
## This would solve...

Currently you only get an opaque responses using `fetch(..., { redirect: 'manual' })`, for example:

```js
import { fetch } from 'undici'
fetch('https://httpbin.org/status/302', {redirect: 'manual'})
      .then(res => console.log(res.status, res.headers))
// -> 0 HeadersList(0) []
```

This is what the spec says should happen, but it is not very useful for a server-side library...

## The implementation should look like...

`redirect: 'manual'` should allow me to get the 302 request and read it like normal.

## Additional context

Deno changed their fetch API to support this: https://github.com/denoland/deno/pull/8353 https://github.com/denoland/deno/issues/4389

```js
fetch('https://httpbin.org/status/302', {redirect: 'manual'})
      .then(res => console.log(res.status, res.headers))
/* -> 302 Headers {
        "access-control-allow-credentials": "true",
        "access-control-allow-origin": "*",
        "content-length": "0",
        date: "Tue, 01 Feb 2022 13:51:40 GMT",
        location: "/redirect/1",
        server: "gunicorn/19.9.0"
      }
*/
```

Cloudflare Workers' fetch API also supports this.

adamburgess:
Solved: https://github.com/nodejs/node/pull/41749#issuecomment-1025473227

benjamingr:
I think it's worth keeping an issue open to make sure this doesn't get missed :]

(If you feel strongly the issue should stay closed let me know and I'll re-close it and open a new one)

mcollina:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

RaisinTen:
Sent a PR to implement this - https://github.com/nodejs/undici/pull/1210

PR #1210, opened 2022-02-07
feat(fetch): allow `manual` redirect handling

RaisinTen:
It doesn't make sense to return `opaqueredirect` on server side because it
is not manually possible to follow an HTTP redirect. Hence, this aligns
fetch with the other implementations in Deno and Cloudflare Workers.

Fixes: https://github.com/nodejs/undici/issues/1193


ronag (approved):

benjamingr (approved):

lib/fetch/index.js
```diff
@@ -1042,9 +1042,10 @@ async function httpFetch (fetchParams) {
       // Set response to a network error.
       response = makeNetworkError()
     } else if (request.redirect === 'manual') {
-      // Set response to an opaque-redirect filtered response whose internal
-      // response is actualResponse.
-      response = filterResponse(actualResponse, 'opaqueredirect')
+      // On the web this would return an `opaqueredirect` response, but
+      // those don't make sense server side.
+      // See https://github.com/nodejs/undici/issues/1193.
+      response = actualResponse
```
ronag, line 1048:
```suggestion
      // Set response to an opaque-redirect filtered response whose internal
      // response is actualResponse.
      // NOTE(spec): On the web this would return an `opaqueredirect` response, but
      // those don't make sense server side.
      // See https://github.com/nodejs/undici/issues/1193.
      response = actualResponse
```

RaisinTen: thanks, done!

RaisinTen:
@ronag sure, I've updated the readme with the required info, PTAL

ronag:
Thanks!

adamburgess:
how come the set-cookie test was changed? it was failing before... ideally the forbidden headers list is also removed for undici

(it was allowed to fail in https://github.com/nodejs/undici/pull/928/commits/c3822cfd64581177bb412b1852e01cb9985e1040 with the `|| echo Skipping`?)

RaisinTen:
@adamburgess I was actually running the tests locally with `./node_modules/.bin/mocha test/node-fetch/main.js` which doesn't skip that and sets the exit code to `1` (Node.js v17.4.0) unlike `npm run test:node-fetch`, so I modified that test too unintentionally.

PR #1211, opened 2022-02-08
feat(mock): add support reply option callbacks

suneettipirneni:
This PR is similar to #1200, except it allows dynamic mocking of *all* reply options rather than just the `data`.

```js
mockPool.intercept({
  path: '/echo',
  method: 'GET',
  headers: {
    'User-Agent': 'undici',
    Host: 'example.com'
  }
}).reply(({ headers }) => ({ statusCode: 200, data: { message: headers.get('message') }})))

const { statusCode, body, headers } = await request('http://localhost:3000', {
  headers: {
    message: 'hello world!'
  }
})

console.log('response received', statusCode) // response received 200
console.log('headers', headers) // { 'content-type': 'application/json' }

  console.log('data', data.toString('utf8')) // { "message":"hello world!" }
}
```

This is useful in situations where the status code and the response headers need to be dynamically mocked.

### DIFF
--- a/test/mock-interceptor.js
+++ b/test/mock-interceptor.js
@@ -53,6 +53,23 @@ test('MockInterceptor - reply callback', t => {
     t.throws(() => mockInterceptor.reply(), new InvalidArgumentError('statusCode must be defined'))
     t.throws(() => mockInterceptor.reply(200, () => {}, 'hello'), new InvalidArgumentError('responseOptions must be an object'))
   })
+})
+
+test('MockInterceptor - reply options callback', t => {
+  t.plan(1)
+
+  t.test('should return MockScope', t => {
+    t.plan(1)
+    const mockInterceptor = new MockInterceptor({
+      path: '',
+      method: ''
+    }, [])
+    const result = mockInterceptor.reply(() => ({
+      statusCode: 200,
+      data: 'hello'
+    }))
+    t.type(result, MockScope)
+  })
 }) 
 
 test('MockInterceptor - replyWithError', t => {
--- a/test/mock-interceptor.js
+++ b/test/mock-interceptor.js
@@ -2,6 +2,7 @@
 
 const { test } = require('tap')
 const { MockInterceptor, MockScope } = require('../lib/mock/mock-interceptor')
+const MockAgent = require('../lib/mock/mock-agent');
 const { InvalidArgumentError } = require('../lib/core/errors')
 
 test('MockInterceptor - reply', t => {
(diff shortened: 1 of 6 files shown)


test/mock-interceptor.js
```diff
@@ -53,6 +53,23 @@ test('MockInterceptor - reply callback', t => {
...
+test('MockInterceptor - reply options callback', t => {
+  t.plan(1)
+
+  t.test('should return MockScope', t => {
+    t.plan(1)
+    const mockInterceptor = new MockInterceptor({
+      path: '',
+      method: ''
+    }, [])
+    const result = mockInterceptor.reply(() => ({
```
mcollina, line 67: Can you add a few assertions on the parameters of this function?

test/types/mock-interceptor.test-d.ts
```diff
@@ -19,6 +19,13 @@ import { MockInterceptor, MockScope } from '../../types/mock-interceptor'
   expectAssignable<MockScope>(mockInterceptor.reply(200, () => ({}), { trailers: { foo: 'bar' }}))
   expectAssignable<MockScope<{ foo: string }>>(mockInterceptor.reply<{ foo: string }>(200, { foo: 'bar' }))
   expectAssignable<MockScope<{ foo: string }>>(mockInterceptor.reply<{ foo: string }>(200, () => ({ foo: 'bar' })))
+  expectAssignable<MockScope>(mockInterceptor.reply(() => ({ statusCode: 200, data: { foo: 'bar' }})))
+  expectAssignable<MockScope>(mockInterceptor.reply(() => ({ statusCode: 200, data: { foo: 'bar' }, responseOptions: {
+    headers: { foo: 'bar' }
+  }})))
+  expectAssignable<MockScope>(mockInterceptor.reply(() => ({ statusCode: 200, data: { foo: 'bar' }, responseOptions: {
+    trailers: { foo: 'bar' }
+  }})))
```
mcollina, line 28: Can you add another assertion for the parameter of this function?

test/mock-interceptor.js
suneettipirneni: So I had a question about this. Since callbacks are late-binding, their validation errors for `status`, `data` and `responseOptions` get thrown later in the actual fetch operation. This leads these validation errors to show up as `fetch failed` errors. Would there be any way to expose the true error during the fetch operation instead of just getting the standard `fetch failed` error?

suneettipirneni: Oh actually never mind, it's attached the the `cause` of the fetched failed error

mcollina (approved):
lgtm

PR #1212, opened 2022-02-08
fix: incorrect http(s) regexp

cjh980402:
`/^http?s/` is matched for `htts` and `https`.
It should fix to `/^https?:/` (matched for `http:` and `https:`)


ronag (approved):

ronag:
Do you think you could add a regression test?

mcollina (approved):
lgtm

cjh980402:
> Do you think you could add a regression test?

I add more test for requestBadPort

PR #1216, opened 2022-02-09
fix(test): fix test skipping logic

RaisinTen:
I had unintentionally modified the cookie test in
https://github.com/nodejs/undici/pull/1210 to fix the failures I got
when I ran them using:
```sh
./node_modules/.bin/mocha test/node-fetch/main.js
```
instead of
```sh
npm run test:node-fetch
```
However, the test was failing even before that change landed, so I
reverted the change I made to the test and allowed it to get skipped.

Also, we were skipping all of the fetch tests for all Node.js versions,
< v16 as well as >= v16, so this change fixes the test skipping logic.

Refs: https://github.com/nodejs/undici/pull/1210#issuecomment-1032854813


ronag (approved):

mcollina (approved):
lgtm

PR #1219, opened 2022-02-10
chore: ignore wasm file for npm publishing

Kikobeats:
After https://github.com/nodejs/undici/pull/1183, WASM files are no more necessary to be shipped as part of the NPM bundle


ronag (approved):

mcollina:
I thiink we need to add a bit more things from `.gitignore`, have you tried looking inside the generated package?

mcollina (approved):
lgtm

PR #1224, opened 2022-02-11
Add missing ts import

### LINKED ISSUE #1223: TypeScript failure
The [last release](https://github.com/nodejs/undici/releases/tag/v4.14.0) introduced a TypeScript bug that does not allow TypeScript to compile. It looks like the bug has been introduced in https://github.com/nodejs/undici/pull/1200.

Error:
```
node_modules/undici/types/mock-interceptor.d.ts:74:14 - error TS2304: Cannot find name 'Headers'.

74     headers: Headers;
                ~~~~~~~
```

To reproduce, create a empty project, install `undici`, `typescript` and `@types/node` and try to compile the following:
```ts
```

I don't know how to fix it, if anyone has an idea I can open a pr.

delvedor:
As titled. I don't know why `tsd` didn't catch this.

Fix #1223

### DIFF
--- a/package.json
+++ b/package.json
@@ -109,7 +109,8 @@
   "tsd": {
     "directory": "test/types",
     "compilerOptions": {
-      "esModuleInterop": true
+      "esModuleInterop": true,
+      "lib": ["esnext"]
     }
   }
 }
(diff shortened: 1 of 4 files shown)


ronag (approved):

delvedor:
[`30f4b12` (#1224)](https://github.com/nodejs/undici/pull/1224/commits/30f4b12d52c75c71da634020ca5acf6eb41b1026) removes the `dom` lib from tsd config. If I disable the fix I did in the first commit, the test will fail.
I'm not 100% this is the right solution tho.

Issue #958, opened 2021-08-14
Implement fetch methods


ronag:
Under `lib/fetch/util.js` we have a few helper methods that are pretty well documented in spec and and needs to be implemented:

- [x] appendRequestOriginHeader
- [x] appendFetchMetadata
- [x] TAOCheck
- [x] corsCheck
- [x] crossOriginResourcePolicyCheck
- [x] makePolicyContainer
- [x] clonePolicyContainer
- [x] determineRequestsReferrer

ronag:
@szmarczak

wong2:
I'd like to help!

wong2:
Though I have a question: do we need to implement those browser related features such as cors, content security policy in undici fetch?

ronag:
Some of those are indeed only browser related, you can either implement them if possible or add a note explaining why we don't need to implement them and return a appropriate value or an assertion.

wong2:
I'd like to confirm what we think about `cors` related features, for example, `mode` is currently supported, while it's unimplemented in node-fetch. Do we want to support them as much as we can? Or are they unnecessary complexities?

szmarczak:
Browser-related features might be useful when used with https://github.com/jsdom/jsdom but I don't have any use case for this.

ronag:
> I'd like to confirm what we think about `cors` related features, for example, `mode` is currently supported, while it's unimplemented in node-fetch. Do we want to support them as much as we can? Or are they unnecessary complexities?

We implement them because it's easy to just follow the spec literally.

btakita:
It also supports isomorphic javascript.

Viper-space:
Hey there is it okay if I try to implement some of them? :D

ronag:
@KhafraDev is there anything left to implement?

KhafraDev:
No, everything is done.

PR #1236, opened 2022-02-15
feat(fetch#Request): Implements `determineRequestReferrer`

metcoder95:
Relates to #958 

Pendings:
- [x] Tests

### DIFF
--- a/lib/fetch/util.js
+++ b/lib/fetch/util.js
@@ -450,33 +450,41 @@ function isURLPotentiallyTrustworthy (url) {
     return false
   }
 
+  // If child of about, return true
   if (url.href === 'about:blank' || url.href === 'about:srcdoc') {
     return true
   }
 
+  // If scheme is data, return true
   if (url.protocol === 'data:') return true
 
   return isOriginPotentiallyTrustworthy(url.origin)
 
-  function isOriginPotentiallyTrustworthy(origin) {
+  function isOriginPotentiallyTrustworthy (origin) {
+    // If origin is explicitly null, return false
     if (origin == null || origin === 'null') return false
 
     let originAsURL
 
+    // If not valid because not semantically correct, return false
     try { originAsURL = new URL(origin) } catch (e) { return false }
 
+    // If secure, return true
     if (originAsURL.protocol === 'https:' || originAsURL.protocol === 'wss:') {
       return true
     }
 
+    // If localhost or variants, return true
     if (/^127(?:\.[0-9]+){0,2}\.[0-9]+$|^(?:0*:)*?:?0*1$/.test(originAsURL.hostname) ||
      (originAsURL.hostname === 'localhost' || originAsURL.hostname.includes('localhost.')) ||
      (originAsURL.hostname.endsWith('.localhost'))) {
       return true
     }
 
+    // If file, return true
     if (originAsURL.protocol === 'file:') return true
 
+    // If any other, return false
     return false
   }
 }
--- a/lib/fetch/util.js
+++ b/lib/fetch/util.js
@@ -770,6 +770,8 @@ module.exports = {
   responseURL,
   responseLocationURL,
   isBlobLike,
+  isFileLike,
+  isURLPotentiallyTrustworthy,
   isValidReasonPhrase,
   sameOrigin,
   normalizeMethod,
--- a/lib/fetch/util.js
+++ b/lib/fetch/util.js
@@ -457,6 +457,9 @@ function isURLPotentiallyTrustworthy (url) {
 
   // If scheme is data, return true
   if (url.protocol === 'data:') return true
+  
+  // If file, return true
+  if (url.protocol === 'file:') return true
 
   return isOriginPotentiallyTrustworthy(url.origin)
 
@@ -475,14 +478,11 @@ function isURLPotentiallyTrustworthy (url) {
     }
 
     // If localhost or variants, return true
-    if (/^127(?:\.[0-9]+){0,2}\.[0-9]+$|^(?:0*:)*?:?0*1$/.test(originAsURL.hostname) ||
+    if (/^127(?:\.[0-9]+){0,2}\.[0-9]+$|^\[(?:0*:)*?:?0*1\]$/.test(originAsURL.hostname) ||
      (originAsURL.hostname === 'localhost' || originAsURL.hostname.includes('localhost.')) ||
      (originAsURL.hostname.endsWith('.localhost'))) {
       return true
     }
-
-    // If file, return true
-    if (originAsURL.protocol === 'file:') return true
 
     // If any other, return false
     return false
--- a/lib/fetch/util.js
+++ b/lib/fetch/util.js
@@ -332,7 +332,6 @@ function clonePolicyContainer () {
 
 // https://w3c.github.io/webappsec-referrer-policy/#determine-requests-referrer
 function determineRequestsReferrer (request) {
-  // TODO
   // 1. Let policy be request's referrer policy.
   const policy = request.referrerPolicy
 
@@ -345,14 +344,29 @@ function determineRequestsReferrer (request) {
   const environment = request.client
   let referrerSource = null
 
-  // TODO
-  // 3(a) If policy is "origin", then
-  // 3(a)(I) If environment's global object is Window then
+  /**
+   * 3, Switch on request’s referrer:
+    "client"
+      If environment’s global object is a Window object, then
+        Let document be the associated Document of environment’s global object.
+        If document’s origin is an opaque origin, return no referrer.
+        While document is an iframe srcdoc document,
+        let document be document’s browsing context’s browsing context container’s node document.
+        Let referrerSource be document’s URL.
+
+      Otherwise, let referrerSource be environment’s creation URL.
+
+    a URL
+    Let referrerSource be request’s referrer.
+   */
   if (request.referrer === 'client') {
     if (environment?.globalObject instanceof Window) {
       const origin = environment.globalObject.self?.origin ?? environment.globalObject.location?.origin
 
+      // If document’s origin is an opaque origin, return no referrer.
       if (origin == null || origin === 'null') return 'no-referrer'
+
+      // Let referrerSource be document’s URL.
       referrerSource = new URL(environment.globalObject.location.href)
     } else {
       // 3(a)(II) If environment's global object is not Window,
--- a/lib/fetch/util.js
+++ b/lib/fetch/util.js
@@ -336,6 +336,7 @@ function determineRequestsReferrer (request) {
   // 1. Let policy be request's referrer policy.
   const policy = request.referrerPolicy
 
+  // Return no-referrer when empty or policy says so
   if (policy === '' || policy === 'no-referrer') return 'no-referrer'
 
   // 2. Let environment be the request client
@@ -390,7 +391,7 @@ function determineRequestsReferrer (request) {
     : temp
   const areSameOrigin = sameOrigin(request, referrerUrl)
 
-  // NOTE: how shall we treat step 7?
+  // NOTE: How to treat step 7?
   // 8. Execute the switch statements corresponding to the value of policy:
   switch (policy) {
     case 'origin': return referrerOrigin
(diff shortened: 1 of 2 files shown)


ronag (changes requested):
Please paste in the spec test similar to our other implementations.

MetCoder95:
Sure thing, I actually have two doubts in here:
1. How do we treat step 7 of the step, is not something super clear to me. It quotes:

>The user agent MAY alter referrerURL or referrerOrigin at this point to enforce arbitrary policy considerations in the interests of minimizing data leakage. For example, the user agent could strip the URL down to an origin, modify its host, replace it with an empty string, etc.

AFAIU means the agent has full control about the `referrerURL`, and treats it accordingly to our policies. Then, this step should be totally ignored or just leave it with a note to indicate this part of the spec?

2. In order to cover all the `referrer policy` cases is needed to assess if the `referrerURL` is `potentially trustworthy` and the same applies for the `currentURL`. Shall I make a step forward and implement the asses algorithm by the spec? I think is also needed in other parts of Undici when handling the URL

I'll cover your suggestions, add documentation (on code and for implementation) 👍

MetCoder95:
I'm not sure about those errors, any hint? 🤔

MetCoder95:
Thanks for the hint @mcollina, but it seems it was a bad check on the Request environment, as I didn't check for possible undefined from the root object. Now shall be fixed, looking forward for feedback 🙂

lib/fetch/util.js
```diff
@@ -286,7 +286,160 @@ function clonePolicyContainer () {
...
+  if (policy === '' || policy === 'no-referrer') return 'no-referrer'
+
+  // 2. Let environment be the request client
+  const environment = request.client
+  let referrerSource = null
+
+  // TODO
+  // 3(a) If policy is "origin", then
+  // 3(a)(I) If environment's global object is Window then
+  if (request.referrer === 'client') {
```
ronag, line 302: Why this? Not in spec?

lib/fetch/util.js
```diff
@@ -286,7 +286,160 @@ function clonePolicyContainer () {
...
+      if (origin == null || origin === 'null') return 'no-referrer'
+      referrerSource = new URL(environment.globalObject.location.href)
+    } else {
+      // 3(a)(II) If environment's global object is not Window,
+      // Let referrerSource be environments creationURL
+      if (environment?.globalObject?.location == null) {
+        return 'no-referrer'
+      }
+
+      referrerSource = new URL(environment.globalObject.location.href)
```
ronag, line 315: Also not part of spec?

lib/fetch/util.js
```diff
@@ -286,7 +286,160 @@ function clonePolicyContainer () {
 // https://w3c.github.io/webappsec-referrer-policy/#determine-requests-referrer
 function determineRequestsReferrer (request) {
   // TODO
-  return 'no-referrer'
+  // 1. Let policy be request's referrer policy.
+  const policy = request.referrerPolicy
+
+  // Return no-referrer when empty or policy says so
+  if (policy === '' || policy === 'no-referrer') return 'no-referrer'
```
ronag, line 293:
```suggestion
  if (policy === '' || policy === 'no-referrer') {
    return 'no-referrer'
  }
```

lib/fetch/util.js
```diff
@@ -286,7 +286,160 @@ function clonePolicyContainer () {
 // https://w3c.github.io/webappsec-referrer-policy/#determine-requests-referrer
 function determineRequestsReferrer (request) {
   // TODO
```
ronag, line 288:
```suggestion
```

lib/fetch/util.js
```diff
@@ -286,7 +286,160 @@ function clonePolicyContainer () {
...
+    case 'origin-when-cross-origin':
+      return areSameOrigin ? referrerUrl : referrerOrigin
+    case 'no-referrer-when-downgrade':
+      /**
+       * 1. If referrerURL is a potentially trustworthy URL and
+       * request’s current URL is not a potentially trustworthy URL,
+       * then return no referrer.
+       * 2. Return referrerOrigin
+      */
+      return isNonPotentiallyTrustWorthy ? 'no-referrer' : referrerOrigin
```
ronag, line 383: default?

lib/fetch/util.js
MetCoder95: Accordingly to the third step of the spec, states a switch on the `referrer` value, but wasn’t sure if we want it to totally cover it or now, so I just applied as spec stated:

>3, Switch on request’s [referrer](https://fetch.spec.whatwg.org/#concept-request-referrer):
- "client"
  - If environment’s [global object](https://html.spec.whatwg.org/multipage/webappapis.html#concept-settings-object-global) is a [Window](https://html.spec.whatwg.org/multipage/window-object.html#window) object, then
     - Let document be the [associated Document](https://html.spec.whatwg.org/multipage/browsers.html#concept-document-window) of environment’s [global object](https://html.spec.whatwg.org/multipage/webappapis.html#concept-settings-object-global).
       - If document’s [origin](https://dom.spec.whatwg.org/#concept-document-origin) is an [opaque origin](https://html.spec.whatwg.org/multipage/browsers.html#concept-origin-opaque), return no referrer.
       - While document is [an iframe srcdoc document](https://html.spec.whatwg.org/multipage/embedded-content.html#an-iframe-srcdoc-document), let document be document’s [browsing context](https://html.spec.whatwg.org/multipage/browsers.html#concept-document-bc)’s [browsing context container](https://html.spec.whatwg.org/multipage/browsers.html#browsing-context-container)’s [node document](https://dom.spec.whatwg.org/#concept-node-document).
      - Let referrerSource be document’s [URL](https://dom.spec.whatwg.org/#concept-document-url).
  - Otherwise, let referrerSource be environment’s [creation URL](https://html.spec.whatwg.org/multipage/webappapis.html#concept-environment-creation-url).
a [URL](https://url.spec.whatwg.org/#concept-url)
- Let [referrer](https://fetch.spec.whatwg.org/#concept-request-referrer)Source be request’s referrer.

lib/fetch/util.js
ronag: This is not part of spec?

lib/fetch/util.js
ronag: Look at how it's done in the rest of the existing code.

MetCoder95: Sure thing, I'll do that 👍

lib/fetch/util.js
MetCoder95: Accordingly to the spec, it quite is 🤔 

>Switch on request’s [referrer](https://fetch.spec.whatwg.org/#concept-request-referrer):
- "client"
  - If environment’s [global object](https://html.spec.whatwg.org/multipage/webappapis.html#concept-settings-object-global) is a [Window](https://html.spec.whatwg.org/multipage/window-object.html#window) object, then
      - Let document be the [associated Document](https://html.spec.whatwg.org/multipage/browsers.html#concept-document-window) of environment’s [global object](https://html.spec.whatwg.org/multipage/webappapis.html#concept-settings-object-global).
    - If document’s [origin](https://dom.spec.whatwg.org/#concept-document-origin) is an [opaque origin](https://html.spec.whatwg.org/multipage/browsers.html#concept-origin-opaque), return no referrer.
    - While document is [an iframe srcdoc document](https://html.spec.whatwg.org/multipage/embedded-content.html#an-iframe-srcdoc-document), 
let document be document’s [browsing context](https://html.spec.whatwg.org/multipage/browsers.html#concept-document-bc)’s [browsing context container](https://html.spec.whatwg.org/multipage/browsers.html#browsing-context-container)’s [node document](https://dom.spec.whatwg.org/#concept-node-document).
    - Let referrerSource be document’s [URL](https://dom.spec.whatwg.org/#concept-document-url).
  - Otherwise, let referrerSource be environment’s [creation URL](https://html.spec.whatwg.org/multipage/webappapis.html#concept-environment-creation-url).
- a [URL](https://url.spec.whatwg.org/#concept-url)
  - Let [referrer](https://fetch.spec.whatwg.org/#concept-request-referrer)Source be request’s referrer.


In specific this statement:
> Let referrerSource be document’s [URL](https://dom.spec.whatwg.org/#concept-document-url).

Maybe I understood it wrong?

lib/fetch/util.js
MetCoder95: Not explicitly, but if there's no value or the `policy` explicitly says that there's `no-referrer`, there's no point in continuing executing the spec algorithm. Any suggestion on how to handle it? 🙂

metcoder95:
Hi @ronag, changes applied and conflicts fixed, let me know if there's anything else to change 👍

ronag:
@metcoder95 CI is still failing.

lib/fetch/util.js
```diff
@@ -332,8 +332,178 @@ function clonePolicyContainer () {
...
+        Let referrerSource be document’s URL.
+
+      Otherwise, let referrerSource be environment’s creation URL.
+
+    a URL
+    Let referrerSource be request’s referrer.
+   */
+  if (request.referrer === 'client') {
+    // Not defined in Node but part of the spec
+    if (environment?.globalObject instanceof Window) { // eslint-disable-line
```
KhafraDev, line 364: This should be changed to something similar to https://github.com/nodejs/undici/blob/e5e5b97d83010cfb752eff8d9c8d488399cc3ab7/lib/fetch/index.js#L414

lib/fetch/util.js
```diff
@@ -332,8 +332,178 @@ function clonePolicyContainer () {
...
+    return 'no-referrer'
+  }
+
+  const urlProtocol = referrerSource.protocol
+
+  // If url's scheme is a local scheme (i.e. one of "about", "data", "javascript", "file")
+  // then return "no-referrer".
+  if (
+    urlProtocol === 'about:' || urlProtocol === 'data:' ||
+    urlProtocol === 'javascript:' || urlProtocol === 'file:'
```
KhafraDev, line 397: `javascript` isn't a local scheme
https://fetch.spec.whatwg.org/#local-scheme

KhafraDev (changes requested):

lib/fetch/util.js
metcoder95: Sure, applied in [6ed09c1](https://github.com/nodejs/undici/pull/1236/commits/6ed09c17be399e459f9174dbaf4fd30206bcb8c5)

lib/fetch/util.js
metcoder95: Sure, applied in [6ed09c1](https://github.com/nodejs/undici/pull/1236/commits/6ed09c17be399e459f9174dbaf4fd30206bcb8c5)

metcoder95:
It seems the testing was failing due to some bad import. Fixed 🙂

lib/fetch/util.js
KhafraDev: this doesn't seem to have been done

lib/fetch/util.js
```diff
@@ -332,8 +332,178 @@ function clonePolicyContainer () {
...
+  return isOriginPotentiallyTrustworthy(url.origin)
+
+  function isOriginPotentiallyTrustworthy (origin) {
+    // If origin is explicitly null, return false
+    if (origin == null || origin === 'null') return false
+
+    let originAsURL
+
+    // If not valid because not semantically correct, return false
+    try { originAsURL = new URL(origin) } catch (e) { return false }
```
KhafraDev, line 490: this shouldn't be able to throw

KhafraDev (changes requested):
is it possible to add tests that use `fetch` directly, instead of just testing the utility methods?

lib/fetch/util.js
metcoder95: True, sorry for that. Change was made but didn't stage it for the commit 😓 
Here it is the new one: 681a755

lib/fetch/util.js
metcoder95: You're right, a bad `bad origin` cannot happen without being priorly identified. Removed in 2ca0941

metcoder95:
>is it possible to add tests that use fetch directly, instead of just testing the utility methods?

Sure, by just changing to a different value from `no-referrer` on the `Request#referrer` when using Fetch should do the work. I'll add some during the day 🙂

ronag:
@KhafraDev are we good here?

KhafraDev (approved):

PR #1237, opened 2022-02-16
feat: add factory option to BalancedPool

### LINKED ISSUE #1220: BalancedPool is missing factory option
`BalancedPool` should have the `factory` option similar to `Pool` and `Agent`.

nipeshkc7:
closes #1220

### DIFF
--- a/docs/api/BalancedPool.md
+++ b/docs/api/BalancedPool.md
@@ -11,14 +11,15 @@ Requests are not guaranteed to be dispatched in order of invocation.
 Arguments:
 
 * **upstreams** `URL | string | string[]` - It should only include the **protocol, hostname, and port**.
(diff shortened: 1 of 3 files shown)


ronag (commented):
Can you add a test for the factory actually being used?

ronag (commented):
Also I see a problem here. The factory options can overlap with the constructor options. Not sure how to resolve that. I think we might have similar issues in the other agents as well. @mcollina wdyt?

nipeshkc7:
> Can you add a test for the factory actually being used?

Sure can do that.

mcollina:
Could you add the option to our docs?

I'm ok with the behavior - I assume the factory can override any of the agent options.

mcollina (approved):
lgtm

PR #1239, opened 2022-02-17
feat(fetch): add support for file urls in fetch

Linkgoron:
This PR adds support for file urls in fetch

refs: https://github.com/nodejs/node/issues/42003

According to the fetch spec, file urls are "left as an exercise for the reader."
https://fetch.spec.whatwg.org/#scheme-fetch

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -52,6 +52,7 @@ const { PassThrough, pipeline } = require('stream')
 const { isErrored, isReadable } = require('../core/util')
 const { kIsMockActive } = require('../mock/mock-symbols')
 const { dataURLProcessor } = require('./dataURL')
+const fs = require('fs')
 
 /** @type {import('buffer').resolveObjectURL} */
 let resolveObjectURL
@@ -882,7 +883,7 @@ async function schemeFetch (fetchParams) {
     case 'file:': {
       // For now, unfortunate as it is, file URLs are left as an exercise for the reader.
       // When in doubt, return a network error.
-      return makeNetworkError('not implemented... yet...')
+      return fileFetch.call(this, fetchParams)
     }
     case 'http:':
     case 'https:': {
@@ -1988,6 +1989,74 @@ function httpNetworkFetch (
       }
     )
   })
+}
+
+async function fileFetch (fetchParams) {
+  const { request } = fetchParams
+  const url = requestCurrentURL(request)
+  const context = this
+  if (request.method !== 'GET') {
+    return makeNetworkError(`Fetching files only supports the GET method. Received ${request.method}`)
+  }
+
+  let stream
+  let file
+
+  try {
+    file = await fs.promises.open(url)
+    if (context.terminated) {
+      if (context.terminated.aborted) {
+        throw new AbortError()
+      } else {
+        throw new Error('terminated')
+      }
+    }
+    stream = file.createReadStream({ autoClose: true })
+  } catch (originalError) {
+    if (stream) {
+      stream.destroy(originalError)
+    } else if (file) {
+      try {
+        await file.close()
+      } catch (closeError) {
+        return createAggregateNetworkError(originalError, closeError)
+      }
+    }
+    return makeNetworkError(originalError)
+  }
+
+  try {
+    function onTerminated () {
+      const aborted = context.terminated.aborted
+      if (aborted) {
+        response.aborted = true
+        stream.destroy(new AbortError())
+      } else {
+        stream.destroy(new TypeError('terminated'))
+      }
+    }
+    function streamClose () {
+      context.off('terminated', onTerminated)
+    }
+    const response = makeResponse({
+      status: 200,
+      statusText: 'OK',
+      headersList: [],
+      body: safelyExtractBody(stream)[0]
+    })
+    context.on('terminated', onTerminated)
+    stream.on('close', streamClose)
+    return response
+  } catch (err) {
+    stream.destroy(err)
+    return makeNetworkError(err)
+  }
+}
+
+function createAggregateNetworkError (originalError, secondError) {
+  const err = new AggregateError([originalError, secondError], originalError.message)
+  err.code = originalError.code
+  return makeNetworkError(err)
 }
 
 module.exports = fetch
--- a/test/fetch/client-fetch.js
+++ b/test/fetch/client-fetch.js
@@ -7,6 +7,8 @@ const { createServer } = require('http')
 const { ReadableStream } = require('stream/web')
 const { Blob } = require('buffer')
 const { fetch, Response, Request, FormData, File, FileLike } = require('../..')
+const { join } = require('path')
+const { pathToFileURL } = require('url')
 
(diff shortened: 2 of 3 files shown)


lib/fetch/index.js
```diff
@@ -882,7 +883,7 @@ async function schemeFetch (fetchParams) {
     case 'file:': {
       // For now, unfortunate as it is, file URLs are left as an exercise for the reader.
```
benjamingr, line 884: remove comment?

lib/fetch/index.js
```diff
@@ -1990,4 +1991,72 @@ function httpNetworkFetch (
   })
 }
 
+async function fileFetch (fetchParams) {
+  const { request } = fetchParams
+  const url = requestCurrentURL(request)
+  const context = this
```
benjamingr, line 1997: why is context just not `this`?

lib/fetch/index.js
```diff
@@ -1990,4 +1991,72 @@ function httpNetworkFetch (
...
+  if (request.method !== 'GET') {
+    return makeNetworkError(`Fetching files only supports the GET method. Received ${request.method}`)
```
benjamingr, line 1999: What do browsers do with HEAD?

lib/fetch/index.js
```diff
@@ -1990,4 +1991,72 @@ function httpNetworkFetch (
...
+  }
+
+  try {
+    function onTerminated () {
+      const aborted = context.terminated.aborted
+      if (aborted) {
+        response.aborted = true
+        stream.destroy(new AbortError())
+      } else {
+        stream.destroy(new TypeError('terminated'))
```
benjamingr, line 2035: Not specific to these PR but these errors should have codes?

(That seems orthogonal though)

test/fetch/client-fetch.js
```diff
@@ -332,3 +334,23 @@ test('post FormData with File', (t) => {
...
+  t.plan(2)
+  const fileToRead = pathToFileURL(join(__dirname, '..', 'fixtures', 'does-not-exist.text'))
+  try {
+    await fetch(fileToRead)
+    t.fail('fetch should have thrown')
+  } catch (err) {
+    t.ok(err.cause)
+    t.equal(err.cause.code, 'ENOENT')
+  }
+})
```
benjamingr, line 356: test AbortSignal?

benjamingr (commented):
Good progress

lib/fetch/index.js
Linkgoron: I looked at what's done in other methods.

lib/fetch/index.js
Linkgoron: I looked at what Deno does, I event kept the message the same.

mcollina (changes requested):
We need a way to control the security of this. There are quite a few possible attacks when reading files from disk on servers.

My opinion is that we should aim to a "close by default" approach vs "open by default".

I would recommend we add a "FileSystemEnabledAgent" that can direct requests to disk in case we need it to be. This entity will be responsible to configure what paths could be served.

benjamingr:
> We need a way to control the security of this. There are quite a few possible attacks when reading files from disk on servers.

So your concern is that we get an external URL from the user and they send over a `file` url and that lets them somehow access/manipulate data on the server?

mcollina:
> This kind of change would make this less portable (which was one of the reasons why this was requested), as I assume that other implementation don't have such constructions (I'll check them out). Is going out of spec and adding a flag or forcing localURLsOnly when using file urls an option?

That's not correct. In the browser this is managed by CSP and others. Unfortunately Node.js is not a browser and we have to deal with more strict security constraints, so I don't think this can be enabled by default without checks.

mcollina:
> So your concern is that we get an external URL from the user and they send over a file url and that lets them somehow access/manipulate data on the server?

Exactly that. Moreover I might want to send a URL to read a file that is restricted in a given folder.

Linkgoron:
> > This kind of change would make this less portable (which was one of the reasons why this was requested), as I assume that other implementation don't have such constructions (I'll check them out). Is going out of spec and adding a flag or forcing localURLsOnly when using file urls an option?
> 
> That's not correct. In the browser this is managed by CSP and others. Unfortunately Node.js is not a browser and we have to deal with more strict security constraints, so I don't think this can be enabled by default without checks.

Yes, of course. What I meant was that we I didn't think the fetch _method_ had such parameters. I edited my message after you already quoted me, where I asked if you meant that we should add some kind of global configuration for fetch that would allow such behaviour?

mcollina:
We usually include those configuration inside the `Dispatcher` system. That's the right place where the retrieving of file should happen and where the checks should be.

benjamingr:
Asked about this in Deno btw: https://github.com/denoland/deno/issues/11925#issuecomment-1043217870

KhafraDev:
I think there should also be some warning that this behavior can change at any times; once (or if - file uris have been left to the implementation for years now) the spec includes steps for handling file:// urls, this will have to be rewritten and it's very likely that the behavior will change.

I believe that only Firefox and Deno handle file urls at the moment as well, which means it's not technically spec compliant and not widely available.

Jamesernator:
While it's not a web option, could we just have an option to `fetch` to allow file urls? Browsers would simply ignore such an option so it would still be possible to write a compatible subset of `fetch` that works in both for loading package resources:

```js
// file.js
const myResource = new URL("./resource.txt", import.meta.url);
 
const resourceBlob = await fetch(myResource, {
    // If file.js is loaded in Node.js then myResource will be some file:// url in which
    // case Node.js will see this option and allow it,
    // In browsers myResource will generally be a https:// url and will just ignore this
    // options proceeding per normal
    allowFileURLs: true,
});
```

Linkgoron:
> While it's not a web option, could we just have an option to `fetch` to allow file urls? Browsers would simply ignore such an option so it would still be possible to write a compatible subset of `fetch` that works in both for loading package resources:
> 
> ```js
> // file.js
> const myResource = new URL("./resource.txt", import.meta.url);
>  
> const resourceText = await fetch(myResource, {
>     // If file.js is loaded in Node.js then myResource will be some file:// url in which
>     // case Node.js will see this option and allow it,
>     // In browsers myResource will generally be a https:// url and will just ignore this
>     // options proceeding per normal
>     allowFileURLs: true,
> }).then(res => res.text());
> ```
> 
> By having such an option, other `fetch` calls could be presumed to be "safe" (to the extent arbitrary fetches are actually safe in node).

I agree that this feature is useful, and I'd be happy to get this into the undici if we can agree on how to make everyone happy with its security.

mcollina:
Go ahead and add this option, I think it's a good addition. I would use a more ugly-and-prefixed name for it to avoid any potential conflicts with whatwg. Anyhow, I'd recommend you to open an issue at wintercg/fetch.

KhafraDev:
I would prefer to leave this to https://github.com/wintercg/fetch/issues/4 instead of implementing a non-spec compliant solution.

Issue #1213, opened 2022-02-08
Raw Header support


chasetec:
Can we get raw header support in the high level client APIs? The node http(s) clients support it and migrating to undici isn't practical if there is a dependency on having raw header support. Some origins incorrectly depend on header case. If there is a header that I need to return to the origin on subsequent requests I can't know what case the origin used right now.

Not picky on how its supported. A new rawHeaders property in the promised responses....

````
    const {
        statusCode,
        headers,
        rawHeaders,
        trailers,
        body
    } = await undici.request('http://url');
````

Or an option in DispatchOptions that would change the existing headers value.

````
    const {
        statusCode,
        headers,
        trailers,
        body
    } = await undici.request('http://url', {'headersPreserveCase': true});
````

The rawHeaders route might be easier since it could be an array with duplicates.

mcollina:
Would you like to send a Pull Request to address this issue? Remember to add unit tests.

MoLow:
@chasetec @mcollina I'l take a look into this

PR #1240, opened 2022-02-17
feat: pass rawHeaders

MoLow:
add rawHeaders whenever we return headers.
hope this PR makes sense
closes https://github.com/nodejs/undici/issues/1213

### DIFF
--- a/lib/core/util.js
+++ b/lib/core/util.js
@@ -200,6 +200,10 @@ function parseHeaders (headers, obj = {}) {
   return obj
 }
 
+function parseRawHeaders (headers) {
+  return headers.map(header => header.toString());
+}
+
 function isBuffer (buffer) {
   // See, https://github.com/mcollina/undici/pull/319
   return buffer instanceof Uint8Array || Buffer.isBuffer(buffer)
@@ -339,6 +343,7 @@ module.exports = {
   isIterable,
   isAsyncIterable,
   isDestroyed,
+  parseRawHeaders,
   parseHeaders,
   parseKeepAliveTimeout,
   destroy,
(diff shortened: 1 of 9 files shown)


ronag (commented):
I'm concerned that this causes reduced performance for an edge case.

ronag:
Can we make this an option on the request instead? i.e. whether or not to return raw headers. That way it doesn't affect performance and you get exactly what you want in the fastest way possible.

e.g.

```js
    const {
        statusCode,
        headers,
        trailers,
        body
    } = await undici.request('http://url', { headers: 'raw' });
```

MoLow:
> Can we make this an option on the request instead? i.e. whether or not to return raw headers. That way it doesn't affect performance and you get exactly what you want in the fastest way possible.
> 
> e.g.
> 
> ```js
>     const {
>         statusCode,
>         headers, // rawHeaders
>         trailers,
>         body
>     } = await undici.request('http://url', { responeHeaders: 'raw' });
> ```
> 
> I'm not sure what the option should be called but I think something like this.

@ronag  makes sense to me,

mcollina (approved):
lgtm

lib/core/util.js
```diff
@@ -200,6 +200,10 @@ function parseHeaders (headers, obj = {}) {
   return obj
 }
 
+function parseRawHeaders (headers) {
+  return headers.map(header => header.toString());
+}
```
ronag, line 205: Do we even need this? I think it's fine to leave this conversion to user space, no? That way we don't have to add another option when someone wants the "raw" headers as buffers.

Issue #1225, opened 2022-02-11
Fetch options do not allow explicit `undefined`


chlonel:
## Bug Description

It's hard to say for sure that this is a bug, but it feels like unintended behavior, so I'm reporting it as a bug.

Passing `redirect: undefined` as an option of `fetch` will throw an error.
At the time of executing the following code, `init.redirect` may be assigned `undefined`, which will be considered as an invalid value.
This does not only happen with `redirect`, but also with other options such as `method` and `cache`.

```javascript
// 22. If init["redirect"] exists, then set request’s redirect mode to it.
if ('redirect' in init) {
  request.redirect = init.redirect
  if (!requestRedirect.includes(request.redirect)) {
    throw new TypeError(
      `Failed to construct 'Request': The provided value '${request.redirect}' is not a valid enum value of type RequestRedirect.`
    )
  }
}
```

## Reproducible By

```javascript
fetch("https://github.com", { redirect: undefined });
// TypeError: Failed to construct 'Request': The provided value 'undefined' is not a valid enum value of type RequestRedirect.
```

## Expected Behavior

If `undefined` is passed, it will simply be ignored and the default value will be assigned.
Although `undefined` should not be passed by right, it is useful if it works as well, since Google Chrome and node-fetch implementations do not throw errors.
## Environment

Manjaro 21.2.1
Node.js v16.13.2

ronag:
What does chrome do?

ronag:
Good spot. I think we should replace `x in init` with `init[x] != null` everywhere. Feel free to open a PR.

chlonel:
There is a difference between Chrome and node-fetch.
Chrome does not allow `redirect: null`, while node-fetch ignores it as well as `undefined`.
I think we should only allow `undefined` if we want to be compatible with Chrome, but which one is better?

mcollina:
What does the spec says?

mcollina:
I would treat a field with undefined value and no field in the same way.

ronag:
Also I think Chrome treats `null` the same way as well...

chlonel:
When I run the following code in the console, it throws an error.
Unfortunately, I don't think it is treated the same way.
```javascript
fetch('https://github.com', { redirect: null })
// Uncaught (in promise) TypeError: Failed to execute 'fetch' on 'Window': Failed to read the 'redirect' property from 'RequestInit': The provided value 'null' is not a valid enum value of type RequestRedirect.
//     at <anonymous>:1:1
```

[...]

PR #1242, opened 2022-02-19
fix: `RequestInit` doesn't allow explicit `undefined`

chlonel:
Fixes: https://github.com/nodejs/undici/issues/1225

### DIFF
--- a/lib/fetch/request.js
+++ b/lib/fetch/request.js
@@ -127,12 +127,12 @@ class Request {
     }
 
     // 10. If init["window"] exists and is non-null, then throw a TypeError.
-    if ('window' in init && window != null) {
+    if (init.window !== undefined && window != null) {
       throw new TypeError(`'window' option '${window}' must be null`)
     }
 
     // 11. If init["window"] exists, then set window to "no-window".
-    if ('window' in init) {
+    if (init.window !== undefined) {
       window = 'no-window'
     }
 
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

lib/fetch/request.js
```diff
@@ -127,12 +127,12 @@ class Request {
     }
 
     // 10. If init["window"] exists and is non-null, then throw a TypeError.
-    if ('window' in init && window != null) {
+    if (init.window !== undefined && window != null) {
```
targos, line 130: shouldn't this have `init.window != null` ?

ronag: @chlonel can you fix?

ronag: Actually it doesn't matter here. It's the same.

targos: It's not. `window` is a local variable created a few lines above (`let window = 'client'`).

targos: > If init["window"] exists and is non-null, then throw a TypeError.

To me, this means that we should check if `init["window"]` in non-null, not `window`. But maybe the spec or the comment is wrong.

ronag: Ah, sorry I missed that

ronag: [Fixed](https://github.com/nodejs/undici/commit/ec2813498dac2534a53590b5e81d92c3ce334597)

PR #1243, opened 2022-02-19
feat: implement `makePolicyContainer` and `clonePolicyContainer`

suneettipirneni:
Implements:
- https://html.spec.whatwg.org/multipage/origin.html#policy-container
- https://html.spec.whatwg.org/multipage/origin.html#clone-a-policy-container

### DIFF
--- a/lib/fetch/util.js
+++ b/lib/fetch/util.js
@@ -274,13 +274,16 @@ function createOpaqueTimingInfo (timingInfo) {
 // https://html.spec.whatwg.org/multipage/origin.html#policy-container
 function makePolicyContainer () {
   return {
+    // A CSP list, which is a CSP list. It is initially empty. https://w3c.github.io/webappsec-csp/#csp-list
     CSPList: [],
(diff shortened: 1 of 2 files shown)


mcollina (changes requested):
What's the purpose of these? Isn't it better than we throw?

lib/fetch/util.js
```diff
@@ -273,14 +273,36 @@ function createOpaqueTimingInfo (timingInfo) {
...
+  return {
+    CSPList: [],
+    embedderPolicy: {
+      value: 'unsafe-none',
+      reportingEndpoint: '',
+      reportOnlyValue: 'unsafe-none',
+      reportOnlyReportingEndpoint: ''
+    },
+    referrerPolicy: 'strict-origin-when-cross-origin'
+  }
```
ronag, line 285: This is missing spec comments

suneettipirneni:
> What's the purpose of these? Isn't it better than we throw?

Sorry, I'm unsure what you mean by this. Can you explain?

ronag (approved):

mcollina:
What is the purpose of those functions in the context of undici? We do not have a CSP in Node.js, so why are they here?
IMHO we should not have those functions or throw in them when called.

PR #1244, opened 2022-02-21
Allow string[] as Record value on HeadersInit

dgadelha:
This is a TypeScript-only change, undici already handles Records with string[] values: https://runkit.com/embed/972d5c6lsor2

A test was created to ensure this feature works properly

### DIFF
--- a/test/node-fetch/headers.js
+++ b/test/node-fetch/headers.js
@@ -279,4 +279,24 @@ describe('Headers', () => {
     // eslint-disable-next-line quotes
     expect(format(headers)).to.equal("{ a: [ '1', '3' ], b: '2', host: 'thehost' }")
   })
+
+  it('shoud accept headers as objects with array values', () => {
+    const headers = new Headers({
+      a: ['1', '2'],
+      b: ['3', '4'],
+      c: '5'
+    })
+    expect(headers).to.have.property('forEach')
+
+    const result = []
+    for (const [key, value] of headers.entries()) {
+      result.push([key, value])
+    }
+
+    expect(result).to.deep.equal([
+      ['a', '1,2'],
+      ['b', '3,4'],
+      ['c', '5']
+    ])
+  })
 })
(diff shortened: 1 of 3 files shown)


ronag (approved):

types/fetch.d.ts
```diff
@@ -42,7 +42,7 @@ export interface BodyMixin {
   readonly text: () => Promise<string>
 }
 
-export type HeadersInit = string[][] | Record<string, string> | Headers
+export type HeadersInit = string[][] | Record<string, string | string[]> | Headers
```
targos, line 45: Should it be `Readonly<string>`? I think the array is not mutated?

mcollina (approved):
lgtm

types/fetch.d.ts
dgadelha: It shouldn't really make any much difference, and I also just kept the same behavior it was using on the same line (no `Readonly` on `string[][]`)

test/node-fetch/headers.js
```diff
@@ -279,4 +279,24 @@ describe('Headers', () => {
     // eslint-disable-next-line quotes
     expect(format(headers)).to.equal("{ a: [ '1', '3' ], b: '2', host: 'thehost' }")
   })
+
+  it('shoud accept headers as objects with array values', () => {
```
Ethan-Arrowood, line 283: Can we move this test to a non node-fetch suite? I'm worried we may overright this in the future if we ever resync with node-fetch.

Ethan-Arrowood (changes requested):
Type change is fine but the test should exist somewhere else I think

types/fetch.d.ts
targos: The difference is that it allows to pass a readonly array. It was a non-blocking suggestion, though.

dgadelha: Makes sense, less restrictive. Should I also adjust `string[][]`?

test/node-fetch/headers.js
dgadelha: done :)

Ethan-Arrowood (approved):

Issue #1245, opened 2022-02-21
undici.fetch does not append url fragments (url.hash)


jennyEckstein:
## Bug Description

At the moment undici.fetch does not support GET requests with fragments

## Reproducible By

See repository to reproduce [here](https://github.com/jennyEckstein/undici.fetch-url-hash)

## Expected Behavior

undici.fetch should be able to handle fragments in url.
Fix: add `url.hash` [here](https://github.com/nodejs/undici/blob/main/lib/fetch/index.js#L1811)

## Logs & Screenshots

See repo ☝️

## Environment

macOS Catalina 10.15.7
Node v16.10.0

### Additional context
I would be happy to open PR.

ronag:
PR welcome

ronag:
I think we have the same problem in all undici api's

targos:
I thought that the fragment should not be sent to the remote server?

MoLow:
taking a look

ronag:
> I thought that the fragment should not be sent to the remote server?

I actually don't know. Anyone know what the spec says?

jennyEckstein:
> I thought that the fragment should not be sent to the remote server?

@targos I think you are right, I found this in wiki.
> Clients are not supposed to send URI fragments to servers when they retrieve a document, and without help from a local application (see below) fragments do not participate in [HTTP](https://en.wikipedia.org/wiki/Hypertext_Transfer_Protocol) redirections.
I am testing another approach I can take.

jennyEckstein:
Upon further investigation, I can confirm that undici.fetch works correctly. Server receives path with no fragment (as it is supposed to). The issue is strictly related to the use of MockAgent and the way it stores and compares intercepted path [here](https://github.com/nodejs/undici/blob/main/lib/mock/mock-utils.js#L16). The values must be strictly equal and do not account for stripping fragments or query strings. This function needs to be modified to behave just like it would with real server.

galvez:
I think this interpretation is somewhat dubious. In the Fetch standard, it says it should with a work a URL as per its definition, and its definitions [includes fragments](https://url.spec.whatwg.org/#concept-url). So maybe the URI spec is outdated in relation to the Fetch spec? cc @ronag

ronag:
What does Chrome do?

galvez:
Chrome's fetch() doesn't send #hash indeed. I think we can safely close this then.

mcollina:
[...]

PR #1249, opened 2022-02-25
fix: include URL hash in fetch calls

galvez:
Fixes https://github.com/nodejs/undici/issues/1245


mcollina (approved):
lgtm

ronag (changes requested):
I don’t think the hash fragment is supposed to be sent to the server?

galvez:
See my comment — think this is potentially still warranted.

mcollina:
> See my comment — think this is potentially still warranted.

which comment?

PR #1252, opened 2022-02-25
feat: normalize origin for BalancedPool addUpstream/removeUpstream

### LINKED ISSUE #1250: Normalize BalancedPool addUpstream / removeUpstream
## This would solve...

If I call `pool.addUpstream("http://localhost:80")`, I'd expect to later be able to call `pool.removeUpstream("http://localhost:80")` using the same URL but it silently fails and the upstream continues to exist.

The issue is that `addUpstream` passes the URL through to the Pool constructor which uses `parseOrigin` to normalize it, while `removeUpstream` just looks up the pool directly using the origin URL passed in. In the above case, the passed string is `http://localhost:80` while the normalized string is `http://localhost`.

Additionally, while `addUpstream` has a [check to ensure we don't add the same backend twice](https://github.com/nodejs/undici/blob/main/lib/balanced-pool.js#L52), this check is bypassed when the  passed URL and normalized origin do not match.

We might consider having `addUpstream` and `removeUpstream` call `parseOrigin` to normalize the URL.

## I have also considered...

One potential downside to this approach would be if someone was setting a `kFactory` that expected non-normalized origin URLs. I can't think of a great way to mitigate that, though I also think undici doesn't intend to have factories modify origins given the implementation. One (probably undesirable) way might be to store the `rawOrigin` property on the Pool and use that for upstream deduplication/removal.

This issue isn't blocking my use case; knowing how `addUpstream` and `removeUpstream` are implemented, I can call `parseOrigin` myself as a workaround. It was just unexpected that the nodes were not removed and no warnings/errors were returned.

## Additional context

I was running into this issue while trying to pair an undici BalancedPool with service discovery using the Consul node client. This involves keeping a list of upstreams, periodically polling Consul for upstreams added/removed, calculating the delta, and calling `addUpstream`/`removeUpstream`. This initially appeared to be working as expected but testing showed that upstreams were not removed properly.

If this change sounds acceptable I would be happy to open a PR.

mnutt:
Fixes #1250.

Calls `parseOrigin` on upstream addition/removal to ensure that finding an existing upstream by `origin` works when deduplicating/removing an upstream.

### DIFF
--- a/lib/balanced-pool.js
+++ b/lib/balanced-pool.js
@@ -14,6 +14,7 @@ const {
 } = require('./pool-base')
 const Pool = require('./pool')
 const { kUrl } = require('./core/symbols')
+const { parseOrigin } = require('./core/util')
 const kFactory = Symbol('factory')
 
 const kOptions = Symbol('options')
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

Issue #693, opened 2021-04-07
runtime performance metrics


ronag:
What do we need for introspection and metrics?

I propose the following events:

- 'pending' - on request queued
- 'running' - on request running
- 'complete'- on request complete
- 'failed' - on request failed

Refs: https://github.com/nodejs/undici/discussions/673

szmarczak:
1. `pending` -> `queued` (in order to avoid confusing w/ `running`)
2. Does the `failed` event have an error argument?

ronag:
> 1. `pending` -> `queued` (in order to avoid confusing w/ `running`)

Fixed:

> 2. Does the `failed` event have an error argument?

Yes. Alternatively we could combine complete/failed and have an err argument.

mcollina:
I think we might want to provide a "stats provider" of sort that could be attached to a Agent or a Dispatcher so that we could track all those metrics automatically - and potentially be easily collectable by prometheus.

artur-ma:
Since undici does not use node agent, I guess it means tools like Newrelic will not track network requests correctly
Will this be enough for such tools as NR?

mcollina:
> Since undici does not use node's built-in agent, I guess it means tools like Newrelic will not track network requests correctly

Exactly. We should look into providing some hooks for APMs in the future. 

> Will this be enough for such tools as NR?

I do not think so. This issue will provide metrics about how the process operates - data that is relatively difficult to extract from Node.js own client.

NR (and all others) will need a way to inject custom headers for each request.

ronag:
> NR (and all others) will need a way to inject custom headers for each request.

That should be reasonably simple. Just extend `Agent` and pass another handler wrapping the user's handler. Similar to how we do redirects.

artur-ma:
It looks great, but currently, as far as I can see, none of the methods on the handler can modify the request (headers).

```
  export interface DispatchHandlers {
    /** Invoked before request is dispatched on socket. May be invoked multiple times when a request is retried when the request at the head of the pipeline fails. */
    onConnect?(abort: () => void): void;
    /** Invoked when an error has occurred. */
    onError?(err: Error): void;
    /** Invoked when request is upgraded either due to a `Upgrade` header or `CONNECT` method. */
    onUpgrade?(statusCode: number, headers: string[] | null, socket: Duplex): void;
    /** Invoked when statusCode and headers have been received. May be invoked multiple times due to 1xx informational headers. */
[...]

PR #1255, opened 2022-02-26
feat: expose pool aggregate statistics

mnutt:
## This relates to...

In achieving parity with the node.js agent, I'd like to be able to get aggregate statistics about requests and connections in undici. This is possible today, but requires using private symbols.

Addresses #1254, relates to #693 -- I just now noticed that one, so will review it to see if there's any feedback that can be applied here.

## Rationale

I currently use the node.js agent and send its stats to prometheus; I'd like to send the same ones for undici to be able to instrument the switchover and ensure everything is working properly.

## Changes

* Expose a `stats` getter on pools.

### Features

* Access aggregate statistics about undici Pools via a `stats` object.

### Bug Fixes

N/A

N/A

## Status

- [x] Tested
  - I didn't benchmark it because the new methods aren't in the normal request path and are only intended to be called periodically; eyeballing it I believe it should scale roughly O(n) with number of active requests.
- [x] Documented
- [] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/docs/api/BalancedPool.md
+++ b/docs/api/BalancedPool.md
@@ -34,6 +34,10 @@ Implements [Client.closed](Client.md#clientclosed)
 
 Implements [Client.destroyed](Client.md#clientdestroyed)
 
+### `Pool.stats`
+
+Returns [`PoolStats`](PoolStats.md) instance for this pool.
+
 ## Instance Methods
 
 ### `BalancedPool.addUpstream(upstream)`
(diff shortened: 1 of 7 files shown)


mnutt:
That's an interesting idea, I guess in metrics terms these are more like gauges, while diagnostics channels are counters. I think it's possible that one could use diagnostic channels counters to emulate some of these by doing some bookkeeping, though I think it might require quite a few more diagnostic channels since some of these are overlapping?

mcollina (approved):
lgtm

ronag (commented):
I still think for best performance and reactive API adding more diagnostic channels and implementing this on top of that is a better idea.

ronag:
@mnutt Do you think you could add something similar to `Client` and maybe `Agent`?

PR #1258, opened 2022-03-01
Fix for ignoring explicit empty host header values

### LINKED ISSUE #1257: A host header value of the empty string is ignored
## Bug Description

As noted, `undici` will ignore the specified host header value and instead will use the host derived from the remote connection target.

## Reproducible By

```js
const Assert = require('node:assert');
const Http = require('node:http');
const Events = require('node:events');
const Undici = require('undici');

async function main() {
  const server = Http.createServer((req, res) => {
    res.writeHead(200, {
      'content-type': 'text/plain',
    });
    res.end(req.headers['host']);
  });

  server.listen(0, 'localhost');

  await Events.once(server, 'listening');

  const address = server.address();
  const serverHost = `${address.address}:${address.port}`;

  const check = async (host) => {
    const res = await Undici.request(`http://${serverHost}`, {
      headers: {
        host,
      },
    });
    return res.body.text();
  };

  try {
    Assert.strictEqual(await check('good'), 'good');
    Assert.strictEqual(await check(undefined), serverHost);
    Assert.strictEqual(await check(''), '');
  } finally {
    server.close();
  }
}

main().catch((err) => {
  console.error(err);
  process.exit(1);
});
```

## Expected Behavior

If a value is specified for the `host` header that is a string, it should be respected even if it is empty.

This could probably be solved by changing the following line to be `if (typeof host === 'string') {`: https://github.com/nodejs/undici/blob/901f9d6f93322e45285868299f2689acf27ef1ac/lib/client.js#L1484

## Logs & Screenshots

N/A

## Environment

node@16.14.0

### Additional context

This issue was detected when using `undici` as the client in a node-based reverse proxy while running integration tests for `CVE-2019-15606` based on the [tests in node core](https://github.com/nodejs/node/commit/2eee90e959).

ggoodman:
## This relates to...

#1257

## Rationale

Includes a failing test and a suggestion for how that test can be made to pass.

## Changes

- Added test confirming that the correct host header is being sent when: 1) it is a non-empty string; 2) it is `undefined`; and 3) it is an empty string.
- Added a change to distinguish empty strings from other falsy values when building the request.

### Features

N/A

### Bug Fixes

Fixes #1257

Anyone relying on being able to explicitly set a header to an empty string and have that default to the target host will observe a change in behaviour.

## Status

- [x] Tested
- [S] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1481,7 +1481,7 @@ function write (client, request) {
 
   let header = `${method} ${path} HTTP/1.1\r\n`
 
-  if (host) {
+  if (typeof host === 'string') {
     header += `host: ${host}\r\n`
   } else {
     header += client[kHostHeader]
--- a/test/client-request.js
+++ b/test/client-request.js
@@ -281,6 +281,34 @@ test('request text', (t) => {
   })
 })
 
+test('empty host header', (t) => {
+  t.plan(3)
+
+  const server = createServer((req, res) => {
+    res.end(req.headers.host)
+  })
+  t.teardown(server.close.bind(server))
+
+  server.listen(0, async () => {
+    const serverAddress = `localhost:${server.address().port}`;
+    const client = new Client(`http://${serverAddress}`)
+    t.teardown(client.destroy.bind(client))
+
+    const getWithHost = async (host, wanted) => {
+      const { body } = await client.request({
+        path: '/',
+        method: 'GET',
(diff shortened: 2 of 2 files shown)


mcollina (approved):
lgtm

PR #1259, opened 2022-03-01
fix: Improve BodyReadable type definition

### LINKED ISSUE #1256: `body.dump()` is not available in type definitions
## This would solve...

Allow dumping the body if it is not being used. The body type is an intersection between the [`BodyMixin`](https://github.com/nodejs/undici/blob/17e6a0f464b32ef59c27b1b8fe9d37c71fa16c4a/types/dispatcher.d.ts#L160) and [`Readable`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/stream.d.ts#L50).

Create a new interface for `BodyReadable` and extend it from Readable. Include any undici specific add-ons.

I am not sure if it makes sense to add it to the [`BodyMixin`](https://github.com/nodejs/undici/blob/17e6a0f464b32ef59c27b1b8fe9d37c71fa16c4a/types/dispatcher.d.ts#L160) as it represents that the [`fetch` Body mixin](https://fetch.spec.whatwg.org/#body-mixin)

I can try adding this type definition. I just wanted to get some feedback on ideas. If anyone else can do it, feel free.

## I have also considered...

For now, I am just doing `body.text()` to release the connection.

umarov:
Updates the `body` to use `BodyReadable`. Currently, body was using an intersection between `Readable` and `BodyMixin`. 

This actually exposes the `dump` function in TypeScript now.

Let me know if the `resume` and `abort` are actually nullable. I looked at the code and `abort` is `null` initially. `resume` seems to be optional for `Readable`

Fixes #1256

### DIFF
--- a/test/types/readable.test-d.ts
+++ b/test/types/readable.test-d.ts
@@ -1,0 +1,34 @@
+import { expectAssignable } from 'tsd'
+import BodyReadable from '../../types/readable'
+import { Blob } from 'buffer'
+
+expectAssignable<BodyReadable>(new BodyReadable())
+
+{
+  const readable = new BodyReadable()
+
+  // dump
+  expectAssignable<Promise<void>>(readable.dump())
+  expectAssignable<Promise<void>>(readable.dump({ limit: 123 }))
+
+  // text
+  expectAssignable<Promise<string>>(readable.text())
+
+  // json
+  expectAssignable<Promise<any>>(readable.json())
+
+  // blob
+  expectAssignable<Promise<Blob>>(readable.blob())
+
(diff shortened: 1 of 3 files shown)


mcollina (commented):
we tests our types using tsd. Could you add a test?

umarov:
@mcollina does it make sense to add types from the `BodyMixin` into this `BodyReadable`? I tried adding type tests for those, but the class type I created does not have them. I saw that the `BodyMixin` already has some types. I wonder if it's redundant to have them in two places.

umarov:
I can try pushing up what I have

Issue #1263, opened 2022-03-04
Mac OS X runners always fail


mcollina:
It seems Mac OS X runners are always failing with the current setup.

![Screenshot 2022-03-04 at 16 31 45](https://user-images.githubusercontent.com/52195/156793020-4a753b5a-1dfe-4511-95fd-602ddf621a1a.png)]

https://github.com/nodejs/undici/actions/runs/1934500731

cc @dominykas @nodejs/build could you please take a look?

richardlau:
Potentially linked to 
- https://github.community/t/ubuntu-latest-runner-not-found-when-using-workflow-call-between-repositories/204957
- https://github.community/t/called-workflows-cannot-be-queued-onto-self-hosted-runners-across-organisations-enterprises-failed-to-queue-this-job-labels-ubuntu-latest/229355/9

PR #1265, opened 2022-03-04
chore: add back macos to Node CI

richardlau:
Match the case of the `macos-latest` workflow label.

Refs: https://github.com/nodejs/undici/issues/1263
Refs: https://github.com/nodejs/undici/pull/1264
Refs: https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources


richardlau:
Apparently that didn't work 😞 .

PR #1273, opened 2022-03-11
fix: remove hash from intercept dispatch key

galvez:
See https://github.com/nodejs/undici/issues/1245

As per RFC 3986, clients are not supposed to send URI  fragments to servers when they retrieve a document.


ronag (approved):

mcollina (approved):
lgtm

PR #1274, opened 2022-03-11
feat: allow mock matching missing headers

### LINKED ISSUE #1269: Mocks should allow to match headers that are meant to be missing
## This would solve...

Being able to match only when a given header is not present.

I dunno, ...
- prefix the header name with a unique symbol outside the range of allowed header names to indicate it's meant to be missing in a to be matched request?
- special symbol instead of a string|regex|function as a matcher to be provided
- function matcher would get undefined passed as a value?

## I have also considered...

I've tried to match the missing header with a function that checks that the value is undefined but the matcher doesn't even run because the header is not in the request.

panva:
resolves #1269

Seeking feedback and help on making sure the change is properly propagated to the type declarations.

### DIFF
--- a/index.js
+++ b/index.js
@@ -15,6 +15,7 @@ const MockAgent = require('./lib/mock/mock-agent')
 const MockPool = require('./lib/mock/mock-pool')
 const mockErrors = require('./lib/mock/mock-errors')
 const ProxyAgent = require('./lib/proxy-agent')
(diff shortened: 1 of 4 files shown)


panva:
Come to think or it null or undefined as match value would also work.

panva:
Actually, like body I think it would make sense to allow matching headers as a whole rather than individual ones. WDYT @mcollina

PR #1275, opened 2022-03-11
feat(mock): allow matching the complete header list

### LINKED ISSUE #1269: Mocks should allow to match headers that are meant to be missing

panva:
resolves #1269 and (subjectively) a preferred alternative to #1274

### DIFF
--- a/types/mock-interceptor.d.ts
+++ b/types/mock-interceptor.d.ts
@@ -49,7 +49,7 @@ declare namespace MockInterceptor {
     /** Body to intercept on. */
     body?: string | RegExp | ((body: string) => boolean);
     /** Headers to intercept on. */
-    headers?: Record<string, string | RegExp | ((body: string) => boolean)> | ((headers: Headers) => boolean)
+    headers?: Record<string, string | RegExp | ((body: string) => boolean)> | ((headers: Record<string, string>) => boolean)
   }
   export interface MockDispatch<TData extends object = object, TError extends Error = Error> extends Options {
     times: number | null;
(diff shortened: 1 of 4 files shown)


panva:
Might need to use a regular object instead of Headers for node < 16 compat. Happy to do so given otherwise non blocking feedback.

mcollina:
code looks great!

mcollina (approved):
lgtm

types/mock-interceptor.d.ts
```diff
@@ -49,7 +49,7 @@ declare namespace MockInterceptor {
     /** Body to intercept on. */
     body?: string | RegExp | ((body: string) => boolean);
     /** Headers to intercept on. */
-    headers?: Record<string, string | RegExp | ((body: string) => boolean)>;
+    headers?: Record<string, string | RegExp | ((body: string) => boolean)> | ((headers: Record<string, string>) => boolean);
```
mcollina, line 52: Could you add an assertion for this type? We use tsd?

panva: Sure, where?

mcollina: test/types/mock-interceptor.test-d.ts

mcollina (approved):
lgtm

PR #1278, opened 2022-03-14
refactor: unify close/destroy

ronag:


lib/dispatcher-base.js
```diff
@@ -0,0 +1,158 @@
...
+        })
+      })
+    }
+
+    if (typeof callback !== 'function') {
+      throw new InvalidArgumentError('invalid callback')
+    }
+
+    if (this[kDestroyed]) {
+      queueMicrotask(() => callback(new ClientDestroyedError(), null))
```
ronag, line 47: I don't think this was an error before (hence the updates to test). Not sure what is most correct.

ronag:
@mcollina

ronag:
Benchmarks have only one sample so I don't think they are very accurate but it looks good to me.

ronag:
I'm good with semver major

mcollina (approved):
lgtm as semver-major

Issue #1011, opened 2021-08-31
Documentation: RequestOptions


rudolfbyker:
## Bug Description

Documentation about RequestOptions is inconsistent.

## Reproducible By

Comparing the following:

https://github.com/nodejs/undici#undicirequesturl-options-promise
![image](https://user-images.githubusercontent.com/10025342/131479955-832006c1-e304-4613-ac55-1630d6f4e6e9.png)

https://github.com/nodejs/undici/blob/main/docs/api/Dispatcher.md#parameter-requestoptions
![image](https://user-images.githubusercontent.com/10025342/131480063-0fe61374-11c8-422a-94be-b85a7227ffbf.png)

https://github.com/nodejs/undici/blob/main/docs/api/Dispatcher.md#parameter-dispatchoptions
![image](https://user-images.githubusercontent.com/10025342/131480139-b267e50b-3d12-4480-953a-c2cd82cb9dbd.png)

https://github.com/nodejs/undici/blob/main/types/api.d.ts#L16
![image](https://user-images.githubusercontent.com/10025342/131480249-601340cf-e8fc-42a2-9cb5-dd1e8e127ff0.png)

## Expected Behavior

Expected consistent types in code and in documentation. Specifically, are we allowed to specify `origin` and `path` in the options object passed as the second argument to `request` and friends? Why or why not?

PR #1282, opened 2022-03-16
fix: throws on unsupported option `path`

is2ei:
Refs: #1011

`undici.request` accepts `path` option but the type definition omits `path`.
POC: https://github.com/is2ei/POC-nodejs-undici-1011
code: https://github.com/nodejs/undici/blob/d7eac3e4e57c387bfd954da7a28e84138c243d4c/index.js#L63-L75

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,9 +106,8 @@ This section documents our most commonly used API methods. Additional APIs are d
 Arguments:
 
 * **url** `string | URL | UrlObject`
-* **options** [`RequestOptions`](./docs/api/Dispatcher.md#parameter-requestoptions)(`origin` is omitted)
+* **options** [`RequestOptions`](./docs/api/Dispatcher.md#parameter-requestoptions)(`path` and `origin` are omitted)
   * **dispatcher** `Dispatcher` - Default: [getGlobalDispatcher](#undicigetglobaldispatcher)
-  * **path** `String` (optional)
   * **method** `String` - Default: `PUT` if `options.body`, otherwise `GET`
   * **maxRedirections** `Integer` - Default: `0`
 
--- a/README.md
+++ b/README.md
@@ -106,8 +106,9 @@ This section documents our most commonly used API methods. Additional APIs are d
 Arguments:
 
 * **url** `string | URL | UrlObject`
-* **options** [`RequestOptions`](./docs/api/Dispatcher.md#parameter-requestoptions)
+* **options** [`RequestOptions`](./docs/api/Dispatcher.md#parameter-requestoptions)(`origin` is omitted)
   * **dispatcher** `Dispatcher` - Default: [getGlobalDispatcher](#undicigetglobaldispatcher)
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -92,8 +92,9 @@ This section documents our most commonly used API methods. Additional APIs are d
 Arguments:
 
 * **url** `string | URL | UrlObject`
-* **options** [`RequestOptions`](./docs/api/Dispatcher.md#parameter-requestoptions)
+* **options** [`RequestOptions`](./docs/api/Dispatcher.md#parameter-requestoptions)(`origin` is omitted)
   * **dispatcher** `Dispatcher` - Default: [getGlobalDispatcher](#undicigetglobaldispatcher)
+  * **path** `String` (optional)
```
ronag, line 97: `path` is already part of `RequestOptions`

ronag (changes requested):
path is already part of these through extending parent structure

is2ei:
@ronag 
Thanks for your comment. I basically try to solve 2 problems related to #1011.

1. wrong type definition
2. document for `undici.request` lacks some information.

When using TypeScript, you can not pass the `path` option to `undici.request` because the type definition omits the `path` option. But when using JavaScript, you can pass the `path` option to `undici.request` and it works. So I fixed the type definition.

https://github.com/nodejs/undici/blob/d7eac3e4e57c387bfd954da7a28e84138c243d4c/index.js#L63-L68

Also, I found `method` option is documented even though `DispatchOptions` already has it. I guess the difference is `method` is mandatory in `DispatchOptions` but optional in `undici.request` option. `path` option has the same situation so I documented it.

https://github.com/nodejs/undici/blob/d7eac3e4e57c387bfd954da7a28e84138c243d4c/types/api.d.ts#L13-L17

ronag:
> When using TypeScript, you can not pass the path option to undici.request because the type definition omits the path option. But when using JavaScript, you can pass the path option to undici.request and it works. So I fixed the type definition.

I think the typescript version is correct and reflects intended/recommended usage.

is2ei:
@ronag 

> Maybe we should throw?

Ok. I'll try.

is2ei:
@ronag 
I changed the code. Could you review it?
I basically agree with your opinion,  but I'd be worried if it would be a breaking change.

hebertcisco (changes requested):
I saw that your pull request has conflicts, have you tried to resolve these conflicts?

hebertcisco (changes requested):
I saw that your pull request has conflicts, have you tried to resolve these conflicts?

hebertcisco (commented):
@ronag What do you think of this pull request and the changes made to it? Would it be mergeable and adequate?

is2ei:
@hebertcisco 
I fixed the conflict, but it looks `path` option is actually used now.
Do we still need this change?

mcollina:
I think this can be closed, thanks!

PR #1285, opened 2022-03-16
fix: default mock interceptor to GET

JustinBeckwith:
I am in the process of migrating an open source project from `node-fetch` to `undici`.  I am making heavy use of `nock`, which doesn't exactly work here :) One of the things I noticed is that nock tends to default to assuming `GET` on interceptors, and I think it's a fairly accepted norm in terms of request libraries.  This change makes `method` optional on the `MockClient.intercept` method, and defaults it to a `GET` in that case.  

```js
const mock = new MockAgent();
const scope = mock
      .get('http://fake.local')
      .intercept({path: '/'}) // look ma, no {method: 'GET'}!
      .reply(200);
```

Not sure if this is a desirable API change or not, but thought it would aid in making the transition from `nock` to here a tad easier.

### DIFF
--- a/types/mock-interceptor.d.ts
+++ b/types/mock-interceptor.d.ts
@@ -44,8 +44,8 @@ declare namespace MockInterceptor {
   export interface Options {
     /** Path to intercept on. */
     path: string | RegExp | ((path: string) => boolean);
-    /** Method to intercept on. */
-    method: string | RegExp | ((method: string) => boolean);
+    /** Method to intercept on. Defaults to GET. */
+    method?: string | RegExp | ((method: string) => boolean);
     /** Body to intercept on. */
     body?: string | RegExp | ((body: string) => boolean);
     /** Headers to intercept on. */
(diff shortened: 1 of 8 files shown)


types/mock-interceptor.d.ts
```diff
@@ -44,8 +44,8 @@ declare namespace MockInterceptor {
   export interface Options {
     /** Path to intercept on. */
     path: string | RegExp | ((path: string) => boolean);
-    /** Method to intercept on. */
-    method: string | RegExp | ((method: string) => boolean);
+    /** Method to intercept on. Defaults to GET. */
+    method?: string | RegExp | ((method: string) => boolean);
```
mcollina, line 48: Can you add a test for the type? We use tsd!

mcollina:
could you also update the docs?

JustinBeckwith:
Updated docs and added a type test!  Sorry for the delay.

mcollina (approved):
lgtm

PR #1288, opened 2022-03-17
chore: fix duplicated error import in test/agent.js

is2ei:
This PR refactors `test/agent.js`.

- It imports `../` and `../index` but those are the same.
- It uses `InvalidArgumentError` and `errors.InvalidArgumentError` but those are the same.


ronag (approved):

mcollina (approved):
lgtm

PR #1290, opened 2022-03-18
fix: Do not use WeakRef on mocks.

ShogunPanda:
This PR fixes a very hard to spot bug that happens under these conditions:

1. You use mocks and setup some interceptors.
2. The pool/client (`mockAgent.get`) is setup within a function and no reference to the pool/client is returned.
3. A lot of request (same or different url doesn't matter) are performed and a lot of data is returned.
4. Some time passes (either with setTimeout or just with regular application I/O delays)

What happens is that the original pool/client is garbage collected and thus all the set interceptors are gone, leading to either test failure or unwanted network activity.

The PR fixes that by using a `FakeWeakRef` in the `MockAgent` that keeps a reference to the pool and thus no garbage collection happens.

### DIFF
--- a/lib/mock/mock-agent.js
+++ b/lib/mock/mock-agent.js
@@ -20,6 +20,16 @@ const { InvalidArgumentError } = require('../core/errors')
 const Dispatcher = require('../dispatcher')
 const { WeakRef } = require('../compat/dispatcher-weakref')()
 
+class FakeWeakRef {
+  constructor (value) {
+    this.value = value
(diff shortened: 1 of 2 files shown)


ronag (approved):

ShogunPanda:
@ronag I fixed linting errors. I saw some failures somewhere else. Are them related to my fix or is the CI flaky?

mcollina (approved):
lgtm

PR #1295, opened 2022-03-20
fix(fetch): spec

ronag:


ronag:
@KhafraDev interested in helping out with this one?

lib/fetch/index.js
```diff
@@ -928,33 +929,77 @@ function fetchFinale (fetchParams, response) {
...
+      transform: identityTransformAlgorithm,
+      flush: processResponseEndOfBody
+    })
 
-  // TODO: This is a workaround. Until the above has been implemented, i.e.
-  // we need to either fully consume the body or terminate the fetch.
-  if (response.type === 'error') {
-    context.terminate({ reason: response.error })
+    // 4. Set response’s body to the result of piping response’s body through transformStream.
+    // TODO
```
ronag, line 975: @KhafraDev help here?

test/fetch/data-uri.js
```diff
@@ -1,215 +1,215 @@
-'use strict'
```
ronag, line 1: @KhafraDev help here? these fail...

test/node-fetch/main.js
```diff
@@ -1347,42 +1347,42 @@ describe('node-fetch', () => {
...
-      const r1 = res.clone()
-      return r1.text().then(result => {
-        expect(result).to.equal('{"name":"value"}')
-        return res.json().then(result => {
-          expect(result).to.deep.equal({ name: 'value' })
-        })
-      })
-    })
-  })
+  // it('should allow cloning a json response and log it as text response', () => {
```
ronag, line 1350: @KhafraDev

test/fetch/data-uri.js
KhafraDev:
```
\undici\lib\fetch\index.js:971
    response.body = { stream: response.body.stream.pipeThrough(transformStream) }
                                                   ^

TypeError: Cannot read properties of undefined (reading 'pipeThrough')
    at fetchFinale (\undici\lib\fetch\index.js:971:52)
    at mainFetch (\undici\lib\fetch\index.js:761:5)

Node.js v17.7.1
```

KhafraDev: `response.body` is a Uint8Array

test/node-fetch/main.js
KhafraDev: stream is being locked somewhere

test/fetch/data-uri.js
KhafraDev: Okay this is quite an easy fix, but difficult to track down (my favorite)!

`dataURLProcessor` returns either a string or a Uint8Array
https://github.com/nodejs/undici/blob/6a0388f38da39538368eed5ef16d218f8510515c/lib/fetch/dataURL.js#L57
https://github.com/nodejs/undici/blob/6a0388f38da39538368eed5ef16d218f8510515c/lib/fetch/dataURL.js#L67

So then, possibly either a spec oversight or my mistake, the body is never extracted:
https://github.com/nodejs/undici/blob/6a0388f38da39538368eed5ef16d218f8510515c/lib/fetch/index.js#L879

Which means it's now a Uint8Array (most likely). The fix is just adding `extractBody(dataURLStruct.body)[0]`. All tests pass as well.

KhafraDev: @ronag

test/node-fetch/main.js
KhafraDev: Specifically just calling `response.clone()` will throw an error. Happens in `cloneBody`'s first step (teeing the stream). No idea how the stream is locked beforehand as it shouldn't be used.

KhafraDev:
```diff
diff --git a/lib/fetch/response.js b/lib/fetch/response.js
index 2e7e42e..53d9794 100644
--- a/lib/fetch/response.js
+++ b/lib/fetch/response.js
@@ -316,12 +316,6 @@ function cloneResponse (response) {
   // 1. If response is a filtered response, then return a new identical
   // filtered response whose internal response is a clone of response’s
   // internal response.
-  if (response.internalResponse) {
-    return filterResponse(
-      cloneResponse(response.internalResponse),
-      response.type
-    )
-  }

   // 2. Let newResponse be a copy of response, except for its body.
   const newResponse = makeResponse({ ...response, body: null })

```

`response.body` and `response.internalResponse.body` are mismatched; response's body's stream is not locked and works properly (yay!), but internalResponse's body's stream is locked, causing the error. The above is NOT a good solution but all tests do pass...

KhafraDev:
```diff
diff --git a/lib/fetch/response.js b/lib/fetch/response.js
index 2e7e42e..9756273 100644
--- a/lib/fetch/response.js
+++ b/lib/fetch/response.js
@@ -316,11 +316,11 @@ function cloneResponse (response) {
   // 1. If response is a filtered response, then return a new identical
   // filtered response whose internal response is a clone of response’s
   // internal response.
-  if (response.internalResponse) {
-    return filterResponse(
-      cloneResponse(response.internalResponse),
-      response.type
-    )
+  if (response.internalResponse != null) {
+    return filterResponse({
+      ...response.internalResponse,
+      body: cloneBody(response.body)
+    }, response.type)
```
This also works but I'm just throwing stuff against the wall and hoping it sticks tbh.

test/fetch/data-uri.js
ronag: @KhafraDev thanks! Seems like a spec bug that it's missing extract body here. @dominic

ronag: @annevk

test/node-fetch/main.js
ronag: I feel like we are miss-reading the spec somewhere somehow...

ronag: I think I figured it out. A filtereredResponse should be a proxy for the response.

lib/fetch/response.js
```diff
@@ -369,6 +370,50 @@ function makeNetworkError (reason) {
...
+            }
+          }
+          return arr
+        }
+      } else {
+        assert(false)
+      }
+    }
+  })
+}
```
ronag, line 415: We could just freeze it but I suspect in the future we will need to be able to mutate the headers list for e.g. set-cookie.

test/fetch/data-uri.js
KhafraDev: This one is on us I'm pretty sure:

```
A body consists of:

*    A stream (a [ReadableStream](https://streams.spec.whatwg.org/#readablestream) object).

*    A source (null, a [byte sequence](https://infra.spec.whatwg.org/#byte-sequence), a [Blob](https://w3c.github.io/FileAPI/#dfn-Blob) object, or a [FormData](https://xhr.spec.whatwg.org/#formdata) object), initially null.

*    A length (null or an integer), initially null.
```

undici only works if the body is a stream, but it should work for all of the other types as well. `dataURLProcessor` returns a `byte sequence` body (Uint8Array in our case).

KhafraDev:
I would assume so, `Proxy` is pretty slow (and looking at the PR one or more are created every request).

ronag:
I’m currently prioritizing correctness and maintenance over perf. For perf I recommend the much fast undici api’s.

That being said. I’m happy to accept pr’s that improve perf as long as it doesn’t conflict with the prior priorities.

PR #1298, opened 2022-03-21
fetch: misc. bug fixes and small changes

KhafraDev:
Remove references to `window`, as there is no window in node's environment.

Remove references to `service workers` are there are none in node.

Remove some utility functions that have no purpose in node (CORs, TAO).

Fix `util.responseURL` returning a string where its single occurrence expected a URL (checked for `url.hash` when `url` is a string).

----

Each PR is separated into its own section to make it easier to follow :D

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -567,7 +567,7 @@ async function mainFetch (fetchParams, recursive = false) {
         // error.
         if (request.redirect !== 'follow') {
           return makeNetworkError(
-            'redirect mode cannot be "follow" for this request'
+            'redirect mode cannot be "follow" for "no-cors" request'
           )
         }
 
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -991,9 +991,6 @@ async function httpFetch (fetchParams) {
   const timingInfo = fetchParams.timingInfo
 
   // 5. If request’s service-workers mode is "all", then:
-  if (request.serviceWorkers === 'all') {
-    // TODO
-  }
 
   // 6. If response is null, then:
   if (response === null) {
@@ -1002,9 +999,6 @@ async function httpFetch (fetchParams) {
 
     // 2. If request’s redirect mode is "follow", then set request’s
     // service-workers mode to "none".
-    if (request.redirect === 'follow') {
-      request.serviceWorkers = 'none'
-    }
 
     // 3. Set response and actualResponse to the result of running
     // HTTP-network-or-cache fetch given fetchParams.
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -14,7 +14,6 @@ const { Request, makeRequest } = require('./request')
 const zlib = require('zlib')
 const {
   ServiceWorkerGlobalScope,
-  Window,
   matchRequestIntegrity,
   makePolicyContainer,
   clonePolicyContainer,
@@ -98,7 +97,7 @@ class Fetch extends EE {
 async function fetch (...args) {
   if (args.length < 1) {
     throw new TypeError(
-      `Failed to execute 'fetch' on 'Window': 1 argument required, but only ${args.length} present.`
+      `Failed to execute 'fetch': 1 argument required, but only ${args.length} present.`
     )
   }
   if (
@@ -107,7 +106,7 @@ async function fetch (...args) {
     args[1] !== undefined
   ) {
     throw new TypeError(
-      "Failed to execute 'fetch' on 'Window': cannot convert to dictionary."
+      "Failed to execute 'fetch': cannot convert to dictionary."
     )
   }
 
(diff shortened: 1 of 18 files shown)


lib/fetch/index.js
```diff
@@ -570,43 +555,31 @@ async function mainFetch (fetchParams, recursive = false) {
...
-            'redirect mode cannot be "follow" for "no-cors" request'
-          )
-        }
+      // 1. If request’s redirect mode is not "follow", then return a network
+      // error.
+      if (request.redirect !== 'follow') {
+        return makeNetworkError(
+          'redirect mode cannot be "follow" for this request'
+        )
+      }
```
ronag, line 566: This is from inside an impossible branch?

lib/fetch/request.js
```diff
@@ -608,18 +550,6 @@ class Request {
...
-  // whether the request will use CORS, or will be restricted to same-origin
-  // URLs.
-  get mode () {
-    if (!(this instanceof Request)) {
-      throw new TypeError('Illegal invocation')
-    }
-
-    // The mode getter steps are to return this’s request’s mode.
-    return this[kState].mode
-  }
```
ronag, line 621: I don’t agree with changes to the public api. That should follow spec strictly.

ronag (changes requested):
I'm sorry but I think most of these changes makes it hard to maintain things. Instead of just following the spec language verbatim now we have to consider whether or not each case is relevant and if we make incorrect assumption we introduce bugs.

lib/fetch/util.js
```diff
@@ -16,13 +16,16 @@ const badPorts = [
...
+ * @returns {null|URL}
+ */
 function responseURL (response) {
   // https://fetch.spec.whatwg.org/#responses
   // A response has an associated URL. It is a pointer to the last URL
   // in response’s URL list and null if response’s URL list is empty.
   const urlList = response.urlList
   const length = urlList.length
-  return length === 0 ? null : urlList[length - 1].toString()
+  return length === 0 ? null : urlList[length - 1]
```
ronag, line 28: This is good change.

lib/fetch/util.js
```diff
@@ -77,7 +80,7 @@ function isFileLike (object) {
     object &&
     (typeof object.stream === 'function' ||
      typeof object.arrayBuffer === 'function') &&
-    /^(File)$/.test(object[Symbol.toStringTag])
+     object[Symbol.toStringTag] === 'File'
```
ronag, line 83: 👍

ronag:
Alternative https://github.com/nodejs/undici/pull/1300

ronag:
I'm fine with replacing the content of impossible branches with an `assert(false)`.

KhafraDev:
I'll revert that commit then 👍

lib/fetch/index.js
```diff
@@ -96,7 +93,7 @@ class Fetch extends EE {
 async function fetch (...args) {
   if (args.length < 1) {
     throw new TypeError(
-      `Failed to execute 'fetch' on 'Window': 1 argument required, but only ${args.length} present.`
+      `Failed to execute 'fetch': 1 argument required, but only ${args.length} present.`
```
ronag, line 96: This is externally observable. I would prefer if the error is the exact same as in Chrome.

lib/fetch/index.js
```diff
@@ -105,7 +102,7 @@ async function fetch (...args) {
     args[1] !== undefined
   ) {
     throw new TypeError(
-      "Failed to execute 'fetch' on 'Window': cannot convert to dictionary."
+      "Failed to execute 'fetch': cannot convert to dictionary."
```
ronag, line 105: This is externally observable. I would prefer if the error is the exact same as in Chrome.

lib/fetch/index.js
```diff
@@ -594,12 +580,9 @@ async function mainFetch (fetchParams, recursive = false) {
 
         // 4. If noCorsResponse is a filtered response or the CORB check with
         // request and noCorsResponse returns allowed, then return noCorsResponse.
-        if (noCorsResponse.status === 0 || CORBCheck(request, noCorsResponse) === 'allowed') {
-          return noCorsResponse
-        }
+        return noCorsResponse
 
         // 5. Return a new response whose status is noCorsResponse’s status.
-        return makeResponse({ status: noCorsResponse.status })
```
ronag, line 602: I think this changes externally observable behavior

lib/fetch/index.js
```diff
@@ -407,13 +400,6 @@ function fetching ({
   // 8. If request’s window is "client", then set request’s window to request’s
   // client, if request’s client’s global object is a Window object; otherwise
   // "no-window".
-  if (request.window === 'client') {
-    // TODO: What if request.client is null?
-    request.window =
-      request.client?.globalObject?.constructor?.name === 'Window'
-        ? request.client
-        : 'no-window'
-  }
```
ronag, line 416: The problem with this kind of change is that it's hard to tell whether some other code indirectly inspects this and causes some externally observable changes... which is why IMHO it's better to just keep it and then we don't have to wonder.

ronag (changes requested):
I'm sorry but it's just to hard to tell whether or not these changes have externally observable side-effects or not. IMHO it's easier to just follow the spec verbatim even if it may not have an relevance in node.

lib/fetch/index.js
KhafraDev: `CORBCheck` would only return `allowed` so noCorsResponse was always being returned.

PR #1299, opened 2022-03-22
fetch: replace `Proxy` objects

KhafraDev:
Should alleviate most perf regressions in #1295, I still have no idea what a filtered response is and it's most likely wrong, but all of the tests pass.

I'll add tests later... it's late here

### DIFF
--- a/lib/fetch/headers.js
+++ b/lib/fetch/headers.js
@@ -2,6 +2,7 @@
 
 'use strict'
 
+const assert = require('assert')
 const { validateHeaderName, validateHeaderValue } = require('http')
 const { kHeadersList } = require('../core/symbols')
 const { kGuard } = require('./symbols')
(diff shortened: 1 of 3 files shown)


lib/fetch/headers.js
```diff
@@ -149,6 +150,60 @@ class HeadersList extends Array {
   }
 }
 
+class FilteredHeadersList extends HeadersList {
+  /** @type {(name: string) => boolean} */
+  filter = () => true;
```
KhafraDev, line 155: Needed `babel-eslint` for standard because it couldn't parse this properly. There's an issue on github with this solution I can pull up if you need.

lib/fetch/headers.js
```diff
@@ -149,6 +150,60 @@ class HeadersList extends Array {
...
+
+  /**
+   * @param {HeadersList} headersList
+   * @param {typeof this['filter'] | undefined} filter
+   */
+  constructor (headersList, filter) {
+    super()
+    this.push(...headersList) // copy values
```
ronag, line 163: You can’t copy.

ronag (changes requested):
A filtered object is a view on the internal object. It’s not enough to have a copy.

ronag:
So what you need to do here is:

Implement a `FilteredResponse` that acts like an explicit proxy forwarding everything except what is filtered to the internalResponse.

Implement a `FiltereredHeadersList` which does the same as above. In order to do that we probably need to change so that `HeadersList` doesn't extend `Array` but instead has a array store property which you can share references to.

KhafraDev:
That makes sense and I definitely see why you chose proxies. Think I can work around it (already have a better `FilteredResponseView` ready.

KhafraDev:
@ronag I've thought about this quite a bit and came up with a possibly acceptable solution (I know this is an old issue...);

I think `response` (the internal one, not to be confused with the fetch `Response`) is better suited as a class. In doing so, we can easily make a "response" filtered by, for example, setting `response.filtered = true`. Of course it would also need some method for property lookups and handle the filtering logic there. Same could be done for `HeadersList` without that much change, simply by adding a `.filtered` property and running a pre-set filtering function if it's filtered.

However, I don't think server-environments (deno, cf) even do this filtering in the first place and I don't see the benefit of keeping it in.

PR #1300, opened 2022-03-22
fix(fetch): remove code for currently impossible conditions

ronag:


ronag:
@KhafraDev

ronag:
Actually it is possible to reach these by creating a Request with explicit mode.

PR #1301, opened 2022-03-23
fix(mock): matching the complete header list when using fetch

panva:
Fixes the ability to match a mock on the complete header when using fetch. This ability was introduced in [v4.16.0](https://github.com/nodejs/undici/releases/tag/v4.16.0), specifically #1275.

The issue is in the fact that when using fetch the matcher argument is not a plain object but rather HeadersList.


mcollina (approved):
lgtm

PR #1303, opened 2022-03-28
fetch: avoid async iteration when request.body is null

RafaelGSS:
It avoids treating `request.body` as an async iterator when it's null.

In my local tests, it increases the performance by almost 10%.

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -1664,7 +1664,13 @@ async function httpNetworkFetch (
   //        2. Otherwise, return a network error.
 
   // To transmit request’s body body, run these steps:
-  const requestBody = (async function * () {
+  let requestBody = null
+  // 1. If body is null and fetchParams’s process request end-of-body is
+  // non-null, then queue a fetch task given fetchParams’s process request
+  // end-of-body and fetchParams’s task destination.
+  if (request.body == null && fetchParams.processRequestEndOfBody) {
+    queueMicrotask(() => fetchParams.processRequestEndOfBody())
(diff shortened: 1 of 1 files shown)


mcollina (approved):

mcollina:
lgtm

ronag (changes requested):
I need to think about this

ronag (approved):
@RafaelGSS I slightly modified your PR. Hope you don't mind. LGTM

lib/fetch/util.js
```diff
@@ -195,7 +195,7 @@ function appendFetchMetadata (httpRequest) {
   header = httpRequest.mode
 
   //  4. Set a structured field value `Sec-Fetch-Mode`/header in r’s header list.
-  httpRequest.headersList.append('sec-fetch-mode', header)
```
RafaelGSS, line 198: is this change expected?

lib/mock/mock-interceptor.js
```diff
@@ -64,7 +64,7 @@ class MockInterceptor {
       throw new InvalidArgumentError('opts.path must be defined')
     }
     if (typeof opts.method === 'undefined') {
-      throw new InvalidArgumentError('opts.method must be defined')
```
RafaelGSS, line 67: is this change expected?

RafaelGSS (commented):
It seems there are some pushes from other branches. Overall LGTM.

lib/fetch/util.js
ronag: Just missing a rebase.

PR #1304, opened 2022-03-28
fetch: bypass headerlist.append for known headers

RafaelGSS:
It avoids the `headerList.append` checks: https://github.com/nodejs/undici/blob/main/lib/fetch/headers.js#L97, when the key/value is set by spec.

It increases the performance by 5 ~ 10% in my tests.


RafaelGSS:
This won't work due to the `binarySearch` implementation assuming the string are sorted by `localeCompare` function. I'll think about a better implementation (maybe using `Set`/`Map`) and open a fresh PR.

PR #1309, opened 2022-03-28
fetch: replace HeadersList Array inheritance to Map

RafaelGSS:
This PR replaces the `HeadersList` Array inheritance to Map, which basically is more suitable for headers use case. ~It also increases the performance by 10% (it's not done yet, so it might be more/less)~.

### DIFF
--- a/lib/fetch/headers.js
+++ b/lib/fetch/headers.js
@@ -12,7 +12,8 @@ const {
 } = require('./constants')
 
 const kHeadersMap = Symbol('headers map')
-const kSorted = Symbol('headers map sorted')
+const kSorted = Symbol('kSorted')
+const kHeadersSortedMap = Symbol('headers map sorted')
 
 function normalizeAndValidateHeaderName (name) {
   if (name === undefined) {
@@ -82,9 +83,11 @@ class HeadersList {
     if (init instanceof HeadersList) {
       this[kHeadersMap] = new Map(init[kHeadersMap])
       this[kSorted] = init[kSorted]
+      this[kHeadersSortedMap] = init[kHeadersSortedMap]
     } else {
       this[kHeadersMap] = new Map(init)
       this[kSorted] = false
+      this[kHeadersSortedMap] = new Map()
     }
   }
 
@@ -118,12 +121,13 @@ class HeadersList {
   }
 
   sortMap () {
-    this[kHeadersMap] = new Map([...this[kHeadersMap]].sort((a, b) => a[0] < b[0] ? -1 : 1))
+    this[kHeadersSortedMap] = new Map([...this[kHeadersMap]].sort((a, b) => a[0] < b[0] ? -1 : 1))
     this[kSorted] = true
   }
 
   // bypass normalization
   set (key, val) {
+    this[kSorted] = false
     return this[kHeadersMap].set(key, val)
   }
 
@@ -132,7 +136,7 @@ class HeadersList {
       this.sortMap()
     }
 
-    for (const [key, val] of this[kHeadersMap].entries()) {
+    for (const [key, val] of this[kHeadersSortedMap].entries()) {
       yield [key, val]
     }
   }
@@ -304,26 +308,26 @@ class Headers {
       this[kHeadersList].sortMap()
     }
 
-    return new HeadersList(this[kHeadersList])
+    return this[kHeadersList][kHeadersSortedMap]
   }
 
   * keys () {
-    const listFreezed = this._getHeadersListIterable()
-    for (const key of listFreezed[kHeadersMap].keys()) {
+    const sortedHeaders = this._getHeadersListIterable()
+    for (const key of sortedHeaders.keys()) {
       yield key
     }
   }
 
   * values () {
-    const listFreezed = this._getHeadersListIterable()
-    for (const val of listFreezed[kHeadersMap].values()) {
+    const sortedHeaders = this._getHeadersListIterable()
+    for (const val of sortedHeaders.values()) {
       yield val
     }
   }
 
   * entries () {
-    const listFreezed = this._getHeadersListIterable()
-    for (const [key, val] of listFreezed[kHeadersMap]) {
+    const sortedHeaders = this._getHeadersListIterable()
+    for (const [key, val] of sortedHeaders) {
       yield [key, val]
     }
   }
@@ -345,8 +349,8 @@ class Headers {
     const callback = args[0]
     const thisArg = args[1]
 
-    const listFreezed = this._getHeadersListIterable()
-    listFreezed[kHeadersMap].forEach((value, index) => {
+    const sortedHeaders = this._getHeadersListIterable()
+    sortedHeaders.forEach((value, index) => {
       callback.apply(thisArg, [value, index, this])
     })
   }
--- a/lib/fetch/headers.js
+++ b/lib/fetch/headers.js
@@ -78,7 +78,7 @@ function fill (headers, object) {
 }
 
 class HeadersList {
-  constructor(init) {
+  constructor (init) {
     if (init instanceof HeadersList) {
       this[kHeadersMap] = new Map(init[kHeadersMap])
       this[kSorted] = init[kSorted]
@@ -117,7 +117,7 @@ class HeadersList {
     return this[kHeadersMap].has(normalizedName)
   }
 
-  sortMap() {
+  sortMap () {
     this[kHeadersMap] = new Map([...this[kHeadersMap]].sort((a, b) => a[0] < b[0] ? -1 : 1))
     this[kSorted] = true
   }
@@ -127,7 +127,7 @@ class HeadersList {
     return this[kHeadersMap].set(key, val)
   }
 
-  *[Symbol.iterator]() {
+  * [Symbol.iterator] () {
     if (!this[kSorted]) {
       this.sortMap()
     }
@@ -299,7 +299,7 @@ class Headers {
 
   // The value pairs to iterate over are the return value of running sort
   // and combine with this’s header list.
-  _getHeadersListIterable() {
+  _getHeadersListIterable () {
     if (!this[kHeadersList][kSorted]) {
       this[kHeadersList].sortMap()
     }
--- a/lib/fetch/headers.js
+++ b/lib/fetch/headers.js
@@ -127,7 +127,7 @@ class HeadersList {
     return this[kHeadersMap].has(normalizedName)
   }
 
-  keys() {
+  keys () {
     return this[kHeadersMap].keys()
   }
 
--- a/lib/fetch/headers.js
+++ b/lib/fetch/headers.js
@@ -84,11 +84,13 @@ class HeadersList {
       this[kHeadersSortedMap] = init[kHeadersSortedMap]
     } else {
       this[kHeadersMap] = new Map(init)
-      this[kHeadersSortedMap] = undefined
+      this[kHeadersSortedMap] = null
     }
   }
 
   append (name, value) {
+    this[kHeadersSortedMap] = null
+
     const normalizedName = normalizeAndValidateHeaderName(name)
     const normalizedValue = normalizeAndValidateHeaderValue(name, value)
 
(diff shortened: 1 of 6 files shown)


lib/fetch/headers.js
```diff
@@ -91,61 +91,33 @@ function fill (headers, object) {
...
-    const normalizedName = normalizeAndValidateHeaderName(name)
-    const normalizedValue = normalizeAndValidateHeaderValue(name, value)
-
-    const index = binarySearch(this, normalizedName)
-    if (this[index] === normalizedName) {
-      this[index + 1] = normalizedValue
-    } else {
-      this.splice(index, 0, normalizedName, normalizedValue)
-    }
+    return super.has(normalizedName)
```
RafaelGSS, line 120: I did a benchmark and `super` is inconsistent in performance terms. In case of agreement in this PR, I'll improve it.

lib/fetch/headers.js
```diff
@@ -309,23 +290,25 @@ class Headers {
...
+      yield val
     }
   }
 
   * entries () {
-    const clone = this[kHeadersList].slice()
-    for (let index = 0; index < clone.length; index += 2) {
-      yield [clone[index], clone[index + 1]]
+    console.log('called')
+    const listFreezed = new HeadersList(this[kHeadersList])
```
RafaelGSS, line 309: I had to create a frozen list because the `Map` iterator doesn't freeze the values/keys.

ronag:
@Ethan-Arrowood

Ethan-Arrowood:
From my investigations the Map wasn't as good for reading the header values (as in per spec you must return the values in sorted order) but I do know v8 is always making magic happen; especially with objects like Map. 

Wouldn't be shocked if your new benchmark results are even better perf wise. 

One thing to keep in mind is memory footprint too; a flat array of strings should be more efficient space wise but it's a hard one to verify as well.

mcollina (approved):
lgtm

ronag:
> > as in per spec you must return the values in sorted order
> 
> Can you point me to this statement in the spec? Yeah, in my benchmarks it was better than Array implementation, mostly due to the fact of needing `binarySearch` to do any operation in headers using `Array` inheritance

https://fetch.spec.whatwg.org/#headers-class

> The [value pairs to iterate over](https://webidl.spec.whatwg.org/#dfn-value-pairs-to-iterate-over) are the return value of running [sort and combine](https://fetch.spec.whatwg.org/#concept-header-list-sort-and-combine) with [this](https://webidl.spec.whatwg.org/#this)’s [header list](https://fetch.spec.whatwg.org/#concept-headers-header-list).

AlttiRi:
Just run a server that logs the incoming request's headers:

Chrome headers

The headers are **not** alphabetically ordered. 
Neither in requests from `fetch` function, nor in any other request from the browser.

Also I don't see any clear mention in the specs that the headers should be alphabetically ordered.

BTW, an example of _"sorted"_ array: `["1", "2", "3", "4", "5"].sort(() => Math.random() >= 0.5 ? -1 : 1)`.

---

I expect that it (undici/fetch) will allow to a user decide the order of the headers, if the user will decide to overwrite the default headers, or add new ones.

RafaelGSS:
In fact, the iteration is sorted: 

image

I'll adjust the PR to match this requirement.

RafaelGSS:
I have adjusted, but still non-optimal. I'll implement a orderedmap soon.

lib/fetch/headers.js
```diff
@@ -91,64 +91,37 @@ function fill (headers, object) {
   }
 }
 
-// TODO: Composition over inheritence? Or helper methods?
-class HeadersList extends Array {
+class HeadersList extends Map {
```
ronag, line 94: I'd much prefer composition over inheritence...

lib/fetch/headers.js
```diff
@@ -308,24 +280,28 @@ class Headers {
...
 
+  // The value pairs to iterate over are the return value of running sort
+  // and combine with this’s header list.
   * keys () {
-    const clone = this[kHeadersList].slice()
-    for (let index = 0; index < clone.length; index += 2) {
-      yield clone[index]
+    const listFreezed = new HeadersList([...this[kHeadersList]].sort())
+    for (const key of listFreezed.keys()) {
+      yield key
```
ronag, line 288: No point in converting back into headers list. Also I don't think the sort works here...

lib/fetch/headers.js
```diff
@@ -308,24 +280,28 @@ class Headers {
...
     }
   }
 
+  // The value pairs to iterate over are the return value of running sort
+  // and combine with this’s header list.
   * entries () {
-    const clone = this[kHeadersList].slice()
-    for (let index = 0; index < clone.length; index += 2) {
-      yield [clone[index], clone[index + 1]]
+    const listFreezed = new HeadersList([...this[kHeadersList]].sort())
```
ronag, line 302:
```suggestion
    const listFreezed = new HeadersList([...this[kHeadersList]].sort((a, b) => a[0] > b[0] ? -1 : 1)
```

or something...

ronag (commented):
Would it make sense to cache the sorted list?

RafaelGSS:
> Would it make sense to cache the sorted list?

Yes, I thought about it yesterday. However, I'm thinking of a solution for an insertion sort (like the one before, but with less overhead).

Also, it's better(for the memory footprint) to have a flag telling if the list is currently sorted. Example:

```js
if (this[kSorted]) {
  listFreezed = new HeadersList(this[kHeadersList])
} else {
  this[kHeadersList] = new HeadersList([...this[kHeadersList]].sort())
  this[kSorted] = true
  listFreezed = new HeadersList(this[kHeadersList])
}

for (const val of listFreezed.values()) {
  yield val
}
```

ronag:
@RafaelGSS Please don't stop working on this. 😄  I was looking forward to having this fixed.

ronag (changes requested):
Getting close!

This should not require any (almost) changes outside of the headers stuff, e.g. replacing `append` with `set` is wrong and unnecessary.

Also the store for the headers should never be sorted. You would need to create a separate sorted list copy used *only* for iteration. We want to send the headers to the server as specified by the users (i.e. without sorting).

lib/fetch/headers.js
```diff
@@ -91,64 +77,68 @@ function fill (headers, object) {
...
 
-    const index = binarySearch(this, normalizedName)
+  sortMap () {
+    this[kHeadersMap] = new Map([...this[kHeadersMap]].sort((a, b) => a[0] < b[0] ? -1 : 1))
+    this[kSorted] = true
+  }
 
-    return this[index] === normalizedName
+  // bypass normalization
+  set (key, val) {
```
ronag, line 126: Why are we bypassing normalizations?

RafaelGSS: It's a way to improve the performance once it wouldn't pass by validations. I did it just in cases where we assure the key/value are valid already

RafaelGSS:
> Also the store for the headers should never be sorted. You would need to create a separate sorted list copy used only for iteration. We want to send the headers to the server as specified by the users (i.e. without sorting).

Good spot. Working on this.

lib/fetch/headers.js
ronag: Can you make a safe and unsafeSet? To make it a bit more explicit.

ronag: I'd prefer if you keep `append` where used so that it matches the spec text.

RafaelGSS: That's fair. This micro-optimization shouldn't affect too much. I'll adjust to using `append` everywhere.

lib/fetch/headers.js
```diff
@@ -91,64 +78,71 @@ function fill (headers, object) {
...
+class HeadersList {
+  constructor (init) {
+    if (init instanceof HeadersList) {
+      this[kHeadersMap] = new Map(init[kHeadersMap])
+      this[kSorted] = init[kSorted]
+      this[kHeadersSortedMap] = init[kHeadersSortedMap]
+    } else {
+      this[kHeadersMap] = new Map(init)
+      this[kSorted] = false
+      this[kHeadersSortedMap] = new Map()
```
ronag, line 90:
```suggestion
      this[kHeadersSortedMap] = null
```

If we have a sortedMap then it's sorted otherwise it's not... no need to have to vars here

RafaelGSS:
@ronag After all the changes, looks like we will get no performance improvement. The Benchmark CI looks not consistent, so, I'll perform those tests in my dedicated machine and share the results soon.

In case of no performance improvements, I'm not sure how to proceed with that. Maybe just close?

RafaelGSS:
`Main`:

```
│ Tests               │ Samples │          Result │ Tolerance │ Difference with slowest │
│ undici - fetch      │      35 │ 1391.03 req/sec │  ± 2.80 % │                       - │
│ undici - fetch      │      25 │ 1455.34 req/sec │  ± 2.82 % │                       - │
│ undici - fetch      │      30 │ 1400.48 req/sec │  ± 2.62 % │                       - │
```

This branch:

```
│ Tests               │ Samples │          Result │ Tolerance │ Difference with slowest │
│ undici - fetch      │      15 │ 1414.86 req/sec │  ± 2.38 % │                       - │
│ undici - fetch      │      10 │ 1394.15 req/sec │  ± 2.37 % │                       - │
│ undici - fetch      │      10 │ 1431.36 req/sec │  ± 2.62 % │                       - │
```

TL;DR: The performance is the same, just should just choose the best implementation.

ronag:
I still think this PR is good since it's more correct in terms of the ordering @AlttiRi mentions.

lib/fetch/headers.js
```diff
@@ -91,64 +77,68 @@ function fill (headers, object) {
...
+    this[kHeadersSortedMap] = undefined
+    return this[kHeadersMap].set(key, val)
   }
 
-  set (name, value) {
-    const normalizedName = normalizeAndValidateHeaderName(name)
-    const normalizedValue = normalizeAndValidateHeaderValue(name, value)
+  * [Symbol.iterator] () {
+    if (!this[kHeadersSortedMap]) {
+      this.sortMap()
```
ronag, line 132: I think you can inline sortMap?

lib/fetch/headers.js
```diff
@@ -91,64 +77,68 @@ function fill (headers, object) {
...
-    const index = binarySearch(this, normalizedName)
+  sortMap () {
+    this[kHeadersSortedMap] = new Map([...this[kHeadersMap]].sort((a, b) => a[0] < b[0] ? -1 : 1))
+  }
 
-    return this[index] === normalizedName
+  // bypass normalization
+  set (key, val) {
+    this[kHeadersSortedMap] = undefined
+    return this[kHeadersMap].set(key, val)
```
ronag, line 127: Where is set used?

lib/fetch/headers.js
RafaelGSS: You mean:
```js
if (!this[kHeadersSortedMap]) this.sortMap()
```

lib/fetch/headers.js
RafaelGSS: It was used in two pieces of code https://github.com/nodejs/undici/pull/1309/commits/0fdf0a698b495d6148396a41ffd4682584203187. Now, it's used in several tests only. Should I remove it?

ronag: Yea

mcollina (approved):
lgtm, the perf boost would bave been nice

ronag:
@RafaelGSS I helped a little. Hope you don't mind. PTAL.

RafaelGSS:
LGTM overall.

lib/fetch/headers.js
```diff
@@ -84,11 +84,13 @@ class HeadersList {
       this[kHeadersSortedMap] = init[kHeadersSortedMap]
     } else {
       this[kHeadersMap] = new Map(init)
-      this[kHeadersSortedMap] = undefined
+      this[kHeadersSortedMap] = null
     }
   }
 
   append (name, value) {
+    this[kHeadersSortedMap] = null
```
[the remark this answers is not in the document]
ronag: Yea, but I don't think the complexity is worth it. IMHO

RafaelGSS:
This is a breaking change, right?

Issue #1306, opened 2022-03-28
Missing Interceptable type in export ?


fraxken:
## Bug Description

While working on UT (with TypeScript) for one of our project we were unable to found the `Interceptable` type in the undici export. 

https://github.com/nodejs/undici/blob/main/types/mock-interceptor.d.ts#L88

## Reproducible By

//

## Expected Behavior

Export the Interceptable type.

## Environment

Node.js 16.13.x on Windows 10.

fraxken:
I'm ok to send a PR if maintainer confirm this should be exported :)

Uzlopak:
@nodejs/undici 

I guess we can close this as solved.

PR #1311, opened 2022-03-29
chore: export Interceptable interface

fraxken:
PR related to https://github.com/nodejs/undici/issues/1306

Not sure if i had to add a new TSD test somewhere for this new export.


mcollina (commented):
Can you add a test for this? We use tsd.

fraxken:
Done, not sure if i also need to add an expectAssignable in `mock-interceptor.test-d.ts` ? Since it was about a new exported type i focused on index.test-d.ts.

Issue #1308, opened 2022-03-28
[types] `Response.body` does not extend `ReadableStream`


AlttiRi:
The browser's  [`Response.body`](https://developer.mozilla.org/en-US/docs/Web/API/Response/body) is [`ReadableStream`](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream).

Based on `fetch.d.ts` `Response.body` of `undici` is just  `ControlledAsyncIterable`.

---

The code below is valid and workable.
However, because of `fetch.d.ts` I see the warning in my IDE (IntelliJ IDEA).

```js
const response = await fetch(url);

/** @type {ReadableStream} */
const body = response.body; // Warning: "Initializer type ControlledAsyncIterable is not assignable to variable type ReadableStream"
body.getReader();
```

```js
const response = await fetch(url);

const body = response.body;
body.getReader(); // Warning: "Unresolved function or method getReader()"
```

ronag:
Sounds like a bug in the type definitions. PR welcome.

fraxken:
I will work on it and make a PR

dheeraj1997:
Can I contribute to it?

mcollina:
sure, go ahead!

mcollina:
Ah your PR did it! It was just not tagged with "closes #issue" so this would get closed automatically.

PR #1314, opened 2022-03-30
fix(fetch): add missing ReadableStream interface to response.body

fraxken:
Add missing ReadableStream type to fetch response.body (relate to #1308).

Hope i have not missed something, ~~maybe we need a seperate type for `(ControlledAsyncIterable & ReadableStream) | null` ?~~

### DIFF
--- a/test/types/fetch.test-d.ts
+++ b/test/types/fetch.test-d.ts
@@ -1,5 +1,6 @@
 import { URL } from 'url'
 import { Blob } from 'buffer'
+import { ReadableStream } from 'stream/web'
 import { expectType, expectError } from 'tsd'
 import {
   BodyInit,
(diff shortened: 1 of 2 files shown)


test/types/fetch.test-d.ts
```diff
@@ -17,7 +18,8 @@ import {
   Response,
   ResponseInit,
   ResponseType,
-  ReferrerPolicy
+  ReferrerPolicy,
+  ControlledAsyncIterable
```
ronag, line 22: What is `ControlledAsyncIterable` and do we even need it?

fraxken: Not sure. This does not seem to be used or tested.

ronag: I suggest you simply remove it...

ronag (approved):

test/types/fetch.test-d.ts
fraxken: Done 👍

KhafraDev: When the types were published I don't think @types/node had a ReadableStream type, so this pseudo-readablestream one was added.

distracteddev:
This change broke undici's support for working with @types/node < 16.x.x 

Not sure if this project considers this a problem, since it technically still runs on Node 12 and 14, but cannot be compiled without using a different set of types.

PR #1315, opened 2022-03-30
feat(fecth): corsCheck

hemanth:
CORS check.

### DIFF
--- a/lib/fetch/util.js
+++ b/lib/fetch/util.js
@@ -168,9 +168,38 @@ function crossOriginResourcePolicyCheck () {
 }
 
 // https://fetch.spec.whatwg.org/#concept-cors-check
-function corsCheck () {
-  // TODO
-  return 'success'
+function corsCheck (request, response) {
+  // 1.  Get `Access-Control-Allow-Origin` from response’s header list.
+  const allowedOrigin = response.headersList.get('access-control-allow-origin')
+
+  // 2. If origin is null, then return failure.
+  if (allowedOrigin === null) {
+    return 'failure'
+  }
+
+  // 3. If request’s credentials mode is not "include" and origin is `*`, then return success.
+  if (request.credentials !== 'include' && allowedOrigin === '*') {
+    return 'success'
+  }
+
+  // 4. If the result of byte-serializing a request origin with request is not origin, then return failure.
+  // TODO: not very clear what byte-serializing means.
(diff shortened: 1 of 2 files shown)


lib/fetch/util.js
```diff
@@ -168,9 +168,38 @@ function crossOriginResourcePolicyCheck () {
...
+
+  // 3. If request’s credentials mode is not "include" and origin is `*`, then return success.
+  if (request.credentials !== 'include' && allowedOrigin === '*') {
+    return 'success'
+  }
+
+  // 4. If the result of byte-serializing a request origin with request is not origin, then return failure.
+  // TODO: not very clear what byte-serializing means.
+  // https://fetch.spec.whatwg.org/#byte-serializing-a-request-origin
+  // Should we do the isomorphic encoding?
```
hemanth, line 188: Seeking some clarity on this.

mcollina (commented):
Why are we adding this? We'd
never support cors.

ronag:
@benjamingr

KhafraDev:
If this PR isn't merged 83defac59ca50494e88b10ff22226b524f1f7dee should probably get reverted too as it deals with cross origin stuff as well

> Cross-origin read blocking, better known as CORB, is an algorithm which identifies dubious cross-origin resource fetches (e.g., fetches that would fail anyway like attempts to render JSON inside an img element) and blocks them before they reach a web page. CORB reduces the risk of leaking sensitive data by keeping it further from cross-origin web pages.

benjamingr:
@ronag I agree with Matteo on this

ronag:
We should replace the implementation parts with a clear comment that this stub is out of scope.

hemanth:
Ah, it makes more sense, was doubtful too. 

Should we revisit the other unimplemented methods?

hemanth:
We have the following to be implemented:

1. crossOriginResourcePolicyCheck
2. TAOCheck
3. coarsenedSharedCurrentTime
4. makePolicyContainer
5. clonePolicyContainer
6. determineRequestsReferrer
7. tryUpgradeRequestToAPotentiallyTrustworthyURL

mcollina (changes requested):
We should be checking CORS in fetch() in Undici.

Could you please document this divergence from the spec instead?

mcollina:
We are not going to support CORS in Node.js.

PR #1332, opened 2022-04-12
feat: allow `FormData` bodies in `request`

### LINKED ISSUE #1330: feature request: allow sending FormData in `undici.request`?
## This would solve...

In a codebase I am working on, it is a requirement to be able to send FormDatas to the server (to upload files, etc.). It is preferable to use `request` because it is currently much faster than `fetch`.

Undici's fetch currently parses the FormData to a stringified version:
https://github.com/nodejs/undici/blob/c0961f4c7ef5dd6e64dbb7ffbd3f6cb1f78f106a/lib/fetch/body.js#L75-L124

## I have also considered...

Currently I have an implementation identical to this that just removed the async generator and manually sets the body & header properly. This is very tedious, but it does work.

## Additional context

This is basically all that is missing from making a seamless transition from `node-fetch` to `undici.fetch` to `undici.request`.

KhafraDev:
Fixes #1330 

Allows sending a FormData body to `undici.request`.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -67,6 +67,12 @@ const {
 const kClosedResolve = Symbol('kClosedResolve')
 
 const channels = {}
+const [nodeMajor, nodeMinor] = process.version
+  .slice(1) // remove 'v'
+  .split('.', 2)
+  .map(v => Number(v))
+
+let extractBody
 
 try {
   const diagnosticsChannel = require('diagnostics_channel')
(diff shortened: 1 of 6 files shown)


lib/client.js
```diff
@@ -1580,10 +1586,26 @@ async function writeIterable ({ body, client, request, socket, contentLength, he
     .on('close', onDrain)
     .on('drain', onDrain)
 
+  let bodyAsyncIterable = body
+
+  if (util.isFormDataLike(body)) {
+    if (nodeMajor < 16 || (nodeMajor > 16 && nodeMinor < 5)) {
+      throw new InvalidArgumentError('Form-Data bodies are only supported in node v16.5 and newer.')
+    }
```
almeidx, line 1594: This will throw on 17.5, 18.5, 19.5, etc.

KhafraDev: good catch, fixed in 5169b98022b20c8dfc16fafdcaa707636e2df051

mcollina:
go ahead and add busboy!

mcollina (approved):
lgtm

cjh980402:
Although it has already been merged, it seems a better way to use `process.version.node` instead of `process.version.slice(1)`.

PR #1334, opened 2022-04-13
fix(types): callback body can be any input type

### LINKED ISSUE #1322: types(MockResponseCallbackOptions): body is not always a string
## Bug Description

When mocking `undici.fetch`, and supplying a function to `.intercept`, the parameter's `body` can be an object (tested with FormData, maybe it falls back to not parsing on unsupported types?).

## Reproducible By

```js
import { fetch, FormData, MockAgent, setGlobalDispatcher } from 'undici';
import { Blob } from 'buffer';

const mockPool = mockAgent.get('http://localhost:3000');

mockPool.intercept({
	path: '/test',
	method: 'POST'
}).reply((t) => {
	console.log({ b: t.body }); // FormData 🤯

	return {
		data: 'Hello!',
		statusCode: 200
	}
});

const fd = new FormData();
fd.append('a', new Blob(['key']), 'and_name');

const r = await fetch('http://localhost:3000/test', {
	method: 'POST',
	body: fd
});
```

## Expected Behavior

It would be very helpful to get a stringified version of the FormData (ie. the value actually sent to a server) 🙏 

## Logs & Screenshots
image

## Environment

windows 10 node v17.7.1

### Additional context

I wasn't sure if you guys preferred separate issues or for me to cram them all into 1, sorry if you preferred the latter ❤️

KhafraDev:
Partially fixes #1322

This sets the callback option body type to the input types supported by undici. Replication with `Uint8Array`, `Buffer`, etc. can be shown in the code provided in the issue.

(this is a types-only change)

### DIFF
--- a/test/types/mock-interceptor.test-d.ts
+++ b/test/types/mock-interceptor.test-d.ts
@@ -1,6 +1,10 @@
 import { expectAssignable } from 'tsd'
-import { MockAgent, MockPool } from '../..'
+import { MockAgent, MockPool, BodyInit, Dispatcher } from '../..'
 import { MockInterceptor, MockScope } from '../../types/mock-interceptor'
+
+declare const mockResponseCallbackOptions: MockInterceptor.MockResponseCallbackOptions;
+
+expectAssignable<BodyInit | Dispatcher.DispatchOptions['body']>(mockResponseCallbackOptions.body)
 
 {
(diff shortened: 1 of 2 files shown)


ronag (approved):

mcollina (approved):
lgtm

PR #1335, opened 2022-04-13
Unify version check code with index.js

cjh980402:
Prevent using unnecessary slice method and unify version check code.


ronag:
Could we move this into some kind of utils file?

cjh980402:
The code for the node version check is duplicated, so it would be better using utils file.
But I can't sure of the best idea for where to implement it.

PR #1336, opened 2022-04-13
fix: formdata cleanup

ronag:


ronag:
@KhafraDev

lib/core/request.js
```diff
@@ -79,7 +86,7 @@ class Request {
       this.body = body.byteLength ? body : null
     } else if (typeof body === 'string') {
       this.body = body.length ? Buffer.from(body) : null
-    } else if (util.isIterable(body) || util.isBlobLike(body)) {
+    } else if (util.isFormDataLike(this.body) || util.isIterable(body) || util.isBlobLike(body)) {
```
KhafraDev, line 89: I don't think `this.body` is defined yet so it will never be a FormData, and `util.isIterable` returns true for FormData which is why passing a FormData previously didn't throw

ronag, line 89:
```suggestion
    } else if (util.isFormDataLike(body) || util.isIterable(body) || util.isBlobLike(body)) {
```

ronag: Thanks!

KhafraDev:
Much nicer now!

lib/core/request.js
```diff
@@ -11,6 +11,13 @@ const kHandler = Symbol('handler')
 
 const channels = {}
 
+let extractBody
+
+const [nodeMajor, nodeMinor] = process.version
```
KhafraDev, line 16: maybe #1335 can be implemented here as well?

PR #1337, opened 2022-04-13
fix(Headers): filter out forbidden headers & `get` should return null on unknown header name

### LINKED ISSUE #1328: Strange result with `set-cookie` header of `undici.fetch`
## Bug Description
Result of set-cookie, Set-cookie, and sEt-CoOkIe header value should be same, but undici.fetch has different result.

## Reproducible By

```js

const response = await fetch('https://maple.gg');

console.log(response.headers); // has set-cookie
console.log(response.headers.get('set-cookie')); // undefined
console.log(response.headers.get('Set-cookie')); // some cookie key and value
console.log(response.headers.get('sEt-CoOkIe')); // some cookie key and value
```

## Expected Behavior
I think all result of get should be same.

## Environment

Ubuntu 20.04
Node v16.14.2

KhafraDev:
Fixes #1328

Two issues with the recently added filtered lists and using Proxy.

### DIFF
--- a/lib/fetch/response.js
+++ b/lib/fetch/response.js
@@ -393,7 +393,8 @@ function makeFilteredHeadersList (headersList, filter) {
     get (target, prop) {
       // Override methods used by Headers class.
       if (prop === 'get' || prop === 'has') {
-        return (name) => filter(name) ? target[prop](name) : undefined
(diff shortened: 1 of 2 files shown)


ronag (approved):

mcollina (approved):
lgtm

Issue #491, opened 2020-12-05
Lifecycle hook support


kibertoad:
It would be helpful to support pre/post-processing of requests/responses. Some of the use-cases that I see are:

- authentication
- logging
- custom container wrapping (e. g. by using Either)

Ideally this feature should come with a certain level of granularity, so that it is either possible to instantiate separate Client pools with different hooks, or pass aggregated hook configuration as an entity when making a request, as well as being able to set it on a global level.

If this is accepted, I can work on implementation.

mcollina:
I'm positive. This technique can be implemented with very little overhead (see Fastify).

However I would not do global, only per-client and per-pool.

kibertoad:
@mcollina Can you point out places in fastify code you would recommend to use as reference? Also should I go for full-blown plugin style as fastify does, or implement minimal useful subset of features?

mcollina:
I would go with minimal features. Here is the hooks implementation: https://github.com/fastify/fastify/blob/master/lib/hooks.js

kibertoad:
Thanks a lot!

kibertoad:
@mcollina BTW, there is no `engines` constraint for Node.js version in package.json, but CI is running on 10+. Do any older Node versions have to be supported?

ronag:
> Do any older Node versions have to be supported?

No only LTS versions.

arontsang:
I would like to suggest the following signature for extending Dispatchers

    function DispatchExtender(dispatcher: Dispatcher, opts: RequestOptions, handler: DispatchHandlers): [RequestOptions, DispatchHandlers]

This way, we can reduce a list of DispatchExtenders when calling dispatch on each dispatcher.

    for (const decorator of opts.dispatchExtenders ?? []) {
        [opts, handler] = decorator(this, opts, handler)
    }

For example, we can wrap the existing RedirectHandler like this:

    function RedirectDecorator (dispatcher, opts, innerHandler) {
      if (dispatcher instanceof Agent) {
        const { maxRedirections = dispatcher[kMaxRedirections] } = opts
        if (maxRedirections != null && maxRedirections !== 0) {
          opts = { ...opts, maxRedirections: 0 } // Stop sub dispatcher from also redirecting.
          return [opts, new RedirectHandler(this, maxRedirections, opts, innerHandler)]
        }
      }
      if (dispatcher instanceof Client) {
        const { maxRedirections = dispatcher[kMaxRedirections] } = opts
        if (maxRedirections) {
[...]

PR #1338, opened 2022-04-13
Extensible DispatchHandler

arontsang:
This adds a plugin system that allows decorators to be added to the DispatchHandler as part of a pipeline.

I expect to see in future plugins such as (but not limited to):

- Windows Authentication NTLM/Negotiate/Kerberos
- Client side caching (https://github.com/nodejs/undici/issues/1146)
- Retry logic
- Telemetry/Logging

This should support https://github.com/nodejs/undici/issues/491


ronag:
I don't really see what this brings other than maybe better ergonomics? The dispatcher interface is already quite extensible? Could this live outside as an npm package?

mcollina (changes requested):
I see the need for better ergonomics for this, however I would not add any new globals for this. Could this be a feature of Agent instead?

arontsang:
At the end of the day, this is simply a proposal.

I expect some back and forth before a final revision.

Personally, I'm not happy with the shape of the API, but I'm happy to take
suggestions.

Additionally, I'm considering switching to a DispatcherDecorator model,
with a AgentBuilder for setting the global dispatcher.

arontsang:
Also, I'm having trouble coming up with an API that would not create a
breaking change on a raw Client, whilst also extracting out the
RedirectHandler.

mcollina:
How would the DispatcherDecorator work?

arontsang:
@mcollina Here is another commit showing what I mean https://github.com/arontsang/undici/commit/d54794f34480819103bf304476f7e26da4938a88?diff=unified

In effect we add to the constructor of Dispatcher `{ interceptors: Record<string, DispatchInterceptor[]> }`

Where `type DispatchInterceptor = Dispatcher[kDispatch] => Dispatcher[kDispatch]`
Where `type Dispatcher[kDispatch] = (options: Dispatcher.DispatchOptions, handler: Dispatcher.DispatchHandlers) => boolean`

We then reduce the array of `DispatchInterceptors` over `this[kDispatch]` to produce an intercepted dispatch method.

arontsang:
I've just noticed that ProxyAgent could be replaced with a DispatchInterceptor quite trivially.

lib/dispatcher-base.js
```diff
@@ -125,6 +130,15 @@ class DispatcherBase extends Dispatcher {
     })
   }
 
+  [kBuildInterceptedDispatch] () {
+    let dispatch = this[kDispatch].bind(this)
+    for (let i = this[kInterceptors].length - 1; i >= 0; i--) {
+      dispatch = this[kInterceptors][i](dispatch)
+    }
+
+    return dispatch
```
mcollina, line 139: Can we cache this?

types/dispatcher.d.ts
```diff
@@ -36,6 +37,59 @@ declare class Dispatcher extends EventEmitter {
...
+
+  rawListeners(eventName: 'connect'): ((origin: URL, targets: Array<Dispatcher>) => void)[]
+  rawListeners(eventName: 'disconnect'): ((origin: URL, targets: Array<Dispatcher>, error: UndiciError) => void)[];
+  rawListeners(eventName: 'connectionError'): ((origin: URL, targets: Array<Dispatcher>, error: UndiciError) => void)[];
+  rawListeners(eventName: 'drain'): ((origin: URL) => void)[];
+
+  emit(eventName: 'connect', origin: URL, targets: Array<Dispatcher>): boolean;
+  emit(eventName: 'disconnect', origin: URL, targets: Array<Dispatcher>, error: UndiciError): boolean;
+  emit(eventName: 'connectionError', origin: URL, targets: Array<Dispatcher>, error: UndiciError): boolean;
+  emit(eventName: 'drain', origin: URL): boolean;
```
mcollina, line 92: Can you add the tests for those?

lib/interceptor/redirect.js
```diff
@@ -0,0 +1,20 @@
+
```
mcollina, line 1: Add use strict to the top.

types/pool.d.ts
```diff
@@ -18,5 +19,7 @@ declare namespace Pool {
     factory?(origin: URL, opts: object): Dispatcher;
     /** The max number of clients to create. `null` if no limit. Default `null`. */
     connections?: number | null;
+
+    interceptors?: { Pool?: DispatchInterceptor[] } & Client.Options["interceptors"]
```
mcollina, line 23: can you please add a test for these new types?

mcollina (commented):
Good work! Could you run the benchmarks before/after? Could you document this new feature?

types/dispatcher.d.ts
arontsang: Sorry I am really having trouble with `tsd`.

I am much more comfortable writing Typescript than this mix. I am finding that the usage of `tsd` instead of writing the library in typescript is actually creating a lot of bugs in the `.d.ts` files (such as `DispatchHandler.onHeaders`).

Would you guys be open to converting the library to typescript instead?

types/pool.d.ts
arontsang: I've added a few Jest tests to run some interceptors. Mainly in manipulating requests and responses.

types/dispatcher.d.ts
mcollina: We are ok with Undici being JavaScript, thanks for the offer.

lib/dispatcher-base.js
arontsang: Cached

types/dispatcher.d.ts
arontsang: @mcollina I am unable to write the tests for these events. I am not familiar enough with `tsd`. How would you like to proceed?

mcollina: check out https://github.com/nodejs/undici/tree/main/test/types and tsd README. If you can write TypeScript, you could learn tsd quickly.

docs/api/DispatchInterceptor.md
```diff
@@ -0,0 +1,60 @@
...
+
+### Basic Example 2
+
+Here is a contrived example of an interceptor stripping the headers from a response.
+
+```js
+'use strict'
+
+const clearHeadersInterceptor = dispatch => {
+  const DecoratorHandler = require('undici/lib/handler/decorator')
```
mcollina, line 45: This is an internal library, let's not document it. If it needs to be exposed, please use the module entry point.

lib/dispatcher-base.js
```diff
@@ -143,7 +148,26 @@ class DispatcherBase extends Dispatcher {
         throw new ClientClosedError()
       }
 
-      return this[kDispatch](opts, handler)
+      if (!this[kInterceptedDispatch]) {
+        function buildInterceptedDispatch () {
```
mcollina, line 152: Could you move this to a top level function?

mcollina (commented):
Good work!

mcollina:
@ronag PTAL

docs/api/DispatchInterceptor.md
arontsang: So? Anything needed to be done here?

arontsang:
> @ronag PTAL

@ronag Any chance you can take a look?

docs/api/DispatchInterceptor.md
arontsang: I've also noticed that http headers being of type ByteArray means that each extension is likely to be to decode utf8 once each...

ronag:
I'll have. A look next week.

arontsang:
Thanks

ronag:
I think this can be achieved without modifying existing internals and I think I would maybe prefer that, i.e. implement a `compose/decorate/intercept` (or other name) function that implements a compose dispatcher that applies this logic. So it would essentially be possible to do the following without any modification of existing code:

```js
const insertHeaderInterceptor = dispatch => {
  return function InterceptedDispatch(opts, handler){
    opts.headers.push('Authorization', 'Bearer [Some token]')
  }
}

const client = undici.intercept(new Client('https://localhost:3000', [ insertHeaderInterceptor ])
```

ronag:
Or maybe:

```js
const client = undici.compose([
  () => new Client('https://localhost:3000', 
  dispatch => function insertHeader(opts, handler){
    opts.headers.push('Authorization', 'Bearer [Some token]')
  }
])
```

ronag:
```js
undici.request(url, [
   decorate: [insertHeaders({ foo: 'bar' }, redirect({ maxRetries: 3 )]
])
```

ronag:
This might need a few iterations and since it doesn't require any internal modifications we should consider whether or not we should first have it as an external package. Not saying we should. Just to consider.

arontsang:
> ```js
> undici.request(url, {
>    decorate: [insertHeaders({ foo: 'bar' }, redirect({ maxRetries: 3 )]
> })
> ```

This would not work for my use case. I NEED to apply the decorator AT the client level of an Agent.

The decorator needs to be scoped to a single TCP connection to be able to implement NTLM authentication.

arontsang:
> Or maybe:
> 
> ```js
> const client = undici.compose(
>   () => new Client('https://localhost:3000', 
>   dispatch => function insertHeader(opts, handler){
>     opts.headers.push('Authorization', 'Bearer [Some token]')
>     return dispatch(opts, handler)
>   }
> )
> ```

This works, but it would require an inordinate amount of boiler plate to bring up a full Agent with decorators at the right levels, Agent, Pool, Client etc...

ronag:
Good point. Please give me a little time to mull and experiment with this.

docs/api/DispatchInterceptor.md
mcollina: Go ahead and expose it in this PR.

> I've also noticed that http headers being of type ByteArray means that each extension is likely to be to decode utf8 once each...

What do you mean? Can you make an example?

arontsang:
> Good point. Please give me a little time to mull and experiment with this.

Any thoughts?

ronag:
Sorry for the delay. This is quite tricky and not something that's easy to land. It's good work but this kind major API change is something that we can't change back and forth and needs thought.

I will spend more time on this before I go on vacation next week.

arontsang:
Well, if you are already thinking about it. I would like to change the
interface for the DispatchHandler to return headers as string[] rather than
ByteArray[].

This would help reduce the number of Utf8 decodes.

lib/dispatcher-base.js
```diff
@@ -125,6 +130,24 @@ class DispatcherBase extends Dispatcher {
...
+  [kInterceptedDispatch] (opts, handler) {
+    if (!this[kInterceptors] || this[kInterceptors].length === 0) {
+      this[kInterceptedDispatch] = this[kDispatch]
+      return this[kDispatch](opts, handler)
+    }
+
+    let dispatch = this[kDispatch].bind(this)
+    for (let i = this[kInterceptors].length - 1; i >= 0; i--) {
+      const interceptor = this[kInterceptors][i]
+      if (typeof interceptor !== 'function') {
```
kibertoad, line 142: Can we validate this once on addition and not on every dispatch?

lib/handler/decorator.js
```diff
@@ -0,0 +1,35 @@
+'use strict'
+
+module.exports = class DecoratorHandler {
```
kibertoad, line 3: file name is inconsistent with class name

lib/interceptor/redirect.js
```diff
@@ -0,0 +1,21 @@
+'use strict'
+
+const RedirectHandler = require('../handler/redirect')
+
+function RedirectInterceptor ({ maxRedirections: defaultMaxRedirections }) {
```
kibertoad, line 5: this has class naming but is not a class. maybe it should be renamed to indicate that this is a factory function for a partially resolved function?

kibertoad, line 5: this kind of renaming is fairly confusing, do we need it?

test/jest/interceptor.test.js
```diff
@@ -0,0 +1,207 @@
...
+    const origin = new URL(`http://localhost:${server.address().port}`)
+    await Promise.all([
+      request(origin, { dispatcher: agent }),
+      request(origin, { dispatcher: agent })
+    ])
+
+    // Assert that the requests are run on different interceptors (different Clients)
+    const requestCounts = interceptors.map(x => x.requestCount)
+    expect(requestCounts).toEqual([1, 1])
+  } finally {
```
kibertoad, line 36: why? can't this be handled in jest `after` block?

types/client.d.ts
```diff
@@ -41,6 +42,8 @@ declare namespace Client {
     tls?: TlsOptions | null;
     /** */
     maxRequestsPerClient?: number;
+
+    interceptors?: {Client: DispatchInterceptor[] | undefined}
```
kibertoad, line 46: do you need to mention `undefined` explicitly if it's an optional parameter already?

types/dispatcher.d.ts
```diff
@@ -36,6 +37,59 @@ declare class Dispatcher extends EventEmitter {
   destroy(err: Error | null): Promise<void>;
   destroy(callback: () => void): void;
   destroy(err: Error | null, callback: () => void): void;
+
+  on(eventName: 'connect', callback: (origin: URL, targets: Array<Dispatcher>) => void): this;
```
kibertoad, line 41: probably readonly arrays should suffice in all these cases?

types/dispatcher.d.ts
```diff
@@ -145,7 +199,7 @@ declare namespace Dispatcher {
     /** Invoked when request is upgraded either due to a `Upgrade` header or `CONNECT` method. */
     onUpgrade?(statusCode: number, headers: string[] | null, socket: Duplex): void;
     /** Invoked when statusCode and headers have been received. May be invoked multiple times due to 1xx informational headers. */
-    onHeaders?(statusCode: number, headers: string[] | null, resume: () => void): boolean;
+    onHeaders?(statusCode: number, headers: Buffer[] | null, resume: () => void): boolean;
```
kibertoad, line 202: why this change?

types/client.d.ts
```diff
@@ -1,8 +1,9 @@
 import { URL } from 'url'
 import { TlsOptions } from 'tls'
 import Dispatcher = require('./dispatcher')
-import { DispatchOptions, RequestOptions } from './dispatcher'
-import buildConnector = require('./connector')
+import {DispatchInterceptor} from './dispatcher'
+import buildConnector, {connector} from "./connector";
+import {Socket} from "net"
```
kibertoad, line 6: where is this used?

kibertoad:
Is there something I could help with that would advance this PR further?

types/dispatcher.d.ts
arontsang: Because the typescript is wrong. I am merely correcting the type definition file.

To be honest, this is why I hate javascript libraries, nobody knows what the correct types are anywhere.

test/jest/interceptor.test.js
arontsang: Because I am not a javascript programmer and I don't use jest. I had enough trouble getting this work...

ronag (commented):
@mcollina I'm not convinced by the direction of this at the moment but I'm having difficulty getting time to focus on this and give it the thought that is required. I'm ok with merging this but I might propose semver major changes in the future. I'm not sure what the right thing to do here is. I don't want to be blocking here for unreasonably long time either.

mcollina:
Let's go ahead and land. The code seems reasonable to me so far, minus all the comments.
I'm 100% ok with semver-major changes on this one.

Maybe we could classify interceptors as experimental?

Could you address them @arontsang? Sorry for the long wait.

kibertoad:
@arontsang if you need any help, can you give me write permissions on your repo?

arontsang:
> @arontsang if you need any help, can you give me write permissions on your repo?

Added

kibertoad:
I'll finally have some free time next week and will try to resolve conflicts here, sorry for the delay

ci010:
I think PR is really useful for extensibility.

I'm interested in implementing retry logic with it, and I did some POCs.
I found current design might have some gaps to implement retry.

Consider following retry handler (implemented by DecoratorHandler):
```ts

export class RetryHandler extends DecoratorHandler {
    private attempt = 0

    constructor(handler: Dispatcher.DispatchHandlers, private policy: RetryPolicy['retry'], private dispatcher: Dispatcher['dispatch'], private opts: Dispatcher.DispatchOptions) {
        super(handler)
    }

    onError(err: Error): void {
        const shouldRetry = this.policy(this.opts, ++this.attempt, err)
        if (typeof shouldRetry === 'boolean') {
            if (shouldRetry) {
                this.dispatcher(this.opts, this)
            } else {
                super.onError(err)
            }
        } else {
            shouldRetry.then((shouldRetry) => {
                if (shouldRetry) {
                    this.dispatcher(this.opts, this)
                } else {
                    super.onError(err)
                }
            })
        }
    }
}
```

Basically, it will re-dispatch the http request once the error happened and the error should be retried (by user logic).

This works totally fine if the error is thrown before the client receive the response `headers` and `statusCode`.
However, if the error is thrown during body receiving progress, the retry won't work, as the `Request`'s `callback` is set to null.
(Like body timeout error)

https://github.com/nodejs/undici/blob/main/lib/api/api-request.js#L90
```ts
onHeaders() {
...
    this.callback = null // set callback to null
    this.res = body
...
}
```

And that will make the further retries always failed:

https://github.com/nodejs/undici/blob/main/lib/api/api-request.js#L68
```ts
onConnect (abort, context) {
    if (!this.callback) {
      throw new RequestAbortedError() // no callback will throw abort error
    }

    this.abort = abort
    this.context = context
  }
```

The `stream` and `pipeline` have different behaviors. `stream` will clear the `callback` after the `response` stream finished. I didn't test it, but it might work. Maybe we can do the similar things in `request`.

I just wonder will this PR also deal this issue. Or it's better to be another PR.

kibertoad:
@arontsang Can you provide any details on
```
Well, if you are already thinking about it. I would like to change the
interface for the DispatchHandler to return headers as string[] rather than
ByteArray[].

This would help reduce the number of Utf8 decodes.
```
? I can't find where in the code this is happening

kibertoad:
@mcollina @ronag I've resolved conflicts and addressed all the outstanding comments, could you please rereview?

kibertoad:
@mcollina @ronag All green and ready for rereview.

docs/api/DispatchInterceptor.md
mcollina: @kibertoad this needs fixing, we should not be deep-requiring.

docs/api/Pool.md
```diff
@@ -19,6 +19,7 @@ Extends: [`ClientOptions`](Client.md#parameter-clientoptions)
 
 * **factory** `(origin: URL, opts: Object) => Dispatcher` - Default: `(origin, opts) => new Client(origin, opts)`
 * **connections** `number | null` (optional) - Default: `null` - The number of `Client` instances to create. When set to `null`, the `Pool` instance will create an unlimited amount of `Client` instances.
+* **interceptors.Pool** `Array<DispatchInterceptor>` - Default: `[]` - A list of interceptors that are applied to the dispatch method. Additional logic can be applied (such as, but not limited to: 302 status code handling, authentication, cookies, compression and caching).
```
mcollina, line 22: It's not clear this is `{ interceptors: { Pool: [] } }`

kibertoad:
@mcollina Comments addressed!

lib/proxy-agent.js
```diff
@@ -42,7 +62,7 @@ class ProxyAgent extends DispatcherBase {
 
     const connect = buildConnector({ ...opts.proxyTls })
     this[kConnectEndpoint] = buildConnector({ ...opts.requestTls })
-    this[kClient] = new Client({ origin: opts.origin, connect })
+    this[kClient] = new Agent({ origin: opts.origin, connect })
```
mcollina, line 65: Why this change?

mcollina (approved):
I would avoid introducing any more global settings, could you avoid them? Thanxk

lib/proxy-agent.js
kibertoad:
```
const Client = require('./agent')
const Agent = require('./agent')
```

Naming was confusing, they were exactly same thing. Am I missing anything?

ronag: Seems more like the import was incorrect?

kibertoad: ok, I'll fix it then

kibertoad:
> I would avoid introducing any more global settings, could you avoid them? Thanxk

@mcollina Do you mean new options for Client/Agent/Pool? Where would you place them instead?

lib/proxy-agent.js
kibertoad: done

docs/api/Pool.md
kibertoad: fixed in all docs

mcollina (approved):
lgtm, that text was a leftover from a previous half-botched review

ronag (approved):

kibertoad:
@mcollina @ronag What's left before this can be merged?
Also it would be great to release a new version sometime after the merge, would allow our team to start working on actual Undici interceptors :D

wong2:
Does this support async logic in interceptors? For example, read credentials from the disk in a request interceptor and then add them to request headers.

wong2:
Also I find it annoying when extending request headers in a request interceptor, the `opts.headers` type is vague, according to typescript declaration they could be `IncomingHttpHeaders | string[] | null`, which makes it hard when extending...

arontsang:
I would need help with some of this.

I am a dotnet Dev by trade, so a lot of it isn't my wheelhouse, for
example, I don't know the Jest library inside and out...

Would it be better if we put this into a branch?

arontsang:
It can, but, that would get very complicated when we get to the point of
having half a dozen of these things

PR #1343, opened 2022-04-17
fetch: remove unnecessary new URL calls

RafaelGSS:
Those `new URL` calls seem unnecessary. Its basically doing `new URL(new URL(x))`

It bumps the performance by 10% on avg.


mcollina (approved):
lgtm

PR #1344, opened 2022-04-18
feat(diagnostics_channel): add `bodyReceived` property to trailers

### LINKED ISSUE #1342: [diagnostics_channel] support body received in `undici:request:trailers`
## This would solve...

In a library I am working on, they had 2 events (request & response) that would fire before and after a request was made. The purpose of these events was for analytics (ie. receiving a "weird" response code and checking the body). With undici, I'm hoping that these events can be removed entirely and instead use the `diagnostics_channel` support (the library only supports v16.9.0+).

In the `request` property, the body is sent but under a symbol property and uncloned. Exposing it publicly (maybe under `bodyReceived` or some other name) and cloning it so it can be used twice would be exactly what I need. Receiving a buffer would also be very useful.

Since it is using `request`, rather than `fetch`, it is using a node Readable stream so we can't just tee the stream.

## Additional context

The implementation previously used was not performant and didn't need to be because it was only supposed to be used when something majorly messed up.

KhafraDev:
Fixes #1342 

This adds a property to the object in the `undici:request:trailers` called "response" which currently only contains the `body` received from the server (as an array of buffers).

Downsides:
- Probably uses more memory if subscribed to this event now.


ronag (changes requested):
I don’t think this is a good idea. The body can be huge.

KhafraDev:
Yeah that's fair, would it be better if I created a new stream instead?

ronag:
Why not just send the body stream as is?

KhafraDev:
Can't re-use it. It's very helpful for analytics, etc. to be able to see what body was received. For example an API I work with will send 2 types of 429s: a "real" one with JSON and rarely a cloudfare one. When getting a 429, it's more-or-less required to know "which type" it is, as one could be something wrong on your end, and one could be something wrong with a server.

lib/api/api-request.js
```diff
@@ -52,6 +61,10 @@ class RequestHandler extends AsyncResource {
     this.context = null
     this.onInfo = onInfo || null
 
+    if (channels.trailers.hasSubscribers) {
+      this.bodyReceived = []
+    }
```
mcollina, line 66: Unfortunately keeping all the body in memory for all the users of the trailers event is significantly too expensive. Why can't you pass the `this.res` object to a previous event?

KhafraDev: If you consume the body, it can't be consumed later on (I have converted it to a Readable stream, just haven't pushed it yet). Ie:

```js
import diagnosticsChannel from 'diagnostics_channel'
import { text } from 'stream/consumers'

diagnosticsChannel.channel('undici:request:trailers').subscribe(({ response }) => {
    // let's say it was the same readable stream
    // consume the body somehow
    text(response.body).then(...)
    // will likely throw because the body has been used already
})

const { body } = await reques('...')
await body.text() // do something with body after
```

mcollina: How about?

```js

import diagnosticsChannel from 'diagnostics_channel'
import { text } from 'stream/consumers'

diagnosticsChannel.channel('undici:request:trailers').subscribe(({ response }) => {
   const push = response.push
   response.push = (chunk) => {
     // do whatever you want here
     push.call(response, chunk)
   }
})

const { body } = await reques('...')
await body.text() // do something with body after

```

KhafraDev: You receive the diagnostics_channel message after the body has been consumed (in this example and others I tested), so we only get a chunk that is null.

KhafraDev: After looking into this, we can get the body - as an array of buffers, on the main branch - using:

```js
import diagnosticsChannel from 'diagnostics_channel'
import { request } from './index.js'

diagnosticsChannel.channel('undici:request:trailers').subscribe(({ response }) => {
  const kConsume = Object.getOwnPropertySymbols(response).find(s => s.description === 'kConsume');
  console.log(response[kConsume].body);
})

const { body } = await request('https://www.google.com/')
await body.text() // do something with body after

```

KhafraDev:
```js
let body1 = '', body2 = ''
diagnosticsChannel.channel('undici:request:trailers').subscribe(({ response }) => {
  const kConsume = Object.getOwnPropertySymbols(response).find(s => s.description === 'kConsume');
  body2 = Buffer.concat(response[kConsume].body).toString();
})

const { body } = await request('https://www.google.com/')
body1 = await body.text() // do something with body after

console.log(body1 === body2); // true
```

The bodies are the same 🤯

KhafraDev: I have 0 idea how this property works 😅. When fetching stuff on localhost, this property doesn't exist, but when fetching any external server, it's there.. This is making writing tests pretty difficult.

KhafraDev:
I think the best idea would be to add a new diagnostics channel (maybe `undici:body:consumed` or something) that sends the entirety of the body once it is consumed.

mcollina:
That would be a great idea, but it would be possible only for the cases where a body is fully consumed by undici.

KhafraDev:
I think that's acceptable if it's documented (would fit my usecase perfectly)

PR #1349, opened 2022-04-20
chore: add `lint-staged` for running the lint & test scripts

vinayakkulkarni:
- [x] Brings a better DX as the `npm run lint` & the `npm run test` scripts run only when the `.js` or `.ts` files are staged for pushing
- [x] Also enables the [`pre-push`](https://git-scm.com/docs/githooks#_pre_push) git hook to run lint-staged on the staged files

### DIFF
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npx --no-install lint-staged
--- a/.husky/pre-push
+++ b/.husky/pre-push
@@ -1,4 +1,0 @@
-#!/bin/sh
-. "$(dirname "$0")/_/husky.sh"
-
-npx --no-install lint-staged
--- a/package.json
+++ b/package.json
@@ -80,6 +80,7 @@
     "husky": "^7.0.2",
     "jest": "^27.2.0",
     "jsfuzz": "^1.0.15",
+    "lint-staged": "^12.3.8",
     "mocha": "^9.1.1",
     "p-timeout": "^3.2.0",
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -80,6 +80,7 @@
     "husky": "^7.0.2",
     "jest": "^27.2.0",
     "jsfuzz": "^1.0.15",
+    "lint-staged": "^12.3.8",
```
ronag, line 83: If you are using npx why is the dep needed?

.husky/pre-push
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
+. "$(dirname "$0")/_/husky.sh"
+
+npx --no-install lint-staged
```
ronag, line 4: Why pre-push? Why isn't pre-commit sufficient?

package.json
vinayakkulkarni: - [x] Fixed in 10614b0

mcollina (approved):
lgtm

.husky/pre-push
vinayakkulkarni: [`pre-push`](https://git-scm.com/docs/githooks#_pre_push) provides another layer of validation ?

ronag: But there is nothing that pre-push can catch that pre-commit doesn't? Is there? You can't push something that's not commited. Or am I missing something?

vinayakkulkarni: Agreed, `pre-push` does nothing special than `pre-commit` but I personally use it everywhere, so it was a opinionated opinion I guess 😉 

- [x] Fixed in f1a4d5e

PR #1353, opened 2022-04-21
fix(fetch): makeNetworkError called twice when httpNetworkFetch run error

meixg:
fixes: https://github.com/nodejs/node/issues/42804

the error will be warped again by `makeNetworkError` at L1751:

https://github.com/nodejs/undici/blob/b6a72620a67236ea84160a0ae3d505b3d77a2214/lib/fetch/index.js#L1734-L1752


ronag (approved):
Nice catch!

mcollina (approved):
lgtm

Issue #1002, opened 2021-08-25
Add `nock`-like "expectations" to mocks


SimenB:
## This would solve...

Currently migrating from `nock` based tests to using `undici`'s built-in mocks (thank you for including those!). Our tests currently assert that no mocks that were set up was not called. We do this via checking `nock.isDone()` and sometimes `nock.activeMocks().length === 0`. `nock` ships with a few of these they call "expectations", see https://github.com/nock/nock/blob/main/README.md#expectations.

I like being able to verify we have called all the mocks we set up, and nothing is dangling. Extra useful if using `times()` to ensure we've actually hit the endpoint the number of times expected

## The implementation should look like...

Not sure which of them makes sense to add, but `pendingMocks` and `activeMocks` would be nice (as `isDone()` is basically `pendingMocks().length === 0`).

I don't really have an opinion if it should be added to a `MockScope` or to `MockAgent` or something else.

## I have also considered...

Deleting the assertions when migrating, or trying to count in `headers` callback or some such.

## Additional context

N/A

mcollina:
This is actually a good feature to add. Would you like to send a PR?

SimenB:
Cool! I'll see if I can find the time this weekend or next week 🙂 But just to make sure the design is _somewhat_ ready:

This is nock's implementation:
https://github.com/nock/nock/blob/83131b796e2e1ca16bc9563c1aea4555d7da5162/lib/scope.js#L138-L170

Their docs says that "You probably don't need to use this" about `activeMocks()` and the `isDone` and `done` is just `pending().length === 0` with an `assert` for `done`, so I guess we just need `pendingMocks()`?

And `pendingMocks()` would then return any interceptors not used (or used up in the `times` case), plus any persisted ones? Or should we skip persisted as longs as they're called at least once (like `nock` does)?

In addition, some way of removing and interceptor would be nice - that way if we we can make sure one test doesn't affect any other by clearing any interceptors between tests. For `nock` we used `nock.cleanAll()` - I guess here it makes sense for an `Interceptable` (`MockAgent` & `MockPool`) to clear any? Or should that be on `MockAgent` which is where `enableNetConnect` is etc.?

SimenB:
Cool, I'll go for some exploratory approach then and see where we land 👍

PR #1358, opened 2022-04-21
feat: add pending interceptor check functions to mock agent

theneva:
Hi! :wave:

This adds functions to check any pending interceptors on a mock agent, as well as a function that throws with a friendly, emoji-ful error message if there are any pending interceptors.

I think this closes https://github.com/nodejs/undici/issues/1002

I've never contributed here before, so please let me know if I should change anything! 🙌

(Huge thanks to @SimenB for helping me port it from our own repo)


lib/mock/mock-agent.js
```diff
@@ -134,6 +137,92 @@ class MockAgent extends Dispatcher {
...
+      clients.totals.tooFewUses.length
+    )
+    const consumed = interceptorPluralizer.pluralize(
+      clients.totals.consumed.length
+    )
+    const persistent = interceptorPluralizer.pluralize(
+      clients.totals.persistent.length
+    )
+
+    throw new Error(`
```
mcollina, line 210: Could you use our error facility instead? https://github.com/nodejs/undici/blob/main/lib/core/errors.js

lib/mock/table-formatter.js
```diff
@@ -0,0 +1,24 @@
+const { Transform } = require('stream')
```
mcollina, line 1: add use strict

mcollina (commented):
Good work!!!

theneva:
Thanks for the review! I'm aware of some other nits and I'll address the review comments soon :)

mcollina:
Also CI is failing

theneva:
Yep, I think CI fails because I drop ansi colouring in the output if process.env.CI is set. Gonna take a look!

(Is reading that env var okay, or should I add a separate one? As an unfortunate Google Cloud Build user, I really appreciate it when tools default to no colours on CI.)

theneva:
I think it should be good now, but the linter will trip until https://github.com/nodejs/undici/pull/1361 lands I think

mcollina:
Can you rebase? The linter landed

theneva:
I merged master into my branch :) everything passed locally, so hopefully it all works now

mcollina (approved):
lgtm

@SimenB could you take a look?

docs/api/MockAgent.md
```diff
@@ -445,3 +445,71 @@ mockAgent.disableNetConnect()
 await request('http://example.com')
 // Will throw
 ```
+
```
SimenB, line 448:
```suggestion
```

lib/mock/mock-agent.js
```diff
@@ -134,6 +137,92 @@ class MockAgent extends Dispatcher {
...
+        const scope = scopeWeakRef.deref()
+
+        if (scope == null) {
+          throw new Error('scope was null; this should not happen')
+        }
+
+        return { origin, scope }
+      })
+      .map(({ origin, scope }) => {
+        // @ts-expect-error TypeScript doesn't understand the symbol use
```
SimenB, line 155:
```suggestion
```

lib/mock/mock-agent.js
```diff
@@ -134,6 +137,92 @@ class MockAgent extends Dispatcher {
...
+    const mockAgentClients = this[kClients]
+
+    return Array.from(mockAgentClients.entries())
+      .map(([origin, scopeWeakRef]) => {
+        const scope = scopeWeakRef.deref()
+
+        if (scope == null) {
+          throw new Error('scope was null; this should not happen')
+        }
+
```
SimenB, line 151:
```suggestion
        if (scope == null) {
          throw new Error('scope was null; this should not happen')
        }

```

this is just to please TS

lib/mock/mock-agent.js
```diff
@@ -134,6 +137,92 @@ class MockAgent extends Dispatcher {
...
+  pendingInterceptors () {
+    return this[kGetMockAgentClientsByConsumedStatus]().totals.tooFewUses
+  }
+
+  assertNoUnusedInterceptors (options = {}) {
+    const clients = this[kGetMockAgentClientsByConsumedStatus]()
+
+    if (clients.totals.tooFewUses.length === 0) {
+      return clients
```
SimenB, line 196: why a return value?

test/mock-interceptor-unused-assertions.js
```diff
@@ -0,0 +1,138 @@
+'use strict'
```
SimenB, line 1: there are no tests for `persist`, can you add?

Also, I think a persisted interceptor should error if it's _never_ called (I don't know if the logic does that today)

SimenB (commented):
This nicely covers the use case I had in mind when opening the issue. Thanks! 👍

ronag (approved):

test/mock-interceptor-unused-assertions.js
theneva: > there are no tests for persist, can you add?

Yep!

> Also, I think a persisted interceptor should error if it's never called (I don't know if the logic does that today)

I don't think that's currently possible with the way Undici counts requests, since the `times` value is only set if `times(<number>)` has been called. In this PR I "cheat" by treating `times` as `1` if neither `persist` nor `times` is set.

It could definitely be solved by tracking the call count counting upwards to either a `limit` set by `times()`, `1` if neither `persist` nor `times` is set, or forever if `persist` is set, before it stops serving the mocked response.

I can give it a try, but that feels like it might belong in a separate pull request. On the other hand, it would trip on a new assertion and thus be a breaking change to the API introduced in this PR, so maybe it belongs here…

Just so I don't waste time on code that will be rejected: @mcollina or @ronag are you open to me changing the call count tracking in this PR so it can all land at once?

mcollina: Go for it!

SimenB (approved):
nice!

lib/mock/mock-utils.js
```diff
@@ -107,9 +107,9 @@ function getMockDispatch (mockDispatches, key) {
 }
 
 function addMockDispatch (mockDispatches, key, data) {
-  const baseData = { times: null, persist: false, consumed: false }
+  const baseData = { timesInvoked: 0, times: 1, persist: false, consumed: false }
```
theneva, line 110: we default `times` to `1` here so we don't have to do null checking everywhere

lib/mock/mock-utils.js
```diff
@@ -107,9 +107,9 @@ function getMockDispatch (mockDispatches, key) {
...
   const replyData = typeof data === 'function' ? { callback: data } : { ...data }
-  const newMockDispatch = { ...baseData, ...key, data: { error: null, ...replyData } }
+  const newMockDispatch = { ...baseData, ...key, pending: true, data: { error: null, ...replyData } }
```
theneva, line 112: All interceptors are `pending` as they are first registered. I think `consumed` belongs here as well (and not in `baseData`), but… meh. I suppose it might be possible to register an interceptor that should not be `pending` by registering it with `.times(0)`, but that sounds too illogical to account for.

lib/mock/mock-utils.js
```diff
@@ -230,25 +230,20 @@ function mockDispatch (opts, handler) {
   const key = buildKey(opts)
   const mockDispatch = getMockDispatch(this[kDispatches], key)
 
+  mockDispatch.timesInvoked++
```
theneva, line 233: We now increase `timesInvoked` instead of decreasing `times` (which is now constant).

(Is `invoked` a good name, or would something like `timesIntercepted` be better?)

lib/mock/mock-utils.js
```diff
@@ -230,25 +230,20 @@ function mockDispatch (opts, handler) {
...
-  // If persist is true, skip
-  // Or if times is a number and > 0, skip
-  // Otherwise, mark as consumed
-
-  if (!(persist === true || (typeof times === 'number' && times > 0))) {
-    mockDispatch.consumed = true
-  }
+  // If it's used up and not persistent, mark as consumed
+  mockDispatch.consumed = !persist && timesInvoked >= times
+  mockDispatch.pending = timesInvoked < times
```
theneva, line 246: `consumed` is the same as before, but `pending` is new.

lib/mock/pending-interceptors-formatter.js
```diff
@@ -0,0 +1,40 @@
...
+    const withPrettyHeaders = pendingInterceptors.map(
+      ({ method, path, data: { statusCode }, persist, times, timesInvoked, origin }) => ({
+        Method: method,
+        Origin: origin,
+        Path: path,
+        'Status code': statusCode,
+        Persistent: persist ? '✅' : '❌',
+        Invocations: timesInvoked,
+        Remaining: persist ? Infinity : times - timesInvoked
+      }))
```
theneva, line 35: we moved the header name mapping here so that the formatter gets proper lowerCamelCased variable names passed in. I think that makes it less annoying to build a new formatter. Otherwise, this does the same thing as before

test/mock-interceptor-unused-assertions.js
theneva: Done 😅

lib/mock/mock-agent.js
```diff
@@ -134,6 +136,30 @@ class MockAgent extends Dispatcher {
...
+  }
+
+  assertNoPendingInterceptors ({ pendingInterceptorsFormatter = new PendingInterceptorsFormatter() } = {}) {
+    const pending = this.pendingInterceptors()
+
+    if (pending.length === 0) {
+      return
+    }
+
+    const pluralizer = new Pluralizer('interceptor', 'interceptors').pluralize(pending.length)
```
theneva, line 155: the `pluralizer` API ends up being a little more complicated than it maybe needs to now that we only track one array (instead of `consumed`, `tooFewUses`, and `persistent` as we did before), but I think it's still a nice little utility.

mcollina (approved):
lgtm

SimenB:
>Okay, @SimenB and I sat down together

Exclude my name from the commit, this is 99.99% @theneva. I just shouted encouragement from the back seat! 😀

That said, this PR is an awesome implementation of my feature request! ♥️

theneva:
Awesome! Is there anything else I can do to help this land? 😄

SimenB:
Yay!

SimenB:
Testing this out, works great 👍 

image

not sure if `[Function: path]` should be attempted to prettify somehow? Looks fine to me, and not sure of any other way it could be better

Issue #1359, opened 2022-04-22
CI Failing on Node 18


ronag:
@mcollina FYI

targos:
CITGM didn't see it?

ronag:
@RafaelGSS

dominykas:
Its reproducible locally, although I only get one test failing.

PR #1360, opened 2022-04-22
fix: do not set decoders when there will be no body (#1359)

ShogunPanda:
This PR makes sure no decoding attempt is done when there will be no body.

Fixes: #1359


ronag (approved):

ronag:
Any chance for a regression test?

PR #1362, opened 2022-04-22
fix: connector types

S222em:
This PR fixes an issue with the type `buildConnector.BuildOptions`.

The `buildConnector.BuildOptions` are used for 2 different methods, `tls.connect(opts)` for https, and `net.connect(opt)` for all other protocols. `TlsOptions` does not include all possible options that can be passed to those methods.

- `tls.connect(opt)` accepts the type `ConnectOptions`
- `net.connect(opt)` accepts the type `NetConnectOpts`

### DIFF
--- a/test/types/connector.test-d.ts
+++ b/test/types/connector.test-d.ts
@@ -1,6 +1,7 @@
-import { expectAssignable } from 'tsd'
+import {expectAssignable} from 'tsd'
 import { Client, buildConnector } from '../..'
-import { TLSSocket } from 'tls'
+import {ConnectionOptions, TLSSocket} from 'tls'
+import {NetConnectOpts} from "net";
 
 const connector = buildConnector({ rejectUnauthorized: false })
 expectAssignable<Client>(new Client('', {
@@ -19,3 +20,11 @@ expectAssignable<Client>(new Client('', {
     })
   }
 }))
+
+expectAssignable<buildConnector.BuildOptions>({
+  checkServerIdentity: () => undefined,
+  port: 80,
+});
+// This needs to be partial as required arguments (port) have a default in lib/core/connect.js
+expectAssignable<ConnectionOptions>({} as buildConnector.BuildOptions);
+expectAssignable<Partial<NetConnectOpts>>({} as buildConnector.BuildOptions);
--- a/test/types/connector.test-d.ts
+++ b/test/types/connector.test-d.ts
@@ -1,7 +1,7 @@
 import {expectAssignable} from 'tsd'
 import { Client, buildConnector } from '../..'
 import {ConnectionOptions, TLSSocket} from 'tls'
-import {NetConnectOpts} from "net";
+import {IpcNetConnectOpts, NetConnectOpts, TcpNetConnectOpts} from "net";
 
 const connector = buildConnector({ rejectUnauthorized: false })
 expectAssignable<Client>(new Client('', {
@@ -22,9 +22,6 @@ expectAssignable<Client>(new Client('', {
 }))
 
 expectAssignable<buildConnector.BuildOptions>({
-  checkServerIdentity: () => undefined,
-  port: 80,
+  checkServerIdentity: () => undefined, // Test if ConnectionOptions is assignable
+  localPort: 1234, // Test if TcpNetConnectOpts is assignable
 });
-// This needs to be partial as required arguments (port) have a default in lib/core/connect.js
-expectAssignable<ConnectionOptions>({} as buildConnector.BuildOptions);
-expectAssignable<Partial<NetConnectOpts>>({} as buildConnector.BuildOptions);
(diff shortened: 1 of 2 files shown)


types/connector.d.ts
```diff
@@ -1,16 +1,14 @@
...
-import { Socket } from 'net'
+import {TLSSocket, ConnectionOptions} from 'tls'
+import {NetConnectOpts, Socket} from 'net'
 
 export = buildConnector
 declare function buildConnector (options?: buildConnector.BuildOptions): typeof buildConnector.connector
 
 declare namespace buildConnector {
-  export interface BuildOptions extends TlsOptions {
+  export type BuildOptions = (ConnectionOptions | NetConnectOpts) & {
```
KhafraDev, line 8:
```suggestion
  export interface BuildOptions extends ConnectionOptions, NetConnectOpts {
```

see: https://github.com/microsoft/TypeScript/wiki/Performance#preferring-interfaces-over-intersections

mcollina (commented):
We test our types! Can you add an assertion for this using tsd?

mcollina:
Also the build is failing

S222em:
Okay, so @KhafraDev the interface can't actually extend both of the types, it must be `type some = 1 | 2`. Will it really be a problem? Explained below why it cant be like that:

The type of `NetConnectOpts` is actually not a interface but a `type some = 1 | 2`, which could be made into this:
```ts
export interface BuildOptions extends ConnectionOptions, TcpNetConnectOpts, IpcNetConnectOpts {
    maxCachedSessions?: number | null;
    socketPath?: string | null;
    timeout?: number | null;
  }
```
But this gives a TS error that properties of Tcp and Ipc opts can't be both used due to conflicting types.

Then we would get back to this:
```ts
export type BuildOptions = (ConnectionOptions | NetConnectOpts) & {
    maxCachedSessions?: number | null;
    socketPath?: string | null;
    timeout?: number | null;
  }
```

Which could also be:
```ts
export type BuildOptions = (ConnectionOptions | TcpNetConnectOpts | IpcNetConnectOpts ) & {
    maxCachedSessions?: number | null;
    socketPath?: string | null;
    timeout?: number | null;
  }
```

S222em:
@mcollina what's your opinion on this? ^

mcollina:
I think the following look correct:

```typescript
export type BuildOptions = (ConnectionOptions | TcpNetConnectOpts | IpcNetConnectOpts ) & {
    maxCachedSessions?: number | null;
    socketPath?: string | null;
    timeout?: number | null;
  }
```

however I'm not a big expert on TS, so I defer to whatever pleases the tests.

mcollina:
Can you add a couple of assertions for this in the types tests?

mcollina (approved):
lgtm

kibertoad:
@ronag Should this be merged already?

mcollina:
@kibertoad is this ok? I'm waiting for somebody to better knowledge of TS than myself.

test/types/connector.test-d.ts
```diff
@@ -19,3 +20,11 @@ expectAssignable<Client>(new Client('', {
     })
   }
 }))
+
```
kibertoad, line 23: can we also have some more explicit tests for `TcpNetConnectOpts`  and `IpcNetConnectOpts` part?

S222em: NetConnectOpts is `TcpNetConnectOpts | IpcNetConnectOpts`. I can make the tests for those separately if needed.

kibertoad: it currently only has two fields, so I suspect if you remove some of the concatenated types, tests currently will not fail.
they should fail if any of concatenated types are removed

S222em: Okay, so I ended up with:

```ts
expectAssignable<buildConnector.BuildOptions>({
  checkServerIdentity: () => undefined, // Test if ConnectionOptions is assignable
  localPort: 1234, // Check if TcpNetConnectOpts is assigable
});
```

The only problem here is that the only difference with `IpcNetConnectOpts` is that `port` is required. It does not have additional params as the other 2 do. Which would make adding `IpcNetConnectOpts` completely useless anyways probably, should I keep that one?

The ones I had indeed did not do anything.

kibertoad: let's keep it, as conceptually we are supporting both, even if they overlap

kibertoad (approved):

test/types/connector.test-d.ts
S222em: This one?
```ts
expectAssignable<buildConnector.BuildOptions>({
  checkServerIdentity: () => undefined, // Test if ConnectionOptions is assignable
  localPort: 1234, // Check if TcpNetConnectOpts is assigable
});
```
Or the one I had before?

PR #1363, opened 2022-04-22
fix: ProxyAgent dispatch options receives a header array when using fetch

### LINKED ISSUE #1355: Fetch garbles headers when used with a proxy global dispatcher
## Bug Description

When a ProxyAgent is set as the global dispatcher, the headers sent in fetch requests are completely broken: they're transformed from `header-name: header-value` into `1: header-name`, `2: header-value`, basically flattening them into an array and using the indexes as the header names.

## Reproducible By

Use any proxy, for quick testing I'd suggest launching tinyproxy on port 8888 locally via Docker:

```
docker run -it --rm -p 8888:8888 monokal/tinyproxy ANY
```

Then use Undici with that proxy:

```javascript
let Undici = require('undici');

// Use the proxy as the global dispatcher:
Undici.setGlobalDispatcher(new Undici.ProxyAgent('http://localhost:8888'))

// Send a fetch request to httpbin, which echoes the incoming headers:
Undici.fetch('http://httpbin.org/get', {
  headers: { 'Test-header': 'value' }
}).then(r => r.text()).then(console.log)
```

This prints:

```json
{
  "args": {}, 
  "headers": {
    "0": "accept", 
    "1": "*/*", 
    "10": "user-agent", 
    "11": "undici", 
    "2": "accept-encoding", 
    "3": "gzip, deflate", 
    "4": "accept-language", 
    "5": "*", 
    "6": "sec-fetch-mode", 
    "7": "cors", 
    "8": "test-header", 
    "9": "value",
    "Host": "httpbin.org", 
    "X-Amzn-Trace-Id": "Root=1-6170a3-4c398f08ae255c83"
  }, 
  "origin": "ip.ip.ip.ip", 
  "url": "http://httpbin.org/get"
}
```

I've tested this with a couple of different proxies and looked at the traffic in wireshark. This isn't being broken on the proxy site - those are the real headers that Undici is sending to the proxy itself.

## Expected Behavior

The headers should be sent in their original format. `Undici.request` does have the correct behaviour:

```javascript
// (after setting the global dispatcher as above)
Undici.request('http://httpbin.org/get', {
  headers: { 'Test-header': 'value' }
}).then(r => r.body.pipe(process.stdout))
```

That correct prints:

```json
{
  "args": {}, 
  "headers": {
    "Host": "httpbin.org", 
    "Test-Header": "value", 
    "X-Amzn-Trace-Id": "Root=1-62624e-c8afe258cd33244f8"
  }, 
  "origin": "ip.ip.ip.ip", 
  "url": "http://httpbin.org/get"
}
```

Using `Undici.fetch` without a proxy also works correctly, it's only when a global dispatcher is configured that this breaks.

## Environment

Ubuntu 20.04.4, Node 16.14.2, Undici v5.0.0

KhafraDev:
Fixes #1355

types were already correct (I'm guessing this issue has been around for a little bit):
https://github.com/nodejs/undici/blob/1dc2977c94583efbecd5b6791b5a091d18a2ceeb/types/dispatcher.d.ts#L49

### DIFF
--- a/docs/api/Dispatcher.md
+++ b/docs/api/Dispatcher.md
@@ -193,7 +193,7 @@ Returns: `Boolean` - `false` if dispatcher is busy and further dispatch calls wo
 * **path** `string`
 * **method** `string`
 * **body** `string | Buffer | Uint8Array | stream.Readable | Iterable | AsyncIterable | null` (optional) - Default: `null`
-* **headers** `UndiciHeaders` (optional) - Default: `null`
+* **headers** `UndiciHeaders | string[]` (optional) - Default: `null`.
 * **idempotent** `boolean` (optional) - Default: `true` if `method` is `'HEAD'` or `'GET'` - Whether the requests can be safely retried or not. If `false` the request won't be sent until all preceding requests in the pipeline has completed.
 * **blocking** `boolean` (optional) - Default: `false` - Whether the response is expected to take a long time and would end up blocking the pipeline. When this is set to `true` further pipelining will be avoided on the same connection until headers have been received.
 * **upgrade** `string | null` (optional) - Default: `null` - Upgrade the request. Should be used to specify the kind of upgrade i.e. `'Websocket'`.
--- a/lib/proxy-agent.js
+++ b/lib/proxy-agent.js
@@ -66,7 +66,7 @@ function buildHeaders (headers) {
     /** @type {Record<string, string>} */
     const headersPair = {}
 
-    for (let i = 0; i < headers.length; i++) {
+    for (let i = 0; i < headers.length; i += 2) {
       headersPair[headers[i]] = headers[i + 1]
     }
 
--- a/lib/proxy-agent.js
+++ b/lib/proxy-agent.js
@@ -23,7 +23,7 @@ class ProxyAgent extends DispatcherBase {
         origin: this[kProxy].uri,
         path: opts.origin + opts.path,
         headers: {
-          ...opts.headers,
+          ...buildHeaders(opts.headers),
           host
         }
       },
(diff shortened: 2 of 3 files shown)


docs/api/Dispatcher.md
```diff
@@ -193,7 +193,7 @@ Returns: `Boolean` - `false` if dispatcher is busy and further dispatch calls wo
 * **path** `string`
 * **method** `string`
 * **body** `string | Buffer | Uint8Array | stream.Readable | Iterable | AsyncIterable | null` (optional) - Default: `null`
-* **headers** `UndiciHeaders` (optional) - Default: `null`
+* **headers** `UndiciHeaders | string[]` (optional) - Default: `null`.
```
RafaelGSS, line 196: Are you sure that is `string[]`? I mean, looks like the headers must be `string[][]` then you can just `new Headers(headers)`

KhafraDev: Yes it is, or else I could have already used Object.fromEntries :/

KhafraDev:
I need a better test for this.

docs/api/Dispatcher.md
KhafraDev: I'll expand a little (so that without the test you can verify); if we used the spread operator on a 2d array, the results would be very different:

```js
 // 2d array:
 { ...[ ['a', 'b'], ['c', 'd'] ] } // -> { '0': [ 'a', 'b' ], '1': [ 'c', 'd' ] }
 
 // 1d array
 { ... ['a', 'b', 'c', 'd'] } // -> { '0': 'a', '1': 'b', '2': 'c', '3': 'd' }
}
```

ronag:
Lint failure

KhafraDev:
@ronag that is from 950a78cd8248db8362229af97c6939bc3b6632c8

PR #1369, opened 2022-04-24
feat: add `undici:body:consumed` diagnostics channel

### LINKED ISSUE #1342: [diagnostics_channel] support body received in `undici:request:trailers`

KhafraDev:
Fixes #1342

Superseded #1344 with a slightly more reliable solution (relies on consuming via the body mixins).

### DIFF
--- a/docs/api/DiagnosticsChannel.md
+++ b/docs/api/DiagnosticsChannel.md
@@ -135,3 +135,21 @@ diagnosticsChannel.channel('undici:client:connectError').subscribe(({ error, soc
   // connector is a function that creates the socket
   console.log(`Connect failed with ${error.message}`)
 })
+```
+
+## `undici:body:consumed`
+
+This message is published once a body is consumed using a [body mixin](https://github.com/nodejs/undici#body-mixins).
+
+```js
+diagnosticsChannel.channel('undici:body:consumed').subscribe(({ body, request }) => {
+  // request is the same object undici:request:create
+
+  // body is the body consumed (this corresponds with how the body was consumed. For example, if the body was consumed using .blob(), this property will be a blob)
+  if (typeof body === 'string') {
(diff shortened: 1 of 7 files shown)


ronag (changes requested):
I'm sorry but I don't like this. Can't you just have a event for each body chunk? and then trailers will signal end of body?

KhafraDev:
I can look into it, but this is also opt-in (rather than the old PR where it was added onto the trailers channel)

mcollina (approved):
lgtm, good work

KhafraDev:
I'm getting mixed signals here 😅

mcollina:
> I'm sorry but I don't like this. Can't you just have a event for each body chunk? and then trailers will signal end of body?

I have a feeling this could impact perf in the wrong way as that's a fairly hot path. Anyway if that does not regress the perf, it would be ok.

KhafraDev:
I think this way is more convenient and faster. It also only requires 1 message rather than x amount of chunks received and doesn't require the user to parse the body themselves.

PR #1370, opened 2022-04-25
feat: add support for dependabot 🥳

vinayakkulkarni:
- [x] Enables support for `dependabot` so that all the package dependencies are kept up-to-date with `npm` so as to mitigate any vulnerabilities in older versions of packages.


ronag:
We only have dev deps so there is a risk this just leads to annoying notifications with little value.

mcollina (approved):
lgtm

PR #1384, opened 2022-04-26
Change timeouts to 120 seconds

kyrylkov:


mcollina (approved):
lgtm

We will ship this when we cut a new major.

ronag (changes requested):
If prefer if we only change this for fetch.

kyrylkov:
@ronag I looked through fetch code but I wasn't able to figure out how to change timeouts there.

ronag:
> @ronag I looked through fetch code but I wasn't able to figure out how to change timeouts there. 

You can pass timeout in the `.dispatch` call in the fetch code.

> Also for server requests 30 seconds is just too short. We deal with lots of long running requests and that's how we ran into this problem in the first place.

I disagree and would prefer the 30 second. Maybe you could change my mind but I would need some more convincing.

kyrylkov:
> I disagree and would prefer the 30 second. Maybe you could change my mind but I would need some more convincing.

30 second timeout exists only in Firefox.  Node default timeout is 120.  Chrome is 300.  

Otherwise why 30?  Why not 10?  There should be some consistency.  Aligning with Node or Chrome seems like a good idea.

ronag:
The default request timeout for a node server is 30s.

ronag:
> Node default timeout is 120. 

As far as I can see from the code and docs there is no default timeout for Node?

kyrylkov:
@ronag Then how about another PR for fetch timeout for 300 seconds (like in Chrome)?

ronag:
> @ronag Then how about another PR for fetch timeout for 300 seconds (like in Chrome)?

SGTM!

PR #1386, opened 2022-04-27
feat: set fetch timeout to 300 seconds

kyrylkov:


ronag (approved):
LGTM. Would be nice with a test.

kyrylkov:
@ronag Like no timeout after 60 seconds for example?

kyrylkov:
@ronag added tests

test/node-fetch/main.js
```diff
@@ -1641,4 +1641,21 @@ describe('node-fetch', () => {
...
+      expect(res.ok).to.be.true
+      return res.text().then(result => {
+        expect(result).to.equal('text')
+      })
+    })
+
+  })
+
+  it('should time out waiting for more than 300 seconds', function () {
+    this.timeout(305_000);
```
ronag, line 1658:
```suggestion
    this.timeout(305_000)
```

ronag (commented):
Can you implement the tests using FakeTimers? See our other timeout tests.

kyrylkov:
@ronag I don't really know how to make it work.  This naive doesn't work:

```js
    if (p === '/timeout300s') {
      const clock = FakeTimers.install()
      clock.setTimeout(() => {
        res.statusCode = 200
        res.end('text')
      }, 300_000)
      clock.tick(300_000)
      clock.uninstall()
    }```

**Comment by ronag** [comment]
> @ronag I don't really know how to make it work. This naive doesn't work:
> 
> ```js
>     if (p === '/timeout300s') {
>       const clock = FakeTimers.install()
>       clock.setTimeout(() => {
>         res.statusCode = 200
>         res.setHeader('Content-Type', 'text/plain')
>         res.end('text')
>       }, 300_000)
>       clock.tick(300_000)
>       clock.uninstall()
>     }
> ```

No worries. Let's skip the test then and land this.

**Comment by kyrylkov** [comment]
I didn't commit FakeTimers.  PR has tests with regular timers.

**Comment by ronag** [comment]
> I didn't commit FakeTimers. PR has tests with regular timers.

Can you add `{ skip: true }` to them then. Otherwise, it will take very long to complete the tests?

**Comment by kyrylkov** [comment]
Done

File: test/node-fetch/main.js [hunk 2]
```diff
@@ -1643,6 +1643,7 @@ describe('node-fetch', () => {
   })
 
   it('should not time out waiting for a response 60 seconds', function () {
+    this.skip()
```
ronag, line 1646:
```suggestion
  it('should not time out waiting for a response 60 seconds', { skip: true }, function () {
```

test/node-fetch/main.js
```diff
@@ -1654,6 +1655,7 @@ describe('node-fetch', () => {
   })
 
   it('should time out waiting for more than 300 seconds', function () {
+    this.skip()
```
ronag, line 1658:
```suggestion
  it('should time out waiting for more than 300 seconds', { skip: true }, function () {
```

kyrylkov:
@ronag It looks like your change is wrong.  Fetch uses mocha which requires `this.skip()`

kyrylkov:
And that the reason I cannot make FakeTimers work, cause with fetch has different testing infrastructure

PR #1390, opened 2022-04-30
fix: don't enqueue bytes if the request was aborted

### LINKED ISSUE #1389: Error with parallel `fetch()` and the same abort signal
## Bug Description

There seems to be a problem with abort signal handling if the signal is aborted during fetch requests.
I don't know if the bug is in Node.js or undici.

## Reproducible By

```js
const abortController = new AbortController();

async function makeRequest() {
  console.log('Fetching...');
  const result = await fetch('https://jsonplaceholder.typicode.com/todos/1', {
    signal: abortController.signal
  }).then(response => response.text());
  console.log('Got response')
  // Abort other requests.
  abortController.abort();
  return result;
}

const requests = Array.from({ length: 10 }, makeRequest);
Promise.allSettled(requests).then((result) => {
  console.log('All settled');
});

```

## Expected Behavior

This should print "Fetching..." 10 times, then "Got response", then "All settled" and exit without errors.

## Logs & Screenshots

```
> node --no-warnings fetch.js
Fetching...
Fetching...
Fetching...
Fetching...
Fetching...
Fetching...
Fetching...
Fetching...
Fetching...
Fetching...
Got response
All settled
node:internal/errors:465
    ErrorCaptureStackTrace(err);
    ^

TypeError [ERR_INVALID_STATE]: Invalid state: Controller is already closed
    at new NodeError (node:internal/errors:372:5)
    at ReadableStreamDefaultController.enqueue (node:internal/webstreams/readablestream:992:13)
    at fetchParams.controller.resume (node:internal/deps/undici/undici:7863:45)
  code: 'ERR_INVALID_STATE'
}

Node.js v18.0.0
```

## Environment

Node.js 18.0.0 with builtin `fetch`

KhafraDev:
Fixes #1389

The output is now similar to Firefox:
image
image

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -1867,7 +1867,9 @@ async function httpNetworkFetch (
 
       // 7. Enqueue a Uint8Array wrapping an ArrayBuffer containing bytes
       // into stream.
-      fetchParams.controller.controller.enqueue(new Uint8Array(bytes))
+      if (!isAborted(fetchParams)) {
+        fetchParams.controller.controller.enqueue(new Uint8Array(bytes))
+      }
 
       // 8. If stream is errored, then terminate the ongoing fetch.
       if (isErrored(stream)) {
--- a/test/fetch/abort.js
+++ b/test/fetch/abort.js
@@ -1,0 +1,59 @@
+'use strict'
+
+const { test } = require('tap')
+const { fetch } = require('../..')
+const { createServer } = require('http')
+const { once } = require('events')
+
+/* global AbortController */
+
+test('parallel fetch with the same AbortController works as expected', async (t) => {
+  const body = {
+    fixes: 1389,
+    bug: 'Ensure request is not aborted before enqueueing bytes into stream.'
+  }
+
+  const server = createServer((req, res) => {
+    res.statusCode = 200
+    res.end(JSON.stringify(body))
+  })
+
+  t.teardown(server.close.bind(server))
+
+  const abortController = new AbortController()
+
+  async function makeRequest () {
+    const result = await fetch(`http://localhost:${server.address().port}`, {
+      signal: abortController.signal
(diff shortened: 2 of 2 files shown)


lib/fetch/index.js
```diff
@@ -1867,7 +1867,9 @@ async function httpNetworkFetch (
 
       // 7. Enqueue a Uint8Array wrapping an ArrayBuffer containing bytes
       // into stream.
-      fetchParams.controller.controller.enqueue(new Uint8Array(bytes))
+      if (!isAborted(fetchParams)) {
+        fetchParams.controller.controller.enqueue(new Uint8Array(bytes))
+      }
```
ronag, line 1872: Is this a spec bug or are we doing something wrong somewhere else?

mcollina (approved):
lgtm

lib/fetch/index.js
KhafraDev: Possibly a spec bug or way undici handles it. When a request is aborted, the fetchparams.controller.resume can still run (causing the bug). It could also be step 18 of https://fetch.spec.whatwg.org/#http-network-fetch but I think "if aborted" refers to *previous* steps (although I'm still confused by it).

ronag:
Why was this closed?

targos:
It's weird. GitHub doesn't say why.

lib/fetch/index.js
ronag: @KhafraDev can you move this to after await next(). That's the part where the check is relevant since we go async there.

KhafraDev:
In my fork dependabot started merging PRs into the main branch and by the time I realized I did other things that messed up the git history so much I thought it would be easier just to re-create the fork without realizing it would close my PRs.

I can't access this PR at all so I'll need to re-create it to apply ronag's suggestion.

PR #1391, opened 2022-05-01
fix: use `unknown` statusText by default in mocks

KhafraDev:
When encountering an unknown status code, an error should not be thrown. This behavior now matches fetch's behavior (tested with undici & node-fetch, unsure about browsers but I doubt it is different).


mcollina (approved):
lgtm

Issue #1392, opened 2022-05-01
[fetch] 5.1.0 regressed ability to intercept headers


panva:
## Bug Description

Intercepting upon individual headers when using `fetch` is broken in 5.1.0, worked fine in 5.0.0

#1309 and possibly lack of test coverage for this is imho the likely suspect

## Reproducible By

```js
import * as undici from 'undici'
import * as assert from 'node:assert'

const mockAgent = new undici.MockAgent()
mockAgent.disableNetConnect()
undici.setGlobalDispatcher(mockAgent)
const pool = mockAgent.get('https://example.com')
pool.intercept({
  path: '/foo',
  method: 'GET',
  headers: {
    accept: 'application/json'
  }
}).reply(200, { ok: 1 })

const response = await undici.fetch('https://example.com/foo', {
  headers: {
    accept: 'application/json'
  }
})

assert.deepStrictEqual(await response.json(), { ok: 1 })
```

## Expected Behavior

fetch gets matched intercepted

## Logs & Screenshots

```
TypeError: fetch failed
  cause: MockNotMatchedError: Mock dispatch not matched for headers '["accept","application/json","accept-language","*","sec-fetch-mode","cors","user-agent","undici","accept-encoding","br, gzip, deflate"]': subsequent request to origin https://example.com was not allowed (net.connect disabled)
```

## Environment

Node 18.0.0

### Additional context

Matching headers with a function works as expected.

panva:
cc @RafaelGSS

mcollina:
@KhafraDev could you take a look?

KhafraDev:
I'll look into it 👍

PR #1393, opened 2022-05-01
fix(mock-utils): match header when using fetch

KhafraDev:
Fixes: #1392


ronag (approved):

RafaelGSS (approved):
LGTM. These ones are hard to find in the implementation of the new header without a test, good catch.

PR #1394, opened 2022-05-02
fix: add index-fetch.js to npm distribution

targos:
See https://github.com/nodejs/node/runs/6253304838?check_suite_focus=true

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,6 +35,7 @@
   "files": [
     "*.d.ts",
     "index.js",
+    "index-fetch.js",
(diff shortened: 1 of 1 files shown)


targos:
@mcollina can we have a small v5.1.1 with this? I'd like to update undici in Node.js for v18.1.0

ronag:
> @mcollina can we have a small v5.1.1 with this? I'd like to update undici in Node.js for v18.1.0

Done

PR #1398, opened 2022-05-02
fix: set correct expectations on supported Node.js versions

n1ru4l:


saihaj (approved):

mcollina:
Why do you think so?

@dominykas for some reasons the test are not run on Node v14. Can you check!

n1ru4l:
See https://github.com/nodejs/undici/pull/1399 and https://github.com/nodejs/undici/pull/1318

I am not sure if there are other usages of modern language features that are used within the codebase. 

In both PRs I recommended an alternative solution for ensuring people from committing unsupported language features (https://github.com/weiran-zsd/eslint-plugin-node/blob/HEAD/docs/rules/no-unsupported-features/es-syntax.md).

Is this something you wanna consider? For us it is a bit frustrating building on top of undici, as we have users of all LTS Node.js versions

n1ru4l:
Okay, after some more research I noticed that this code should never run on Node.js <= 16.5 😅

Sorry for the confusion! Maybe it is possible to have a sub package.json within the fetch directory with correct engines 🤔 .

PR #1399, opened 2022-05-02
fix: do not use the logicial assignment operator

n1ru4l:
Alternative solution for https://github.com/nodejs/undici/pull/1398

See https://github.com/nodejs/undici/pull/1318

> The package.json of this project lists ["node": ">=12.18" in the engines](https://github.com/nodejs/undici/blob/8b1dcd44755d7c9bfba06066518f4c01629ad1c0/package.json#L96-L98).
>
> [Logical Assignment is only supported since Node.js 15.14](https://node.green/#ES2021-features-Logical-Assignment).
>
> To me, it seems like this usage here got added unexpectedly.
>
> We would love to prevent such things in the future by setting up a GitHub action that runs eslint with the following rule: https://github.com/weiran-zsd/eslint-plugin-node/blob/HEAD/docs/rules/no-unsupported-features/es-syntax.md
>
> Would that be something you are interested in?

### DIFF
--- a/lib/fetch/headers.js
+++ b/lib/fetch/headers.js
@@ -302,7 +302,9 @@ class Headers {
   }
 
   get [kHeadersSortedMap] () {
-    this[kHeadersList][kHeadersSortedMap] ??= new Map([...this[kHeadersList]].sort((a, b) => a[0] < b[0] ? -1 : 1))
+    if (this[kHeadersList][kHeadersSortedMap] === undefined) {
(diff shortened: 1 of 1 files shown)


saihaj (approved):

ronag (approved):

mcollina:
I don't understand why this is needed. That code is not even loaded in Node < 16: https://github.com/nodejs/undici/blob/48bb0822e052c63b40fd1fe41ce60139f03a4609/index.js#L95.

n1ru4l:
@mcollina As mentioned in https://github.com/nodejs/undici/pull/1398#issuecomment-1115247154, this is a misunderstanding! Sorry for the inconvenience.

PR #1401, opened 2022-05-02
fix(Request): allow passing Headers instance to init.headers

### LINKED ISSUE #1400: Unable to pass a `Headers` instance to the `Request` constructor
## Bug Description

When creating a `new Request`, if a `Headers` instance is passed within the init param, it does not become a part of the new request object.

## Reproducible By

```js
const { Request, Headers } = require("undici");

const request = new Request("http:localhost", {
  headers: new Headers({ wow: "hi" }),
});
assert.equal(request.headers.get("wow"), "hi");
```

## Expected Behavior

I expect the returned header to exist, instead the assertion throws because `null` is returned.

## Environment
MacOS 12.3.1 on an M1. Node v16.14.2

### Additional context
Everything works fine if instead of passing `new Header()`, a vanilla object is used.

```js
const { Request, Headers } = require("undici");

const request = new Request("http:localhost", {
  headers: { wow: "hi" },
});
assert.equal(request.headers.get("wow"), "hi");
```

I also tried this in Chrome, Firefox, and Safari, and passing a `Headers` instance to the `Response` constructor works as expected.

KhafraDev:
Fixes #1400


lib/fetch/request.js
```diff
@@ -420,10 +420,9 @@ class Request {
       // 4. If headers is a Headers object, then for each header in its header
       // list, append header’s name/header’s value to this’s headers.
       if (headers instanceof Headers) {
-        // TODO (fix): Why doesn't this work?
-        // for (const [key, val] of headers[kHeadersList]) {
-        //   this[kHeaders].append(key, val)
-        // }
+        for (const [key, val] of headers) {
+          this[kHeaders].append(key, val)
+        }
```
ronag, line 425: I think the `this[kState].headersList = new HeadersList([` below should be removed then.

KhafraDev: Done in 2569544a67c30d1e74177c51b2105a4181f09c74 👍 

I also found a way to remove the `kInit` overload in the constructor, so I'll probably make a new PR for that at some point.

lib/fetch/request.js
```diff
@@ -420,15 +420,9 @@ class Request {
...
-        // TODO (fix): Why doesn't this work?
-        // for (const [key, val] of headers[kHeadersList]) {
-        //   this[kHeaders].append(key, val)
-        // }
-
-        this[kState].headersList = new HeadersList([
-          ...this[kState].headersList,
-          ...headers[kHeadersList]
-        ])
+        for (const [key, val] of headers) {
```
ronag, line 423: Also the spec says `for each header in its header list` so would maybe `of (headers[kHeadersList] || headers)`? Be an idea?

ronag: And I think we can relax `headers instanceof Headers` to `headers.constructor.name === 'Headers'`

KhafraDev: I'm 95% sure that the Headers' iterator is the header list's iterator, although I'm not 100% sure with the new changes

KhafraDev: I can actually make the `headers instanceof Headers` a `headers !== init.headers` I'm pretty sure! I'll let the CI run before I commit the change.

lib/fetch/request.js
KhafraDev: So removing this entirely caused node-fetch tests to fail, however just using the newly constructed headers' header list did work.

mcollina (approved):
lgtm

lib/fetch/request.js
```diff
@@ -419,16 +419,12 @@ class Request {
...
-        this[kState].headersList = new HeadersList([
-          ...this[kState].headersList,
-          ...headers[kHeadersList]
-        ])
+      if (headers.constructor.name === 'Headers') {
+        for (const [key, val] of headers) {
+          this[kHeaders].append(key, val)
+        }
+
+        this[kState].headersList = headers[kHeadersList]
```
ronag, line 427: Why is this needed?

ronag (changes requested):

lib/fetch/request.js
KhafraDev: node-fetch tests without it 🤷 
see: https://github.com/nodejs/undici/runs/6262358645?check_suite_focus=true

ronag: This is weird. I'm not comfortable landing before we know what's going on here.

KhafraDev: So this is why... I think...

this[kState]'s headersList is set to this[kHeaders][kHeadersList] so when appending to 1, both are being appended to. So in the diff below I separate them and fill each separately. That fixes the issue, however I noticed that the spec doesn't say to have 2 different headersList so I think that's part of the confusion.

```diff
diff --git a/lib/fetch/request.js b/lib/fetch/request.js
index 9993b81..6478eda 100644
--- a/lib/fetch/request.js
+++ b/lib/fetch/request.js
@@ -415,19 +415,19 @@ class Request {

       // 3. Empty this’s headers’s header list.
       this[kState].headersList = new HeadersList()
-      this[kHeaders][kHeadersList] = this[kState].headersList
+      this[kHeaders][kHeadersList] = new HeadersList()

       // 4. If headers is a Headers object, then for each header in its header
       // list, append header’s name/header’s value to this’s headers.
       if (headers.constructor.name === 'Headers') {
         for (const [key, val] of headers) {
           this[kHeaders].append(key, val)
+          this[kState].headersList.append(key, val)
         }
-
-        this[kState].headersList = headers[kHeadersList]
       } else {
         // 5. Otherwise, fill this’s headers with headers.
         fillHeaders(this[kState].headersList, headers)
+        fillHeaders(this[kHeaders][kHeadersList], headers)
       }
     }
```

KhafraDev: I don't know if this is an acceptable fix because it seems slower than just letting it be weird lol

ronag: I found the problem.

PR #1405, opened 2022-05-03
feat: make dispatcher truly global

### LINKED ISSUE #1331: Make the global dispatcher truly global
The current implementation of the global dispatcher is that it stores within the library the global agent. This works well when the user is directly using `undici`. Still, suppose the user is using a library that internally uses `undici` and the user didn't install the exact same version of the library. In that case, the global dispatcher API won't work as the user would expect.

To clarify what I think a typical user expects: it should work in the same way as `nock`.
`nock` does not suffer from this problem because it will always use the same version of Node.js, while for `undici` this is not true (yet).

Even if it's kinda ugly, I would suggest attaching the global dispatcher to the Node.js's [`global`](https://nodejs.org/api/globals.html#global) object, so different versions (withing the major possibly) of `undici` can use the same global dispatcher.

If this solution is fine for you (or you have a better idea), I can work on it.

mcollina:
Fixes #1331


targos (approved):

ronag (approved):

delvedor (approved):
Great work! I still think it would be useful to encode in the global dispatcher the version, even if we are not using it right now.
In this case *if* in the future there is the need to do a breaking change, we already have the code to handle it.

mcollina:
@delvedor done.

delvedor (approved):
LGTM

djmadeira:
@delvedor @mcollina this doesn't seem to work. It works fine when using undici's fetch, but elasticsearch uses a pool, which doesn't seem to pick up the global dispatcher at all. Any thoughts on a fix?

markdemich-fl:
If I'm reading this correct, it seems we can use this to set the global agent.  If so, can someone explain how I would do this.  I'm currently npm installing undici and setting that way, but that had complications with my next.js app where I had to modify the my webpack settings to ignore undici.  All I was trying to do was to optionally (based on an environment) set up a proxy like fiddler on my local dev box.  I don't need anything special built into production.

PR #1408, opened 2022-05-03
chore: revert #1349

Fdawgs:
There was nothing wrong with the original pre-commit hook. The only thing #1349 does is introduce an extra dependency that needs to be downloaded (which could be compromised down the road).


mcollina (approved):
thanks, this was very annoying.

LGTM

PR #1410, opened 2022-05-03
fix(Request): URLSearchParams body w/ Headers obj

### LINKED ISSUE #1407: Content type header set incorrectly (5.1.0 regression)
## Bug Description

When sending a URLSearchParams AND a 'headers' to the Request (or fetch) constructor the resulting content type is duplicated.

## Reproducible By

```js
import { URLSearchParams } from 'node:url'

const request = new Request(
  'http:localhost',
  {
    method: 'post',
    body: new URLSearchParams({
      abc: 123
    }),
    headers: {
      Authorization: 'test'
    }
  }
)

console.log('request', {
  headers: Object.fromEntries(request.headers.entries())
})
```

```js
request {
  headers: {
    authorization: 'test',
    'content-type': 'application/x-www-form-urlencoded;charset=UTF-8, application/x-www-form-urlencoded;charset=UTF-8'
  }
}
```

## Expected Behavior
The correct result is returned for undici 5.0.0:
```js
request {
  headers: {
    authorization: 'test',
    'content-type': 'application/x-www-form-urlencoded;charset=UTF-8'
  }
}
```

## Environment

Node 16.15.0

KhafraDev:
Fixes #1407

Although the spec says to "append" the header, setting it should make no difference as the check beforehand ensures that no 'content-type' header exists.

p.s.: I have no idea how the header was being duplicated 😕

### DIFF
--- a/lib/fetch/request.js
+++ b/lib/fetch/request.js
@@ -467,8 +467,8 @@ class Request {
       // not contain `Content-Type`, then append `Content-Type`/Content-Type to
       // this’s headers.
       if (contentType && !this[kHeaders].has('content-type')) {
-        this[kHeaders].append('content-type', contentType)
-        this[kState].headersList.append('content-type', contentType)
+        this[kHeaders].set('content-type', contentType)
+        this[kState].headersList.set('content-type', contentType)
       }
     }
 
--- a/test/fetch/request.js
+++ b/test/fetch/request.js
@@ -7,6 +7,7 @@ const {
   Request
 } = require('../../')
 const { kState } = require('../../lib/fetch/symbols.js')
+const { URLSearchParams } = require('url')
 
 test('arg validation', (t) => {
   // constructor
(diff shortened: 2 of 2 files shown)


lib/fetch/request.js
```diff
@@ -467,8 +467,8 @@ class Request {
       // not contain `Content-Type`, then append `Content-Type`/Content-Type to
       // this’s headers.
       if (contentType && !this[kHeaders].has('content-type')) {
-        this[kHeaders].append('content-type', contentType)
-        this[kState].headersList.append('content-type', contentType)
+        this[kHeaders].set('content-type', contentType)
+        this[kState].headersList.set('content-type', contentType)
```
jefbarn, line 471: I think the reason for the duplication is this[kHeaders][kHeadersList] === this[kState].headersList, making this second call redundant.

See
https://github.com/nodejs/undici/blob/0babdb1371085ff5b1f3fcec28256df2384fb177/lib/fetch/headers.js#L215

KhafraDev: I thought so too, but removing the second call caused some tests (from the node-fetch test suite) to fail.

RafaelGSS: Yes, I think here we should use `.set`. If I remember correctly, the old implementation was using `.set/.push` as well.

RafaelGSS (approved):
Sorry to add those issues. The good part is that we are adding tests to these missing asserts smile

KhafraDev:
No problem at all, there's no way anyone would have noticed these things without tests.

I also discovered a bunch of bugs in the Headers class while working on this PR so this issue will end up making Headers more spec compliant!

mcollina (approved):
lgtm

PR #1419, opened 2022-05-06
fix: Remove unnecessary mismatched trailers check

### LINKED ISSUE #1418: Undici incorrectly rejects responses with trailer header but no matching trailer
## Bug Description

When a server sends a `Trailer: X` header, it's saying that it _may_ send an `X` trailer field after the response. Undici treats this as a commitment, and throws an error if the response has no such trailer, for no apparent reason.

This seems to be an intentional check [here](https://github.com/nodejs/undici/blob/4b5205309f308836885576aa2c6954faf305d50b/lib/client.js#L852-L867) discussed originally in https://github.com/nodejs/undici/issues/432. There's no reference to any spec there that says this is necessary though (and I can't find one) just discussion suggesting maybe it's odd. Undici should follow the spec, not reject responses that seem weird.

There are real servers (e.g. all IPFS nodes, as noted in that issue at the time) that are broken by this in many cases, since they return `trailer: X-Stream-Error` with all streaming responses, to indicate that they _may_ add stream errors after the body in some cases.

Browser fetch & node HTTP don't enforce this, and no spec says this is invalid - this check should just be removed imo.

## Reproducible By

```javascript
const http = require('http');

    res.writeHead(200, {
        trailer: 'test'
    });
    res.end('body');
}).listen(8008);

// Global Undici fetch from Node 18:
fetch('http://localhost:8008').then((res) => {
    console.log('got status', res.status);
    return res.text();
}).then((body) => {
    console.log('got body', body);
})
.catch(console.error);
```

This prints:

```javascript
TypeError: terminated
    at Fetch.onAborted (/tmp/tmp.EolblgSMPd/node_modules/undici/lib/fetch/index.js:1900:49)
    at Fetch.emit (node:events:527:28)
    at Fetch.terminate (/tmp/tmp.EolblgSMPd/node_modules/undici/lib/fetch/index.js:79:10)
    at Object.onError (/tmp/tmp.EolblgSMPd/node_modules/undici/lib/fetch/index.js:2034:34)
    at Request.onError (/tmp/tmp.EolblgSMPd/node_modules/undici/lib/core/request.js:233:27)
    at errorRequest (/tmp/tmp.EolblgSMPd/node_modules/undici/lib/client.js:1728:13)
    at Socket.onSocketClose (/tmp/tmp.EolblgSMPd/node_modules/undici/lib/client.js:1002:5)
    at Socket.emit (node:events:527:28)
    at TCP.<anonymous> (node:net:715:12) {
  [cause]: TrailerMismatchError: Trailers does not match trailer header
      at Parser.onMessageComplete (/tmp/tmp.EolblgSMPd/node_modules/undici/lib/client.js:864:30)
      at wasm_on_message_complete (/tmp/tmp.EolblgSMPd/node_modules/undici/lib/client.js:383:30)
      at wasm://wasm/0002afd2:wasm-function[45]:0x8dc
      at wasm://wasm/0002afd2:wasm-function[56]:0x1ad3
      at wasm://wasm/0002afd2:wasm-function[55]:0xcd7
      at wasm://wasm/0002afd2:wasm-function[21]:0x4e4
      at Parser.execute (/tmp/tmp.EolblgSMPd/node_modules/undici/lib/client.js:515:22)
      at Parser.readMore (/tmp/tmp.EolblgSMPd/node_modules/undici/lib/client.js:484:12)
      at Socket.onSocketReadable (/tmp/tmp.EolblgSMPd/node_modules/undici/lib/client.js:925:10)
      at Socket.emit (node:events:527:28) {
    code: 'UND_ERR_TRAILER_MISMATCH'
  }
}
```

## Expected Behavior

This should work as normal, as if the `trailer` field was not present.

This works correctly in Node HTTP, node-fetch & browser fetch, so this produces regressions for everybody using a fetch polyfill with Node 18.

## Environment

Ubuntu 20.04, using global from Node 18.0.0 or Node 16.14.2 with Undici 5.1.1.

### Additional context

Looks like this was causing errors in the past too, e.g. https://github.com/nodejs/undici/issues/880 says it broke all requests to pinterest.

pimterry:
Fixes #1418.

As in that issue, I don't think there's any reason for this, and it doesn't match other implementations, so this removes the error and check entirely, and flips the error tests to check it works correctly.

### DIFF
--- a/docs/api/Errors.md
+++ b/docs/api/Errors.md
@@ -19,7 +19,6 @@ import { errors } from 'undici'
 | `RequestContentLengthMismatchError`  | `UND_ERR_REQ_CONTENT_LENGTH_MISMATCH` | request body does not match content-length header  |
 | `ResponseContentLengthMismatchError` | `UND_ERR_RES_CONTENT_LENGTH_MISMATCH` | response body does not match content-length header |
 | `InformationalError`                 | `UND_ERR_INFO`                        | expected error with reason                         |
-| `TrailerMismatchError`               | `UND_ERR_TRAILER_MISMATCH`            | trailers did not match specification               |
 
 ### `SocketError`
 
--- a/lib/client.js
+++ b/lib/client.js
@@ -11,7 +11,6 @@ const RedirectHandler = require('./handler/redirect')
 const {
   RequestContentLengthMismatchError,
   ResponseContentLengthMismatchError,
-  TrailerMismatchError,
   InvalidArgumentError,
   RequestAbortedError,
   HeadersTimeoutError,
@@ -425,7 +424,6 @@ class Parser {
 
     this.bytesRead = 0
 
-    this.trailer = ''
     this.keepAlive = ''
     this.contentLength = ''
   }
@@ -615,8 +613,6 @@ class Parser {
     const key = this.headers[len - 2]
     if (key.length === 10 && key.toString().toLowerCase() === 'keep-alive') {
       this.keepAlive += buf.toString()
-    } else if (key.length === 7 && key.toString().toLowerCase() === 'trailer') {
-      this.trailer += buf.toString()
     } else if (key.length === 14 && key.toString().toLowerCase() === 'content-length') {
       this.contentLength += buf.toString()
     }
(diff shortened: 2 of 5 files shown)


mcollina (approved):
lgtm

PR #1423, opened 2022-05-08
feat(Headers): implement `Headers Iterator`

KhafraDev:
See:
- https://github.com/node-fetch/node-fetch/issues/1119#issuecomment-100222049
- https://bugs.chromium.org/p/chromium/issues/detail?id=793204

This implements `Headers Iterator` which is a generator-like... thing... that only has a ` .next()` method. Chrome doesn't implement this entirely correct so I recommend using Firefox to see the correct behavior.

Notes:
- Typescript's Header types are incorrect! You can't use `.return()` or `.throw()` on a Headers iterator (.keys(), .entries(), .values(), Symbol.iterator()).
- Deno does implement this correctly.
- The [`Headers`](https://fetch.spec.whatwg.org/#headers-class) standard doesn't mention Headers Iterator at all so I made this implementation through ✨ trial and error ✨.

### DIFF
--- a/lib/fetch/headers.js
+++ b/lib/fetch/headers.js
@@ -77,6 +77,22 @@ function fill (headers, object) {
   }
 }
 
+function makeHeadersIterator (iterator) {
+  const i = {
+    next: () => {
+      // TODO(@KhafraDev): brand check
+      return iterator.next()
+    },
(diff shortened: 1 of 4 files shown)


mcollina (approved):
lgtm

PR #1424, opened 2022-05-09
fix: import `BodyReadable` and `Dispatcher` correctly

RA80533:
_This PR is a continuation of the changes made in #1059._

`BodyReadable` is imported with the assumption that `allowSyntheticDefaultImports` is enabled. This is restrictive to consumers wishing to use undici's type definitions without its enablement.

The following imports have been fixed:

- `BodyReadable` in test/types/readable.test-d.ts
- `BodyReadable` in types/dispatcher.d.ts
- `Dispatcher` in types/mock-interceptor.d.ts

### DIFF
--- a/test/types/readable.test-d.ts
+++ b/test/types/readable.test-d.ts
@@ -1,5 +1,5 @@
 import { expectAssignable } from 'tsd'
-import BodyReadable from '../../types/readable'
+import BodyReadable = require('../../types/readable')
 import { Blob } from 'buffer'
 
 expectAssignable<BodyReadable>(new BodyReadable())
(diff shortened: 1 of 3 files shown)


test/types/readable.test-d.ts
```diff
@@ -1,5 +1,5 @@
 import { expectAssignable } from 'tsd'
-import BodyReadable from '../../types/readable'
+import BodyReadable = require('../../types/readable')
```
Ethan-Arrowood, line 2: is it possible for us to test both ways? I want to ensure we are not breaking `allowSyntheticDefaultImports=true` users

RA80533: It is! There is no impact for configurations with the setting enabled.

mcollina:
@Ethan-Arrowood could you review?

mcollina (approved):
lgtm

Ethan-Arrowood:
Oh sorry I was waiting for another test case - but if it works still no worries

PR #1425, opened 2022-05-09
fix(Headers): forEach

KhafraDev:
`Headers.forEach` had a bunch of things wrong with it:
- Parameter order was named weird (where `key` was named as `index` in the code).
- `thisArg` was not set to `globalThis` by default.
- `Headers.prototype.forEach.length` was returning 0, instead of 1 (this is another issue I was going to make a PR for eventually).

I have 2 more PRs planned for Headers fixes, so I'm wondering if you would prefer combining those 2 changes into a single PR, or if you would prefer 2 separate PRs?

### DIFF
--- a/lib/fetch/headers.js
+++ b/lib/fetch/headers.js
@@ -354,26 +354,24 @@ class Headers {
     return makeHeadersIterator(this[kHeadersSortedMap])
   }
 
-  forEach (...args) {
+  /**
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

mcollina:
go for them separately!

PR #1431, opened 2022-05-10
fix(Headers Iterator): make spec compliant

KhafraDev:
Follow-up to #1423 that provides hopefully 100% spec compliance.

See: https://github.com/whatwg/fetch/issues/1437#issuecomment-1122701985


mcollina (approved):
lgtm

PR #1441, opened 2022-05-14
feat: implement `CookieStore`

KhafraDev:
Closes https://github.com/nodejs/node/issues/42183

Implements:
- a mostly spec compliant `CookieStore` *
- ability to create a `CookieStore` with 1 or more `set-cookie` headers while remaining spec compliant.
- rfc6265 compliant `set-cookie` header parsing **

\* I follow Chrome's behavior where it mismatches from the spec. This can be seen when setting & getting a cookie name with bom characters in it. Chrome ignores these (?), so likewise the behavior is the same here.
** Without header parsing, `CookieStore` is entirely useless in node. With it, I can see it getting used a fair bit for parsing cookies server-side.

I haven't exposed it yet and it needs more tests, but nyc wasn't giving me good coverage reports anymore.

`CookieStoreFrom` also needs a way to set a hostname to check for valid/invalid cookies (would solve the issue of node not having an origin).


mcollina:
how would this work in combination with fetch?

KhafraDev:
Something along the lines of https://github.com/whatwg/fetch/issues/1384

I think it has more 'substantial' use server-side though.

```js

   const cookieStore = req.headers['set-cookie'] ? CookieStoreFrom(req.headers['set-cookie'] : null

  if (cookieStore === null ) {
    res.statusCode = 404
    res.end('no cookies')
    return
  }

  // ...
}).listen(0)

File: lib/cookie-store/constants.js [hunk 1]
```diff
@@ -0,0 +1,14 @@
+// https://wicg.github.io/cookie-store/#cookie-maximum-attribute-value-size
```
mcollina, line 1: add use strict at the top

lib/cookie-store/cookie-store.js
```diff
@@ -0,0 +1,350 @@
...
+const { createDeferredPromise } = require('../fetch/util')
+const { kEnumerableProperty } = require('../core/util')
+const { queryCookies, setCookie, deleteCookie } = require('./util')
+const { parseSetCookie, parseUnparsedAttributes } = require('./parse')
+const { toUSVString } = require('util')
+
+const kCookieStoreConstructable = Symbol('CookieStore constructable')
+
+class CookieStore extends EventTarget {
+  #cookieStore = []
```
mcollina, line 12: avoid private properties, they are slow in Node v16 (should be ok in v18), prefer "private" symbols.

mcollina (commented):
good work!

mcollina:
Could you add some test for fetch integration and docs?

lib/cookie-store/cookie-store.js
KhafraDev: Ever since switching to v18, I've been using them everywhere now that they're fast! I'll change it to a symbol

KhafraDev:
Fetch integration hasn't been done yet, I wasn't sure if it would be in scope for this PR.

mcollina:
I would restrain from adding this without further fetch() integration.

KhafraDev:
That's fair, I'll convert it to a draft PR until there's more discussion with integrating into fetch or there's been more discussion.

blittle:
@KhafraDev I'm not sure if your example above would work with multiple cookies:

```js
const server = createServer(async (req, res) => {
   // there might be multiple set-cookie headers, and req.headers['set-cookie'] currently does not properly return them
   const cookieStore = req.headers['set-cookie'] ? CookieStoreFrom(req.headers['set-cookie']) : null

  // instead what if `cookieStore` was just a property on the request?
  const cookieStoreReq = req.cookieStore;

  if (cookieStore === null ) {
    res.statusCode = 404
    res.end('no cookies')
    return
  }

  const sessionCookie = await cookieStore.get('sessionId')

  if (sessionCookie && await matchesFromSomewhere(sessionCookie)) {
    res.statusCode = 200
    res.end('Welcome!')
    return
  }

  // ...
}).listen(0)
```

jimmywarting:
What's weird is that you create a server and listen for incomming `set-cookies` headers
```js
const server = createServer(async (req, res) => {
   // there might be multiple set-cookie headers, and req.headers['set-cookie'] currently does not properly return them
   const cookieStore = req.headers['set-cookie'] ? CookieStoreFrom(req.headers['set-cookie']) : null
```
when clients makes request to a server then you are only sending cookies.
so it should rather be something like `CookieStoreFrom(req.header.cookies)`

```js
  // instead what if `cookieStore` was just a property on the request?
  const cookieStoreReq = req.cookieStore;
```
i like that, and it was what i proposed in my original feature request. it would always exist on the request (even if no cookie are being sent) it should only populate the newly created cookiestore with new values. so you don't even have to check if it's null. 

you would only have to do:
```js
  const sessionCookie = await req.cookieStore.get('sessionId')
  if (!sessionCookie) return res.end('not logged in')
```

blittle:
@KhafraDev oh yes, you are right. It's the node request object, not undici's `Request`.

PR #1449, opened 2022-05-18
feat: Query string params

### LINKED ISSUE #1448: Support for query params
## This would solve...

Supporting query params would decrease amount of boilerplate code that developers need to write

```
    const { statusCode, headers, trailers, body } = await request("localhost:8080/", {
      method: 'GET',
      params: {
        pageNum: 1,
        pageSize: 10
    }
    })
```

## I have also considered...

Embedding query params directly into the URL, but it is very cumbersome or requires writing a wrapper in every project.

## Additional context

If this is approved, I can contribute a PR.

kibertoad:
fixes #1448


test/client.js
```diff
@@ -251,6 +294,16 @@ test('head with host header', (t) => {
   })
 })
 
+function serverRequestParams (t, expected) {
+  return function (req, res) {
+    const queryParams = querystring.parse(req.url.replace('/?', '')) // We remove opening /? because parser assumes it to be a part of param name
+    t.strictSame(JSON.parse(JSON.stringify(queryParams)), expected)
```
kibertoad, line 300: For some reason TAP is unhappy here without JSON back-and-forth conversion, because it find a mismatch between `Object` (expected, object literal) and `Null Object` (actual, result of querystring.parse). My googling was unsuccessful here, I have no idea what is going on and how to address it, help from tape gurus would be most welcome here.

lib/core/request.js
```diff
@@ -97,7 +99,7 @@ class Request {
 
     this.upgrade = upgrade || null
 
-    this.path = path
+    this.path = params && Object.keys(params).length > 0 ? `${path}?${querystring.stringify(params)}` : path
```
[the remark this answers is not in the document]
kibertoad: Addressed in the same way how Axios does it.

One thing to be aware of: URLSearchParams stringifies array into a single string `ELEMENT_1,ELEMENT_2`, which is non-reversible and is not how majority of users send arrays (typical pattern is x=ELEMENT_1,&x=ELEMENT_2). Should I keep the behaviour as implemented by `URLSearchParams` and leave it to users to prestringify their data the way they need, should I add optional array stringifier param, or implement more conventional array conversion within undici itself?

kibertoad:
Failed test seems to be flaky.

lib/core/request.js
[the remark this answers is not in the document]
targos: > URLSearchParams stringifies array into a single string ELEMENT_1,ELEMENT_2

That's just because URLSearchParams doesn't care about the value type. I guess it does something equivalent to `String(value)` and that's arrays are stringified. I don't think we necessarily need to do the same thing as URLSearchParams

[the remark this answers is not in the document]
kibertoad: @targos What would be your preference? `x=ELEMENT_1&x=ELEMENT_2` conversion for arrays?

lib/core/request.js
```diff
@@ -97,7 +98,13 @@ class Request {
 
     this.upgrade = upgrade || null
 
-    this.path = path
+    if (params && Object.keys(params).length > 0) {
```
ronag, line 101:
```suggestion
    if (params) {
```

lib/core/request.js
```diff
@@ -97,7 +98,13 @@ class Request {
...
+      const searchParams = new URLSearchParams(params)
+      const delimiter = path.indexOf('?') === -1 ? '?' : '&'
```
ronag, line 103: What if path just ends with ? without any params?

ronag (changes requested):

lib/core/request.js
kibertoad: Tried `https://openlibrary.org/api/books?&bibkeys=ISBN:0201558025` and `https://fakestoreapi.com/products?&limit=3`, it still works as expected, so it seems like it's a generally acceptable input. Axios is not protecting against this case either.

lib/core/util.js
```diff
@@ -26,6 +26,57 @@ function isBlobLike (object) {
...
+
+    if (!Array.isArray(val)) {
+      val = [val]
+    }
+
+    for (let v of val) {
+      if (isDate(v)) {
+        v = v.toISOString()
+      } else if (isObject(v)) {
+        v = JSON.stringify(v)
```
mcollina, line 60: I would prefer if we did not do the JSON.stringify in this case. It's too opinionated.

docs/api/Dispatcher.md
```diff
@@ -194,6 +194,7 @@ Returns: `Boolean` - `false` if dispatcher is busy and further dispatch calls wo
 * **method** `string`
 * **body** `string | Buffer | Uint8Array | stream.Readable | Iterable | AsyncIterable | null` (optional) - Default: `null`
 * **headers** `UndiciHeaders | string[]` (optional) - Default: `null`.
+* **params** `Record<string, any | null` (optional) - Default: `null` - Query string params to be embedded in the request URL.
```
mcollina, line 197: I would call it `query` instead.

lib/core/util.js
kibertoad: Should we throw on object or ignore it?

lib/core/request.js
targos: `path.indexOf('?') === -1` -> `path.includes('?')`

lib/core/request.js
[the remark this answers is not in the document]
targos: yes

lib/core/request.js
kibertoad: isn't indexOf faster?

targos: I don't see why it should be faster, but anyway it's unlikely to matter compared to the rest of the work done here.

lib/core/util.js
mcollina: I think throwing would be a predictable behavior.

lib/core/util.js
```diff
@@ -26,6 +26,57 @@ function isBlobLike (object) {
...
+    const hashmarkIndex = url.indexOf('#')
+    if (hashmarkIndex !== -1) {
+      url = url.slice(0, hashmarkIndex)
+    }
+
+    url += (url.indexOf('?') === -1 ? '?' : '&') + serializedParams
+  }
+
+  return url
+}
```
ronag, line 78: This is getting too complicated. What was wrong with the previous version with URLSearchParams?

kibertoad: It did not handle all of the relevant cases: e. g. arrays, dates or objects: implicit toString() is not a particularly good solution.

I'm happy to drop any parts of this logic that you believe to be complicated; I've already replaced the most opinionated part (object serialization) and happy to remove anything else that you believe to be excessive (e. g. hashmark handling)

lib/core/util.js
kibertoad: Done.

lib/core/request.js
kibertoad: Did the change, but I can share a benchmark later, there is a difference, but yeah, should be negligible in the context.

lib/core/request.js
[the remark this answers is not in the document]
kibertoad: Done.

lib/core/util.js
ronag: I'd prefer to just do the exact same thing that `URLSearchParams` does.

ronag (commented):
Not blocking but I don't feel this belongs in undici.

lib/core/util.js
kibertoad: @ronag It is significantly less useful than the current implementation, especially with regards to the handling of arrays, and is going to be a better option in no cases: no framework is capable of automatically parsing `[1,2]` as an array, while current implementation will be automatically decoded by pretty much anything with the support of arrays.

@targos @mcollina Could you please provide a second/third opinion?

mcollina (approved):
lgtm

I think this is a useful utility. This is simple and we could add it.

I see @ronag point on this becoming too complex, however this would be really nice to have.

ronag (commented):
If we land this I at least ask that we have full test coverage of the added code and more tests in general. I can see many potential edge cases here. I don't mind if we adapt tests from other client libs.

lib/core/util.js
```diff
@@ -26,6 +26,57 @@ function isBlobLike (object) {
   )
 }
 
+function encode (val) {
+  return encodeURIComponent(val).replace(/%3A/gi, ':').replace(/%24/g, '$').replace(/%2C/gi, ',').replace(/%20/g, '+').replace(/%5B/gi, '[').replace(/%5D/gi, ']')
+}
```
ronag, line 31: What does this do? Can we add a comment explaining?

lib/core/util.js
```diff
@@ -26,6 +26,57 @@ function isBlobLike (object) {
...
+
+  const serializedParams = parts.join('&')
+
+  if (serializedParams) {
+    const hashmarkIndex = url.indexOf('#')
+    if (hashmarkIndex !== -1) {
+      url = url.slice(0, hashmarkIndex)
+    }
+
+    url += (url.includes('?') ? '&' : '?') + serializedParams
```
ronag, line 74: What happens if the same query param is twice? Should we maybe just throw if the path already includes `?`?

kibertoad: Same query param being included multiple times is perfectly valid, that's a common pattern for passing arrays.
Why throw an error if we can handle this use-case with this one-liner easily? No strong opinion here, can drop if you feel strongly about it, I just see it as decreased value.

lib/core/util.js
kibertoad: Whole buildURL block is simplified adaptation of how Axios is doing it. I can add tests for when this is actually used as well as a comment, including the reference to source.

kibertoad:
Full test coverage coming up :)

lib/core/util.js
ronag: > Same query param being included multiple times is perfectly valid, that's a common pattern for passing arrays.

I see.

> No strong opinion here, can drop if you feel strongly about it, I just see it as decreased value.

I'd prefer no surprises. Please throw for now and we can always relax it later (easier than the other direction).

kibertoad:
@ronag This is ready for rereview. Please let me know if you want additional tests for any cases.

lib/core/util.js
```diff
@@ -26,6 +26,60 @@ function isBlobLike (object) {
...
+
+function isObject (val) {
+  return val !== null && typeof val === 'object'
+}
+
+// based on https://github.com/axios/axios/blob/63e559fa609c40a0a460ae5d5a18c3470ffc6c9e/lib/helpers/buildURL.js (MIT license)
+// this escapes all non-browser-url friendly characters, but restores the ones that are supported by browsers
+function encode (val) {
+  return encodeURIComponent(val).replace(/%3A/gi, ':').replace(/%24/g, '$').replace(/%2C/gi, ',').replace(/%20/g, '+').replace(/%5B/gi, '[').replace(/%5D/gi, ']')
+}
```
ronag, line 41: Why do we need this?

> ll non-browser-url friendly characters, but restores the ones that are supported by browsers

We are not in a browser?

kibertoad: Something glitched in my head, sorry. I think restoring non-escaped symbols was purely to conserve bandwidth, and I'm OK to just use encodeURIComponent for everything. Will do the change.

lib/core/request.js
```diff
@@ -97,7 +98,11 @@ class Request {
 
     this.upgrade = upgrade || null
 
-    this.path = path
+    if (query) {
+      this.path = util.buildURL(path, query)
+    } else {
+      this.path = path
+    }
```
ronag, line 105:
```suggestion
    this.path = query ? util.buildURL(path, query) : path
```

lib/core/util.js
```diff
@@ -26,6 +26,59 @@ function isBlobLike (object) {
...
+  return toString.call(val) === '[object Date]'
+}
+
+function isObject (val) {
+  return val !== null && typeof val === 'object'
+}
+
+// this escapes all non-uri friendly characters, but restores : which is commonly used for date formatting
+function encode (val) {
+  return encodeURIComponent(val).replace(/%3A/gi, ':')
```
ronag, line 39: Why restore `:`? I don't understand the use case?

lib/core/util.js
```diff
@@ -26,6 +26,59 @@ function isBlobLike (object) {
...
+      parts.push(encode(key) + '=' + encode(v))
+    }
+  }
+
+  const serializedParams = parts.join('&')
+
+  if (serializedParams) {
+    const hashmarkIndex = url.indexOf('#')
+    if (hashmarkIndex !== -1) {
+      url = url.slice(0, hashmarkIndex)
```
ronag, line 73: This is a bit magical...

kibertoad: We can drop it, if you like

lib/core/util.js
```diff
@@ -26,6 +26,59 @@ function isBlobLike (object) {
...
+    }
+
+    url += '&' + serializedParams
```
ronag, line 76:
```suggestion
    url += '?' + serializedParams
```

lib/core/util.js
```diff
@@ -26,6 +26,59 @@ function isBlobLike (object) {
...
+      continue
+    }
+
+    if (!Array.isArray(val)) {
+      val = [val]
+    }
+
+    for (let v of val) {
+      if (isDate(v)) {
+        v = v.toISOString()
```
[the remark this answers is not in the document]
ronag: I guess it's the same a `JSON.stringify` does so maybe not too magical.

lib/core/util.js
kibertoad: Escaping `:` is going to mess up date format and complicate rather frequent use-case.
I'm having second thoughts about escaping anything, TBH, especially since Undici is not supposed to be used from browser. What do you think about removing escaping completely and let user do it if they need to?

lib/core/util.js
ronag: Throw if `#`?

lib/core/util.js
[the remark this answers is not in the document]
kibertoad: It is more on opinionated side, I agree. Should we throw an error, same way as with Objects, and let users format themselves?

lib/core/util.js
```diff
@@ -26,6 +26,59 @@ function isBlobLike (object) {
...
+// this escapes all non-uri friendly characters, but restores : which is commonly used for date formatting
+function encode (val) {
+  return encodeURIComponent(val).replace(/%3A/gi, ':')
+}
+
+// based on https://github.com/axios/axios/blob/63e559fa609c40a0a460ae5d5a18c3470ffc6c9e/lib/helpers/buildURL.js (MIT license)
+function buildURL (url, queryParams) {
+  if (url.includes('?')) {
+    throw new Error('Query params cannot be passed when url contains "?" already.')
+  }
```
ronag, line 46:
```suggestion
  if (url.includes('?') || url.includes('#')) {
    throw new Error('Query params cannot be passed when url contains "?" or "#" already.')
  }
```

lib/core/util.js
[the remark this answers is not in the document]
ronag: I'd prefer throwing.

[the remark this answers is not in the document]
ronag: Just to be clear I'm happy to revisit these throwing cases later. But in order to land this PR sooner rather than later I think throwing is appropriate.

lib/core/util.js
ronag: I'm ok with removing escaping. What does got do?

ronag: I think we actually need to escape or we might end up with invalid URLs, e.g. if they contain `?`, `&` etc...

ronag: I think `encodeURIComponent` is sufficient for now.

ronag: Just need to clarify in the docs that we expect unencoded keys and values.

kibertoad: I'll compare with got implementation later today.

kibertoad: What I'm concerned about is that there is not much of an escape hatch there. If receiving API is not capable of parsing escaped query params, user will be forced to just hardcode entire URL, which we are not encoding; and it is likely going to be confusing for the user, why are we treating query params differently.

ronag: This is why I'd prefer not to have these kinds of things in undici. There is no perfect solution. I think not escaping is worse than escaping. If the receiving API needs something else then the user can always build the path themselves.

lib/core/util.js
```diff
@@ -26,6 +26,59 @@ function isBlobLike (object) {
...
+      continue
+    }
+
+    if (!Array.isArray(val)) {
+      val = [val]
+    }
+
+    for (const v of val) {
+      if (isDate(v)) {
+        throw new Error('Passing date as a query param is not supported, please serialize to string up-front')
```
ronag, line 60: Date is an object so the check below is sufficient.

lib/core/util.js
kibertoad: got is using URLSearchParams, and there are already issues with that: https://github.com/sindresorhus/got/issues/1180
If we go with escaping, I would suggest following axios example instead of got - they were around for much longer, and had all the time in the world to work out the kinks of real world usage.

targos: I'm not sure I understand the use case. Is it that some server implementations don't decode query parameters at all?

ronag: I'd prefer if we just keep it simple and use `encodeURIComponent`.

lib/core/util.js
ronag: This is not resolved yet.

lib/core/util.js
```diff
@@ -26,6 +26,59 @@ function isBlobLike (object) {
...
+function encode (val) {
+  return encodeURIComponent(val).replace(/%3A/gi, ':')
+}
+
+// based on https://github.com/axios/axios/blob/63e559fa609c40a0a460ae5d5a18c3470ffc6c9e/lib/helpers/buildURL.js (MIT license)
+function buildURL (url, queryParams) {
+  if (url.includes('?') || url.includes('#')) {
+    throw new Error('Query params cannot be passed when url already contains "?" or "#".')
+  }
+
```
ronag, line 47: Check that queryParams is a POJO?

lib/core/util.js
targos: I agree

lib/core/util.js
kibertoad: done

lib/core/util.js
kibertoad: done

lib/core/util.js
kibertoad: sorry, change was done and test was added, but some garbage left behind

lib/core/util.js
kibertoad: done

Issue #1205, opened 2022-02-05
Error handling?


calvintwr:
## This would solve...

It doesn’t look like there’s anything in the documentation on error handling, except perhaps only statusCode.

For comprehensiveness, I would suggest it be considered important to include error handling in the first example.

ronag:
PR is welcome

calvintwr:
Sure. Can you point me to the code that does error handling?

kibertoad:
@ronag Would a PR implementing (opt-in) error throwing for 4xx and 5xx responses be accepted?

mcollina:
I do not see a problem with having it as an option.

PR #1453, opened 2022-05-18
feat: Option to throw on error status codes

kibertoad:
refs #1205


ronag:
I think this is outside of the scope of what undici should be doing.

kibertoad:
@ronag What is the downside? It is a very simple piece of logic that significantly improves developer experience. Reimplementing it for every project is wasteful boilerplate.

kibertoad:
@mcollina Could use your perspective here as well.

test/client.js
```diff
@@ -315,6 +315,64 @@ test('basic get with query params partially in path', (t) => {
...
+      keepAliveTimeout: 300e3
+    })
+    t.teardown(client.close.bind(client))
+
+    const signal = new EE()
+    client.request({
+      signal,
+      path: '/',
+      method: 'GET',
+      throwOnError: true
```
mcollina, line 338: We need a better name for the option, in this example is not throwing at all.

types/dispatcher.d.ts
```diff
@@ -57,6 +57,8 @@ declare namespace Dispatcher {
     headersTimeout?: number | null;
     /** The timeout after which a request will time out, in milliseconds. Monitors time between receiving body data. Use 0 to disable it entirely. Defaults to 30 seconds. */
     bodyTimeout?: number | null;
+    /** Whether Undici should throw an error upon receiving a 4xx or 5xx response from the server. Defaults to false */
+    throwOnError?: boolean;
```
mcollina, line 61: the additional type needs a test

mcollina (commented):
I'm in favor for this change too.

ronag:
I'm on the losing end again 😢 ... can we at least make the error signature match that of https://www.npmjs.com/package/http-errors?

test/client.js
kibertoad: Any ideas? What aspects would you like emphasized? Axios uses `validateStatus` for similar config, got uses `throwHttpErrors`.

kibertoad:
@ronag Hopefully everyone will win after this feature is properly refined. Can you elaborate on what changes would you like to be made on error structure? I followed structure of other Undici errors, should this one be different for some reason?

kibertoad:
@ronag I gave it a shot, please let me know if revised error structure is what you wanted.

types/dispatcher.d.ts
kibertoad: done

lib/api/api-request.js
```diff
@@ -89,6 +91,13 @@ class RequestHandler extends AsyncResource {
     const headers = this.responseHeaders === 'raw' ? util.parseRawHeaders(rawHeaders) : util.parseHeaders(rawHeaders)
 
     if (callback !== null) {
+      if (this.throwOnError && statusCode > 399) {
```
ronag, line 94:
```suggestion
      if (this.throwOnError && statusCode >= 400) {
```

lib/core/request.js
```diff
@@ -71,6 +72,8 @@ class Request {
 
     this.bodyTimeout = bodyTimeout
 
+    this.throwOnError = throwOnError !== undefined ? throwOnError : false
```
ronag, line 75:
```suggestion
    this.throwOnError = throwOnError === true
```

ronag (approved):

test/client.js
kibertoad: @ronag any preference on the name?

mcollina (approved):
lgtm

kibertoad:
@mcollina What is the release cadence? Any estimates on when new version can be published?

kibertoad:
I think some flakiness for Windows/Node 16 build was introduced again with this PR. I'll try to fiddle with the new tests added here, helped in the past.

mcollina:
thanks, once that's stable I'll ship a release.

s97712:
This will lose the error stack.

kibertoad:
@s97712 can you elaborate? there is no error stack on error code replies per se

s97712:
If I use throwOnError to throw an error, I will get the following stack trace. 
This will be difficult to pinpoint the exact location of the error.
```bash
  ResponseStatusCodeError: Bad Request

  - util.js:36 getResolveErrorBodyCallback
    [admin]/[undici@5.23.0]/[undici]/lib/api/util.js:36:34

  - task_queues:95 processTicksAndRejections
    node:internal/process/task_queues:95:5

```

PR #1454, opened 2022-05-18
fix: get `set-cookie` header with `credentials: include`

### LINKED ISSUE #1262: Undici strips out set-cookie headers, even when "credentials: 'include'" is set
## Bug Description
Undici strips out `set-cookie` headers, which is expected per the `fetch` API spec. What is not expected though is that the `set-cookie` headers are stripped out in the presence of `{ credentials: 'include' }` in a `fetch` request. Here is the relevant portion of the `fetch` API spec:
https://fetch.spec.whatwg.org/#example-cors-with-credentials

### server.js
```
const express = require('express')
const app = express()

app.get('/', (req, res) => {
  res.set({
    abc: '123',
    'set-cookie': 'hello=world',
  });
  res.send('Hello World!');
})

app.listen(3333, () => {
  console.log(`Example app listening on port ${port}`)
})
```

### client.js
```

async function get() {
  const headers = await fetch('http://localhost:3333/', { credentials: 'include' })
    .then(async res => {
      for await (const chunk of res.body) {}
      return res.headers
    });

  console.log(`abc: ${headers.get('abc')}`);
  console.log(`set-cookie: ${headers.get('set-cookie')}`);
}

get();
```

With the server script running, the client script returns:
```
abc: 123
set-cookie: null
```

## Expected Behavior

In the presence of `{ credentials: 'include' }` I would expect `set-cookie` headers to _not_ be stripped from the response. The expected output of the client script should be:
```
abc: 123
set-cookie: hello=world
```

## Environment

Mac OS 12.2.1
Node v16.13.1

### Additional context

There is a `TODO` in Undici's `fetch` code where this behavior should be implemented:
https://github.com/nodejs/undici/blob/main/lib/fetch/index.js#L1369-L1375

KhafraDev:
Fixes #1262 

When `credentials` is equal to `include`, the `set-cookie` header(s) should not be stripped.

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -654,7 +654,7 @@ async function mainFetch (fetchParams, recursive = false) {
     // Set response to the following filtered response with response as its
     // internal response, depending on request’s response tainting:
     if (request.responseTainting === 'basic') {
-      response = filterResponse(response, 'basic')
+      response = filterResponse(response, 'basic', request.credentials)
     } else if (request.responseTainting === 'cors') {
       response = filterResponse(response, 'cors')
     } else if (request.responseTainting === 'opaque') {
--- a/lib/fetch/response.js
+++ b/lib/fetch/response.js
@@ -407,7 +407,7 @@ function makeFilteredHeadersList (headersList, filter) {
 }
 
 // https://fetch.spec.whatwg.org/#concept-filtered-response
-function filterResponse (response, type) {
+function filterResponse (response, type, credentials) {
   // Set response to the following filtered response with response as its
   // internal response, depending on request’s response tainting:
   if (type === 'basic') {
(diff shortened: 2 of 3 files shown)


mcollina (approved):
lgtm

ronag:
IMHO this is wrong and does not follow the spec. I believe the correct thing to do is to revert this and implement this section https://github.com/nodejs/undici/blob/5478cdd38eb057a1e86c15cf418eeb90486470b2/lib/fetch/index.js#L1806-L1811.

ronag:
@KhafraDev

mcollina:
Ok I'll push a revert, seemed ok at first look.

mcollina:
revert pushed.

ronag:
The logic added here is not part of the spec. So either we are doing something differently (not good) or there is a bug in the spec.

The third option is that we have not implemented something in the spec or done it incorrectly. However this PR does not address that.

KhafraDev:
Yep, the spec doesn't allow `set-cookie` headers by design. However, [this comment](https://github.com/nodejs/undici/issues/1262#issuecomment-1131839918) shows how just about every other non-browser platform allows this. Similar alterations from the spec regarding headers has been done with the `redirect: "manual"` change.

ronag:
Can we raise this to wintercg?

ronag:
What does deno and cloudflare do?

ronag:
This conversation is happening in two places now. Maybe continue in https://github.com/nodejs/undici/issues/1262.

PR #1456, opened 2022-05-20
docs: Add documentation for reply accepting a function as the data parameter

p16:
Following up from this discussion https://github.com/nodejs/undici/discussions/1185 I thought there was a missing bit in the doc around the fact that the `reply` `data` parameter can be a function that will have the mocked request options as input .

### DIFF
--- a/docs/best-practices/mocking-request.md
+++ b/docs/best-practices/mocking-request.md
@@ -103,7 +103,7 @@ const badRequest = await bankTransfer('1234567890', '100')
 
 ## Reply with data based on request
 
-If the mocked response needs to be dinamically derived from the reuqest parametes, you can provide a function instead of an object to `reply`
+If the mocked response needs to be dynamically derived from the request parameters, you can provide a function instead of an object to `reply`
 
 ```js
 mockPool.intercept({
--- a/docs/best-practices/mocking-request.md
+++ b/docs/best-practices/mocking-request.md
@@ -101,4 +101,36 @@ const badRequest = await bankTransfer('1234567890', '100')
 // subsequent request to origin http://localhost:3000 was not allowed (net.connect disabled)
 ```
 
+## Reply with data based on request
 
+If the mocked response needs to be dinamically derived from the reuqest parametes, you can provide a function instead of an object to `reply`
+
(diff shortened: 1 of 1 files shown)


ronag (approved):

p16:
The failed test

```
not ok 9 - test/fetch/headers.js # time=4684.681ms
  ---
  env: {}
  file: test/fetch/headers.js
  timeout: 60000
  command: C:\hostedtoolcache\windows\node\16.15.0\x64\node.exe
  args:
    - --expose-gc
    - --experimental-wasm-simd
    - test/fetch/headers.js
  stdio:
    - 0
    - pipe
    - 2
  cwd: D:\a\undici\undici
  exitCode: null
  signal: SIGTERM
  ...
```

is it a know "flaky" test? it seems something related to the platform more than to the test 🤔

mcollina (approved):
lgtm

KhafraDev (approved):

docs/best-practices/mocking-request.md
```diff
@@ -101,4 +101,36 @@ const badRequest = await bankTransfer('1234567890', '100')
 // subsequent request to origin http://localhost:3000 was not allowed (net.connect disabled)
 ```
 
+## Reply with data based on request
 
+If the mocked response needs to be dinamically derived from the reuqest parametes, you can provide a function instead of an object to `reply`
```
Fdawgs, line 106:
```suggestion
If the mocked response needs to be dynamically derived from the request parameters, you can provide a function instead of an object to `reply`
```

Fdawgs (changes requested):

mcollina:
Please update the PR title an action fails.

p16:
@mcollina @Fdawgs It seems that the build on latest windows and node 16 is failing on`test/redirect-stream.js` (same test result shown in previous message). There is nothing more than what I reported above in the log of that failure. I do not have a window machine ready to test this, is there anyone else that can have a look?

kibertoad:
Documentation deployment is failing now: https://github.com/nodejs/undici/runs/6558409571?check_suite_focus=true

PR #1457, opened 2022-05-20
fix(fetch): replace `instanceof FormData` check

NMinhNguyen:
See https://github.com/nodejs/undici/pull/1451#issuecomment-1131619657

I've tested this using `esbuild` with `--minify` and `util.isFormData` still works on a minified `FormData` class

### DIFF
--- a/lib/fetch/body.js
+++ b/lib/fetch/body.js
@@ -71,7 +71,7 @@ function extractBody (object, keepalive = false) {
 
     // Set source to a copy of the bytes held by object.
     source = new Uint8Array(object)
-  } else if (object instanceof FormData || util.isFormDataLike(object)) {
+  } else if (util.isFormDataLike(object)) {
     const boundary = '----formdata-undici-' + Math.random()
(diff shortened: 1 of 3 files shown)


ronag (commented):
I don't have an opinion on this. +/- 0. @mcollina @KhafraDev @Ethan-Arrowood @targos thoughts?

mcollina (approved):
lgtm

mcollina (commented):
Could you add a test?

NMinhNguyen:
> Could you add a test?

I'm not sure what sort of test I could add that tests this edge case. Also `util.isFormDataLike` is not part of the public API. The original PR that was merged didn't have a test either, as I suspect it's quite tricky to test.

NMinhNguyen:
Added a simple test, let me know if it's sufficient @mcollina: https://github.com/nodejs/undici/compare/213706c71daa48705cf5a4da096ac0f23c8cbefb..2e5814b077e2695eadcc7b9b87e8e8899031e27c

mcollina:
Once we have some time. At least once a month, ideally every two weeks.

PR #1459, opened 2022-05-20
fix: Attempt to reproduce flakiness

kibertoad:
There were random failures under Windows/Node 16 lately, let's see if we can reproduce them consistently.

### DIFF
--- a/test/client.js
+++ b/test/client.js
@@ -316,7 +316,7 @@ test('basic get with query params partially in path', (t) => {
 })
 
 test('basic get returns 400 when configured to throw on errors (callback)', (t) => {
-  t.plan(6)
+  t.plan(5)
 
   const server = createServer((req, res) => {
     res.statusCode = 400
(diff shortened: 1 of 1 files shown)


kibertoad:
Attempt 1/3: green.

kibertoad:
Attempt 2/3: green.

kibertoad:
Attempt 3/3 - fail.
@mcollina I've looked at past commits, and I don't think that flakiness was introduced by any of my PRs, actually. The earliest that I see is this: https://github.com/nodejs/undici/issues/1359 (https://github.com/nodejs/undici/commit/ac3191e5430756e7d83aa039ffca13dcf3e35b54)

Considering that it is extremely arbitrary (only Windows, only Node 16) and I can't reproduce it locally, I have no idea how to approach it :(

kibertoad:
Opened #1460

PR #1464, opened 2022-05-22
fix: mock node fetch with `options.body` fn

### LINKED ISSUE #1447: unexpected behaviours between MockAgent and node's bundled fetch
## Bug Description

Intercepting undici's own fetch body works but node's doesn't. This is building on top of the global global dispatcher shipped with 5.2.0 which is then used by Node 18.2.0 bundled fetch.

## Reproducible By

```js
import * as undici from 'undici' // 5.2.0

const mockAgent = new undici.MockAgent()
mockAgent.disableNetConnect()
undici.setGlobalDispatcher(mockAgent)
const pool = mockAgent.get('https://example.com')

pool.intercept({
  path: '/path',
  method: 'POST',
  body(body) {
    console.log(body)
    return true
  }
}).reply(200, { ok: 1 }).times(2)

const url = new URL('https://example.com/path').href

// undici fetch v5.2.0 from node_modules
await undici.fetch(url, {
  method: 'POST',
  body: JSON.stringify({ foo: 'bar' })
})

// undici fetch v5.2.0 from node 18.2.0 global
await globalThis.fetch(url, {
  method: 'POST',
  body: JSON.stringify({ foo: 'bar' })
})
```

The above logs

```
{"foo":"bar"} // undici fetch v5.2.0 from node_modules
Object [AsyncGenerator] {} // undici fetch v5.2.0 from node 18.2.0 global
```

## Expected Behavior

It is expected that the string body is passed into the matching function both times.

## Environment

Node.js 18.2.0
Undici 5.2.0

KhafraDev:
Fixes #1447 

No tests but I could theoretically add them in (by calling `npm run node:build` and then using that file, but seems pretty hacky).

Verify the fix works:

test.mjs:
```js
import * as undici from './index.js' // 5.2.0
import { fetch } from './undici-fetch.js'

const mockAgent = new undici.MockAgent()
mockAgent.disableNetConnect()
undici.setGlobalDispatcher(mockAgent)
const pool = mockAgent.get('https://example.com')

pool.intercept({
  path: '/path',
  method: 'POST',
  body (body) {
    console.log(body)
    return true
  }
}).reply(200, { ok: 1 }).times(2)

const url = new URL('https://example.com/path').href

// undici fetch v5.2.0 from node_modules
await undici.fetch(url, {
  method: 'POST',
  body: JSON.stringify({ foo: 'bar' })
})

// undici fetch v5.2.0 from node 18.2.0 global
await fetch(url, {
  method: 'POST',
  body: JSON.stringify({ foo: 'bar' })
})
```

```
npm run build:node && node test.mjs
```

Output:
```
{"foo":"bar"}
{"foo":"bar"}
{"foo":"bar"}
```
The issue of the function being called 3 times is not an issue with this fix (see comment in the issue)

Everything is done in the repo's root directory.

### DIFF
--- a/lib/mock/mock-agent.js
+++ b/lib/mock/mock-agent.js
@@ -96,6 +96,12 @@ class MockAgent extends Dispatcher {
     this[kNetConnect] = false
   }
 
+  // This is required to bypass issues caused by using global symbols - see:
+  // https://github.com/nodejs/undici/issues/1447
+  get isActive () {
+    return this[kIsMockActive]
+  }
+
   [kMockAgentSet] (origin, dispatcher) {
     this[kClients].set(origin, new FakeWeakRef(dispatcher))
   }
--- a/lib/mock/mock-agent.js
+++ b/lib/mock/mock-agent.js
@@ -98,7 +98,7 @@ class MockAgent extends Dispatcher {
 
   // This is required to bypass issues caused by using global symbols - see:
   // https://github.com/nodejs/undici/issues/1447
-  get isActive () {
+  get isMockActive () {
     return this[kIsMockActive]
   }
 
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -1911,7 +1911,7 @@ async function httpNetworkFetch (
         path: url.pathname + url.search,
         origin: url.origin,
         method: request.method,
-        body: fetchParams.controller.dispatcher.isActive ? request.body && request.body.source : body,
+        body: fetchParams.controller.dispatcher.isMockActive ? request.body && request.body.source : body,
         headers: [...request.headersList].flat(),
         maxRedirections: 0,
         bodyTimeout: 300_000,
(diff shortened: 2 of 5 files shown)


lib/mock/mock-agent.js
```diff
@@ -96,6 +96,12 @@ class MockAgent extends Dispatcher {
     this[kNetConnect] = false
   }
 
+  // This is required to bypass issues caused by using global symbols - see:
+  // https://github.com/nodejs/undici/issues/1447
+  get isActive () {
```
mcollina, line 101:
```suggestion
  get isMockActive () {
```

This is more descriptive.

mcollina (approved):
lgtm

This would really benefit from having a test, as it's a major use case of undici. Maybe open an issue?

mcollina:
cc @ronag

KhafraDev:
@mcollina added a test in a not so terrible way

mcollina (approved):
lgtm

PR #1466, opened 2022-05-25
feat: add support for request params in mock client

kibertoad:
refs #1448

### DIFF
--- a/types/mock-interceptor.d.ts
+++ b/types/mock-interceptor.d.ts
@@ -50,6 +50,8 @@ declare namespace MockInterceptor {
     body?: string | RegExp | ((body: string) => boolean);
     /** Headers to intercept on. */
     headers?: Record<string, string | RegExp | ((body: string) => boolean)> | ((headers: Record<string, string>) => boolean);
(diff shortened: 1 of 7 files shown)


mcollina (changes requested):
good work! Docs are missing!

types/mock-interceptor.d.ts
```diff
@@ -50,6 +50,8 @@ declare namespace MockInterceptor {
     body?: string | RegExp | ((body: string) => boolean);
     /** Headers to intercept on. */
     headers?: Record<string, string | RegExp | ((body: string) => boolean)> | ((headers: Record<string, string>) => boolean);
+    /** Query params to intercept on */
+    query?: Record<string, any>;
```
mcollina, line 54: please add a test for this type

kibertoad: Done!

mcollina (approved):
lgtm

kibertoad:
@mcollina Thank you! Would it be feasible to release a new version next week? New query param functionality is not very compatible with tests without this piece.

PR #1467, opened 2022-05-26
fix: ensure Blob and File `type` are included in FormData

petebacondarwin:
When appending a Blob-like or File-like object to FormData,
the item is wrapped in a `File` object. But this was causing
any options, such as `type` to be hidden in the wrapped type
and not exposed at the top level.

This fix copies the approach used in `formdata-polyfill` and
so also `node-fetch` where the wrapped object is also passed
as the `options` argument when creating a new `File` object.

Fixes https://github.com/cloudflare/wrangler2/issues/1085

### DIFF
--- a/lib/fetch/formdata.js
+++ b/lib/fetch/formdata.js
@@ -243,8 +243,8 @@ function makeEntry (name, value, filename) {
   // object, representing the same bytes, whose name attribute value is "blob".
   if (isBlobLike(value) && !isFileLike(value)) {
     value = value instanceof Blob
-      ? new File([value], 'blob')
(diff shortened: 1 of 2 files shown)


petebacondarwin:
I can't tell if the Node 16 Windows test is a flake or a real failure?

mcollina (approved):
LGTM

the Windows v16 is actually failing for a reason. However it's unrelated to this PR.

KhafraDev (approved):

ronag:
Confirmed that Chrome has the same behaviour.

PR #1469, opened 2022-05-28
feat: remove headers filtering

### LINKED ISSUE #1262: Undici strips out set-cookie headers, even when "credentials: 'include'" is set

### LINKED ISSUE #1463: `cookie` no longer able to be set in the Headers API as of v5.2.0
## Bug Description

Upgrading Node from 18.1 to 18.2 we are no longer able to set the `cookie` header in the `Headers` API.

Seems this is a result of https://github.com/nodejs/undici/pull/1337

## Reproducible By

```js
const headers = new Headers();
headers.append('cookie', 'bar=baz; foo=bar');

fetch('https://httpbin.org/cookies', {
  headers,
})
  .then(res => res.json())
  .then(console.log);
```

Running the above code on Node 18.1:

```js
{ cookies: { bar: 'baz', foo: 'bar' } }
```

On Node 18.2 (and undidi 5.2):

```js
{ cookies: {} }
```

Curiously if I don't use the `Headers` API and set the `cookie` header directly it works:

```js
fetch('https://httpbin.org/cookies', {
  headers: {
    cookie: 'bar=baz; foo=bar',
  },
})
  .then(res => res.json())
  .then(console.log);

// { cookies: { bar: 'baz', foo: 'bar' } }
```

## Expected Behavior

Would expect to be able to still set the `cookie` header because we can't use `document.cookie` like you can use in browser environments.

## Environment

* OSX 11.6
* Node 18.1 and Node 18.2 using `nvm`

KhafraDev:
Fixes #1463 
Fixes #1262 

@ronag's checklist [here](https://github.com/nodejs/undici/issues/1262#issuecomment-1132013882) should be completed:

1. "A WinterCG issue" - https://github.com/wintercg/fetch/issues/6
2. "It's clear in the code that all related logic is outside of spec and a reference motivating it." - Areas of the code where the spec mentioned "filtered" or "safelisted" headers has been commented.
3. "Clearly added to the README."
4. "Investigate what exactly deno, node-fetch and cloudflare does so that we don't introduce another variant with slight differences." - 

- Cloudflare has a [`Headers.prototype.getAll`](https://developers.cloudflare.com/workers/runtime-apis/headers#differences) method (deprecated & removed from the spec) that only works with `set-cookie` headers.
- Deno [doesn't do this filtering at all](https://deno.land/manual/runtime/web_platform_apis#spec-deviations) (confirmed on my end, a Deno maintainer also mentions this in the wintercg issue).
- node-fetch has a [`Headers.prototype.raw`](https://github.com/node-fetch/node-fetch#extract-set-cookie-header) method that returns the raw headers.

Clearly, following Deno is the best idea and closest to the spec itself. We aren't adding anything in, but simply removing constraints that don't apply to a server environment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,6 +293,7 @@ const headers = await fetch(url)
 * https://fetch.spec.whatwg.org/#cors-safelisted-response-header-name
 * https://fetch.spec.whatwg.org/#forbidden-header-name
 * https://fetch.spec.whatwg.org/#forbidden-response-header-name
+* https://github.com/wintercg/fetch/issues/6
 
 The [Fetch Standard](https://fetch.spec.whatwg.org) requires implementations to exclude certain headers from requests and responses. Undici does not filter out these headers.
 
--- a/README.md
+++ b/README.md
@@ -295,7 +295,7 @@ const headers = await fetch(url)
 * https://fetch.spec.whatwg.org/#forbidden-response-header-name
 * https://github.com/wintercg/fetch/issues/6
 
-The [Fetch Standard](https://fetch.spec.whatwg.org) requires implementations to exclude certain headers from requests and responses. Undici does not filter out these headers.
+The [Fetch Standard](https://fetch.spec.whatwg.org) requires implementations to exclude certain headers from requests and responses. In browser environments, some headers are forbidden so the user agent remains in full control over them. In Undici, these constraints are removed to give more control to the user.
 
 ### `undici.upgrade([url, options]): Promise`
 
--- a/README.md
+++ b/README.md
@@ -288,6 +288,14 @@ const headers = await fetch(url)
   .then(res => res.headers)
 ```
 
+##### Forbidden and Safelisted Header Names
+
+* https://fetch.spec.whatwg.org/#cors-safelisted-response-header-name
+* https://fetch.spec.whatwg.org/#forbidden-header-name
+* https://fetch.spec.whatwg.org/#forbidden-response-header-name
+
+The [Fetch Standard](https://fetch.spec.whatwg.org) requires implementations to exclude certain headers from requests and responses. Undici does not filter out these headers.
+
 ### `undici.upgrade([url, options]): Promise`
 
 Upgrade to a different protocol. See [MDN - HTTP - Protocol upgrade mechanism](https://developer.mozilla.org/en-US/docs/Web/HTTP/Protocol_upgrade_mechanism) for more details.
--- a/lib/fetch/request.js
+++ b/lib/fetch/request.js
@@ -383,10 +383,10 @@ class Request {
     // 30. Set this’s headers to a new Headers object with this’s relevant
     // Realm, whose header list is request’s header list and guard is
     // "request".
-    this[kHeaders] = new Headers(request.headersList)
+    this[kHeaders] = new Headers()
+    this[kHeaders][kHeadersList] = request.headersList
     this[kHeaders][kGuard] = 'request'
     this[kHeaders][kRealm] = this[kRealm]
-    this[kState].headersList = this[kHeaders][kHeadersList]
 
     // 31. If this’s request’s mode is "no-cors", then:
     if (mode === 'no-cors') {
@@ -414,8 +414,7 @@ class Request {
       }
 
       // 3. Empty this’s headers’s header list.
-      this[kHeaders] = new Headers()
-      this[kState].headersList = this[kHeaders][kHeadersList]
+      this[kHeaders][kHeadersList].clear()
 
       // 4. If headers is a Headers object, then for each header in its header
       // list, append header’s name/header’s value to this’s headers.
--- a/lib/fetch/request.js
+++ b/lib/fetch/request.js
@@ -383,10 +383,10 @@ class Request {
     // 30. Set this’s headers to a new Headers object with this’s relevant
     // Realm, whose header list is request’s header list and guard is
     // "request".
-    this[kHeaders] = new Headers()
+    this[kHeaders] = new Headers(request.headersList)
     this[kHeaders][kGuard] = 'request'
-    this[kHeaders][kHeadersList] = request.headersList
     this[kHeaders][kRealm] = this[kRealm]
+    this[kState].headersList = this[kHeaders][kHeadersList]
 
     // 31. If this’s request’s mode is "no-cors", then:
     if (mode === 'no-cors') {
@@ -406,7 +406,7 @@ class Request {
     if (Object.keys(init).length !== 0) {
       // 1. Let headers be a copy of this’s headers and its associated header
       // list.
-      let headers = new Headers(this.headers)
+      let headers = new Headers(this[kHeaders])
 
       // 2. If init["headers"] exists, then set headers to init["headers"].
       if (init.headers !== undefined) {
@@ -414,18 +414,18 @@ class Request {
       }
 
       // 3. Empty this’s headers’s header list.
-      this[kState].headersList = new HeadersList()
-      this[kHeaders][kHeadersList] = this[kState].headersList
+      this[kHeaders] = new Headers()
+      this[kState].headersList = this[kHeaders][kHeadersList]
 
       // 4. If headers is a Headers object, then for each header in its header
       // list, append header’s name/header’s value to this’s headers.
       if (headers.constructor.name === 'Headers') {
-        for (const [key, val] of headers[kHeadersList] || headers) {
-          this[kState].headersList.append(key, val)
+        for (const [key, val] of headers) {
+          this[kHeaders].append(key, val)
         }
       } else {
         // 5. Otherwise, fill this’s headers with headers.
-        fillHeaders(this[kState].headersList, headers)
+        fillHeaders(this[kHeaders], headers)
       }
     }
 
@@ -461,7 +461,7 @@ class Request {
       // not contain `Content-Type`, then append `Content-Type`/Content-Type to
       // this’s headers.
       if (contentType && !this[kHeaders].has('content-type')) {
-        this[kState].headersList.append('content-type', contentType)
+        this[kHeaders].append('content-type', contentType)
       }
     }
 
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -288,6 +288,14 @@ const headers = await fetch(url)
...
 ```
 
+##### Forbidden and Safelisted Header Names
+
+* https://fetch.spec.whatwg.org/#cors-safelisted-response-header-name
+* https://fetch.spec.whatwg.org/#forbidden-header-name
+* https://fetch.spec.whatwg.org/#forbidden-response-header-name
+
+The [Fetch Standard](https://fetch.spec.whatwg.org) requires implementations to exclude certain headers from requests and responses. Undici does not filter out these headers.
+
```
ronag, line 298: A link to the wintercg issue.

lib/fetch/request.js
```diff
@@ -421,7 +421,7 @@ class Request {
       // list, append header’s name/header’s value to this’s headers.
       if (headers.constructor.name === 'Headers') {
         for (const [key, val] of headers[kHeadersList] || headers) {
-          this[kHeaders].append(key, val)
+          this[kState].headersList.append(key, val)
```
ronag, line 424: Why this change?

KhafraDev: Without it, doing something along the lines of 
```js
import { fetch, Headers } from 'undici'

const headers = new Headers([['cookie', 'value']])

await fetch('http://example.com', { headers })
```

wouldn't work. My best guess is removing the filtered headers but I'm not entirely sure. I believe I also made a similar change for the same reason.

README.md
```diff
@@ -288,6 +288,15 @@ const headers = await fetch(url)
...
 ```
 
+##### Forbidden and Safelisted Header Names
+
+* https://fetch.spec.whatwg.org/#cors-safelisted-response-header-name
+* https://fetch.spec.whatwg.org/#forbidden-header-name
+* https://fetch.spec.whatwg.org/#forbidden-response-header-name
+* https://github.com/wintercg/fetch/issues/6
+
+The [Fetch Standard](https://fetch.spec.whatwg.org) requires implementations to exclude certain headers from requests and responses. Undici does not filter out these headers.
```
mcollina, line 298: Could you add a bit of an explanation on _why_ we do it?

KhafraDev: Not sure how to word it entirely...

"In browser environments, these headers are forbidden so the user agent remains in full control over them. In Undici, these constraints are removed to give more control to the user."? I took some of the wording directly from the spec without making it seem like this is insecure, but I'm not sure if it's enough.

KhafraDev: Also noticed that undici's filtering wasn't really spec compliant as it allowed `sec-` and `proxy-` headers anyways.

lib/fetch/request.js
ronag: I'm iffy about landing this before we understand what's going on here. Last time we were in This situation I found a rather serious bug.

KhafraDev: I think it's mostly confusing with having multiple Headers/HeadersList for each Request and confusion with distinguishing one from another.

https://github.com/nodejs/undici/blob/0e6427467e5c2d604dca32b7cb01c0b4979d25de/lib/fetch/request.js#L421
The spec says to use "this's headers" and undici uses `this[kHeaders]`.

https://github.com/nodejs/undici/blob/0e6427467e5c2d604dca32b7cb01c0b4979d25de/lib/fetch/request.js#L427
Then a few lines down the spec says to also fill "this's Headers" and `this[kState].headersList` is being used.

Without looking too far into it, removing `this[kHeaders]` would probably be beneficial, and use `this[kState].headersList` (maybe change it to a Headers instance?) instead. Or remove one of them, so there's less confusion on which should be used?

KhafraDev: With looking into it, we can set `this[kHeaders]` to a Headers class and set `request.headersList` to `this[kHeaders].headersList`?

KhafraDev: Using `.set` and `.append` on the `Headers` instance doesn't seem to update its HeadersList. Updating the list itself, however, does work.

ronag: I'm open for changes that are consistent and follow the wording of the spec.

KhafraDev: Found a solution and was able to remove the weird hack that was added last time this happened.

lib/fetch/request.js
```diff
@@ -406,26 +406,26 @@ class Request {
...
       // 2. If init["headers"] exists, then set headers to init["headers"].
       if (init.headers !== undefined) {
         headers = init.headers
       }
 
       // 3. Empty this’s headers’s header list.
-      this[kState].headersList = new HeadersList()
-      this[kHeaders][kHeadersList] = this[kState].headersList
+      this[kHeaders] = new Headers()
+      this[kState].headersList = this[kHeaders][kHeadersList]
```
ronag, line 418: This doesn’t seem to be quite the same as the spec?

KhafraDev: `Headers` is just a wrapper around `HeadersList` so emptying the headers list is as easy as creating a new instance of headers.

`this[kState]`'s headersList has to be the same as `this[kHeaders][kHeadersList]` so we need to re-assign the value after creating a new Headers instance.

ronag: I’m a little worried that changing the reference might have unintended side effects. Why not just empty the list as the spec says?

ronag: Spliec(0)

KhafraDev: The list is a class with a symbol key that's a map but even calling `.clear()` on it or setting it to a new Map causes the tests to break. This is the only solution that works. I don't think changing the reference matters because this only happens in the constructor before we can even access the properties.

ronag: Something is happening since the tests fail…

KhafraDev: Tests pass when setting it to a new Headers instance (as it currently is) but everything else causes them to fail.

mcollina (approved):
lgtm

ronag (commented):
I'm still concerned about the 2 places where this PR diverges from the spec. However, I'm going on vacation tomorrow and won't have time to dig into it myself. Merge it if you wish but I'd be happy if you could resolve this before merging.

lib/fetch/request.js
```diff
@@ -383,10 +383,10 @@ class Request {
     // 30. Set this’s headers to a new Headers object with this’s relevant
     // Realm, whose header list is request’s header list and guard is
     // "request".
-    this[kHeaders] = new Headers()
+    this[kHeaders] = new Headers(request.headersList)
```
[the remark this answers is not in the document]
KhafraDev: We set this's headers to a new Headers instance and then below we set this's headers headersList to this's request's headersList... Backwards, but it works and follows the spec.

lib/fetch/request.js
```diff
@@ -406,26 +406,26 @@ class Request {
...
-      this[kState].headersList = new HeadersList()
-      this[kHeaders][kHeadersList] = this[kState].headersList
+      this[kHeaders] = new Headers()
+      this[kState].headersList = this[kHeaders][kHeadersList]
 
       // 4. If headers is a Headers object, then for each header in its header
       // list, append header’s name/header’s value to this’s headers.
       if (headers.constructor.name === 'Headers') {
-        for (const [key, val] of headers[kHeadersList] || headers) {
+        for (const [key, val] of headers) {
```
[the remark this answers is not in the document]
KhafraDev: header's Symbol.iterator is a reference to a map's Symbol.iterator, seemed like a weird hack that was added last time this was an issue (since `headers[kHeadersList]` will always be set cuz it's a Headers instance).

lib/fetch/request.js
KhafraDev: Found this issue and I'll give you a hint about it, pretty funny...

look at the changes to Headers guard...

KhafraDev:
@ronag Enjoy your vacation 😄, you were completely right - something was wrong with how the `guard` checks were being done now that the forbidden headers were no longer being checked as well (`append` and `set` were returning early so the headers weren't getting set). Only reason my changes worked was because `new Headers(...)` sets the guard to `none` by default, which doesn't have special behavior attached to it.

Everything follows the spec entirely now, and issues in the future should be easier now that I understand entirely what's happening!

dsine-de:
Should this be available already in a current Node.js version? I don't see a `cookie` header being set on a Node.js fetch request when following a redirect with a `set-cookie` response header in Node.js v.18.8.0, like in the browser or e.g. in the Postman client.

KhafraDev:
Yes, this is bundled in the current node version. If you make an issue or have a minimal reproduction I'd be glad to look into it.

dsine-de:
Ok thank you, I don't know what's the decision if this should work with Node.js native fetch, but it seems Postman and the browser fetch have something like a cookie jar and if they get a `set-cookie` header in a request which redirects, they save that cookie and add it as a `cookie` header when following the redirect.

If this should work the same with Node.js fetch, I'll make a minimal repro example & issue.

PR #1472, opened 2022-05-31
fix: support `TypedArray` and `DataView` as Request body

LiviaMedeiros:
- `ArrayBuffer.isView(BufferInstance)` didn't allow `isBuffer()` branch to be reached
- Fixed `DataView` dropping whole buffer into request body
- Fixed the rest of `ArrayBufferView`s


mcollina (approved):
lgtm

PR #1476, opened 2022-06-01
test: skip IPv6 tests if not supported

LiviaMedeiros:
This would unbreak testing in environments without IPv6 support.


mcollina (approved):
lgtm

PR #1477, opened 2022-06-01
ci: remove git credentials after checkout

Fdawgs:
See option description at https://github.com/actions/checkout#usage.
The repo does not use git in any `run` commands after checking out the repo in any of the workflow jobs.
This just removes the git credentials for security sake; should any used actions become compromised then at least they won't be exposed.


mcollina (approved):
lgtm

PR #1478, opened 2022-06-01
ci: skip automerge job if user is not dependabot

Fdawgs:
This stops the job from running if the user is not Dependabot, saving a few seconds CI run time.


mcollina (approved):
lgtm

PR #1488, opened 2022-06-10
fix: readme example codestyle

mainarthur:
Fixed a bit code style of example in readme

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,13 +185,13 @@ Help us improve the test coverage by following instructions at [nodejs/undici/#9
 Basic usage example:
 
 ```js
-    import {fetch} from 'undici';
+import { fetch } from 'undici';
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -185,13 +185,13 @@ Help us improve the test coverage by following instructions at [nodejs/undici/#9
...
 ```js
-    import {fetch} from 'undici';
+import { fetch } from 'undici';
 
-    async function fetchJson() {
-        const res = await fetch('https://example.com')
-        const json = await res.json()
-        console.log(json);
-    }
+async function fetchJson() {
```
mcollina, line 190: I think we can drop this function and just use top-level-await.

mcollina (approved):
lgtm

KhafraDev (approved):

PR #1491, opened 2022-06-13
test: add rejectUnauthorized for agent1 cert

RafaelGSS:
It will fix the main CI. However, it's not the best approach.

I'll be working to adjust the certificates for `servername`. FYI: this is happening because the host order has been changed.


mcollina (approved):
lgtm

PR #1494, opened 2022-06-14
fix(types): import `URL`

SimenB:
For various reasons, we are stuck on a slightly older `@types/node` which [doesn't have the global `URL` declaration](https://www.runpkg.com/?@types/node@16.11.21/url.d.ts). Importing it explicitly fixes that, without any negative side-effects


mcollina (approved):
lgtm

PR #1495, opened 2022-06-14
feat: implement (basic) webidl & headers comments

### LINKED ISSUE #933: Add spec comments to Header class
It is useful to add comment references to the relevant parts of the whatwg fetch spec for the Headers class.

KhafraDev:
Fixes #933 

WebIdl is pretty cool and it's used rather heavily in the fetch spec. Implementing a subset of it (that we need) will make everything much easier once it's finished. I can explain how it'll help if you want, just don't want to write out an essay if you don't care lol

I only did the `Headers` class as a PoC and in the process discovered a *load* of bugs and decided just to add the spec comments since I was already adding a ton to it and fixing stuff anyways 😄.

p.s.: it's very clear that `HeadersList` was not supposed to be Map-like structure, as each key can be duplicated and the actual appending doesn't occur until you use `.get(...)`. Not a big deal, just something I noticed when working on it.

### DIFF
--- a/lib/fetch/webidl.js
+++ b/lib/fetch/webidl.js
@@ -1,0 +1,54 @@
+const { toUSVString } = require('util')
+
+const webidl = {}
+webidl.converters = {}
+
+// https://webidl.spec.whatwg.org/#es-DOMString
+webidl.converters.DOMString = function DOMString (V, opts = {}) {
+  // 1. If V is null and the conversion is to an IDL type
+  //    associated with the [LegacyNullToEmptyString]
+  //    extended attribute, then return the DOMString value
+  //    that represents the empty string.
+  if (V === null && opts.legacyNullToEmptyString) {
+    return ''
+  }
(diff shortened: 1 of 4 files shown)


ronag (approved):

lib/fetch/webidl.js
```diff
@@ -0,0 +1,54 @@
+const { toUSVString } = require('util')
```
mcollina, line 1:
```suggestion
'use strict'

const { toUSVString } = require('util')
```

mcollina (changes requested):
Could you add some tests for the new code?

mcollina (approved):
lgtm

mcollina (changes requested):
Could you add some tests for the new code?

KhafraDev:
added tests

mcollina (approved):
lgtm

Issue #1496, opened 2022-06-15
Version 5.5.1 breaks ProxyAgent with port in the URL (on windows)


fvictorio:
## Bug Description

We have a test that creates a proxy and connects to it by doing:

```
new ProxyAgent(process.env.HTTPS_PROXY);
```

The env var here is something like `http://127.0.0.1:12345`.

Before v5.5.1, this was working fine, but starting from that version we get this error **only on windows**:

```
      Uncaught RangeError [ERR_SOCKET_BAD_PORT]: Port should be >= 0 and < 65536. Received NaN.
       at validatePort (internal/validators.js:182:11)
       at lookupAndConnect (net.js:1002:5)
       at Socket.connect (net.js:978:5)
       at Object.connect (net.js:187:17)
       at D:\a\hardhat\hardhat\node_modules\proxy\proxy.js:431:16
       at authenticate (D:\a\hardhat\hardhat\node_modules\proxy\proxy.js:456:3)
       at Server.onconnect (D:\a\hardhat\hardhat\node_modules\proxy\proxy.js:415:2)
       at onParserExecuteCommon (_http_server.js:641:14)
       at onParserExecute (_http_server.js:575:3)
```

As extra context, we are using [this package](https://www.npmjs.com/package/proxy) to create the proxy.

## Reproducible By

Hopefully the description is enough to reproduce this. If it isn't, I can try to create a minimal reproduction.

## Expected Behavior

`ProxyAgent` should work on windows when the given URL has a port.

## Environment

Windows 10
Node v12.22.12

### Additional context

This is happening on GitHub actions and I could reproduce it in a virtual machine with windows 10.

RafaelGSS:
@fvictorio I'm trying to reproduce, but it's working for me. Can you create a reproducible code?

fvictorio:
Sure:

```js
const Proxy = require("proxy");
const { request, ProxyAgent } = require("undici");

function getProxy() {
  const proxy = new Proxy();
  return new Promise((resolve) => {
    proxy.listen(() => resolve(proxy));
  });
}

async function main() {
  const proxy = await getProxy();

  const agent = new ProxyAgent(`http://127.0.0.1:${proxy.address().port}`);

  const response = await request(
    "http://solc-bin.ethereum.org/wasm/list.json",
    {
      dispatcher: agent,
      headersTimeout: 1000,
      maxRedirections: 10,
      method: "GET",
    }
  );

  console.log(response);
}

main().catch(console.error);
```

[...]

PR #1497, opened 2022-06-15
fix(proxy-agent): include host port in the path

RafaelGSS:
fixes: #1496 

I've tried to include a test, but this happens when the requested URL doesn't contain the port. Usually, external requests, so adding that might lead our tests to flaky in some moment.

### DIFF
--- a/lib/proxy-agent.js
+++ b/lib/proxy-agent.js
@@ -14,6 +14,10 @@ const kRequestTls = Symbol('request tls settings')
 const kProxyTls = Symbol('proxy tls settings')
 const kConnectEndpoint = Symbol('connect endpoint function')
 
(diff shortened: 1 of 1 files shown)


ronag (approved):

lib/proxy-agent.js
```diff
@@ -42,11 +46,15 @@ class ProxyAgent extends DispatcherBase {
     this[kAgent] = new Agent({
       ...opts,
       connect: async (opts, callback) => {
+        let requestedHost = opts.host
+        if (!opts.port) {
+          requestedHost += `:${defaultProtocolPort(opts.protocol)}`
```
ronag, line 51: Does host contain port if url has port?

mcollina (approved):
lgtm

PR #1436, opened 2022-05-11
fix(fetch): remove brand checks from Symbol.toStringTag

KhafraDev:
Continuation of #1430

I also removed the FormData export resolving to globalthis.FormData by default as it made testing the fixes impossible.

PR #1430, opened 2022-05-10
fix(Headers): do not brand-check toString & Symbol.toStringTag

KhafraDev:
image

Fixes 2 bugs:
1. `Headers.toString` should be inherited from Object's prototype.
2. The `Symbol.toStringTag` getter should not be brand checked.

PR #1498, opened 2022-06-16
fix(fetch): missing body-mixin brand checks

KhafraDev:
This adds brand checks to body mixin methods & properties. These are webidl checks done throughout the spec. This disallows the following:
```
// accessing these methods without a proper "this" value
Response.prototype.bodyUsed
Response.prototype.body
Request.prototype.bodyUsed
Request.prototype.body

Request.prototype.text.call({ some: 'this value' })
Response.prototype.text.call({ some: 'this value' })
// same for all other body mixin methods
// "this" must be a Request/Response object now.
```

These checks are already done everywhere else.

Refs: #1436 #1430

### DIFF
--- a/lib/fetch/body.js
+++ b/lib/fetch/body.js
@@ -262,100 +262,132 @@ function cloneBody (body) {
   }
 }
 
-const methods = {
-  async blob () {
-    const chunks = []
+function bodyMixinMethods (instance) {
+  const methods = {
+    async blob () {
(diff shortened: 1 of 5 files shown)


mcollina (changes requested):
Could you clarify in the PR description what this does? It's a bit unclear:

KhafraDev:
Done, sorry 😄

mcollina (approved):
lgtm

PR #1500, opened 2022-06-18
feat(File): proper webidl checks

KhafraDev:
Makes `File` completely spec-compliant and adds in a bunch of needed webidl checks.

# File:
- Fixes 0 arguments being allowed in File constructor.
- Add webidl checks for File constructor parameters.
- Add missing steps from the spec.

# WebIdl:
- Add `sequence` converters
- Add dictionary matching & optional key assigning with default & required keys.
- Add TypedArray, (Shared)ArrayBuffer, DataView, and IDL long long converters (required by File).

p.s. most of the webidl apis were inspired by Deno because they're just done well.

### DIFF
--- a/lib/fetch/file.js
+++ b/lib/fetch/file.js
@@ -1,19 +1,28 @@
 'use strict'
 
 const { Blob } = require('buffer')
+const { types } = require('util')
 const { kState } = require('./symbols')
+const { isBlobLike } = require('./util')
+const { webidl } = require('./webidl')
+const { collectASequenceOfCodePoints } = require('./dataURL')
 
 class File extends Blob {
   constructor (fileBits, fileName, options = {}) {
-    // TODO: argument idl type check
-
     // The File constructor is invoked with two or three parameters, depending
     // on whether the optional dictionary parameter is used. When the File()
     // constructor is invoked, user agents must run the following steps:
+    if (arguments.length < 2) {
+      throw new TypeError('2 arguments required')
+    }
+
+    fileBits = webidl.converters['sequence<BlobPart>'](fileBits)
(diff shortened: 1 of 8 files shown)


KhafraDev:
I'm unsure what the `FileLike` class' purpose is so I haven't changed it. Should it be removed? The few instances of it throughout the code doesn't make it seem necessary, but I'm guessing there's a reason it was implemented that I don't know of.

ronag (approved):

mcollina (approved):
lgtm

jimmywarting:
I don't know who created file like class. But I think File like class was created to allow for 3th party Blob that isn't a instance of NodeJS blobs (like fetch-blob). has a feature to get Blob/Files from the file system without reading anything into memory.   Only when you call any of the read methods on the Blob will it actually read the file.
This is necessary to allow for uploading large files with eg FormData

I'm still very much missing the feature from nodejs to retrieve a Blob/File backed up by the file system. And that is one of the main issue why i simply can not extend NodeJS blobs and why fetch-blob is still necessary to this day. Without it blobs are pretty useless you might as well use arraybuffer

Blobs main purpose is to offload data to the FS

NodeJS Blob is pretty bad written in comparison to fetch-blob when it comes to streaming and creating text of blobs that are above 500mb. I have created issue about it but it haven't been resolved for a long time now.

I really want to depreciate fetch-blob but i can't until my issues on nodejs have been resolved

ronag:
I second everything @jimmywarting. I'm the one that added `FileLike`.

jimmywarting:
A File is more like a file handle/reference point to where the data is located with an offset and size property of where it should start reading from and where it should stop

PR #1506, opened 2022-06-20
feat(Response): webidl checks

KhafraDev:
- Adds idl checks to `Response`
- Adds `unsigned short` idl conversion (required by Response).
- Fixes 2 bugs in `webidl.dictionaryConverter` (tests added for these)

All tests pass.

### DIFF
--- a/test/node-fetch/response.js
+++ b/test/node-fetch/response.js
@@ -1,6 +1,7 @@
 /* eslint no-unused-expressions: "off" */
 
 const chai = require('chai')
+const stream = require('stream')
 const { Response } = require('../../lib/fetch/response.js')
 const TestServer = require('./utils/server.js')
 const { Blob } = require('buffer')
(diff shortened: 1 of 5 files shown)


test/node-fetch/response.js
```diff
@@ -137,13 +136,13 @@ describe('Response', () => {
...
-      expect(result).to.equal('a=1')
-    })
-  })
+  // it('should support stream as body', () => {
+  //   const body = stream.Readable.from('a=1')
+  //   const res = new Response(body)
+  //   return res.text().then(result => {
+  //     expect(result).to.equal('a=1')
+  //   })
+  // })
```
ronag, line 145: Shouldn’t it support async iterable as body?

KhafraDev: That would make sense - fetch spec doesn't explicitly forbid/allow async iterables so it wouldn't be included in a webidl check, but undici does. I'll add support for them back.

mcollina (approved):
lgtm

PR #1507, opened 2022-06-20
feat(Request): webidl checks

KhafraDev:
- Remove `abort-controller` npm package support in `fetch`. I could add it back, but unlike `fetch-blob`, doesn't necessarily offer anything over node's built-in AbortController.
- Add in `webidl.converters.any` (just a noop)
- Add in `webidl.converters.boolean` (required for RequestInit.keepalive)
- Add in `webidl.nullableConverter` which allows null as a value in converters.
- Add in `allowedValues` in `webidl.dictionaryConverter` which allows specifying a list of values an input must be (this is needed for webidl enums)
- `webidl.converters.XMLHttpRequestBodyInit` now defaults to using the DOMString converter. Fixed a bug in the node-fetch suite when passing an object as RequestInit's body.
- Add in webidl converters for `Request`.

### DIFF
--- a/lib/fetch/request.js
+++ b/lib/fetch/request.js
@@ -8,7 +8,6 @@ const util = require('../core/util')
 const {
   isValidHTTPToken,
   sameOrigin,
-  toUSVString,
   normalizeMethod
 } = require('./util')
 const {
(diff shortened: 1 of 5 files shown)


ronag (approved):

mcollina (approved):
lgtm

PR #1510, opened 2022-06-22
fix(fetch): provide reason upon rejecting when redirected

### LINKED ISSUE #1509: Unknown cause error when fetch fails due to a redirect
## Bug Description

When you make a fetch request with the `redirect` option set to "error", upon receiving a redirect, fetch will throw an error with an undefined cause and simply saying "fetch failed"

## Reproducible By

```js
fetch("https://i.imgur.com/BLUyAX0.png", { redirect: "error" })
```

## Expected Behavior

The error is thrown with a more explicit reason such as "Unexpected redirect"

## Logs & Screenshots

image

## Environment

Windows 11, Node.js v18.4.0

ImRodry:
Fixes #1509. This is my first PR here so I'm not sure if I did the tests correctly or if this is the right approach to this issue. If it isn't, please comment suggesting how I should do it and I'll gladly change it :)

### DIFF
--- a/test/fetch/request.js
+++ b/test/fetch/request.js
@@ -5,7 +5,8 @@
 const { test } = require('tap')
 const {
   Request,
-  Headers
+  Headers,
+  fetch
 } = require('../../')
 const { kState } = require('../../lib/fetch/symbols.js')
 
@@ -265,6 +266,13 @@ test('null integrity', t => {
 test('undefined signal', t => {
   const req = new Request('http://asd', { signal: undefined })
   t.equal(req.signal.aborted, false)
+  t.end()
+})
+
+test('error on redirect', async t => {
+  const errorCause = await fetch('http://nodejs.org/', { redirect: 'error' }).catch(e => e.cause)
+
+  t.equal(errorCause.message, 'unexpected redirect')
   t.end()
 })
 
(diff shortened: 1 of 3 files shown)


ImRodry:
Seems like CI was already failing before my commit so I assume there’s nothing I can do about that ?

test/fetch/request.js
```diff
@@ -268,6 +269,13 @@ test('undefined signal', t => {
   t.end()
 })
 
+test('error on redirect', async t => {
+  const errorCause = await fetch('http://nodejs.org/', { redirect: 'error' }).catch(e => e.cause)
```
KhafraDev, line 273: This shouldn't use an external url, should use a node http server instead (examples can be found in other tests).

ImRodry: The other tests all use invalid URLs and I need this one to be a valid one and to redirect somewhere else. I chose the nodejs website but it can be any website really, as long as we put HTTP instead of https. Any suggestions?

KhafraDev: The other tests don't depend on a valid url to work; the behavior of the nodejs site could change at any time to not redirect, or the CI or site could have connection issues.

```js
    res.statusCode = 302
    res.end()
})
// ...
```

There's examples elsewhere of how to setup node http servers elsewhere in the tests.

ImRodry: Thanks! I've followed the patterns on other tests and mine passes now with a custom server. Feel free to take a look again!

ronag (approved):

PR #1515, opened 2022-06-25
fix: don't check socket writing on header timeout

fengmk2:


lib/client.js
```diff
@@ -889,10 +889,8 @@ function onParserTimeout (parser) {
 
   /* istanbul ignore else */
   if (timeoutType === TIMEOUT_HEADERS) {
-    if (!socket[kWriting]) {
```
fengmk2, line 892: The server side does not consume and client socket is always writing state.

mcollina (approved):
lgtm

PR #1516, opened 2022-06-26
docs: Improve documentation for dispatcher response

### LINKED ISSUE #1480: Correct way to identify empty response body
## This would solve...

Even if there is no response body, undici handles it by returning a stream with 0 length. Currently the only way to tell if you should be trying to parse the body is `response.body._readableState.length`, which feels a lot like delving deep into internals.

Additional getter on response would allow to solve this problem without performance overhead:

```
response.isBodyPresent // getter that checks _readableState.length
```

## I have also considered...

This can also be addressed via documentation, if it is recommended to resolve this in a different way, e. g. by trusting statusCode === 204.

kibertoad:
fixes #1480

### DIFF
--- a/docs/api/Dispatcher.md
+++ b/docs/api/Dispatcher.md
@@ -461,14 +461,14 @@ Arguments:
 * **options** `RequestOptions`
 * **callback** `(error: Error | null, data: ResponseData) => void` (optional)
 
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

hebertcisco (approved):

kibertoad:
What is left before we can merge?

PR #1518, opened 2022-06-29
fix: don't leak connections on throwOnError

ronag:


mcollina (changes requested):
CI is failing, also tests are missing.

mcollina (approved):
lgtm

kibertoad:
@ronag Was `body.dump()` intentional there? 
Considering that in many cases retrieving body of error for programmatic processing is important, can I contribute alternate implementation which will immediately read and deserialize error response body if available?

PR #1522, opened 2022-07-02
docs: add example with `HEAD` method to garbage collection section

### LINKED ISSUE #1505: GC example could be improved
if you are only after the headers then it might be better to just use `HEAD` request.
```js
// Do
const headers = await fetch(url)
  .then(async res => {
    for await (const chunk of res.body) {
      // force consumption of body
    }
    return res.headers
  })

// Do not
const headers = await fetch(url)
  .then(res => res.headers)
```
alternative is to just cancel it with abort signal and abort the request
but i also understand if it's just for explaining how GC works. but many ppl just blindly copy code in readme's that isn't always the best

mrkldshv:
Adds example of getting headers using `HEAD` method to garbage collection section.

Closes #1505.

### DIFF
--- a/README.md
+++ b/README.md
@@ -283,6 +283,13 @@ const headers = await fetch(url)
   .then(res => res.headers)
 ```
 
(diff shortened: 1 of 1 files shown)


mrkldshv:
@mcollina I guess it's failing because of the flaky tests on Windows. Can we restart CI?

mcollina (approved):
lgtm

PR #1523, opened 2022-07-04
fix: improper handling of relative location header

### LINKED ISSUE #1521: the redirect handler does not correctly honor relative paths in the location header
## Bug Description

As outlined in [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.2), the `Location` header can feature full URLs, absolute paths, and relative paths. Undici treats relative paths like absolute paths.

## Reproducible By

I noticed this while trying to use the Wikipedia API, which returns relative location headers. Here's example code:

```js
const undici = require('undici');

(async () => {
    const res = await undici.request('https://en.wikipedia.org/api/rest_v1/page/summary/bean', { maxRedirections: 10 });

    console.log(res.context.history.map(item => item.href));

    const { headers } = await undici.request('https://en.wikipedia.org/api/rest_v1/page/summary/bean', { maxRedirections: 0 });

    console.log(headers['location']);

})();
```

## Expected Behavior

\> GET /a/b/c
\> Host: example.com
\>
< HTTP/1.1 301 Moved Permanently
< location: d

Should redirect to `example.com/a/b/d`, but undici will redirect to `example.com/d`.

(in the particular case above, it should return `https://en.wikipedia.org/api/rest_v1/page/summary/Bean` instead of `https://en.wikipedia.org/Bean`).

## Logs & Screenshots

For the code example:
```js
[
  'https://en.wikipedia.org/api/rest_v1/page/summary/bean',
  'https://en.wikipedia.org/Bean'
]
Bean
```

## Environment

node: 16.15.1
undici: ^5.6.0

### Additional context

I suspect the line where this happens is [this one](https://github.com/nodejs/undici/blob/main/lib/handler/redirect.js#L102). Here, only the `origin` is passed into the URL constructor, discarding any previous path and therefore breaking for relative paths.
I was unable to produce a passing build with a possible fix locally -- that's why this is an issue and not a PR.

PrincessRavy:
fixes #1521

### DIFF
--- a/lib/handler/redirect.js
+++ b/lib/handler/redirect.js
@@ -99,7 +99,7 @@ class RedirectHandler {
       return this.handler.onHeaders(statusCode, headers, resume, statusText)
     }
 
-    const { origin, pathname, search } = util.parseURL(new URL(this.location, this.opts.origin))
+    const { origin, pathname, search } = util.parseURL(new URL(this.location, this.opts.origin && new URL(this.opts.path, this.opts.origin)))
     const path = search ? `${pathname}${search}` : pathname
 
     // Remove headers referring to the original URL.
--- a/test/redirect-relative.js
+++ b/test/redirect-relative.js
@@ -1,0 +1,22 @@
+'use strict'
+
+const t = require('tap')
+const { request } = require('..')
+const {
+  startRedirectingWithRelativePath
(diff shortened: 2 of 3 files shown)


EthanDevelops (approved):
LGTM :D

PR #1525, opened 2022-07-04
feat(headers): expose `headerValueNormalize`

Kikobeats:
Hello,

We want to upgrade from undici [5.5.1 to 5.6.0](https://github.com/vercel/edge-runtime/pull/14), and we were using `normalizeAndValidateHeaderValue` that is now `headerValueNormalize`, so we need it to be exported too.

### DIFF
--- a/lib/fetch/headers.js
+++ b/lib/fetch/headers.js
@@ -497,5 +497,6 @@ webidl.converters.HeadersInit = function (V) {
 module.exports = {
   fill,
   Headers,
-  HeadersList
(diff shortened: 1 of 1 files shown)


ronag (commented):
I dont' think we should be exporting these kind of internal uitilites.

mcollina (changes requested):
I don't think we should be exporting them either.

jimmywarting:
I don't think we should be exporting them either.

PR #1526, opened 2022-07-04
fix: require toUSVString from fetch utils

Kikobeats:
Since `util.toUSVString` doesn't exist at node12, in this way the fallback implementation will be used.

### DIFF
--- a/lib/fetch/webidl.js
+++ b/lib/fetch/webidl.js
@@ -1,6 +1,7 @@
 'use strict'
 
-const { toUSVString, types } = require('util')
+const { toUSVString } = require('./util')
+const { types } = require('util')
 
 const webidl = {}
 webidl.converters = {}


mcollina (changes requested):
why this change?

Kikobeats:
mainly to make `webidl.converters.USVString` addressable at Node12
https://github.com/vercel/edge-runtime/runs/7177696222?check_suite_focus=true#step%3A7%3A110=

mcollina:
I'm a bit lost. None of fetch should run on v12.

mcollina:
I'm a bit lost. None of fetch should run on v12.

Rich-Harris:
Presently, undici's package.json specifies 12.18 as the minimum version:

https://github.com/nodejs/undici/blob/5ca25c23500d649a629865d9a7a28ce653d5e991/package.json#L99-L101

In reality, it requires >16.9. Merging this PR wouldn't change that, since undici also uses `Object.hasOwn`. For that reason, I'd advocate updating the `engines` field and removing this line in favour of using the built-in `util.toUSVString` everywhere: https://github.com/nodejs/undici/blob/5ca25c23500d649a629865d9a7a28ce653d5e991/lib/core/util.js#L385

mcollina:
> In reality, it requires >16.9. 

Can you articulate? We have our suite testing everything on 12, minus fetch - which is v16+ and clearly documented as such.

I'd be happy to work on whatever incompatibility there is as long as I get a reproduction.

mcollina:
I'd be happy to land this as long as we get a test that is failing on Node v12 - so we know we won't regress.

Rich-Harris:
Sure — using Node 16.7:

```js
const { Response } = require('undici');
new Response('ok');
/*
Uncaught TypeError: webidl.converters.USVString is not a function
    at Object.webidl.converters.XMLHttpRequestBodyInit (/[dir]/node_modules/undici/lib/fetch/response.js:516:30)
    at Object.webidl.converters.BodyInit (/[dir]/node_modules/undici/lib/fetch/response.js:554:28)
    at new Response (/[dir]/node_modules/undici/lib/fetch/response.js:138:32)
*/
```

Using Node 16.8:

```js
const { Response } = require('undici');
new Response('ok');
/*
Uncaught TypeError: Object.hasOwn is not a function
    at Object.ResponseInit (/[dir]/node_modules/undici/lib/fetch/webidl.js:327:33)
    at new Response (/[dir]/node_modules/undici/lib/fetch/response.js:141:30)
*/
```

KhafraDev:
> do we want to support v16.7 at all?

Support should be easy enough, I don't see a reason not to. I'll work on a PR to support v16.5+ again.

mcollina:
Just add 16.5.0 to CI so we don't regress again.

Kikobeats:
closing since https://github.com/nodejs/undici/pull/1534 addressed it

PR #1529, opened 2022-07-04
fix(fetch): do not assign default value to `RequestInit.method`

### LINKED ISSUE #1527: `fetch` no longer respects `Request` object as a parameter
## Bug Description

Until 5.6.0, you could use `Request` objects as an input for `fetch`. But now it doesn't respect and ignores the parameters already given with previously created `Request`.

## Reproducible By
It can be easily reproduced like below. You will get an error that says it is not possible to have a `body` for HTTP `GET` and `HEAD` requests.
```ts
  const body = JSON.stringify({ foo: "bar" });
  const request = new undici.Request("https://www.google.com", {
        method: "POST",
        body,
    });

  await undici.fetch(request);
```

Also see https://github.com/nodejs/undici/pull/1528

## Expected Behavior

Like other `fetch` implementations, it should be expected to use `Request` objects as an input.

Related part of the spec;
https://fetch.spec.whatwg.org/#requestinfo

## Environment

Ubuntu 18.04 LTS,
Node v16.13

KhafraDev:
Fixes #1527 

---

The fetch docs confused me; for the `method` getter steps, GET is the default value. It should have been unset if the value was not provided in the constructor, though.

See:
1. https://fetch.spec.whatwg.org/#ref-for-dom-requestinit-method
2. https://fetch.spec.whatwg.org/#concept-request-method

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -95,7 +95,7 @@ class Fetch extends EE {
 }
 
 // https://fetch.spec.whatwg.org/#fetch-method
-async function fetch (input, init = undefined) {
+async function fetch (input, init = {}) {
   if (arguments.length < 1) {
     throw new TypeError(
       `Failed to execute 'fetch' on 'Window': 1 argument required, but only ${arguments.length} present.`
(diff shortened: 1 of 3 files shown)


GauBen:
Hello @ronag, this fix is quite critical for undici to work correctly, could you please release it to npm?

ronag:
@GauBen done

GauBen:
Thanks!

PR #1534, opened 2022-07-08
fix(fetch): re-add support for node v16.8.0+

KhafraDev:
Fetch support incorrectly claimed it would work on node >= v16.5.0.

- Data urls would not work until v16.7.0 (requires `URL.createObjectURL`).
- Body could be infinitely consumed until v16.8.0 (requires [`stream.isDisturbed`](https://nodejs.org/api/stream.html#streamreadableisdisturbedstream)).
- DOMException has some issues where it would extend a TypeError (or at least its name property would be "TypeError")? No idea when this was fixed.

See #1526 (specifically [this](https://github.com/nodejs/undici/pull/1526#issuecomment-1178207405) comment) for more context.

Also adds node v16.8.0 to the test matrix.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -19,7 +19,7 @@ jobs:
           uses: codecov/codecov-action@v2
       include: |
         - runs-on: ubuntu-latest
-          node-version: 16.7
+          node-version: 16.8
   automerge:
     if: >
         github.event_name == 'pull_request' &&
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -17,6 +17,9 @@ jobs:
       post-test-steps: |
         - name: Coverage Report
           uses: codecov/codecov-action@v2
+      include: |
+        - runs-on: ubuntu-latest
+          node-version: 16.5
   automerge:
     if: >
         github.event_name == 'pull_request' &&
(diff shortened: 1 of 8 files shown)


KhafraDev:
@ronag - isDisturbed doesn't seem to work on node v16.5 (`stream.isDisturbed doesn't yet exist)
https://github.com/nodejs/undici/blob/c1a0490f20760ce700b28ddcddd493758d3f54a1/lib/core/util.js#L295

mcollina (approved):
lgtm

mcollina:
I don't think we can do without `isDisturbed`, have you tried lifting the code from core? If not we'd have to support a version that at least has that.

KhafraDev:
The code is mostly similar to core other than core checks if a symbol is true on the object and does a `stream.readableAborted` check (which didn't fix the issue). This is a really niche issue afaict, and only happens on versions that don't have `stream.isDisturbed`.

mcollina:
That's why we added it in core. You can't implement fetch() with only the public API from WHATWG streams. That Symbol expose some internal state.

KhafraDev:
There were a few existing issues with node v16.5 that make fetch largely incompatible (up to 16.7, where all tests pass):

1. `URL.createObjectURL` was added in 16.7. Without it, fetching any `date:` url would fail.
2. When a request was aborted, it would throw a `TypeError` instead of a DOMException (?).
3. The `isDisturbed` util didn't work properly on this version.

mcollina:
Why did you close this?

KhafraDev:
Didn't realize lol. Need to bump the min version for fetch to 16.8.

mcollina (approved):
lgtm

PR #1535, opened 2022-07-10
fix(Headers): lowercase name in `Headers.prototype.set`

KhafraDev:
Fixes a regression in a WPT from https://github.com/nodejs/node/pull/43752

This is only an issue in `Headers.prototype.set` when using a header name that contains uppercase characters.


ronag (approved):

mcollina (approved):
lgtm

PR #1536, opened 2022-07-10
fix: allow optional body for mock reply

JustinBeckwith:
Quick quality of life fix.  The `reply` method today on the mock interceptor requires both an HTTP status code and a body to send.  In a lot of cases, it's common to mock out an empty body.  Trying to make this as smooth as `nock` by defaulting the body to `''` :)


mcollina (approved):
lgtm

PR #1537, opened 2022-07-10
fix: faster direct read approach

### LINKED ISSUE #1533: Switch out `body.text()` blob reading method to use `TextDecoder` directly instead
_Originally posted by @jimmywarting in https://github.com/nodejs/undici/issues/1524#issuecomment-1173783768_

jimmywarting:
This will saves some memory and make decoding responses faster and avoids some unnecessary overhead that needs to clone the hole body using Blob that in turns requires everything to be immutable and requires a lots of data being copied.

using `body.text()` became like 2x faster, compared to the prev commit 👍 

closes #1533

### DIFF
--- a/lib/fetch/body.js
+++ b/lib/fetch/body.js
@@ -258,6 +258,29 @@ function cloneBody (body) {
   }
 }
 
+async function * consumeBody (body) {
+  if (body) {
+    if (isUint8Array(body)) {
+      yield body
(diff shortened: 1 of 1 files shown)


ronag (approved):

KhafraDev (approved):

mcollina (approved):
lgtm

Issue #1490, opened 2022-06-12
Undici throws a terminated error


Eddy84:
## Bug Description

I have a bug with graphl-codegen which is very likely caused by undici closing the connection and sending a terminated signal?

I have double checked my backend with various graphql clients and plain http requests doing the introspection query, the bug is definitely in undici.

The server receives the request, but it is terminated.

A stacktrace is included below:

Here is the related issue
https://github.com/dotansimha/graphql-code-generator/issues/7934

## Reproducible By

See the linked issue above

## Expected Behavior

Return a response

## Logs & Screenshots

See below

## Environment

- Ubuntu 20 LTS
- undici 5.1

### Additional context

Stacktrace:

```
` Exited with an error DetailedError: Failed to load schema
    at loadSchema (/home/eduard/Projects/ecommerce/frontend/node_modules/@graphql-codegen/cli/bin.js:504:15)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
    at async /home/eduard/Projects/ecommerce/frontend/node_modules/@graphql-codegen/cli/bin.js:1034:65 {
  details: '\n' +
    '        Failed to load schema from http://localhost/service:\n' +
    '\n' +
    '        terminated\n' +
    '        TypeError: terminated\n' +
    '    at Fetch.onAborted (/home/eduard/Projects/ecommerce/frontend/node_modules/undici/lib/fetch/index.js:1893:49)\n' +
    '    at Fetch.emit (node:events:390:28)\n' +
    '    at Fetch.terminate (/home/eduard/Projects/ecommerce/frontend/node_modules/undici/lib/fetch/index.js:77:10)\n' +
    '    at Object.onError (/home/eduard/Projects/ecommerce/frontend/node_modules/undici/lib/fetch/index.js:2027:34)\n' +
    '    at Request.onError (/home/eduard/Projects/ecommerce/frontend/node_modules/undici/lib/core/request.js:237:27)\n' +
    '    at errorRequest (/home/eduard/Projects/ecommerce/frontend/node_modules/undici/lib/client.js:1711:13)\n' +
    '    at Socket.onSocketClose (/home/eduard/Projects/ecommerce/frontend/node_modules/undici/lib/client.js:985:5)\n' +
    '    at Socket.emit (node:events:390:28)\n' +
    '    at TCP.<anonymous> (node:net:687:12)\n' +
    '    \n' +
    '        GraphQL Code Generator supports:\n' +
[...]

Issue #1412, opened 2022-05-04
`fetch` throws UND_ERR_SOCKET error


regseb:
## Bug Description

A request on this [page of the Daily Mail](https://www.dailymail.co.uk/sciencetech/article-8057229/Scientists-create-stunning-gifs-Mars-sand-dunes-understand-conditions-impact-them.html) throws an _UND_ERR_SOCKET_ error. There is no problem with `https`.

## Reproducible By

Execute this script:

```JavaScript
import https from "node:https";
import zlib from "node:zlib";

const testWithHttps = (url) => {
    return new Promise((resolve, reject) => {
        const req = https.request(new URL(url), (res) => {
            const gunzip = zlib.createGunzip();
            res.pipe(gunzip);

            const buffer = [];
            gunzip.on("data", (chunk) => buffer.push(chunk))
                  .on("error", (err) => reject(err))
                  .on("end", () => resolve(buffer.join("")));
        });
        req.on("error", (err) => reject(err));
        req.setHeader("Accept", "*/*");
        req.setHeader("Accept-Encoding", "br, gzip, deflate");
        req.setHeader("Accept-Language", "*");
        req.setHeader("Sec-Fetch-Mode", "cors");
        req.setHeader("User-Agent", "undici");
        req.end();
    });
};

const testWithFetch = async (url) => {
    const response = await fetch(url);
    return response.text();
};


console.log("PTS (https://ptsv2.com/t/undici)");
let url = "https://ptsv2.com/t/undici/post";
console.log("HTTPS: " + (await testWithHttps(url)).substring(0, 100));
console.log("FETCH: " + (await testWithFetch(url)).substring(0, 100));

console.log("\nDaily Mail");
url = "https://www.dailymail.co.uk/sciencetech/article-8057229" +
                                  "/Scientists-create-stunning-gifs-Mars-sand" +
                                "-dunes-understand-conditions-impact-them.html";
console.log("HTTPS: " + (await testWithHttps(url)).substring(0, 100));
console.log("FETCH: " + (await testWithFetch(url)).substring(0, 100));
```

## Expected Behavior

```
PTS (https://ptsv2.com/t/undici)
HTTPS: Thank you for this dump. I hope you have a lovely day!
(node:22125) ExperimentalWarning: The Fetch API is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
FETCH: Thank you for this dump. I hope you have a lovely day!

Daily Mail
[...]

PR #1540, opened 2022-07-12
fix: Improve spec compliance by improve handling of responses without content-length or transfer-encoding

### LINKED ISSUE #1414: Undici fails to read bodies from responses with implicit lengths
## Bug Description

Paraphrasing the HTTP spec ([RFC 7230 3.3.3](https://www.rfc-editor.org/rfc/rfc7230#section-3.3.3)), the message length for a normal response body can be defined in a few ways:

* An explicit `content-length` header with a fixed length (point 3 in the RFC list)
* Given a `transfer-encoding` header, a transfer coding chunk that explicitly ends the body (point 5 in the list)
* If otherwise unspecified, all remaining bytes on the connection until the connection is closed (point 7 in the list)

Undici handles the last case incorrectly, making it impossible to read the HTTP response body for this simple "respond and then close the connection" case.

This doesn't come up much for big fancy servers, which tend to use keep-alive etc and explicit framing wherever they can, but it is common on quick simple HTTP server implementations, which avoid state and complexity by just streaming responses and closing the connection when they're done. It's a legitimate way to send responses according to the HTTP spec, and it is supported correctly by browsers and Node's `http` module, but not by Undici.

## Reproducible By

Running in Node 18.0.0:

```javascript

    res.removeHeader('transfer-encoding');
    res.writeHead(200, {
        // Header isn't actually necessary, but tells node to close after response
        'connection': 'close'
    });
    res.end('a response body');
}).listen(8008);

fetch('http://localhost:8008').then(async (response) => { // Global fetch from Undici
    console.log('got response', response.status, response.headers);

    try {
        const responseText = await response.text()
        console.log('got body', responseText); // Should log the body OK here
    } catch (e) {
        console.log('error reading body', e); // Throws a 'terminated' error instead
    }
});
```

The server returns a simple response, just containing the default `date` header and a `connection: close` header, then sends the body, then closes the connection.

The `connection: close` header is for clarity - this should work equally well without that, just using `res.socket.end()` explicitly instead.

## Expected Behavior

The above should print the status, headers, and then the response body, which is everything after the headers until the connection is closed.

This does work using fetch in browsers. To test this, run the script above (which will print the fetch error, but then keep running) then load `localhost:8008` in your browser - the string loads successfully.

With that page loaded (for CORS) you can also run the equivalent command in the browser dev console, which also works and prints the response body correctly:

```javascript
fetch('http://localhost:8008').then(async (response) =>
    console.log(await response.text()) // Logs the body correctly, no errors
);
```

This also works using Node's built-in HTTP module:

```javascript
const http = require('http');
const streamConsumers = require('stream/consumers');

http.get('http://localhost:8008', async (res) => {
    const body = await streamConsumers.text(res);
    console.log('GET got body:', body); // Logs the body correctly
});
```

## Logs & Screenshots

Undici's fetch does not print the response body, instead it throws an error:

```javascript
error reading body TypeError: terminated
    at Fetch.onAborted (node:internal/deps/undici/undici:7881:53)
    at Fetch.emit (node:events:527:28)
    at Fetch.terminate (node:internal/deps/undici/undici:7135:14)
    at Object.onError (node:internal/deps/undici/undici:7968:36)
    at Request.onError (node:internal/deps/undici/undici:696:31)
    at errorRequest (node:internal/deps/undici/undici:2774:17)
    at Socket.onSocketClose (node:internal/deps/undici/undici:2236:9)
    at Socket.emit (node:events:527:28)
    at TCP.<anonymous> (node:net:715:12) {
  [cause]: SocketError: closed
      at Socket.onSocketClose (node:internal/deps/undici/undici:2224:35)
      at Socket.emit (node:events:527:28)
      at TCP.<anonymous> (node:net:715:12) {
    code: 'UND_ERR_SOCKET',
    socket: {
      localAddress: undefined,
      localPort: undefined,
      remoteAddress: undefined,
      remotePort: undefined,
      remoteFamily: 'IPvundefined',
      timeout: undefined,
      bytesWritten: 171,
      bytesRead: 90
    }
  }
}
```

## Environment

Ubuntu 
* Node v18.0.0 with built-in fetch
* Node v16.14.2 with Undici 5.1.1

evanderkoogh:
This fixes #1414 #1412 #1490

From [RFC 7230 3.3.3](https://www.rfc-editor.org/rfc/rfc7230#section-3.3.3)
```
       Otherwise, this is a response message without a declared message
       body length, so the message body length is determined by the
       number of octets received prior to the server closing the
       connection.
```

So now if we get a SocketClosed, we check if we have seen a `content-length` or `transfer-encoding` header and if not, we complete the message parsing.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -419,6 +419,7 @@ class Parser {
     this.headersSize = 0
     this.headersMaxSize = client[kMaxHeadersSize]
     this.shouldKeepAlive = false
+    this.hasTransferEncoding = false
     this.paused = false
     this.resume = this.resume.bind(this)
 
@@ -615,6 +616,8 @@ class Parser {
       this.keepAlive += buf.toString()
     } else if (key.length === 14 && key.toString().toLowerCase() === 'content-length') {
       this.contentLength += buf.toString()
+    } else if (key.length === 16 && key.toString().toLowerCase() === 'transfer-encoding') {
+      this.hasTransferEncoding = true
     }
 
     this.trackHeader(buf.length)
@@ -957,10 +960,16 @@ function onSocketEnd () {
 }
 
 function onSocketClose () {
-  const { [kClient]: client } = this
+  let { [kClient]: client, [kParser]: parser } = this
 
-  this[kParser].destroy()
-  this[kParser] = null
+  // without content-length or transfer-encoding header, the body is defined as everything
+  // the server sends before closing the connection. RFC 7230 3.3.3 point 7.
+  if (!parser.contentLength && !parser.hasTransferEncoding) {
+    parser.onMessageComplete()
+  }
+
+  parser.destroy()
+  parser = null
 
   const err = this[kError] || new SocketError('closed', util.getSocketInfo(this))
 
--- a/test/client.js
+++ b/test/client.js
@@ -1972,6 +1972,43 @@ test('async iterator yield object error', (t) => {
   })
 })
 
+test('Successfully get a Response when neither a Transfer-Encoding or Content-Length header is present', (t) => {
+  t.plan(2)
+  const server = createServer((req, res) => {
+    req.on('data', (data) => {
+    })
+    req.on('end', () => {
+      res.removeHeader('transfer-encoding')
+      res.writeHead(200, {
+        // Header isn't actually necessary, but tells node to close after response
+        connection: 'close',
+        foo: 'bar'
+      })
+      res.end('a response body')
+    })
+  })
+  t.teardown(server.close.bind(server))
+
+  server.listen(0, () => {
+    const client = new Client(`http://localhost:${server.address().port}`)
+    t.teardown(client.close.bind(client))
+
+    client.request({ path: '/', method: 'GET' }, (err, { body }) => {
+      t.error(err)
+      const bufs = []
+      body.on('error', () => {
+        t.fail('Closing the connection is valid')
+      })
+      body.on('data', (buf) => {
+        bufs.push(buf)
+      })
+      body.on('end', () => {
+        t.equal('a response body', Buffer.concat(bufs).toString('utf8'))
+      })
+    })
+  })
+})
+
 function buildParams (path) {
   const cleanPath = path.replace('/?', '').replace('/', '').split('&')
   const builtParams = cleanPath.reduce((acc, entry) => {


ronag (commented):
Needs a test. Also I need to think about this. Feels like there might be some edge cases here.Especially with pipelining.

evanderkoogh:
It does have a test: https://github.com/nodejs/undici/pull/1540/commits/ac2fbb96e612b6cbb999a956fbe03a66673d937f#diff-ecc2f3b14c967f02fd074ac2d0e0b876e683c84f2755d251bbaa74a813a82130

It is really weird that it says [line 620](https://app.codecov.io/gh/nodejs/undici/compare/1540/diff?src=pr&el=tree#D1L620) is not covered by test because [line 928](https://app.codecov.io/gh/nodejs/undici/compare/1540/diff?src=pr&el=tree#D1L968) can never be executed if `this.hasTransferEncoding` has not been set to true?

evanderkoogh:
Oh and the `Node CI / Test / test (18, ubuntu-latest, experimental) (pull_request) Failing after 2m` failing test is a timeout in some unrelated code.

Ahh.. I think I see where your concern is around the edge cases and pipelining.. I would need to check, but it could be indeed possible that when you are pipelining, it would error out the next request. I am going to investigate.

mcollina:
I think we should just error all the pipelined requests. Given they are GETs, they will be picked up whenever another socket is connected.

mcollina (changes requested):
Could you add a test with more than one `client.request()` executing in parallel with pipelining enabled and disabled? So we verify this works ok in all situations.

Guilledll:
Hi everyone, any news about merging this? It will be very usefull for me to have it, thanks for the great work.

mcollina:
This still needs one more test: https://github.com/nodejs/undici/pull/1540#pullrequestreview-1038547402

@evanderkoogh are you still interesting in completing this PR?

lib/client.js
```diff
@@ -615,6 +616,8 @@ class Parser {
       this.keepAlive += buf.toString()
     } else if (key.length === 14 && key.toString().toLowerCase() === 'content-length') {
       this.contentLength += buf.toString()
+    } else if (key.length === 16 && key.toString().toLowerCase() === 'transfer-encoding') {
```
kyranet, line 619: Judging by the line 617 (since I'm not very familiar with Undici's code), shouldn't this be `17`? `transfer-encoding` has 3 more characters than the previous header, and `.length` does indeed evaluate to `17`.

ronag: Sounds like a test is missing for this?

lib/client.js
```diff
@@ -957,10 +960,16 @@ function onSocketEnd () {
...
-  this[kParser].destroy()
-  this[kParser] = null
+  // without content-length or transfer-encoding header, the body is defined as everything
+  // the server sends before closing the connection. RFC 7230 3.3.3 point 7.
+  if (!parser.contentLength && !parser.hasTransferEncoding) {
+    parser.onMessageComplete()
+  }
+
+  parser.destroy()
+  parser = null
```
ronag, line 972: This doesn't properly null the parser. Please don't use a local var.

ronag (changes requested):

mcollina:
Would you mind sending a PR finishing this?

DamianGlowala:
Affects me (a Nuxt user) as well; receiving an error when proxying the request through a Nitro server:

image

garand:
😢 So is this not getting fix for now?

xgqfrms:
Any updates?

metcoder95:
PR has been closed for a while, but this seems to be still something to be considered to add. If interested, feel free to open a new PR with a proposal and we can outline something 🙂

evanderkoogh:
The problem is that when I made that change, I fixed the actual issue, but then realised I introduced a bunch of edge-cases where there were still problems and I didn't have any bandwidth to figure out how to solve for all of those.

DamianGlowala:
I would have assumed a non-compliance with [RFC 7230 3.3.3](https://www.rfc-editor.org/rfc/rfc7230#section-3.3.3) should gain a bit more attention from the Node.js team than simply 'feel free to open a new PR'...

Clearly this might be not a trivial problem to solve, which author of this PR already expressed in the above comment, hence any form of *support* from the team would be (very!) appreciated.

metcoder95:
The team has limited bandwidth for addressing all the issues opened at the time, hence we asked the community if they are willing to contribute to fixing the problem meanwhile we provide all the support and guidance through reviews.

Since the PR has been closed for a while, it might be better to open a new one where we can continue the efforts if someone volunteers 🙂

PR #1544, opened 2022-07-14
Fix up `exclude` in CI

dominykas:
Unfortunately, shared worklfows can't take objects as params - this needs to be a string...


mcollina (approved):
lgtm

PR #1545, opened 2022-07-14
fix: use FinalizationRegistry compat if global is missing

Kikobeats:
Similar to https://github.com/nodejs/undici/blob/main/lib/agent.js#L10=


mcollina:
How is this a problem? We have FinalizationRegistry in all versions of Node.js that are running that code. _In case it's a bundling issue_, please add a test that bundles Undici and loads it in a previous node runtime.

mcollina:
Could you add a unit test that would fail with this change? So it's clear we don't regress.

PR #1550, opened 2022-07-15
fix(body mixin): only allow Uint8Array chunks

KhafraDev:
I was writing a regression test for https://github.com/nodejs/node/issues/43838 and discovered that the body mixin methods were still not spec-compliant. According to the stream standards, every chunk must be a Uint8Array when reading a body.

1. [body-mixin .text()](https://fetch.spec.whatwg.org/#dom-body-text) (for example), calls [consume body](https://fetch.spec.whatwg.org/#concept-body-consume-body)
2. consume body calls [fully reading body as promise](https://fetch.spec.whatwg.org/#fully-reading-body-as-promise)
3. fully reading body as promise calls [ReadableStreamDefaultReader read all bytes](https://streams.spec.whatwg.org/#readablestreamdefaultreader-read-all-bytes)
4. and then [read loop](https://streams.spec.whatwg.org/#read-loop) is called
5. read loop chunk steps says "If chunk is not a [Uint8Array](https://tc39.es/ecma262/#sec-typedarray-objects) object, [reject](https://webidl.spec.whatwg.org/#reject) promise with a [TypeError](https://tc39.es/ecma262/#sec-native-error-types-used-in-this-standard-typeerror) and abort these steps."

Writing this out because it's very convoluted any might help someone understand this in the future.

p.s.: `.formData()` likely isn't correct still, but I left it untouched.

### DIFF
--- a/lib/fetch/body.js
+++ b/lib/fetch/body.js
@@ -291,6 +291,10 @@ function bodyMixinMethods (instance) {
       const chunks = []
 
       for await (const chunk of consumeBody(this[kState].body)) {
+        if (!isUint8Array(chunk)) {
+          throw new TypeError('Expected Uint8Array chunk')
+        }
+
         // Assemble one final large blob with Uint8Array's can exhaust memory.
         // That's why we create create multiple blob's and using references
         chunks.push(new Blob([chunk]))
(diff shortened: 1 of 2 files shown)


ronag (approved):

mcollina (approved):
lgtm

Issue #1484, opened 2022-06-08
Undici throws a ConnectTimeoutError when there is no connection


mdmitry01:
## Bug Description

If you try to send an HTTP request, but the DNS lookup for the requested domain name fails, Undici may throw a `ConnectTimeoutError`.

## Reproducible By

I wrote a simple script which reproduces the issue

```js
const { Pool } = require('undici')

const agent = new Pool('http://foobar.bar')

for (let i = 0; i < 500; i++) {
  agent.request({ method: 'GET', path: '/foobar' })
    .then(async ({ body }) => {
      for await (const data of body) {
        console.log('data', data.toString('utf8'))
      }
    })
    .catch(console.error)
}
```

I used undici@5.4.0

## Expected Behavior

I'd expect to get only errors which indicate that the DNS lookup has failed.

## Logs & Screenshots

[Video](https://user-images.githubusercontent.com/69001428/172602292-44b429bc-029e-4bc4-bbf8-b82edb98930c.mp4)

## Environment

Ubuntu 20.04.4 LTS, Node v16.13.0

mcollina:
I cannot reproduce the error at all on Linux or Mac.

@ronag could you take a look too?

jodevsa:
@mcollina  I can reproduce it with

`
const { Pool } = require('undici')

const agent = new Pool('http://foobar.bar')

for (let i = 0; i < 50000; i++) {
  agent.request({ method: 'GET', path: '/foobar' })
    .then(async ({ body }) => {
      for await (const data of body) {
        console.log('data', data.toString('utf8'))
      }
    })
    .catch(console.error)
}
`


then try to write the output on a file and search for TO. something like

`node a.js 2> a.txt`


I think the problem here is that the call stack blocks the event loop from executing stuff in the callback queue until the for loop is done executing and as soon as the callbacks are executed we reached TO

jodevsa:
Seems like the for loop execution is blocking the event loop from picking executing callbacks from the callback queue and for some reason the TO callback comes before the error callback.

using process.nextTick for the setTimeout fixes it 

`let timeoutId = null
    socket
      .setNoDelay(true)
      .once(protocol === 'https:' ? 'secureConnect' : 'connect', function () {
        clearTimeout(timeoutId)

        if (callback) {
          const cb = callback
          callback = null
          cb(null, this)
        }
      })
      .on('error', function (err) {
        clearTimeout(timeoutId)

        if (callback) {
          const cb = callback
          callback = null
[...]

PR #1551, opened 2022-07-15
fix: prioritise error events over timeouts

jodevsa:
fix: prioritise error events over timeouts

Error events in the implementation of `net.connect` are defered to the next tick: https://github.com/nodejs/node/blob/main/lib/net.js#L1144
This will result in the TO callback always being executed before the error events callback in the callback queue.

This fix makes sure that error event callbacks are executed before the TO socket destroy logic.

fixes https://github.com/nodejs/undici/issues/1484


test/connect-timeout.js
```diff
@@ -4,7 +4,31 @@ const { test } = require('tap')
...
 const net = require('net')
 
-// Never connect
+function blockingSleep (ms) {
+  let current = Date.now()
+  const blockUntil = current + ms
+  while (current < blockUntil) {
+    current = Date.now()
+  }
+}
```
mcollina, line 13: Use https://www.npmjs.com/package/atomic-sleep instead

mcollina (commented):
Good spot!

jodevsa:
wasted 1 hour trying to find out why the test isn't working and realised that starbucks DNS resolves any hostname.!!
Screenshot 2022-07-16 at 20 22 05

jodevsa:
@mcollina  please check again

lib/core/connect.js
```diff
@@ -76,7 +76,7 @@ function buildConnector ({ maxCachedSessions, socketPath, timeout, ...opts }) {
     }
 
     const timeoutId = timeout
-      ? setTimeout(onConnectTimeout, timeout, socket)
+      ? setTimeout(() => setImmediate(() => onConnectTimeout(socket)), timeout)
```
ronag, line 79: I'm a little unsure about this... but at least we need a comment here explaining why.

jodevsa: sure, how about something like 

"setImmediate is needed to make sure that error event callbacks are executed before the TO socket destroy logic."?

jodevsa:
tests are not working for windows

`Transformation error for D:\a\undici\undici\lib\core\connect.js ; return original code
EPERM: operation not permitted, rename 'D:\a\undici\undici\node_modules\.cache\nyc\fbfece793b45952b8a9635dc1bc6687ac6fb676105e3851ee22696fdcb8c0a14.js.349094085' -> 'D:\a\undici\undici\node_modules\.cache\nyc\fbfece793b45952b8a9635dc1bc6687ac6fb676105e3851ee22696fdcb8c0a14.js'
Transformation error for D:\a\undici\undici\lib\mock\mock-symbols.js ; return original code`

not sure what this is about

jodevsa:
this change also seems to fix the example from the original issue

`const { Pool } = require('./')

const agent = new Pool('http://foobar.bar')

for (let i = 0; i < 50000; i++) {
  agent.request({ method: 'GET', path: '/foobar' })
    .then(async ({ body }) => {
        console.log('data', data.toString('utf8'))
      }
    })
    .catch(console.error)
}
`

lib/core/connect.js
jodevsa: > I'm a little unsure about this...
I was also not very sure about this but thinking of it again: Deferring throwing a TO in the hope that we already got a response makes sense.

jodevsa:
my new test is failing on windows :(

jodevsa:
@mcollina  seems like on windows we'll need 2 setImmediates I still don't know why but I noticed we have different implementation for windows in net.js. is it fine if we add another setImmediate?

mcollina:
Detect if we are on windows and add another immediate in that case. I would also recommend moving that logic in a function because it's a bit hard to read.

jodevsa:
@mcollina  please check again

jodevsa:
It would be a good idea to cancel the setImmedates when the response/error is fired.. let me do that first and then I'll pinh you for revewing

mcollina (approved):
lgtm

PR #1554, opened 2022-07-18
Do not decode the body while we are following a redirect

mcollina:
Ref: https://github.com/nodejs/node/issues/43868

Signed-off-by: Matteo Collina <hello@matteocollina.com>

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -1942,6 +1942,7 @@ async function httpNetworkFetch (
           }
 
           let codings = []
+          let location = ''
 
           const headers = new Headers()
           for (let n = 0; n < headersList.length; n += 2) {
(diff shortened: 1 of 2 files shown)


ronag (approved):
I still don't think this fixes the missing error handler. I don't understand how it became unhandled.

mcollina:
> I still don't think this fixes the missing error handler. 

It doesn't. That's a node core bug from my point of view.

This "hotfix" removes the decoding of a body that's not going to be decoded anyway.

This originates from https://github.com/nodejs/undici/blob/26f60b7b6e612bb831133d7f85914963d1955011/lib/fetch/index.js#L1053-L1057.. it's the old https://github.com/whatwg/fetch/issues/1288.

mcollina:
I think it's https://github.com/nodejs/node/pull/41954, specifically https://github.com/nodejs/node/pull/41954/files#diff-95c23fb6674af85f6f23a374851af6cccf58fb4617d73897df5078ba537a862cR243. We are not consuming that decoded stream completely. Therefore we are crashing.

This fix looks correct now.

ronag:
I feel there is a more fundamental problem here but this should fix the most immediate issue.

PR #1563, opened 2022-07-19
fix: x-www-form-urlencoded parser keep the BOM

cola119:
According to [WHATWG spec](https://fetch.spec.whatwg.org/#body-mixin:~:text=Otherwise%2C%20if%20mimeType%E2%80%99s%20essence%20is%20%22application/x%2Dwww%2Dform%2Durlencoded%22%2C%20then%3A), if mimeType is "application/x-www-form-urlencoded", `.formData()` will parse the body with [application/x-www-form-urlencoded parser](https://url.spec.whatwg.org/#concept-urlencoded-parser). It uses [UTF-8 decode without BOM](https://encoding.spec.whatwg.org/#utf-8-decode-without-bom) which preserves the BOM.
In contrast, `.text()` will use [UTF-8 decode](https://fetch.spec.whatwg.org/#body-mixin:~:text=Return%20the%20result%20of%20running%20UTF%2D8%20decode%20on%20bytes.) , which removes the BOM.

Currently Node.js fails this test case from WPT.
https://github.com/web-platform-tests/wpt/blob/fd64a55ccd24da176a82c1d9cc4e777f1832ab30/url/urlencoded-parser.any.js#L3

### DIFF
--- a/lib/fetch/body.js
+++ b/lib/fetch/body.js
@@ -404,7 +404,18 @@ function bodyMixinMethods (instance) {
         // 1. Let entries be the result of parsing bytes.
         let entries
         try {
-          entries = new URLSearchParams(await this.text())
+          let text = ''
+          // application/x-www-form-urlencoded parser will keep the BOM.
+          // https://url.spec.whatwg.org/#concept-urlencoded-parser
+          const textDecoder = new TextDecoder('utf-8', { ignoreBOM: true })
+          for await (const chunk of consumeBody(this[kState].body)) {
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

mcollina:
CI seems failing

cola119:
`test/fixtures/client-ca-crt.pem` is expired?

KhafraDev:
tests are failing from #1565

cola119:
@mcollina @KhafraDev Thank you! All CI are green.

Issue #1574, opened 2022-07-25
fetch a long base64 url will crash and nothing happens


ahaoboy:
### Version

v18.6.0

### Platform

Darwin C02G87AYMD6R 21.5.0 Darwin Kernel Version 21.5.0: Tue Apr 26 21:08:22 PDT 2022; root:xnu-8020.121.3~4/RELEASE_X86_64 x86_64

### Subsystem

macos

### What steps will reproduce the bug?

node test.js
```
async function test(n) {
  const s = "a".repeat(n);
  const base64 = Buffer.from(s).toString("base64");
  const dataURIPrefix = "data:application/octet-stream;base64,";
  const url = dataURIPrefix + base64;
  const resp = await fetch(url);
  const buf = await resp.arrayBuffer();
  const str = Buffer.from(buf).toString("ascii");
  console.log(str === s);
}

test(1 << 10); // true
test(1 << 20); // nothing happens
test(1 << 15); // true

``` 

### How often does it reproduce? Is there a required condition?

always reproduce

### What is the expected behavior?

expect log true always or throw error

### What do you see instead?

when url is long enough, fetch is crash and nothing happen

mcollina:
Moved to the undici repo.

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

ahaoboy:
@mcollina 
https://github.com/nodejs/undici/pull/1575

PR #1575, opened 2022-07-25
fix: fetch a long base64 url will crash and nothing happens (close: #1574)

ahaoboy:
When data is too large, function argument number will overflow
```
return Uint8Array.of(...output)
```


ronag (approved):

mcollina (approved):
lgtm

KhafraDev:
Nice find, surprised that WPTs don't have a test for this!

PR #1584, opened 2022-07-28
fix: support `ArrayBufferView` and `ArrayBuffer` as body

LiviaMedeiros:
Implementation-wise, do we really need to make a copy?

### DIFF
--- a/lib/fetch/body.js
+++ b/lib/fetch/body.js
@@ -57,16 +57,16 @@ function extractBody (object, keepalive = false) {
 
     // Set Content-Type to `application/x-www-form-urlencoded;charset=UTF-8`.
     contentType = 'application/x-www-form-urlencoded;charset=UTF-8'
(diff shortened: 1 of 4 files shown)


ronag (changes requested):
Needs a test

KhafraDev:
> Implementation-wise, do we really need to make a copy?

Yes, see [extract](https://fetch.spec.whatwg.org/#concept-bodyinit-extract):
```
BufferSource]

    Set source to a copy of the bytes held by object.
```

LiviaMedeiros:
> Set source to a copy of the bytes held by object.

This is true spec-wise, but seems costly in implementation. I wonder if there is a workaround with better performance, considering that most other implementations do not expose `Request.body` directly.

Anyway, even if it's possible to optimize later, working code and spec compliance goes first.

KhafraDev (approved):

mcollina (approved):
lgtm

mcollina:
ping @ronag

PR #1588, opened 2022-08-03
docs: mock different endpoints in a single file

ritvik130:


mcollina (changes requested):
Could you avoid all the linting change in this PR? It looks like you have prettier or similar automatically formatting markdown.

ritvik130:
Do you mean disable prettier and then recommit?

mcollina:
Yes, thanks. Make sure to undo all the previous changes.

Issue #1590, opened 2022-08-03
`Argument is not a ByteString` when setting headers with non-ASCII characters


benmccann:
## Bug Description

It looks like https://github.com/nodejs/undici/issues/1317 wasn't actually fixed

## Reproducible By

I got this while setting the `link` header to the following value:
```
</_app/immutable/assets/__layout-a4ea7e9d.css>; rel="stylesheet"; nopush,</_app/immutable/start-cf8d1dd9.js>; rel="modulepreload"; nopush,</_app/immutable/chunks/index-bf47b8b5.js>; rel="modulepreload"; nopush,</_app/immutable/chunks/paths-98950022.js>; rel="modulepreload"; nopush,</_app/immutable/chunks/preload-helper-c28b9807.js>; rel="modulepreload"; nopush,</_app/immutable/chunks/singletons-1e28953d.js>; rel="modulepreload"; nopush,</_app/immutable/chunks/env-public-887c86aa.js>; rel="modulepreload"; nopush,</_app/immutable/pages/__layout.svelte-b3b95660.js>; rel="modulepreload"; nopush,</_app/immutable/chunks/navigation-c4179065.js>; rel="modulepreload"; nopush,</_app/immutable/pages/encoded/苗条.svelte-7e7988f6.js>; rel="modulepreload"; nopush,</_app/immutable/chunks/stores-2e0a9bb3.js>; rel="modulepreload"; nopush
```

## Expected Behavior

I think this should work. If it's invalid in some way, the error message should certainly be clarified. I passed a `string` which is valid, so it shouldn't be complaining about wanting a `ByteString`

## Logs & Screenshots

```
TypeError: Argument is not a ByteString
    at Object.webidl.converters.ByteString (node_modules/.pnpm/undici@5.8.1/node_modules/undici/lib/fetch/webidl.js:404:11)
    at Headers.set (node_modules/.pnpm/undici@5.8.1/node_modules/undici/lib/fetch/headers.js:362:31)
    at render_response (packages/kit/test/apps/basics/.svelte-kit/output/server/_app/immutable/chunks/index-fb484aa3.js:1090:13)
    at respond$1 (packages/kit/test/apps/basics/.svelte-kit/output/server/_app/immutable/chunks/index-fb484aa3.js:2013:25)
    at resolve (packages/kit/test/apps/basics/.svelte-kit/output/server/_app/immutable/chunks/index-fb484aa3.js:2261:114)
[...]

PR #1591, opened 2022-08-03
feat(webidl): better error message for ByteString converter

KhafraDev:
Closes https://github.com/nodejs/undici/issues/1590

Adds a better error message to the ByteString converter. This is close to Firefox's error message.


mcollina (approved):
lgtm

PR #1596, opened 2022-08-10
fix: add support for `integrity` option to Fetch

jelmervdl:
Fixes #1594

Issue #1594, opened 2022-08-10
Fetch with `integrity` option fails on "TypeError: response.arrayBuffer is not a function"


jelmervdl:
### Version

18.4.0

### Platform

Darwin MPB16.local 21.6.0 Darwin Kernel Version 21.6.0: Sat Jun 18 17:07:25 PDT 2022; root:xnu-8020.140.41~1/RELEASE_X86_64 x86_64

### Subsystem

fetch

### What steps will reproduce the bug?

Using `fetch()` with the `integrity` option seems to consistently fail, throwing from an internal library function.

Simplest example I could come up with:

```js
async function test() {
	const response = await fetch('https://storage.googleapis.com/bergamot-models-sandbox/0.3.3/enes/qualityModel.enes.bin', {
		integrity: 'sha256-zhQfjp5Qpe9NjjJDonSxc03FMvaWN5Sohp3ONay1Q8I='
	})

	console.log(response);
}

test();
```

### How often does it reproduce? Is there a required condition?

Seems to be consistent. Without the `integrity` option for the request, it succeeds. But with it, it throws. I have not tried it with different hosts.

### What is the expected behavior?

The above script retuning and printing the Response object instead of throwing.

### What do you see instead?

```
(node:22831) ExperimentalWarning: The Fetch API is an experimental feature. This feature could change at any time
(Use `node --trace-warnings ...` to show where the warning was created)
node:internal/deps/undici/undici:5555
          p.reject(Object.assign(new TypeError("fetch failed"), { cause: response.error }));
                                 ^

TypeError: fetch failed
    at Object.processResponse (node:internal/deps/undici/undici:5555:34)
    at node:internal/deps/undici/undici:5877:42
    at node:internal/process/task_queues:140:7
    at AsyncResource.runInAsyncScope (node:async_hooks:203:9)
    at AsyncResource.runMicrotask (node:internal/process/task_queues:137:8)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  cause: TypeError: response.arrayBuffer is not a function
      at mainFetch (node:internal/deps/undici/undici:5773:38)
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
}
```

mcollina:
Thanks for reoporting! I don't think integrity is implemented or tested at all:

https://github.com/nodejs/undici/blob/2944587da53ec394b6f1fde946d78763eecc6896/lib/fetch/util.js#L342-L344

[...]

PR #1597, opened 2022-08-11
fix(fetch): implement fully read body algorithm

KhafraDev:
This fixes the incorrect error that is thrown in #1594; it will now throw an integrity-mismatch check instead because it hasn't been implemented yet.

This is a weird case where neither uses of `fully read body` had code coverage and I still can't add a test to this area of the code because an error is still being thrown. I did include a test that worked on my end by removing the integrity checks, so once it is implemented the test can be enabled & should run without an issue. I would implement this myself, but there is already a WIP PR at #1596.

In either case, the `response` object referenced in these locations is **not** a WHATWG Response, but an internal representation of a response. It doesn't include the body mixin, which means calling `.arrayBuffer()` on it was throwing an error (edit: also calling `.arrayBuffer()` on a Readable web stream throws an error). This fixes that by implementing the algorithm in a more... efficient way than the spec while also still remaining compliant w/ the spec.

### DIFF
--- a/lib/fetch/util.js
+++ b/lib/fetch/util.js
@@ -4,6 +4,7 @@ const { redirectStatus } = require('./constants')
 const { performance } = require('perf_hooks')
 const { isBlobLike, toUSVString, ReadableStreamFrom } = require('../core/util')
 const assert = require('assert')
+const { isUint8Array } = require('util/types')
 
 let File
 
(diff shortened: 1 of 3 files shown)


ronag (approved):
Good job!

jelmervdl:
@KhafraDev Feel free to copy over parts of #1596 that are useful and make that pull request redundant.

lib/fetch/util.js
```diff
@@ -438,6 +439,53 @@ function makeIterator (iterator, name) {
...
+    //    taskDestination.
+    const fulfilledSteps = (bytes) => queueMicrotask(() => {
+      processBody(bytes)
+    })
+
+    fulfilledSteps(Buffer.concat(chunks, length))
+  } catch {
+    // 4. Let rejectedSteps be to queue a fetch task to run
+    //    processBodyError, with taskDestination.
+    queueMicrotask(() => processBodyError())
```
ronag, line 483: I think processBOdy should take a reason (err?).

KhafraDev: Spec doesn't tell us to but it makes sense. Will update it.

KhafraDev: done in a365c37971b8d94d9d55a135fa9e44ab426b1d1e

PR #1601, opened 2022-08-13
fix(File): respect typed array `byteOffset` and `byteLength`

mrbbot:
`File` ignored the `byteOffset` and `byteLength` properties of typed arrays, meaning the entire backing `ArrayBuffer` was included. Notably, this caused issues when passing `Buffer`s to the `File` constructor.

```js
import { File } from "undici";

const arrayBuffer = new TextEncoder().encode("abcdef").buffer;
const array = new Uint8Array(arrayBuffer, 1, 3);
const arrayFile = new File([array], "file.txt");
console.log(await arrayFile.text()); // expected "bcd", got "abcdef"

const buffer = Buffer.from("test");
const bufferFile = new File([buffer], "file.txt");
console.log(await bufferFile.text()); // expected "test", got garbage
```

### DIFF
--- a/lib/fetch/file.js
+++ b/lib/fetch/file.js
@@ -278,7 +278,9 @@ function processBlobParts (parts, options) {
       if (!element.buffer) { // ArrayBuffer
         bytes.push(new Uint8Array(element))
       } else {
-        bytes.push(element.buffer)
+        bytes.push(
+          new Uint8Array(element.buffer, element.byteOffset, element.byteLength)
+        )
       }
     } else if (isBlobLike(element)) {
       // 3. If element is a Blob, append the bytes it represents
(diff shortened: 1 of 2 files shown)


ronag (approved):

lib/fetch/file.js
```diff
@@ -278,7 +278,9 @@ function processBlobParts (parts, options) {
       if (!element.buffer) { // ArrayBuffer
         bytes.push(new Uint8Array(element))
       } else {
-        bytes.push(element.buffer)
+        bytes.push(
+          new Uint8Array(element.buffer, element.byteOffset, element.byteLength)
+        )
       }
```
ronag, line 284: Actually this is still wrong. We need to make a copy, not just a new reference. Could you fix that?

ronag: @KhafraDev is this something you could look into? The copy part.

KhafraDev: Will do. On vacation for the next 6 days but I'll see if I can get to it

PR #1604, opened 2022-08-17
fix: don't timeout while waiting for client to send request

ronag:
When e.g. uploading a large file from the client one would always
get a headers timeout even though everything was working and the file
was being in progress of sending.


ronag:
Note, fake-timers doesn't seem to work so well with `refresh`.

mcollina (approved):
lgtm

Issue #974, opened 2021-08-17
Implement multipart parsing


ronag:
Our `extractBody` method is missing support for parsing `multipart/form-data` responses.

ronag:
I'm not sure who/why someone would send multipart/form-data from a server so would appreciate some requests on this. Doesn't seem to be a very useful feature.

jimmywarting:
I think it seems useful, i mean https://www.npmjs.com/package/form-data has 4.5M download / week

MrWaip:
Hello. Telegram bot api accept only FormData for files
https://core.telegram.org/bots/api#inputfile

blittle:
> I'm not sure who/why someone would send multipart/form-data from a server so would appreciate some requests on this. Doesn't seem to be a very useful feature.

Correct me if I'm wrong, but isn't the bigger use case the server parsing multipart form data from the client? The example from the node-fetch PR iml is:

```js
import { Response } from 'node-fetch'

http.createServer(async function (req, res) {
  const formData = await new Response(req, { headers: req.headers }).formData()
  const file = formData.get('avatar')
  
  var arrayBuffer = await file.arrayBuffer()
  var text = await file.text()
  var whatwgReadableStream = file.stream()
});
```

It would be great if undici had the same support for multipart form data.

jimmywarting:
the example should actually be something like this:
```js
http.createServer(async function (req, res) {
  const url = `${req.protocol}://${req.get('host')}${req.originalUrl}`
  const request new Request(url, { body: req, method: req.method, headers: req.headers })
  const formdata = request.formData()
  const file = formData.get('avatar')
  
  var arrayBuffer = await file.arrayBuffer()
  var text = await file.text()
  var whatwgReadableStream = file.stream()
});
```
But it's easier/quicker to construct a Response that don't need any url or extra options
the idea of having a formdata parsing and shipping the `body.formData()` comes from service workers
a service worker is meant to intercept the request and be able to parse the payload when needed

> I'm not sure who/why someone would send multipart/form-data from a server so would appreciate some requests on this. Doesn't seem to be a very useful feature.

It's not about responding with a multipart/form-data response like in: `fetch(url).then(res => {res.formData()})`
cuz again that would be weird if a server responded with a multipart/form-data. It's rather more about parsing incoming request to the http server which alot of folk use nodejs for. 

[...]

PR #1606, opened 2022-08-17
feat: add support for multipart/form-data

cameron-robey:
- Add support for multipart/form-data

Closes https://github.com/nodejs/undici/issues/974

### DIFF
--- a/lib/fetch/body.js
+++ b/lib/fetch/body.js
@@ -402,7 +402,7 @@ function bodyMixinMethods (instance) {
         for (const [key, value] of this.headers) headers[key.toLowerCase()] = value
 
         const responseFormData = new FormData()
-        
+
         try {
           const busboy = Busboy({ headers })
           busboy.on('field', (name, value) => {
--- a/lib/fetch/body.js
+++ b/lib/fetch/body.js
@@ -402,29 +402,34 @@ function bodyMixinMethods (instance) {
         for (const [key, value] of this.headers) headers[key.toLowerCase()] = value
 
         const responseFormData = new FormData()
-
-        const busboy = Busboy({ headers })
-        busboy.on('field', (name, value) => {
-          responseFormData.append(name, value)
-        })
-        busboy.on('file', (name, value, info) => {
-          const { filename, encoding, mimeType } = info
-          const base64 = encoding.toLowerCase() === 'base64'
-          const chunks = []
-          value.on('data', (chunk) => {
-            if (base64) chunk = Buffer.from(chunk.toString(), 'base64')
-            chunks.push(chunk)
+        
+        try {
+          const busboy = Busboy({ headers })
+          busboy.on('field', (name, value) => {
+            responseFormData.append(name, value)
           })
-          value.on('end', () => {
-            const file = new File(chunks, filename, { type: mimeType })
-            responseFormData.append(name, file)
+          busboy.on('file', (name, value, info) => {
+            const { filename, encoding, mimeType } = info
+            const base64 = encoding.toLowerCase() === 'base64'
+            const chunks = []
+            value.on('data', (chunk) => {
+              if (base64) chunk = Buffer.from(chunk.toString(), 'base64')
+              chunks.push(chunk)
+            })
+            value.on('end', () => {
+              const file = new File(chunks, filename, { type: mimeType })
+              responseFormData.append(name, file)
+            })
           })
-        })
-        const busboyResolve = new Promise((resolve) => busboy.on('finish', resolve))
+          const busboyResolve = new Promise((resolve) => busboy.on('finish', resolve))
 
-        if (this.body !== null) for await (const chunk of consumeBody(this[kState].body)) busboy.write(chunk)
-        busboy.end()
-        await busboyResolve
+          if (this.body !== null) for await (const chunk of consumeBody(this[kState].body)) busboy.write(chunk)
+          busboy.end()
+          await busboyResolve
+        } catch (err) {
+          // busboy may fail to parse a malformed body, so throw a type error
+          throw Object.assign(new TypeError(), { cause: err })
+        }
 
         return responseFormData
       } else if (/application\/x-www-form-urlencoded/.test(contentType)) {
--- a/lib/fetch/body.js
+++ b/lib/fetch/body.js
@@ -403,33 +403,40 @@ function bodyMixinMethods (instance) {
 
         const responseFormData = new FormData()
 
-        try {
-          const busboy = Busboy({ headers })
-          busboy.on('field', (name, value) => {
-            responseFormData.append(name, value)
-          })
-          busboy.on('file', (name, value, info) => {
-            const { filename, encoding, mimeType } = info
-            const base64 = encoding.toLowerCase() === 'base64'
-            const chunks = []
-            value.on('data', (chunk) => {
-              if (base64) chunk = Buffer.from(chunk.toString(), 'base64')
-              chunks.push(chunk)
-            })
-            value.on('end', () => {
-              const file = new File(chunks, filename, { type: mimeType })
-              responseFormData.append(name, file)
-            })
-          })
-          const busboyResolve = new Promise((resolve) => busboy.on('finish', resolve))
+        let busboy
 
-          if (this.body !== null) for await (const chunk of consumeBody(this[kState].body)) busboy.write(chunk)
-          busboy.end()
-          await busboyResolve
-        } catch (err) {
-          // busboy may fail to parse a malformed body, so throw a type error
+        try {
+          busboy = Busboy({ headers })
+        } catch(err) {
+          // Error due to headers:
           throw Object.assign(new TypeError(), { cause: err })
         }
+
+        busboy.on('field', (name, value) => {
+          responseFormData.append(name, value)
+        })
+        busboy.on('file', (name, value, info) => {
+          const { filename, encoding, mimeType } = info
+          const base64 = encoding.toLowerCase() === 'base64'
+          const chunks = []
+          value.on('data', (chunk) => {
+            if (base64) chunk = Buffer.from(chunk.toString(), 'base64')
+            chunks.push(chunk)
+          })
+          value.on('end', () => {
+            const file = new File(chunks, filename, { type: mimeType })
+            responseFormData.append(name, file)
+          })
+        })
+        busboy.on('error', (err) => {
+          throw Object.assign(new TypeError(), { cause: err })
+        })
+        
+        const busboyResolve = new Promise((resolve) => busboy.on('finish', resolve))
+
+        if (this.body !== null) for await (const chunk of consumeBody(this[kState].body)) busboy.write(chunk)
+        busboy.end()
+        await busboyResolve
 
         return responseFormData
       } else if (/application\/x-www-form-urlencoded/.test(contentType)) {
--- a/lib/fetch/body.js
+++ b/lib/fetch/body.js
@@ -407,7 +407,7 @@ function bodyMixinMethods (instance) {
 
         try {
           busboy = Busboy({ headers })
-        } catch(err) {
+        } catch (err) {
           // Error due to headers:
           throw Object.assign(new TypeError(), { cause: err })
         }
@@ -428,11 +428,11 @@ function bodyMixinMethods (instance) {
             responseFormData.append(name, file)
           })
         })
-        busboy.on('error', (err) => {
-          throw Object.assign(new TypeError(), { cause: err })
+
+        const busboyResolve = new Promise((resolve, reject) => {
+          busboy.on('finish', resolve)
+          busboy.on('error', (err) => reject(err))
         })
-        
-        const busboyResolve = new Promise((resolve) => busboy.on('finish', resolve))
 
         if (this.body !== null) for await (const chunk of consumeBody(this[kState].body)) busboy.write(chunk)
         busboy.end()
--- a/lib/fetch/body.js
+++ b/lib/fetch/body.js
@@ -1,5 +1,6 @@
 'use strict'
 
+const Busboy = require('busboy')
 const util = require('../core/util')
 const { ReadableStreamFrom, toUSVString, isBlobLike } = require('./util')
 const { FormData } = require('./formdata')
@@ -8,9 +9,9 @@ const { webidl } = require('./webidl')
 const { Blob } = require('buffer')
 const { kBodyUsed } = require('../core/symbols')
 const assert = require('assert')
-const { NotSupportedError } = require('../core/errors')
 const { isErrored } = require('../core/util')
 const { isUint8Array, isArrayBuffer } = require('util/types')
+const { File } = require('./file')
 
 let ReadableStream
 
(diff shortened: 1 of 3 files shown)


JacobMGEvans:
Exciting!

ronag (commented):
What about errors?

mcollina (changes requested):

mcollina (changes requested):
busbuy will __emit__ an `'error'` event: https://github.com/mscdex/busboy/blob/9aadb7afbcb8c70c81c93b1018313c1b1835afb0/lib/types/multipart.js#L398.

This also needs tests for the error path.

lib/fetch/body.js
```diff
@@ -397,7 +398,40 @@ function bodyMixinMethods (instance) {
...
+          busboy.on('field', (name, value) => {
+            responseFormData.append(name, value)
+          })
+          busboy.on('file', (name, value, info) => {
+            const { filename, encoding, mimeType } = info
+            const base64 = encoding.toLowerCase() === 'base64'
+            const chunks = []
+            value.on('data', (chunk) => {
+              if (base64) chunk = Buffer.from(chunk.toString(), 'base64')
+              chunks.push(chunk)
```
ronag, line 417: This will buffer a lot of data in memory. It needs some form of limit or backpressure.

ronag (changes requested):

lib/fetch/body.js
mcollina: I don't think we can easily implement or put a limit. That's the fundamental weakness of parsing multipart with `fetch()`. We should document this.

The actual solution is to have disk-backed `Blob` objects, but that would need to be implemented in Node.js core. There is only so much we can do ourselves here.

jimmywarting: Think this is good as a MVP for now. At least it will be spec compatible and also making it at least possible to use `.fromData()` for those who might need it. It will at least solve the problem for ppl who might just handle some small tiny files (like profile picture uploads). or just only using simple `<form>`s without any files and still uses `new FormData()` to upload just text inputs as multipart payloads.

I agree with @mcollina that disc-backed `Blob`s are necessary. 
Think we can try to improve this later once disc-backed `Blob`s are in place. 
But right now I just wish `res.formData()` to at least work, even if it meant that it was handled badly. 

---

I think and i believe whenever you fetch something with either `xhr.responseType = 'blob'` or do `response.blob()` in browser that they just do pipe the data to the disc rather than reading it all to the memory, but i'm not sure... Browser can do all kind of magic to blobs. And they might all works differently in all browsers. They can try to offload some of "in memory blobs" to the disc if something is taking up too much memory. 

i have tried to make since of this old [presentation](https://docs.google.com/presentation/d/1MOm-8kacXAon1L2tF6VthesNjXgx0fp5AP17L7XDPSM/edit#slide=id.g75d5729ce_0_145) when building `fetch-blob`

---

Is there any reason why we can't just bring in the `Blob` and `File` class from v8 itself instead of building our own kind of polyfilled File/Blob class written in javascript?

think Bun.js was able to do this with JavaScriptCore (webkit). 
So when you pass a file/blob to `Bun.write(blob)`, Bun automatically uses a faster system call:
```js
const blob = Bun.file("input.txt");
await Bun.write("output.txt", blob);
```
On Linux, this uses the [copy_file_range](https://man7.org/linux/man-pages/man2/copy_file_range.2.html) syscall and on macOS, this becomes clonefile (or [fcopyfile](https://developer.apple.com/library/archive/documentation/System/Conceptual/ManPages_iPhoneOS/man3/copyfile.3.html)).

mcollina: > Is there any reason why we can't just bring in the Blob and File class from v8 itself instead of building our own kind of polyfilled File/Blob class written in javascript?

V8 does not have them, they are from Chromium. They should be implemented in Node.js.

JacobMGEvans: Should this thread be a discussion?

mcollina (approved):
lgtm

ronag (changes requested):
CI failure

lib/fetch/body.js
```diff
@@ -397,7 +398,47 @@ function bodyMixinMethods (instance) {
...
+            if (base64) chunk = Buffer.from(chunk.toString(), 'base64')
+            chunks.push(chunk)
+          })
+          value.on('end', () => {
+            const file = new File(chunks, filename, { type: mimeType })
+            responseFormData.append(name, file)
+          })
+        })
+        busboy.on('error', (err) => {
+          throw Object.assign(new TypeError(), { cause: err })
```
ronag, line 432: You can't just throw here... this will crash the process. Is there tests for this?

ronag (changes requested):

lib/fetch/body.js
JacobMGEvans: > You can't just throw here... this will crash the process.

Do you have a suggestion?

mcollina: You'll need to reject this promise https://github.com/nodejs/undici/pull/1606/files#diff-bd6db9a8bceb4e9fce3f93ba5284c81f46a44a1187909e8609006e1efeab2570R435

mcollina (approved):
lgtm

lib/fetch/body.js
```diff
@@ -397,7 +398,47 @@ function bodyMixinMethods (instance) {
...
+          })
+          value.on('end', () => {
+            const file = new File(chunks, filename, { type: mimeType })
+            responseFormData.append(name, file)
+          })
+        })
+
+        const busboyResolve = new Promise((resolve, reject) => {
+          busboy.on('finish', resolve)
+          busboy.on('error', (err) => reject(err))
```
mcollina, line 434: Would you mind adding a test for the error path too?

lib/fetch/body.js
```diff
@@ -1,5 +1,6 @@
 'use strict'
 
+const Busboy = require('busboy')
```
kibertoad, line 3: shouldn't busboy be added to package.json as a runtime dependency? it is likely to break in systems without it now.

kibertoad (changes requested):

ronag:
@mcollina I think @KhafraDev is working on a multipart parser without any deps. Should we maybe wait and see how that go before landing this?

kibertoad:
@ronag Do you expect new custom solution to work faster/smoother than a solution that was used across entire Node ecosystem for years?

mcollina:
I would prefer we use busboy at this stage.

lib/fetch/body.js
```diff
@@ -397,7 +398,47 @@ function bodyMixinMethods (instance) {
...
+
+        busboy.on('field', (name, value) => {
+          responseFormData.append(name, value)
+        })
+        busboy.on('file', (name, value, info) => {
+          const { filename, encoding, mimeType } = info
+          const base64 = encoding.toLowerCase() === 'base64'
+          const chunks = []
+          value.on('data', (chunk) => {
+            if (base64) chunk = Buffer.from(chunk.toString(), 'base64')
```
mcollina, line 423: Please add a test to cover this line

KhafraDev:
> I would prefer we use busboy at this stage.

Yes, this was my sentiment from my comment in #974.

lib/fetch/body.js
mcollina: @cameron-robey ping ^

lib/fetch/body.js
KhafraDev: Allowing this to be an optional feature would be nice, although npm has a few issues with optional dependencies. :(

lib/fetch/body.js
repsac-by: It is not possible to decode every chunk from base64, since the chunk may be incomplete from the base64 point of view

ronag (approved):

lib/fetch/body.js
mcollina: @cameron-robey can you add busboy to dependencies?

lib/fetch/body.js
mcollina: The error path still need a unit test.

mcollina (changes requested):
Good work, there are a couple of comments to fix and we'd be good for landing.

lib/fetch/body.js
repsac-by, line 423: It is not possible to decode each chunk from `base64' because the chunk may be an incomplete sequence of `base64`, which will lead to data corruption.

mcollina: I don't understand.

repsac-by:
```js
const data = Buffer.from(Buffer.from('hello world').toString('base64'));

{
	const chunks = [];
	for (let offset = 0, step = 6; offset <= data.length; offset += step) {
		const chunk = data.slice(offset, offset + step);
		chunks.push(Buffer.from(chunk.toString(), 'base64'));
	}

	console.log(await new Blob(chunks).text());
	// hell�v�ld
}

{
	const chunks = [];
	for (let offset = 0, step = 6; offset <= data.length; offset += step) {
		const chunk = data.slice(offset, offset + step);
		chunks.push(chunk);
	}

	const buffer = Buffer.from(Buffer.concat(chunks).toString(), 'base64');

	console.log(await new Blob([buffer]).text());
	// hello world
}
```

mcollina: We have the https://nodejs.org/api/string_decoder.html for this purpose.

cameron-robey:
Will get to finishing the last couple of things by the end of the week 👍

lib/fetch/body.js
cameron-robey: Added in https://github.com/nodejs/undici/pull/1606/commits/0a9785bd7ee28446b551b82b8ab4ba1012bea3c7

lib/fetch/body.js
cameron-robey: Added test in https://github.com/nodejs/undici/pull/1606/commits/0b4c50587ee1d77582e4702b1fec2af1ef0e9d86

lib/fetch/body.js
```diff
@@ -397,7 +398,47 @@ function bodyMixinMethods (instance) {
 
       // If mimeType’s essence is "multipart/form-data", then:
       if (/multipart\/form-data/.test(contentType)) {
-        throw new NotSupportedError('multipart/form-data not supported')
+        const headers = {}
+        for (const [key, value] of this.headers) headers[key.toLowerCase()] = value
+
+        const responseFormData = new FormData()
+
+        let busboy
```
kibertoad, line 406: since we throw when failing to resolve anyway, maybe we should move entire try-catch block into an internal function to make busboy const?

lib/fetch/body.js
kibertoad: @mcollina does string decoder need to be applied somewhere in this code explicitly, or it will be called down the line automatically?

lib/fetch/body.js
JacobMGEvans: Unless that is critical behavior to have, I would say this is a good place for MVP to iterate from and meets the initial criteria discussed with @mcollina

kibertoad: Agreed

lib/fetch/body.js
mcollina: I think in this case we should be using it manually.

test/fetch/client-fetch.js
```diff
@@ -165,21 +165,97 @@ test('unsupported formData 1', (t) => {
...
+
+  const server = createServer((req, res) => {
+    res.setHeader('content-type', 'multipart/form-data; boundary=' + boundary)
+    res.write(formRaw)
+    res.end()
+  })
+  t.teardown(server.close.bind(server))
+
+  await new Promise((resolve) => {
+    server.listen(0, async () => {
```
mcollina, line 189: I would recommend to avoid promises and callbacks.

mcollina (commented):
I think this is mostly good to go.

test/fetch/client-fetch.js
cameron-robey: I rewrote the tests here: https://github.com/nodejs/undici/pull/1606/commits/b079acb0948e32c656f576c7eeadf53db5c5c21b

Let me know if there is anything else that needs doing before we can get this landed

lib/fetch/body.js
kibertoad: @cameron-robey string decoder needs to be used here, based on the above

repsac-by: https://nodejs.org/api/string_decoder.html this is not what is needed. Because it decodes from buffer to a base64 string. In our case, it is supposed to decode base64 string into a buffer.

kibertoad: @repsac-by any alternate suggestions?

repsac-by: I haven't found any helpers without using external dependencies. We can collect all the chunks and decode them at the end, but this will entail double memory consumption. The second approach is as follows:

```js
busboy.on('file', (name, value, info) => {
          const { filename, encoding, mimeType } = info
          const base64 = encoding.toLowerCase() === 'base64'
          const chunks = []
          let base64buf = ''
          value.on('data', (chunk) => {
            if (base64) {
                base64buf += chunk.toString();
                const end = base64buf.length - base64buf.length % 4;
                chunk = Buffer.from(base64buf.slice(0, end), 'base64')
                base64buf = base64buf.slice(end);
            }
            chunks.push(chunk)
          })
          value.on('end', () => {
            if (base64) chunks.push(Buffer.from(base64buf, 'base64'))
            const file = new File(chunks, filename, { type: mimeType })
            responseFormData.append(name, file)
          })
```

kibertoad: @repsac-by what external dependencies would help?

repsac-by: @kibertoad https://github.com/mazira/base64-stream

kibertoad: Considering that the library wasn't updated for years, and decoder is [really small](https://github.com/mazira/base64-stream/blob/master/lib/decode.js), wonder if we should just inline it in the repo.
@mcollina thoughts?

kibertoad: @repsac-by btw, am I missing something, or https://github.com/mazira/base64-stream/blob/master/lib/decode.js does for the most part exact same thing as this PR?

repsac-by: @kibertoad  it does what I suggested in the second case, only with the difference that it applies to [stream.Transform](https://nodejs.org/dist/latest-v18.x/docs/api/stream.html#implementing-a-transform-stream). This is convenient because `value` in our case is a stream

test/fetch/client-fetch.js
```diff
@@ -165,24 +165,83 @@ test('unsupported formData 1', (t) => {
...
+    res.end()
+  })
+  t.teardown(server.close.bind(server))
+
+  server.listen(0, async () => {
+    const res = await fetch(`http://localhost:${server.address().port}`)
+    res.formData().catch(err => {
+      t.equal(err.message, 'Unexpected end of form')
+    })
+  })
```
mcollina, line 242: Try something like this, treat like it like pseudocode:

```suggestion
  const { promisify } = require('util')
  const listen = util.promisify(server.listen.bind(server))
  await listen(0)
  const res = await fetch(`http://localhost:${server.address().port}`)
  await t.rejects(res.formData(), 'Unexpected end of form')
```

lib/fetch/body.js
mcollina: I think this is the only comment left to resolve.

ronag (approved):

lib/fetch/body.js
cameron-robey: What exactly needs to be implemented - the thread is unclear @mcollina

kibertoad: @cameron-robey see what @repsac-by is suggesting, plus test to cover that case. inlining the code should be fine

mcollina: What's the problem exactly? You need 4 bytes to decode base64 at minimum, so why don't you accumulate until you have 4 bytes of multiples?

meyfa: @mcollina I'm assuming that you mean: push chunks until the total length of unprocessed data is divisible by 4, then decode and clear the chunk array, and repeat with the next pieces of data. Accumulating like that seems unreliable, since it cannot be known when that multiple of 4 will be reached. It might happen only after a huge number of chunks. [The streaming method suggested above](https://github.com/nodejs/undici/pull/1606#discussion_r967644495) is superior because the number of unprocessed bytes that are kept in memory never exceeds 3.

mcollina: The one in the comment also looks good.

kibertoad: @cameron-robey Do you need any help with the remaining part?

mcollina (approved):
lgtm

kibertoad:
@ronag why did we merge this without addressing all comments? should I create a follow-up PR?

RangerMauve:
Is there somewhere we can track the inclusion of this update into Node?

kibertoad:
https://github.com/nodejs/node/pulls?q=is%3Apr+undici+is%3Aclosed

Issue #1600, opened 2022-08-11
Query assertions on MockAgent intercept have ordering significance


James1x0:
## Bug Description
Query parameter assertions have ordering significance. Objects do have key ordering significance, however query parameters should not.

## Reproducible By

Consider this example:

```typescript
 mockAgent = new MockAgent();
 mockAgent.disableNetConnect();
 setGlobalDispatcher(mockAgent);

mockAgent.get('someHost').intercept({
    path: '/some/path',
    method: 'get',
    query: {
        somevalue: ['abc'], // note key ordering here.
        key: 'test',
    },
}).reply(200, 'yay');

const response = await request('someHost/some/path', {
    query: {
        key: 'test', // keys in different order
        somevalue: ['abc'],
    }
});
// ^ ❌ miss

const response = await request('someHost/some/path', {
    query: {
        somevalue: 'test', // keys in same order
        key: ['abc'],
    }
});
// ^ ✅ hit
```

## Expected Behavior
Request `query` key ordering should have no significance on assertions.

## Logs & Screenshots
N/A

## Environment
macOS 12.3.1, Node v16.13.2

### Additional context
I'm happy to submit a PR fixing this if the contributors agree with the expected behavior. Thanks!

mateonunez:
When the `getMockDispatch` method resolves the path, the **mockDispatches** are filtered by `consumed` props and the result of `matchValue` method, checking the following variables:

```js
path: "/foo?foo=bar&key=get";
resolvedPath: "/foo?key=get&foo=bar"
```

https://github.com/nodejs/undici/blob/deed6284ad97df01df92cee7576384bff7da1ad9/lib/mock/mock-utils.js#L110

mcollina:
I think the behavior is fundamentally correct. We cannot test or verify all permutations.
Having said that, I think we should provide better diagnostics to let people know where things got stuck.

James1x0:
> I think the behavior is fundamentally correct. We cannot test or verify all permutations.
> 
> Having said that, I think we should provide better diagnostics to let people know where things got stuck.

Hm, makes sense if for some reason it's not advisable to deserialize the query parameters.

How would you imagine diagnostics? Logging mismatched requests?

mcollina:
> How would you imagine diagnostics? Logging mismatched requests?

Having a callback in MockAgent that produce some kind of reporting object saying _why_ a request was not matched. Folks could then add a `console.log()` there if they want to.

PR #1607, opened 2022-08-18
fix: make mock intercept agnostic to query key ordering

James1x0:
Proposed solution to #1600 

--- 

- Added test for mis-ordered keys, verified failure against current functionality
- Added parsing for path comparison

Background: intercept does not match requests with query parameters if the keys are in a different order. From an http-standpoint query parameter key ordering does not matter. This causes some confusion if the intercept doesn't work for someone using the library.

Proposed solution: Sort the keys to build a "safeUrl" to prepare a comparable value using URLSearchParams.

Based on Matteo's feedback we may also want to introduce some diagnostic hooks if requests are missed. I'll follow up with another PR.

### DIFF
--- a/lib/mock/mock-utils.js
+++ b/lib/mock/mock-utils.js
@@ -85,6 +85,22 @@ function matchHeaders (mockDispatch, headers) {
   return true
 }
 
+function safeUrl (path) {
+  if (typeof path !== 'string') {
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

PR #1609, opened 2022-08-18
fix: allow third party AbortControllers

### LINKED ISSUE #1605: Can't bring custom `abort-controller` implementation to undici.fetch
## Bug Description

We are using `abort-controller` from npm in our codebase with undici. undici fails with:

```
TypeError: AbortSignal: Expected [object AbortSignal] to be an instance of AbortSignal.
```

## Reproducible By

https://replit.com/@SukkaW/ImaginaryDishonestCommunication#index.js

```js
const { AbortController } = require('abort-controller');

const controller = new AbortController()
controller.abort()

try {
    signal: controller.signal,
  })
} catch (error) {
  console.log(error.message)
}
```

## Expected Behavior

The error `AbortError: The operation is aborted` should be shown instead.

## Environment

Node.js 18.7.0 on macOS 13 Beta 4

## Additional Context

I have searched the codebase of undici and I notice this:

https://github.com/nodejs/undici/blob/5890e16ddd2703151ce0be0a468e13d685b89f60/lib/fetch/request.js#L838-L840

KhafraDev:
Complements #1608 (totally fixes #1605)

When better checks are performed in the `interfaceConverter` function, this partial revert will be more useful. From my comment in #1608: `Of course the webidl interface converter could also be improved such that when passing a strict: false option, it would check the object's Symbol.toStringTag/constructor name [instead of allowing any value].`

### DIFF
--- a/lib/fetch/request.js
+++ b/lib/fetch/request.js
@@ -835,6 +835,10 @@ webidl.converters.RequestInfo = function (V) {
   return webidl.converters.USVString(V)
 }
 
+webidl.converters.AbortSignal = webidl.interfaceConverter(
+  AbortSignal
+)
+
 // https://fetch.spec.whatwg.org/#requestinit
 webidl.converters.RequestInit = webidl.dictionaryConverter([
   {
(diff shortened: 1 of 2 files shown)


ronag (approved):

mcollina (approved):
lgtm

PR #1613, opened 2022-08-22
feat(fetch): implement spec compliant integrity checks

### LINKED ISSUE #1612: Undici crashes on node compiled without crypto
@nodejs/undici this seems to break with a container/machine without OpenSSL:
https://ci.nodejs.org/job/node-test-commit-linux-containered/33340/nodes=ubuntu1804_sharedlibs_withoutssl_x64/
```
17:44:45     [UNEXPECTED_FAILURE][FAIL] Importing an i32 global with an incorrectly-typed value: ""
17:44:45     Node.js is not compiled with OpenSSL crypto support
17:44:45     Error [ERR_NO_CRYPTO]: Node.js is not compiled with OpenSSL crypto support
```

_Originally posted by @MoLow in https://github.com/nodejs/node/issues/44319#issuecomment-1221580468_

KhafraDev:
Fixes #1612 

Bugs fixed:
- Disabled when nodejs is not built with openssl (see issue above)
- Disallows algorithms outside of the ones listed in the spec (sha256, sha384, sha512)
- Checks are ignored if an invalid hash/algorithm are provided (Firefox follows the spec; Chrome throws an error. Left a comment about this behavior.)
- Checks if the algorithms are supported by the user's node.
Tests were added in #1596 but I could add more if needed.

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -13,7 +13,7 @@ const { Headers } = require('./headers')
 const { Request, makeRequest } = require('./request')
 const zlib = require('zlib')
 const {
-  matchRequestIntegrity,
+  bytesMatch,
   makePolicyContainer,
   clonePolicyContainer,
   requestBadPort,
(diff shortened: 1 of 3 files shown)


ronag (approved):

mcollina (approved):
lgtm

PR #1631, opened 2022-09-02
feat(fetch): allow setting base urls

KhafraDev:
Notes:
1. ~~Will not work between different undici versions (as `setGlobalDispatcher`/`getGlobalDispatcher` do), because of https://github.com/nodejs/node/issues/43157.~~
2. Only allows `http:` or `https:` urls.
3. Allows strings, URL objects, or undefined, as ts lists all of them as a second parameter for the URL constructor.
4. Setting undefined resets the base url (relative paths will throw again).
5. No way to set for individual `fetch` calls. This could come later, but I'd rather not add a million undici-only keys to RequestInit.

Usage:
```js
import { setGlobalOrigin, getGlobalOrigin } from 'undici'

getGlobalOrigin() // undefined

getGlobalOrigin() // equal to new URL('http://localhost:3000')
```

### DIFF
--- a/lib/fetch/global.js
+++ b/lib/fetch/global.js
@@ -1,0 +1,36 @@
+'use strict'
+
+/** @type {undefined|URL} */
+let origin
+
+function getGlobalOrigin () {
+  return origin
+}
+
+function setGlobalOrigin (newOrigin) {
+  if (
+    newOrigin !== undefined &&
+    typeof newOrigin !== 'string' &&
+    !(newOrigin instanceof URL)
+  ) {
+    throw new Error('Invalid base url')
+  }
+
+  if (newOrigin === undefined) {
+    origin = newOrigin
+    return
+  }
+
+  const parsedURL = new URL(newOrigin)
+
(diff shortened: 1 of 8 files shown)


KhafraDev:
Is this change even wanted, or does it add ambiguity? I believe Deno offers a cli flag for doing something similar.

ronag (approved):
OK with me.

lib/fetch/global.js
```diff
@@ -0,0 +1,36 @@
+'use strict'
+
+/** @type {undefined|URL} */
+let origin
```
mcollina, line 4: I would use the same logic setGlobalDispatcher uses for this.

mcollina (changes requested):

lib/fetch/global.js
KhafraDev: Is it fine to add another symbol? Wasn't sure because of the issue linked above

mcollina: Sure thing.

KhafraDev: done

mcollina (approved):
lgtm

PR #1632, opened 2022-09-02
feat: export some fetch utilities

KhafraDev:
These fetch utils are required for a spec-compliant version of XMLHttpRequest.

I'd like to make https://github.com/khafradev/undici/tree/xmlhttprequest into its own package and exporting these would allow me to do so. 😄


mcollina (approved):
lgtm

PR #1646, opened 2022-09-13
ci: re-enable testing in Node.js 12

### LINKED ISSUE #1645: Node.js 12 support?
It seems as part of https://github.com/nodejs/undici/pull/1604, Node.js 12 was removed from the support matrix - not sure if that was intentional?

https://github.com/nodejs/undici/blob/88eca8d967897d68e9ce31b452e573c6eac1d278/.github/workflows/nodejs.yml#L23-L26

If Node.js 12 is no longer supported - then probably the `engines` field needs to be updated (incl. a major version release) - but quite possibly that was a typo and the matrix needs to be adjusted?

dominykas:
Closes #1645

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "test": "npm run test:tap && npm run test:node-fetch && npm run test:fetch && npm run test:jest && tsd",
     "test:node-fetch": "node scripts/verifyVersion.js 16 || mocha test/node-fetch",
     "test:fetch": "node scripts/verifyVersion.js 16 || (npm run build:node && tap test/fetch/*.js && tap test/webidl/*.js)",
-    "test:jest": "jest",
+    "test:jest": "node scripts/verifyVersion.js 14 || jest",
     "test:tap": "tap test/*.js test/diagnostics-channel/*.js",
     "test:tdd": "tap test/*.js test/diagnostics-channel/*.js -w",
     "test:typescript": "tsd",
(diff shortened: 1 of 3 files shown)


test/request-timeout.js
```diff
@@ -55,7 +57,7 @@ test('request timeout with readable body', (t) => {
       t.type(err, errors.HeadersTimeoutError)
     })
   })
-})
+}, { skip: nodeMajor < 14 })
```
dominykas, line 60: As suggested by Matteo: https://github.com/nodejs/undici/issues/1645#issuecomment-1246988294

package.json
```diff
@@ -49,7 +49,7 @@
     "test": "npm run test:tap && npm run test:node-fetch && npm run test:fetch && npm run test:jest && tsd",
     "test:node-fetch": "node scripts/verifyVersion.js 16 || mocha test/node-fetch",
     "test:fetch": "node scripts/verifyVersion.js 16 || (npm run build:node && tap test/fetch/*.js && tap test/webidl/*.js)",
-    "test:jest": "jest",
+    "test:jest": "node scripts/verifyVersion.js 14 || jest",
```
dominykas, line 52: `jest@29+`, used here, does not support Node.js 12...

dominykas:
Seems GH actions are stuck for this one job, wasting precious build minutes? I don't have access to stop/restart that job...

mcollina:
PTAL https://github.com/nodejs/undici/pull/1652

mcollina (approved):
lgtm

Issue #1470, opened 2022-05-29
docs: forbidden headers not documented


shellscape:
_(We don't have a documentation template here, so please forgive the use of a blank issue)_

I recently ran into an `invalid connection header` error on refactoring some older code from axios to undici. I found this error to be particularly curious, because it wasn't documented in this repo and I couldn't find any reference to that error in the Node docs. For what it's worth, here's some of the code being called:

```ts
const headers = {
  Accept:
    'text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.9',
  'Accept-Language': 'en-US,en;q=0.9,he-IL;q=0.8,he;q=0.7',
  'Cache-Control': 'no-cache',
  Connection: 'keep-alive',
  Pragma: 'no-cache',
  'Sec-Fetch-Dest': 'document',
  'Sec-Fetch-Site': 'same-origin',
  'Sec-Fetch-Mode': 'navigate',
  'Sec-Fetch-User': '?1',
  'Upgrade-Insecure-Requests': '1',
  'User-Agent':
    'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_3) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/80.0.3987.163 Safari/537.36'
};

...

const { body, statusCode } = await request(uri, { headers, bodyTimeout: 5e3 });
```

Tracing this in code, it looks like `undici` marks a `connection` header as invalid in two places:

https://github.com/nodejs/undici/blob/9e5316c8b04a7b35522d0d5b8de71f67fa2a3c34/lib/fetch/constants.js#L3

and

https://github.com/nodejs/undici/blob/1f2cca68f0608e31eb02f0d08956786dd1260ef9/lib/core/request.js#L281-L285

With the latter being the spot that raised the error I ran into. Now, I don't believe the original dev was correct in their use of the `keep-alive` connection header here and it'll be removed, but I was really surprised that _none_ of these forbidden headers were documented anywhere easily reachable, nor the conditions in which they are invalid. Perhaps this is another spot of documentation that was overlooked due to institutional knowledge of the contributors?

I _love_ what's being done with this project, but documentation here seems to be more of an afterthought and that's a bummer.

mcollina:
Thanks for reporting the issue! Indeed, this project was born out of a proof-of-concept and the docs where added later.

[...]

Issue #1307, opened 2022-03-28
Allow user provided connection header


ronag:
Override the pipelining option accordingly in case user passes a valid connection header.

Refs: https://github.com/nodejs/undici/issues/1305

metcoder95:
Just to recap out of the discussion of #1305 

The goal should be:

- Allow pass custom `connection-header` so no error is thrown when someone wants to override the default one
- Disable pipelining in case custom header is passed

Is correct, please let me know if I'm missing something 🙂

Issue #1305, opened 2022-03-28
Can't remove default headers. Can't overwrite `"sec-fetch-mode"`, `"connection"` headers.


AlttiRi:
- I can't remove any default header.
- Setting `"connection": "keep-alive"` throws an error.
- I can't overwrite default `"sec-fetch-mode"` header **_correctly_**. 

### 1. I can't remove the default headers

I can't remove the default headers: `host`, `connection`, `accept`, `accept-encoding`, `accept-language`, `sec-fetch-mode`, `user-agent`.

For example, `curl` allows to remove any header (even `host` header):
`curl http://127.0.0.1:3000 -H 'User-Agent:' -H 'Accept:' -H 'Host:'`
And yes, it works.

Using of `""` will send `""` as the header's value, using of `null` will send `"null"` as the header's value.

### 2. `"connection": "keep-alive"`

Also, if I manually set `connection: "keep-alive"` I get the error (`code: 'UND_ERR_INVALID_ARG'`).
```js
const resp = await fetch(url, {
    headers: {
        "connection": "keep-alive"
    }
});
```

It's important for code compatibility reason — since `node-fetch` uses `connection: close` by default.

### 3. `"sec-fetch-mode"`

It uses `"sec-fetch-mode": "cors"` by default.

As I said above, I can't remove this header _(as well as any other default header)_.

More over, **I can't change it correctly.**

For example, using of:
```js
await fetch(url, {
    headers: {
        "sec-fetch-mode": "navigate"
    }
});
```
Will produce a request with `sec-fetch-mode: navigate, cors`, while it should be `sec-fetch-mode: navigate`.

---

`"undici": "4.16.0"`

ronag:
Can you check what Chrome does? I think most of the behavior you are seeing is regulated by the spec.

ronag:
**2** and **3** are obviously bugs.

For **1**, for me:
- Either add a way to remove the default headers, ~for example, by setting its value to `null` (not `"null"`). It's the best choose especially if it will be a part of Node.js.~ (UPD: However it will not be compatible with native `fetch`, so no.)
- Or at least remove `sec-fetch-mode` from the default headers

https://fetch.spec.whatwg.org/#http-network-or-cache-fetch

Step 13

> Append the Fetch metadata headers for httpRequest. [FETCH-METADATA]

https://w3c.github.io/webappsec-fetch-metadata/#sec-fetch-mode-header

>  (and maybe `accept-language` too).

[...]

PR #1649, opened 2022-09-14
refactor: allow server-side headers

Kikobeats:
Hello,

At [Edge Runtime](https://github.com/vercel/edge-runtime/blob/main/packages/primitives/src/patches/undici-core-request.js#L2), we are patching the undici `Request` implementation to be more server-side friendly, allowing to pass `connection` header.

I guess `connection` and the rest of headers are forbidden to follow with `fetch` spec. Potentially there are more headers that should to be forbidden according to https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name

but since undici can be used in servers too, is it any chance to be less restrictive?

MDN is pointing something that could be key:

> Modifying such headers is forbidden because the user agent retains full control over them.

Maybe setup `user-agent: undefined` could be the way to discriminate these headers in any way?

Related:

- https://github.com/nodejs/undici/issues/1305
- https://github.com/nodejs/undici/issues/1307
- https://github.com/nodejs/undici/issues/1470

### DIFF
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -303,26 +303,6 @@ function processHeader (request, key, val) {
     request.contentType = val
     request.headers += `${key}: ${val}\r\n`
   } else if (
-    key.length === 17 &&
-    key.toLowerCase() === 'transfer-encoding'
-  ) {
-    throw new InvalidArgumentError('invalid transfer-encoding header')
-  } else if (
-    key.length === 10 &&
-    key.toLowerCase() === 'connection'
-  ) {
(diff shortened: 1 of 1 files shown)


ronag:
I'm not so sure about this. Need to think a bit.

mcollina:
I think the best avenue to discuss this is https://github.com/wintercg/fetch

ronag:
I don't understand the use case. Most of these headers is something the user should not be touching or using. Do you have some examples?

PR #1654, opened 2022-09-16
feat(fetch): better errors messages for body-mixin methods

### LINKED ISSUE #1653: Provide a hint how to solve a `disturbed` exception
I was using Node to fetch some JSON off a server and ran into an `disturbed` Exception after trying to call the `.json` method twice on the server response. At first, I could not locate what caused that in my application. Eventually, I found #1499 and realized my mistake. However, the error message could provide a hint towards what to fix to improve developer experience.

## This would solve...

Developer confusion when confronted with the `disturbed` exception from a package that is not even included in their `package-lock.json` 🤯

Here is an example error message:

```js
if (util.isDisturbed(object)) {
  throw new TypeError("Object is disturbed.\n\n" +
    "You may have already tried to consume this response. " +
    "At most one method that can consume a body may be called on a response."
  )
}
```

[This function](https://github.com/nodejs/undici/blob/dcd4e81b83c709f78ad2b28c844e9dac67f9078f/lib/fetch/body.js#L220) may currently be the culprit, but I'm not sure.

## I have also considered...

An alternative could be to clarify the concept of "disturbed" bodies within the documentation, a FAQ section, etc. However, placing helpful messages right inside the exception IMHO places them in front of the programmer when they need it.

KhafraDev:
Fixes #1653

The error messages are closer to chrome/firefox now. I don't think they need to be super descriptive.
image
image

edit: I considered adding the error header (ie. `Response.text`) but it would be relatively annoying to add the message to `.json()` as it just calls `.text()`, or the message would be completely wrong.

### DIFF
--- a/lib/fetch/body.js
+++ b/lib/fetch/body.js
@@ -230,9 +230,9 @@ function safelyExtractBody (object, keepalive = false) {
   if (object instanceof ReadableStream) {
     // Assert: object is neither disturbed nor locked.
     // istanbul ignore next
-    assert(!util.isDisturbed(object), 'disturbed')
+    assert(!util.isDisturbed(object), 'The body has already been consumed.')
     // istanbul ignore next
-    assert(!object.locked, 'locked')
+    assert(!object.locked, 'The stream is locked.')
   }
 
   // 2. Return the results of extracting object.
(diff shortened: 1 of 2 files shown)


ronag (approved):

mcollina (approved):
lgtm

PR #1662, opened 2022-09-22
feat: WPT runner

KhafraDev:
Implements a *very* basic WPT runner that should allow every test to run (theoretically - this PR only enables a single test to ensure that the runner is working).

The implementation is taken from https://github.com/KhafraDev/node-xmlhttprequest/tree/main/tests with some minor adjustments (setting exitCode on test failure, exiting if there are no tests to run).

### Pros:
1. Uses the WPT test harness, unlike node's builtin wpt runner. See https://github.com/nodejs/node/blob/098eac7f272f5e720c9aff8b232501928a88b82a/test/common/wpt.js#L12-L14.
2. Each test runs in its own context, and the tests are run in a Worker. The environment these tests run in should be identical to an actual node.js runtime, so hopefully there won't be any unexpected jest-like issues.
3. Has a built-in server that allows serving resources for tests and enabling a very basic routing system. Note that the server runs in its own process (using `child_process.fork`) and communicates via ipc. This is because the xhr tests (from my package mentioned above) test synchronous requests that would block the thread and cause requests to never complete. This workaround probably isn't needed here but I decided not to make many changes.

### Cons
1. Uses esm because of personal preference. Also makes use of top level await and other esm-only things, however it could be re-written to use cjs.
2. Probably only works on node v18 or above; hasn't been tested in v16.
3. Doesn't log test completions; only test failures. Currently there isn't a way to enable flaky tests or expected test failures either.

### DIFF
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "build:wasm": "node build/wasm.js --docker",
     "lint": "standard | snazzy",
     "lint:fix": "standard --fix | snazzy",
-    "test": "npm run test:tap && npm run test:node-fetch && npm run test:fetch && npm run test:jest && tsd",
+    "test": "npm run test:tap && npm run test:node-fetch && npm run test:fetch && npm run test:wpt && npm run test:jest && tsd",
     "test:node-fetch": "node scripts/verifyVersion.js 16 || mocha test/node-fetch",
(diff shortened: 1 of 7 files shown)


ronag (approved):
Nice!

mcollina:
Can you add it to CI?

KhafraDev:
added to the CI

mcollina (approved):
lgtm

PR #1664, opened 2022-09-23
feat(WPTRunner): parse `META` tags

KhafraDev:
- Allows parsing `META` tags and injecting individual scripts into each test. Example of `META` tags: https://github.com/web-platform-tests/wpt/blob/master/fetch/api/abort/general.any.js#L1-L5
- Runs each test in its own Worker.
- Throws an AbortError DOMException in the body mixin methods if the request has been aborted, rather than a TypeError. This behavior, albeit *not* mentioned in the spec, is tested for in the WPTs.
- Supports enabling flaky/expected failures for certain tests.
- Logs stats (# completed, # succeeded, # failed, etc.) once every test is completed.


KhafraDev:
This test (taken from wpts) should pass but currently fails. It also fails on Deno, but all 4 major browsers pass it: https://wpt.fyi/results/fetch/api/abort/general.any.html?label=experimental&label=master&aligned

```js
const assert_true = (a, message) => {
	if (a !== true) {
		throw new Error(message)
	}
}

const signal = controller.signal;

const response = await fetch(`https://example.com`, { signal });

// Read whole response to ensure close signal has sent.
await response.clone().text();

const reader = response.body.getReader();

controller.abort();

const item = await reader.read();

assert_true(item.done, "Stream is done");
```

KhafraDev:
This test is failing and I'm not really sure how to fix it 😕 

```js
import { fetch, setGlobalOrigin } from 'undici'
import assert from 'assert'

const signal = controller.signal;
controller.abort();

const log = [];

await Promise.all([
	fetch('../resources/data.json', { signal }).then(
		() => assert_unreached("Fetch must not resolve"),
		() => log.push('fetch-reject')
	),
	Promise.resolve().then(() => log.push('next-microtask'))
]);

assert.deepStrictEqual(log, ['fetch-reject', 'next-microtask']);
```

mcollina (approved):
lgtm

KhafraDev:
The `throwIfAborted` steps in the body mixin methods aren't correct - there's a bug somewhere in undici I haven't managed to figure out just yet. However it could also be fixed in a separate pr as the workaround doesn't seem to cause issues.

mcollina (approved):
lgtm

PR #1665, opened 2022-09-24
feat: expose socket info on request

fengmk2:


ronag (changes requested):
Isn't it better to send this info in the `onConnect` handler? Also I think sending 

```
    timeout: socket.timeout,
    bytesWritten: socket.bytesWritten,
    bytesRead: socket.bytesRead
```

here is a bad idea...

fengmk2:
@ronag request error will contains socket info https://github.com/nodejs/undici/blob/main/lib/client.js#L967 , should keep the same behaviour.

Issue #1672, opened 2022-09-29
Improve buildURL performance


anonrig:
## This would solve...

The current implementation is quite slow. I recommend using [fast-querystring](https://github.com/anonrig/fast-querystring) and replacing the existing querystring builder.

## The implementation should look like...

Replacing buildURL function with [fast-querystring](https://github.com/anonrig/fast-querystring) and improve the performance (with a proper benchmark)
 https://github.com/nodejs/undici/blob/0d1419ce8980e8c0f289e876448bf1e6eb755a8d/lib/core/util.js#L39

## I have also considered...

Just copying the code of [fast-querystring](https://github.com/anonrig/fast-querystring) and putting inside the undici, but I believe it will bloat the codebase.

## Additional context

I'm the author of [fast-querystring](https://github.com/anonrig/fast-querystring) which makes me biased towards this decision. That's why I wanted to open an issue before a pull request.

kibertoad:
This is a hot path for the large number of GET requests, so having this would be nice.

ronag:
PR welcome

anonrig:
I created a pull request and added a comment regarding the benchmarks @ronag https://github.com/nodejs/undici/pull/1673

PR #1673, opened 2022-09-29
feat(util): use fast-querystring for building url

anonrig:
Solves https://github.com/nodejs/undici/issues/1672.

**Important Note**: Even though none of the benchmarks focus on query-string building (can test it with just adding `new Error('hello')` inside buildURL function), the results of the benchmarks show significant degregation in ops/sec. My educated guess would be: Due to requiring & resolving a new package, and how `cronometro` creates an isolated v8 environment every time a benchmark is run, the require is never cached, and therefore the amount of time requiring the `fast-querystring` package is also included in the benchmark and as a result creates a **false** outcome.

### DIFF
--- a/lib/core/util.js
+++ b/lib/core/util.js
@@ -8,6 +8,7 @@ const net = require('net')
 const { InvalidArgumentError } = require('./errors')
 const { Blob } = require('buffer')
 const nodeUtil = require('util')
+const queryString = require('fast-querystring')
 
 function nop () {}
 
@@ -26,47 +27,12 @@ function isBlobLike (object) {
   )
 }
 
-function isObject (val) {
-  return val !== null && typeof val === 'object'
-}
-
-// this escapes all non-uri friendly characters
-function encode (val) {
-  return encodeURIComponent(val)
-}
-
-// based on https://github.com/axios/axios/blob/63e559fa609c40a0a460ae5d5a18c3470ffc6c9e/lib/helpers/buildURL.js (MIT license)
 function buildURL (url, queryParams) {
   if (url.includes('?') || url.includes('#')) {
     throw new Error('Query params cannot be passed when url already contains "?" or "#".')
   }
-  if (!isObject(queryParams)) {
-    throw new Error('Query params must be an object')
-  }
 
-  const parts = []
-  for (let [key, val] of Object.entries(queryParams)) {
-    if (val === null || typeof val === 'undefined') {
-      continue
-    }
-
-    if (!Array.isArray(val)) {
-      val = [val]
-    }
-
-    for (const v of val) {
-      if (isObject(v)) {
-        throw new Error('Passing object as a query param is not supported, please serialize to string up-front')
-      }
-      parts.push(encode(key) + '=' + encode(v))
-    }
-  }
-
-  const serializedParams = parts.join('&')
-
-  if (serializedParams) {
-    url += '?' + serializedParams
-  }
+  url += '?' + queryString.stringify(queryParams)
 
   return url
 }
--- a/lib/core/util.js
+++ b/lib/core/util.js
@@ -8,7 +8,7 @@ const net = require('net')
 const { InvalidArgumentError } = require('./errors')
 const { Blob } = require('buffer')
 const nodeUtil = require('util')
-const queryString = require('fast-querystring')
+const { stringify } = require('fast-querystring')
 
 function nop () {}
 
@@ -32,7 +32,7 @@ function buildURL (url, queryParams) {
     throw new Error('Query params cannot be passed when url already contains "?" or "#".')
   }
 
-  url += '?' + queryString.stringify(queryParams)
+  url += '?' + stringify(queryParams)
 
   return url
 }
--- a/test/client.js
+++ b/test/client.js
@@ -97,6 +97,8 @@ test('basic get with query params', (t) => {
       bool: 'true',
       foo: '1',
       bar: 'bar',
+      nullVal: '',
+      undefinedVal: '',
       '%60~%3A%24%2C%2B%5B%5D%40%5E*()-': '%60~%3A%24%2C%2B%5B%5D%40%5E*()-',
       multi: ['1', '2']
     })
@@ -134,95 +136,6 @@ test('basic get with query params', (t) => {
       t.equal(statusCode, 200)
     })
     t.equal(signal.listenerCount('abort'), 1)
-  })
-})
-
-test('basic get with query params with object throws an error', (t) => {
-  t.plan(1)
-
-  const server = createServer((req, res) => {
-    t.fail()
-  })
-  t.teardown(server.close.bind(server))
-
-  const query = {
-    obj: { id: 1 }
-  }
-
-  server.listen(0, () => {
-    const client = new Client(`http://localhost:${server.address().port}`, {
-      keepAliveTimeout: 300e3
-    })
-    t.teardown(client.close.bind(client))
-
-    const signal = new EE()
-    client.request({
-      signal,
-      path: '/',
-      method: 'GET',
-      query
-    }, (err, data) => {
-      t.equal(err.message, 'Passing object as a query param is not supported, please serialize to string up-front')
-    })
-  })
-})
-
-test('basic get with non-object query params throws an error', (t) => {
-  t.plan(1)
-
-  const server = createServer((req, res) => {
-    t.fail()
-  })
-  t.teardown(server.close.bind(server))
-
-  const query = '{ obj: { id: 1 } }'
-
-  server.listen(0, () => {
-    const client = new Client(`http://localhost:${server.address().port}`, {
-      keepAliveTimeout: 300e3
-    })
-    t.teardown(client.close.bind(client))
-
-    const signal = new EE()
-    client.request({
-      signal,
-      path: '/',
-      method: 'GET',
-      query
-    }, (err, data) => {
-      t.equal(err.message, 'Query params must be an object')
-    })
-  })
-})
-
-test('basic get with query params with date throws an error', (t) => {
-  t.plan(1)
-
-  const date = new Date()
-  const server = createServer((req, res) => {
-    t.fail()
-  })
-  t.teardown(server.close.bind(server))
-
-  const query = {
-    dateObj: date
-  }
-
-  server.listen(0, () => {
-    const client = new Client(`http://localhost:${server.address().port}`, {
-      keepAliveTimeout: 300e3
-    })
-    t.teardown(client.close.bind(client))
-
-    const signal = new EE()
-    client.request({
-      signal,
-      path: '/',
-      method: 'GET',
-      query
-    }, (err, data) => {
-      t.equal(err.message, 'Passing object as a query param is not supported, please serialize to string up-front')
-    })
   })
 })
 
--- a/test/client.js
+++ b/test/client.js
@@ -99,6 +99,8 @@ test('basic get with query params', (t) => {
       bar: 'bar',
       nullVal: '',
       undefinedVal: '',
+      functionVal: '',
+      objectVal: '',
       '%60~%3A%24%2C%2B%5B%5D%40%5E*()-': '%60~%3A%24%2C%2B%5B%5D%40%5E*()-',
       multi: ['1', '2']
     })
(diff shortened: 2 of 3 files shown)


anonrig:
Destructuring fast-querystring and calling it using `const { stringify } = require('fast-querystring')` improved the benchmarks. 

Screen Shot 2022-09-29 at 6 10 11 PM

test/client.js
```diff
@@ -137,95 +139,6 @@ test('basic get with query params', (t) => {
   })
 })
 
-test('basic get with query params with object throws an error', (t) => {
```
kibertoad, line 140: if this no longer throws an error, tests need to be adjusted to document new behaviour.
what was the handling of these edge cases based on?

lib/core/util.js
```diff
@@ -26,47 +27,12 @@ function isBlobLike (object) {
...
 
-    if (!Array.isArray(val)) {
-      val = [val]
-    }
-
-    for (const v of val) {
-      if (isObject(v)) {
-        throw new Error('Passing object as a query param is not supported, please serialize to string up-front')
-      }
-      parts.push(encode(key) + '=' + encode(v))
```
kibertoad, line 61: is the same style of ?x=1&x=2 preserved?

ronag:
@mcollina wdyt? Another external dependency?

lib/core/util.js
anonrig: Yes, it's preserved. A similar test exists in `fast-querystring` test module: https://github.com/anonrig/fast-querystring/blob/main/test/node.ts#L57

test/client.js
anonrig: Well, this was unorthodox. `new URL`, `node:querystring` and `fast-querystring` supports both array and object as an input.

kibertoad: can you add args of those type into existing happy-path test which used to pass args of all supported types?

anonrig: Thanks for the review. I truly appreciate it. I added a new test for this particular happy path.

test/client.js
```diff
@@ -97,6 +97,8 @@ test('basic get with query params', (t) => {
       bool: 'true',
       foo: '1',
       bar: 'bar',
+      nullVal: '',
```
kibertoad, line 100: can you also add an object and a function to the mix?

kibertoad: those tests are in place to document how undici itself behaves. what library or algorithm is used underneath is an implementation detail.

anonrig: That's a really good argument. I added a new commit and updated the tests.

ronag (approved):

KhafraDev:
There are very minor differences between undici's current implementation and fast-querystring. These include:
1. Using Object.keys instead of Object.entries
2. Using a traditional for loop instead of a for..of
3. Joining each part in a string, not an array.

Implementing these minor changes, the performance of the current impl. is faster/on par with fast-querystring.

```diff
diff --git a/lib/core/util.js b/lib/core/util.js
index e9a8384..92e0079 100644
--- a/lib/core/util.js
+++ b/lib/core/util.js
@@ -44,9 +44,13 @@ function buildURL (url, queryParams) {
     throw new Error('Query params must be an object')
   }

-  const parts = []
-  for (let [key, val] of Object.entries(queryParams)) {
-    if (val === null || typeof val === 'undefined') {
+  let parts = ''
+  const keys = Object.keys(queryParams)
+  for (let i = 0; i < keys.length; i++) {
+    let key = keys[i]
+    let val = queryParams[key]
+
+    if (val == null) {
       continue
     }

@@ -58,14 +62,13 @@ function buildURL (url, queryParams) {
       if (isObject(v)) {
         throw new Error('Passing object as a query param is not supported, please serialize to string up-front')
       }
-      parts.push(encode(key) + '=' + encode(v))
+      parts += encode(key) + '='
+      parts += encode(v) + '&'
     }
   }

-  const serializedParams = parts.join('&')
-
-  if (serializedParams) {
-    url += '?' + serializedParams
+  if (parts) {
+    url += '?' + parts
   }

   return url
```

```
╔═════════════════╤═════════╤═══════════════════╤═══════════╗
║ Slower tests    │ Samples │            Result │ Tolerance ║
╟─────────────────┼─────────┼───────────────────┼───────────╢
║ fastquerystring │   10000 │ 3486506.00 op/sec │ ± 11.53 % ║
╟─────────────────┼─────────┼───────────────────┼───────────╢
║ Fastest test    │ Samples │            Result │ Tolerance ║
╟─────────────────┼─────────┼───────────────────┼───────────╢
║ undici          │   10000 │ 3892565.20 op/sec │ ±  2.43 % ║
╚═════════════════╧═════════╧═══════════════════╧═══════════╝
```

However I also noticed that the benchmark never gives the same results between results. One run might have undici destroy fast-querystring, and another run will show the exact opposite. However the perf gains from the modified impl. destroy the current one.

kibertoad:
@KhafraDev Could you open an alternative PR with these changes?

anonrig:
@KhafraDev Your code change does not reflect the truth, since it is not compliant with `new URL` and `node:querystring`. I recommend testing your change on the test suite of `fast-querystring`

1. The `&` character should only be appended to parts if `i !== keys.length - 1`. If you include an additional if check there, I think the result would be different. Incorrect line: `parts += encode(v) + '&'`
2. `if (val == null) {` check is invalid since, both `new URL`, `node:querystring` adds empty string `''` to keys that have falsy values. You should not skip those key value pairs.

Can you share with us the updated benchmarks after these changes?

anonrig:
@KhafraDev Can you also share your input for the benchmarks? Here's `fast-querystring`s benchmark input: https://github.com/anonrig/fast-querystring/blob/main/benchmark/stringify.mjs#L11

KhafraDev:
The idea I was going for was to match undici's current behavior, rather than mimicking fast-querystring's. It would take slightly more work in the future to add support for bigints (etc.) which seemed out of scope for the current pr (improving perf).

The & getting appended to the last value is a mistake on my part, I've amended it 👍.

Here's an update benchmark for the changes; as mentioned previously the benchmark results skew heavily between runs.
```
╔═════════════════╤═════════╤═══════════════════╤═══════════╗
║ Slower tests    │ Samples │            Result │ Tolerance ║
╟─────────────────┼─────────┼───────────────────┼───────────╢
║ undici          │   10000 │ 3829950.21 op/sec │  ± 1.93 % ║
╟─────────────────┼─────────┼───────────────────┼───────────╢
║ Fastest test    │ Samples │            Result │ Tolerance ║
╟─────────────────┼─────────┼───────────────────┼───────────╢
║ fastquerystring │   10000 │ 3851338.34 op/sec │  ± 2.92 % ║
╚═════════════════╧═════════╧═══════════════════╧═══════════╝
```

Uzlopak:
Just a remark: When you run your benchmarks you should imho set in your cpu governor all cpu cores to performance mode. Your benchmark tolerance is imho too high. It should be about 1-2%. Tolerance of 18 % indicates to me that your cpu cores are in on-demand mode and need to "spin up". So your benchmark is kind of tainted because e.g. your cpu starts with e.g. 800 Mhz and then goes up to 3 Ghz while benchmarking and that results in the bigger tolerance. 

But that is only a remark and maybe your benchmarks are valid.

anonrig:
> This is the main scenario that we should be comparing, TBH. And it feels like we should add it to our benchmark.

You're correct @kibertoad Can you direct me to the correct place in the code to define a complex query-string that directly triggers `buildURL` instead of `parseURL`?

kibertoad:
@anonrig Main place that calls `buildURL` is `this.path = query ? util.buildURL(path, query) : path`
This is invoked when you are passing query params like this
```
client.request({
      path: '/',
      method: 'GET',
      query: { /* some query params */ }
    }
```

and not like this:
```
client.request({
      path: '/?someQueryParams=xyz',
      method: 'GET',
    }
```

So we need to add cases like that into https://github.com/nodejs/undici/blob/main/benchmarks/benchmark.js

KhafraDev:
> Just a remark: When you run your benchmarks you should imho set in your cpu governor all cpu cores to performance mode. Your benchmark tolerance is imho too high. It should be about 1-2%. Tolerance of 18 % indicates to me that your cpu cores are in on-demand mode and need to "spin up". So your benchmark is kind of tainted because e.g. your cpu starts with e.g. 800 Mhz and then goes up to 3 Ghz while benchmarking and that results in the bigger tolerance.
> 
> But that is only a remark and maybe your benchmarks are valid.

Unfortunately stuck using wsl. However I don't think any benchmarks here are necessarily valid regardless.

Since the benchmarks do not ever call `buildURL`, performance should have been equal with and without moving to fast-querystring. Then there's the question of if `buildURL` is even a performance bottleneck - in the opening issue it mentioned it was "slow", but that was about it. Would switching to fast-querystring improve performance in a real world situation?

kibertoad:
@KhafraDev considering that it is faster at some cases and slower at none, and also reduces complexity on undici side, isn't that a net positive?

KhafraDev:
I'm personally for using as few third party dependencies as possible, especially in cases where it's feasible to make such an area in the code more performant without it being complex.

However, I decided to run the benchmarks in fast-querystring and found that if the object being stringified only consists of strings and arrays, node's builtin querystring is actually on par with fast-querystring (the benchmarks vary with one never being 10% slower than the other). For undici's current behavior of only accepting strings and arrays, it seems like the better choice is to use querystring instead of adding another dependency.

```js
const value = {
  frappucino: "muffin",
  goat: "scone",
  pond: "moose",
  foo: ["bar", "baz", "bal"],
  bool: `${true}`,
  bigIntKey: BigInt(100).toString(),
  numberKey: `${256}`,
};
```
Object modified from the fast-qs benchmark to only contain string and an array.

p.s. the modified buildURL came in third place in all attempts, but isn't that important to include.

kibertoad:
@KhafraDev built-in querystring is deprecated (for reasons that I do not understand), should we be depending on that?

Uzlopak:
@kibertoad 
querystring is not deprecated. It is legacy.

kibertoad:
True, which still means "discouraged to use":
"The querystring API is considered Legacy. While it is still maintained, new code should use the [<URLSearchParams>](https://nodejs.org/api/url.html#class-urlsearchparams) API instead", according to docs.

mcollina:
`URLSearchParams` is way too slow for use in this library. I'm ok to use `querystring` or `fast-querystring`.

kibertoad:
personally I'm +1 on fast-querystring

ronag:
If fast-querystring is only negligable faster then I don't see any reason to use it over querystring.

kibertoad:
It seems that majority opinion is in favour of switching to querystring
@anonrig I take it it would be a drop-in replacement for this PR, as API is the same?

kibertoad:
@KhafraDev what about booleans and numbers? undici accepts these just fine too, and they are quite common

KhafraDev:
Same result. Both are nearly equally fast. I can post benchmark results if you'd like.

anonrig:
Thank you for all the comments. I was on my way to Ireland and just had the time to review and read the responses.

From my perspective, here are the steps we need to take in order to improve `undici` query-string parsing.

1. Add `query` parameter to benchmark in a separate pull-request to make sure the benchmarks include the proper input. This will give us the ability to receive proper benchmark results for the future. **(I've already opened a pull request [for this change](https://github.com/nodejs/undici/pull/1676))**
2. Open a separate pull-request, when `step #1` is merged, to replace existing implementation with node's legacy `querystring` module, to support additional value types such as `bigint`, `null`, `undefined` etc. The reason of making this a different pull request is that, I actually want to see the `ops/sec` difference in the Github workflow, since a [readme of fast-querystring](https://github.com/anonrig/fast-querystring) does not mean anything, unless its justified with a real-world scenario. My only concern in this step is that, Node's recommendation of not using legacy modules.
3. Open a separate pull-request when querystring module is merged into the core, to make sure the performance impact of `fast-querystring` is justifiable to include into the `undici` library as an external dependency

Even though, I don't agree with some of the comments made into this pull request, I think that we are on the correct path to improve the existing implementation.

PS: I didn't want to update and replace this pull request with a `node:querystring` implementation because I believe that this is a really good example of good communication, engineering, and receiving feedback. (and I just want to preserve that)

PR #1676, opened 2022-10-02
perf: add query parameter to benchmark

anonrig:
Referencing https://github.com/nodejs/undici/pull/1673, this pull request adds `query` as a parameter to benchmarks.

### DIFF
--- a/benchmarks/benchmark.js
+++ b/benchmarks/benchmark.js
@@ -32,6 +32,14 @@ const httpBaseOptions = {
   hostname: 'localhost',
   method: 'GET',
   path: '/',
+  query: {
+    frappucino: 'muffin',
+    goat: 'scone',
+    pond: 'moose',
+    foo: ['bar', 'baz', 'bal'],
+    bool: true,
+    numberKey: 256
+  },
   ...dest
 }
 


ronag:
Not sure this is very indicative for the most common use cases...

anonrig:
Thank you for the comment @ronag. This is the worst-case scenario, since `URL` and `node:querystring` supports all parameters in this object, but I'm happy to adjust it if you have any suggestions.

anonrig:
Here are the benchmark inputs from the Node repository: https://github.com/nodejs/node/blob/main/benchmark/querystring/querystring-stringify.js#L11

ronag:
I think we should optimize for most common case rather than worst case. I'm not sure what that is but I'm pretty sure e.g. BigInt is quite unusual...

anonrig:
> I think we should optimize for most common case rather than worst case. I'm not sure what that is but I'm pretty sure e.g. BigInt is quite unusual...

That seems reasonable. I'll remove it, and update the pull-request.

ronag (approved):

KhafraDev (approved):

PR #1677, opened 2022-10-02
feat(util): use node:querystring for buildURL

KhafraDev:


ronag (commented):
Why remove tests?

KhafraDev:
the tests were checking error messages that were no longer applicable, I added tests for a bunch of types as a replacement. I could add them back but i'm not sure how useful they would be

ronag (approved):

ronag:
LGTM assuming we are pretty sure this is not breaking

lib/core/util.js
```diff
@@ -26,49 +27,12 @@ function isBlobLike (object) {
...
-  }
-
-  const serializedParams = parts.join('&')
-
-  if (serializedParams) {
-    url += '?' + serializedParams
-  }
-
-  return url
+  return url + '?' + stringify(queryParams)
```
anonrig, line 35: What happens if stringify returns empty string? Can you handle that scenario?

anonrig (changes requested):

anonrig:
@KhafraDev Can you rebase your branch?

mcollina (approved):
lgtm

lib/core/util.js
ronag: Do we need to check if the url already contains a query string?

anonrig: We don't need to do that since we already do that at the beginning of the string by checking for `?` or `#`.

KhafraDev: an empty string is handled now 👍

anonrig (approved):

ronag (approved):

kibertoad:
@ronag @mcollina It's been forever since the last release, shall a new version be cut?

anonrig:
I'll open a new pull request to see the benchmark results for before/after `fast-querystring`.

PR #1678, opened 2022-10-02
fix(fetch): multiple data: url fixes & move tests to wpt runner

KhafraDev:
# Changes:

1. Correctly passes the `extract value` flag in `collectAnHttpQuotedString` which caused at least 1 test to fail.
2. Does not trim leading http whitespace in `parseMIMEType` which caused at least 1 test to fail.
3. Replaces custom serialization with spec-compliant one which caused multiple tests to fail.
4. Moves WPT data: url and base64 tests to the WPT runner.
5. Changes the url structure in the wpt runner's server to match the folder structure.
6. Update wpt data: url tests to include new tests.


test/fetch/data-uri.js
```diff
@@ -121,7 +119,7 @@ test('https://mimesniff.spec.whatwg.org/#parse-a-mime-type', (t) => {
   t.same(parseMIMEType('text/html;charset="shift_jis"iso-2022-jp'), {
     type: 'text',
     subtype: 'html',
-    parameters: new Map([['charset', '"shift_jis"']])
+    parameters: new Map([['charset', 'shift_jis']])
```
KhafraDev, line 122: not really sure where I got the quotes from but this is even the wrong output as shown in https://mimesniff.spec.whatwg.org/#example-mime-type-parser-trailing-garbage

ronag (approved):

PR #1679, opened 2022-10-03
perf: replace querystring with fast-querystring

anonrig:
I wanted to see the performance difference between legacy and fast-querystring benchmarks in undici.

### DIFF
--- a/lib/core/util.js
+++ b/lib/core/util.js
@@ -8,7 +8,7 @@ const net = require('net')
 const { InvalidArgumentError } = require('./errors')
 const { Blob } = require('buffer')
 const nodeUtil = require('util')
-const { stringify } = require('querystring')
+const stringify = require('fast-querystring/lib/stringify')
 
 function nop () {}
 
--- a/package.json
+++ b/package.json
@@ -126,6 +126,7 @@
     ]
   },
   "dependencies": {
-    "busboy": "^1.6.0"
+    "busboy": "^1.6.0",
+    "fast-querystring": "^1.0.0"
   }
 }


anonrig:
Here's the benchmark of just importing the packages. Since `node:querystring` ships with node itself, the requiring is a lot faster (cached), but for others it's not... I wonder how we can improve the benchmarks to reflect real-world usage, instead of including require times.

I'll also update the `fast-querystring` repository with the benchmarks for this. 

```
➜  fast-querystring git:(main) ✗ node benchmark/import.mjs
╔═════════════════════════════╤═════════╤══════════════════╤═══════════╗
║ Slower tests                │ Samples │           Result │ Tolerance ║
╟─────────────────────────────┼─────────┼──────────────────┼───────────╢
║ @aws-sdk/querystring-parser │    1000 │  10154.56 op/sec │  ± 0.43 % ║
║ querystringparser           │    1000 │  10624.80 op/sec │  ± 0.53 % ║
║ query-string                │    1000 │  11004.50 op/sec │  ± 0.47 % ║
║ qs                          │    1000 │  11077.09 op/sec │  ± 0.37 % ║
║ querystringify              │    1000 │  11147.54 op/sec │  ± 0.51 % ║
║ fast-querystring            │    1000 │  38184.61 op/sec │  ± 0.84 % ║
╟─────────────────────────────┼─────────┼──────────────────┼───────────╢
║ Fastest test                │ Samples │           Result │ Tolerance ║
╟─────────────────────────────┼─────────┼──────────────────┼───────────╢
║ node:querystring            │   10000 │ 224756.38 op/sec │  ± 1.14 % ║
╚═════════════════════════════╧═════════╧══════════════════╧═══════════╝
```

Kikobeats:
maybe better to just use a native API? https://github.com/nodejs/undici/pull/1682

anonrig:
> maybe better to just use a native API? https://github.com/nodejs/undici/pull/1682

Native API is slower. Benchmarks are available at github.com/anonrig/fast-querystring

silverwind:
> > maybe better to just use a native API? #1682
> 
> Native API is slower. Benchmarks are available at github.com/anonrig/fast-querystring

But is it 100% compatible with the standardized `URLSearchParams` interface? Imho, speed does not matter if it comes at the cost of compatibilty.

mcollina:
> Thought I do recall URLSearchParams does for example deal oddly with falsy values which may often not be desirable, so maybe 100% compatibilty to URLSearchParams is not a goal.

Compatiblity with `URLSearchParams` is not a goal. Essentially `URLSearchParams` should be avoided as it's significantly slower than the old `querystring` module.

ronag (approved):

ronag:
@mcollina @KhafraDev

anonrig:
Benchmarks does not reflect the truth because requiring node:querystring versus requiring fast-querystring is not the same, due to snapshotting. If you compare the functionality, it is 2x faster.

PR #1681, opened 2022-10-04
feat(fetch): add `Request{Init}.duplex` and add WPTs

KhafraDev:
- Adds [`Request.duplex`](https://github.com/whatwg/fetch/commit/1fbc40c1df1a353ce1a49de8e4b2c753563e265d) and [`RequestInit.duplex`](https://github.com/whatwg/fetch/commit/edf07e58abdf6fbef0e1b0c3a6788dbfe93e14ad)
- Handles `title` attributes in META tags.
- Better expected fails - only fails tests in a file with the test name now. Can also be adjusted to allow skipping entire files, etc. in the future (similar to node core).
- Most importantly, fixes an issue with using a dynamic import that would cause multiple `test(...)` functions to fail due to the internal state being set. I could not tell you why this happened.

Both tests that were added to the expected fails could likely be fixed, but I haven't looked into it.

### DIFF
--- a/lib/fetch/request.js
+++ b/lib/fetch/request.js
@@ -472,7 +472,13 @@ class Request {
     // 38. If inputOrInitBody is non-null and inputOrInitBody’s source is
     // null, then:
     if (inputOrInitBody != null && inputOrInitBody.source == null) {
-      // 1. If this’s request’s mode is neither "same-origin" nor "cors",
+      // 1. If initBody is non-null and init["duplex"] does not exist,
+      //    then throw a TypeError.
+      if (initBody != null && init.duplex == null) {
+        throw new TypeError(`RequestInit: duplex option is required when sending a body.`)
+      }
+
+      // 2. If this’s request’s mode is neither "same-origin" nor "cors",
       // then throw a TypeError.
       if (request.mode !== 'same-origin' && request.mode !== 'cors') {
         throw new TypeError(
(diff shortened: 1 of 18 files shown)


ronag (approved):

mcollina (approved):
lgtm

benmccann:
This appears to be causing a regression. We got a report that requests in SvelteKit are now failing: https://github.com/sveltejs/kit/issues/7409. It seems that the `duplex` field is now required. Someone has proposed always providing it in SvelteKit requests though that does not seem correct to me: https://github.com/sveltejs/kit/pull/7412

mcollina:
can you open a new issue to track the regression, ideally with a repro?

KhafraDev:
This is part of the spec now. When you have a ReadableStream body, the `duplex` option is required.

benmccann:
Thanks! We'll get SvelteKit updated. I wonder if this should have been a major release as it's a breaking change?

KhafraDev:
Fetch in undici and node core are both considered experimental, so breaking changes should be expected.

See https://github.com/nodejs/undici#undicifetchinput-init-promise and https://nodejs.org/dist/latest-v18.x/docs/api/all.html#all_globals_fetch

It would likely be a good idea to pin undici's version and do manual updates to ensure nothing breaks between versions

PR #1682, opened 2022-10-04
use URLSearchParams instead of querystring

Kikobeats:
querystring is legacy API


ronag:
I don't see any advantage of this? It will just be slower...

anonrig:
You can find a benchmark that compares alternatives from https://github.com/anonrig/fast-querystring#benchmark

PR #1686, opened 2022-10-05
feat: add abort reason support

KhafraDev:
Implements https://github.com/whatwg/fetch/commit/c9039b97a502b1d5fea46b55b9166762c3479837

Allows aborting with custom errors. For example, when using `AbortSignal.timeout`, a TimeoutError DOMException will now be thrown instead of an AbortError.


mcollina (approved):
lgtm

PR #1687, opened 2022-10-06
feat: add all File WPTs

KhafraDev:
- Fixes bug when using boxed primitives, for example: `new File([new String('bits')], 'dummy')`
- Fixes bug where null/undefined values would try to be converted to a webidl dictionary, causing an error.

After this I don't think there's any reason not to make `File` a global in node?

+95 tests, no failures. All test files are from https://github.com/web-platform-tests/wpt/tree/master/FileAPI/file

p.s. sorry for all the PRs lately


mcollina:
> After this I don't think there's any reason not to make File a global in node?

No, let's do it.

KhafraDev:
if it is made a global, should [FileReader](https://w3c.github.io/FileAPI/#APIASynch) also be implemented?

mcollina:
I think we should implement that interface anyway.

jimmywarting:
I don't think the FileReader is necessary. We have read methods on the blob itself

I consider it to be kind of legacy now

KhafraDev:
I think adding it is good for the same reason that `btoa` and `atob` were added: browser (& deno) compatibility. It was insanely easy to implement too.

It also allows us to enable the idlharness for File (the alternative is disabling ~100 tests) which allowed me to find 2 minor bugs in the File implementation. It also prevents regressions in the future, which is great.

PR #1689, opened 2022-10-07
feat(WPTs): add in idlharness.any.js test

KhafraDev:
Fixes ~20 'bugs' in fetch. Note that every change was done for a reason and I could explain why if you'd like.

---

1 failed test. I could easily fix it but the test itself makes no sense (`Response.json.length` should be 0). This test also fails in chrome and Deno (only other environments that implement Response.json that I bothered checking).

However node now passes more of fetch's idlharness tests than any other platform, so that's cool.

### DIFF
--- a/lib/fetch/body.js
+++ b/lib/fetch/body.js
@@ -530,32 +530,8 @@ function bodyMixinMethods (instance) {
   return methods
 }
 
-const properties = {
-  body: {
-    enumerable: true,
-    get () {
-      if (!this || !this[kState]) {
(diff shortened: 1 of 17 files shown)


mcollina (approved):
lgtm

KhafraDev:
https://wpt.fyi/results/fetch/api/idlharness.any.html?label=master&label=experimental&product=chrome&product=edge&product=firefox&product=safari&product=deno&aligned&view=subtest

With this PR 142/143 pass in undici

PR #1690, opened 2022-10-07
feat: implement FileReader

KhafraDev:
Implements [FileReader](https://w3c.github.io/FileAPI/#APIASynch)

Benefits:
- Allowed me to enable the idlharness tests for the FileAPI section (Blob, URL, File, etc.) - found 2 issues in File, 6 in Blob, and 2 in URL. The alternative was going through and disabling over 100 tests.
- Deno and every browser implement this. The same arguments for adding btoa and atob could be made here as well.

Also see:
https://github.com/nodejs/undici/pull/1687#issuecomment-1272016124

The API, although outdated, is still perfectly usable in modern environments. Unlike XMLHttpRequest, does not come with any foot guns (synchronous requests).


mcollina (approved):
lgtm

index.js
```diff
@@ -98,9 +98,9 @@ if (nodeMajor > 16 || (nodeMajor === 16 && nodeMinor >= 8)) {
     if (!fetchImpl) {
       fetchImpl = require('./lib/fetch').fetch
     }
-    const dispatcher = (arguments[1] && arguments[1].dispatcher) || getGlobalDispatcher()
+
     try {
-      return await fetchImpl.apply(dispatcher, arguments)
+      return await fetchImpl(...arguments)
```
ronag, line 103: Why these changes?

ronag (commented):
Amazing work! 

@KhafraDev I see you don't have any public email address. Would you mind sending me an email?

ronag (approved):

targos (changes requested):
Please also update `index-fetch.js`

targos:
Not objecting to it landing here, but I feel like it would be nice to have it in Node.js core independently of `undici`

index.js
KhafraDev: I was experiencing a flaky test, I can revert if needed

KhafraDev:
It would require a lot of work to port it to node core because it uses parts from the webidl, mime, and encoding spec.

I'm also not sure if node's WPT runner can handle idlharness tests as it currently does not do so.

jimmywarting:
think PR should rather be made to WPT to change the test to using the blobs new read methods instead...
The FileReader should be a own isolated independent test.

jimmywarting:
The Blob class in NodeJS core is kind of B imo, specially when it comes to streaming the content of the blob. (see https://github.com/nodejs/node/issues/42264) + it also lacks all those webidl stuff. 

I think I also saw some changes/fixes you did to the Blob/File class that I think should rather be reflected onto the NodeJS Core Blob.... 

So I somewhat agree with @targos on this. either we have all the Blob stuff in NodeJS core, or everything is moved to from Core into Undici. But then we would have to move a bunch of things such as object URL and copy/postMessage stuff. 

I also wish to see some form of "create a file backed up by the filesystem" to be implemented by NodeJS also. That would require the blob to keep track of certain file metadata. This is `fetch-blob` only main reason why it's still needed/around and why I haven't deprecated it yet and said: just use NodeJS built in stuff.

I also can't extend the NodeJS core Blob in `fetch-blob` and create blobs backed up by the filesystem. wish is a bit sad b/c my `fetch-blob` can't be transfered over with PostMessage + it's not an instance of NodeJS core Blob, and they can't work along with ObjectURLs or Worker 😞 

As long as Blobs can only be constructed by some stuff in the memory, then they are pretty useless IMO. And you might just as well just use a ArrayBuffer/uint8array instead. Blob main purpose is kind of helping you to offload some of the data and have them being backed up by some file handle and offload some of the memory to a temporary location on the disc.

KhafraDev:
> it also lacks all those webidl stuff.

The problem with missing webidl is that it would be a pain to convert it to the node equivalent - it's simply easier to implement it in undici.

Regarding URL and Blob, I believe both of them interface with c++ at some point. It's not feasible to re-write both in js. Both File and FileReader are implemented in js (same with Deno).

---

If anything, File and FileReader could be implemented in node core directly, however, considering that this FileReader implementation passes 95% of the WPTs already, it will incur very little maintenance burden (1 test with event ordering and 1 due to differences between a microtask and a html task queue).

mcollina:
@KhafraDev this has quite a bit of conflicts.

mcollina (approved):
lgtm

package.json
```diff
@@ -116,7 +116,8 @@
     "compilerOptions": {
       "esModuleInterop": true,
       "lib": [
-        "esnext"
+        "esnext",
+        "DOM"
```
SimenB, line 120: this means the types of undici only works if dependent project also adds the dom types. That shouldn't be necessary for a module which explicitly _doesn't_ work in the browser

ronag: Whether it works in the browser or not is not the relevant part. The types are and fetch is based on DOM types per the spec. So it's not weird that we also need to import DOM types as well...

SimenB: this PR uses `DOMException` (and possibly more) from DOM types

>Whether it works in the browser or not is not the relevant part.

It's relevant for consumers that `undici` doesn't work in a typescript environment that isn't configured to be run in the browser. But I can agree it's more of a symptom than a cause 🙂

KhafraDev: unfortunately @types/node doesn't seem to include quite a bit. If you remove that line and run `npm run test:typescript` we get errors about missing DOMException and EventInit types.

SimenB: `DOMException` doesn't exist in Node, so that makes sense. `EventInit` seems to be an interface? Not something in MDN at least

SimenB: But just like this project comes with types for most of `fetch` (https://github.com/nodejs/undici/blob/main/types/fetch.d.ts) that are duplicated from TS core dom types, `DOMException` and `EventInit` should also be defined here

KhafraDev: DOMException exists in core since v17 (somewhere around there) globally, and [EventInit](https://dom.spec.whatwg.org/#dictdef-eventinit) is the name of the options passed to Event.

SimenB: Ah. I tried to run `typeof DOMException` and got undefined, but I see I wrote `typeof DOMEXception` 🙈 

Regardless, I'm sure definetely typed would welcome a PR adding `DOMException`, `EventInit` and any other missing interfaces. But they shouldn't be pulled from `dom`.

EDIT: I do see `EventInit` there already: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/488ef79fe14ea591166813a01ba7f34dfd955b68/types/node/dom-events.d.ts#L78

KhafraDev: EventInit doesn't seem like a problem regardless, we can create our own and it won't overlap with the existing type (although I'm not sure why I can't access it). If we could "feature detect" DOMException, and only use our own type when it doesn't exist, that would be good. That way, the types won't overlap when/if DOMException gets added to node types.

KhafraDev: I tried adding `/// <reference types="node/dom-events" />` but the type still couldn't be used. I'll play around a little when I get home.

KhafraDev: I opened a PR to fix this https://github.com/nodejs/undici/pull/1762

jimmywarting: I do think this is kind of a TypeScript issue: https://github.com/microsoft/TypeScript/issues/43972

Issue #1666, opened 2022-09-24
fetch: wpt test is failing


KhafraDev:
Signal gets aborted here: https://github.com/nodejs/undici/blob/9c3f34c97dfce7f7f6cffd04b7d4dab9921dc40d/lib/fetch/index.js#L126
and the promise gets rejected here: https://github.com/nodejs/undici/blob/9c3f34c97dfce7f7f6cffd04b7d4dab9921dc40d/lib/fetch/index.js#L304

```js
import { fetch, setGlobalOrigin } from 'undici'
import assert from 'assert'

setGlobalOrigin('http://localhost:3000')

const controller = new AbortController();
const signal = controller.signal;
controller.abort();

const log = [];

await Promise.all([
	fetch('../resources/data.json', { signal }).then(
		() => assert_unreached("Fetch must not resolve"),
		() => log.push('fetch-reject')
	),
	Promise.resolve().then(() => log.push('next-microtask'))
]);

assert.deepStrictEqual(log, ['fetch-reject', 'next-microtask']);
```

_Originally posted by @KhafraDev in https://github.com/nodejs/undici/issues/1664#issuecomment-1256836575_

KhafraDev:
This case also fails:

```js
const request = new Request('https://example.com', { method: 'POST', body: 'body' })
const originalBody = request.body
assert(request.bodyUsed === false)

const otherRequest = new Request(request, { body: 'another body' })
assert(request.bodyUsed === true)
assert(request.body === originalBody)
assert(originalBody !== undefined)
assert(originalBody !== null)
assert(otherRequest.body !== originalBody)
```

KhafraDev:
This case fails:

```js
const body = new ReadableStream()

assert.throws(() => {
  new Request('https://example.com', { method: 'POST', body })
}, TypeError)
```

KhafraDev:
```js
const initialBuffer = new Int8Array(new ArrayBuffer(16))

const stream = new ReadableStream({
  start (controller) {
    controller.enqueue(initialBuffer)
    controller.close()
  }
})
const [out1, out2] = stream.tee()

const { value } = await out2.getReader().read()

console.log(value === initialBuffer) // this should be false
```

The same subset of tests are failing here (https://staging.wpt.fyi/results/fetch/api/response/response-clone.any.html?label=master&label=experimental&aligned&view=subtest)

[...]

PR #1697, opened 2022-10-13
fix(fetch): use structuredClone in clone body steps

KhafraDev:
Refs: https://github.com/nodejs/undici/issues/1666#issuecomment-1275142274

Enables the following WPTs (only Safari currently passes these tests, and the same 2 tests fail here):
https://wpt.fyi/results/fetch/api/response/response-clone.any.html?label=master&label=experimental&aligned&view=subtest

---

The spec does tell implementations to clone the stream, but you have to go down a few layers:
1. Let « out1, out2 » be the result of [teeing](https://streams.spec.whatwg.org/#readablestream-tee) body’s [stream](https://fetch.spec.whatwg.org/#concept-body-stream).
2. To tee a [ReadableStream](https://streams.spec.whatwg.org/#readablestream) stream, return ? [ReadableStreamTee](https://streams.spec.whatwg.org/#readable-stream-tee)(stream, true).
3. ReadableStreamTee(stream, cloneForBranch2) will [tee](https://streams.spec.whatwg.org/#tee-a-readable-stream) a given readable stream.

3. The second argument, cloneForBranch2, governs whether or not the data from the original stream will be cloned (using HTML’s [serializable objects](https://html.spec.whatwg.org/multipage/structured-data.html#serializable-objects) framework) before appearing in the second of the returned branches.

### DIFF
--- a/lib/fetch/body.js
+++ b/lib/fetch/body.js
@@ -6,7 +6,7 @@ const { ReadableStreamFrom, toUSVString, isBlobLike } = require('./util')
 const { FormData } = require('./formdata')
 const { kState } = require('./symbols')
 const { webidl } = require('./webidl')
-const { DOMException } = require('./constants')
+const { DOMException, structuredClone } = require('./constants')
 const { Blob } = require('buffer')
 const { kBodyUsed } = require('../core/symbols')
 const assert = require('assert')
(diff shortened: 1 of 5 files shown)


ronag (approved):

mcollina (approved):
lgtm

PR #1702, opened 2022-10-14
feat: add minesniff WPTs

KhafraDev:
Undici implements a subset of the [mimesniff](https://mimesniff.spec.whatwg.org/) spec for data URLs.

Currently passes 909 tests and fails 989. This might seem pretty bad, but undici currently passes more of these tests than Chrome and Safari ([results](https://wpt.fyi/results/mimesniff/mime-types/parsing.any.html?label=master&label=experimental&aligned&view=subtest)). There's likely more we can do, but most of the failures are caused by node's mimetype parser (or whatever is used in Blob).


mcollina (approved):
lgtm

PR #1703, opened 2022-10-14
feat: make File's `type` more strict

KhafraDev:
This might be a bad change because, although the spec says to make `type` a "parseable MIME type", browsers have seemingly ignored this for years (see: https://github.com/nodejs/node/issues/45008#issuecomment-1279557343).

Not only this, it
1. causes more tests to fail (as Blob's attribute is still 'compliant' with browsers I guess?)
2. requires us to implement our own `type` getter, rather than using Blob's.
3. causes a FileAPI test to fail. I don't think it was meant to pass, as it is literally named "nonparseable" though.

With this change, if you remove the Blob tests from the parsing.any.js test, ~400 more tests succeed. This is because the `File` and `Blob` tests are grouped together and failing an assert causes the File test not to run. 😞

### DIFF
--- a/lib/fetch/file.js
+++ b/lib/fetch/file.js
@@ -5,6 +5,7 @@ const { types } = require('util')
 const { kState } = require('./symbols')
 const { isBlobLike } = require('./util')
 const { webidl } = require('./webidl')
+const { parseMIMEType, serializeAMimeType } = require('./dataURL')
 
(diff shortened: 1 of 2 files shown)


ronag (approved):

mcollina (approved):
LGTM

Issue #1680, opened 2022-10-03
bug: header value checks are too strict in fetch


KhafraDev:
## Bug Description

https://github.com/nodejs/undici/blob/0964a83710467b994cc108d096e97f69dd54ac90/lib/core/request.js#L26

The regex here is too strict for `fetch` which causes a bunch of tests to fail:
```
{ fail: '\x01' }
{ fail: '\x02' }
{ fail: '\x03' }
{ fail: '\x04' }
{ fail: '\x05' }
{ fail: '\x06' }
{ fail: '\x07' }
{ fail: '\b' }
{ fail: '\x0E' }
{ fail: '\x0F' }
{ fail: '\x10' }
{ fail: '\x11' }
{ fail: '\x12' }
{ fail: '\x13' }
{ fail: '\x14' }
{ fail: '\x15' }
{ fail: '\x16' }
{ fail: '\x17' }
{ fail: '\x18' }
{ fail: '\x19' }
{ fail: '\x1A' }
{ fail: '\x1B' }
{ fail: '\x1C' }
{ fail: '\x1D' }
{ fail: '\x1E' }
{ fail: '\x1F' }
```

## Reproducible By

https://gist.github.com/KhafraDev/cc178abba0a89580008f44c7ebee76a3

Sorry for the messy code it was taken almost directly from wpt (https://github.com/web-platform-tests/wpt/blob/master/fetch/api/headers/header-values-normalize.any.js)

## Expected Behavior

The tests should pass.

### Additional context

These checks were added in a29a151d to fix an issue where headers weren't being sanitized.

cc @mcollina

mcollina:
My understanding is that the use case of fetch in the browser vs in the server is different and it might be that those chars are _safe_ to render on the wire from browsers but not from servers. However, that regexp might be too strict.

If you'd like to loose it a bit, it's fine as long as the tests pass.

KhafraDev:
The server is responding with the headers sent from fetch so there is an expectation that both the server and fetch can receive these headers. I will get around to fixing this, but wasn't entirely sure. 👍

PR #1712, opened 2022-10-17
wpt: add remaining fetch/api tests

KhafraDev:
These were mostly tests that had a large amount of failures or required a non-zero amount of effort to implement.

1036 -> 1281 tests; 123 more failures (50% of tests added fail).

Most failures have to deal with not having FileReader/XMLHttpRequest.
- 26 tests fail because of #1680 
- 36 tests fail because undici doesn't filter headers
- 15 tests fail because undici handles `redirect: 'manual'` differently than the spec


mcollina (approved):
lgtm

mcollina:
It might be good to have a summary page generated automatically with all those stats.

KhafraDev:
it should be possible to do so

PR #1715, opened 2022-10-18
fix(fetch): remove terminated event listener on redirect

### LINKED ISSUE #1711: redirecting >11 times causes "MaxListenersExceededWarning"
## Bug Description

Each redirect adds one `terminated` listener to fetchParams and it's never removed.

## Reproducible By

```mjs
import { once } from 'events'

let redirects = 0

  if (redirects === 15) {
    res.end('Okay goodbye')
    return
  }

  res.writeHead(302, {
    Location: `/${redirects++}`
  })
  res.end()
}).listen(0)

const url = `http://localhost:${server.address().port}`

fetch(url, { redirect: 'follow' })
```

## Logs & Screenshots

```
MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 terminated listeners added to [Fetch]. Use emitter.setMaxListeners() to increase limit
```

KhafraDev:
Disclaimer: this is a pretty terrible solution, but I couldn't find another way of fixing this.

Fixes #1711 

- The listener gets added here
https://github.com/nodejs/undici/blob/7102eb0b7f544cee7669aaf06eb94f38b22aba86/lib/fetch/index.js#L1813
- This is where it should be removed (& onError). If we remove it here it doesn't work, my best guess is that it gets gc'ed? I can't move `onAborted` outside of this scope as it relies on `response`, and if we use a closure (etc.) the same problem would happen.
https://github.com/nodejs/undici/blob/7102eb0b7f544cee7669aaf06eb94f38b22aba86/lib/fetch/index.js#L2028

If you have any idea of how to fix this, please let me know. I spent maybe 45 minutes trying anything I could think of.

Also not sure how I would add tests for this.

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -73,6 +73,11 @@ class Fetch extends EE {
     this.connection = null
     this.dump = false
     this.state = 'ongoing'
+    // 2 terminated listeners get added per request,
+    // but only 1 gets removed. If there are 20 redirects,
+    // 21 listeners will be added.
+    // See https://github.com/nodejs/undici/issues/1711
+    this.setMaxListeners(21)
   }
 
   terminate (reason) {
@@ -1068,13 +1073,6 @@ async function httpFetch (fetchParams) {
       // See https://github.com/nodejs/undici/issues/1193.
       response = actualResponse
     } else if (request.redirect === 'follow') {
-      /** @type {Fetch} */
-      const controller = fetchParams.controller
-
-      if (controller.listenerCount('terminated') > 0) {
-        controller.removeAllListeners('terminated')
-      }
-
       // Set response to the result of running HTTP-redirect fetch given
       // fetchParams and response.
       response = await httpRedirectFetch(fetchParams, response)
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -77,6 +77,7 @@ class Fetch extends EE {
     // but only 1 gets removed. If there are 20 redirects,
     // 21 listeners will be added.
     // See https://github.com/nodejs/undici/issues/1711
+    // TODO (fix): Find and fix root cause for leaked listener.
     this.setMaxListeners(21)
   }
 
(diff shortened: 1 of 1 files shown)


mcollina:
Could you add a unit test?

mcollina:
The code looks good. I would remove the individual listener and not all.

ronag:
For now... I think it's easier just to increase the listener limit, i.e. `setMaxListeners`.

lib/fetch/index.js
```diff
@@ -73,6 +73,11 @@ class Fetch extends EE {
     this.connection = null
     this.dump = false
     this.state = 'ongoing'
+    // 2 terminated listeners get added per request,
+    // but only 1 gets removed. If there are 20 redirects,
+    // 21 listeners will be added.
+    // See https://github.com/nodejs/undici/issues/1711
+    this.setMaxListeners(21)
```
ronag, line 80:
```suggestion
    // TODO (fix): Find and fix root cause for leaked listener.
    this.setMaxListeners(21)
```

jonseymour:
Did this PR actually fix anything or did it just mask the problem so that an annoying message got printed less frequently?

KhafraDev:
Masked the issue.

KhafraDev:
I'll have a PR that actually fixes it, wasn't that complicated.

jonseymour:
Could I clarify, Is there is an actual issue with fetch listeners being added but not removed in the presence of redirect responses?

Uzlopak:
its not like you have a memory leak.

jonseymour:
ok, so the emission of the messages was simply caused by bursty acquisition of listeners prior to GC occurring?

Uzlopak:
Basically. An EventEmitter could have hundreds of listeners for a single event without being an issue. But the case is very rare, so the node core devs decided to warn the devs if they add more than 10 listeners to an event without maybe actively noticing it, as it indicates a bug, which could slow down the performance, without being that obvious while you are developing.

PR #1717, opened 2022-10-19
feat: implement `8.2 Set request’s referrer policy on redirect`

metcoder95:


KhafraDev (changes requested):

lib/fetch/util.js
```diff
@@ -200,12 +212,24 @@ function setRequestReferrerPolicyOnRedirect (request, actualResponse) {
...
-  // TODO:  https://w3c.github.io/webappsec-referrer-policy/#parse-referrer-policy-from-header
-  const policy = ''
+
+  // 8.1 Parse a referrer policy from a Referrer-Policy header
+  // 1. Let policy-tokens be the result of extracting header list values given `Referrer-Policy` and response’s header list.
+  const { headersList } = actualResponse
+  // 2. Let policy be the empty string.
+  // 3. For each token in policy-tokens, if token is a referrer policy and token is not the empty string, then set policy to token.
+  // 4. Return policy.
+  const token = headersList.get('referrer-policy') ?? ''
```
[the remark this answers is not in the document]
metcoder95: Sure, done in [0b4c8e1](https://github.com/nodejs/undici/pull/1717/commits/0b4c8e1c8452366673e2943d64ad97c74ab9414c)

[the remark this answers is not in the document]
KhafraDev: It should pick the first valid one, this is similar to how the `integrity` option on RequestInit works.

There has to be some validation that the first referrer policy is actually a referrer policy. Otherwise a referrer policy like `invalid, origin` wouldn't work.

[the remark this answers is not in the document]
metcoder95: Addressed and added tests in [683ef3a](https://github.com/nodejs/undici/pull/1717/commits/683ef3aaa32577d43d82a6e8f27c9afa58c94d1e)

lib/fetch/util.js
```diff
@@ -200,12 +212,24 @@ function setRequestReferrerPolicyOnRedirect (request, actualResponse) {
...
-  if (policy !== '') {
-    request.referrerPolicy = policy
+  if (token !== '') {
+    for (const policyToken of referrerPolicyTokens) {
+      // if token is a referrer policy and token is not an empty string, then set policy to token.
+      if (token === policyToken) {
+        request.referrerPolicy = token
+        break
+      }
+    }
```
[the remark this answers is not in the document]
metcoder95: Addressed in: [683ef3a](https://github.com/nodejs/undici/pull/1717/commits/683ef3aaa32577d43d82a6e8f27c9afa58c94d1e)

mcollina (approved):
lgtm

lib/fetch/util.js
```diff
@@ -200,8 +190,31 @@ function setRequestReferrerPolicyOnRedirect (request, actualResponse) {
...
+
+  // Note: As the referrer-policy can contain multiple policies
+  // separated by comma, we need to loop through all of them
+  // and pick the first valid one.
+  // Ref: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy#specify_a_fallback_policy
+  let policy = ''
+  if (policyHeader.length > 0) {
+    // The right-most policy takes precedence.
+    // The left-most policy is the fallback.
+    for (let i = policyHeader.length; i !== 0; i--) {
```
KhafraDev, line 210: this seems to pick the last valid one?
ie. `same-origin, origin` will pick origin, when it should pick same-origin

lib/fetch/util.js
```diff
@@ -200,8 +190,31 @@ function setRequestReferrerPolicyOnRedirect (request, actualResponse) {
...
+      const token = policyHeader[i - 1].trim()
```
KhafraDev, line 211: does this match how the browser handles it?

KhafraDev (changes requested):

lib/fetch/util.js
KhafraDev: with the integrity option, the first valid one is always picked (assuming the algorithm "weight" is equal) so for example
`integrity: sha512-abc, sha512-def` will always choose sha512-abc assuming abc is a valid hash.

metcoder95: >ie. same-origin, origin will pick origin, when it should pick same-origin

Yes the right-most

>with the integrity option, the first valid one is always picked (assuming the algorithm "weight" is equal) so for example
integrity: sha512-abc, sha512-def will always choose sha512-abc assuming abc is a valid hash.

I based the implementation on the MDN docs (see [ref](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy#specify_a_fallback_policy)) which states:

>>`Referrer-Policy: no-referrer, strict-origin-when-cross-origin`
>>
>>In the above scenario, `no-referrer` is used only if the browser does not support the `strict-origin-when-cross-origin` policy.

Now accordingly to the spec Unknown Policy Values (ref: https://w3c.github.io/webappsec-referrer-policy/#unknown-policy-values), and following the _Example 18_:

>>Suppose older user agents don’t understand the ["unsafe-url"](https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-unsafe-url) policy. A site can specify an ["origin"](https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-origin) policy followed by an ["unsafe-url"](https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-unsafe-url) policy: older user agents will ignore the unknown ["unsafe-url"](https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-unsafe-url) value and use ["origin"](https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-origin), while newer user agents will use ["unsafe-url"](https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-unsafe-url) because it is the last to be processed.

The preferred policy should be appended at the very last of the list (right-most) while the fallbacks should be enlisted in order of priority (from right-to-left) being the left-most the final fallback to be chosen. This as the response on redirect can send the `Referrer Policy` header either independently or in a comma-separated list. Accordingly to _Example 19_:

>>To specify multiple policy values in the Referrer-Policy header, a site can send multiple Referrer-Policy headers:
`Referrer-Policy: no-referrer`
`Referrer-Policy: unsafe-url`
or, equivalently, multiple comma-separated header values:
>>
`Referrer-Policy: no-referrer,unsafe-url`

Where again, the right-most takes prescedence

lib/fetch/util.js
metcoder95: Didn't test it, but I can do it if you think is needed 🙂

lib/fetch/util.js
KhafraDev: In example 18, `unsafe-url` isn't a supported policy so it is ignored. If it was known, it would be picked.

In 19 I don't see where the right most would be picked first, unless no-referrer wasn't a valid policy.

metcoder95: The 19th was mostly as a reference and its theory was keeping the same logic of `unsafe-url` not being supported, though reading twice I think it doesn't 🤔 

For the 18th, after a third read, it is true that doesn't seem really clear whereas the order within the list, as this sentence makes it harder:

>older user agents will ignore the unknown ["unsafe-url"](https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-unsafe-url) value and use ["origin"](https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-origin), while newer user agents will use ["unsafe-url"](https://w3c.github.io/webappsec-referrer-policy/#referrer-policy-unsafe-url) because it is the last to be processed.

Which I understand as: `origin, unsafe-url`, so still quite ambiguous.
Let me move forward with testing how Chrome does it and change it accordingly 👍

metcoder95: After doing the test on Chrome, it seems the first assumption was correct, it will take precedence over the right-most token on the `Referrer-Policy` header, and fallback to the `left-most` progressively.

I tested two scenarios with the same token policies, `no-referrer` and `origin` coming as responses from the Response on Redirect.

### Redirect response with `Referrer-Policy: origin,no-referrer`:

**Redirect Request/Response**
image

**Final Request/Response**
image

**Result**: `no-referrer` took precede as being the left-most option (first to be processed)


### Redirect response with `Referrer-Policy: no-referrer,origin`:

**Redirect Request/Response**
image


**Final Request/Response**
image


**Result**: `origin` took precede as being the left-most option (first to be processed)

Based on Chrome, I think is possible to move forward with the current behavior, or wdyt? 🙂

KhafraDev: In firefox it is the left most, https://github.com/mozilla/gecko-dev/blob/76ccfc801e6b736c844cde3fddeab7a748fc8515/dom/security/ReferrerInfo.cpp#L165

Having it the right most goes against every other part of the spec and just generally doesn't make sense

KhafraDev: https://source.chromium.org/chromium/chromium/src/+/main:net/url_request/redirect_info.cc;l=50;bpv=0;bpt=1
https://github.com/mozilla/gecko-dev/blob/76ccfc801e6b736c844cde3fddeab7a748fc8515/dom/security/ReferrerInfo.cpp#L165

You seem to be correct; it chooses the last valid policy.

Specifically, the spec says `the value of the latest one will be used`.

lib/fetch/util.js
KhafraDev: this is fine, chrome trims it, firefox doesn't. doesn't seem to matter

KhafraDev (commented):
Could you add some tests that use fetch directly?

metcoder95:
> Could you add some tests that use fetch directly?

Sure, I'll take care of it Tomorrow

metcoder95:
When working on adding the testing over `fetch` of the `8.2` I noticed that I was always getting the `no-referrer` Referrer's value, even on scenarios where this should not happen (e.g. referrer is a valid same-origin URL, and the policy is `origin`). 

Then, I was able to notice that the implementation of the spec [`Determine Request Referrer`](https://w3c.github.io/webappsec-referrer-policy/#determine-requests-referrer) is making a wrong assumption around the `Request#referrer`, as by spec the `Request#referrer` getter is set to return a serialized URL in the form of a `string`([ref](https://fetch.spec.whatwg.org/#dom-request-referrer)) but the implementation is assuming the `Request#referrer` is an instance of URL class ([ref](https://github.com/nodejs/undici/blob/683ef3aaa32577d43d82a6e8f27c9afa58c94d1e/lib/fetch/util.js#L394)).

I'll prepare a fix on a separated PR and wait till it is approved and merge to continue with the testing of this one, just to keep the scope clear about the boundaries of each change. wdyt @KhafraDev?

KhafraDev (approved):

KhafraDev:
> spec [Determine Request Referrer](https://w3c.github.io/webappsec-referrer-policy/#determine-requests-referrer) is making a wrong assumption

The spec is correct.

> as by spec the Request#referrer getter

The spec isn't referring to a Request object, but an internal [request](https://fetch.spec.whatwg.org/#requests) object, where the [referrer](https://fetch.spec.whatwg.org/#concept-request-referrer) is one of "'no-referrer', 'client', or a [URL](https://url.spec.whatwg.org/#concept-url)."

The issue here is either `referrer` isn't being set somewhere and is defaulted to client in [`makeRequest`](https://github.com/nodejs/undici/blob/7827118d494e189819af552f33e1a9da30ab6620/lib/fetch/request.js#L781) or it is being set to `client` somewhere else incorrectly.

metcoder95:
>The spec isn't referring to a Request object, but an internal [request](https://fetch.spec.whatwg.org/#requests) object, where the [referrer](https://fetch.spec.whatwg.org/#concept-request-referrer) is one of "'no-referrer', 'client', or a [URL](https://url.spec.whatwg.org/#concept-url)."

The issue here is either referrer isn't being set somewhere and is defaulted to client in [makeRequest](https://github.com/nodejs/undici/blob/7827118d494e189819af552f33e1a9da30ab6620/lib/fetch/request.js#L781) or it is being set to client somewhere else incorrectly.

You're right, the spec implementation works as expected. I might touched something by mistake that broke the tests I was doing. Sorry for that 🙇  
Though, I'll add couple of more testing for `determineRequestReferrer` and `setRequestReferrerPolicyOnRedirect` to set proper baselines. Thanks for the review and the chat!

KhafraDev:
There are some WPTs that can be added to ensure referrerPolicy works as intended:
https://github.com/web-platform-tests/wpt/blob/master/fetch/api/basic/request-referrer.any.js
https://github.com/web-platform-tests/wpt/blob/master/fetch/api/redirect/redirect-referrer-override.any.js
https://github.com/web-platform-tests/wpt/blob/master/fetch/api/redirect/redirect-referrer.any.js

Don't worry about it, there's a very strong possibility that this hasn't been working as expected for some time now, and just hasn't been noticeable without your work.

metcoder95:
Thanks! I'll look at those and follow your work to add them. Looking forward 🙂

PR #1719, opened 2022-10-19
wpt: add two tests

KhafraDev:
This is probably it


ronag:
CI fails?

KhafraDev:
flaky test

KhafraDev:
there actually seems to be some flakiness with windows & the WPT workers. I guess limiting the number of ones that can be open would fix it?

PR #1721, opened 2022-10-21
fetch: implement new spec changes

KhafraDev:
Implements
1. https://github.com/whatwg/fetch/commit/813788a7d7b4bfba9d4398efd22fdae13125ee8a
2. https://github.com/whatwg/fetch/commit/4cd70cf5aecf777a9f6d451c6381423153913da9
3. https://github.com/whatwg/fetch/commit/4a7bf3536c4bd4cd30aedba5d0a2ba865c587699

### DIFF
--- a/lib/fetch/body.js
+++ b/lib/fetch/body.js
@@ -19,33 +19,45 @@ const { parseMIMEType, serializeAMimeType } = require('./dataURL')
 /** @type {globalThis['ReadableStream']} */
 let ReadableStream
 
-async function * blobGen (blob) {
-  yield * blob.stream()
-}
-
 // https://fetch.spec.whatwg.org/#concept-bodyinit-extract
 function extractBody (object, keepalive = false) {
   if (!ReadableStream) {
     ReadableStream = require('stream/web').ReadableStream
   }
 
-  // 1. Let stream be object if object is a ReadableStream object.
-  // Otherwise, let stream be a new ReadableStream, and set up stream.
-  let stream = null
+  // 1. Let stream be null.
+  let stream
 
-  // 2. Let action be null.
+  // 2. If object is a ReadableStream object, then set stream to object.
+  if (object instanceof ReadableStream) {
+    stream = object
+  } else if (isBlobLike(object)) {
+    // 3. Otherwise, if object is a Blob object, set stream to the
+    //    result of running object’s get stream.
+    stream = object.stream()
+  }
+
+  // 4. Otherwise, set stream to a new ReadableStream object, and set
+  //    up stream.
+
+  // 5. Assert: stream is a ReadableStream object.
+
+  // TODO(@KhafraDev): setting stream to an empty stream here causes
+  // the process to unexpectedly exit. An error is likely being swallowed.
+
+  // 6. Let action be null.
   let action = null
 
-  // 3. Let source be null.
+  // 7. Let source be null.
   let source = null
 
-  // 4. Let length be null.
+  // 8. Let length be null.
   let length = null
 
-  // 5. Let Content-Type be null.
-  let contentType = null
+  // 9. Let type be null.
+  let type = null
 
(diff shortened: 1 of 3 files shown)


lib/fetch/index.js
```diff
@@ -850,13 +858,13 @@ async function schemeFetch (fetchParams) {
...
       //    header list is « (`Content-Type`, mimeType) »,
-      //    and body is dataURLStruct’s body.
+      //    and body is dataURLStruct’s body as a body.
       return makeResponse({
         statusText: 'OK',
         headersList: [
           ['content-type', mimeType]
         ],
-        body: extractBody(dataURLStruct.body)[0]
+        body: safelyExtractBody(dataURLStruct.body)[0]
```
ronag, line 867: Why safely?

ronag (approved):

lib/fetch/body.js
```diff
@@ -19,33 +19,45 @@ const { parseMIMEType, serializeAMimeType } = require('./dataURL')
...
+  } else if (isBlobLike(object)) {
+    // 3. Otherwise, if object is a Blob object, set stream to the
+    //    result of running object’s get stream.
+    stream = object.stream()
+  }
+
+  // 4. Otherwise, set stream to a new ReadableStream object, and set
+  //    up stream.
+
+  // 5. Assert: stream is a ReadableStream object.
```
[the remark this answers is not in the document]
KhafraDev: See my note below:
```
// TODO(@KhafraDev): setting stream to an empty stream here causes
  // the process to unexpectedly exit. An error is likely being swallowed.
```
for some reason tap starts failing when I implement 4 & 5. No idea why 😕

lib/fetch/index.js
KhafraDev: https://fetch.spec.whatwg.org/#byte-sequence-as-a-body

"To get a [byte sequence](https://infra.spec.whatwg.org/#byte-sequence) bytes as a body, return the [body](https://fetch.spec.whatwg.org/#body-with-type-body) of the result of [safely extracting](https://fetch.spec.whatwg.org/#bodyinit-safely-extract) bytes."

lib/fetch/body.js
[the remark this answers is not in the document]
KhafraDev: Seems like a node core bug...

```js
const readable = new ReadableStream({
  start () {},
  type: undefined
})

while (true) {
  const { done, value } = await reader.read()

  console.log({ done, value })

  if (done) {
    break
  }
}

console.log('hi')
```

'hi' is never logged.

lib/fetch/body.js
```diff
@@ -19,33 +19,65 @@ const { parseMIMEType, serializeAMimeType } = require('./dataURL')
...
+    stream = object.stream()
+  } else {
+    // 4. Otherwise, set stream to a new ReadableStream object, and set
+    //    up stream.
+    stream = new ReadableStream({
+      async pull (controller) {
+        controller.enqueue(
+          typeof source === 'string' ? new TextEncoder().encode(source) : source
+        )
+        queueMicrotask(() => {
```
ronag, line 46: Yes, this is a known issue with node readable streams. I think we have the same queueMicrotask hack somewhere else.

ronag: Maybe we could move this into a `closeReadableStream` helper function and re-use? Then we can link additional information why that is needed in one place.

KhafraDev: Yeah I see it somewhere else, I actually copied it from there lol

Do you know why it happens?

ronag: I don't remember. Maybe @jasnell has an idea. I believe there is some issue over at the node repo.

KhafraDev: added the helper

ronag (approved):

PR #1722, opened 2022-10-21
wpt: add FormData wpts

KhafraDev:
Fixes 2 issues in FormData:
- `new FormData(null)` should throw
- have `File` entries inherit the lastModified property

These tests only take ~330 ms to run.


mcollina (approved):
lgtm

Issue #1691, opened 2022-10-08
`Error: Query params cannot be passed when url already contains "?" or "#".` at redirection


nandenjin:
## Bug Description

When use `options.query`, `options.maxRedirections` and the target server returns redirection URL contains `?` like `https://example.com?param=here`, undici raises `Error: Query params cannot be passed when url already contains "?" or "#".` to redirected URL.

It seems that undici tries to add query parameters for URLs after redirections. (To be clear, I think it should not - it is not intended behavior for undici)

## Reproducible By

I found this issue when I tried to send request to WebApp built with Google Apps Script, so I created an endpoint just for testing. It redirects you once then shows query params you passed first.

https://script.google.com/macros/s/AKfycbzf56MFzf5Mdvop2eldVkhVf-vBvazOA86Noc0R7Km6vUgYUd0Ufm7_zIKd0yx2VpO1mA/exec

And you can repoduce the issue by:

```javascript
const { request } = require('undici');

request(
  `https://script.google.com/macros/s/AKfycbzf56MFzf5Mdvop2eldVkhVf-vBvazOA86Noc0R7Km6vUgYUd0Ufm7_zIKd0yx2VpO1mA/exec`,
  {
    maxRedirectioins: 1,
    query: { some_params: 'here' }
  }
)
```

## Expected Behavior

No error raised up. `options.query` works only for the first URL, not for URLs as redirected target.

## Logs & Screenshots

Logs for sample code to reproduce (given above):

```
nandenjin@nandenjin-hp:~/workspace$ node sample.js
/home/nandenjin/workspace/node_modules/undici/lib/core/util.js:33
    throw new Error('Query params cannot be passed when url already contains "?" or "#".')
          ^

Error: Query params cannot be passed when url already contains "?" or "#".
    at Object.buildURL (/home/nandenjin/workspace/node_modules/undici/lib/core/util.js:33:11)
    at new Request (/home/nandenjin/workspace/node_modules/undici/lib/core/request.js:132:30)
    at Client.[dispatch] (/home/nandenjin/workspace/node_modules/undici/lib/client.js:263:21)
    at Intercept (/home/nandenjin/workspace/node_modules/undici/lib/interceptor/redirectInterceptor.js:11:16)
    at Client.[Intercepted Dispatch] (/home/nandenjin/workspace/node_modules/undici/lib/dispatcher-base.js:157:12)
[...]

PR #1724, opened 2022-10-23
fix(redirect): Ignore query config after redirection

IlyasShabi:
Ignore query option on redirection, please check https://github.com/nodejs/undici/issues/1691 for more information.
The issue can be reproduced with a simple node/express server:

```javascript
app.get('/', (req, res) => { res.send('Hello world');});

app.get('/endpoint', (req, res) => { res.redirect('/redirection?param1=first');});

app.get('/redirection', (req, res) => { res.redirect('/');});

app.listen(4000, () => {})
```

Then we can call it with undici:

```javascript
undici.request(
'http://localhost:4000/endpoint',
{
   maxRedirection: 4,
   query: { param: 'test' }
}
).then(res => console.log('Good')).catch(err => console.log(err));
```
The code snippet will produce an error as described in the issue.

Solution: Remove query from options after redirection

### DIFF
--- a/lib/handler/RedirectHandler.js
+++ b/lib/handler/RedirectHandler.js
@@ -109,6 +109,7 @@ class RedirectHandler {
     this.opts.path = path
     this.opts.origin = origin
     this.opts.maxRedirections = 0
+    this.opts.query = null
 
     // https://tools.ietf.org/html/rfc7231#section-6.4.4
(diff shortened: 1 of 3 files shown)


IlyasShabi:
cc @mcollina

ronag (approved):

mcollina (approved):
Lgtm

Issue #1692, opened 2022-10-09
MAX_RESPONSE_BODY_SIZE is missing


nosovk:
## This would solve...

Sometimes we have to fetch user provided URL. In that case some users can provide a link to something big, like 4k video.
In that cases we would like to limit body size accepted by fetch.
In `node-fetch` there was option [`size`: maximum response body size in bytes. 0 to disable](https://github.com/node-fetch/node-fetch/blob/main/README.md#options). It would be nice to have a similar option in `undici`, because fetching user provided URLs without limitations seems to be a bad idea.

## The implementation should look like...

Add additional parameter to options, implementation in `node-fetch` seems to be clear to understand.

## I have also considered...

We can go to more low level implementation, but having option in fetch implementations seems to be a good idea.

KhafraDev:
This would be better suited as an Agent option, as a counterpart to maxHeaderSize & bodyTimeout.

Fetch is a [web spec](https://fetch.spec.whatwg.org/) which means we should fundamentally keep modifications to it at a minimum.

mcollina:
I'm +1 to add it to `Client`. Would you like to send a PR?

nosovk:
Unfortunately won't fix now. This issue blocks migration from `node-fetch` to `undici`, but till node-fetch is not deprecated it could be postponed.

jodevsa:
Hi @mcollina, can this issue be closed now?

PR #1729, opened 2022-10-25
feat(Client): add max response size option

mateonunez:
This PR implements #1692 by introducing a new client option: `maxResponseSize`.

The `maxResponseSize` parameter is set to `0` which means whatever size is allowed, otherwise, the size will be considered in bytes.

There is a new check before the _Parser_ computes the `bytesRead`. If that computation is greater than the `maxResponseSize` the socket is destroyed by the corresponding error.

```js
    if (maxResponseSize !== 0 && this.bytesRead + buf.length > maxResponseSize) {
      util.destroy(socket, new ResponseExceededMaxSizeError())
      return -1
    }
```

This approach enforces the `Client` to verify large responses and to close the socket before the buffer is completed.

This implementation surf in a low-level way, so tests (such as _ResponseContentLengthMismatchError_) are covered by `scenarios` in `test/errors.js`


mcollina (changes requested):

KhafraDev:
Should 0 be the default - is there ever a possibility that someone would want to ensure a response has an empty body? If so, -1 would be a better default for unlimited body size.

mateonunez:
> Should 0 be the default - is there ever a possibility that someone would want to ensure a response has an empty body? If so, -1 would be a better default for unlimited body size.

I fully agree with you, I changed the default value to **-1** and added a test to check even an empty payload.

test/max-response-size.js
```diff
@@ -0,0 +1,103 @@
...
+        })
+      })
+    })
+  })
+
+  t.test('invalid max response size should throw an error', (t) => {
+    t.plan(2)
+
+    t.throws(() => {
+      Client('http://localhost:3000', { maxResponseSize: 'hello' })
```
KhafraDev, line 97:
```suggestion
      new Client('http://localhost:3000', { maxResponseSize: 'hello' })
```

test/max-response-size.js
```diff
@@ -0,0 +1,103 @@
...
+    }, 'maxResponseSize must be a number')
+    t.throws(() => {
+      Client('http://localhost:3000', { maxResponseSize: -2 })
```
KhafraDev, line 100:
```suggestion
      new Client('http://localhost:3000', { maxResponseSize: -2 })
```

KhafraDev (changes requested):

lib/client.js
```diff
@@ -209,6 +216,7 @@ class Client extends DispatcherBase {
     this[kMaxRedirections] = maxRedirections
     this[kMaxRequests] = maxRequestsPerClient
     this[kClosedResolve] = null
+    this[kMaxResponseSize] = maxResponseSize || -1
```
ronag, line 219: This won't allow 0

ronag (changes requested):

test/max-response-size.js
mateonunez: In this way I got a `standard` error:

```bash
❯ npm run lint

> undici@5.11.0 lint
> standard | snazzy

standard: Use JavaScript Standard Style (https://standardjs.com)

/Users/mateonunez/dev/undici/test/max-response-size.js
   97:7  error  Do not use 'new' for side effects  no-new
  100:7  error  Do not use 'new' for side effects  no-new

✖ 2 problems
```

There is a way to avoid it without disabling the lint rule?

ronag: Assign the object to a variable or use eslint-disable-line

mateonunez:
Thanks for the review @ronag and @KhafraDev 🙏. I made the requested changes, disabling the lint rule and fixing the `maxResponseSize` assignment.

nosovk:
@mateonunez can you please add `maxResponseSize` to fetch options also, to close https://github.com/nodejs/undici/issues/1692?

KhafraDev:
That should not be added to fetch options

weijarz:
It appears to only limit the size before compression and has no effect on the gziped response. In other words, this cannot be used to prevent ZipBomb attacks.

metcoder95:
I'd suggest to add that as part of an interceptor

Issue #951, opened 2021-08-13
Fetch Test Coverage


ronag:
Current test coverage on fetch is too low for a stable release.

Please refer [lib/fetch](https://codecov.io/gh/nodejs/undici/tree/main/lib/fetch) folder on CodeCov to check which components of fetch implementation lack unit tests. Comment on this issue on which component you would like to add test and post a PR!

Please refer [CONTRIBUTING.md](https://github.com/nodejs/undici/blob/main/CONTRIBUTING.md) on how to run tests.

teezzan:
@ronag, I would love to work on this. However, I am a little new to this. Can you give me more information about the issue and a little bit of direction?

ronag:
Run ’npm run coverage’ see what parts of the code lacks test and make one in the test folder.

trivikr:
> Run ’npm run coverage’ see what parts of the code lacks test and make one in the test folder.

I landed on this issue after reading about lack of test coverage in https://github.com/nodejs/node/pull/41749#issuecomment-1024970687

Assuming this is a tracking issue to improve test coverage, would it be helpful to provide detailed instructions on how folks can help in the issue description? cc @ronag 

Also, it would be helpful if a link to this issue is provided in README at https://github.com/nodejs/undici#undicifetchinput-init-promise

mcollina:
Could you propose some text?

trivikr:
> Could you propose some text?

How about this for issue description?

---

Please refer [lib/fetch](https://codecov.io/gh/nodejs/undici/tree/main/lib/fetch) folder on CodeCov to check which components of fetch implementation lack unit tests. Comment on this issue on which component you would like to add test and post a PR!

Please refer [CONTRIBUTING.md](https://github.com/nodejs/undici/blob/main/CONTRIBUTING.md) on how to run tests.

---

I've posted the following PRs to improve discoverability/documentation:
* https://github.com/nodejs/undici/pull/1191
* https://github.com/nodejs/undici/pull/1192

mcollina:
Done!

RaisinTen:
Sent https://github.com/nodejs/undici/pull/1206 to complete the coverage of the `File` class in its current state.

Artrix9095:
any updates on this becoming stable?

ronag:
@KhafraDev, any idea on the coverage stats from wpt?

KhafraDev:
I couldn't get any tools to work with it, but it's likely pretty high with the current test coverage.

[...]

PR #1734, opened 2022-10-26
fix: add wpts to code coverage

KhafraDev:
Refs: #951 

I have no idea if this works.


KhafraDev:
Looks like I can't test this until the CI works for node 19 😕 
https://app.codecov.io/gh/nodejs/undici/pull/1734

```
You will see an error stating No coverage uploaded for pull request head. when the pull request has:
    ...
    or the uploaded reports are from failed CI builds.

```

ronag:
Try removing 19 from the test matrix

KhafraDev:
well I guess it doesn't work

KhafraDev:
honestly I don't know

KhafraDev:
Here's the predicament: 
- nyc doesn't support esm (which is why it shows no coverage for WPTs)
- tap doesn't support c8, only nyc

So the only option is to replace esm in the WPT runner with cjs.

PR #1736, opened 2022-10-27
wpt: fail runner if expected failures don't fail

KhafraDev:
The runner will now fail if a test that is marked as an expected failure doesn't fail.

The way I did this is either ingenious or horrible, depending on how you look at it. It maps tests that failed to an object and then compares that object to a slightly modified version of the status file (without including other flags or flaky tests).

`assert` gives a very nice error message with exactly which test(s) did not fail (ie. tests that were not in both objects)

I should also add that the order in which tests fail in is entirely deterministic such that the order in which tests fail in will never differ. The failures are mapped by the name of the file itself, similarly to the status files.


mcollina (approved):
lgtm

mcollina (approved):
lgtm

PR #1745, opened 2022-10-28
fetch: update extractBody to better match spec

KhafraDev:


lib/fetch/body.js
```diff
@@ -172,17 +181,10 @@ function extractBody (object, keepalive = false) {
 
     stream =
       object instanceof ReadableStream ? object : ReadableStreamFrom(object)
-  } else {
-    // TODO: byte sequence?
-    // TODO: scalar value string?
-    // TODO: else?
-    source = toUSVString(object)
-    type = 'text/plain;charset=UTF-8'
```
ronag, line 180: What if none of the above is true? assert(false)?

KhafraDev: I think the spec lists all the potential values it can be. If the `body` passed isn't one of the allowed values, webidl converts the value to a USVString for us.

KhafraDev: See https://github.com/nodejs/undici/blob/d23f5dd00fed2d6fb85baeef9b714f3a0d1bad88/lib/fetch/response.js#L524-L578

KhafraDev:
hoping to get rid of most of the TODOs now that we can almost guarantee changing the behavior of things doesn't break

Issue #1746, opened 2022-10-29
undici doesnt support keep-alive connection header?


Automized:
I am trying to use the keep-alive connection header but it returns this error (see below)

```
(node:65464) UnhandledPromiseRejectionWarning: InvalidArgumentError: invalid connection header
```

Not sure if I am doing something wrong, maybe the keep-alive is called something different in undici or that it is not supported.

Please let me know, thank you!

ronag:
You should not be setting the keep-alive header yourself. undici already does this for you.

Automized:
> You should not be setting the keep-alive header yourself. undici already does this for you.

okay great, thank you for the quick response :)

Issue #1310, opened 2022-03-29
How to disconnect after fetch?


daron4ever:
I would like to know how to close the connection after using fetch(URL) immediately?

mcollina:
This is currently not possible. You can configure a global Agent that will do that.

jimmywarting:
Use abortSignal?

mcollina:
@jimmywarting it's not that simple. Undici maintains a persistent connection pool to the server with keep-alive sockets.

metcoder95:
Hey! I've been taking a look into this one but have some doubts about the possible shape of the API to control this behavior.

My initial idea was to provide a new option property to the `init` param of the `fetch` function, that when `true` it calls `dispatcher.destroy()` after a fetch. Something like:

```js
const result = await fetch('https://example.com', { autoClose: true });

// this results in a `dispatcher.destroy` getting called after the full body is consumed
// depending on the user's implementation
```

But it didn't seem too ergonomic as it adds a new prop outside of the `fetch` spec and also makes the implementation to be fully aware of the body's consumption (if any) to properly start to destroy either the pool or standalone `agent`.

The second option I have in mind is to add the `autoClose` option directly to the `agent` that allows the agent to self-destroy/close after any request. 

Would like to hear your thoughts, so I can draft some proposal afterwards 🙂

mcollina:
I would do both.

metcoder95:
> I think using pipelining: 0 will always close the request

That's true, but that's not always obvious (I think that is not present in the docs). I found myself quite a few times looking at why `tap` finished the tests but was hanging at an open handle until I figured out/remembered that it was persisting the connection.
Having it more explicit can improve ergonomics, but also making it more explicit in the documentation will solve the issue. Not sure which one is better to be honest, matter of tradeoffs maybe?

>I would do both.

Sure, I'll work on PoC then 🙂

silverwind:
I have a odd case where the `vitest` process keeps hanging randomly after `fetch` was invoked (about 1 in 10 runs). I'm already passing `pipelining: 0` but it seems to have no effect. Switching to `node-fetch` resolves the issue. I have not yet tried `dispatcher.destroy` but it seems quiet unergonomic if that is necessary for the process to cleanly shut down.

silverwind:
Node 19.1.0. Yes this should be possible to make a repro for.

Issue #444, opened 2020-10-01
`Client.reset` to reset connection without destroying client


ronag:
Provide a way in the API to force a reconnect.

PR #1747, opened 2022-10-29
feat: allow connection header in request

ronag:
This seems to come up quite often. I'm opening this PR as stub for anyone that wants to finish it up. I think it mostly needs test and docs.

Refs: https://github.com/nodejs/undici/issues/1746
Refs: https://github.com/nodejs/undici/issues/1310
Refs: https://github.com/nodejs/undici/issues/1305
Refs: https://github.com/nodejs/undici/issues/444

### DIFF
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -65,6 +65,7 @@ class Request {
     upgrade,
     headersTimeout,
     bodyTimeout,
+    reset,
     throwOnError
   }, handler) {
     if (typeof path !== 'string') {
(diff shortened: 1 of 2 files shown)


lib/core/request.js
```diff
@@ -311,7 +314,11 @@ function processHeader (request, key, val) {
     key.length === 10 &&
     key.toLowerCase() === 'connection'
   ) {
-    throw new InvalidArgumentError('invalid connection header')
+    if (val.toLowerCase() === 'close') {
+      this.reset = true
+    } else if (val.toLowerCase() !== 'keep-alive') {
+      throw new InvalidArgumentError('invalid connection header')
```
ronag, line 320: What if there is a conflict between header and reset option?

PR #1749, opened 2022-11-02
chore(docs/assets): compress image

Fdawgs:
Losslessly compress the lifecycle image using [Pngcrush](https://pmt.sourceforge.io/pngcrush/). Should reduce the build size by a tiny amount, as this is published in the package.


mcollina (approved):
lgtm

PR #1752, opened 2022-11-04
fix: `DiagnosticsChannel` namespace not exported

PhilipTrauner:
## Rationale

One can already subscribe to `diagnostics_channel` messages, but because the types of said messages aren't exported, they are awkward to deal with.

I'm assuming that the missing export declaration is an oversight, as multiple interfaces inside `DiagnosticsChannel` are already exported.

## Changes

* Add export declaration for `DiagnosticsChannel` namespace

### Bug Fixes

N/A

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/types/diagnostics-channel.d.ts
+++ b/types/diagnostics-channel.d.ts
@@ -3,6 +3,8 @@ import { URL } from "url";
 import { connector } from "./connector";
 import { HttpMethod } from "./dispatcher";
 
+export = DiagnosticsChannel;
+
 declare namespace DiagnosticsChannel {
   interface Request {
     origin?: string | URL;
--- a/types/diagnostics-channel.d.ts
+++ b/types/diagnostics-channel.d.ts
@@ -3,9 +3,7 @@ import { URL } from "url";
 import { connector } from "./connector";
 import { HttpMethod } from "./dispatcher";
 
-export = DiagnosticsChannel;
-
-declare namespace DiagnosticsChannel {
+export declare namespace DiagnosticsChannel {
   interface Request {
     origin?: string | URL;
     completed: boolean;


mcollina (changes requested):
Thanks for opening a PR! Can you please add a unit test? We use tsd for types

PhilipTrauner:
Thanks for hopping onto this so quickly @mcollina. Is the  [existing](https://github.com/nodejs/undici/blob/main/test/types/diagnostics-channel.test-d.ts) type test sufficient, or do you have something more extensive in mind?

mcollina:
I suspect some code was failing for you to suggest this change. What was failing? Could you add that?

PhilipTrauner:
> I'm assuming that the missing export declaration is an oversight, as multiple interfaces inside `DiagnosticsChannel` are already exported.

This PR is based on the assumption that the messages published with `diagnostics_channel` are [part of the public API](https://github.com/nodejs/undici/blob/main/docs/api/DiagnosticsChannel.md), and that the types of said messages are mistakenly not being exported. 

If that is not the case, this PR can be closed 😅.

mcollina:
They are part of the public API. I don't understand why this change is needed given that the following works:

https://github.com/nodejs/undici/blob/main/test/types/diagnostics-channel.test-d.ts

PhilipTrauner:
While I thought that I had upgraded to the newest version before opening this PR, it seems that I was mistaken. 
Due to a faulty version constraint I was testing against a version that did not yet include 5478cdd38eb057a1e86c15cf418eeb90486470b2.

Sorry for wasting your time 😬

PR #1754, opened 2022-11-05
webidl: small changes

KhafraDev:
*No behavior changes*

1. Add `webidl.brandCheck` which performs:
```js
if (!(x instanceof Y)) {
  throw new TypeError('Illegal invocation')
}
```
2. Add typescript types which make working with webidl nicer. These types aren't exported and are only used for dev UX. Next step: add types for converters added outside of the webidl file.
3. Throw errors outside of the webidl error utilities.

### DIFF
--- a/lib/fetch/body.js
+++ b/lib/fetch/body.js
@@ -331,9 +331,7 @@ function bodyMixinMethods (instance) {
     },
 
     async formData () {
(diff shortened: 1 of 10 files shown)


mcollina (approved):
lgtm

PR #1755, opened 2022-11-05
webidl: add `argumentLengthCheck` guard

KhafraDev:
No behavior changes

Adds `webidl.argumentLengthCheck(arguments, min, ...)` check which performs
```js
const { length } = arguments
if (length < min) {
  throw new TypeError('...')
}
```

Also checks for -0 using `x === 0` in some cases which removes some Object.is calls.


ronag:
One CI failure?

KhafraDev:
Guessing flaky test, running it again just in case. Although it wouldn't make sense for any of the changes here to fail just on windows

KhafraDev:
yeah it was flaky, now the ci is stuck on the node-fetch tests. Just windows things.

PR #1758, opened 2022-11-06
feat(MockInterceptor): allow async reply callbacks

### LINKED ISSUE #1757: Allow async callbacks in `reply`
## This would solve...

I am using a library that proxies `undici` in an unintended way, changing the type of `opts.body` to `AsyncGenerator`. I would like to read the generator and reply with the result, but it's not possible because reply callback isn't awaited. This was discussed more in-depth here https://github.com/nodejs/undici/issues/1756.

This capability would also allow awaiting other async functions in the callback 

In my specific use-case, I'd be able to read `AsyncGenerator` and return the body string:

```
.reply(200, async (opts) => {
  const buffers = [];
  for await (const data of opts.body) {
    buffers.push(data);
  }
  const body = Buffer.concat(buffers).toString('utf8');
    
  return body;
});
```

KhafraDev:
Fixes #1757

Not the cleanest solution ever, but one that involves no breaking changes.

In the future it would probably be a good idea to switch everything to async so all mock functions can be asynchronous (option matchers, reply callback, etc) without needing ugly workarounds. Didn't feel like making such a breaking change today, though.

### DIFF
--- a/lib/mock/mock-utils.js
+++ b/lib/mock/mock-utils.js
@@ -10,6 +10,11 @@ const {
 } = require('./mock-symbols')
 const { buildURL, nop } = require('../core/util')
 const { STATUS_CODES } = require('http')
+const {
+  types: {
+    isPromise
+  }
+} = require('util')
 
 function matchValue (match, value) {
   if (typeof match === 'string') {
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

Issue #1753, opened 2022-11-04
undici.stream ignores throwOnError


mdashlw:
## Bug Description

`undici.stream`/`Dispatcher.stream` accepts [`RequestOptions`](https://undici.nodejs.org/#/docs/api/Dispatcher?id=parameter-requestoptions) which extends [`DispatchOptions`](https://undici.nodejs.org/#/docs/api/Dispatcher?id=parameter-dispatchoptions) which includes `throwOnError` to throw on 4xx and 5xx HTTP errors, but completely ignores it and calls the factory regardless of the status code.

## Reproducible By

```js
import fs from "node:fs";
import undici from "undici";

await undici.stream(
  "https://httpbin.org/status/502",
  { throwOnError: true },
  () => fs.createWriteStream("httpbin_502.txt")
);
```

Successfully creates an empty `httpbin_502.txt` (because httpbin.org sends empty body on `/status`)

## Expected Behavior

```
ResponseStatusCodeError: Response status code 502: BAD GATEWAY
```

### Additional context

* `throwOnError` was introduced in https://github.com/nodejs/undici/pull/1453
* the current code for `request`: https://github.com/nodejs/undici/blob/0a7ba7226ce8cdd6bc67527bcd3427665ff7edd2/lib/api/api-request.js#L95
* `stream` lacks that code: https://github.com/nodejs/undici/blob/0a7ba7226ce8cdd6bc67527bcd3427665ff7edd2/lib/api/api-stream.js#L92

ronag:
PR welcome!

ronag:
Should do the same on undici.pipeline

MattBidewell:
If no one else is looking, I've started to take a look at this. :) Have a WIP draft PR here, not finished yet. 

https://github.com/nodejs/undici/pull/1766

assalielmehdi:
Hi, I implemented a [fix](https://github.com/nodejs/undici/pull/1767). Can you take a look?

PR #1767, opened 2022-11-09
fix(api-stream): Add throwOnError support in undici.stream

assalielmehdi:
[#1753](https://github.com/nodejs/undici/issues/1753)

### DIFF
--- a/test/client-stream.js
+++ b/test/client-stream.js
@@ -788,12 +788,13 @@ test('stream legacy needDrain', (t) => {
 })
 
 test('steam throw if statusCode >= 400', (t) => {
-  t.plan(1)
+  t.plan(3)
 
   const expectedBodyContent = 'expected_body_content'
+  const expectedStatus = 400
 
   const server = createServer((req, res) => {
-    res.writeHead(400, { 'Content-Type': 'text/plain' })
+    res.writeHead(expectedStatus, { 'Content-Type': 'text/plain' })
     res.end(expectedBodyContent)
   })
   t.teardown(server.close.bind(server))
@@ -802,26 +803,27 @@ test('steam throw if statusCode >= 400', (t) => {
     const client = new Client(`http://localhost:${server.address().port}`)
     t.teardown(client.close.bind(client))
 
-    const passThrough = new PassThrough()
-    const chunks = []
-
-    passThrough.on('data', (chunk) => {
-      chunks.push(chunk)
-    })
+    let written = false
 
     try {
       await client.stream({
         path: '/',
         method: 'GET',
         throwOnError: true,
-        opaque: passThrough
-      }, ({ opaque }) => opaque)
+        opaque: new PassThrough()
+      }, ({ opaque }) => {
+        opaque.on('data', (chunk) => {
+          written = true
+        })
+
+        return opaque
+      })
 
       t.fail('No Error')
     } catch (e) {
-      const actualBodyContent = Buffer.concat(chunks).toString()
-
-      t.equal(actualBodyContent, expectedBodyContent)
+      t.false(written)
+      t.equal(e.status, expectedStatus)
+      t.equal(e.body, expectedBodyContent)
     }
   })
 })
--- a/test/client-stream.js
+++ b/test/client-stream.js
@@ -786,3 +786,31 @@ test('stream legacy needDrain', (t) => {
     })
   })
 })
+
+test('steam throw if statusCode >= 400', (t) => {
+  t.plan(1)
+
+  const server = createServer((req, res) => {
+    res.writeHead(400, { 'Content-Type': 'text/plain' })
+    res.end()
+  })
+  t.teardown(server.close.bind(server))
+
+  server.listen(0, async () => {
+    const client = new Client(`http://localhost:${server.address().port}`)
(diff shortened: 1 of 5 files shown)


mdashlw:
This doesn't handle body properly, see differences between `api-request.js` and `api-stream.js` in `onData` and `onComplete`
```js
await undici.request('https://api.github.com/notfound', {
  throwOnError: true
})
```
```
ResponseStatusCodeError: Response status code 403: Forbidden {
  code: 'UND_ERR_RESPONSE_STATUS_CODE',
  body: '\r\n' +
    'Request forbidden by administrative rules. Please make sure your request has a User-Agent header (https://docs.github.com/en/rest/overview/resources-in-the-rest-api#user-agent-required). Check https://developer.github.com for other possible causes.\r\n',
  status: 403,
  statusCode: 403,
  headers: {
    'cache-control': 'no-cache',
    'content-type': 'text/html; charset=utf-8',
    'strict-transport-security': 'max-age=31536000',
    'x-content-type-options': 'nosniff',
    'x-frame-options': 'deny',
    'x-xss-protection': '0',
    'content-security-policy': "default-src 'none'; style-src 'unsafe-inline'",
    connection: 'close'
  }
}
```
```js
await undici.stream(
  "https://api.github.com/notfound",
  { throwOnError: true },
  () => fs.createWriteStream("response.txt")
);
```
```
ResponseStatusCodeError: Response status code 403: Forbidden {
  code: 'UND_ERR_RESPONSE_STATUS_CODE',
  body: undefined,
  status: 403,
  statusCode: 403,
  headers: {
    'cache-control': 'no-cache',
    'content-type': 'text/html; charset=utf-8',
    'strict-transport-security': 'max-age=31536000',
    'x-content-type-options': 'nosniff',
    'x-frame-options': 'deny',
    'x-xss-protection': '0',
    'content-security-policy': "default-src 'none'; style-src 'unsafe-inline'",
    connection: 'close'
  }
}
```

mcollina (commented):
good work, can you add docs?

assalielmehdi:
Hi @mdashlw, thank you for your comment.
I think that it's pretty normal for the body to be `undefined` in the case of `undici.stream()` since the `HTTP` body is written to the `writable stream` created by the factory. However, I had to make sure that the `writable` in finished before throwing an error.
We can retrieve the body from a `PassThrough` opaque if needed (see unit test).

assalielmehdi:
Hi @mcollina, thank you 🙏🏼.
In docs, `options` is already of type [`RequestOptions`](https://undici.nodejs.org/#/docs/api/Dispatcher?id=parameter-requestoptions) that extends [`DispatchOptions`](https://undici.nodejs.org/#/docs/api/Dispatcher?id=parameter-dispatchoptions), and `throwOnError` is already explained in DispatchOptions.
I see nothing more to add in docs. WDYT?

mdashlw:
https://github.com/nodejs/undici/pull/1767/commits/8f7a6d79ff2d43af6997770a710e0086057e9ba3 kind of re-introduces the entire issue.
This will create `response.txt` with the actual response and still throw `ResponseStatusCodeError`.
```js
await undici.stream(
  "https://api.github.com",
  { throwOnError: true },
  () => fs.createWriteStream("response.txt")
);
```
```
ResponseStatusCodeError: Response status code 403: Forbidden {
  code: 'UND_ERR_RESPONSE_STATUS_CODE',
  body: undefined
  ...
}
```
```
response.txt
Request forbidden by administrative rules. Please make sure your request has a User-Agent header (https://docs.github.com/en/rest/overview/resources-in-the-rest-api#user-agent-required). Check https://developer.github.com for other possible causes.
```
IMO the most intuitive and convenient way would be to throw an error with special body handling, as in `.request()`, and not to call the factory.

test/client-stream.js
```diff
@@ -786,3 +786,42 @@ test('stream legacy needDrain', (t) => {
...
+      chunks.push(chunk)
+    })
+
+    try {
+      await client.stream({
+        path: '/',
+        method: 'GET',
+        throwOnError: true,
+        opaque: passThrough
+      }, ({ opaque }) => opaque)
```
mcollina, line 818: I would not expect this factory to be called at all.

mcollina (changes requested):

assalielmehdi:
Since in `undici.request()`, the body is consumed before throwing the error, I figured we want the same from `undici.stream()`. I fixed the code to throw an error before even calling the factory, and updated the unit test accordingly.

@mdashlw for the body handling in `undici.request()`, the body is consumed before throwing, and we can't do so in `undici.stream()` unless we call the factory. So `body` will always be undefined in the error object.

mcollina:
You can consume the body without calling the factory.

assalielmehdi:
I updated the logic, to throw before calling the factory, and I handled the body so the thrown error contains it. I updated the tests accordingly. I hope that is the behavior we expect.

assalielmehdi:
Hi @mcollina, can you review the changes please? 😊

KhafraDev:
there are ci failures caused by a node-fetch test (`npm run test:node-fetch`)

ronag:
This will be fixed once we move `throwOnError` to an interceptor.

https://github.com/nodejs/undici/issues/2835

PR #1771, opened 2022-11-16
fix: remove unnecessary toLowerCase

anonrig:
Line 135: `this.contains(name)` already runs `name.toLowerCase()` there is no need to run it twice.


KhafraDev (approved):

mcollina (approved):
lgtm

PR #1772, opened 2022-11-16
perf: simplify isValidHeaderName

anonrig:
`isValidHTTPToken` already checks character by character. There is no need for the additional for loop.


KhafraDev (approved):

mcollina (approved):
lgtm

PR #1773, opened 2022-11-16
fix(fetch): remove unnecessary encode operation

anonrig:


KhafraDev (approved):

ronag:
Could also allocate that as a global?

mcollina (approved):
lgtm

silverwind:
> Could also allocate that as a global?

You mean as a constant at top-level? Sure, why not.

silverwind:
Probably put it in `lib/fetch/constants.js`.

kibertoad:
@anonrig wasn't the idea to move it to a constant to avoid unnecessary repeated allocation?

PR #1774, opened 2022-11-16
perf(fetch): simplify url serializer

anonrig:
There is no need for the `URLSerializer` since we already have a reference to the `URL` object. `url.toString()` produces the same result

### DIFF
--- a/lib/fetch/dataURL.js
+++ b/lib/fetch/dataURL.js
@@ -1,5 +1,6 @@
 const assert = require('assert')
 const { atob } = require('buffer')
+const { format } = require('url')
 const { isValidHTTPToken, isomorphicDecode } = require('./util')
 
 const encoder = new TextEncoder()
@@ -111,73 +112,7 @@ function dataURLProcessor (dataURL) {
  * @param {boolean} excludeFragment
  */
 function URLSerializer (url, excludeFragment = false) {
-  // 1. Let output be url’s scheme and U+003A (:) concatenated.
-  let output = url.protocol
-
-  // 2. If url’s host is non-null:
-  if (url.host.length > 0) {
-    // 1. Append "//" to output.
-    output += '//'
-
-    // 2. If url includes credentials, then:
-    if (url.username.length > 0 || url.password.length > 0) {
-      // 1. Append url’s username to output.
-      output += url.username
-
-      // 2. If url’s password is not the empty string, then append U+003A (:),
-      // followed by url’s password, to output.
-      if (url.password.length > 0) {
-        output += ':' + url.password
-      }
-
-      // 3. Append U+0040 (@) to output.
-      output += '@'
-    }
-
-    // 3. Append url’s host, serialized, to output.
-    output += decodeURIComponent(url.hostname)
-
-    // 4. If url’s port is non-null, append U+003A (:) followed by url’s port,
-    // serialized, to output.
-    if (url.port.length > 0) {
-      output += ':' + url.port
-    }
-  }
-
-  // 3. If url’s host is null, url does not have an opaque path,
-  // url’s path’s size is greater than 1, and url’s path[0]
-  // is the empty string, then append U+002F (/) followed by
-  // U+002E (.) to output.
-  // Note: This prevents web+demo:/.//not-a-host/ or web+demo:/path/..//not-a-host/,
-  // when parsed and then serialized, from ending up as web+demo://not-a-host/
-  // (they end up as web+demo:/.//not-a-host/).
-  // Undici implementation note: url's path[0] can never be an
-  // empty string, so we have to slightly alter what the spec says.
-  if (
-    url.host.length === 0 &&
-    url.pathname.length > 1 &&
-    url.href.slice(url.protocol.length + 1)[0] === '.'
-  ) {
-    output += '/.'
-  }
-
-  // 4. Append the result of URL path serializing url to output.
-  output += url.pathname
-
-  // 5. If url’s query is non-null, append U+003F (?),
-  // followed by url’s query, to output.
-  if (url.search.length > 0) {
-    output += url.search
-  }
-
-  // 6. If exclude fragment is false and url’s fragment is non-null,
-  // then append U+0023 (#), followed by url’s fragment, to output.
-  if (excludeFragment === false && url.hash.length > 0) {
-    output += url.hash
-  }
-
-  // 7. Return output.
-  return output
+  return format(url, { fragment: !excludeFragment })
 }
 
 // https://infra.spec.whatwg.org/#collect-a-sequence-of-code-points
--- a/test/fetch/data-uri.js
+++ b/test/fetch/data-uri.js
@@ -177,3 +177,17 @@ test('too long base64 url', async (t) => {
     t.fail(`failed to fetch ${dataURL}`)
   }
 })
+
+test('https://domain.com/#', (t) => {
+  t.plan(1)
+  const domain = 'https://domain.com/#'
+  const serialized = URLSerializer(new URL(domain))
+  t.equal(serialized, domain)
+})
+
+test('https://domain.com/?', (t) => {
+  t.plan(1)
+  const domain = 'https://domain.com/?'
+  const serialized = URLSerializer(new URL(domain))
+  t.equal(serialized, domain)
+})
--- a/test/node-fetch/main.js
+++ b/test/node-fetch/main.js
@@ -1532,34 +1532,12 @@ describe('node-fetch', () => {
     })
   })
 
-  it('should NOT keep `?` sign in URL when no params are given', () => {
-    const url = `${base}question?`
-    const urlObject = new URL(url)
-    const request = new Request(urlObject)
-    return fetch(request).then(res => {
-      expect(res.url).to.equal(url.slice(0, -1))
-      expect(res.ok).to.be.true
-      expect(res.status).to.equal(200)
-    })
-  })
-
   it('if params are given, do not modify anything', () => {
     const url = `${base}question?a=1`
     const urlObject = new URL(url)
     const request = new Request(urlObject)
     return fetch(request).then(res => {
       expect(res.url).to.equal(url)
-      expect(res.ok).to.be.true
-      expect(res.status).to.equal(200)
-    })
-  })
-
-  it('should NOT preserve the hash (#) symbol', () => {
-    const url = `${base}question?#`
-    const urlObject = new URL(url)
-    const request = new Request(urlObject)
-    return fetch(request).then(res => {
-      expect(res.url).to.equal(url.slice(0, -2))
       expect(res.ok).to.be.true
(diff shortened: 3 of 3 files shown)


KhafraDev (approved):
I'm fine with the change (URL's toString implements the url serializer algorithm) but performance of the node core version is almost definitely slower due to its seemingly unnecessary complexity.

https://github.com/nodejs/node/blob/5fad0b93667ffc6e4def52996b9529ac99b26319/lib/internal/url.js#L699

---

Totally unrelated, but removing the empty object from [`.toString`](https://github.com/nodejs/node/blob/5fad0b93667ffc6e4def52996b9529ac99b26319/lib/internal/url.js#L750) would probably help performance as it would remove the `validateObject` call.

anonrig:
If URL's `toString` method is slow, let's fix it on the Node side.

anonrig:
I opened a pull request and started the benchmark CI. We'll have the results in a couple of hours. @KhafraDev

KhafraDev:
Thanks, I've found an issue with the current undici serialization that I will fix once this is merged as well.

ronag (changes requested):
I kind of prefer the literal spec implementation... how much faster is this in practice?

kibertoad:
@ronag What's the benefit of including complexity within the lib if upstream produces same result? Shouldn't it be safe swap as long as there are tests covering existing behaviour?

anonrig:
> I kind of prefer the literal spec implementation... how much faster is this in practice?

I understand the preference for spec implementation, but `new URL` is also spec-compliant. The performance is neglectable.

ronag:
As far as I know `URL.toString` is not standard and just happens to be similar but could be changed.

anonrig:
> As far as I know `URL.toString` is not standard and just happens to be similar but could be changed.

Then, I recommend making sure that it's spec compliant in the node core, instead of implementing it twice. I'll be happy to keep track of it. @Trott has more experience on this I believe. What do you think?

ronag:
I think making `URL.toString` implement the spec is breaking due to:
```
if (serialized.endsWith('?')) {
   return serialized.slice(0, -1)
}
```

I'm pretty sure the echo system relies on that.

ronag:
`URL.href` is standardised. Can we base it on that?

ronag:
`The href getter steps and the toJSON() method steps are to return the [serialization](https://url.spec.whatwg.org/#concept-url-serializer) of [this](https://webidl.spec.whatwg.org/#this)’s [URL](https://url.spec.whatwg.org/#concept-url-url).`

lib/fetch/dataURL.js
```diff
@@ -111,73 +111,17 @@ function dataURLProcessor (dataURL) {
...
-  // Undici implementation note: url's path[0] can never be an
-  // empty string, so we have to slightly alter what the spec says.
-  if (
-    url.host.length === 0 &&
-    url.pathname.length > 1 &&
-    url.href.slice(url.protocol.length + 1)[0] === '.'
-  ) {
-    output += '/.'
+  if (excludeFragment) {
+    url.hash = ''
```
ronag, line 115: We can't mutate the input.

lib/fetch/dataURL.js
```diff
@@ -111,73 +111,17 @@ function dataURLProcessor (dataURL) {
...
-    output += url.search
-  }
+  const serialized = url.toString()
 
-  // 6. If exclude fragment is false and url’s fragment is non-null,
-  // then append U+0023 (#), followed by url’s fragment, to output.
-  if (excludeFragment === false && url.hash.length > 0) {
-    output += url.hash
+  if (serialized.endsWith('?')) {
+    return serialized.slice(0, -1)
```
ronag, line 121: actually why is this needed? Point 5 says `?`  should be appended...

lib/fetch/dataURL.js
```diff
@@ -111,73 +111,17 @@ function dataURLProcessor (dataURL) {
...
 
-  // 4. Append the result of URL path serializing url to output.
-  output += url.pathname
-
-  // 5. If url’s query is non-null, append U+003F (?),
-  // followed by url’s query, to output.
-  if (url.search.length > 0) {
-    output += url.search
-  }
+  const serialized = url.toString()
```
ronag, line 118:
```suggestion
  const serialized = url.href
```

lib/fetch/dataURL.js
ronag, line 115:
```suggestion
    url = new URL(url)
    url.hash = ''
```

anonrig: I'm updating the pull request to use `format(URL, { fragment: !excludeFragment })`

lib/fetch/dataURL.js
anonrig: 2 node-fetch tests are failing without this change.

```
  1) node-fetch
       should NOT keep `?` sign in URL when no params are given:

      AssertionError: expected 'http://localhost:60850/question?' to equal 'http://localhost:60850/question'
      + expected - actual

      -http://localhost:60850/question?
      +http://localhost:60850/question

      at /Users/yagiz/Developer/undici/test/node-fetch/main.js:1540:26

  2) node-fetch
       should NOT preserve the hash (#) symbol:

      AssertionError: expected 'http://localhost:60850/question?' to equal 'http://localhost:60850/question'
      + expected - actual

      -http://localhost:60850/question?
      +http://localhost:60850/question

      at /Users/yagiz/Developer/undici/test/node-fetch/main.js:1562:26
```

lib/fetch/dataURL.js
```diff
@@ -111,73 +112,13 @@ function dataURLProcessor (dataURL) {
...
-  ) {
-    output += '/.'
-  }
-
-  // 4. Append the result of URL path serializing url to output.
-  output += url.pathname
-
-  // 5. If url’s query is non-null, append U+003F (?),
-  // followed by url’s query, to output.
-  if (url.search.length > 0) {
```
anonrig, line 169: This code is wrong. `non-null` does not mean `.length > 0`

anonrig:
@KhafraDev @ronag I removed 2 faulty `node-fetch` tests, which are not spec-compliant, and updated my implementation.

ronag (approved):

KhafraDev:
> I believe either we or node is implementing URL serialisation incorrectly

Yes, this *was* the issue I was going to fix. 😄

lib/fetch/dataURL.js
KhafraDev: This can never be null in js land though?

KhafraDev (changes requested):
can you add tests for urls with an empty fragment and an empty search query?

lib/fetch/dataURL.js
```diff
@@ -111,73 +112,7 @@ function dataURLProcessor (dataURL) {
...
-
-  // 6. If exclude fragment is false and url’s fragment is non-null,
-  // then append U+0023 (#), followed by url’s fragment, to output.
-  if (excludeFragment === false && url.hash.length > 0) {
-    output += url.hash
-  }
-
-  // 7. Return output.
-  return output
+  return format(url, { fragment: !excludeFragment })
```
ronag, line 115:
```suggestion
  return format(url, { fragment: excludeFragment !== false })
```

test/node-fetch/main.js
```diff
@@ -1554,17 +1543,6 @@ describe('node-fetch', () => {
...
-  it('should NOT preserve the hash (#) symbol', () => {
-    const url = `${base}question?#`
-    const urlObject = new URL(url)
-    const request = new Request(urlObject)
-    return fetch(request).then(res => {
-      expect(res.url).to.equal(url.slice(0, -2))
-      expect(res.ok).to.be.true
-      expect(res.status).to.equal(200)
-    })
-  })
```
mcollina, line 1566: why those tests were removed?

kibertoad: apparently, they were against the spec

KhafraDev (approved):

lib/fetch/dataURL.js
ronag: Yea. I think this is a spec bug.

ronag:
CI fails?

lib/fetch/dataURL.js
anonrig: According to the spec, it is null initially.

A [URL](https://url.spec.whatwg.org/#concept-url)’s query is either null or an [ASCII string](https://infra.spec.whatwg.org/#ascii-string). It is initially null.

KhafraDev: That's an internal representation of a url, which the fetch spec assumes we have access to (similar to other parts of the code). For us, search can never be null: "if [this](https://webidl.spec.whatwg.org/#this)’s [URL](https://url.spec.whatwg.org/#concept-url-url)’s [query](https://url.spec.whatwg.org/#concept-url-query) is either null or the empty string, then return the empty string." (https://url.spec.whatwg.org/#dom-url-search)

anonrig:
I couldn't find the CI error in the logs, and the tests succeeded locally. Does anybody have more in-depth knowledge?

KhafraDev:
Looks like https://github.com/nodejs/undici/blob/93f150f5bbf0ce66d50d9b419277b2b13141b8b3/test/fetch/data-uri.js#L52 and https://github.com/nodejs/undici/blob/93f150f5bbf0ce66d50d9b419277b2b13141b8b3/test/fetch/data-uri.js#L57 are failing from the logs.

test/fetch/data-uri.js
```diff
@@ -177,3 +177,14 @@ test('too long base64 url', async (t) => {
     t.fail(`failed to fetch ${dataURL}`)
   }
 })
+
+test('https://domain.com#', (t) => {
+  const domain = 'https://domain.com#'
```
KhafraDev, line 182:
```suggestion
  const domain = 'https://domain.com/#'
```

unless this is expected, and then there's a bug in url serializer?

test/fetch/data-uri.js
```diff
@@ -177,3 +177,14 @@ test('too long base64 url', async (t) => {
...
+  const serialized = URLSerializer(new URL(domain))
+  t.same(serialized, domain)
+})
+
+test('https://domain.com?', (t) => {
+  const domain = 'https://domain.com?'
```
KhafraDev, line 188: maybe the same here too

test/fetch/data-uri.js
anonrig: I think this is correct. In chrome, the result has `/` too.

KhafraDev:
Okay so the last hopefully is that you need to call `t.end()` after the test has run to signify that it's over, or use `t.plan(1)` to plan a single test. This is only needed when the tests are synchronous, and not very fun to debug from experience.

anonrig:
I understood what the issue is. If I used a `async` function in the test callback, I wouldn't need `t.plan(1)` or `t.end()`. Thanks for the help @KhafraDev

PR #1775, opened 2022-11-16
fetch(HeadersList): s/has/contains

KhafraDev:
[Contains](https://fetch.spec.whatwg.org/#header-list-contains) is the spec version of "has". It was added a while ago and *contains* the same logic as has did.


ronag (approved):

ronag:
@mcollina breaking

KhafraDev:
this contains no breaking changes, only the internal HeadersList class is affected.

PR #1779, opened 2022-11-19
feat(fetch): add in native File class support

KhafraDev:
I added File to node core in https://github.com/nodejs/node/pull/45139

This PR prioritizes using it if it exists, and falling back to the undici implementation if it doesn't exist.

---

I tested this PR with node v18.11.0 and https://github.com/nodejs/node/commit/b18ec0433e602d17693a316ddcd60357fa1892d8

I only tested the WPTs, so node-fetch and the undici tests may still fail.

### DIFF
--- a/lib/fetch/body.js
+++ b/lib/fetch/body.js
@@ -7,17 +7,20 @@ const { FormData } = require('./formdata')
 const { kState } = require('./symbols')
 const { webidl } = require('./webidl')
 const { DOMException, structuredClone } = require('./constants')
-const { Blob } = require('buffer')
+const { Blob, File: NativeFile } = require('buffer')
 const { kBodyUsed } = require('../core/symbols')
 const assert = require('assert')
 const { isErrored } = require('../core/util')
 const { isUint8Array, isArrayBuffer } = require('util/types')
-const { File } = require('./file')
+const { File: UndiciFile } = require('./file')
 const { StringDecoder } = require('string_decoder')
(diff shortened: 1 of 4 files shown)


mcollina (approved):
lgtm

lib/fetch/body.js
```diff
@@ -7,18 +7,21 @@ const { FormData } = require('./formdata')
...
 const { webidl } = require('./webidl')
 const { DOMException, structuredClone } = require('./constants')
-const { Blob } = require('buffer')
+const { Blob, File: NativeFile } = require('buffer')
 const { kBodyUsed } = require('../core/symbols')
 const assert = require('assert')
 const { isErrored } = require('../core/util')
 const { isUint8Array, isArrayBuffer } = require('util/types')
-const { File } = require('./file')
+const { File: UndiciFile } = require('./file')
```
anonrig, line 15: Can we just change the `./file` and dynamically change the module.exports if NativeFile exists?

anonrig (changes requested):

lib/fetch/body.js
KhafraDev: The issue with that is that undici's File class won't be accessible in node 19.x and above.

ie. you can import the File class by doing
```js
import { File } from 'undici'
```

KhafraDev:
The CI is still disabled for v19 but File was just released in v19.2.0 so maybe it should be merged regardless?

mcollina (approved):
lgtm

ronag (approved):

PR #1782, opened 2022-11-23
fix: use readable-stream

ronag:


ronag:
Could have impact when vendoring into node. Should probably use proper stream in that case... is there an easy fix for that? i.e. alias `readable-stream` with `stream` when required by core?

mcollina:
We should definitely _not_ bundle readable-stream when preparing the code for core.

ronag:
@targos

targos:
What about this:

```diff
diff --git a/package.json b/package.json
index a02ae10..222a235 100644
--- a/package.json
+++ b/package.json
@@ -41,7 +41,8 @@
     "docs"
   ],
   "scripts": {
-    "build:node": "npx esbuild@0.14.38 index-fetch.js --bundle --platform=node --outfile=undici-fetch.js",
+    "build:node": "npx esbuild@0.15.15 index-fetch.js --bundle --external:readable-stream --platform=node --outfile=undici-fetch.js",
+    "postbuild:node": "sed -i 's/require(\"readable-stream\")/require(\"stream\")/g' undici-fetch.js",
     "prebuild:wasm": "docker build -t llhttp_wasm_builder -f build/Dockerfile .",
     "build:wasm": "node build/wasm.js --docker",
     "lint": "standard | snazzy",
```

ronag:
Thank you! Applied.

ronag:
@mcollina wdyt?

mcollina:
CI seems really unhappy with this change.

ronag:
@mcollina yea, assuming I managed to fix that, do you have any further objections?

mcollina:
Not, it should have been like this from the beginning but we were stuck with a very old and legacy version of `readable-stream`.

targos:
Isn't the fix simply to `npm install readable-stream` ?

jimmywarting:
I would prefer if you just used `node:stream` instead of depending on a npm package

ronag:
> I would prefer if you just used `node:stream` instead of depending on a npm package

Why?

jimmywarting:
> > I would prefer if you just used `node:stream` instead of depending on a npm package
> 
> Why?

- Fewer dependencies
- faster to download
- `node:stream` is always up to date with newest features
- `readable-stream` have lagged behind in features
- `readable-stream` also depends on buffer which itself haven't been updated for 2y
- so it will use a npm:buffer instead of nodes which is also slower than node:buffer when it comes to things like `buf.toString()` as it dosen't use native code and so on.

ronag:
@mcollina wdyt?

mcollina:
Why did you want to do this change in the first place?

ronag:
Make sure we have the latest stream fixes even when not running latest version of Node.

ronag:
> * `node:stream` is always up to date with newest features

This is a truth with modification. Only if/when you are running latest version of Node. Otherwise, the opposite should be true.

> * `readable-stream` have lagged behind in features

I think and hope that situation is no longer true.

> * `readable-stream` also depends on buffer which itself haven't been updated for 2y

Yea, I guess that is a problem.

> * so it will use a npm:buffer instead of nodes which is also slower than node:buffer when it comes to things like `buf.toString()` as it dosen't use native code and so on.

Also true.

jimmywarting:
also it uses abort-controller polyfill instead of using built in whenever available. and it too lacks things like `AbortSignal.prototype.reason`

ronag:
> I mostly saw it as "just to bring it outside of nodejs". all those legacy package never updated their own code after a while and those started to lack newer features such as async iterable

For me. I want to have all the latest fixes and features from Node streams even when running LTS.

ronag:
Would be nice if it only used the polyfills when required, i.e. running an older Node version than expected.

mcollina:
@jimmywarting those sounds good improvements to do to `readable-stream`. Maybe open a few issues there / or send PRs?

ronag:
So it seems to me like this could be a good idea but would require some more things from readable-streams:

- [ ] https://github.com/nodejs/readable-stream/issues/501
- [ ] Only use buffer package when Node version < expected
- [ ] Only use abort-controller package when Node version < expected

jimmywarting:
> Maybe open a few issues there / or send PRs?

I'm not so keen at working on buffer and/or stream. i have a huge cross compatible code mindset, such as (async)Iterators + uint8array instead (and now web streams). also got a huge bundlephobia for large packages 😝 

But you can go ahead and do that...
conditional import when required is a 👍 

I just recently did a config.js in one of my project that only exported the things i need to work with such as:
```js
const config = {
  ReadableStream: globalThis.ReadableStream,
  WritableStream: globalThis.WritableStream,
  TransformStream: globalThis.TransformStream,
  Promise: globalThis.Promise,
  DOMException: globalThis.DOMException,
  Blob: globalThis.Blob,
  File: globalThis.File
}

export default config
```
and if developers wanted to they could override all those things before i even start using anything of it...
```js
import config from 'mod/config.js'
config[xyz] = zxy
import mod
```
so if they needed to support older version then they could bring their own polyfill and the version they needed/wanted/or already had

package.json
```diff
@@ -41,7 +41,8 @@
     "docs"
   ],
   "scripts": {
-    "build:node": "npx esbuild@0.14.38 index-fetch.js --bundle --platform=node --outfile=undici-fetch.js",
+    "build:node": "npx esbuild@0.15.15 index-fetch.js --bundle --external:readable-stream --platform=node --outfile=undici-fetch.js",
+    "postbuild:node": "sed -i 's/require(\"readable-stream\")/require(\"stream\")/g' undici-fetch.js",
```
targos, line 45:
```suggestion
    "build:node": "npx esbuild@0.15.16 index-fetch.js --bundle --alias:readable-stream=stream --platform=node --outfile=undici-fetch.js",
```

esbuild just added a flag for this: https://github.com/evanw/esbuild/releases/tag/v0.15.16

PR #1784, opened 2022-11-25
fix(fetch): treat headers as case sensitive

KhafraDev:
This fixes an outstanding issue in fetch where header names were treated as case insensitive, always being sent to the server as lowercase. It's been on my radar for a while, but I could never find a decent approach to fixing it. Hopefully this solution doesn't sacrifice a ton of performance as it simply replaces the `lowercase_key ⇔ value` HeadersList structure with a `lowercase_key ⇔ { key, value }` one.


mcollina:
This is not an issue. Headers are case insensitive by the HTTP specs. Lowercasing all of them makes processing faster.

mcollina (changes requested):
I'm -1 with this change

ronag:
Unless the fetch spec says otherwise, I agree with @mcollina. What does Chrome do?

mcollina:
[RFC2516](https://www.ietf.org/rfc/rfc2616.txt), 4.2:

>  HTTP header fields, which include general-header (section 4.5),
   request-header (section 5.3), response-header (section 6.2), and
   entity-header (section 7.1) fields, follow the same generic format as
   that given in Section 3.1 of RFC 822 [9]. Each header field consists
   of a name followed by a colon (":") and the field value. __Field names
   are case-insensitive.__ ...

silverwind:
Woudn't it be faster if `.toLowerCase` is not performed at all?

Generally, I'd lean towards what other implementations do, either let headers through with unmodified case or lowercase to favor faster processing on the receiving end.

KhafraDev:
@mcollina there's a WPT for this that all browsers pass, and deno fails - https://wpt.fyi/results/fetch/api/basic/request-headers-case.any.html?label=master&label=experimental&product=chrome&product=edge&product=firefox&product=safari&product=deno&aligned&view=subtest

mcollina:
I can't find a link to that test source. Can you add it?

---

My -1 remain.

lowercasing all headers is one of the reasons undici.request is faster than node http.request.

This change will slow us down further and our performance is already subpar.

Finally, there is absolutely no reason to do this as the current behavior is in line with the HTTP spec, and ultimately the servers needs to understand HTTP, not the fetch() WPT.

KhafraDev:
The test exists at https://github.com/web-platform-tests/wpt/blob/master/fetch/api/basic/request-headers-case.any.js and I added it in this commit.

Regarding undici.request, it doesn't lowercase headers automatically, here's a small repro using node's http server & undici.request:

```js
import { once } from 'events'
import { request } from './index.js'

  console.log(req.rawHeaders)
  res.end()
}).listen(0)

await request(`http://localhost:${server.address().port}`, {
  headers: {
    'ABC-def': 'value'
  }
}).finally(() => server.close())

````

Where it outputs

```js
[
  'host',
  'localhost:51291',
  'connection',
  'keep-alive',
  'ABC-def',
  'value'
]
```

KhafraDev:
> Field names are case-insensitive

For this, the RFC doesn't say that headers should lowercase - just that `ABC` is the same as `abc` (which is still the behavior in this PR). The only difference is that we don't send the lowercased header to the server anymore, but the first one the user sent. To explain it more, if someone did `headers.append('A', 'b')` and `headers.append('a', 'b')`, the server will receive `A: 'b, b'`.

mcollina:
This kind of processing of Headers is one of the bottlenecks of OutgoingMessage. Let's not repeat the same mistakes.
Given this is cosmetic, let's not do this.

ronag:
I'm +- 0. When it comes to fetch I personally care more about spec than perf.

KhafraDev:
> This kind of processing of Headers is one of the bottlenecks of OutgoingMessage.

Would you be willing to explain this further? Undici doesn't send lowercase headers to the server, but sends them in the case that they are set (other than fetch currently). This PR matches that behavior, where we send headers as we receive them. See the `undici.request` example above.

PR #1785, opened 2022-11-25
fix(fetch): set content-length header for FormData body

### LINKED ISSUE #1783: content-length never sent
Are there any plans to optionally send the content-length header?

Some API's reject requests that are missing the header, for example sending FormData to AWS S3 returns 411.

KhafraDev:
Fixes #1783 

If you have a better solution feel free to let me know.

WPTs test this behavior so I only added 1 normal test.

---

I also checked what Deno did; they convert the FormData to a Blob and then read the `.size` property for the length. If we did that here, we would set `action` to `blob.stream()`. I considered doing this but decided against it because we'd have to load the entirety of the FormData into memory first.


mcollina:
Should the WPT test be enabled or something?

test/fetch/content-length.js
```diff
@@ -0,0 +1,31 @@
...
+const { test } = require('tap')
+const { Blob } = require('buffer')
+const { FormData, fetch } = require('../..')
+const { once } = require('events')
+const { createServer } = require('http')
+
+const isV16x = process.version.startsWith('v16.')
+
+// https://github.com/nodejs/undici/issues/1783
+test('Sending a FormData body sets Content-Length header', { skip: isV16x }, async (t) => {
```
mcollina, line 12: why skipped on node 16?

KhafraDev: The length is different on ubuntu v16.x for some reason (3 extra characters somewhere).

KhafraDev:
> Should the WPT test be enabled or something?

They already are! https://github.com/nodejs/undici/tree/main/test/wpt/tests/FileAPI/file (`send-file-formdata-*` tests)

test/fetch/content-length.js
JamesKyburz: > The length is different on ubuntu v16.x for some reason (3 extra characters somewhere).

Could it be the random identifier in boundary?

KhafraDev: yes, most likely. Although locally tests passed in v16 so I decided just to disable this one as there are other tests that run and will fail if there's an issue with the content-length calculation in the future.

ronag: I would prefer if we understood what's going on here

KhafraDev: A weird ubuntu-only ci-only (?) 'bug' where the length calculated is different than other platforms. The length was calculated correctly and the body was sent, otherwise we'd get an error for mismatched content-length.

ronag: How can the length be calculated differently?

jimmywarting: maybe you are calling `Math.random()` twice? once when you are calculating the length and another time when you are sending the bits to the api endpoint?

Anyway Math.random() can return different length.
```js
i=1000; while(i--) console.log(Math.random().toString().length) // 17-20 can vary more...
```

> I considered doing this but decided against it because we'd have to load the entirety of the FormData into memory first.

A good Blob impl wouldn't have to read everything into memory if it can just make references points to other blob parts.
so: `new Blob([file1, file2])` would only hold 2 references points to other blob barts without reading anything into memory basically.
only the strings would allocate new memory. (btw, doing it as you have currently done requires to basically do the same logic twice `escape + normalizeLinefeeds` where as a formData to blob would already have pre computed the data that needs to be sent. 

The history of Deno's formData to blob originates from my formdata package where the old browsers wouldn't be able to write a stream to the request (so had to have all data ready at `xhr.send(blob)`

I think it's a decent solution...

jimmywarting: In my formdata impl (where i did generate the boundary i did something like 
```js
`${ Math.random() }${ Math.random() }`.slice(-28).padStart(32, '-')`
```
so i always had the same length...

KhafraDev: I'll remove this test since we have coverage for sending POST requests with FormData bodies already, since it relies on a length which can/might vary(?).

ronag: Why not fix it?

KhafraDev: Nothing is really broken - otherwise we'd get a content-length mismatch error. We also have both normal tests and WPTs that ensure that the length is being calculated properly.

Maybe in a future PR we could use a fixed length boundary.

ronag (approved):

mcollina:
> > Should the WPT test be enabled or something?
> 
> 
> 
> They already are! https://github.com/nodejs/undici/tree/main/test/wpt/tests/FileAPI/file (`send-file-formdata-*` tests)

I might not understand our herness, the point of tests is that they fail the build if they are not passing.

KhafraDev:
The WPTs don't test for an exact content-length header value because it'd be impossible to do with the varying boundary between platforms. However those tests ensure that utf-8 and ascii files/file names work correctly and the content-length header value is correct (because undici checks the content-length value and throws an error if they mismatch).

The WPT runner will cause the CI to fail if any tests fail. All the tests are passing with this change, which means that with ascii & utf-8 file names and bodies, the length is correct.

mcollina:
> The WPTs don't test for an exact content-length header value because it'd be impossible to do with the varying boundary between platforms. However those tests ensure that utf-8 and ascii files/file names work correctly and the content-length header value is correct (because undici checks the content-length value and throws an error if they mismatch).

The overall goal of this feature is to send a content-length to the server. A test therefore is needed.

KhafraDev:
yeah, that's a good point. I'll add in a test that ensures the header is sent & is a numeric value, and leave a TODO to check the length once the boundary is a fixed length.

mcollina (approved):
lgtm

PR #1790, opened 2022-11-29
fix: connect option types

Kaciras:
This PR fixes `buildConnector.Options` issues:

* The `port` property passed from `URL.port` is a string.
* Missing `localAddress` and `httpSocket`.


mcollina (approved):
lgtm

PR #1792, opened 2022-11-30
perf: reduce number of native calls from fetch.body

anonrig:
This pull request includes 2 changes around improving the performance:

1. Each `Buffer.byteLength` makes a C++ call. If we concat them and then calculate pass it to ByteLength, it is faster.

2. `toLowerCase()` is a slow method. That's why I divided the if statement into 2 parts.

### DIFF
--- a/lib/fetch/body.js
+++ b/lib/fetch/body.js
@@ -151,22 +151,18 @@ function extractBody (object, keepalive = false) {
         if (typeof value === 'string') {
           bodyLength +=
             prefixLength +
-            Buffer.byteLength(`; name="${escape(normalizeLinefeeds(name))}"`) +
-            Buffer.byteLength(`\r\n\r\n${normalizeLinefeeds(value)}\r\n`)
+            Buffer.byteLength(`; name="${escape(normalizeLinefeeds(name))}"\r\n\r\n${normalizeLinefeeds(value)}\r\n`)
         } else {
           bodyLength +=
             prefixLength +
-            Buffer.byteLength(`; name="${escape(normalizeLinefeeds(name))}"`) +
-            (value.name ? Buffer.byteLength(`; filename="${escape(value.name)}"`) : 0) +
+            Buffer.byteLength(`; name="${escape(normalizeLinefeeds(name))}"` + (value.name ? `; filename="${escape(value.name)}"` : '')) +
             2 + // \r\n
             `Content-Type: ${
               value.type || 'application/octet-stream'
             }\r\n\r\n`.length
 
-          // value is a Blob or File
-          bodyLength += value.size
-
-          bodyLength += 2 // \r\n
+          // value is a Blob or File, and \r\n
+          bodyLength += value.size + 2
         }
       }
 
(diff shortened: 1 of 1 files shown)


lib/fetch/body.js
```diff
@@ -151,22 +151,18 @@ function extractBody (object, keepalive = false) {
...
             prefixLength +
-            Buffer.byteLength(`; name="${escape(normalizeLinefeeds(name))}"`) +
-            Buffer.byteLength(`\r\n\r\n${normalizeLinefeeds(value)}\r\n`)
+            Buffer.byteLength(`; name="${escape(normalizeLinefeeds(name))}"\r\n\r\n${normalizeLinefeeds(value)}\r\n`)
         } else {
           bodyLength +=
             prefixLength +
-            Buffer.byteLength(`; name="${escape(normalizeLinefeeds(name))}"`) +
-            (value.name ? Buffer.byteLength(`; filename="${escape(value.name)}"`) : 0) +
+            Buffer.byteLength(`; name="${escape(normalizeLinefeeds(name))}"` + value ? `; filename="${escape(value.name)}"` : '') +
```
KhafraDev, line 158:
```suggestion
            Buffer.byteLength(`; name="${escape(normalizeLinefeeds(name))}"` + value.name ? `; filename="${escape(value.name)}"` : '') +
```

To match the current behavior.

lib/fetch/body.js
```diff
@@ -445,7 +441,6 @@ function bodyMixinMethods (instance) {
             }
             text += textDecoder.decode(chunk, { stream: true })
           }
-          text += textDecoder.decode()
```
KhafraDev, line 448: this is needed to flush the queue

KhafraDev (changes requested):

lib/fetch/body.js
anonrig: Ah, true! Thanks.

KhafraDev (approved):

jimmywarting:
It's good that you want to optimize things... but I don't know how much this micro optimization will benefit... 
seems like a small win...

if we really want to optimize something like 
```diff
- // value is a Blob or File
- bodyLength += value.size
- bodyLength += 2 // \r\n

into:

+ // value is a Blob or File, and \r\n
+ bodyLength += value.size + 2
```

Then maybe we should just minify all of our code and bundle everything into one file instead. 
And then i'm not talking about using a minifier like uglify, tensor, or anything that ships with typescript or rollup. I'm talking about going full on valid jsdoc and then using closure compiler with advance mode turned on... it will not just minify your code but also increase the speed of it, by inlining functions and evaluating what an piece of section dose...

for example this pice of code: 
```js
function unusedFunction(note) {
  alert(note['text']);
}

function displayNoteTitle(note) {
  alert(note['title']);
}

var flowerNote = {};
flowerNote['title'] = "Flowers";
displayNoteTitle(flowerNote);
```
would normally turn into something like this (with common minifiers)
```js
function unusedFunction(a){alert(a.text)}function displayNoteTitle(a){alert(a.title)}var flowerNote={};flowerNote.title="Flowers";displayNoteTitle(flowerNote);
```
while with adv mode you would get:
```js
alert("Flowers");
```

mcollina (approved):
lgtm

ronag (approved):

PR #1793, opened 2022-11-30
fetch: prefer global over lazy loading

KhafraDev:
ReadableStream and TransformStream were added as globals in v18, this PR uses them if available, and falls back to lazy loading them. This pattern is already used elsewhere in the code:

https://github.com/nodejs/undici/blob/232905f476913da9c73df03bac41a0a96d48b128/lib/fetch/util.js#L853


mcollina (approved):
lgtm

PR #1795, opened 2022-12-01
feat: implement websockets

### LINKED ISSUE #932: WebSocket
https://fetch.spec.whatwg.org/#websocket-protocol

KhafraDev:
Fixes #932 
Refs: https://github.com/nodejs/node/issues/19308

- [x] Initial handshake
- [x] Integrating with fetch (`await fetch('wss://whatever')`)
- [x] Setting websocket state
- [x] Implementing WebSocket events
- [x] Implementing WebSocket getters
- [x] Sending data to the server ~~& buffering~~ **edit:** since the socket is a duplex, buffering should be handled already for us
- [ ] [pings and pongs](https://websockets.spec.whatwg.org/#ping-and-pong-frames) **edit:** https://github.com/whatwg/websockets/issues/10
- [x] [CloseEvent](https://websockets.spec.whatwg.org/#the-closeevent-interface), [MessageEvent](https://html.spec.whatwg.org/multipage/comms.html#messageevent), and ErrorEvent
- [x] [web-platform-tests](https://github.com/web-platform-tests/wpt/tree/master/websockets)
- [x] diagnostics channel support (opening connection, closing, receiving a message)
- [ ] permessage-deflate support
- [ ] https://github.com/nodejs/undici/pull/1795#discussion_r1038928224

---

To test it out:

```js
import { WebSocket } from './lib/websocket/websocket.js'

const ws = new WebSocket('wss://echo.websocket.events/')

ws.addEventListener('message', (event) => {
  console.log(event.data) // "echo.websocket.events sponsored by Lob.com"

  ws.send('my message')
  
  setTimeout(() => ws.close(), 2000)
})

ws.addEventListener('close', ({ code }) => {
  console.log(code) // 1005
})

```


lib/core/request.js
```diff
@@ -307,21 +307,11 @@ function processHeader (request, key, val) {
...
     key.length === 10 &&
     key.toLowerCase() === 'keep-alive'
   ) {
     throw new InvalidArgumentError('invalid keep-alive header')
-  } else if (
-    key.length === 7 &&
-    key.toLowerCase() === 'upgrade'
-  ) {
-    throw new InvalidArgumentError('invalid upgrade header')
   } else if (
```
ronag, line 315: Not sure about these... why are they required?

KhafraDev:
```
   5.   The request MUST contain an |Upgrade| header field whose value
        MUST include the "websocket" keyword.

   6.   The request MUST contain a |Connection| header field whose value
        MUST include the "Upgrade" token.
```

lib/websocket/websocket.js
```diff
@@ -218,25 +235,72 @@ class WebSocket extends EventTarget {
...
+      // be buffered but the buffer is full, the user agent must flag the
+      // WebSocket as full and then close the WebSocket connection. The data
+      // to be sent is the raw data represented by the Blob object. Any
+      // invocation of this method with a Blob argument that does not throw
+      // an exception must increase the bufferedAmount attribute by the size
+      // of the Blob object’s raw data, in bytes.
+
+      frame.opcode = opcodes.BINARY
+
+      data.arrayBuffer().then((ab) => {
```
lpinca, line 297: If you have concurrent writes this might change the order of messages:

```js
websocket.send(blob);
websocket.send(string);
```

The string message is sent before the blob.

lib/core/request.js
ronag: The client adds them automatically through upgrade option

lib/websocket/connection.js
```diff
@@ -0,0 +1,413 @@
...
+    mode: 'websocket',
+    credentials: 'include',
+    cache: 'no-store',
+    redirect: 'error'
+  })
+
+  // 3. Append (`Upgrade`, `websocket`) to request’s header list.
+  // TODO: what if the headersList already has this header?
+  // in the browser, this header is filtered, but here it isn't.
+  request.headersList.append('upgrade', 'websocket')
```
jodevsa, line 52: Hey, do we need this? currently we are sending 2 upgrade headers:

Screenshot 2022-12-04 at 15 13 44
de headers:

KhafraDev: Weird, I'm unsure why this happens (the `upgrade` header was previously a forbidden header). Maybe setting `upgrade: websocket` in the dispatcher option sets this header?

KhafraDev: Yes it does. https://github.com/nodejs/undici/blob/68c269c4144c446f3f1220951338daef4a6b5ec4/lib/client.js#L1397

lib/core/request.js
KhafraDev: Reverting these changes.

jodevsa:
Hi, is this statement correct:
The message event should only be fired when: 

1- we have a complete un-fragmented frame (so, even if we got multiple chunks with some data.. we still don't want to send that to the user until we get the whole frame?)
2- we have a complete message that consists of multiple complete fragmented frames

KhafraDev:
Yes,

```
   If the frame is part of
   a fragmented message, the "Application data" of the subsequent data
   frames is concatenated to form the /data/.  When the last fragment is
   received as indicated by the FIN bit (frame-fin), it is said that _A
   WebSocket Message Has Been Received_ with data /data/ (comprised of
   the concatenation of the "Application data" of the fragments)
```

When [a WebSocket message has been received](https://datatracker.ietf.org/doc/html/rfc6455#page-66:~:text=_A%20WebSocket%20Message%20Has%20Been%20Received_) with type type and data data,
...
[Fire an event](https://dom.spec.whatwg.org/#concept-event-fire) named message at the [WebSocket](https://websockets.spec.whatwg.org/#websocket) object, using [MessageEvent](https://html.spec.whatwg.org/multipage/comms.html#messageevent), with the [origin](https://html.spec.whatwg.org/multipage/comms.html#dom-messageevent-origin) attribute initialized to the [serialization](https://url.spec.whatwg.org/#concept-url-serializer) of the [WebSocket](https://websockets.spec.whatwg.org/#websocket) object’s [url](https://url.spec.whatwg.org/#concept-url)'s [origin](https://url.spec.whatwg.org/#concept-url-origin), and the [data](https://html.spec.whatwg.org/multipage/comms.html#dom-messageevent-data) attribute initialized to dataForEvent.

jimmywarting:
Just wonder why `node:buffer` is being used so much... why not just use Uint8Arrays? 

ref: https://github.com/nodejs/node/issues/41588

KhafraDev:
Nothing that uses Buffer is exposed, so I don't think it's an issue. The [data](https://nodejs.org/api/net.html#event-data) event sends us a buffer, so it's easiest to use buffers here.

KhafraDev:
Believe it or not, just about every wpt for websockets is running. Out of 137 tests, only 7 are currently being skipped.
- 2 rely on features Node doesn't have (XMLHttpRequest and Web workers)
- 3 rely on sending > 65k bytes
- 2 fail because I haven't implemented the routes in the WebSocketServer yet.

Once payloads > 65k bytes can be sent, I think this can actually land. Of course there's still a few bugs, but it can be improved/fixed over time, similar to fetch. Similarly, fetch landed while being almost nothing like it is today, and having quite a few bugs.

lib/websocket/websocket.js
KhafraDev: I don't know of any good way of fixing this without needing a Promise for every send.. Preferably node would allow writing a blob to a writable stream.

lpinca: This is the reason why there is no `Blob` support in `ws`. This and because the user can already do

```js
websocket.send(await blob.arrayBuffer());
```

I agree with https://github.com/ricea/websocketstream-explainer/blob/832f88d7f8eaecbe51da318d802951a3f8aadb2f/README.md?plain=1#L152-L158

KhafraDev: Yes, when I was writing out my comment is kinda just clicked - "oh... this is why ws doesn't support blobs", lol. I still think undici should implement it for the sake of complying with the spec.

When I get around to implementing websocketstream this shouldn't be a problem at least!

lpinca: Well, `ws` already uses a queue (only when `permessage-deflate` is enabled and negotiated since compression is asynchronous) so support for `Blob` could be added, but I think it does not worth the effort. I think that `Blob` is of little use in Node.js if not natively supported by streams.

lib/websocket/websocket.js
```diff
@@ -0,0 +1,534 @@
...
+      // of the Blob object’s raw data, in bytes.
+
+      const frame = new WebsocketFrameSend()
+
+      data.arrayBuffer().then((ab) => {
+        const value = Buffer.from(ab)
+        frame.frameData = value
+        const buffer = frame.createFrame(opcodes.BINARY)
+
+        this.#bufferedAmount += value.byteLength
```
lpinca, line 326: Is `bufferedAmount` ever decreased? I can't see where.

lib/websocket/websocket.js
jimmywarting: hmm, but `send(await blob.arrayBuffer())` would require reading the hole blob into memory before sending...
sure you could do something like: 
```js
for await (const chunk of blob.stream()) {
  websocket.send(chunk)
}
```
But then you would get multiple messages and you would have to assemble them yourself on the receiving side as well 😞 

if you can accept blobs then it could as well be piped to a tmp location on the disc and wouldn't hold up any memory when receiving blobs

jimmywarting: Ah, just went into the github repo and learned about a new `WebSocketStream` forgot that it existed, only head about it when it was in early proposal. Well, then you could instead maybe do: `blob.stream().pipeTo(websocket)`?

lib/websocket/websocket.js
KhafraDev: No it isn't, meant to fix it and forgot about it.

lib/websocket/websocket.js
lpinca: > But then you would get multiple messages and you would have to assemble them yourself on the receiving side as well 😞

That's up you. With `ws` you could send each chunk as a message fragment and receive the whole thing as a single message on the receiving end.

```js
for await (const chunk of blob.stream()) {
  websocket.send(chunk, { fin: false}); // For the sake of simplicity backpressure is not handled.
}
websocket.send(Buffer.alloc(0), { fin: true });
```

> if you can accept blobs then it could as well be piped to a tmp location on the disc and wouldn't hold up any memory when receiving blobs

AFAIK the whole message is buffered in memory before the `'message'` event is emitted. That message is then returned to the user as a `Blob`.

lib/websocket/connection.js
```diff
@@ -0,0 +1,365 @@
...
+      // a single frame with the FIN bit set and an opcode of 0.
+
+      const fragmentedFrame = WebsocketFrame.from(chunk)
+
+      if (fragmentedFrame.fin && fragmentedFrame.opcode === opcodes.CONTINUATION) {
+        frame.terminated = true
+      }
+
+      frame.addFrame(fragmentedFrame.data)
+    }
```
lpinca, line 258: At a glance this does not handle the case where a control frame (ping/pong) is received in the middle of a fragmented message. If it is, ignore this comment.

KhafraDev: I had no idea this could happen, I will fix it.

lib/websocket/connection.js
```diff
@@ -0,0 +1,365 @@
...
+      }
+
+      // 2. Let dataForEvent be determined by switching on type and binary type:
+      let dataForEvent
+
+      if (frame.opcode === opcodes.TEXT) {
+        // - type indicates that the data is Text
+        //      a new DOMString containing data
+        dataForEvent = new TextDecoder().decode(frame.data)
+      } else if (frame.opcode === opcodes.BINARY) {
```
lpinca, line 294: You can also receive a control (ping/pong) frame. You can ignore them but you don't want to return them as messages via the `'message'` event to the user.

Also you should handle pings. A peer must respond with a pong when it receives a ping. Yes, browsers also do this.

lib/websocket/connection.js
lpinca: As per https://github.com/KhafraDev/undici/pull/30#discussion_r1039009237, another problem I see is that the WHATWG spec advertises support for the `permessage-deflate` extension. If the server you are connecting to also supports `permessage-deflate`, you might receive frames whose payload data is compressed.

KhafraDev: I don't know if I'll implement permessage deflate support in this pr; the ping/pong issue has been fixed.

lpinca: Ok, then I think it would be better to not send the `Sec-WebSocket-Extensions` header. The connection would be surprisingly closed every time the server sends a compressed message. Plus there is currently no validation and negotiation of the server sent `Sec-WebSocket-Extensions` header. If you enable `permessage-deflate` on the server, some of web-platform-tests fails.

lib/websocket/connection.js
```diff
@@ -0,0 +1,409 @@
...
+        frame.terminated = true
+      }
+
+      frame.addFrame(receivedFrame.data)
+    }
+
+    // If a control frame (Section 5.5) is
+    // received, the frame MUST be handled as defined by Section 5.5.
+    if (frame.opcode === opcodes.PING) {
+      handlePing(response.socket, ws)
```
lpinca, line 264: This should be done after all frame data is read because the pong frame must have the same data of the ping frame being replied to.

lib/websocket/connection.js
```diff
@@ -0,0 +1,409 @@
...
+
+  /** @type {WebsocketFrame|undefined} */
+  let frame
+
+  response.socket.on('data', (chunk) => {
+    const receivedFrame = WebsocketFrame.from(chunk)
+
+    if (!frame) {
+      frame = receivedFrame
+      frame.terminated = frame.fin // message complete
```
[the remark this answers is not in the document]
lpinca: Also, I get that this is a client and the assumption is that the server is correctly implemented but nothing prevents a client from connecting to a fake/evil server. The connection should be closed if:

- The opcode is 0 and the message is not fragmented.
- The opcode is 1 or 2 and the message is fragmented.
- The opcode is in the range [0x8 - 0x0a] (control frame) and the `FIN` bit is not set.
- The opcode is in the range [0x8 - 0x0a] and the payload length is bigger than 127 bytes.
- The opcode is 8 and the payload length is 1.
- The opcode is none of the above.

lib/websocket/connection.js
```diff
@@ -0,0 +1,409 @@
...
+        return
+      }
+
+      // 2. Let dataForEvent be determined by switching on type and binary type:
+      let dataForEvent
+
+      if (frame.opcode === opcodes.TEXT) {
+        // - type indicates that the data is Text
+        //      a new DOMString containing data
+        dataForEvent = new TextDecoder().decode(frame.data)
```
lpinca, line 306: Connection should be closed if `frame.data` is not valid UTF-8.

lib/websocket/connection.js
```diff
@@ -0,0 +1,409 @@
...
+ * @param {import('./websocket').WebSocket} ws
+ */
+function socketClosed (ws) {
+  const { [kResponse]: response } = ws
+
+  response.socket.on('close', () => {
+    const wasClean = ws[kReadyState] === states.CLOSING || isEstablished(ws)
+
+    /** @type {ReturnType<WebsocketFrame['parseCloseBody']} */
+    const { code, reason } = ws[kClosingFrame]
```
lpinca, line 348: `code` must be a valid status code. `reason` must be valid UTF-8.

lib/websocket/connection.js
```diff
@@ -0,0 +1,409 @@
...
+    } else if (frame.opcode === opcodes.PONG) {
+      handlePong(frame)
+
+      return
+    } else if (frame.opcode === opcodes.CLOSE) {
+      // Upon either sending or receiving a Close control frame, it is said
+      // that _The WebSocket Closing Handshake is Started_ and that the
+      // WebSocket connection is in the CLOSING state.
+      ws[kReadyState] = states.CLOSING
+      ws[kClosingFrame] = frame
```
lpinca, line 276: When you receive a close frame, you should send back another close frame. The connection should be closed only when both peers have sent and received a close frame.

lib/websocket/connection.js
KhafraDev: I've disabled it

lib/websocket/connection.js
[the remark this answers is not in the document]
KhafraDev: Currently not much validation is performed; just the most basic outline of a working implementation. Thanks for that list - it really is helpful.

lib/websocket/connection.js
KhafraDev: The implementation here is currently lacking - the server doesn't necessarily need to send the same data it's responding to, just the data from the latest ping it received. The server *can* respond to both pings, but it's not required.

```
   If an endpoint receives a Ping frame and has not yet sent Pong
   frame(s) in response to previous Ping frame(s), the endpoint MAY
   elect to send a Pong frame for only the most recently processed Ping
   frame.
```

Also, a server doesn't need to send one right away, just when it's "practical" to do so. I was thinking of adding a ping queue with a ms timestamp and deleting all earlier pings, but as you said, an evil/bad server just might not send any pongs.

lpinca: I was referring to this https://www.rfc-editor.org/rfc/rfc6455#section-5.5.3

> A Pong frame sent in response to a Ping frame must have identical
   "Application data" as found in the message body of the Ping frame
   being replied to.

You can't send a pong without reading the ping data.

KhafraDev: Currently we send an empty ping frame so the payload length of the pong frame must be 0 (validation is done to check). Not the greatest solution ever, but it should be okay?

KhafraDev: Interesting, I was confusing "endpoint" in the rfc with "server"

lib/websocket/connection.js
KhafraDev: Thanks! This actually fixed an issue where the wpt runner would take ~35 seconds to run (I'm assuming the server was staying open because it wasn't receiving a close frame). Now it takes 4 seconds to run.

lib/websocket/connection.js
KhafraDev: Is there an easy way to validate utf-8 without taking ws' implementation? Code is easy enough.

lpinca: There is https://github.com/websockets/utf-8-validate, https://github.com/micnic/uv, https://github.com/hcodes/isutf8.

lpinca: > Or hopefully we could eventually have an implementation inside of node, similar to toUSVString, which would benefit anyone using websockets.

A native implementation in core could use https://github.com/simdutf/simdutf. This is used by Bun and is the state-of-art in terms of performance, see https://github.com/websockets/utf-8-validate/pull/101. I did not merge that PR because the library has a lot of unneeded features. It would be great to add compile time flags to create a binary only with the wanted features.

KhafraDev: There's a possibility of simdutf getting added as a dependency to node, I'm talking to anonrig and he already has a fork with it added. He just didn't have a reason to add it in node.

anonrig: @lpinca @KhafraDev I opened the PR. Appreciate, if you can share your reasoning the pull request so that TSC members can have a better understanding of the need. https://github.com/nodejs/node/pull/45803

lib/websocket/connection.js
KhafraDev: Just wondering, how does ws handle pongs? What if the user pings twice and only receives a pong for the last ping? What if a server sends pongs for both? I have an idea of how to fix it, but it seems as though ws 

> the endpoint MAY
   elect to send a Pong frame for only the most recently processed Ping
   frame.

```js
const ws = new WebSocket('...')

ws.ping('...')
ws.ping('...')

// assume that the server either responds to both pings, or only 1 of them

```

lpinca: > What if the user pings twice and only receives a pong for the last ping? 

That's ok, it does not matter, pongs need no special handling.

> What if a server sends pongs for both?

Same answer.

> I have an idea of how to fix it, but it seems as though ws handles this much easier than I'm thinking. Am I overthinking this? Am I overthinking this?

I think so, `ws` responds to all pings, not just the most recent one. If the implementation sends a pong for every received ping there is no `If an endpoint receives a Ping frame and has not yet sent Pong frame(s) in response to previous Ping frame(s)` case.

KhafraDev: > pongs need no special handling

I guess that's where my issue was. I was assuming that I'd need to validate the application data of the pong frame, to ensure it was replying to a ping we sent. Although it gets complicated since pong frames could also be unsolicited.

test/wpt/status/websockets.status.json
```diff
@@ -0,0 +1,20 @@
...
+    "skip": true
+  },
+  "bufferedAmount-unchanged-by-sync-xhr.any.js": {
+    "skip": true,
+    "//": "Node.js doesn't have XMLHttpRequest nor does this test make sense regardless"
+  },
+  "referrer.any.js": {
+    "skip": true
+  },
+  "Send-65K-data.any.js": {
```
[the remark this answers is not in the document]
lpinca: > it seems more like an issue with ws to me, but I am really unsure

I doubt. Run it with the `ws` client and see if you can reproduce.

lib/websocket/connection.js
lpinca: I've just remembered that there is `TextDecoder` now in Node.js. It is possible to use it to "validate" UTF-8.

```js
const decoder = new TextDecoder('utf-8', { fatal: true });

try {
  decoder.decode(Buffer.from([0xf8]));
} catch {
  console.log('Invalid UTF-8');
}
```

It can also be used in a streaming fashion which is what the spec recommends.

```js
const decoder = new TextDecoder('utf-8', { fatal: true });

const chunks = [
  0xf0, 0x90, 0x8c, 0x88, // 𐍈
  0xe0, 0x80, 0xa0, // Overlong
  0xe2, 0x82, 0xac  // €
];

let string = ''

for (let i = 0; i < chunks.length; i++) {
  const chunk = chunks[i];
  const isLastChunk = i === chunks.length - 1;

  try {
    string += decoder.decode(
      Buffer.from([chunk]),
      { stream: !isLastChunk }
    );

    if (isLastChunk) console.log(string);
  } catch {
    console.log('Invalid UTF-8 detected at index %d', i + 1);
    break;
  }
}
```

The only problem is that the `fatal` option is not supported when ICU is disabled.

test/wpt/status/websockets.status.json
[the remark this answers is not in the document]
KhafraDev: I was calculating the payload length incorrectly, reading the entire byte instead of the 7 bits the length is contained in. Now, I have a different problem with a different test, and still only on linux. It's really weird that windows has no issues.

lib/websocket/connection.js
KhafraDev: TIL the fatal option exists.

lib/websocket/connection.js
```diff
@@ -0,0 +1,504 @@
...
+ * @param {import('./websocket').WebSocket} ws
+ */
+function receiveData (ws) {
+  const { [kResponse]: response } = ws
+
+  /** @type {WebsocketFrame|undefined} */
+  let frame
+
+  response.socket.on('data', (chunk) => {
+    const receivedFrame = WebsocketFrame.from(chunk)
```
lpinca, line 259: I'm pretty sure this still crashes if `chunk` is `Buffer.from([0x82, 0x7F])` as you assume that chunks can't be smaller than 8 bytes. That's a wrong assumption. Keep in mind that you can also receive a single byte so you have to wait until you have at least 2 bytes before reading the payload length, otherwise `buffer[1] & 0x7F` gives you 0 when it could actually be 127.

lib/websocket/connection.js
```diff
@@ -0,0 +1,551 @@
...
+            return
+          }
+        }
+      }
+    }
+
+    let receivedFrame
+
+    if (partialFrame.payloadLength !== undefined) {
+      receivedFrame = WebsocketFrame.from(Buffer.concat(partialFrame.chunks, partialFrame.length))
```
lpinca, line 299: What happens if you receive a chunk like `82008200` that is two empty frames in the same data chunk? I did not test it but it does not seem correct.

KhafraDev: to be honest I'd prefer to fix this in a followup pr. Behavior like this doesn't seem common and there's currently a lot of "weird behavior" things we aren't testing (the wpts are extremely lacking).

KhafraDev:
Fetch landed with 3 tests and eventually node-fetch's suite was ported. I think this PR is complete enough to land as a *very* basic websocket implementation that can be improved on over time.

lib/websocket/util.js
```diff
@@ -0,0 +1,128 @@
...
+      char === ':' ||
+      char === '\\' ||
+      char === '"' ||
+      char === '/' ||
+      char === '[' ||
+      char === ']' ||
+      char === '?' ||
+      char === '=' ||
+      char === '{' ||
+      char === '}' ||
```
anonrig, line 94: Would storing these values in a string, and later doing a `indexOf` operation be faster?

lib/fetch/index.js
```diff
@@ -2060,6 +2070,30 @@ async function httpNetworkFetch (
...
+        onUpgrade (status, headersList, socket) {
+          if (status !== 101) {
+            return
+          }
+
+          const headers = new Headers()
+
+          for (let n = 0; n < headersList.length; n += 2) {
+            const key = headersList[n + 0].toString('latin1')
+            const val = headersList[n + 1].toString('latin1')
```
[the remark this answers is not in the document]
KhafraDev: (talked about this in dms)

lib/websocket/frame.js
```diff
@@ -0,0 +1,222 @@
...
+    data
+  } = {}) {
+    this.fin = fin
+    this.rsv1 = rsv1
+    this.rsv2 = rsv2
+    this.rsv3 = rsv3
+    this.opcode = opcode
+
+    this.#buffers.push(data)
+    this.#offset += data.length
```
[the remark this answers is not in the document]
KhafraDev: doesn't really matter, as it's initialized with a value of 0

lib/websocket/frame.js
```diff
@@ -0,0 +1,222 @@
...
+    if (reason[0] === 0xEF && reason[1] === 0xBB && reason[2] === 0xBF) {
+      reason = reason.subarray(3)
+    }
+
+    if (code !== undefined && !isValidStatusCode(code)) {
+      return null
+    }
+
+    try {
+      reason = new TextDecoder('utf-8', { fatal: true }).decode(reason)
```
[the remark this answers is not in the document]
KhafraDev: The reason here can only be <= 125 bytes, I'll add a comment where the body being received is validated instead (the payload can be much larger)

lib/websocket/frame.js
```diff
@@ -0,0 +1,222 @@
...
+    if (payloadLength === 126) {
+      buffer.writeUInt16BE(bodyLength, 2)
+    } else if (payloadLength === 127) {
+      buffer.writeUIntBE(bodyLength, 4, 6)
+    }
+
+    buffer[1] |= 0x80 // MASK
+
+    // mask body
+    for (let i = 0; i < this.frameData?.length ?? 0; i++) {
```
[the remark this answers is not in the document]
KhafraDev: A frame doesn't necessarily need to have a body, for example a close frame could just consist of `0x88 0x00`.

lib/websocket/util.js
KhafraDev: This pattern is used frequently (see lib/fetch/util.js) and I believe all subprotocols must be <= 125 bytes in length.

anonrig: In that scenario, instead of comparing string to string, comparing number to number would be faster.

KhafraDev: This validation is only used if the user passed protocols to the WebSocket constructor (so it only happens when a connection is being opened). I don't think this needs to be the most performant check ever because WebSocket connections typically aren't opened very often.

```js
const ws = new WebSocket('wss://') // no validation

const ws = new WebSocket('wss://', ['chat']) // validation
```

anonrig: 👍

anonrig (approved):

ronag (commented):
Docs? With a note that it's experimental.

KhafraDev:
It emits an experimental warning on first use if that's good enough. If not I'll add docs tomorrow.

ronag:
Please add note in docs

lib/websocket/frame.js
```diff
@@ -0,0 +1,223 @@
...
```
ronag, line 212: If mask is 0 then skip all of this. It's a common optimisation to set mask to 0 for this reason.

lib/websocket/connection.js
lpinca: Your call, but trust me, this is common. Not two empty frames in the same data chunk, that is just an example for easy reproduction, but multiple frames in the same data chunk and parts of them. A working frame parser is a basic thing, otherwise you might miss messages, leak memory and sockets, etc. It's like shipping HTTP with a broken HTTP parser.

Try to subscribe to the websocket feed of a popular cryptocurrency exchange:

```js
const ws = new WebSocket('wss://ws-feed.exchange.coinbase.com');

ws.addEventListener('open', function () {
  ws.send(
    JSON.stringify({
      type: 'subscribe',
      product_ids: ['BTC-USD'],
      channels: ['full']
    })
  );
});

ws.addEventListener('message', function (event) {
  console.log(event.data);
});
```

lib/websocket/frame.js
```diff
@@ -0,0 +1,223 @@
...
+
+    this.#buffers = [data]
+    this.#offset += data.length
+
+    this.payloadLength = payloadLength
+    this.fragmented = !this.fin && this.opcode !== opcodes.CONTINUATION
+  }
+
+  get data () {
+    return Buffer.concat(this.#buffers, this.#offset)
```
ronag, line 68: if buffers.length === 1 we don't need to make a copy? Also should we memoize this once it is done?

lib/websocket/frame.js
```diff
@@ -0,0 +1,223 @@
...
+    // Data sent from an endpoint cannot be masked.
+    assert(!masked)
+
+    let payloadLength = buffer[1] & 0x7F
+    let data
+
+    if (payloadLength <= 125) {
+      data = buffer.subarray(2)
+    } else if (payloadLength === 126) {
+      payloadLength = buffer.subarray(2, 4).readUInt16BE(0)
```
ronag, line 145: This can be optimized. Add TODO?

lib/websocket/frame.js
```diff
@@ -0,0 +1,223 @@
...
+      data = buffer.subarray(4)
+    } else if (payloadLength === 127) {
+      payloadLength = buffer.subarray(2, 10).readBigUint64BE(0)
```
ronag, line 148: This can be optimized. Add TODO?

lib/websocket/frame.js
```diff
@@ -0,0 +1,223 @@
...
+
+    if (bodyLength > maxUnsigned16Bit) {
+      offset += 8 // payload length is next 8 bytes
+      payloadLength = 127
+    } else if (bodyLength > 125) {
+      offset += 2 // payload length is next 2 bytes
+      payloadLength = 126
+    }
+
+    const buffer = Buffer.alloc(bodyLength + offset)
```
ronag, line 190:
```suggestion
    const buffer = Buffer.allocUnsafe(bodyLength + offset)
```

lib/websocket/frame.js
```diff
@@ -0,0 +1,223 @@
...
+      return { code }
+    }
+
+    // https://datatracker.ietf.org/doc/html/rfc6455#section-7.1.6
+    /** @type {Buffer} */
+    let reason = data.subarray(2)
+
+    // Remove BOM
+    if (reason[0] === 0xEF && reason[1] === 0xBB && reason[2] === 0xBF) {
+      reason = reason.subarray(3)
```
ronag, line 108: Faster subarray:

```
const FastBuffer = Buffer[Symbol.species]

reason = new FastBuffer(reason.buffer, 3)

```

lib/websocket/frame.js
```diff
@@ -0,0 +1,223 @@
...
+    assert(!masked)
+
+    let payloadLength = buffer[1] & 0x7F
+    let data
+
+    if (payloadLength <= 125) {
+      data = buffer.subarray(2)
+    } else if (payloadLength === 126) {
+      payloadLength = buffer.subarray(2, 4).readUInt16BE(0)
+      data = buffer.subarray(4)
```
ronag, line 146: This can be optimised further. TODO?

lib/websocket/frame.js
```diff
@@ -0,0 +1,223 @@
...
+    } else if (payloadLength === 127) {
+      payloadLength = buffer.subarray(2, 10).readBigUint64BE(0)
+      data = buffer.subarray(10)
```
ronag, line 149: This can be optimised further. TODO?

lib/websocket/frame.js
```diff
@@ -0,0 +1,223 @@
...
+  }
+}
+
+class WebsocketFrameSend {
+  /**
+   * @param {Buffer|undefined} data
+   */
+  constructor (data) {
+    this.frameData = data
+    this.maskKey = randomBytes(4)
```
ronag, line 172: This can be slow and often unnecessary... is there a way we can force 0 mask?

lib/websocket/frame.js
lpinca: It does not seem to be much faster:

```
$ cat index.js
'use strict';

const benchmark = require('benchmark');

const FastBuffer = Buffer[Symbol.species];
const buf = Buffer.alloc(200).fill('a');

let res;

suite.add('Buffer#subarray()', function () {
  res = buf.subarray(100);
});

suite.add('FastBuffer', function () {
  res = new FastBuffer(buf.buffer, buf.byteOffset + 100, buf.length - 100);
});

suite.on('cycle', function (event) {
});

suite.on('complete', function () {
  console.log('Fastest is ' + this.filter('fastest').map('name'));
});

suite.run({ async: true });
```

```
$ node -v
v18.12.1
```

```
$ node index.js
Buffer#subarray() x 22,040,974 ops/sec ±2.21% (86 runs sampled)
FastBuffer x 22,715,035 ops/sec ±2.60% (85 runs sampled)
Fastest is FastBuffer
```

ronag: but it is a little faster :D

lib/websocket/frame.js
KhafraDev: Setting mask to 0 makes it a continuation frame, basically sending a fragmented message. We currently don't support support sending fragmented messages, and even if we did, we would still need to mask the body.

lib/websocket/frame.js
KhafraDev: I've ensured that this getter is never used more than once per code path, I'll add a fast path for 1 buffer.

lib/websocket/frame.js
lpinca: @ronag is suggesting skip masking if the mask is `00000000`.

lib/websocket/frame.js
KhafraDev: I will look into all the subarray optimizations in a future pr

lib/websocket/frame.js
KhafraDev: Unsure of what you mean - it's possible to create the key lazily, but it has to be secure.

> Clients MUST choose a new masking key for each frame, using an
   algorithm that cannot be predicted by end applications that provide
   data.  For example, each masking could be drawn from a
   cryptographically strong random number generator.

ronag: The mask is not useful/necessary for server side websocket applications and a significant performance bottleneck. Without the possibility for the 0 mask optimization, this implementation would not be useful at least for us. The only time the mask is useful and the reason it exists is for when you non trusted proxies in the path.

KhafraDev: I see now. I'd need to give it more thought as it goes against everything the spec tells us to do. Does ws have a similar ability?

lpinca: Yes, but it uses `crypto.randomBytes()` by default.

lib/websocket/connection.js
KhafraDev: Thanks! We aren't leaking memory/sockets, but messages are being missed. I need to move the implementation to something that supports buffering & reading the body only until the application data length.

lib/websocket/frame.js
ronag: >> Does ws have a similar ability?

Yes, @lpinca was kind enough to add the option.

```js
new ws.WebSocket(url, {
   generateMask() {},
})
```

KhafraDev: This would be better off in its own PR so it can be fleshed out more. The spec WebSocket takes 2 arguments - neither of which is an object - so it's not as easy to add custom things like it is with fetch. Also considering how it violates the rfc I'd need to give it much more thought as mentioned above.

This PR is fine for implementing just a bare-bones version of WebSockets, so we can add in more stuff later.

KhafraDev:
I'm working on a better parsing system in order to fix https://github.com/nodejs/undici/pull/1795#discussion_r1045471864 and other issues.

lib/websocket/connection.js
KhafraDev: Okay, this is finally fixed. I took a lot of inspiration from ws (which I am becoming ever more grateful for) so everything is working pretty smoothly. Still needs validation and a lot of work, but much better than the old solution.

This should also fix the issue where sending < 2 bytes would cause an issue.

lib/websocket/connection.js
```diff
@@ -256,7 +256,10 @@ function socketClosed (ws) {
   const { [kResponse]: response } = ws
 
   response.socket.on('close', () => {
-    const wasClean = ws[kReadyState] === states.CLOSING || isEstablished(ws)
+    // If the TCP connection was closed after the
+    // WebSocket closing handshake was completed, the WebSocket connection
+    // is said to have been closed _cleanly_.
+    const wasClean = ws[kSentClose]
```
lpinca, line 262: You are probably already aware of this but receiving the socket `'close'` event does not mean that the closing handshake was completed.

lib/websocket/receiver.js
```diff
@@ -0,0 +1,223 @@
...
+
+        websocketMessageReceived(this.ws, this.#info.opcode, this.#info.data)
+
+        this.#info = {}
+        this.#state = parserStates.INFO
+      }
+    }
+
+    if (this.#byteOffset > 0) {
+      return this.run(callback)
```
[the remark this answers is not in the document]
KhafraDev: Something would have to go horribly wrong for this to occur - even the coinbase feed (which sends quite a few messages fragmented/unfragmented/partial frames) works fine.

However I would still like to move this to a while loop like ws does, just to rely on recursion less. Probably a separate PR though.

lib/websocket/connection.js
KhafraDev: I was on the fence about where setting this was. Thinking it over, it should have been set after the server receives the close frame.

lib/websocket/receiver.js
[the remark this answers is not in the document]
lpinca: https://github.com/websockets/ws/issues/990 and the fix https://github.com/websockets/ws/pull/992.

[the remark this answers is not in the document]
KhafraDev: > This makes the parser work with buffers containing thousands of frames.

oh.

lib/websocket/receiver.js
```diff
@@ -94,9 +94,13 @@ class ByteParser extends Writable {
           const closeFrame = new WebsocketFrameSend(body)
 
           this.ws[kResponse].socket.write(
-            closeFrame.createFrame(opcodes.CLOSE)
+            closeFrame.createFrame(opcodes.CLOSE),
+            (err) => {
+              if (!err) {
+                this.ws[kSentClose] = true
```
lpinca, line 100: 👍 but commit message is misleading:

1. It does not set `wasClean` but `Symbol('kSentClose')`.
2. When the callback is called it does not mean that the server received the frame.

lib/websocket/connection.js
```diff
@@ -0,0 +1,314 @@
...
+ * @param {import('./websocket').WebSocket} ws
+ */
+function socketClosed (ws) {
+  const { [kResponse]: response } = ws
+
+  response.socket.on('close', () => {
+    // If the TCP connection was closed after the
+    // WebSocket closing handshake was completed, the WebSocket connection
+    // is said to have been closed _cleanly_.
+    const wasClean = ws[kSentClose]
```
lpinca, line 262: A sent close frame is not sufficient. A close frame should also be received. It should be `ws[kSentClose] && ws[kReceivedClose]` assuming that `ws[kReceivedClose]` is set when a close frame is received.

lib/websocket/websocket.js
```diff
@@ -0,0 +1,553 @@
...
+        frame.frameData.write(reason, 2, 'utf-8')
+      } else {
+        frame.frameData = Buffer.alloc(0)
+      }
+
+      /** @type {import('stream').Duplex} */
+      const socket = this[kResponse].socket
+
+      socket.write(frame.createFrame(opcodes.CLOSE))
+      this[kSentClose] = true
```
lpinca, line 217:
```suggestion
      socket.write(frame.createFrame(opcodes.CLOSE), (err) => {
        if (!err) {
          this[kSentClose] = true
        }
      })
```

lib/websocket/connection.js
```diff
@@ -0,0 +1,315 @@
...
+      // once this happens, the connection is open
+      ws[kResponse] = response
+
+      const parser = new ByteParser(ws)
+      response.socket.ws = ws // TODO: use symbol
+      ws[kByteParser] = parser
+
+      whenConnectionEstablished(ws)
+
+      response.socket.on('data', onSocketData)
```
lpinca, line 188: I don't know if undici already adds an `'error'` event listener on the socket. If it does not one should be added.

lib/websocket/connection.js
```diff
@@ -0,0 +1,315 @@
...
+  //     processResponse given response being these steps:
+  const controller = fetching({
+    request,
+    useParallelQueue: true,
+    dispatcher: getGlobalDispatcher(),
+    processResponse (response) {
+      // 1. If response is a network error or its status is not 101,
+      //    fail the WebSocket connection.
+      if (response.type === 'error' || response.status !== 101) {
+        failWebsocketConnection(ws)
```
lpinca, line 101: I think no error is returned to the user. If so, without an `'error'` or `'open'` event, there is not much a user can do.

KhafraDev: I had no idea when an error event should be emitted, but reading over the rfc, an error event should be emitted whenever fail the WebSocket connection occurs.

KhafraDev: ... but we're not supposed to make the error message useful. However the reasoning is that by `[a]llowing a script to distinguish these cases would allow a script to probe the user’s local network in preparation for an attack.` I think I'll ignore this and give useful errors.

lib/websocket/receiver.js
```diff
@@ -0,0 +1,229 @@
+const { Writable } = require('stream')
```
mcollina, line 1: Use strict is missing

mcollina (approved):
good work here, apart from the outstanding comments this is ready for me

lib/websocket/connection.js
```diff
@@ -0,0 +1,321 @@
...
+      code,
+      reason
+    })
+  }
+}
+
+function onSocketError (error) {
+  const { ws } = this
+
+  fireEvent('error', ws, ErrorEvent, {
```
lpinca, line 314: This might cause multiple `'error'` events because there is another place where this is called. It does not cause issues in practice because `ws` is an `EventTarget` but I'm not sure if it is wanted. To match the behavior of browsers and the `http` module, in `websockets/ws` socket errors are not returned to the user.

KhafraDev: Added diagnostic channel support instead.

KhafraDev:
Any last minute issues to fix? I'd like to merge this by tonight and subsequently fix/add the remaining features over the next few days.

My next PRs (if this is merged, otherwise they'll be here):
- pings and pongs (very easy to add, I want to add tests though)
- tests (WPTs are "fine", but very lacking -- only 220 tests total)
- concurrently sends with a blob and string
- permessage-deflate support (likely not for a while)

Regarding the concurrent sends with a blob, we'd need core support for writing blobs to a writable stream. Then we could do:
```js
const frame = new WebsocketFrameSend(Buffer.allocUnsafe(0))
const buffer = frame.createFrame(opcodes.TEXT) // or binary

socket.cork()
socket.write(buffer)
socket.write(blob)
socket.uncork()

socket.write(string)
```

lpinca:
You have to mask the blob data so support for `Blob` in `Writable.prototype.write()` is not sufficient.

KhafraDev:
Right, completely forgot about masking. There would need to be a way of resolving blobs synchronously, which I doubt has much hope of landing in core.

lpinca:
> There would need to be a way of resolving blobs synchronously, which I doubt has much hope of landing in core.

When permessage-deflate will be implemented, if compression will be done asynchronously (there is currently no support for sync deflate streams in Node.js core) you will have the same problem. You can then use the same queue used for compressed messages for blobs.

KhafraDev:
Also found a way to read blobs synchronously, just took a deep dive into node internals. Obviously not going to implement it because it's a terrible hack lol.

```js
const blob = new Blob(['hello world'])

const handle = Reflect.ownKeys(blob)[0] // kHandle

console.log(blob[handle].toArrayBuffer())

```

anonrig (approved):

ronag (approved):

PR #1798, opened 2022-12-05
Add test for mock agent with fetch in jest

hansottowirtz:
(I'm also sorry for using jest)

MockAgent or setGlobalDispatcher seem to be bypassed when used with Node 18 native fetch or with fetch imported from undici, but only when running in Jest.

I added a simple test, but I'm not familiar with undici internals, any suggestions where I can look?

### DIFF
--- a/test/jest/mock-agent.test.js
+++ b/test/jest/mock-agent.test.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const { request, setGlobalDispatcher, MockAgent } = require('../..')
+const { request, setGlobalDispatcher, MockAgent, fetch } = require('../..')
 const { getResponse } = require('../../lib/mock/mock-utils')
 
 /* global describe, it, expect */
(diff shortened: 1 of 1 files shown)


mcollina:
I bet it's another case of https://github.com/facebook/jest/issues/2549 we need to work around :(.

silverwind:
At this point, I can not recommend using jest any more. If you can, try vitest which in simple cases is a drop-in replacement.

KhafraDev:
I seemed to have missed this issue. Here's how to fix this test:
```diff
diff --git a/test/jest/mock-agent.test.js b/test/jest/mock-agent.test.js
index 8e6966ff..977a2be3 100644
--- a/test/jest/mock-agent.test.js
+++ b/test/jest/mock-agent.test.js
@@ -48,15 +48,22 @@ describe('MockAgent', () => {
     expect.assertions(1)

     mockAgent = new MockAgent()
-    setGlobalDispatcher(mockAgent)
+    const mockPool = mockAgent.get('http://localhost:9999')
+
+    mockPool.intercept({
+      path: '/',
+      method: 'GET'
+    }).reply(200, 'hello')
+
+    setGlobalDispatcher(mockPool)

     let failed = false
     try {
       await fetch('http://localhost:9999')
     } catch (e) {
-      if (e.cause?.code === 'UND_MOCK_ERR_MOCK_NOT_MATCHED') failed = true
+      failed = true
     }

-    expect(failed).toBe(true)
+    expect(failed).toBe(false)
   })
 })
```

Issue #1806, opened 2022-12-14
request Does Not Support Multiple Headers


jd-carroll:
## Bug Description
Even though the type definitions [clearly](https://github.com/nodejs/undici/blob/main/types/dispatcher.d.ts#L103) state multiple headers (of the same name) are supported, it would appear they are not. Any time I try to set an [string] array value to a header an exception is thrown.

Note: Here are the types for `IncomingHttpHeaders `, non-single valued headers (ie. the ones explicitly listed) can appear multiple times in a request.
https://github.com/DefinitelyTyped/DefinitelyTyped/blob/878f77e0bbce48cc9c528484e525d7ad1432297c/types/node/http.d.ts#L47-L49

## Reproducible By
```typescript
const { request } = require('undici');
request('https://example.com', {
  method: 'GET',
  headers: {
    'X-Test-Header': ['a']
  }
}).then(() => console.log('success'))
.catch((err) => console.error(err));
```

## Expected Behavior
The request to be sent with the `X-Test-Header`.

## Environment
macOS 13.0.1
node: v18.12.1
undici: 5.14.0

### Additional context
The issue is this line
https://github.com/nodejs/undici/blob/f6323f67e64721e86b9c02b718c7d370b18ca25b/lib/core/request.js#L274-L276

Note: An array is an object! (`typeof [] === 'object' // true`)

I could probably work-up a fix for this, but at this point I am way too frustrated to deal with this any more (sorry).

**Important Note:** It is also more likely than not, that I am doing something wrong. So please tell me where I've gone awry 😎

PR #1807, opened 2022-12-15
fix(request): request headers array

jd-carroll:
Fix for #1806

### DIFF
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -279,7 +279,7 @@ function processHeaderValue (key, val) {
 }
 
 function processHeader (request, key, val) {
-  if (val && !(typeof val === 'string' || Array.isArray(val))) {
+  if (val && !(typeof val === 'string' || typeof val === 'number' || typeof val === 'boolean' || Array.isArray(val))) {
     throw new InvalidArgumentError(`invalid ${key} header`)
   } else if (val === undefined) {
     return
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -271,8 +271,15 @@ class Request {
   }
 }
 
+function processHeaderValue (key, val) {
+  if (headerCharRegex.exec(val) !== null) {
+    throw new InvalidArgumentError(`invalid ${key} header`)
+  }
+  return `${key}: ${val}\r\n`
+}
+
 function processHeader (request, key, val) {
-  if (val && typeof val === 'object') {
+  if (val && !(typeof val === 'string' || Array.isArray(val))) {
     throw new InvalidArgumentError(`invalid ${key} header`)
   } else if (val === undefined) {
     return
@@ -329,10 +336,14 @@ function processHeader (request, key, val) {
     throw new NotSupportedError('expect header not supported')
   } else if (tokenRegExp.exec(key) === null) {
     throw new InvalidArgumentError('invalid header key')
-  } else if (headerCharRegex.exec(val) !== null) {
-    throw new InvalidArgumentError(`invalid ${key} header`)
   } else {
-    request.headers += `${key}: ${val}\r\n`
+    if (Array.isArray(val)) {
+      for (let i = 0; i < val.length; i++) {
+        request.headers += processHeaderValue(key, val[i])
+      }
+    } else {
+      request.headers += processHeaderValue(key, val)
+    }
   }
 }
 
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -272,7 +272,9 @@ class Request {
 }
 
 function processHeaderValue (key, val) {
-  if (headerCharRegex.exec(val) !== null) {
+  if (val && (typeof val === 'object' && !Array.isArray(val))) {
+    throw new InvalidArgumentError(`invalid ${key} header`)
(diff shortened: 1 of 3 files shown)


ronag (approved):

jd-carroll:
Happy to, to be honest wasn't really sure those were the correct tests in the first place (but wanted something).

Could you point me to a good place to add them?

KhafraDev:
https://github.com/nodejs/undici/blob/main/test/headers-as-array.js Would be fine

jd-carroll:
@KhafraDev thanks, added

jd-carroll:
Seeing the test failures, not exactly sure what is going on. I was getting all kinds of failure locally (from unrelated code).

Do I need to look into anything?

KhafraDev:
The failures are likely caused by the change. Locally no errors should occur either.

jd-carroll:
Tests should be all set, though I was having a problem committing locally (error code 13?)

Turns out a number of the tests (incorrectly?) use numbers for header values and not strings. While this is against the "types", my guess is that many people are already doing this, so best not to break them.

lib/core/request.js
```diff
@@ -271,8 +271,15 @@ class Request {
...
+function processHeaderValue (key, val) {
+  if (headerCharRegex.exec(val) !== null) {
+    throw new InvalidArgumentError(`invalid ${key} header`)
+  }
+  return `${key}: ${val}\r\n`
+}
+
 function processHeader (request, key, val) {
-  if (val && typeof val === 'object') {
+  if (val && !(typeof val === 'string' || typeof val === 'number' || typeof val === 'boolean' || Array.isArray(val))) {
```
ronag, line 282: This can be simpler. Check for object and not array.

ronag (changes requested):

lib/core/request.js
jd-carroll: Not sure that I follow your comment (and agree that there is likely a more simple / clean way to express that line).

The root of the original issue is that there was a `typeof val === 'object'`. The reason this is problematic is that an array will be treated the same as an object.

What I am trying to express is that the only allow header values should be allowed:
* string
* string[]
* number (cast to string) (_unofficially supported, Node's `IncomingHttpRequest` lists header values as `string` | `string[]`_)
* boolean (cast to string) (_unofficially supported, Node's `IncomingHttpRequest` lists header values as `string` | `string[]`_)

So Object's should not be allowed, except for Arrays.

How to express all of that, open to suggestions / criticism...

ronag: (typeof val === 'object' && !Array.isArray(obj))

jd-carroll: Ahh, yeah... :see_no_evil:

ronag (approved):

lib/core/request.js
```diff
@@ -329,10 +336,14 @@ function processHeader (request, key, val) {
     throw new NotSupportedError('expect header not supported')
   } else if (tokenRegExp.exec(key) === null) {
     throw new InvalidArgumentError('invalid header key')
-  } else if (headerCharRegex.exec(val) !== null) {
-    throw new InvalidArgumentError(`invalid ${key} header`)
   } else {
-    request.headers += `${key}: ${val}\r\n`
+    if (Array.isArray(val)) {
+      for (let i = 0; i < val.length; i++) {
```
ronag, line 341: Don't we need to check the types of the values in the array?

jd-carroll: How would you like to think about "types" generally? When we took out the string | number | boolean typecheck above, it opens the door to Symbols, functions, and bigints. 

What may make more sense is to collapse line 282-336 into `processHeaderValue` and have all value type checking there. But then that would bring us back to what types should / should not throw.

Should we take the stance:
* Don't be dumb and we'll only throw an error if the value is `typeof === 'object'`

OR

* We'll throw an error any time value is not a string or string equivalent (string | number | boolean)

?

ronag: Just check for object like already do.

jd-carroll:
Anything else either of you would like changed?

test/headers-as-array.js
```diff
@@ -25,6 +25,69 @@ test('handle headers as array', (t) => {
...
+    }, () => {})
+  })
+})
+
+test('handle multi-valued headers', (t) => {
+  t.plan(1)
+  const headers = { a: '1', b: '2', c: '3', d: ['4', '5'] }
+
+  const server = createServer((req, res) => {
+    t.match(req.headers, { a: '1', b: '2', c: '3', d: '4, 5' })
```
KhafraDev, line 75: One thing to note: this behaves differently from fetch. Fetch's webidl checks will stringify the array, giving us `4,5` (no space), but node's http parser joins them with `, `, giving us `4, 5` (with space).

ronag (approved):

test/headers-as-array.js
jd-carroll: Is this something you think needs to be fixed / changed?

KhafraDev: No, just wanted to point it out since it might have been missed in the original review. 👍

PR #1809, opened 2022-12-16
feat(websocket): handle ping/pong frames & fix fragmented frames

KhafraDev:
- Adds ping and pong frame handling, including diagnostics_channel support
- Fixes a bug where receiving a close frame wouldn't remove its frame from the cached buffers
- Uses correct way of identifying binary frames in the websocket server for the WPTs
- Fixes parsing fragmented frames


test/websocket/ping-pong.js
```diff
@@ -0,0 +1,46 @@
...
+test('Receives ping and parses body', (t) => {
+  t.plan(1)
+
+  const server = new WebSocketServer({ port: 0 })
+
+  server.on('connection', (ws) => {
+    ws.ping('Hello, world')
+
+    ws._socket.destroy()
+    server._server.close()
```
KhafraDev, line 17: I tried closing the websocket connection we get from ws, closed it with undici's WebSocket, and just about everything else you can think of. Even though ws receives a close frame from us, or we receive a close frame from ws and send one to the server, it refused to close the server & destroy the socket. Even using `t.plan` or `t.end` wouldn't end the test because of open active handles.

KhafraDev: Also couldn't figure out a way of sending a ping frame in the middle of a fragmented message using ws.

ronag (approved):

test/websocket/ping-pong.js
lpinca: > Even though ws receives a close frame from us, or we receive a close frame from ws and send one to the server, it refused to close the server & destroy the socket.

Are you calling `socket.end()` when a close frame is both received and sent? Here is a simple test with a browser client.

```js
const http = require('http');
const WebSocket = require('ws');

const data = `<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8">
  </head>
  <body>
    <script>
      (function () {
        var ws = new WebSocket('ws://127.0.0.1:8080');

        ws.onmessage = function (event) {
          console.log(event.data);

          if (event.data === 'close') {
            ws.close();
          }
        };
      })();
    </script>
  </body>
</html>`;

const server = http.createServer();

server.on('request', function (req, res) {
  res.setHeader('Content-Type', 'text/html');
  res.end(data);
});

const wss = new WebSocket.Server({ server });

wss.on('connection', function (ws) {
  ws.on('close', function () {
    wss.close();
    server.close();
  });

  ws.on('pong', function (data) {
    console.log('Received a pong with payload data: %s', data);
  });

  ws.ping('foo');
  ws.send('close');
});

server.listen(8080, function () {
  console.log('Open http://127.0.0.1:8080 in a browser');
});
```

> Also couldn't figure out a way of sending a ping frame in the middle of a fragmented message using ws.

```js
const WebSocket = require('ws');

const wss = new WebSocket.Server({ port: 0 }, function () {
  const ws = new WebSocket(`ws://127.0.0.1:${wss.address().port}`);

  ws.on('message', function (data) {
    console.log(data.toString());
  });

  ws.on('ping', function (data) {
    console.log(data.toString());
  });

  ws.on('close', function() {
    wss.close();
  });
});

wss.on('connection', function (ws) {
  ws.send('Hel', { fin: false });
  ws.ping('ping');
  ws.send('lo', { fin: true});

  ws.close();
});
```

KhafraDev: The second issue was 100% due to fragmented messages not being handled correctly - this has been fixed. Although I opted to using raw buffers because they were in the rfc.

Regarding the first issue, I finally fixed it thanks to your help once again. :)

Issue #1811, opened 2022-12-17
WebSockets


KhafraDev:
# `permessage-deflate` support

- Is this something we *need*/*want* to support? Servers that are setup correctly should work even if the client doesn't support it.
- Memory fragmentation issues when using zlib?
- Added complexity of decompressing frames asynchronously

---

# Performance

- Consume the least amount of bytes possible, rather than concatenating every chunk available. See [ws'](https://github.com/websockets/ws/blob/fb1dfd217861757d60c1d02c6e66b4da3630cc93/lib/receiver.js#L94) implementation.
- Switch to Buffer.allocUnsafe in [`lib/websocket/frame.js`](https://github.com/nodejs/undici/blob/860c2617ba55d911286f967dbc915f03e89e0d94/lib/websocket/frame.js#L31)
- Handle TODO comments labeled as "optimize this".
- Benchmarks

---

# Tests

- Add ws' test suite. Note this isn't easy because ws' api does not strictly follow the spec and it occasionally uses internal, underscored properties. Also note that a lot of the validation tests are already handled by the WPTs.
- 100% code coverage for lib/websockets
- Test more strange/error conditions:
    * Chunks that contain thousands of frames
    * Chunks that receive a pong/close frame in the middle of a fragmented message
    * Sending invalid frames
 
---

# Bugs

- using `WebSocket.send` with a Blob asynchronously writes the blob data to the socket. This can cause issues when concurrently sending a blob with anything else. Note: we need support in node core to read a Blob synchronously.
- `ByteParser.run` runs recursively, meaning the max call stack can be exceeded under certain conditions (ie. receiving thousands of frames in a single chunk).

jimmywarting:
> Note: we need support in node core to read a Blob synchronously.

Hmm, would that be really be any good?
if we ever get something like blob backed up by the fs and you would be able to create them from a network mounted cloud drive. then that would block quite a bit.

KhafraDev:
> would that be really be any good?

No, but it's the only way that issue could be fixed without adding complexity.

jimmywarting:
where thinking about trying out websocket today but couldn't find any documentation of how to use it.
eg how to upgrade a request connection

KhafraDev:
[...]

PR #1812, opened 2022-12-18
feat(websocket): only consume necessary bytes

KhafraDev:
Refs: #1811 

Before, for every parsing stage, we would concatenate every buffer, even if we only needed to read the first few bytes (for most cases). Now, we can consume `n` bytes from the buffers, removing them & handling the total length. This simplifies the logic quite a bit and should improve performance (although there aren't any benchmarks to confirm this).

ws implements something similar:
https://github.com/websockets/ws/blob/fb1dfd217861757d60c1d02c6e66b4da3630cc93/lib/receiver.js#L94-L131

### DIFF
--- a/lib/websocket/receiver.js
+++ b/lib/websocket/receiver.js
@@ -7,6 +7,11 @@ const { kReadyState, kSentClose, kResponse, kReceivedClose } = require('./symbol
 const { isValidStatusCode, failWebsocketConnection, websocketMessageReceived } = require('./util')
 const { WebsocketFrameSend } = require('./frame')
 
+// This code was influenced by ws released under the MIT license.
+// Copyright (c) 2011 Einar Otto Stangvik <einaros@gmail.com>
+// Copyright (c) 2013 Arnout Kazemier and contributors
+// Copyright (c) 2016 Luigi Pinca and contributors
+
 const channels = {}
 channels.ping = diagnosticsChannel.channel('undici:websocket:ping')
 channels.pong = diagnosticsChannel.channel('undici:websocket:pong')
--- a/lib/websocket/receiver.js
+++ b/lib/websocket/receiver.js
@@ -49,7 +49,7 @@ class ByteParser extends Writable {
         return callback()
       }
 
-      const buffer = Buffer.concat(this.#buffers, this.#byteOffset)
+      const buffer = this.consume(2)
 
       this.#info.fin = (buffer[0] & 0x80) !== 0
       this.#info.opcode = buffer[0] & 0x0F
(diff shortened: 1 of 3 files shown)


lpinca:
Can you add the copyright notice? `ByteParser` is a slightly modified copy of `ws` `Receiver`.

FWIW `Receiver` is exposed by `ws` for third party usage.

```
const { Receiver } = require('ws');
```

KhafraDev:
Sure, although none of the code is copied directly from ws. Everything was written by hand, which is probably the reason it took so long to get right.

lpinca:
There are so many similarities here and in https://github.com/nodejs/undici/pull/1795/commits/a0094a49b3e110609f080c3dae3c656849feabea that I think it is deserved even if `ws` `Receiver` was used only a reference. You could add something along the lines "This code was adapted from XXX released under the MIT license". It is the only requirement of the MIT license.

Extending `stream.Writable` is a good choice for a future `permessage-deflate` implementation where decompression is async and is slower than incoming data. At the moment, without `permessage-deflate`, extending `stream.Writable` is only unnecessary overhead as everything is synchronous.

KhafraDev:
I've added a notice, let me know if there's anything!

lpinca:
That's ok, thanks. If you think no notice should be added and want to remove it, feel free to do that, I won't get in the way.

KhafraDev:
No, you deserve recognition - without your help the last week, there's no way I'd have as good of an understanding of websockets, and undici's implementation, without ws, would be incomplete and full of bugs. A notice is the least I can do, all things considered.

lib/websocket/receiver.js
```diff
@@ -254,6 +233,43 @@ class ByteParser extends Writable {
...
+  /**
+   * Take n bytes from the buffered Buffers
+   * @param {number} n
+   * @returns {Buffer|null}
+   */
+  consume (n) {
+    if (n > this.#byteOffset) {
+      return null
+    } else if (n === 0) {
+      return Buffer.allocUnsafe(0)
```
lpinca, line 245: Consider adding an empty buffer constant so that you can reuse it every time you need a zero length buffer without calling the constructor every time.

lib/websocket/receiver.js
```diff
@@ -254,6 +233,43 @@ class ByteParser extends Writable {
...
+
+    const buffer = Buffer.allocUnsafe(n)
+    let offset = 0
+
+    while (offset !== n) {
+      const next = this.#buffers[0]
+      const { length } = next
+
+      if (length + offset === n) {
+        buffer.set(this.#buffers.shift(), offset)
```
lpinca, line 256: If you get here with `offset === 0` no copy and no `Buffer.allocUnsafe(n)` is needed, you can just return `this.#buffers.shift()`.

ronag (approved):

PR #1814, opened 2022-12-18
make formdata body immutable and encode it only once

jimmywarting:
with this changes you only have to encode the formdata once. 
it also creates an array of blob parts to make it immutable so that it can't be changed afterwards

This basically solves the following problem:
```js
const fd = new FormData()
fd.set('foo', 'bar')
const res = new Response(fd)
fd.set('foo', 'foo')
// fd.set('foo', 'foooo') could also set the field to a other size making req content-length invalid.
const fd2 = await res.formData()
fd2.get('foo')
```
reason why i didn't just do `action = new Blob(parts)` was due to 3rd party files that can be backed up by fs

### DIFF
--- a/lib/fetch/body.js
+++ b/lib/fetch/body.js
@@ -99,7 +99,7 @@ function extractBody (object, keepalive = false) {
     // Set source to a copy of the bytes held by object.
     source = new Uint8Array(object.buffer.slice(object.byteOffset, object.byteOffset + object.byteLength))
   } else if (util.isFormDataLike(object)) {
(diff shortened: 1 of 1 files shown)


KhafraDev:
Can you add a test for the immutable formdata?

jimmywarting:
sure

jimmywarting:
hmm, Wondering if i could add the test to WPT.

KhafraDev:
It'd make more sense as a WPT, but I don't think there's a reason to hold up this PR waiting for the test to be accepted.

ronag (approved):

PR #1818, opened 2022-12-19
fix(websocket): deprecation warning & 64-bit unsigned int body length

KhafraDev:


lib/websocket/frame.js
```diff
@@ -45,7 +45,8 @@ class WebsocketFrameSend {
     if (payloadLength === 126) {
       new DataView(buffer.buffer).setUint16(2, bodyLength)
     } else if (payloadLength === 127) {
-      // TODO: optimize this once tests are added for payloads >= 2^16 bytes
+      // Clear extended payload length
+      buffer[2] = buffer[3] = 0
       buffer.writeUIntBE(bodyLength, 4, 6)
```
KhafraDev, line 50: I removed the todo comment because there likely isn't a faster way, unless I'm missing something obvious.

ronag: Assert body length fits in 6 bytes?

ronag: that's why it's an assertion?

KhafraDev: I don't think it's possible to allocate a buffer/typedarray/string that long.

KhafraDev: Although we should add an assertion on the receiving side, to ensure that we don't get a malicious frame.

PR #1822, opened 2022-12-20
fix(websocket): validate payload length received

KhafraDev:
Refs: https://github.com/nodejs/undici/pull/1818#discussion_r1053575150

### DIFF
--- a/lib/websocket/receiver.js
+++ b/lib/websocket/receiver.js
@@ -197,6 +197,18 @@ class ByteParser extends Writable {
 
       const buffer = this.consume(8)
       const upper = buffer.readUInt32BE(0)
+
+      // 2^31 is the maxinimum bytes an arraybuffer can contain
+      // on 32-bit systems. Although, on 64-bit systems, this is
+      // 2^53-1 bytes.
+      // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Invalid_array_length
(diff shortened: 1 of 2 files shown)


lib/websocket/receiver.js
```diff
@@ -197,6 +197,18 @@ class ByteParser extends Writable {
...
       const buffer = this.consume(8)
       const upper = buffer.readUInt32BE(0)
+
+      // 2^31 is the maxinimum bytes an arraybuffer can contain
+      // on 32-bit systems. Although, on 64-bit systems, this is
+      // 2^53-1 bytes.
+      // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Errors/Invalid_array_length
+      // https://source.chromium.org/chromium/chromium/src/+/main:v8/src/common/globals.h;drc=1946212ac0100668f14eb9e2843bdd846e510a1e;bpv=1;bpt=1;l=1275
+      // https://source.chromium.org/chromium/chromium/src/+/main:v8/src/objects/js-array-buffer.h;l=34;drc=1946212ac0100668f14eb9e2843bdd846e510a1e
+      if (upper > 2 ** 31 - 1) {
```
KhafraDev, line 207: Although 2^53 is the theoretical max length, it depends on the system (at least from information I could find). I wasn't able to allocate anything close to 2^53 bytes without a range error being thrown, and it's unlikely a client would *want* such a large payload. This seemed like a fine middle ground solution that will work on every system.

ronag (approved):

Issue #1823, opened 2022-12-21
unidict fetch leaked if using Request object as argument


stalkerg:
## Bug Description

unidict leaked if using Request, for example:
```js
export function create_fetch({ event, options, state, get_cookie_header }) {
	return async (info, init) => {
		return await fetch(new Request(info, init));
	};
}
```

I found it in SvelteKit inside this issue https://github.com/sveltejs/kit/issues/8239

## Reproducible By
See SvelteKit issue. 

## Environment
Ubuntu and Gentoo with the latest node 18

stalkerg:
Seems like we trying to copy AbortController signal from the income Request here:
https://github.com/nodejs/undici/blob/main/lib/fetch/request.js#L106

and after locking into the arrow function, such signal from the previous Request:
https://github.com/nodejs/undici/blob/main/lib/fetch/request.js#L351


seems like we never run finalization for the original signal. Not sure here, but a bug exists... maybe we should do `clone` instead copy only a few fields?

ronag:
Isn't this just a case of the finalisation registry not yet running the finaliser?

stalkerg:
hmm seems like it's never running, and at the same time this:
```js
export function create_fetch({ event, options, state, get_cookie_header }) {
	return async (info, init) => {
		return await fetch(info, init);
	};
}
```
working prefect.

stalkerg:
Seems like this https://github.com/nodejs/undici/blob/main/lib/fetch/request.js#L337 is needed only if `init.signal` exist, for the case when we just copy the ac signal from the original Request we shouldn't follow it.

stalkerg:
As I understand, we shouldn't recreate the Request inside `fetch` if we have already received the Request. 
Even if we made Request based on a Request, it should be copied without the following:
https://developer.mozilla.org/en-US/docs/Web/API/Request/Request#parameters

Maybe I do not fully understand the concept behind this, but it seems like we are doing something wrong now.

stalkerg:
Sorry, for too many words here. I checked the specification https://fetch.spec.whatwg.org/#fetch-method and it seems like, currently, everything is implemented right. I still have questions about such specifications but this question is outside this issue. 
Also, seems like SveltKit `fetch` like wrapper is not fully correct just because it's not a `fetch` anymore, but it's also a different topic. 

@ronag big thanks for a fast fix.

stalkerg:
@ronag can you reopen? The latest changes make memory leak again.

[...]

PR #1824, opened 2022-12-21
fix(fetch): weaker refs

ronag:
Refs: https://github.com/nodejs/undici/issues/1823

### DIFF
--- a/lib/fetch/request.js
+++ b/lib/fetch/request.js
@@ -348,7 +348,10 @@ class Request {
       if (signal.aborted) {
         ac.abort(signal.reason)
       } else {
-        const abort = () => ac.abort(signal.reason)
+        const acRef = new WeakRef(ac)
+        const abort = function () {
+          acRef.deref()?.abort(this.reason)
+        }
         signal.addEventListener('abort', abort, { once: true })
         requestFinalizer.register(this, { signal, abort })
       }


stalkerg:
Should I check your PR on my test case or you already can reproduce it?

stalkerg:
Thanks! My test case working correctly now. 
It should close https://github.com/sveltejs/kit/issues/8239 after release.

ronag:
> PS but can we just copy the whole ac instead of following the signal in that case?

I don't understand.

stalkerg:
> I don't understand.

sorry, it's a confusing concept:
```
let req1 = Request("");
let req2 = Request(req1);
```
Now we have two similar instances of requests (copy?) and two AbortController inside, but the signal from req1 will be leaked into req2 but not the opposite.  
Anyway seems like ok for Request and it's just my misunderstanding. 
At the same time, I don't think we must generate a new Request inside fetch if the state of the Request was initial (nobody tried to run it). But again, it's maybe my fundamental misunderstanding.

ronag:
I think we currently follow the spec as intended.

mcollina (commented):
Can we add a test for this?

ronag:
> Can we add a test for this?

Not sure how.

KhafraDev:
I don't think a test is needed, maybe just a comment linking to the issue

stalkerg:
I can't find a good and mature way to unit test the memory leaks in JS for Node; it seems like it's not trivial.

PR #1828, opened 2022-12-22
fix(websocket): run parser in loop, instead of recursively

KhafraDev:
Refs: #1811

Unsure if I can add a test case, but I've confirmed that this patch fixes the following repro:

```mjs
import assert from 'assert'
import { WebSocketServer } from 'ws'
import { WebSocket } from './index.js'

const helloFrame = Buffer.from([0x81, 0x05, 0x48, 0x65, 0x6c, 0x6c, 0x6f])

// largest body size that is divisible by the frame length
const buffer = Buffer.allocUnsafe(2 ** 31 - 2).fill(helloFrame)

const server = new WebSocketServer({ port: 0 })

server.on('connection', (ws) => {
  ws._socket.write(buffer)
})

const ws = new WebSocket(`ws://localhost:${server.address().port}`)

ws.addEventListener('message', ({ data }) => {
  assert.equal(data, 'Hello')
})
```

Before, a RangeError would be thrown; now, it succeeds. The problem is that the test takes forever to complete because it's parsing about 306,783,378 frames. I have no idea what the v8 limit is for how many times a function can recurse, or if it's dependent upon hardware/specs.

### DIFF
--- a/lib/websocket/receiver.js
+++ b/lib/websocket/receiver.js
@@ -48,201 +48,204 @@ class ByteParser extends Writable {
    * or not enough bytes are buffered to parse.
    */
   run (callback) {
-    if (this.#state === parserStates.INFO) {
-      // If there aren't enough bytes to parse the payload length, etc.
-      if (this.#byteOffset < 2) {
-        return callback()
-      }
+    while (true) {
+      if (this.#state === parserStates.INFO) {
+        // If there aren't enough bytes to parse the payload length, etc.
+        if (this.#byteOffset < 2) {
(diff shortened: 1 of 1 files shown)


lpinca:
> I have no idea what the v8 limit is for how many times a function can recurse, or if it's dependent upon hardware/specs.

It's around 15k, this is sufficient https://github.com/websockets/ws/blob/d412358ccb5320bc00c8993ecd5d9b992df0753e/test/receiver.test.js#L437-L455

PR #1829, opened 2022-12-22
feat: allow connection header in request

### LINKED ISSUE #444: `Client.reset` to reset connection without destroying client

### LINKED ISSUE #1305: Can't remove default headers. Can't overwrite `"sec-fetch-mode"`, `"connection"` headers.

### LINKED ISSUE #1310: How to disconnect after fetch?

### LINKED ISSUE #1746: undici doesnt support keep-alive connection header?

metcoder95:
Follow up from #1747 

### References
- Closes #1746
- Closes #1310
- Closes #1305
- Closes #444


lib/client.js
```diff
@@ -115,6 +117,10 @@ class Client extends DispatcherBase {
       throw new InvalidArgumentError('unsupported keepAlive, use pipelining=0 instead')
     }
 
+    if (reset != null && reset.constructor !== Boolean) {
```
KhafraDev, line 120: typeof reset === 'boolean'?

mcollina (approved):
lgtm

metcoder95:
Sorry for the noise, finger issue 😅 
I'll add the documentation soon, in the meanwhile is ready for review. Thanks 🙂

lib/client.js
metcoder95: Done in [bf16edd](https://github.com/nodejs/undici/pull/1829/commits/bf16edd8db8f377257c806d88a4bfc2ff89dfdb7)

ronag (changes requested):
Why isn't it enough to have reset on the request? I don't think you need a reset option on the client.

metcoder95:
> Why isn't it enough to have reset on the request? I don't think you need a reset option on the client.

Is it true that the `Client` does not need it at all and for instance, the parent implementations neither as well, my rationing was to allow the implementation side to have granular control over it.
Meaning that if they want to set it as the default behavior or just for a couple of requests, they have the chance to do so without having to keep track of appending the `reset` option on every request if they want to make this the default one.

e.g.

> Circuit Breaker that transitions into half-open so it starts assessing the connection by sending requests with `connection: close;` and transitions to `closed` if the connection becomes stable.

But happy to revert it if everybody agrees this might be an overhead 🙂

ronag:
You can set pipelining to 0 to achieve the same effect.

lib/client.js
```diff
@@ -115,6 +117,10 @@ class Client extends DispatcherBase {
       throw new InvalidArgumentError('unsupported keepAlive, use pipelining=0 instead')
     }
 
+    if (reset != null && typeof reset !== 'boolean') {
+      throw new InvalidArgumentError('invalid reset')
```
anonrig, line 121: in order to have a consistent error message, this should be the same as the previous error message `reset must be a boolean`

lib/client.js
```diff
@@ -1381,10 +1389,15 @@ function write (client, request) {
     socket[kReset] = true
   }
 
+  if (reset) {
+    socket[kReset] = true
+  }
```
anonrig, line 1394: We can move this to line 1388

lib/client.js
```diff
@@ -1381,10 +1389,15 @@ function write (client, request) {
...
+
   if (blocking) {
     socket[kBlocking] = true
   }
 
+  const shouldReset = request.reset != null ? request.reset : client[kClientResetValue]
```
anonrig, line 1400: Nit: `request.reset ?? client[kClientResetValue]`

lib/pool.js
```diff
@@ -40,6 +41,10 @@ class Pool extends PoolBase {
       throw new InvalidArgumentError('invalid connections')
     }
 
+    if (reset != null && reset.constructor !== Boolean) {
```
anonrig, line 44: Why not `typeof reset !== 'boolean'`?

anonrig (changes requested):

lib/client.js
KhafraDev: Note: this would drop support for v12 and you'd need to remove it from the ci.

anonrig: Ah, that's true. I recommend doing it in a new pull request, if we are going to drop it (and I think we should)

lib/pool.js
metcoder95: Was mostly about a way of working, but reverted it to keep in sync with previous @KhafraDev suggestion as well 👍

metcoder95:
I addressed the changes requested and reverted the extension to `Client` to support the `reset` property. Now, this is only accessible when executing any call that promotes the `DispatchOptions`-like object. I also added the documentation on top. 🙂 

cc: @ronag @KhafraDev @anonrig

lib/client.js
```diff
@@ -1361,16 +1361,11 @@ function write (client, request) {
     return false
   }
```
ronag, line 1362: Should be enough to just do:

```
if (reset) {
  socket[kReset] = true
}
```

I don't think you need the rest of the changes here.

metcoder95: Sadly it doesn't work without it, as we need to use the `Request#reset` to see if this request in specific will be reset after response. If relying only on the `pipelining` option, we'll ignore the Request options and continue sending the `connection: keep-alive` header.

Or do you refer to something else? 🤔

ronag: https://github.com/nodejs/undici/pull/1846

lib/core/request.js
```diff
@@ -54,19 +54,22 @@ try {
...
+      headers,
+      query,
+      idempotent,
+      blocking,
+      upgrade,
+      headersTimeout,
+      bodyTimeout,
+      reset,
+      throwOnError
+    } = options
```
ronag, line 71: Why change this?

lib/core/request.js
```diff
@@ -320,7 +330,10 @@ function processHeader (request, key, val) {
     key.length === 10 &&
     key.toLowerCase() === 'connection'
   ) {
-    throw new InvalidArgumentError('invalid connection header')
+    if (val.toLowerCase() === 'close') {
+      request.reset = true
+    }
+    request.headers += `${key}: ${val}\r\n`
```
ronag, line 336: We should not set the header here...

lib/client.js
metcoder95: Got it, let me rebase and adjust 👍

docs/api/Dispatcher.md
```diff
@@ -192,6 +192,7 @@ Returns: `Boolean` - `false` if dispatcher is busy and further dispatch calls wo
 * **origin** `string | URL`
 * **path** `string`
 * **method** `string`
+* **reset** `null | boolean` (optional) - Default: `null` - Indicates whether the request should attempt to create a long-living connection by sending the `connection: keep-alive` header, or close it immediately after response by sending `connection: close`.
```
ronag, line 195:
```suggestion
* **reset** `boolean` (optional) - Default: `false` - Indicates whether the request should attempt to create a long-living connection by sending the `connection: keep-alive` header, or close it immediately after response by sending `connection: close`.
```

lib/core/request.js
metcoder95: Better to do it in `Client` or where exactly?

lib/core/request.js
metcoder95: Do you mean the reset?

Mostly to validate the `reset` value and carry over the setting during the request lifecycle, as it is attached to a single request.

Another option is to pass it as a third-parameter to the `write` function on `lib/client`. What are your thoughts? 🤔

lib/core/request.js
metcoder95: Instead, just switching the value on the `reset` will do what's needed, and just check for either one of the `close` or `keep-alive` options. The settings usually come in the `keep-alive` header if custom behavior is desired. Thoughts? 🤔

lib/client.js
```diff
@@ -1361,16 +1361,11 @@ function write (client, request) {
...
+  if (upgrade || method === 'CONNECT' || method === 'HEAD' || reset) {
     // https://github.com/mcollina/undici/issues/258
-
     // Close after a HEAD request to interop with misbehaving servers
     // that may send a body in the response.
 
-    socket[kReset] = true
-  }
-
-  if (upgrade || method === 'CONNECT') {
```
ronag, line 1373: Unecessary changes... just set:

```
if (reset) {
  socket[kReset] = true
}
```

lib/core/request.js
ronag: Just add `reset` to the previous list. I don't understand why you moved out options.

lib/core/request.js
ronag: Just remove `request.headers += `${key}: ${val}\r\n`` and make sure val is either keep-alive or close.

test/request.js
```diff
@@ -149,3 +149,61 @@ test('Absolute URL as pathname should be included in req.path', async (t) => {
...
+      t.equal(req.headers.connection, 'keep-alive')
+      res.statusCode = 200
+      res.end('hello')
+    })
+
+    t.plan(3)
+
+    t.teardown(server.close.bind(server))
+
+    await server.listen(0)
```
KhafraDev, line 201: this doesn't return a promise

```js
const { once } = require('events')

server.listen(0)

```

lib/core/request.js
```diff
@@ -250,6 +259,7 @@ class Request {
     if (channels.trailers.hasSubscribers) {
       channels.trailers.publish({ request: this, trailers })
     }
+
```
ronag, line 262: Unnecessary change...

mcollina (approved):
lgtm

docs/api/Dispatcher.md
```diff
@@ -192,6 +192,7 @@ Returns: `Boolean` - `false` if dispatcher is busy and further dispatch calls wo
 * **origin** `string | URL`
 * **path** `string`
 * **method** `string`
+* **reset** `boolean` (optional) - Default: `false` - Indicates whether the request should attempt to create a long-living connection by sending the `connection: keep-alive` header, or close it immediately after response by sending `connection: close`.
```
kibertoad, line 195: @metcoder95 This is phrased in a confusing way. Documentation implies that "true" means creating long-lived connection, while code seems to indicate the opposite.

metcoder95: You're right, let me fix it opening a new PR, thanks for the heads up 🙇

robhogan:
Looks like this incorrectly closed #1305 - it only partly addresses it. It still doesn't appear to be possible to customise `sec-fetch-mode`.

metcoder95:
Please, feel free to open a new issue to reevaluate and see further actions 🙂

robhogan:
Will do, but also, please don't mark issues closed in a PR unless the issue is actually closed - it just creates unnecessary churn and fragmentation.

KhafraDev:
setting a custom sec-fetch-mode is probably a wontfix anyways.

PR #1831, opened 2022-12-22
websocket: add tests for opening handshake

KhafraDev:
Should be close to 100% coverage for [`/lib/websocket/connection.js`](https://github.com/nodejs/undici/blob/main/lib/websocket/connection.js) barring 1 or 2 branches which aren't yet possible to reach (no websocket extension support).

### DIFF
--- a/test/websocket/diagnostics-channel.js
+++ b/test/websocket/diagnostics-channel.js
@@ -1,0 +1,37 @@
+'use strict'
+
+const { test } = require('tap')
+const dc = require('diagnostics_channel')
+const { WebSocketServer } = require('ws')
+const { WebSocket } = require('../..')
+
+test('undici:websocket:open and undici:websocket:close', (t) => {
+  t.plan(6)
+
+  const server = new WebSocketServer({ port: 0 })
+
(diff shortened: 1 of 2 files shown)


test/websocket/diagnostics-channel.js
```diff
@@ -0,0 +1,37 @@
+'use strict'
+
+const { test } = require('tap')
+const dc = require('diagnostics_channel')
+const { WebSocketServer } = require('ws')
+const { WebSocket } = require('../..')
+
+test('undici:websocket:open and undici:websocket:close', (t) => {
```
KhafraDev, line 8: Is it possible to run tap tests sequentially within a file? Otherwise the diagnostics channel were being triggered by both tests, causing them to fail.

Ethan-Arrowood: try `t.jobs = 1` it should force sequential execution

KhafraDev: The issue was that after each test I wasn't removing the listener 😐.

ronag (approved):

PR #1835, opened 2022-12-26
fix(WPTs): flaky abort test

### LINKED ISSUE #1834: fetch/api/abort/general.any.js WPT is flaky on Windows
[`fetch/api/abort/general.any.js`](https://github.com/nodejs/undici/blob/main/test/wpt/tests/fetch/api/abort/general.any.js) has a test that occasionally prevents the worker from exiting, causing the CI to fail after it times out. This file is a conglomerate of ~50 tests, so skipping the file entirely is not a feasible solution. There is no way to disable individual tests either since undici uses the WPT test harness.

The test is only flaky on windows, and is most likely an issue where something isn't getting unref'd (considering it's testing aborts, it's likely AbortController/AbortSignal). Using powershell, you can run the wpt testing script x amount of times by doing `1..50 | % { npm run test:wpt }`

It's possible to determine when a test times out using the following diff:

```diff
diff --git a/test/wpt/runner/runner/runner.mjs b/test/wpt/runner/runner/runner.mjs
index 8dfb21e8..a99b94fc 100644
--- a/test/wpt/runner/runner/runner.mjs
+++ b/test/wpt/runner/runner/runner.mjs
@@ -112,6 +112,13 @@ export class WPTRunner extends EventEmitter {
       })

       activeWorkers.add(worker)
+      const timeout = setTimeout(
+        () => {
+          console.log('test timed out', { worker, test })
+        },
+        meta.timeout === 'long' ? 60_000 : 10_000
+      )

       worker.on('message', (message) => {
         if (message.type === 'result') {
@@ -123,6 +130,7 @@ export class WPTRunner extends EventEmitter {

       worker.once('exit', () => {
         activeWorkers.delete(worker)
+        clearTimeout(timeout)

         if (activeWorkers.size === 0) {
           this.handleRunnerCompletion()
```

Calling `worker.terminate` doesn't do anything in this case because there may still be active refs; terminate works "as soon as possible". Another option is to send a message to the worker, and force it to shutdown, but there may be other implications by doing this.

KhafraDev:
Fixes #1834 

Merry Christmas! 🎅

The flaky WPT on Windows should now be fixed. I detailed the cause/fix for this in the issue above. Testing locally (on windows), this fixed the issue for me.

Also, tests which timeout in the future will now be logged. Should help discover any issues in the runner.

### DIFF
--- a/test/wpt/server/server.mjs
+++ b/test/wpt/server/server.mjs
@@ -115,18 +115,19 @@ const server = createServer(async (req, res) => {
       while (true) {
         if (!res.write('.')) {
           break
-        } else if (abortKey && stash.take(abortKey, fullUrl.pathname)) {
+        } else if (abortKey && stash.take(abortKey)) {
           break
         }
 
-        await sleep(10)
+        await sleep(100)
       }
 
       if (stateKey) {
         stash.put(stateKey, 'closed', fullUrl.pathname)
       }
 
-      return res.end()
+      res.end()
+      return
     }
     case '/fetch/api/resources/stash-take.py': {
       // https://github.com/web-platform-tests/wpt/blob/6ae3f702a332e8399fab778c831db6b7dca3f1c6/fetch/api/resources/stash-take.py
--- a/test/wpt/runner/runner/runner.mjs
+++ b/test/wpt/runner/runner/runner.mjs
@@ -112,6 +112,12 @@ export class WPTRunner extends EventEmitter {
       })
 
       activeWorkers.add(worker)
+      const timeout = setTimeout(
+        () => {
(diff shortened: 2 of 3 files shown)


test/wpt/server/server.mjs
```diff
@@ -115,18 +115,19 @@ const server = createServer(async (req, res) => {
       while (true) {
         if (!res.write('.')) {
           break
-        } else if (abortKey && stash.take(abortKey, fullUrl.pathname)) {
+        } else if (abortKey && stash.take(abortKey)) {
           break
         }
 
-        await sleep(10)
+        await sleep(100)
```
KhafraDev, line 122: This is the fix.

PR #1846, opened 2023-01-02
fix: don't send keep-alive if we want reset

ronag:


ronag:
Question: Should CONNECT even have a connection header?

mcollina (approved):
lgtm

PR #1848, opened 2023-01-05
feat: add cookie parsing ability

KhafraDev:
Refs: https://github.com/nodejs/node/pull/45684#issuecomment-1371954007

It turns out my old [CookieStore branch](https://github.com/khafradev/undici/tree/cookie-store) came in use! Most of the `set-cookie` parsing came from there, making it much easier for me.

- adds `deleteCookie`, `getCookies`, `getSetCookies`, and `setCookie` (ie. everything from https://deno.land/std@0.170.0/http/cookie.ts). The APIs are (almost?) a 1:1 match and 95% of deno's tests passed. There were a few tests that Deno had where they couldn't parse certain cookies, but my parser could. It's likely because they use a lazy solution (no offense) while I followed the CookieStore spec/RFC.
- this version uses webidl, while deno's doesn't. Webidl is more strict than doing no validation, so there might be differences there as well.
- this is only exposed in node > 18 because I was lazy and didn't test it in v16/v14. Other than some syntax I use, there's nothing stopping anyone from adding support for node 12/14/16.


ronag (approved):
RSLGTM

test/cookie/cookies.js
```diff
@@ -0,0 +1,594 @@
...
+const { test } = require('tap')
+const {
+  deleteCookie,
+  getCookies,
+  getSetCookies,
+  setCookie,
+  Headers
+} = require('../..')
+
+// https://raw.githubusercontent.com/denoland/deno_std/b4239898d6c6b4cdbfd659a4ea1838cf4e656336/http/cookie_test.ts
```
mcollina, line 12: You must include the license for this file at the top.

mcollina (changes requested):

mcollina:
I think you should include the full MIT license of Deno there.

KhafraDev:
Added the full license

mcollina (approved):
lgtm

PR #1849, opened 2023-01-06
fix(cookie): add docs & expose in node v16

KhafraDev:
Since the cookie handling uses some stuff from the fetch spec, it can only be exposed in node v16+.

### DIFF
--- a/docs/api/Cookies.md
+++ b/docs/api/Cookies.md
@@ -1,0 +1,101 @@
+# Cookie Handling
+
+## `Cookie` interface
+
+* **name** `string`
+* **value** `string`
+* **expires** `Date|number` (optional)
+* **maxAge** `number` (optional)
+* **domain** `string` (optional)
+* **path** `string` (optional)
+* **secure** `boolean` (optional)
+* **httpOnly** `boolean` (optional)
+* **sameSite** `'String'|'Lax'|'None'` (optional)
+* **unparsed** `string[]` (optional) Left over attributes that weren't parsed.
+
+## `deleteCookie(headers, name[, attributes])`
+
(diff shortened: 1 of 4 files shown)


ronag (approved):

mcollina (commented):
How will we expose those in Node.js core?

KhafraDev:
I'm not sure, there's a few way I could think of.

1. with the rest of undici. I know there's an issue open somewhere asking to expose other parts of undici in node core.
2. As `http.cookies`/`https.cookies`/`http2.cookies`, or somewhere in util?
3. Recommend using undici in node docs for cookie handling with Headers.

I also need to note that it doesn't currently work with the global Headers because we need to access internals (`set-cookie` headers are stored in an array rather than being joined together).

mcollina:
Once undici is updated in core, will it work with the global headers?

mcollina (approved):
lgtm

KhafraDev:
> will it work with the global headers

For `deleteCookie`, `getCookies`, and `setCookie` yes, but `getSetCookies` is handled differently, unlike Deno, to be more correct & not introduce a breaking change. Deno stores `set-cookie` headers in an array, and when getting them, will return an array. The spec tells us to return a concatenated string though.

To bypass this, I added a property to HeadersList that stores `set-cookie` headers in an array, and we use the value of that rather than doing `headers.get('set-cookie')`, which wouldn't be possible to parse (using `headers.get('set-cookie')` still returns a string). Since we can guarantee that node won't add in their own symbols to Headers (etc), I could add support for it, but I wasn't keen on implementing such a dirty hack.

PR #1851, opened 2023-01-09
feat: implement 0 dependency streaming multipart/form-data parser

KhafraDev:
This PR implements a multipart/form-data parser with a 1:1 match to busboy's API. This took me about a week to complete and many, many hours. I believe docs and types are needed still, but don't feel like adding them right now.

Tests from busboy are included, and every single test works without any modifications!

- [ ] - docs
- [ ] - benchmarks
- [x] - types
- [x] - replace busboy in fetch
- [x] - use ~~c8~~ nyc for coverage in busboy tests (busboy uses vanilla node for testing)
- [ ] - ??

---

Bug(s):

- [x] ~~`path.basename` returns different values in windows & linux (`C:\\files\\1k_b.dat`)~~ fixed in 56052ec8fa47c566f8536b79c26c80eeda51268d
- [x] ~~reading headers is extremely slow~~ fixed in b7b9645e54df7ba5f42eeeca1626d28eadfcb385, parsing headers should now be O(n)
- [x] ~~in certain circumstances FileStream will not emit the 'end' event~~ fixed in dd02fe5930507d67f20acf9044aa7e6036c65df1


kibertoad:
@KhafraDev How does the perf compare?

KhafraDev:
Needs benchmarks but probably slower in some areas, same in others. Busboy uses streamsearch (which implements a Boyer-Moore-Horspool algorithm) to search for boundaries after bodies etc. While this implementation is much... lazier/easier. Rather than doing that,  I join chunks together until one of them contains the boundary, and then split the chunks accordingly. Of course without benchmarks, most of that is just speculation of what's faster.

jimmywarting:
> Busboy uses streamsearch (which implements a Boyer-Moore-Horspool algorithm) to search for boundaries after bodies etc

multipart/form-data is really old and historical. It isn't the best solution for parsing and sending files.
Something that would be better is to have something more like structural data that tells how large a field is in size so that you can read x amount of data instead of scanning a boundary.

I [asked](https://github.com/whatwg/fetch/issues/1130) if we could at least add `content-length` to each and every field. (not just the total body size). that would make boundary a bit more obsolete. but it got very idle. 

you know what would be awesome? having `cbor` support baked into fetch! `await new Resoponse(data).cbor()`
it would totally kick out formdata, json, text and blobs out of existence by having more structural data support. it would more or less support the same things that you could do with an object and `structuralClone` on. (just wishful thinking)

lib/formdata/constants.js
```diff
@@ -0,0 +1,41 @@
...
+
+const chars = {
+  '-': '-'.charCodeAt(0),
+  cr: '\r'.charCodeAt(0),
+  lf: '\n'.charCodeAt(0),
+  ':': ':'.charCodeAt(0),
+  ' ': ' '.charCodeAt(0), // 0x20
+  ';': ';'.charCodeAt(0),
+  '=': '='.charCodeAt(0),
+  '"': '"'.charCodeAt(0)
```
anonrig, line 25: Nit: Let’s use the actual ascii code for this.

lib/formdata/util.js
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Collect a string inside of quotes. Unlike
+ * `collectAnHTTPQuotedString` from dataURL utilities,
+ * this one does not remove backslashes.
+ * @param {string} str
+ */
+function collectHTTPQuotedStringLenient (str) {
+  if (str[0] !== '"') {
+    return str
+  }
```
anonrig, line 14: charCodeAt numeric comparison would be faster

lib/formdata/util.js
```diff
@@ -0,0 +1,55 @@
...
+
+  return collectASequenceOfCodePoints(
+    (char) => char !== '"',
```
anonrig, line 17: charCode suggestion applies to here too

lib/formdata/constants.js
```diff
@@ -0,0 +1,41 @@
...
+}
+
+const emptyBuffer = Buffer.alloc(0)
+
+const crlfBuffer = Buffer.from('\r\n')
```
anonrig, line 30: Nit: Can we pre calculate this and write it here to reduce the boot time overhead, even though it might be negligible?

lib/fetch/dataURL.js
```diff
@@ -117,25 +117,37 @@ function URLSerializer (url, excludeFragment = false) {
...
   // 1. Let result be the empty string.
   let result = ''
 
   // 2. While position doesn’t point past the end of input and the
   // code point at position within input meets the condition condition:
   while (position.position < input.length && condition(input[position.position])) {
     // 1. Append that code point to the end of result.
-    result += input[position.position]
+    if (inputIsString) {
+      result += input[position.position]
```
anonrig, line 138: If inputIsString you might skip the while loop and provide a fast path for this function.

lib/formdata/util.js
```diff
@@ -0,0 +1,55 @@
...
+}
+
+function hasHeaderValue (chunk, start) {
+  for (let i = start; i < chunk.length; i++) {
+    if (!isWhitespace(chunk[i])) {
+      return true
+    }
+  }
+
+  return false
```
anonrig, line 47: Would chunk.find be faster? Inlined functions are most of the time faster due to severe v8 optimizations.

lib/fetch/dataURL.js
KhafraDev: We can't skip the loop entirely, we need to increase the index and check if the character matches the condition

lib/formdata/util.js
KhafraDev: I doubt it, you can't specify a position to start at so I'd have to split the buffer first.

lib/formdata/parser.js
```diff
@@ -0,0 +1,672 @@
...
+      ) {
+        // The multipart/form-data media type does not support any MIME header
+        // fields in parts other than Content-Type, Content-Disposition, and (in
+        // limited circumstances) Content-Transfer-Encoding.  Other header
+        // fields MUST NOT be included and MUST be ignored.
+
+        // Go to the next header if one is available. Headers are split by \r\n,
+        // so we skip all characters until the sequence is found.
+        collectASequenceOfCodePoints(
+          (char) => char !== chars.cr && buffer[position.position + 1] !== chars.lf,
```
maximelkin, line 561: feels like here should be `||`
maybe rewrite to `!(char === char.cr && buffer[position.position +1] === char.lf)`
583 line same

KhafraDev:
I'm much more interested in replacing the asynchronous parsing with a synchronous parser. Not only is it actually spec compliant, but it will finally cleanup the body consumers. It doesn't make sense to parse a multipart/form-data body asynchronously if it's already in memory. It's also many times slower than Deno/bun (https://twitter.com/jarredsumner/status/1625067963656335361).

cc @jimmywarting I know you disagree with this, do you have any counterarguments before I replace busboy with the proposed alternative?

mcollina:
I don't think a synchronous multipart parser would work for us long term.
Node.js just gained support for disk-baked Blobs. This means we can stream files to disk and then create Blobs that way and avoid memory leaks. However we can do that only if we have a streaming multipart parser.

KhafraDev:
This doesn't seem to be an issue with Deno/bun (or any of the browsers, although their use case is different). Since `request.body` is a ReadableStream, we could document formData's downsides and show an example of asynchronously parsing the body with busboy (or this async parser). It makes it much harder to maintain this section of the code, and many of the WPTs regarding the formdata body mixin are disabled :(.

There is also a proposal to add a limit to the max body size a FormData can have (https://github.com/whatwg/fetch/issues/1592) and it's already an option on undici's Agent. There are alternative apis that undici has that are much better suited for large bodies. The main users of .formData IIRC are library authors who want cross-env compatibility - I wonder how many users are active using it?

One more note: it's incredibly slow. https://twitter.com/jarredsumner/status/1625067963656335361

KhafraDev:
There's also the issue of *how* parsing a multipart/form-data should be parsed and there's no actual spec for it. The WPT coverage sucks as well, while it's really good for *every* other mixin.

Personally I'd like to follow the spec and other platforms who have all implemented the parsing synchronously. See #1694 for proof.

mcollina:
I don't think we could ever be in agreement on this one. Crashing for out-of-memory issues when receiving files is not safe.
It's better to not ship a feature than shipping a feature that will crash everything: the number of vulnerabilities we will receive for this one would be massive.

Last time I checked, it seemed that Chrome used disk-based Blobs for at least some of those cases.

The difference with `.json()` and the others exists because we need all the data to process a JS Objects, which is not the case for this one. 

Node.js has just landed support for disk-baker Blobs, so we will soon be able to use this ;).

jimmywarting:
> I don't think a synchronous multipart parser would work for us long term.
Node.js just gained support for disk-baked Blobs. This means we can stream files to disk and then create Blobs that way and avoid memory leaks. However we can do that only if we have a streaming multipart parser.

I agree that disk based blob/files would be the best option and that async streaming is best for not hitting the RAM limit. would rather want to have something that's working and runs a bit slower. then rather (as @mcollina puts it) "not ship a feature at all" with the risk for out-of-memory issues.

it would also be technically possible to also write all of formdata payload to one single file first and then slice it (virtually - by just changing the offset + size, of where it should start/stop reading from)

```js
const blob = fs.openAsBlob(path) // could include all formdata entries
const file1 = new File( [ blob.slice(100, 1024) ], entry.name, { type: entry.mimeType } )
```
so if you want to write all the data to the disc first and then synchronous search the file after each boundary to lookup where each file entry begin/ends then that could also be an option. 

---

currently undici url encoded formdata decoder is synchronous.
b/c it dose not include files, and probably isn't as large as formdata payloads are. and that it uses URLSearchParams to decode everything.

So i'm wondering is this: https://twitter.com/jarredsumner/status/1625067963656335361 measuring URL encoded payload or multipart formdata encoded payload, cuz it's two completely different path of solving the issue at hand.

> we could document formData's downsides 

if we are going to document the downside of formdata, then we shouldn't provide them with an alternative solution to how they should do it themself using busboy and streams. it should just work as it's intended to work.
if we want to say that multipart/formdata it's slower then that's fine, recommend that ppl instead use URL encoded payload for faster parsing or use json. multipart/formdata is intended of file uploads so that is what it should be used for. 

so if we should document anything at all, then it should be:
> if you intend to send files then by all means you can use `FormData()` to send payload. but the multipart/formdata parsing will be slower then URLEncoded payloads. so, if you intend to send just simple forms with just text, then use `new URLSearchParams()` instead of `new FormData()` when you want to decode the payload with `await body.formData()`

KhafraDev:
I agree that we probably won't find middle ground for this 😄.

>  it should just work as it's intended to work.

That's partially the issue - it doesn't work as intended, if by "intended" we're referring to the spec and/or user expectations. There are other slightly less noticeable issues and pretty crappy workarounds we're doing already to match the spec: https://github.com/nodejs/undici/blob/06f77a92087f18151f1ed8c7eb25ad44351ba508/lib/fetch/body.js#L461-L462.

---

Anyways, this branch is mostly done, passes all of busboy's tests, passes the same set of WPTs that are enabled. I'll finish up the docs eventually and I think we're good to go? Although I've been holding off because I know there will inevitably be issues that I won't have motivation to fix...

lib/formdata/parser.js
maximelkin: Sorry for being noisy.
Is this code is intended to skip until meet CR **or** LF?
"blablaCR actual finish here CRLF" => skip only to first CR position.

KhafraDev: It looks for CRLF and returns everything before that

ronag:
@KhafraDev What's the status on this?

KhafraDev:
It's mostly done, just need to spent a couple hours adding docs and cleaning stuff up. Is there a way to opt-in to using the file backed Blobs, or is it automatic?

jimmywarting:
> Is there a way to opt-in to using the file backed Blobs, or is it automatic?

Perhaps if `content-length` is lower than x bytes then it would construct blobs in memory?
Also what if you where to use `response.blob()`?

ronag:
@KhafraDev did you give up on this?

KhafraDev:
yes, the parsing should be done synchronously to match the spec.

PR #1854, opened 2023-01-12
fix(websocket): simplify typedarray copying

KhafraDev:
... duh, really weird "solution" I came up with, I guess.


mcollina (approved):
lgtm

PR #1858, opened 2023-01-16
perf: allow keep alive for HEAD requests

ronag:
One of the performance bottlenecks we have is that we don't allow keep-alive for HEAD requests due to common server-side  implementation problems where we can't reliably detect if the server is properly ending the HEAD request. However, in the cases where we know the target server it would be nice to be able to explicitly allow this.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -441,6 +441,7 @@ class Parser {
 
     this.keepAlive = ''
     this.contentLength = ''
+    this.connection = ''
     this.maxResponseSize = client[kMaxResponseSize]
   }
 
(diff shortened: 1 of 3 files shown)


ronag:
I tried implementing this by setting `reset: false` explicitly allows keep alive for HEAD requests. However, it seems the parser always sets `shouldKeepAlive` to false in the case of HEAD requests. I'm not sure why... might need a little help here...

@KhafraDev

ronag:
In llhttp:

https://github.com/nodejs/llhttp/blob/main/src/native/http.c#L118

ronag:
v18 failure seems unrelated to PR? @KhafraDev can you help?

KhafraDev:
I'll take a look at it, I think I know what's happening

lib/client.js
```diff
@@ -709,7 +712,11 @@ class Parser {
     assert.strictEqual(this.timeoutType, TIMEOUT_HEADERS)
 
     this.statusCode = statusCode
-    this.shouldKeepAlive = shouldKeepAlive
+    this.shouldKeepAlive = (
+      shouldKeepAlive ||
+      // Override llhttp value which does not allow keepAlive for HEAD.
```
anonrig, line 717: Should we add a TODO in here?

mcollina (approved):
lgtm

PR #1868, opened 2023-01-19
fix: don't load websockets w/o crypto support

KhafraDev:


ronag (approved):

ronag (approved):

mcollina (approved):
lgtm

jimmywarting:
Q: Dose web crypto come with every new-ish NodeJS version? is it only `node:crypto` that lands in some build versions? 

Follow up Q (if web crypto is available in all builds): Can we use web crypto instead whenever it's available and perhaps fallback to `node:crypto` instead?

mcollina:
If there is no crypto it's because  Node was built without OpenSSL. This is a remote case, but we also test this in Node CI.

mcollina:
@KhafraDev why are we building Websocket in the Nodejs bundle? We don't use it there (yet).

KhafraDev:
@mcollina we aren't, I don't really know why it's being imported. Thinking it over, the issue might not be fixed since the node bundle shouldn't be using index.js?

merceyz:
> I don't really know why it's being imported

The esbuld bundle analyser says the following:
```
lib/websocket/websocket.js
This file is included in the bundle because:

Entry point index-fetch.js contains:
require("./lib/fetch");

Required file lib/fetch/index.js contains:
require("../../index");

Required file index.js contains:
require("./lib/websocket/websocket");

So required file lib/websocket/websocket.js is included in the bundle.
```

Issue #1864, opened 2023-01-19
UND_ERR_HEADERS_TIMEOUT received even when setting a larger headersTimeout in dispatcher for fetch()


xconverge:
## Bug Description

I am trying to make a fetch request that can take 5+ minutes.

I make a fetch() request with a custom Agent containing atleast these parameters (I have also overridden connect for overriding checkServerIdentity())

```
    headersTimeout: 3000000,
    bodyTimeout: 3000000,
```

fetch() call looks like this:
```
      fetch( url,  {
          dispatcher: this.apiParameters.httpAgent,
          method: "get",
        }
```

yet after about ~200 seconds I am getting a headers timeout

```
TypeError: fetch failed
    at fetch (/node_modules/undici/index.js:105:13)
    at location I called unidici fetch) {
  cause: HeadersTimeoutError: Headers Timeout Error
      at Timeout.onParserTimeout [as _onTimeout] (/node_modules/undici/lib/client.js:902:28)
      at listOnTimeout (node:internal/timers:566:11)
      at process.processTimers (node:internal/timers:507:7) {
    code: 'UND_ERR_HEADERS_TIMEOUT'
  }
}
```

## Reproducible By

Performing a LONG fetch operation

## Expected Behavior

I would expect some combination of me setting headerTimeout to either 0 or a very high number to not trigger this timeout.

## Environment

Ubuntu 20 in WSL, Node 18

### Additional context

I have seen some stuff related to "blocking" https://undici.nodejs.org/#/docs/api/Dispatcher but am unable to figure out where to set it and am unsure if its even related

mcollina:
How is the agent instantiated?

xconverge:
> How is the agent instantiated?

I just got this sample working. Changing the delay at the top to less than 4 is fine, and greater than 5 minutes and it fails

I was able to do this same test with a Client and it seems to work fine, so something specific to fetch() I think

```
"use strict";

const { test } = require("tap");

const { Agent } = require("../../");
const { createServer } = require("http");
const FakeTimers = require("@sinonjs/fake-timers");

// When this is set to less than 4 it works, when set to 5 or greater it fails and returns a header error
const minutes = 6;

const msToDelay = 1000 * 60 * minutes;

test("Realllly long for a single fetch()", (t) => {
  t.plan(1);

  const clock = FakeTimers.install();
  t.teardown(clock.uninstall.bind(clock));

  const server = createServer((req, res) => {
    setTimeout(() => {
      res.end("hello");
[...]

Issue #1373, opened 2022-04-26
Default fetch timeout too short


kyrylkov:
By default fetch times out after 30 seconds.  This causes issues like https://github.com/nodejs/undici/issues/1248

Chrome has 300 second time out.

Please set default unidici fetch timeout to 300 seconds by default.

kyrylkov:
@ronag Would it make sense to change default request timeout undici-wide to 120 seconds (presumably default in Node.js) or 300 seconds (Chrome)?

mcollina:
What's the current timeout?

kyrylkov:
30 seconds

mcollina:
Sure thing, let's bring it at least to 120 to match Node.js.

kyrylkov:
https://github.com/nodejs/undici/pull/1384

kibertoad:
Now that https://github.com/nodejs/undici/pull/1386 was merged, should this be closed?

kyrylkov:
What the default timeout now?  I don't see 300_000 anymore set in `lib/fetch/index.js`?

ronag:
> The default time out is only 30 seconds again? I don't see 5 minutes anymore set in `lib/fetch/index.js`.
> 
> @mcollina

That caused other problems... See https://github.com/nodejs/undici/pull/1870

ronag:
I would suggest we bump the undici defaults then...

kyrylkov:
The problem is that we using undici fetch bundled in Node.js

Is there even a way to override this in such case?

Setting to Chrome default 5 minutes would be great.

In the mean time, can we use `undici` to override like this?

```
fetch(webservicesUrl, {
  dispatcher: new Agent({
    bodyTimeout: 10 * 60e3, // 10 minutes
  }
});
```

targos:
You can make Node.js fetch use a custom dispatcher with:

```js
globalThis[Symbol.for('undici.globalDispatcher.1')] = dispatcher;
```

kyrylkov:
@ronag https://github.com/nodejs/undici/pull/1924

pether-b:
I'm also interested in learning how to override the `headersTimeout` when using Node's `fetch()`. I just tried 
```
globalThis[Symbol.for("undici.globalDispatcher.1")] = new Agent({
  headersTimeout: 10000,
});
```
inspired by @targos ' suggestion above, and 
```
http.globalAgent = new Agent({
  headersTimeout: 10000,
});
```
[...]

PR #1870, opened 2023-01-19
Fix fetch parameters not being applied correctly

xconverge:
Fixes https://github.com/nodejs/undici/issues/1864


lib/fetch/index.js
```diff
@@ -1951,8 +1951,6 @@ async function httpNetworkFetch (
         body: fetchParams.controller.dispatcher.isMockActive ? request.body && request.body.source : body,
         headers: request.headersList[kHeadersCaseInsensitive],
         maxRedirections: 0,
-        bodyTimeout: 300_000,
-        headersTimeout: 300_000,
```
xconverge, line 1955: this would overwrite what was set in the agent a few lines up in ```const agent = fetchParams.controller.dispatcher```

xconverge:
Not sure how this directly undoes https://github.com/nodejs/undici/pull/1386 and the test still passes, will look into it and see if I am missing something obvious?

mcollina:
@ronag @KhafraDev why is `fetch()` completely overriding the agent options? I think it'd be best to stick to what is configured in the agent.

(I think I missed https://github.com/nodejs/undici/pull/1386, otherwise I would have objected then. It was a busy period of my life.).

ronag (changes requested):
I think the fix here is simply that fetch doesn't pass it's own timeout values. Why do we override them in fetch?

KhafraDev:
They are overridden in fetch to match chrome's behavior. I agree that the fix is to remove them.

xconverge:
Do I need to adjust any defaults anywhere to preserve behavior or simply delete the 2 lines in lib/fetch/index.js? And remove my 300 sec test that was aiming to enforce the previous feature request?

KhafraDev:
removing the 2 lines is fine, along with the other changes

xconverge:
Ok this is ready for a re-review, it was my original changeset as well until I realized it was undo-ing #1373 and #1386

If you want me to add the 300 second tests back in from my previous commit and change the defaults elsewhere in the agent, please let me know, but for now at least the ability to override in fetch exists/works and a test that makes sure of that is in (unless the default gets changed to a much larger number in the future)

xconverge:
Oops sorry I didn't know clicking re-request review on a change request removes the other reviewers

KhafraDev (approved):

mcollina (approved):
lgtm

xconverge:
@mcollina @ronag what happens if you set bodyTimeout and headersTimeout to 20 minutes, but do NOT set keepAliveTimeout (default is 10 minutes), and a request takes say 15 minutes.

Looking at this code, it kind of looks to me like the request will just die and eventually timeout at the 20 minutes and return a bodyTimeout or headersTimeout. This is what I am experiencing (I know this is wack to have such long requests... but I am starting by just porting this request from axios which works)

```
    if (this.shouldKeepAlive && client[kPipelining]) {
      const keepAliveTimeout = this.keepAlive ? util.parseKeepAliveTimeout(this.keepAlive) : null

      if (keepAliveTimeout != null) {
        const timeout = Math.min(
          keepAliveTimeout - client[kKeepAliveTimeoutThreshold],
          client[kKeepAliveMaxTimeout]
        )
        if (timeout <= 0) {
          socket[kReset] = true
        } else {
          client[kKeepAliveTimeoutValue] = timeout
        }
      } else {
        client[kKeepAliveTimeoutValue] = client[kKeepAliveDefaultTimeout]
      }
    } else {
      // Stop more requests from being dispatched.
      socket[kReset] = true
    }
```

I will increase setting keepAliveTimeout but am wondering if there is a more elegant way to handle this so that the errors I get make sense

mcollina:
I'm failing to understand how that relates to this PR. If not, please open a new issue.

xconverge:
> I'm failing to understand how that relates to this PR. If not, please open a new issue.

Ok fair enough, I have little to no information to put in a new issue yet so was hoping to just get some guidance. I was hopeful that this PR fixed my long request issue and it does not, so to me there is something about undici that I continue to not understand for a long request. I am glad I was able to write a tangible test for this particular issue, so I will see if I can figure out how to write a test case for what I am still experiencing.

xconverge:
This is why I thought it was related, because it allowed me to expose the next issue basically https://github.com/nodejs/undici/issues/1887

PR #1871, opened 2023-01-20
chore(performance#38): Performance improvements

metcoder95:
While working on [Performance#38](https://github.com/nodejs/performance/issues/38), I decided to use it as starting point Undici for experimenting with possible performance improvements, as Undici has a greatly clean implementation of the WHATWG for parsing a MIMEType.

While experimenting I was able to notice substantial performance improvements while changing smaller pieces, while trying to respect them as much as possible the spec.

Was able to obtain a substantial ~48% improvement compared to the current implementation.

The PR as always is just a suggestion as pointed out in the performance thread, in Undici this is not one of the biggest worries as it is used only while parsing `data:` strings 🙂

Feel free to close it if there's no value added.

### Machine
```
Hardware Overview:
	Model Name: MacBook Pro
	Model Identifier: MacBookPro18,1
	Chip: Apple M1 Pro
	Total Number of Cores: 10 (8 performance and 2 efficiencies)
	Memory: 16 GB
	System Firmware Version: 8419.60.44
	OS Loader Version: 7459.141.1
```

### Results
**Code**
```js
const str = 'application/json; charset=utf-8'

suite
  .add('util#MIMEType', function () {
    new util.MIMEType(str)
  })
  .add('undici#parseMIMEType', function () {
    parseMIMEType(str)
  })
  .add('undici#parseMIMEType(original)', function () {
    parseMIMETypeOriginal(str)
  })
  .add('fast-content-type-parse#parse', function () {
    fastContentType.parse(str)
  })
  .add('fast-content-type-parse#safeParse', function () {
    fastContentType.safeParse(str)
  })
  .on('cycle', function (event) {
    console.log(String(event.target))
  })
  .on('complete', function () {
    console.log('Fastest is ' + this.filter('fastest').map('name'))
  })
  .run({ async: true })
```

```
util#MIMEType x 1,330,919 ops/sec ±0.36% (92 runs sampled)
undici#parseMIMEType x 2,366,624 ops/sec ±0.22% (98 runs sampled)
undici#parseMIMEType(original) x 1,588,546 ops/sec ±0.67% (98 runs sampled)
```

### DIFF
--- a/lib/fetch/dataURL.js
+++ b/lib/fetch/dataURL.js
@@ -264,9 +264,7 @@ function parseMIMEType (input) {
     /** @type {Map<string, string>} */
     parameters: new Map(),
     // https://mimesniff.spec.whatwg.org/#mime-type-essence
-    get essence () {
-      return `${this.type}/${this.subtype}`
-    }
+    essence: `${type}/${subtype}`
   }
 
   // 11. While position is not past the end of input:
@@ -362,7 +360,7 @@ function parseMIMEType (input) {
     if (
       parameterName.length !== 0 &&
       HTTP_TOKEN_CODEPOINTS.test(parameterName) &&
-      !HTTP_QUOTED_STRING_TOKENS.test(parameterValue) &&  // eslint-disable-line
+      !HTTP_QUOTED_STRING_TOKENS.test(parameterValue) &&
       !mimeType.parameters.has(parameterName)
     ) {
       mimeType.parameters.set(parameterName, parameterValue)
--- a/lib/fetch/dataURL.js
+++ b/lib/fetch/dataURL.js
@@ -5,6 +5,12 @@ const { isValidHTTPToken, isomorphicDecode } = require('./util')
 
 const encoder = new TextEncoder()
 
+// Regex
+const HTTP_TOKEN_CODEPOINTS = /^[!#$%&'*+-.^_|~A-z0-9]+$/
+const HTTP_WHITESPACE_REGEX = /(\u000A|\u000D|\u0009|\u0020)/ // eslint-disable-line
+// https://mimesniff.spec.whatwg.org/#http-quoted-string-token-code-point
+const HTTP_QUOTED_STRING_TOKENS = /^(\u0009|\x{0020}-\x{007E}|\x{0080}-\x{00FF})+$/ // eslint-disable-line
+
 // https://fetch.spec.whatwg.org/#data-url-processor
 /** @param {URL} dataURL */
 function dataURLProcessor (dataURL) {
@@ -217,7 +223,7 @@ function parseMIMEType (input) {
   // 4. If type is the empty string or does not solely
   // contain HTTP token code points, then return failure.
   // https://mimesniff.spec.whatwg.org/#http-token-code-point
-  if (type.length === 0 || !/^[!#$%&'*+-.^_|~A-z0-9]+$/.test(type)) {
+  if (type.length === 0 || !HTTP_TOKEN_CODEPOINTS.test(type)) {
     return 'failure'
   }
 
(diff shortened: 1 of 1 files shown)


lib/fetch/dataURL.js
```diff
@@ -258,9 +264,7 @@ function parseMIMEType (input) {
     /** @type {Map<string, string>} */
     parameters: new Map(),
     // https://mimesniff.spec.whatwg.org/#mime-type-essence
-    get essence () {
-      return `${this.type}/${this.subtype}`
-    }
+    essence: `${type}/${subtype}`
```
metcoder95, line 267: This was the main driver for the improvements. Tried to find something in the spec that indicates this is a `getter` but sadly couldn't find any point indicating it. I might overlook it, but @KhafraDev can you confirm this approach is ok?

ronag: This won't work as it can be overwritten...

metcoder95: That was what I was missing then. Does that mean that the mime type object can be altered after creation, right?

ronag: Yes, and it should be immutable.

ronag: Actually there is more stuff here than should get getters?

ronag: I might be wrong, but then it becomes a case of consistency, if you change subtype then essence need to be automatically updated, which is why a getter is necessary either way.

metcoder95: >Actually there is more stuff here than should get getters?

You mean the PR itself, then yes? Otherwise, I might be missing something

>I might be wrong, but then it becomes a case of consistency, if you change subtype then essence need to be automatically updated, which is why a getter is necessary either way

Yeah, once you said that about that must be a getter, that is what I started to see, as most likely the issue is not on the getter itself but the recomputation of the `string` on every `get` call.

metcoder95: It seems is actually more about the string computation, as if I replace it with the following class that computes the `essence` on instantiation, and on set/get (kinda caching), the results remains stable.

**Class**:
```js
class MimeType {
  #type = ''
  #subtype = ''
  #essence = ''
  #parameters = new Map()

  constructor ({ type, subtype }) {
    this.#type = type
    this.#subtype = subtype
    this.#essence = `${type}/${subtype}`
  }

  set type (value) {
    this.#type = value
    this.#essence = `${type}/${subtype}`
  }

  get type () {
    this.#type
  }

  set subtype (value) {
    this.#subtype = value
    this.#essence = `${type}/${subtype}`
  }

  get subtype () {
    return this.#subtype
  }

  get essence () {
    return this.#essence
  }

  get parameters () {
    return this.#parameters
  }
}```

**Scenario**
```
const str = 'application/json; charset=utf-8'

suite
  .add('util#MIMEType', function () {
    new util.MIMEType(str)
  })
  .add('undici#parseMIMEType', function () {
    parseMIMEType(str).essence
  })
  .add('undici#parseMIMEType(original)', function () {
    parseMIMETypeOriginal(str).essence
  })
  .add('fast-content-type-parse#parse', function () {
    fastContentType.parse(str)
  })
  .add('fast-content-type-parse#safeParse', function () {
    fastContentType.safeParse(str)
  })
  .on('cycle', function (event) {
    console.log(String(event.target))
  })
  .on('complete', function () {
    console.log('Fastest is ' + this.filter('fastest').map('name'))
  })
  .run({ async: true })```

H4ad:
I read the changes and these lines get my attention:

https://github.com/nodejs/undici/blob/d7cd095a053539d42c4a65fccea412832c4a2312/lib/fetch/dataURL.js#L325-L331

Is part of the spec but is not used anymore, either the result of either the position variable, could this be removed?

Also in this part:

https://github.com/nodejs/undici/blob/d7cd095a053539d42c4a65fccea412832c4a2312/lib/fetch/dataURL.js#L360-L365

I think that worth to put the validation of `!mimeType.parameters.has(parameterName)` after `parameterName.length !== 0`, maybe we could extract a little bit of performance when the parameter already was defined because _I think_ is faster than run the regex first.

KhafraDev (approved):

mcollina (approved):
lgtm

Issue #1872, opened 2023-01-20
More strict isFormDataLike


ronag:
`isFormDataLike` currently only checks that the `constructor.name` is `'FormData'`. We should also check the interface as there are npm packages that have the correct name but wrong interface causing confusion where it's read as a FOrmData instead of a async generator..

jimmywarting:
> there are npm packages that have the correct name but wrong interface

hmm, which one?

ronag:
https://www.npmjs.com/package/form-data

mateonunez:
Hey @ronag, what about the following check?

```js
function isFormDataLike (chunk) {
  return (
    (chunk &&
      chunk.constructor && chunk.constructor.name === 'FormData' &&
      /^(FormData)$/.test(chunk[Symbol.toStringTag])
    )
  )
}
```

Could this control be sufficient in your opinion? It checks if the instance of `chunk` contains the `Symbol.toStringTag`. It's extendible with other thirds parts (ex: formdata-node) and prevents confusion between packages like: `form-data`.

ronag:
I'm not sure that is sufficient

ronag:
That has the problem in the other direction. Third party module that as FormData like might not work.

PR #1875, opened 2023-01-21
perf(util): improve isFormDataLike checks

mateonunez:
This PR would implement #1872.

## Context

Actually, the `FormData` class needs the following declaration to work:

```js
  class FormData {
    constructor () {}

    append () {}
    delete () {}
    get () {}
    getAll () {}
    has () {}
    set () {}
    entries () {}
    keys () {}
    values () {}
    forEach () {}
  }
```

The `isFormDataLike` method was improved by checking the `constructor.name` and the declaration of those methods. This makes that control more strict and flexible. The check would be linear, there is no way to guess what API is requested by the `chunk`. This check allows **third-party libraries** to be compatible with this control.

## Tests

The tests were updated by adding some cases where the check should return false or true.

### DIFF
--- a/lib/core/util.js
+++ b/lib/core/util.js
@@ -354,8 +354,23 @@ function ReadableStreamFrom (iterable) {
   )
 }
 
+// The chunk should be a FormData instance and contains
+// all the required methods.
 function isFormDataLike (chunk) {
-  return chunk && chunk.constructor && chunk.constructor.name === 'FormData'
+  return (chunk &&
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

ronag:
Can you add a test for the npm package form-data. An instance of that should not be formdata like

mateonunez:
> Can you add a test for the npm package form-data. An instance of that should not be formdata like

Sure! I've installed the package onto `devDependencies` and included it in the tests.

PR #1877, opened 2023-01-22
Add feature to specify custom headers for proxies

### LINKED ISSUE #1876: Allow setting custom kProxyHeaders
## This would solve...

ProxyAgent currently only allows configuring the `Authorization` header for its CONNECT request to the proxy. I'd like to add different custom headers (e.g. User-Agent) as well. This header should only be sent in the CONNECT request, not in the tunneled requests to the target URL.

I think providing a way to add more headers to `kProxyHeaders` - possibly just an arbitrary headers object in the options param?

Sebmaster:
This adds a custom headers option in the ProxyAgent params, which allows specifying arbitrary headers. These will only provided in the initial CONNECT request to the proxy and never to the tunneled requests (they get their own headers specified in the request()).

Fixes #1876.

### DIFF
--- a/lib/proxy-agent.js
+++ b/lib/proxy-agent.js
@@ -53,7 +53,7 @@ class ProxyAgent extends DispatcherBase {
 
     this[kRequestTls] = opts.requestTls
     this[kProxyTls] = opts.proxyTls
-    this[kProxyHeaders] = {}
+    this[kProxyHeaders] = opts.headers || {}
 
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

RafaelGSS (approved):
LGTM.

mcollina (approved):
lgtm

PR #1879, opened 2023-01-24
Make test pass in v19.x

mcollina:
Signed-off-by: Matteo Collina <hello@matteocollina.com>

PR #1880, opened 2023-01-24
fix(wpts): Blob is a global getter in >=v19.x.x

KhafraDev:
refs: https://github.com/nodejs/undici/pull/1879#issuecomment-1401966074

I already knew the bug would occur, just forgot about it honestly.


ronag (approved):

mcollina (approved):
lgtm

PR #1884, opened 2023-01-25
wpt: make runner more resilient

KhafraDev:
I have no idea if any of these changes will have an effect, but hopefully they will help on windows.

- Hide experimental warnings from websocket tests
- Limit number of workers to the number of cpus available (about 2-3 seconds slower locally, but less resource intensive)
- If a message can't be sent to a child process, exit the process rather than doing nothing (this *might* be the hanging issue)
- detect when to shutdown using a more reliable system (listening to the exit event on the child process and exiting once the subprocess exited)


mcollina (approved):
lgtm

anonrig (approved):

PR #1888, opened 2023-01-26
fix(types): update error type definitions

### LINKED ISSUE #1883: [types] some error type definitions are stale or missing
## Bug Description

The undici error types provided [here](https://github.com/nodejs/undici/blob/main/types/errors.d.ts) do not correspond to all errors declared [here](https://github.com/nodejs/undici/blob/main/lib/core/errors.js). 

## Reproducible By

A couple of examples:
* The `ConnectTimeoutError` is not available in the `errors` type definitions
* Using `SocketTimeoutError` in a type check like
    ```typescript
    if (error instanceof errors.SocketTimeoutError)
    ```
    produces an `right-hand side of 'instanceof' is not an object` error.

## Expected Behavior

All core error types are defined.

## Environment

macOS 13.1, node v18.5.0

rafaelcr:
This PR adds missing type definitions for errors and removes others that were no longer in use. It also makes sure all `name`s and `code`s match the JS implementations.

Fixes #1883

### DIFF
--- a/test/types/errors.test-d.ts
+++ b/test/types/errors.test-d.ts
@@ -4,29 +4,44 @@ import Client from '../../types/client'
 
 expectAssignable<errors.UndiciError>(new errors.UndiciError())
 
+expectAssignable<errors.UndiciError>(new errors.ConnectTimeoutError())
+expectAssignable<errors.ConnectTimeoutError>(new errors.ConnectTimeoutError())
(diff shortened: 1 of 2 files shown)


KhafraDev (approved):

mcollina:
Can you please rebase your PR? The test did not run

rafaelcr:
Done @mcollina

KhafraDev (approved):

PR #1893, opened 2023-01-28
Implement the requirement of IncomingHttpHeaders["set-cookie"]

### LINKED ISSUE #1892: The type of set-cookies obtained by heaers is inconsistent with the actual type?
## Bug Description

`headers["set-cookie"]`returned is a string, type declaration is a `string [] | undefined`?

```js
 const { headers } = await request(index_url);
 console.log(headers);
/* {
  date: 'Sat, 28 Jan 2023 09:43:48 GMT',
  'content-type': 'text/html',
  'content-length': '2920',
  connection: 'keep-alive',
  vary: 'Accept-Encoding',
  'last-modified': 'Sat, 17 Sep 2022 08:52:52 GMT',
  'accept-ranges': 'bytes',
  etag: '"1d8ca72dec58168"',
  'set-cookie': 'SERVERID=f7a91655448e700d4049e6f0ca2b6b52|1674899028|1674899028;Path=/'
} */
console.log(headers["set-cookie"]);
/*
SERVERID=f7a91655448e700d4049e6f0ca2b6b52|1674899028|1674899028;Path=/
*/
```

pan93412:
## This relates to...

Fixed #1892

## Rationale

Currently `headers["set-cookie"]` does not split, which can't match the expected type of IncomingHttpHeader.

## Changes

This commit attempts to split `headers["set-cookie"]` by `;` like what 'node:http' does: <https://nodejs.org/api/http.html#class-httpserverresponse>

### Bug Fixes

See above.

- `set-cookie` is now always an `string[]`.

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/core/util.js
+++ b/lib/core/util.js
@@ -209,6 +209,10 @@ function parseKeepAliveTimeout (val) {
   return m ? parseInt(m[1], 10) * 1000 : null
 }
 
+function splitCookie (val) {
+  return val.toString().split(';')
+}
+
 function parseHeaders (headers, obj = {}) {
   for (let i = 0; i < headers.length; i += 2) {
     const key = headers[i].toString().toLowerCase()
@@ -216,6 +220,8 @@ function parseHeaders (headers, obj = {}) {
     if (!val) {
       if (Array.isArray(headers[i + 1])) {
         obj[key] = headers[i + 1]
+      } else if (key === 'set-cookie') {
+        obj[key] = splitCookie(headers[i + 1])
       } else {
         obj[key] = headers[i + 1].toString()
       }
@@ -224,7 +230,12 @@ function parseHeaders (headers, obj = {}) {
         val = [val]
         obj[key] = val
       }
-      val.push(headers[i + 1].toString())
+
+      if (key === 'set-cookie') {
+        val.push(...splitCookie(headers[i + 1]))
+      } else {
+        val.push(headers[i + 1].toString())
+      }
     }
   }
   return obj
--- a/test/util.js
+++ b/test/util.js
@@ -83,12 +83,27 @@ test('validateHandler', (t) => {
 })
 
 test('parseHeaders', (t) => {
-  t.plan(5)
+  t.plan(7)
   t.same(util.parseHeaders(['key', 'value']), { key: 'value' })
   t.same(util.parseHeaders([Buffer.from('key'), Buffer.from('value')]), { key: 'value' })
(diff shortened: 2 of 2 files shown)


pan93412:
Note that there are at least 1 unmatched behavior of [IncomingHttpHeaders](https://nodejs.org/api/http.html#messageheaders):

- [ ] Duplicates of age, authorization, content-length, content-type, etag, expires, from, host, if-modified-since, if-unmodified-since, last-modified, location, max-forwards, proxy-authorization, referer, retry-after, server, or user-agent are discarded. To allow duplicate values of the headers listed above to be joined, use the option joinDuplicateHeaders in [http.request()](https://nodejs.org/api/http.html#httprequestoptions-callback) and [http.createServer()](https://nodejs.org/api/http.html#httpcreateserveroptions-requestlistener). See RFC 9110 Section 5.3 for more information.
- [x] set-cookie is always an array. Duplicates are added to the array.
- [ ] For duplicate cookie headers, the values are joined together with `;` .
- [ ] For all other headers, the values are joined together with `,` .

mcollina:
I'm a bit lost with this PR. The original issue refers a typescript problem. This change alters the current behavior in code to match the types.

Let's do the other way around: the types should follow what the code implements. This change can ship in v5.

A breaking change to match the RFC or the current behavior of node is ok, but it will have to ship on v6.

pan93412:
> I'm a bit lost with this PR. The original issue refers a typescript problem. This change alters the current behavior in code to match the types.
> 
> Let's do the other way around: the types should follow what the code implements. This change can ship in v5.
> 
> A breaking change to match the RFC or the current behavior of node is ok, but it will have to ship on v6.

Sure, let me fix the type to match what `parseCookie` current does. I'll currently close it.

tremby:
I found my way here by way of being confused that `IncomingHttpHeaders` in undici is not the same as `IncomingHttpHeaders` from node:http. In particular I was getting broken behaviour with `set-cookie` headers, which Typescript wasn't catching.

I think the specific problem in my case is a bug in urllib which I reported as https://github.com/node-modules/urllib/issues/467. (I'm using that package only because I need HTTP digest auth, and that was the first working implementation I could find. Open to suggestions.) Urllib makes what I believe is a flawed assumption that the `IncomingHttpHeaders` undici provides is the same as the node:http `IncomingHttpHeaders`. This seems like an easy mistake to make; I wonder if other downstream projects have done the same.

So this is just me adding my upward thumb to the idea of changing the behaviour to match node's in a future version.

PR #1895, opened 2023-01-29
feat: expose content-type parser

### LINKED ISSUE #1891: exposing the content-type parser?
Since undici's parser consistently outperforms node's, should we consider exposing it? It already passes the entirety of the [mimesniff](https://mimesniff.spec.whatwg.org/) [tests](https://github.com/web-platform-tests/wpt/tree/master/mimesniff).

_Originally posted by @KhafraDev in https://github.com/nodejs/performance/issues/38#issuecomment-1407118885_

KhafraDev:
Closes #1891

Exposes parse and stringify utility functions for content-type parsing. It's only exposed in v16+ because it relies on fetch utils (which imports `util/types` which wasn't added until v15.3.0).

### DIFF
--- a/docs/api/ContentType.md
+++ b/docs/api/ContentType.md
@@ -1,0 +1,57 @@
+# MIME Type Parsing
+
+## `MIMEType` interface
+
+* **type** `string`
+* **subtype** `string`
+* **parameters** `Map<string, string>`
+* **essence** `string`
+
+## `parseMIMEType(input)`
+
+Implements [parse a MIME type](https://mimesniff.spec.whatwg.org/#parse-a-mime-type).
+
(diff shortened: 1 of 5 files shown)


ronag (approved):

mcollina (changes requested):
Can you add some docs?

ronag (approved):

KhafraDev:
I've added docs, I knew I forgot something 😄

mcollina (approved):
lgtm

metcoder95 (commented):
Do you think that it might have value to also expose the MIME object as a class, as well for MIMEParams?

KhafraDev:
I don't see the issue with it being an object./params being a Map.

metcoder95:
Me neither, just wanted to see if there was a value of some sort in doing so. As this is the way is exposed by Node.js core, but nothing else 🙂

PR #1896, opened 2023-01-29
Correct the type of DispatchOptions["headers"]

### LINKED ISSUE #1892: The type of set-cookies obtained by heaers is inconsistent with the actual type?

pan93412:
## This relates to...

Fixed #1892 
The simpler solution of #1893

## Rationale

According to core/util@parseHeaders, the current behavior is:

- By default, the entry type is a `string`.
- The entry type turns to a `string[]` when there are duplicated entries.

This behavior is not as same as what `node:http` does currently; therefore, `IncomingHttpHeaders` can't reflect our parsed data.

## Changes

image

This commit attempts to reflect this parsing behavior by redefining it to `Record<string, string | string[]>`.

### Features

N/A

### Bug Fixes

See Changes.

- The old workaround to this wrong declaration should be removed now.

## Status

- [x] Tested
- [ ] ~~Benchmarked (**optional**)~~
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

Note that I can't pass `npm run test`. Seems not related:

image


pan93412:
request for review: @mcollina @ronag

mcollina (approved):
lgtm

types/header.d.ts
```diff
@@ -0,0 +1,4 @@
+/**
+ * The header type declaration of `undici`.
+ */
+export type IncomingHttpHeaders = Record<string, string | string[]>;
```
SimenB, line 4: any particular reason why `undefined` is not allowed? Now it's a type error to pass `headers` from an incoming http request into `undici`

pan93412: I think it is just a mistake – no particular reason there. Maybe we can open another PR to fix it?

SimenB: cool 👍 I personally can't think of a reason to disallow `undefined` at least

PR #1899, opened 2023-02-01
fix: ensure header value is a string

ronag:
I don't think it's something else than a string in any of the branches but the check should still be correct.

### DIFF
--- a/test/client-dispatch.js
+++ b/test/client-dispatch.js
@@ -94,7 +94,7 @@ test('basic dispatch get', (t) => {
     t.equal(`localhost:${server.address().port}`, req.headers.host)
     t.equal(undefined, req.headers.foo)
     t.equal('bar', req.headers.bar)
-    t.equal('null', req.headers.baz)
+    t.equal('', req.headers.baz)
     t.equal(undefined, req.headers['content-length'])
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

ronag:
So this was not as easy as I thought. We allow other header values than strings. I propose:

- If object, throw
- If null, empty string
- Otherwise stringify with `${val}`

ronag:
@mcollina PTAL

test/client-dispatch.js
```diff
@@ -94,7 +94,7 @@ test('basic dispatch get', (t) => {
     t.equal(`localhost:${server.address().port}`, req.headers.host)
     t.equal(undefined, req.headers.foo)
     t.equal('bar', req.headers.bar)
-    t.equal('null', req.headers.baz)
+    t.equal('', req.headers.baz)
```
ronag, line 97: @mcollina do we consider this breaking?

mcollina: It's actually better.

KhafraDev (approved):

mcollina (approved):
lgtm

PR #1901, opened 2023-02-01
fix: throw if invalid content-type header

ronag:
If content-type header value is not a string we should throw.


mcollina:
CI seems not happy

mcollina (approved):
lgtm

PR #1904, opened 2023-02-03
Add ability to set TCP keepalive

### LINKED ISSUE #1887: Multiple parallel long fetches with overridden, non-zero, timeouts fails all the time
## Bug Description

When running multiple LONG running fetches in parallel, some of them seem to just die and hit a timeout that doesnt really seem relevant 

```
name: HeadersTimeoutError
    message: Headers Timeout Error
    stack: >-
      HeadersTimeoutError: Headers Timeout Error
          at onParserTimeout (/src/archive/undici/lib/client.js:908:28)
          at callTimer (/src/archive/undici/node_modules/@sinonjs/fake-timer
  s/src/fake-timers-src.js:743:24)
          at doTickInner (/src/archive/undici/node_modules/@sinonjs/fake-tim
  ers/src/fake-timers-src.js:1311:29)
          at doTick (/src/archive/undici/node_modules/@sinonjs/fake-timers/s
  rc/fake-timers-src.js:1392:20)
          at Object.tick (/src/archive/undici/node_modules/@sinonjs/fake-tim
  ers/src/fake-timers-src.js:1400:20)
          at Server.<anonymous>
  (/src/archive/undici/test/fetch/fetch-timeouts.js:32:11)
          at Server.emit (node:events:513:28)
          at parserOnIncoming (node:_http_server:1065:12)
          at HTTPParser.parserOnHeadersComplete (node:_http_common:117:17)
    code: UND_ERR_HEADERS_TIMEOUT
```

## Reproducible By

Tweaking minutesPerRequest, numParallelRequests, and timeout here makes for some pretty confusing results. When timeout is 0, everything seems to work correctly. I suspect there are a few timers that are overlapping eachother or getting shared somehow.

Simply set timeout to 0 and see that the test passes, and set timeout to 30 minutes and see that it fails

```
"use strict";

const { test } = require("tap");

const { fetch, Agent } = require("../..");
const FakeTimers = require("@sinonjs/fake-timers");

// Tweaking request length can have an impact too potentially (Setting to 1 seems to pass, and setting it to 10 min fails on my machine)
const minutesPerRequest = 10;

// On my machine setting this to 0 passes, and setting it to something very high fails with a UND_ERR_HEADERS_TIMEOUT
const timeout = 1000 * 60 * (minutesPerRequest + 5);
// const timeout = 0;

// Tweaking num requests to either 1 or multiple has different impacts
const numParallelRequests = 5;

test("Fetch very long parallel requests, timeout overridden", (t) => {
  const msToDelay = 1000 * 60 * minutesPerRequest;

  t.setTimeout(undefined);
  t.plan(numParallelRequests);

  const clock = FakeTimers.install();
  t.teardown(clock.uninstall.bind(clock));

    setTimeout(() => {
      res.end("hello");
    }, msToDelay);
    clock.tick(msToDelay + 1);
  });
  t.teardown(server.close.bind(server));

  server.listen(0, () => {
    for (let i = 0; i < numParallelRequests; i++) {
      fetch(`http://localhost:${server.address().port}`, {
        path: "/",
        method: "GET",
        dispatcher: new Agent({
          headersTimeout: timeout,
          connectTimeout: timeout,
          bodyTimeout: timeout,
        }),
      })
        .then((response) => response.text())
        .then((response) => {
          t.equal("hello", response);
        })
        .catch((err) => {
          // This should not happen, a timeout error should not occur
          t.error(err);
        });
    }

    clock.tick(msToDelay - 1);
  });
});

```

## Expected Behavior

The fetches to succeed

## Environment

Node v18.13.0

xconverge:
Fixes #1887

### DIFF
--- a/lib/core/connect.js
+++ b/lib/core/connect.js
@@ -71,7 +71,7 @@ if (global.FinalizationRegistry) {
   }
 }
 
-function buildConnector ({ maxCachedSessions, socketPath, timeout, tcpKeepAliveInterval, ...opts }) {
+function buildConnector ({ maxCachedSessions, socketPath, timeout, ...opts }) {
   if (maxCachedSessions != null && (!Number.isInteger(maxCachedSessions) || maxCachedSessions < 0)) {
     throw new InvalidArgumentError('maxCachedSessions must be a positive integer or zero')
   }
@@ -79,8 +79,6 @@ function buildConnector ({ maxCachedSessions, socketPath, timeout, tcpKeepAliveI
   const options = { path: socketPath, ...opts }
   const sessionCache = new SessionCache(maxCachedSessions == null ? 100 : maxCachedSessions)
   timeout = timeout == null ? 10e3 : timeout
-
-  tcpKeepAliveInterval = tcpKeepAliveInterval == null ? 60e3 : tcpKeepAliveInterval
 
   return function connect ({ hostname, host, protocol, port, servername, localAddress, httpSocket }, callback) {
     let socket
@@ -97,8 +95,6 @@ function buildConnector ({ maxCachedSessions, socketPath, timeout, tcpKeepAliveI
 
       socket = tls.connect({
         highWaterMark: 16384, // TLS in node can't have bigger HWM anyway...
-        keepAlive: tcpKeepAliveInterval !== 0,
-        keepAliveInitialDelay: tcpKeepAliveInterval,
         ...options,
         servername,
         session,
@@ -117,13 +113,17 @@ function buildConnector ({ maxCachedSessions, socketPath, timeout, tcpKeepAliveI
       assert(!httpSocket, 'httpSocket can only be sent on TLS update')
       socket = net.connect({
         highWaterMark: 64 * 1024, // Same as nodejs fs streams.
-        keepAlive: tcpKeepAliveInterval !== 0,
-        keepAliveInitialDelay: tcpKeepAliveInterval,
         ...options,
         localAddress,
         port: port || 80,
         host: hostname
       })
+    }
+
+    // Set TCP keep alive options on the socket here instead of in connect() for the case of assigning the socket
+    if (options.keepAlive ?? true) {
+      const keepAliveInitialDelay = options.keepAliveInitialDelay === undefined ? 60e3 : options.keepAliveInitialDelay
+      socket.setKeepAlive(true, keepAliveInitialDelay)
     }
 
     const cancelTimeout = setupTimeout(() => onConnectTimeout(socket), timeout)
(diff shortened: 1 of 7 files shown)


ronag:
IMO this should be an option on the Client/Pool/Agent. Preferably under the `connect` options.

xconverge:
> IMO this should be an option on the Client/Pool/Agent. Preferably under the `connect` options.

fair enough, I will see if I can find time to assess and get to it in the next few days

I struggle a bit to find and understand the differences and building blocks of all of the elements, this library has more parts to it than I expected

If you could comment which markdown files/sections should be changed for the docs, I can make it a reality. Otherwise I can attempt to give it a spin when I get a minute

ronag:
https://undici.nodejs.org/#/docs/api/Client?id=parameter-connectoptions

types/client.d.ts
```diff
@@ -43,6 +43,8 @@ declare namespace Client {
     maxRequestsPerClient?: number;
     /** Max response body size in bytes, -1 is disabled */
     maxResponseSize?: number | null;
+    /** TCP keepalive interval for the socket in milliseconds. Default: `60e3` milliseconds (60s) */
+    tcpKeepAliveInterval?: number | null;
```
mcollina, line 47:
```suggestion
    tcpKeepAliveInterval?: number | null;
```

Can you please creating a test for this new type? We use tsd.

mcollina:
It would be great to add an automated test for this.

xconverge:
Will add tests and make sure tests pass (I pushed with --no-verify) before this is ready, but I wanted to get the latest changes staged for discussion

ronag (approved):

xconverge:
Pretty happy with where this ended up, makes so much more sense now, thanks for sticking with the discussion. Let me know if there are any changes needed now but it is pretty simple now and logical

ronag:
> Pretty happy with where this ended up, makes so much more sense now, thanks for sticking with the discussion. Let me know if there are any changes needed now but it is pretty simple now and logical

Please fix node 12 compatibility :D

lib/core/connect.js
```diff
@@ -120,6 +120,12 @@ function buildConnector ({ maxCachedSessions, socketPath, timeout, ...opts }) {
       })
     }
 
+    // Set TCP keep alive options on the socket here instead of in connect() for the case of assigning the socket
+    if (options.keepAlive ?? true) {
```
ronag, line 124:
```suggestion
    if (options.keepAlive == null || options.keepAlive) {
```

xconverge: done!

KhafraDev (approved):

PR #1906, opened 2023-02-03
fix(fetch): use semicolon for Cookie header delimiter

### LINKED ISSUE #1905: Fetch request concatenates Cookie headers with comma instead of semicolon
### Version

19.6.0

### Platform

Darwin MBP.local 22.2.0 Darwin Kernel Version 22.2.0: Fri Nov 11 02:08:47 PST 2022; root:xnu-8792.61.2~4/RELEASE_X86_64 x86_64

### Subsystem

_No response_

When specifying multiple `Cookie` headers for a fetch request, the headers will be concatenated using commas instead of semicolons. While this is _technically_ compliant with the WHATWG spec, that spec is based around browser use and causes a risk of incorrect handling by the receiving server.

Example:

```js
fetch('http://localhost:8080/', {
  headers: [
    ['cookie', 'FOO=lorem-ipsum-dolor-sit-amet'],
    ['cookie', 'BAR=the-quick-brown-fox']
  ]
});
```

_No response_

### What is the expected behavior?

Cookies are separated by semicolons (`;`).

```http
GET / HTTP/1.1
host: localhost:8080
connection: keep-alive
cookie: FOO=lorem-ipsum-dolor-sit-amet; BAR=the-quick-brown-fox
...
```

### What do you see instead?

Cookies are separated by commas (`,`).

```http
GET / HTTP/1.1
host: localhost:8080
connection: keep-alive
cookie: FOO=lorem-ipsum-dolor-sit-amet, BAR=the-quick-brown-fox
...
```

### Additional information

Similar issue but on the receiving end: https://github.com/nodejs/node/issues/11256

KhafraDev:
Fixes #1905 

In browsers, the cookie header is filtered out so there's no mention of special handling. While technically being compliant, will lead to incorrect parsing on the receiving side (from the issue). Deno matches this behavior - delimiting cookie headers with a semicolon rather than a comma, and I believe node core does similar.

### DIFF
--- a/lib/fetch/headers.js
+++ b/lib/fetch/headers.js
@@ -104,7 +104,11 @@ class HeadersList {
 
     // 2. Append (name, value) to list.
     if (exists) {
-      this[kHeadersMap].set(lowercaseName, { name: exists.name, value: `${exists.value}, ${value}` })
+      const delimiter = lowercaseName === 'cookie' ? '; ' : ', '
+      this[kHeadersMap].set(lowercaseName, {
+        name: exists.name,
+        value: `${exists.value}${delimiter}${value}`
+      })
     } else {
       this[kHeadersMap].set(lowercaseName, { name, value })
     }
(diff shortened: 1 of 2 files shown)


ronag (approved):

mcollina (approved):
lgtm

PR #1908, opened 2023-02-04
use faster timers

ronag:
Another major bottleneck I found is the constant refreshing of timers, in particular the body timeout timer which is refresh every time we get a packet.

Our current use of timers is that we have a very few timers which we update frequently where high accuracy is not required. This PR tries to implement that.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -84,48 +84,6 @@ try {
   channels.connected = { hasSubscribers: false }
 }
 
-let fastNow = Date.now()
-const fastTimers = new Set()
-const fastNowInterval = setInterval(() => {
-  fastNow = Date.now()
-
-  // TODO (perf): This can probably be optimized.
-  for (const timer of fastTimers) {
-    if (fastNow >= timer.expires) {
-      timer.expires = 0
-      timer.callback(timer.opaque)
-    }
-  }
-}, 1e3)
-if (fastNowInterval.unref) {
-  fastNowInterval.unref()
-}
-
-function setFastTimeout (callback, delay, opaque) {
-  const timer = {
-    callback,
-    delay,
-    expires: fastNow + delay,
-    opaque
-  }
-
-  fastTimers.add(timer)
-
-  return timer
-}
-
-function refreshFastTimeout (timer) {
-  if (timer) {
-    timer.expires = fastNow + timer.delay
-  }
-}
-
-function clearFastTimeout (timer) {
-  if (timer) {
-    fastTimers.delete(timer)
-  }
-}
-
 class Client extends DispatcherBase {
   constructor (url, {
     interceptors,
@@ -462,9 +420,14 @@ class Parser {
     this.ptr = this.llhttp.llhttp_alloc(constants.TYPE.RESPONSE)
     this.client = client
     this.socket = socket
-    this.timeout = null
-    this.timeoutValue = null
+
+    this.timeoutActive = false
+    this.timeoutValue = 0
     this.timeoutType = null
+    this.timeoutExpires = 0
+    this.timeoutNow = Date.now()
+    this.timeoutInterval = null
+
     this.statusCode = null
     this.statusText = ''
     this.upgrade = false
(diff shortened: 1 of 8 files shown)


ronag:
@mcollina This breaks all the tests using fake timers... thoughts on what to do about that?

mcollina:
What benefits is this giving us?

I would prefer this timer to be attached to the agent, so we can actually stop them. If this is significantly faster, why aren't we briging that optimization to Node.js itself?

ronag:
> What benefits is this giving us?

In high throughput scenarios with lots of small packets we are constantly calling `refresh` on node timers which is much more expensive than what this PR does.

> I would prefer this timer to be attached to the agent, so we can actually stop them. If this is significantly faster, why aren't we briging that optimization to Node.js itself?

I don't follow... Node has a more generic timer implementation which works for other less nisch use cases.

ronag:
@mcollina PTAL

ronag:
Note that the timers in this PR are *much* less accurate than note timers, which I believe is fine for our use case.

mcollina:
Have you got a benchmark before/after?

ronag:
Before:

[bench:run] │ undici - stream     │       1 │ 7383.29 req/sec │  ± 0.00 % │              + 395.85 % │

After:

[bench:run] │ undici - dispatch   │       1 │ 8045.52 req/sec │  ± 0.00 % │              + 380.21 % │

lib/client.js
```diff
@@ -84,6 +84,35 @@ try {
   channels.connected = { hasSubscribers: false }
 }
 
+let fastNow = Date.now()
+const fastTimers = []
+const fastNowInterval = setInterval(() => {
```
anonrig, line 89: When should we call clearInterval on this?

mcollina:
Note this is the same bottleneck of the slowloris mitigation we have in core. Works kind of the same too.

mcollina:
cc @ShogunPanda

ronag:
> Note this is the same bottleneck of the slowloris mitigation we have in core. Works kind of the same too.

slowloris?

KhafraDev:
> in particular the body timeout timer which is refresh every time we get a packet.

Rather than refreshing the timer, or making the timers less reliable (although not particularly important), it might be possible to let the timer run down, and re-calculate the delay based on the last packet received, or throw an error when the timer ends. Not sure if that sounds stupid or not lol

```js
lastPacketReceived = Date.now() // set on every packet

let timer = setTimeout(onBodyTimeout, bodyTimeout)

function onBodyTimeout () {
  if (Date.now() - lastPacketReceived >= bodyTimeout)
    throw ...

  timer = setTimeout(onBodyTimeout, bodyTimeout - (Date.now() - lastPacketReceived))
}

File: lib/client.js [hunk 2]
```diff
@@ -84,6 +84,35 @@ try {
...
+      len = fastTimers.length
+    } else {
+      if (timer.timeoutExpires && fastNow >= timer.timeoutExpires) {
+        timer.timeoutExpires = 0
+        onParserTimeout(timer)
+      }
+      idx += 1
+    }
+  }
+}, 1e3)
```
mcollina, line 111: This should be active only if there is an http going on. Given how this is implemented, every node process would get this timer.

mcollina (changes requested):

ShogunPanda:
I wonder. Wouldn't be beneficial for undici to use the same connectionsCheckingInterval I introduced for node?
So we only have one timer on the server side and one on the client side.

ronag:
> I wonder. Wouldn't be beneficial for undici to use the same connectionsCheckingInterval I introduced for node? So we only have one timer on the server side and one on the client side.

Compared to the implementation of this PR? How is that better? Also, this would require node to expose it somehow publicly.

ronag:
I haven't looked into connectionsCheckingInterval so much. But would an alternative be the other way around? Go from a c++ version to an inaccurate but fast js version?

ShogunPanda:
Well, I guess you could try and measure the impact.

For you to consider: connectionsCheckingInterval is a mechanism that every x seconds check the entire pool of opened inbound connections comparing (in the same "inaccurate" way you did here) if the timeout as gone off.
He's written in C++ (leveraging C sorted sets and thus able to easily bail out early in the loop) but I think it can be easily adapted in JS. 

The only pain is to maintain the list of opened/active connections (especially when we think about KeepAlive)

ronag:
I see. I wouldn't mind using that per se. I don't know how that would practically work in undici which is developed outside of core.

ronag:
CI looks ok 👍

lib/timers.js
```diff
@@ -0,0 +1,88 @@
...
+    }
+  }
+
+  if (fastTimers.length > 0) {
+    refreshTimeout()
+  }
+}
+
+function refreshTimeout () {
+  if (fastNowTimeout && fastNowTimeout.refresh) {
```
ronag, line 40:
```suggestion
  if (fastNowTimeout && fastNowTimeout.refresh) {
    fastNow = Date.now()
```

anonrig (approved):

lib/timers.js
```diff
@@ -0,0 +1,89 @@
...
+    fastNow = Date.now()
```
ronag, line 41:
```suggestion
```

lib/timers.js
```diff
@@ -0,0 +1,89 @@
...
+
+    this.refresh()
+  }
+
+  refresh () {
+    if (!this.active) {
+      this.active = true
+      fastTimers.push(this)
+      if (!fastNowTimeout || fastTimers.length === 1) {
+        refreshTimeout()
```
ronag, line 68:
```suggestion
        refreshTimeout()
        fastNow = Date.now()
```

mcollina (approved):
lgtm

mcollina:
CI is still very red :(

ronag:
I will fix that

ronag:
Seems ok now. Same as usual with windows on node 19.

KhafraDev:
v12 CI failed on this test

```
     # Subtest: keep-alive threshold
        1..2
        ok 1 - should not error
        not ok 2 - (unnamed test)
          ---
          at:
            line: 217
            column: 13
            file: test/client-keep-alive.js
            type: Timeout
            method: _onTimeout
          stack: |
            Timeout._onTimeout (test/client-keep-alive.js:217:13)
          source: |2
                    const timeout = setTimeout(() => {
                      t.fail()
            ------------^
                    }, 2e3)
                    client.on('disconnect', () => {
          ...
        
        # failed 1 of 2 tests
    not ok 6 - keep-alive threshold # time=2017.2ms
```

PR #1910, opened 2023-02-06
fix(fetch): raise AbortSignal max event listeners

KhafraDev:
Fixes https://github.com/nodejs/node/issues/46525
Refs https://github.com/nodejs/undici/issues/939#issuecomment-1419471112

### DIFF
--- a/lib/fetch/request.js
+++ b/lib/fetch/request.js
@@ -28,6 +28,7 @@ const { getGlobalOrigin } = require('./global')
 const { URLSerializer } = require('./dataURL')
 const { kHeadersList } = require('../core/symbols')
 const assert = require('assert')
+const { setMaxListeners, getEventListeners, defaultMaxListeners } = require('events')
 
 let TransformStream = globalThis.TransformStream
 
@@ -352,6 +353,11 @@ class Request {
         const abort = function () {
           acRef.deref()?.abort(this.reason)
         }
+
+        if (getEventListeners(signal, 'abort').length >= defaultMaxListeners) {
+          setMaxListeners(100, signal)
+        }
+
         signal.addEventListener('abort', abort, { once: true })
         requestFinalizer.register(this, { signal, abort })
       }
--- a/test/fetch/issue-node-46525.js
+++ b/test/fetch/issue-node-46525.js
@@ -1,0 +1,28 @@
+'use strict'
+
+const { once } = require('events')
+const { createServer } = require('http')
+const { test } = require('tap')
+const { fetch } = require('../..')
+
+// https://github.com/nodejs/node/issues/46525
+test('No warning when reusing AbortController', async (t) => {
+  function onWarning (error) {
+    t.error(error, 'Got warning')
+  }
+
+  const server = createServer((req, res) => res.end()).listen(0)
+
+  await once(server, 'listening')
+
+  process.on('warning', onWarning)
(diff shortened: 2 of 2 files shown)


ronag (approved):

mcollina (approved):
Lgtm

ricardocosta:
@KhafraDev According to https://github.com/remix-run/react-router/issues/10118, this PR introduces errors when using React Router because `getEventListeners` doesn't support signals of type `AbortSignal `, whilst `signal` can be of that type.

Is there any issue/plan to fix this?

KhafraDev:
getEventListeners does work on a nodejs AbortSignal, since it extends EventTarget. If you have a minimal repro that would be helpful.

joshkel:
@KhafraDev, thank you for looking into this. However, I'm still having problems; depending on how quickly my loop runs, Node.js might add >200 event listeners, so I still get these warnings spammed to the console. I tried raising the warning limit  myself (`setMaxListeners(Infinity, this.controller.signal)`); however, this PR overrides any warning limit that was already set.

For a repro for https://github.com/remix-run/react-router/issues/10118, see https://github.com/joshkel/vitest-undici-abort-bug. The problem arises because testing frameworks such as Vitest may cause jsdom's AbortController to be used with undici's fetch.

KhafraDev:
So it does seem like an issue in vitest - overriding some globals is bound to cause issues (see jest...). Although we should offer support for 3rd party AbortControllers, so this is a bug. We also shouldn't be changing the max listeners if the amount was changed, but there's no way of checking so there isn't much to do with that.

KhafraDev:
this was already fixed in undici v5.21.0, node v19.6.0 is using undici v5.19.1.

KhafraDev:
node v19.8+ have the fix

sangaman:
> node v19.8+ has the fix

I'm still seeing this `MaxListenersExceededWarning` with `AbortSignal` on v19.8.1, it appears to be triggered by calling the native `fetch` a handful of times in a test suite.

Himself65:
is there any hotpatch for nodejs 18?

ba55ie:
> is there any hotpatch for nodejs 18?

You can patch `globalThis.fetch` with the latest version from `undici`:

```js
import { fetch } from 'undici';

globalThis.fetch = fetch;
```  

That's how I fixed our tests in node `v18.15.0`:

```js
import { fetch } from 'undici';
import { beforeAll, afterAll, test } from 'your-test-lib';

let originalFetch;

beforeAll(() => {
  originalFetch = globalThis.fetch;
  globalThis.fetch = fetch;
});

afterAll(() => {
  globalThis.fetch = originalFetch;
});

// Run your tests using fetch.
```

IanVS:
Looks like this is now included in node 18.16.0 as well. 👍

thw0rted:
I'm still seeing this with Jest tests on Node 18.16.1:

> (node:10904) MaxListenersExceededWarning: Possible EventTarget memory leak detected. 11 abort listeners added to [AbortSignal]. Use events.setMaxListeners() to increase limit

Is this PR supposed to address the warning automatically, or am I meant to be manually increasing `setMaxListeners`?

Maxlitv:
> 

I love you.

Maxlitv:
> Looks like this is now included in node 18.16.0 as well. 👍

I love you.

PR #1915, opened 2023-02-08
feat: add Headers.prototype.getSetCookie

KhafraDev:
Implements https://github.com/whatwg/fetch/commit/e4d3480e72da5c19c7938a3261a23e5b4b346b49

It got merged finally!

### DIFF
--- a/lib/fetch/headers.js
+++ b/lib/fetch/headers.js
@@ -11,6 +11,7 @@ const {
   isValidHeaderValue
 } = require('./util')
 const { webidl } = require('./webidl')
+const assert = require('assert')
 
 const kHeadersMap = Symbol('headers map')
 const kHeadersSortedMap = Symbol('headers map sorted')
@@ -115,7 +116,7 @@ class HeadersList {
 
     if (lowercaseName === 'set-cookie') {
       this.cookies ??= []
-      this.cookies.push([name, value])
+      this.cookies.push(value)
     }
   }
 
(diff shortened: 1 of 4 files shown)


lib/fetch/headers.js
```diff
@@ -383,18 +384,68 @@ class Headers {
...
+
+    // 1. If this’s header list does not contain `Set-Cookie`, then return « ».
+    // 2. Return the values of all headers in this’s header list whose name is
+    //    a byte-case-insensitive match for `Set-Cookie`, in order.
+
+    return this[kHeadersList].cookies ?? []
+  }
+
+  // https://fetch.spec.whatwg.org/#concept-header-list-sort-and-combine
   get [kHeadersSortedMap] () {
```
KhafraDev, line 399: A Map can't be used anymore because there can now be duplicate set-cookie keys. I could probably work around it but it wouldn't be maintainable. So I chose the spec solution instead, which is probably slower, but much easier to fix issues with in the future.

KhafraDev:
Test failures seem to be due to the server used in the test being down. https://github.com/nodejs/undici/blob/c0ba75f784765e81722d0c2b31c08d273e04b0f6/test/fetch/issue-1137.js#L11

ronag (approved):
LGTM, does this break anything?

KhafraDev:
All tests pass locally but we should wait for CI still. I'm opening another pr to remove the test.

PR #1918, opened 2023-02-09
fix(types): add missing keepAlive params

SkeLLLa:
Add missing `keepAlive*` params that are present were added in #1904 [docs](https://undici.nodejs.org/#/docs/api/Client?id=parameter-connectoptions) and [code](https://github.com/nodejs/undici/blob/main/lib/core/connect.js#L124), but missing in typings.

### DIFF
--- a/test/types/connector.test-d.ts
+++ b/test/types/connector.test-d.ts
@@ -25,6 +25,8 @@ expectAssignable<Client>(new Client('', {
 expectAssignable<buildConnector.BuildOptions>({
   checkServerIdentity: () => undefined, // Test if ConnectionOptions is assignable
   localPort: 1234, // Test if TcpNetConnectOpts is assignable
+  keepAlive: true,
+  keepAliveInitialDelay: 12345,
 });
 
 expectAssignable<buildConnector.Options>({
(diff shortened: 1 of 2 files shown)


KhafraDev (approved):

xconverge:
I was pretty sure these are already included? https://nodejs.org/api/net.html#socketconnectoptions-connectlistener

The main reason for putting them in the documentation explicitly was mostly just to say what the defaults we set were

connector.d.ts
```
declare namespace buildConnector {
  export type BuildOptions = (ConnectionOptions | TcpNetConnectOpts | IpcNetConnectOpts) & {
    maxCachedSessions?: number | null;
    socketPath?: string | null;
    timeout?: number | null;
    port?: number;
  }
```

net.d.ts
```
    interface TcpNetConnectOpts extends TcpSocketConnectOpts, SocketConstructorOpts {
        timeout?: number | undefined;
    }

    interface TcpSocketConnectOpts extends ConnectOpts {
        port: number;
        host?: string | undefined;
        localAddress?: string | undefined;
        localPort?: number | undefined;
        hints?: number | undefined;
        family?: number | undefined;
        lookup?: LookupFunction | undefined;
        noDelay?: boolean | undefined;
        keepAlive?: boolean | undefined;
        keepAliveInitialDelay?: number | undefined;
    }
```

SkeLLLa:
It's not included in all `@types/node` versions that undici supports.
image

mcollina (approved):
lgtm

xconverge:
Ah I see, sorry about that and thanks!

PR #1921, opened 2023-02-10
fix: timeout immediately

elrrrrrrr:
> setTimeout to singleton mode for better performance https://github.com/nodejs/undici/pull/1908
> But it introduces an additional problem when the server timeout is very short,
> it will invalidate the timeout configuration, which will have an unusual effect.

* `lib/timers.js` add a new timeoutType parameter to avoid parser overwriting.
* `lib/timers.js` add first tick check logic to handle quick requests, as in the test case scenario.

### DIFF
--- a/test/request.js
+++ b/test/request.js
@@ -256,15 +256,15 @@ test('should throw timeout immediately', scope => {
     const server = createServer((req, res) => {
       setTimeout(() => {
         res.end('banana')
-      }, 50)
+      }, 800)
     })
     t.teardown(server.close.bind(server))
 
     server.listen(0, () => {
       t.rejects(request(`http://localhost:${server.address().port}`, {
         method: 'GET',
-        headersTimeout: 10,
-        bodyTimeout: 10
+        headersTimeout: 500,
+        bodyTimeout: 400
       }), 'HeadersTimeoutError')
     })
   })
--- a/test/request.js
+++ b/test/request.js
@@ -246,3 +246,26 @@ test('DispatchOptions#reset', scope => {
     })
   })
 })
+
+test('should throw timeout immediately', scope => {
+  scope.plan(1)
+
+  scope.test('throw headersTimeoutError', t => {
+    t.plan(1)
+
+    const server = createServer((req, res) => {
+      setTimeout(() => {
+        res.end('banana')
+      }, 50)
+    })
+    t.teardown(server.close.bind(server))
+
+    server.listen(0, () => {
+      t.rejects(request(`http://localhost:${server.address().port}`, {
+        method: 'GET',
+        headersTimeout: 10,
+        bodyTimeout: 10
+      }), 'HeadersTimeoutError')
+    })
+  })
+})
--- a/lib/client.js
+++ b/lib/client.js
@@ -447,7 +447,7 @@ class Parser {
     if (value !== this.timeoutValue) {
       timers.clearTimeout(this.timeout)
       if (value) {
-        this.timeout = timers.setTimeout(onParserTimeout, value, this)
+        this.timeout = timers.setTimeout(onParserTimeout, value, this, type)
         // istanbul ignore else: only for jest
         if (this.timeout.unref) {
           this.timeout.unref()
@@ -896,8 +896,9 @@ class Parser {
   }
 }
 
-function onParserTimeout (parser) {
-  const { socket, timeoutType, client } = parser
+function onParserTimeout (parser, timeoutType) {
+  const { socket, client } = parser
+  parser.timeoutType = timeoutType
 
   /* istanbul ignore else */
   if (timeoutType === TIMEOUT_HEADERS) {
--- a/lib/client.js
+++ b/lib/client.js
@@ -447,7 +447,7 @@ class Parser {
     if (value !== this.timeoutValue) {
       timers.clearTimeout(this.timeout)
       if (value) {
-        this.timeout = timers.setTimeout(onParserTimeout, value, this, type)
+        this.timeout = timers.setTimeout(onParserTimeout, value, this)
         // istanbul ignore else: only for jest
         if (this.timeout.unref) {
           this.timeout.unref()
@@ -896,9 +896,8 @@ class Parser {
   }
 }
 
-function onParserTimeout (parser, timeoutType) {
-  const { socket, client } = parser
-  parser.timeoutType = timeoutType
+function onParserTimeout (parser) {
+  const { socket, timeoutType, client } = parser
 
   /* istanbul ignore else */
   if (timeoutType === TIMEOUT_HEADERS) {
--- a/lib/timers.js
+++ b/lib/timers.js
@@ -13,18 +13,22 @@ function onTimeout () {
   while (idx < len) {
     const timer = fastTimers[idx]
 
-    // no need to detect
-    if (fastNow < timer.expires) {
-      break
-    }
-
     if (timer.expires && fastNow >= timer.expires) {
-      removeTimeout(timer)
-      timer.callback(timer.opaque, timer.type)
-      len--
+      timer.expires = 0
+      timer.callback(timer.opaque)
     }
 
-    idx++
+    if (timer.expires === 0) {
+      timer.active = false
+      if (idx !== len - 1) {
+        fastTimers[idx] = fastTimers.pop()
+      } else {
+        fastTimers.pop()
+      }
+      len -= 1
+    } else {
+      idx += 1
+    }
   }
 
   if (fastTimers.length > 0) {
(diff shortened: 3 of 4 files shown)


ronag (changes requested):
I don't feel this fixes a realistic problem. When would you ever need/have such a short timeout?

test/request.js
```diff
@@ -246,3 +246,26 @@ test('DispatchOptions#reset', scope => {
...
+
+test('should throw timeout immediately', scope => {
+  scope.plan(1)
+
+  scope.test('throw headersTimeoutError', t => {
+    t.plan(1)
+
+    const server = createServer((req, res) => {
+      setTimeout(() => {
+        res.end('banana')
```
KhafraDev, line 258: ending the request will likely make this test flaky

KhafraDev (changes requested):

elrrrrrrr:
> I don't feel this fixes a realistic problem. When would you ever need/have such a short timeout?

The 50ms return and 10ms timeout in the use case is a boundary case.

In the current implementation, if the return time of a request is less than the time of each detection (< 1000ms), then the timeout will be invalid.

Such as the `800ms` return and `500ms` timeout that I have updated the configuration for in my use case.

This is a more common scenario, such as requesting `www.githubstatus.com` to ensure that the network heartbeat checks and ensures that the network returns at a hight rate.

Also, this is a breakChange and will add a new latent rule to the existing parameter configuration, hopefully keeping the original logic as much as possible, thanks.

elrrrrrrr:
I have refactored this PR, but this seems to extend the scope of the changes.

* Turning fastTimers into an ordered array, sorted by expires.
* Adjust the detection time for the next execution with `Math.min(delay, 1e3)`.
* Modify the `clear` operation to be an array of operations.
* Timeouts within 1 second will still be triggered in bulk

Looking forward to your responses, thanks.

test/request.js
elrrrrrrr: My intention was to expect the results to be returned after `800ms`
This is to test whether the headerTimeout is triggered successfully.

KhafraDev: the test will be flaky if the request is being ended, we've had similar issues in the past.

ronag (commented):
I suspect this will actually slow things down rather than speed it up... do you have any benchmarks?

ronag:
To clarify the old implementation is optimized for insert, remove and update O(1) while being a bit slow during the once a second expiry check O(n). This is the common usage in this case and the once-a-second O(n) is an amortized O(1).

This PR changes this so that insert, remove and update is O(n) or worse for the fast path. Which  is not optimal.

elrrrrrrr:
> I suspect this will actually slow things down rather than speed it up... do you have any benchmarks?

I tried testing this with the bench script that comes with the project.
but the results were very erratic, probably relying on the requests of the runtime.


Compare the results of the `main` branch run:
### main
```
[bench:run] │ Tests               │ Samples │          Result │ Tolerance │ Difference with slowest │
[bench:run] |─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
[bench:run] │ undici - fetch      │       1 │ 1100.87 req/sec │  ± 0.00 % │                       - │
[bench:run] |─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
[bench:run] │ http - no keepalive │       1 │ 1696.40 req/sec │  ± 0.00 % │               + 54.10 % │
[bench:run] |─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
[bench:run] │ http - keepalive    │       1 │ 3143.53 req/sec │  ± 0.00 % │              + 185.55 % │
[bench:run] |─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
[bench:run] │ undici - request    │       1 │ 3315.95 req/sec │  ± 0.00 % │              + 201.21 % │
[bench:run] |─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
[bench:run] │ undici - pipeline   │       1 │ 3996.36 req/sec │  ± 0.00 % │              + 263.02 % │
[bench:run] |─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
[bench:run] │ undici - dispatch   │       1 │ 4428.62 req/sec │  ± 0.00 % │              + 302.28 % │
[bench:run] |─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
[bench:run] │ undici - stream     │       1 │ 5274.53 req/sec │  ± 0.00 % │              + 379.12 % │

```

```
[bench:run] │ Tests               │ Samples │          Result │ Tolerance │ Difference with slowest │
[bench:run] |─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
[bench:run] │ http - no keepalive │       1 │ 1293.92 req/sec │  ± 0.00 % │                       - │
[bench:run] |─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
[bench:run] │ undici - fetch      │       1 │ 1428.64 req/sec │  ± 0.00 % │               + 10.41 % │
[bench:run] |─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
[bench:run] │ undici - request    │       1 │ 3051.56 req/sec │  ± 0.00 % │              + 135.84 % │
[bench:run] |─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
[bench:run] │ undici - pipeline   │       1 │ 3125.12 req/sec │  ± 0.00 % │              + 141.52 % │
[bench:run] |─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
[bench:run] │ http - keepalive    │       1 │ 3220.67 req/sec │  ± 0.00 % │              + 148.91 % │
[bench:run] |─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
[bench:run] │ undici - stream     │       1 │ 4818.47 req/sec │  ± 0.00 % │              + 272.39 % │
[bench:run] |─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
[bench:run] │ undici - dispatch   │       1 │ 5928.51 req/sec │  ± 0.00 % │              + 358.18 % │
```

### timeout immediately

```
[bench:run] │ Tests               │ Samples │          Result │ Tolerance │ Difference with slowest │
[bench:run] |─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
[bench:run] │ undici - fetch      │       1 │ 1566.85 req/sec │  ± 0.00 % │                       - │
[bench:run] |─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
[bench:run] │ http - no keepalive │       1 │ 1583.40 req/sec │  ± 0.00 % │                + 1.06 % │
[bench:run] |─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
[bench:run] │ undici - pipeline   │       1 │ 3958.27 req/sec │  ± 0.00 % │              + 152.63 % │
[bench:run] |─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
[bench:run] │ undici - request    │       1 │ 4226.98 req/sec │  ± 0.00 % │              + 169.78 % │
[bench:run] |─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
[bench:run] │ http - keepalive    │       1 │ 4403.38 req/sec │  ± 0.00 % │              + 181.03 % │
[bench:run] |─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
[bench:run] │ undici - stream     │       1 │ 4438.96 req/sec │  ± 0.00 % │              + 183.30 % │
[bench:run] |─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
[bench:run] │ undici - dispatch   │       1 │ 4638.47 req/sec │  ± 0.00 % │              + 196.04 % │
```

```
[bench:run] │ Tests               │ Samples │          Result │ Tolerance │ Difference with slowest │
[bench:run] |─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
[bench:run] │ undici - fetch      │       1 │ 1395.61 req/sec │  ± 0.00 % │                       - │
[bench:run] |─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
[bench:run] │ http - no keepalive │       1 │ 1561.49 req/sec │  ± 0.00 % │               + 11.89 % │
[bench:run] |─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
[bench:run] │ http - keepalive    │       1 │ 3063.47 req/sec │  ± 0.00 % │              + 119.51 % │
[bench:run] |─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
[bench:run] │ undici - stream     │       1 │ 3900.97 req/sec │  ± 0.00 % │              + 179.52 % │
[bench:run] |─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
[bench:run] │ undici - pipeline   │       1 │ 3948.47 req/sec │  ± 0.00 % │              + 182.92 % │
[bench:run] |─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
[bench:run] │ undici - dispatch   │       1 │ 4246.95 req/sec │  ± 0.00 % │              + 204.31 % │
[bench:run] |─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
[bench:run] │ undici - request    │       1 │ 4897.70 req/sec │  ± 0.00 % │              + 250.94 % │
```

elrrrrrrr:
> To clarify the old implementation is optimized for insert, remove and update O(1) while being a bit slow during the once a second expiry check O(n). This is the common usage in this case and the once-a-second O(n) is an amortized O(1).
> 
> This PR changes this so that insert, remove and update is O(n) or worse. Which is not optimal.


Thanks for the enlightening answer, as the timeout array is constantly being updated, there is a lot of computational overhead involved.

In fact, performance was not a concern before and I was expecting better real time performance.

I have reworked the PR so that for real time timeout scenarios, it can be left out of the queue of timers to get better real time performance, which is a compromise.

For short requests, I'd like to get a more real-time timeout.
But it seems to cause some splitting, I can't think of any other way at the moment.

elrrrrrrr:
I would like to close this PR as it has not been accepted by others despite my efforts to make changes. I want to thank you for your time and feedback during this process.

PR #1937, opened 2023-02-17
feat: change headersTimeout and bodyTimeout to 300s

kyrylkov:


ronag (approved):

docs/api/Client.md
```diff
@@ -18,7 +18,7 @@ Returns: `Client`
 ### Parameter: `ClientOptions`
 
 * **bodyTimeout** `number | null` (optional) - Default: `30e3` - The timeout after which a request will time out, in milliseconds. Monitors time between receiving body data. Use `0` to disable it entirely. Defaults to 30 seconds.
```
mcollina, line 20: Can you also increase the `bodyTimeout` here? Having _headers_ wait for 300s and _body_ for 30s seems incorrect.

mcollina (changes requested):

docs/api/Client.md
ronag: I was the one that asked not to change `bodyTimeout` since I didn't find it necessary. I don't have a strong opinion on it if you prefer it otherwise.

ronag: @kyrylkov sorry for the back and forth

mcollina: Looking at the defaults, I do not think why I would want to wait for 5 minutes for headers and then 30 seconds for the full data transfer. If I waited 4 minutes for the headers, I might as well wait another 4 for the body.

kyrylkov: In our case we wait for headers well above several minutes, but body comes very soon after headers.  So bumping up only headers timeout fixed our case.  But I changed body timeout again now.

ronag:
@KhafraDev flaky fetch failure on node 19?

mcollina (approved):
lgtm

KhafraDev (approved):
node-fetch tests are occasionally flaky

PR #1939, opened 2023-02-17
fix: fix exported formdata type def to have a constructor

JikkyHD:
Relates to: https://github.com/nodejs/undici/discussions/1938

Fixes types/formdata.d.ts to export a constructor for constructing FormData to use in `request.body`

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "undici",
-  "version": "5.19.1",
+  "version": "5.19.2",
   "description": "An HTTP/1.1 client, written from scratch for Node.js",
   "homepage": "https://undici.nodejs.org",
   "bugs": {
--- a/types/formdata.d.ts
+++ b/types/formdata.d.ts
@@ -2,7 +2,7 @@
 /// <reference types="node" />
 
 import { File } from './file'
-import { SpecIterator, SpecIterableIterator } from './fetch'
+import { SpecIterableIterator } from './fetch'
 
 /**
  * A `string` or `File` that represents a single value from a set of `FormData` key-value pairs.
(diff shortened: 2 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "undici",
-  "version": "5.19.1",
+  "version": "5.19.2",
```
ronag, line 3:
```suggestion
  "version": "5.19.1",
```

types/formdata.d.ts
```diff
@@ -21,7 +22,7 @@ export declare class FormData {
    *
    * @param name The name of the field whose data is contained in `value`.
    * @param value The field's value. This can be [`Blob`](https://developer.mozilla.org/en-US/docs/Web/API/Blob)
-    or [`File`](https://developer.mozilla.org/en-US/docs/Web/API/File). If none of these are specified the value is converted to a string.
+   or [`File`](https://developer.mozilla.org/en-US/docs/Web/API/File). If none of these are specified the value is converted to a string.
```
ronag, line 25:
```suggestion
    or [`File`](https://developer.mozilla.org/en-US/docs/Web/API/File). If none of these are specified the value is converted to a string.
```

types/formdata.d.ts
```diff
@@ -32,7 +33,7 @@ export declare class FormData {
...
```
ronag, line 36:
```suggestion
    or [`File`](https://developer.mozilla.org/en-US/docs/Web/API/File). If none of these are specified the value is converted to a string.
```

KhafraDev (changes requested):
If the error is

> TypeError: undici_1.FormData is not a constructor

as mentioned in the discussion, changing the types won't have any effect. Are you by chance using a node version < v16.8.0?

JikkyHD:
Hey @KhafraDev - yes, I've been using node v14.15.
I changed to node v18.13, which resolves the issue.

PR #1944, opened 2023-02-20
test(wpt): add results to an existing WPT Report

panva:
I wish to extend the [daily Node.js WPT Report](https://github.com/nodejs/node/blob/main/.github/workflows/daily-wpt-fyi.yml) with your WPTs.

I plan to extend the workflow file with roughly the following

```
UNDICI_VERSION = cat deps/undici/src/package.json | jq -r '.version'
if (UNDICI_VERSION_INCLUDES_THIS_CHANGE(UNDICI_VERSION))
  actions/checkout nodejs/undici at $UNDICI_VERSION tag
  WPT_REPORT=$(pwd)/out/wpt/wptreport.json npm run test:wpt
```

In order for the WPTs to be comparable with other wpt.fyi results they must respect the variants (1st commit), must sanitize the subtest names and messages the same (part of the 2nd commit).

NB: per commit no whitespace review recommended

### DIFF
--- a/test/wpt/runner/runner/runner.mjs
+++ b/test/wpt/runner/runner/runner.mjs
@@ -1,6 +1,6 @@
 import { EventEmitter, once } from 'node:events'
-import { readdirSync, readFileSync, statSync } from 'node:fs'
 import { isAbsolute, join, resolve } from 'node:path'
+import { existsSync, readdirSync, readFileSync, statSync, writeFileSync } from 'node:fs'
 import { fileURLToPath } from 'node:url'
 import { Worker } from 'node:worker_threads'
 import { colors, handlePipes, normalizeName, parseMeta, resolveStatusPath } from './util.mjs'
@@ -8,6 +8,24 @@ import { colors, handlePipes, normalizeName, parseMeta, resolveStatusPath } from
 const basePath = fileURLToPath(join(import.meta.url, '../../..'))
 const testPath = join(basePath, 'tests')
 const statusPath = join(basePath, 'status')
+
+// https://github.com/web-platform-tests/wpt/blob/b24eedd/resources/testharness.js#L3705
+function sanitizeUnpairedSurrogates (str) {
+  return str.replace(
+    /([\ud800-\udbff]+)(?![\udc00-\udfff])|(^|[^\ud800-\udbff])([\udc00-\udfff]+)/g,
+    function (_, low, prefix, high) {
+      let output = prefix || '' // Prefix may be undefined
+      const string = low || high // Only one of these alternates can match
+      for (let i = 0; i < string.length; i++) {
+        output += codeUnitStr(string[i])
+      }
+      return output
+    })
+}
+
+function codeUnitStr (char) {
+  return 'U+' + char.charCodeAt(0).toString(16)
+}
 
 export class WPTRunner extends EventEmitter {
   /** @type {string} */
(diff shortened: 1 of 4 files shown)


panva:
cc @KhafraDev please give it a rough first look

KhafraDev:
the changes look fine so far, although the runner is erroring out

KhafraDev:
jest, unfortunately (https://github.com/nodejs/undici/issues/1945)

KhafraDev:
I'd love to, but we need to make sure undici works with jest 😵‍💫

panva:
@KhafraDev Do you have a handy command in the undici repo to update the WPT test files?

panva:
Ok, so with updated WPT files the `test:wpt` doesn't even run. That is an issue seeing how the daily job runs with the latest wpt files...

panva:
I've pushed the one commit that can go alone as #1951.

As for the WPT Report, this is also good to go but I cannot use it unless the Runner is ready for the fixtures to be replaced by a full `web-platform-tests/wpt` checkout.

Issues i've encountered while doing so:

- new `{{ }}` GET replace block to be handled by `handlePipes`
- rather than logging wpt setup failures the runner stops the run with no logs

Feel free to close this as needed, I don't unfortunately have the time needed to fix the runner issues and for fetch related WPTs to be included in the daily wpt.fyi upload they need to represent a given wpt checkout, not the cherry picked one the runner expects at the moment.

KhafraDev:
I've dmed you on twitter, I'd like to take care of those issues (including a way to automatically update the WPTs). What are you doing to update/replce them?

anonrig:
Is there a specific reason for why we are not running fetch wpt in node.js core?

KhafraDev:
@anonrig I outlined some difficulties with doing so here: https://github.com/nodejs/node/pull/46498#issuecomment-1418216449

KhafraDev:
I can remove the blob tests, but undici implements its own File class (predecessor to node's), so we should keep those in. Maybe the easiest fix is to disable reports being generated for File/Blob tests?

panva:
Yeah, I'll pass the WPT_REPORT optionally straight from the start file and skip doing so for FileAPI.

KhafraDev:
We do also implement FileReader (at the time it made sense), but I guess reporting coverage from it isn't that important.

panva:
> We do also implement FileReader (at the time it made sense), but I guess reporting coverage from it isn't that important.

Not unless it's made a global in Node. To be honest a lot of the FileWPTs in Node are failing just because we don't FileReader to support them...

panva:
Also skipped WebSocket since it's not a Node.js global (yet?).

KhafraDev (approved):

panva:
I will do a test submit to staging.wpt.fyi before you land this.

panva:
I found some untitled [subtests](https://staging.wpt.fyi/results/fetch/api/redirect/redirect-schemes.any.html?run_id=5247378118410240&run_id=6086989598162944), looking into it.

KhafraDev:
Just wondering, since our tests (& node's) run in a worker, should the *.worker.html tests be marked as passing too?

panva:
> Just wondering, since our tests (& node's) run in a worker, should the *.worker.html tests be marked as passing too?

It's not the same and just duplicating the results based on the path seems wrong. You can use the rich wpt query language to filter only tests that are present in one of the products to compare.

panva:
> I found some untitled [subtests](https://staging.wpt.fyi/results/fetch/api/redirect/redirect-schemes.any.html?run_id=5247378118410240&run_id=6086989598162944), looking into it.

Worked around in the runner.

KhafraDev (approved):

ronag (approved):
RSLGTM, @KhafraDev I will have to trust you on this one.

PR #1947, opened 2023-02-20
fix(fetch): improve Headers and Request type-compatibility

### LINKED ISSUE #1943: Fetch types are incompatible with TypeScript
## Bug Description

The `Request` interface exposed by Undici does not satisfy the same type from the global type definitions (`lib/dom`), and also missing properties according to the Fetch API specification:

https://github.com/nodejs/undici/blob/d8d9a96b6b8f856d9e01a1edfe22c3d83c6996ec/types/fetch.d.ts#L138

## Reproducible By

As an example, consider the following code: 

```ts
import { Request } from 'undicit'

function transform(req: globalThis.Request) {}

transform(new Request('/hello'))
// ^^^ Property 'referrer' is missing in type 'import(".../node_modules/undici/types/fetch").Request' but required in type 'Request'.ts(2345)
```

The `referrer` property is not present in the Undici's `Request` type definition while it's both required in `lib/dom` definitions and the [Fetch API specification](https://fetch.spec.whatwg.org/#request-class).

```ts
readonly attribute [USVString](https://webidl.spec.whatwg.org/#idl-USVString) [referrer](https://fetch.spec.whatwg.org/#dom-request-referrer);
```

## Expected Behavior

Fetch API primitives such as Request, Response and Headers, are compatible with the global 

## Logs & Screenshots

No applicable logs.

## Environment

- Node: v16.16.0
- Undici: ^5.20.0
- TypeScript: ^4.9.4

### Additional context

I understand that I'm setting `lib/dom` as the source of truth here but I expect that shouldn't matter since:

1. Both Undici and `lib/node` aim to implement the Fetch API specification.
2. Undici is objectively missing some properties in its type definitions defined in the spec, like `referrer`. 

I'd like to open a pull request to ensure this compliance after discussing this with the maintainers of this lib.

kettanaito:
- Fixes #1943

I'm using a simple reproduction example to test type congruence of Undici against `lib/dom`:

```ts
import { Request as UndiciRequest } from 'undici'

function transform(r: Request): void {}

transform(new UndiciRequest('/hello'))
// ^^^ This won't compile in TypeScript!
```

### DIFF
--- a/types/fetch.d.ts
+++ b/types/fetch.d.ts
@@ -68,7 +68,7 @@ export declare class Headers implements SpecIterable<[string, string]> {
   readonly keys: () => SpecIterableIterator<string>
   readonly values: () => SpecIterableIterator<string>
   readonly entries: () => SpecIterableIterator<[string, string]>
-  readonly [Symbol.iterator]: () => SpecIterator<[string, string]>
+  readonly [Symbol.iterator]: () => SpecIterableIterator<[string, string]>
 }
 
 export type RequestCache =
@@ -146,7 +146,8 @@ export declare class Request implements BodyMixin {
   readonly method: string
   readonly mode: RequestMode
   readonly redirect: RequestRedirect
-  readonly referrerPolicy: string
+  readonly referrer: string
+  readonly referrerPolicy: ReferrerPolicy
   readonly url: string
 
   readonly keepalive: boolean


types/fetch.d.ts
```diff
@@ -68,7 +68,7 @@ export declare class Headers implements SpecIterable<[string, string]> {
   readonly keys: () => SpecIterableIterator<string>
   readonly values: () => SpecIterableIterator<string>
   readonly entries: () => SpecIterableIterator<[string, string]>
-  readonly [Symbol.iterator]: () => SpecIterator<[string, string]>
+  readonly [Symbol.iterator]: () => SpecIterableIterator<[string, string]>
```
kettanaito, line 71: It appears that the iterator symbol on `Headers` must be iterable.

types/fetch.d.ts
```diff
@@ -146,7 +146,8 @@ export declare class Request implements BodyMixin {
   readonly method: string
   readonly mode: RequestMode
   readonly redirect: RequestRedirect
-  readonly referrerPolicy: string
+  readonly referrer: string
+  readonly referrerPolicy: ReferrerPolicy
```
kettanaito, line 150: Also narrowing down the referrer policy as we already have the `ReferrerPolicy` type used for `RequestInit`. It cannot really be anything else, to my best knowledge.

kettanaito:
With these changes, I can eliminate most of the type-related errors in my usage project. However, there seems to be an incongruence between the `Request.body` types. The `ReadableStream<R>` type from `stream/web` is incompatible with the `ReadableStream<T>` type from `lib/dom`. Here, I'm hesitant to call `stream/web` the issue because it looks like `lib/dom` contains outdated types and has issues raised regarding this (https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/58252). Sadly, those were never addressed. 

There are a few waves of errors that happen:

## 1. `ReadableStream<any>` (Undici) is incompatible with `ReadableStream<Uint8Array>` (lib/dom)

This can be rather simply addressed by:

```diff
-body: ReadableStream | null
+body: ReadableStream<Uint8Array> | null
```

However, another issues arises from this. 

## 2. `ReadableStream<Uint8Array>` (Undici*) is not compatible with `ReadableStream<Uint8Array>`

```
Argument of type 'import("/undici/types/fetch").Request' is not assignable to parameter of type 'Request'.
  Types of property 'body' are incompatible.
    Type 'import("stream/web").ReadableStream<Uint8Array> | null' is not assignable to type 'ReadableStream<Uint8Array> | null'.
      Type 'import("stream/web").ReadableStream<Uint8Array>' is not assignable to type 'ReadableStream<Uint8Array>'.
        The types returned by 'getReader(...).read(...)' are incompatible between these types.
          Type 'Promise<ReadableStreamDefaultReadResult<Uint8Array>>' is not assignable to type 'Promise<ReadableStreamReadResult<T>>'.
            Type 'ReadableStreamDefaultReadResult<Uint8Array>' is not assignable to type 'ReadableStreamReadResult<T>'.
              Type 'ReadableStreamDefaultReadValueResult<Uint8Array>' is not assignable to type 'ReadableStreamReadResult<T>'.
                Type 'ReadableStreamDefaultReadValueResult<Uint8Array>' is not assignable to type 'ReadableStreamReadValueResult<T>'.
                  Types of property 'value' are incompatible.
                    Type 'Uint8Array' is not assignable to type 'T'.
                      'Uint8Array' is assignable to the constraint of type 'T', but 'T' could be instantiated with a different subtype of constraint 'ArrayBufferView'.
```

This has to do with proper generics usage, my head is too cloudy right now to remember how to resolve it. Will circle back to it later.

KhafraDev (approved):
changes look fine, the body fix can wait for another PR (although ReadableStream<Uint8Array> is correct)

kettanaito:
By the way, I don't mind lending a hand with type testing in this repo. I've worked some time with it and it has proven rather useful. You can see an example of it in [here](https://github.com/mswjs/msw/tree/2a1ca4379f20a0fc4ce5519621bd9ae3b298f628/test/typings). Basically, with a bit of a configuration, we can automate the type-checking of Undici examples against different versions of TypeScript even, and use those as tests to catch type-related regressions. I think in the light that Undici is not authored in TypeScript, type tests stand twice as useful.

mcollina:
> By the way, I don't mind lending a hand with type testing in this repo. I've worked some time with it and it has proven rather useful. You can see an example of it in [here](https://github.com/mswjs/msw/tree/2a1ca4379f20a0fc4ce5519621bd9ae3b298f628/test/typings). Basically, with a bit of a configuration, we can automate the type-checking of Undici examples against different versions of TypeScript even, and use those as tests to catch type-related regressions. I think in the light that Undici is not authored in TypeScript, type tests stand twice as useful.

Any help is appreciated! Go for it.

---

Can this PR land? It's still marked as draft.

kettanaito:
Hey, @mcollina. It's ready at its current state. Is it okay that some tests are failing on CI? They also fail for me locally (on `main`) so I assume that's not related to the changes.

mcollina:
Could you add a test for the types? We use tsd.

kettanaito:
I'm a bit overwhelmed with work, I will add the test this/next week.

kettanaito:
I love GitHub. Let's put two buttons with totally opposite actions 3px away from each other. Sorry, will re-open.

PR #1960, opened 2023-02-22
fix: remove unneeded fetch tests

KhafraDev:
the WPTs have adequate coverage for referrer & referrerPolicy, and these tests were incorrect.


KhafraDev:
I mixed these tests up with some other tests that were failing from the same file, my bad

mcollina (approved):
lgtm

PR #1964, opened 2023-02-22
fix(fetch): improve Headers and Request type-compatibility

kettanaito:
- Reopened #1947 because GitHub UI is fantastic.


kettanaito:
As mentioned in the previous review, a type test has to be added here. I will add it this/next week due to the work load.

KhafraDev:
type tests are failing (`npm run test:typescript`)

```
   test/types/fetch.test-d.ts:141:0
  ✖  141:0  Parameter type string is declared too wide for argument type ReferrerPolicy.  
```

MikeZeDev:
Suddenly after an update in my package i got the problem with the referrer property missing and i just found this library and this PR is related. 

I manually edited fetch.d.ts to test this PR. I can get rid of the test error you mentioned if i revert 

```  readonly referrerPolicy: ReferrerPolicy ```

to

```  readonly referrerPolicy: string ```

* Would that be acceptable ? I fear this is not strict enough or just a bad solution
* Should i make another PR because the person who submitted it seems to be busy?

**Comment by kettanaito** [comment]
@MikeZeDev, not sure if `string` isn't too broad for the referrer policy. It is annotated as `RefererPolicy` type in lib.dom, I think it'd be best if Undici introduced compatibility with that as well. Value-wise, you cannot set unknown values anyway. Can you tell me more about the use case that resulted in the error for you?

**Comment by kettanaito** [comment]
@KhafraDev, I've fixed the test by adjusting the expected type. It's no longer a wider `string` but a narrower `ReferrerPolicy`, which is compatible with how [lib.dom](https://github.com/microsoft/TypeScript/blob/33b156147b78486ebebd83ce3de90a635d737bbb/src/lib/dom.generated.d.ts#L1696) annotates the `referrerPolicy` request property.

**Comment by kettanaito** [comment]
Type tests are passing now.

```
> undici@6.15.0 test:typescript
> tsd && tsc --skipLibCheck test/imports/undici-import.ts
```

@KhafraDev, may I please get the workflow approved on this pull request to see if there are any other pending things to resolve? Thanks.

mcollina (approved):
lgtm

kettanaito:
@mcollina, note that this change can be technically breaking if consumers relied on `referrerPolicy` being a plain string. A narrower type shipped by this change will error on the type level in some cases. 

My personal take on situations like this is if you rely on a bug, you should expect breaking changes. I'd release this as a patch fix since the type annotations were previously incorrect.

Issue #1949, opened 2023-02-21
TypeScript Type Fixes (for *.d.ts files)


joshxyzhimself:
Hi, thought of submitting a couple type fixes for the *.d.types to improve some VSCode IntelliSense.

For example, `Client.Options.socketPath` type is missing when the following lines use it:

https://github.com/nodejs/undici/blob/d04ec2b1b4145f9068fb97a43ea35f0bb3197d15/lib/client.js#L103

https://github.com/nodejs/undici/blob/d04ec2b1b4145f9068fb97a43ea35f0bb3197d15/benchmarks/benchmark.js#L27

A question before I proceed, am I correct that the types below:

https://github.com/nodejs/undici/blob/d04ec2b1b4145f9068fb97a43ea35f0bb3197d15/types/client.d.ts#L20-L52

..are intended for the constructor below?

https://github.com/nodejs/undici/blob/d04ec2b1b4145f9068fb97a43ea35f0bb3197d15/lib/client.js#L88-L115

I'm thinking of adding types for `options.socketPath`, but also noticed missing types for other parameters like `options.socketTimeout`, is it alright to include those?

Thanks.

mcollina:
Go ahead!

PR #1968, opened 2023-02-23
TypeScript type fixes, for #1949

joshxyzhimself:
TypeScript type fixes, for #1949.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -91,11 +91,10 @@ try {
  * @type {import('../types/client').default}
  */
 class Client extends DispatcherBase {
-
   /**
-   * 
-   * @param {string|URL} url 
-   * @param {import('../types/client').Client.Options} options 
+   *
+   * @param {string|URL} url
+   * @param {import('../types/client').Client.Options} options
    */
   constructor (url, options = {}) {
     let {
@@ -124,7 +123,7 @@ class Client extends DispatcherBase {
       maxResponseSize,
       autoSelectFamily,
       autoSelectFamilyAttemptTimeout
-    } = options;
+    } = options
 
     super()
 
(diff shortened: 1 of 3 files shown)


joshxyzhimself:
at lines 118 to 136, I added `@deprecated` for `keepAlive`, `socketTimeout`, `requestTimeout`, `idleTimeout`, `maxKeepAliveTimeout`.. is this correct?

https://github.com/nodejs/undici/blob/d04ec2b1b4145f9068fb97a43ea35f0bb3197d15/lib/client.js#L118-L136

KhafraDev:
I wouldn't add types for properties that are completely unusable, not sure how everyone else feels about it

joshxyzhimself:
@KhafraDev replaced it with `never`, what do you think?

mcollina:
Could you add a few test for the types changes?

lib/client.js
```diff
@@ -228,7 +238,7 @@ class Client extends DispatcherBase {
     this[kMaxRedirections] = maxRedirections
     this[kMaxRequests] = maxRequestsPerClient
     this[kClosedResolve] = null
-    this[kMaxResponseSize] = maxResponseSize > -1 ? maxResponseSize : -1
+    this[kMaxResponseSize] = (typeof maxResponseSize === 'number' && maxResponseSize > -1) ? maxResponseSize : -1
```
mcollina, line 241: why this change?

joshxyzhimself: Ah, I added `// @ts-check` at the first-line to confirm that the types work.

It showed me the following error on that line 231:

```
'maxResponseSize' is possibly 'undefined'.
```

image

Should I keep it or revert it? `undefined > -1` is `false` anyways.

joshxyzhimself: reverted it

mcollina (approved):
lgtm

PR #1973, opened 2023-02-25
websocket: separate connection logic from websocket

KhafraDev:
this is needed *if* we want to implement websocketstream, which i've kinda been working on for fun


mcollina (approved):
lgtm. I think implementing WebSocketStream is a good goal.

KhafraDev:
The problem is that WebSocketStream doesn't have a spec, this document which hasn't been updated since April 2020 is all there is https://github.com/ricea/websocketstream-explainer#websocketstream-explained

PR #1974, opened 2023-02-27
wptrunner: expose gc

KhafraDev:
some tests rely on this https://github.com/nodejs/undici/blob/49c6b48fe67c5c11d6693e3a8f8618dd269e26d5/test/wpt/tests/common/gc.js#L16


ronag (approved):

mcollina (approved):
lgtm

Issue #1976, opened 2023-03-01
Shorter than expected headers timeout exception is fired


kajiiiro:
## Bug Description

i update v5.15.0 -> v5.18.0,
it increased unexpected headers timeout exception.
my condition is
- set `headersTimeout ` 1000msec
- under high load
- destination server response in about 50msec
but it fired headers timeout.

i see `lib/timers.js` set invalid expired time.
https://github.com/nodejs/undici/blob/main/lib/timers.js#L72

`fastNow` is expected current time,`fastNow` is updated in onTimeout function.
Normally onTimeout is called in 1000msec.
But under high load onTimeout is called in over 1000msec.

<img width="443" alt="image" src="https://user-images.githubusercontent.com/2986250/222027236-6afa2193-dfba-4f4b-9543-f8cb0c213e47.png">

## Reproducible By

I checked `fastNow` and `Date.now()`.

add console.log after this line.
`console.log('diff', Date.now() - fastNow)`
https://github.com/nodejs/undici/blob/main/lib/timers.js#L72

exec the following code.
```
"use strict";
const request = require("undici").request;

const sleep = msec => new Promise(resolve => setTimeout(resolve, msec));

const port = 3000;
const server = require("express")()
server.get('/', (_, res) => { res.send('ok') });
server.listen(port, async () => {
    for (let j = 0; j < 100; ++j) {
        for (let i = 0; i < 50; ++i) {
            request(`http://localhost:${port}`, { headersTimeout: 500 })
                .catch(error => { console.error(error); process.exit(1) })
        }
        await sleep(100)
    }
})

setTimeout(() => { process.exit(0) }, 10 * 1000)
```

output

```
diff 918
...
```

## Expected Behavior

expired = current time(timers.setTimeout called time) + delay.

## Environment

Node v18
undici ^5.18.0

### Additional context

it is fixed the following code.

```diff
diff --git a/lib/timers.js b/lib/timers.js
index 4f4fcea4..955c4f34 100644
--- a/lib/timers.js
+++ b/lib/timers.js
@@ -65,10 +65,10 @@ class Timeout {
       fastTimers.push(this)
       if (!fastNowTimeout || fastTimers.length === 1) {
         refreshTimeout()
-        fastNow = Date.now()
       }
     }
 
+    fastNow = Date.now()
     this.expires = fastNow + this.delay
   }

```

ronag:
Can you try with https://github.com/nodejs/undici/pull/1977?

[...]

PR #1977, opened 2023-03-01
fix: fast timers and event loop lag

ronag:
When there is significant event loop lag timers can prematurely expire.

Fixes: https://github.com/nodejs/undici/issues/1976

### DIFF
--- a/lib/timers.js
+++ b/lib/timers.js
@@ -13,13 +13,15 @@ function onTimeout () {
   while (idx < len) {
     const timer = fastTimers[idx]
 
-    if (timer.expires && fastNow >= timer.expires) {
-      timer.expires = 0
+    if (timer.state === 0) {
+      timer.state = fastNow + timer.delay
+    } else if (timer.state > 0 && fastNow >= timer.state) {
+      timer.state = -1
       timer.callback(timer.opaque)
     }
 
-    if (timer.expires === 0) {
-      timer.active = false
+    if (timer.state === -1) {
+      timer.state = -2
       if (idx !== len - 1) {
         fastTimers[idx] = fastTimers.pop()
       } else {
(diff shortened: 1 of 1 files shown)


ronag:
@kajiiiro would you like helping adding a test?

mcollina (commented):
Can you add a test?

lib/timers.js
```diff
@@ -13,13 +13,15 @@ function onTimeout () {
   while (idx < len) {
     const timer = fastTimers[idx]
 
-    if (timer.expires && fastNow >= timer.expires) {
-      timer.expires = 0
+    if (timer.state === 0) {
+      timer.state = fastNow + timer.delay
+    } else if (timer.state > 0 && fastNow >= timer.state) {
```
kajiiiro, line 18: @ronag 
expired check is executed in next event loop(after 1000msec).
it is executed in same event loop?
```suggestion
    }
    if (timer.state > 0 && fastNow >= timer.state) {
```

kajiiiro:
is this stopped?

> would you like helping adding a test?

Sorry, I have no idea. 
the test should confirm 'request timeout is not fired shorter than delay'
but it is hard...

- event loop interval is about 1sec
- built-in setimeout is used under 1sec

kajiiiro (approved):

ronag:
I'll work on a test when I have time.

ronag:
@mcollina looking into this, I realize that making a test for this is quite tricky and might require workers, etc... I suggest we either land this without a test or skip it. The case it is trying to fix is when you have significant event loop lag, which is an edge case...

mcollina (approved):
lgtm

(I trust you, but I would prefer to have a test anyway).

Let's land this and create an issue about writing this test?

PR #1984, opened 2023-03-03
wpt: update tests

KhafraDev:
- Uses the testharness from the wpt tests/resources folder, rather than moving it to /runner/resources/ folder.
- Moves some other things around
- Updates the WPTs in accordance to the [docs](https://github.com/nodejs/undici/blob/main/CONTRIBUTING.md#update-wpts)

cc @panva since the harness was just updated with your fix, is there anything else I need to do?


KhafraDev:
I will also make a PR in the node core util repo, so `git node wpt` can be used rather than git. We need an option to change the hardcoded, node-specific paths to our own.

panva:
> since the harness was just updated with your fix, is there anything else I need to do?

You may clean up the "Untitled" test name fix with the updated testharness.

ronag (approved):

Issue #1978, opened 2023-03-01
body.dump() is missing abort signal support


ronag:
We should support `body.dump({ signal })`

debadree25:
Should it throw an error when aborted with signal?

PR #1993, opened 2023-03-06
feat: add abort signal to body.dump()

debadree25:
Attempt to implement body.dump({ abort });

fixes: https://github.com/nodejs/undici/issues/1978

### DIFF
--- a/lib/api/readable.js
+++ b/lib/api/readable.js
@@ -147,14 +147,15 @@ module.exports = class BodyReadable extends Readable {
   async dump (opts) {
     let limit = opts && Number.isFinite(opts.limit) ? opts.limit : 262144
     const signal = opts && opts.signal
+    const abortFn = () => {
+      this.destroy()
+    }
     if (signal) {
       if (typeof signal !== 'object' || !('aborted' in signal)) {
         throw new InvalidArgumentError('signal must be an AbortSignal')
       }
       util.throwIfAborted(signal)
-      signal.addEventListener('abort', () => {
-        this.destroy()
-      }, { once: true })
+      signal.addEventListener('abort', abortFn, { once: true })
     }
     try {
       for await (const chunk of this) {
@@ -166,6 +167,10 @@ module.exports = class BodyReadable extends Readable {
       }
     } catch {
       util.throwIfAborted(signal)
+    } finally {
+      if (signal) {
+        signal.removeEventListener('abort', abortFn)
+      }
     }
   }
 }
--- a/lib/api/readable.js
+++ b/lib/api/readable.js
@@ -166,9 +166,7 @@ module.exports = class BodyReadable extends Readable {
       }
     } catch (err) {
       signal.throwIfAborted()
-      if (err.name === 'AbortError') {
         throw err
-      }
     }
   }
 }
--- a/lib/api/readable.js
+++ b/lib/api/readable.js
@@ -147,40 +147,38 @@ module.exports = class BodyReadable extends Readable {
   async dump (opts) {
     let limit = opts && Number.isFinite(opts.limit) ? opts.limit : 262144
     const signal = opts && opts.signal
+    function throwIfAborted () {
+      if (!signal) { return }
+      if (signal.throwIfAborted) {
+        signal.throwIfAborted()
+      } else {
+        if (signal.aborted) {
+          // DOMException not available < v17.0.0
+          const err = new Error('The operation was aborted')
+          err.name = 'AbortError'
+          throw err
+        }
+      }
+    }
     if (signal) {
       if (typeof signal !== 'object' || !('aborted' in signal)) {
         throw new InvalidArgumentError('signal must be an AbortSignal')
       }
-      if (!signal.throwIfAborted) {
-        // add in a throwIfAborted method to the signal
-        signal.throwIfAborted = function () {
-          if (signal.aborted) {
-            // DOMException not available < v17.0.0
-            const err = new Error('The operation was aborted')
-            err.name = 'AbortError'
-            throw err
-          }
-        }
-      }
-      signal.throwIfAborted()
+      throwIfAborted()
       signal.addEventListener('abort', () => {
         this.destroy()
       })
     }
     try {
       for await (const chunk of this) {
-        if (signal) {
-          signal.throwIfAborted()
-        }
+        throwIfAborted()
         limit -= Buffer.byteLength(chunk)
         if (limit < 0) {
           return
         }
       }
     } catch {
-      if (signal) {
-        signal.throwIfAborted()
-      }
+      throwIfAborted()
     }
   }
 }
--- a/lib/api/readable.js
+++ b/lib/api/readable.js
@@ -148,6 +148,7 @@ module.exports = class BodyReadable extends Readable {
     let limit = opts && Number.isFinite(opts.limit) ? opts.limit : 262144
     const signal = opts && opts.signal
     if (signal) {
+      signal.throwIfAborted()
       if (typeof signal !== 'object' || !('aborted' in signal)) {
         throw new InvalidArgumentError('signal must be an AbortSignal');
       }
--- a/lib/api/readable.js
+++ b/lib/api/readable.js
@@ -147,13 +147,13 @@ module.exports = class BodyReadable extends Readable {
   async dump (opts) {
     let limit = opts && Number.isFinite(opts.limit) ? opts.limit : 262144
     const signal = opts && opts.signal && opts.signal instanceof AbortSignal ? opts.signal : null
-    if (signal && signal.aborted) {
-      throw new RequestAbortedError()
+    if (signal) {
+      signal.throwIfAborted()
     }
     try {
       for await (const chunk of this) {
-        if (signal && signal.aborted) {
-          throw new RequestAbortedError()
+        if (signal) {
+          signal.throwIfAborted()
         }
         limit -= Buffer.byteLength(chunk)
         if (limit < 0) {
@@ -161,7 +161,7 @@ module.exports = class BodyReadable extends Readable {
         }
       }
     } catch (err) {
-      if (err instanceof RequestAbortedError) {
+      if (err.name === 'AbortError') {
         throw err
       }
     }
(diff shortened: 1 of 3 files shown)


KhafraDev:
there are failures in v12 and v14 due to the optional chaining

debadree25:
Removed @KhafraDev

debadree25:
Still fails 😕, looks like unrelated tests?

debadree25:
hmm probably not unrelated, investigating

ronag:
https://developer.mozilla.org/en-US/docs/Web/API/AbortSignal/throwIfAborted

ronag (approved):
Not perfect. As it won't cancel the iteration promise. But good enough.

Maybe a .destroy() call on abort can help?

debadree25:
Something like 
```js
signal.addEventListener('abort', () => {
        this.destroy()
})
```

debadree25:
Also not sure how to address the failures,
I tried using AbortController from `require('abort-controller')` but when using that the test simply hangs

lib/api/readable.js
```diff
@@ -146,15 +146,23 @@ module.exports = class BodyReadable extends Readable {
 
   async dump (opts) {
     let limit = opts && Number.isFinite(opts.limit) ? opts.limit : 262144
+    const signal = opts && opts.signal && opts.signal instanceof AbortSignal ? opts.signal : null
```
ronag, line 149:
```suggestion
    const signal = opts && opts.signal 
```

lib/api/readable.js
```diff
@@ -146,15 +146,23 @@ module.exports = class BodyReadable extends Readable {
...
+    if (signal) {
```
ronag, line 150:
```suggestion
    if (signal) {
      signal.throwIfAborted()
```

lib/api/readable.js
```diff
@@ -146,15 +146,23 @@ module.exports = class BodyReadable extends Readable {
...
     try {
       for await (const chunk of this) {
         limit -= Buffer.byteLength(chunk)
         if (limit < 0) {
           return
         }
       }
-    } catch {
-      // Do nothing...
+    } catch (err) {
```
ronag, line 162:
```suggestion
    } catch (err) {
      signal.throwIfAborted()
```

lib/api/readable.js
```diff
@@ -146,15 +146,23 @@ module.exports = class BodyReadable extends Readable {
...
   async dump (opts) {
     let limit = opts && Number.isFinite(opts.limit) ? opts.limit : 262144
+    const signal = opts && opts.signal && opts.signal instanceof AbortSignal ? opts.signal : null
+    if (signal) {
+      signal.addEventListener('abort', () => {
+        this.destroy()
+      })
+    }
     try {
       for await (const chunk of this) {
```
ronag, line 156:
```suggestion
      for await (const chunk of this) {
        signal.throwIfAborted()
```

ronag (commented):
from node:

```js
const validateAbortSignal = hideStackFrames((signal, name) => {
  if (signal !== undefined &&
      (signal === null ||
       typeof signal !== 'object' ||
       !('aborted' in signal))) {
    throw new ERR_INVALID_ARG_TYPE(name, 'AbortSignal', signal);
  }
});
```

lib/api/readable.js
debadree25: since when it is aborted we are calling .destroy, wont this automatically error?

ronag: depends on the timing, not guaranteed

debadree25: understood, adding these back

lib/api/readable.js
```diff
@@ -146,15 +146,30 @@ module.exports = class BodyReadable extends Readable {
 
   async dump (opts) {
     let limit = opts && Number.isFinite(opts.limit) ? opts.limit : 262144
+    const signal = opts && opts.signal
+    if (signal) {
+      signal.throwIfAborted()
```
ronag, line 151:
```suggestion
```

lib/api/readable.js
```diff
@@ -146,15 +146,30 @@ module.exports = class BodyReadable extends Readable {
...
           return
         }
       }
-    } catch {
-      // Do nothing...
+    } catch (err) {
+      signal.throwIfAborted()
+      if (err.name === 'AbortError') {
+        throw err
+      }
```
ronag, line 172:
```suggestion
        throw err
```

debadree25:
Quite weird failures again, looking into these in a while

lib/api/readable.js
```diff
@@ -146,15 +146,32 @@ module.exports = class BodyReadable extends Readable {
 
   async dump (opts) {
     let limit = opts && Number.isFinite(opts.limit) ? opts.limit : 262144
+    const signal = opts && opts.signal
+    if (signal) {
+      if (typeof signal !== 'object' || !('aborted' in signal)) {
+        throw new InvalidArgumentError('signal must be an AbortSignal')
+      }
+      console.log(signal.throwIfAborted)
```
ronag, line 154:
```suggestion
```

lib/api/readable.js
```diff
@@ -146,15 +146,32 @@ module.exports = class BodyReadable extends Readable {
...
           return
         }
       }
-    } catch {
-      // Do nothing...
+    } catch (err) {
+      if (signal) {
+        signal.throwIfAborted()
+      }
+      throw err
```
ronag, line 174:
```suggestion
```

lib/api/readable.js
```diff
@@ -146,15 +146,32 @@ module.exports = class BodyReadable extends Readable {
...
+          signal.throwIfAborted()
+        }
         limit -= Buffer.byteLength(chunk)
         if (limit < 0) {
           return
         }
       }
-    } catch {
-      // Do nothing...
+    } catch (err) {
```
ronag, line 170:
```suggestion
    } catch {
```

test/client-request.js
```diff
@@ -1,3 +1,5 @@
+/* globals AbortController */
```
debadree25, line 1: Does this not somehow inject `AbortController` for tests on older versions, i see this pattern being used in tests for eg: `test/fetch/client-fetch.js`

debadree25:
Is there a way to rerun CI, the failures in 18 and 19 look unrelated

test/client-request.js
KhafraDev: no, it's just telling prettier that AbortController is a global

debadree25: Any ideas then @KhafraDev what could we do about this test?

debadree25:
Yes doing so

debadree25:
Seems everything passing finally 🎉

lib/api/readable.js
```diff
@@ -146,15 +146,41 @@ module.exports = class BodyReadable extends Readable {
...
   async dump (opts) {
     let limit = opts && Number.isFinite(opts.limit) ? opts.limit : 262144
+    const signal = opts && opts.signal
+    if (signal) {
+      if (typeof signal !== 'object' || !('aborted' in signal)) {
+        throw new InvalidArgumentError('signal must be an AbortSignal')
+      }
+      if (!signal.throwIfAborted) {
+        // add in a throwIfAborted method to the signal
+        signal.throwIfAborted = function () {
```
KhafraDev, line 156: I don't think adding a public method here is a good solution, it'd be better to add it as a utility functions

```js
function throwIfAborted (signal) {
  if (signal.throwIfAborted) {
    signal.throwIfAborted()
  } else {
    // ...
  }
}
```

KhafraDev (changes requested):

lib/api/readable.js
debadree25: Updating

debadree25: Updated, does it look ok?

KhafraDev (approved):

lib/api/readable.js
```diff
@@ -146,15 +146,39 @@ module.exports = class BodyReadable extends Readable {
...
+      }
+    }
+    if (signal) {
+      if (typeof signal !== 'object' || !('aborted' in signal)) {
+        throw new InvalidArgumentError('signal must be an AbortSignal')
+      }
+      throwIfAborted()
+      signal.addEventListener('abort', () => {
+        this.destroy()
+      })
```
ronag, line 170: This event listener needs to be cleaned up or we have a memory leak.

lib/api/readable.js
```diff
@@ -146,15 +146,39 @@ module.exports = class BodyReadable extends Readable {
 
   async dump (opts) {
     let limit = opts && Number.isFinite(opts.limit) ? opts.limit : 262144
+    const signal = opts && opts.signal
+    function throwIfAborted () {
```
ronag, line 150: Could we move this into the util file? `throwIfAborted(signal)`

debadree25: Done!

lib/api/readable.js
debadree25: Made this `{ once: true }`

lib/api/readable.js
```diff
@@ -147,38 +147,25 @@ module.exports = class BodyReadable extends Readable {
...
     if (signal) {
       if (typeof signal !== 'object' || !('aborted' in signal)) {
         throw new InvalidArgumentError('signal must be an AbortSignal')
       }
-      throwIfAborted()
+      util.throwIfAborted(signal)
       signal.addEventListener('abort', () => {
         this.destroy()
-      })
+      }, { once: true })
```
ronag, line 157: this will still leak one we completed without abort

ronag: add a finally clause to remove it

lib/core/util.js
```diff
@@ -389,6 +389,20 @@ function isFormDataLike (object) {
   )
 }
 
+function throwIfAborted (signal) {
+  if (!signal) { return }
+  if (signal.throwIfAborted) {
```
ronag, line 394:
```suggestion
  if (typeof signal.throwIfAborted === 'function') {
```

lib/api/readable.js
debadree25: Ah sorry, completely distracted need more chai sigh, anyway have updated

PR #1995, opened 2023-03-08
fix: undici stream throwOnError

### LINKED ISSUE #1753: undici.stream ignores throwOnError

dancastillo:
undici.stream throwOnError fixes #1753

### DIFF
--- a/lib/api/api-stream.js
+++ b/lib/api/api-stream.js
@@ -117,6 +117,9 @@ class StreamHandler extends AsyncResource {
             )
           )
         })
+        .on('error', (err) => {
+          this.onError(err)
+        })
       this.res = pt
       return
     }
--- a/lib/api/api-stream.js
+++ b/lib/api/api-stream.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const { finished } = require('stream')
+const { finished, PassThrough } = require('stream')
 const {
   InvalidArgumentError,
   InvalidReturnValueError,
@@ -16,7 +16,7 @@ class StreamHandler extends AsyncResource {
       throw new InvalidArgumentError('invalid opts')
     }
 
-    const { signal, method, opaque, body, onInfo, responseHeaders } = opts
+    const { signal, method, opaque, body, onInfo, responseHeaders, throwOnError } = opts
 
     try {
       if (typeof callback !== 'function') {
@@ -57,6 +57,7 @@ class StreamHandler extends AsyncResource {
     this.trailers = null
     this.body = body
     this.onInfo = onInfo || null
+    this.throwOnError = throwOnError || false
 
     if (util.isStream(body)) {
       body.on('error', (err) => {
@@ -76,8 +77,8 @@ class StreamHandler extends AsyncResource {
     this.context = context
   }
 
-  onHeaders (statusCode, rawHeaders, resume) {
-    const { factory, opaque, context } = this
+  onHeaders (statusCode, rawHeaders, resume, statusMessage) {
+    const { factory, opaque, context, callback } = this
 
     if (statusCode < 200) {
       if (this.onInfo) {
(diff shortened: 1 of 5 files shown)


mcollina (approved):
lgtm

ronag:
Wait with landing this.

lib/api/api-stream.js
```diff
@@ -96,6 +97,24 @@ class StreamHandler extends AsyncResource {
...
 
+    if (this.throwOnError && statusCode >= 400) {
+      const headers = this.responseHeaders === 'raw' ? util.parseRawHeaders(rawHeaders) : util.parseHeaders(rawHeaders)
+      const chunks = []
+      const pass = new PassThrough()
+      pass
+        .on('data', (chunk) => chunks.push(chunk))
+        .on('end', () => {
+          const payload = Buffer.concat(chunks).toString('utf8')
+          this.runInAsyncScope(
```
ronag, line 108: This won't work since `getResolveErrorBodyCallback` expects a `BodyReadable`.

ronag (changes requested):
`getResolveErrorBodyCallback` needs to be rewritten to work without the `body` parameter. Also, seems to be missing tests as the current broken implementation seems to pass?

lib/api/api-stream.js
ronag: Should also automatically parse json like request does.

lib/core/util.js
```diff
@@ -342,6 +342,38 @@ function getSocketInfo (socket) {
   }
 }
 
+async function getResolveErrorBodyCallback ({ callback, body, contentType, statusCode, statusMessage, headers, payload }) {
+  if (typeof payload === 'string' && statusCode >= 400) {
+    process.nextTick(callback, new ResponseStatusCodeError(`Response status code ${statusCode}${statusMessage ? `: ${statusMessage}` : ''}`, statusCode, headers, payload))
+    return
+  }
```
ronag, line 349: This shouldn't be necessary.

test/client-stream.js
```diff
@@ -785,4 +785,38 @@ test('stream legacy needDrain', (t) => {
...
+      t.teardown(client.close.bind(client))
+
+      client.stream({
+        path: '/',
+        method: 'GET',
+        throwOnError: true,
+        opaque: new PassThrough()
+      }, ({ opaque: pt }) => {
+        pt.on('data', () => {
+          t.not.pass()
```
KhafraDev, line 812: I don't think this is a thing, the equivalent being t.fail.

dancastillo: updated to `t.fail()`

dancastillo:
@ronag @KhafraDev addressed changes ptal

KhafraDev (approved):

lib/api/api-stream.js
```diff
@@ -96,6 +98,29 @@ class StreamHandler extends AsyncResource {
...
+            callback,
+            null,
+            new ResponseStatusCodeError(
+              `Response status code ${statusCode}${statusMessage ? `: ${statusMessage}` : ''}`,
+              statusCode,
+              headers,
+              payload
+            )
+          )
+        })
```
ronag, line 119: Needs to handle error event

ronag (changes requested):

dancastillo:
@ronag added error event handling ptal

Issue #1962, opened 2023-02-22
Make ProxyAgent take another Agent as an option


mcollina:
In https://github.com/nodejs/undici/blob/0beedf9911a6104aec33ef6b84b54e48f0716e95/lib/proxy-agent.js#L41, we create a new `Agent` for every `ProxyAgent`. I think we should make `ProxyAgent` take an instance of `Agent` as an option, making something like the following possible:

```ts
import { ProxyAgent, request, setGlobalDispatcher } from "undici"
import { CookieAgent } from "http-cookie-agent/undici"
import { CookieJar } from "tough-cookie"


async function test() {
  const proxyAgent = new ProxyAgent("http://127.0.0.1:10810")
  setGlobalDispatcher(proxyAgent)
  const jar = new CookieJar()
  const agent = new CookieAgent({ cookies: { jar } })
  await request("https://test.com", {
    dispatcher: agent,
  })
}
```
Help would be greatly appreciated! 🙏

_Originally posted by @tgioer in https://github.com/nodejs/undici/discussions/1957_

RishabhKodes:
Hey @mcollina, I'm taking up this issue.

jabali2004:
Hey @mcollina 👋🏼
Can we close that one? Since #2003 has been merged.

PR #1998, opened 2023-03-09
enabling `ProxyAgent` take an instance of `Agent` as an option

RishabhKodes:
In the updated constructor, the `opts` parameter is checked to see if it contains an `agent` property that is an instance of `Agent`. If it is, then the `[kAgent]` is set to the provided instance. Otherwise, a new `Agent` instance is created and assigned to the `[kAgent]`.

fixes: https://github.com/nodejs/undici/issues/1962

### DIFF
--- a/lib/proxy-agent.js
+++ b/lib/proxy-agent.js
@@ -38,7 +38,6 @@ class ProxyAgent extends DispatcherBase {
   constructor (opts) {
     super(opts)
     this[kProxy] = buildProxyOptions(opts)
-    this[kAgent] = new Agent(opts)
     this[kInterceptors] = opts.interceptors && opts.interceptors.ProxyAgent && Array.isArray(opts.interceptors.ProxyAgent)
       ? opts.interceptors.ProxyAgent
       : []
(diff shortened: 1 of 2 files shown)


KhafraDev:
Can you add in a test?

RishabhKodes:
> Can you add in a test?

Added the tests for both scenarios, take a look.

mcollina (commented):
Could you add docs for this change?

ronag:
How is this different from #2003

RishabhKodes:
> How is this different from #2003



> How is this different from #2003

#2003 majorly targets the broader issue of successive requests to ProxyAgent getting blocked, thus introducing a new `Client` for solving the issue. This pr also does somewhat of the same thing enabling the user to create an instance of Agent (similar but less complex than creating a `Client`) inside the `ProxyAgent`. 

This pr solves the inability to use ProxyAgent and any other Agent at the same time (example mentioned here, https://github.com/nodejs/undici/discussions/1957).

ronag:
Do you think we could use the same pattern with a factory?

RishabhKodes:
Closing this as https://github.com/nodejs/undici/pull/2003 solves the issue in almost the same way (and that solution has been approved)

PR #2002, opened 2023-03-10
fix(fetch): third party abortcontrollers throwing errors

KhafraDev:
Refs: https://github.com/nodejs/undici/pull/1910#issuecomment-1464495619

I'm unsure why our tests didn't catch this since we do test third party abortcontrollers. I added in another test where I could replicate the issue regardless. Maybe this was the flaky test, but then it should have consistently failed?! So many questions...


lib/fetch/request.js
```diff
@@ -358,9 +358,13 @@ class Request {
           ac.abort(this.reason)
         }
 
-        if (getEventListeners(signal, 'abort').length >= defaultMaxListeners) {
-          setMaxListeners(100, signal)
-        }
+        // Third-party AbortControllers may not work with these.
+        // See https://github.com/nodejs/undici/pull/1910#issuecomment-1464495619
+        try {
```
KhafraDev, line 363: I wasn't sure if vitest would override EventTarget as well, meaning an instanceof check wouldn't work. This way also shouldn't slow down anyone not overriding globals too so I'd call it a half-shitty solution lol

Issue #2001, opened 2023-03-10
Slow http CONNECT requests will block other requests when using ProxyAgent


andrewfecenko:
## Bug Description

When using ProxyAgent, a slow CONNECT request to the proxy will block the Client's queue (this[kClient] in proxy-agent.js) and can cause delays in issues any requests with that ProxyAgent.

## Reproducible By

We found that the easiest way to reproduce this with a test was to add a delay for the first connect request to the proxy via proxy.authenticate:

```
test('delayed proxy-agent request blocks queue', async (t) => {
  t.plan(2)
  const server = await buildServer()
  const proxy = await buildProxy()

  const serverUrl = `http://localhost:${server.address().port}`
  const proxyUrl = `http://localhost:${proxy.address().port}`
  const proxyAgent = new ProxyAgent({auth: Buffer.from('user:pass').toString('base64'), uri: proxyUrl});
  let firstConnect = true;
  await new Promise(async (resolve) => {
    let connectCount = 0
    console.log('running test');
    proxy.authenticate = function (req, res) {
      if (++connectCount === 2) {
        console.log('allowing connect');
        firstConnect = false;
        res(null, true)
      } else {
        console.log('delaying connect');
        setTimeout(() => {
          console.log('delayed connect');
          t.equal(firstConnect, true);
          res(null, true);
        }, 1000);
      }
    }

    server.on('request', (req, res) => {
      res.end()
    })

    request(`${serverUrl}`, { dispatcher: proxyAgent })
      .then((resp) => {
        resp.body.on('error', () => {}).destroy()
      })
      .catch(() => {})
    request(`${serverUrl}`, { dispatcher: proxyAgent })
      .then((resp) => {
        resp.body.on('error', () => {}).destroy()
        resolve();
      })
      .catch(() => {})
  })
  console.log('done');
  t.pass()
  server.close()
  proxy.close()
  proxyAgent.close()
})
```

## Expected Behavior

A single request issued with a ProxyAgent should not block other requests also using that ProxyAgent.

## Environment

Ran the above test using node v19.4.0 on macOS Ventura 13.2.1

### Additional context

We discovered this issue after investigating delays in requests that use undici's request function with a ProxyAgent passed in as the dispatcher. We were instantiating a single ProxyAgent and using it in all requests, but we have since made a change to instantiate a new ProxyAgent for every single `request` call. This appears to have fixed the issue where a single request blocks other successive requests.

[...]

PR #2003, opened 2023-03-11
Add clientFactory option to ProxyAgent

andrewfecenko:
The default use of a Client means that HTTP CONNECT requests to the Proxy will block successive requests. Giving consumers the option to use a Pool ensures that these requests will not block.

fixes: #2001
https://github.com/nodejs/undici/issues/2001

### DIFF
--- a/docs/api/ProxyAgent.md
+++ b/docs/api/ProxyAgent.md
@@ -19,6 +19,7 @@ Extends: [`AgentOptions`](Agent.md#parameter-agentoptions)
 * **uri** `string` (required) - It can be passed either by a string or a object containing `uri` as string.
 * **token** `string` (optional) - It can be passed by a string of token for authentication.
 * **auth** `string` (**deprecated**) - Use token.
+* **clientFactory** `(origin: URL, opts: Object) => Dispatcher` - Default: `(origin, opts) => new Pool(origin, opts)`
 
 Examples:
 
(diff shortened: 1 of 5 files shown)


Sebmaster:
I think it'd be good to change the default behavior to allocate a Pool instead of using a Client. Imo the blocking behavior of the tunnel setup should be considered a bug to be fixed.

mcollina (commented):
Good work! Could you add the docs for this change?

I would also switch this to use a Pool by default.

ronag:
All other dispatchers call the option just factory. Should we maybe be consistent?

andrewfecenko:
@ronag We could do that but I chose clientFactory as I wasn't sure if we were okay with the factory being passed into the Agent instantiation. I thought it would be best to be explicit in that this factory only affects the instantiation of this[kClient]. I'm happy to change the naming to factory if we think that it's okay to pass it into factory as well (or I could explicitly exclude it from the Agent instantiation so that we use the default Agent factory). Also happy to rename clientFactory to proxyClientFactory if we only want it passed in there and want a slightly more applicable name.

I will also change the default factory to Pool as suggested 👍

andrewfecenko:
Cool, just pushed up two commits.
First commit uses Pool by default and allows clientFactory to return any Dispatcher (was explicitly restricted to just Client and Pool in the prior commit)
The second commit adds documentation for this new ProxyAgent argument. If there are other places I should add docs, please let me know and I'd be happy to add em

anonrig (approved):

mcollina (approved):
lgtm

PR #2013, opened 2023-03-17
fix: issue 2009

### LINKED ISSUE #2009: Race condition w/ body cancel, and strange mixing results
## Bug Description

Calling response.body.cancel — but only with a message — can lead to controller already closed errors.

## Reproducible By

```js

await fetch("https://httpbin.org", {
  method: "POST",
  body: "{}",
  headers: { "Content-Type": "application/json" },
}).then(
  async (resp) => {
    await resp.body.cancel("Some message");
  },
  (err) => {}
);
```

## Expected Behavior

HTTPBin returns a 405 with an HTML body for this request. Calling `resp.body.cancel(reasonString)` should not error.

## Logs & Screenshots

About 80% of the time, the code above gives the following (uncaught) rejection:

```
TypeError [ERR_INVALID_STATE]: Invalid state: Controller is already closed
    at new NodeError (node:internal/errors:399:5)
    at ReadableStreamDefaultController.enqueue (node:internal/webstreams/readablestream:1036:13)
    at fetchParams.controller.resume (/Users/ethanresnick/Desktop/node_modules/undici/lib/fetch/index.js:1888:41)
  code: 'ERR_INVALID_STATE'
```

It seems like there might be some sort of race condition.

Interestingly, if the code calls `resp.body.cancel()` (i.e., leaving the reason string out), the call never seems to throw.

## Environment

MacOS 13.2
Nodejs 19.6.0
undici 5.21.0

KhafraDev:
Fixes #2009

It ended up being basically exactly what they said in the issue, the body was being canceled with a string, which wasn't getting caught by the isError check. Also a race condition.

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -1845,6 +1845,7 @@ async function httpNetworkFetch (
       // 4. Set bytes to the result of handling content codings given
       // codings and bytes.
       let bytes
+      let isFailure
       try {
         const { done, value } = await fetchParams.controller.next()
 
@@ -1859,6 +1860,10 @@ async function httpNetworkFetch (
           bytes = undefined
         } else {
           bytes = err
+
+          // err may be propagated from the result of calling readablestream.cancel,
+          // which might not be an error. https://github.com/nodejs/undici/issues/2009
+          isFailure = true
         }
       }
 
@@ -1878,7 +1883,7 @@ async function httpNetworkFetch (
       timingInfo.decodedBodySize += bytes?.byteLength ?? 0
 
       // 6. If bytes is failure, then terminate fetchParams’s controller.
-      if (isErrorLike(bytes)) {
+      if (isFailure) {
         fetchParams.controller.terminate(bytes)
         return
       }
--- a/test/fetch/issue-2009.js
+++ b/test/fetch/issue-2009.js
@@ -1,0 +1,28 @@
+'use strict'
+
+const { test } = require('tap')
+const { fetch } = require('../..')
+const { createServer } = require('http')
+const { once } = require('events')
+
+test('issue 2009', async (t) => {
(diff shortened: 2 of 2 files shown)


KhafraDev:
test failures are unrelated

ethanresnick:
Thanks for such a quick fix @KhafraDev!

In your debugging, did you figure out anything about why this issue would’ve been happening only for some origins? (Ie, even when `cancel(reasonString)` is used, the error never seems to happen if the request is to protegoapi.com rather than httpbin, like I mentioned in the original ticket.) Is that something to do with a race condition?

lib/fetch/index.js
```diff
@@ -1845,6 +1845,7 @@ async function httpNetworkFetch (
       // 4. Set bytes to the result of handling content codings given
       // codings and bytes.
       let bytes
+      let isFailure
```
ronag, line 1848:
```suggestion
      let isFailure = false
```

ronag (approved):

lib/fetch/index.js
```diff
@@ -1859,6 +1860,10 @@ async function httpNetworkFetch (
           bytes = undefined
         } else {
           bytes = err
```
ronag, line 1862:
```suggestion
          bytes = isErrorLike(err) ? err : new Error(err)
```

ronag: alternative solution...

KhafraDev:
@ethanresnick 

> did you figure out anything about why this issue would’ve been happening only for some origins

It probably had something to do with the race condition, but it's most likely an issue with how httpbin was closing/ending the response. Note that in the test I use `res.socket.end`, which replicated the issue, instead of `res.end`, which didn't.

ethanresnick:
Makes sense! Thanks again for the quick fix :)

Issue #2021, opened 2023-03-22
When Content-Length is specified, fetch() raises RequestContentLengthMismatchError on redirection


geryogam:
## Bug Description

When the `Content-Length` request header field is manually specified (instead of automatically generated), `fetch()` raises a `RequestContentLengthMismatchError` in case of redirection.

## Reproducible By

We send a `POST` request to the `/sirene/public/recherche` target URI to search for a non-existing company whose name `foobarbaz` is provided in the request body:

```javascript
const uri = 'https://www.sirene.fr/sirene/public/recherche';
const method = 'POST';
const body = 'recherche.sirenSiret=&recherche.raisonSociale=foobarbaz&recherche.adresse=&recherche.commune=&recherche.excludeClosed=true&__checkbox_recherche.excludeClosed=true&recherche.captcha=';
const headers = {'Content-Type': 'application/x-www-form-urlencoded', 'Content-Length': Buffer.byteLength(body)};
const response = await fetch(uri, {method, headers, body});
```

The target URI `/sirene/public/recherche` replies with a `302` response with a `Location: /sirene/error/autre.action` header field. So `fetch` automatically (by default) redirects the original request to the target URI `/sirene/error/autre.action` and raises a `RequestContentLengthMismatchError`:

```
Uncaught TypeError: fetch failed
    at Object.fetch (node:internal/deps/undici/undici:11413:11)
    at async REPL5:1:47 {
  cause: RequestContentLengthMismatchError: Request body length does not match content-length header
      at write (node:internal/deps/undici/undici:9907:41)
      at _resume (node:internal/deps/undici/undici:9885:33)
      at resume (node:internal/deps/undici/undici:9787:7)
      at connect (node:internal/deps/undici/undici:9776:7) {
    code: 'UND_ERR_REQ_CONTENT_LENGTH_MISMATCH'
  }
}
```

## Expected Behavior

The specified `Content-Length` request header field is correct in the original request so I don’t expect a `RequestContentLengthMismatchError` after automatic redirection.

## Logs & Screenshots

I assume that during redirection, `fetch()` resends the original request but with the following modifications:

- the target URI is changed from `/sirene/public/recherche` to `/sirene/error/autre.action`;
- the `POST` request method is changed to `GET`;
- the request body is removed.

[...]

PR #2022, opened 2023-03-23
fix(fetch): remove content-length header on redirect

KhafraDev:
Fixes https://github.com/nodejs/undici/issues/2021

### DIFF
--- a/lib/fetch/constants.js
+++ b/lib/fetch/constants.js
@@ -48,11 +48,17 @@ const requestCache = [
   'only-if-cached'
 ]
 
+// https://fetch.spec.whatwg.org/#request-body-header-name
 const requestBodyHeader = [
   'content-encoding',
   'content-language',
   'content-location',
-  'content-type'
+  'content-type',
+  // See https://github.com/nodejs/undici/issues/2021
+  // 'Content-Length' is a forbidden header name, which is typically
+  // removed in the Headers implementation. However, undici doesn't
+  // filter out headers, so we add it here.
+  'content-length'
 ]
 
 // https://fetch.spec.whatwg.org/#enumdef-requestduplex
--- a/test/fetch/issue-2021.js
+++ b/test/fetch/issue-2021.js
@@ -1,0 +1,32 @@
+'use strict'
+
+const { test } = require('tap')
+const { once } = require('events')
+const { createServer } = require('http')
+const { fetch } = require('../..')
+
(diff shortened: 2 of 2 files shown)


annevk:
If you don't forbid `content-length`, how do you deal with conflicts when someone passes `content-length` and a string that's quite a bit longer or shorter? The layering here is quite a bit different from how Fetch does it.

ronag (commented):
Remind me again why we can't just check for this in `Headers.append` like the spec does?

KhafraDev:
> how do you deal with conflicts when someone passes content-length and a string that's quite a bit longer or shorter?

An error is thrown when a content-length header mismatches

> Remind me again why we can't just check for this in Headers.append like the spec does?

There's a few reasons I could think of:

- If we filter out every header that the spec does, we bring back issues with cookie headers (linked in the issue).
- If choose to only filter some headers, it will create subtle bugs for people, who can no longer use the spec or another environment (Deno, cfw, bun) as reference. This will probably cause issues w/ wintercg in the future? Basically node would be doing its own thing, while browsers follow the spec and other server environments have the same behavior.
- It'll break setting some headers with node's [setHeaders](https://nodejs.org/dist/latest-v19.x/docs/api/all.html#all_http_outgoingmessagesetheadersheaders).

```js
  const headers = new Headers({ 'content-length': '5' })
  res.setHeaders(headers)
  res.write('hello')
  res.end()
}).listen(0)
```

ronag:
I don't think all (?) of those are valid though. Since this logic would only occur if the header guard is set to "request" which is only the case internally inside of fetch?

ronag:
1 partly, and yea I guess 2 is still relevant.

ronag:
I'm not sure... as soon as we go away from the spec, things get complicated.

KhafraDev:
We've been away from the spec for a while now in regards to Headers, it hasn't caused too many issues.

Relevant tweet: https://twitter.com/matteocollina/status/1537466032490700802

KhafraDev:
It's an unfortunate situation since neither option is good. Filtering headers causes issues and not filtering them also causes issues. But I also think the spec, regarding headers, is too restrictive for the backend.

KhafraDev:
not sure if it's needed but if we need to discuss filtering headers a new issue might be a better place

geryogam:
Thank you for the PR @KhafraDev!

Issue #1926, opened 2023-02-13
fetch() AbortSignal.timeout() does not work as expected


mcollina:
## Bug Description

It seems passing `{ signal: AbortSignal.timeout(50) }` to `fetch()` is not taken into consideration.

## Reproducible By

Run `node test/fetch/abort.js`. The `AbortSignal.timeout()` test should pass immediately but it waits for 30s. This is the output:

```
 Subtest: Allow the usage of custom implementation of AbortController
    ok 1 - should be equal
    1..1
ok 1 - Allow the usage of custom implementation of AbortController # time=6.82ms

# Subtest: allows aborting with custom errors
    # Subtest: Using AbortSignal.timeout with cause
        1..2
        ok 1 - should be equal
        ok 2 - should be equal
    ok 1 - Using AbortSignal.timeout with cause # time=30057.744ms

    # Subtest: Error defaults to an AbortError DOMException
        ok 1 - expect rejected Promise
        1..1
    ok 2 - Error defaults to an AbortError DOMException # time=7.816ms

    1..2
ok 2 - allows aborting with custom errors # time=30072.88ms

1..2
# time=30085.421ms
```

## Expected Behavior

The `fetch` request should be completed in the expected timeout.

## Environment

Node 18 and 19, on both Windows and Mac OS X.

mcollina:
cc @KhafraDev

KhafraDev:
Did a bisect, this is caused by f035bbc6480883db07b79006202f13662895e702. cc @ronag

debadree25:
It seems, that the gc cleans up `ac` lot before it is used

from some console logs 
```console
TAP version 13
# Subtest: Allow the usage of custom implementation of AbortController
    ok 1 - should be equal
    1..1
ok 1 - Allow the usage of custom implementation of AbortController # time=13.181ms

# Subtest: allows aborting with custom errors
    # Subtest: Using AbortSignal.timeout with cause
        1..2
cleaned
aborting undefined
aborted
```

I was wondering if using the `aborted()` utility introduced in https://github.com/nodejs/node/pull/46494 would be helpful here? I tried something like this

```diff
diff --git a/lib/fetch/request.js b/lib/fetch/request.js
index 9764871..5a0e17d 100644
--- a/lib/fetch/request.js
+++ b/lib/fetch/request.js
@@ -30,6 +30,7 @@ const { URLSerializer } = require('./dataURL')
 const { kHeadersList } = require('../core/symbols')
[...]

PR #2025, opened 2023-03-26
fix: avoid mem leak

ronag:
Refs: https://github.com/nodejs/node/issues/46435
Refs: https://github.com/nodejs/undici/issues/1926

### DIFF
--- a/lib/fetch/request.js
+++ b/lib/fetch/request.js
@@ -354,8 +354,12 @@ class Request {
       if (signal.aborted) {
         ac.abort(signal.reason)
       } else {
+        const weakSelf = new WeakRef(this)
         const abort = function () {
-          ac.abort(this.reason)
+          const self = weakSelf.deref()
+          if (self) {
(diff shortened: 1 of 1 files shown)


ronag:
@jamesdiacono @stalkerg Can you try and see if this fixes your issue?

ronag:
If this helps it almost feels like a V8 bug.

stalkerg:
I will try today, not so easy however in my case.

stalkerg:
No, it's not helped me.

stalkerg:
I think the reason is somewhere here:
```
signal.addEventListener('abort', abort, { once: true })
requestFinalizer.register(this, { signal, abort })
```

Because our `abort` function borrows `ac` object, which is bounded to `this` by `this[kSignal] = ac.signal`.

stalkerg:
Very interesting! If I comment this `requestFinalizer.register(this, { signal, abort })` it's also help.

Issue #1557, opened 2022-07-18
"ReferenceError: AbortSignal is not defined" during jest execution


JacopoPatroclo:
## Bug Description

When I try to run with jest a test suite that uses a file that import `undici` I get this error

    ReferenceError: AbortSignal is not defined

      at Object.<anonymous> (../node_modules/undici/lib/fetch/request.js:839:3)
      at Object.<anonymous> (../node_modules/undici/index.js:94:28)

## Reproducible By

Create a basic jest configuration with `testEnvironment: 'node',`
Create a js file that imports the `undici` library
Create a spec file to test somthing exported from that file
Run jest

## Expected Behavior

To not throw that exception

## Environment

Node v16.14.0
Jest v27.2.1

### Additional context

Googling the error I found out that another lib had to solve a [similar issue](https://github.com/isaacs/node-lru-cache/issues/239). They think it's related on how jest is mocking AbortSignal and AbortController.

ronag:
Feels more like a jest issue

mcollina:
We are also running some jest tests to verify compatibility, so this is a bit unexpected. There is likely something else at play.

Can you provide steps to reproduce? We often need a [reproducible example](https://stackoverflow.com/help/minimal-reproducible-example), e.g. some code that allows someone else to recreate your problem by just copying and pasting it. If it involves more than a couple of different file, create a new repository on GitHub and add a link to that.

watson:
Odd, I'm getting this issue in a just test suite while upgrading to the newest version of undici. I know it's not the fault of undici, but just thought I'd mention it in this issue in case anybody had found the culprit since this issue was closed.

feugy:
In case this helps, the regression started with version 5.6.0. Working with 5.5.1 works. Trying to assemble a reproduction case.

mcollina:
A PR would be helpful.

LaurentZuijdwijk:
Still no solution for this? I imagine this breaking a lot of projects and it prevents adoption within my organisation.

mcollina:
@LaurentZuijdwijk would you like to send a PR?

noel-dolan:
Also seeing this by simply adding @fastify/http-proxy to project, jest test then start failing:

ReferenceError: AbortSignal is not defined
        at Object.<anonymous> (/Users/ND/development/nebula/_temp/chive/node_modules/undici/lib/fetch/request.js:890:3)
[...]

PR #2029, opened 2023-03-27
'ReferenceError: AbortSignal is not defined' during jest execution

DariaGritsienko:
Fixes https://github.com/nodejs/undici/issues/1557

### DIFF
--- a/lib/fetch/request.js
+++ b/lib/fetch/request.js
@@ -30,6 +30,7 @@ const { URLSerializer } = require('./dataURL')
 const { kHeadersList } = require('../core/symbols')
 const assert = require('assert')
 const { setMaxListeners, getEventListeners, defaultMaxListeners } = require('events')
+const { AbortSignal } = require('node-abort-controller')
 
 let TransformStream = globalThis.TransformStream
 
--- a/lib/fetch/request.js
+++ b/lib/fetch/request.js
@@ -30,15 +30,20 @@ const { URLSerializer } = require('./dataURL')
 const { kHeadersList } = require('../core/symbols')
 const assert = require('assert')
 const { setMaxListeners, getEventListeners, defaultMaxListeners } = require('events')
-const { AbortSignal } = require('node-abort-controller')
 
 let TransformStream = globalThis.TransformStream
+let AbortSignal = global.AbortSignal;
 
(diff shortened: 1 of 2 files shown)


mcollina:
@DariaGritsienko thanks for opening this! https://github.com/nodejs/undici/issues/2021 seems already fixed, are you sure you linked to the right PR?

DariaGritsienko:
yes, thanks, fixed it - https://github.com/nodejs/undici/issues/1557
@mcollina

lib/fetch/request.js
```diff
@@ -30,6 +30,7 @@ const { URLSerializer } = require('./dataURL')
 const { kHeadersList } = require('../core/symbols')
 const assert = require('assert')
 const { setMaxListeners, getEventListeners, defaultMaxListeners } = require('events')
+const { AbortSignal } = require('node-abort-controller')
```
mcollina, line 33: We should be using Node.js core `AbortSignal` if it is available

package.json
```diff
@@ -69,6 +69,7 @@
   "devDependencies": {
     "@sinonjs/fake-timers": "^10.0.2",
     "@types/node": "^18.0.3",
+    "node-abort-controller": "^3.1.1",
```
mcollina, line 72: this must be a dependency.

mcollina (changes requested):

mcollina (approved):
lgtm, let's see if this works!

DariaGritsienko:
I studied this question a little more and found that jest adds in the jest-environment-node package global values, and they fixed it to auto in https://github.com/facebook/jest/pull/12642/files (v28.0.0)

`const contextGlobals = new Set(Object.getOwnPropertyNames(global));
    for (const nodeGlobalsKey of nodeGlobals) {
      if (!contextGlobals.has(nodeGlobalsKey)) {
        // @ts-expect-error
        global[nodeGlobalsKey] = globalThis[nodeGlobalsKey];
      }
    }`
    
I think undiсi will be better without this code... :) 

if people run into this problem they will need to install jest-environment-node package version >=28.0.0 and update jest >=28.0.0

I'll go try to update jest in my project, if anything else I'll be back

@mcollina

Issue #2026, opened 2023-03-26
`fetch` can make `worker_threads` stuck and process hang


AriPerkkio:
## Bug Description

When `undici` is run inside `worker_threads` it may prevent the worker from terminating and makes the whole process hang.

## Reproducible By

This Node script runs `undici` in worker threads and does 4x requests there before exiting. If this does not reproduce on your fast environment, try increasing the requests and `const ROUNDS`. 

```
"undici": "^5.21.0"
```

<details>

```js
// Save as `undici.mjs` and run as `node uncidi.mjs`
import { isMainThread, parentPort, Worker } from "node:worker_threads";
import { fileURLToPath } from "node:url";
import { cpus } from "node:os";

const THREADS = cpus().length - 1;
const ROUNDS = 10;
const filename = fileURLToPath(import.meta.url);

if (isMainThread) {
  const stuckWorkers = [];
  const interval = setInterval(() => {
    if (stuckWorkers.length > 0) {
      console.log(`There are ${stuckWorkers.length} stuck workers`);
    }
  }, 1000);

  const timeout = setTimeout(() => {
    setTimeout(() => console.log("Still running"), 5_000); // This is never called so process.exit does something, but it's still unable to terminate worker running undici
    console.log("Forcing process.exit on main thread");
    process.exit();
  }, 30_000);

  const task = async () => {
    const worker = new Worker(filename, {});

    return new Promise((resolve) => {
      worker.on("message", (message) => {
        if (message !== "DONE") return;

        const timer = setTimeout(() => {
          console.log("Unable to terminate");
          stuckWorkers.push(worker);
        }, 10_000);
        worker.terminate().then(() => {
          clearTimeout(timer);
          resolve();
        });
      });
    });
  };

  const pool = Array(ROUNDS).fill(task);

  async function execute() {
    const task = pool.shift();

    if (task) {
      await task();
      return execute();
    }
  }

  await Promise.all(
    Array(THREADS)
      .fill(execute)
      .map((task) => task())
  );
  console.log("All done!");
  clearTimeout(timeout);
  clearInterval(interval);
} else {
  const { fetch } = await import("undici");

  const request = async (pathWithParams) => {
    const url = new URL(pathWithParams, "https://api.spacexdata.com");

    const timer = setTimeout(
      () => console.log("Request taking more than 10s"), // This is never logged so requests are not stuck
      10_000
    );
    const output = await fetch(url.href, {
      method: "GET",
[...]

PR #2032, opened 2023-03-29
test: issue #2026

AriPerkkio:
- Failing test case for #2026

image

Example of failure on CI: https://github.com/AriPerkkio/undici/actions/runs/4551514748/jobs/8025677893#step:14:12832

Somehow `tap` is able to kill the process even though workers cannot terminate. I'm not sure how it does that as the reproduction case of #2026 cannot exit main thread even when it calls `process.exit()`.

Replace `fetch` with `request` and test case will pass.

```diff
const requests = Array(count)
-  .fill(urls.map((url) => fetch(url).then((r) => r.json())))
+  .fill(urls.map((url) => request(url).then((r) => r.body.json())))
```

### DIFF
--- a/test/fetch/issue-2026.js
+++ b/test/fetch/issue-2026.js
@@ -1,0 +1,75 @@
+'use strict'
+
+const { cpus } = require('os')
+const { isMainThread, Worker, parentPort } = require('worker_threads')
+const { test } = require('tap')
+const { fetch } = require('../..')
+
+const THREADS = cpus().length - 1
+const ROUNDS = 50
+const PARALLEL_REQUESTS = 6
+
+if (isMainThread) {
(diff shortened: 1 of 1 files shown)


test/fetch/issue-2026.js
```diff
@@ -0,0 +1,75 @@
...
+      Array(THREADS)
+        .fill(execute)
+        .map((task) => task())
+    )
+
+    clearTimeout(timeout)
+  })
+} else {
+  const urls = [
+    new URL('/v4/starlink', 'https://api.spacexdata.com'),
```
mcollina, line 62: Please avoid doing any calls to external servers

mcollina: I think figuring that out will get us the fix.

PR #2037, opened 2023-03-31
fetch: treat content-encoding as case-insensitive & remove x-deflate

KhafraDev:
RFC 7231 says that content-condings should be treated as case-insensitive.

I also removed x-deflate support because it doesn't seem to be an official content-coding, feel free to correct me if I'm wrong.

Also removes the regexes that would match invalid codings such as `agzip` etc., although I think this was caught elsewhere as the bodies weren't being decompressed from my testing.


ronag (approved):

mcollina (approved):
lgtm

KhafraDev:
thanks, codecov...

PR #2048, opened 2023-04-04
fix: set window option properly

KhafraDev:
The line above we check that it's either null (only valid option) or undefined (missing), so for it to exist, it must be `null` here.

### DIFF
--- a/lib/fetch/request.js
+++ b/lib/fetch/request.js
@@ -133,7 +133,7 @@ class Request {
     }
 
     // 11. If init["window"] exists, then set window to "no-window".
-    if (init.window === null) {
+    if (window in init) {
       window = 'no-window'
     }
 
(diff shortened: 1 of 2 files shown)


lib/fetch/request.js
```diff
@@ -133,7 +133,7 @@ class Request {
     }
 
     // 11. If init["window"] exists, then set window to "no-window".
-    if (init.window !== undefined) {
+    if (init.window === null) {
```
ronag, line 136:
```suggestion
    if (window in init) {
```

lib/fetch/request.js
```diff
@@ -133,7 +133,7 @@ class Request {
     }
 
     // 11. If init["window"] exists, then set window to "no-window".
-    if (init.window !== undefined) {
+    if (window in init) {
```
ronag, line 136:
```suggestion
    if ('window' in init) {
```

KhafraDev:
I'm going to add a test considering that issue wasn't caught

Issue #2020, opened 2023-03-22
Test failures on Node v19 (and likely v20


mcollina:
We have a few tests failures on Node v19 which we should work to iron out before the v20 release.

RishabhKodes:
Failure logs https://github.com/nodejs/undici/actions/runs/4477206842/jobs/7888223614#step:14:7115

`# Subtest: test/issue-1903.js
    # Subtest: should parse content-disposition consistencely
        1..5
        not ok 1 - Invalid character in header content ["content-disposition"]
          ---
          stack: |
            Server.<anonymous> (test/issue-1903.js:27:9)
          at:
            line: 572
            column: 5
            file: node:_http_outgoing
            function: storeHeader
          type: TypeError
          code: ERR_INVALID_CHAR
          tapCaught: uncaughtException
          test: should parse content-disposition consistencely`

climba03003:
Related to https://github.com/nodejs/node/pull/46528
Seems like either side turns out provide or parsing the header wrongly.

mcollina:
cc @marco-ippolito

marco-ippolito:
With https://github.com/nodejs/node/pull/46528 if there is explicit `content-lenght`, the `content-disposition` header is encoded in latin1 as per https://www.rfc-editor.org/rfc/rfc6266#section-4.3.
There is probably a mismatching in the implementation, I'll have a look

marco-ippolito:
I dont think this was ever supposed to work:
https://github.com/nodejs/undici/blob/f0271d46cbdc47faebfc28579b54431292e09087/test/issue-1903.js#L26-L30 you are passing a header value with non-ascii characters without encoding it.
https://github.com/nodejs/node/pull/46528 this PR solved the behaviour mismatch when using content length.

KhafraDev:
Even once the header is encoded it still fails. I noticed that undici *always* encodes the content-disposition header to latin1, even if there is no content-length header too, but fixing that still doesn't pass the test either ☹️.

KhafraDev:
I'm mostly sure this is a bug in node.

The test is triggering this condition and throwing an error:
https://github.com/nodejs/node/blob/57caf1356229bd24c6129574863fa414ff6862b2/lib/_http_outgoing.js#L627
where `value` is a Buffer:
```js
buffer = Buffer.from(`66 69 6c 65 6e 61 6d 65 3d 27 fd 72 2e 70 64 66 27`.split(' ').map(v => parseInt(v, 16)))
```

[...]

PR #2049, opened 2023-04-07
fetch: fix leak

ronag:
Fixes: https://github.com/nodejs/undici/issues/1823
Fixes: https://github.com/nodejs/node/issues/46435


ronag:
@mcollina now with a test

mcollina:
CI seems to be failing badly :(.

test/fetch-leak.js
```diff
@@ -0,0 +1,42 @@
...
+    global.gc()
+    const next = process.memoryUsage().heapUsed
+    if (next <= prev) {
+      t.pass()
+    } else if (count++ > 10) {
+      t.fail()
+    } else {
+      prev = next
+    }
+  }, 1e3)
```
ronag, line 40: @mcollina Do you have any idea how to do this better?

ronag:
@mcollina PTAL

mcollina (approved):
lgtm

are the v19 failures related? We need to sort those too.

KhafraDev (approved):
test failures are from #2020

test/fetch-leak.js
stalkerg: it's smart! I think it's better than nothing.

KhafraDev:
Just got this failure on the windows v19 ci:

```
# Subtest: test/fetch/fetch-leak.js
    # Subtest: do not leak
        1..1
        ok 1 - (unnamed test)
    ok 1 - do not leak # time=1071.239ms
    
    not ok 2 - fetch failed
      ---
      at:
        line: 1
        column: 17427
        file: index.js
        function: fetch
      stack: |
        fetch (index.js:1:17427)
      type: TypeError
      cause:
        !error
        name: Error
        message: read ECONNRESET
        stack: |-
          Error: read ECONNRESET
              at TCP.onStreamRead (node:internal/stream_base_commons:217:20)
        errno: -4077
        code: ECONNRESET
        syscall: read
      tapCaught: unhandledRejection
      test: TAP
      source: "'use strict'\r
      
        \r
      
        const Client = require('./lib/client')\n"
      ...
    
    1..2
    # failed 1 of 2 tests
    # time=1112.355ms
not ok 12 - test/fetch/fetch-leak.js # time=1112.355ms
```

test/fetch/fetch-leak.js
```diff
@@ -0,0 +1,43 @@
...
+      .then(attack)
+  })
+
+  let prev = Infinity
+  let count = 0
+  const interval = setInterval(() => {
+    done = true
+    global.gc()
+    const next = process.memoryUsage().heapUsed
+    if (next <= prev) {
```
youssefm, line 34: @ronag maybe i'm missing something, but if `prev` starts at Infinity, won't the first iteration of this interval always pass the test? It doesn't look like this test can ever fail.

PR #2057, opened 2023-04-12
feat: allow overriding hwm

ronag:


mcollina (changes requested):
Needs tests, but +1 on the code

PR #2058, opened 2023-04-12
fix: don't leak socket if client is destroyed while connecting

ronag:


mcollina:
CI is failing :( and this might need a test.

ronag:
@mcollina PTAL

mcollina (approved):
lgtm

types/connector.d.ts
```diff
@@ -21,7 +21,7 @@ declare namespace buildConnector {
     port: string
     servername?: string
     localAddress?: string | null
-    httpSocket?: Socket
+    httpSocket?: Socket,
```
ronag, line 24:
```suggestion
    httpSocket?: Socket
```

PR #2061, opened 2023-04-12
feat: Add H2 support

### LINKED ISSUE #399: HTTP2/3 Support

metcoder95:
Superseeds #1014 
Closes #399 

**Checklist - APIs**

- [x] Support for `Client#Connect`
- [x] Support for `Client#Dispatch`
- [x] Support for `Client#Request`
- [x] Support for `Client#Stream`
- [x] Support for `Client#Pipeline`
- [x] Support for `Client#Destroy`
- [x] Support for `Client#Close`
- [x] Support for H2 settings when creating a stream at the different APIs
- [x] Support for disabling H2 whenever ALPN says is available
- [ ] ~~Internal stream management (max concurrent, and usage of them)~~
- [x] Documentation
- [x] Experimental warning
- [x] Calculate `content-length` before triggering the request

**Checklist - HTTP2**
- [x] `100 continue`
- [x] ~~`SETTINGS` - Client~~
  - Not planned
- [x] ~~`SETTINGS` - Server (triggered by)~~
  - Does not needed, as is automatically applied to the session
- [ ] ~~`GOAWAY` - Client~~
- [x] `GOAWAY` - Server (triggered by)
- [ ] ~~`PING` - Client~~
- [ ] ~~`PING` - Server (triggered by)~~


lib/client.js
```diff
@@ -1474,6 +1515,155 @@ function write (client, request) {
...
+      writeIterable({ client, request, stream, contentLength, expectsPayload })
+    })
+  } else {
+    assert(false)
+  }
+
+  console.log('http/2 request')
+  // TODO(HTTP/2): ref only if current streams count is 0
+  session.ref()
+  // TODO(HTTP/2): The handler expects an array but the native http2 module provides us with an object. What should we do?
```
metcoder95, line 1625: It is safe to assume here that we should just transform the array of headers into objects before triggering the request?

lib/client.js
```diff
@@ -1474,6 +1515,155 @@ function write (client, request) {
...
+  } else if (util.isIterable(body)) {
+    process.nextTick(() => {
+      writeIterable({ client, request, stream, contentLength, expectsPayload })
+    })
+  } else {
+    assert(false)
+  }
+
+  console.log('http/2 request')
+  // TODO(HTTP/2): ref only if current streams count is 0
```
metcoder95, line 1623: Shall we handle the number of streams based on user input?

lib/core/connect.js
```diff
@@ -99,6 +99,8 @@ function buildConnector ({ maxCachedSessions, socketPath, timeout, ...opts }) {
         servername,
         session,
         localAddress,
+        // TODO(HTTP/2): Should we support h2c?
```
metcoder95, line 102: Is this something that we should explore or focus only on `h2`?

metcoder95:
Hi!
I just started the work by rebasing and adjusting the set of changes made previously with the current state of the main, this allowed me to understand in deep the changes related to that enabled the support of the H2 protocol.

I just left a couple of questions on some of the topics, as well I would like to understand more or less what other expectations are outside of the comments and just plain h2 support.
This includes UPGRADE requests with a positive response from the server, and other aspects (I'll create a checklist for readiness and keeping track of the features).

As well, there was a concern from @ronag about better use of the branch from @jasnell on QUIC, but not sure if this is really wanted, but is still something to check.

Let me know your thoughts and I'll come back to this later this week 🙂 

cc: @mcollina

lib/client.js
KhafraDev: alternatively, allow the handler to accept an object, rather than having to transfer to <-> from multiple times?

lib/core/connect.js
KhafraDev: this can be handled in a later pr

ronag:
I'm not a fan of using the node core http2 module. I won't block, but I will discourage its usage.

lib/client.js
metcoder95: I think already does, within the handler, unless I overlooked?

metcoder95:
>I'm not a fan of the node core http2 module. I won't block, but I will discourage its usage. I am a fan of the quic/http3 work though.

I don't have a strong opinion on this one, so I opened it again to refresh my ideas and thoughts about it 🙂 

Is this still valid?
>Maybe would be better to target @jasnell's node http3/quic branch? I'm not too comfortable with node core http2. https://github.com/nodejs/node/pull/38233

ronag (commented):
I think the best way to implement this is to only modify `connect` to set a socket:like object on the client. Any further changes than that seems to me like complicating things. But maybe I'm missing something.

metcoder95:
>I think the best way to implement this is to only modify connect to set a socket:like object on the client. Any further changes than that seems to me like complicating things. But maybe I'm missing something.

Now that you mention it, I think we can point in that direction. Though, the `writeH2` basically reuses the biggest portion of the normal `write` one with some tweaks.
The key differentiators are:
1. The parser (`llhttp` for H1, `HTTP2ClientSession` for H2), for response body and headers, as well as requests ones.
2. Session events adjustments, including some weird behaviors of one stream compared to the other

I'll try to bring support for the bare basics of the H2, and do cleanup either here or in another PR. But so far, I think we can consolidate them into a single set of adjustments to mimic the `socket` shape expected by undici

metcoder95:
**Update**
I'm afraid already reach the most extent that can be done for H2 independently. As suggested by @ronag, I'll start working on an abstraction on top of H2 Node.js module that plays compatibility with Undici interface expectations. 
I might require to also adjust some of the parts that undici already has to cope with the different scenarios, but at least String/Buffer and Streams are fully supported. For instance, the bare basics are there, good enough to start the abstraction.

Please @KhafraDev, @mcollina , if there's feedback at the way the things for the next steps, please let me know 🙂

mkaufmaner:
> **Update** I'm afraid already reach the most extent that can be done for H2 independently. As suggested by @ronag, I'll start working on an abstraction on top of H2 Node.js module that plays compatibility with Undici interface expectations. I might require to also adjust some of the parts that undici already has to cope with the different scenarios, but at least String/Buffer and Streams are fully supported. For instance, the bare basics are there, good enough to start the abstraction.
> 
> Please @KhafraDev, @mcollina , if there's feedback at the way the things for the next steps, please let me know 🙂

@metcoder95 Thank you for your hard work on supporting http/2. Keep it up!

I know the interfaces are completely different for undici and the node http/http2 modules but have you considered leveraging other http2 wrappers as a basis?

For example;
https://github.com/szmarczak/http2-wrapper
https://github.com/grantila/fetch-h2

metcoder95:
Hi @mkaufmaner, thanks for the kind words! 🙇 

The interface "struggles" mainly were how `undici` handles TCP/HTTP connections.
> TL;DR It does not use `node:http` but writes directly against the TCP socket, handling the parsing itself (on top of `llhttp`).

Though, in the end, wasn't that complicated, and found a small way to do it. I'll clean up in the next few days and set the PR ready for review

mkaufmaner:
> Hi @mkaufmaner, thanks for the kind words! 🙇
> 
> The interface "struggles" mainly were how `undici` handles TCP/HTTP connections.
> 
> > TL;DR It does not use `node:http` but writes directly against the TCP socket, handling the parsing itself (on top of `llhttp`).
> 
> Though, in the end, wasn't that complicated, and found a small way to do it. I'll clean up in the next few days and set the PR ready for review

Amazing, really looking forward to seeing the PR when it is ready for review!

I know undici doesn't use `node:http` (the whole reason undici is fast) but guess I haven't really comprehended the intricacies of interfacing with `llhttp`. I'll spend some time myself reading through undici source code to improve my comprehension.

Thanks again!

metcoder95:
> @metcoder95, do you need some help with this? I'm not an undici contributor, but I'd like to use H2 and I can try to contribute with it

Hi @mateus4k, thanks for volunteering for it!
It would actually will be nice, but for this work I’ll be wrapping up the work for initial review. I would say that I won’t cover all the corner cases so quite few things will pop-up out of it 👍

metcoder95:
I think the PR is in a reviewable state. Please let me know your thoughts 🙂 

I tried to make the H2 implementation match Undici expectations of streams/sockets, but the implementation is heavily tight to H1 that making an effort and trying to do so will require several rewrites that might be better to have them done in a separate PR, or maybe not at all.

There are several other pieces of H2 support that might be worth it to chat about it to evaluate whether to add it or not (maybe even create its own issues for others to contribute).

> __Note__: On the testing, it seems something is failing over `async-hooks` on 20 and 19, couldn't narrow them down, but maybe you have a better idea of what can I do to fix it 🙂

lib/api/api-request.js
```diff
@@ -95,7 +95,7 @@ class RequestHandler extends AsyncResource {
 
     this.callback = null
     this.res = body
-
+    // console.log('called', callback !== null)
```
mcollina, line 98: remove this

mcollina (commented):
Good work. This is a completely different implementation to what I would have done. I have a few questions:

1. how many concurrent requests are possible for a given `Client`?
2. could you add a test with a HTTPS+HTTP2 server, showing that the connection uses HTTP2?

mkaufmaner:
> 2. could you add a test with a HTTPS+HTTP2 server, showing that the connection uses HTTP2?

@mcollina Hopefully you wouldn't mind if we used fastify for this since writing a one off ALPN negotiator seems a little redundant.

@metcoder95 if @mcollina doesn't mind using fastify, I have taken care of this and opened a PR into your branch. See https://github.com/metcoder95/undici/pull/34

KhafraDev:
I'm a strong -1 on adding more dependencies.

metcoder95:
>Good work. This is a completely different implementation to what I would have done. I have a >few questions:
>
>- how many concurrent requests are possible for a given Client?
>- could you add a test with a HTTPS+HTTP2 server, showing that the connection uses HTTP2?

Thanks!
Going one by one:
- Haven't made that check yet, so far is not limited and we are under the limits of the `node:http2` module. Let me test that out.
  - This leads me to this [comment](https://github.com/nodejs/undici/pull/2061/files#diff-ca407d959d33ce92a7f7efee354a5ea5e6e61fcd797d32c04205e429a074ed54R1317), how do we handle stream concurrency?
- Not yet, but saw that @mkaufmaner did already some research and showed good results. Based on @KhafraDev comments, I think we can adjust it to use native `node:http(2)` to construct a bare basic server that does the ALPN negotiation

lib/client.js
```diff
@@ -1092,29 +1112,43 @@ async function connect (client) {
...
+        createConnection: () => socket
+      })
+
+      session[kError] = null
+      session[kClient] = client
+      session.on('error', onHttp2SessionError)
+      session.on('close', onSocketClose)
+      session.unref()
+
+      client[kHTTP2Session] = session
```
mkaufmaner, line 1130: I think concurrency should be handled through an LRU cache scoped to the `Client` class. This approach would be very similar to @szmarczak  https://github.com/szmarczak/http2-wrapper library.

### Public Interface
First I think the addition of http2 should expose the `http2.connect` options to whomever is instantiating the `Client` class through an `http2` constructor option. Maybe even possibly omitting the `createConnection` option? 

### Performance
One key aspect of performant http2 is setting the appropriate window sizing. That being said the default `http2` window sizing should be resolved at some point. See https://github.com/nodejs/node/issues/38426

Until the default window sizing issue is resolved the initial window size should be set;
```js
const session = http2.connect(client[kUrl], {
  createConnection: () => socket,
  initialWindowSize: 1024 * 1024 * 32
})
```

### Concurrency Events
If the LRU cache for concurrency approach is taken, the `session` events for `remoteSettings` and `goaway` need to be handled.

### Remote Settings
Once the session receives a `remoteSettings` settings event, the session should set a`Client` class `kMaxConcurrentStreams` limit and set the local window size of the session.

```js
// See https://github.com/nodejs/node/issues/38426
if (session.setLocalWindowSize) {
  session.setLocalWindowSize(1024 * 1024 * 4); // 4 MB
}
```

Within the handler for once the session receives a `remoteSettings` settings event, another listener should be created for on `remoteSettings` as the session `session.remoteSettings.maxConcurrentStreams`  needs to update the value for the `kMaxConcurrentStreams`.

### Goaway
If concurrency is implemented through an LRU cache, once a `goaway` event is received from the remote the session should be removed from the LRU cache so it won't be reused.

mkaufmaner (commented):
Added some thoughts about concurrency and performance.

lib/client.js
mkaufmaner: Ah, I see, the client instances are maintained in the `Pool` class. Thus the remote settings and http2 connection state/settings need to be bubbled up.

lib/client.js
```diff
@@ -1489,9 +1530,235 @@ function write (client, request) {
...
+
+  if (request.aborted) {
+    return false
+  }
+
+  // TODO: Content length needs to be calculated at this stage
+  const headers = { ...request[kRawRequestHeaders] }
+  headers[':authority'] = host || client[kHost]
+  headers[':path'] = path
+  headers[':method'] = method
```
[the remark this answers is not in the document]
metcoder95: Good catch 👍

lib/client.js
metcoder95: Thanks for all the feedback, incredibly insightful 🙇 

Going one by one:

>First I think the addition of http2 should expose the http2.connect options to whomever is instantiating the Client class through an http2 constructor option. 

I would love to hear @mcollina, @ronag, and @KhafraDev thoughts to see which options would be ok to move to the public interface, and based on that expose the one required (at least the bare minium). It is also worth it to say that Undici already exposes configs required to manipulate the `Socket` connection

> Maybe even possibly omitting the createConnection option?

Not so sure about this one. By the nature of Undici, the initial connection is made through a socket assuming an H1 connection. If then, the server negotiates and serves a H2, is when this feature comes into play.


>One key aspect of performant http2 is setting the appropriate window sizing. That being said the default http2 window sizing should be resolved at some point. See https://github.com/nodejs/node/issues/38426

Good catch, seems is already approved. I'll update it then 👍 


**For parallel streams**
Though I agree with your points, I didn't start yet with the parallel streams supports, even tho is something that should be treated with care as you suggest.
But this requires also come with a good default, in combination with an API to update several settings accordingly for concurrency.
I would like to hear thoughts from the team, in the meanwhile I'll implement the smaller pieces already pointed out and kick-off benchmarks as well 😄

mkaufmaner:
@metcoder95 @KhafraDev @mcollina I created some simple benchmarks against the http/2 implementation and opened a PR https://github.com/metcoder95/undici/pull/35 into the http2 branch.

The http/2 benchmark results are **_promising_**! However, there are certainly some bugs that need to be worked through as the benchmark only seems to work sporadically.

Initial Results;
```
│ Tests             │ Samples │          Result │ Tolerance │ Difference with slowest │
|───────────────────|─────────|─────────────────|───────────|─────────────────────────|
│ http2 - request   │       1 │  438.80 req/sec │  ± 0.00 % │                       - │
|───────────────────|─────────|─────────────────|───────────|─────────────────────────|
│ undici - fetch    │       1 │  486.82 req/sec │  ± 0.00 % │               + 10.94 % │
|───────────────────|─────────|─────────────────|───────────|─────────────────────────|
│ undici - pipeline │       1 │ 1114.42 req/sec │  ± 0.00 % │              + 153.97 % │
|───────────────────|─────────|─────────────────|───────────|─────────────────────────|
│ undici - stream   │       1 │ 1319.30 req/sec │  ± 0.00 % │              + 200.66 % │
|───────────────────|─────────|─────────────────|───────────|─────────────────────────|
│ undici - request  │       1 │ 1482.31 req/sec │  ± 0.00 % │              + 237.81 % │
```

metcoder95:
Nice work @mkaufmaner 🚀 
Thanks for the support on the benchmarks. I have a rough idea of what might be missing. I'll go through them later on 👍

metcoder95:
I'm having some doubts about how to support the `Dispatcher#connect`. As the concept of `CONNECT` is not that similar to H1, the concept of session mimics when an initial `CONNECT` has been done. After the session has been established, the next is to start creating streams to send/receive data over the wire.

My biggest doubt is whether to return the `socket` as is after the `session` is established (just as is now, triggering the `onUpgrade` for propagating the rest to the end-user), or trigger a request that ends with a stream to the path specified. Once the `stream` is ready, we forward the `ClientH2Stream` instead of the `socket` through the `onUpgrade` handler to continue with the process.

In my head, it seems that the second option fits better, but wanted to doube-check if my assumptions were correct or there is something better to do 🤔

mcollina:
> Once the stream is ready, we forward the ClientH2Stream instead of the socket through the onUpgrade handler to continue with the process.

This seems more natural, yes.

mcollina:
If I understand this correctly, by default it would use http2 whenever the ALPN says it's available. I think we need an option in the Agent/Client to tell "please do not use HTTP/2".

We should also do some rough benchmark to see if this is worse/better than plain text http.

metcoder95:
>If I understand this correctly, by default it would use http2 whenever the ALPN says it's available. I think we need an option in the Agent/Client to tell "please do not use HTTP/2".

Added to the checklist 👍 

>We should also do some rough benchmark to see if this is worse/better than plain text http.

The initial numbers showed good results but were incomplete due to some APIs missing support with H2. Let me finish the APIs support and bring some numbers. I'll use https://github.com/metcoder95/undici/pull/35 as base

lib/core/connect.js
```diff
@@ -99,6 +99,8 @@ function buildConnector ({ maxCachedSessions, socketPath, timeout, ...opts }) {
         servername,
         session,
         localAddress,
+        // TODO(HTTP/2): Add support for h2c
+        ALPNProtocols: allowH2 ? ['http/1.1', 'h2'] : ['http/1.1'],
```
metcoder95, line 103: Whenever the flag is passed, the `h2` is advertised as supported, though the  `h1` has higher priority.

cc: @mcollina

ronag:
I'm swamped at the moment. From a quick glance this looks promising. But will need some significant time for review.

ronag (approved):
LGTM. Needs to be documented as experimental.

metcoder95:
Sure thing. I'll close up the stream management first, and continue with documentation. The rest of the features can be implemented in separated PRs

lib/core/request.js
```diff
@@ -150,19 +156,27 @@ class Request {
...
     if (Array.isArray(headers)) {
       if (headers.length % 2 !== 0) {
         throw new InvalidArgumentError('headers array must be even')
       }
       for (let i = 0; i < headers.length; i += 2) {
-        processHeader(this, headers[i], headers[i + 1])
+        const headerKey = headers[i]
+        const headerValue = headers[i + 1]
+        processHeader(this, headerKey, headerValue)
+        this[kRawRequestHeaders] = {
```
KhafraDev, line 169: can you explain this part to me? This seems to only add the last header pair whereas below all of the headers are stored

metcoder95: Hmm, let me check something as I believe it will have problems with the headers as an array.

But in theory, what it should do is:

1. If the `headers` are passed as an array, it will iterate as usual, process them, and store them as an `object`. This is due to how `node:http2` expects the request headers.
2. If headers are passed as `object`, we allow the flow to continue as is, and stores the headers as plain object within the symbol.

Let me test the headers as an array, and harden the headers to make a copy of them instead of attaching the reference

lib/api/api-request.js
```diff
@@ -95,7 +95,6 @@ class RequestHandler extends AsyncResource {
 
     this.callback = null
     this.res = body
-
```
ronag, line 98: unnecessary

lib/client.js
```diff
@@ -1273,12 +1361,14 @@ function _resume (client, sync) {
...
-    if (!socket) {
+    if (!socket && !client[kHTTP2Session]) {
       connect(client)
       return
     }
 
-    if (socket.destroyed || socket[kWriting] || socket[kReset] || socket[kBlocking]) {
+    if ((socket.destroyed || socket[kWriting] || socket[kReset] || socket[kBlocking]) ||
+      (client[kHTTP2Session] && client[kHTTP2Session].destroyed)) {
+      // TODO(HTTP/2): what if exceeds max concurrent streams or can't accept new
```
ronag, line 1371: if possible I'd like to avoid special cases here... socket `socket` be `client[kHTTP2Session]` in case of http2?

lib/core/request.js
ronag: This will affect non h2 negatively.

ronag: I'd prefer not to change the code here and instead do whaterver is need for h2 as close to h2 as possible

KhafraDev: yes, maybe my question wasn't worded good... right now the code is equal to:

```js
 // process headers
}

this[kRawRequestHeaders] = {
  [headers.at(-2)]: headers.at(-1)
}
```

while below, kRawRequestHeaders is set to the entire headers object. I don't understand the purpose of doing that, could you explain it to me?

lib/client.js
metcoder95: It can but it touches several places and might lead to more changes, especially as we still handle the socket lifetime (and in userland, a custom socket factory can be provided).
Although destroying the socket will eventually lead to destroying the session as well, so it might not be needed after all 🤔

Let me test it out

lib/core/request.js
metcoder95: >I'd prefer not to change the code here and instead do whaterver is need for h2 as close to h2 as possible

The biggest issue comes with the processing of the header, as it gets stuck early in the request lifecycle, but maybe there's something I'm missing to make it closer to H2 implementation 👍 


>while below, kRawRequestHeaders is set to the entire headers object. I don't understand the purpose of doing that, could you explain it to me?

I'm seeing more clearly what are you suggestion, but let me know if I'm still beyond what you are suggesting

The change is not that far from the current implementation.
Users can provide the headers in a form of an array with a pair length.
When iterating over the pair length, we move forward two positions, and we take the `n` position as the header name, and `n+1` as the value.
The change I made here was merely to just store them as variables so they can be used to construct the headers object.

In the implementation below, the user passed the headers. We just iterate over its keys to validate each header and its value, 

If everything passes correctly, we just attach them directly into the `kRawRequestHeaders`.

Although I noticed an optimization here as the `processHeader` is already of the `this` context 😅

Nonetheless, I'll explore how to couple that part directly into H2 implementation

lib/core/util.js
```diff
@@ -199,7 +199,13 @@ function destroy (stream, err) {
       // See: https://github.com/nodejs/node/pull/38505/files
       stream.socket = null
     }
-    stream.destroy(err)
+
+    // HTTP/2 - It causes to throw uncaught exceptions due to the
+    // error passed down the socket.destroy
+    const isHTTP2 = stream.alpnProtocol === 'h2' &&
+      Object.getPrototypeOf(err).constructor === ClientDestroyedError
```
ronag, line 206: Are we missing an `error` handler?

metcoder95: Fixed, good catch. When creating the H2 session, the socket didn't have its event handlers

lib/client.js
```diff
@@ -298,7 +333,9 @@ class Client extends DispatcherBase {
   [kDispatch] (opts, handler) {
     const origin = opts.origin || this[kUrl].origin
 
-    const request = new Request(origin, opts, handler)
+    let request
+    if (this[kHTTP2Session]) request = Request[kHTTP2BuildRequest](origin, opts, handler)
```
metcoder95, line 337: I've tried to handle the headers fact in a way that can be decoupled from the original H1 implementation, but the implementation is quite tight and the refactor will require to touch several pieces.

In a best effort, will assume at the first instance that a request is made through H1 building the request as usual.
As at this point, we don't know yet the ALPN protocol, we'll convert to H2 headers only once we already negotiated this with the server, where we'll convert them from string to object one time only.
After the initial check, the rest of the Request through the Client will go as H2 requests, using the `headers` as an object.

Let me know if I missed something or if there are improvements that can be done

cc: @ronag @mcollina

mcollina: sgtm

metcoder95:
@ronag @KhafraDev @mcollina 

I think we can wrap up the implementation at this stage.
Tests with `fetch` are missing, I'll add a few to have some baseline for the future.

From now on, I'll move forward on fixing the tests that are failing, adding some for `fetch`, and creating the docs around it for its usage.

ronag (commented):
LGTM

Hopefully will have time to do a more in depth review soon.

mcollina:
code is good so far! Can't wait for the final work

KhafraDev:
> Tests with fetch are missing, I'll add a few to have some baseline for the future.

there should be a bunch of WPTs that use http2. Might be a little difficult setting up the server but could save a lot of time writing tests. Also LGTM so far, this is really great.

lib/api/api-connect.js
```diff
@@ -50,7 +50,13 @@ class ConnectHandler extends AsyncResource {
     removeSignal(this)
 
     this.callback = null
-    const headers = this.responseHeaders === 'raw' ? util.parseRawHeaders(rawHeaders) : util.parseHeaders(rawHeaders)
+
+    let headers
+    // Indicates is an HTTP2Session
+    if (headers != null) {
```
ronag, line 56: Isn't this always true?

metcoder95: Yes. For two reasons; there's no `upgrade` on H2, and when using `Client#connect`, the intention is to return the H2 Stream, as soon as it has an identifier assigned.

As we do not wait for the response, but rather just there's an `identifier` assigned, we do not read the `headers` at this point but rather is up to the implementation that has control of the stream to decide what to do with it.

metcoder95:
> > Tests with fetch are missing, I'll add a few to have some baseline for the future.
> 
> there should be a bunch of WPTs that use http2. Might be a little difficult setting up the server but could save a lot of time writing tests. Also LGTM so far, this is really great.

We can aim to have just a minimum here to ensure it works, and we explore together the WPYs with H2 in another PR 👍

lib/api/api-connect.js
ronag: What's the point of the condition then?

metcoder95: Handle the requests for `http/1` as well

KhafraDev: the issue is that header is always undefined there because its not being assigned to anything

metcoder95: Ah... My bad, sorry, I read it wrong 😓 
It is `rawHeaders` the one that should be checked.

lib/client.js
```diff
@@ -241,6 +266,15 @@ class Client extends DispatcherBase {
     this[kClosedResolve] = null
     this[kMaxResponseSize] = maxResponseSize > -1 ? maxResponseSize : -1
 
+    // HTTP/2
+    this[kHTTP2Session] = null
+    this[kHTTP2SessionState] = {
```
ronag, line 271:
```suggestion
    this[kHTTP2SessionState] = !allowH2 ? null : {
```

lib/client.js
```diff
@@ -298,7 +332,9 @@ class Client extends DispatcherBase {
   [kDispatch] (opts, handler) {
     const origin = opts.origin || this[kUrl].origin
 
-    const request = new Request(origin, opts, handler)
+    let request
+    if (this[kHTTP2Session]) request = Request[kHTTP2BuildRequest](origin, opts, handler)
+    else request = new Request(origin, opts, handler)
```
ronag, line 337:
```suggestion
    else request = new Request[kHTTP1BuildRequest](origin, opts, handler)
```

lib/client.js
```diff
@@ -952,7 +1060,7 @@ function onSocketError (err) {
 
   // On Mac OS, we get an ECONNRESET even if there is a full body to be forwarded
   // to the user.
-  if (err.code === 'ECONNRESET' && parser.statusCode && !parser.shouldKeepAlive) {
+  if (err.code === 'ECONNRESET' && parser && parser.statusCode && !parser.shouldKeepAlive) {
```
ronag, line 1063:
```suggestion
  if (err.code === 'ECONNRESET' && parser?.statusCode && !parser.shouldKeepAlive) {
```

lib/client.js
```diff
@@ -1088,28 +1198,50 @@ async function connect (client) {
     })
 
     if (client.destroyed) {
-      util.destroy(socket.on('error', () => {}), new ClientDestroyedError())
+      util.destroy(socket.on('error', () => { }), new ClientDestroyedError())
```
ronag, line 1201:
```suggestion
      util.destroy(socket.on('error', () => {}), new ClientDestroyedError())
```

lib/client.js
```diff
@@ -1208,7 +1340,7 @@ function _resume (client, sync) {
 
     const socket = client[kSocket]
 
-    if (socket && !socket.destroyed) {
+    if (socket && !socket.destroyed && socket.alpnProtocol !== 'h2') {
```
ronag, line 1343: Don't we need to ref/unref the h2 sockets? Is it harmful if we do ref/unref?

ronag (commented):
Took a better look now. My main concern is the number of ways we check for H2:

- allowH2
- client[kHTTP2Session]
- socket.alpnProtocol === 'h2'

lib/client.js
metcoder95: Do you mean build a new one, or?

lib/client.js
metcoder95: We are using the same socket for the `h2` connection, and then we can follow the same pattern as per `h1`.
It should not be a problem, as no more clients running is the criteria; no more traffic is flowing although the streams are still open.
As we are not supporting `push` nor `ping` or similar, this should not be a problem, once we decide to introduce it we we'll need to check this once more.

metcoder95:
> Took a better look now. My main concern is the number of ways we check for H2:
> 
> * allowH2
> * client[kHTTP2Session]
> * socket.alpnProtocol === 'h2'

Wanted to keep it quiet "_simple_", and use what we have at hand at each step to decide wether or not is an `h2` connection. If you agree, I can better use a flag within `client` to decide wether or not is an `h2` connection. e.g.
```js
client[kHTTPVersion]: '<h1 | h2>'
```

And use it across all the steps, wdyt?

mcollina (approved):
lgtm

lib/client.js
```diff
@@ -1092,24 +1200,47 @@ async function connect (client) {
...
+      socket[kReset] = false
+      socket[kBlocking] = false
+      socket[kParser] = new Parser(client, socket, llhttpInstance)
+      socket[kCounter] = 0
+      socket[kMaxRequests] = client[kMaxRequests]
+    }
+
     socket[kClient] = client
-    socket[kCounter] = 0
-    socket[kMaxRequests] = client[kMaxRequests]
```
ronag, line 1112: Why removed?

metcoder95: They were not removed, just moved a level above. AFAIW it is not applicable within H2, not beyond at a socket level. But it is quite used for handling the requests within the `Client`. Let me put them back one level outside

package.json
```diff
@@ -133,4 +133,4 @@
   "dependencies": {
     "busboy": "^1.6.0"
   }
-}
+}
```
ronag, line 136: unecessary change

KhafraDev:
@metcoder95 I have a few questions:
- will this work with websockets?
- could you add a test that uses fetch (just to make sure it works)

metcoder95:
Hey, replying to your questions 👋 

>will this work with websockets?

Sadly, it won't. Thanks for pointing that out, we'll need to document that, if WebSockets is desired, `h2` won't be supported.
This is due to the fact that the semantics to establish a `ws` connection differs substantially from `h1`, in the sense that `h2` does not have the concept of `upgrade`, and that the initial `CONNECT` frame.
We can explore add this further once the feature is more stable 👍 

>could you add a test that uses fetch (just to make sure it works)

Do you refer to `fetch` over `h2`?
I already added a few [here](https://github.com/nodejs/undici/pull/2061/files#diff-15bb8133b2c65012b3da0ed17e259f6df029f2d42a6855887baa3ed90dc73899R892). Do you want me to add a more complex one?

KhafraDev:
the fetch tests need to be moved to `test/fetch` otherwise they'll run on v14, which will fail

metcoder95:
> the fetch tests need to be moved to `test/fetch` otherwise they'll run on v14, which will fail

Has sense, let me move them then 👍

lib/client.js
```diff
@@ -298,7 +344,9 @@ class Client extends DispatcherBase {
   [kDispatch] (opts, handler) {
     const origin = opts.origin || this[kUrl].origin
 
-    const request = new Request(origin, opts, handler)
+    let request
+    if (this[kHTTPConnVersion] === 'h2') request = Request[kHTTP2BuildRequest](origin, opts, handler)
+    else request = Request[kHTTP1BuildRequest](origin, opts, handler)
```
ronag, line 349: nit. ternary?

metcoder95:
I'm seeing that in all the testing failed, are due to `fetch` testing failures. Not sure if they are really related to H2 as H2 is disabled by default, and only enabled for `fetch/http2.js` testing.

Any suggestion on how to proceed?

cc @KhafraDev @ronag

01101sam:
> I'm seeing that in all the testing failed, are due to `fetch` testing failures. Not sure if they are really related to H2 as H2 is disabled by default, and only enabled for `fetch/http2.js` testing.
> 
> Any suggestion on how to proceed?
> 
> cc @KhafraDev @ronag

Upon checking the log, I found the testing failing was related to `test/fetch/encoding.js`

<deatils>
<summary>The error log</summary>

```
2023-08-23T08:43:08.9645777Z # Subtest: test/fetch/encoding.js
2023-08-23T08:43:08.9646791Z     # Subtest: content-encoding header is case-iNsENsITIve
2023-08-23T08:43:08.9647252Z         not ok 1 - terminated
2023-08-23T08:43:08.9647587Z           ---
2023-08-23T08:43:08.9647876Z           stack: |
2023-08-23T08:43:08.9648267Z             Fetch.onAborted (lib/fetch/index.js:678:115)
2023-08-23T08:43:08.9648700Z             Fetch.terminate (lib/fetch/index.js:7:358)
2023-08-23T08:43:08.9649222Z             Fetch.fetchParams.controller.resume (lib/fetch/index.js:665:113)
2023-08-23T08:43:08.9649646Z           at:
2023-08-23T08:43:08.9649930Z             line: 678
2023-08-23T08:43:08.9650245Z             column: 115
2023-08-23T08:43:08.9650569Z             file: lib/fetch/index.js
2023-08-23T08:43:08.9650929Z             function: Fetch.onAborted
2023-08-23T08:43:08.9651264Z           type: TypeError
2023-08-23T08:43:08.9651647Z           tapCaught: returnedPromiseRejection
2023-08-23T08:43:08.9652422Z           test: content-encoding header is case-iNsENsITIve
2023-08-23T08:43:08.9652812Z           source: |2
2023-08-23T08:43:08.9653297Z                   response = filterResponse(response, 'basic')
2023-08-23T08:43:08.9653931Z                 } else if (request.responseTainting === 'cors') {
2023-08-23T08:43:08.9654459Z                   response = filterResponse(response, 'cors')
2023-08-23T08:43:08.9655001Z                 } else if (request.responseTainting === 'opaque') {
2023-08-23T08:43:08.9655362Z           ...
2023-08-23T08:43:08.9663190Z         
2023-08-23T08:43:08.9663648Z         1..1
2023-08-23T08:43:08.9664015Z         # failed 1 test
2023-08-23T08:43:08.9664787Z     not ok 1 - content-encoding header is case-iNsENsITIve # time=133.024ms
2023-08-23T08:43:08.9665169Z     
2023-08-23T08:43:08.9665792Z     # Subtest: response decompression according to content-encoding should be handled in a correct order
2023-08-23T08:43:08.9666300Z         not ok 1 - terminated
2023-08-23T08:43:08.9666626Z           ---
2023-08-23T08:43:08.9666890Z           stack: |
2023-08-23T08:43:08.9667272Z             Fetch.onAborted (lib/fetch/index.js:678:115)
2023-08-23T08:43:08.9667697Z             Fetch.terminate (lib/fetch/index.js:7:358)
2023-08-23T08:43:08.9668212Z             Fetch.fetchParams.controller.resume (lib/fetch/index.js:665:113)
2023-08-23T08:43:08.9668584Z           at:
2023-08-23T08:43:08.9668863Z             line: 678
2023-08-23T08:43:08.9669146Z             column: 115
2023-08-23T08:43:08.9669465Z             file: lib/fetch/index.js
2023-08-23T08:43:08.9670154Z             function: Fetch.onAborted
2023-08-23T08:43:08.9670482Z           type: TypeError
2023-08-23T08:43:08.9670871Z           tapCaught: returnedPromiseRejection
2023-08-23T08:43:08.9671550Z           test: response decompression according to content-encoding should be handled in
2023-08-23T08:43:08.9671976Z             a correct order
2023-08-23T08:43:08.9672276Z           source: |2
2023-08-23T08:43:08.9672745Z                   response = filterResponse(response, 'basic')
2023-08-23T08:43:08.9673295Z                 } else if (request.responseTainting === 'cors') {
2023-08-23T08:43:08.9673826Z                   response = filterResponse(response, 'cors')
2023-08-23T08:43:08.9674374Z                 } else if (request.responseTainting === 'opaque') {
2023-08-23T08:43:08.9674723Z           ...
2023-08-23T08:43:08.9674994Z         
2023-08-23T08:43:08.9675273Z         1..1
2023-08-23T08:43:08.9675570Z         # failed 1 test
2023-08-23T08:43:08.9676236Z     not ok 2 - response decompression according to content-encoding should be handled in a correct order # time=14.161ms
2023-08-23T08:43:08.9676708Z     
2023-08-23T08:43:08.9676967Z     1..2
2023-08-23T08:43:08.9677247Z     # failed 2 of 2 tests
2023-08-23T08:43:08.9677526Z     # time=5053.951ms
2023-08-23T08:43:08.9678202Z not ok 11 - test/fetch/encoding.js # time=5053.951ms
```

</deatils>

KhafraDev:
```
test/fetch/encoding.js 1> {
test/fetch/encoding.js 1>   reason: Error: incorrect header check
test/fetch/encoding.js 1>       at Zlib.zlibOnError [as onerror] (node:zlib:189:17) {
test/fetch/encoding.js 1>     errno: -3,
test/fetch/encoding.js 1>     code: 'Z_DATA_ERROR'
test/fetch/encoding.js 1>   }
test/fetch/encoding.js 1> }
```

this doesn't seem like an issue with fetch

metcoder95:
>Upon checking the log, I found the testing failing was related to test/fetch/encoding.js

Yeah, I noticed the same. I was wondering if there was an already known issue or if some other PR was applying a similar fix.
Locally, it is reproducible, so I'll try to debug it a bit

test/fetch/encoding.js
```diff
@@ -17,10 +17,10 @@ test('content-encoding header is case-iNsENsITIve', async (t) => {
     res.setHeader('Content-Encoding', contentCodings)
     res.setHeader('Content-Type', 'text/plain')
 
-    gzip.pipe(brotli).pipe(res)
+    brotli.pipe(gzip).pipe(res)
```
metcoder95, line 20: The order of the encoding was wrong

metcoder95:
I think the v20 ones are expected, so the test should be fixed now (🤞)

test/fetch/encoding.js
KhafraDev: this partially undoes c415fbbb59e2b898c5db6a681265cf3da865d02c, I'm not sure it's right. It should be gzipped first and then brotli'd?

metcoder95: Accordingly to RFC-9110:
>If one or more encodings have been applied to a representation, the sender that applied the encodings MUST generate a Content-Encoding header field that lists the content codings in the order in which they were used.

At least, in my interpretation, the leftmost is the last applied encoding.
Using `deflate, gzip` as an example, to encode it we first
1. Encode it to `gzip`
2. Encode it to `deflate`

Then to decode it we should do it the other way around, first `gzip`, later `deflate`.
The tests where highlighting this as the header for `gzip` has often wrong when `fetch` was trying to decompress it

mcollina:
(side note: we need to fix those broken tests in v20)

KhafraDev:
some of the v20 tests are not expected:
```
     # Subtest: Request should fail if allowH2 is false and server advertises h2 only
        not ok 1 - Client network socket disconnected before secure TLS connection was established
          ---
          at:
            line: 720
            column: 14
            file: node:internal/errors
            function: connResetException
          code: ECONNRESET
          path: null
          host: localhost
          port: "34227"
          localAddress: null
          tapCaught: returnedPromiseRejection
          test: Request should fail if allowH2 is false and server advertises h2 only
          ...
        
        1..1
        # failed 1 test
    not ok 6 - Request should fail if allowH2 is false and server advertises h2 only # time=19.33ms
```

metcoder95:
Hmm, I missed that. I'll go through them 👍

test/http2.js
```diff
@@ -1,32 +1,949 @@
...
+    t.equal(
+      error.message,
+      'maxConcurrentStreams must be a possitive integer, greater than 0'
+    )
+  }
+})
+
+test(
+  'Request should fail if allowH2 is false and server advertises h1 only',
+  { skip: isGreaterThanv20 },
```
metcoder95, line 217: It seems the behaviour changed between v18 and v20. On v18 we receive a `403` while on v20 now it throws that the initial handshake (the ALPN negotiation step) failed before upgrading to TLS

test/fetch/encoding.js
hazycora: I believe it's the rightmost which is the last applied encoding, it's in the order the sender compressed it not the order used to decompress it by the client.

> MUST generate a Content-Encoding header field that lists the content codings in the order in which they *were* used

"were" being past tense, this is not in reference to the client who hasn't yet received the content but the server which would be generating the header. And I assume this would be left to right. So your example, ``deflate, gzip`` should be, for the server,
1. Encode it to ``deflate``
2. Encode it to ``gzip``

I've looked around a bit and besides the RFC which is maybe a little unclear, the phrasing everywhere I can see that this is explained (MDN, various tutorials) it's said that it's meant to be the order of the encoding, not the intended decoding order.

KhafraDev:
```
           compare: ===
          at:
            line: 126
            column: 5
            file: test/fetch/http2.js
            type: Test
          stack: |
            Test.<anonymous> (test/fetch/http2.js:126:5)
          source: "  t.equal(responseBody, 'hello h2!')\r
          
            \  t.equal(Buffer.concat(requestChunks).toString('utf-8'), expectedBody)\r
          
            ----^
          
            })\r
          
            test('Should handle h2 request with body (Blob)', { skip: !Blob }, async t
            => {\n"
          ...
        
        # failed 1 of 8 tests
    not ok 2 - [Fetch] Should handle h2 request with body (stream) # time=47.939ms
```

https://github.com/nodejs/undici/actions/runs/6022709926/job/16342298970?pr=2061#step:14:13589

metcoder95:
> ```
>            compare: ===
>           at:
>             line: 126
>             column: 5
>             file: test/fetch/http2.js
>             type: Test
>           stack: |
>             Test.<anonymous> (test/fetch/http2.js:126:5)
>           source: "  t.equal(responseBody, 'hello h2!')\r
>           
>             \  t.equal(Buffer.concat(requestChunks).toString('utf-8'), expectedBody)\r
>           
>             ----^
>           
>             })\r
>           
>             test('Should handle h2 request with body (Blob)', { skip: !Blob }, async t
>             => {\n"
>           ...
>         
>         # failed 1 of 8 tests
>     not ok 2 - [Fetch] Should handle h2 request with body (stream) # time=47.939ms
> ```
> 
> [nodejs/undici/actions/runs/6022709926/job/16342298970?pr=2061#step:14:13589](https://github.com/nodejs/undici/actions/runs/6022709926/job/16342298970?pr=2061#step:14:13589)

Is this still happening? I'm seeing the test green.
Although I faced it several times only when running all the tests in the band, while running the test individually it passed; even through several runs. I tried to simplify its consumption so is less "susceptible" to race conditions but stears just appear sporadically.

KhafraDev:
a few tests are still failing, on the most recent ubuntu v20 ci run

metcoder95:
I was checking it out and for some unknown reason this test on `test/fetch/blob-uri.js` seems failing:
```js
  t.test('a normal fetch request works', async (t) => {
    const res = await fetch(objectURL)

    t.plan(3)
    t.equal(blobContents, await res.text())
    t.equal(blob.type, res.headers.get('Content-Type'))
    t.equal(`${blob.size}`, res.headers.get('Content-Length'))
  })
```

I've tried to narrow down the possible cause of the issue but so far couldn't fully determine the problem. While debugging the issue directed me to `readAllBytes` in `lib/fetch/util.js`.
This one uses a `while` loop to iterate over the blob stream until fully consumed; although in two iterations should be fully consumed given its size, it seems that something cuts the processing in the middle and does not allow it to fully finish the last iteration.

```js
async function readAllBytes (reader) {
  const bytes = []
  let byteLength = 0

  while (true) { // Enters the loop, it gets called once
    const { done, value: chunk } = await reader.read() // This is not read at the second iteration

    if (done) { // This is never true as its processing gets cut at the end of the first iteration
      // 1. Call successSteps with bytes.
      return Buffer.concat(bytes, byteLength)
    }

    // 1. If chunk is not a Uint8Array object, call failureSteps
    //    with a TypeError and abort these steps.
    if (!isUint8Array(chunk)) {
      throw new TypeError('Received non-Uint8Array chunk')
    }

    // 2. Append the bytes represented by chunk to bytes.
    bytes.push(chunk)
    byteLength += chunk.length

    // 3. Read-loop given reader, bytes, successSteps, and failureSteps.
  }
}
```

I'm not sure at what extent this has to do with `H2` as only the `header` parsing was altered 🤔

KhafraDev:
it's a node bug, fixed by https://github.com/nodejs/node/pull/48935 which hasn't landed yet

metcoder95:
Nice, thanks! Wasn't aware of that. Then, waiting for its approval to land 👍

mcollina:
@KhafraDev could we skip that test for now? It won't work in Node 18 for a while anyway.

KhafraDev:
It seems to work in node 18, I guess whatever commit broke it never got back ported. I don't love the idea of skipping it completely because it's a very basic test that should work. Plus we would probably also have to disable a bunch of WPTs to get v20 to pass... is there something keeping v20.6.0 from releasing? If there is it's fine to skip them since it'd be unfair to keep this pr in review for weeks while waiting for an unrelated fix

mcollina:
Let's wait then ;)

metcoder95:
Perfect, SGTM 👍

01101sam:
tested on prod, works very great 👍🏻

01101sam (approved):

KhafraDev:
@metcoder95 can u push an empty commit so we can rerun the ci? Blob issue should be fixed now that v20.6.0 landed

Issue #2062, opened 2023-04-12
Error: Must use import to load ES Module for undici/lib/llhttp/llhttp.wasm


jeromemacias:
## Bug Description

Error when loading undici into jest unit tests

## Logs

Tests run using:
`NODE_OPTIONS=--experimental-vm-modules npx jest --run-in-band --verbose --ci`

Error log:
```
/home/runner/work/my-project/my-project/node_modules/jest-runtime/build/index.js:841
      const error = new Error(
                    ^

Error: Must use import to load ES Module: /home/runner/work/my-project/my-project/node_modules/undici/lib/llhttp/llhttp.wasm
    at Runtime.requireModule (/home/runner/work/my-project/my-project/node_modules/jest-runtime/build/index.js:841:21)
    at Runtime.requireModuleOrMock (/home/runner/work/my-project/my-project/node_modules/jest-runtime/build/index.js:1039:21)
    at lazyllhttp (/home/runner/work/my-project/my-project/node_modules/undici/lib/client.js:362:55)
    at Object.<anonymous> (/home/runner/work/my-project/my-project/node_modules/undici/lib/client.js:424:21)
    at Runtime._execModule (/home/runner/work/my-project/my-project/node_modules/jest-runtime/build/index.js:1429:24)
    at Runtime._loadModule (/home/runner/work/my-project/my-project/node_modules/jest-runtime/build/index.js:1013:12)
    at Runtime.requireModule (/home/runner/work/my-project/my-project/node_modules/jest-runtime/build/index.js:873:12)
    at Runtime.requireModuleOrMock (/home/runner/work/my-project/my-project/node_modules/jest-runtime/build/index.js:1039:21)
    at Object.<anonymous> (/home/runner/work/my-project/my-project/node_modules/undici/index.js:3:16)
    at Runtime._execModule (/home/runner/work/my-project/my-project/node_modules/jest-runtime/build/index.js:1429:24)
    at Runtime._loadModule (/home/runner/work/my-project/my-project/node_modules/jest-runtime/build/index.js:1013:12)
    at Runtime.requireModule (/home/runner/work/my-project/my-project/node_modules/jest-runtime/build/index.js:873:12)
    at Runtime.requireModuleOrMock (/home/runner/work/my-project/my-project/node_modules/jest-runtime/build/index.js:1039:21)
    at Runtime.loadCjsAsEsm (/home/runner/work/my-project/my-project/node_modules/jest-runtime/build/index.js:717:22)
[...]

Issue #1878, opened 2023-01-23
can't run unit tests with jest when using esm modules, typescript and undici v5.16.0


marinrusu1997:
## Bug Description

When trying to run unit tests with jest for files which use undici an error is thrown.

## Reproducible By

You can reproduce this error by cloning [this repo](https://github.com/marinrusu1997/jest-undici-esm-bug). [README](https://github.com/marinrusu1997/jest-undici-esm-bug/blob/master/README.md) contains steps to reproduce the bug.

## Expected Behavior

Unit tests should complete successfully.

## Logs & Screenshots
This is the error shown by jest:
```
 Must use import to load ES Module: /Users/marin/Development/jest-undici-esm-bug/node_modules/undici/lib/llhttp/llhttp.wasm

      at Runtime.requireModule (node_modules/jest-runtime/build/index.js:817:21)
          at async Promise.all (index 0)
          at async Promise.all (index 0)
```

Also notice that if you lower the undici version, let's say to v5.15.0, a different issue will be thrown
```
/Users/marin/Development/jest-undici-esm-bug/node_modules/undici/lib/client.js:1184
        if (socket[kParser].timeoutType !== TIMEOUT_IDLE) {
                            ^

TypeError: Cannot read properties of undefined (reading 'timeoutType')
    at _resume (/Users/marin/Development/jest-undici-esm-bug/node_modules/undici/lib/client.js:1184:29)
    at resume (/Users/marin/Development/jest-undici-esm-bug/node_modules/undici/lib/client.js:1148:3)
    at connect (/Users/marin/Development/jest-undici-esm-bug/node_modules/undici/lib/client.js:1133:3)
```

## Environment

MacOS Monterey
Node v16.14.2
Undici v5.16.0

KhafraDev:
This is an issue with `jest.config.cjs`:

```diff
diff --git a/jest.config.cjs b/jest.config.cjs
index 3841fe0..cb6e7d5 100644
--- a/jest.config.cjs
+++ b/jest.config.cjs
@@ -8,7 +8,7 @@ module.exports = {
                },
        },
        moduleNameMapper: {
-               '^(\\.{1,2}/.*)\\.js$': '$1',
+               '^(\\.{1,2}/.*)\\.js$': '$1.js',
        },
        testMatch: ["**/test/**/*.spec.ts"]
-};
\ No newline at end of file
+};
diff --git a/test/index.spec.ts b/test/index.spec.ts
[...]

PR #2064, opened 2023-04-12
fix: rename .wasm to -wasm to appease jest

KhafraDev:
Fixes: https://github.com/nodejs/undici/issues/2062
Refs: https://github.com/nodejs/undici/issues/1878#issuecomment-1421484960

ts-jest recommends a really dumb default that drops (rewrites) the `.js` extension, this makes jest think that the files are wasm and subsequently fails to import them.


KhafraDev:
node-fetch test is failing, not because of this PR tho

```js
it('should allow manual redirect handling', function () {
    this.timeout(5000)
    const url = 'https://httpbin.org/status/302'
    const options = {
      redirect: 'manual'
    }
    return fetch(url, options).then(res => {
      expect(res.status).to.equal(302)
      expect(res.url).to.equal(url)
      expect(res.type).to.equal('basic')
      expect(res.headers.get('Location')).to.equal('/redirect/1')
      expect(res.ok).to.be.false
    })
  })
```

ronag (approved):

PR #2074, opened 2023-04-16
DO NOT MERGE: fix: node 19 fail

ronag:


package.json
```diff
@@ -104,7 +104,7 @@
     "ws": "^8.11.0"
   },
   "engines": {
-    "node": ">=14.0"
+    "node": ">=18"
```
mcollina, line 107: are we bumping a major?

ronag: No. Just don't want all the version test to run while testing

PR #2076, opened 2023-04-17
Cache storage

### LINKED ISSUE #2071: implement CacheStorage?
I was wondering if you would be willing to look into implementing the [Cache Storage](https://developer.mozilla.org/en-US/docs/Web/API/CacheStorage) api

it was something i was looking into in the past... but it got abandoned.
it is pretty much tightly coupled with Request/Response classes 

and i would like to use it to avoid having to make extra requests if i have the key (Request) cached locally.

it's also in fact tightly coupled with the `cache` option in Request too. 
it would be neat if this could just work with some default storage and Request / Responses where saved locally on the disc somewhere. 
```js
fetch(url, {
  cache: "no-cache", // *default, no-cache, reload, force-cache, only-if-cached
})
```
manual working with caches works to...
```js
    caches.match(request).then((response) => {
      // caches.match() always resolves
      // but in case of success response will have value
      if (response !== undefined) {
        return response;
      } else {
        return fetch(event.request)
          .then((response) => {
            // response may be used only once
            // we need to save clone to put one copy in cache
            // and serve second one
            let responseClone = response.clone();

            caches.open("v1").then((cache) => {
              cache.put(event.request, responseClone);
            });
            return response;
          })
          .catch(() => caches.match("/gallery/myLittleVader.jpg"));
      }
    })
```
(taken from mdn)

maybe `caches.open('node:undici')` could store the default request / responses from fetch when using the cache option?

KhafraDev:
Closes #2071

Current test status (okay):
```
[service-workers/cache-storage]: Completed: 92, failed: 13, success: 79, expected failures: 13, unexpected failures: 0, skipped: 3
```


KhafraDev:
I think we're passing more tests than Deno now, and I haven't even implemented `CacheStorage.prototype.match` 😆

KhafraDev:
Not everything is finished, but it is usable currently. Once this lands I'll make a followup PR that adds docs and exposes a CacheStorage instance because this PR is already big enough... and I'm sick of w3c specs atm 😆

lib/cache/cache.js
```diff
@@ -25,13 +25,12 @@ const assert = require('assert')
...
-class Cache {
-  /**
-   * The name of the cache
-   * @type {string}
-   */
-  #id
+/**
+ * @type {(o: Cache) => requestResponseList}
+ */
+let getCacheRequestResponseList
```
mcollina: This module global seems very very dangerous and a source of leaks.

KhafraDev: node has a similar pattern in its mimetype parser, https://github.com/nodejs/node/blob/d3b0a2a68b5cb03de1f254cefa46f4c4023b1f89/lib/internal/mime.js#L296

KhafraDev: I could change to a symbol if you want, this is just more private (Object.getOwnPropertySymbols returns an empty array)

mcollina: Node.js usage is not a cache. This looks like an unbounded list.

KhafraDev: The list generally has to be unbounded - every time .open is called a new Cache is created with the requestResponseList associated with the cache name. Basically:

```js
// the two caches are different instances
const c1 = await caches.open('v1')
const c2 = await caches.open('v1')

assert.notEqual(c1, c2) // same name, different instances
assert.deepStrictEqual(await c1.keys(), await c2.keys()) // same cache names
assert.deepStrictEqual(await c1.match('/req'), await c2.match('/req'))
```

KhafraDev: I guess we can store the Cache instances themselves and then access the list inside the constructor...

mcollina: So you did remove the global cache, am I correct?

mcollina: where in the code is this logic? I can't find it.

I don't think we can support this feature in Node.js, as this would result in a massive memory leak.

KhafraDev: https://github.com/nodejs/undici/pull/2076/files#diff-138705835e0968882bd15c371854df97c448d16e5ef1352ede0f57634df917caR57-R74

In a future iteration we'd use a file based cache in place of this, right now it's just trying to match the spec.

mcollina: That code does not have a global list of caches or response, right?

KhafraDev: no it doesn't

mcollina (approved):
lgtm

jimmywarting:
Wow, what a speedy delivery from a feature request to something usable! 🚀

KhafraDev:
I like working on spec things :)

jimmywarting:
Me too, have written quite a few polyfills myself.

PR #2082, opened 2023-04-20
Cache storage cleanup

KhafraDev:
- adds docs
- exports types & adds tests for types
- adds `CacheStorage.prototype.match` (all WPTs for it pass)
- exposes it as `caches`, similar to browsers' [`globalThis.caches`](https://developer.mozilla.org/en-US/docs/Web/API/caches)

Test status:
```
[service-workers/cache-storage]: Completed: 103, failed: 13, success: 90, expected failures: 13, unexpected failures: 0, skipped: 2
```

### DIFF
--- a/docs/api/CacheStorage.md
+++ b/docs/api/CacheStorage.md
@@ -1,0 +1,30 @@
+# CacheStorage
+
+Undici exposes a W3C spec-compliant implementation of [CacheStorage](https://developer.mozilla.org/en-US/docs/Web/API/CacheStorage) and [Cache](https://developer.mozilla.org/en-US/docs/Web/API/Cache).
+
+## Opening a Cache
+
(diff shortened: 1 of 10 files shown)


KhafraDev:
the remaining failing wpts seem to be either a race condition in the test harness (unlikely?) or an issue with Cache.put, I just can't find the issue and can't repro the failures unless running tests from the runner

mcollina (approved):
lgtm

jimmywarting:
Q: is it available in NodeJS core as well?

KhafraDev:
1. only available in undici right now, not sure if http caching is something that's desirable in node core
2. no, unless by accident.

jimmywarting:
Hmm.... then, is `cacheStorage` something NodeJS core could consider exposing and adding to `globalThis` namespace?
maybe even as an experimental flag? you know as it's memory-only based for now... Deno and Web has it globally. and i would really like to use it in NodeJS as well right now.

> no, unless by accident.

i figured as such, due to strict brand check being used like instanceof checks and private symbol stuff being used here and there.

KhafraDev:
probably not, considering that only fetch uses Request/Response so it'd be pretty limited with node's current http/s/2 server

PR #2083, opened 2023-04-21
Cache storage fixes

KhafraDev:
fixes a cachestorage WPT and fixes the tests on linux

### DIFF
--- a/lib/cache/cache.js
+++ b/lib/cache/cache.js
@@ -32,7 +32,7 @@ class Cache {
    * @see https://w3c.github.io/ServiceWorker/#dfn-relevant-request-response-list
    * @type {requestResponseList}
    */
-  #relevantRequestResponseList = []
+  #relevantRequestResponseList
 
   constructor () {
     if (arguments[0] !== kConstruct) {
@@ -314,7 +314,7 @@ class Cache {
     let innerRequest = null
 
     // 2.
-    if (typeof request !== 'string') {
+    if (request instanceof Request) {
       innerRequest = request[kState]
     } else { // 3.
       innerRequest = new Request(request)[kState]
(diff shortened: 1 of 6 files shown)


ronag (approved):

lib/cache/cache.js
```diff
@@ -314,7 +314,7 @@ class Cache {
     let innerRequest = null
 
     // 2.
-    if (typeof request !== 'string') {
+    if (request instanceof Request) {
```
ronag, line 317: Is this fetch request?

ronag (commented):
This looks very fetchy... you sure it doesn't go under fetch dir?

lib/cache/cache.js
KhafraDev: yes it is, I import it so it's not the global one

KhafraDev:
It's tightly coupled with fetch, but fetch is a different spec controller by WHATWG, while Cache/CacheStorage is part of the service workers spec by the w3c (https://w3c.github.io/ServiceWorker/)

KhafraDev:
I think there's integration with spec with the cache option fetch takes, but I haven't looked into it yet cc @jimmywarting

jimmywarting:
I can't remember. 
It's true that browsers have a built in cache that they use themself internally. 
Otherwise we would not have a option for `cache`- mode.

I tried to implement `CacheStorage` ~6 years ago... I think i remembered reading something about "query a match" or origin partition store and RFC HTTP-CACHING logic or something like that. And it shared many similarities of checking cache headers and validating that it was okey to use.

I'm not sure if `fetch` even uses `Cache` API to handle it's own internal cache or not, it could be... but if it isn't then then i think it could be a good foundation to build a own internal cache based upon the `Cache` api.

I think it's like @KhafraDev said. `Cache/CacheStorage` is tightly coupled with `fetch` with the `Request/Response` classes and how it fetch things, but `fetch` isn't so much involved with `Cache/CacheStorage`.

I wonder what would happen if a service worker responded with something from `CacheStorage` and the main thread initiated a request with `{ cache: 'no-cache' }` option. would it make a request to the server or not then?

PR #2085, opened 2023-04-23
test: improve test coverage for ErrorEvent and MessageEvent

KhafraDev:
takes tests from the WPT repo that can't/aren't being run for whatever reason (being in an html file, being part of a spec undici doesn't implement, whatever) and ports them to tap. As expected there weren't any new bugs found in either of them.


KhafraDev:
also it looks like tap is having issues on node 20, had to revert to v19.9.0 locally to run the tests

PR #2088, opened 2023-04-23
websocket: add websocketinit

KhafraDev:
Refs: https://github.com/nodejs/undici/issues/1811#issuecomment-1518940816

This allows someone to supply an options bag to the WebSocket constructor :
```js
const ws = new WebSocket('ws://whatever', {
  dispatcher: new MyCustomDispatcher(),
  protocols: ['chat', 'echo']
})
```

Allowing it as the second param matches Bun and a proposal by a Deno maintainer (not implemented in Deno to my knowledge). Browsers (specifically Chrome) have no interest in expanding the spec to allow an options bag so this is the only way one will be added. The spec is unfortunately very limited, for both browser & server environments, so we do need to add this for any chance of having a useful WebSocket implementation.

### DIFF
--- a/docs/api/WebSocket.md
+++ b/docs/api/WebSocket.md
@@ -1,17 +1,40 @@
 # Class: WebSocket
 
-> ⚠️ Warning: the WebSocket API is experimental and has known bugs.
+> ⚠️ Warning: the WebSocket API is experimental.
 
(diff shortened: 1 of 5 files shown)


KhafraDev:
also conveniently lets people set headers for the request

ronag (approved):

PR #2091, opened 2023-04-24
fix: skip set content-length when FormData value is stream

fengmk2:
https://github.com/nodejs/undici/pull/2066/files#r1174932605

### DIFF
--- a/lib/fetch/body.js
+++ b/lib/fetch/body.js
@@ -123,6 +123,7 @@ function extractBody (object, keepalive = false) {
     const blobParts = []
     const rn = new Uint8Array([13, 10]) // '\r\n'
     length = 0
+    let hasUnknownSizeValue = false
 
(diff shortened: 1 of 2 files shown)


KhafraDev:
I'd also like to mention that this is also an issue in fetch so I'd prefer a solution that fixes the logic in `extractBody` or `isBlobLike` instead of patching this 1 case.

fengmk2:
@KhafraDev I made some modifications according to your suggestion. If there is a value with unknown size in the formdata, we will not set the content length.

KhafraDev (approved):
LGTM but just wondering, what's your use case for setting Blob-like things in FormData?

fengmk2:
@KhafraDev BlobFromStream example come from https://github.com/octet-stream/form-data#usage
> You can still use files sourced from any stream

PR #2095, opened 2023-04-25
feat: add message to fetch failed error

MaxLeiter:
The `fetch failed` error is annoying to hit and little information is provided if the response error is unhelpful. This should help point the user in the right direction if they don't think to check the source.


KhafraDev (changes requested):
this will break libraries that use the message to detect a fetch error, see https://github.com/sindresorhus/p-retry/ (https://github.com/sindresorhus/p-retry/pull/68). Plus, fetch could fail to reasons unrelated to a network error.

PR #2097, opened 2023-04-26
test: fix failing tests

KhafraDev:
The tests aren't failing according to the CI, they're just timing out. Not sure if this will fix them because they don't timeout locally for me 🤷.


KhafraDev:
no this doesn't work

mcollina (approved):
lgtm

KhafraDev:
Okay it passed without having to re-run any tests. Most issues are within the tests/tap, not with undici. They're failing, afaict, because some handle/resource is being kept open. So the fix is just to set a global teardown that manually exits the process. This way the child process(es) don't exit w/ SIGTERM, which is what happens when they timeout. I'm not sure why they're being marked as timing out because they're not.

anonrig (approved):

PR #2099, opened 2023-04-28
doc: update outdated command in contributing.md

jazelly:
llhttp no longer has `npm run build:wasm`

https://github.com/nodejs/llhttp/blob/c0f4e754b75c033fd3bd2018c8f3ffb673d43d90/package.json#L16


jazelly:
actually, does it make more sense to update undici's `build:wasm` to `build-wasm`?

Issue #2101, opened 2023-04-29
Could not open '/lib64/ld-linux-x86-64.so.2'


jazelly:
## Bug Description

When running `npm run build:wasm` on Mac M1 Pro, I get an error `qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory`. By looking at the Dockerfile, we hardcoded the `wasi-sdk` version to be linux.

## Reproducible By

`npm run build:wasm`

## Expected Behavior

The build script should pass

## Logs & Screenshots

`qemu-x86_64: Could not open '/lib64/ld-linux-x86-64.so.2': No such file or directory`

## Environment

Mac M1 Pro

### Additional context

Same as [this issue in llhttp](https://github.com/nodejs/llhttp/issues/208)
we might need to apply [the fix](https://github.com/nodejs/llhttp/pull/215/files) to undici

PR #2102, opened 2023-04-29
feat: allow build:wasm to auto detect platform

jazelly:
Fixes: https://github.com/nodejs/undici/issues/2101

Apply this fix https://github.com/nodejs/llhttp/pull/215/files

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -1,4 +1,4 @@
-FROM node:16-alpine
+FROM node:18-alpine
 
 ARG UID=1000
 ARG GID=1000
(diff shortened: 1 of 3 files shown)


build/Dockerfile
```diff
@@ -1,16 +1,9 @@
-FROM node:16.13.0-buster
```
jazelly, line 1: I hope there is no particular reason using buster

ronag (approved):

build/Dockerfile
mcollina: while we are at it, could you update this to Node v18 instead?

jazelly: @mcollina sure thing! done

KhafraDev (approved):

PR #2103, opened 2023-04-29
mimesniff: fix many broken tests

KhafraDev:
Before, after
```
[mimesniff]: Completed: 1898, failed: 989, success: 909, expected failures: 989, unexpected failures: 0, skipped: 0

[mimesniff]: Completed: 1898, failed: 872, success: 1026, expected failures: 872, unexpected failures: 0, skipped: 0
```

+117 tests passing.

https://wpt.fyi/results/mimesniff/mime-types/parsing.any.html?label=master&label=experimental&product=chrome&product=firefox&product=safari&product=deno&product=node.js&aligned

### DIFF
--- a/lib/fetch/dataURL.js
+++ b/lib/fetch/dataURL.js
@@ -42,14 +42,12 @@ function dataURLProcessor (dataURL) {
 
   // 6. Strip leading and trailing ASCII whitespace
   // from mimeType.
-  // Note: This will only remove U+0020 SPACE code
-  // points, if any.
   // Undici implementation note: we need to store the
   // length because if the mimetype has spaces removed,
   // the wrong amount will be sliced from the input in
   // step #9
   const mimeTypeLength = mimeType.length
-  mimeType = mimeType.replace(/^(\u0020)+|(\u0020)+$/g, '')
+  mimeType = removeASCIIWhitespace(mimeType, true, true)
 
   // 7. If position is past the end of input, then
   // return failure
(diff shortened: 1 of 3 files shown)


KhafraDev:
cc @panva the duration on wpt.fyi shows as N/A, seems like an issue with the reports?

lib/fetch/dataURL.js
```diff
@@ -561,6 +564,33 @@ function serializeAMimeType (mimeType) {
   return serialization
 }
 
+/**
+ * @see https://fetch.spec.whatwg.org/#http-whitespace
+ * @param {string} char
+ */
+function isHTTPWhiteSpace (char) {
+  return char === '\r' || char === '\n' || char === '\t' || char === ' '
```
anonrig, line 572: Can you compare ascii numbers?

lib/fetch/dataURL.js
```diff
@@ -561,6 +564,33 @@ function serializeAMimeType (mimeType) {
...
+/**
+ * @see https://fetch.spec.whatwg.org/#http-whitespace
+ * @param {string} str
+ */
+function removeHTTPWhitespace (str, leading = true, trailing = true) {
+  let lead = 0
+  let trail = str.length - 1
+
+  if (leading) {
+    for (; lead < str.length && isHTTPWhiteSpace(str[lead]); lead++);
```
anonrig, line 584: findIndex would be faster than this I assume?

lib/fetch/response.js
```diff
@@ -467,7 +467,7 @@ function initializeResponse (response, init, body) {
 
   // 5. If init["headers"] exists, then fill response’s headers with init["headers"].
   if ('headers' in init && init.headers != null) {
-    fill(response[kState].headersList, init.headers)
+    fill(response[kHeaders], init.headers)
```
KhafraDev, line 470: I think we've had some discussion about this line before, using headersList here was always incorrect, but a workaround for some bugs with data urls. However it'd also bypass header name/header value validation which *almost* sounds like a vulnerability...

lib/fetch/dataURL.js
KhafraDev: Performance wasn't really my goal here, believe it or not with this monstrosity. We actually parse the WPT data set 1.5x faster than deno, and 2-4x faster than browsers lol

lib/fetch/dataURL.js
KhafraDev: I don't think it's needed, or at least I'd prefer to wait for there to be benchmarks before switching over.

ronag (approved):

panva:
> cc @panva the duration on wpt.fyi shows as N/A, seems like an issue with the reports?

Never bothered to figure out what triggers its inclusion.

mcollina (approved):
lgtm

PR #2106, opened 2023-05-02
Fix bug in 16-bit frame length when buffer is a subarray

jawj:
I was just bitten by this issue.

Because `buffer` could be a [subarray](https://nodejs.org/api/buffer.html#bufsubarraystart-end), the `byteOffset` and `byteLength` arguments to `new DataView()` are required, otherwise the `bodyLength` value may be written at the wrong offset.

Alternatively, it looks like it would be simpler (and I've tested it works fine) to replace the patched line with `buffer.writeUInt16BE(bodyLength, 2)`.

### DIFF
--- a/lib/websocket/frame.js
+++ b/lib/websocket/frame.js
@@ -43,7 +43,7 @@ class WebsocketFrameSend {
     buffer[1] = payloadLength
 
     if (payloadLength === 126) {
-      new DataView(buffer.buffer, buffer.byteOffset, buffer.byteLength).setUint16(2, bodyLength)
+      buffer.writeUInt16BE(bodyLength, 2)
     } else if (payloadLength === 127) {
       // Clear extended payload length
       buffer[2] = buffer[3] = 0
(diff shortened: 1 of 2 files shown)


mcollina (commented):

KhafraDev (approved):

anonrig (approved):

jawj:
I wasn't exactly sure what you kind of test you wanted, but I've added a simple new test for `WebsocketFrameSend` that (almost certainly) fails without my fix.

I also switched to using `writeUInt16BE` instead of correcting the `DataView` creation, because it's simpler and more efficient and I couldn't see any reason why not. :)

It would be much appreciated if this fix could be released ASAP: a project I'm working on is blocked on this.

mcollina (approved):
lgtm

jawj:
Fab! Do you know how soon this will make it into the npm package?

jawj:
@mcollina Sorry to chase, but very keen to see this released to npm soon, because WebSockets are badly broken in the meantime. Any idea when that might happen?

PR #2112, opened 2023-05-07
fix: update error definitions

dfilatov:
This PR adds missing parts of typescript definitions to match with js implementation


mcollina (commented):
Could you add a couple of tests for those types? We used tsd.

dfilatov:
@mcollina added

mcollina (approved):
lgtm

PR #2119, opened 2023-05-12
bump engines to node >= 16

ronag:
Node 14 is EOL and no longer supported.


KhafraDev (approved):
We can finally use optional chaining and nullish coalescing 👀

anonrig (approved):

mcollina:
note that this is a server-major change.

mcollina:
this is a semver major change. Given v16 is out of support at the end of September, I think we could just “jump” to v18 minimum.

KhafraDev:
I think we should wait to drop v16, otherwise there will be people on a supported version of node that can't use undici. I thought about dropping v16 support too but decided against it until it's EOL.

Ethan-Arrowood:
Yeah, please keep 16 support for now. I see little harm in doing another Undici major release when 16 hits EOL. 

At Vercel we use Undici and fetch from core heavily. We can't drop Node 14 or 16 support as quickly and so will have to pin these older versions of Undici until we can properly drop support for 14 and 16 ourselves.

mcollina:
I would recommend we revert this change then. Let's drop v14 and v16 when v16 enters EOL.

PR #2120, opened 2023-05-12
fix: remove unneeded version checks

KhafraDev:


ronag (commented):
Why don't we need the v16 checks anyomore?

KhafraDev:
Which ones specifically? The min node version is v16 so making sure it's v16 / checking if the node version is less than 16 isn't needed anymore.

ronag (approved):

mcollina:
I recommend we keep support for v14 until v16 is EOL.

PR #2126, opened 2023-05-15
fix: support truncated gzip

### LINKED ISSUE #2123: support slightly invalid gzip response
Here is one test we have in node-fetch, and i run it on undici fetch but it failed...
our test:
```js
	it.only('should decompress slightly invalid gzip response', async () => {
		const url = `${base}gzip-truncated`;
		const res = await fetch(url);
		expect(res.headers.get('content-type')).to.equal('text/plain');
		const result = await res.text();
		expect(result).to.be.a('string');
		expect(result).to.equal('hello world');
	});
```
```js
// our response

		if (p === '/gzip-truncated') {
			res.statusCode = 200;
			res.setHeader('Content-Type', 'text/plain');
			res.setHeader('Content-Encoding', 'gzip');
			zlib.gzip('hello world', (err, buffer) => {
				if (err) {
					throw err;
				}

				// Truncate the CRC checksum and size check at the end of the stream
				res.end(buffer.slice(0, -8)); // Buffer.from('H4sIAAAAAAAAE8tIzcnJVyjPL8pJAQA=')
			});
		}
```

this works fine in browsers and some server ignore this CRC checksum and size at the end

jimmywarting:
- fix #2123

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -2007,7 +2007,14 @@ async function httpNetworkFetch (
             for (const coding of codings) {
               // https://www.rfc-editor.org/rfc/rfc9112.html#section-7.2
               if (coding === 'x-gzip' || coding === 'gzip') {
-                decoders.push(zlib.createGunzip())
+                decoders.push(zlib.createGunzip({
+                  // Be less strict when decoding compressed responses, since sometimes
+                  // servers send slightly invalid responses that are still accepted
+                  // by common browsers.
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

KhafraDev (approved):

jimmywarting:
Just wondering. if the gzip response do contain a integrity checks and isn't truncated. dose it still validate/throw an error if it's incorrect?
ignoring flush kind of feel like you have to do that check yourself manually afterwards

PR #2145, opened 2023-06-01
added tracing channels on fetch

ida613:
We added tracing channels on fetch to support fetch instrumentation. The [channels](https://nodejs.org/api/diagnostics_channel.html#class-tracingchannel) added emit the following events:

- tracing:undici:start emits input (url) and init settings
- tracing:undici:end emits input (url), init settings and error
- tracing:undici:asyncStart emits input (url), init settings, result (response) and error
- tracing:undici:asyncEnd emits input (url), init settings, result (response) and error

Use case: APM tools could provide support for fetch by using the tracing channels.

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -62,6 +62,7 @@ const { TransformStream } = require('stream/web')
 const { getGlobalDispatcher } = require('../global')
 const { webidl } = require('./webidl')
 const { STATUS_CODES } = require('http')
+const diagnosticsChannel = require('diagnostics_channel')
 
 /** @type {import('buffer').resolveObjectURL} */
 let resolveObjectURL
@@ -122,7 +123,6 @@ class Fetch extends EE {
 
 let tracingChannel
 try {
-  const diagnosticsChannel = require('diagnostics_channel')
   tracingChannel = diagnosticsChannel.tracingChannel('undici')
 } catch {
   tracingChannel.start = { hasSubscribers: false }
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -120,137 +120,151 @@ class Fetch extends EE {
   }
 }
 
+let tracingChannel
+try {
+  const diagnosticsChannel = require('diagnostics_channel')
+  tracingChannel = diagnosticsChannel.tracingChannel('undici')
+} catch {
+  tracingChannel.start = { hasSubscribers: false }
+  tracingChannel.end = { hasSubscribers: false }
+  tracingChannel.asyncStart = { hasSubscribers: false }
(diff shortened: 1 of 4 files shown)


lib/fetch/index.js
```diff
@@ -120,137 +120,151 @@ class Fetch extends EE {
...
+let tracingChannel
+try {
+  const diagnosticsChannel = require('diagnostics_channel')
+  tracingChannel = diagnosticsChannel.tracingChannel('undici')
+} catch {
+  tracingChannel.start = { hasSubscribers: false }
+  tracingChannel.end = { hasSubscribers: false }
+  tracingChannel.asyncStart = { hasSubscribers: false }
+  tracingChannel.asyncEnd = { hasSubscribers: false }
+  tracingChannel.error = { hasSubscribers: false }
```
KhafraDev, line 132: We don't need this try/catch because fetch is only supported in node >= 16.8.0. Diagnostics_channel was added in some version of node 14.

KhafraDev (changes requested):
tracePromise was added in [v19.9.0](https://nodejs.org/api/diagnostics_channel.html#tracingchanneltracepromisefn-context-thisarg-args) and is experimental. We shouldn't rely on it until it's stable and we're willing to drop v16 and v18 support.

lib/fetch/index.js
ida613: just updated the PR to remove try/catch on DC

mcollina (changes requested):
Why are not the channels defined in https://github.com/nodejs/undici/blob/main/docs/api/DiagnosticsChannel.md enough?
Can we amend those so all undici APIs are supported? I see quite a bit of duplication.

In case new ones are needed, those needs to be prefixed following the same pattern `undici:fetch`.

mcollina:
last but not least, using plain diagnostics_channel ensure maximum compatibility. There is no need to use tracingChannel.

metcoder95:
Is this something we want to persue or we can close it up?
@ida613 @KhafraDev @mcollina ?

mcollina:
I'm ok to close this given the low interest. Note that I've explained how to do this in  https://github.com/nodejs/undici/pull/2145#pullrequestreview-1473322600. It's 100% possible to write tracing-channel compabile diagnostics_channels. The first one maximises compatibility.

metcoder95:
Let's close for now then 👍 

@ida613, if still interested in persuing its implementation feel free to open a new PR to see discuss about it 🙂

ida613:
@mcollina @metcoder95 thank you for reviewing the PR! Sorry I haven't been on top of it. I'm working up a new PR including the changes you mentioned

PR #2149, opened 2023-06-02
Additional WebSocket send tests to cover all payload size categories

jawj:
As a follow-up to https://github.com/nodejs/undici/pull/2106, some additional tests. Sorry about the multiple commits listed — think you can just squash these?


KhafraDev (approved):
thank you!

we can squash your commits, dw.

Issue #2161, opened 2023-06-16
"Error: Request was cancelled" when using ProxyAgent


NBNGaming:
## Bug Description

When using ProxyAgent and http proxy responds with status code !== 200, Undici throws "Error: Request was cancelled". This error message is confusing and irrelevant to problem cause.
This line throws error:
https://github.com/nodejs/undici/blob/c415fbbb59e2b898c5db6a681265cf3da865d02c/lib/proxy-agent.js#L103

## Expected Behavior

Throwing error with meaningful error "Proxy response !== 200 when HTTP Tunneling".

## Logs & Screenshots
```
TypeError: fetch failed
    at fetch (node_modules/undici/index.js:109:13)
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  cause: Error: Request was cancelled.
      at makeNetworkError (node_modules/undici/lib/fetch/response.js:351:9)
      at makeAppropriateNetworkError (node_modules/undici/lib/fetch/response.js:437:7)
      at httpNetworkFetch (node_modules/undici/lib/fetch/index.js:1763:14)
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
      at async httpNetworkOrCacheFetch (node_modules/undici/lib/fetch/index.js:1450:29)
      at async httpFetch (node_modules/undici/lib/fetch/index.js:1030:33)
      at async schemeFetch (node_modules/undici/lib/fetch/index.js:886:14)
      at async node_modules/undici/lib/fetch/index.js:599:16
      at async mainFetch (node_modules/undici/lib/fetch/index.js:583:16)
}
```

## Environment
Ubuntu 22.04.2 and Windows 11
Node.js 18.16.0
Undici v5.22.1

ronag:
The cause message should probably not be `Request was cancelled.` I'm not sure why the proper message isn't forwarded.

NBNGaming:
Because this line replaces proper error with `Request was cancelled`
https://github.com/nodejs/undici/blob/c415fbbb59e2b898c5db6a681265cf3da865d02c/lib/fetch/response.js#L437

ronag:
Hm. Then I don't know what we can do without violating the spec. @KhafraDev

KhafraDev:
a possible solution is to add an optional cause to makeNetworkError

KhafraDev:
[...]

PR #2162, opened 2023-06-16
fix: pass ProxyAgent proxy status code error

NBNGaming:
Fixes: #2161

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -1759,7 +1759,12 @@ async function httpNetworkFetch (
       // 1. If connection uses HTTP/2, then transmit an RST_STREAM frame.
       fetchParams.controller.connection.destroy()
 
-      // 2. Return the appropriate network error for fetchParams.
+      // 2. If error is ProxyAgent related, then pass it.
+      if (err.message === 'Proxy response !== 200 when HTTP Tunneling') {
+        return makeNetworkError(err.message)
+      }
+
+      // 3. Return the appropriate network error for fetchParams.
       return makeAppropriateNetworkError(fetchParams)
     }
 


ronag (changes requested):
This doesn't follow spec.

lib/fetch/index.js
```diff
@@ -1759,7 +1759,12 @@ async function httpNetworkFetch (
       // 1. If connection uses HTTP/2, then transmit an RST_STREAM frame.
       fetchParams.controller.connection.destroy()
 
-      // 2. Return the appropriate network error for fetchParams.
+      // 2. If error is ProxyAgent related, then pass it.
```
KhafraDev, line 1762: this comment isn't related to the fetch spec, the comment changes should be reverted. Instead, add a comment explaining why this is needed or link to the issue.

lib/fetch/index.js
```diff
@@ -1759,7 +1759,12 @@ async function httpNetworkFetch (
...
+      if (err.message === 'Proxy response !== 200 when HTTP Tunneling') {
```
KhafraDev, line 1763:
```suggestion
      if (err.code === 'UND_ERR_ABORTED') {
```

KhafraDev (changes requested):
also needs a test

metcoder95:
Any update @NBNGaming ?

NBNGaming:
> Any update @NBNGaming ?

I'm sorry, I don't have free time right now to make the required changes to the commit.

lib/fetch/index.js
NBNGaming: In my code, I check whether the request was aborted exactly because of a proxy error. And there is no need for a second check that the request has been aborted, there is already one.
https://github.com/nodejs/undici/blob/00f40cdc8b102053f2c6f8fac9b5bf09df4c4fe3/lib/fetch/index.js#L1758

KhafraDev: checking error messages is bound to cause issues

NBNGaming: Then I suggest adding a new UndiciError subclass - ProxyError. There is no other way to separate the errors related to the ProxyAgent.

ronag (commented):
Why are we not just passing the error as the cause?

NBNGaming:
I didn't quite understand your remark, but [RequestAbortedError](https://github.com/nodejs/undici/blob/403c249f103b8305ed7ee72a7395fcc959e699a5/lib/core/errors.js#L85) constructor doesn't have a cause argument and [makeAppropriateNetworkError](https://github.com/nodejs/undici/blob/403c249f103b8305ed7ee72a7395fcc959e699a5/lib/fetch/response.js#L429) overrides error.

ronag:
```
diff --git a/lib/fetch/index.js b/lib/fetch/index.js
index d615f07e..4cb63807 100644
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -1760,10 +1760,10 @@ async function httpNetworkFetch (
       fetchParams.controller.connection.destroy()
 
       // 2. Return the appropriate network error for fetchParams.
-      return makeAppropriateNetworkError(fetchParams)
+      return Object.assign(makeAppropriateNetworkError(fetchParams), { cause: err })
     }
 
-    return makeNetworkError(err)
+    return Object.assign(makeNetworkError(err), { cause: err })
   }
 
   // 11. Let pullAlgorithm be an action that resumes the ongoing fetch

```

NBNGaming:
I implemented the required changes and added a test.

ronag (approved):

KhafraDev:
```
Subtest: pass ProxyAgent proxy status code error when using fetch - \#2161
        not ok 1 - should have the named property
          ---
          found:
            !error
            name: TypeError
            message: fetch is not a function
            stack: >-
              TypeError: fetch is not a function
                  at Test.<anonymous> (/home/runner/work/undici/undici/test/proxy-agent.js:459:11)
                  at processTicksAndRejections (internal/process/task_queues.js:95:5)
          pattern: cause
          at:
            line: 461
            column: 7
            file: test/proxy-agent.js
            type: Test
          stack: |
            Test.<anonymous> (test/proxy-agent.js:461:7)
          source: |2
              } catch (e) {
                t.hasProp(e, 'cause')
            ------^
              }
          ...
        
        1..1
        # failed 1 test
```

this test fails on v14

PR #2181, opened 2023-07-05
stricter types for bodymixin.json

KhafraDev:
When we first added types for these methods, I thought it would have been best to follow the official DOM types. Since then, I learned that the DOM types are garbage, and that using unknown would have been the better choice. When types were added for fetch (or soon thereafter), `.json()` was changed to use unknown instead of any, providing much better type safety.

IMO this is not a major change, ignoring that it is a types-only change, and someone can revert to using the old `any` behavior very easily:
```diff
- const value = await body.json()
+ const value: any = await body.json()
```


metcoder95 (approved):

mcollina (approved):
lgtm

PR #2184, opened 2023-07-09
change `statusCode` to `status`

renhiyama:
I dont see `statusCode` in the request. Instead I see `status`:
image

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,13 +49,13 @@ The benchmarks below have the [simd](https://github.com/WebAssembly/simd) featur
 import { request } from 'undici'
 
 const {
-  statusCode,
+  status,
   headers,
   trailers,
   body
 } = await request('http://localhost:3000/foo')
 
-console.log('response received', statusCode)
+console.log('response received', status)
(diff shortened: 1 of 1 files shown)


metcoder95:
Thanks for the contribution!
Do you have a reproducible example we can take a look at?

The `status` is scoped to the usage of `fetch` due to the spec, you should find `statusCode` otherwise.

renhiyama:
> Thanks for the contribution! Do you have a reproducible example we can take a look at?
> 
> The `status` is scoped to the usage of `fetch` due to the spec, you should find `statusCode` otherwise.

```js
let req = await fetch("https://httpstat.us/500");
console.log(req);
```
I'm running this code on node v18+

It logs `status` again instead of `statusCode`

KhafraDev (changes requested):
the docs are using `undici.request`, not fetch.

renhiyama:
Alright.

PR #2188, opened 2023-07-14
fix: fetch resource timing performance entry names should be strings

GaryWilber:
The performance resource timings api expects the name parameter to be a string. Undici's fetch implementation is currently passing a URL object when calling `markResourceTiming`. This is outside the spec and makes it impossible to call `getEntriesByName(name)` to find the performance entry for a given url.

The fix is simple and I added a test to verify `getEntriesByName` can find the performance entries created by undici.

References:
- https://nodejs.org/docs/latest-v18.x/api/perf_hooks.html#performancemarkresourcetimingtiminginfo-requestedurl-initiatortype-global-cachemode - `requestedUrl` should be a `string`
- https://w3c.github.io/resource-timing/#dfn-requested-url - `A PerformanceResourceTiming has an associated DOMString requested URL.`
- https://www.w3.org/TR/performance-timeline/#the-performanceentry-interface - `PerformanceEntry.name` should be a `DOMString`.

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -320,7 +320,7 @@ function finalizeAndReportTiming (response, initiatorType = 'other') {
 // https://w3c.github.io/resource-timing/#dfn-mark-resource-timing
 function markResourceTiming (timingInfo, originalURL, initiatorType, globalThis, cacheState) {
   if (nodeMajor > 18 || (nodeMajor === 18 && nodeMinor >= 2)) {
-    performance.markResourceTiming(timingInfo, originalURL.toString(), initiatorType, globalThis, cacheState)
+    performance.markResourceTiming(timingInfo, originalURL.href, initiatorType, globalThis, cacheState)
   }
 }
 
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -320,7 +320,7 @@ function finalizeAndReportTiming (response, initiatorType = 'other') {
 // https://w3c.github.io/resource-timing/#dfn-mark-resource-timing
 function markResourceTiming (timingInfo, originalURL, initiatorType, globalThis, cacheState) {
   if (nodeMajor > 18 || (nodeMajor === 18 && nodeMinor >= 2)) {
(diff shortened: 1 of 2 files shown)


KhafraDev (commented):
I won't block this, but this seems like a bug in node. Node should be doing some sort of webidl conversions here to match the rest of performance.*

mcollina (changes requested):
I'm -1 with this implementation. The problem is that adding a `url.toString()` is going to be relatively _expensive_, and it conflicts with the goal of #1203.

Maybe I'm overly paranoid, but this should be benchmarked.

mcollina:
@anonrig how much a `url.toString()` would cost us?

KhafraDev:
it doesn't seem expensive at all, toString returns the href that is parsed in the URL constructor. I doubt it adds any overhead.

mcollina:
If @anonrig confirm, happy to +1 this

anonrig:
> @anonrig how much a `url.toString()` would cost us?

@mcollina  toString just returns a href we already get from C++. There is no cost. I agree with @KhafraDev

lib/fetch/index.js
```diff
@@ -320,7 +320,7 @@ function finalizeAndReportTiming (response, initiatorType = 'other') {
 // https://w3c.github.io/resource-timing/#dfn-mark-resource-timing
 function markResourceTiming (timingInfo, originalURL, initiatorType, globalThis, cacheState) {
   if (nodeMajor > 18 || (nodeMajor === 18 && nodeMinor >= 2)) {
-    performance.markResourceTiming(timingInfo, originalURL, initiatorType, globalThis, cacheState)
+    performance.markResourceTiming(timingInfo, originalURL.toString(), initiatorType, globalThis, cacheState)
```
anonrig, line 323: i prefer calling url.href since toString returns href.

mcollina (approved):
lgtm

anonrig (approved):

mcollina (approved):
lgtm

PR #2196, opened 2023-07-28
fix: make multipart/form-data boundary string more consistent

LiviaMedeiros:
Stringifying `Math.random()` value directly sometimes produces very short boundary string.
Also there is an extreme corner case of it returning exactly `0.0`.


ronag (approved):

metcoder95 (approved):

mcollina (approved):
lgtm

PR #2218, opened 2023-08-18
allow http & https websocket urls

KhafraDev:
https://github.com/whatwg/websockets/commit/771cd4ffa1385c0b66ac17d0e82015f354359e36


KhafraDev:
cc @nodejs/undici

lib/websocket/websocket.js
```diff
@@ -57,38 +58,48 @@ class WebSocket extends EventTarget {
...
     } catch (e) {
-      // 2. If urlRecord is failure, then throw a "SyntaxError" DOMException.
+      // 3. If urlRecord is failure, then throw a "SyntaxError" DOMException.
       throw new DOMException(e, 'SyntaxError')
     }
 
-    // 3. If urlRecord’s scheme is not "ws" or "wss", then throw a
-    //    "SyntaxError" DOMException.
+    // 4. If urlRecord’s scheme is "http", then set urlRecord’s scheme to "ws".
+    if (urlRecord.protocol === 'http:') {
```
anonrig, line 75: Can you store the protocol in a variable so it doesn't trigger the getter everytime?

lib/websocket/websocket.js
```diff
@@ -100,12 +111,12 @@ class WebSocket extends EventTarget {
       throw new DOMException('Invalid Sec-WebSocket-Protocol value', 'SyntaxError')
     }
 
-    // 7. Set this's url to urlRecord.
-    this[kWebSocketURL] = urlRecord
+    // 10. Set this's url to urlRecord.
+    this[kWebSocketURL] = new URL(urlRecord.href)
```
anonrig, line 115: Why do you need this? This seems unnecessary

KhafraDev: test/wpt/tests/websockets/Create-http-urls.any.js would fail without this. It's important to note that websocket connections aren't typically opened very much and it's very unlikely that this is a bottleneck if many connections are opened

ronag (approved):

lib/websocket/websocket.js
```diff
@@ -57,38 +58,48 @@ class WebSocket extends EventTarget {
...
-    //    "SyntaxError" DOMException.
+    // 4. If urlRecord’s scheme is "http", then set urlRecord’s scheme to "ws".
+    if (urlRecord.protocol === 'http:') {
+      urlRecord.protocol = 'ws:'
+    } else if (urlRecord.protocol === 'https:') {
+      // 5. Otherwise, if urlRecord’s scheme is "https", set urlRecord’s scheme to "wss".
+      urlRecord.protocol = 'wss:'
+    }
+
+    // 6. If urlRecord’s scheme is not "ws" or "wss", then throw a "SyntaxError" DOMException.
```
anonrig, line 82: The next line triggers the protocol getters, and can just be replaced buy a boolean variable. The getters of URLs splices the string, and avoiding them is a good practice imho.

lib/websocket/websocket.js
anonrig: Nit: Line 75 and 77 triggers the protocol getters, which can just be removed into one.

lib/websocket/websocket.js
```diff
@@ -57,38 +58,48 @@ class WebSocket extends EventTarget {
...
         `Expected a ws: or wss: protocol, got ${urlRecord.protocol}`,
         'SyntaxError'
       )
     }
 
-    // 4. If urlRecord’s fragment is non-null, then throw a "SyntaxError"
+    // 7. If urlRecord’s fragment is non-null, then throw a "SyntaxError"
     //    DOMException.
-    if (urlRecord.hash) {
+    if (urlRecord.hash || urlRecord.href.endsWith('#')) {
```
anonrig, line 92: I think we should have a method in the WHATWG url to get if the fragment is `non-null`.

Nit: `urlRecord.hash.length > 0 || urlRecord.href.endsWith('#')`

anonrig (approved):

lib/websocket/websocket.js
KhafraDev: p.s.: I have no idea why the test fails, it took me a little while to track down the issue to here

KhafraDev:
I think we can leave possible performance improvements to a dedicated PR, maybe once or if this lands in core when there might be more contributors

metcoder95:
After the `h2` work, I'll try to spend some time on the performance topics.

mcollina:
how does this work? May I ask for a quick example? It's hard to grasp from the wpts.

KhafraDev:
it's the same as using a `ws` or `wss` url:
```js
new WebSocket('http://websocket')
// is the same as
new WebSocket('ws://websocket')

new WebSocket('https://websocket')
// is the same as
new WebSocket('wss://websocket')
```

PR #2225, opened 2023-08-25
fix: adjust number of arguments in `Response.json()`

LiviaMedeiros:
`data` is not optional

Refs: https://fetch.spec.whatwg.org/#ref-for-dom-response-json

PR #2226, opened 2023-08-25
Update wpts

KhafraDev:
Closes #2225


mcollina (approved):
lgtm

lib/fetch/util.js
```diff
@@ -556,16 +556,37 @@ function bytesMatch (bytes, metadataList) {
...
     //    return true.
     if (actualValue === expectedValue) {
       return true
     }
+
+    let actualBase64URL = crypto.createHash(algorithm).update(bytes).digest('base64url')
+
+    if (actualBase64URL.endsWith('==')) {
+      actualBase64URL = actualBase64URL.slice(0, -2)
+    }
```
panva, line 585: This will always eval as `false`

panva, line 585: @KhafraDev This will always eval as `false`

PR #2234, opened 2023-08-31
feat: respect `--max-http-header-size` Node.js flag

balazsorban44:
## This relates to...

Fixes: https://github.com/nodejs/node/issues/47246

## Rationale

Rationale is https://github.com/nodejs/node/issues/47246#issuecomment-1700892661

### Features

See the rationale

### Bug Fixes

N/A

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [x] In review
- [x] Merge ready


lib/client.js
```diff
@@ -223,7 +224,7 @@ class Client extends DispatcherBase {
     this[kConnector] = connect
     this[kSocket] = null
     this[kPipelining] = pipelining != null ? pipelining : 1
-    this[kMaxHeadersSize] = maxHeaderSize || 16384
+    this[kMaxHeadersSize] = maxHeaderSize || http.maxHeaderSize || 16384
```
metcoder95, line 227: `http.maxHeaderSize` has a default value of `16KiB`, meaning that the property will always have a value.

Would it be better to just default to `http.maxHeaderSize` if we are going to be depending on this variable? Anyways is the same default value we have

cc: @ronag @mcollina

balazsorban44: That makes sense since that value is also read-only. I removed the default 16384. Let me know if it looks good. :+1:

test/client-node-max-header-size.js
```diff
@@ -0,0 +1,20 @@
+const { execSync } = require('node:child_process')
```
mcollina, line 1: use strict is missing

mcollina (approved):
lgtm with a nit

docs/api/Client.md
```diff
@@ -22,7 +22,7 @@ Returns: `Client`
...
 * **keepAliveTimeout** `number | null` (optional) - Default: `4e3` - The timeout after which a socket without active requests will time out. Monitors time between activity on a connected socket. This value may be overridden by *keep-alive* hints from the server. See [MDN: HTTP - Headers - Keep-Alive directives](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Keep-Alive#directives) for more details. Defaults to 4 seconds.
 * **keepAliveTimeoutThreshold** `number | null` (optional) - Default: `1e3` - A number subtracted from server *keep-alive* hints when overriding `keepAliveTimeout` to account for timing inaccuracies caused by e.g. transport latency. Defaults to 1 second.
-* **maxHeaderSize** `number | null` (optional) - Default: `16384` - The maximum length of request headers in bytes. Defaults to 16KiB.
+* **maxHeaderSize** `number | null` (optional) - Default: `--max-http-header-size` or `16384` - The maximum length of request headers in bytes. Defaults to Node.js' --max-http-header-size or 16KiB.
```
metcoder95, line 25: Doc's update and we are good to go 👍

types/client.d.ts
```diff
@@ -24,7 +24,7 @@ export declare namespace Client {
   export interface Options {
     /** TODO */
     interceptors?: OptionsInterceptors;
-    /** The maximum length of request headers in bytes. Default: `16384` (16KiB). */
```
metcoder95, line 27: Same here

metcoder95 (approved):

docs/api/Client.md
balazsorban44: What's missing here?

types/client.d.ts
balazsorban44: What's missing?

docs/api/Client.md
metcoder95: Never mind, I misread it 👍

types/client.d.ts
metcoder95: Same here, I my wires crossed

metcoder95 (approved):

mcollina:
Unfortunately there are a significant amount of CI failures on old nodes.

balazsorban44:
- Node.js 14 is EOL, it's just a syntax error because of `??=` in the source code that I import in the test.

- Node.js 16 does not have `globalThis.fetch` so updated the test to use the imported version although arguably it's also EOL in 1 week.

- Node.js 18 passes

- Node.js 20 tests fail for me locally, even on `main` so seems to be unrelated

test/client-node-max-header-size.js
```diff
@@ -0,0 +1,22 @@
+'use strict'
+
+const { execSync } = require('node:child_process')
+const { test } = require('tap')
+
+const command = 'node -e "require(\'./index-fetch\').fetch(\'https://httpbin.org/get\')"'
```
metcoder95, line 6: If you want to test it under fetch, you need to put the testing suite under `test/fetch` as that suite runs with `Node v16` and above.

I missed that at a first glance, but it would be good to have a new one just for `fetch` and a second one that does the same but with pure `undici`.

metcoder95 (commented):
>Node.js 14: it's just a syntax error because of ??= in the source code that I imported in the test. To make this work, build:node would have to run before tests in the CI and use the built version. Is it worth it since v14 is EOL? Maybe just remove it from the CI matrix?

I believe that will be better to just separate the tests as suggested, and just add a new one that focuses on `undici`.

test/client-node-max-header-size.js
balazsorban44: OK, copied the test over to `test/fetch`, what do you mean by a pure `undici` test?

Ethan-Arrowood: test that this behaves the same with the `request` method or something like that.

mcollina:
linting is failing

mcollina:
linting is still failing

diaz-hfc:
@balazsorban44 need this, can you help me fixing the linting issues? Thank you :)

balazsorban44:
Node.js 20 on windows seems to be flaky. It ran successfully here: https://github.com/nodejs/undici/actions/runs/6419488071/job/17433770898?pr=2234

Ethan-Arrowood (approved):

metcoder95 (approved):

Ethan-Arrowood:
Test failure is unrelated. Merging 😄 thank you @balazsorban44 !

PR #2251, opened 2023-09-08
Fix fetch in node v20.6.0

mcollina:
Node v20.6.0 added `ReadableStream.from()` and we broke.

### DIFF
--- a/lib/core/util.js
+++ b/lib/core/util.js
@@ -355,6 +355,15 @@ function getSocketInfo (socket) {
   }
 }
 
+async function * convertIterableToBuffer (iterable) {
+  for await (let chunk of iterable) {
+    if (!Buffer.isBuffer(chunk)) {
+      chunk = Buffer.from(chunk)
+    }
+    yield chunk
+  }
+}
+
 let ReadableStream
 function ReadableStreamFrom (iterable) {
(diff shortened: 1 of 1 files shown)


ronag (approved):

lib/core/util.js
```diff
@@ -355,15 +355,23 @@ function getSocketInfo (socket) {
   }
 }
 
+async function * convertIterableToBuffer (iterable) {
+  for await (let chunk of iterable) {
+    if (!Buffer.isBuffer(chunk)) {
+      chunk = Buffer.from(chunk)
+    }
+    yield chunk
```
ronag, line 363:
```suggestion
    yield Buffer.isBuffer(chunk) ? chunk : Buffer.from(chunk)
```

ronag: .nit

ronag (commented):
I'm a little confused why web stream from requires `Buffer` when `Buffer` is not part of the web standard. Is it to convert strings to `Uint8Array`s?

mcollina:
> I'm a little confused why web stream from requires Buffer when Buffer is not part of the web standard. Is it to convert strings to Uint8Arrays?

Yes, we need UInt8Arrays

PR #2253, opened 2023-09-08
Fix http2 fetch test

mcollina:
fixup

### DIFF
--- a/test/fetch/http2.js
+++ b/test/fetch/http2.js
@@ -11,6 +11,8 @@ const pem = require('https-pem')
 
 const { Client, fetch } = require('../..')
 
+const hasToWeb = typeof Readable.toWeb === 'function'
+
 plan(4)
 
 test('[Fetch] Should handle h2 request with body (string or buffer)', async t => {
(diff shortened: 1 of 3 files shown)


test/fetch/http2.js
```diff
@@ -65,7 +66,8 @@ test('[Fetch] Should handle h2 request with body (string or buffer)', async t =>
   t.equal(responseBody, expectedRequestBody)
 })
 
-test('[Fetch] Should handle h2 request with body (stream)', async t => {
+// Skipping for now, there is something odd in the way the body is handled
+test('[Fetch] Should handle h2 request with body (stream)', { skip: true }, async t => {
```
mcollina, line 70: @metcoder95 @KhafraDev I'm skipping this now, but this needs some investigation.

ronag (approved):

test/fetch/http2.js
metcoder95: Yeah, it is kinda of flaky. I couldn't reproduce it all the time locally. I'll focus on that.

I can try to move it to use an AsyncIterator instead as I believe the way I set the stream to handle the data is the culprit. I'll take a look at it

PR #2258, opened 2023-09-08
fix: h2 without body

### LINKED ISSUE #2257: http2 demo not working as expected
Consider the following:

```js
import { request, fetch, Agent, Client, setGlobalDispatcher } from './index.js'

setGlobalDispatcher(new Agent({
  allowH2: true
}))

// Not working as expected
// the request is done using HTTP/1.1
const res1 = await fetch(`https://http2.pro/api/v1`)
console.log(await res1.json())

// Not working as expected, the response is never received
const client = new Client('https://http2.pro', {
  allowH2: true
})

const res2 = await request(`https://http2.pro/api/v1`, { dispatcher: client })
console.log(await res2.body.json())
```

metcoder95:
## This relates to...

- H2 Support #2061
- Fixes #2257

## Rationale

- Fix H2 connection with empty body.
A method such as `GET` or `HEAD` is not meant to expect a body to be sent. For instance, the `writable` side of the duplex stream can be closed by default.
Not doing so, hangs the request as the server is expecting to finish delivering a body for the request that was not meant to be delivered in the first place.

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -320,10 +320,12 @@ class Request {
 
   static [kHTTP2CopyHeaders] (raw) {
     const rawHeaders = raw.split('\r\n')
-
     const headers = {}
+
     for (const header of rawHeaders) {
       const [key, value] = header.split(': ')
+
+      if (value == null || value.length === 0) continue
 
       if (headers[key]) headers[key] += `,${value}`
       else headers[key] = value
--- a/lib/client.js
+++ b/lib/client.js
@@ -1762,17 +1762,21 @@ function writeH2 (client, session, request) {
 
   session.ref()
 
+  const shouldEndStream = method === 'GET' || method === 'HEAD'
   if (expectContinue) {
     headers[HTTP2_HEADER_EXPECT] = '100-continue'
     /**
      * @type {import('node:http2').ClientHttp2Stream}
      */
-    stream = session.request(headers, { signal })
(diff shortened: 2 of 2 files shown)


lib/core/request.js
```diff
@@ -320,11 +320,13 @@ class Request {
 
   static [kHTTP2CopyHeaders] (raw) {
     const rawHeaders = raw.split('\r\n')
-
     const headers = {}
+
     for (const header of rawHeaders) {
       const [key, value] = header.split(': ')
 
+      if (value == null || value.length === 0) continue
```
KhafraDev, line 328: why is this needed?

metcoder95: This is a small fix, basically if the content of the `headers` was an empty string, the object was initialized with `{ '': undefined }`

KhafraDev: I don't understand why the header would ever be an empty string

ShenHongFei (approved):
I tried these modifications. Before the modification, I used request to send a GET request, and after enabling allowH2, I always waited and never returned. After the modification, it returned normally.

ShenHongFei:
fetch with connections: null | > 1 still not use h2. see https://github.com/nodejs/undici/issues/2257#issuecomment-1712554669

PR #2265, opened 2023-09-15
ci: remove duplicated runs

metcoder95:
## This relates to...

- Duplicated CI runs

## Rationale

I've noticed that the CI runs twice for each PR. Once following the `push` event and once following the `pull_request` event. We can scope the `push` to only `main`, `next`, and `v*` branches, meanwhile keeping `pull_request` events open.

image

## Changes

- Remove duplicated CI triggers

## Status

- [ ] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -4,7 +4,13 @@
 
 name: Node CI
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+     - current
+     - next
+     - 'v*'
+  pull_request:
 
 jobs:
   build:


.github/workflows/nodejs.yml
```diff
@@ -4,7 +4,13 @@
 
 name: Node CI
 
-on: [push, pull_request]
+on:
```
metcoder95, line 7: Should we also consider the same for other workflows?

mcollina (approved):
let's try this

PR #2272, opened 2023-09-19
fix undici when node is built with --without-ssl

KhafraDev:
Refs: https://github.com/nodejs/node/pull/49559#issuecomment-1712238921

tested and confirmed undici can be required


mcollina (approved):
lgtm

metcoder95 (commented):
Late to the party, but LGTM 👍

PR #2279, opened 2023-09-21
fix: fix request with readable mode is object

killagu:
In the stream Readable.from(buf) state, the length of 1 refers to the number of objects, not the length of buf. Therefore, when the stream mode is set to 'object,' should not use the length from the state.

### Bug Fixes

In the stream Readable.from(buf) state, the length
of 1 refers to the number of objects, not the length of
buf. Therefore, when the stream mode is set to 'object,'
should not use the length from the state.

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/core/util.js
+++ b/lib/core/util.js
@@ -168,7 +168,7 @@ function bodyLength (body) {
     return 0
   } else if (isStream(body)) {
     const state = body._readableState
-    return state && state.ended === true && Number.isFinite(state.length)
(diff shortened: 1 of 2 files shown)


fengmk2 (approved):
LGTM

ronag (approved):

Ethan-Arrowood (approved):

PR #2281, opened 2023-09-22
feat(#2264): Expose Retry Handler

### LINKED ISSUE #2264: Retry
Similar to https://github.com/nodejs/undici/issues/2256

Sharing some internal code we have been using in case someone wants to work on it and add it as a util in undici.

```js
const stream = require('node:stream')
const { isReadableNodeStream } = require('../stream')
const { parseHeaders } = require('../http')
const createError = require('http-errors')

function parseRange(range) {
  const m = range ? range.match(/^bytes=(\d+)-(\d+)?$/) : null
  return m ? { start: parseInt(m[1]), end: m[2] ? parseInt(m[2]) : null } : null
}

module.exports = class RetryHandler {
  /** @type {Object} */ opts
  /** @type {Object} */ dispatcher
  /** @type {Error | undefined} */ retryError
  /** @type {Number} */ retryCount = 0
  /** @type {Number} */ retryMax
  /** @type {Array<Number>} */ retryCode
  /** @type {Array<String>} */ retryMessage
  /** @type {Array<Number>} */ retryStatus
  /** @type {Array<String>} */ retryMethod
  /** @type {Boolean} */ idempotent
  /** @type {Number} */ position = 0
  /** @type {NodeJS.Timeout | null} */ timeout
  /** @type {{start: Number, end: Number | null} | null} */ range

  constructor (opts, { dispatcher, handler }) {
    const range = opts.headers?.range ?? opts.headers?.range

    this.dispatcher = dispatcher
    this.handler = handler
    this.opts = opts
    this.statusCode = 0
    this.idempotent = opts.idempotent

    const {
      count: retryMax = 8,
      method: retryMethod = ['GET', 'HEAD', 'OPTIONS', 'PUT', 'DELETE', 'TRACE', 'PATCH'],
      status: retryStatus = [420, 429, 502, 503, 504],
      code: retryCode = [
        'ECONNRESET',
        'ECONNREFUSED',
        'ENOTFOUND',
        'ENETDOWN',
        'ENETUNREACH',
        'EHOSTDOWN',
        'EHOSTUNREACH',
        'EPIPE',
      ],
      message: retryMessage = ['other side closed'],
    } = opts.retry ?? {}

    this.retryMax = retryMax
    this.retryCount = retryCode
    this.retryMessage = retryMessage
    this.retryStatus = retryStatus
    this.retryMethod = retryMethod

    this.range = opts.method === 'GET'
      ? range
        ? parseRange(range)
        : { start: 0, end: null }
      : null
  }

  onConnect(abort) {
    this.handler.onConnect(err => {
      if (this.timeout) {
        clearTimeout(this.timeout)
      }
      abort(err)
    })
  }

  onHeaders(statusCode, rawHeaders, resume, statusMessage) {
    if (statusCode >= 300) {
      throw createError(statusCode, { headers: parseHeaders(rawHeaders) })
    }

    let contentLength
    for (let i = 0; i < rawHeaders.length; i += 2) {
      const key = rawHeaders[i].toString()
      if (key.length === 'content-length'.length && key.toLowerCase() === 'content-length') {
        contentLength = rawHeaders[key]
      }
    }

    if (statusCode === 200 && this.range?.end === null && contentLength) {
      this.range.end = Number(contentLength)
      assert(Number.isFinite(this.range.end), 'invalid content-length')
    }

    if (this.statusCode) {
      assert(this.retryError)

      if (statusCode === 200) {
        // TODO (feat): Resume 200 if distance is not too far?
        throw this.retryError
      } else if (statusCode === 206) {
        // TODO (fix): Check content-range.
        return true // TODO (fix): return false if body is full...
      } else {
        throw this.retryError
      }
    } else {
      this.statusCode = statusCode
      return this.handler.onHeaders(statusCode, rawHeaders, resume, statusMessage)
    }
  }

  onData(chunk) {
    this.position += chunk.length
  }

  onComplete(rawTrailers) {
    return this.handler.onComplete(rawTrailers)
  }

  onError(err) {
    if (this.statusCode) {
      if (this.range && this.retryCount < this.retryMax && this.opts.methid === 'GET') {
        // this.logger?.warn({ err, retryCount: this.retryCount }, 'upstream response retrying')

        this.retryError = err
        this.retryCount += 1

        this.dispatcher.dispatch(
          {
            ...this.opts,
            headers: {
              ...this.opts.headers,
              range: `bytes=${this.range.start + this.position}-${
                this.range.end ? this.range.end : ''
              }`,
            },
          },
          this,
        )
      } else {
        return this.handler.onError(err)
      }
    } else if (
      err.name !== 'AbortError' &&
      this.retryCount < this.retryMax &&
      (this.opts.body == null ||
        typeof this.opts.body === 'string' ||
        Buffer.isBuffer(this.opts.body) ||
        // @ts-ignore: isDisturbed is not in typedefs
        (isReadableNodeStream(this.opts.body) && !stream.isDisturbed(this.opts.body))) &&
      (this.idempotent || this.retryMethod.includes(this.opts.method)) &&
      (this.retryCode.includes(err.code) ||
        this.retryMessage.includes(err.message) ||
        this.retryStatus.includes(err.statusCode))
    ) {
      const delay =
        parseInt(err.headers?.['Retry-After']) * 1e3 ||
        Math.min(10e3, this.retryCount * 1e3 + 1e3)

      // this.logger?.warn({ err, retryCount: this.retryCount, delay }, 'upstream request retrying')

      this.retryError = err
      this.retryCount += 1

      this.timeout = setTimeout(() => {
        this.timeout = null
        this.dispatcher.dispatch(this.opts, this)
      }, delay).unref()
    } else {
      // this.logger?.error({ err }, 'upstream request failed')
      this.handler.onError(err)
    }
  }
}

```

metcoder95:
## This relates to...
- Closes #2264

## Changes

- [x] Retry Agent
- [x] Documentation
- [x] Testing

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready


ronag:
Unfortunately I think retry is more complicated than this. See:

- https://github.com/nxtedition/nxt-lib/blob/e3ac11edbd265838cce537b0f4ab8ea298588f79/undici/interceptor/response-body-retry.js
- https://github.com/nxtedition/nxt-lib/blob/e3ac11edbd265838cce537b0f4ab8ea298588f79/undici/interceptor/response-retry.js
- https://github.com/nxtedition/nxt-lib/blob/e3ac11edbd265838cce537b0f4ab8ea298588f79/undici/interceptor/response-status-retry.js

lib/retry.js
```diff
@@ -0,0 +1,175 @@
...
+    // TODO: improve the erro handler for retry
+    if (
+      this.retryAfter == null ||
+      this.aborted ||
+      isDisturbed(this.opts.body)
+    ) {
+      return this.handler.onError(err)
+    }
+
+    this.timeout = setTimeout(() => {
```
ronag, line 164: This is only valid if we errored before forwarding any headers or body,

metcoder95: Or failed while connecting to the origin

metcoder95:
Hey @ronag! Yes, I’m basing most of this work on your implementations, I just the work and opened the draft so I do not miss it and not make a mess with my local branches nor `stash` when switching between them. I’ll ping you once more when I have something more solid 😉

lib/retry.js
```diff
@@ -0,0 +1,175 @@
+const assert = require('node:assert')
+
+const createError = require('http-errors')
```
KhafraDev, line 3: not a fan, this adds 6 dependencies

metcoder95: Yeah, agree. I’d better create an util or something 👍

kibertoad:
[mockttp](https://www.npmjs.com/package/mockttp) provides a very convenient way for testing weirder edge cases, such as connection lost or connection cancelled, highly recommend it.

You can look up the code examples here: https://github.com/kibertoad/undici-retry/blob/main/test/undiciRetry.spec.ts

lib/handler/RetryHandler.js
```diff
@@ -0,0 +1,324 @@
...
+      maxTimeout: maxTimeout ?? 30 * 1000, // 30s,
+      timeout: minTimeout ?? 500, // .5s
+      timeoutFactor: timeoutFactor ?? 2,
+      max: max ?? 8,
+      // What errors we should retry
+      methods: methods ?? ['GET', 'HEAD', 'OPTIONS', 'PUT', 'DELETE', 'TRACE'],
+      // States weather or not retry on idempotent methods
+      idempotent: idempotent ?? false,
+      // Indicates which errors to retry
+      status: statusCodes ?? [500, 502, 503, 504],
```
kibertoad, line 44: why no 429?

metcoder95: I haven't implemented that one yet; as its nature is coming from a rate-limit logic, we need to handle its `Retry-After` header; I'm planning to add settings specifically to it and its own default logic

metcoder95:
More testing and docs/types is missed, but should be good already for review 👍 

cc: @ronag

ronag (commented):
Test and docs. LGTM otherwise.

ronag (approved):

lib/handler/RetryHandler.js
```diff
@@ -0,0 +1,345 @@
...
+      this.resume = null
+
+      if (statusCode !== 206) {
+        return true
+      }
+
+      // TODO (fix): etag
+      // if (this.etag == null || this.etag !== etag) {
+      //   throw this.error
+      // }
```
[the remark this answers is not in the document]
metcoder95: Missing feature, good catch; let me add it as well

lib/handler/RetryHandler.js
```diff
@@ -0,0 +1,347 @@
...
+      if (statusCode !== 206) {
+        return true
+      }
+
+      let contentRange
+      if (
+        // Let's start with a weak etag check
+        (this.etag != null && this.etag !== headers.etag) ||
+        // If no content range
+        !(contentRange = parseRangeHeader(headers['content-range']))
```
ronag, line 191: This is a bit too much syntax magic

lib/handler/RetryHandler.js
```diff
@@ -0,0 +1,347 @@
...
+        'ENOTFOUND',
+        'ENETDOWN',
+        'ENETUNREACH',
+        'EHOSTDOWN',
+        'EHOSTUNREACH',
+        'EPIPE'
+      ]
+    }
+
+    // TODO: hidde behind a symbol
```
ronag, line 67: Just do it :)

lib/handler/RetryHandler.js
```diff
@@ -0,0 +1,347 @@
...
+  onHeaders (statusCode, rawHeaders, resume, statusMessage) {
+    const headers = parseHeaders(rawHeaders)
+    const err = new RequestRetryError(
+      'Request failed after retry',
+      statusCode,
+      {
+        headers,
+        count: this.count
+      }
+    )
```
ronag, line 169: Only create error if failed?

lib/handler/RetryHandler.js
```diff
@@ -0,0 +1,347 @@
...
+
+    this.count += 1
+
+    if (statusCode > 299) {
```
ronag, line 173:
```suggestion
    if (statusCode >= 300) {
```

lib/handler/RetryHandler.js
```diff
@@ -0,0 +1,347 @@
...
+      )
+    }
+
+    // Meant to just indicate weather or not trigger a retry
+    // It should return null to indicate we exhausted the retries
+    const retryAfter = this.retryOpts.retry(
+      err,
+      { counter: this.count, currentTimeout: this.retryAfter },
+      { retry: this.retryOpts, ...this.opts }
+    )
```
ronag, line 265: I would like the retry method to be asynchronous and the timeout lives inside of it, i.e. instead of returning a duration.

ronag: Either by returning a promise or taking a callback

lib/handler/RetryHandler.js
```diff
@@ -0,0 +1,347 @@
...
+        'EPIPE'
+      ]
+    }
+
+    // TODO: hidde behind a symbol
+    this.count = 0
+    this.timeout = null
+    this.retryAfter = null
+    this.head = 0
+    this.tail = null
```
ronag, line 72: start and end is a bit clearer? Took me a bit to figure out what these were for. Maps better to existing vocabulary in regards to ranges.

types/retry-handler.d.ts
```diff
@@ -0,0 +1,109 @@
...
+     * It enables automatic retry between requets on idempotent methods.
+     * The method should be defined as safe accordingly to https://developer.mozilla.org/en-US/docs/Glossary/Safe/HTTP
+     *
+     * Note: It takes presedence over the `methods` option.
+     * 
+     * @type {boolean}
+     * @memberof RetryOptions
+     * @default false
+     */
+    idempotent?: boolean;
```
ronag, line 89: Shouldn't this be true by default?

types/retry-handler.d.ts
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * Error codes to be retried. e.g. `ECONNRESET`, `ENOTFOUND`, `ETIMEDOUT`, `ECONNREFUSED`, etc.
+     *
+     * @type {string[]}
+     */
+    codes?: string[];
```
ronag, line 95: Missing default?

types/retry-handler.d.ts
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * HTTP status codes to be retried.
+     *
+     * @type {number[]}
+     * @memberof RetryOptions
+     */
+    status?: number[];
```
ronag, line 102: Missing default?

types/retry-handler.d.ts
```diff
@@ -0,0 +1,109 @@
...
+     * @default true
+     */
+    retryAfter?: boolean;
+    /**
+     * HTTP methods to retry.
+     *
+     * @type {Dispatcher.HttpMethod[]}
+     * @memberof RetryOptions
+     */
+    methods?: Dispatcher.HttpMethod[];
```
ronag, line 78: Missing default?

types/retry-handler.d.ts
```diff
@@ -0,0 +1,109 @@
...
+     */
+    retry?: RetryCallback;
+    /**
+     * Maximum number of retries to allow.
+     *
+     * @type {number}
+     * @memberof RetryOptions
+     * @default 5
+     */
+    max?: number;
```
ronag, line 39:
```suggestion
    maxRetries?: number;
```

types/retry-handler.d.ts
```diff
@@ -0,0 +1,109 @@
...
+  export interface RetryOptions {
+    /**
+     * Callback to be invoked on every retry iteration.
+     * It receives the error, current state of the retry object and the options object
+     * passed when instantiating the retry handler.
+     *
+     * @type {RetryCallback}
+     * @memberof RetryOptions
+     */
+    retry?: RetryCallback;
```
ronag, line 31: All the options below are only relevant if `retry` is null. Maybe this is better represented as an union?

lib/handler/RetryHandler.js
```diff
@@ -0,0 +1,347 @@
...
+
+function calculateRetryAfterHeader (retryAfter) {
+  const current = Date.now()
+  const diff = new Date(retryAfter).getTime() - current
+
+  return diff
+}
+
+class RetryHandler {
+  constructor (opts, handlers, retryOpts) {
```
ronag, line 20: I would expect things to live inside opts, i.e. no retryOpts.

```suggestion
  constructor ({ retry: retryOpts, ...opts }, handlers) {
    if (typeof retryOpts === 'function') {
      retryOpts = { retry: retryOpts }
    } else if (typeof retryOpts === 'number') {
      retryOpts = { maxRetries: retryOpts }
    }
  }
```

ronag: Of course this has the downside of potential conflict with other handlers... so maybe a separate handler opts is not bad either.

lib/handler/RetryHandler.js
metcoder95: haha, left over, I forgot about that 😅

lib/handler/RetryHandler.js
metcoder95: Hmm, has sense; it can enable several use cases

I'm imagine something like
```suggestion
    const retryAfter = this.retryOpts.retry(
      err,
      { counter: this.count, currentTimeout: this.retryAfter },
      { retry: this.retryOpts, ...this.opts }
    ).then(timeout => { /* timeout goes here */}, err => this.handler.onError(er))
```

Thoughts?

lib/handler/RetryHandler.js
metcoder95: SGTM
I don't recall exactly why I went in that direction, I kind of recall because wanted to handle those two differently; but yeah your suggestion goes better 👍

types/retry-handler.d.ts
metcoder95: Have some conflicts deciding that; especially as the `methods` is also enabled by default, so it might make sense have one of them enabled. Maybe push for `idempotent` and disabled `methods`?

types/retry-handler.d.ts
metcoder95: Had hard times coping with adding the array for the JSDoc 😅 
But let me try once more

lib/handler/RetryHandler.js
metcoder95: TBH, had the same thoughts as you, but the conflict won on my side so I decided to go with the weird opts. I'd prefer it to keep it there but maybe change the order first `opts` and later `handlers`?

Just wanted to avoid collisions, specially as the `opts` are reused between retries for the `dispatch`

lib/handler/RetryHandler.js
metcoder95: Decided to not do it, do not see much value on doing so; implementation cannot access it, resolving for now

types/retry-handler.d.ts
metcoder95: Kind of, they `retryOpts` are also passed down to the custom `retry` fn; the goal was to allow them also make use of them if they want, so the settings they set, can also be consumed by the same implementation; I can see the potential but do not have strong opinion wether to keep it or not, thoughts?

lib/handler/RetryHandler.js
ronag: ACtually you don't need it to return any value, when the promise finishes then that's the retry.

metcoder95: Yeah, I was actually about to ask you the same; basically the promise handles the retry; after it resolves we call `this.dispatch`, otherwise we handle the error

ronag: TBH I would probably prefer a callback api, rather than promises, we don't use so much promises in our core code.

metcoder95: Yeah, SGTM. Found some hiccups while refactoring to callback, I'll go through them later on but so far so good 👍

lib/handler/RetryHandler.js
ronag: Would you be fine with my suggestion?

types/retry-handler.d.ts
ronag: Hm, since there is overlap, maybe we can make it so methods is either an array of methods or the string `'idempotent'`?

ronag: Not sure, maybe easiest to just remove this option?

lib/handler/RetryHandler.js
```diff
@@ -0,0 +1,347 @@
...
+
+  onBodySent (chunk) {
+    return this.handler.onBodySent(chunk)
+  }
+
+  static [kRetryHandlerDefaultRetry] (err, { counter, currentTimeout }, opts) {
+    const { statusCode, code, headers } = err
+    const { method } = opts
+    const {
+      max,
```
ronag, line 103:
```suggestion
      maxRetries,
```

lib/handler/RetryHandler.js
```diff
@@ -0,0 +1,347 @@
...
+      timeout,
+      maxTimeout,
+      timeoutFactor,
+      status,
```
ronag, line 107:
```suggestion
      statusCodes,
```

lib/handler/RetryHandler.js
```diff
@@ -0,0 +1,347 @@
...
+      idempotent,
+      codes,
```
ronag, line 109:
```suggestion
      errorCodes,
```

lib/handler/RetryHandler.js
```diff
@@ -0,0 +1,347 @@
...
+  static [kRetryHandlerDefaultRetry] (err, { counter, currentTimeout }, opts) {
+    const { statusCode, code, headers } = err
+    const { method } = opts
+    const {
+      max,
+      timeout,
+      maxTimeout,
+      timeoutFactor,
+      status,
+      idempotent,
```
ronag, line 108:
```suggestion
```

lib/handler/RetryHandler.js
```diff
@@ -0,0 +1,347 @@
...
+    const err = new RequestRetryError(
+      'Request failed after retry',
+      statusCode,
+      {
+        headers,
+        count: this.count
+      }
+    )
+
+    this.count += 1
```
ronag, line 171:
```suggestion
    this.retryCount += 1
```

lib/handler/RetryHandler.js
```diff
@@ -0,0 +1,347 @@
...
+        !contentRange
+      ) {
+        this.abort(err)
+        return false
+      }
+
+      const { start, size, end = size } = contentRange
+
+      assert(this.head === start, 'content-range mismatch')
+      assert(this.tail == null || this.tail === end, 'content-range mismatch')
```
ronag, line 200:
```suggestion
      assert(this.start === start, 'content-range mismatch')
      assert(this.end == null || this.end === end, 'content-range mismatch')
```

lib/handler/RetryHandler.js
metcoder95: Sure thing 👍

lib/handler/RetryHandler.js
metcoder95: Weird, I remember doing those changes, maybe never pushed? 🤔

metcoder95: Yeah, never pushed

types/retry-handler.d.ts
metcoder95: I'd say we better remove it, it will make our lifes and implementers easier. Anyways, if they want to automatically retry on certain method, they can set it in the `method` array

metcoder95:
It seems its done, @ronag can you take a look?
Specially at the retry async function. I have some doubts about the signature, and possibly the handling. Open to suggestions 👍

docs/api/RetryHandler.md
```diff
@@ -0,0 +1,106 @@
...
+
+Returns: `retryHandler`
+
+### Parameter: `Dispatch.DispatchOptions & RetryOptions`
+
+Extends: [`Dispatch.DispatchOptions`](Dispatcher.md#parameter-dispatchoptions).
+
+#### `RetryOptions`
+
+- **retry** `(err: Error, context: RetryContext, done: (result?: boolean) => void) => void` (optional) - Function to be called after every retry. It should call the `done` parameter to try the request once more.
```
ronag, line 22:
```suggestion
- **retry** `(err: Error, context: RetryContext, done: () => void) => void` (optional) - Function to be called after every retry. It should re-throw the error if no more retries should be performed.
```

docs/api/RetryHandler.md
```diff
@@ -0,0 +1,106 @@
...
+    retryOptions: {
+      // custom retry function
+      retry: function (err, state, done) {
+        counter++;
+
+        if (err.code && err.code === "UND_ERR_DESTROYED") {
+          return null;
+        }
+
+        if (err.statusCode === 206) return done(false);
```
ronag, line 61:
```suggestion
        if (err.statusCode === 206) throw err;
```

docs/api/RetryHandler.md
```diff
@@ -0,0 +1,106 @@
...
+const handler = new RetryHandler(
+  {
+    ...dispatchOptions,
+    retryOptions: {
+      // custom retry function
+      retry: function (err, state, done) {
+        counter++;
+
+        if (err.code && err.code === "UND_ERR_DESTROYED") {
+          return null;
```
ronag, line 58:
```suggestion
          throw err;
```

lib/handler/RetryHandler.js
```diff
@@ -0,0 +1,344 @@
...
+      code &&
+      code !== 'UND_ERR_REQ_RETRY' &&
+      code !== 'UND_ERR_SOCKET' &&
+      !errorCodes.includes(code)
+    ) {
+      return done(false)
+    }
+
+    // If a set of method are provided and the current method is not in the list
+    if (Array.isArray(methods) && !methods.includes(method)) return done(false)
```
ronag, line 129:
```suggestion
    if (Array.isArray(methods) && !methods.includes(method)) {
      return done(false)
    }
```

I would hope we don't have these one-liners

lib/handler/RetryHandler.js
```diff
@@ -0,0 +1,344 @@
...
+
+    // If a set of status code are provided and the current status code is not in the list
+    if (
+      statusCode != null &&
+      Array.isArray(statusCodes) &&
+      !statusCodes.includes(statusCode)
+    ) {
+      return done(false)
```
ronag, line 137:
```suggestion
      throw err
```

docs/api/RetryHandler.md
```diff
@@ -0,0 +1,106 @@
...
+
+```js
+const client = new Client(`http://localhost:${server.address().port}`);
+const chunks = [];
+const handler = new RetryHandler(
+  {
+    ...dispatchOptions,
+    retryOptions: {
+      // custom retry function
+      retry: function (err, state, done) {
```
ronag, line 54:
```suggestion
      retry: function (err, state, callback) {
```

docs/api/RetryHandler.md
```diff
@@ -0,0 +1,106 @@
...
+        counter++;
+
+        if (err.code && err.code === "UND_ERR_DESTROYED") {
+          return null;
+        }
+
+        if (err.statusCode === 206) return done(false);
+        
+        setTimeout(done, 800);
```
ronag, line 63:
```suggestion
        setTimeout(() => done(null), 800);
```

lib/handler/RetryHandler.js
```diff
@@ -0,0 +1,344 @@
...
+      currentTimeout != null && currentTimeout > 0 ? currentTimeout : timeout
+
+    // Any code that is not a Undici's originated and allowed to retry
+    if (
+      code &&
+      code !== 'UND_ERR_REQ_RETRY' &&
+      code !== 'UND_ERR_SOCKET' &&
+      !errorCodes.includes(code)
+    ) {
+      return done(false)
```
ronag, line 125:
```suggestion
      return done(err)
```

lib/handler/RetryHandler.js
```diff
@@ -0,0 +1,344 @@
...
+    if (
+      statusCode != null &&
+      Array.isArray(statusCodes) &&
+      !statusCodes.includes(statusCode)
+    ) {
+      return done(false)
+    }
+
+    // If we reached the max number of retries
+    if (counter > maxRetries) return done(false)
```
ronag, line 141:
```suggestion
    if (counter > maxRetries) return done(err)
```

lib/handler/RetryHandler.js
```diff
@@ -0,0 +1,344 @@
...
+    }
+
+    const retryTimeout =
+      retryAfterHeader > 0
+        ? Math.min(retryAfterHeader, maxTimeout)
+        : Math.min(currentTimeout * timeoutFactor ** counter, maxTimeout)
+
+    state.currentTimeout = retryTimeout
+
+    setTimeout(done, retryTimeout)
```
ronag, line 158:
```suggestion
    setTimeout(() => done(null), retryTimeout)
```

lib/handler/RetryHandler.js
```diff
@@ -0,0 +1,344 @@
...
+
+      const contentRange = parseRangeHeader(headers['content-range'])
+      if (
+        // Let's start with a weak etag check
+        (this.etag != null && this.etag !== headers.etag) ||
+        // If no content range
+        !contentRange
+      ) {
+        this.abort(
+          new RequestRetryError('Content-Range mismatch', statusCode, {
```
ronag, line 192: Separate error for etag mismatch?

lib/handler/RetryHandler.js
```diff
@@ -0,0 +1,344 @@
...
+    this.retryOpts.retry(
+      err,
+      {
+        state: { counter: this.retryCount++, currentTimeout: this.retryAfter },
+        opts: { retryOptions: this.retryOpts, ...this.opts }
+      },
+      onRetry.bind(this)
+    )
+
+    function onRetry (next) {
```
ronag, line 320:
```suggestion
    function onRetry (err) {
```

lib/handler/RetryHandler.js
```diff
@@ -0,0 +1,344 @@
...
+      if (next === false || this.aborted || isDisturbed(this.opts.body)) {
+        return this.handler.onError(err)
+      }
```
ronag, line 323:
```suggestion
      if (err || this.aborted || isDisturbed(this.opts.body)) {
        return this.handler.onError(err)
      }
```

metcoder95:
Applied on [1ed45f3](https://github.com/nodejs/undici/pull/2281/commits/1ed45f37b0cae527d49a86bba14407bb46ed6f7f) @ronag  👍

ronag:
@metcoder95 are we ready here?

metcoder95:
Yes 👍

Ethan-Arrowood:
Nice!

jiyuu-jin:
Is this exposed anywhere in the current NodeJS version?

metcoder95:
This is exposed only through `undici`

PR #2282, opened 2023-09-22
fix loading websockets when node is built w/ --without-ssl

KhafraDev:
I didn't catch this in my other PR because the main entry point (index.js) only exports WebSocket if crypto is available. The fetch bundle doesn't do this, which causes issues when node tries to require undici in an environment that doesn't have crypto.

Refs: https://ci.nodejs.org/job/node-test-commit-linux-containered/nodes=ubuntu1804_sharedlibs_withoutssl_x64/39416/

### DIFF
--- a/lib/websocket/connection.js
+++ b/lib/websocket/connection.js
@@ -1,6 +1,5 @@
 'use strict'
 
-const { randomBytes, createHash } = require('crypto')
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

PR #2285, opened 2023-09-24
fix: don't call onConnect multiple times for sub handler

ronag:
Many sub-handlers (e.g. api-request) don't expect onConnect (or any other handler other than onData or onBodySent) to be called multiple times.

I'm not sure why/how we haven't caught this earlier. Should fail at https://github.com/nodejs/undici/blob/4013c4b8932e73728809e4106d5c9d9d40648031/lib/api/api-request.js#L73

### DIFF
--- a/lib/handler/RedirectHandler.js
+++ b/lib/handler/RedirectHandler.js
@@ -34,6 +34,8 @@ class RedirectHandler {
     this.dispatch = dispatch
     this.location = null
     this.abort = null
+    this.aborted = false
+    this.reason = null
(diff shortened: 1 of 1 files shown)


KhafraDev (approved):

metcoder95 (approved):
Interesting, never noticed that. It might be worth it to check out the other handlers. I imagine similar patterns might be the cause

The PR LGTM 👍

mcollina (commented):
Can you add a test?

PR #2286, opened 2023-09-24
perf: improve parse-url implementation

anonrig:
Reduced the number of checks for where we know the variable is an instance of `URL`.

### DIFF
--- a/lib/core/util.js
+++ b/lib/core/util.js
@@ -58,31 +58,31 @@ function parseURL (url) {
     throw new InvalidArgumentError('Invalid URL: The URL argument must be a non-null object.')
   }
 
-  if (url.port != null && url.port !== '' && !Number.isFinite(parseInt(url.port))) {
-    throw new InvalidArgumentError('Invalid URL: port must be a valid integer or a string representation of an integer.')
-  }
-
-  if (url.path != null && typeof url.path !== 'string') {
-    throw new InvalidArgumentError('Invalid URL path: the path must be a string or null/undefined.')
-  }
-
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

Ethan-Arrowood (approved):

lib/core/util.js
```diff
@@ -58,31 +58,31 @@ function parseURL (url) {
...
-  if (url.port != null && url.port !== '' && !Number.isFinite(parseInt(url.port))) {
-    throw new InvalidArgumentError('Invalid URL: port must be a valid integer or a string representation of an integer.')
+  if (!/^https?:/.test(url.origin || url.protocol)) {
+    throw new InvalidArgumentError('Invalid URL protocol: the URL must start with `http:` or `https:`.')
   }
 
-  if (url.path != null && typeof url.path !== 'string') {
-    throw new InvalidArgumentError('Invalid URL path: the path must be a string or null/undefined.')
-  }
+  if (!(url instanceof URL)) {
```
metcoder95, line 65: Small, `nit`.

Does `url.constructor` is slower compared to `instanceof`?

Also, `instance tampering` is possible when using `instanceof`, where using the `well-known` symbol can allow non-valid instances of URL to return `true`.

Just wanted to point that out, not a blocker 🙂

metcoder95 (approved):

lib/core/util.js
anonrig: Compared to new URL which takes around 300-400ns, I think constructor vs instanceof is negligible

PR #2298, opened 2023-10-01
Disable finalizationregistry if node code cov

mcollina:
See https://github.com/nodejs/node/issues/49344#issuecomment-1741776308 for more details, this has been the source of some of the flakiness we have seen.

### DIFF
--- a/lib/compat/dispatcher-weakref.js
+++ b/lib/compat/dispatcher-weakref.js
@@ -31,6 +31,14 @@ class CompatFinalizer {
 }
 
 module.exports = function () {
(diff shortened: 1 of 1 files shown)


types/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "undici-types",
-  "version": "5.25.1",
+  "version": "5.25.2",
```
mcollina, line 3: I don't understand why I have a commit with this files in.

KhafraDev (approved):

Issue #2046, opened 2023-04-04
Sending a DELETE request with "Content-Length: 0" fails with a RequestContentLengthMismatchError


ouuan:
## Bug Description

Sending a DELETE request with "content-length: 0" fails with a RequestContentLengthMismatchError.

The specification says "A user agent SHOULD NOT send a Content-Length header field when the request message does not contain a payload body and the method semantics do not anticipate such a body", but according to [RFC 2119](https://www.rfc-editor.org/rfc/rfc2119), "SHOULD NOT" means "NOT RECOMMENDED", so it's not reasonable to fail a request for this reason.

Even if the Content-Length header "MUST NOT" be sent, the current implementation would still be problematic:

1. A better behavior could be removing the header instead of failing the request.
2. The request only fails when Content-Length is set to zero, but not when it is not zero and matches the length of the body. Note that the DELETE method does not anticipate a body.
3. The error message is unhelpful. It says "Request body length does not match content-length header" instead of "Content-Length should not be set for the DELETE method".

## Reproducible By

```javascript
fetch('https://example.com', { method: 'DELETE', headers: { 'content-length': '0' } })
```

## Expected Behavior

The request succeeds, either as it is or with the Content-Length header removed.

## Logs & Screenshots

```
> Uncaught TypeError: fetch failed
    at Object.fetch (node:internal/deps/undici/undici:11413:11) {
  cause: RequestContentLengthMismatchError: Request body length does not match content-length header
      at write (node:internal/deps/undici/undici:9907:41)
      at _resume (node:internal/deps/undici/undici:9885:33)
      at resume (node:internal/deps/undici/undici:9787:7)
      at connect (node:internal/deps/undici/undici:9776:7) {
    code: 'UND_ERR_REQ_CONTENT_LENGTH_MISMATCH'
  }
}
```

## Environment

Reproducible on both Node v18.15.0 and v19.8.1

## Additional Context

Related codes:

https://github.com/nodejs/undici/blob/5f3b8e16a9aa33fe72d640fe271d8e7a5748321a/lib/client.js#L1333-L1366

KhafraDev:
RFC 2119 also says about "SHOULD NOT":
```
   there may exist valid reasons in particular circumstances when the
   particular behavior is acceptable or even useful, but the full
   implications should be understood and the case carefully weighed
   before implementing any behavior described with this label.
````

Do you have a use case that justifies this?

ronag:
PR welcome

[...]

PR #2305, opened 2023-10-03
fix: define conditions when content-length should be sent 

pxue:
Fixes: https://github.com/nodejs/undici/issues/2046


lib/client.js
```diff
@@ -1538,7 +1538,12 @@ function write (client, request) {
     contentLength = null
   }
 
-  if (request.contentLength !== null && request.contentLength !== contentLength) {
+  // https://github.com/nodejs/undici/issues/2046
+  // A user agent may send a Content-Length header with 0 value, this should be
+  // allowed.
+  const bypassDelete = method === 'DELETE' && contentLength === null && request.contentLength === 0
+
+  if (!bypassDelete && request.contentLength !== null && request.contentLength !== contentLength) {
```
metcoder95, line 1546: Two things:

1. We need to replicate this to `writeH2` otherwise will lead to inconsistencies
2. I have mixed feelings about doing this **only for DELETE** as does feel like a one-time-only fix; overall several methods allow an empty body or completely forbid it. I'm open to relaxing it for those payloads that do not expect a body or the body is optional. Should we consider a better strategy for these scenarios?

cc: @mcollina @ronag @KhafraDev

ronag: Yea. I don't like this approach either.

KhafraDev: would it make more sense not to remove the content-length header if it's 0?

KhafraDev:
```
   A user agent SHOULD NOT send a
   Content-Length header field when the request message does not contain
   a payload body and the method semantics do not anticipate such a
   body.
```

I think it's fine for DELETE only (or maybe re-evaluate which methods this should apply to, because the RFC doesn't forbid DELETE requests from having a body).

```
content received in a DELETE request has no generally defined semantics, cannot alter the meaning or target of the request, and might lead some implementations to reject the request and close the connection because of its potential as a request smuggling attack
```

pxue: FWIW, here's the Go's logic for sending content-length
https://github.com/golang/go/blob/e9379a8f8bd5f00463f67fe7e052cbcb262b169a/src/net/http/transfer.go#L253

metcoder95: >maybe re-evaluate which methods this should apply to, because the RFC doesn't forbid DELETE requests from having a body

I'm pro this approach; I believe is better to reevaluate and relax it for those scenarios where the `payload` is not expected or is optional (e.g. `DELETE` is optional, as well as `GET` and even `HEAD`).

The approach from `Go` doesn't seem bad, it can be a good starting point

pxue: I made one more pass based on your suggestion, @metcoder95 @KhafraDev

lib/client.js
```diff
@@ -1495,6 +1495,23 @@ function _resume (client, sync) {
...
+  if (method === 'POST' || method === 'PUT' || method === 'PATCH') {
+    return true
+  }
+  if (contentLength === 0) {
+    if (method === 'GET' || method === 'HEAD') {
+      return false
+    }
+    return true
+  }
+  return false
```
metcoder95, line 1512:
```suggestion
	return contentLength !== 0 ? method !== 'GET' && method !== 'HEAD' : true
```

We just want to entirely exclude the `GET`, `HEAD`, `TRACE`, `OPTIONS`, and `CONNECT` methods, which do not expect a body. If we see a content length different than 0 and have any method other than them, we should attach it. wdyt?

metcoder95 (commented):
Tests seem failing.
Can we also port this to `writeH2`?

mcollina (approved):
lgtm

lib/client.js
pxue: Agreed here, for future reference, we're interpreting RFC7230 and RFC7231

```
   A payload within a GET request message has no defined semantics;

...

   A user agent SHOULD NOT send a
   Content-Length header field when the request message does not contain
   a payload body and the method semantics do not anticipate such a
   body.

```

KhafraDev (approved):

lib/client.js
```diff
@@ -1495,6 +1495,16 @@ function _resume (client, sync) {
...
 }
 
+function shouldSendContentLength (contentLength, method) {
+  if (contentLength > 0) {
+    if (method === 'POST' || method === 'PUT' || method === 'PATCH') {
+      return true
+    }
+    return false
+  }
+  return (method === 'POST' || method === 'PUT' || method === 'PATCH')
```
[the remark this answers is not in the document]
pxue: You're right!

Is this what we want? We're basically saying all non Post/put/patch methods have no defined semantics for a body.

Therefore we should not send content-length

lib/client.js
```diff
@@ -1538,7 +1548,9 @@ function write (client, request) {
     contentLength = null
   }
 
-  if (request.contentLength !== null && request.contentLength !== contentLength) {
+  // https://github.com/nodejs/undici/issues/2046
+  // A user agent may send a Content-Length header with 0 value, this should be allowed.
+  if (shouldSendContentLength(contentLength, method) && request.contentLength !== null && request.contentLength !== contentLength) {
```
metcoder95, line 1553: Following previous comment, it is better if we offload the `contentLength` to the `if` statement instead. In that sense **if** `contentLength` is greater than `0`, and the `method` has explicit semantics to handle the request's body; we attach it. Otherwise, we just ignore it.

lib/client.js
```diff
@@ -1495,6 +1495,16 @@ function _resume (client, sync) {
   }
 }
 
+function shouldSendContentLength (contentLength, method) {
+  if (contentLength > 0) {
+    if (method === 'POST' || method === 'PUT' || method === 'PATCH') {
```
metcoder95, line 1500: Thinking twice is better if we exclude those who don't explicitly support the request's body `semantics`, so methods like `DELETE`, whose bodies are up to the server to decide what to do with them, can still have a `content-length` header.

pxue: Made the changes, glad we're getting closer.

test/content-length.js
```diff
@@ -329,3 +288,129 @@ test('request streaming with Readable.from(buf)', (t) => {
...
+    client.on('disconnect', () => {
+      t.pass()
+    })
+  })
+})
+
+test('content-length shouldSendContentLength=false', (t) => {
+  t.plan(9)
+  const server = createServer((req, res) => {
+    res.end()
```
metcoder95, line 337: Can we verify that the `content-length` is `undefined` or `0` in any case?

test/content-length.js
```diff
@@ -329,3 +288,129 @@ test('request streaming with Readable.from(buf)', (t) => {
     })
   })
 })
+
+test('request DELETE, content-length=0, with body', (t) => {
+  t.plan(3)
+  const server = createServer((req, res) => {
+    res.end()
```
metcoder95, line 295: Same here

pxue: done!

test/content-length.js
metcoder95: What about the `request` being handled at the server; shouldn't we test that also its `0`?

pxue: i see what you're saying. test added for both this and the DELETE below.

Do we want to strictly NOT send content-length header for GET/HEAD/OPTION requests even if the request contains body? This most likely means we'll have to check `shouldSendContentLength` here as well https://github.com/nodejs/undici/blob/741b8118e1452d9c00358b2c33d7068683e9dcc6/lib/client.js#L1641

metcoder95: Hmm, no you are right that will be a breaking change and we should not need that. Let's go with your suggestion then, of adding an extra guard for the `content-length`, in that way we keep the current semantics and APIs valid 👍

Let's try adding your changes, and see if related; `redirect-pipeline` is flaky, but worth it check if with the changes fixes

pxue: Agreed, i've backed out the last batch of changes.

I think I'm going to leave things as is for backwards compat sake.

From testing it seems like forcing `content-length: 0` on `shouldSendContentLength=false` methods will also force any body written via `socket.write(body)` to be empty, this will cause tests in `test/client-post.js` to fail.

I think this is a decision for another PR? @metcoder95

metcoder95: Yeah, can you open an issue based on that; it will be worth it to chat as it will be a breaking change for sure 👍

pxue: For sure, https://github.com/nodejs/undici/issues/2357

pxue:
@metcoder95 
anything else needed to get this PR across the finish line? I'll have some time this weekend.

lib/client.js
```diff
@@ -1538,7 +1543,9 @@ function write (client, request) {
     contentLength = null
   }
 
-  if (request.contentLength !== null && request.contentLength !== contentLength) {
+  // https://github.com/nodejs/undici/issues/2046
+  // A user agent may send a Content-Length header with 0 value, this should be allowed.
+  if (shouldSendContentLength(method) && contentLength > 0 && request.contentLength !== null && request.contentLength !== contentLength) {
```
metcoder95, line 1548: `nit`:

Let's store it in some var so we avoid call it twice

metcoder95 (approved):
Small `nit` but LGTM 👍

lib/client.js
pxue: I think we're ok since one's in `write` and other in `writeH2`!
if all good, no need to respond. thanks!

metcoder95: Missed that, thanks!

PR #2310, opened 2023-10-04
change default header to `node`

### LINKED ISSUE #2306: Change default "user-agent" from `undici`

Currently, Node.js uses `undici` for its `fetch` implementation. This sets `undici` as the default user agent.

From Node.js' perspective, `undici` is an implementation detail and so it should have its own user agent header.

**For context** at Next.js, we stopped polyfilling `globalThis.fetch` in favor of the platform-provided alternative (previously using `node-fetch`), and started getting reports where users did not know where the requests with the `undici` user agent were coming from.

I've tested the current behavior in different runtimes, here is Node.js, Bun and Deno for comparison:

- Node.js `node node.mjs` prints: undici:
```js
// node.mjs
  console.log(req.headers['user-agent'])
  res.end()
}).listen(3000)
await fetch('http://localhost:3000')
server.close()
process.exit(0)
```

- Bun `bun bun.js` prints: Bun/1.0.2:
```js
// bun.js
const server = Bun.serve({
  port: 3000,
  fetch: (req) => console.log(req.headers.get('user-agent')),
})
await fetch('http://localhost:3000')
server.stop()
```

- Deno `deno run --allow-net deno.js` prints: Deno/1.37.1:
```js
// deno.js
Deno.serve({ port: 3000 }, (request) => {
  console.log(request.headers.get('user-agent'))
  return new Response()
})
await fetch('http://localhost:3000')
Deno.exit(0)
```

I propose that the default user agent in Node.js changes to the format used by Deno, and Bun for consistency.

Example: `Node/18.18.0`

Related: https://github.com/nodejs/node/pull/43852 https://github.com/nodejs/node/issues/43851

Ethan-Arrowood:
Closes #2306

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -1344,7 +1344,7 @@ async function httpNetworkOrCacheFetch (
   //    user agents should append `User-Agent`/default `User-Agent` value to
   //    httpRequest’s header list.
   if (!httpRequest.headersList.contains('user-agent')) {
-    httpRequest.headersList.append('user-agent', 'undici')
+    httpRequest.headersList.append('user-agent', 'node')
   }
 
   //    15. If httpRequest’s cache mode is "default" and httpRequest’s header
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -1344,7 +1344,7 @@ async function httpNetworkOrCacheFetch (
   //    user agents should append `User-Agent`/default `User-Agent` value to
   //    httpRequest’s header list.
   if (!httpRequest.headersList.contains('user-agent')) {
-    httpRequest.headersList.append('user-agent', 'node')
+    httpRequest.headersList.append('user-agent', __filename.endsWith('index.js') ? 'undici' : 'node')
   }
 
   //    15. If httpRequest’s cache mode is "default" and httpRequest’s header
--- a/test/fetch/user-agent.js
+++ b/test/fetch/user-agent.js
@@ -1,0 +1,25 @@
+'use strict'
+
+const { test } = require('tap')
+const events = require('events')
+const http = require('http')
+const undici = require('../../')
+const nodeBuild = require('../../undici-fetch.js')
+
+test('user-agent defaults correctly', async (t) => {
+  const server = http.createServer((req, res) => {
+    res.end(JSON.stringify({ userAgentHeader: req.headers['user-agent'] }))
+  })
+  t.teardown(server.close.bind(server))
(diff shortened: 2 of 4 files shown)


lib/fetch/index.js
```diff
@@ -1344,7 +1344,7 @@ async function httpNetworkOrCacheFetch (
   //    user agents should append `User-Agent`/default `User-Agent` value to
   //    httpRequest’s header list.
   if (!httpRequest.headersList.contains('user-agent')) {
-    httpRequest.headersList.append('user-agent', 'undici')
+    httpRequest.headersList.append('user-agent', 'node')
```
ronag, line 1347: Is there any way we can see whether we are node internal or not? I'm a little off about doing this for the npm package.

ronag (approved):

lib/fetch/index.js
Ethan-Arrowood: I agree. And I couldn't think of anything. Maybe someone else has an idea. Is there potentially some kinda `process.env` value?

KhafraDev: Maybe something like if __filename ends with index-fetch.js?

Ethan-Arrowood: How would we test that? Something in the Node.js side?

Ethan-Arrowood: And is `__filename` guaranteed to always be assigned? Regardless if the user's code is ESM or CJS for example?

KhafraDev: undici-fetch.js is the bundled build of undici that node core uses. I believe there are tests that use it so you'd test the fetch exported from there? I'm not home atm so I can't do any tests myself

Ethan-Arrowood: Looks like we were already building `undici-fetch.js` pre test, but it wasn't obvious another test was using it so I created a new test file.

mcollina (approved):
lgtm

ronag (approved):

KhafraDev:
this is causing test failures in node `__filename is not defined`. I have no good ideas on how to detect if we're in node core.

Ethan-Arrowood:
Why is `__filename` undefined? Is this some kinda esm issue?

Ethan-Arrowood:
Let's chat in the PR I just opened. We can discuss alternatives there. I think maybe just defaulting it when it's undefined is sufficient.

KhafraDev:
I found something that worked

PR #2314, opened 2023-10-06
fix(#2311): End stream after body sent

### LINKED ISSUE #2311: bug: fetch causes http2 server to hang
## Bug Description

Reading the `body` of an http2 POST call causes `node:http2` to hang when called from this `fetch`, but not from Deno's fetch or curl.

## Reproducible By

```js
import http2 from "node:http2";
import assert from "node:assert";
import { exec } from "node:child_process";

import { Client, fetch } from "undici";
import devcert from "devcert";

const ssl = await devcert.certificateFor("localhost");
const server = http2.createSecureServer(ssl, async (req, res) => {
  assert.equal(req.method, "POST");
  console.log("beginning to read body...");
  let body = [];
  for await (const chunk of req) {
    console.log({ chunk });
    body.push(chunk);
  }
  let buffer = Buffer.concat(body);
  console.log("end of reading body");
  res.end(buffer);
}).listen(3000);

const METHOD = "fetch";
const h2Client = new Client("https://localhost:3000", {
  connect: {
    rejectUnauthorized: false,
  },
  allowH2: true,
});

server.on("listening", async () => {
  console.log("listening on 3000");
  if (METHOD === "fetch") {
    const response = await fetch("https://localhost:3000", {
      dispatcher: h2Client,
      method: "POST",
      body: "12",
      headers: {
        "content-type": "text/plain",
        "content-length": "2",
      },
    });
    console.log("DONE", { response });
  } else if (METHOD === "curl") {
    const curl =
      `curl --no-progress-meter -k -X POST -H "content-type: text/plain" -H "content-length: 2" -d "12" https://localhost:3000`;
    exec(curl, (err, stdout, stderr) => {
      if (err) {
        console.error("DONE", err);
        return;
      }
      console.log("DONE", stdout);
    });
  } else {
    throw new Error();
  }
});
```

Play around with the value of `METHOD` to see what I'm talking about

## Expected Behavior

both `curl` and `fetch` should receive a response quickly.

Instead, `curl` gets a response immediately but `fetch` never returns.

## Logs & Screenshots
```js
// with curl
beginning to read body...
{ chunk: <Buffer 31 32> }
end of reading body
DONE 12
```

```js
// with fetch
beginning to read body...
{ chunk: <Buffer 31 32> }
// never gets passed here
```

## Environment

node v20.3.1
Linux maxs-pc 5.19.0-43-generic #44~22.04.1-Ubuntu SMP PREEMPT_DYNAMIC Mon May 22 13:39:36 UTC 2 x86_64 x86_64 x86_64 GNU/Linux

### Additional context

I have tested this code with
- node's builtin fetch ❌
- undici fetch ❌
- undici `Client#request` ❌
- curl ✅
- Deno fetch ✅

metcoder95:
## This relates to...

Fixes #2311

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -2095,13 +2095,17 @@ async function writeIterable ({ h2stream, body, client, request, socket, content
           throw socket[kError]
         }
 
-        if (!h2stream.write(chunk)) {
+        const res = h2stream.write(chunk)
+        request.onBodySent(chunk)
+        if (!res) {
           await waitForDrain()
         }
       }
     } catch (err) {
       h2stream.destroy(err)
     } finally {
+      request.onRequestSent()
+      h2stream.end()
       h2stream
         .off('close', onDrain)
         .off('drain', onDrain)
--- a/test/fetch/http2.js
+++ b/test/fetch/http2.js
@@ -13,7 +13,61 @@ const { Client, fetch } = require('../..')
 
 const nodeVersion = Number(process.version.split('v')[1].split('.')[0])
 
-plan(5)
+plan(6)
+
+test('[Fetch] Issue#2311', async t => {
+  const expectedBody = 'hello from client!'
+
+  const server = createSecureServer(pem, async (req, res) => {
+    let body = ''
+
+    req.setEncoding('utf8')
+
+    res.writeHead(200, {
+      'content-type': 'text/plain; charset=utf-8',
+      'x-custom-h2': req.headers['x-my-header']
+    })
+
+    for await (const chunk of req) {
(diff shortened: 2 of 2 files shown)


metcoder95:
It seems all are failing over the redirect, except one. Re-running to see if is transitive

ronag (approved):

mcollina (approved):
lgtm

KhafraDev (approved):

PR #2322, opened 2023-10-09
disallow setting host header in fetch

### LINKED ISSUE #2318: Fetch failed when setting host: header correctly
## Reproducible By

~~~~
$ node
Welcome to Node.js v18.13.0.
> const { fetch } = require('undici');
undefined
> await fetch("https://bing.com/translator", { headers: { host: 'bing.com' } });
Uncaught TypeError: fetch failed
    at fetch ($PWD/node_modules/undici/index.js:110:15)
    at async REPL2:1:33 {
  cause: Error: Client network socket disconnected before secure TLS connection was established
      at connResetException (node:internal/errors:718:14)
      at TLSSocket.onConnectEnd (node:_tls_wrap:1600:19)
      at TLSSocket.emit (node:events:525:35)
      at TLSSocket.emit (node:domain:552:15)
      at endReadableNT (node:internal/streams/readable:1359:12)
    code: 'ECONNRESET',
    path: undefined,
    host: 'www.bing.com',
    port: 443,
    localAddress: null
  }
}
> await fetch("https://bing.com/translator");
Response {
  [Symbol(realm)]: null,
[ .. all good without explicit host: .. ]
~~~~

The bug happens with bing.com because it tries to redirect to [www.bing.com](http://www.bing.com/); it doesn't happen for domains that don't redirect, such as fanyi.baidu.com.

## Environment

| software         | version
| ---------------- | -------
| undici       | v5.25.4
| node             |  v18.13.0
| Operating System | Debian GNU/Linux trixie/sid

KhafraDev:
Fixes #2318

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -1406,6 +1406,8 @@ async function httpNetworkOrCacheFetch (
     }
   }
 
+  httpRequest.headersList.delete('host')
+
   //    20. If includeCredentials is true, then:
   if (includeCredentials) {
     // 1. If the user agent is not configured to block cookies for httpRequest
--- a/test/fetch/issue-2318.js
+++ b/test/fetch/issue-2318.js
@@ -1,0 +1,25 @@
+'use strict'
+
+const { test } = require('tap')
+const { once } = require('events')
+const { createServer } = require('http')
+const { fetch } = require('../..')
+
+test('Undici overrides user-provided `Host` header', async (t) => {
+  t.plan(1)
+
+  const server = createServer((req, res) => {
+    t.equal(req.headers.host, `localhost:${server.address().port}`)
+
+    res.end()
+  }).listen(0)
+
+  t.teardown(server.close.bind(server))
+  await once(server, 'listening')
+
+  await fetch(`http://localhost:${server.address().port}`, {
+    headers: {
+      host: 'www.idk.org'
+    }
+  })
+})


KhafraDev:
this matches what cloudflare workers and deno do

mcollina (approved):
lgtm

paulrutter:
> this matches what cloudflare workers and deno do

@KhafraDev it's not, as Deno offers a way to allow setting the host header still.
We need something similar for Node.js, as setting the `host` header is valid in some usecases (see [#50305](https://github.com/nodejs/node/issues/50305)).

Ethan-Arrowood:
This is unfortunate, but this change did impact Vercel's dev server implementation. I do recognize we were inappropriately using fetch when we should've been using `request`. I think we need to do a better job at documentation appropriate use cases for `fetch` and the other methods.

Ref: https://github.com/vercel/vercel/pull/10767

kentonv:
> this matches what cloudflare workers and deno do

FWIW, the reason Cloudflare Workers does this is arguably pretty specific to Cloudflare: Many origin servers that sit behind Cloudflare are configured to allowlist Cloudflare IPs and use host header verification to make sure that requests coming from Cloudflare actually came through their Cloudflare account. This was true before Workers was introduced. So allowing Workers to rewrite host headers would have allowed bypassing security configuration on existing origin servers.

Browsers, of course, need to restrict setting of the Host header because otherwise all kinds of CORS bypass would be possible. However, CORS really only applies to browsers, not servers. (Even Cloudflare doesn't implement CORS restrictions.)

I'm not sure if there's any need for Node to implement the same restriction, and indeed the ability to rewrite the host header is usually necessary when implementing reverse proxies, since you typically want the upstream app server to receive the original host header as sent by the browser, not a host header specifying its internal hostname.

Netail:
> and indeed the ability to rewrite the host header is usually necessary when implementing reverse proxies

We are indeed making use of a reverse proxy and are currently facing this issue from switching from node-fetch to native fetch... :/

PR #2327, opened 2023-10-10
chore: simplify creation of `fetch error`

fedyk:
## Simplify creation of `fetch error`

This relates to the way how `fetch error` is created. `TypeError` accept `cause` in constructor, so `Object.assign` can be removed.

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready


ronag (approved):

metcoder95 (approved):

KhafraDev:
A few things:
- this will drop support for node v16.8 (or make the error entirely useless). Not really an issue but the README should be changed.
- there's at least one more place where this pattern is used, I think in `lib/fetch/body.js`? It should be changed there also.

mcollina (changes requested):
I don't think we should be landing this without bumping a major.

metcoder95:
Changed, the branch but it seems `next` needs to be rebased

fedyk:
@KhafraDev thanks for your comment. I've adjusted `lib/fetch/body.js` and `lib/fetch/response.js` as well.

As for changes in README, I didn't get the point with dropping support for node v16.8. If it is critical for this MR, please elaborate 🙂

mcollina:
This commit should target the `next` branch

KhafraDev:
> As for changes in README, I didn't get the point with dropping support for node v16.8. If it is critical for this MR, please elaborate 

error.cause was added in node v16.9, at least according to mdn. On node v16.8, which is currently supported by undici, the error message will be completely useless with this change, essentially dropping support for it IMO.

fedyk:
> > As for changes in README, I didn't get the point with dropping support for node v16.8. If it is critical for this MR, please elaborate
> 
> error.cause was added in node v16.9, at least according to mdn. On node v16.8, which is currently supported by undici, the error message will be completely useless with this change, essentially dropping support for it IMO.

You are right - `cause` is supported since [v16.9.0](https://nodejs.org/api/errors.html#errorcause)

fedyk:
I've created a new clean MR for the changes https://github.com/nodejs/undici/pull/2347. This one got glutted

PR #2339, opened 2023-10-11
fix: user-agent by making __filename check more resilient

Ethan-Arrowood:
Rel: https://github.com/nodejs/undici/pull/2310 https://github.com/nodejs/node/pull/50145

For some reason `__filename` is undefined when this gets loaded into node.js

Trying something else that might be more resilient. This also just falls back to `index.js` so that if all goes wrong at least we don't get a runtime error. WDYT? Should we switch the default to result in `'node'`?

### DIFF
--- a/build/fetch.js
+++ b/build/fetch.js
@@ -1,0 +1,36 @@
+const esbuild = require('esbuild')
+const fs = require('fs')
+const path = require('path')
+
+// Copied from: https://github.com/evanw/esbuild/issues/859#issuecomment-829154955
+const nodeModules = /^(?:.*[\\/])?node_modules(?:[\\/].*)?$/
+
+const dirnamePlugin = {
+  name: 'dirname',
+  setup (build) {
+    build.onLoad({ filter: /.*/ }, ({ path: filePath }) => {
+      if (!filePath.match(nodeModules)) {
+        let contents = fs.readFileSync(filePath, 'utf8')
+        const loader = path.extname(filePath).substring(1)
+        const dirname = path.dirname(filePath)
+        contents = contents
+          .replace('__dirname', `"${dirname}"`)
+          .replace('__filename', `"${filePath}"`)
+        return {
+          contents,
+          loader
+        }
+      }
+    })
+  }
+}
+
+esbuild.build({
+  platform: 'node',
+  entryPoints: [path.resolve('index-fetch.js')],
+  bundle: true,
+  outfile: path.resolve('undici-fetch.js'),
+  plugins: [dirnamePlugin]
+}).then(() => {
+  console.log('Build complete.')
+})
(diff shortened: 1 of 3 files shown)


Ethan-Arrowood:
@KhafraDev previously commented that there isn't a difference between `__filename` and `globalThis.__filename`.

Unless someone else has a better idea, I think we rely on this undefined behavior to set the agent to `'node'`. We can do:

```js
const filename = __filename ?? ''
filename.endsWith('index.js') ? 'undici' : 'node'
```

or 

```js
__filename ? 'undici' : 'node'
```

I prefer the first so that if node ever enables __filename in its deps, then the code won't magically switch to `'undici'` user-agent. 

WDYT?

Ethan-Arrowood:
I prefer this with the `globalThis.__filename` change. that way no random change to globals could break it.

KhafraDev:
a few tests are failing

Ethan-Arrowood:
lol `globalThis` doesn't have `__filename` on it 

```
TAP version 13
globalThis <ref *1> Object [global] {
  global: [Circular *1],
  clearInterval: [Function: clearInterval],
  clearTimeout: [Function: clearTimeout],
  setInterval: [Function: setInterval],
  setTimeout: [Function: setTimeout] {
    [Symbol(nodejs.util.promisify.custom)]: [Getter]
  },
  queueMicrotask: [Function: queueMicrotask],
  performance: Performance {
    nodeTiming: PerformanceNodeTiming {
      name: 'node',
      entryType: 'node',
      startTime: 0,
      duration: 112.42291688919067,
      nodeStart: 1.5486669540405273,
      v8Start: 3.308124542236328,
      bootstrapComplete: 22.103166580200195,
      environment: 12.799707889556885,
      loopStart: -1,
      loopExit: -1,
      idleTime: 0
    },
    timeOrigin: 1697039951604.925
  },
  clearImmediate: [Function: clearImmediate],
  setImmediate: [Function: setImmediate] {
    [Symbol(nodejs.util.promisify.custom)]: [Getter]
  }
}
```

KhafraDev:
I can't believe I forgot this, but __filename isn't a global because it wouldn't work (how would it change for every file?), cjs modules are wrapped in an iife that has __filename, __dirname, etc. So I guess node core isn't doing that for dependencies.

KhafraDev:
I think we should just make an esbuild plugin that injects a variable into `lib/fetch/index.js` then we can do the typeof check

Ethan-Arrowood:
Have you written one of those before? I have not but can start looking into it.

ronag (approved):

KhafraDev:
I will have a PR fixing this soon

KhafraDev:
#2341

Ethan-Arrowood:
I found this thread: https://github.com/evanw/esbuild/issues/859#issuecomment-829154955

I don't love this solution mainly because I don't understand what the plugin is really doing. Would love some more input before we ship this

Ethan-Arrowood:
Closing in favor of #2341

PR #2340, opened 2023-10-11
fix user-agent detection

KhafraDev:
refs: https://github.com/nodejs/undici/pull/2310#issuecomment-1757937704

I confirmed that this fixes the issue in node core.

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -1344,7 +1344,9 @@ async function httpNetworkOrCacheFetch (
   //    user agents should append `User-Agent`/default `User-Agent` value to
   //    httpRequest’s header list.
   if (!httpRequest.headersList.contains('user-agent')) {
-    httpRequest.headersList.append('user-agent', __filename.endsWith('index.js') ? 'undici' : 'node')
+    // __commonJS is injected by esbuild when bundled. The esbuild bundle is used in node core.
+    // Therefore if we're running the esbuild bundle, we're likely in node core.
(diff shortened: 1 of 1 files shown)


KhafraDev:
esbuild will rename it to something else when rebundled, this doesn't work

lib/fetch/index.js
```diff
@@ -1344,7 +1344,9 @@ async function httpNetworkOrCacheFetch (
   //    user agents should append `User-Agent`/default `User-Agent` value to
   //    httpRequest’s header list.
   if (!httpRequest.headersList.contains('user-agent')) {
-    httpRequest.headersList.append('user-agent', __filename.endsWith('index.js') ? 'undici' : 'node')
+    // __commonJS is injected by esbuild when bundled. The esbuild bundle is used in node core.
+    // Therefore if we're running the esbuild bundle, we're likely in node core.
+    httpRequest.headersList.append('user-agent', typeof __commonJS !== 'function' ? 'undici' : 'node')
```
Ethan-Arrowood, line 1349: not a fan of relying on an unspecified global. Someone could decide to set this themselves and undici would break itself.

KhafraDev: fwiw it's also possible to set __filename

Ethan-Arrowood: true but there is at least an assumption/definition for `__filename`. If someone decides to mess with that - it's on them.

PR #2341, opened 2023-10-11
Fix node detection omfg

KhafraDev:
I confirmed tests pass and it works in node.

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -1344,7 +1344,7 @@ async function httpNetworkOrCacheFetch (
   //    user agents should append `User-Agent`/default `User-Agent` value to
   //    httpRequest’s header list.
   if (!httpRequest.headersList.contains('user-agent')) {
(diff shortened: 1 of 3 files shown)


Ethan-Arrowood (approved):
I like this much more. ship it

KhafraDev:
test seems to fail on windows for some reason

merceyz:
Instead of a plugin couldn't you use the [`--define`](https://esbuild.github.io/api/#define) flag for this?

Ethan-Arrowood:
Didn't even know that was an option! I think that should work nicely.

KhafraDev:
We only bundle undici for node core, I don't see the difference between the two unless we started bundling undici for npm.

Ethan-Arrowood:
It might just be a simpler solution. I'm fine with either way

KhafraDev:
Let's merge this as is so we can land the undici update and then we can look into other solutions. I'm not home anymore so would you mind merging and releasing an update? Thanks

Ethan-Arrowood:
Done here: https://github.com/nodejs/undici/releases/tag/v5.26.1

PR #2342, opened 2023-10-11
use esbuild define/hooks

KhafraDev:
The bundle file now compiles the relevant code to:

```js
if (!httpRequest.headersList.contains("user-agent")) {
    httpRequest.headersList.append("user-agent", false ? "undici" : "node");
}
```

while for npm users it will be:

```js
  httpRequest.headersList.append('user-agent', typeof esbuildDetection === 'undefined' ? 'undici' : 'node')
}
```

much simpler


Ethan-Arrowood (approved):
great work! definitely keeping this `--define` option in mind for future esbuild usage. it works really well!

mcollina (approved):
lgtm

PR #2347, opened 2023-10-15
simplify error and cause

fedyk:
This relates to https://github.com/nodejs/undici/pull/2327. Due to rebase of changes from master to next, the changes looks weird. I decided to create a new clean MR.

## Rationale

Since [Node v16.9.0](https://nodejs.org/api/errors.html#errorcause), `Error` supports `cause` property.

## Changes

Replace `Object.assign` with `new Error(.., { cause })`.

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [x] In review
- [ ] Merge ready

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,7 +180,7 @@ Implements [fetch](https://fetch.spec.whatwg.org/#fetch-method).
 * https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch
 * https://fetch.spec.whatwg.org/#fetch-method
 
-Only supported on Node 16.8+.
+Only supported on Node 16.9+.
 
 Basic usage example:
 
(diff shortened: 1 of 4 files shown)


mcollina (changes requested):
I have a loose recall that v16.8 is what lambda deploys, so changing this would be detrimental.

fedyk:
@mcollina thanks for your comment.

I believe you are referring to AWS Lambda. According to the [docs](https://docs.aws.amazon.com/lambda/latest/dg/lambda-nodejs.html), they support Node.js `v16.x`. Brief grep in the [base nodejs image](https://github.com/aws/aws-lambda-base-images/tree/nodejs16.x), shows that the current version is `16.20.2` (they probably keep the most recent version live).

metcoder95 (approved):

ronag:
Can we make a test for node 16?

mcollina:
There is already one for node v16.8 in CI

fedyk:
If support of v16.8 is important, then those changes are breaking due to `cause` only available in v16.9. I believe this MR can be closed for now.

PR #2351, opened 2023-10-17
test: only include WebSocket in WPT Report where it's landed

panva:
This makes the inclusion of WebSocket in the daily WPT Report uploaded depend on whether it has landed in the respective tested release line or not (or whether it's already made global by default).

[Currently](https://wpt.fyi/results/websockets?run_id=6203937658503168&run_id=5195270930825216&run_id=5093410379137024) it's reported even for releases where the experimental flag is not available (yet).


mcollina (approved):
lgtm

PR #2361, opened 2023-10-20
Update namespace type with Fetch exports

Ethan-Arrowood:
Now that 16 is EOL I think it is okay to update our default types to include the Fetch related exports. 

Corresponds to the exports in `index.js`.


mcollina (approved):
lgtm

kibertoad:
@Ethan-Arrowood Maybe support for Node 16 should be dropped, and semver major bumped then?

PR #2368, opened 2023-10-23
Use sets and reusable TextEncoder/TextDecoder instances

kibertoad:
## Rationale

Minor perf improvements

## Changes

* Use sets over array.indexOf (seems to be consistently faster even for small datasets, see https://www.measurethat.net/Benchmarks/Show/28018/0/sethas-vs-arrayincludes-vs-arrayindexof-small-set-of-st)

* Do not instantiate TextDecoder and TextEncoder over and over

### Features

N/A

### Bug Fixes

N/A

No breaking changes

## Status

- [X] Tested
- [ ] Documented
- [X] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/fetch/body.js
+++ b/lib/fetch/body.js
@@ -27,7 +27,6 @@ let ReadableStream = globalThis.ReadableStream
 /** @type {globalThis['File']} */
 const File = NativeFile ?? UndiciFile
 const textEncoder = new TextEncoder()
-const textDecoderNoBom = new TextDecoder('utf-8', { ignoreBOM: true })
 const textDecoder = new TextDecoder()
 
 // https://fetch.spec.whatwg.org/#concept-bodyinit-extract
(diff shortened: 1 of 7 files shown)


mcollina (approved):
lgtm

lib/fetch/body.js
```diff
@@ -443,14 +445,13 @@ function bodyMixinMethods (instance) {
           let text = ''
           // application/x-www-form-urlencoded parser will keep the BOM.
           // https://url.spec.whatwg.org/#concept-urlencoded-parser
-          const textDecoder = new TextDecoder('utf-8', { ignoreBOM: true })
           for await (const chunk of consumeBody(this[kState].body)) {
             if (!isUint8Array(chunk)) {
               throw new TypeError('Expected Uint8Array chunk')
             }
-            text += textDecoder.decode(chunk, { stream: true })
+            text += textDecoderNoBom.decode(chunk, { stream: true })
```
merceyz, line 452: Is it safe to reuse the decoder in stream mode for multiple requests?

kibertoad: Good question. Documentation says this:

`Decodes the input and returns a string. If options.stream is true, any incomplete byte sequences occurring at the end of the input are buffered internally and emitted after the next call to textDecoder.decode().`

That does make it sound like streaming decoding is, indeed, stateful.

@mcollina any comments on this?

ronag: Yea, don't re-use in the streaming case.

ronag (changes requested):
- Don't re-use for streaming case.
- You are setting `ignoreBOM` in cases where it was not so before.

lib/fetch/body.js
kibertoad: gotcha, will adjust the streaming case shortly

kibertoad: fixed

kibertoad:
> You are setting ignoreBOM in cases where it was not so before.

@ronag Can you point out where that is happening? I've went over the code and it seems to be the same way as it was before (streaming case has its own decoder before, and now it's not reused anymore either)

ronag (approved):

metcoder95 (approved):

PR #2373, opened 2023-10-24
improve `HeadersInit` type

Ethan-Arrowood:
This PR improves the `HeadersInit` type. The spec sorta gets this wrong (https://fetch.spec.whatwg.org/#typedefdef-headersinit) It only specifies `record<ByteString, ByteString>`, but `record<ByteString, ByteString | undefined` also works. I've tested it in multiple browsers and our implementation, and it works fine. 

This allows me to pass an `IncomingHTTPHeaders` object directly to the constructor (which is valid for all other reasons except for this missing `| undefined`).

### DIFF
--- a/types/fetch.d.ts
+++ b/types/fetch.d.ts
@@ -50,7 +50,7 @@ export interface SpecIterable<T> {
   [Symbol.iterator](): SpecIterator<T>;
 }
 
-export type HeadersInit = string[][] | Record<string, string | ReadonlyArray<string>> | Headers
+export type HeadersInit = string[][] | Record<string, string | ReadonlyArray<string> | undefined> | Headers
 
 export declare class Headers implements SpecIterable<[string, string]> {
   constructor (init?: HeadersInit)


KhafraDev (changes requested):
The type isn't wrong here because `HeadersInit` cannot be undefined, but it is optional in the Headers constructor.

Technically webidl will handle any type we pass (`null`, numbers, objects, arrays, etc.) but we should be explicit in what types/values are expected.

KhafraDev:
IMO the ReadonlyArray<string> should be removed because it's not an expected value, I'm really not sure how that ended up there. Webidl will handle it, but probably not in an expected way, and we probably shouldn't encourage people to misuse it.

Ethan-Arrowood:
I hear ya but we should match browser behavior. And honestly our types should match our implementation. If the spec does say to fail on undefined, and if our implementation doesn't fail on it either, I see little harm in adjusting the types so that we can avoid an unnecessary `@ts-expect-error`

KhafraDev:
If you pass `undefined` as the value it'll get serialized to `"undefined"`. It'll only lead to unexpected behavior. Plus the type, if we're following webidl, would be something similar to `[any, any][] | Headers | Record<any, any>` which is less explicit and less useful than stricter typings.

Uzlopak:
Closing due too inactivity.

PR #2376, opened 2023-10-25
add regression test

KhafraDev:
refs: https://github.com/nodejs/node/issues/50263#issuecomment-1779670104

### DIFF
--- a/test/fetch/bundle.js
+++ b/test/fetch/bundle.js
@@ -18,3 +18,17 @@ test('bundle sets constructor.name and .name properly', (t) => {
 
   t.end()
 })
+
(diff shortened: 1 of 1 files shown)


metcoder95 (approved):

mcollina (approved):
lgtm

mcollina:
@KhafraDev my understanding is that was caused by bundling?

ronag (approved):

KhafraDev:
Yes it was

KhafraDev:
I think it was because we're using `blah.constructor.name === 'Headers'` somewhere which then broke when esbuild decided to rename the classes for us. Not really sure though.

PR #2379, opened 2023-10-25
fix windows CI

### LINKED ISSUE #2378: Windows CI: The syntax of the command is incorrect.
windows 18 & 20 consistently fail 80% of the way into the WPTs with ` The syntax of the command is incorrect.`

the fix might be to replace the forward slashes for paths with `\\`, but then it won't work on other platforms.

https://github.com/nodejs/undici/blob/3ec35ee23a7f2693bd6f9f707f6a20dc6de7a6b0/package.json#L87

KhafraDev:
Fixes #2378

### DIFF
--- a/package.json
+++ b/package.json
@@ -84,7 +84,7 @@
     "test:tdd": "tap test/*.js test/diagnostics-channel/*.js -w",
     "test:typescript": "node scripts/verifyVersion.js 14 || tsd && tsc --skipLibCheck test/imports/undici-import.ts",
     "test:websocket": "node scripts/verifyVersion.js 18 || tap test/websocket/*.js",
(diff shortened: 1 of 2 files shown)


KhafraDev:
yeah so that works, I have no idea why removing the options passed to node would matter, but...

panva:
FYI This PR has removed websockets from the WPT report again.

panva:
And adding the `--experimental-websocket` option to your npm script will not work because presence of that option will make the daily WPT job error on older versions of node where the option didn't land.

KhafraDev:
why did we need the flag?

panva:
To allow detecting that the feature was present on a particular node build that's being used in the WPT job matrix

KhafraDev:
I think it can be replaced with a version check instead

panva:
> I think it can be replaced with a version check instead

If you want to maintain that, sure, I didn't.

Issue #2364, opened 2023-10-21
[HTTP/2] Pipelined requests might fail unexpectedly if another request toward the same domain has been aborted


SukkaW:
## Bug Description

As the title, pipelined requests might fail unexpectedly if another request toward the same domain has been aborted:

```
node:assert:399
    throw err;
    ^

AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:

  assert(!this.aborted)

    at Request.onHeaders (/home/runner/undici-h2-promise-any/node_modules/undici/lib/core/request.js:235:5)
    at ClientHttp2Stream.<anonymous> (/home/runner/undici-h2-promise-any/node_modules/undici/lib/client.js:1804:17)
    at Object.onceWrapper (node:events:628:26)
    at ClientHttp2Stream.emit (node:events:513:28)
    at emit (node:internal/http2/core:330:3)
    at process.processTicksAndRejections (node:internal/process/task_queues:85:22) {
  generatedMessage: true,
  code: 'ERR_ASSERTION',
  actual: false,
  expected: true,
  operator: '=='
}

Node.js v18.16.1
```

## Reproducible By

https://replit.com/@isukkaw/undici-h2-promise-any

```js
const undici = require('undici');

undici.setGlobalDispatcher(new undici.Agent({
  allowH2: true,
  pipelining: 10
}));

const { fetch } = undici;

const fetchUbuntuXenialReleaseGpg = () => {
  const controller = new AbortController();
  return Promise.any(
    [
      'https://mirror-cdn.xtom.com/ubuntu/dists/xenial/Release.gpg',
      'https://mirrors.xtom.com/ubuntu/dists/xenial/Release.gpg',
      'https://mirrors.xtom.com.hk/ubuntu/dists/xenial/Release.gpg',
      'https://mirrors.xtom.de/ubuntu/dists/xenial/Release.gpg',
      'https://mirrors.xtom.nl/ubuntu/dists/xenial/Release.gpg',
      'https://mirrors.xtom.ee/ubuntu/dists/xenial/Release.gpg',
      'https://mirrors.xtom.jp/ubuntu/dists/xenial/Release.gpg',
      'https://mirrors.xtom.au/ubuntu/dists/xenial/Release.gpg'
    ].map(async url => {
      try {
        const resp = await fetch(url, { signal: controller.signal });
        const text = await resp.text();
        // The body should have been consumed, abort the requests
        controller.abort();
        return text;
      } catch(e) {
[...]

PR #2387, opened 2023-10-29
fix(HTTP/2): handle consumption of aborted request

### LINKED ISSUE #2386: Aborting HTTP2 to streaming endpoints throws an assertion error
## Bug Description

Aborting streaming endpoints immediately the after response is ready results in an assertion failure
This might be related to https://github.com/nodejs/undici/issues/2364 but the assert is appearing in a different spot
And the repro doesnt require multiple requests to the same domain
But if these are deemed the same, happy for yall to consolidate to that one issue

## Reproducible By

https://replit.com/@JS7/undici-h2-abort-failure

```javascript
const undici = require('undici');

undici.setGlobalDispatcher(new undici.Agent({
  allowH2: true,
}));

(async () => {
  const response = await undici.fetch('https://mirror-cdn.xtom.com/ubuntu/dists/xenial/Release.gpg', { signal: controller.signal });
  controller.abort()
  setTimeout(()=>console.log("done"), 100);
})();
```

## Expected Behavior

At some point I see the console.log printing "done". Nothing is thrown.

## Actual Behavior

I see this and no other output
```
node:assert:399
    throw err;
    ^

AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:

  assert(!this.aborted)

    at Request.onData (/home/runner/undici-h2-abort-failure/node_modules/undici/lib/core/request.js:246:5)
    at ClientHttp2Stream.<anonymous> (/home/runner/undici-h2-abort-failure/node_modules/undici/lib/client.js:1814:17)
    at ClientHttp2Stream.emit (node:events:513:28)
    at addChunk (node:internal/streams/readable:324:12)
    at readableAddChunk (node:internal/streams/readable:297:9)
    at Readable.push (node:internal/streams/readable:234:10)
    at Http2Stream.onStreamRead (node:internal/stream_base_commons:190:23) {
  generatedMessage: true,
  code: 'ERR_ASSERTION',
  actual: false,
  expected: true,
  operator: '=='
}
```

## Environment

Node.js v18.16.1
undici@5.26.4

metcoder95:
## This relates to...

Fixes #2386

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready


lib/client.js
```diff
@@ -1820,6 +1820,11 @@ function writeH2 (client, session, request) {
   })
 
   stream.on('data', (chunk) => {
+    // Aborting a request does not abort
+    // the stream, so we need to check and destroy it if request
+    // is aborted.
+    if (request.aborted) util.destroy(stream, new RequestAbortedError())
```
metcoder95, line 1826: I'm not super convinced with the solution but couldn't find other way. So I'm open to suggestions

ronag: Why not register this logic in the closure sent in onConnect?

metcoder95: Do you mean here https://github.com/nodejs/undici/blob/3a77cbb471b76175898d8b094beb9b40ceadd671/lib/client.js#L1684?

mcollina: I think it would be better to remove the listener instead

metcoder95: >Do you mean here

Hmm, yeah it makes sense; maybe we can also call it just and only if the stream is ready or wait until the `ready` event triggers.

>I think it would be better to remove the listener instead

Will the data remain buffered there, or what will be the direct consequence of removing the listener? Just to understand the reason behind better 😅

jackschu: The current patch seems like we'd immediately call request.onData in the next line (since destroy only schedules an 'error' event), which would then hit the same `assert(!this.aborted)`, and experimentally this is what I run into

jackschu:
Any update here @metcoder95? We're running with 
```
   stream.on('data', (chunk) => {
-    if (request.onData(chunk) === false) stream.pause()
+      if (request.aborted) util.destroy(stream, new RequestAbortedError())
+      else if (request.onData(chunk) === false) stream.pause()
   })
```

But we understandably still hit https://github.com/nodejs/undici/issues/2364 since this has no effect on `request.onHeaders`

Right now to unblock myself the path im going down is to patch every stream event listener :/

@mcollina you mentioned this, what listener are  you referring to?
> I think it would be better to remove the listener instead

metcoder95:
Haven't had the time this week to work on this; open to delegate if someone is interested in. Planning to spend some time on the task this week 🙂

mcollina (approved):
lgtm

mcollina:
Can you merge main in again?

PR #2391, opened 2023-11-01
fix: stream body handling

ronag:
- Make sure stream body always has an error handler.
- Make sure not to leave dangling event listeners.


KhafraDev (approved):

metcoder95 (commented):
LGTM 👍

PR #2397, opened 2023-11-02
perf(headers): Improve Headers

tsctx:
## Rationale

This eliminates the `Headers` performance bottleneck.

- `for...of` is slower than `for(;;)`.
- `Object.entries` is slower than  `Object.keys`.
- Reduced the number of variables in the function `ByteString`.
- When trimming `HTTP whitespace`, using RegExp is slow, so change to using `charCodeAt` instead.

### Features

N/A

### Bug Fixes

N/A

N/A

## Benchmark

```js
// bench.js
const Headers = require("./lib/fetch/headers.js").Headers;

const headers = {
  "X-Content-Type": "application/json",
  "X-NodeJS-Version": "v21.1.0",
  "Powered-By": "NodeJS",
  "Content-Encoding": "gzip",
  "Content-Length": "150",
};

const headersEntries = Object.entries(headers);


suite.add("new Headers - record", () => {
  new Headers(headers);
});

suite.add("new Headers - entries", () => {
  new Headers(headersEntries);
});

});

// run async
suite.run({ async: true });
```

- main
```
new Headers - record x 144,160 ops/sec ±1.91% (85 runs sampled)
new Headers - entries x 156,706 ops/sec ±1.88% (83 runs sampled)
```

- this patch
```
new Headers - record x 179,333 ops/sec ±4.49% (72 runs sampled)
new Headers - entries x 170,316 ops/sec ±2.39% (80 runs sampled)
```

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/fetch/headers.js
+++ b/lib/fetch/headers.js
@@ -56,7 +56,7 @@ function fill (headers, object) {
       }
 
       // 2. Append (header’s first item, header’s second item) to headers.
-      headers.append(header[0], header[1])
+      appendHeader(headers, header[0], header[1])
     }
   } else if (typeof object === 'object' && object !== null) {
     // Note: null should throw
@@ -65,7 +65,7 @@ function fill (headers, object) {
     //    append (key, value) to headers
     const keys = Object.keys(object)
     for (let i = 0; i < keys.length; ++i) {
-      headers.append(keys[i], object[keys[i]])
+      appendHeader(headers, keys[i], object[keys[i]])
     }
   } else {
     throw webidl.errors.conversionFailed({
@@ -74,6 +74,49 @@ function fill (headers, object) {
       types: ['sequence<sequence<ByteString>>', 'record<ByteString, ByteString>']
     })
   }
+}
+
+/**
+ * @see https://fetch.spec.whatwg.org/#concept-headers-append
+ */
+function appendHeader (headers, name, value) {
+  // 1. Normalize value.
+  value = headerValueNormalize(value)
+
+  // 2. If name is not a header name or value is not a
+  //    header value, then throw a TypeError.
+  if (!isValidHeaderName(name)) {
+    throw webidl.errors.invalidArgument({
+      prefix: 'Headers.append',
+      value: name,
+      type: 'header name'
+    })
+  } else if (!isValidHeaderValue(value)) {
+    throw webidl.errors.invalidArgument({
+      prefix: 'Headers.append',
+      value,
+      type: 'header value'
+    })
+  }
+
+  // 3. If headers’s guard is "immutable", then throw a TypeError.
+  // 4. Otherwise, if headers’s guard is "request" and name is a
+  //    forbidden header name, return.
+  // Note: undici does not implement forbidden header names
+  if (headers[kGuard] === 'immutable') {
+    throw new TypeError('immutable')
+  } else if (headers[kGuard] === 'request-no-cors') {
+    // 5. Otherwise, if headers’s guard is "request-no-cors":
+    // TODO
+  }
+
+  // 6. Otherwise, if headers’s guard is "response" and name is a
+  //    forbidden response-header name, return.
+
+  // 7. Append (name, value) to headers’s header list.
+  // 8. If headers’s guard is "request-no-cors", then remove
+  //    privileged no-CORS request headers from headers
+  return headers[kHeadersList].append(name, value)
 }
 
 class HeadersList {
@@ -221,43 +264,7 @@ class Headers {
     name = webidl.converters.ByteString(name)
     value = webidl.converters.ByteString(value)
 
-    // 1. Normalize value.
-    value = headerValueNormalize(value)
-
-    // 2. If name is not a header name or value is not a
-    //    header value, then throw a TypeError.
-    if (!isValidHeaderName(name)) {
-      throw webidl.errors.invalidArgument({
-        prefix: 'Headers.append',
-        value: name,
-        type: 'header name'
-      })
-    } else if (!isValidHeaderValue(value)) {
-      throw webidl.errors.invalidArgument({
-        prefix: 'Headers.append',
-        value,
-        type: 'header value'
-      })
-    }
-
-    // 3. If headers’s guard is "immutable", then throw a TypeError.
-    // 4. Otherwise, if headers’s guard is "request" and name is a
-    //    forbidden header name, return.
-    // Note: undici does not implement forbidden header names
-    if (this[kGuard] === 'immutable') {
-      throw new TypeError('immutable')
-    } else if (this[kGuard] === 'request-no-cors') {
-      // 5. Otherwise, if headers’s guard is "request-no-cors":
-      // TODO
-    }
-
-    // 6. Otherwise, if headers’s guard is "response" and name is a
-    //    forbidden response-header name, return.
-
-    // 7. Append (name, value) to headers’s header list.
-    // 8. If headers’s guard is "request-no-cors", then remove
-    //    privileged no-CORS request headers from headers
-    return this[kHeadersList].append(name, value)
+    return appendHeader(this, name, value)
   }
 
   // https://fetch.spec.whatwg.org/#dom-headers-delete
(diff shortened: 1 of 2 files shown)


lib/fetch/webidl.js
```diff
@@ -399,6 +399,11 @@ webidl.nullableConverter = function (converter) {
 
 // https://webidl.spec.whatwg.org/#es-DOMString
 webidl.converters.DOMString = function (V, opts = {}) {
+  // Note: avoid re-stringify
+  if (typeof V === 'string') {
+    return V
+  }
```
KhafraDev, line 405: The String constructor calls [`ToString(value)`](https://tc39.es/ecma262/#sec-tostring) in which the first step is `If argument is a String, return argument.` This almost certainly won't improve performance in a meaningful way (and if it does, there's likely somewhere else it could be improved, like initializing opts to undefined).

KhafraDev (changes requested):
How much does performance improve by only removing the regex?

I don't really like the idea of removing the `for..of` loops because I don't find it maintainable, in the sense that no one will block a PR that uses it over the standard for loop. I don't mind the change and I'm not asking you to revert it though, just mentioning.

metcoder95:
It seems there's a small regression:

**Branch**
image


**Main**
image

Running once more to verify

lib/fetch/headers.js
```diff
@@ -16,6 +16,21 @@ const assert = require('assert')
...
+  switch (code) {
+    case 0x00a:
+    case 0x00d:
+    case 0x009:
+    case 0x020:
+      return true
+    default:
+      return false
+  }
+}
```
Ethan-Arrowood, line 32: If we are going to micro optimize I recommend giving the optimizer a break. Something like this should be a single expression.

```suggestion
function isHTTPWhiteSpaceCharCode (code) {
  return code == 0x00a || code == 0x00d || code == 0x009 || code == 0x020;
}
```

lib/fetch/headers.js
```diff
@@ -55,8 +71,10 @@ function fill (headers, object) {
 
     // 2. Otherwise, object is a record, then for each key → value in object,
     //    append (key, value) to headers
-    for (const [key, value] of Object.entries(object)) {
-      headers.append(key, value)
+    const keys = Object.keys(object)
+    for (let i = 0; i < keys.length; i++) {
+      const key = keys[i]
+      headers.append(key, object[key])
```
Ethan-Arrowood, line 77:
```suggestion
    const entries = Object.entries(object)
    for (let i = 0; i < entries.length; i++) {
      headers.append(entries[i][0], entries[i][1])
```

remove unnecessary `const` declaration

Ethan-Arrowood (changes requested):

lib/fetch/headers.js
KhafraDev: Object.entries is very slow in v8

Ethan-Arrowood: Good to know. I wonder what the diff is between the declaration here. Maybe just removing that in favor of inlining `keys[i]` then

lib/fetch/webidl.js
tsctx: Hello!
I looked into it and Deno already does that.
Please let me know again if I should change it.
https://github.com/denoland/deno/blob/621440017a3fb93dcba6f9153ad150994592de02/ext/webidl/00_webidl.js#L386C66-L386C66

lib/fetch/headers.js
tsctx: Hello!
How about this way?
```js
const kIsHTTPWhitespaceBit = (1 << ' '.charCodeAt(0)) |
(1 << '\n'.charCodeAt(0)) |
(1 << '\r'.charCodeAt(0)) |
(1 << '\t'.charCodeAt(0))

/**
 * @see https://fetch.spec.whatwg.org/#concept-header-value-normalize
 * @param {string} potentialValue
 */
function headerValueNormalize (potentialValue) {
  //  To normalize a byte sequence potentialValue, remove
  //  any leading and trailing HTTP whitespace bytes from
  //  potentialValue.
  let i = 0
  let l = potentialValue.length

  while (l > i && ((1 << potentialValue.charCodeAt(l - 1)) & kIsHTTPWhitespaceBit) !== 0) --l
  while (i > l && ((1 << potentialValue.charCodeAt(i)) & kIsHTTPWhitespaceBit) !== 0) ++i

  return i === 0 && l === potentialValue.length ? potentialValue : potentialValue.substring(i, l)
}
```

lib/fetch/headers.js
tsctx: Hello!
I have removed the variables.

lib/fetch/webidl.js
KhafraDev: I'd prefer if it was removed

tsctx:
Updated benchmarks.

tsctx:
> How much does performance improve by only removing the regex?
> 
> I don't really like the idea of removing the for..of loops because I don't find it maintainable, in the sense that no one will block a PR that uses it over the standard for loop. I don't mind the change and I'm not asking you to revert it though, just mentioning.

This is just a benchmark for removing RegExp.

```
new Headers - record x 161,831 ops/sec ±2.69% (83 runs sampled)
new Headers - entries x 168,051 ops/sec ±4.50% (80 runs sampled)
```

KhafraDev:
so it looks like a majority of the improvement is from removing the regex and removing Object.entries?

tsctx:
@KhafraDev 
You’re right.

lib/fetch/headers.js
tsctx: Thanks for letting me know.

tsctx:
Hello!
New improvements have been made.
The conversion to `ByteString` seems to have already been completed by the webidl converter `sequence<sequence<ByteString>> or record<ByteString, ByteString>`, but it is being converted again.
See the new benchmark below.
[`9d838c5`](https://github.com/tsctx/undici/tree/4592203e7b71721de1d46a8921e00b9687a5c092)
```
new Headers - record x 185,007 ops/sec ±2.16% (82 runs sampled)
new Headers - entries x 165,584 ops/sec ±1.75% (83 runs sampled)
```
[`f0aaa32`](https://github.com/tsctx/undici/tree/f0aaa32ca5925b1ad7d56f4c3bc09685ab34f232)
```
new Headers - record x 255,866 ops/sec ±2.29% (80 runs sampled)
new Headers - entries x 224,652 ops/sec ±1.87% (88 runs sampled)
```

tsctx:
Hi @KhafraDev @Ethan-Arrowood,
Could you please review the code at a time that is convenient for you?

ronag:
@KhafraDev @Ethan-Arrowood

lib/fetch/headers.js
```diff
@@ -67,6 +76,49 @@ function fill (headers, object) {
...
+  }
+
+  // 6. Otherwise, if headers’s guard is "response" and name is a
+  //    forbidden response-header name, return.
+
+  // 7. Append (name, value) to headers’s header list.
+  // 8. If headers’s guard is "request-no-cors", then remove
+  //    privileged no-CORS request headers from headers
+  return headers[kHeadersList].append(name, value)
+}
```
ronag, line 120:
```suggestion
  // 6. Otherwise, if headers’s guard is "response" and name is a
  //    forbidden response-header name, return.

  // 7. Append (name, value) to headers’s header list.
  headers[kHeadersList].append(name, value)
  
  // 8. If headers’s guard is "request-no-cors", then remove
  //    privileged no-CORS request headers from headers
}
```

ronag (approved):

Ethan-Arrowood (approved):
lgtm

KhafraDev:
I have no time to do an actual review for the next 2 days but if the tests pass there is *very* little that can break.

lib/fetch/headers.js
tsctx: Changed in [`11765a`](https://github.com/nodejs/undici/pull/2397/commits/11765a001cf31af1352e869d544c143389c3c4fb)

PR #2403, opened 2023-11-06
feat: allow customization of build environment

khardix:
This allows for the WASM artifacts to be built elsewhere than only in the alpine-based node container.

## Rationale

Pre-built WASM artifacts are often a no-go when creating packages for Linux distributions; they need to be rebuilt from source on trusted build system by the package maintainer. This PR allows me to easily tweak the build parameters and flags for the WASM build to do just that.

## Changes

The `build/wasm.js` script now checks for presence and content of several related environment variables (`WASM_CC`, `WASM_CFLAGS`, `WASM_LDFLAGS`, and `WASM_LDLIBS`; the names mirror those traditionally used in C Makefiles). If present, content of these variables override some of the compilation flags that are directly dependent on the build environment (i.e. `--target` and `--sysroot`).

## Status

- [x] Tested
- [S] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

---

Note: This is a minimal change that does what I need. I was also toying with the idea to convert the build process (without container management) to separate Makefile; this would enable the WASM build to not require nodejs on the build machine. I'm not sure this is worth it or if it's anything you would want to maintain; let me know what you think.

### DIFF
--- a/build/wasm.js
+++ b/build/wasm.js
@@ -9,6 +9,18 @@ const WASM_SRC = resolve(__dirname, '../deps/llhttp')
 const WASM_OUT = resolve(__dirname, '../lib/llhttp')
 const DOCKERFILE = resolve(__dirname, './Dockerfile')
 
+// These are defined by build environment
+const WASM_CC = process.env.WASM_CC || 'clang'
+let WASM_CFLAGS = process.env.WASM_CFLAGS || '--sysroot=/usr/share/wasi-sysroot -target wasm32-unknown-wasi'
+let WASM_LDFLAGS = process.env.WASM_LDFLAGS || ''
+const WASM_LDLIBS = process.env.WASM_LDLIBS || ''
+
+// These are relevant for undici and should not be overridden
+WASM_CFLAGS += ' -Ofast -fno-exceptions -fvisibility=hidden -mexec-model=reactor'
+WASM_LDFLAGS += ' -Wl,-error-limit=0 -Wl,-O3 -Wl,--lto-O3 -Wl,--strip-all'
+WASM_LDFLAGS += ' -Wl,--allow-undefined -Wl,--export-dynamic -Wl,--export-table'
+WASM_LDFLAGS += ' -Wl,--export=malloc -Wl,--export=free -Wl,--no-entry'
+
 let platform = process.env.WASM_PLATFORM
 if (!platform && process.argv[2]) {
   platform = execSync('docker info -f "{{.OSType}}/{{.Architecture}}"').toString().trim()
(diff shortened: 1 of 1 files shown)


metcoder95 (commented):
How are you building `undici` currently?

khardix:
> How are you building `undici` currently?

Currently we are not building it. We are using the bundled blobs in nodejs and attaching corresponding source tarball. I'm working on changing that to rebuild undici as a separate package (rpm).

The work-in-progress way of rebuilding (with the proposed change already applied) then looks something like this:

```sh
rm -rf lib/llhttp/llhttp*.{wasm,js}
export WASM_CC=clang WASM_CFLAGS='--target=wasm32-wasi --sysroot=/usr/wasm32-wasi' WASM_LDFLAGS=-nodefaultlibs WASM_LDLIBS=-lc
node build/wasm.js
npm --offline pack
```

metcoder95:
Sorry if I pop up many questions, but I'm trying to understand what's the use-case behind. What I'm understanding is that you are trying to bundle `undici` as a standalone executable?
Then, to achieve so you require to customize the build process, is that all right?

khardix:
I'm trying to rebuild undici *from source* on "my machine" (Fedora build system). Emphasis on from source – anything pre-built (in this case, the WASM-compiled llhttp) has to go before build and be re-created during the build. This is mainly due to security concerns (preventing supply chain attacks, etc.).

On the build system, I already have some parts natively available (i.e. wasi-libc), while others are not (i.e. docker or network access). I need a way to hook into the build process so I can tell it where to look for the parts it needs and how to invoke them (mainly compiler and it's flags).

---

This effort is a part of larger process of enhancing NodeJS packaging in Fedora. The separate executable/rpm is actually an implementation detail; I would need basically the same changes if I just wanted to rip out and rebuild all pre-build WASM artifacts in NodeJS itself (which bundles undici and thus it's pre-build artifacts).

metcoder95:
Perfect! Thanks, now is clearer the intentions.

On my side, I believe the changes are good, and the defaults are respected; but the are some security concerns given that the paths are now customizable. On my side, LGTM, but would like to hear the thoughts @nodejs/undici

khardix:
No one chimed in for a few weeks now. Tagging @mhdawson which whom I've previously talked about either looking at this or poking others into looking. :slightly_smiling_face:

mhdawson (approved):
LGTM

mhdawson:
Reviewing the PR I can see it should no effect on what is run without additional environment settings versus what will be run before.  

In terms of the CI the failures seem to in `73 - test/redirect-request.js # time=868.622ms`  which I can see in many of the other failings CIs runs (which seems to be most of them), so I don't believe the changes in this PR are related to the failures.

PR #2410, opened 2023-11-08
fix: HTTPToken check

### LINKED ISSUE #2409: Non-HTTPToken characters can be used in header names
## Bug Description

The following character `"(),/:;<=>?@[\]{}` is not allowed in the specification, but is allowed in Undici.

References:
<https://fetch.spec.whatwg.org/#header-name>
<https://tools.ietf.org/html/rfc7230#section-3.2.6>
<https://tools.ietf.org/html/rfc9110#section-5.6.2>

## Reproducible By

There is a simple test.

```js
// test.js
function pass(name) {
    try {
        return !!new Headers({ [name]: "test" });
    } catch (e) {
        return false;
    }
}

function assert(cond, message) {
    if (!cond) console.error(message);
}

function test() {
    for (const char of `"(),/:;<=>?@[\\]{}`.split("")) {
        assert(!pass(char), `The string \`${char}\` should throw an error.`);
    }
}

test();
```

## Expected Behavior

The above test should succeed.

## Logs & Screenshots

```sh
> $ node test.js
The string `"` should throw an error.
The string `(` should throw an error.
The string `)` should throw an error.
The string `,` should throw an error.
The string `/` should throw an error.
The string `:` should throw an error.
The string `;` should throw an error.
The string `<` should throw an error.
The string `=` should throw an error.
The string `>` should throw an error.
The string `?` should throw an error.
The string `@` should throw an error.
The string `[` should throw an error.
The string `\` should throw an error.
The string `]` should throw an error.
The string `{` should throw an error.
The string `}` should throw an error.
```

## Environment

Windows 11
Node.js v21.1.0
Undici v5.27.2

### Additional context

None

tsctx:
## This relates to...

Fixes #2409 

### Features

N/A

### Bug Fixes

#2409 

N/A

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -35,6 +35,7 @@ const {
   isCancelled,
   isAborted,
   isErrorLike,
+  isValidHeaderName,
   fullyReadBody,
   readableStreamClose,
   isomorphicEncode,
@@ -2016,7 +2017,7 @@ async function httpNetworkFetch (
                 location = val
               }
 
-              headers.append(key, val)
+              if (isValidHeaderName(key)) headers.append(key, val)
             }
           }
 
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -2003,7 +2003,7 @@ async function httpNetworkFetch (
                 location = val
               }
 
-              headers.append(key, val)
+              if (isValidHeaderName(key)) headers.append(key, val)
             }
           } else {
             const keys = Object.keys(headersList)
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -35,7 +35,6 @@ const {
   isCancelled,
   isAborted,
   isErrorLike,
-  isValidHeaderName,
   fullyReadBody,
   readableStreamClose,
   isomorphicEncode,
@@ -2003,7 +2002,7 @@ async function httpNetworkFetch (
                 location = val
               }
 
-              if (isValidHeaderName(key)) headers.append(key, val)
+              headers.append(key, val)
             }
           } else {
             const keys = Object.keys(headersList)
@@ -2017,7 +2016,7 @@ async function httpNetworkFetch (
                 location = val
               }
 
-              if (isValidHeaderName(key)) headers.append(key, val)
+              headers.append(key, val)
             }
           }
 
--- a/lib/fetch/util.js
+++ b/lib/fetch/util.js
@@ -134,7 +134,7 @@ function isTokenCharCode (c) {
 }
 
 /**
- * @param {string} characters 
+ * @param {string} characters
  */
 function isValidHTTPToken (characters) {
   if (characters.length === 0) {
@@ -150,7 +150,7 @@ function isValidHTTPToken (characters) {
 
 /**
  * @see https://fetch.spec.whatwg.org/#header-name
- * @param {string} potentialValue 
+ * @param {string} potentialValue
  */
 function isValidHeaderName (potentialValue) {
   return isValidHTTPToken(potentialValue)
--- a/lib/fetch/util.js
+++ b/lib/fetch/util.js
@@ -122,6 +122,7 @@ function isTokenCharCode (c) {
     case 0x3f:
     case 0x40:
     case 0x5b:
+    case 0x5c:
     case 0x5d:
     case 0x7b:
     case 0x7d:
--- a/lib/fetch/util.js
+++ b/lib/fetch/util.js
@@ -103,52 +103,56 @@ function isValidReasonPhrase (statusText) {
   return true
 }
 
-function isTokenChar (c) {
-  return !(
-    c >= 0x7f ||
-    c <= 0x20 ||
-    c === '(' ||
-    c === ')' ||
-    c === '<' ||
-    c === '>' ||
-    c === '@' ||
-    c === ',' ||
-    c === ';' ||
-    c === ':' ||
-    c === '\\' ||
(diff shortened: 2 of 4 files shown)


metcoder95 (commented):
The changes affect `fetch` only, but I wonder if we want to extend it to `undici` itself.

lib/fetch/util.js
```diff
@@ -103,52 +103,57 @@ function isValidReasonPhrase (statusText) {
...
 }
 
-// See RFC 7230, Section 3.2.6.
-// https://github.com/chromium/chromium/blob/d7da0240cae77824d1eda25745c4022757499131/third_party/blink/renderer/platform/network/http_parsers.cc#L321
+/**
+ * @param {string} characters
+ */
 function isValidHTTPToken (characters) {
-  if (!characters || typeof characters !== 'string') {
+  if (characters.length === 0) {
```
ronag, line 141: What's the point of this change?

tsctx:
@metcoder95 
Do you mean checking all append headers?

lib/fetch/util.js
tsctx: If the length of the empty string was checked explicitly, it would make it easier to understand.
Also, The string check was removed because only `isValidHeaderName` was used.

tsctx: Since it is called after being stringified, I don't think there is a problem.

lib/fetch/index.js
```diff
@@ -2002,7 +2003,7 @@ async function httpNetworkFetch (
                 location = val
               }
 
-              headers.append(key, val)
+              if (isValidHeaderName(key)) headers.append(key, val)
```
KhafraDev, line 2006: are these headers expected to be ignored, or should an error be thrown? It doesn't make sense to double validate every header we receive.

lib/fetch/index.js
```diff
@@ -2016,7 +2017,7 @@ async function httpNetworkFetch (
...
```
KhafraDev, line 2020: same here

lib/fetch/util.js
```diff
@@ -103,52 +103,57 @@ function isValidReasonPhrase (statusText) {
...
-    c === '?' ||
-    c === '=' ||
-    c === '{' ||
-    c === '}'
-  )
+/**
+ * @see https://tools.ietf.org/html/rfc7230#section-3.2.6
+ * @param {number} c
+ */
+function isTokenCharCode (c) {
```
KhafraDev, line 110: this seems to be the only required change here (since the issue seems to be comparing the char code to the actual characters would always return false). I'd prefer if the old checks were added back and only this change made.

lib/fetch/index.js
tsctx: I too think double checking is useless.
If `:status` is passed as an example of an http2 header, the http2 test fails.
I will comment the solution later.

tsctx: Some solutions,
- Change core so that http2 headers, e.g. `:status`, are not passed.
- Use try-catch.
- Check only the leading colon.

lib/fetch/index.js
tsctx: Explained in the comment above.

lib/fetch/index.js
metcoder95: None of the specific `h2` headers should be passed as those are automatically appended upon request. Although it can be passed, they are going to be overwritten.
On Chrome, `fetch` throws when trying to append them as headers

tsctx: @metcoder95
Is it correct to identify pseudo-headers with a leading colon and not pass them to Headers#append?
Or change the core?

tsctx: Probably due to the :status pseudo-header.

tsctx: Probably due to the `:status` pseudo-header.

tsctx: The pseudo-header `:status` seems to be the cause

KhafraDev: please undo both of these changes, we can reevaluate in a dedicated issue

tsctx: Sure, changes will be undone.

metcoder95 (approved):
LGTM

lib/fetch/util.js
KhafraDev: @tsctx I don't believe this should be blocked by anything. The handling of pseudo-headers has nothing to do with the original issue that this fixes, right?

tsctx: @KhafraDev 
This change will fail the H2 test and should wait.

tsctx: But, I understand your thinking.

KhafraDev: The test isn't sound with my understanding of the fetch spec, I think the failing test(s) can be disabled for now.

KhafraDev: I posted more about my understanding of the fetch spec in the linked issue.

tsctx: @KhafraDev 
After the errands are completed, change the failing H2 test to be skipped.

tsctx:
Hi @metcoder95 @KhafraDev,
I've changed it to skip the failed H2 test.
Could you please take a look at it?

KhafraDev (approved):

metcoder95 (approved):

PR #2414, opened 2023-11-09
refactor: remove await before return

ywave620:
## This relates to...

no issues

## Rationale

await before return is not necessary and introduce a little overhead.

## Status

- [x] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -504,7 +504,7 @@ async function lazyllhttp () {
     mod = await WebAssembly.compile(Buffer.from(llhttpWasmData || require('./llhttp/llhttp-wasm.js'), 'base64'))
   }
 
-  return await WebAssembly.instantiate(mod, {
+  return WebAssembly.instantiate(mod, {
     env: {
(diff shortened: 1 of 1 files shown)


lib/client.js
```diff
@@ -504,7 +504,7 @@ async function lazyllhttp () {
     mod = await WebAssembly.compile(Buffer.from(llhttpWasmData || require('./llhttp/llhttp-wasm.js'), 'base64'))
   }
 
-  return await WebAssembly.instantiate(mod, {
+  return WebAssembly.instantiate(mod, {
```
metcoder95, line 507: @ronag I imagine this is intentional to keep the stacktrace of the `lazyhttp` function, or it is safe to remove?

ronag: I guess it should be safe, but don't really see any reason to do so. It's a cold path.

ronag:
Thanks for the PR but I don't think this is a good change.

Issue #2415, opened 2023-11-09
The pseudo-header `:status` is included in h2 fetch.


tsctx:
## Bug Description

The pseudo-header `:status` can be obtained by h2 fetch.
A fix for this problem is needed in #2410.

## Reproducible By

There is a simple test.

```js
// test.js
(async () => {
  const http2 = require("node:http2");

  const undici = require("undici");

  const serverOptions = {
    key:
      "-----BEGIN RSA PRIVATE KEY-----\r\n" +
      "MIICXAIBAAKBgQCqK4GJEFelA5sWipPsBgONPuwxAL3io9GexKlTv6YPsXkfbgkP\r\n" +
      "a8P6zbzqt4eOVjFtDyqrMkJDmc4gRqGNk3+VzOPqaTLcrvOgDAFfY9eF+Z4KmVut\r\n" +
      "JYTYE/3antTOtcP5yAfVVxkpErvvNhE+0+Vtro6FpfGu93iWGsALtM6p5QIDAQAB\r\n" +
      "AoGAfjyB/pTR/8baAPHSZA/eCKT6ACwlsQCH23jWiBsu7ItiaA9vxzHRk7eWAqUd\r\n" +
      "eVi1l8rdg1xoTc/GWqkq5WTbCqbfBE23UQF65wXcBg+sB7aPcpK48PsnIKM3x0ty\r\n" +
      "Xe33hkSQUn9WzH6TTN0b9pFOHb66bHpmY0bM+3g6xxA4ByECQQDTZuw3opoKgt8t\r\n" +
      "tVNnOeOCgcqwIsZZAv97WhKLPJw1iVj5zSJA+fOnskGHevHBByhTPG2Fxjosv4kB\r\n" +
      "wbFAatktAkEAzhHIjyLWG55LF8V3wqlb9IQj4ufy9Oz+GgUSfU83+w2ZsPRW8ywY\r\n" +
      "Nn40dZe4H3cgjPg2vF2Iw3R0QtafWw0WmQJBAJDfR6IY3Fe3uBoq2S00jqUxLSXn\r\n" +
      "ruXxGAQAhhTFRIeryD3oRkEhCckyb5XYuzP9I7VnA+dZISTpXYrGoiHzF0ECQFGG\r\n" +
      "p5o5Zgu3LG5TKXZSGpc6UBhNO8m6mS0iJJZ+6N7iGpPvlxbNsyab95UuWSHZt38O\r\n" +
[...]

PR #2423, opened 2023-11-11
fix: `:status` is not a string

tsctx:
## Rationale

The pseudo-headers `:status` numeric and must be converted to a string.

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -2002,7 +2002,7 @@ async function httpNetworkFetch (
                 location = val
               }
 
-              headers[kHeadersList].append(key, val)
+              headers[kHeadersList].append(key, String(val))
             }
           } else {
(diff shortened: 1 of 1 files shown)


tsctx:
cc @KhafraDev

KhafraDev (changes requested):
As I mentioned in [my comment](https://github.com/nodejs/undici/issues/2415#issuecomment-1806508566), pseudo (pseudo being synonymous with "fake") headers are not header fields, and shouldn't be treated as such. The issue is therefore no longer with fetch, but with how undici handles pseudo headers.

tsctx:
@KhafraDev
I understand. I will close this PR.

PR #2424, opened 2023-11-11
perf(headers): if the guard is immutable

tsctx:
## Rationale

If the guard is immutable, it cannot be deleted or added, so it can be safely cached.
It is also a 1.5x performance improvement over *main*.
This change speeds up the header loop of the fetch response.

## Status

- [x] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [x] Merge ready


KhafraDev (approved):
makes sense

PR #2434, opened 2023-11-14
ci: unpin nodejs workflow version

dominykas:
Not 100% sure if it was pinned deliberately earlier - there are several other things that are pinned to major only, and I think that's fine. When this is pinned, it might sometimes not receive new versions of Node.js (as the release dates are cached inside the action, as opposed to getting resolved each time).

For some reason, dependabot did not open a PR here 🤷

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -15,7 +15,7 @@ on:
 jobs:
   build:
     name: Test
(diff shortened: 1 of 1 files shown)


metcoder95:
@KhafraDev @mcollina @ronag @Ethan-Arrowood any thoughts?

mcollina (approved):
lgtm

ronag (approved):

KhafraDev (approved):

Issue #2453, opened 2023-11-23
The test of `test/client-dispatch.js` fails.


tsctx:
## Bug Description

After the https://github.com/nodejs/undici/commit/e495094b3700690f771089f1a3c338dd4da01b63, the test fails.

## Reproducible By

```
> $ npx tap test/client-dispatch.js
```
See #Logs & Screenshots

## Logs & Screenshots

- https://github.com/nodejs/undici/commit/e495094b3700690f771089f1a3c338dd4da01b63

```
(node:11348) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.                                                                                              
(Use `node --trace-deprecation ...` to show where the warning was created)                                  
test/client-dispatch.js 2> npm exec tap test/client-dispatch.js[14112]: c:\ws\src\api\callback.cc:145:      
test/client-dispatch.js 2> _async_id()) == (0)' failed.
​ FAIL ​ test/client-dispatch.js
 ✖ fail

  test/client-dispatch.js

              onBodySent (chunk) {
  802 |         throw new Error('fail')
      | --------------^
                         
              onError (err) {
                t.type(err, Error)

  test: dispatch onBodySent throws error
  stack: |
    Object.onBodySent (test/client-dispatch.js:802:15)
    Request.onBodySent (lib/core/request.js:232:29)
    write (lib/client.js:1634:13)
    _resume (lib/client.js:1478:29)
    resume (lib/client.js:1352:3)
    connect (lib/client.js:1337:3)
  tapCaught: unhandledRejection

test/client-dispatch.js 2>  1: 00007FF649E6C1FF node::SetCppgcReference+13695
test/client-dispatch.js 2>  2: 00007FF649DE4E16 DSA_meth_get_flags+75686
test/client-dispatch.js 2>  3: 00007FF649DE5212 DSA_meth_get_flags+76706
test/client-dispatch.js 2>  4: 00007FF649EA4C99
test/client-dispatch.js 2> ackScope+841
test/client-dispatch.js 2>  5: 00007FF649EA49BE
test/client-dispatch.js 2> ackScope+110
test/client-dispatch.js 2>  6: 00007FF649E62A1B node_api_throw_syntax_error+166091
 
test/client-dispatch.js 2>  7: 00007FF649E4AE7A node_api_throw_syntax_error+68906
test/client-dispatch.js 2>  8: 00007FF649EC6D71 uv_pipe_pending_type+849
[...]

PR #2454, opened 2023-11-23
perf(request): optimize if headers are given

tsctx:
## Rationale

For performance improvement:
1. Do not call `Object.keys` more than once.
2. When copying the `headersList`, use `HeadersList` instead of `Headers`.
3. Avoid call `HeadersList` if init["headers"] is not undefined
4. When looping through headers, use `HeadersList` instead of `Headers`, which eliminates the sorting process and is faster.

## Benchmark

```js
// bench.js
const { Request } = require(".");

const headers = {
  "X-Content-Type": "application/json",
  "X-NodeJS-Version": "v21.1.0",
  "Powered-By": "NodeJS",
  "Content-Encoding": "gzip",
  "Content-Length": "150",
};

const headersEntries = Object.entries(headers);
suite.add("new Request() - record", () => {
  new Request("https://example.com/post", {
    headers: headers,
  });
});

suite.add("new Request() - record without constructor", () => {
  const request = new Request("https://example.com/post", {});
  const h = request.headers
  const keys = Object.keys(headers);
  for (let i = 0; i < keys.length; ++i) {
    h.append(keys[i], headers[keys[i]]);
  } 
})

suite.add("new Request() - entires", () => {
  new Request("https://example.com/post", {
    headers: headersEntries,
  });
});

suite.add("new Request() - entires without constructor", () => {
  const request = new Request("https://example.com/post", {});
  const h = request.headers
  for (let i = 0; i < headersEntries.length; ++i) {
    h.append(headersEntries[i][0], headersEntries[i][1]);
  }
});

});

// run async
suite.run({ async: true });
```

- *v5.27.2*
```
new Request() - record x 34,814 ops/sec ±3.33% (75 runs sampled)
new Request() - record without constructor x 64,510 ops/sec ±2.22% (85 runs sampled)
new Request() - entires x 34,316 ops/sec ±3.31% (77 runs sampled)
new Request() - entires without constructor x 61,662 ops/sec ±3.00% (79 runs sampled)
```

- *main*
```
new Request() - record x 41,193 ops/sec ±5.16% (75 runs sampled)
new Request() - record without constructor x 65,217 ops/sec ±2.69% (80 runs sampled)
new Request() - entires x 38,771 ops/sec ±2.78% (78 runs sampled)
new Request() - entires without constructor x 68,280 ops/sec ±2.86% (83 runs sampled)
```

- *this patch*
```
new Request() - record x 68,534 ops/sec ±2.08% (80 runs sampled)
new Request() - record without constructor x 69,161 ops/sec ±2.31% (83 runs sampled)
new Request() - entires x 65,713 ops/sec ±2.13% (84 runs sampled)
new Request() - entires without constructor x 67,820 ops/sec ±2.57% (79 runs sampled)
```

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/fetch/request.js
+++ b/lib/fetch/request.js
@@ -440,7 +440,7 @@ class Request {
         for (const [key, val] of headers) {
           headersList.append(key, val)
         }
-        // Note: Copy the `set-cooke` meta-data.
+        // Note: Copy the `set-cookie` meta-data.
         headersList.cookies = headers.cookies
       } else {
         // 5. Otherwise, fill this’s headers with headers.
--- a/lib/fetch/request.js
+++ b/lib/fetch/request.js
@@ -423,6 +423,7 @@ class Request {
       // 1. Let headers be a copy of this’s headers and its associated header
       // list.
       let headers
+
       // 2. If init["headers"] exists, then set headers to init["headers"].
       if (init.headers !== undefined) {
         headers = init.headers
--- a/lib/fetch/request.js
+++ b/lib/fetch/request.js
@@ -183,8 +183,10 @@ class Request {
       urlList: [...request.urlList]
     })
 
+    const initHasKey = Object.keys(init).length !== 0
+
     // 13. If init is not empty, then:
-    if (Object.keys(init).length > 0) {
+    if (initHasKey) {
       // 1. If request’s mode is "navigate", then set it to "same-origin".
       if (request.mode === 'navigate') {
         request.mode = 'same-origin'
@@ -415,23 +417,23 @@ class Request {
     }
 
     // 32. If init is not empty, then:
-    // Note: Avoid call `Object.keys` if init["headers"] is not undefined.
-    if (init.headers !== undefined || Object.keys(init).length !== 0) {
+    if (initHasKey) {
+      const headersList = this[kHeaders][kHeadersList]
       // 1. Let headers be a copy of this’s headers and its associated header
       // list.
       // 2. If init["headers"] exists, then set headers to init["headers"].
       // Note: Avoid call `HeadersList` if init["headers"] is not undefined
-      const headers = init.headers !== undefined ? init.headers : new HeadersList(request.headersList)
+      const headers = init.headers !== undefined ? init.headers : new HeadersList(headersList)
 
       // 3. Empty this’s headers’s header list.
       // Note: Avoid call `HeadersList#clear` if already empty
-      if (request.headersList.mapSize !== 0) request.headersList.clear()
+      if (headersList.mapSize !== 0) headersList.clear()
 
       // 4. If headers is a Headers object, then for each header in its header
       // list, append header’s name/header’s value to this’s headers.
       // Note: If init["headers"] is undefined, then it is `HeadersList`.
       if (init.headers === undefined) {
-        headersListCopy(request.headersList, headers)
+        headersListCopy(headersList, headers)
       } else {
         // 5. Otherwise, fill this’s headers with headers.
         fillHeaders(this[kHeaders], headers)
--- a/lib/fetch/request.js
+++ b/lib/fetch/request.js
@@ -437,8 +437,8 @@ class Request {
       // 4. If headers is a Headers object, then for each header in its header
       // list, append header’s name/header’s value to this’s headers.
       if (headers instanceof HeadersList) {
-        for (const [key, value] of headers) {
-          headersList.set(key, value)
+        for (const [key, val] of headers) {
+          headersList.set(key, val)
         }
         headersList.cookies = headers.cookies
       } else {
--- a/lib/fetch/request.js
+++ b/lib/fetch/request.js
@@ -440,6 +440,7 @@ class Request {
         for (const [key, val] of headers) {
           headersList.set(key, val)
         }
+        // Note: Copy the `set-cooke` meta-data.
         headersList.cookies = headers.cookies
       } else {
         // 5. Otherwise, fill this’s headers with headers.
--- a/lib/fetch/request.js
+++ b/lib/fetch/request.js
@@ -422,8 +422,8 @@ class Request {
       const headersList = this[kHeaders][kHeadersList]
       // 1. Let headers be a copy of this’s headers and its associated header
       // list.
-      // 2. If init["headers"] exists, then set headers to init["headers"].
       let headers
+      // 2. If init["headers"] exists, then set headers to init["headers"].
       if (init.headers !== undefined) {
         headers = init.headers
       } else {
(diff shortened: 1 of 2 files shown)


tsctx:
For test failures, see https://github.com/nodejs/undici/issues/2453.

KhafraDev (changes requested):
Don't have time to review but going against the spec is a -1 from me

mcollina:
Does this pass the same number of wpts?

tsctx:
@mcollina 
please wait a moment

tsctx:
@mcollina 
See here
- *main*
```
[fetch]: Completed: 1453, failed: 165, success: 1288, expected failures: 164, unexpected failures: 1, skipped: 71
[FileAPI]: Completed: 406, failed: 16, success: 390, expected failures: 16, unexpected failures: 0, skipped: 3
[mimesniff]: Completed: 1898, failed: 872, success: 1026, expected failures: 872, unexpected failures: 0, skipped: 0
[xhr/formdata]: Completed: 39, failed: 0, success: 39, expected failures: 0, unexpected failures: 0, skipped: 0
[websockets]: Completed: 500, failed: 18, success: 482, expected failures: 18, unexpected failures: 0, skipped: 10
```
- *this patch*
```
[fetch]: Completed: 1453, failed: 165, success: 1288, expected failures: 164, unexpected failures: 1, skipped: 71
[FileAPI]: Completed: 406, failed: 16, success: 390, expected failures: 16, unexpected failures: 0, skipped: 3
[mimesniff]: Completed: 1898, failed: 872, success: 1026, expected failures: 872, unexpected failures: 0, skipped: 0
[xhr/formdata]: Completed: 39, failed: 0, success: 39, expected failures: 0, unexpected failures: 0, skipped: 0
[websockets]: Completed: 500, failed: 18, success: 482, expected failures: 18, unexpected failures: 0, skipped: 10
```

tsctx:
It is no different from *main*.

mcollina (approved):
lgtm

@KhafraDev where is the spec breakage?

KhafraDev:
Spec breakage isn't the concern here, it's not following the spec directly that worries me. If for example those lines are changed in the spec, how do we upstream the changes here? It makes maintaining it a nightmare. If the change is completely compatible with the spec, I don't see a reason that the changes couldn't be made in the fetch spec itself, and then made here.

KhafraDev:
I have no problem with something like setting the value of `Object.keys(init).length` to a variable, but right now the comments from the spec are no longer what undici is doing. I promise that if we don't follow the spec directly, it'll lead to an unmaintainable mess (ie. look at basically any spec node core implements).

mcollina:
@KhafraDev these APIs are called by a _lot_, making a small improvement noticeable perf-wise.

lib/fetch/request.js
```diff
@@ -418,22 +418,28 @@ class Request {
...
+      if (init.headers !== undefined) {
+        headers = init.headers
+      } else {
+        init.headers = new HeadersList(headersList)
+      }
 
       // 3. Empty this’s headers’s header list.
-      // Note: Avoid call `HeadersList#clear` if already empty
-      if (headersList.mapSize !== 0) headersList.clear()
+      headersList.clear()
```
ronag, line 435: Does  this actaully make a difference?

tsctx: Checking the Map size and not clearing is twenty times faster if the map is empty.
However, to lower maintenance costs, this change has been reverted.
If you want it back, I will gladly undo it.

tsctx:
The new changes lowered maintenance costs by making fewer changes while maintaining performance.
It is now better suited to the specifications than previous PR implementations.

lib/fetch/request.js
```diff
@@ -415,25 +417,31 @@ class Request {
...
-      this[kHeaders][kHeadersList].clear()
+      headersList.clear()
 
       // 4. If headers is a Headers object, then for each header in its header
       // list, append header’s name/header’s value to this’s headers.
-      if (headers.constructor.name === 'Headers') {
+      if (headers instanceof HeadersList) {
         for (const [key, val] of headers) {
-          this[kHeaders].append(key, val)
+          headersList.set(key, val)
```
ronag, line 441: This doesn't follow the spec though. It says append? Are we sure this is the same?

tsctx: There is no implementation problem, but append should be used according to the specification.
p.s. I missed it.

ronag (approved):

lib/fetch/request.js
```diff
@@ -415,25 +417,31 @@ class Request {
...
 
     // 32. If init is not empty, then:
-    if (Object.keys(init).length !== 0) {
+    if (initHasKey) {
+      /** @type {HeadersList} */
+      const headersList = this[kHeaders][kHeadersList]
       // 1. Let headers be a copy of this’s headers and its associated header
       // list.
-      let headers = new Headers(this[kHeaders])
+      let headers
```
KhafraDev, line 425: this comment no longer matches what is happening. Instead, you can combine steps 1 & 2 into a single step and have both comments come before this:

```js
const headers = init.headers !== undefined ? init.headers : new HeadersList(headersList)
```

lib/fetch/request.js
```diff
@@ -415,25 +417,31 @@ class Request {
...
       }
 
       // 3. Empty this’s headers’s header list.
-      this[kHeaders][kHeadersList].clear()
+      headersList.clear()
 
       // 4. If headers is a Headers object, then for each header in its header
       // list, append header’s name/header’s value to this’s headers.
-      if (headers.constructor.name === 'Headers') {
+      if (headers instanceof HeadersList) {
```
KhafraDev, line 439: I wouldn't be opposed to cloning the map itself, and avoid the loop. (ie. `new Map(new Map([['a', 'b']]))`)

lib/fetch/request.js
tsctx: Change with 47d95b793e479e3b1a8b3115aecef918716a060e

lib/fetch/request.js
tsctx: I tried to avoid loops too, but I wasn't very wise, so I stopped.
If you do this, it will look like this.
```js
    // 32. If init is not empty, then:
    if (initHasKey) {
      /** @type {HeadersList} */
      const headersList = this[kHeaders][kHeadersList]
      // 1. Let headers be a copy of this’s headers and its associated header
      // list.
      // 2. If init["headers"] exists, then set headers to init["headers"].
      const headers = init.headers !== undefined ? init.headers : new HeadersList(headersList)

      headersList.clear()

      if (headers instanceof HeadersList) {
        this[kHeaders][kHeaders] = headers
        // Note: important !!
        request.headersList = headers
      } else {
        // 5. Otherwise, fill this’s headers with headers.
        fillHeaders(this[kHeaders], headers)
      }
    }
```

KhafraDev (approved):
I don't feel comfortable landing this while the CI isn't working, but the change looks fine to me now.

PR #2455, opened 2023-11-25
fix: Add `null` type to `signal` in `RequestInit`

### LINKED ISSUE #2448: `signal` is not nullable in types of `RequestInit`
## Bug Description

According to the Fetch standard, the [`signal`](https://fetch.spec.whatwg.org/#dom-requestinit-signal) property in `RequestInit` might be nullable. However, currently, `@types/node` that use typings from this package do not allow setting that property to `null` and cause TypeScript errors.

## Reproducible By

1. Create a new package with the following dependencies: `@types/node` and `typescript`.
2. Create a `tsconfig.json` file with the following content:

    ```ts
    {
        "compilerOptions": {
            "lib": ["esnext"],
            "module": "node16",
            "strict": true
            "target": "esnext"
        },
        "files": ["./main.mts"]
    }
    ```
3. Create a `main.mts` file with the following content:

    ```ts
    const response = await fetch('http://localhost:3000', { signal: null });
    ```
4. Compile the project using `tsc`.

## Expected Behavior

There should be no error and TypeScript should compile `main.mts`.

## Logs & Screenshots

This produces the following error:

```txt
main.mts:1:57 - error TS2322: Type 'null' is not assignable to type 'AbortSignal | undefined'.

1 const response = await fetch('http://localhost:3000', { signal: null });
```

## Environment

Node v18.18.2, `@types/node` v20.9.3 (with `undici-types` v5.26.5)

### Additional context

This works fine when using adding `"dom"` to the `lib` option in `tsconfig.json` (or when using [`@types/web`](https://www.npmjs.com/package/@types/web)) as the built-in DOM types allow setting `signal` to `null`:

```ts
// Other properties omitted for brevity.
interface RequestInit {
    /** An AbortSignal to set request's signal. */
    signal?: AbortSignal | null;
}
```

This error makes it impossible to write concise app-specific `fetch` wrappers:

```ts
async function sendFile(file: File, signal?: AbortSignal): Promise<undefined> {
    return fetch('http://localhost:3000/upload, {
        method: 'POST',
        body: file,
        signal: signal ?? null,
    });
}
```

`undefined` can be used instead of `null` as the right-hand side of the nullish coalescing operator in the preceding snippet, but only when the `exactOptionalPropertyTypes` TypeScript compiler option is disabled. When combined with this flag, I'm forced to create fetch options in a verbose way:

```ts
async function sendFile(file: File, signal?: AbortSignal): Promise<undefined> {
    const options: RequestInit = {
        method: 'POST',
        body: file,
    };

    if ( signal !== undefined ) {
        options.signal = signal;
    }

    return fetch('http://localhost:3000/upload, options);
}
```

gebsh:
## This relates to...

Fixes #2448.

### Features

N/A

### Bug Fixes

- Add the missing `null` type to the `signal` property in `RequestInit`

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/test/types/fetch.test-d.ts
+++ b/test/types/fetch.test-d.ts
@@ -42,7 +42,7 @@ expectType<HeadersInit | undefined>(requestInit.headers)
 expectType<BodyInit | undefined>(requestInit.body)
 expectType<RequestRedirect | undefined>(requestInit.redirect)
 expectType<string | undefined>(requestInit.integrity)
-expectType<AbortSignal | undefined>(requestInit.signal)
+expectType<AbortSignal | null | undefined>(requestInit.signal)
 expectType<RequestCredentials | undefined>(requestInit.credentials)
 expectType<RequestMode | undefined>(requestInit.mode)
 expectType<string | undefined>(requestInit.referrer);
--- a/types/fetch.d.ts
+++ b/types/fetch.d.ts
@@ -108,7 +108,7 @@ export interface RequestInit {
   body?: BodyInit
   redirect?: RequestRedirect
   integrity?: string
-  signal?: AbortSignal
+  signal?: AbortSignal | null
   credentials?: RequestCredentials
   mode?: RequestMode
   referrer?: string


KhafraDev (approved):

metcoder95 (commented):
Tests seems failing, can you also rebase with `main`?

gebsh:
> Tests seems failing, can you also rebase with `main`?

I've rebased my branch.

PR #2456, opened 2023-11-25
perf: Improve `normalizeMethod`

tsctx:
## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/fetch/util.js
+++ b/lib/fetch/util.js
@@ -713,7 +713,7 @@ const normalizeMethodRecord = {
   PUT: 'PUT'
 }
 
-// Note: object prototypes should not be able to be referenced. e.g. `Object#hasOwnProperty`
+// Note: object prototypes should not be able to be referenced. e.g. `Object#hasOwnProperty`.
 Object.setPrototypeOf(normalizeMethodRecord, null)
 
 /**
@@ -721,7 +721,6 @@ Object.setPrototypeOf(normalizeMethodRecord, null)
  * @param {string} method
  */
 function normalizeMethod (method) {
-  // Note: `String#toLowerCase` faster than `String#toUpperCase`.
   return normalizeMethodRecord[method] ?? normalizeMethodRecord[method.toLowerCase()] ?? method
 }
 
--- a/lib/fetch/util.js
+++ b/lib/fetch/util.js
@@ -721,7 +721,7 @@ Object.setPrototypeOf(normalizeMethodRecord, null)
  * @param {string} method
  */
 function normalizeMethod (method) {
-  return normalizeMethodRecord[method] ?? normalizeMethodRecord[method.toLowerCase()] ?? method
+  return normalizeMethodRecord[method.toLowerCase()] ?? method
 }
 
 // https://infra.spec.whatwg.org/#serialize-a-javascript-value-to-a-json-string
@@ -1066,5 +1066,6 @@ module.exports = {
   urlIsLocal,
   urlHasHttpsScheme,
   urlIsHttpHttpsScheme,
-  readAllBytes
+  readAllBytes,
+  normalizeMethodRecord
 }
--- a/lib/fetch/util.js
+++ b/lib/fetch/util.js
@@ -698,7 +698,7 @@ function isCancelled (fetchParams) {
     fetchParams.controller.state === 'terminated'
 }
 
-const normalizeMethodRecord = {
+const normalizeMethodMap = new Map(Object.entries({
   delete: 'DELETE',
   DELETE: 'DELETE',
   get: 'GET',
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

ronag:
Benchmark?

lib/fetch/util.js
```diff
@@ -698,11 +698,30 @@ function isCancelled (fetchParams) {
...
+  options: 'OPTIONS',
+  OPTIONS: 'OPTIONS',
+  post: 'POST',
+  POST: 'POST',
+  put: 'PUT',
+  PUT: 'PUT'
+}
+
+// Note: object prototypes should not be able to be referenced. e.g. `Object#hasOwnProperty`.
+Object.setPrototypeOf(normalizeMethodRecord, null)
```
KhafraDev, line 717: Can you add a test that uses fetch that attempts to access something on Object.prototype?

KhafraDev (changes requested):

lib/fetch/util.js
tsctx: Unit tests are added with 41c1e743ad55a63fc87d8d9c0873a9cbb7caa059

KhafraDev (approved):
Changes look fine but lets wait for CI to be fixed before landing.

tsctx:
Hi @ronag, 
See this benchmark.
```js
// bench.js
const { Request } = require("./lib/fetch/request");

const input = "https://example.com/post";

const init = {
  method: "POST",
};

suite.add("new Request()", () => {
  new Request(input, init);
});

});

// run async
  suite.run({ async: true })
);
```

- *main*
```
new Request() x 42,952 ops/sec ±5.67% (81 runs sampled)
```
- *this patch*
```
new Request() x 43,595 ops/sec ±3.76% (81 runs sampled)
```

ronag (approved):

tsctx:
Listing `['delete', 'DELETE', 'get', 'GET', 'head', 'HEAD', 'options', 'OPTIONS', 'post', 'POST', 'put', 'PUT']` below is 2.5 times faster than the current regex implementation.
But otherwise it is about 1.6 times slower.
Most of the time, I use the ones on this list, so it's not a problem.

ronag:
Patch is missing

tsctx:
Furthermore, if you want to pursue performance, you can use cache in Map.
This is the fastest way.
```js
const normalizeMethodMap = new Map(
  Object.entries({
    delete: "DELETE",
    DELETE: "DELETE",
    get: "GET",
    GET: "GET",
    head: "HEAD",
    HEAD: "HEAD",
    options: "OPTIONS",
    OPTIONS: "OPTIONS",
    post: "POST",
    POST: "POST",
    put: "PUT",
    PUT: "PUT",
  })
);

let methodMapCache = new Map(normalizeMethodMap);

function impl_map_cache(method) {
  const cache = methodMapCache.get(method);
  if (cache !== undefined) return cache;
  if (methodMapCache.size > 4096) methodMapCache = new Map(normalizeMethodMap);
  const c = normalizeMethodMap.get(method.toLowerCase()) ?? method;
  methodMapCache.set(method, c);
  return c;
}
```

tsctx:
@ronag 
Adding a PATCH is nice, but you need two objects.

ronag:
> @ronag Adding a PATCH is nice, but you need two objects.

Sorry? I don't understand.

ronag (approved):

tsctx:
@ronag 
In short, it comes down to this.
```js
const normalizeMethodRecordBase = {
  delete: "DELETE",
  DELETE: "DELETE",
  get: "GET",
  GET: "GET",
  head: "HEAD",
  HEAD: "HEAD",
  options: "OPTIONS",
  OPTIONS: "OPTIONS",
  post: "POST",
  POST: "POST",
  put: "PUT",
  PUT: "PUT",
};

const normalizeMethodRecord = {
  ...normalizeMethodRecordBase,
  patch: "patch",
  PATCH: "PATCH",
};

Object.setPrototypeOf(normalizeMethodRecordBase, null);

Object.setPrototypeOf(normalizeMethodRecord, null);

function impl_object(method) {
  return (
    normalizeMethodRecord[method] ??
    // Here, don't use normalizeMethodRecord, otherwise `PATCH` becomes `patch`.
    normalizeMethodRecordBase[method.toLowerCase()] ??
    method
  );
}
```

ronag:
I don't understand. Why not just add it to the first object?

ronag:
```
const normalizeMethodRecordBase = {
  delete: "DELETE",
  DELETE: "DELETE",
  get: "GET",
  GET: "GET",
  head: "HEAD",
  HEAD: "HEAD",
  options: "OPTIONS",
  OPTIONS: "OPTIONS",
  post: "POST",
  POST: "POST",
  put: "PUT",
  PUT: "PUT",
  patch: "PATCH",
  PATCH: "PATCH",
};
```

tsctx:
The specification reads as follows.
```
To normalize a [method](https://fetch.spec.whatwg.org/#concept-method), if it is a [byte-case-insensitive](https://infra.spec.whatwg.org/#byte-case-insensitive) match for `DELETE`, `GET`, `HEAD`, `OPTIONS`, `POST`, or `PUT`, [byte-uppercase](https://infra.spec.whatwg.org/#byte-uppercase) it.
```
It does not say that `patch` is to be `PATCH`.

tsctx:
An example specification states.
Using `patch` is highly likely to result in a `405 Method Not Allowed`. `PATCH` is much more likely to succeed.

PR #2457, opened 2023-11-25
perf: optimize HeadersList.clear

tsctx:
## Rationale

Checking the Map size and not clearing is thirty times faster if the map is empty.

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/fetch/headers.js
+++ b/lib/fetch/headers.js
@@ -146,7 +146,7 @@ class HeadersList {
   }
 
   clear () {
-    this[kHeadersMap].clear()
+    if (this[kHeadersMap].size) this[kHeadersMap].clear()
     this[kHeadersSortedMap] = null
     this.cookies = null
   }


KhafraDev (commented):
I'm all for optimizing, but when would someone ever clear an empty Headers? It seems like a case that would never happen.

tsctx:
@KhafraDev https://github.com/nodejs/undici/pull/2454#discussion_r1404768634

tsctx:
The headersList is empty unless the first argument of the Request constructor is Request

tsctx:
@KhafraDev The same logic is used here.
https://github.com/nodejs/undici/blob/0437f69d186e7144e726f0657e5cda08b418606b/lib/fetch/headers.js#L230-L234

ronag:
> @KhafraDev The same logic is used here.
> 
> https://github.com/nodejs/undici/blob/0437f69d186e7144e726f0657e5cda08b418606b/lib/fetch/headers.js#L227-L237

That's not quite the same. It avoids creating an unnecessary iterator. But yea, I would argue it's unnecessary there as well.

tsctx:
It is the same in that it avoids unnecessary function execution.

tsctx:
On second thought, this does not seem necessary.

KhafraDev:
That optimization makes much more sense because headers received from the server are immutable

tsctx:
I'm sorry. This PR thing?

tsctx:
I'm sorry. Are you referring to this PR?

PR #2460, opened 2023-11-26
perf: add `patch` and `PATCH`

tsctx:
cc @ronag 
I reconsidered. I still think it should be added.


ronag:
@KhafraDev not in spec, wdyt?

KhafraDev (changes requested):
no, it's explicitly not normalized in the spec for a reason. I can find the issue if you want.

PR #2461, opened 2023-11-26
perf(request): window check once

tsctx:


KhafraDev:
I don't see the benefit in doing this.

tsctx:
@KhafraDev 
Since `init.window` is not present in most cases, one less if statement can be used, and micro-optimization can be expected.

KhafraDev:
that is exactly why I don't see the benefit in this change.

tsctx:
You should do everything you can.

tsctx:
@KhafraDev 
I think we should do what we can.

KhafraDev:
I don't think we should make unnecessary changes

PR #2462, opened 2023-11-26
refactor: better integrity check

tsctx:
If the integrity is null, it is safe to delete it because it will be converted to the string 'null' by `DOMString`.

### DIFF
--- a/lib/fetch/request.js
+++ b/lib/fetch/request.js
@@ -300,7 +300,7 @@ class Request {
     }
 
     // 23. If init["integrity"] exists, then set request’s integrity metadata to it.
-    if (init.integrity !== undefined) {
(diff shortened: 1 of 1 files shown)


KhafraDev (commented):
this should be the other way around. Remove the !== undefined and keep the != null

KhafraDev (approved):

mcollina (approved):
lgtm

metcoder95 (approved):

ronag:
@tsctx I don't intend to discourage you. Some of the contributions you have been making have been absolutely great and I hope you continue doing so.

However, these small changes with limited value add extra work for us. I would appreciate it if we could reduce the amount of such minor refactorings and optimizations PRs.

PR #2463, opened 2023-11-27
chore: less async await

Uzlopak:
Why do we use so much async await?

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -1808,10 +1808,10 @@ async function httpNetworkFetch (
         fetchParams.controller.controller = controller
       },
       async pull (controller) {
-        await pullAlgorithm(controller)
+        pullAlgorithm(controller)
       },
       async cancel (reason) {
-        await cancelAlgorithm(reason)
+        cancelAlgorithm(reason)
       }
     },
     {
--- a/lib/proxy-agent.js
+++ b/lib/proxy-agent.js
@@ -82,40 +82,38 @@ class ProxyAgent extends DispatcherBase {
     this[kClient] = clientFactory(resolvedUrl, { connect })
     this[kAgent] = new Agent({
       ...opts,
-      connect: async (opts, callback) => {
+      connect: (opts, callback) => {
         let requestedHost = opts.host
         if (!opts.port) {
           requestedHost += `:${defaultProtocolPort(opts.protocol)}`
         }
-        try {
-          const { socket, statusCode } = await this[kClient].connect({
-            origin,
-            port,
-            path: requestedHost,
-            signal: opts.signal,
-            headers: {
-              ...this[kProxyHeaders],
-              host
+        this[kClient].connect({
+          origin,
+          port,
+          path: requestedHost,
+          signal: opts.signal,
+          headers: {
+            ...this[kProxyHeaders],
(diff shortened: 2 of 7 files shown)


KhafraDev (approved):

mcollina (approved):
lgtm

lib/proxy-agent.js
```diff
@@ -136,14 +134,12 @@ class ProxyAgent extends DispatcherBase {
     )
   }
 
-  async [kClose] () {
-    await this[kAgent].close()
-    await this[kClient].close()
+  [kClose] () {
+    return Promise.all([this[kAgent].close, this[kClient].close])
```
webcarrot, line 138: @Uzlopak @mcollina @KhafraDev ?:
```js
return Promise.all([this[kAgent].close(), this[kClient].close()])
```

And 👇🏻

lib/fetch/index.js
```diff
@@ -1808,10 +1808,10 @@ async function httpNetworkFetch (
         fetchParams.controller.controller = controller
       },
       async pull (controller) {
-        await pullAlgorithm(controller)
+        pullAlgorithm(controller)
```
ronag, line 1811: !!! This is wrong? needs a return at least?

lib/proxy-agent.js
```diff
@@ -82,40 +82,38 @@ class ProxyAgent extends DispatcherBase {
     this[kClient] = clientFactory(resolvedUrl, { connect })
     this[kAgent] = new Agent({
       ...opts,
-      connect: async (opts, callback) => {
+      connect: (opts, callback) => {
```
ronag, line 85: this was unnecessary and made it less maintainable...

ronag:
@KhafraDev also ☝️

Uzlopak:
First of all, I expected a little bit more discussion before merging. 
Secondly, I expect that for a reason that we have alot of Promises for a fetch call, is the unecessary wrapping of Promises in async functions.And awaiting Promises in internally used async functions. 

As the tests pass, I was like: "Ok, seems to be right" and was hoping to get some thoughts, and maybe revert some parts but still get some changes through.

lib/fetch/index.js
Uzlopak: It was before not returning anything. So it is fire and forget? So why await anyway?

lib/proxy-agent.js
Uzlopak: This one was also questionable for me. But  it should save a Promise?

mcollina:
Done, sorry

mcollina:
Can you reopen @Uzlopak ?

giacomorebonato:
I thought that the point of `return await` was to have cleaner stacktraces about where the error resides.
Also seeing a function marked as `async` is a clear statement about something, while omitting `async` removes this clarity.

lib/fetch/index.js
kibertoad: wasn't it at least informing the caller that the operation is done? I'm not sure where this is used, this can't possible cause racing conditions or break someone's tests?

PR #2464, opened 2023-11-27
chore: less async await

Uzlopak:
Reopening #2463 for a better discussion ;). 

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/fetch/body.js
+++ b/lib/fetch/body.js
@@ -235,7 +235,7 @@ function extractBody (object, keepalive = false) {
         return controller.desiredSize > 0
       },
       async cancel (reason) {
-        await iterator.return()
+        return iterator.return()
       },
       type: undefined
     })
--- a/lib/fetch/body.js
+++ b/lib/fetch/body.js
@@ -234,7 +234,7 @@ function extractBody (object, keepalive = false) {
         }
         return controller.desiredSize > 0
       },
-      async cancel (reason) {
+      cancel (reason) {
         return iterator.return()
       },
       type: undefined
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -1808,10 +1808,10 @@ async function httpNetworkFetch (
         fetchParams.controller.controller = controller
       },
       async pull (controller) {
-        await pullAlgorithm(controller)
+        pullAlgorithm(controller)
       },
       async cancel (reason) {
-        await cancelAlgorithm(reason)
+        cancelAlgorithm(reason)
       }
     },
     {
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -1807,11 +1807,11 @@ async function httpNetworkFetch (
       async start (controller) {
         fetchParams.controller.controller = controller
       },
-      async pull (controller) {
-        pullAlgorithm(controller)
+      pull (controller) {
+        return pullAlgorithm(controller)
       },
-      async cancel (reason) {
-        cancelAlgorithm(reason)
+      cancel (reason) {
+        return cancelAlgorithm(reason)
       }
     },
     {
--- a/lib/proxy-agent.js
+++ b/lib/proxy-agent.js
@@ -82,40 +82,38 @@ class ProxyAgent extends DispatcherBase {
     this[kClient] = clientFactory(resolvedUrl, { connect })
     this[kAgent] = new Agent({
       ...opts,
-      connect: async (opts, callback) => {
+      connect: (opts, callback) => {
         let requestedHost = opts.host
         if (!opts.port) {
           requestedHost += `:${defaultProtocolPort(opts.protocol)}`
         }
-        try {
-          const { socket, statusCode } = await this[kClient].connect({
-            origin,
-            port,
-            path: requestedHost,
-            signal: opts.signal,
-            headers: {
-              ...this[kProxyHeaders],
-              host
+        this[kClient].connect({
+          origin,
+          port,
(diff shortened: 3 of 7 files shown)


mcollina (approved):
lgtm

lib/fetch/index.js
```diff
@@ -1808,10 +1808,10 @@ async function httpNetworkFetch (
         fetchParams.controller.controller = controller
       },
       async pull (controller) {
-        await pullAlgorithm(controller)
+        pullAlgorithm(controller)
       },
       async cancel (reason) {
-        await cancelAlgorithm(reason)
+        cancelAlgorithm(reason)
```
ronag, line 1814:
```suggestion
        return cancelAlgorithm(reason)
```

lib/proxy-agent.js
```diff
@@ -82,40 +82,38 @@ class ProxyAgent extends DispatcherBase {
     this[kClient] = clientFactory(resolvedUrl, { connect })
     this[kAgent] = new Agent({
       ...opts,
-      connect: async (opts, callback) => {
+      connect: (opts, callback) => {
```
ronag, line 85: I think this is unnecessary change

lib/fetch/index.js
```diff
@@ -1808,10 +1808,10 @@ async function httpNetworkFetch (
         fetchParams.controller.controller = controller
       },
       async pull (controller) {
-        await pullAlgorithm(controller)
+        pullAlgorithm(controller)
```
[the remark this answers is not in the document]
Uzlopak: So it was a bug in the first place?

lib/fetch/index.js
Uzlopak: like above, a bug?

lib/fetch/index.js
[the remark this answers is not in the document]
ronag: no, you need to await it?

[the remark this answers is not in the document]
ronag: you can either return or await it for it to be resolved.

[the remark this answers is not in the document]
Uzlopak: can we remove the async then?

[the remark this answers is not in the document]
ronag: yes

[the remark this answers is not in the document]
Uzlopak: done

lib/fetch/index.js
Uzlopak: done

lib/proxy-agent.js
Uzlopak: tbh. This is for discussing. I agree this is reducing the maintainability. I am unsure if we save one Promise by this.

lib/proxy-agent.js
```diff
@@ -136,14 +134,12 @@ class ProxyAgent extends DispatcherBase {
     )
   }
 
-  async [kClose] () {
-    await this[kAgent].close()
-    await this[kClient].close()
+  [kClose] () {
+    return Promise.all([this[kAgent].close, this[kClient].close])
```
tonivj5, line 138:
```suggestion
    return Promise.all([this[kAgent].close(), this[kClient].close()])
```
are they functions, they aren't?

lib/proxy-agent.js
```diff
@@ -136,14 +134,12 @@ class ProxyAgent extends DispatcherBase {
...
   }
 
-  async [kDestroy] () {
-    await this[kAgent].destroy()
-    await this[kClient].destroy()
+  [kDestroy] () {
+    return Promise.all([this[kAgent].destroy, this[kClient].destroy])
```
tonivj5, line 142:
```suggestion
    return Promise.all([this[kAgent].destroy(), this[kClient].destroy()])
```
are they functions, they aren't?

ronag (commented):
Except for the connect case I don't mind the changes per se. I just don't think this will make any difference whatsoever in real world application. Maybe **the consume change** might have some value in terms of performance.

lib/proxy-agent.js
Uzlopak: Uff right.

lib/fetch/body.js
```diff
@@ -234,8 +234,8 @@ function extractBody (object, keepalive = false) {
         }
         return controller.desiredSize > 0
       },
-      async cancel (reason) {
-        await iterator.return()
+      cancel (reason) {
+        return iterator.return()
```
metcoder95, line 238: This might hide it from the stacktrace, not sure the attempt is to keep that function hidden as is part of the `RedeableStream` API and can be helpful while debugging.

lib/fetch/index.js
```diff
@@ -1807,11 +1807,11 @@ async function httpNetworkFetch (
       async start (controller) {
         fetchParams.controller.controller = controller
       },
-      async pull (controller) {
-        await pullAlgorithm(controller)
+      pull (controller) {
```
metcoder95, line 1810: cc: @KhafraDev

KhafraDev: the streams spec will wrap this in a promise, so it does nothing

KhafraDev (changes requested):
I've changed my mind, I just had to track down an error that was caused by similar changes. For example, if something synchronously throws an error, it can now get swallowed - this is what happened in #2482. It's super annoying to debug and makes more work for us.

PR #2468, opened 2023-11-27
perf: avoid Headers initialization

tsctx:
## Rationale

Improves performance by avoiding the initialization process of `Headers`.

## Benchmark

```js
// bench.js
const { Request, Response } = require(".");

const input = "https://example.com/post";

suite.add("new Request()", () => {
  new Request(input, undefined);
});

suite.add("new Response()", () => {
  new Response(null, undefined);
});

});

// run async
  suite.run({ async: true })
);

```

- *main*
```
new Request() x 100,116 ops/sec ±4.53% (74 runs sampled)
new Response() x 1,922,146 ops/sec ±2.12% (83 runs sampled)
```

- *this patch*
```
new Request() x 106,769 ops/sec ±3.77% (80 runs sampled)
new Response() x 2,330,854 ops/sec ±1.48% (89 runs sampled)
```

### DIFF
--- a/lib/cache/symbols.js
+++ b/lib/cache/symbols.js
@@ -1,5 +1,5 @@
 'use strict'
 
 module.exports = {
-  kConstruct: Symbol('constructable')
+  kConstruct: require('../core/symbols').kConstruct
 }
--- a/lib/fetch/headers.js
+++ b/lib/fetch/headers.js
@@ -2,7 +2,7 @@
 
 'use strict'
 
-const { kHeadersList } = require('../core/symbols')
+const { kHeadersList, kConstruct } = require('../core/symbols')
 const { kGuard } = require('./symbols')
 const { kEnumerableProperty } = require('../core/util')
 const {
@@ -240,6 +240,7 @@ class HeadersList {
 // https://fetch.spec.whatwg.org/#headers-class
 class Headers {
   constructor (init = undefined) {
+    if (init === kConstruct) return
     this[kHeadersList] = new HeadersList()
 
     // The new Headers(init) constructor steps are:
--- a/lib/fetch/request.js
+++ b/lib/fetch/request.js
@@ -28,7 +28,7 @@ const { kHeaders, kSignal, kState, kGuard, kRealm } = require('./symbols')
 const { webidl } = require('./webidl')
 const { getGlobalOrigin } = require('./global')
 const { URLSerializer } = require('./dataURL')
-const { kHeadersList } = require('../core/symbols')
+const { kHeadersList, kConstruct } = require('../core/symbols')
 const assert = require('assert')
 const { getMaxListeners, setMaxListeners, getEventListeners, defaultMaxListeners } = require('events')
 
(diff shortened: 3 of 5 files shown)


tsctx:
Hi @KhafraDev @ronag @mcollina,
What do you think about this?

lib/cache/symbols.js
```diff
@@ -1,5 +1,5 @@
 'use strict'
 
 module.exports = {
-  kConstruct: Symbol('constructable')
+  kConstruct: require('../core/symbols').kConstruct
```
tsctx, line 4: There were similar symbols, so I decided to make them common.

ronag (commented):
Have you benchmarked it? I would have guessed V8 already optimizes away things like this.

lib/fetch/headers.js
```diff
@@ -240,6 +240,7 @@ class HeadersList {
 // https://fetch.spec.whatwg.org/#headers-class
 class Headers {
   constructor (init = undefined) {
+    if (init === kConstruct) return
```
ronag, line 243:
```suggestion
    if (init === kConstruct) {
      return
    }
```

ronag (approved):
LGTM

lib/fetch/request.js
```diff
@@ -728,7 +728,7 @@ class Request {
     const clonedRequestObject = new Request(kInit)
     clonedRequestObject[kState] = clonedRequest
     clonedRequestObject[kRealm] = this[kRealm]
-    clonedRequestObject[kHeaders] = new Headers()
+    clonedRequestObject[kHeaders] = new Headers(kConstruct)
```
ronag, line 731:
```suggestion
    clonedRequestObject[kHeaders] = Object.create(Headers.prototype)
```

Wouldn't this achieve the same?

lib/fetch/headers.js
tsctx: changed with 9e1a7cd52b8ba9485a3289ce147587d10cf5bad0

lib/fetch/request.js
tsctx: It is faster to use the new operator.

tsctx:
```js

// bench.js
const kSkip = Symbol("skip");

class FakeClass {
  constructor(skip) {
    if (skip !== kSkip) throw new TypeError("Illegal constructor");
  }
}

suite.add("new operator", () => {
  new FakeClass(kSkip);
});

suite.add("Object.create", () => {
  Object.create(FakeClass.prototype);
});

});

// run async
  suite.run({ async: true })
);
```

```
new operator x 720,079,182 ops/sec ±1.36% (83 runs sampled)
Object.create x 710,966,585 ops/sec ±1.43% (85 runs sampled)
```

ronag: The difference is within the margin of error. I would prefer Object.create since it's less "hacky".

KhafraDev (approved):

PR #2475, opened 2023-11-28
fix: correctly handle data URL with hashes.

### LINKED ISSUE #2474: data URL that contain hashes are not handled properly.
## Bug Description

data URL that contain hashes are not handled properly.

## Reproducible By

```shell
> $ node
Welcome to Node.js v21.1.0.
> await (await fetch("data:,node#js#")).text();
'node#js'
```

## Expected Behavior

```shell
> $ node
Welcome to Node.js v21.1.0.
> await (await fetch("data:,node#js#")).text();
'node'
```

tsctx:
Fixes #2474

### DIFF
--- a/lib/fetch/dataURL.js
+++ b/lib/fetch/dataURL.js
@@ -124,8 +124,9 @@ function URLSerializer (url, excludeFragment = false) {
   }
 
   const href = url.href
+  const hashLength = url.hash.length
 
-  return href.substring(href, href.length - url.hash.length)
+  return hashLength === 0 ? href : href.substring(href, href.length - hashLength)
 }
 
 // https://infra.spec.whatwg.org/#collect-a-sequence-of-code-points
--- a/lib/fetch/dataURL.js
+++ b/lib/fetch/dataURL.js
@@ -125,8 +125,8 @@ function URLSerializer (url, excludeFragment = false) {
     return href
   }
 
-  // Note: Since there are at least two characters in the url, 
-  // the scheme and the U+003A (:) character, 
+  // Note: Since there are at least two characters in the url,
+  // the scheme and the U+003A (:) character,
   // the hash searches after these.
   const hash = href.indexOf('#', 2)
 
--- a/lib/fetch/dataURL.js
+++ b/lib/fetch/dataURL.js
@@ -125,11 +125,12 @@ function URLSerializer (url, excludeFragment = false) {
     return href
   }
 
-  const hash = href.lastIndexOf('#')
-  if (hash === -1) {
-    return href
-  }
-  return href.slice(0, hash)
+  // Note: Since there are at least two characters in the url, 
+  // the scheme and the U+003A (:) character, 
+  // the hash searches after these.
+  const hash = href.indexOf('#', 2)
+
+  return hash === -1 ? href : href.substring(0, hash)
 }
 
 // https://infra.spec.whatwg.org/#collect-a-sequence-of-code-points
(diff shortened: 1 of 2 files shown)


ronag (approved):

test/fetch/data-uri.js
```diff
@@ -191,3 +191,17 @@ test('https://domain.com/?', (t) => {
   const serialized = URLSerializer(new URL(domain))
   t.equal(serialized, domain)
 })
+
+// https://github.com/nodejs/undici/issues/2474
+test('data url that includes the hash', async (t) => {
```
ronag, line 196: Does this test fail before the fix?

lib/fetch/dataURL.js
```diff
@@ -125,11 +125,12 @@ function URLSerializer (url, excludeFragment = false) {
...
-  if (hash === -1) {
-    return href
-  }
-  return href.slice(0, hash)
+  // Note: Since there are at least two characters in the url,
+  // the scheme and the U+003A (:) character,
+  // the hash searches after these.
+  const hash = href.indexOf('#', 2)
+
+  return hash === -1 ? href : href.substring(0, hash)
```
ronag, line 133: Isn't something like this more reliable?

```
if (!url.hash) {
  return href
}

return href.substring(0, -href.hash.length)
```

test/fetch/data-uri.js
tsctx: Of course.
```shell
> $ npx tap test/fetch/data-uri.js
(node:4844) [DEP0040] DeprecationWarning: The `punycode` module is deprecated. Please use a userland alternative instead.
(Use `node --trace-deprecation ...` to show where the warning was created)
​ FAIL ​ test/fetch/data-uri.js
 ✖ should be equal

  test/fetch/data-uri.js
                          'data:,node#js#'
  199 |   const serialized = URLSerializer(new URL(dataURL), true)
                         node', serialized)
      | ----^

                        await fetch(dataURL)
                       it res.text(), 'node')

  --- expected    
  +++ actual      
  @@ -1,1 +1,1 @@ 
  -data:,node#js  
  +data:,node     

  test: test/fetch/data-uri.js data url that includes the hash
  stack: |
    Test.<anonymous> (test/fetch/data-uri.js:200:5)

​ FAIL ​ test/fetch/data-uri.js
 ✖ should be equal

  test/fetch/data-uri.js

      |     const res = await fetch(dataURL)
      |     t.equal(await res.text(), 'node')
      | ------^
                       r) {
  205 |     t.fail(`failed to fetch ${dataURL}`)
                                       

  --- expected    
  +++ actual      
  @@ -1,1 +1,1 @@ 
  -node
  +node#js        

  test: test/fetch/data-uri.js data url that includes the hash
  stack: |
    Test.<anonymous> (test/fetch/data-uri.js:203:7)

​ FAIL ​ test/fetch/data-uri.js 2 failed of 33 807.704ms
 ✖ should be equal
 ✖ should be equal




  🌈 SUMMARY RESULTS 🌈  
                         
​ FAIL ​ test/fetch/data-uri.js 2 failed of 33 807.704ms
 ✖ should be equal
 ✖ should be equal


Suites:   ​1 failed​, ​1 of 1 completed​
Asserts:  ​​​2 failed​, ​31 passed​, ​of 33​
​Time:​   ​5s
```

lib/fetch/dataURL.js
tsctx: I used `String#indexOf` because I felt that you were using `String#lastIndexOf` intentionally.
I was looking at the commit log and it doesn't seem intentional, so I will do so.

ronag (approved):

PR #2477, opened 2023-11-29
fix: check response for timinginfo allow flag

### LINKED ISSUE #2476: Logic error in performance timing filtering
## Bug Description

When performing a request which should allow performance timing, only opaque timing info is returned.

## Reproducible By

1. Perform a fetch request
2. Check performance timing info for that request
3. See that we only get opaque timing info, with transferSize and encodedBodySize 0.

## Expected Behavior

I would expect to get timing info for the request

## Environment

Node v20.9.0

### Additional context

[This comment](https://github.com/nodejs/undici/blob/main/lib/fetch/index.js#L288) describes that the timingInfo should be replaced with an opaqueTimingInfo if the "**response**'s timing allow passed flag is not set", but the code checks `timingInfo.timingAllowPassed`, which doesn't seem to be set anywhere. 
However, the response's timingAllowPassed flag is set [here](https://github.com/nodejs/undici/blob/main/lib/fetch/index.js#L702).

Could it be that the code should as follows? This seems to fix my problem.
```
  if (!response.timingAllowPassed) {
    //  1. Set timingInfo to a the result of creating an opaque timing info for timingInfo.
    timingInfo = createOpaqueTimingInfo({
      startTime: timingInfo.startTime
    })
    ...
  }
```

ToshB:
## This relates to...

Fixes #2476 

## Rationale

Explained in #2476

## Changes

Fix property, add test verifying fixed behavior

### Features

N/A

### Bug Fixes

Performance entry now includes decodedBodySize, encodedBodySize and transferSize values

## Status

- [x] Tested
- [S] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/test/fetch/resource-timing.js
+++ b/test/fetch/resource-timing.js
@@ -46,3 +46,27 @@ test('should create a PerformanceResourceTiming after each fetch request', { ski
 
   t.teardown(server.close.bind(server))
 })
+
+test('should include encodedBodySize in performance entry', { skip }, (t) => {
+  t.plan(4)
+  const obs = new PerformanceObserver(list => {
+    const [entry] = list.getEntries()
+    t.equal(entry.encodedBodySize, 2)
+    t.equal(entry.decodedBodySize, 2)
+    t.equal(entry.transferSize, 2+300)
+
+    obs.disconnect()
+    performance.clearResourceTimings()
+  })
+
+  obs.observe({ entryTypes: ['resource'] })
+
+  const server = createServer((req, res) => {
+    res.end('ok')
+  }).listen(0, async () => {
+    const body = await fetch(`http://localhost:${server.address().port}`)
+    t.strictSame('ok', await body.text())
+  })
+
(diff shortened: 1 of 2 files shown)


ronag (approved):

test/fetch/resource-timing.js
```diff
@@ -46,3 +46,27 @@ test('should create a PerformanceResourceTiming after each fetch request', { ski
...
   t.teardown(server.close.bind(server))
 })
+
+test('should include encodedBodySize in performance entry', { skip }, (t) => {
+  t.plan(4)
+  const obs = new PerformanceObserver(list => {
+    const [entry] = list.getEntries()
+    t.equal(entry.encodedBodySize, 2)
+    t.equal(entry.decodedBodySize, 2)
+    t.equal(entry.transferSize, 2+300)
```
KhafraDev, line 56:
```suggestion
    t.equal(entry.transferSize, 2 + 300)
```

PR #2479, opened 2023-11-30
perf(fetch): Improve fetch of detaurl

tsctx:
## Rationale

1.5~6.0 times faster than current implementation.

## Benchmark

```js
// bench.js
const { fetch } = require(".");

const small = "data:," + "x".repeat(50);
const middle = "data:," + "x".repeat(3000);
const long = "data:," + "x".repeat(45000);
const longest = "data:," + "x".repeat(70000);

const small_base64 =
  "data:;base64," + Buffer.from("x".repeat(50)).toString("base64");
const middle_base64 =
  "data:;base64," + Buffer.from("x".repeat(3000)).toString("base64");
const long_base64 =
  "data:;base64," + Buffer.from("x".repeat(45000)).toString("base64");
const longest_base64 =
  "data:;base64," + Buffer.from("x".repeat(70000)).toString("base64");

suite.add("small", {
  fn: async (deferred) => {
    await (await fetch(small)).arrayBuffer();
    deferred.resolve();
  },
  defer: true,
});

suite.add("middle", {
  fn: async (deferred) => {
    await (await fetch(middle)).arrayBuffer();
    deferred.resolve();
  },
  defer: true,
});

suite.add("long", {
  fn: async (deferred) => {
    await (await fetch(long)).arrayBuffer();
    deferred.resolve();
  },
  defer: true,
});

suite.add("longest", {
  fn: async (deferred) => {
    await (await fetch(longest)).arrayBuffer();
    deferred.resolve();
  },
  defer: true,
});

suite.add("base64 - small", {
  fn: async (deferred) => {
    await (await fetch(small_base64)).arrayBuffer();
    deferred.resolve();
  },
  defer: true,
});

suite.add("base64 - middle", {
  fn: async (deferred) => {
    await (await fetch(middle_base64)).arrayBuffer();
    deferred.resolve();
  },
  defer: true,
});

suite.add("base64 - long", {
  fn: async (deferred) => {
    await (await fetch(long_base64)).arrayBuffer();
    deferred.resolve();
  },
  defer: true,
});

suite.add("base64 - longest", {
  fn: async (deferred) => {
    await (await fetch(longest_base64)).arrayBuffer();
    deferred.resolve();
  },
  defer: true,
});

});

// run async
  suite.run({ async: true })
);
```

- *main*
```
small x 12,535 ops/sec ±11.00% (71 runs sampled)
middle x 9,651 ops/sec ±1.82% (76 runs sampled)
long x 844 ops/sec ±2.25% (74 runs sampled)
longest x 539 ops/sec ±2.08% (70 runs sampled)
base64 - small x 12,414 ops/sec ±4.33% (72 runs sampled)
base64 - middle x 1,911 ops/sec ±0.74% (80 runs sampled)
base64 - long x 114 ops/sec ±1.63% (70 runs sampled)
base64 - longest x 92.35 ops/sec ±2.93% (65 runs sampled)
```

- *this patch*
```
small x 12,314 ops/sec ±12.63% (66 runs sampled)
middle x 11,968 ops/sec ±2.59% (75 runs sampled)
long x 2,110 ops/sec ±2.72% (68 runs sampled)
longest x 1,593 ops/sec ±4.15% (66 runs sampled)
base64 - small x 14,139 ops/sec ±3.88% (73 runs sampled)
base64 - middle x 7,324 ops/sec ±1.84% (75 runs sampled)
base64 - long x 740 ops/sec ±1.60% (80 runs sampled)
base64 - longest x 462 ops/sec ±2.28% (76 runs sampled)
```


lib/fetch/dataURL.js
```diff
@@ -429,14 +434,8 @@ function forgivingBase64 (data) {
...
-  const binary = atob(data)
-  const bytes = new Uint8Array(binary.length)
-
-  for (let byte = 0; byte < binary.length; byte++) {
-    bytes[byte] = binary.charCodeAt(byte)
-  }
-
-  return bytes
+  const buffer = Buffer.from(data, 'base64')
+  return new Uint8Array(buffer.buffer, buffer.byteOffset, buffer.byteLength)
```
tsctx, line 438: I don't see a problem with the change here. Because all we are doing is base64 decoding.
Changing this section has greatly improved performance.

tsctx:
The test for Node 16.8 fails because of the formdata-node version upgrade.

KhafraDev (commented):
can you remove all the substring changes?

KhafraDev:
> Would be nice with more test cases or even a fuzzer to ensure this doesn't introduce bugs.

@ronag the WPTs cover data urls extensively. However they only ran on 1 platform because every other CI flaked before reaching them

lib/fetch/dataURL.js
KhafraDev: These are *very* unlikely to be equivalent.

tsctx: The only difference between buffer and atob is that atob only supports latin1.
So I have to do this hackly thing to get back to utf8.

lib/fetch/dataURL.js
```diff
@@ -425,18 +437,12 @@ function forgivingBase64 (data) {
   //  U+002F (/)
   //  ASCII alphanumeric
   // then return failure.
-  if (/[^+/0-9A-Za-z]/.test(data)) {
+  if (/[^+/0-9A-Za-z]/.test(data.length === dataLength ? data : data.substring(0, dataLength))) {
```
tsctx, line 440: You can do this, but what do you do?
```js
  if (/^[+/0-9A-Za-z]*=*$/.test(data)) {
```

lib/fetch/util.js
```diff
@@ -914,12 +912,37 @@ function isomorphicDecode (input) {
...
+  }
+  // FIXME: Is it really possible that an array of numbers is passed on?
+  while (i + 32 < length) {
+    result += String.fromCharCode(
+      input[i], input[i + 1], input[i + 2], input[i + 3], input[i + 4], input[i + 5], input[i + 6], input[i + 7],
+      input[i + 8], input[i + 9], input[i + 10], input[i + 11], input[i + 12], input[i + 13], input[i + 14], input[i + 15],
+      input[i + 16], input[i + 17], input[i + 18], input[i + 19], input[i + 20], input[i + 21], input[i + 22], input[i + 23],
+      input[i + 24], input[i + 25], input[i + 26], input[i + 27], input[i + 28], input[i + 29], input[i + 30], input[i + 31]
+    )
+    i += 32
```
metcoder95, line 939: It does the work but it feels like a codesmell. It also seems harder to maintain without a previous context of why we are doing that.
Is there any way to simplify this?

tsctx: I see two options: delete this part or delete the non-TypedArray part. Currently only the TypedArray seems to be used.

tsctx: It appears that only TypedArray is currently being used.
So this part can be removed.
Let me know if you have any other opinions.

tsctx: One idea is to comment out the code and raise an error if it is not a TypedArray.

tsctx: It would be easier to understand in the future if we did so.
```js
/**
 * @see https://infra.spec.whatwg.org/#isomorphic-decode
 * @param {Uint8Array} input
 */
function isomorphicDecode (input) {
  // 1. To isomorphic decode a byte sequence input, return a string whose code point
  //    length is equal to input’s length and whose code points have the same values
  //    as the values of input’s bytes, in the same order.
  const length = input.length
  if (!('buffer' in input)) {
    throw new TypeError('Nothing but TypedArray are supported. Please uncomment to make it work.')
  }
  if ((2 << 15) - 1 > length) {
    return String.fromCharCode.apply(null, input)
  }
  let result = ''; let i = 0
  // if ('buffer' in input) {
  let addition = (2 << 15) - 1
  while (i < length) {
    if (i + addition > length) {
      addition = length - i
    }
    result += String.fromCharCode.apply(null, input.subarray(i, i += addition))
  }
  return result
  // }
  // while (i + 32 < length) {
  //   result += String.fromCharCode(
  //     input[i], input[i + 1], input[i + 2], input[i + 3], input[i + 4], input[i + 5], input[i + 6], input[i + 7],
  //     input[i + 8], input[i + 9], input[i + 10], input[i + 11], input[i + 12], input[i + 13], input[i + 14], input[i + 15],
  //     input[i + 16], input[i + 17], input[i + 18], input[i + 19], input[i + 20], input[i + 21], input[i + 22], input[i + 23],
  //     input[i + 24], input[i + 25], input[i + 26], input[i + 27], input[i + 28], input[i + 29], input[i + 30], input[i + 31]
  //   )
  //   i += 32
  // }
  // // Decode the remaining characters.
  // while (i < length) {
  //   result += String.fromCharCode(input[i++])
  // }
  // return result
}
```

tsctx: @metcoder95 Doing this is easily accomplished.
```js
if (!('buffer' in input)) throw new TypeError("Supported only for TypedArray.")
```

tsctx: What I am currently thinking:
```js
function isomorphicDecode (input) {
  // 1. To isomorphic decode a byte sequence input, return a string whose code point
  //    length is equal to input’s length and whose code points have the same values
  //    as the values of input’s bytes, in the same order.
  const length = input.length
  if ((2 << 15) - 1 > length) {
    return String.fromCharCode.apply(null, input)
  }
  let result = ''; let i = 0
  // isTypedArray
  if ('buffer' in input) {
    let addition = (2 << 15) - 1
    while (i < length) {
      if (i + addition > length) {
        addition = length - i
      }
      result += String.fromCharCode.apply(null, input.subarray(i, i += addition))
    }
    return result
  }
  // while (i + 32 < length) {
  //   result += String.fromCharCode(
  //     input[i], input[i + 1], input[i + 2], input[i + 3], input[i + 4], input[i + 5], input[i + 6], input[i + 7],
  //     input[i + 8], input[i + 9], input[i + 10], input[i + 11], input[i + 12], input[i + 13], input[i + 14], input[i + 15],
  //     input[i + 16], input[i + 17], input[i + 18], input[i + 19], input[i + 20], input[i + 21], input[i + 22], input[i + 23],
  //     input[i + 24], input[i + 25], input[i + 26], input[i + 27], input[i + 28], input[i + 29], input[i + 30], input[i + 31]
  //   )
  //   i += 32
  // }
  // // Decode the remaining characters.
  // while (i < length) {
  //   result += String.fromCharCode(input[i++])
  // }
  // return result
  throw new TypeError("Supported only for TypedArray. If you want it to work, uncomment it.")
}
```

tsctx: > that's not code we want even if something other than a typed array was passed

Okay, I'll delete this part.

tsctx:
Would you like to see it again?
cc @ronag @KhafraDev @metcoder95

PR #2481, opened 2023-11-30
feat: Add resource timing entries for connection, request and response

ToshB:
## Rationale

Currently the timestamps included in the resource timing data are startTime, duration, fetchStart and responseEnd. By adding timestamps for requestStart and responseStart the resource timing values can be used to more accurately report on connection speed. 

## Changes

This PR adds the missing values by setting timestamps in existing callbacks, with one addition of an onMessageBegin callback for getting the requestStart timestamp. It also adds an initial version of connection timing info, with default values.

### Features

Added requestStart and responseStart resource timings.

### Bug Fixes

N/A

## Status

- [x] Tested
- [S] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready


ToshB:
> depends on #2480 because of the optional chaining

Should I change the few places using optional chaining, or do you want to finish #2480 first? 
Assuming this PR is accepted.

KhafraDev:
Let's wait, using optional chaining is fine

lib/client.js
```diff
@@ -740,6 +740,7 @@ class Parser {
     if (!request) {
       return -1
     }
+    request.onMessageBegin()
```
metcoder95, line 743: If we are following the path of extending the request, I believe we should sync this with the `Dispatcher` API and document it accordingly, as it opens a new API.

ToshB: Hm, good point. According to [spec](https://fetch.spec.whatwg.org/#http-network-fetch) we want to "Set timingInfo’s [final network-response start time](https://fetch.spec.whatwg.org/#fetch-timing-info-final-network-response-start-time) to the [coarsened shared current time](https://w3c.github.io/hr-time/#dfn-coarsened-shared-current-time) given fetchParams’s [cross-origin isolated capability](https://fetch.spec.whatwg.org/#fetch-params-cross-origin-isolated-capability), _immediately after the user agent’s HTTP parser receives the first byte of the response_ (e.g., frame header bytes for HTTP/2 or response status line for HTTP/1.x).". 
So I think this is the closest we get with the current llhttp API. Unless there are some alternative hooks where we can add this timing info, without opening a new API?

ronag: Isn't onHeaders sufficient in a practical sense?

ToshB: It would be close, I guess it depends on how closely we want to adhere to the spec.

I do think it's unfortunate that responseStart (which is time to first byte) is dependent on header size. 
Having accurate responseStart and responseEnd values would also be good for benchmarking message parsing.

ronag: onResponseStarted?

ToshB: I was also considering if it would be better to call this from outside of the llhttp hooks, as it doesn't really depend on the payload parsing. But I couldn't find an obvious place to hook in, that would also work with the pause/receive functionality. If I understand correctly llhttp keeps track of all that.

ronag: I think where you have it is good enough for any practical case.

lib/client.js
```diff
@@ -740,6 +740,7 @@ class Parser {
     if (!request) {
       return -1
     }
+    request.onResponseStarted()
```
metcoder95, line 743: Can you add it to `http2` as well, there it should be within the following line: https://github.com/nodejs/undici/blob/107664ab220dc389c6dbb9c0f85a20cf0ba1ed3d/lib/client.js#L1784

As the `HEADERS` frame states the transition of the stream from `idle` to open, I believe it should be good to put it there.

lib/fetch/index.js
```diff
@@ -2098,12 +2099,21 @@ async function httpNetworkFetch (
           // TODO (fix): Do we need connection here?
           const { connection } = fetchParams.controller
 
+          // TODO: pass connection timing info
```
metcoder95, line 2102: Is there something missed to add this?

ToshB: According to [2.6](https://fetch.spec.whatwg.org/#connections) in the spec, each connection should have an associated timing info, with values for "domain lookup start time", "domain lookup end time", "connection start time" and "connection end time". I wasn't able to find any place where these values are tracked today, and I didn't dig deep enough into the connection process to understand where to add them.

Step 7.2 under [HTTP-network fetch](https://fetch.spec.whatwg.org/#http-network-fetch), says we should use the connection timing info together with a "defaultStartTime" equal to postRedirectStartTime which is used as the lower bound for each of these values. So without these values I understand that the postRedirectStartTime (which we have) is a sensible default until these values are tracked in more detail.

lib/client.js
ToshB: 👍 I'll add it.

lib/fetch/util.js
```diff
@@ -311,9 +311,39 @@ function appendRequestOriginHeader (request) {
...
+    connectionStartTime: coarsenTime(connectionTimingInfo.connectionStartTime, crossOriginIsolatedCapability),
+    connectionEndTime: coarsenTime(connectionTimingInfo.connectionEndTime, crossOriginIsolatedCapability),
+    secureConnectionStartTime: coarsenTime(connectionTimingInfo.secureConnectionStartTime, crossOriginIsolatedCapability),
+    ALPNNegotiatedProtocol: connectionTimingInfo.ALPNNegotiatedProtocol
+  }
+}
+
+// https://w3c.github.io/hr-time/#dfn-coarsened-shared-current-time
 function coarsenedSharedCurrentTime (crossOriginIsolatedCapability) {
   // TODO
```
KhafraDev, line 345: is this TODO needed?

KhafraDev (changes requested):
please add a test that ensures onResponseStarted is called before onHeaders, both with http1 and http2

lib/fetch/util.js
ToshB: No, probably not. I forgot to remove it after extracting the more general coarsenTime function. I'll remove it.

lib/fetch/index.js
metcoder95: Can you add those as comments, it can provide great context to understand what's left

lib/fetch/util.js
ToshB: I added tests for ensuring ordering of all dispatches

lib/fetch/index.js
ToshB: Added comments to this and the two other places where I add timing info

KhafraDev (approved):

metcoder95 (approved):

Ethan-Arrowood:
Does this resolve the issue I opened last week? https://github.com/nodejs/undici/issues/2515

Timing couldn't be better! Was going to contribute this sometime this week.

ToshB:
> Does this resolve the issue I opened last week? #2515

It adds several of the missing values, but not the connection and dns-resolve timings (except for defaulting them to valid values).

Ethan-Arrowood:
Okay looks good to me. I'll go through and add those later then.

Ethan-Arrowood (approved):
LGTM 🚀

PR #2482, opened 2023-12-01
update spec & wpts

KhafraDev:


KhafraDev:
Also getting a bunch of test failures with this message `TypeError [ERR_INVALID_STATE]: Invalid state: chunk ArrayBuffer is zero-length or detached"` (what's the best way of getting around this? none of my ideas are particularly good)

lib/fetch/dataURL.js
```diff
@@ -126,7 +126,13 @@ function URLSerializer (url, excludeFragment = false) {
   const href = url.href
   const hashLength = url.hash.length
 
-  return hashLength === 0 ? href : href.substring(0, href.length - hashLength)
+  const serialized = hashLength === 0 ? href : href.substring(0, href.length - hashLength)
```
KhafraDev, line 129: this is a recent bug where serializing `https://localhost/#` would result in `https://localhost/#` instead of `https://localhost/`

debadree25:
One of the cases to consider in bytes stream is to call ```c.byobRequest?.respond(0);``` after calling controller.close() you can take a look at https://github.com/nodejs/node/pull/49713/files maybe could be helpful?

lib/fetch/dataURL.js
debadree25: I dont think endsWith would be slow any longer https://chromium-review.googlesource.com/c/v8/v8/+/4603612 :-)

tsctx: I forgot to add. I meant slower than the way I presented it.

KhafraDev:
Thank you! That fixed most of the errors!

KhafraDev:
We have one more error here:

```js
// 1. Let transformStream be a new TransformStream.
    // 2. Let identityTransformAlgorithm be an algorithm which, given chunk, enqueues chunk in transformStream.
    // 3. Set up transformStream with transformAlgorithm set to identityTransformAlgorithm and flushAlgorithm
    //    set to processResponseEndOfBody.
    const transformStream = new TransformStream({
      start () {},
      transform (chunk, controller) {
        controller.enqueue(chunk)
      },
      flush: processResponseEndOfBody
    })

    // 4. Set internalResponse’s body’s stream to the result of internalResponse’s body’s stream piped through transformStream.
    internalResponse.body.stream.pipeThrough(transformStream)
    internalResponse.body.stream = transformStream.readable
```

because the readable side of the transform stream isn't a byte stream... is there anything we could do about this? Or do we have to implement the equivalent of the transformstream?

lib/fetch/dataURL.js
KhafraDev: I'm not really interested in optimizing against v8, and the link above mostly proves why

KhafraDev:
as many as I needed to for byob support, I'm probably missing some but this is fast forwarding us a couple months

ronag:
CI failing?

KhafraDev:
windows 20: ```
 not ok 74 - test/redirect-request.js # time=1809.245ms

```

windows 18: ```
 not ok 74 - test/redirect-request.js # time=10253.305ms
```

KhafraDev:
(not related to the WPTs, they all passed on ubuntu 18/20)

PR #2495, opened 2023-12-04
perf: twice faster method check

tsctx:
We wondered if we should submit it with the PR we are currently working on, but decided to split it up since we expect significant improvement.

## Benchmark
```js

const method = "GET"

/**
 * @see https://tools.ietf.org/html/rfc7230#section-3.2.6
 * @param {number} c
 */
function isTokenCharCode (c) {
  switch (c) {
    case 0x22:
    case 0x28:
    case 0x29:
    case 0x2c:
    case 0x2f:
    case 0x3a:
    case 0x3b:
    case 0x3c:
    case 0x3d:
    case 0x3e:
    case 0x3f:
    case 0x40:
    case 0x5b:
    case 0x5c:
    case 0x5d:
    case 0x7b:
    case 0x7d:
      // DQUOTE and "(),/:;<=>?@[\]{}"
      return false
    default:
      // VCHAR %x21-7E
      return c >= 0x21 && c <= 0x7e
  }
}

/**
 * @param {string} characters
 */
function isValidHTTPToken (characters) {
  if (characters.length === 0) {
    return false
  }
  for (let i = 0; i < characters.length; ++i) {
    if (!isTokenCharCode(characters.charCodeAt(i))) {
      return false
    }
  }
  return true
}

const tokenRegExp = /^[\^_`a-zA-Z\-0-9!#$%&'*+.|~]+$/

suite.add("RegExp#test", () => {
  tokenRegExp.test(method)
})

suite.add("RegExp#exec", () => {
  tokenRegExp.exec(method) !== null
})

suite.add("charCodeAt approach", () => {
  isValidHTTPToken(method)
})

});

// run async
  suite.run({ async: true })
);
```

```
RegExp#test x 24,317,904 ops/sec ±3.85% (76 runs sampled)
RegExp#exec x 15,239,797 ops/sec ±2.97% (76 runs sampled)
charCodeAt approach x 45,102,665 ops/sec ±4.63% (76 runs sampled)
```

### DIFF
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -80,7 +80,7 @@ class Request {
 
     if (typeof method !== 'string') {
       throw new InvalidArgumentError('method must be a string')
-    } else if (tokenRegExp.exec(method) === null) {
+    } else if (!util.isValidHTTPToken(method)) {
       throw new InvalidArgumentError('invalid request method')
     }
 
--- a/lib/core/util.js
+++ b/lib/core/util.js
@@ -449,6 +449,52 @@ function toUSVString (val) {
   return `${val}`
 }
 
+/**
+ * @see https://tools.ietf.org/html/rfc7230#section-3.2.6
+ * @param {number} c
+ */
+function isTokenCharCode (c) {
+  switch (c) {
+    case 0x22:
+    case 0x28:
+    case 0x29:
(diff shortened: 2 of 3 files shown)


mcollina (approved):
lgtm

PR #2497, opened 2023-12-05
fix: stream error timings

ronag:
This is rather bad...


ronag:
I have multiple failures in production I need to sort out so if you want me to make tests it will take a day or two.

mcollina:
I would prefer if we had a test for this

mcollina (approved):
lgtm

mcollina:
CI does not seem happy, could you fix?

PR #1700, opened 2022-10-13
fix: allow cloned stream to be unref'd

KhafraDev:

PR #2499, opened 2023-12-05
fix: terminate the process correctly

### LINKED ISSUE #2026: `fetch` can make `worker_threads` stuck and process hang

### LINKED ISSUE #2413: Tests not terminating after execution with msw when using FormData and Blob
## Bug Description

I'm encountering an issue where the process does not terminate after running tests with `msw` when using `FormData` and `Blob`. Removing the line where the `Blob` is appended to the `FormData` allows the tests to terminate as expected.

## Reproducible By

Below is a simplified test case to illustrate the issue:

```
~/work/msw-formdata-issue/ npm test

> msw-formdata-issue@1.0.0 test
> mocha



  test
{}
    ✔ test


  1 passing (22ms)
```

When the above code is executed, the process does not terminate after running the test. However, if the line with `formData.append(...)` is commented out, the process terminates correctly. It seems that the use of FormData in combination with Blob in msw setup is causing some resources to not be released properly.

## Expected Behavior

exit the process after running all tests.

## Environment

Node.js 21.1.0

### Additional context

At first, I noted this issue to msw repo, but msw author suggest to post this issue to undici.
ref. https://github.com/mswjs/msw/issues/1843

tsctx:
Fixes #2413


tsctx:
Sorry, but I could not create a unit test.

lib/fetch/body.js
```diff
@@ -277,7 +277,8 @@ function cloneBody (body) {
   const out2Clone = structuredClone(out2, { transfer: [out2] })
   // This, for whatever reasons, unrefs out2Clone which allows
   // the process to exit by itself.
-  const [, finalClone] = out2Clone.tee()
+  const [finalClone, final2Clone] = out2Clone.tee()
+  final2Clone.cancel()
```
KhafraDev, line 281: this promise isn't being handled

KhafraDev (commented):
A test that just makes a fetch call in a worker sounds useful. FWIW I couldn't repro it locally either, but maybe it is in the CIs.

Is there a reason you're using the first tee'd stream instead of the second? IIRC I had issues using the first one.

tsctx:
> A test that just makes a fetch call in a worker sounds useful. FWIW I couldn't repro it locally either, but maybe it is in the CIs.
> 
> Is there a reason you're using the first tee'd stream instead of the second? IIRC I had issues using the first one.

No particular reason. What problems did you have?

KhafraDev:
Don't recall exactly, probably related to the comment

tsctx:
Is this? https://github.com/nodejs/undici/pull/1700

KhafraDev:
that's why we needed the tee, not why I used the second stream. Just wondering, how do you know that this fixes the issue?

tsctx:
I found that I could run a reproducible script of 2 issues and fix it.

KhafraDev:
why can't that be made into a test?

tsctx:
> why can't that be made into a test?

This is because it requires access to an external server.
That test did not work when I set up a local server.

If you want the test, you need a few days. Still, you may not be able to create the test.

tsctx:
My meaning of did not work is that it worked correctly even before the modification.

> as I mentioned before, it might not work locally but it might run on the CI, which is better than nothing (even if it doesn't work on the CI either)

I can't do that right now. I will do some research

mcollina (approved):
lgtm

AriPerkkio:
@tsctx does this fix reproduction case from #2026 for you locally? I can still reproduce it after applying the changes from this PR.

tsctx:
Hi @AriPerkkio,
Changes are added. Is this fixed?

ronag:
@KhafraDev I'd like you approval on this before merging.

tsctx:
> > Hi @AriPerkkio,
> 
> > Changes are added. Is this fixed?
> 
> 
> 
> No, #2026 still gets stuck with these changes applied locally:
> 
> 
> 
> ```sh
> 
> ari ~/Git/undici (main) $ node repro.mjs 
> 
> Unable to terminate
> 
> ...
> 
> There are 1 stuck workers
> 
> Forcing process.exit on main thread
> 
> ```
> 
> 

Thanks for the report! I thought I had fixed it.

KhafraDev:
@tsctx it's most likely a bug in node, not really sure how much of it can be fixed here. If you find a fix for https://github.com/nodejs/node/issues/44985 it would probably help.

PR #2503, opened 2023-12-06
fix: more sensible stack trace from dump error

ronag:
## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready


lib/core/errors.js
```diff
@@ -82,7 +82,16 @@ class InvalidReturnValueError extends UndiciError {
   }
 }
 
-class RequestAbortedError extends UndiciError {
+class AbortError extends UndiciError {
+  constructor (message) {
+    super(message)
+    Error.captureStackTrace(this, RequestAbortedError)
```
metcoder95, line 88: Shouldn't be?
```suggestion
    Error.captureStackTrace(this, AbortError)
```

ronag:
> @ronag class, the `Error.captureStackTrace` is not necessary since the stack trace is automatically collected by the Error constructor. source: https://github.com/v8/v8/blob/8d0ffc5559c0aabe6551096d3e52386feec0aea4/src/execution/messages.cc#L636-L645
> 
> Within the Error constructor, it is set to skip frames above the function specified in `new.target`. This is the same behavior as specifying a custom class name (AbortError) for `Error.captureStackTrace`. source: https://github.com/v8/v8/blob/8d0ffc5559c0aabe6551096d3e52386feec0aea4/src/execution/messages.cc#L550-L556

PR welcome. Don't think this is relevant for this PR as it's using the existing practice.

tsctx:
> > @ronag class, the `Error.captureStackTrace` is not necessary since the stack trace is automatically collected by the Error constructor. source: https://github.com/v8/v8/blob/8d0ffc5559c0aabe6551096d3e52386feec0aea4/src/execution/messages.cc#L636-L645
> > Within the Error constructor, it is set to skip frames above the function specified in `new.target`. This is the same behavior as specifying a custom class name (AbortError) for `Error.captureStackTrace`. source: https://github.com/v8/v8/blob/8d0ffc5559c0aabe6551096d3e52386feec0aea4/src/execution/messages.cc#L550-L556
> 
> PR welcome. Don't think this is relevant for this PR as it's using the existing practice.

PR will be issued after this is merged because of conflicts.

metcoder95 (approved):

PR #2507, opened 2023-12-06
refactor: version cleanup

tsctx:


ronag (approved):

lib/core/util.js
```diff
@@ -417,10 +409,7 @@ function throwIfAborted (signal) {
     signal.throwIfAborted()
```
KhafraDev, line 409: throwIfAborted was added in node v17.3.0 according to MDN. That branch can't be reached anymore, so the entire thing can be replaced with just signal.throwIfAborted

mcollina (approved):
lgtm

PR #2510, opened 2023-12-07
perf: optimize consumeEnd

### LINKED ISSUE #2505: Optimize `consumeEnd` json/text
In `function consumeEnd (consume)`

We do:

`Buffer.concat(body).toString()`

I think there can be faster or at least more memory efficient alternatives, e.g. using a streaming `TextDecoder`

tsctx:
Closes #2505
use `Buffer.concat` length hints.

### DIFF
--- a/lib/api/readable.js
+++ b/lib/api/readable.js
@@ -299,9 +299,18 @@ function consumeEnd (consume) {
       case 'json':
         resolve(JSON.parse(utf8DecodeBytes(Buffer.concat(body, length))))
         break
-      case 'arrayBuffer':
-        resolve(Buffer.concat(body, length).buffer)
+      case 'arrayBuffer': {
+        const dst = new Uint8Array(length)
+
+        let pos = 0
+        for (const buf of body) {
+          dst.set(buf, pos)
+          pos += buf.byteLength
+        }
+
+        resolve(dst.buffer)
         break
+      }
       case 'blob':
         resolve(new (Blob ??= require('buffer').Blob)(body, { type: stream[kContentType] }))
         break
--- a/lib/api/readable.js
+++ b/lib/api/readable.js
@@ -274,6 +274,7 @@ function chunksDecode (chunks, length) {
   return buffer.toString(
     'utf8',
     buffer.length >= 3 &&
+    // Skip BOM.
       buffer[0] === 0xef &&
       buffer[1] === 0xbb &&
       buffer[2] === 0xbf
--- a/lib/api/readable.js
+++ b/lib/api/readable.js
@@ -6,7 +6,7 @@ const assert = require('assert')
 const { Readable } = require('stream')
 const { RequestAbortedError, NotSupportedError, InvalidArgumentError, AbortError } = require('../core/errors')
 const util = require('../core/util')
-const { ReadableStreamFrom, toUSVString } = require('../core/util')
+const { ReadableStreamFrom } = require('../core/util')
 
 let Blob
 
@@ -267,29 +267,44 @@ function consumeStart (consume) {
   }
 }
 
+const decoder = new TextDecoder()
+
+/**
+ * @see https://encoding.spec.whatwg.org/#utf-8-decode
+ * @param {Buffer} buffer
+ */
+function utf8DecodeBytes (buffer) {
+  if (buffer.length === 0) {
+    return ''
+  }
+  // 1.
+  // 2.
+  if (buffer[0] === 0xEF && buffer[1] === 0xBB && buffer[2] === 0xBF) {
+    buffer = buffer.subarray(3)
+  }
+  // 3.
+  const output = decoder.decode(buffer)
+  // 4.
+  return output
+}
+
 function consumeEnd (consume) {
   const { type, body, resolve, stream, length } = consume
 
   try {
-    if (type === 'text') {
-      resolve(toUSVString(Buffer.concat(body)))
-    } else if (type === 'json') {
-      resolve(JSON.parse(Buffer.concat(body)))
-    } else if (type === 'arrayBuffer') {
-      const dst = new Uint8Array(length)
-
-      let pos = 0
-      for (const buf of body) {
-        dst.set(buf, pos)
-        pos += buf.byteLength
-      }
-
-      resolve(dst.buffer)
-    } else if (type === 'blob') {
-      if (!Blob) {
-        Blob = require('buffer').Blob
-      }
-      resolve(new Blob(body, { type: stream[kContentType] }))
+    switch (type) {
+      case 'text':
+        resolve(utf8DecodeBytes(Buffer.concat(body, length)))
+        break
+      case 'json':
+        resolve(JSON.parse(utf8DecodeBytes(Buffer.concat(body, length))))
+        break
+      case 'arrayBuffer':
+        resolve(Buffer.concat(body, length).buffer)
+        break
+      case 'blob':
+        resolve(new (Blob ??= require('buffer').Blob)(body, { type: stream[kContentType] }))
+        break
     }
 
     consumeFinish(consume)
--- a/lib/api/readable.js
+++ b/lib/api/readable.js
@@ -287,28 +287,22 @@ function consumeEnd (consume) {
   const { type, body, resolve, stream, length } = consume
 
   try {
-    switch (type) {
-      case 'text':
-        resolve(chunksDecode(body, length))
-        break
-      case 'json':
-        resolve(JSON.parse(chunksDecode(body, length)))
-        break
-      case 'arrayBuffer': {
-        const dst = new Uint8Array(length)
+    if (type === 'text') {
+      resolve(chunksDecode(body, length))
+    } else if (type === 'json') {
+      resolve(chunksDecode(body, length))
+    } else if (type === 'arrayBuffer') {
+      const dst = new Uint8Array(length)
 
-        let pos = 0
-        for (const buf of body) {
-          dst.set(buf, pos)
-          pos += buf.byteLength
-        }
-
-        resolve(dst.buffer)
-        break
+      let pos = 0
+      for (const buf of body) {
+        dst.set(buf, pos)
+        pos += buf.byteLength
       }
-      case 'blob':
-        resolve(new Blob(body, { type: stream[kContentType] }))
-        break
+
+      resolve(dst.buffer)
+    } else if (type === 'blob') {
+      resolve(new Blob(body, { type: stream[kContentType] }))
     }
 
     consumeFinish(consume)
--- a/lib/api/readable.js
+++ b/lib/api/readable.js
@@ -266,13 +266,20 @@ function consumeStart (consume) {
 }
 
 /**
- * @see https://encoding.spec.whatwg.org/#utf-8-decode
- * @param {Buffer} buffer
+ * @param {Buffer[]} chunks
+ * @param {number} length
  */
-function utf8DecodeBytes (buffer) {
-  return buffer.length >= 3
-    ? buffer.toString('utf8', buffer[0] === 0xEF && buffer[1] === 0xBB && buffer[2] === 0xBF ? 3 : 0)
-    : buffer.toString('utf8')
+function chunksDecode (chunks, length) {
+  const buffer = Buffer.concat(chunks, length)
+  return buffer.toString(
+    'utf8',
+    buffer.length >= 3 &&
+      buffer[0] === 0xef &&
+      buffer[1] === 0xbb &&
+      buffer[2] === 0xbf
+      ? 3
+      : 0
+  )
 }
 
 function consumeEnd (consume) {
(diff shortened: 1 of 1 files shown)


tsctx:
cc @ronag

lib/api/readable.js
```diff
@@ -267,29 +267,44 @@ function consumeStart (consume) {
...
-      if (!Blob) {
-        Blob = require('buffer').Blob
-      }
-      resolve(new Blob(body, { type: stream[kContentType] }))
+    switch (type) {
+      case 'text':
+        resolve(utf8DecodeBytes(Buffer.concat(body, length)))
+        break
+      case 'json':
+        resolve(JSON.parse(utf8DecodeBytes(Buffer.concat(body, length))))
```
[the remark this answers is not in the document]
ronag: Isn't this faster?

```js
let ret = ''
for (const buf of body) {
  ret += decoder.decode(buf, { stream: true })
}
ret += decoder.decode(undefined, { stream: false })
```

lib/api/readable.js
```diff
@@ -267,29 +267,53 @@ function consumeStart (consume) {
...
+          dst.set(buf, pos)
+          pos += buf.byteLength
+        }
+
+        resolve(dst.buffer)
+        break
       }
-      resolve(new Blob(body, { type: stream[kContentType] }))
+      case 'blob':
+        resolve(new (Blob ??= require('buffer').Blob)(body, { type: stream[kContentType] }))
```
KhafraDev, line 315:
```suggestion
        resolve(new Blob(body, { type: stream[kContentType] }))
```

lib/api/readable.js
```diff
@@ -267,29 +267,53 @@ function consumeStart (consume) {
   }
 }
 
+const decoder = new TextDecoder()
+
+/**
+ * @see https://encoding.spec.whatwg.org/#utf-8-decode
+ * @param {Buffer} buffer
+ */
+function utf8DecodeBytes (buffer) {
```
KhafraDev, line 276: import it from the fetch utils instead of duplicating

KhafraDev (changes requested):

lib/api/readable.js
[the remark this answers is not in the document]
tsctx: No.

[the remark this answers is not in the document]
tsctx: No, it is not. It's only faster after concat.

lib/api/readable.js
tsctx: I just changed the implementation and it is ok.

lib/api/readable.js
tsctx: Thank you.

lib/api/readable.js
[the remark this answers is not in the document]
tsctx: It is faster after cancat.

lib/api/readable.js
KhafraDev: not really, the code is duplicated when it doesn't need to be

tsctx: Sorry for the misunderstanding. 
Just wanted to say that this is not a problem since we have changed the implementation.

tsctx: @KhafraDev 
is this all right with you?https://github.com/nodejs/undici/pull/2510/commits/8b07631f4626e46451d23d8bf1bfbeca4648a492

lib/api/readable.js
```diff
@@ -267,29 +265,49 @@ function consumeStart (consume) {
...
+ * @param {number} length
+ */
+function chunksDecode (chunks, length) {
+  const buffer = Buffer.concat(chunks, length)
+  return buffer.toString(
+    'utf8',
+    buffer.length >= 3 &&
+      buffer[0] === 0xef &&
+      buffer[1] === 0xbb &&
+      buffer[2] === 0xbf
```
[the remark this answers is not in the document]
tsctx: Nice suggestion. Thank you!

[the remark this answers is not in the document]
ronag: and why?

lib/api/readable.js
```diff
@@ -267,29 +265,49 @@ function consumeStart (consume) {
...
-      for (const buf of body) {
-        dst.set(buf, pos)
-        pos += buf.byteLength
-      }
-
-      resolve(dst.buffer)
-    } else if (type === 'blob') {
-      if (!Blob) {
-        Blob = require('buffer').Blob
+    switch (type) {
```
ronag, line 290: I don't think changing to a switch case here is strictly necessary...

ronag: I would personally like to avoid switch break statements.

lib/api/readable.js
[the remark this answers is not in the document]
tsctx: The [spec](https://encoding.spec.whatwg.org/#utf-8-decode) reads.
```
2. If buffer is 0xEF 0xBB 0xBF, then [read](https://encoding.spec.whatwg.org/#concept-stream-read) three bytes from ioQueue. (Do nothing with those bytes.)
```

lib/api/readable.js
```diff
@@ -267,14 +265,32 @@ function consumeStart (consume) {
   }
 }
 
+/**
+ * @param {Buffer[]} chunks
+ * @param {number} length
+ */
+function chunksDecode (chunks, length) {
+  const buffer = Buffer.concat(chunks, length)
```
ronag, line 273:
```suggestion
  const buffer = chunks.length === 1 ? chunks[0] : Buffer.concat(chunks, length)
```

lib/api/readable.js
```diff
@@ -267,14 +265,32 @@ function consumeStart (consume) {
...
+    'utf8',
+    buffer.length >= 3 &&
+    // Skip BOM.
+      buffer[0] === 0xef &&
+      buffer[1] === 0xbb &&
+      buffer[2] === 0xbf
+      ? 3
+      : 0
+  )
+}
```
ronag, line 284:
```suggestion
  return buffer.utf8Slice(
    0,
    buffer.length >= 3 &&
    // Skip BOM.
      buffer[0] === 0xef &&
      buffer[1] === 0xbb &&
      buffer[2] === 0xbf
      ? 3
      : 0
  )
}
```

ronag (approved):

lib/api/readable.js
```diff
@@ -267,14 +265,35 @@ function consumeStart (consume) {
...
+
+  const start =
+   buffer.length >= 3 &&
+   // Skip BOM.
+      buffer[0] === 0xef &&
+      buffer[1] === 0xbb &&
+      buffer[2] === 0xbf
+     ? 3
+     : 0
+  return buffer.utf8Slice(start, buffer.length - start)
```
ronag, line 286: @KhafraDev Do we need to do a `toWellFormed` here?

tsctx: @ronag It looks like no problem.
```js
import { request, MockAgent, setGlobalDispatcher } from "./index.js";

const Q = "\ud800\uda04"

setGlobalDispatcher(agent);

const mockPool = agent.get("http://localhost:3000");
mockPool.intercept({ path: "/foo" }).reply(200, Q);

const requestResponse = await (
  await request("http://localhost:3000/foo")
).body.text();

console.log(requestResponse === Q); // false
console.log(requestResponse === Q.toWellFormed()); // true
```

tsctx: @ronag It looks like no problem.
```js
import { request, MockAgent, setGlobalDispatcher, Response } from "./index.js";
const Q = "\ud800\uda04";
setGlobalDispatcher(agent);
const mockPool = agent.get("http://localhost:3000");
mockPool.intercept({ path: "/foo" }).reply(200, Q);
const requestResponse = await (
  await request("http://localhost:3000/foo")
).body.text();
console.log(requestResponse === Q); // false
console.log(requestResponse === Q.toWellFormed()); // true
console.log(requestResponse === await new Response(Q).text()); // true
console.log(requestResponse === new TextDecoder().decode(new TextEncoder().encode(Q))) // true
console.log(requestResponse === Buffer.from(Q).toString()) // true
```

ronag, line 286:
```suggestion
  return buffer.utf8Slice(start, buffer.length - start).toWellFormed()
```

ronag (changes requested):
I think we need to keep `toWellFormed/toUSVString ` unfortunately.

lib/api/readable.js
KhafraDev: we do

KhafraDev: it was added in node v20, @tsctx you need to use the wrapper for it which falls back to toUSVString

lemire: @KhafraDev Why is it needed? Can you give an example?

lib/api/readable.js
```diff
@@ -267,14 +265,35 @@ function consumeStart (consume) {
...
 function consumeEnd (consume) {
   const { type, body, resolve, stream, length } = consume
 
   try {
     if (type === 'text') {
-      resolve(toUSVString(Buffer.concat(body)))
+      resolve(chunksDecode(body, length))
     } else if (type === 'json') {
-      resolve(JSON.parse(Buffer.concat(body)))
+      resolve(chunksDecode(body, length))
```
lemire, line 296: So here the `JSON.parse` was irrelevant? Wow.

lib/api/readable.js
KhafraDev: maybe utf8slice does the same, idk

ronag: No idea. Just following specs.

ronag: What is that for then? If it doesn't do anything?

lib/api/readable.js
ronag: Are we missing a test?

lib/api/readable.js
lemire: I submit to you that if you need it then there should be a corresponding test. 

@anonrig can you help?

metcoder95: `toWellFormed`:
>The toWellFormed() method of [String](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String) values returns a string where all [lone surrogates](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String#utf-16_characters_unicode_code_points_and_grapheme_clusters) of this string are replaced with the Unicode replacement character U+FFFD.

The difference seems to be that `toUSVString` removes all surrogates, meanwhile `toWellFormed` is just the `lone-surrogates`. I'd imagine we can choose based on the use case we are covering.

KhafraDev: so we should either stop calling these spec compliant body mixins or remove the utf8slice because the 2 aren't compatible?

lemire: @metcoder95 The input is UTF-8 which does not have a concept of surrogate. The closest you can come to is an invalid UTF-8 byte sequence that would translate into a lone surrogate if you just did UTF-8 to UTF-16 transcoding in a broken manner (e.g., using the non-standard WTF-8 interpretation) but `utf8Slice` does not accept invalid UTF-8, it uses replacement characters instead... So it cannot produce something that would lead to a lone surrogate.

```
> Buffer.from(Buffer.from([0xed,0xb8,0x80]).utf8Slice(0,3).toWellFormed())
<Buffer ef bf bd ef bf bd ef bf bd>
> Buffer.from(Buffer.from([0xed,0xb8,0x80]).utf8Slice(0,3))
<Buffer ef bf bd ef bf bd ef bf bd>
> Buffer.from(Buffer.from([0xed,0xb8,0x80]).toString())
<Buffer ef bf bd ef bf bd ef bf bd>
> Buffer.from(util.toUSVString(Buffer.from([0xed,0xb8,0x80])))
<Buffer ef bf bd ef bf bd ef bf bd>
```

Lone surrogates are a feature of UTF-16. The Unicode standard specifically forbids having a lone surrogate in UTF-8:

https://www.unicode.org/faq/utf_bom.html#utf8-5

You may argue that the input might be ill-formed, and indeed, it might be, but the functions we are using (`toString()`, `utf8Slice()`) do transcoding with replacement, so they 'fix' the input.

Here is a surrogate pair:
```
> Buffer.from([0x3d,0xd8, 0x00,0xde]).toString("UTF-16LE")
'😀'
```

Here is a lone surrogate pair:

```
> Buffer.from([0x3d,0xd8]).toString("UTF-16LE")
'\ud83d'
```

But in the code that we are considering here, we seem to assume an UTF-8 input. So I don't understand.

lemire: @KhafraDev I personally do not know, that's why I am raising the question.

lib/api/readable.js
tsctx: Sorry, it was accidentally deleted.

tsctx:
The node v18 `test/balanced-pool.js` test is failing, but this seems to be happening only with the latest v18.19.0

ronag (approved):

PR #2516, opened 2023-12-09
perf: reduce `String#toLowerCase` call

tsctx:
```js
import { fetch } from "./index.js";

const toLowerCase = String.prototype.toLowerCase

let call = 0

String.prototype.toLowerCase = function () {
    ++call
    return toLowerCase.call(this)
}

await fetch("https://google.com")

// before -> 75
// after -> 4
console.log(call)

```

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -79,6 +79,7 @@ const {
   kHTTP2CopyHeaders,
   kHTTP1BuildRequest
 } = require('./core/symbols')
+const { tree } = require('./core/tree')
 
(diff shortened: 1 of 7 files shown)


tsctx:
Hi @KhafraDev @ronag @metcoder95,
What do you think about this?

ronag:
Can you make a helper so that we don't need to know about tree

ronag (approved):

mcollina (approved):
lgtm

PR #2519, opened 2023-12-13
fix: clear cache

tsctx:


lib/fetch/headers.js
```diff
@@ -241,6 +241,8 @@ class HeadersList {
    * @param {string} value
    */
   lowerCaseSet (name, value) {
+    this[kHeadersSortedMap] = null
```
tsctx, line 244: I'm sorry. I deleted it by mistake.

ronag (approved):

lib/fetch/headers.js
ronag: Are we missing a test that can cache this?

tsctx: No, I forgot to add here. https://github.com/nodejs/undici/pull/2516

tsctx: > Are we missing a test that can catch this?

Should I be adding tests?

tsctx: There seems to be no HeadersList testing.

metcoder95: Please 🙏

metcoder95 (approved):

lib/fetch/headers.js
KhafraDev: why do we need lowercaseGet/lowercaseDelete? Problems like these occur when we deviate from the spec...

lib/fetch/headers.js
```diff
@@ -257,7 +259,14 @@ class HeadersList {
    * @param {string} name
    */
   delete (name) {
```
KhafraDev, line 261: this is missing comments

tsctx: There were no comments originally.

KhafraDev (changes requested):
I do want an answer to my question - why do we need the lowercase_ methods? I didn't have time to review the original PR.

lib/fetch/headers.js
tsctx: > why do we need lowercaseGet/lowercaseDelete? Problems like these occur when we deviate from the spec...

As shown in #2516, too many toLowerCase are called.
A simple fetch would be called at least 30 times.

tsctx:
@KhafraDev Sorry I'm late in answering.

lib/fetch/headers.js
KhafraDev: I understand that, but I don't understand why they have to be separate. As in, why can't lowerCaseAppend just be append?

tsctx: Another advantage is that this method, added in #2501, can be used.

tsctx: > As in, why can't lowerCaseAppend just be append?

I didn't do that because I was afraid I might accidentally call it that. Even though it's capitalized.

tsctx: > As in, why can't lowerCaseAppend just  be append?

I tried to change the append, I didn't do that because I was afraid I might accidentally call it that. Even though it's capitalized.

KhafraDev: if append is identical to lowerCaseAppend, why are you afraid of calling it that...? it's following the steps from the spec? What's capitalized?

KhafraDev: Okay I disagree heavily on implementation but I understand what it's doing. Maintaining 2 versions of all of these is not maintainable and should be removed. Instead of the `lowerCase_` methods, why not add a third parameter to them called "isLowercase"/something along the lines of that (and then handling casing there)? It's closer to the spec and much easier to maintain that way.

tsctx: Okey I'll do it that way.

tsctx:
@KhafraDev Can you take a look at it?

KhafraDev (approved):
thank you

PR #2525, opened 2023-12-19
feat: expose headerNameToString

tsctx:
This is useful when processing header name.

### DIFF
--- a/types/util.d.ts
+++ b/types/util.d.ts
@@ -1,0 +1,31 @@
+export namespace util {
+  /**
+   * Retrieves a header name and returns its lowercase value.
+   * @param value Header name
+   */
+  export function headerNameToString(value: string | Buffer): string;
+
+  /**
+   * Receives a header object and returns the parsed value.
+   * @param headers Header object
+   */
+  export function parseHeaders(
+    headers:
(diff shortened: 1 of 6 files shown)


types/util.d.ts
```diff
@@ -0,0 +1,42 @@
...
+    headers: H,
+    obj?: H extends any[] ? T : never
+  ): T extends Record<string, infer V>
+    ? unknown extends V
+      ? Record<string, string | string[]>
+      : H extends any[]
+      ? Record<string, string | string[] | V>
+      : Record<string, string | string[]>
+    : never;
+}
```
tsctx, line 42: It is a little more complicated, but it can handle all cases.

tsctx, line 42: It is a little more complicated, but it can handle all cases.

ronag (approved):

metcoder95:
@ronag @mcollina did we make changes recently to the BalancedPool?
I'm seeing other PRs starting to fail because of its tests

tsctx:
@metcoder95 
This is due to node. This happens in the latest version (v18.19.0), but not in the previous version (v18.18.2).

PR #2528, opened 2023-12-20
perf: use tree

tsctx:


mcollina:
Can you rebase? CI should be better

tsctx:
@mcollina I was able to fix it.

metcoder95 (approved):

mcollina (approved):
lgtm

Can you rebase again? CI should be green now

KhafraDev (approved):
LGTM but I missed the original discussion - is converting it to a string faster than checking if the buffers are equal?

tsctx:
@KhafraDev It is a little faster, but the difference is negligible.

PR #2530, opened 2023-12-20
Remove FinalizationRegistry from Agent

mcollina:
Given https://github.com/nodejs/node/issues/49344 and https://github.com/nodejs/node/issues/47748, it's better to reconsider our usage of `FinalizationRegistry` in `Agent`.


mcollina:
The happy part is is that this _fixes_ our flaky CI too.

metcoder95 (approved):

Issue #2531, opened 2023-12-20
Dispatcher.connect doesn't include `origin` field


dvoytenko:
## Bug Description

[`ConnectOptions`](https://github.com/nodejs/undici/blob/871baa74ec326518a3de3ee4591459e25617402e/types/dispatcher.d.ts#L123) interface doesn't include an `origin` field that's required for the [HTTP CONNECT](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/CONNECT).

Furthermore, it appears that the `agent.connect({})` currently fails to execute without the `origin` field.

## Reproducible By

```
const agent = new Agent();
agent.connect({
  origin: "https://example.com",
  path: "/robots.txt"
 });
```

This gives a TypeScript error for unknown "origin" field.

## Expected Behavior

Origin is accepted.

## Environment

Any

metcoder95:
Would you like to send a PR?

dvoytenko:
@metcoder95 I took a stub at it in https://github.com/nodejs/undici/pull/2532. But not sure what to do with backward compat, so I left it an optional field.

KhafraDev:
Type changes are always considered minor changes, it's fine to make the field required (especially since it doesn't work without it).

PR #2532, opened 2023-12-20
ConnectOptions should include 'origin' field

dvoytenko:
## Rationale

[HTTP CONNECT](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/CONNECT) requires host and port that already appear to be derived from the `origin` field. It's simply not declared in a type. Furthermore, `agent.connect()` currently fails without the `origin` field in [this line](https://github.com/nodejs/undici/blob/0c4c4504852c71dac1a6eb8dfae0f2411b6f2fc6/lib/agent.js#L90).

## Changes

The `origin` field is being added to the `ConnectOptions`.

This should actually be a required field, but that would make the API non-backward compatible.

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

Fixes https://github.com/nodejs/undici/issues/2531

### DIFF
--- a/test/types/agent.test-d.ts
+++ b/test/types/agent.test-d.ts
@@ -86,8 +86,13 @@ expectAssignable<Agent>(new Agent({ factory: () => new Dispatcher() }))
   }))
 
   // connect
-  expectAssignable<Promise<Dispatcher.ConnectData>>(agent.connect({ path: '' }))
-  expectAssignable<void>(agent.connect({ path: '' }, (err, data) => {
+  expectAssignable<Promise<Dispatcher.ConnectData>>(agent.connect({ origin: '', path: '' }))
+  expectAssignable<Promise<Dispatcher.ConnectData>>(agent.connect({ origin: new URL('http://localhost'), path: '' }))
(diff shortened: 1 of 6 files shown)


KhafraDev:
Can you add a test? You can run the typescript tests by running `npm run test:typescript`. The test can be added [here](https://github.com/nodejs/undici/blob/main/test/types/agent.test-d.ts).

dvoytenko:
@KhafraDev 

> Can you add a test? You can run the typescript tests by running `npm run test:typescript`. The test can be added [here](https://github.com/nodejs/undici/blob/main/test/types/agent.test-d.ts).

I added tests. 

I also updated `origin` to be a required field in the `Dispatcher`. That created a small blast radius, so I had to update a few more APIs. For instance, `Client.connect()` doesn't need `origin` since it's specified in the `Client`'s constructor.

metcoder95 (approved):

KhafraDev (approved):

Ethan-Arrowood (approved):
Excellent! Thank you Dima!

KhafraDev:
thank you!

PR #2533, opened 2023-12-21
chore: reduce dependencies

tsctx:


package.json
```diff
@@ -119,8 +118,6 @@
     "jsfuzz": "^1.0.15",
     "mitata": "^0.1.6",
     "mocha": "^10.0.0",
-    "mockttp": "^3.9.2",
```
tsctx, line 122: This is not actually used.
https://github.com/search?q=repo%3Anodejs%2Fundici+mockttp&type=code

ronag (changes requested):
mockttp ok, delay I don't see any point in re-implementing ourselves.

tsctx:
> mockttp ok, delay I don't see any point in re-implementing ourselves.

I did so simply to replace it.

@ronag I made some changes, can you take a look?

ronag:
Don't add more code :/. For dev deps it's fine to have dependencies.

package.json
metcoder95: Thanks, I left it when working on the `RedirectHandler` and forget to remove it

tsctx:
@ronag Is that all right with you?

metcoder95 (approved):

KhafraDev (approved):

Issue #2116, opened 2023-05-11
Some test take long time to run


ronag:
e.g. jsdom-abortcontroller-1910-1464495619.js takes several seconds to complete.

ronag:
@KhafraDev something you could investigate?

ronag:
https://www.npmjs.com/package/why-is-node-running could be useful

Uzlopak:
@ronag isnt this fixed with PR #2534 by @tsctx?

PR #2534, opened 2023-12-21
fix: call explicitly unregister

tsctx:
Fix flaky CI test of `jsdom-abortcontroller-1910-1464495619.js`

### DIFF
--- a/lib/fetch/request.js
+++ b/lib/fetch/request.js
@@ -371,6 +371,18 @@ class Request {
         const abort = function () {
           const ac = acRef.deref()
           if (ac !== undefined) {
+            // Currently, there is a problem with FinalizationRegistry.
+            // https://github.com/nodejs/node/issues/49344
+            // https://github.com/nodejs/node/issues/47748
+            // In the case of abort, the first step is to unregister from it.
+            // If the controller can refer to it, it is still registered.
+            // It will be removed in the future.
+            requestFinalizer.unregister(abort)
+
+            // Unsubscribe a listener.
+            // FinalizationRegistry will no longer be called, so this must be done.
+            this.removeEventListener('abort', abort)
+
             ac.abort(this.reason)
           }
         }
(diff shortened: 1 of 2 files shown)


tsctx:
@ronag 
I think I could fix #2116, what do you think?

lib/fetch/request.js
```diff
@@ -388,7 +393,7 @@ class Request {
         } catch {}
 
         util.addAbortListener(signal, abort)
-        requestFinalizer.register(ac, { signal, abort })
+        requestFinalizer.register(ac, { signal, abort, controller: ac })
```
mcollina, line 396: This does not work because it will make the finalizer _nerver_ called, because the `ac` is always referenced. You would need to call `.unregister` elsewhere.

tsctx: I will find a way to resolve this.

tsctx:
@mcollina 
Can you take a look at it? This should not be a problem.
I have run CI several times and never once has this test failed.

lib/fetch/request.js
```diff
@@ -35,6 +35,12 @@ const { getMaxListeners, setMaxListeners, getEventListeners, defaultMaxListeners
 const kAbortController = Symbol('abortController')
 
 const requestFinalizer = new FinalizationRegistry(({ signal, abort }) => {
+  // Currently FinalizationRegistry has a problem and will explicitly call unregister.
+  // https://github.com/nodejs/node/issues/49344
+  // https://github.com/nodejs/node/issues/47748
+  // It will be removed in the future.
+  // The unregister key is abort.
+  requestFinalizer.unregister(abort)
```
mcollina, line 43: you don't need to call unregister here. The finalizationregistry already fired.

mcollina (changes requested):
good work!

lib/fetch/request.js
tsctx: Thank you for the notice.

tsctx:
@mcollina Can you please confirm?

mcollina (approved):
lgtm

PR #2535, opened 2023-12-21
Clear all timeout on destroy and close

mcollina:
Followup from https://github.com/nodejs/undici/pull/2530


ronag (approved):

metcoder95 (approved):

mcollina:
This should not have landed

KhafraDev:
my bad, there wasn't any indication that it shouldn't

mcollina:
please revert, this will crash things now

PR #2541, opened 2023-12-25
fix: check the content-type of invalid formData

tsctx:
The Content-Type is not strictly checked.
If `multipart/form-data` or `application/x-www-form-urlencoded` is included, Error is not thrown.
```shell
> $ node
Welcome to Node.js v21.5.0.
> [...await (new Response("x=y", { headers: { "content-type": "application/x-www-form-urlencoded_" } }).formData())]
[ [ 'x', 'y' ] ]
```


KhafraDev (approved):
in the future, I wouldn't bother with .formData. It's a disaster that I purposefully avoid.

mcollina (approved):
lgtm

PR #2546, opened 2023-12-27
Draft: abort for maximum redirects rather than null

VigneshMurugan:
## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/handler/constants.js
+++ b/lib/handler/constants.js
@@ -1,0 +1,7 @@
+'use strict'
+
+const MAX_REDIRECTS = 'Max Redirects'
+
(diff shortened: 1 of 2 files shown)


lib/handler/constants.js
metcoder95, line 1: This is not needed, feel free to keep the `message` within the handler

metcoder95 (commented):
Please, do not forget to add unit testing

mcollina (changes requested):

Issue #2267, opened 2023-09-15
Move from tap to the node core test runner


mcollina:
As titled

This could be part of the next semver major that drops old nodes.

ashutoshdhande:
Will give a shot at this

anurag-roy:
@mcollina There are a lot of test files using `tap` (around 150).

Are we looking to migrate all of them in one PR or migrating part by part is fine?

pmarchini:
Hey!

Given the large number of tests and the need to meticulously avoid false positives, would it be more effective to split the test suites into separate PRs?

For instance, `test:cookies` appears to be easily portable independently of other suites.

How about we create a list of these suites to distribute the workload and more effectively monitor the outcomes?

Uzlopak:
Is there something like t.plan in node:test?

anurag-roy:
@Uzlopak There isn't, we are using `@matteo.collina/tspl` package instead of t.plan

Example usage [here](https://github.com/nodejs/undici/blob/main/test/node-test/abort-controller.js#L25).

Uzlopak:
I am on this now. 

I opened a PR #2719 for an example transformation. 
Also I will split the necessary work into smaller PRs to reduce the review work.

Uzlopak:
Opened #2720 and #2721 to incrementally prepare the following PR.

Uzlopak:
LOL. I created basically 6 batches and lost the following stuff, which I stashed in git. 

So please review and merge the open PRs. Then I can continue to migrate it.

Uzlopak:
Nice.

PR #2547, opened 2023-12-27
feat: port cookies tests to node runner

pmarchini:
### This relates to...

https://github.com/nodejs/undici/issues/2267

### Rationale

Port cookie tests to node test runner

### Changes

Remove tap and use node test runner for the cookie suite

## Status

- [x] Tested
- [S] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/package.json
+++ b/package.json
@@ -100,6 +100,7 @@
     "@types/node": "^18.0.3",
     "abort-controller": "^3.0.0",
     "atomic-sleep": "^1.0.0",
+    "borp": "^0.4.2",
     "chai": "^4.3.4",
     "chai-as-promised": "^7.1.1",
     "chai-iterator": "^3.0.2",
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     "lint": "standard | snazzy",
     "lint:fix": "standard --fix | snazzy",
     "test": "node scripts/generate-pem && npm run test:tap && npm run test:node-fetch && npm run test:fetch && npm run test:cookies && npm run test:wpt && npm run test:websocket && npm run test:jest && npm run test:typescript",
-    "test:cookies": "borp --coverage test/cookie/*.js",
+    "test:cookies": "borp --coverage \"test/cookie/*.js\"",
     "test:node-fetch": "mocha --exit test/node-fetch",
     "test:fetch": "npm run build:node && tap --expose-gc test/fetch/*.js && tap test/webidl/*.js",
     "test:jest": "jest",
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     "lint": "standard | snazzy",
     "lint:fix": "standard --fix | snazzy",
     "test": "node scripts/generate-pem && npm run test:tap && npm run test:node-fetch && npm run test:fetch && npm run test:cookies && npm run test:wpt && npm run test:websocket && npm run test:jest && npm run test:typescript",
-    "test:cookies": "tap test/cookie/*.js",
+    "test:cookies": "borp --coverage test/cookie/*.js",
     "test:node-fetch": "mocha --exit test/node-fetch",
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

metcoder95 (approved):

package.json
```diff
@@ -76,7 +76,7 @@
     "lint": "standard | snazzy",
     "lint:fix": "standard --fix | snazzy",
     "test": "node scripts/generate-pem && npm run test:tap && npm run test:node-fetch && npm run test:fetch && npm run test:cookies && npm run test:wpt && npm run test:websocket && npm run test:jest && npm run test:typescript",
-    "test:cookies": "tap test/cookie/*.js",
+    "test:cookies": "borp --coverage test/cookie/*.js",
```
mcollina, line 79:
```suggestion
    "test:cookies": "borp --coverage \"test/cookie/*.js\"",
```

This will fix windows

mcollina: big question why this is needed for borp and not for tap

test/cookie/cookies.js
```diff
@@ -35,28 +36,26 @@ const {
 
 test('Cookie parser', (t) => {
```
ilteoood, line 37: `t` is unused

test/cookie/global-headers.js
```diff
@@ -16,55 +17,47 @@ if (!globalThis.Headers) {
...
           secure: true
         }
       ])
     }
-
-    t.end()
   })
 
-  t.test('setCookie', (t) => {
+  await t.test('setCookie', (t) => {
```
ilteoood, line 57: why await? this test is not async 🤔

pmarchini: It's an alternative way to `describe` / `test` :

```
/**
     * The `test()` function is the value imported from the `test` module. Each
     * invocation of this function results in reporting the test to the `TestsStream`.
     *
     * The `TestContext` object passed to the `fn` argument can be used to perform
     * actions related to the current test. Examples include skipping the test, adding
     * additional diagnostic information, or creating subtests.
     *
     * `test()` returns a `Promise` that resolves once the test completes.
     * if `test()` is called within a `describe()` block, it resolve immediately.
     * The return value can usually be discarded for top level tests.
     * However, the return value from subtests should be used to prevent the parent
     * test from finishing first and cancelling the subtest
     * as shown in the following example.
     *
     * ```js
     * test('top level test', async (t) => {
     *   // The setTimeout() in the following subtest would cause it to outlive its
     *   // parent test if 'await' is removed on the next line. Once the parent test
     *   // completes, it will cancel any outstanding subtests.
     *   await t.test('longer running subtest', async (t) => {
     *     return new Promise((resolve, reject) => {
     *       setTimeout(resolve, 1000);
     *     });
     *   });
     * });
     * ```
     *
     * The `timeout` option can be used to fail the test if it takes longer than`timeout` milliseconds to complete. However, it is not a reliable mechanism for
     * canceling tests because a running test might block the application thread and
     * thus prevent the scheduled cancellation.
     * @since v18.0.0, v16.17.0
     * @param [name='The name'] The name of the test, which is displayed when reporting test results.
     * @param options Configuration options for the test. The following properties are supported:
     * @param [fn='A no-op function'] The function under test. The first argument to this function is a {@link TestContext} object. If the test uses callbacks, the callback function is passed as the
     * second argument.
     * @return Resolved with `undefined` once the test completes, or immediately if the test runs within {@link describe}.
     */
    function test(name?: string, fn?: TestFn): Promise<void>;
```

package.json
pmarchini: > big question why this is needed for borp and not for tap

I'll take a look at it

test/cookie/cookies.js
pmarchini: fixed 🚀

PR #2549, opened 2023-12-27
drop verifyVersion in scripts

KhafraDev:
It's no longer needed since we dropped node 14/16. Didn't remove the script because it might be needed in the future.


mcollina (approved):
lgtm

PR #2550, opened 2023-12-28
perf(websocket): avoid toLowerCase call

tsctx:


KhafraDev (commented):
Handshake performance *doesn't matter* because you rarely connect to a websocket server (lpinca mentioned this to me when I implemented websocket). This is optimizing (is it?) for a very, very cold path.

tsctx:
Thanks for the explanation. I agree with you.

Issue #2440, opened 2023-11-16
Add dns caching by default


ronag:
Something like: https://github.com/szmarczak/cacheable-lookup#readme

Should be an option to disable.

SimoneDutto:
Hi,
I would like to work on this.
What do you mean by adding this to undici?
Is the behavior needed or the whole library?

> This would be my first time contributing to an Open Source Project, but I want to start contributing!

MrSharpp:
can you state some instructions, where its code should be written?

Medhansh404:
Hey!
I'm currently tackling issues #2440 and #1066. My focus is on enhancing the DNS cache lookup functionality. I plan to intercept DNS requests, check the cache before reaching out to the server, and address other caching aspects.

However, I'm facing difficulty locating the specific code related to fetch requests and DNS resolution commands within the files. Your help in navigating and identifying these components would be immensely appreciated!

Medhansh404:
"I can come up with ways to approach this issue Planning to add DNS cache lookup. Considering two ways: 1) Middleware like HTTP dispatcher in 'lib,' or 2) Custom Resolver globally. would appreciate if I can get a head's up on the approach and which one will be better(if any) !!! @ronag

metcoder95:
The more pragmatic the approach, the better; I'd suggest you take a look at how [Socket#connect](https://nodejs.org/docs/latest-v18.x/api/net.html#socketconnectoptions-connectlistener) handles the `lookup` aspect when establishing a connection; and maybe seek something down that line

antoinegomez:
I had a look last evening and the easiest way I think is to do like @metcoder95 suggest by using the lookup option in the socket module.

My PR is not yet ready, need to finish implementation and refactor a bit. This was mostly to play and see if it works.
I will not implement all options like in this module, otherwise why not simply copy/paste it or add it as a peer dependency and provide an option to configure.

Not sure how to make it configurable in the fetch method to keep it rfc compliant.
Could also have a global config for all requests made via undici.

KhafraDev:
@ronag feel free to reopen but as you mentioned it's already possible to do this

```mjs
import CacheableLookup from 'cacheable-lookup';

const cacheable = new CacheableLookup(opts)

const agent = new Agent({
  connect: { lookup: cacheable.lookup }
})
```

metcoder95:
No strong opinion on this, but I see the potential in having it within core; maybe it will require more thought but as @KhafraDev suggests, is already doable by third-party

Issue #2515, opened 2023-12-08
Missing fields on Fetch PerformanceResourceTiming instance


Ethan-Arrowood:
Does anyone know why the PerformanceResourceTiming instance is missing information regarding DNS and what not? 

Here is an example:

```
PerformanceResourceTiming {
    name: 'https://example.com/',
    entryType: 'resource',
    startTime: 257.69891691207886,
    duration: 98.86862516403198,
    initiatorType: 'fetch',
    nextHopProtocol: undefined,
    workerStart: 0,
    redirectStart: 0,
    redirectEnd: 0,
    fetchStart: 257.69891691207886,
    domainLookupStart: undefined,
    domainLookupEnd: undefined,
    connectStart: undefined,
    connectEnd: undefined,
    secureConnectionStart: undefined,
    requestStart: 0,
    responseStart: 0,
    responseEnd: 356.56754207611084,
    transferSize: 300,
    encodedBodySize: 0,
    decodedBodySize: 0
  }
```

That you get from a simple script like:
```js
const url = new URL('https://example.com')
const res1 = await fetch(url);

await res1.text();

const fetchEntry = perfHooks.performance.getEntriesByName(url);
console.log(fetchEntry);
```

I tracked down the places in the source code where this gets created:
1. https://github.com/nodejs/undici/blob/8422aa988243fb4c6c37b78519954d7337cc240b/lib/fetch/index.js#L311-L325
2. https://github.com/nodejs/node/blob/main/lib/internal/perf/resource_timing.js#L205-L228

But it's still not immediately clear to me how to fix this. Any ideas?

Ethan-Arrowood:
I think I found it: https://github.com/nodejs/undici/blob/main/lib/fetch/util.js#L274-L288

These are all `0` - we need to figure out how to populate these values correctly.

Ethan-Arrowood:
And the relevant part of the spec that would be setting some of these values (such as `domainLookupStartTime` (https://fetch.spec.whatwg.org/#ref-for-connection-timing-info-domain-lookup-start-time)

Ethan-Arrowood:
Found it for real: https://github.com/nodejs/undici/blob/main/lib/fetch/index.js#L1581 its just not implemented. 

Someone is welcome to take this on if they'd like. Otherwise I may be able to next week.

Ethan-Arrowood:
Some relevant places to completely implement performanceresourcetiming:
[...]

PR #2552, opened 2023-12-28
feat: add dns cache #2440

antoinegomez:
## Rationale

Implement a dns cache lookup feature as requested.

Copy code from https://github.com/szmarczak/cacheable-lookup/blob/master/source/index.mjs

## Changes

- add lookup in core/connect.js

### Features

- dns cache lookup

### Bug Fixes

N/A

N/A

## Status

- [x] Tested
- [x] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready


antoinegomez:
Note that this is just an initial draft, anyone can jump in and use this as a base or if in wrong direction simply discard it.

- if ipv6 is available on host and returned by lookup use this
- this is a quick copy of the mentioned cacheable-lookup module and it's clearly not finished
- the difficulty is to know how to set the options for this module and pass them down to the core/connect method

mcollina (commented):
I would create a new class DNSResolver and add a `dnsResolver` option to pool and Agent. An Agent or pool would allocate one if it's not passed in as an option. The Agent would pass down its cache to all created pools.

antoinegomez:
Benchmarks:

```
Linux KDE Neon
AMD Ryzen 5 5600X 6-Core
4 x 8GB RAM
```

main branch 9b9ca602bcc33803200507280c2462b60e21b642

```
│ Tests               │ Samples │          Result │ Tolerance │ Difference with slowest │
|─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
│ http - no keepalive │       1 │ 1962.85 req/sec │  ± 0.00 % │                       - │
|─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
│ http - keepalive    │       1 │ 2387.90 req/sec │  ± 0.00 % │               + 21.65 % │
|─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
│ undici - pipeline   │       1 │ 3035.14 req/sec │  ± 0.00 % │               + 54.63 % │
|─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
│ undici - request    │       1 │ 3160.77 req/sec │  ± 0.00 % │               + 61.03 % │
|─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
│ undici - stream     │       1 │ 3807.23 req/sec │  ± 0.00 % │               + 93.96 % │
|─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
│ undici - dispatch   │       1 │ 4833.26 req/sec │  ± 0.00 % │              + 146.24 % │
```


Feature branch:

```
│ Tests               │ Samples │          Result │ Tolerance │ Difference with slowest │
|─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
│ http - no keepalive │       1 │ 1824.48 req/sec │  ± 0.00 % │                       - │
|─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
│ http - keepalive    │       1 │ 2483.32 req/sec │  ± 0.00 % │               + 36.11 % │
|─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
│ undici - request    │       1 │ 3087.63 req/sec │  ± 0.00 % │               + 69.23 % │
|─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
│ undici - pipeline   │       1 │ 3289.09 req/sec │  ± 0.00 % │               + 80.28 % │
|─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
│ undici - stream     │       1 │ 3935.02 req/sec │  ± 0.00 % │              + 115.68 % │
|─────────────────────|─────────|─────────────────|───────────|─────────────────────────|
│ undici - dispatch   │       1 │ 5294.26 req/sec │  ± 0.00 % │              + 190.18 % │
```

lib/core/dns-resolver.js
```diff
@@ -266,6 +266,10 @@ class DNSResolver {
       ignoreNoResultErrors(this._resolve6(hostname, ttl))
     ])
 
+    if (entries[0].status === 'rejected' && entries[1].status === 'rejected') {
+      throw entries[0].reason
+    }
```
KhafraDev, line 271: Why only throw the first error? Wouldn't it be better to throw an AggregateError that contains both?

lib/core/dns-resolver.js
```diff
@@ -0,0 +1,443 @@
+// source: https://raw.githubusercontent.com/szmarczak/cacheable-lookup/9e60c9f6e74a003692aec68f3ddad93afe613b8f/source/index.mjs
```
mcollina, line 1: please include the license at the top of the file

test/dns-resolver.js
```diff
@@ -0,0 +1,935 @@
+const { promises: dnsPromises, V4MAPPED, ADDRCONFIG, ALL } = require('node:dns')
```
mcollina, line 1: add 'use strict' at the top everywhere

lib/core/dns-resolver.js
```diff
@@ -0,0 +1,443 @@
...
+      try {
+        await this._cache.set(hostname, data, cacheTtl)
+      } catch (error) {
+        this.lookupAsync = async () => {
+          const cacheError = new Error(
+            'Cache Error. Please recreate the DNSResolver instance.'
+          )
+          cacheError.cause = error
+
+          throw cacheError
```
mcollina, line 350: I don't think this kind of error handling is correct. Error situations should auto heal when possible

lib/core/dns-resolver.js
```diff
@@ -0,0 +1,443 @@
...
+      }
+    } else if (options.family === 4) {
+      cached = cached.filter((entry) => entry.family === 4)
+    }
+
+    if (options.hints & ADDRCONFIG) {
+      const { _iface } = this
+      cached = cached.filter((entry) =>
+        entry.family === 6 ? _iface.has6 : _iface.has4
+      )
```
mcollina, line 212: Please avoid the use of filter, map etc, they are an order of magnitude (or more) slower than using normal iteration.

mcollina (changes requested):

lib/core/dns-resolver.js
antoinegomez: Here it is because it's possible to specify a custom cache object, if the set method throws it is considered as a fatal and unrecoverable error.

Using the default cache which is a simple Map() this would not happen.

Also happy to remove any form of custom caching, that's was part of the copied module.

lib/core/dns-resolver.js
antoinegomez: doing that, using AggregateError and modifying tests

lib/core/dns-resolver.js
antoinegomez: I translated the code with loops, no optimizations yet. To have the code passing all tests. Made the function less readable. Also note that here it will be common to have one or two items returns and I am yet to see anyone using more than 4-5 items per record. 
Agreeing to keep an eye on the perf side.

lib/core/dns-resolver.js
antoinegomez: yes no probs

test/dns-resolver.js
antoinegomez: done

antoinegomez:
**Update:**

- As mentioned above, happy to refactor a bit more and to remove some functionalities like the custom caching if we want to slim down the code and responsibilities. in the end people could easily use `cacheable-lookup` module in Undici if they want
- Added a file `global-dns-resolver` to get/set the global instance of the resolver, as my guess is that the default behavior we want is to cache across all requests
- There is no option/method to disable the DNS cache globally easily
  - for now you have to create and set a new global Agent with the option `dnsResolver.disable: true`
- By default and for the first release should we not enable it by default to test? This kind of feature could break people production if not tested more carefully. DNS sometimes can behaves in unexpected ways.

mcollina:
I don't think there should be a global dns resolver. We already have a global agent, which is enough to handle this use case.

lib/agent.js
```diff
@@ -22,6 +23,18 @@ function defaultFactory (origin, opts) {
...
 }
 
+function getDNSResolver (options) {
+  if (options.DNSResolver instanceof DNSResolver) {
+    return options.DNSResolver
+  }
+
+  if (options?.dnsResolver?.disable) {
+    return undefined
+  }
```
KhafraDev, line 33: If someone provides a DNSResolver option *and* a `dnsResolver` option, the former takes precedence. Having `disable: true` should take precedence instead. It might also be better off being renamed to dnsResolverOptions or something so there isn't any confusion.

lib/agent.js
```diff
@@ -22,6 +23,18 @@ function defaultFactory (origin, opts) {
     : new Pool(origin, opts)
 }
 
+function getDNSResolver (options) {
+  if (options.DNSResolver instanceof DNSResolver) {
```
KhafraDev, line 27: I think it's okay to assume that if this option is present, it's an instance of DNSResolver

lib/dns-resolver.js
```diff
@@ -0,0 +1,501 @@
...
+} = require('node:dns')
+const { promisify } = require('node:util')
+const os = require('node:os')
+
+const { Resolver: AsyncResolver } = dnsPromises
+
+const kExpires = Symbol('expires')
+const roundRobinStrategies = ['first', 'random']
+
+const supportsALL = typeof ALL === 'number'
```
KhafraDev, line 46: ALL was added in node 12/13, it's always supported. This can be removed.

lib/dns-resolver.js
```diff
@@ -0,0 +1,501 @@
...
+    throw error
+  })
+}
+
+const ttl = { ttl: true }
+const all = { all: true }
+const all4 = { all: true, family: 4 }
+const all6 = { all: true, family: 6 }
+
+class DNSResolver {
```
KhafraDev, line 107: I would prefer using private fields and methods here rather than using underscore-prefixed methods/props.

lib/dns-resolver.js
```diff
@@ -0,0 +1,501 @@
...
+      ignoreNoResultErrors(this._resolve4(hostname, ttl)),
+      ignoreNoResultErrors(this._resolve6(hostname, ttl))
+    ])
+
+    if (entries[0].status === 'rejected' && entries[1].status === 'rejected') {
+      const error = new AggregateError(`All resolvers failed for hostname: ${hostname}`)
+      error.errors = [
+        entries[0].reason,
+        entries[1].reason
+      ]
```
KhafraDev, line 319:
```suggestion
      const error = new AggregateError([
        entries[0].reason,
        entries[1].reason
      ], `All resolvers failed for hostname: ${hostname}`)
```

KhafraDev (changes requested):

lib/agent.js
metcoder95: But we allow custom `DNSResolver` instances that do not necessarily extend from `DNSResolver`; maybe we should check if the required methods are provided and fail if not?

metcoder95 (commented):
Can we add tests that covers the usage within the Agent/Pool?

lib/agent.js
antoinegomez: Done

lib/agent.js
antoinegomez: The option is not always present, it allows people to set a custom DNSResolver instead of our one. They could inject another module or develop their own.

I added the check on the function needed.

lib/dns-resolver.js
antoinegomez: Done

lib/dns-resolver.js
antoinegomez: Done

lib/dns-resolver.js
antoinegomez: Done

lib/dns-resolver.js
```diff
@@ -0,0 +1,516 @@
...
+      this.lookup = (hostname, _options, callback) => this.#lookup(hostname, lookupOptions, callback)
+    } else {
+      this.lookup = this.#lookup.bind(this)
+    }
+    this.lookupAsync = this.lookupAsync.bind(this)
+  }
+
+  get _cache () {
+    return this.#cache
+  }
```
KhafraDev, line 184: from the tests, it looks like this is only needed to check how many entries are in the cache. Would it make sense to only expose the number of cached entries, rather than the cache itself? If not, changing these to a symbol property to deter users from using this would be better.

lib/dns-resolver.js
```diff
@@ -0,0 +1,516 @@
...
+
+  get _hostnamesToFallback () {
+    return this.#hostnamesToFallback
+  }
```
KhafraDev, line 188: ditto

KhafraDev (changes requested):
I forgot with my other review but this does need types as well.

lib/agent.js
```diff
@@ -22,6 +23,18 @@ function defaultFactory (origin, opts) {
     : new Pool(origin, opts)
 }
 
+function getDNSResolver (options) {
+  if (options?.dnsResolverOptions?.disable) {
+    return undefined
```
metcoder95, line 28: Should be enough

```suggestion
    return
```

lib/pool.js
```diff
@@ -57,6 +58,7 @@ class Pool extends PoolBase {
         maxCachedSessions,
         allowH2,
         socketPath,
+        lookup: dnsResolver ? dnsResolver.lookup : undefined,
```
metcoder95, line 61:
```suggestion
        lookup: dnsResolver?.lookup
```

lib/dns-resolver.js
```diff
@@ -0,0 +1,516 @@
...
+  #resolve6
+
+  constructor ({
+    cache = new Map(),
+    maxTtl = Infinity,
+    fallbackDuration = 3600,
+    errorTtl = 0.15,
+    resolver = new AsyncResolver(),
+    lookup = dnsLookup,
+    lookupOptions = undefined,
```
metcoder95, line 125:
```suggestion
    lookupOptions,
```

lib/dns-resolver.js
antoinegomez: Yes agree

antoinegomez:
> Can we add tests that covers the usage within the Agent/Pool?

I did add two tests, one that will check that the cache is incrementing when using the DNS resolver and one that check that no cache is being used if disabled.

Since Agent is not exposing the DNSResolver it is not possible to get it and check the cache, but I will have a look for a solution to see if I can intercept/mock it.

mcollina:
linting is failing

README.md
```diff
@@ -365,6 +365,52 @@ Returns: `URL`
...
+All arguments match [`cacheable-lookup`](https://github.com/szmarczak/cacheable-lookup/tree/9e60c9f6e74a003692aec68f3ddad93afe613b8f) package
+
+Extra arguments available in Undici:
+
+* **lookupOptions**
+  * **family** `4 | 6 | 0` - Default: `0`
+  * **hints** [`getaddrinfo flags`](https://nodejs.org/api/dns.html#supported-getaddrinfo-flags)
+  * **all** `Boolean` - Default: `false`
+
+* **roundRobinStrategy** `'first' | 'random'` - Default: `'first'`
```
mcollina, line 412: this should be called `scheduling`. What does `first` vs `random` do?

lib/dns-resolver.js
```diff
@@ -0,0 +1,516 @@
...
+
+      throw error
+    }
+
+    if (options.all) {
+      return cached
+    }
+
+    if (this.roundRobinStrategy === 'first') {
+      return cached[0]
```
mcollina, line 290: if we do not rotate the cache, then it's not round robin.

mcollina (changes requested):
We would need to implement some kind of solution of "Happy Eyeballs" similar to what Node Core did.

The more I look at this, the more I'm thinking this should live in Node.js core and not here, but let's chat about it.

lib/dns-resolver.js
antoinegomez: Done

lib/dns-resolver.js
antoinegomez: Done

README.md
antoinegomez: done

lib/dns-resolver.js
antoinegomez: yes true, I will think of a way to rotate the cache

antoinegomez:
> I forgot with my other review but this does need types as well.

first time I add types in non typescript project, is that ok?

https://github.com/nodejs/undici/pull/2552/files#diff-6704654b8ba0b45cd2cc752d380e7c9f70c6c8aac678524044da44446db6c976

mcollina:
cc @ShogunPanda

lib/dns-resolver.js
```diff
@@ -0,0 +1,516 @@
...
+      error.hostname = hostname
+
+      throw error
+    }
+
+    if (options.all) {
+      return cached
+    }
+
+    if (this.#scheduling === 'first') {
```
metcoder95, line 289: Wondering, what are the intentions with the `scheduling` option?
Its not added within `cacheable-lookup` as it already caches the records

lib/agent.js
```diff
@@ -22,6 +23,18 @@ function defaultFactory (origin, opts) {
     : new Pool(origin, opts)
 }
 
+function getDNSResolver (options) {
+  if (options?.dnsResolverOptions?.disable) {
```
ShogunPanda, line 27: I wonder if it would be a better DX to have `{ dnsResolverOptions: false }` instead to completely disable it.
Not a blocker, just a thought.

types/agent.d.ts
```diff
@@ -22,6 +23,8 @@ declare namespace Agent {
     maxRedirections?: number;
 
     interceptors?: { Agent?: readonly Dispatcher.DispatchInterceptor[] } & Pool.Options["interceptors"]
+
+    dnsResolverOptions: DNSResolver.Options;
```
ShogunPanda, line 27: Similarly, what about having just `dnsResolver` here?

ShogunPanda (commented):
Minor suggestions, but not blockers.
LGTM!

ShogunPanda:
In regard of Happy Eyeballs, let me clarify our options here.

From what I see now, with this new feature undici will attempt the records in the order node (and thus the DNS server) resolves them). This is a good thing, at least as default.

The next step would be to implement Happy Eyeballs. I would add a specific implementation in undici since we can be explicit and control the sockets we open and thus we can implement the full Happy Eyeballs specification. In other words, in Node I was forced to implement a serialized version of the algorithm (not to break too many things), here we don't have to.

If we do this, we have two options:
1. Implement Happy Eyeballs on the Dispatcher class.
2. Implement Happy Eyeballs on the DNSResolver class. Maybe this is too much for it, but it would open the possibility to return to the caller an IP which we already verified via Happy Eyeballs. And/or, potentially, to return the socket we were already able to open.

mcollina:
I think we would need happy eyeballs to enable this by default.

metcoder95:
> The next step would be to implement Happy Eyeballs. I would add a specific implementation in undici since we can be explicit and control the sockets we open and thus we can implement the full Happy Eyeballs specification. In other words, in Node I was forced to implement a serialized version of the algorithm (not to break too many things), here we don't have to.

We should also not forget to add an option to disable it if required

> 1. Implement Happy Eyeballs on the Dispatcher class.

I'd like to see if we can keep the `Dispatcher` class lean, and leverage all this logic to the `DNSResolver` class, so we can also enable users to customize the behaviour as they need.

> 2. Implement Happy Eyeballs on the DNSResolver class. Maybe this is too much for it, but it would open the possibility to return to the caller an IP which we already verified via Happy Eyeballs. And/or, potentially, to return the socket we were already able to open.

Like this pretty much, do you think this can enable features such as #2515?

ShogunPanda:
> > The next step would be to implement Happy Eyeballs. I would add a specific implementation in undici since we can be explicit and control the sockets we open and thus we can implement the full Happy Eyeballs specification. In other words, in Node I was forced to implement a serialized version of the algorithm (not to break too many things), here we don't have to.
> 
> We should also not forget to add an option to disable it if required
> 

Absolutely yes!
> > 1. Implement Happy Eyeballs on the Dispatcher class.
> 
> I'd like to see if we can keep the `Dispatcher` class lean, and leverage all this logic to the `DNSResolver` class, so we can also enable users to customize the behaviour as they need.
> 

I'm fine with it. Maybe a separate class used by the the `DNSResolver`.

> > 2. Implement Happy Eyeballs on the DNSResolver class. Maybe this is too much for it, but it would open the possibility to return to the caller an IP which we already verified via Happy Eyeballs. And/or, potentially, to return the socket we were already able to open.
> 
> Like this pretty much, do you think this can enable features such as #2515?

Probably yes, and I would love it!

lib/dns-resolver.js
antoinegomez: So with `first` it is the same behavior as `cacheable-lookup` and returns the first cached entry from cache.

With `random` it will return a random item from the cache.

This is a way to achieve some sort of round robin without to much complexity.

types/agent.d.ts
antoinegomez: It was asked before to rename the options to `dnsResolverOptions` and have a separate one called `dnsResolver` to specify a custom DNSResolver class.

I changed the disable check on `dnsResolver` instead of putting it in the options.

So to disable it will be:

```js
new Agent({
  dnsResolver: false,
})
```

types/pool.d.ts
```diff
@@ -34,6 +35,8 @@ declare namespace Pool {
     /** The max number of clients to create. `null` if no limit. Default `null`. */
     connections?: number | null;
 
-    interceptors?: { Pool?: readonly Dispatcher.DispatchInterceptor[] } & Client.Options["interceptors"]
+    interceptors?: { Pool?: readonly Dispatcher.DispatchInterceptor[] } & Client.Options["interceptors"];
+
+    dnsResolver?: typeof DNSResolver;
```
mcollina, line 40: please add tests for this types

antoinegomez:
> > > The next step would be to implement Happy Eyeballs. I would add a specific implementation in undici since we can be explicit and control the sockets we open and thus we can implement the full Happy Eyeballs specification. In other words, in Node I was forced to implement a serialized version of the algorithm (not to break too many things), here we don't have to.
> > 
> > 
> > We should also not forget to add an option to disable it if required
> 
> Absolutely yes!
> 
> > > 1. Implement Happy Eyeballs on the Dispatcher class.
> > 
> > 
> > I'd like to see if we can keep the `Dispatcher` class lean, and leverage all this logic to the `DNSResolver` class, so we can also enable users to customize the behaviour as they need.
> 
> I'm fine with it. Maybe a separate class used by the the `DNSResolver`.
> 
> > > 2. Implement Happy Eyeballs on the DNSResolver class. Maybe this is too much for it, but it would open the possibility to return to the caller an IP which we already verified via Happy Eyeballs. And/or, potentially, to return the socket we were already able to open.
> > 
> > 
> > Like this pretty much, do you think this can enable features such as #2515?
> 
> Probably yes, and I would love it!

I added some tests for types, tried to resolved your comments.

For the Happy Eyeballs I would be glad to do it too if that's ok. Any source or link to existing material?

Since I changed the behavior for this to be disabled by default it can be done in another PR.

metcoder95:
Yeah, another PR will be good. So far I believe this is the implementation: https://github.com/ShogunPanda/node/blob/ffb326c583c6a23bc94ff4989fe8a06edb92b011/lib/net.js#L1401

and this is the standard that is based on: https://datatracker.ietf.org/doc/html/rfc8305

Are those resources all right @ShogunPanda?

As pointed out, we should seek to stick to the spec as much as we can as we have better ergonomics here to do so.

ShogunPanda:
@metcoder95 Yes, all good links sir. :)

antoinegomez:
Hello @ShogunPanda, I have read the RFC and also the code linked.

Some thoughts.

The code you linked could be used. For now this PR is creating a custom lookup function to cache dns queries.
So in my understanding if happy eye balls is already implemented in node/net module then it seems redundant to implement it again here.

By default the options for lookup are ```{ family: undefined, hints: 32 }```
And in node/net the happy eye balls happens when: ```if (dnsopts.family !== 4 && dnsopts.family !== 6 && !localAddress && autoSelectFamily)```.
I can see that `autoSelectFamily` is undefined by default in undici.

In the RFC it suggest that we should not wait for all DNS queries to resolve to start connecting. Starting to query ipv6 then ipv4. Would be a bit more complex to make it work.

I also understand that we could implement the happy eye balls in undici using a similar approach than node/net module but since we are relying on it I am wondering if it worth it.

ShogunPanda:
In Node.js it is going to use Happy Eyeballs only if using a hostname when doing `net.connect`, which is not the case in undici since we're using a DNS cache.
We don't have to implement Happy Eyeballs exactly as the RFC mandates, for instance not even Node is doing it.

osuka:
Great to see work on this, we are very interested in the possibility of DNS caching as in kubernetes in particular the DNS lookups become a big issue at scale (even with local caching)

DarkGL:
@antoinegomez could I help with that pr ?

ronag:
Closing due to staleness.

Issue #2441, opened 2023-11-16
HashWithOptions regex matches invalid attributes


rtkdpalozzi:
## Bug Description

it looks like the regex at https://github.com/nodejs/undici/blob/main/lib/fetch/util.js#L604 (`/((?<algo>sha256|sha384|sha512)-(?<hash>[A-z0-9+/]{1}.*={0,2}))( +[\x21-\x7e]?)?/i`) is incorrect in a few different ways. `A-z` will also match `[\]^_` which are invalid `base64` characters, and the range `[A-z0-9+/]` is only matched once anyways - the second character and all others until any padding are matched via `.*`. i didn't fully read the spec but i believe the latter half, `( +[\x21-\x7e]?)?`, is also incorrect and won't accurately match the `option-expression`

## Reproducible By

N/A

## Expected Behavior

The regex should only match valid integrity attributes and no other values.

## Logs & Screenshots

N/A

## Environment

N/A

### Additional context

* https://w3c.github.io/webappsec-subresource-integrity/#the-integrity-attribute

KhafraDev:
Would you like to send in a PR? Remember to add tests.

Uzlopak:
@nodejs/undici 

The PR got merged, so this issue is solved ;)

PR #2561, opened 2023-12-29
Fix parseHashWithOptions regex

flapenna:
## This relates to...

#2441 

## Rationale

`[A-z]` will also match `[\]^_` which are invalid base64 characters.

The current regex `[A-z0-9+/]{1}.*={0,2})` will match the initial valid characters in the hash group, even if the rest of the string contains an invalid Base64 string like **zifp5hE1Xl5LQQqQz[]Bq/iaq9Wb6jVb/** where the matched part is **zifp5hE1Xl5LQQqQz**

## Changes

change the hash group to `(?<hash>[A-Za-z0-9+/]+={0,2}(?=\s|$))` where `[A-Za-z0-9+/]`+ are valid base64 chars and `(?=\s|$))` is a positive lookahead assertion.

`( +[!-~]*)` looks for a group of printable ASCII chars as documented in https://datatracker.ietf.org/doc/html/rfc5234#appendix-B.1

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/fetch/util.js
+++ b/lib/fetch/util.js
@@ -584,7 +584,7 @@ function bytesMatch (bytes, metadataList) {
 // https://w3c.github.io/webappsec-subresource-integrity/#grammardef-hash-with-options
 // https://www.w3.org/TR/CSP2/#source-list-syntax
 // https://www.rfc-editor.org/rfc/rfc5234#appendix-B.1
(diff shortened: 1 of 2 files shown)


KhafraDev (approved):

metcoder95 (approved):

mcollina (approved):
lgtm

KhafraDev:
thank you!

Issue #2436, opened 2023-11-14
Redirect Handler throwOnMaxRedirect


ronag:
Would like to have an option where the redirect handler throws on max redirects instead of just forwarding the result.

ronag:
@metcoder95

metcoder95:
It should be a matter of calling `this.abort(new Error('max redirects')` if the flag is set to true, right here: https://github.com/nodejs/undici/blob/ba4ca327843de62a83c1f9c32acc303bd6b8545f/lib/handler/RedirectHandler.js#L90-L92

Do you have something else to cover in mind?

VigneshMurugan:
https://github.com/nodejs/undici/pull/2546 

@metcoder95

Lewiscowles1986:
See #951 as commented in #2260 the redirect interceptor and handler seem to lack any meaningful direct unit tests

mertcanaltin:
landed: https://github.com/nodejs/undici/commit/a7551ce4e2f9d0f8ee1b5a44d4c429aae6d85abb

can I turn this problem off ?

PR #2563, opened 2023-12-30
feat: implement throwOnMaxRedirect option for RedirectHandler

mertcanaltin:
This commit introduces a new enhancement to the RedirectHandler. The issue #2436 reported by @ronag requested the ability to throw an error when the maximum number of redirects is reached, rather than just forwarding the result. To implement this, a new option 'throwOnMaxRedirect' has been added.

The enhancement allows users to set the 'throwOnMaxRedirect' flag to true, and if the maximum redirection limit is reached, the RedirectHandler will now throw an error with the message 'max redirects' instead of forwarding the result.

This improvement is designed to provide better control and error handling in scenarios where strict redirection limits need to be enforced.

issue: #2436


mertcanaltin:
I want to add a test, I need some support, I would appreciate if anyone can help me. ❤️

lib/handler/RedirectHandler.js
```diff
@@ -91,6 +91,11 @@ class RedirectHandler {
       ? null
       : parseLocation(statusCode, headers)
 
+    if (this.history.length >= this.maxRedirections) {
```
metcoder95, line 94: Remember to add the control flag, otherwise, this will throw every time we reach max redirections, and might not be expected by the implementer.

metcoder95 (commented):
To test it out you can do something similar to what's done for `redirect-request` testing. Check this out: https://github.com/nodejs/undici/blob/0f7769bc947d9e6a69dc3d7f9ca67bbd4c68496d/test/redirect-request.js#L27

lib/handler/RedirectHandler.js
mertcanaltin: thank you very much, as you said, I reviewed and improved the test and then I updated my code and added a flag 

first I solved my errors in the test/redirect-request.js file, but then I encountered an error in the test/redirect-stream.js file, I added a few logs and the error output was like this 

```js
 FAIL ​ test/redirect-stream.js
 ✖ max redirects


        this.abort(new Error('max redirects'))
-------------------^
      }

  test: should follow a redirect chain up to the allowed
number of times
  stack: |
    RedirectHandler.onHeaders
(lib/handler/RedirectHandler.js:106:20)
    Request.onHeaders (lib/core/request.js:269:29)
    Parser.onHeadersComplete (lib/client.js:924:27)
    wasm_on_headers_complete (lib/client.js:536:30)
    null.<anonymous> (wasm://wasm/0003626a:1:1173)
    null.<anonymous> (wasm://wasm/0003626a:1:4100)
    null.<anonymous> (wasm://wasm/0003626a:1:28303)
    null.<anonymous> (wasm://wasm/0003626a:1:5481)
    null.<anonymous> (wasm://wasm/0003626a:1:1363)
    Parser.execute (lib/client.js:677:22)
  at:
    line: 106
    column: 20
    file: lib/handler/RedirectHandler.js
    function: RedirectHandler.onHeaders
  tapCaught: returnedPromiseRejection

test/redirect-request.js 1> History Length: 2
test/redirect-request.js 1> Max Redirects: 2
test/redirect-request.js 1> Redirection Limit Reached: true
test/redirect-request.js 1> History Length: 2
test/redirect-request.js 1> Max Redirects: 2
test/redirect-request.js 1> Redirection Limit Reached: true
```

lib/handler/RedirectHandler.js
```diff
@@ -38,6 +38,7 @@ class RedirectHandler {
     this.maxRedirections = maxRedirections
     this.handler = handler
     this.history = []
+    this.redirectionLimitReached = false
```
metcoder95, line 41: This flag is the one that the implementer should provide to enable the feature. If the caller sets it to `false` it means that doesn't want to have that feature and would rather prefer the old behaviour.

For instance we'll need to document it as we are exposing it, also remember to add it to the TS types of the `RedirectHandler`

test/redirect-request.js
```diff
@@ -255,16 +255,24 @@ for (const factory of [
   t.test('should follow a redirect chain up to the allowed number of times', async t => {
     const server = await startRedirectingServer(t)
 
-    const { statusCode, headers, body: bodyStream, context: { history } } = await request(t, server, undefined, `http://${server}/300`, {
-      maxRedirections: 2
-    })
+    try {
+      const { statusCode, headers, body: bodyStream, context: { history } } = await request(t, server, undefined, `http://${server}/300`, {
```
metcoder95, line 259: Good work with the tests, but let's create a new test and keep this one as is. Feel free to copy/paste this test and just change the name; but let's revert this current tests to its original form;

docs/api/RedirectHandler.md
```diff
@@ -0,0 +1,96 @@
+# Class: RedirectHandler
```
mcollina, line 1: add this to the sidebar too

types/handlers.d.ts
```diff
@@ -1,7 +1,7 @@
 import Dispatcher from "./dispatcher";
 
 export declare class RedirectHandler implements Dispatcher.DispatchHandlers{
-  constructor (dispatch: Dispatcher, maxRedirections: number, opts: Dispatcher.DispatchOptions, handler: Dispatcher.DispatchHandlers)
+  constructor (dispatch: Dispatcher, maxRedirections: number, opts: Dispatcher.DispatchOptions, handler: Dispatcher.DispatchHandlers, redirectionLimitReached: boolean)
```
mcollina, line 4: can you add a test for this type?

mcollina (changes requested):

lib/handler/RedirectHandler.js
```diff
@@ -91,6 +92,15 @@ class RedirectHandler {
       ? null
       : parseLocation(statusCode, headers)
 
+    if (this.history.length >= this.maxRedirections && !this.redirectionLimitReached) {
+      if (this.request) {
+        this.request.abort()
+      }
+
+      this.redirectionLimitReached = true
```
metcoder95, line 100: I believe this flag should be handled by the implementation, not necessarily by us (although we can use it as a control flag).

What I mean is, that a use case should look like this:

```js
await request(t, server, undefined, `http://${server}/300`, {
    maxRedirections: 2,
    throwOnMaxRedirect: true
})
``

If `throwOnMaxRedirect` is set to `true`,  whenever we reach this section, we will abort the request with an error (as you are already doing). Otherwise, we continue with the old behaviour (which is to do nothing).

Wdyt?

File: docs/api/RedirectHandler.md
Comment (metcoder95) [inline docs/api/RedirectHandler.md:1]: Thanks for this!

**Review by mcollina — APPROVED** [review]
lgtm

File: lib/handler/RedirectHandler.js (hunk 6 above)
Comment (mertcanaltin) [reply lib/handler/RedirectHandler.js:100]: yes that would be much better thank you very much I tried to implement this in my next commit @metcoder95

File: lib/handler/RedirectHandler.js [hunk 7]
```diff
@@ -91,6 +92,15 @@ class RedirectHandler {
       ? null
       : parseLocation(statusCode, headers)
 
+    if (this.opts.throwOnMaxRedirect && this.history.length >= this.maxRedirections) {
+      if (this.request) {
+        this.request.abort()
+      }
+
+      this.redirectionLimitReached = true
+      this.abort(new Error('max redirects'))
```
metcoder95, line 101: Do not forget to cut the flow with a `return` in here

lib/handler/RedirectHandler.js
```diff
@@ -91,6 +92,15 @@ class RedirectHandler {
       ? null
       : parseLocation(statusCode, headers)
 
+    if (this.opts.throwOnMaxRedirect && this.history.length >= this.maxRedirections) {
+      if (this.request) {
+        this.request.abort()
```
metcoder95, line 97: Let's also abort it with the same error as we are aborting this `RedirectHandler`
```suggestion
        this.request.abort(new Error('max redirects')
```

**Optional**
- You can maybe create an error within [Errors](https://github.com/nodejs/undici/blob/main/lib/core/errors.js) to have a fully fledge undici error

mertcanaltin: thank you very much I will try to create a full thank you error message today @metcoder95

mcollina:
The types tests are not ok.

mertcanaltin:
I corrected the test 🚀

mertcanaltin:
error detail says all tests passed but I wonder if there is something I missed  🤔

mcollina (approved):
lgtm

metcoder95 (approved):

PR #2569, opened 2023-12-31
feat: port first half of fetch tests to node test runner

anurag-roy:
## This relates to...

#2267 

## Rationale

Port first chunk of tests inside the `test/fetch` folder from tap to `node:test` runner.

There are 39 files in the `test/fetch` folder, this PR ports the first 20 of those tests. Doing it in two chunks to reduce code review pressure and to get some early feedback.

## Changes

This PR ports the following tests:
- 407-statuscode-window-null.js
- abort.js
- abort2.js
- about-uri.js
- blob-uri.js
- bundle.js
- client-error-stack-trace.js
- client-fetch.js
- client-node-max-header-size.js
- content-length.js
- cookies.js
- data-uri.js
- encoding.js
- fetch-leak.js
- fetch-timeouts.js
- file.js
- formdata.js
- general.js
- headers.js
- http2.js
- issue-2171.js

## Status

- [x] Tested
- [S] Documented
- [X] Review ready
- [ ] In review
- [ ] Merge ready


mcollina (approved):
lgtm

package.json
```diff
@@ -78,7 +78,7 @@
     "test": "node scripts/generate-pem && npm run test:tap && npm run test:node-fetch && npm run test:fetch && npm run test:cookies && npm run test:wpt && npm run test:websocket && npm run test:jest && npm run test:typescript",
     "test:cookies": "borp --coverage -p \"test/cookie/*.js\"",
     "test:node-fetch": "mocha --exit test/node-fetch",
-    "test:fetch": "npm run build:node && tap --expose-gc test/fetch/*.js && borp --coverage -p \"test/webidl/*.js\"",
+    "test:fetch": "npm run build:node && node --expose-gc --test test/fetch/*.js && borp --coverage -p \"test/webidl/*.js\"",
```
anurag-roy, line 81: Tests are failing on Windows due to this line. Node versions before `v21` do not support the glob pattern and Windows does not resolve the pattern whereas Linux does.

So, I tried a few things:

1. 
    ```
    borp --coverage -p "test/fetch/*.js"
    ```
    But now we aren't able to pass the `--expose-gc` flag, so `fetch-leak.js` which depends on gc fails.
    Also, strangely `issue-2171.js` also fails when run through borp, but not when run individually through `node --test`

2. 
    ```
    node --expose-gc node_modules/borp/borp.js --coverage -p "test/fetch/*.js"
    ```
    This does work, but it is worse in other ways (depends on a file name in installed dep)

### Options

We have a couple of options to go for:

1. Add `--expose-gc` flag directly to borp, so we are able to do
    ```
    borp --expose-gc --coverage -p "test/fetch/*.js"
    ```
    I am working with @mcollina [here](https://github.com/mcollina/borp/pull/4) to see if that can be done

2. Remove dependency from `--expose-gc` flag
    We can update `fetch-leak.js` to dynamically enable gc
    ```
    import { setFlagsFromString } from 'node:v8';
    import { runInNewContext } from 'node:vm';

    setFlagsFromString('--expose-gc');
    const gc = runInNewContext('gc');
    gc() // instead of calling global.gc() directly
    ```
    
Happy to hear anyone's thoughts on this!

test/fetch/http2.js
```diff
@@ -172,27 +171,29 @@ test('[Fetch] Should handle h2 request with body (string or buffer)', async t =>
 
   const responseBody = await response.text()
 
-  t.teardown(server.close.bind(server))
-  t.teardown(client.close.bind(client))
+  t.after(server.close.bind(server))
+  t.after(() => client.close())
```
anurag-roy, line 175: When I write this as
```js
t.after(client.close.bind(client))
```
it fails with the following error
```
Error [InvalidArgumentError]: invalid callback
      at Client.close (/home/anurag/dev/hobby/undici/lib/dispatcher-base.js:62:13)
      at TestHook.runInAsyncScope (node:async_hooks:206:9)
      at TestHook.run (node:internal/test_runner/test:631:25)
      at TestHook.run (node:internal/test_runner/test:856:18)
      at TestHook.run (node:internal/util:531:12)
      at node:internal/test_runner/test:565:20
      at async Test.runHook (node:internal/test_runner/test:563:7)
      at async after (node:internal/test_runner/test:591:9)
      at async Test.run (node:internal/test_runner/test:641:7) {
    code: 'UND_ERR_INVALID_ARG'
  }
```

On digging a little I found that an empty `TestContext {}` is being passed as callback to the `close` method. Not too sure why that is happening.

package.json
metcoder95: I'd prefer option 1, does `borp` has been updated already?

anurag-roy: @metcoder95 Yes just landed in `borp`, will update the code here to option 1

mcollina (approved):
lgtm

metcoder95 (approved):

PR #2580, opened 2024-01-02
fix data url test

### LINKED ISSUE #2579: fetch: data url test fails
Caused by: de7dc7fabbbf7b97edb583b99003cb2222d4166f

```js
const { fetch } = require('./')

const input = "data:  ;charset=x   ;  base64,WA"
const expectedMimeType = "text/plain;charset=x"
const expectedBody = [88]

fetch(input).then(res => {
  return res.arrayBuffer().then(body => {
    assert.deepEqual(new Uint8Array(body), new Uint8Array(expectedBody))
    assert.equal(res.headers.get("content-type"), expectedMimeType)
  })
})

```

@tsctx

KhafraDev:
Fixes #2579 

fix was easier than I expected, I just reverted the changes here for the most part.

### DIFF
--- a/lib/fetch/dataURL.js
+++ b/lib/fetch/dataURL.js
@@ -604,14 +604,18 @@ function isHTTPWhiteSpace (char) {
  * @param {boolean} [trailing=true]
  */
 function removeHTTPWhitespace (str, leading = true, trailing = true) {
-  let i = 0; let j = str.length
+  let lead = 0
+  let trail = str.length - 1
+
   if (leading) {
-    while (j > i && isHTTPWhiteSpace(str.charCodeAt(i))) --i
+    while (lead < str.length && isHTTPWhiteSpace(str.charCodeAt(lead))) lead++
   }
+
   if (trailing) {
-    while (j > i && isHTTPWhiteSpace(str.charCodeAt(j - 1))) --j
+    while (trail > 0 && isHTTPWhiteSpace(str.charCodeAt(trail))) trail--
   }
-  return i === 0 && j === str.length ? str : str.substring(i, j)
+
+  return lead === 0 && trail === str.length - 1 ? str : str.slice(lead, trail + 1)
 }
 
 /**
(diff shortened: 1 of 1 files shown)


lib/fetch/dataURL.js
```diff
@@ -630,14 +630,18 @@ function isASCIIWhitespace (char) {
  * @param {boolean} [trailing=true]
  */
 function removeASCIIWhitespace (str, leading = true, trailing = true) {
-  let i = 0; let j = str.length
+  let lead = 0
+  let trail = str.length - 1
+
   if (leading) {
-    while (j > i && isASCIIWhitespace(str.charCodeAt(i))) --i
```
KhafraDev, line 635: so unless `str.length` was 0 this would always be true... and we're decrementing from i, which means if isASCIIWhitespace(...) was true, `i` would be negative? Not sure why I didn't realize this before.

KhafraDev: also using `i` and `j` instead of `lead` and `trail` shouldn't have landed

mcollina (approved):
lgtm

lib/fetch/dataURL.js
```diff
@@ -604,14 +604,18 @@ function isHTTPWhiteSpace (char) {
  * @param {boolean} [trailing=true]
  */
 function removeHTTPWhitespace (str, leading = true, trailing = true) {
-  let i = 0; let j = str.length
+  let lead = 0
+  let trail = str.length - 1
+
   if (leading) {
-    while (j > i && isHTTPWhiteSpace(str.charCodeAt(i))) --i
+    while (lead < str.length && isHTTPWhiteSpace(str.charCodeAt(lead))) lead++
```
tsctx, line 611: It may be better to save the length.

lib/fetch/dataURL.js
```diff
@@ -630,14 +634,18 @@ function isASCIIWhitespace (char) {
  * @param {boolean} [trailing=true]
  */
 function removeASCIIWhitespace (str, leading = true, trailing = true) {
-  let i = 0; let j = str.length
+  let lead = 0
+  let trail = str.length - 1
+
   if (leading) {
-    while (j > i && isASCIIWhitespace(str.charCodeAt(i))) --i
+    while (lead < str.length && isASCIIWhitespace(str.charCodeAt(lead))) lead++
```
tsctx, line 641: same here

lib/fetch/dataURL.js
KhafraDev: this code is very rarely run twice, it'd be a very pointless optimization (as in, I don't think I've ever seen a library or whatever put whitespaces in data urls).

PR #2581, opened 2024-01-02
mark wpt as failing

KhafraDev:
this test has been failing for a while. Once my other PR lands all tests we run will pass.


mcollina (approved):
lgtm

PR #2582, opened 2024-01-02
fix: trimming issue

tsctx:


tsctx:
@KhafraDev

KhafraDev:
`i` is always 0 when being decremented which means we end up with a negative number. I also prefer my solution because it doesn't rely on the ordering of leading/trailing.

tsctx:
@KhafraDev 
I was wrong. The order doesn't matter.
You can merge whatever you want.

PR #2587, opened 2024-01-03
feat: port remaining fetch tests to node test runner

anurag-roy:
## This relates to...

#2267 

## Changes

Ported the remaining tests in the `fetch` folder which were left over from #2569 

## Status

- [X] Tested
- [S] Documented
- [X] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/test/fetch/integrity.js
+++ b/test/fetch/integrity.js
@@ -1,6 +1,7 @@
 'use strict'
 
-const { test } = require('tap')
+const { test } = require('node:test')
+const assert = require('node:assert')
 const { createServer } = require('http')
 const { createHash, getHashes } = require('crypto')
 const { gzipSync } = require('zlib')
@@ -14,7 +15,7 @@ setGlobalDispatcher(new Agent({
   keepAliveMaxTimeout: 1
 }))
 
-test('request with correct integrity checksum', (t) => {
+test('request with correct integrity checksum', (t, done) => {
   const body = 'Hello world!'
   const hash = createHash('sha256').update(body).digest('base64')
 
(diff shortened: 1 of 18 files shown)


test/fetch/integrity.js
```diff
@@ -14,49 +15,49 @@ setGlobalDispatcher(new Agent({
...
 
-  t.teardown(server.close.bind(server))
+  t.after(server.close.bind(server))
 
   server.listen(0, () => {
     fetch(`http://localhost:${server.address().port}`, {
       integrity: `sha256-${hash}`
     }).then(response => {
-      t.pass('request did not fail')
+      assert.ok(true)
```
KhafraDev, line 51: this test isn't equal - it doesn't look like it will fail if fetch throws an error here?

KhafraDev (changes requested):

test/fetch/integrity.js
KhafraDev: looking at this test, is this meant to be t.fail? We pass if fetch rejects and resolves...?

anurag-roy: @KhafraDev

You're right, we are expecting it to fail, so if it passes, we should fail. I think it was meant to be t.fail.
Should I change the `assert.ok(true)` to something like `throw new Error('request did not fail')`?

KhafraDev: sounds good

anurag-roy: @KhafraDev Updated the test to use `assert.rejects` and match the error, think this is better. Wdyt?

KhafraDev (approved):

PR #2588, opened 2024-01-04
deprecate support for 3rd party blobs

KhafraDev:
Blob has been global since node v18 and File is in node v20. When we drop support for node 18, we should drop support for third party blobs/files.

### DIFF
--- a/lib/core/util.js
+++ b/lib/core/util.js
@@ -6,7 +6,6 @@ const { IncomingMessage } = require('http')
 const stream = require('stream')
 const net = require('net')
 const { InvalidArgumentError } = require('./errors')
-const { Blob } = require('buffer')
 const nodeUtil = require('util')
 const { stringify } = require('querystring')
 const { headerNameLowerCasedRecord } = require('./constants')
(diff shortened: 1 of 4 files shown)


mcollina (changes requested):
I'm not convinced we should ever stop supporting those.

KhafraDev:
could you explain why? It leaves a very large hole in FormData where we perform next to no validation, which allows for things like this:

```js
const fd = new FormData()

fd.set('a', {
  [Symbol.toStringTag]: 'Blob',
  arrayBuffer () { },
  stream () { return 'whatever you want' }
})

fd.get('a') // FileLike

new (fd.get('a').constructor)() // FileLike...

```

mcollina:
Folks uses all sort of old polyfills for this, and they might not even control it (might be an old dependency). I don't think the warning would be ok for Node.js itself on the released lines.

I would just drop this in when we cut support for Node v18 (early 2025).

KhafraDev:
> I would just drop this in when we cut support for Node v18 (early 2025).

sgtm

PR #2592, opened 2024-01-04
abort request with reason if one is provided

### LINKED ISSUE #2590: request throws `AbortError` rather than `TimeoutError` when passed AbortSignal.timeout(1000) as signal
## Bug Description
```
AbortError: Request aborted
    at abort (file:///D:/0/lib/deps.bundle.js:19008:18)
    at self.<computed> (file:///D:/0/lib/deps.bundle.js:19027:5)
    at [nodejs.internal.kHybridDispatch] (node:internal/event_target:822:20)
    at EventTarget.dispatchEvent (node:internal/event_target:757:26)
    at abortSignal (node:internal/abort_controller:374:10)
    at Timeout._onTimeout (node:internal/abort_controller:128:7)
    at listOnTimeout (node:internal/timers:573:17)
    at process.processTimers (node:internal/timers:514:7)
```

## Reproducible By

await request('http://localhost', { signal: AbortSignal.timeout(1000) })

## Expected Behavior

```
TimeoutError: xxx
```

## Logs & Screenshots

image

## Environment

latest node.js
latest windows

KhafraDev:
fixes #2590

### DIFF
--- a/lib/api/abort-signal.js
+++ b/lib/api/abort-signal.js
@@ -6,7 +6,7 @@ const kSignal = Symbol('kSignal')
 
 function abort (self) {
   if (self.abort) {
-    self.abort()
+    self.abort(self[kSignal]?.reason)
   } else {
     self.onError(self[kSignal]?.reason ?? new RequestAbortedError())
   }
--- a/lib/api/abort-signal.js
+++ b/lib/api/abort-signal.js
@@ -8,7 +8,7 @@ function abort (self) {
   if (self.abort) {
     self.abort()
   } else {
-    self.onError(new RequestAbortedError())
+    self.onError(self[kSignal]?.reason ?? new RequestAbortedError())
   }
 }
 
--- a/test/issue-2349.js
+++ b/test/issue-2349.js
@@ -2,7 +2,7 @@
 
 const { test } = require('tap')
 const { Writable } = require('stream')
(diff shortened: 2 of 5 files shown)


KhafraDev:
might be a major change, marking it as such just to be safe

mcollina:
I don't think this is a major change.

ronag (approved):

KhafraDev:
hopefully a flaky test?

```
 not ok 2 - test/ca-fingerprint.js # time=515.802ms
  ---
  env: {}
  file: test/ca-fingerprint.js
  timeout: 60000
  command: C:\hostedtoolcache\windows\node\18.19.0\x64\node.exe
  args:
    - --expose-gc
    - test/ca-fingerprint.js
  stdio:
    - 0
    - pipe
    - 2
  cwd: D:\a\undici\undici
  exitCode: null
  signal: SIGTERM
  ...
```

KhafraDev:
I'm not sure if it's a major change, it does change the errors you get when a reason isn't passed to abort. We made this change/fix in fetch when it was still experimental.

metcoder95 (approved):
I'm a little bit over the fence, but I believe it can pass under a `minor`.
I don't see we are breaking contracts with current usage, in any case we are extending it as we are adding now the reason if exists, otherwise we fallback to the `RequestAbortedError` which they must handle already

PR #2594, opened 2024-01-05
fix: defer error accross micro task queue execution

ronag:
Just deferring a micro task is not sufficient. I still encounter the original issue in production.

### DIFF
--- a/lib/api/readable.js
+++ b/lib/api/readable.js
@@ -62,9 +62,13 @@ module.exports = class BodyReadable extends Readable {
     // never get a chance and will always encounter an unhandled exception.
     // - tick => process.nextTick(fn)
     // - micro tick => queueMicrotask(fn)
-    queueMicrotask(() => {
+    if (err) {
+      setImmediate(() => {
+        callback(err)
+      })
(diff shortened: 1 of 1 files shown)


lib/api/readable.js
```diff
@@ -62,9 +62,13 @@ module.exports = class BodyReadable extends Readable {
     // never get a chance and will always encounter an unhandled exception.
     // - tick => process.nextTick(fn)
     // - micro tick => queueMicrotask(fn)
-    queueMicrotask(() => {
+    if (err) {
+      setImmediate(() => {
+        callback(err)
+      })
```
mcollina, line 68: you can avoid the closure

mcollina (changes requested):
CI does not seem happy. Could you add a test?

lib/api/readable.js
ronag: How?

mcollina: `setImmediate(callback, err)`

ronag:
Looking furhter into it, this might not be necessary after all.

PR #2596, opened 2024-01-07
cacheStorage: fix bugs make wpts pass

KhafraDev:
CacheStorage is weird. All tests pass again.

### DIFF
--- a/lib/cache/cache.js
+++ b/lib/cache/cache.js
@@ -55,7 +55,7 @@ class Cache {
       return
     }
 
-    return p[0]
+    return p[0].clone()
   }
 
   async matchAll (request = undefined, options = {}) {
@@ -112,14 +112,12 @@ class Cache {
     // 5.5.2
     for (const response of responses) {
       // 5.5.2.1
-      const responseObject = new Response(response.body?.source ?? null)
-      const body = responseObject[kState].body
+      const responseObject = new Response(null)
       responseObject[kState] = response
-      responseObject[kState].body = body
       responseObject[kHeaders][kHeadersList] = response.headersList
       responseObject[kHeaders][kGuard] = 'immutable'
 
-      responseList.push(responseObject)
+      responseList.push(responseObject.clone())
     }
 
     // 6.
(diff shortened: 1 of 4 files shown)


lib/cache/cache.js
```diff
@@ -146,16 +144,24 @@ class Cache {
     webidl.brandCheck(this, Cache)
     webidl.argumentLengthCheck(arguments, 1, { header: 'Cache.addAll' })
 
-    requests = webidl.converters['sequence<RequestInfo>'](requests)
```
KhafraDev, line 149: this is weird: any value is technically allowed in RequestInfo (it gets stringified and then parsed as a URL with the window's location typically), but undefined is very explicitly not allowed here for some reason).

lib/cache/cache.js
```diff
@@ -112,14 +112,12 @@ class Cache {
...
-      const responseObject = new Response(response.body?.source ?? null)
-      const body = responseObject[kState].body
+      const responseObject = new Response(null)
       responseObject[kState] = response
-      responseObject[kState].body = body
       responseObject[kHeaders][kHeadersList] = response.headersList
       responseObject[kHeaders][kGuard] = 'immutable'
 
-      responseList.push(responseObject)
+      responseList.push(responseObject.clone())
```
KhafraDev, line 120: this is weird: this whole thing i'm kinda iffy about, somehow this was setting a 0-length uint8array as the source and was locking the stream.

KhafraDev: the cloning itself makes sense, if you matched a Response and then consumed its body, the cached stream would be locked if you matched it again.

ronag (approved):

lib/cache/cache.js
mcollina: How is the body assigned to the new response?

metcoder95 (approved):

lib/cache/cache.js
KhafraDev:
```js
      responseObject[kState] = response
```

`response` here is actually an internal response, not a Response object. It contains the body.

PR #2598, opened 2024-01-07
cleanup index

KhafraDev:
- crypto check is done directly in WebSocket now, it's not needed anymore
- I don't think we need to lazy load fetch anymore?
- default the second argument to `undefined` in fetch because we don't need it to be an object


mcollina (approved):
lgtm

metcoder95 (approved):

PR #2599, opened 2024-01-07
cacheStorage: separate matchAll logic

KhafraDev:
The biggest benefit can be seen from Cache.match where we can now limit the number of responses fetched, instead of returning the entire list while match only requires the first.


mcollina (approved):
lgtm

Issue #2589, opened 2024-01-04
Non-GET http2 requests without a body hang indefninitely


ktaekwon000:
## Bug Description

When using fetch to make non-GET/HEAD http2 requests, the function call hangs indefinitely. Passing in an empty string in place of the body fixes this issue.

## Reproducible By

The following js script triggers the issue on undici v6.2.1
```javascript
'use strict'

const { fetch, Client } = require('undici')

async function main () {
    const client = new Client("https://google.com", { allowH2: true });
    const res = await fetch("https://google.com", {
        dispatcher: client,
        baseURL: 'https://google.com',
        url: 'https://google.com/',
        method: "POST",
    })

    const data = await res.text()
    console.log('response received', res.status)
    console.log('headers', res.headers)
    console.log('data', data)
}
main()
```

## Expected Behavior

The function call should send the request with an empty body.

## Environment

macOS 14.0, node v20.8.1, undici v6.2.1

### Additional context

I believe the issue is closely related to #2258 , but while this PR had only fixed GET/HEAD requests, this actually affects all kinds of requests.

metcoder95:
Hey! Thanks for the report

I believe this should be scope to validate wether or not there's a body at all being sent.

Would you like to submit a PR?

ronag:
> I believe this should be scope to validate wether or not there's a body at all being sent.

I'm confused. It should be possible to send a fetch without body? Also this sounds like we are missing something in the spec language? @KhafraDev

KhafraDev:
this isn't an issue with fetch afaict

timursevimli:
While I was trying to investigate this problem, I discovered a completely different problem.

When we pass body the following object appears:
```js
{
   finalBody: {
     stream: ReadableStream { locked: false, state: 'readable', supportsBYOB: true },
     source: 'helloWorld',
     length: 10
   }
}
```
(For this output I used output a variable called finalBody in undici/lib/fetch/request.js:505)

Well, if we pass an object to body, then we get the following output:
```js
{
   finalBody: {
     stream: ReadableStream { locked: false, state: 'readable', supportsBYOB: true },
     source: '[object Object]',
     length: 15
   }
}
```
In this case, regardless of the size of our object, length will always be 15 and source will be incorrect.

taekwon-datature:
[...]

PR #2601, opened 2024-01-07
fix: h2 hang issue with empty body

timursevimli:
#2589 

## Status

- [x] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1769,7 +1769,7 @@ function writeH2 (client, session, request) {
 
   session.ref()
 
-  const shouldEndStream = method === 'GET' || method === 'HEAD'
+  const shouldEndStream = method === 'GET' || method === 'HEAD' || body === null
   if (expectContinue) {
     headers[HTTP2_HEADER_EXPECT] = '100-continue'
     stream = session.request(headers, { endStream: shouldEndStream, signal })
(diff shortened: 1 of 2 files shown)


KhafraDev (changes requested):
Can you add a test?

lib/client.js
```diff
@@ -1769,7 +1769,7 @@ function writeH2 (client, session, request) {
 
   session.ref()
 
-  const shouldEndStream = method === 'GET' || method === 'HEAD'
+  const shouldEndStream = method === 'GET' || method === 'HEAD' || body === null
```
metcoder95, line 1772:
```suggestion
  const shouldEndStream = method === 'GET' || method === 'HEAD' || body == null
```

To also handle `undefined`

mcollina: A test for that would be needed as well

tsctx: Here, if body is undefined, it will be null, so this change does not seem necessary.
https://github.com/nodejs/undici/blob/03f419f3ccee0c7779cf0dbc7b8d3a16e9eee0a2/lib/core/request.js#L95-L96

metcoder95: Good catch, although I'll still add the test for both cases 👍

KhafraDev (approved):

mcollina (approved):
lgtm

mcollina:
@metcoder95 ptal

metcoder95 (approved):

Issue #2520, opened 2023-12-13
fetch url now show the final url after all redirect


fancy45daddy:
in nodejs v21.2.0
await globalThis.fetch('https://bit.ly/48b82fP').then(_ => _.url) just show the first redirect url:
https://l.facebook.com/l.php?u=https://f.theloadedbaze.com/%3Ffbclid%3DIwAR09e4iW8r6DAGuJbtEfepgIwLKm1zsIcavDGWT7gbFqyX4E56N7fL9pQNQ&h=AT1dnnHhh3K5JfViDpurjF0I833b7Fs5ukXQIDMNju_dSkagEbhuGB1zjjrFKuOZF2k6IU9MSTTwz0FlSoW761RP7xoRXBay0mUmD5bJYggSXdVlLwb4vOVlNXcM3oRnYMlwr_sh4RQ

But in node-fetch:
import fetch from 'node-fetch'
await globalThis.fetch('https://bit.ly/48b82fP').then(_ => _.url) show the final redirect url:
https://www.facebook.com/flx/warn/?u=https%3A%2F%2Ff.theloadedbaze.com%2F%3Ffbclid%3DIwAR09e4iW8r6DAGuJbtEfepgIwLKm1zsIcavDGWT7gbFqyX4E56N7fL9pQNQ&h=AT3Lwqzvo-lauSwVjMfI9Q5fw0WWGVojnuGIDzB_6p8f_8TP-A4MTOdU8LcjH_2OZ6o8auwyrtzF_XZOwH9EQ9ZBb5KlZz4D_x6DoR58t-FSfF3bNDoXfC8iqpPm-mAp-8aqMYwuFOQ

I want to know how to get the final url after all redirect in undici fetch link that in node-fetch?

KhafraDev:
fetch will return the final redirect's url. If you can provide a minimal reproduction that doesn't use third party services I can look into it.

fancy45daddy:
OK, the minimal test is I want to get the final URL after all the redirect https://bit.ly/48b82fP
[...]

PR #2608, opened 2024-01-11
feat: Implement EventSource

Uzlopak:
This is my attempt to implement eventsource into undici and thus into nodejs. 

I am currently working on it, of course. EventSourceStream is not working perfect, so that would be my priority. After that I would focus on spec compliance.

But my biggest issue still needs clarification:

EventSource in a Browser UserAgent expects that you can use "withCredentials" option to send cookies with the request. So this is kind of limiting as I assume that node native fetch is not storing any cookies. So maybe we should deviate from the spec and allow maybe like npm:eventsource to pass headers via the options parameter. Thus you could also use the authorization header to connect to a EventSource server.change

@KhafraDev 
I am looking forward for some hints regarding expected coding style in undici. E.g. should I replace the private properties to symbols?

Should I integrate wpt tests? Tbh. they are not working out of the box, and needs python. 

Looking forward for your feedback.

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready


KhafraDev (commented):
private properties are fine and preferred :)

regarding the WPTs we have our own test runner, but it's not as easy to use (you'd have to implement the routes for yourself, [for example](https://github.com/nodejs/undici/commit/eead2b8209519c4ad0ad1668f1e8181a46914540#diff-9db6492b1ab92ac1a53d7a45081cd885943b1ff413f5568d3e1eded6f03867cb)). There is also a subsequent [section](https://github.com/nodejs/undici/blob/main/CONTRIBUTING.md#update-wpts) of the contributing guide that may help. I'd be willing to do this for you if you'd prefer, it'd save you a lot of work 😅.

Uzlopak:
@KhafraDev 

I would be very grateful if you would add the wpt tests :). I invited you to my fork. So should be easier for you to directly pushing to it. 

It feels strange, that some attributes are private and some are symbols. Also comparing to other parts of the codebase looks not very tidy. 

Anyway: Looking forward for every nitpick.

KhafraDev:
added the tests, you can run them by doing `node test/wpt/start-eventsource.mjs`

KhafraDev:
I can start reviewing now if you'd like, but currently the WPTs are failing because the WPTs use relative urls but you're not passing a base URL, for example:

https://github.com/nodejs/undici/blob/d3d825e21ddea12b893bbf7ba0a4c1e5899c85e4/lib/websocket/websocket.js#L61-L67

Uzlopak:
@KhafraDev 

Awesome! Thanks for the snippet with getGlobalOrigin. Now 3 only wpt tests fail, which are related to the EventSourceStream, which I already said is not up to standard.

Please review the code, so that I can tackle them asap you write that you finished the review :). 

Please be nitpicky as possible :)

Thank you!

lib/eventsource/eventsource.js
```diff
@@ -0,0 +1,326 @@
...
+    const baseURL = getGlobalOrigin()
+
+    // 2. Let urlRecord be the result of applying the URL parser to url with baseURL.
+    let urlRecord
+
+    try {
+      urlRecord = new URL(url, baseURL)
+    } catch (e) {
+      // 3. If urlRecord is failure, then throw a "SyntaxError" DOMException.
+      throw new DOMException(e, 'SyntaxError')
```
KhafraDev, line 56: In browsers this technically can't fail here so we can use any error we'd like. We should probably match Deno here, do you know what they're doing when URL parsing fails?

lib/eventsource/eventsource.js
```diff
@@ -0,0 +1,326 @@
...
+      reconnectionTime: defaultReconnectionTime
+    }
+
+    this[kEvents] = {
+      message: null,
+      error: null,
+      open: null
+    }
+
+    this[kState].origin = new URL(this.#url).origin
```
KhafraDev, line 74:
```suggestion
    this[kState].origin = urlRecord.origin
```

lib/eventsource/eventsource.js
```diff
@@ -0,0 +1,326 @@
...
+  #connection = null
+  #reconnectionTimer = null
+  #controller = new AbortController()
+
+  /**
+   * Creates a new EventSource object.
+   * @param {string} url
+   * @param {EventSourceInit} [eventSourceInitDict]
+   */
+  constructor (url, eventSourceInitDict) {
```
KhafraDev, line 41: eventSourceInitDict needs its own webidl conversions, here's how it's done for MessageEvent: https://github.com/nodejs/undici/blob/d3d825e21ddea12b893bbf7ba0a4c1e5899c85e4/lib/websocket/events.js#L219

lib/eventsource/eventsource.js
```diff
@@ -0,0 +1,326 @@
...
+      // https://html.spec.whatwg.org/multipage/server-sent-events.html#server-sent-events-intro
+      switch (this.#connection.status) {
+        // Redirecting status codes
+        case 301: // 301 Moved Permanently
+        case 302: // 302 Found
+        case 307: // 307 Temporary Redirect
+        case 308: // 308 Permanent Redirect
+          if (!this.#connection.headers.has('Location')) {
+            this.close()
+            this.dispatchEvent(new Event('error'))
```
KhafraDev, line 150: is this meant to use ErrorEvent? If so, we have an implementation with the other WebSocket events

lib/eventsource/eventsource.js
```diff
@@ -0,0 +1,326 @@
...
+            return
+          }
+          this.#url = new URL(this.#connection.headers.get('Location'), new URL(this.#url).origin).href
+          this[kState].origin = new URL(this.#url).origin
```
KhafraDev, line 154: parsing `this.#url` twice here

lib/eventsource/eventsource.js
```diff
@@ -0,0 +1,326 @@
...
+      }
+
+      const self = this
+
+      pipeline(this.#connection.body,
+        new EventSourceStream({
+          eventSourceState: this[kState],
+          push: function push (chunk) {
+            self.dispatchEvent(chunk)
+          }
```
KhafraDev, line 189:
```suggestion
          push: (chunk) => {
            this.dispatchEvent(chunk)
          }
```

can get rid of self afterwards

lib/eventsource/eventsource.js
```diff
@@ -0,0 +1,326 @@
...
+      }
+      this.dispatchEvent(new Event('error'))
+
+      // Always set to CONNECTING as the readyState could be OPEN
+      this.#readyState = CONNECTING
+      this.#connection = null
+
+      this.#reconnectionTimer = setTimeout(() => {
+        this.#connect()
+      }, this[kState].reconnectionTime)
```
KhafraDev, line 212: unref?

KhafraDev (changes requested):

lib/eventsource/eventsource.js
KhafraDev: `url` should use `webidl.converters.USVString` as well

lib/eventsource/eventsource.js
Uzlopak: If I remember correctly SyntaxError is expected from a wpt test.

Uzlopak:
I will continue tonight to work on this. It feels much cleaner to implement it in undici than in node core itself. 

Stay tuned ;)

lib/eventsource/eventsource.js
KhafraDev: @Uzlopak this isn't resolved afaict

lib/eventsource/eventsource.js
```diff
@@ -0,0 +1,324 @@
...
+    }
+
+    if (this.#lastEventId) {
+      options.headers['Last-Event-ID'] = this.#lastEventId
+    }
+
+    options.credentials = this.#withCredentials ? 'include' : 'omit'
+
+    try {
+      this.#connection = await fetch(this.#url, options)
```
KhafraDev, line 140: this should use undici's fetch, not the global one

KhafraDev (changes requested):

lib/eventsource/eventsource.js
KhafraDev: I can't see how this would throw an error in browsers 🤔

Uzlopak: e.g. new URL('https://_') could result in an error of URL constructor.

Uzlopak: also i copied it from the websocket implementation

KhafraDev: yeah, wasn't blocking, was mostly curious - Deno is throwing a SyntaxError DOMException too so it's good 👍 https://github.com/denoland/deno/blob/f45ceb23206b4aaeca123c06bd3d59770f728468/ext/fetch/27_eventsource.js#L178

lib/eventsource/eventsource.js
Uzlopak: I didnt resolve it yet 🤔

lib/eventsource/eventsource-stream.js
metcoder95, line 1: Let's not forget to mark it as experimental 👍

metcoder95 (commented):
So far LGTM, nice work @Uzlopak 🚀 
I'd like to go on a second review later this weekend, to also have the time to read the spec a bit more; if that's ok?

Uzlopak:
I have following tasks to do:

- [ ] Make it more like the fetch implementation where every step is commented implemented according to the spec
- [ ] Make the parser feature complete
- [ ] Check if the state parameter, shared between the EventSource and EventSourceStream is complete, or if it should be modified
- [ ] 100% test coverage, if possible

But you are invited to review and nitpick as much as possible as early as possible ;)

Uzlopak:
Ok, I figured now out how to it like you did it for fetch. Obviously I have to implement it differently. My local code is not working, so I wont push. Tomorrow/Sunday I willl take care of it.

lib/eventsource/events.js
```diff
@@ -0,0 +1,27 @@
+'use strict'
+
+const { webidl } = require('../fetch/webidl')
+const {
+  MessageEvent,
+  ErrorEvent,
+  eventInit
+} = require('../websocket/events')
+
+class OpenEvent extends Event {
```
Uzlopak, line 10: OpenEvent could be moved to websocket/events.js as Websocket is also having an OpenEvent. wdyt?

lib/eventsource/index.js
```diff
@@ -0,0 +1,408 @@
...
+      experimentalWarned = true
+      process.emitWarning('EventSource is experimental, expect them to change at any time.', {
+        code: 'UNDICI-ES'
+      })
+    }
+
+    url = webidl.converters.USVString(url)
+    eventSourceInitDict = webidl.converters.EventSourceInitDict(eventSourceInitDict)
+
+    // 2. Let settings be ev's relevant settings object.
```
Uzlopak, line 70: actually i dont know what the spec means with relevant settings object. Is this right?

lib/eventsource/index.js
```diff
@@ -0,0 +1,408 @@
...
+    const initRequest = {
+      redirect: 'follow',
+      keepalive: true,
+      // @see https://html.spec.whatwg.org/multipage/urls-and-fetching.html#cors-settings-attributes
+      mode: 'cors',
+      credentials: corsAttributeState === 'anonymous'
+        ? 'same-origin'
+        : 'omit',
+      signal: this.#controller.signal,
+      referrer: 'no-referrer',
```
Uzlopak, line 116: This referrer settings were annoying. I wanted to use first makeRequest, but makeRequest does not work with AbortController. So I use new Request(). See the changes in the fetch code.

lib/eventsource/index.js
```diff
@@ -0,0 +1,408 @@
...
+    // 11. Set request's cache mode to "no-store".
+    initRequest.cache = 'no-store'
+
+    // 12. Set request's initiator type to "other".
+    initRequest.initiator = 'other'
+
+    // 13. Set ev's request to request.
+    this.#request = new Request(this.#url, initRequest)
+
+    this.#request.headersList = initRequest.headers
```
Uzlopak, line 135: I needed to assign headersList explicitly to initRequest.headers, because it would throw in https://github.com/nodejs/undici/blob/3372aed902951f135d710d0fb430f39b8571e547/lib/fetch/index.js#L472

lib/eventsource/index.js
```diff
@@ -0,0 +1,408 @@
...
+    this.#request.urlList = this.#request.urlList || [new URL(this.#url)]
```
Uzlopak, line 136: very strange. Why does it not assign it by default?

lib/fetch/index.js
```diff
@@ -555,7 +555,7 @@ async function mainFetch (fetchParams, recursive = false) {
 
   // 8. If request’s referrer is not "no-referrer", then set request’s
   // referrer to the result of invoking determine request’s referrer.
-  if (request.referrer !== 'no-referrer') {
+  if (request[kState].referrer !== 'no-referrer') {
```
Uzlopak, line 558: Bug?

This is maybe not correct, but the issue is, that we have a custom getter for request.referrer. So if request.referrer is 'no-referrer' than request.referrer returns empty string (''). 

https://github.com/nodejs/undici/blob/3372aed902951f135d710d0fb430f39b8571e547/lib/fetch/request.js#L609

lib/fetch/request.js
```diff
@@ -220,7 +220,7 @@ class Request {
       const referrer = init.referrer
 
       // 2. If referrer is the empty string, then set request’s referrer to "no-referrer".
-      if (referrer === '') {
+      if (referrer === '' || referrer === 'no-referrer') {
```
Uzlopak, line 223: Bug?

I think the spec is here wrong or not clear.

If referrer is empty string, set request.referrer to no-referrer. But what if it is 'no-referrer', because we explicitly set it to no-referrer in makeRequest? Here it will be set to no-referrer, and then go to the else branch and try to parse the 'no-referrer' like an URL.
https://github.com/nodejs/undici/blob/3372aed902951f135d710d0fb430f39b8571e547/lib/fetch/request.js#L231

Obviously this will throw.

lib/fetch/request.js
```diff
@@ -823,7 +823,7 @@ function makeRequest (init) {
       ? new HeadersList(init.headersList)
       : new HeadersList()
   }
-  request.url = request.urlList[0]
+  request.url = request.urlList[request.urlList.length - 1]
```
Uzlopak, line 826: Bug?

I assume that urlList lenght - 1 has to be used?

test/wpt/runner/worker.mjs
```diff
@@ -113,7 +119,7 @@
 `)
 
 if (meta.title) {
-  runInThisContext(`globalThis.META_TITLE = "${meta.title}"`)
+  runInThisContext(`globalThis.META_TITLE = "${meta.title.replace(/"/g, '\\"')}"`)
```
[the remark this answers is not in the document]
Uzlopak: @KhafraDev
Should it be?
```suggestion
  runInThisContext(`globalThis.META_TITLE = "${meta.title.replace(/\"/g, '\\"')}"`)
```

lib/eventsource/index.js
```diff
@@ -0,0 +1,408 @@
...
+
+      // 3. Otherwise, [...] if res's
+      // `Content-Type` is not `text/event-stream`, then fail the
+      // connection.
+
+      const contentType = response.headersList.get('content-type', true)
+      const mimeType = contentType !== null ? parseMIMEType(contentType) : 'failure'
+      if (mimeType === 'failure' || mimeType.essence !== 'text/event-stream') {
+        this.close()
+        this.dispatchEvent(new ErrorEvent('error', { error: 'Invalid content-type' }))
```
Uzlopak, line 250: Note to myself: should be message and not error attribute

lib/fetch/index.js
```diff
@@ -1361,6 +1361,7 @@ function httpRedirectFetch (fetchParams, response) {
 
   // 18. Append locationURL to request’s URL list.
   request.urlList.push(locationURL)
+  // TODO: not enough?
```
Uzlopak, line 1364: Bug?

After redirecting, you dont have the url of the request object. Does it have to be in the response? Does reponse even have a url field?

This does not change request.url, which is probably connected with bug report #2520

request.url is using a getter:

https://github.com/nodejs/undici/blob/3372aed902951f135d710d0fb430f39b8571e547/lib/fetch/request.js#L577

So either urlList gets a Proxy to set the corresponding request[kState].url or url is retrieved dynamically by taking the last element of request[kState].urlList

Uzlopak:
@KhafraDev 
@metcoder95 

I created various comments. The implementation is now more following the spec and uses the in the spec implied fetching algorith.

Looking forward to your feedback.

lib/eventsource/events.js
KhafraDev: what's the difference between this one and WebSocket's?

lib/eventsource/index.js
KhafraDev: yes it is, it's basically just default options but worded in a very confusing way

lib/eventsource/index.js
KhafraDev: `makeRequest` returns an internal request, not a Request object - ie. `new Request(...)[kState]`.

lib/eventsource/index.js
KhafraDev: same reason as above

lib/eventsource/events.js
Uzlopak: Nothing they are basically just Events. I just made a dedicated class for it.

lib/fetch/index.js
KhafraDev: this should be reverted

lib/eventsource/index.js
Uzlopak: Should I use makeRequest or new Request?

Uzlopak: does fetchParam expect an internal request?

lib/fetch/request.js
KhafraDev: I think it's more likely that the EventSource spec is wrong here, and rather than making the change in fetch, you should instead just set the referrer to an empty string, and let fetch handle it.

lib/fetch/request.js
KhafraDev: request's URL is the first one, not the last

test/wpt/runner/worker.mjs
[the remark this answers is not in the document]
KhafraDev: the regex I added works

lib/eventsource/index.js
KhafraDev: makeRequest, all of the internals require it instead of a Request

lib/eventsource/events.js
KhafraDev: I haven't taken a look at the spec yet, but there's a difference between "an event named 'open'" and an OpenEvent. If WebSocket doesn't have its own OpenEvent, it's likely just `new Event('open')` that's required

lib/eventsource/index.js
KhafraDev: https://fetch.spec.whatwg.org/#concept-request

lib/fetch/index.js
Uzlopak:
```suggestion
```

Uzlopak:
The pipeline is broken because of latest node 21. 
#2615

lib/eventsource/index.js
```diff
@@ -0,0 +1,448 @@
...
+    this.#settings = {
+      lastEventId: '',
+      reconnectionTime: defaultReconnectionTime
+    }
+
+    // 1. Let baseURL be this's relevant settings object's API base URL.
+    const baseURL = getGlobalOrigin()
+
+    // 2. Let urlRecord be the result of applying the URL parser to url with baseURL.
+    let urlRecord
```
KhafraDev, line 89: step 2 is being done where step 3 is. I don't think you need to do anything special for step 3, so just putting the step 2 comment above it will be fine

lib/eventsource/index.js
```diff
@@ -0,0 +1,448 @@
...
+        : 'omit',
+      referrer: 'no-referrer',
+      referrerPolicy: 'no-referrer'
+    }
+
+    // 9. Set request's client to settings.
+
+    // 10. User agents may set (`Accept`, `text/event-stream`) in request's header list.
+    initRequest.headers = new HeadersList()
+    initRequest.headers.set('accept', 'text/event-stream', true)
```
KhafraDev, line 131: This isn't correct, the internal response has a `headersList` property. You should add this to initRequest:

```js
    const initRequest = {
      // ...
      headersList: [['accept', 'text/event-stream']] 
    }
```

lib/eventsource/index.js
```diff
@@ -0,0 +1,448 @@
...
+
+    // 12. Set request's initiator type to "other".
+    initRequest.initiator = 'other'
+
+    initRequest.urlList = [new URL(this.#url)]
+
+    // 13. Set ev's request to request.
+    this.#request = makeRequest(initRequest)
+
+    this.#request.headersList = initRequest.headers
```
KhafraDev, line 144: I'm kinda confused here 😅. Once the change I recommended above it made you can remove this.

lib/eventsource/index.js
```diff
@@ -0,0 +1,448 @@
...
+      if (isNetworkError(response)) {
+        // 1. When a user agent is to fail the connection, the user agent
+        // must queue a task which, if the readyState attribute is set to a
+        // value other than CLOSED, sets the readyState attribute to CLOSED
+        // and fires an event named error at the EventSource object. Once the
+        // user agent has failed the connection, it does not attempt to
+        // reconnect.
+        if (response.aborted) {
+          this.close()
+          this.dispatchEvent(new ErrorEvent('error', { error: response.error }))
```
KhafraDev, line 212: I think I mentioned it somewhere but is this meant to be an ErrorEvent, or a `new Event('error')`? The spec text makes it seem like the latter.

lib/eventsource/index.js
```diff
@@ -0,0 +1,448 @@
...
+      // `Content-Type` is not `text/event-stream`, then fail the
+      // connection.
+
+      const contentType = response.headersList.get('content-type', true)
+      const mimeType = contentType !== null ? parseMIMEType(contentType) : 'failure'
+      if (mimeType === 'failure' || mimeType.essence !== 'text/event-stream') {
+        this.close()
+        this.dispatchEvent(new ErrorEvent('error', { message: 'Invalid content-type' }))
+        return
+      }
```
KhafraDev, line 244: It would probably be easier to read if the ifs were merged and the step not split in two

lib/eventsource/index.js
```diff
@@ -0,0 +1,448 @@
...
+
+      // 3. Otherwise, [...] if res's
+      // `Content-Type` is not `text/event-stream`, then fail the
+      // connection.
+
+      const contentType = response.headersList.get('content-type', true)
+      const mimeType = contentType !== null ? parseMIMEType(contentType) : 'failure'
+      if (mimeType === 'failure' || mimeType.essence !== 'text/event-stream') {
+        this.close()
+        this.dispatchEvent(new ErrorEvent('error', { message: 'Invalid content-type' }))
```
KhafraDev, line 242: ditto regarding ErrorEvent vs. event named error

lib/eventsource/index.js
```diff
@@ -0,0 +1,448 @@
...
+      // 4. Otherwise, announce the connection and interpret res's body
+      // line by line.
+
+      // When a user agent is to announce the connection, the user agent
+      // must queue a task which, if the readyState attribute is set to a
+      // value other than CLOSED, sets the readyState attribute to OPEN
+      // and fires an event named open at the EventSource object.
+      // @see https://html.spec.whatwg.org/multipage/server-sent-events.html#sse-processing-model
+      this.#readyState = OPEN
+      this.dispatchEvent(new Event('open', {}))
```
KhafraDev, line 254:
```suggestion
      this.dispatchEvent(new Event('open'))
```

lib/eventsource/index.js
```diff
@@ -0,0 +1,448 @@
...
+          ))
+        }
+      })
+
+      pipeline(response.body.stream,
+        eventSourceStream,
+        (error) => {
+          if (
+            error &&
+            error.aborted === false
```
KhafraDev, line 271:
```suggestion
            error?.aborted === false
```

lib/eventsource/index.js
```diff
@@ -0,0 +1,448 @@
...
+        })
+    }
+
+    fetchParam.timingInfo = createOpaqueTimingInfo({})
+
+    fetchParam.controller = new Fetch(getGlobalDispatcher())
+
+    this.#controller = fetchParam.controller
+
+    await mainFetch(fetchParam)
```
KhafraDev, line 285: Is there a reason that you using mainFetch over `fetching`? This doesn't seem correct to me.

KhafraDev: fetching will create the fetchParams for you and return the controller. No need to do it yourself!

lib/eventsource/index.js
```diff
@@ -0,0 +1,448 @@
...
+    if (this.#readyState === CLOSED) return
+
+    //   2. Set the readyState attribute to CONNECTING.
+    this.#readyState = CONNECTING
+
+    //   3. Fire an event named error at the EventSource object.
+    this.dispatchEvent(new ErrorEvent('error', { message: 'Reconnecting' }))
+
+    // 2. Wait a delay equal to the reconnection time of the event source.
+    await setTimeout(this.#settings.reconnectionTime)
```
KhafraDev, line 310:
```suggestion
    await setTimeout(this.#settings.reconnectionTime, { ref: false })
```

lib/eventsource/index.js
```diff
@@ -0,0 +1,448 @@
...
+  }
+
+  get onopen () {
+    webidl.brandCheck(this, EventSource)
+
+    return this.#events.open
+  }
+
+  set onopen (fn) {
+    webidl.brandCheck(this, EventSource)
```
KhafraDev, line 357: no need, the private properties act as a brand check for us

lib/eventsource/index.js
```diff
@@ -0,0 +1,448 @@
...
+    clearTimeout(this.#settings.reconnectionTimer)
+    this.#controller.abort()
+
+    if (this.#request) {
+      this.#request = null
+    }
+  }
+
+  get onopen () {
+    webidl.brandCheck(this, EventSource)
```
KhafraDev, line 351: no need, the private properties act as a brand check for us

lib/eventsource/index.js
```diff
@@ -0,0 +1,448 @@
...
+    //   4. Fetch request and process the response obtained in this fashion, if any, as described earlier in this section.
+    this.#connect()
+  }
+
+  /**
+   * Closes the connection, if any, and sets the readyState attribute to
+   * CLOSED.
+   */
+  close () {
+    webidl.brandCheck(this, EventSource)
```
KhafraDev, line 338: no need, the private properties act as a brand check for us

lib/eventsource/index.js
```diff
@@ -0,0 +1,448 @@
...
+    if (typeof fn === 'function') {
+      this.#events.open = fn
+      this.addEventListener('open', fn)
+    } else {
+      this.#events.open = null
+    }
+  }
+
+  get onmessage () {
+    webidl.brandCheck(this, EventSource)
```
KhafraDev, line 372: ditto regarding brand check

lib/eventsource/index.js
```diff
@@ -0,0 +1,448 @@
...
+
+    return this.#events.message
+  }
+
+  set onmessage (fn) {
+    webidl.brandCheck(this, EventSource)
```
KhafraDev, line 378: ditto regarding brand check

lib/eventsource/index.js
```diff
@@ -0,0 +1,448 @@
...
+    if (typeof fn === 'function') {
+      this.#events.message = fn
+      this.addEventListener('message', fn)
+    } else {
+      this.#events.message = null
+    }
+  }
+
+  get onerror () {
+    webidl.brandCheck(this, EventSource)
```
KhafraDev, line 393: ditto regarding brand check

lib/eventsource/index.js
```diff
@@ -0,0 +1,448 @@
...
+
+    return this.#events.error
+  }
+
+  set onerror (fn) {
+    webidl.brandCheck(this, EventSource)
```
KhafraDev, line 399: ditto regarding brand check

lib/eventsource/index.js
```diff
@@ -0,0 +1,448 @@
...
+    configurable: false,
+    enumerable: true,
+    value: CLOSED,
+    writable: false
+  }
+})
+
+EventSource.prototype.CONNECTING = CONNECTING
+EventSource.prototype.OPEN = OPEN
+EventSource.prototype.CLOSED = CLOSED
```
KhafraDev, line 440: I believe these similarly have to be defined with Object.defineProperties making them not configurable & writable

lib/fetch/response.js
```diff
@@ -364,6 +364,19 @@ function makeNetworkError (reason) {
...
+// @see https://fetch.spec.whatwg.org/#concept-network-error
+function isNetworkError (response) {
+  return (
+    // A network error is a response whose type is "error",
+    response.type === 'error' &&
+    //  status is 0
+    response.status === 0 &&
+    'error' in response && (
+      'aborted' in response ? typeof response.aborted === 'boolean' : true
+    )
```
KhafraDev, line 376: what's the purpose of this? type == 'error' and a status of 0 should be enough here.

KhafraDev (changes requested):

lib/eventsource/index.js
Uzlopak: This was regarding OpenEvent. I removed that class. ErrorEvent was already existing in websocket. I think i forgot to remove the types for OpenEvent.

lib/eventsource/index.js
Uzlopak: I had the feeling that it has less complexity than putting everything into one block. Will investigate into this

lib/eventsource/index.js
Uzlopak: I will check it. The headers were actually really problematic.

lib/eventsource/index.js
Uzlopak: I doubt it. But i will investigate again.

lib/eventsource/index.js
Uzlopak: I did it analogue to websocket

https://github.com/nodejs/undici/blob/3372aed902951f135d710d0fb430f39b8571e547/lib/websocket/websocket.js#L400

Uzlopak:
@KhafraDev 

Thank you for the review so far. I will resolve them in about 18 hours.

Tbh I feel much better with this implementation as it feels much cleaner.
Regarding fetching, vs. fetch vs. mainfetch... Tbh i spent multiple hours to get it work. i still believe there is a bug regarding the headers and the headersList.
In the end i did it the deno way and used mainFetch and suddenly it worked everything properly. 
If then I would actually switch to fetch, as the spec says Fetch, but as i said there are issues. 

But again, I feel much more confident regarding the implementation than the original code which i wanted to add to node core. lol.

I will focus on the EventSourceStream, as it is not working properly. Till now i focussed only on EventSource class. If the Stream works properly i would focus again on correctness.

KhafraDev:
I'll take a look at using fetching tomorrow, mainFetch is more 'internal' and I've never seen it used in another spec (websocket & cachestorage both use fetching). Implementation is definitely close to being complete :)

KhafraDev:
@Uzlopak I changed the implementation to use fetching, but I didn't want to infringe on your work so I kept changes minimal. The same WPTs pass/fail so there should be no noticeable difference. I took a look at the spec and the spec did mention to use [fetching](https://fetch.spec.whatwg.org/#fetching), so that's good. Btw, Deno doesn't follow the spec as closely as we do, and while their implementation is a good reference point, I wouldn't assume something is incorrect with your work if it doesn't match verbatim 😉.

Uzlopak:
@KhafraDev 

Can you clarify regarding brandincheck?
https://github.com/nodejs/undici/pull/2608#discussion_r1452924525

I orientated myself on the websocket implementation, which also uses brandingcheck despite using getters and setters.

KhafraDev:
It's not needed when using private properties -

```js
class A {
  #priv

  getPriv () { return this.#priv }
}

A.prototype.getPriv() // throws TypeError
```

lib/eventsource/index.js
Uzlopak: This is not a private property?! Are you sure?

lib/eventsource/index.js
Uzlopak: I could implement it like this:
Do you like it better?

```js
      // 3. Otherwise, if res's status is not 200, or if res's `Content-Type`
      // is not `text/event-stream`, then fail the connection.
      const contentType = response.headersList.get('content-type', true)
      const mimeType = contentType !== null ? parseMIMEType(contentType) : 'failure'
      const contentTypeValid = mimeType !== 'failure' && mimeType.essence === 'text/event-stream'
      if (
        response.status !== 200 ||
        contentTypeValid === false
      ) {
        let message

        if (response.status !== 200) {
          message = 'No content'
        } else if (contentTypeValid === false) {
          message = `Invalid content-type: ${contentType}`
        } else {
          message = `Invalid status code: ${response.status}`
        }

        this.close()
        this.dispatchEvent(new ErrorEvent('error', { message }))
        return
      }

```

mcollina (approved):
lgtm

KhafraDev:
There appears to be quite a few parsing bugs which I noticed after implementing one of the routes for the tests. I fixed 2 bugs unrelated to the parsing, which were quite annoying tracking down.

Uzlopak:
@KhafraDev 
@metcoder95 

1. I assume that this PR still needs few days of development to be "perfect".

1. The code coverage is now near 100%. So changing code will hopefully immediatly show regressions. 

1. Now the wpt tests show 1 failed. But this is a lie ;).

First of all, `tests/eventsource/eventsource-onmessage-trusted.any.js` can not be done it seems. This doesnt seem to work
```js
event.options[Symbol.for('kTrustEvent')] = true
```

This has two possible reasons:
1. maybe we cant retrieve kTrustEvent from node code (unlikely)
2. the webidl converter for MessageEventInit is filtering the symbol out. 

Anyhow, this is something which needs some deep digging. 

1. Some wpt-tests are timing out. Have to see why and how I can resolve them.

1. Also your commit where you basically added mocking logic for message.py helped me to understand the wpt tests better. I was actually wondering why the tests were not failing. So thats why some tests are copies of the wpt test suite. Anyhow there are some more routes, which need to be mocked.

1. I am also looking into other eventsource implementations for useful unit tests. 

1. Something which is not clear to me rn, is the question if an event is dispatched if an event contains only a retry field. 

1. There is  one obvious performance bottleneck, which can be resolved when the above mentioned issues are resolved. Currently in the _transform method of EventSourceStream I do alot of Buffer instantiations. This can be significantly optimized. But I am focussing on getting all the other unclear aspects clarified. There is no gain in optimizing the performance, if the code is still not working according to the spec.

lib/eventsource/eventsource.js
```diff
@@ -0,0 +1,493 @@
...
+
+    // 12. Set request's initiator type to "other".
+    initRequest.initiator = 'other'
+
+    initRequest.urlList = [new URL(this.#url)]
+
+    // 13. Set ev's request to request.
+    this.#request = makeRequest(initRequest)
+
+    this.#request.headersList = initRequest.headers
```
Uzlopak, line 197: @KhafraDev 
You remarked that I should remove this. Please test locally to remove this line and then run the example in examples/eventsource.js 

You will see that the connection will break, because the headers are not set properly. 
Can you please investigate what the issue is? I dont think, that it is something I introduced. 

Thank you :)

lib/eventsource/index.js
Uzlopak: Please see my new remark below

lib/eventsource/eventsource.js
KhafraDev: Changed it, my suggestion was a little off

KhafraDev:
There's nothing we can do about isTrusted, the test can marked as failing in its status file.

lib/eventsource/eventsource.js
Uzlopak: great!

mcollina:
Can you add the docs to the website sidebar too?

Uzlopak:
@mcollina 
Added it to the sidebar.

@KhafraDev 
The wpt tests pass now. 

**Please do not merge yet!**

I still want to investigate some aspects. I wrote an simple sse server and test it against Firefox and Chrome and see how it behaves in some edge cases, to understand

lib/eventsource/eventsource.js
```diff
@@ -0,0 +1,490 @@
...
+
+  /**
+   * Returns a boolean indicating whether the EventSource object was
+   * instantiated with CORS credentials set (true), or not (false, the default).
+   */
+  get withCredentials () {
+    return this.#withCredentials
+  }
+
+  async #connect () {
```
KhafraDev, line 223: this no longer needs to be async after removing mainFetch

lib/fetch/index.js
```diff
@@ -2279,5 +2282,6 @@ module.exports = {
   fetch,
   Fetch,
   fetching,
+  mainFetch,
```
KhafraDev, line 2285: this is no longer required

lib/websocket/events.js
```diff
@@ -297,6 +297,7 @@ webidl.converters.ErrorEventInit = webidl.dictionaryConverter([
 ])
 
 module.exports = {
+  eventInit,
```
KhafraDev, line 300: I don't think this is needed either

package.json
```diff
@@ -78,6 +78,7 @@
     "test": "node scripts/generate-pem && npm run test:tap && npm run test:node-fetch && npm run test:fetch && npm run test:cookies && npm run test:wpt && npm run test:websocket && npm run test:jest && npm run test:typescript && npm run test:node-test",
     "test:cookies": "borp --coverage -p \"test/cookie/*.js\"",
     "test:node-fetch": "mocha --exit test/node-fetch",
+    "test:eventsource": "npm run build:node && borp --expose-gc --coverage -p \"test/eventsource/*.js\"",
```
KhafraDev, line 81: this should be added to `test` so that it runs in the CI

test/wpt/status/eventsource.status.json
```diff
@@ -0,0 +1,10 @@
+{
+  "eventsource-onmessage-trusted.any.js": {
+    "note": "An Event created in userspace can not be set to trusted.",
+    "skip": true
+  },
+  "format-field-event-empty.any.js": {
+    "note": "To be investigated.",
```
KhafraDev, line 7: If you investigated, do you know why this test is failing?

KhafraDev (changes requested):
just some things to clean up

test/wpt/status/eventsource.status.json
Uzlopak: Forgot to remove that case from the exception list.

Uzlopak: It is now also tested

Uzlopak:
@KhafraDev 
There are still some older remarks, which i did not resolve and maybe need again review by you and give feedback to the feedback from me. 

Thank you ;)

Now I go to the gym. When I come back I will continue my work on this PR ;)

lib/eventsource/index.js
KhafraDev: yes, I prefer this way (specifically the comment, makes it easier to search for)

lib/eventsource/index.js
KhafraDev: From what I remember the spec uses the same language "an event named [open/error]". I checked Deno's code and it seemed to use `new Event("error")`.

KhafraDev:
If I missed anything else, lmk.

Uzlopak:
According to this

https://html.spec.whatwg.org/multipage/indices.html#event-error

It could have been an ErrorEvent, which was kind of helpful, as it could be enriched with a message for debugging purposes. But I removed the usage of ErrorEvent ;)

Uzlopak:
@KhafraDev 
@metcoder95 
@mcollina 

What do you think about this security issue in older version of the npm package eventsource?

https://github.com/EventSource/eventsource/pull/273

Uzlopak:
I think for a first implementation this is good enough. Maybe the Buffer.subarray calls can be reduced. But it is currently working. So would be happy if you do a review now ;).

lib/eventsource/eventsource-stream.js
```diff
@@ -0,0 +1,389 @@
...
+  _transform (chunk, _encoding, callback) {
+    if (chunk.length === 0) {
+      callback()
+      return
+    }
+
+    // We cache the chunk in the buffer, as the data might not be complete
+    // while processing it
+    if (this.buffer) {
+      this.buffer = Buffer.concat([this.buffer, chunk])
```
mcollina, line 112: please add a TODO to remove this

lib/eventsource/eventsource-stream.js
```diff
@@ -0,0 +1,389 @@
...
+    }
+
+    let field = ''
+    let value = ''
+
+    // If the line contains a U+003A COLON character (:)
+    if (colonPosition !== -1) {
+      // Collect the characters on the line before the first U+003A COLON
+      // character (:), and let field be that string.
+      field = line.subarray(0, colonPosition).toString('utf8')
```
mcollina, line 304: please add TODO to remove this

lib/eventsource/eventsource-stream.js
```diff
@@ -0,0 +1,389 @@
...
+
+      // Collect the characters on the line after the first U+003A COLON
+      // character (:), and let value be that string.
+      // If value starts with a U+0020 SPACE character, remove it from value.
+      let valueStart = colonPosition + 1
+      if (line[valueStart] === SPACE) {
+        ++valueStart
+      }
+      value = line.subarray(valueStart).toString('utf8')
```
mcollina, line 313: ditto

mcollina (commented):
Can you add TODOs to remove subarrays and concats for performance reasons?

KhafraDev:
@Uzlopak we already drop those headers on cross-origin redirect

Uzlopak:
@KhafraDev 
perfect!

@mcollina 
Added the TODOs and created an issue with some more information regarding it

KhafraDev (approved):

mcollina:
I don't think it'd be so easy ;).

mcollina:
I'd prefer to wait until the new v21.6.1 release is out so we can test this there too.

metcoder95 (approved):
Nice work 🚀

Uzlopak:
@mcollina 
@KhafraDev 
@metcoder95 
The CI/CD pipeline is green now ;).

PR #2612, opened 2024-01-14
fix(cache): set AbortSignal

tsctx:
AbortSignal was not set.

### DIFF
--- a/lib/cache/cache.js
+++ b/lib/cache/cache.js
@@ -8,7 +8,7 @@ const { webidl } = require('../fetch/webidl')
 const { Response, cloneResponse } = require('../fetch/response')
 const { Request } = require('../fetch/request')
 const { Headers } = require('../fetch/headers')
-const { kState, kHeaders, kGuard, kRealm } = require('../fetch/symbols')
+const { kState, kHeaders, kGuard, kRealm, kSignal } = require('../fetch/symbols')
 const { fetching } = require('../fetch/index')
 const { urlIsHttpHttpsScheme, createDeferredPromise, readAllBytes } = require('../fetch/util')
 const assert = require('assert')
@@ -509,6 +509,8 @@ class Cache {
         requestObject[kHeaders][kHeadersList] = request.headersList
         requestObject[kHeaders][kGuard] = 'immutable'
         requestObject[kRealm] = request.client
+        requestObject[kSignal] = new AbortController().signal
+        requestObject[kSignal][kRealm] = requestObject[kRealm]
 
         // 5.4.2.1
         requestList.push(requestObject)
--- a/lib/cache/cache.js
+++ b/lib/cache/cache.js
@@ -6,9 +6,9 @@ const { kEnumerableProperty, isDisturbed } = require('../core/util')
 const { kHeadersList } = require('../core/symbols')
 const { webidl } = require('../fetch/webidl')
 const { Response, cloneResponse } = require('../fetch/response')
-const { Request } = require('../fetch/request')
+const { Request, fromInnerRequest } = require('../fetch/request')
 const { Headers } = require('../fetch/headers')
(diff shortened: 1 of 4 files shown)


lib/cache/cache.js
```diff
@@ -509,6 +509,8 @@ class Cache {
         requestObject[kHeaders][kHeadersList] = request.headersList
         requestObject[kHeaders][kGuard] = 'immutable'
         requestObject[kRealm] = request.client
+        requestObject[kSignal] = new AbortController().signal
+        requestObject[kSignal][kRealm] = requestObject[kRealm]
```
ronag, line 513: this doesn't actually do anything? The abort contorller is just thrown away?

tsctx: This is necessary because AbortSignal is not set [here](https://github.com/nodejs/undici/blob/1b1ae6b60a058252dcaa655e9498b5bbad77bbde/lib/fetch/request.js#L363) because the new Request step was skipped in #2610.

The signal property reads kSignal.

https://github.com/nodejs/undici/blob/1b1ae6b60a058252dcaa655e9498b5bbad77bbde/lib/fetch/request.js#L715-L720

metcoder95 (approved):

lib/cache/cache.js
KhafraDev: this is why I didn't bother originally, causes too many issues

KhafraDev (approved):

mcollina:
Can you add a test?

tsctx:
I don't found any cache test.

tsctx:
I will implement the cache test when I have time.

KhafraDev:
The WPTs fairly make the assumption that constructing a Request will not have issues, as that would have major ramifications in other tests. Since we skip setting up a request, and do so manually, it caused this bug. I don't personally think we need a test here; if one is/you think otherwise, it'd probably be a better idea to revert the kConstruct commit, which guarantees we're doing everything correctly.

tsctx:
I don't think this kind of mistake will happen when we make utilities, so I'll make them later.

tsctx:
@mcollina ptal

lib/cache/cache.js
```diff
@@ -503,15 +503,12 @@ class Cache {
...
-        requestObject[kRealm] = request.client
-        requestObject[kSignal] = new AbortController().signal
-        requestObject[kSignal][kRealm] = requestObject[kRealm]
-
+        const requestObject = fromInnerRequest(
+          request,
+          new AbortController().signal,
+          'immutable',
+          { settingsObject: request.client }
+        )
```
tsctx, line 511: The use of utility can prevent such mistakes.

mcollina:
Code is ok, test is missing ;)

tsctx:
@mcollina
I was told that the test was not necessary, so I did not create it.

https://github.com/nodejs/undici/pull/2612#issuecomment-1897010136

KhafraDev:
tests for the new utility are fine, I don't see any value of testing Request in Cache/CacheStorage tests

lib/fetch/request.js
```diff
@@ -844,6 +844,26 @@ function cloneRequest (request) {
...
 }
 
+/**
+ * @param {any} innerRequest
+ * @param {AbortSignal} signal
+ * @param {'request' | 'immutable' | 'request-no-cors' | 'response' | 'none'} guard
+ * @param {any} [realm]
+ * @returns {Request}
+ */
+function fromInnerRequest (innerRequest, signal, guard, realm) {
```
metcoder95, line 854: Out of curiosity, why use a function for a single place that uses it?
I might be missing the holistic view, but how does this solve the problem?

metcoder95: Ok, now I see the usage; resolving my own comment 😅

mcollina (approved):
lgtm

PR #2619, opened 2024-01-16
fix(mock-agent): split set-cookie

### LINKED ISSUE #2616: MockAgent unexpected multi header behaviour with fetch
First of all thank you for maintaining this library ❤️

## Bug Description

I encountered an unexpected behaviour when using `fetch` with multi value headers via `MockAgent`.
I would expect the behaviour from `fetch` to be the same when using `request` [(reference test)](https://github.com/nodejs/undici/blob/3372aed902951f135d710d0fb430f39b8571e547/test/mock-agent.js#L2564-L2593) with `MockAgent`.
Is this behaviour expected and I am just missing something?

## Reproducible By

save the following content as `test/repro.js` in the undici repository and then run `npx tap test/repro.js`
```js
'use strict'

const { test } = require('tap')
const { setGlobalDispatcher, MockAgent } = require('..')
const { fetch } = require('..')

test("MockAgent - headers should be array of strings (fetch)", async (t) => {

  const mockPool = mockAgent.get("http://localhost:3000");

  mockPool
    .intercept({
      path: "/foo",
      method: "GET",
    })
    .reply(200, "foo", {
      headers: {
        "set-cookie": ["foo=bar", "bar=baz", "baz=qux"],
      },
    });

  const { headers } = await fetch("http://localhost:3000/foo", {
    method: "GET",
  });

  t.equal(headers.getSetCookie(), ["foo=bar", "bar=baz", "baz=qux"]);
});
```

## Expected Behavior

`headers.getSetCookie()` should return an array (`["foo=bar", "bar=baz", "baz=qux"]`) but instead it just returns `["foo=bar,bar=baz,baz=qux"]` (array with one entry and all values concatenated)

## Logs & Screenshots

```
 FAIL ​ test/repro.js
 ✖ should be equal

  test/repro.js                                                               
  27 |   });                                                                  
  28 |                                                                        
> 29 |   t.equal(headers.getSetCookie(), ["foo=bar", "bar=baz", "baz=qux"]);  
     | ----^                                                                  
  30 | });                                                                    

  --- expected                  
  +++ actual                    
  @@ -1,5 +1,3 @@               
   Array [                      
  -  "foo=bar",                 
  -  "bar=baz",                 
  -  "baz=qux",                 
  +  "foo=bar,bar=baz,baz=qux", 
   ]                            

  test: test/repro.js MockAgent - headers should be array of strings (fetch)
```

## Environment

- `Node v20.11.0`
- `ubuntu-22.04`

tsctx:
Fixes #2616

### DIFF
--- a/lib/mock/mock-utils.js
+++ b/lib/mock/mock-utils.js
@@ -188,11 +188,21 @@ function buildKey (opts) {
 }
 
 function generateKeyValues (data) {
-  return Object.entries(data).reduce((keyValuePairs, [key, value]) => [
-    ...keyValuePairs,
-    Buffer.from(`${key}`),
-    Array.isArray(value) ? value.map(x => Buffer.from(`${x}`)) : Buffer.from(`${value}`)
-  ], [])
+  const keys = Object.keys(data)
+  const result = []
+  for (let i = 0; i < keys.length; ++i) {
+    const key = keys[i]
+    const value = data[key]
+    const name = Buffer.from(`${key}`)
+    if (Array.isArray(value)) {
+      for (let j = 0; j < value.length; ++j) {
+        result.push(name, Buffer.from(`${value[j]}`))
+      }
+    } else {
+      result.push(name, Buffer.from(`${value}`))
+    }
+  }
+  return result
 }
 
 /**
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

KhafraDev (approved):

metcoder95 (approved):
Thanks!

PR #2621, opened 2024-01-16
exit with 1 if WPT runner has unexpected errors

KhafraDev:
cc @mcollina @panva

### DIFF
--- a/test/wpt/runner/runner.mjs
+++ b/test/wpt/runner/runner.mjs
@@ -313,7 +313,7 @@ export class WPTRunner extends EventEmitter {
       `skipped: ${skipped}`
     )
 
-    process.exit(failed - expectedFailures ? 1 : 0)
+    process.exit(failed - expectedFailures ? 1 : process.exitCode)
   }
 
   addInitScript (code) {
--- a/test/wpt/runner/runner.mjs
+++ b/test/wpt/runner/runner.mjs
@@ -313,7 +313,7 @@ export class WPTRunner extends EventEmitter {
       `skipped: ${skipped}`
     )
 
-    process.exit(0)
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

test/wpt/runner/worker.mjs
```diff
@@ -89,6 +90,11 @@ Object.defineProperties(globalThis, {
   CacheStorage: {
     ...globalPropertyDescriptors,
     value: CacheStorage
+  },
+  // TODO: remove once node 18 is dropped
+  crypto: {
+    ...globalPropertyDescriptors,
+    value: webcrypto
```
panva, line 97: While this may satisfy the WPT requirements it's not a proper WebCryptoAPI polyfill. Couldn't you just set NODE_OPTIONS to `--experimental-global-webcrypto` instead?

KhafraDev: I have vague recollections of having issues with windows when adding options

panva: Right, proceed as you see best fit then?

KhafraDev: I conditionally add it

test/wpt/runner/runner.mjs
```diff
@@ -313,7 +313,7 @@ export class WPTRunner extends EventEmitter {
       `skipped: ${skipped}`
     )
 
-    process.exit(0)
+    process.exit(failed - expectedFailures ? 1 : 0)
```
panva, line 316: Just to ensure any otherwise set non-zero exitCode is not lost.

```suggestion
    process.exit(failed - expectedFailures ? 1 : process.exitCode)
```

PR #2626, opened 2024-01-18
add webSocket example

mertcanaltin:
I have attached a websocket example

### DIFF
--- a/examples/proxy/websocket.js
+++ b/examples/proxy/websocket.js
@@ -1,0 +1,16 @@
+const WebSocket = require('ws');
+
+function createWebSocketServer() {
+  const wss = new WebSocket.Server({ noServer: true });
+
+  wss.on('connection', ws => {
+    ws.on('message', message => {
+      console.log(`Received message: ${message}`);
+      ws.send('Received your message!');
+    });
+  });
+
+  return wss;
+}
+
+module.exports = createWebSocketServer;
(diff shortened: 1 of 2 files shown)


mertcanaltin:
@ronag, @mcollina  Greetings, I wonder if I am proceeding correctly here, I would be very happy if you could take a look 🚀

metcoder95:
Are we solving any specific issue in here?

mertcanaltin:
> Are we solving any specific issue in here?

no, it was just a todo I saw and I wanted to fill in the gaps @metcoder95

metcoder95 (commented):
The example LGTM, I would just put it in maybe another file within `/proxy` directory just to not overlap the with the current `proxy` example

mertcanaltin:
> The example LGTM, I would just put it in maybe another file within `/proxy` directory just to not overlap the with the current `proxy` example

now I am implementing it immediately, I will try to implement it in a separate file instead of writing it inside.

mertcanaltin:
I did an update, I hope I went the right way @metcoder95 🚀

examples/proxy/websocket.js
metcoder95, line 1: Maybe move the whole WebSocket example here from the index, and delete it there to keep it clean?

mertcanaltin: yes, I applied this

metcoder95 (approved):

PR #2254, opened 2023-09-08
docs: fix tables in README

regseb:
## This relates to...

N/A

## Rationale

Tables are displayed incorrectly in the [README](https://github.com/nodejs/undici/blob/932e5f6/README.md#benchmarks). https://github.com/nodejs/undici/pull/2250

## Changes

- Replace `│` (`U+2502` / Box Drawings Light Vertical) by `|` (`U+007C` / Vertical Line).
- Replace `─` (`U+2500` / Box Drawings Light Horizontal) by `-` (`U+002D` / Hyphen-Minus).

### Features

N/A

### Bug Fixes

N/A

### Breaking Changes and Deprecations

N/A

## Status

- [x] I have read and agreed to the [Developer's Certificate of Origin][cert]
- [ ] Tested
- [ ] Benchmarked (**optional**)
- [x] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

[cert]: https://github.com/nodejs/undici/blob/main/CONTRIBUTING.md

PR #2629, opened 2024-01-19
Use correct http Agent for node-fetch, axios, got and request

mcollina:
After a few people asked me questions on X, I took a closer look at the results, and I've found a discrepancy: axios, got and node-fetch did not have their agent set.  This resulted in skewed results.

The main difference is that our `fetch()` implementation is faster than node-fetch.


Ethan-Arrowood (approved):
LGTM. IMO we should benchmark the default behavior of these clients. Regardless I'm fine with this change too.

mcollina:
The default behavior differs greatly due to different default settings, it'd be unfair and unpredictable.

regseb:
@mcollina The table in the [README](https://github.com/nodejs/undici#benchmarks) is incorrectly displayed. This is because the _Vertical Line_ and _Hyphen-Minus_ have been replaced by _Box Drawings Light Vertical_ and _Horizontal_. I had corrected the problem in the https://github.com/nodejs/undici/pull/2254 pull request. But the problem has reappeared because of this pull request.

PR #2633, opened 2024-01-23
Allow offline run of build:node script

khardix:
## This relates to...

https://github.com/nodejs/undici/pull/2342

## Rationale

During packaging for a Linux distribution, the build systems are usually/sometimes disconnected from the internet,
and can not use anything not present in source tarball or installed on the system.

To facilitate NodeJS packages, in Fedora we cache the `node_modules` directory when creating a package update, and then re-use this cached directory for all subsequent builds. That means that we need all the build-time dependencies installed after invoking `npm install --include=dev`. This is currently not true for the `build:node` script, which pulls `esbuild@0.19.4` at the point of invocation. If invoked in offline mode (`npm --offline run build:node`), this fails.

## Changes

- Include esbuild in `devDependencies`, so that it is installed on `npm install --include=dev`.
- Use the installed version when running the `build:node` script.

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/package.json
+++ b/package.json
@@ -70,7 +70,7 @@
     "docs"
   ],
   "scripts": {
-    "build:node": "npx esbuild@0.19.4 index-fetch.js --bundle --platform=node --outfile=undici-fetch.js --define:esbuildDetection=1 --keep-names",
+    "build:node": "npx esbuild index-fetch.js --bundle --platform=node --outfile=undici-fetch.js --define:esbuildDetection=1 --keep-names",
     "prebuild:wasm": "node build/wasm.js --prebuild",
     "build:wasm": "node build/wasm.js --docker",
     "lint": "standard | snazzy",
@@ -113,6 +113,7 @@
     "delay": "^5.0.0",
     "dns-packet": "^5.4.0",
     "docsify-cli": "^4.4.3",
+    "esbuild": "^0.19.4",
     "form-data": "^4.0.0",
     "formdata-node": "^6.0.3",
     "https-pem": "^3.0.0",


khardix:
Note: I'm not that versed in invocation of `npx`, so if there is a better way to achieve what I need, let me know.

KhafraDev:
It was added by @targos in 946ea781e

The description may apply here:
> It breaks AIX in CITGM. Use npx instead.

khardix:
Hm, no idea what AIX is, but your comment tells me this was a deliberate change and not easily reverted. Since I can patch this downstream without too much of an effort, I'm withdrawing this PR. Thanks for the explanation!

targos:
It's not that one invocation works while the other does not. It's that the `npm install` command failed on the AIX platform when `esbuild` was a dependency.

KhafraDev:
It is, we test the node bundle in the fetch suite.

targos:
Maybe it wasn't back in 2022

richardlau:
> It's not that one invocation works while the other does not. It's that the `npm install` command failed on the AIX platform when `esbuild` was a dependency. This particular command is not used to run `undici` test suite.

It used to be the case that `esbuild` didn't provide binaries for AIX but as of https://github.com/evanw/esbuild/pull/3550 and `esbuild@0.19.10` it does.
Refs: https://github.com/evanw/esbuild/issues/3549

mhdawson:
From my perspective if we can somehow test that CITGM will run, then that would be best and would confirm that this change would be ok.

PR #2635, opened 2024-01-24
refactor: use fromInnerResponse

tsctx:
Make the code easy to read.

### DIFF
--- a/lib/cache/cache.js
+++ b/lib/cache/cache.js
@@ -781,7 +781,7 @@ class Cache {
     // 5.5.2
     for (const response of responses) {
       // 5.5.2.1
-      const responseObject = fromInnerResponse(response, 'immutable')
+      const responseObject = fromInnerResponse(response, 'immutable', { settingsObject: {} })
 
       responseList.push(responseObject.clone())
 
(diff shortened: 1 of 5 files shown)


lib/cache/cache.js
```diff
@@ -783,11 +781,7 @@ class Cache {
     // 5.5.2
     for (const response of responses) {
       // 5.5.2.1
-      const responseObject = new Response(kConstruct)
-      responseObject[kState] = response
-      responseObject[kHeaders] = new Headers(kConstruct)
-      responseObject[kHeaders][kHeadersList] = response.headersList
-      responseObject[kHeaders][kGuard] = 'immutable'
+      const responseObject = fromInnerResponse(response, 'immutable')
```
KhafraDev, line 784: no realm

tsctx: Is this correct?
```suggestion
      const responseObject = fromInnerResponse(response, 'immutable', { settingsObject: {} })
```

tsctx: Resolved.

mcollina (approved):
lgtm

lib/fetch/response.js
```diff
@@ -512,6 +487,23 @@ function initializeResponse (response, init, body) {
   }
 }
 
+/**
+ * @param {any} innerResponse
```
KhafraDev, line 491: Can you link to https://fetch.spec.whatwg.org/#response-create?

KhafraDev (changes requested):

lib/fetch/response.js
tsctx:
```suggestion
 * @see https://fetch.spec.whatwg.org/#response-create
 * @param {any} innerResponse
```

PR #2640, opened 2024-01-25
Use data event to consume body

### LINKED ISSUE #2639: Large body is truncated
If I make a second request before awaiting for the first request large body, it looks like it truncates the large body. Maybe I'm doing something wrong.

```js

'use strict'

const { createServer } = require('http')

const LARGE_BODY = 'x'.repeat(10000000)

async function run () {
  // Works fine
  // const fooRes = await request('http://localhost:3000/foo')
  // const fooBody = await fooRes.body.text()

  // const barRes = await request('http://localhost:3000/bar')
  // await barRes.body.text()

  // Fails with:
  const fooRes = await request('http://localhost:3000/foo')
  const barRes = await request('http://localhost:3000/bar')

  const fooBody = await fooRes.body.text()
  await barRes.body.text()

  if (fooBody !== LARGE_BODY) {
    const contentLength = fooRes.headers['content-length']
    console.log('content length:', contentLength)
    console.log('received body length:', fooBody.length)
    console.log('expected body length:', LARGE_BODY.length)
    return
  }
  console.log('done')
}

  if (req.url === '/foo') {
    res.end(LARGE_BODY)
    return
  }
  if (req.url === '/bar') {
    res.end('bar')
    return
  }
  throw new Error('Unexpected request url: ' + req.url)
})

server.listen(3000, async () => {
  await run()
  process.exit(0)
})
```

mcollina:
Fixes #2639

### DIFF
--- a/lib/api/readable.js
+++ b/lib/api/readable.js
@@ -93,14 +93,6 @@ module.exports = class BodyReadable extends Readable {
     return this.off(ev, ...args)
   }
 
-  push (chunk) {
-    if (this[kConsume] && chunk !== null && this.readableLength === 0) {
-      consumePush(this[kConsume], chunk)
-      return this[kReading] ? super.push(chunk) : true
-    }
-    return super.push(chunk)
-  }
-
   // https://fetch.spec.whatwg.org/#dom-body-text
   async text () {
     return consume(this, 'text')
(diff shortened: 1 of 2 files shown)


ronag (changes requested):
This doesn't look right. I'm sick atm. Not sure if I will have energy to look into before weekend.

mcollina:
Why? The code was very odd before, it bypassed all internal stream buffering before, and it's passing all our tests.

ronag:
> Why? The code was very odd before, it bypassed all internal stream buffering before, and it's passing all our tests.

The point of the code is to bypass the stream logic to improve performance. If we don't want that then I believe the code can be significantly simplified.

mcollina:
Well, I guess that code was missing some handling of an edge case because it was truncating some data in the condition included in the test.

I can try to take a look if a fix that preserve the fast path could be done.

PR #2645, opened 2024-01-26
fix: expose EventSource for nodejs

Uzlopak:
I would actually recommend to rename index-fetch.js to index-node.js as it is not only exposing fetch but also Websocket and now EventSource. 

But this would mean that we also have to patch the dependency updater for undici in nodejs

This is afaik blocking https://github.com/nodejs/node/pull/51575

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/index-fetch.js
+++ b/index-fetch.js
@@ -14,4 +14,7 @@ module.exports.FormData = require('./lib/fetch/formdata').FormData
 module.exports.Headers = require('./lib/fetch/headers').Headers
 module.exports.Response = require('./lib/fetch/response').Response
 module.exports.Request = require('./lib/fetch/request').Request
+
 module.exports.WebSocket = require('./lib/websocket/websocket').WebSocket
+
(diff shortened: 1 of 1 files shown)


Uzlopak:
@KhafraDev 
@metcoder95 
@mcollina

anonrig (approved):

KhafraDev (approved):
This does need approval and discussion beforehand, same as when I added WebSocket here.

mcollina:
@nodejs/tsc are you ok in principle in adding EventSource to Node.js?

richardlau:
Re. the commit message -- this is a feature rather than "fix"?

Uzlopak:
When I created this PR I thought it would be a fix as I expected it to work in the first place. You want to get the commit message fixed? We can rename the PR title and it should be taken over when squash merging in the github ui.

PR #2650, opened 2024-01-27
fix: support deflate raw responses

### LINKED ISSUE #2128: Handle the infamous raw deflate response from old servers
## Bug Description

Something node-fetch and browser could handle was the infamous raw deflate response from old servers
basically if the coding is either: `coding === 'deflate' || coding === 'x-deflate'`
then read the first byte. if it match `chunk[0] & 0x0F) === 0x08` then use `zlib.createInflate()` otherwise use `zlib.createInflateRaw()`

## Reproducible By

You already have a test for it here:
https://github.com/nodejs/undici/blob/51fa0fe3c62a770e8b1bbead0665f0bf9c5b81a2/test/node-fetch/main.js#L663-L672

### Additional context

Relevant part that needs fixing:
https://github.com/nodejs/undici/blob/51fa0fe3c62a770e8b1bbead0665f0bf9c5b81a2/lib/fetch/index.js#L2018-L2019

how node-fetch solved it:
https://github.com/node-fetch/node-fetch/blob/7b86e946b02dfdd28f4f8fca3d73a022cbb5ca1e/src/index.js#L314-L350

### LINKED ISSUE #2603: Error: incorrect header check Z_DATA_ERROR
## Bug Description

I am trying to fetch a URL but it fails. It works just fine when I use `node-fetch`, `curl` or the native browser `fetch`. It's similar to #1271 but there is no redirect happening.

These are the response headers:

```
{
  'cache-control': 'no-cache',
  'content-encoding': 'deflate',
  'content-type': 'text/xml; charset=utf-8',
  date: 'Mon, 08 Jan 2024 06:12:38 GMT',
  server: 'lighttpd/1.4.45',
  'transfer-encoding': 'chunked'
}
```

## Reproducible By

```js
const response = await fetch( 'https://egauge90013.egaug.es/cgi-bin/egauge-show?a&E' );

const text = await response.text();

console.log( text );
```

## Expected Behavior

I would expect the response to log into console

## Logs

Instead I get this error:

```
TypeError: terminated
    at Fetch.onAborted (node:internal/deps/undici/undici:11279:53)
    at Fetch.emit (node:events:513:28)
    at Fetch.emit (node:domain:489:12)
    at Fetch.terminate (node:internal/deps/undici/undici:10534:14)
    at Fetch.fetchParams.controller.resume (node:internal/deps/undici/undici:11258:36)
    at processTicksAndRejections (node:internal/process/task_queues:95:5) {
  [cause]: Error: incorrect header check
      at Zlib.zlibOnError [as onerror] (node:zlib:189:17) {
    errno: -3,
    code: 'Z_DATA_ERROR'
  }
}
```

## Environment

Node v20.7.0 and v18.14.2

Uzlopak:
Closes #1091 
Closes #2128
Closes #2603 

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -43,7 +43,8 @@ const {
   urlHasHttpsScheme,
   clampAndCoursenConnectionTimingInfo,
   simpleRangeHeaderValue,
-  buildContentRange
+  buildContentRange,
+  createInflate
 } = require('./util')
 const { kState, kHeaders, kGuard, kRealm } = require('./symbols')
 const assert = require('assert')
@@ -1096,7 +1097,7 @@ function fetchFinale (fetchParams, response) {
     // 3. Set up transformStream with transformAlgorithm set to identityTransformAlgorithm and flushAlgorithm
     //    set to processResponseEndOfBody.
     const transformStream = new TransformStream({
-      start () {},
+      start () { },
       transform (chunk, controller) {
         controller.enqueue(chunk)
       },
@@ -2186,7 +2187,7 @@ async function httpNetworkFetch (
                   finishFlush: zlib.constants.Z_SYNC_FLUSH
                 }))
               } else if (coding === 'deflate') {
-                decoders.push(zlib.createInflate())
+                decoders.push(createInflate())
               } else if (coding === 'br') {
                 decoders.push(zlib.createBrotliDecompress())
               } else {
@@ -2202,7 +2203,7 @@ async function httpNetworkFetch (
             headersList,
             body: decoders.length
               ? pipeline(this.body, ...decoders, () => { })
-              : this.body.on('error', () => {})
+              : this.body.on('error', () => { })
           })
 
           return true
--- a/lib/fetch/util.js
+++ b/lib/fetch/util.js
@@ -1184,6 +1184,10 @@ function buildContentRange (rangeStart, rangeEnd, fullLength) {
 class InflateStream extends Transform {
   _transform (chunk, encoding, callback) {
     if (!this._inflateStream) {
+      if (chunk.length === 0) {
+        callback()
+        return
+      }
       this._inflateStream = (chunk[0] & 0x0F) === 0x08
         ? zlib.createInflate()
         : zlib.createInflateRaw()
--- a/lib/fetch/util.js
+++ b/lib/fetch/util.js
@@ -1189,8 +1189,8 @@ class InflateStream extends Transform {
         : zlib.createInflateRaw()
 
       this._inflateStream.on('data', this.push.bind(this))
-      this._inflateStream.on('end', callback)
-      this._inflateStream.on('error', callback)
+      this._inflateStream.on('end', () => this.push(null))
+      this._inflateStream.on('error', (err) => this.destroy(err))
     }
 
     this._inflateStream.write(chunk, encoding, callback)
--- a/lib/fetch/util.js
+++ b/lib/fetch/util.js
@@ -1,5 +1,7 @@
 'use strict'
 
+const { Transform } = require('node:stream')
+const zlib = require('node:zlib')
 const { redirectStatusSet, referrerPolicySet: referrerPolicyTokens, badPortsSet } = require('./constants')
 const { getGlobalOrigin } = require('./global')
 const { performance } = require('perf_hooks')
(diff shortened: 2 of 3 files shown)


Uzlopak:
are the failing tests valid?

lib/fetch/index.js
```diff
@@ -2186,7 +2186,35 @@ async function httpNetworkFetch (
                   finishFlush: zlib.constants.Z_SYNC_FLUSH
                 }))
               } else if (coding === 'deflate') {
-                decoders.push(zlib.createInflate())
+                // Instantiate a Stream, which pipes the response to zlib.createInflate()
+                // or zlib.createInflateRaw() depending on the first byte of the Buffer.
+                // If the lower byte of the first byte is 0x08, then the stream is
+                // interpreted as a zlib stream, otherwise it's interpreted as a
+                // raw deflate stream.
+                const deflateDecoder = new Transform({
```
mcollina, line 2194: please move this into its own file, or at the top of the file

Uzlopak:
@mcollina 
Is this better?

Can transform actually be called with an empty chunk? If so, i would need to add a check for chunk not being an empty buffer.

ronag (changes requested):
Inflate stream looks wrong

lib/fetch/util.js
```diff
@@ -1174,6 +1176,39 @@ function buildContentRange (rangeStart, rangeEnd, fullLength) {
...
+// raw deflate stream.
+class InflateStream extends Transform {
+  _transform (chunk, encoding, callback) {
+    if (!this._inflateStream) {
+      this._inflateStream = (chunk[0] & 0x0F) === 0x08
+        ? zlib.createInflate()
+        : zlib.createInflateRaw()
+
+      this._inflateStream.on('data', this.push.bind(this))
+      this._inflateStream.on('end', callback)
```
ronag, line 1192: Push null

lib/fetch/util.js
```diff
@@ -1174,6 +1176,39 @@ function buildContentRange (rangeStart, rangeEnd, fullLength) {
...
+      this._inflateStream.on('error', callback)
```
ronag, line 1193: Destory(err)

ronag (changes requested):

Uzlopak:
@ronag 
Sorry streams are not easy to grok. Now correct?

lib/fetch/util.js
```diff
@@ -1174,6 +1176,39 @@ function buildContentRange (rangeStart, rangeEnd, fullLength) {
...
+      this._inflateStream = (chunk[0] & 0x0F) === 0x08
+        ? zlib.createInflate()
+        : zlib.createInflateRaw()
+
+      this._inflateStream.on('data', this.push.bind(this))
+      this._inflateStream.on('end', () => this.push(null))
+      this._inflateStream.on('error', (err) => this.destroy(err))
+    }
+
+    this._inflateStream.write(chunk, encoding, callback)
```
ronag, line 1196:
```suggestion
    if (this._inflateStream.write(chunk, encoding)) {
      callback(null)
    } else {
      // TODO: Optimize by registering a single callback that always invokes this[kCallback] which is set here.
      this._inflateStream.once('drain', callback) 
    }
```

ronag (approved):
Looks correct now, if not optimal in terms of performance

metcoder95 (approved):

Uzlopak:
@ronag 
I have unfortunately no better idea, on how to improve the performance on this. I mean... it is probably halving the performance because it is an additional stream in the "middle". 

What about _transform? can it happen that chunk has size of 0 when calling _transform the first time?

lib/fetch/util.js
```diff
@@ -1174,6 +1176,39 @@ function buildContentRange (rangeStart, rangeEnd, fullLength) {
...
 
+// A Stream, which pipes the response to zlib.createInflate() or
+// zlib.createInflateRaw() depending on the first byte of the Buffer.
+// If the lower byte of the first byte is 0x08, then the stream is
+// interpreted as a zlib stream, otherwise it's interpreted as a
+// raw deflate stream.
+class InflateStream extends Transform {
+  _transform (chunk, encoding, callback) {
+    if (!this._inflateStream) {
+      this._inflateStream = (chunk[0] & 0x0F) === 0x08
```
Uzlopak, line 1187: @mcollina 
@ronag 
I would like to add this to the code, just to be sure. wdyt?

```suggestion
      if (chunk.length === 0) {
        callback()
        return
      }
      this._inflateStream = (chunk[0] & 0x0F) === 0x08
```

ronag: Doesn't hurt

ronag (approved):

ronag (approved):

mcollina (approved):
lgtm

---

This would be _slow_.

A solution would be to defer resolving the promise until we get the first chunk, get the first byte, unshift the rest and  assemble the pipeline correctly.

Uzlopak:
@mcollina 

I think this would need some heavier refactoring.

ronag (approved):

Uzlopak:
We could open a follow up issue, regarding the performance aspect.

PR #2654, opened 2024-01-27
chore: reduce noise in test logs test/client-node-max-header-size.js

Uzlopak:
Does not print the output of the execSync call into the console. Also refactored to use node core test runner.


mcollina (approved):
lgtm

PR #2660, opened 2024-01-28
fix: resolve event-loop blocking when aborting a fetch

### LINKED ISSUE #2198: Memory leak when aborting fetch requestors
## Bug Description

A memory leak occurs when frequently aborting fetch requests.

## Reproducible By

Run the following code in Node.js.

```javascript
import undici from "undici";

(function attack() {
    undici.fetch(
        "http://idonotexist:9999",
        {signal: controller.signal}
    ).catch(attack);
    controller.abort();
}());

```

In less than a minute, I see memory usage of the Node.js process climb to several gigabytes.

## Expected Behavior

Memory usage to remain stable, as is the case when you comment out the `controller.abort();` statement.

## Environment

```
$ uname -a
Darwin MacBook-Pro.local 22.5.0 Darwin Kernel Version 22.5.0: Thu Jun  8 22:22:20 PDT 2023; root:xnu-8796.121.3~7/RELEASE_ARM64_T6000 arm64 arm Darwin
$ node -v
v20.4.0
```

Uzlopak:
Resolves #2198

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -313,16 +313,14 @@ function finalizeAndReportTiming (response, initiatorType = 'other') {
 }
 
 // https://w3c.github.io/resource-timing/#dfn-mark-resource-timing
-function markResourceTiming (timingInfo, originalURL, initiatorType, globalThis, cacheState) {
-  if (nodeMajor > 18 || (nodeMajor === 18 && nodeMinor >= 2)) {
-    performance.markResourceTiming(timingInfo, originalURL.href, initiatorType, globalThis, cacheState)
-  }
-}
+const markResourceTiming = (nodeMajor > 18 || (nodeMajor === 18 && nodeMinor >= 2))
+  ? performance.markResourceTiming
+  : () => {}
 
 // https://fetch.spec.whatwg.org/#abort-fetch
 function abortFetch (p, request, responseObject, error) {
   // 1. Reject promise with error.
-  p.reject(error)
+  setImmediate(() => p.reject(error))
 
   // 2. If request’s body is not null and is readable, then cancel request’s
   // body with error.
--- a/lib/pool.js
+++ b/lib/pool.js
@@ -76,18 +76,17 @@ class Pool extends PoolBase {
   }
 
   [kGetDispatcher] () {
-    let dispatcher = this[kClients].find(dispatcher => !dispatcher[kNeedDrain])
-
-    if (dispatcher) {
-      return dispatcher
+    for (const client of this[kClients]) {
+      if (!client[kNeedDrain]) {
+        return client
+      }
     }
 
     if (!this[kConnections] || this[kClients].length < this[kConnections]) {
-      dispatcher = this[kFactory](this[kUrl], this[kOptions])
+      const dispatcher = this[kFactory](this[kUrl], this[kOptions])
       this[kAddClient](dispatcher)
+      return dispatcher
     }
-
-    return dispatcher
   }
 }
 


lib/fetch/index.js
```diff
@@ -313,16 +313,14 @@ function finalizeAndReportTiming (response, initiatorType = 'other') {
...
 
 // https://w3c.github.io/resource-timing/#dfn-mark-resource-timing
-function markResourceTiming (timingInfo, originalURL, initiatorType, globalThis, cacheState) {
-  if (nodeMajor > 18 || (nodeMajor === 18 && nodeMinor >= 2)) {
-    performance.markResourceTiming(timingInfo, originalURL.href, initiatorType, globalThis, cacheState)
-  }
-}
+const markResourceTiming = (nodeMajor > 18 || (nodeMajor === 18 && nodeMinor >= 2))
+  ? performance.markResourceTiming
+  : () => {}
```
Uzlopak, line 318: This showed hot in the flame graphs, low hanging fruit.

lib/pool.js
```diff
@@ -76,18 +76,17 @@ class Pool extends PoolBase {
...
+      if (!client[kNeedDrain]) {
+        return client
+      }
     }
 
     if (!this[kConnections] || this[kClients].length < this[kConnections]) {
-      dispatcher = this[kFactory](this[kUrl], this[kOptions])
+      const dispatcher = this[kFactory](this[kUrl], this[kOptions])
       this[kAddClient](dispatcher)
+      return dispatcher
```
Uzlopak, line 88: this was also kind of slow while benchmarking

lib/fetch/index.js
```diff
@@ -313,16 +313,14 @@ function finalizeAndReportTiming (response, initiatorType = 'other') {
...
 
 // https://fetch.spec.whatwg.org/#abort-fetch
 function abortFetch (p, request, responseObject, error) {
   // 1. Reject promise with error.
-  p.reject(error)
+  setImmediate(() => p.reject(error))
```
ronag, line 323: This shouldn't be necessary according to spec. Are we sure this is a spec bug and not something else?

ronag (changes requested):

lib/fetch/index.js
Uzlopak: Something is clogging the event loop. As I see other tests fail. So this is not the solution for the issue.

calling setImmediate resulted in declogging it. Still investigating.

lib/fetch/index.js
```diff
@@ -313,16 +313,14 @@ function finalizeAndReportTiming (response, initiatorType = 'other') {
 }
 
 // https://w3c.github.io/resource-timing/#dfn-mark-resource-timing
-function markResourceTiming (timingInfo, originalURL, initiatorType, globalThis, cacheState) {
-  if (nodeMajor > 18 || (nodeMajor === 18 && nodeMinor >= 2)) {
-    performance.markResourceTiming(timingInfo, originalURL.href, initiatorType, globalThis, cacheState)
```
KhafraDev, line 318: originalURL is being passed instead of the stringified url

KhafraDev (changes requested):

ronag (commented):
maybe leave optimization to separate PR?

lib/fetch/index.js
KhafraDev: I think this is an issue in v8, I vaguely recall an issue in node core where promises weren't being GC'ed in a loop

KhafraDev: https://github.com/nodejs/node/issues/43655

Uzlopak:
Yes it is an upstream issue.

KhafraDev:
I don't think this needed to be closed.

Uzlopak:
Lets discuss this first in the issue. I put the other optimizations into a different PR.

PR #2661, opened 2024-01-29
chore: small perf improvements

Uzlopak:
as discussed in #2660 these two places were hot in the flamegraphs. 

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -305,7 +305,7 @@ function finalizeAndReportTiming (response, initiatorType = 'other') {
   // global, and cacheState.
   markResourceTiming(
     timingInfo,
-    originalURL,
+    originalURL.href,
     initiatorType,
     globalThis,
     cacheState
@@ -313,11 +313,9 @@ function finalizeAndReportTiming (response, initiatorType = 'other') {
 }
 
 // https://w3c.github.io/resource-timing/#dfn-mark-resource-timing
-function markResourceTiming (timingInfo, originalURL, initiatorType, globalThis, cacheState) {
-  if (nodeMajor > 18 || (nodeMajor === 18 && nodeMinor >= 2)) {
-    performance.markResourceTiming(timingInfo, originalURL.href, initiatorType, globalThis, cacheState)
-  }
-}
+const markResourceTiming = (nodeMajor > 18 || (nodeMajor === 18 && nodeMinor >= 2))
+  ? performance.markResourceTiming
+  : () => {}
 
 // https://fetch.spec.whatwg.org/#abort-fetch
 function abortFetch (p, request, responseObject, error) {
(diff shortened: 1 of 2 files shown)


lib/fetch/index.js
```diff
@@ -305,19 +305,17 @@ function finalizeAndReportTiming (response, initiatorType = 'other') {
...
   )
 }
 
 // https://w3c.github.io/resource-timing/#dfn-mark-resource-timing
-function markResourceTiming (timingInfo, originalURL, initiatorType, globalThis, cacheState) {
-  if (nodeMajor > 18 || (nodeMajor === 18 && nodeMinor >= 2)) {
-    performance.markResourceTiming(timingInfo, originalURL.href, initiatorType, globalThis, cacheState)
-  }
-}
+const markResourceTiming = (nodeMajor > 18 || (nodeMajor === 18 && nodeMinor >= 2))
```
KhafraDev, line 316: this seems highly unlikely to improve performance here, unless i'm missing something

Uzlopak:
@KhafraDev 

Yeah probably doesnt make a big difference. Just removes one layer.

before:
image

after:
image

mcollina (approved):
lgtm, both those changes are sound

lib/fetch/index.js
mcollina: it removes an if from an hot path. Given it's not the main bottleneck   this is unlikely to yield massive improvements, but it's still something.

ronag (commented):
I'm surprised this makes a difference. Do you have flamegraphs for before and after?

lib/fetch/index.js
```diff
@@ -1044,7 +1042,7 @@ function fetchFinale (fetchParams, response) {
       //    and responseStatus.
       if (fetchParams.request.initiatorType != null) {
         // TODO: update markresourcetiming
-        markResourceTiming(timingInfo, fetchParams.request.url, fetchParams.request.initiatorType, globalThis, cacheState, bodyInfo, responseStatus)
+        markResourceTiming(timingInfo, fetchParams.request.url.href, fetchParams.request.initiatorType, globalThis, cacheState, bodyInfo, responseStatus)
```
mcollina, line 1045: Can we detect if somebody is listening to these resourceTiminig? This seems like an expensive call that we could potentially skip if there is no one watching.

Uzlopak:
@mcollina 

If I read it right, then there is currently no way to access the kObservers Set in nodejs to figure out if there is a Observer listening on specific events. 

https://github.com/nodejs/node/blob/0f461aad4cbf747fa053714f08e5030ddf019977/lib/internal/perf/observe.js#L118

mcollina:
@ronag @KhafraDev do you object in landing this?

KhafraDev (approved):

PR #2670, opened 2024-01-30
chore: use arrow functions

Uzlopak:


mcollina (changes requested):
I'm -1 with this change.

Arrow function signify that we are retaining `this`. Don't use it for anything else, as you might not want to retain `this`, and therefore increase memory usage.

Some of those closures could be refactored into top level functions too.

Uzlopak:
Sure. I am just happy that I split the PRs, so that it can be accepted in "bundles"...

PR #2672, opened 2024-01-30
enforce latin-1 encoding in source

Ethan-Arrowood:
Related to: https://github.com/nodejs/node/pull/51605

this replaces all `’` with `'` and the `ä` with `a` in a licensing comment.

Based on my research, the licensing comment change does not impact its validity. It would actually be more appropriate to have a complete License file along with the code instead of having it inlined.  

I also tried out the `--minify-whitespace` flag for esbuild. This would remove comments from `node:build`, but also remove all other whitespace. There does not seem to be an esbuild configuration option that only removes comments. We could consider another tool for that so we can retain other whitespace.

### DIFF
--- a/lib/fetch/body.js
+++ b/lib/fetch/body.js
@@ -36,7 +36,7 @@ function extractBody (object, keepalive = false) {
     stream = object
   } else if (isBlobLike(object)) {
     // 3. Otherwise, if object is a Blob object, set stream to the
-    //    result of running object’s get stream.
+    //    result of running object's get stream.
     stream = object.stream()
   } else {
     // 4. Otherwise, set stream to a new ReadableStream object, and set
@@ -87,7 +87,7 @@ function extractBody (object, keepalive = false) {
     // See: https://github.com/nodejs/node/blob/e46c680bf2b211bbd52cf959ca17ee98c7f657f5/lib/internal/url.js#L490
     // and https://github.com/nodejs/node/blob/e46c680bf2b211bbd52cf959ca17ee98c7f657f5/lib/internal/url.js#L1100
 
-    // Set source to the result of running the application/x-www-form-urlencoded serializer with object’s list.
+    // Set source to the result of running the application/x-www-form-urlencoded serializer with object's list.
     source = object.toString()
 
     // Set type to `application/x-www-form-urlencoded;charset=UTF-8`.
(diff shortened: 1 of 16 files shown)


KhafraDev (dismissed):
> If we remove those comments it'll make debugging fetch 10x harder.

We can no longer copy and paste these comments and find where they originate from in the spec (I do this frequently when updating the impl. or fixing bugs). Since esbuild should be removing comments, it's likely a bug there. We can alternatively add a post processing script that strips these only in the bundled file.

I'm very much a -1 against this change. It'll make maintaining fetch harder for little benefit.

Ethan-Arrowood:
Not a bug in esbuild: https://github.com/evanw/esbuild/issues/3117#issuecomment-1550565092

Ethan-Arrowood:
Just tried it and changing the character has no effect on search/find on the spec 😄

Ethan-Arrowood:
But I hear you. I agree too. I think this is an unnecessary change. I think we should minify the build and use source maps for proper stack traces. Not sure how Node.js internal deps handles that so I have to look into it further.

KhafraDev:
Internal deps are kinda a nightmare: they get their own version of require that doesn't work well with multiple files (which is why undici is bundled unfortunately).

KhafraDev:
The issue with the require we get there is that it only prepends 'internal' to the path, which means we'd have to replace every require in undici with `require('undici' + oldPath)` (which then gets resolved to `internal + undici + oldPath` which should theoretically work?). I did look into it a while back, when we had the --keep-names issue with the bundle, but quickly found out it was definitely the correct answer.

mcollina (approved):
lgtm

KhafraDev:
@Ethan-Arrowood I made an alternate PR that strips the comments from the node build only.

joyeecheung:
> they get their own version of require that doesn't work well with multiple files (which is why undici is bundled unfortunately).

I am pretty sure that can be special cased for undici, though I am not sure what undici uses as module id if not unbundled. The internal require will always be just minimal so nothing fancy like module path monkey patching would ever be supported, but prepending certain prefixes when the request comes from under `internal/deps/undici` wouldn't be that hard.

mcollina:
@Ethan-Arrowood @KhafraDev what's the status of this? There are conflicts now.

Ethan-Arrowood:
I think @KhafraDev preferred another methodology. Since he's modifying this source more than I am these days, I'm okay deferring to his preference 😄

PR #2673, opened 2024-01-30
strip comments from node build

KhafraDev:
this is a less invasive alternative to #2672

### DIFF
--- a/package.json
+++ b/package.json
@@ -71,7 +71,7 @@
     "docs"
   ],
   "scripts": {
-    "build:node": "npx esbuild@0.19.4 index-fetch.js --bundle --platform=node --outfile=undici-fetch.js --define:esbuildDetection=1 --keep-names",
+    "build:node": "npx esbuild@0.19.4 index-fetch.js --bundle --platform=node --outfile=undici-fetch.js --define:esbuildDetection=1 --keep-names && node scripts/strip-comments.js",
     "prebuild:wasm": "node build/wasm.js --prebuild",
     "build:wasm": "node build/wasm.js --docker",
     "lint": "standard | snazzy",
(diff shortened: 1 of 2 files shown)


KhafraDev:
Before:
// Returns response’s status.

After:
// Returns response?s status.

metcoder95 (approved):
LGTM

package.json
```diff
@@ -71,7 +71,7 @@
     "docs"
   ],
   "scripts": {
-    "build:node": "npx esbuild@0.19.4 index-fetch.js --bundle --platform=node --outfile=undici-fetch.js --define:esbuildDetection=1 --keep-names",
+    "build:node": "npx esbuild@0.19.4 index-fetch.js --bundle --platform=node --outfile=undici-fetch.js --define:esbuildDetection=1 --keep-names && node scripts/strip-comments.js",
```
Uzlopak, line 74: what about using `--charset=latin1` ?

https://esbuild.github.io/api/#charset

KhafraDev:
```
This flag does not apply to comments. I believe preserving non-ASCII data in comments should be fine because even if the encoding is wrong, the run time environment should completely ignore the contents of all comments. For example, the [V8 blog post](https://v8.dev/blog/scanner) mentions an optimization that avoids decoding comment contents completely. And all comments other than license-related comments are stripped out by esbuild anyway.
```

Uzlopak: Oki dokey

Uzlopak (approved):
I would just rename strip-comments.js to enforce-latin1.js or something like that, as strip-comments indicates that it removes comments from the file.

Ethan-Arrowood (approved):

mcollina (approved):
lgtm

PR #2643, opened 2024-01-26
Support building for externally shared js builtins

mochaaP:
## This relates to...

#2633
<https://src.fedoraproject.org/rpms/nodejs-undici/pull-request/3>

Reference:

https://github.com/nodejs/node/commit/ca5be26b318
https://github.com/nodejs/node/pull/44376

## Rationale

Initial support for loading unbundled module in `AddExternalizedBuiltin`.

- Reduces downstream distribution package size (by not shipping wasm twice and not base64-encoding it)
- Provides a cleaner stacktrace
- Easier to patch

## Changes

- Refactored build process to reflect above changes
- Move buffer decoding logic to encoded .js files
- Build loader for externally shared js builtins

### Features

Initial support for loading unbundled module in `AddExternalizedBuiltin`.

To enable this, pass `EXTERNAL_PATH=/path/to/global/node_modules/undici` to `build/wasm.js`.
You shall also pass this path to `--shared-builtin-undici/undici-path` in Node.js's `configure.py`.

### Bug Fixes

<https://bugzilla.redhat.com/show_bug.cgi?id=2259320>

### Breaking Changes and Deprecations

N/A

## Status

- [x] I have read and agreed to the [Developer's Certificate of Origin][cert]
- [x] Tested
- [x] Benchmarked (**optional**)
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

[cert]: https://github.com/nodejs/undici/blob/main/CONTRIBUTING.md

PR #2677, opened 2024-01-31
chore: refine esbuild & node detection

mochaaP:
When using the loader for external builtins, `esbuildDetection` is undefined. This commit defines `__UNDICI_IS_NODE__` on `globalThis` in the loader and deletes it after loading Undici. `esbuildDetection` has also been extracted as a variable at the top level of the module, to support deleting `__UNDICI_IS_NODE__` on `globalThis` to avoid polluting the global namespace.

Follow up of #2643.

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -66,6 +66,10 @@ const { webidl } = require('./webidl')
 const { STATUS_CODES } = require('node:http')
 const GET_OR_HEAD = ['GET', 'HEAD']
 
+const defaultUserAgent = typeof __UNDICI_IS_NODE__ !== 'undefined' || typeof esbuildDetection !== 'undefined'
+  ? 'node'
+  : 'undici'
+
 /** @type {import('buffer').resolveObjectURL} */
 let resolveObjectURL
 
@@ -1473,7 +1477,7 @@ async function httpNetworkOrCacheFetch (
   //    user agents should append `User-Agent`/default `User-Agent` value to
   //    httpRequest’s header list.
   if (!httpRequest.headersList.contains('user-agent', true)) {
-    httpRequest.headersList.append('user-agent', typeof esbuildDetection === 'undefined' ? 'undici' : 'node', true)
+    httpRequest.headersList.append('user-agent', defaultUserAgent)
   }
 
   //    15. If httpRequest’s cache mode is "default" and httpRequest’s header
--- a/build/wasm.js
+++ b/build/wasm.js
@@ -101,6 +101,8 @@ if (EXTERNAL_PATH) {
   writeFileSync(join(ROOT, 'loader.js'), `
 'use strict'
 
+globalThis.__UNDICI_IS_NODE__ = true
 module.exports = require('node:module').createRequire('${EXTERNAL_PATH}/loader.js')('./index-fetch.js')
+delete globalThis.__UNDICI_IS_NODE__
 `)
 }


mochaaP:
cc @khardix for https://bugzilla.redhat.com/show_bug.cgi?id=2261973

metcoder95 (approved):
LGTM, but I would like to have another pair of eyes from the team on this

KhafraDev (approved):
(same)

lib/fetch/index.js
```diff
@@ -65,6 +65,8 @@ const { webidl } = require('./webidl')
 const { STATUS_CODES } = require('node:http')
 const GET_OR_HEAD = ['GET', 'HEAD']
 
+const isNode = typeof __UNDICI_IS_NODE__ !== 'undefined' || typeof esbuildDetection !== 'undefined'
```
Uzlopak, line 68:
```suggestion
const defaultUserAgent = typeof __UNDICI_IS_NODE__ !== 'undefined' || typeof esbuildDetection !== 'undefined'
  ? 'node'
  : 'undici'
```

lib/fetch/index.js
```diff
@@ -1472,7 +1474,7 @@ async function httpNetworkOrCacheFetch (
   //    user agents should append `User-Agent`/default `User-Agent` value to
   //    httpRequest’s header list.
   if (!httpRequest.headersList.contains('user-agent', true)) {
-    httpRequest.headersList.append('user-agent', typeof esbuildDetection === 'undefined' ? 'undici' : 'node', true)
+    httpRequest.headersList.append('user-agent', isNode ? 'node' : 'undici', true)
```
Uzlopak, line 1477:
```suggestion
    httpRequest.headersList.append('user-agent', isNode ? 'node' : 'undici', true)
```
```suggestion
    httpRequest.headersList.append('user-agent', defaultUserAgent, true)
```

Uzlopak (changes requested):
I think my proposal is reasonable

Uzlopak:
@KhafraDev 
@metcoder95 

I wonder: In node 21, we have navigator.userAgent. Maybe in case of node, we should check if navigator.userAgent is set and use it, if not fallback to `nodejs/${process.versions.node.split('.')[0]}` ?

Uzlopak (approved):

metcoder95:
>I wonder: In node 21, we have navigator.userAgent. Maybe in case of node, we should check if navigator.userAgent is set and use it, if not fallback to nodejs/${process.versions.node.split('.')[0]} ?

Yeah, that SGTM 👍 
But maybe even wait for Node v22

mochaaP:
> But maybe even wait for Node v22

I agree. We could merge this first and make the UA change in v7, since I consider this a breaking change?

mcollina:
I would like to avoid breaking changes for Node.js as much as possible, as it would create multiple release lines for us to maintain and backport security fixes.

That's inescapable, so breaking changes will be batched. I'm not sure we would do a undici v7 before Node.js v22 is out, as it mostly depend on the amount of changes we want to ship.

mochaaP:
IMO, it's not _that_ important to change the UA. I couldn't think of possible benefits for changing the user agent, except for collecting metrics for node.js versions.

Uzlopak:
Yeah, lets not open that can of worms.

mochaaP:
Once this lands, we'd be happy to backport this in Fedora.

mcollina (approved):
lgtm

mcollina:
I'll land this after the security release.

mochaaP:
Hi @mcollina, seems this got missed in the security release?

mcollina:
The goal was to land it _after_ the release.

Issue #2681, opened 2024-01-31
Missed mock request callback


James1x0:
## This would solve...
This would give a cleaner look at what requests were not caught by the mockagent during a particular run

## The implementation should look like...
Long overdue, as discussed in #1600, we could add a callback to allow users to keep tabs on mismatched requests. I'll be adding a draft implementation soon.

## I have also considered...
Another option is to enable a flag to log mismatched requests.

PR #2682, opened 2024-01-31
feat: draft: adds a onMissedRequest hook for seeing missed mocks

James1x0:
## This relates to...
#2681

## Rationale
To enable easier debugging of the mockagent, as discussed in #1600 

## Changes
- docs: added an example of mockAgent onMissedRequest option
- test: added a basic test for onMissedRequest
- feat: added onMissedRequest
  - Added mock symbol for callback
  - Added logic in mock-utils dispatch to emit the hook

### Features
- added onMissedRequest hook for debugging mismatched requests

### Bug Fixes
No bugs fixed

No breaking changes or deprecations

## Status

- [x] Tested
- [ ] Documented (will add specific option documentation in case spec changes)
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/docs/api/MockAgent.md
+++ b/docs/api/MockAgent.md
@@ -38,6 +38,14 @@ const agent = new Agent()
 const mockAgent = new MockAgent({ agent })
 ```
 
+### Example - MockAgent instantiation with missed request callback
+
+```js
+import { MockAgent } from 'undici'
+
+const mockAgent = new MockAgent({ onMissedRequest: (error) => console.log(`A request wasn't handled: ${error.message}`) })
+```
+
 ## Instance Methods
 
 ### `MockAgent.get(origin)`
--- a/lib/mock/mock-agent.js
+++ b/lib/mock/mock-agent.js
@@ -11,7 +11,8 @@ const {
   kNetConnect,
   kGetNetConnect,
   kOptions,
-  kFactory
+  kFactory,
+  kOnMissedRequest
 } = require('./mock-symbols')
 const MockClient = require('./mock-client')
 const MockPool = require('./mock-pool')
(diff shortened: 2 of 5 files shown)


docs/api/MockAgent.md
```diff
@@ -38,6 +38,14 @@ const agent = new Agent()
 const mockAgent = new MockAgent({ agent })
 ```
 
+### Example - MockAgent instantiation with missed request callback
+
+```js
+import { MockAgent } from 'undici'
+
+const mockAgent = new MockAgent({ onMissedRequest: (error) => console.log(`A request wasn't handled: ${error.message}`) })
```
metcoder95, line 46: I like the idea pretty much, but not sure if it should be applied to the `Agent`. Have you considered the interceptor instead?

In that way, I imagine a usage like:

```js
mockPool.intercept({
  path: '/justatest?hello=there&see=ya',
  method: 'GET'
}).reply(200, { wont: 'match' }, {
  headers: { 'content-type': 'application/json' }
// being response the object passed through the intercept
}).onNotMatched((request, response) => console.log(request, response))
```

mcollina (commented):
Could you pass the request that was missed to the callback?

docs/api/MockAgent.md
James1x0: I think I like this usage instead as well. Any potential downsides to this syntax?

James1x0:
> Could you pass the request that was missed to the callback?

@mcollina Good idea. We'd want both the request and the error, right? My argument for the error is that the reasoning is baked into the `MockNotMatchedError` throws.

Also - How do you feel about @metcoder95's alternative syntax?

metcoder95:
SGTM, feel free to add it 🚀

docs/api/MockAgent.md
mcollina: I prefer this actually.

mcollina:
@James1x0 sorry for the delay. I prefer @metcoder95 syntax.

PR #2683, opened 2024-02-01
fix: webidl.brandcheck non strict should throw

Uzlopak:
It is very strange, that in case of strict: false brandCheck would not throw, but in strict cases it would throw. 

We use the non strict brandCheck e.g.
https://github.com/nodejs/undici/blob/0808a721570baaa10e6386a9e50bd34278814afe/lib/cookies/index.js#L56

The non strict check is basically a no op, as nothing happens with the result. I assume this was just an oversight. 

I would actually recommend for performance reasons to implement a webidl.brandCheckLax (or something like that) and make webidl.brandCheck only handle strict cases.

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/test/cookie/cookies.js
+++ b/test/cookie/cookies.js
@@ -599,3 +599,37 @@ test('Set-Cookie parser', () => {
   headers = new Headers()
   assert.deepEqual(getSetCookies(headers), [])
 })
+
+test('Cookie setCookie throws if headers is not of type Headers', () => {
+  class Headers {
+    [Symbol.toStringTag] = 'CustomHeaders'
+  }
+  const headers = new Headers()
+  assert.throws(
+    () => {
+      setCookie(headers, {
+        name: 'key',
+        value: 'Cat',
+        httpOnly: true,
+        secure: true,
+        maxAge: 3
+      })
+    },
+    new TypeError('Illegal invocation')
+  )
+})
+
+test('Cookie setCookie does not throw if headers is an instance of a custom Headers class', () => {
+  class Headers {
+    [Symbol.toStringTag] = 'Headers'
+    append () { }
+  }
+  const headers = new Headers()
+  setCookie(headers, {
+    name: 'key',
+    value: 'Cat',
+    httpOnly: true,
+    secure: true,
+    maxAge: 3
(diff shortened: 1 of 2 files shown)


KhafraDev (approved):

mcollina:
A test would be nice.

Uzlopak:
@mcollina 
Added test

test/cookie/cookies.js
```diff
@@ -599,3 +599,37 @@ test('Set-Cookie parser', () => {
...
+    new TypeError('Illegal invocation')
+  )
+})
+
+test('Cookie setCookie does not throw if headers is an instance of a custom Headers class', () => {
+  class Headers {
+    [Symbol.toStringTag] = 'Headers'
+    append () { }
+  }
+  const headers = new Headers()
```
KhafraDev, line 627:
```suggestion
  const headers = new globalThis.Headers()
```

Uzlopak: No bro, this is not the point of the brandCheck. But I will add one with the global this one. OK?

Uzlopak: done, good? :)

Uzlopak: it does now cover all cases of "Header"-classes

Should I add for all cookie functions these tests? getCookies, getSetCookie etc?

KhafraDev: It definitely is the point of the brand check. I implemented the non-strict mode to work with both the global headers and undici's headers as it didn't require internal state.

Uzlopak: Done

Uzlopak:
added for the other three cookie functions tests.

Uzlopak:
@KhafraDev 

What do you think about having an additional brandCheckLax webidl method for performance reasons?

KhafraDev:
👎

Uzlopak:
Ok, but this PR is fine so far, right?

I guess we could merge it then after the security release?

KhafraDev:
yep, waiting on the security release

PR #2560, opened 2023-12-29
perf: Improve FormData

tsctx:
## Benchmark

<details>
<summary>Benchmark Script</summary>

```js
import { bench, group, run } from 'mitata'
import { FormData } from '../lib/fetch/formdata.js'
import { FormData as FormDataCurrent } from '../lib/fetch/formdata2.js'
import symbols from '../lib/fetch/symbols.js'

const kState = symbols.kState

const data = [
  ['a', 'a'],
  ['b', 'b'],
  ['c', 'c'],
  ['d', 'd'],
  ['e', 'e'],
  ['f', 'f'],
  ['g', 'g'],
  ['h', 'h'],
  ['i', 'i']
]

/*
group('append FormData', () => {
  bench('current', () => {
    const fb = new FormDataCurrent()
    for (let i = 0; i < data.length; ++i) {
      fb.append(data[i][0], data[i][1])
    }
    return fb
  })
  bench('new', () => {
    const fb = new FormData()
    for (let i = 0; i < data.length; ++i) {
      fb.append(data[i][0], data[i][1])
    }
    return fb
  })
})
*/
group('set FormData', () => {
  bench('current', () => {
    const fb = new FormDataCurrent()
    for (let i = 0; i < data.length; ++i) {
      fb.set(data[i][0], data[i][1])
    }
    return fb
  })
  bench('new', () => {
    const fb = new FormData()
    for (let i = 0; i < data.length; ++i) {
      fb.set(data[i][0], data[i][1])
    }
    return fb
  })
})

group('set FormData (already append)', () => {
  const fb = new FormData()
  for (let i = 0; i < data.length; ++i) {
    fb.append(data[i][0], data[i][1])
  }

  const state = fb[kState]

  const stateLength = state.length

  function getNewFormDataCurrent () {
    const fb = new FormDataCurrent()
    fb[kState] = state.slice(0, stateLength)
    return fb
  }

  function getNewFormData () {
    const fb = new FormData()
    fb[kState] = state.slice(0, stateLength)
    return fb
  }
  bench('current', () => {
    const fb = getNewFormDataCurrent()
    for (let i = 0; i < data.length; ++i) {
      fb.set(data[i][0], data[i][1])
    }
    return fb
  })
  bench('new', () => {
    const fb = getNewFormData()
    for (let i = 0; i < data.length; ++i) {
      fb.set(data[i][0], data[i][1])
    }
    return fb
  })
})

group('get FormData', () => {
  const fbc = new FormDataCurrent()
[...]

PR #2692, opened 2024-02-02
perf: optimize Iterator

tsctx:
## Benchmark

Script from #2560

```
benchmark      time (avg)             (min … max)       p75       p99      p999
• FormData forEach
current     6,520 ns/iter   (3,800 ns … 2,551 µs)  4,800 ns   20.3 µs  461.1 µs
new         4,521 ns/iter   (2,933 ns … 12.42 µs)  5,790 ns  10.59 µs  12.42 µs

summary for FormData forEach
  new
   1.44x faster than current

• FormData@@iterator
current     4,391 ns/iter   (3,652 ns … 10.46 µs)  4,332 ns  8,333 ns  10.46 µs
new         3,538 ns/iter   (2,403 ns … 9,227 ns)  4,948 ns  8,872 ns  9,227 ns

summary for FormData@@iterator
  new
   1.24x faster than current
```


lib/fetch/headers.js
```diff
@@ -492,48 +492,33 @@ class Headers {
...
 
-    if (this[kGuard] === 'immutable') {
-      const value = this[kHeadersSortedMap]
-      return makeIterator(() => value, 'Headers',
-        'key+value')
-    }
-
     return makeIterator(
-      () => [...this[kHeadersSortedMap].values()],
+      () => this[kHeadersSortedMap],
```
tsctx, line 518: There is no need to do a shallow clone here.

lib/fetch/formdata.js
```diff
@@ -200,14 +204,25 @@ class FormData {
...
+  delete: kEnumerableProperty,
+  get: kEnumerableProperty,
+  getAll: kEnumerableProperty,
+  has: kEnumerableProperty,
+  set: kEnumerableProperty,
+  entries: kEnumerableProperty,
+  keys: kEnumerableProperty,
+  values: kEnumerableProperty,
+  forEach: kEnumerableProperty,
+  [Symbol.iterator]: { enumerable: false },
```
tsctx, line 225: This is needed but was not done.

lib/fetch/formdata.js
```diff
@@ -200,14 +204,25 @@ class FormData {
     }
 
     for (const [key, value] of this) {
-      callbackFn.apply(thisArg, [value, key, this])
+      callbackFn.call(thisArg, value, key, this)
```
tsctx, line 207: `Function#call` faster than `Function#apply`.

lib/fetch/formdata.js
```diff
@@ -158,29 +159,32 @@ class FormData {
...
+      'name', 'value'
     )
   }
 
   values () {
     webidl.brandCheck(this, FormData)
 
     return makeIterator(
-      () => this[kState].map(pair => [pair.name, pair.value]),
+      () => this[kState],
```
tsctx, line 184: This is the reason why the performance of the loop is poor.

lib/fetch/headers.js
```diff
@@ -492,48 +492,33 @@ class Headers {
   keys () {
     webidl.brandCheck(this, Headers)
 
-    if (this[kGuard] === 'immutable') {
-      const value = this[kHeadersSortedMap]
-      return makeIterator(() => value, 'Headers',
-        'key')
-    }
```
ronag, line 499: Why remove this?

lib/fetch/headers.js
ronag: Actually there is. The headers might change while being iterated?

mcollina:
Can you include some benchmarks?

lib/fetch/util.js
```diff
@@ -783,68 +787,61 @@ function makeIterator (iterator, name, kind) {
...
-
+      const { [keyIndex]: key, [valueIndex]: value } = values[index]
       // 12. Set object’s index to index + 1.
       object.index = index + 1
-
       // 13. Return the iterator result for pair and kind.
-      return iteratorResult(pair, kind)
+      // https://webidl.spec.whatwg.org/#iterator-result
+      let result
+      if (kind === 'key') {
```
metcoder95, line 797: Any reason to not use a switch in here?

lib/fetch/formdata.js
ronag: I actually think we need to create a copy...

ronag (changes requested):
I think per spec we need to have a copy that we pass to the iterator so that it does not see changes that occur after the iterator has been created. @KhafraDev wdyt?

lib/fetch/headers.js
tsctx: I don't think it's necessary to keep it, because it's a fast-path to avoid shallow cloning.
However, It can be left to avoid getter.

tsctx:
> I think per spec we need to have a copy that we pass to the iterator so that it does not see changes that occur after the iterator has been created. @KhafraDev wdyt?

What kind of case is it?

I don't know if this is the case, but if it is, the current implementation does not cover it.
It indicates that it will be reflected even if it was added after creation.

Also, since this clone is created each time next is invoked, it is clear that the clone is virtually unnecessary.

```js
const headers = new Headers([["a", "0"], ["b", "1"]])

const iterator = headers.entries();

headers.append("c", "2");

console.log([...iterator]); // [ [ 'a', '0' ], [ 'b', '1' ], [ 'c', '2' ] ]
```

lib/fetch/util.js
tsctx: Good catch!

lib/fetch/formdata.js
KhafraDev, line 225:
```suggestion

```

KhafraDev (approved):
We have tests for those cases and they're passing

PR #2693, opened 2024-02-03
test: improve cookie tests

Uzlopak:
Check against the specific error to ensue te error is not thrown for another reason


Uzlopak:
@metcoder95 

could you review + merge please :)

PR #2694, opened 2024-02-03
chore: improve test coverage for progressevent

Uzlopak:
Should improve the coverage reported by codecov for this file to 100%


KhafraDev:
this is tested in the WPTs... the WPT runner doesn't work with c8

Uzlopak:
The thing is I actually want to explore potential performance gains. For that I need anyway visible 100% test coverage to not introduce regressions. So if I anyway implement tests, then imho it should be ok to upstream them ;).

KhafraDev:
Effort may be allocated better to getting the WPTs to work with code coverage. Issues in the past with node coverage and FinalizationRegistry made it hard to do, but now it should be doable.

KhafraDev:
I should also mention that I'm planning to drop File/FileReader once node 18 is EOL

PR #2695, opened 2024-02-04
test: more information from debug tests

Uzlopak:
While investigating the code coverage issues, I got some strange error in the debug.js tests, as for some reason the stdout was containing colored output. This PR removes potential color in stdout, so that the tests wont fail because of that. 

Also in case the debug test is again flaky, the chunks should be now logged so we can maybe investigate further. 

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready


mcollina (approved):
lgtm

PR #2696, opened 2024-02-05
chore: remove mocha and chai

Uzlopak:
replaces chai and mocha with native test runner of node.

found a bug in Fetch Request.


lib/fetch/request.js
```diff
@@ -887,6 +889,50 @@ webidl.converters.Request = webidl.interfaceConverter(
...
+
+  // Note: the spec doesn't include async iterables,
+  // this is an undici extension.
+  if (V?.[Symbol.asyncIterator]) {
+    return V
+  }
+
+  return webidl.converters.XMLHttpRequestBodyInit(V)
+}
+
```
Uzlopak, line 935: This is copied from response.js. Imho it belongs to request and not reponse. Looking for a recommendation how to keep it DRY.

test/node-fetch/main.js
```diff
@@ -529,143 +511,144 @@ describe('node-fetch', () => {
...
   })
 
+  // TODO: Should we pass through the error message?
   it('should reject invalid json response', () => {
     const url = `${base}error/json`
     return fetch(url).then(res => {
-      expect(res.headers.get('content-type')).to.equal('application/json')
-      return expect(res.json()).to.eventually.be.rejectedWith(Error)
+      assert.strictEqual(res.headers.get('content-type'), 'application/json')
+      return assert.rejects(res.json(), SyntaxError)
```
Uzlopak, line 538: this exposes some information about the invalid json. Should we do that (Chrome does it too)

test/node-fetch/main.js
```diff
@@ -63,111 +49,110 @@ describe('node-fetch', () => {
...
+    assert.strictEqual(new Headers().toString(), '[object Headers]')
+    assert.strictEqual(new Response().toString(), '[object Response]')
+    assert.strictEqual(new Request(base).toString(), '[object Request]')
   })
-
+  // TODO Should we reflect the input?
   it('should reject with error if url is protocol relative', () => {
     const url = '//example.com/'
-    return expect(fetch(url)).to.eventually.be.rejectedWith(TypeError)
+    return assert.rejects(fetch(url), new TypeError('Failed to parse URL from //example.com/'))
```
Uzlopak, line 70: Should we reflect the input?

KhafraDev:
What is the issue in fetch? Why is the webidl stuff needed?

Uzlopak:
@KhafraDev 

webidl.converters.BodyInit is not defined in request.js but we are using it in this line

https://github.com/nodejs/undici/blob/a0812d8aed8752f76cdbd11f14693abd43689fa6/lib/fetch/request.js#L920

The unit test 'should throw error with GET/HEAD requests with body' was before just checking if a error is thrown. When i changed it to also check the message, it showed 'converter not defined'(or something similar). This was because BodyInit was not defined. 
Now it throws the correct error.

KhafraDev:
I'm assuming you imported `lib/fetch/request.js` and not `index.js`?

Uzlopak:
@KhafraDev Yes you are right. If I import it from index-fetch.js it will work as expected. But would this not mean, that we have the risk of breaking something if we import directly from fetch/request.js and not from index-fetch?

Uzlopak:
@KhafraDev 

Done, but feels error prone as the hoisting order is important.

KhafraDev:
It's either that or circular imports... I opted for this way. After all, you can't have Request by itself.

Uzlopak:
How about adding a simple 

if (webidl.converters.BodyInit === undefined) {
  throw new Error('should have required correctly')
}

Or in webidl.nullableConverter it should check if the coverter is a function, and throw accordingly.

This should result in no runtime overhead but improves the chance of finding a wrong require/import

KhafraDev:
Meh, realistically it shouldn't happen and tests should just require index.js. I don't see it being useful.

Uzlopak:
@KhafraDev 

But overall you like the changes?

KhafraDev:
yeah, I just have no time to do a full review on it

Uzlopak:
@KhafraDev 
No worries. I just want to make sure, that there is no big reason for rejection or something i can fix now as long as it is still "warm" in my head.

metcoder95 (approved):
LGTM

mcollina (approved):
lgtm

test/node-fetch/main.js
```diff
@@ -63,111 +49,110 @@ describe('node-fetch', () => {
...
   })
 
   it('should reject with error if url is relative path', () => {
     const url = '/some/path'
-    return expect(fetch(url)).to.eventually.be.rejectedWith(TypeError)
+    return assert.rejects(fetch(url), new TypeError('Failed to parse URL from /some/path'))
   })
 
+  // TODO: This seems odd
```
styfle, line 78: Can you explain why this seems odd?

I am seeing the following:

```
TypeError: fetch failed
    at node:internal/deps/undici/undici:12502:13
  [cause]: Error: unknown scheme
```

Perhaps this test should check for `unknown scheme` instead?

Uzlopak: Yeah, I think this is what made it odd to me.

KhafraDev: fetch should *always* reject with "fetch failed" and the cause should be the error. Basically, it's working properly.

styfle: I see. In that case we should remove the comment?

Uzlopak: I think are lacking the test that error.cause.message should be "unknown scheme".

PR #2703, opened 2024-02-06
fix: fix the linux patch

Uzlopak:
https://developer.mozilla.org/en-US/docs/Web/API/ReadableByteStreamController/desiredSize

> The value will be null if the stream has errored and 0 if it is closed.

<s>So when the stream was closed we tried to read again from it. </s>

<s>Well... at this point i dont know...</s>

I guess the problem was, that the readable stream can get suspended. So when the readable stream is unsuspended, and pull gets called a second time, than we  try to get again a reader. But you can have only one reader on the stream, thus it breaks. Now it gets the reader only once.

### DIFF
--- a/lib/fetch/index.js
+++ b/lib/fetch/index.js
@@ -1099,16 +1099,12 @@ function fetchFinale (fetchParams, response) {
 
     const byteStream = new ReadableStream({
       readableStream: transformStream.readable,
+      async start () {
+        this._bodyReader = this.readableStream.getReader()
+      },
       async pull (controller) {
-        // TODO(mcollina): removen this block, not sure why pull() is called twice
-        if (this.readableStream.locked) {
-          return
-        }
-
-        const reader = this.readableStream.getReader()
-
         while (controller.desiredSize >= 0) {
-          const { done, value } = await reader.read()
+          const { done, value } = await this._bodyReader.read()
 
           if (done) {
             queueMicrotask(() => readableStreamClose(controller))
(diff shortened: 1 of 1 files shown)


lib/fetch/index.js
```diff
@@ -1910,8 +1906,8 @@ async function httpNetworkFetch (
 
   // 11. Let pullAlgorithm be an action that resumes the ongoing fetch
   // if it is suspended.
-  const pullAlgorithm = () => {
-    fetchParams.controller.resume()
+  const pullAlgorithm = async () => {
+    await fetchParams.controller.resume()
```
Uzlopak, line 1910: While debugging I realized that we call it with an await but as pullAlgorithm is not a async function it basically "decouples" the promise logic.

https://github.com/nodejs/undici/blob/cf4d90dff47dd9d1be9241fe9a324f6b7f671dda/lib/fetch/index.js#L1941

So with this change the await pullAlgorithm will actually have something to await.

lib/fetch/index.js
```diff
@@ -2039,7 +2033,7 @@ async function httpNetworkFetch (
 
       // 9. If stream doesn’t need more data ask the user agent to suspend
       // the ongoing fetch.
-      if (!fetchParams.controller.controller.desiredSize) {
+      if (fetchParams.controller.controller.desiredSize <= 0) {
```
Uzlopak, line 2036: According to MDN desiredSize can go negative if the stream is oversaturized. !(-1) is false. And if desiredSize is 0, then we know we can not push data, anyway. 0 would actually mean, that the stream is closed. So is my change correct?

mcollina (approved):
lgtm

PR #2705, opened 2024-02-06
chore: upgrade llhttp to 9.2.0

### LINKED ISSUE #2678: HPE_INVALID_CHUNK_SIZE but works without any issue in curl, browser and other runtimes excluding Node.js
## Bug Description

Whenever I run `.text()` or try to read the body, I get and HPE_INVALID_CHUNK_SIZE error on Node.js on every requests made to my University's VPN domain (`u-vpn.unilim.fr`)

## Reproducible By

```typescript
const response = await fetch("https://u-vpn.unilim.fr/remote/logincheck", {
  "headers": {
    "accept": "*/*",
    "cache-control": "no-store, no-cache, must-revalidate",
    "content-type": "text/plain;charset=UTF-8",
    "if-modified-since": "Sat, 1 Jan 2000 00:00:00 GMT",
    "pragma": "no-cache",
    "Referer": "https://u-vpn.unilim.fr/remote/login?lang=en",
    "Referrer-Policy": "strict-origin-when-cross-origin"
  },
  // Fake credentials, still throws the error though
  "body": "ajax=1&username=a&realm=&credential=b",
  "method": "POST"
});

// Only happens here.
console.log(await r.text());
```

## Expected Behavior

Every other runtimes and browsers can give me the actual text response : `ret=0,redir=/remote/login?&err=sslvpn_login_permission_denied&lang=en`

## Environment

Ubuntu 22.04 LTS,
Node v20.11.0

### Additional context

I can still get the data by catching the error and doing error.cause.data but this data is truncated on long responses so not usable for me on further requests.

Uzlopak:
<s>To check if this closes #2678 you can run

```js
import { fetch } from './index-fetch.js'

const response = await fetch('https://u-vpn.unilim.fr/remote/logincheck', {
  headers: {
    accept: '*/*',
    'cache-control': 'no-store, no-cache, must-revalidate',
    'content-type': 'text/plain;charset=UTF-8',
    'if-modified-since': 'Sat, 1 Jan 2000 00:00:00 GMT',
    pragma: 'no-cache',
    Referer: 'https://u-vpn.unilim.fr/remote/login?lang=en',
    'Referrer-Policy': 'strict-origin-when-cross-origin'
  },
  // Fake credentials, still throws the error though
  body: 'ajax=1&username=a&realm=&credential=b',
  method: 'POST'
})

// Only happens here.
console.log(await response.text())

```
</s>
@mcollina 

Can your run some benchmarks on this please?

@ShogunPanda


hpe.mjs
```diff
@@ -0,0 +1,19 @@
...
+    pragma: 'no-cache',
+    Referer: 'https://u-vpn.unilim.fr/remote/login?lang=en',
+    'Referrer-Policy': 'strict-origin-when-cross-origin'
+  },
+  // Fake credentials, still throws the error though
+  body: 'ajax=1&username=a&realm=&credential=b',
+  method: 'POST'
+})
+
+// Only happens here.
```
Uzlopak, line 18: now it does throw anymore

lib/client.js
```diff
@@ -555,6 +555,13 @@ class Parser {
 
     this.llhttp = exports
     this.ptr = this.llhttp.llhttp_alloc(constants.TYPE.RESPONSE)
+
+    // this.llhttp.llhttp_set_lenient_chunked_length(this.ptr, 1)
+    // this.llhttp.llhttp_set_lenient_headers(this.ptr, 1)
+    // this.llhttp.llhttp_set_lenient_keep_alive(this.ptr, 1)
+    // this.llhttp.llhttp_set_lenient_transfer_encoding(this.ptr, 1)
+    this.llhttp.llhttp_set_lenient_spaces_after_chunk_size(this.ptr, 1)
```
Uzlopak, line 563: we could activate all lenien flags. but actually we needed only `llhttp_set_lenient_spaces_after_chunk_size` How about hiding it behind a n optional field?

metcoder95: What would be the use cases of it?

Uzlopak: lenient flag?

mcollina: I would set all of those to `true`. Wdyt @ShogunPanda?

ShogunPanda: I agree on both; let's create a `insecureHTTPParser` flag like we do for node. If enabled, it will set all flags to `true`.

See:

https://github.com/nodejs/node/blob/8a41d9b636be86350cd32847c3f89d327c4f6ff7/lib/_http_client.js#L800
https://github.com/nodejs/node/blob/main/src/node_http_parser.cc#L94C1-L98C63

ShogunPanda: @Uzlopak @mcollina  Do you want to do this or do we land this PR (with no lenient flags set) and I'll do the work in another PR?

Uzlopak: I changed this PR to only upgrade llhttp. The changes in client.js are now remove.

I deactivated two tests. Are these regressions or is llhttp handling these two cases more gracefully?

If we upgrade llhttp then I think your PR should be following immediatly? We could also park this PR in a different branch in this repo. Then we would not have heavy pressure.

ShogunPanda: The two PR are independent: when talking to HTTP compliant server undici will still work. 
I'll try to get this PR today.

ShogunPanda (approved):
LGTM!

Uzlopak:
fixed the tests

imho ready for merge.

Uzlopak:
@mcollina

Maybe we should first release a undici 6.6.2 before we merge this?

ShogunPanda:
I'm still not sure about the right semver change but yes, let's merge this first.

KhafraDev (approved):

metcoder95 (approved):

ShogunPanda:
Sure, I'll release llhttp after lunch.

ShogunPanda:
@Uzlopak llhttp 9.2.0. has just been released. You can quickly rework this PR and then we're ready to go.

Uzlopak:
@ShogunPanda 

Thanks. I just patched the version in the header file.

mcollina (approved):
lgtm

I think we should add support/expose the leniency flags, and potentially read those values from node.js core itself. That can be done in a follow-up PR.

Uzlopak:
@mcollina 
@ShogunPanda 
@metcoder95 
@KhafraDev 

Target changed to next branch

mcollina (approved):
lgtm

PR #2708, opened 2024-02-06
Add support for passing iterable objects as headers

### LINKED ISSUE #2700: Issue with undici.request and Headers Global Object
## Bug Description
Undici's `request` function appears to be skipping headers that are constructed using the `Headers` global object.

## Reproducible By

1. Create an instance of `Headers` class.
2. Add header with set/append.
3. Pass headers instance to request.

```js
const headers = new Headers()
headers.append('random-header', 'hello world')

const res = await request('http://localhost:3001', { headers })
```

Full reproducable example can be found in following [repo](https://github.com/JaoodxD/undici-request-headers)

## Expected Behavior

I expect the `undici.request` function to treat instances of the `Headers` class the same way it handles plain JavaScript objects passed to options.  
Specifically, I anticipate that the headers, when constructed using the `Headers` class, should be seamlessly attached to the HTTP request.

## Logs & Screenshots

```
{
  host: 'localhost:3000',
  connection: 'keep-alive',
  'random-header': 'hello world' // passed as plain object
}
{ host: 'localhost:3001', connection: 'keep-alive' } // as Headers instance
```

## Environment
OS: Windows 10
Nodejs: v20.8.0
undici: v6.6.1

JaoodxD:
## This relates to 

Resolves #2700 

## Rationale

Improve flexibility and user expirience by allowing directly passing iterable objects as headers to `undici.request`.  
It will add support for [Headers](https://developer.mozilla.org/en-US/docs/Web/API/Headers), [Maps](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Map) and custom objects with [Symbol.iterator](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/iterator) that yields the key-value pairs.

## Changes

 * add for..of iterating in core/request.js

### Features

 * Support for iterable object headers

### Bug Fixes

N/A

N/A

## Status

- [x] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -174,7 +174,7 @@ class Request {
         }
       } else {
         const keys = Object.keys(headers)
-        for (let i = 0; i < keys.length; i++) {
+        for (const key of keys) {
           const key = keys[i]
           processHeader(this, key, headers[key])
         }
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -169,13 +169,12 @@ class Request {
           if (!Array.isArray(header) || header.length !== 2) {
             throw new InvalidArgumentError('headers must be in key-value pair format')
           }
-   const [key, value] = header;
+          const [key, value] = header
           processHeader(this, key, value)
         }
       } else {
         const keys = Object.keys(headers)
         for (const key of keys) {
-          const key = keys[i]
           processHeader(this, key, headers[key])
         }
       }
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -166,6 +166,9 @@ class Request {
     } else if (headers && typeof headers === 'object') {
       if (headers[Symbol.iterator]) {
         for (const header of headers) {
+          if (!Array.isArray(header) || header.length !== 2) {
+            throw new InvalidArgumentError('headers must be in key-value pair format')
+          }
           const key = header[0]
           const value = header[1]
           processHeader(this, key, value)
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -169,8 +169,7 @@ class Request {
           if (!Array.isArray(header) || header.length !== 2) {
             throw new InvalidArgumentError('headers must be in key-value pair format')
           }
-          const key = header[0]
-          const value = header[1]
+   const [key, value] = header;
           processHeader(this, key, value)
         }
       } else {
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -164,10 +164,18 @@ class Request {
         processHeader(this, headers[i], headers[i + 1])
       }
     } else if (headers && typeof headers === 'object') {
-      const keys = Object.keys(headers)
-      for (let i = 0; i < keys.length; i++) {
-        const key = keys[i]
-        processHeader(this, key, headers[key])
+      if (headers[Symbol.iterator]) {
+        for (const header of headers) {
+          const key = header[0]
+          const value = header[1]
+          processHeader(this, key, value)
+        }
+      } else {
+        const keys = Object.keys(headers)
+        for (let i = 0; i < keys.length; i++) {
+          const key = keys[i]
+          processHeader(this, key, headers[key])
+        }
       }
     } else if (headers != null) {
       throw new InvalidArgumentError('headers must be an object or an array')
--- a/test/request.js
+++ b/test/request.js
@@ -2,7 +2,7 @@
 
 const { createServer } = require('node:http')
 const { test } = require('tap')
-const { request } = require('..')
+const { request, errors } = require('..')
 
 test('no-slash/one-slash pathname should be included in req.path', async (t) => {
   const pathServer = createServer((req, res) => {
@@ -248,7 +248,7 @@ test('DispatchOptions#reset', scope => {
 })
 
 test('Should include headers from iterable objects', scope => {
-  scope.plan(3)
+  scope.plan(4)
 
   scope.test('Should include headers built with Headers global object', async t => {
     const server = createServer((req, res) => {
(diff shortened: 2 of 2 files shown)


JaoodxD:
Hey @ronag and @KhafraDev, I've got a pull request up that I'd appreciate your eyes on. Your feedback would be incredibly helpful!

mcollina (approved):
lgtm

as far as I recall, Headers had some special handling logic for cookies. Can you check that everything works correctly with those?

lib/core/request.js
```diff
@@ -164,10 +164,18 @@ class Request {
         processHeader(this, headers[i], headers[i + 1])
       }
     } else if (headers && typeof headers === 'object') {
-      const keys = Object.keys(headers)
-      for (let i = 0; i < keys.length; i++) {
-        const key = keys[i]
-        processHeader(this, key, headers[key])
+      if (headers[Symbol.iterator]) {
+        for (const header of headers) {
```
KhafraDev, line 168: there should be an error if header isn't an array or its length isn't 2

JaoodxD:
> lgtm
> 
> as far as I recall, Headers had some special handling logic for cookies. Can you check that everything works correctly with those?

Thank you for noticing potencial issue, will check it out.

lib/core/request.js
JaoodxD: Is there necessity to explicitly mention array of two elements or could I just leave something like `Headers must be in key-value pair format.`?

KhafraDev: your suggestion is good

JaoodxD: Fixed

JaoodxD:
> lgtm
> 
> as far as I recall, Headers had some special handling logic for cookies. Can you check that everything works correctly with those?

What I've found is `Headers` is filtering out and forbid to access to `Set-Cookie` header, but this is true only for client-side javascript.  
Done some checks and everything in this case seems absolutely fine on server-side for node.  
```js
const headers = new Headers()
headers.append('Set-Cookie', 'key=value')
headers.append('Set-Cookie', 'key2=value2')

for (const [key, value] of headers) {
  console.log(key, value)
}

console.log(headers.getSetCookie())
```
Should I look for any other potential issues with cookies?

KhafraDev (approved):

lib/core/request.js
```diff
@@ -164,10 +164,21 @@ class Request {
...
+          if (!Array.isArray(header) || header.length !== 2) {
+            throw new InvalidArgumentError('headers must be in key-value pair format')
+          }
+          const key = header[0]
+          const value = header[1]
```
mertcanaltin, line 173: I wonder if we can use it this way
```suggestion
   const [key, value] = header;
```

lib/core/request.js
```diff
@@ -164,10 +164,21 @@ class Request {
...
+          processHeader(this, key, value)
+        }
+      } else {
+        const keys = Object.keys(headers)
+        for (let i = 0; i < keys.length; i++) {
```
mertcanaltin, line 178: would it be healthier?
```suggestion
        for (const key of keys) {
```

lib/core/request.js
JaoodxD: Long story short, this approach is absolutely fine.
For a long time I've been thaught that your suggestion is slower due to internal iterator creation.
Now I double checked and find out that v8 devs already fixed it.
Everything is good 👍

mertcanaltin: It's great 🚀

lib/core/request.js
JaoodxD: It's a bit debatable.  
The main purpose of this PR is to introduce a new feature. Making changes to code that isn't directly related feels like it may violate the principle of separation of concerns.

mertcanaltin: I agree with you, it will be more performant to use it this way, can we try it if it does not pose any problem?

mertcanaltin: The for...of loop assigns each element of the array directly to the key variable and you don't need to check the indexes as you traverse the array. This makes the code more readable and reduces the risk of errors.

With a for loop, you have to check the index and increment the index at each step. This type of loop can lead to potential sources of errors. Also, if the number of elements of the array is very large, keys.length must be calculated at each loop step, which negatively affects performance.

mertcanaltin: I did some research, I hope it will be useful to us, you can see it in my comment above.

JaoodxD: That's reasonable. Let's add it.

mertcanaltin (approved):
LGTM! 🚀

lib/core/request.js
mertcanaltin: Thank you very much for your contribution

mcollina (approved):
lgtm

lforst:
Hi, we got a [bug report](https://github.com/getsentry/sentry-javascript/issues/10936) that seems strongly related to this change. I generally think it is a good change, however, the thought struck me that this might have been a breaking change (at least in combination with `diagnostics_channel`). Should we as library maintainers be more defensive moving forward?

jessezhang91:
I traced this a bit more and recognized it as related to this PR:
https://github.com/nodejs/undici/pull/2874

I *think* this PR is just related to the chain of updates.

KhafraDev:
Is there a minimal repro for the issue? It should be fixed on our end.

lforst:
@KhafraDev What I am referring to is not really a bug, so I am not sure it can be fixed. I am simply saying that this change should potentially have gone into a Node.js major.

But looking at the situation again, Node.js 20 still seems to be on undici 5 so was probably a bit too eager to comment here. As long as this feature is only lands in Node as part of a major, everything is fine. Sorry for the distraction!

KhafraDev:
I would consider a breaking change in a minor release of undici to be a bug. If you provide a minimal repro I'd be willing to take a look.

jessezhang91:
@KhafraDev here's a simple repro demonstrating the change that occurred between 6.6.2 and 6.7.0.

```
import 'undici';
import ds from 'diagnostics_channel';

ds.subscribe('undici:request:create', ({ request }) => {
  console.log(typeof request.headers);
  console.log(JSON.stringify(request.headers));
});

  headers: { 'X-Foo': 'bar' }
});
```


```
$ npm install undici@6.6.2 && node index.js

changed 1 package, and audited 3 packages in 400ms

found 0 vulnerabilities
string
"X-Foo: bar\r\naccept: */*\r\naccept-language: *\r\nsec-fetch-mode: cors\r\nuser-agent: node\r\naccept-encoding: br, gzip, deflate\r\n"
```

```
$ npm install undici@6.7.0 && node index.js

changed 1 package, and audited 3 packages in 408ms

found 0 vulnerabilities
object
["X-Foo","bar","accept","*/*","accept-language","*","sec-fetch-mode","cors","user-agent","node","accept-encoding","br, gzip, deflate"]
```

metcoder95:
Yeah, I can confirm this bug; this changes the contract and might be better if we aim it to next `major` instead.
cc: @ronag @mcollina

Issue #2675, opened 2024-01-30
HTTP2 Requests hang upon receiving 'Stream closed with error code NGHTTP2_INTERNAL_ERROR'


jackschu:
## Bug Description

If node closes a http2stream with `NGHTTP2_INTERNAL_ERROR`, the `await dispatch` in httpNetworkFetch() in fetch/index.js will never resolve

## Reproducible By

The tricky part about reproducing is getting node to throw this error. Reproductions seem related to sending many large http2 post requests via the fetch api to the same endpoint while the event loop is busy.

## Expected Behavior

The promise is rejected rather than left unfulfilled.

## Logs & Screenshots

With NODE_DEBUG=http2 I can see code 2 (NGHTTP2_INTERNAL_ERROR) being thrown
```
[2024-01-30T22:18:35.084Z] [phase_0 saas_local_webhook] [err] HTTP2 14: Http2Stream 3 [Http2Session client]: closed with code 0, closed false, readable false
[2024-01-30T22:18:35.084Z] [phase_0 saas_local_webhook] [err] HTTP2 14: Http2Stream 3 [Http2Session client]: destroying stream
[2024-01-30T22:18:40.109Z] [phase_0 saas_local_webhook] [err] HTTP2 14: Http2Stream 3 [Http2Session client]: shutting down writable on _final
[2024-01-30T22:18:42.118Z] [phase_0 saas_local_webhook] [err] HTTP2 14: Http2Stream 3 [Http2Session client]: closed with code 2, closed false, readable true
[2024-01-30T22:18:42.118Z] [phase_0 saas_local_webhook] [err] HTTP2 14: Http2Stream 3 [Http2Session client]: destroying stream
```

If i add a call to `errorRequest` [here in writeH2()](https://github.com/nodejs/undici/blob/f1d7ada7eb2292932065a7c9ab23569b8a6e1720/lib/client.js#L1817) i can see the following error and my expected behavior of rejecting the promise
```
[2024-01-30T22:18:42.122Z] [phase_0 saas_local_webhook] [err] [1/30/2024, 5:18:41 PM] [PLATFORM_FETCH] Error resulted from fetch for request POST 
{
  "stack": "TypeError: fetch failed\n    at Object.fetch9 (/nix/store/qwj14ylfz8y3ji41rgzssgk25i4wp749-esbuild_node/depengine_worker.js:31730:19)\n    at processTicksAndRejections (node:internal/process/task_queues:96:5)",
  "message": "fetch failed",
[...]

PR #2710, opened 2024-02-07
fix(http2): Fixed HTTP2 stream closure handling

mertcanaltin:
This PR addresses an issue related to handling the closure of HTTP2 streams with the NGHTTP2_INTERNAL_ERROR code. Now, when HTTP2 streams are closed with the NGHTTP2_INTERNAL_ERROR code, the promise is rejected instead of being left unresolved.

#2675

### DIFF
--- a/lib/handler/retry-handler.js
+++ b/lib/handler/retry-handler.js
@@ -1,3 +1,4 @@
+'use strict'
 const assert = require('node:assert')
 
 const { kRetryHandlerDefaultRetry } = require('../core/symbols')
@@ -37,7 +38,7 @@ class RetryHandler {
       retry: retryFn ?? RetryHandler[kRetryHandlerDefaultRetry],
       retryAfter: retryAfter ?? true,
       maxTimeout: maxTimeout ?? 30 * 1000, // 30s,
-      timeout: minTimeout ?? 500, // .5s
+      minTimeout: minTimeout ?? 500, // .5s
       timeoutFactor: timeoutFactor ?? 2,
       maxRetries: maxRetries ?? 5,
       // What errors we should retry
@@ -104,7 +105,7 @@ class RetryHandler {
     const { method, retryOptions } = opts
     const {
       maxRetries,
-      timeout,
+      minTimeout,
       maxTimeout,
       timeoutFactor,
       statusCodes,
@@ -114,7 +115,7 @@ class RetryHandler {
     let { counter, currentTimeout } = state
 
     currentTimeout =
-      currentTimeout != null && currentTimeout > 0 ? currentTimeout : timeout
+      currentTimeout != null && currentTimeout > 0 ? currentTimeout : minTimeout
 
     // Any code that is not a Undici's originated and allowed to retry
     if (
--- a/test/retry-handler.js
+++ b/test/retry-handler.js
@@ -706,3 +706,75 @@ test('should not error if request is not meant to be retried', async t => {
 
   await t.completed
 })
+
+test('Should be able to properly pass the minTimeout to the RetryContext when constructing a RetryCallback function', async t => {
+  t = tspl(t, { plan: 2 })
+
+  let counter = 0
+  const server = createServer()
+  server.on('request', (req, res) => {
+    switch (counter) {
+      case 0:
+        res.writeHead(500)
+        res.end('failed')
+        return
+      case 1:
+        res.writeHead(200)
+        res.end('hello world!')
+        return
+      default:
+        t.fail()
+    }
+  })
+
+  const dispatchOptions = {
+    retryOptions: {
+      retry: (err, { state, opts }, done) => {
+        counter++
+        t.strictEqual(opts.retryOptions.minTimeout, 100)
+
+        if (err.statusCode === 500) {
+          return done()
+        }
+
+        return done(err)
+      },
+      minTimeout: 100
+    },
+    method: 'GET',
+    path: '/',
+    headers: {
+      'content-type': 'application/json'
+    }
+  }
+
+  server.listen(0, () => {
+    const client = new Client(`http://localhost:${server.address().port}`)
+    const handler = new RetryHandler(dispatchOptions, {
+      dispatch: client.dispatch.bind(client),
+      handler: new RequestHandler(dispatchOptions, (err, data) => {
+        t.ifError(err)
+      })
+    })
+
+    after(async () => {
+      await client.close()
+      server.close()
+
+      await once(server, 'close')
+    })
+
+    client.dispatch(
+      {
+        method: 'GET',
+        path: '/',
+        headers: {
+          'content-type': 'application/json'
+        }
+      },
+      handler
+    )
+  })
+
+  await t.completed
+})


lib/fetch/index.js
```diff
@@ -1892,6 +1892,11 @@ async function httpNetworkFetch (
       response = makeResponse({ status, statusText, headersList })
     }
   } catch (err) {
+    if (err.name === 'NGHTTP2_INTERNAL_ERROR') {
+      // Reject pending Promise when HTTP2 stream is closed
+      return Promise.reject(new Error('HTTP2 stream closed with NGHTTP2_INTERNAL_ERROR'))
+    }
```
ronag, line 1898: This is not part of spec. I think this issue should be resolved in undici core. Not fetch.

ronag (changes requested):
I think this issue should be resolved in undici core. Not fetch.

mertcanaltin:
> I think this issue should be resolved in undici core. Not fetch.

greetings I have taken some steps, how can I use this in chore, I would be very happy if you have any suggestions @ronag

lib/core/errors.js
```diff
@@ -195,6 +195,15 @@ class RequestRetryError extends UndiciError {
   }
 }
 
+class NGHTTP2InternalError extends UndiciError {
```
metcoder95, line 198: Let's scope it to only `HTTP2`, so maybe we can do something like `HTTP2Error`?
So we can reuse it in other scenarios that might need similar approach

test/nghttp2-internal-error.js
metcoder95, line 1: Feel free to add them to `test/http2.js` 👍

metcoder95:
Take a look at https://github.com/nodejs/undici/blob/0a069ab1f2d111b8e74f2d444d7f5678e302f39d/lib/client.js#L1639, there we manage everything related to `http2`

mertcanaltin:
@metcoder95 tried a few steps thanks for your support

test/http2.js
```diff
@@ -1244,6 +1244,35 @@ test('The h2 pseudo-headers is not included in the headers', async t => {
   t.equal(response.headers[':status'], undefined)
 })
 
+test('HTTP/2 stream closed with NGHTTP2_INTERNAL_ERROR', async (t) => {
+  t.plan(1)
+
+  // Create a new client instance
+  const client = new Client('http://localhost:8000')
```
jackschu, line 1251: This test fails (you can repro by running `npx tap -t 4 -g NGHTTP2_INTERNAL_ERROR "test/http2.js"`)

This is because the client is trying to hit a non-existent server
```suggestion
  const server = createSecureServer(pem)

  server.on('stream', (stream, headers) => {
    stream.respond({
      ':status': 200,
    })
    stream.end('hello h2!')
  })

  server.listen(0)
  await once(server, 'listening')
  t.teardown(server.close.bind(server))
  
  const client = new Client(`https://localhost:${server.address().port}`, {
  connect: {
    rejectUnauthorized: false,
  },
  allowH2: true,
})
```

The result will be that the test fails with timeout because of the reported issue that this request will never reject or resolve (see other comment)

jackschu (changes requested):

test/http2.js
mertcanaltin: thanks I will update this immediately

lib/client.js
```diff
@@ -1840,7 +1841,9 @@ function writeH2 (client, session, request) {
   })
 
   stream.once('frameError', (type, code) => {
-    const err = new InformationalError(`HTTP/2: "frameError" received - type ${type}, code ${code}`)
+    const errorMessage = `HTTP/2: Frame error received - type: ${type}, code: ${code}.`
```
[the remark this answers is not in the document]
mertcanaltin: that would fit our purpose? @jackschu 
```js
stream.once('error', function (err) {
    if (client[kHTTP2Session] && !client[kHTTP2Session].destroyed && !this.closed && !this.destroyed) {
        errorRequest(client, request, err);
        h2State.streams -= 1;
        util.destroy(stream, err);
    }
});

stream.once('frameError', (type, code) => {
    const errorMessage = `HTTP/2: Frame error received - type: ${type}, code: ${code}.`;
    const err = new HTTP2Error(errorMessage);
    errorRequest(client, request, err);
    
    if (client[kHTTP2Session] && !client[kHTTP2Session].destroyed && !this.closed && !this.destroyed) {
        h2State.streams -= 1;
        util.destroy(stream, err);
    }
});
```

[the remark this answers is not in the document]
jackschu: I'm not sure what the right patch is to be honest, when I saw this in prod `this.closed` was true and the promise was left unresolved

mertcanaltin:
@metcoder95 @ronag @KhafraDev  I would like to update this place, I would appreciate your comments 🙏

lib/client.js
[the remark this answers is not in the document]
metcoder95: Oh wow, sorry. Too late to the party

Let's give it a try, the `frameError` should also error the request that is tight to the given stream that received that error; can you also add some testing around it @mertcanaltin ?

ronag:
This is outside of my expertise. I trust @metcoder95 here.

test/http2.js
```diff
@@ -1265,6 +1265,53 @@ test('The h2 pseudo-headers is not included in the headers', async t => {
   t.strictEqual(response.headers[':status'], undefined)
 })
 
+test('HTTP/2 stream closed with NGHTTP2_INTERNAL_ERROR', async (t) => {
```
Uzlopak, line 1268: you wrote a tap test. we switched to node native test runner

test/http2.js
```diff
@@ -1265,6 +1265,53 @@ test('The h2 pseudo-headers is not included in the headers', async t => {
...
+  t.plan(1)
```
Uzlopak, line 1269:
```suggestion
  t = tspl(t, { plan: 1 })
```

test/http2.js
```diff
@@ -1265,6 +1265,53 @@ test('The h2 pseudo-headers is not included in the headers', async t => {
...
+  server.on('stream', (stream, headers) => {
+    stream.respond({
+      ':status': 200
+    })
+    stream.end('hello h2!')
+  })
+
+  server.listen(0)
+  await once(server, 'listening')
+  t.teardown(server.close.bind(server))
```
Uzlopak, line 1283:
```suggestion
  after(() => server.close())
```

mertcanaltin: thank you ı will used

test/http2.js
```diff
@@ -1265,6 +1265,53 @@ test('The h2 pseudo-headers is not included in the headers', async t => {
...
+    })
+    // If the request succeeds, fail the test
+    t.fail('Request should not succeed')
+  } catch (err) {
+    t.equal(
+      err.message,
+      'HTTP2 internal error'
+    )
+    t.equal(err.code, 'UND_ERR_NGHTTP2_INTERNAL')
+  }
```
Uzlopak, line 1312:
```suggestion
  }
  await t.completed
```

mertcanaltin:
I'll tidy up here, I blew my codes while synchronizing 😢

PR #2711, opened 2024-02-07
fix: HTTP2 tweaks

### LINKED ISSUE #2704: Cannot read properties of null (reading 'destroyed')
## Bug Description

Very occasionally, undici will throw this error as an uncaught rejection, causing the node process to crash. 
The stack trace is as follows:
```
TypeError: Cannot read properties of null (reading 'destroyed')
    at _resume (/node_modules/undici/lib/client.js:1429:16)
    at resume (/node_modules/undici/lib/client.js:1334:3)
    at process.processTicksAndRejections (node:internal/process/task_queues:81:21)
```

## Reproducible By

I am unable to replicate this issue consistently, but this issue pops up frequently, sometimes multiple times in a week.

## Environment

Debian 12.4, node v18.19.0, undici 6.5.0

### Additional context

```
    if (!socket && !client[kHTTP2Session]) {
      connect(client)
      return
    }

    if (socket.destroyed || socket[kWriting] || socket[kReset] || socket[kBlocking]) {
      return
    }
```

The error occurs in the condition of the second if clause, where socket is null. I believe that 

1. client[kHTTP2Session] is not being set to null on socket close, despite socket being set to null. I see another function in client.js, `onUpgrade`, where socket is being set to null but client[kHTTP2Session] is not touched, but I believe this should not be called when client[kHTTP2Session] is truthy.
2. Perhaps first if clause should be rolled back to `if (!socket) {` to account for null socket even if `client[kHTTP2Session]` is truthy, so that http2 connect logic is triggered if socket is null. I'm not entirely sure of the reason this if clause was changed.

metcoder95:
## This relates to...

- Closes #2704

## Changes

- refactor: remove redundant checks
- fix: handle half-closed states

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -808,6 +808,7 @@ class Parser {
       .removeListener('close', onSocketClose)
 
     client[kSocket] = null
+    client[kHTTP2Session] = null
     client[kQueue][client[kRunningIdx]++] = null
     client.emit('disconnect', client[kUrl], [client], new InformationalError('upgrade'))
 
@@ -1421,7 +1422,7 @@ function _resume (client, sync) {
       return
     }
 
-    if (!socket && !client[kHTTP2Session]) {
+    if (!socket) {
       connect(client)
       return
     }
--- a/lib/client.js
+++ b/lib/client.js
@@ -1797,7 +1797,25 @@ function writeH2 (client, session, request) {
   })
 
   stream.once('end', () => {
-    request.onComplete([])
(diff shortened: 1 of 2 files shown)


ronag (approved):
Removing http2 specific logic from core is great.

Uzlopak (approved):
LGTM

Issue #1748, opened 2022-10-31
Add support for Priority Hints


pmeenan:
The Priority Hints spec is in the process of being merged into the fetch and HTML specs and allows for site developers to hint at the relative fetch priority for various resources.

The main changes are:

Adds a "priority" to fetch's Request and RequestInit classes that uses a string enum of "auto", "high" or "low"
Adds a "fetchPriority" attribute to the IMG, PICTURE, SCRIPT, LINK and IFRAME html elements with the same enum values
Existing spec: https://wicg.github.io/priority-hints/

This was shipped in Chromium 103 and Mozilla is working on an implementation so it is now in the process of being moved into the HTML and fetch specs.

For node, the main use case is on the fetch interface to pass through prioritization to the underlying networking stack.

mcollina:
This would be an interesting use case to support. While Node.js does not offer much in terms of prioritization from the network stack, Undici queue logic is a prime candidate for this.

My take is that a high priority request jumps at the top of the queue and it's not pipelineable.

Wdyt?

KhafraDev:
I looked into this but I'm not super familiar with undici's queuing logic. From my understanding, after a request is added to the queue, it's sychronously queued & removed from the queue. So for fetch, assuming that the dispatcher is a default `Agent`, a priority option wouldn't have an affect, as there isn't a time where multiple requests are in the queue.

Test code:
```js
import { fetch } from './index.js'

const result = await Promise.race([
  fetch('https://www.google.com/1').then(() => 'first'),
  fetch('https://www.google.com/2').then(() => 'second')
])

console.log(result)
```

(of course which request resolves first has no relevance here)

Where the queue looks like:

```js
[
  Request {
    headersTimeout: 300000,
    bodyTimeout: 300000,
    throwOnError: false,
    method: 'GET',
    body: null,
    completed: false,
    aborted: false,
    upgrade: null,
    path: '/1',
    origin: 'https://www.google.com',
    idempotent: true,
    blocking: false,
    host: null,
    contentLength: null,
    contentType: null,
    headers: 'accept: */*\r\n' +
      'accept-language: *\r\n' +
      'sec-fetch-mode: cors\r\n' +
      'user-agent: undici\r\n' +
      'accept-encoding: br, gzip, deflate\r\n',
    servername: null,
    [Symbol(handler)]: {
      body: null,
      abort: null,
      onConnect: [Function: onConnect],
      onHeaders: [Function: onHeaders],
      onData: [Function: onData],
[...]

PR #2712, opened 2024-02-07
feat: implement missing Request spec

tsctx:


KhafraDev (changes requested):
this isn't implemented because we don't implement request priority, as the note stated... adding it as an option is not implementing it

mcollina:
Sooo, what's about Request priority? How should it be implemented?

KhafraDev:
https://github.com/nodejs/undici/issues/1748

tsctx:
Okey. I understand.

PR #2718, opened 2024-02-08

tsctx:
Ref: https://github.com/nodejs/node/commit/d3f8db124463e478420d8bba6fa5cf13af7b4ecb


mcollina (approved):
lgtm

KhafraDev:
> committed on Dec 19, 2014

are you sure that v8 hasn't improved splice in 10 years?

Uzlopak:
I doubt that some of these Arrays need to keep the ordering. In the case that the array order is irrelevant, you could avoid all the reassignments of the array and just put the last element of the array into the desired Index and pop the last element.

```
function spliceUnordered (list, index) {
    if (index === list.length - 1) {
        list.pop()
        return
    }
    list[index] = list.pop()
  }

const bla = 'abcdef'.split('')
spliceUnordered(bla, 2)
console.log(bla)
spliceUnordered(bla, 4)
console.log(bla)
spliceUnordered(bla, 0)
console.log(bla)
spliceUnordered(bla, 0)
console.log(bla)
spliceUnordered(bla, 0)
console.log(bla)
spliceUnordered(bla, 0)
console.log(bla)
spliceUnordered(bla, 0)
console.log(bla)
```

https://jsperf.app/kikuvo

image

tsctx:
> > committed on Dec 19, 2014
> 
> 
> 
> are you sure that v8 hasn't improved splice in 10 years?

Yes, it is. Although optimized, it is still much slower than this method.

Uzlopak:
@tsctx 

What is your opinion on my proposal?

tsctx:
@Uzlopak 

Your idea is great. But, unfortunately, these cannot replace all of them, as some of them need to be kept in order.


I consider it sufficient because it is several times faster than `Array#splice`.

lib/core/util.js
```diff
@@ -508,6 +508,16 @@ function parseRangeHeader (range) {
     : null
 }
 
+/**
+ * @param {any[]} list
+ * @param {number} index
+ */
+function spliceOne (list, index) {
+  const length = list.length
+  while (++index < length) list[index - 1] = list[index]
```
metcoder95, line 517:
```suggestion
  while (++index < length) {
  	list[index - 1] = list[index]
  }
```

Just small `nit`

metcoder95 (approved):

lib/core/util.js
```diff
@@ -508,6 +508,18 @@ function parseRangeHeader (range) {
...
 
+/**
+ * @param {any[]} list
+ * @param {number} index
+ */
+function spliceOne (list, index) {
+  const length = list.length
+  while (++index < length) {
+  	list[index - 1] = list[index]
+  }
```
tsctx, line 519:
```suggestion
  while (++index < length) {
  	 list[index - 1] = list[index]
  }
```

lib/core/util.js
```diff
@@ -508,6 +508,18 @@ function parseRangeHeader (range) {
...
 
+/**
+ * @param {any[]} list
+ * @param {number} index
+ */
+function spliceOne (list, index) {
+  const length = list.length
+  while (++index < length) {
+  	 list[index - 1] = list[index]
+  }
```
tsctx, line 519:
```suggestion
  while (++index < length) {
    list[index - 1] = list[index]
  }
```

Uzlopak:
it is strange to propose a performance change but not adapt a change which removes all unnecessary operations.

mcollina (approved):
lgtm

PR #2720, opened 2024-02-08
test: replace t.type with t.ok and instanceof

Uzlopak:
This PR replaces t.type with t.ok and instanceof calls, as node:assert and tspl does not have a type() check.


metcoder95 (approved):

mertcanaltin (approved):
lgtm

mcollina (approved):
Lgtm

Issue #2722, opened 2024-02-09
Semver Major Considerations


ronag:
Based on my experience working with undici through nxt-undici and building a more advanced client on top undici I have come to a few realizations that have semver major implications:

- `onBodySent` and `onRequestSent` are huge footguns. Remove it. (just wrap the body)
- `interceptors` are weird and overengineered. Remove it. (just wrap dispatchers)
- `onConnect` is confusing and should be renamed or something to make it more intuitive. Possibly entirely removed.
-  `body` should support a factory method (important for retries and redirects).
- `onResponseStarted`. Why is onHeaders insufficient?
- `maxRedirections` and `RedirectHandler` should not be part of core/dispatcher APi. Move to the api methods.
- Change hooks signature to accept objects instead of params.
- Rename disatpcher param to dispatch which is just a function (opts, handler) => {}

metcoder95:
>onBodySent and onRequestSent are huge footguns. Remove it. (just wrap the body)

SGTM, left some thoughts on the PR 👍 

>interceptors are weird and overengineered. Remove it. (just wrap dispatchers)

Agree, wrapping dispatches providers' full ergonomics. The interceptors feel like a half-baked feature sadly

>onConnect is confusing and should be renamed or something to make it more intuitive. Possibly entirely removed.

What's the problem that surfaces from it?
Renaming it seems fair, but curious about what's the exact problem there

>body should support a factory method (important for retries and redirects).

Maybe we can refactor it to a base class and do something similar as we do already for the `Dispatcher`?

>onResponseStarted. Why is onHeaders insufficient?

I kind of remember this PR; I believe it is mostly around documenting what `onHeaders` can be used for beyond receiving and parsing the headers.

>maxRedirections and RedirectHandler should not be part of core/dispatcher APi. Move to the api methods.

Do you mean to implement them within each of the APIs or just export a new one?


Do we have a timeline for the next major version?
Or maybe we can make this list the requirements for the next major?

ronag:
@metcoder95 do you think you could help me with a PR removing the interceptor stuff?

Uzlopak:
If we consider undici major version, than maybe target llhttp 9 upgrade to the next branch?!

@ShogunPanda 
@mcollina

metcoder95:
Sure, I can have something prepared for next week 👍

metcoder95:
>If we consider undici major version, than maybe target llhttp 9 upgrade to the next branch?!

[...]

PR #2723, opened 2024-02-09
remove anti-pattern dispatcher hooks

ronag:
onBodySent on onRequestSent are footguns that easily cause bugs
when implementing logic will send multiple requests, e.g. redirect
and retry.

To achieve similar functionality wrap body into a stream and listen
to 'data' and 'end' events.

Refs: https://github.com/nodejs/undici/issues/2722

### DIFF
--- a/docs/api/DiagnosticsChannel.md
+++ b/docs/api/DiagnosticsChannel.md
@@ -27,17 +27,6 @@ diagnosticsChannel.channel('undici:request:create').subscribe(({ request }) => {
 
 Note: a request is only loosely completed to a given socket.
 
-
-## `undici:request:bodySent`
-
-```js
-import diagnosticsChannel from 'diagnostics_channel'
-
-diagnosticsChannel.channel('undici:request:bodySent').subscribe(({ request }) => {
-  // request is the same object undici:request:create
-})
-```
-
 ## `undici:request:headers`
 
 This message is published after the response headers have been received, i.e. the response has been completed.
--- a/docs/api/Dispatcher.md
+++ b/docs/api/Dispatcher.md
@@ -209,11 +209,9 @@ Returns: `Boolean` - `false` if dispatcher is busy and further dispatch calls wo
 * **onConnect** `(abort: () => void, context: object) => void` - Invoked before request is dispatched on socket. May be invoked multiple times when a request is retried when the request at the head of the pipeline fails.
 * **onError** `(error: Error) => void` - Invoked when an error has occurred. May not throw.
 * **onUpgrade** `(statusCode: number, headers: Buffer[], socket: Duplex) => void` (optional) - Invoked when request is upgraded. Required if `DispatchOptions.upgrade` is defined or `DispatchOptions.method === 'CONNECT'`.
-* **onResponseStarted** `() => void` (optional) - Invoked when response is received, before headers have been read.
 * **onHeaders** `(statusCode: number, headers: Buffer[], resume: () => void, statusText: string) => boolean` - Invoked when statusCode and headers have been received. May be invoked multiple times due to 1xx informational headers. Not required for `upgrade` requests.
 * **onData** `(chunk: Buffer) => boolean` - Invoked when response payload data is received. Not required for `upgrade` requests.
 * **onComplete** `(trailers: Buffer[]) => void` - Invoked when response payload and trailers have been received and the request has completed. Not required for `upgrade` requests.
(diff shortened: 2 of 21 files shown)


metcoder95:
The rationale seems fair to me. I had a little bit of an issue understanding its specific use case. However, I think we should provide the same _lifecycle event_ in some other way almost _native_ to `undici`.

Wrapping the body into a `stream` and always listening for its `end` event can be a way to go to trigger the event, we just need to be clear on its intentions.

For the PR, it seems lint is not happy and we need to work on some tests 😅 
Shall we point to next?

metcoder95:
>Sorry, what do you mean?

Point the PR to `next` branch

mcollina (changes requested):
> To achieve similar functionality wrap body into a stream and listen
to 'data' and 'end' events.

This is not the same. The reason those were added was essentially to avoid wrapping the body with a stream, because it adds overhead.

ronag:
> > To achieve similar functionality wrap body into a stream and listen
> > to 'data' and 'end' events.
> 
> This is not the same. The reason those were added was essentially to avoid wrapping the body with a stream, because it adds overhead.

It is the same. But slower. However, they cause more harm than good as it is currently. What happens e.g. during a retry or redirect? When the same body is sent twice, possibly interleaved? The API is badly broken.

mcollina:
Can you add a atest showing how to achieve the same?

mcollina:
That would be ok.

ronag:
No idea how/where to put this. But something like would be the same thing:

```
import undici from 'undici'

function request(url, opts, onBodySent, onRequestSent) {
  let body = !opts?.body || typeof opts.body?.pipe === 'function' ? opts.body : Readable.from([opts.body])
  body
    .on('data', onBodySent)
    .on('end', onRequestSent) 
    .pause()
  undici.request(url, { ...opts, body, })
}
```

mcollina (approved):
Lgtm

tjhiggins:
@ronag what would be the recommended way to get the request body in an interceptor going forward? 

Since fetch's requestOptions.body is an AsyncGenerator and difficult to parse: https://github.com/nodejs/undici/blob/a73fd0970f7b30cdb38b099e71ff2e354500c3b5/lib/web/fetch/index.js#L1845

I'm currently using onBodySent for my LoggingHandler:
```ts

export class LoggingHandler extends DecoratorHandler implements Dispatcher.DispatchHandlers {
  private requestBody = '';
  private responseBody = '';
  private responseHeaders: Buffer[] = [];
  private statusCode?: number;
  private statusText?: string;

  constructor(
    protected options: { log: Log },
    protected requestOptions: Dispatcher.DispatchOptions,
    protected handler: Dispatcher.DispatchHandlers,
  ) {
    super(handler);
  }

  onConnect(abort: (err?: Error) => void): void {
    this.requestBody = '';
    this.responseBody = '';
    this.responseHeaders = [];
    this.statusCode = undefined;
    this.statusText = undefined;
    return this.handler.onConnect!(abort);
  }

  onHeaders(statusCode: number, headers: Buffer[], resume: () => void, statusText: string): boolean {
    this.statusCode = statusCode;
    this.statusText = statusText;
    this.responseHeaders = headers;
    return this.handler.onHeaders!(statusCode, headers, resume, statusText);
  }

  onBodySent(chunk: any, totalBytesSent: number) {
    try {
      this.requestBody += Buffer.from(chunk).toString();
    } catch (err) {
      console.debug('Failed to parse request body:', err);
    }
    if (this.handler.onBodySent) {
      this.handler.onBodySent(chunk, totalBytesSent);
    }
  }

  onData(chunk: Buffer): boolean {
    this.responseBody += chunk.toString();
    return this.handler.onData!(chunk);
  }

  onComplete(trailers: string[] | null) {
    const responseHeaders = util.parseHeaders(this.responseHeaders);

    const logData = {
      url: this.requestOptions.path,
      method: this.requestOptions.method || "GET",
      headers: this.requestOptions.headers,
      responseHeaders,
      body: this.requestBody,
      status: this.statusCode,
      statusText: this.statusText,
      data: this.responseBody,
    };

    let severity = 'INFO';
    if (this.statusCode === 404) {
      severity = 'WARNING';
    } else if (!this.statusCode || this.statusCode > 399) {
      severity = 'ERROR';
    }

    const metadata = {
      resource: { type: 'global' },
      // See: https://cloud.google.com/logging/docs/reference/v2/rest/v2/LogEntry#logseverity
      severity,
    };

    const entry = this.options.log.entry(metadata, logData);
    this.options.log.write(entry).catch((reason) => console.warn('Failed to write log entry:', reason));

    return this.handler.onComplete!(trailers);
  }

  onError(err: Error) {
    return this.handler.onError!(err);
  }
}
```

PR #2727, opened 2024-02-09
remove third party blob support

KhafraDev:


mcollina:
CI is failing

KhafraDev:
I'll update the test actually, to use fs.openAsBlob instead.

mcollina:
I would prefer to land this if openAsBlob is stable.

ronag:
Rebase to next branch?

PR #2729, opened 2024-02-09
remove useless options in web streams

KhafraDev:
ref: https://github.com/nodejs/undici/commit/87a48113f1f68f60aa09abb07276d7c35467c663#r138446550

If we set the strategy correctly as the second parameter, the test starts to fail.


mcollina (approved):
Lgtm

PR #2731, opened 2024-02-10
Let's add superagent to the benchmark. closes #2730

### LINKED ISSUE #2730: Let's add superagent to the benchmark table.
## This would solve...

I'm building a [library](https://github.com/eddienubes/sagetest) similar to **supertest**, but on top of **undici**. As an advantage, I'd like point out that **undici** is faster than **superagent** (which **supertest** uses under the hood).

In my opinion, for this benchmark it makes sense to reside in **undici**'s repo.

By the way, seems like **superagent** doesn't reuse established connections on purpose, so it'll be quite slow.

[Pull Request](https://github.com/nodejs/undici/pull/2731)

eddienubes:
## This relates to...

Adding superagent to the benchmark 
#2730 

## Rationale

I'm building a [library](https://github.com/eddienubes/sagetest) similar to supertest, but on top of undici. As an advantage, I'd like point out that undici is faster than superagent (which supertest uses under the hood).

In my opinion, for this benchmark it makes sense to reside in undici's repo.

## Changes

- added **superagent** call to benchmark.js
- added **superagent** and **@types/superagent** as devDependencies.

Also, I haven't changed the README yet. I'd like someone from maintainers to run this benchmark to make sure proper arguments were passed (samples, connections etc.)

----

- [x] Tested
- [x] Documented
- [x] Review ready
- [x] In review
- [x] Merge ready


benchmarks/benchmark.js
```diff
@@ -318,6 +319,16 @@ if (process.env.PORT) {
       }).catch(console.log)
     })
   }
+
+  experiments.superagent = () => {
+    return makeParallelRequests(resolve => {
+      superagent.get(dest.url).pipe(new Writable({
```
mcollina, line 325: You should be creating and setting and agent, otherwise the benchmark would not be even

mcollina (changes requested):

benchmarks/benchmark.js
eddienubes: @mcollina 
Superagent doesn't expose a verbose API for setting such agent, so I used a workaround.

mcollina (approved):
lgtm

ronag (approved):

eddienubes:
@ronag @mcollina 
I think we didn't update README, or you'd like to do it in a separate PR?

ronag:
It would be have to be in a separate PR now. Someone just needs to run the benchmarks.

eddienubes:
I can handle that.
Do I just run it with defaults set in benchmark.js?

ronag:
> I can handle that. Do I just run it with defaults set in benchmark.js?

I think so.

mcollina:
I'll update the benchmark at the first occasion

PR #2733, opened 2024-02-11
feat: add support for body factory

ronag:
## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -110,6 +110,7 @@ let extractBody
 const FastBuffer = Buffer[Symbol.species]
 
 const kClosedResolve = Symbol('kClosedResolve')
+const kSignal = Symbol('kSignal')
 
 /**
  * @type {import('../types/client').default}
@@ -1460,10 +1461,12 @@ function _resume (client, sync) {
       return
     }
 
-    if (!request.aborted && write(client, request)) {
-      client[kPendingIdx]++
-    } else {
+    write(client, request)
+
+    if (request.aborted) {
       client[kQueue].splice(client[kPendingIdx], 1)
+    } else {
+      client[kPendingIdx]++
     }
   }
 }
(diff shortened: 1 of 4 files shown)


mcollina:
I don't understand this change, why is it needed?

ronag:
It's for e.g enabling retries during pipelining for put request with a streaming body.

ronag:
Ditto with redirects

ronag:
You pass a factory method that can always re create the stream body from 0.

ronag:
Maybe a better example:

```js
request(url, {
  method: 'GET',
  body: () => fs.createReadStream(path)
})
```

types/dispatcher.d.ts
```diff
@@ -98,7 +98,7 @@ declare namespace Dispatcher {
     path: string;
     method: HttpMethod;
     /** Default: `null` */
-    body?: string | Buffer | Uint8Array | Readable | null | FormData;
+    body?: string | Buffer | Uint8Array | Readable | null | FormData | Function;
```
metcoder95, line 101: I believe we would need to document better this 🤔

lib/client.js
```diff
@@ -1472,12 +1474,59 @@ function shouldSendContentLength (method) {
 }
 
 function write (client, request) {
+  const socket = client[kSocket]
+  const { body } = request
+
+  if (typeof body === 'function') {
+    // TODO (fix): What if socket is destroyed while
+    // waiting for body?
```
metcoder95, line 1482: If the factory is meant to be idempotent, destroying the body should be ok to do too.
We should advice to be careful about the state of the body

mcollina:
This seems something I would add to the `dispatcher.stream()` method.

https://github.com/nodejs/undici/blob/main/docs/api/Dispatcher.md#dispatcherstreamoptions-factory-callback

ronag:
Sorry. But I'm not sure how that's relevant here?

mcollina:
This kind of API seemed nice there.

ronag:
Sorry but I still don't understand. This cannot be implemented there. For the pipeline retry to work this has to live in the client. Putting it in stream doesn't do anything?

mcollina (approved):
lgtm

PR #2734, opened 2024-02-11
fix: handle request body as late as possible

ronag:
- Avoid potential unhandled exception from formdata stream
- Allocate new buffers as late as possible


lib/client.js
```diff
@@ -1494,14 +1498,34 @@ function write (client, request) {
     method === 'PATCH'
   )
 
+  if (util.isFormDataLike(body)) {
+    if (!extractBody) {
+      extractBody = require('./fetch/body.js').extractBody
+    }
+
+    const [bodyStream, contentType] = extractBody(body)
+    if (request.contentType == null) {
```
KhafraDev, line 1507: Not a critique of this PR but if the content-type isn't always set here the request will always be invalid. A server can't parse a multipart/form-data body without the boundary type, which is included in the header.

lib/client.js
```diff
@@ -1494,14 +1498,34 @@ function write (client, request) {
...
+      headers += `content-type: ${contentType}\r\n`
+    }
+    body = bodyStream.stream
+    contentLength = bodyStream.length
+  } else if (util.isBlobLike(body) && request.contentType == null && body.type) {
+    headers += `content-type: ${body.type}\r\n`
```
KhafraDev, line 1513: Also... it's possible for body.type to be an empty string here. I don't think that's a valid header value?

lib/client.js
ronag: I guess that's a user error then and it will fail?

KhafraDev: But the request will still work, the server won't be able to process it which can be hard to debug. I had someone ask me a few days ago why a request was failing server side and it ended up being because they were setting a `content-type: multipart/form-data` header manually.

KhafraDev: It seems like curl overwrites the header, which can make it even harder to debug as an equivalent request from undici will fail whereas curl won't. Probably better off for another issue, I was just reminded of it.

lib/client.js
```diff
@@ -1494,14 +1498,34 @@ function write (client, request) {
...
+  }
+
+  if (util.isStream(body) && util.isDestroyed(body)) {
+    errorRequest(client, request, body.errored ?? new RequestAbortedError())
```
metcoder95, line 1517: It seems the errors thrown are `RequestAborted`, possible something happening here?

mcollina (changes requested):
Can you add a test?

mcollina:
> Avoid potential unhandled exception from formdata stream

This should not happen because the stream is not flowing. Why does it happen?

mcollina:
So why are you moving this logic down inside `client.js`?

ronag:
So that we don't create unnecessary stuff if the request is aborted and make sure we don't add more allocation to the old heap.

test/client.js
```diff
@@ -1055,6 +1055,7 @@ test('basic POST with empty stream', (t) => {
       method: 'POST',
       body
     }, (err, { statusCode, headers, body }) => {
+      console.error(err)
```
metcoder95, line 1058: Left over?

lib/client.js
```diff
@@ -1494,14 +1498,29 @@ function write (client, request) {
     method === 'PATCH'
   )
 
+  if (util.isFormDataLike(body)) {
```
metcoder95, line 1501: For another PR, we will need to port this to H2 as well

metcoder95 (approved):

mcollina (approved):
lgtm on green CI

PR #2747, opened 2024-02-12
chore: migration of test/connect-timeout.js

Uzlopak:
This one seems to be not easily migratable?!

### DIFF
--- a/test/connect-timeout.js
+++ b/test/connect-timeout.js
@@ -1,23 +1,26 @@
 'use strict'
 
-const { test } = require('tap')
+const { tspl } = require('@matteo.collina/tspl')
+const { test, after } = require('node:test')
 const { Client, Pool, errors } = require('..')
 const net = require('node:net')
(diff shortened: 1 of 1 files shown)


test/connect-timeout.js
```diff
@@ -1,37 +1,40 @@
...
+test('prioritize socket errors over timeouts', async (t) => {
+  t = tspl(t, { plan: 1 })
   const connectTimeout = 1000
   const client = new Pool('http://foobar.bar:1234', { connectTimeout: 2 })
 
   client.request({ method: 'GET', path: '/foobar' })
     .then(() => t.fail())
     .catch((err) => {
-      t.equal(err.code, 'ENOTFOUND')
+      t.strictEqual(err.code, 'ENOTFOUND')
```
Uzlopak, line 17: Results in UND_ERR_CONNECT_TIMEOUT. 
I guess, this is because they are in the same loop as the sleep call, so it is not enforcing the ENOTFOUND error before the socket timeout.

PR #2753, opened 2024-02-13
chore: remove usage of http-errors in proxy example

Uzlopak:
Currently we have http-errors in the dependency tree because we use docsify. If we ever remove docsify, the proxy example would break.


metcoder95 (approved):

ronag (approved):

mcollina (approved):
lgtm

PR #2757, opened 2024-02-14
core: sync next with main

### LINKED ISSUE #2730: Let's add superagent to the benchmark table.

metcoder95:
## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready


ronag:
isn't it easier to just cherry-pick all commits from main on top of next?

metcoder95:
Kind of, but just wanted to have another pair of eyes on it given that we adjust a lot of tests and other pieces that overlap quite heavily 😅

metcoder95:
The biggest conflicts were the testing, the remains seem to be the same

ronag:
I fixed this.

PR #2762, opened 2024-02-14
run WPTs with c8

KhafraDev:
Tests run ~50-60% slower for me, I don't know why.


KhafraDev:
the codecov report doesn't make any sense...?

Uzlopak:
When i tried to implement it, i added in runner.js that the env variable for NODE_V8_COVERAGE  to the worker and then coverage json files started to also contain the coverage of the code run in the worker.

I did not use c8, but removed the --coverage flags of borp and used cross-env in the coverage:ci npm script to set the NODE_V8_COVERAGE so that it also works in windows. Using /coverage/tmp as target folder.

Also e.g. in the wpt tests of websocket we bundle the code. Do we then use the bundled code? If so the tests will cover that file and not the unbundled source files.

At the end instead run c8 --all report to generate the report from the temp folder.

KhafraDev:
maybe you could take over then, you seem to know more than I do 🤞

Issue #2769, opened 2024-02-16
expose MessageEvent for node core


KhafraDev:
Both node and undici implement MessageEvent. Node's implementation is far from spec compliant, whereas undici's is (we run idl WPTs on MessageEvent). Should we replace the existing MessageEvent with undici's?

On a side note: since I had no idea that node had MessageEvent, it does not work with WebSocket.

metcoder95:
It will most likely mean a major, should we aim for v22?
I'm just not fully aware what's the process on Node.js releases for that 😅

KhafraDev:
Unfortunately error messages do change, but all tests from node core pass.

PR #2770, opened 2024-02-16
expose MessageEvent in fetch bundle

KhafraDev:
Fixes: #2769 

Tests were added from node core (including one for a bug fix that has not landed, and a bug that hasn't yet been reported) and all pass as expected. Error messages and codes do change, unfortunately, so this will end up being a semver-major change in core.


Uzlopak (approved):
LGTM

mcollina (approved):
lgtm

metcoder95 (approved):

Issue #2771, opened 2024-02-16
next branch purged of semver-major changes?


Uzlopak:
@ronag 

Did your force push removed the semver-major changes? E.g. llhttp 9.2.0 is back to v8 in next.

ronag:
What PR did that land in?

ronag:
I did a force push assuming nothing had landed on the next branch. I'm confused.

Uzlopak:
With #2701 llhttp was upgraded to 9.2.0

Also the changes of this PR of yours seems to be removed from next branch #2724

We probably can fix this as we have the commits still in the sync-next branch of @metcoder95 and maybe have to cherrypick them?!

ronag:
I will fix this.

Uzlopak:
Awesome. Thanks

PR #2724, opened 2024-02-09
remove onResponseStarted

ronag:
Refs: https://github.com/nodejs/undici/issues/2722

## Status

- [ ] I have read and agreed to the [Developer's Certificate of Origin][cert]
- [ ] Tested
- [ ] Benchmarked (**optional**)
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

[cert]: https://github.com/nodejs/undici/blob/main/CONTRIBUTING.md

PR #2772, opened 2024-02-16
Missing commits

ronag:
Fixes: https://github.com/nodejs/undici/issues/2771#issuecomment-1949247097


Uzlopak:
Is #2724 missing?

ronag:
That's not necessary. The hooks commit does that as well.

metcoder95:
Thanks! The merge might swallow several commits

Issue #2015, opened 2023-03-18
unidici doesn't play nice with webpack


uriva:
I am writing a package in deno, and compiling it into npm.

Somehow unidici gets in the compiled target.

I then use webpack because I want to use this package in react setup.

webpack refuses to process unidici because of two dependencies: util/types and stream/web

I haven't found a polyfill for them.

Any idea what can be done?

this is my webpack config

```
const NodePolyfillPlugin = require("node-polyfill-webpack-plugin");

const path = require("path");

module.exports = {
  mode: "production",
  entry: "./src/index.js",
  output: {
    filename: "main.js",
    path: path.resolve(__dirname, "dist"),
  },
  plugins: [new NodePolyfillPlugin()],
  resolve: {
    fallback: {
      perf_hooks: false,
      async_hooks: false,
      dns: false,
      tls: false,
      child_process: false,
      fs: false,
      net: require.resolve("net-websocket-polyfill"),
      diagnostics_channel: require.resolve("diagnostics_channel"),
      worker_threads: require.resolve("bthreads"),
    },
  },
};
```

KhafraDev:
this is a duplicate of #1992

Issue #1992, opened 2023-03-06
Request: bump minimum Node.js version in package.json


vetruvet:
Currently `package.json` specifies `>=12.18` in `engines.node`, but the code uses `??=` and `util/types` which are Node.js 15+ features.

## The implementation should look like...

Updating `package.json` to reflect the correct minimum Node.js version.

## I have also considered...

Alternatively, the code could stop using those features to retain compatibility with 12.18+

mcollina:
Could you tell us more about the compatibility issues you are facing? Our test suite (minus fetch, which is 16+) pass on node v12: https://github.com/nodejs/undici/actions/runs/4331585584.

vetruvet:
I guess the issue is when the library is bundled (e.g. by webpack), it causes code that's conditionally required only on certain Node versions to be included in the resulting code bundle, and there are instances of nullish coalesce assignmen t (`??=`) syntax in those parts (specifically `lib/fetch/headers.js`, `lib/cookies/parse.js`, and `lib/websocket/receiver.js`). 

Without bundling those are only required on Node 16+ but when bundled, it causes parsing errors even though those features aren't used.

I've worked around it so far by configuring Babel to parse&transform this library. Perhaps at least a warning about using this library with bundlers in the docs would be helpful.

ronag:
We could also not use `??=` ... PR welcome

KhafraDev:
we've had regressions so I'd rather not accept a PR unless tests are included

trivikr:
Closing as the issues raised in this were with node 12 in undici 4.x, and all undici versions <7.x have been end-of-life at the time of writing.

Issue #1541, opened 2022-07-13
 Can't resolve 'stream/web' ./undici/lib/fetch


ArthurBonsu:
## Logs & Screenshots

./node_modules/hardhat/node_modules/undici/lib/core/util.js
Module not found: Can't resolve 'stream/web' in 'D:\BLOCKCHAIN EXPERIMENTAL LAB\flowstationyarnV2\flowstationyar
n\flowstationkingdom\node_modules\hardhat\node_modules\undici\lib\core'

Import trace for requested module:
./node_modules/hardhat/node_modules/undici/index.js
./node_modules/hardhat/internal/hardhat-network/jsonrpc/server.js
./node_modules/hardhat/builtin-tasks/node.js
./node_modules/hardhat/internal/core/tasks/builtin-tasks.js
./node_modules/hardhat/internal/core/config/config-loading.js
./node_modules/hardhat/internal/lib/hardhat-lib.js
./pages/originaltokenswap.tsx

./node_modules/hardhat/node_modules/undici/lib/fetch/body.js
Module not found: Can't resolve 'util/types' in 'D:\BLOCKCHAIN EXPERIMENTAL LAB\flowstationyarnV2\flowstationyar
n\flowstationkingdom\node_modules\hardhat\node_modules\undici\lib\fetch'

Import trace for requested module:
./node_modules/hardhat/node_modules/undici/lib/fetch/response.js
./node_modules/hardhat/node_modules/undici/index.js
./node_modules/hardhat/internal/hardhat-network/jsonrpc/server.js
./node_modules/hardhat/builtin-tasks/node.js
./node_modules/hardhat/internal/core/tasks/builtin-tasks.js
./node_modules/hardhat/internal/core/config/config-loading.js
./node_modules/hardhat/internal/lib/hardhat-lib.js
./pages/originaltokenswap.tsx

./node_modules/hardhat/node_modules/undici/lib/fetch/body.js
Module not found: Can't resolve 'stream/web' in 'D:\BLOCKCHAIN EXPERIMENTAL LAB\flowstationyarnV2\flowstationyar
n\flowstationkingdom\node_modules\hardhat\node_modules\undici\lib\fetch'

Import trace for requested module:
./node_modules/hardhat/node_modules/undici/lib/fetch/response.js
./node_modules/hardhat/node_modules/undici/index.js
./node_modules/hardhat/internal/hardhat-network/jsonrpc/server.js
./node_modules/hardhat/builtin-tasks/node.js
[...]

PR #2775, opened 2024-02-17
fix: add node prefix for util to fix issue in env with min version node 18

riderx:
## This relates to 

I found 3 old issues talking about this:
https://github.com/nodejs/undici/issues/1992
https://github.com/nodejs/undici/issues/2015
https://github.com/nodejs/undici/issues/1541
Where it talks mainly about anterior version of Node.js before 18.
So my assumption if this import was maybe forgotten to get the prefix, I'm aware that can be wrong, but since that the only blocker for Cloudflare use case I try to bring this

## Rationale

I use undici in Cloudflare workers with Node.js support, it only supports new package nomenclature, and it seems only these 2 imports are malformed compared to all other imports

## Changes

Add node: prefix missing for util/types 

### Features

No new features

### Bug Fixes

This fix undici for Cloudflare wrangler config with 
```yml
compatibility_flags = [ "nodejs_compat" ]
```

None since we already requested Node.js 18 min.

## Status

It passed all tests on Node.js 18

- [x] Tested
- [s] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready


riderx:
CleanShot 2024-02-18 at 00 51 54@2x
I have read the contribution file, and I'm not sure to understand the relation between this repo and the Wasm lib instructions.
I would like to get more clarified about the purpose of this for the contribution.

Uzlopak (approved):
LGTM

Uzlopak:
When I added the node prefix i actually missed it. 

Can you maybe also do following changes?
test/websocket/messageevent.js line 4
test/fetch/pull-dont-push.js line 6 - 9

mcollina (approved):
lgtm

mcollina:
As a side note, I'm not convinced this runs in workerd/cloudflare workers.

metcoder95 (approved):

riderx:
@Uzlopak thanks, I made the change!
Honestly, I didn't expect so many reviews on a Sunday, thanks a lot!

Uzlopak (approved):
Thank you

riderx:
@mcollina I did try my change on workers and there is still one issue, but it's not linked to Undici, it's an error in workers :
https://github.com/cloudflare/workers-sdk/issues/5051
I believe if they fix it, Undici will work with the flag.

mcollina:
I don't think it will, and I'm not 100% why you'd like to do this. This module will try to open a new socket and manage a connection pool.

workers already offer their kmplementation of fetch(), which is definitely more optimized for their environment. Why would you need to use this instead?

ronag (approved):

riderx:
Agree my usecase is a bit special.
My original code is in Deno, and with the import map of Deno and package.json  was able to to replicate my serverless function in many envs.
Netlify functions
Netlify edge.
AWS lambda
cloudflare workers.
Lately I switched from custom implementations for S3 to a package made originally for Deno.

And I ported this package with Deno dnt to NPM.
You can see it here:

https://github.com/bradenmacdonald/deno-s3-lite-client/pull/30

One of my try was to use Undici shim to make it run, it worked on Node but not on workers.

now the final solution is a dynamic import who work everywhere.

i really like the idea in the future the same code will be able to run anywhere, just based on some config differences.

KhafraDev (approved):

mcollina (approved):
lgtm

PR #2778, opened 2024-02-18
test: always exit with 0 when running in Node's Daily WPT Report CI job

panva:
Updates the WPTRunner to always exit with 0 when running in https://github.com/nodejs/node/blob/main/.github/workflows/daily-wpt-fyi.yml

This is so that WPT failures don't block subsequent suites from running and being included in the WPT Report submitted to wpt.fyi.

Example of missing suites: https://wpt.fyi/results/?run_id=5181229028081664&run_id=5071266792341504&run_id=5183242763763712&run_id=5108285182574592


mcollina (approved):
lgtm

panva:
This is finally in Node.js nightly - see https://wpt.fyi/results/?run_id=5175977478455296&run_id=5143455717654528&run_id=4878141998235648&run_id=4853393356488704

PR #2780, opened 2024-02-19
fix: make mock-agent.js test more resilient

Uzlopak:
It seems that the modified test is failing sometimes. This makes the test slightly more resilient...


mcollina (approved):
lgtm

PR #2783, opened 2024-02-19
mock: improve validateReplyParameters

Uzlopak:
if you would set responseOptions to null, it would pass but then it would throw an error in createMockScopeDispatchData because it would try to read headers attribute from null.

https://github.com/nodejs/undici/blob/fedae35f1b71999d2327ef6a28538f5845a71515/lib/mock/mock-interceptor.js#L96

### DIFF
--- a/lib/mock/mock-interceptor.js
+++ b/lib/mock/mock-interceptor.js
@@ -106,7 +106,7 @@ class MockInterceptor {
     if (typeof data === 'undefined') {
       throw new InvalidArgumentError('data must be defined')
     }
-    if (typeof responseOptions !== 'object') {
+    if (typeof responseOptions !== 'object' || responseOptions === null) {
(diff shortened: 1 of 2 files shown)


lib/mock/mock-interceptor.js
```diff
@@ -106,7 +106,7 @@ class MockInterceptor {
     if (typeof data === 'undefined') {
       throw new InvalidArgumentError('data must be defined')
     }
```
Uzlopak, line 108: data can never be undefined because it is always set to empty string

https://github.com/nodejs/undici/blob/fedae35f1b71999d2327ef6a28538f5845a71515/lib/mock/mock-interceptor.js#L151

KhafraDev (approved):

metcoder95 (approved):

PR #2784, opened 2024-02-20
fix: convert HeadersInit to sequence/dictionary correctly

KhafraDev:
## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/fetch/headers.js
+++ b/lib/fetch/headers.js
@@ -528,8 +528,10 @@ Object.defineProperties(Headers.prototype, {
 
 webidl.converters.HeadersInit = function (V) {
   if (webidl.util.Type(V) === 'Object') {
-    if (V[Symbol.iterator]) {
-      return webidl.converters['sequence<sequence<ByteString>>'](V)
+    const iterator = Reflect.get(V, Symbol.iterator)
+
+    if (typeof iterator === 'function') {
+      return webidl.converters['sequence<sequence<ByteString>>'](V, iterator.bind(V))
     }
 
     return webidl.converters['record<ByteString, ByteString>'](V)
--- a/lib/fetch/webidl.js
+++ b/lib/fetch/webidl.js
@@ -218,7 +218,7 @@ webidl.util.IntegerPart = function (n) {
 
 // https://webidl.spec.whatwg.org/#es-sequence
 webidl.sequenceConverter = function (converter) {
-  return (V) => {
+  return (V, Iterable) => {
     // 1. If Type(V) is not Object, throw a TypeError.
     if (webidl.util.Type(V) !== 'Object') {
       throw webidl.errors.exception({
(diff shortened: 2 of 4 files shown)


lib/fetch/webidl.js
```diff
@@ -273,8 +273,8 @@ webidl.recordConverter = function (keyConverter, valueConverter) {
     const result = {}
 
     if (!types.isProxy(O)) {
-      // Object.keys only returns enumerable properties
-      const keys = Object.keys(O)
+      // 1. Let desc be ? O.[[GetOwnProperty]](key).
+      const keys = [...Object.getOwnPropertyNames(O), ...Object.getOwnPropertySymbols(O)]
```
KhafraDev, line 277: Although Object.keys returns enumerable *string* properties, it didn't include enumerable *symbol* properties. This is only noticeable when doing something along the lines of `new Headers({ [Symbol.iterator]: null })`.

lib/fetch/headers.js
```diff
@@ -528,8 +528,10 @@ Object.defineProperties(Headers.prototype, {
 
 webidl.converters.HeadersInit = function (V) {
   if (webidl.util.Type(V) === 'Object') {
-    if (V[Symbol.iterator]) {
```
KhafraDev, line 531: So there are two bugs here:
- The existence of a Symbol.iterator does not mean a value is a sequence.
- Accessing Symbol.iterator this way is detectable via Proxy. Once we pass V to the sequence converted it once again accesses V[Symbol.iterator].

test/fetch/headers.js
```diff
@@ -719,3 +719,30 @@ test('When the value is updated, update the cache', (t) => {
...
+test('Invalid Symbol.iterators', (t) => {
+  const { throws } = tspl(t, { plan: 3 })
+
+  throws(() => new Headers({ [Symbol.iterator]: null }), TypeError)
+  throws(() => new Headers({ [Symbol.iterator]: undefined }), TypeError)
+  throws(() => {
+    const obj = { [Symbol.iterator]: null }
+    Object.defineProperty(obj, Symbol.iterator, { enumerable: false })
+
+    new Headers(obj) // eslint-disable-line no-new
```
KhafraDev, line 746: This throws because Symbol.iterator can't be converted to a string rather than being an invalid dictionary. Technically still a bug, but this also happens in Chrome so I assume no one in the history of mankind will ever notice.

Uzlopak (approved):
LGTM

mcollina (approved):
lgtm

metcoder95 (approved):

test/fetch/headers.js
KhafraDev: As in, we can't implement the spec steps stating to ignore non-enumerable properties because then this case wouldn't throw an error.

PR #2785, opened 2024-02-20
chore: improve getFieldValue

Uzlopak:
- improve test coverage which is not handled by wpt tests

### DIFF
--- a/benchmarks/cacheGetFieldValues.mjs
+++ b/benchmarks/cacheGetFieldValues.mjs
@@ -2,9 +2,14 @@ import { bench, group, run } from 'mitata'
 import { getFieldValues } from '../lib/cache/util.js'
 
 const values = [
+  '',
   'foo',
   'invälid',
-  'foo, '
+  'foo, ',
+  'foo, bar',
+  'foo, bar, baz',
+  'foo, bar, baz, ',
+  'foo, bar, baz, , '
 ]
 
 group('getFieldValues', () => {


lib/cache/util.js
```diff
@@ -19,31 +19,33 @@ function urlEquals (A, B, excludeFragment = false) {
...
+
 /**
  * @see https://github.com/chromium/chromium/blob/694d20d134cb553d8d89e5500b9148012b1ba299/content/browser/cache_storage/cache_storage_cache.cc#L260-L262
  * @param {string} header
  */
-function fieldValues (header) {
+function getFieldValues (header) {
   assert(header !== null)
 
+  if (header === '') {
```
KhafraDev, line 31: header can't be empty here

KhafraDev (changes requested):

lib/cache/util.js
Uzlopak: done

KhafraDev (approved):

benchmarks/cacheGetFieldValues.mjs
```diff
@@ -0,0 +1,18 @@
+import { bench, group, run } from 'mitata'
+import { getFieldValues } from '../lib/cache/util.js'
+
+const values = [
+  'foo',
+  'invälid',
+  'foo, '
+]
```
metcoder95, line 8: Can we add some more values?
I believe the ones you added for testing should be enough 👍

Uzlopak: Like this?

```suggestion
const values = [
  '',
  'foo',
  'invälid',
  'foo, ',
  'foo, bar',
  'foo, bar, baz',
  'foo, bar, baz, ',
  'foo, bar, baz, , '
]
```

Uzlopak: Done

metcoder95 (approved):

mcollina (approved):
lgtm

PR #2795, opened 2024-02-21
Dispatch compose

ronag:
## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready


lib/handler/RedirectHandler.js
```diff
@@ -24,13 +24,18 @@ class BodyAsyncIterable {
 }
 
 class RedirectHandler {
-  static buildDispatch (dispatcher, maxRedirections) {
-    if (maxRedirections != null && (!Number.isInteger(maxRedirections) || maxRedirections < 0)) {
-      throw new InvalidArgumentError('maxRedirections must be a positive number')
-    }
+  static decorate (dispatch) {
```
metcoder95, line 27: Early review, feel free to ignore it if you already considered it.

Should we consider the `decorate` to pass custom options?
I see potential there as it can configure the _decorator_ and pass it down to the actual decoration, and maybe even allow to be overwritten at the moment of using it.

Maybe won't be needed for the `RedirectHandler`, but it will for the `Retry` one, wdyt?

ronag: This is just an experiment PR. Feel free to shoot all ideas. I'll have an update later this week.

metcoder95: Sure, this is what I was having in mind:

```js
static decorate (dispatch, handlerOpts) {
  // We validate the options for the handler are valid
  // These will be the defaults if not passed through the dispatch call
  if (handlerOpts?.maxRedirections == null || handlerOpts.maxRedirections === 0) {
  }

  // By splitting it into two phases, we allow them to override the globals if they need to
  // This can also showcase how user-land can create its own custom implementations
  // and even pass custom options to it (e.g. RetryHandler)
  return (opts, handler) => {
    const redirectHandler = new RedirectHandler(dispatch, handlerOpts);
    return dispatch(opts, redirectHandler.handle(opts, handler))
  }
}
```

mcollina: how that would be used?

metcoder95: There's when I go over the fence; sadly this will require implementers to do use the `dispatch` to make it work.
Or, something like you did, create a custom Agent on top (`RetryAgent`).

Maybe this is something good and we should strive implementers to seek for to implement custom behaviours in that way?

ronag: I kind of like @mcollina's pattern in RetryAgent etc. Will give that more thought. My main concern with the current interceptor design is that unnecessarily lives inside the core instead of just wrapping it.

ronag:
@mcollina @metcoder95 PTAL

metcoder95 (commented):
Pretty much like the concept, it will be interesting as it allows chaining and executing the dispatching one after the other, sequentially.

Just left few comments, I'll give a second look and possibly start doing a review more of the actually implementation within Redirect and Retry handler.

Do you have an idea already for the TS types?

lib/interceptors/redirect.js
```diff
@@ -232,4 +229,31 @@ function cleanRequestHeaders (headers, removeContent, unknownOrigin) {
...
+  constructor (dispatcher, opts) {
+    super()
+
+    this.#dispatcher = dispatcher
+    this.#opts = validateOptions(opts)
+  }
+
+  dispatch (opts, handler) {
+    opts = this.#opts ? { ...this.#opts, ...opts } : opts
+    return opts.maxRedirections
```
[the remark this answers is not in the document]
ronag: This would swallow invalid values that should error.

lib/dispatcher.js
```diff
@@ -14,6 +18,20 @@ class Dispatcher extends EventEmitter {
...
     throw new Error('not implemented')
   }
+
+  intercept (...interceptors) {
+    let dispatcher = this
+    for (const interceptor of interceptors) {
+      if (typeof interceptor !== 'function') {
+        throw new Error('invalid interceptor')
+      }
+      dispatcher = interceptor(dispatcher) ?? dispatcher
```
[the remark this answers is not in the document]
ronag: WHat do you mean?

lib/dispatcher.js
```diff
@@ -14,6 +18,20 @@ class Dispatcher extends EventEmitter {
...
+      if (dispatcher[kDispatcherVersion] !== 1) {
+        throw new Error('invalid dispatcher')
+      }
```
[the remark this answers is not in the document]
ronag: It's mostly if we e.g. change the intercept method in an incompatible way. I'm not using instanceof on purpose so that it works if there are different version of undici at the same time.

ronag:
@mcollina Unfortunately this API conflicts with the mock API. I found a hack around it but not sure... wdyt?

lib/dispatcher.js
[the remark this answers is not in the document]
metcoder95: I'm already seeing issues stating that they tried to chain two or three dispatchers and something didn't work as expected; I'm referring that we should state in the docs that each _interceptor_ should return the dispatcher clearly, otherwise, it can lead to inconsistencies.

More a documentation thing

metcoder95:
> @mcollina Unfortunately this API conflicts with the mock API. I found a hack around it but not sure... wdyt? I believe we could probably re-writen the mock API to use the same intercept API but that's beyond me.

Although I agree it overlaps, I'm not 100% we sure change it, they share similar concepts but have different purposes. The mock simplifies the testing quite a lot, especially for simple scenarios; using the `dispatcher.intercept` might be cumbersome for that (but not essential bad).

ronag:
As I see it we have 3 options.

1. Hacky compat where we check the arguments to decide which intercept the user intends.
2. Rename dispatcher.intercept (maybe compose?)
3. Rename mock.intercept

ronag:
Some ideas:

```js

```js
dispatcher = dispatcher.intercept(redirect(redirectOpts), retry(retryOpts))
dispatcher = dispatcher.compose(redirect(redirectOpts), retry(retryOpts))
dispatcher = compose(redirect(redirectOpts), retry(retryOpts))(dispatcher)
dispatcher = intercept(redirect(redirectOpts), retry(retryOpts))(dispatcher)

dispatcher = dispatcher.dispatch(dispatchOpts)
```
```

mcollina:
I would rename `dispatcher.composer`

metcoder95:
+1 on `dispatcher.compose`

ronag:
@mcollina @metcoder95 wdyt? Are we happy with this API? I'm going to leave documenting it as a TODO for merging next into main.

ronag:
I really like how e.g. `ProxyAgent` nicely folds into this pattern.

metcoder95 (approved):
LGTM 🚀 

Happy to port this to `main`

ronag:
> Happy to port this to main

Can you open a separate PR for that and maybe add some documentation as well update the existing interceptor documentation as deprecated?

mcollina:
Having `next` and `main` having two different implementation of the same thing is a recepie for disaster in maintenance and release. I recommend this is reverted in `next`, implemented in `next` and ported to `next`.

I would need to merge main into next one day, and this will make things _very hard_.

ronag:
> . I recommend this is reverted in next, implemented in next and ported to next.

Sorry that's confusing?

mcollina:
This is based on some other changes in the `next` branch and I think the end result would be a mess.

ronag:
Reverted. @metcoder95 have fun with that 😄

PR #2796, opened 2024-02-21
feat!: enable H2 by default

### LINKED ISSUE #2750: `fetch` support for HTTP/2 by default

This will get node's implementation of `fetch` closer to how it behaves in the browser.

Support protocol upgrade similar to how `fetch-h2` package does. It's weird to try to use a third-party package that uses node's internal http2 package under the hood.

I am now using `fetch-h2` package, but much like `got-fetch` or `node-fetch` I fear it will fall out of spec and not evolve with the ecosystem.

metcoder95:
## This relates to...

Closes #2750 

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready


ronag (changes requested):
I'm strongly against this. I think HTTP2 for most cases is worse both in terms of performance and stability. I'm not a big fan of the quality and lack of maintenance of the http2 module in node core. I would like to even further decouple the http2 stuff from undici core. I would be happy to reconsider this position if and when James's http/3 and quic stuff lands.

ronag:
I would suggest we enable it by default for fetch. It's slow anyway :).

metcoder95:
I believe that most of the inquiries come from `fetch`, so I am happy to scope it for it only 🙂

mcollina (changes requested):
I'm -1 with this implementation.

I think a better approach would be to enable it _only_ if the server does not advertise HTTP/1.1 in TLS. In other terms, if a server is only exposed via HTTP/2, we can support it.

mcollina:
I think we are not ready for turning this on for all requests, there are bugs: https://github.com/nodejs/undici/issues/2801.

metcoder95:
SGTM, let's take this back and circle once we feel we are more prepared to it; I'll revisit the bug and seek for support and possible add a fix there 👍

PR #2800, opened 2024-02-21
Use paralellelRequests instead of connections to calculate req/sec in benchmarks

mcollina:
Preparing some slides I noticed that our benchmarks where a bit off, and I discovered that we were making a mistake, specifically we were multiplying our results for the number of connections and not for the number of parallel requests.

I think the math makes sense now, but PTAL.


ShogunPanda (approved):
LGTM!

Uzlopak (approved):
LGTM

KhafraDev (approved):

metcoder95 (approved):

PR #2809, opened 2024-02-22

tsctx:
Reopening #2718. It was closed by mistake.


ronag (changes requested):
This is way too micro optimization.

tsctx:
Ah. I see.

PR #2816, opened 2024-02-22
refactor: try to reduce amount of http2 specific logic

ronag:
## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready


ronag:
@metcoder95 some more changes, I think you get the gist

lib/client.js
```diff
@@ -1226,21 +1110,39 @@ async function connect (client) {
...
+        onSocketClose.call(session)
+      })
       session.on('close', onSocketClose)
       session.unref()
 
-      client[kHTTP2Session] = session
-      socket[kHTTP2Session] = session
+      session[kOpenStreams] = 0
+
+      client[kSocket] = session
```
metcoder95, line 1145: Do you think it can have potential side-effects?
They are a duplex stream, so they share APIs but the way that they are used is different 🤔

metcoder95 (commented):
So far, the idea LGTM; I would like to see if we can scope the H2 logic even more, but we can do it in another PR, I'll check the test first 👍

lib/client.js
ronag: I *think* in the way it is currently used it does not matter.

lib/core/util.js
```diff
@@ -192,7 +204,7 @@ function isReadableAborted (stream) {
 }
 
 function destroy (stream, err) {
-  if (stream == null || !isStream(stream) || isDestroyed(stream)) {
+  if (stream == null || (!isStream(stream) && !isH2Session(stream)) || isDestroyed(stream)) {
```
ronag, line 207: We don't need to destroy h2 sessions?

metcoder95: We need, now that the H2 session has been moved into a socket-like handling.
As the `isStream` was returning `false`, the sessions were not shutdown and the connection remain opened until timeout

ronag: aha, then maybe the istream check is wrong, maybe `typeof obj.destroy === 'function'` is sufficient?

ronag: I though sessions were streams?

metcoder95: Hmm, yes if we just want to destroy them; but I think `isStream` is used elsewhere and that might be more problematic 🤔 

>I though sessions were streams?

Yeah, me too; had hard time guessing why was not being picked 😅

lib/client.js
```diff
@@ -1226,21 +1108,40 @@ async function connect (client) {
...
+        if (id === 0) {
+          const err = new InformationalError(`HTTP/2: "frameError" received - type ${type}, code ${code}`)
+          util.destroy(this, err)
+          util.destroy(socket, err)
+        }
+      })
+      session.on('end', function () {
+        const err = new SocketError('other side closed')
+        util.destroy(this, err)
+        util.destroy(socket, err)
```
ronag, line 1132:
```suggestion
        this.destroy(err)
        socket.destroy(err)
```

ronag: etc...

ronag:
> do you want me to wait for #2839 to land? In that way we avoid dealing with possible conflicts; anyways we already know where the issue roots to

Sorry for that one. Just got an idea and went with it. Hopefully didn't cause you unnecessary work.

PR #2817, opened 2024-02-22

Uzlopak:
based on #2809 by @tsctx 

While slice keeps the order, sliceOneUnordered doesnt care about the order and can so improve the performance by a factor of 10. 

I think it is a low hanging fruit and not a micro optimization.

### DIFF
--- a/lib/client.js
+++ b/lib/client.js
@@ -1463,7 +1463,7 @@ function _resume (client, sync) {
     if (!request.aborted && write(client, request)) {
       client[kPendingIdx]++
     } else {
-      client[kQueue].splice(client[kPendingIdx], 1)
+      util.spliceOneUnordered(client[kQueue], client[kPendingIdx])
     }
   }
 }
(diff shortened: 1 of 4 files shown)


lib/client.js
```diff
@@ -1463,7 +1463,7 @@ function _resume (client, sync) {
     if (!request.aborted && write(client, request)) {
       client[kPendingIdx]++
     } else {
-      client[kQueue].splice(client[kPendingIdx], 1)
+      util.spliceOneUnordered(client[kQueue], client[kPendingIdx])
```
tsctx, line 1466: It is necessary to keep the order.
This becomes the last request instead of the aborted request, and the order in which requests are executed is out of order.

Uzlopak:
```suggestion
      client[kQueue].splice(client[kPendingIdx], 1)
```

Uzlopak:
@tsctx 

The other places are ok? What about the splice calls in cache? Are they ordered or unorderedß

tsctx:
This is defined as a [tuple](https://infra.spec.whatwg.org/#tuple), and the order must kept.

ronag:
Guys. This is micro optimizing cold paths. No user will practically benefit from this.

PR #2824, opened 2024-02-23
byob WPTs pass now

KhafraDev:


KhafraDev:
I guess v18 has a bug

PR #2826, opened 2024-02-23
feat: add new dispatch compose

metcoder95:
## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready


metcoder95:
Next, time to fight with tests and adapt it

ronag (approved):

lib/interceptor/redirect.js
```diff
@@ -0,0 +1,54 @@
+const { InvalidArgumentError } = require('../core/errors')
```
mcollina, line 1:
```suggestion
'use strict'

const { InvalidArgumentError } = require('../core/errors')
```

lib/interceptor/retry.js
```diff
@@ -0,0 +1,33 @@
+const Dispatcher = require('../dispatcher-base')
```
mcollina, line 1:
```suggestion
'use strict'

const Dispatcher = require('../dispatcher-base')
```

mcollina (changes requested):
Missing docs and tests

lib/interceptor/proxy.js
```diff
@@ -27,36 +27,47 @@ class ProxyAgent extends DispatcherBase {
...
     this[kProxy] = { uri: href, protocol }
     this[kAgent] = new Agent(opts)
-    this[kInterceptors] = opts.interceptors?.ProxyAgent && Array.isArray(opts.interceptors.ProxyAgent)
-      ? opts.interceptors.ProxyAgent
-      : []
+    this[kInterceptors] =
+      opts.interceptors?.ProxyAgent &&
+      Array.isArray(opts.interceptors.ProxyAgent)
+        ? opts.interceptors.ProxyAgent
+        : []
```
ronag, line 53: unrelated change?

lib/interceptor/proxy.js
```diff
@@ -27,36 +27,47 @@ class ProxyAgent extends DispatcherBase {
...
     this[kRequestTls] = opts.requestTls
     this[kProxyTls] = opts.proxyTls
     this[kProxyHeaders] = opts.headers || {}
 
     if (opts.auth && opts.token) {
-      throw new InvalidArgumentError('opts.auth cannot be used in combination with opts.token')
+      throw new InvalidArgumentError(
+        'opts.auth cannot be used in combination with opts.token'
+      )
```
ronag, line 61: unrelated change?

lib/interceptor/proxy.js
```diff
@@ -27,36 +27,47 @@ class ProxyAgent extends DispatcherBase {
...
     } else if (opts.auth) {
       /* @deprecated in favour of opts.token */
       this[kProxyHeaders]['proxy-authorization'] = `Basic ${opts.auth}`
     } else if (opts.token) {
       this[kProxyHeaders]['proxy-authorization'] = opts.token
     } else if (username && password) {
-      this[kProxyHeaders]['proxy-authorization'] = `Basic ${Buffer.from(`${decodeURIComponent(username)}:${decodeURIComponent(password)}`).toString('base64')}`
+      this[kProxyHeaders]['proxy-authorization'] = `Basic ${Buffer.from(
+        `${decodeURIComponent(username)}:${decodeURIComponent(password)}`
+      ).toString('base64')}`
```
ronag, line 70: unrelated change?

lib/interceptor/proxy.js
```diff
@@ -82,7 +93,11 @@ class ProxyAgent extends DispatcherBase {
           })
           if (statusCode !== 200) {
             socket.on('error', () => {}).destroy()
-            callback(new RequestAbortedError(`Proxy response (${statusCode}) !== 200 when HTTP Tunneling`))
+            callback(
+              new RequestAbortedError(
+                `Proxy response (${statusCode}) !== 200 when HTTP Tunneling`
+              )
+            )
```
ronag, line 100: unrelated change?

lib/interceptor/proxy.js
```diff
@@ -94,7 +109,10 @@ class ProxyAgent extends DispatcherBase {
           } else {
             servername = opts.servername
           }
-          this[kConnectEndpoint]({ ...opts, servername, httpSocket: socket }, callback)
+          this[kConnectEndpoint](
+            { ...opts, servername, httpSocket: socket },
+            callback
+          )
```
ronag, line 115: unrelated change?

ronag:
I would also like us to have some kind of benchmark to this pattern to make sure it doesn't add too much overhead.

lib/interceptor/proxy.js
metcoder95: Yeah, renamed the file and the lint did its thing

metcoder95:
@ronag, did the Proxy worked for you?
I'm trying to add some testing, and it seems it will require several pieces to be changed; especially around the way we handle withe dispatch usage for `connect`

ronag:
I never tried it.

metcoder95:
Well, that's because we I've not pushed the tests for the compose feature; still in my machine and constantly failing, on top I forgot to add the extra dispatcher on the constructor for the Proxy.

But its fine, I just wanted to be sure I had the whole picture. Most likely I'll put back the Agent where it was, and copy/paste and adapt a new copy of the Agent proxy to take into account the new dispatcher arg

docs/docs/api/Interceptors.md
```diff
@@ -0,0 +1,68 @@
...
+
+#### Example - Basic Proxy Interceptor
+
+```js
+const { Client, interceptors } = require("undici");
+const { proxy } = interceptors;
+
+const client = new Client("http://example.com");
+
+client.compose(proxy("http://proxy.com"));
```
ronag, line 25:
```suggestion
const client = new Client("http://example.com").compose(proxy("http://proxy.com"));
```

docs/docs/api/Interceptors.md
```diff
@@ -0,0 +1,68 @@
...
+
+#### Example - Basic Redirect Interceptor
+
+```js
+const { Client, interceptors } = require("undici");
+const { redirect } = interceptors;
+
+const client = new Client("http://example.com");
+
+client.compose(redirect({ maxRedirections: 3, throwOnMaxRedirects: true }));
```
ronag, line 42:
```suggestion
const client = new Client("http://example.com").compose(redirect({ maxRedirections: 3, throwOnMaxRedirects: true }));
```

docs/docs/api/Interceptors.md
```diff
@@ -0,0 +1,68 @@
...
+
+client.compose(
+  retry({
+    maxRetries: 3,
+    minTimeout: 1000,
+    maxTimeout: 10000,
+    timeoutFactor: 2,
+    retryAfter: true,
+  })
+);
```
ronag, line 67:
```suggestion
const client = new Client("http://example.com").compose(
  retry({
    maxRetries: 3,
    minTimeout: 1000,
    maxTimeout: 10000,
    timeoutFactor: 2,
    retryAfter: true,
  })
);
```

docs/docs/api/Dispatcher.md
```diff
@@ -817,6 +817,131 @@ try {
...
+
+* **interceptors** `Interceptor[]`: It is an array of `Interceptor` functions passed as only argument, or several interceptors passed as separate arguments.
+
+Returns: `Dispatcher`.
+
+#### Parameter: `Interceptor`
+
+A function that takes a `Dispatcher` instance and returns a `Dispatcher` instance.
+
+#### Example 1 - Basic Compose
```
mcollina, line 839: Why would somebody use this while we provide a the `RedirectDispatcher` ourselves? Let's work on how people would use this API.

lib/dispatcher/dispatcher.js
```diff
@@ -1,8 +1,11 @@
 'use strict'
-
 const EventEmitter = require('node:events')
 
+const kDispatcherVersion = Symbol.for('undici.dispatcher.version')
```
mcollina, line 4: Is this the same simbol we use for the global? I think we should use the same.

lib/dispatcher/dispatcher.js
```diff
@@ -14,6 +17,28 @@ class Dispatcher extends EventEmitter {
...
+    for (const interceptor of interceptors) {
+      if (interceptor == null) {
+        continue
+      }
+
+      if (typeof interceptor !== 'function') {
+        throw new Error('invalid interceptor')
+      }
+
+      dispatcher = interceptor(dispatcher) ?? dispatcher
```
mcollina, line 34: Can you clarify why you are doinng the `??` here? What use case are you trying to support?

As an added benefit, we could support a "shortcut" if the function returns a function `dispatch`, and we automatically add the `close` and `destroy`.

lib/interceptor/proxy.js
```diff
@@ -0,0 +1,39 @@
...
+
+  if (!opts || (!opts.uri && !(opts instanceof URL))) {
+    throw new InvalidArgumentError('Proxy opts.uri or instance of URL is mandatory')
+  }
+
+  if (opts.auth && opts.token) {
+    throw new InvalidArgumentError(
+      'opts.auth cannot be used in combination with opts.token'
+    )
+  }
```
mcollina, line 36: move those in the constructor

lib/interceptor/proxy.js
```diff
@@ -0,0 +1,39 @@
+'use strict'
+
+const { InvalidArgumentError } = require('../core/errors')
+const ProxyAgent = require('../dispatcher/proxy-agent')
+const DispatcherBase = require('../dispatcher/dispatcher-base')
+
+class ProxyInterceptor extends DispatcherBase {
```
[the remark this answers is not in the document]
mcollina: I think this should extend `Dispatcher`, not `DispatcherBase`.

mcollina (changes requested):

lib/interceptor/proxy.js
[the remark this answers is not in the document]
metcoder95: >Please integrate this inside the ProxyAgent. We shouldn't have special classes for these.

Sure, I think now with the changes I already did finally found what I was missing; will check and refactor

lib/interceptor/proxy.js
metcoder95: SGTM, but then we will lose the immediate feedback, as it will be delayed until used by the `dispatcher.compose` logic.
So if implementers do:
```js
const interceptor = proxyInterceptor() // <- this won't fail, until next line
dispatcher.compose(interceptor) // <- here it will be thrown
```

mcollina: I'm proposing we drop those wrapper functions and just expose Dispatchers. They only add noise.

docs/docs/api/Dispatcher.md
ronag: I mean the point is to make it easy to compose these kinds of "interceptor" dispachers, i.e. dispatchers that take another dispatcher + options in its constructor.

ronag (commented):
I slightly disagree with @mcollina. I don't see any point in exposing the dispatcher classes themselves (why would you extend them)? Rather I would prefer exposing curry-able factory methods. That also makes the distinction between composable and non composable dispatcher clearer.

lib/dispatcher/dispatcher.js
metcoder95: Hmm, good point; I believe I brought similar point on the previous PR, overall by doing this we allow implementers to branch the dispatcher chain, which can cause undesired side-effects. Better be safer than sorry and if the compose callback returns either `null` or `undefined` we just ignore it (add a composed validation that checks for a `dispatcher-like` object).

Following the `dispatcher-like`, I'd say that if we introduce this pattern would be better to have some limitations to avoid misunderstanding and constant issue raising.
Just validating the `dispatch` is there, along with `close` and `destroy` should be good to go, but I wouldn't attache automatically `close` and `destroy` as those potentially can cause side-effects users are not aware of.

The last being said, wdyt about validating the `dispatcher-like` shape and documenting it?

lib/interceptor/redirect.js
```diff
@@ -0,0 +1,44 @@
+'use strict'
+
+const { InvalidArgumentError } = require('../core/errors')
+const Dispatcher = require('../dispatcher/dispatcher-base')
+const RedirectHandler = require('../handler/RedirectHandler')
+
+class RedirectDispatcher extends Dispatcher {
```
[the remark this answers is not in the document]
metcoder95: We can add it as a `RedirectAgent`

metcoder95:
> I slightly disagree with @mcollina. I don't see any point in exposing the dispatcher classes themselves (why would you extend them)? Rather I would prefer exposing curry-able factory methods. That also makes the distinction between composable and non composable dispatcher clearer.

I pushed the change to put both options over the table and chat about which one we would like to follow.
On this one, I agree with @ronag the dispatcher classes made for the interceptors might be better to keep internal, as we are looking to introduce the concept of `composable-dispatcher`. Otherwise it kind of becomes more like building Agents we already have the concept of

index.js
```diff
@@ -36,6 +36,11 @@ module.exports.RetryHandler = RetryHandler
 module.exports.DecoratorHandler = DecoratorHandler
 module.exports.RedirectHandler = RedirectHandler
 module.exports.createRedirectInterceptor = createRedirectInterceptor
+module.exports.interceptors = {
+  Proxy: require('./lib/interceptor/proxy'),
```
ronag, line 40:
```suggestion
  proxy: require('./lib/interceptor/proxy'),
```

lib/dispatcher/dispatcher.js
mcollina: Agreed

mcollina:
>  I don't see any point in exposing the dispatcher classes themselves (why would you extend them)? Rather I would prefer exposing curry-able factory methods. That also makes the distinction between composable and non composable dispatcher clearer.

Then we should simplify this API greatly, deprecating/removing a bunch of things that we are currently exposing, and allowing the interceptor functions to only return the `dispatch` function, so that the wrapping of the close and destroy is not required.

ronag:
> Reference

Agreed. The less we expose the less trouble.

metcoder95:
SGTM 👍 

On another topic, I'm starting to see some important highlights or limitations that we need to either document or adjust.

The ones I'm talking about are:

1. `Proxy` interceptor needs to be the very first dispatcher in the chain, otherwise it is not designed to preserve the chain before it does its job.
	- The Proxy Agent/Dispatcher is designed to forward the request with a given Proxy origin, does the required auth work and headers preparations, so we are forced to execute its `dispatch` no matter what if we want to ensure backwards compatibility (ref: [ProxyAgent](https://github.com/nodejs/undici/blob/e343948ec3a8b1052f50f80b109692dfd545f4eb/lib/dispatcher/proxy-agent.js#L65), and [ProxyInterceptor](https://github.com/nodejs/undici/pull/2826/files#diff-81fa2458aa80c4c8efea0c5931b912e7f9f561c1d95d4e205e623059b94ebf72R37))
	- I'm not pretty sure how a possible rework of it will look like yet
2. `api-request` can break some of the handlers as it is designed to forward the control back to the caller after the headers are received
	- This already causes the `retry` interceptor to not work as expected, as in the case of a socket disconnect or similar and the body hasn't been received, the control has been already passed up to the caller, so any further retry and handling of the retry goes out of the stack when it is handled as the promise or callback has been already resolved/called.
	- This problem was already present, but we didn't test it with the actual `request` API, so any change might be a breaking one; you are forced to wait for the body (e.g. `await response.body.text()` to properly wait until it is solved.

What are your thoughts?

Overall, I believe documentation should be enough but happy to hear your thoughts about

metcoder95:
I'll work over the weekend on the refactor to expose only the `dispatch` function instead of the whole dispatcher-like 👍

ronag:
> api-request can break some of the handlers as it is designed to forward the control back to the caller after the headers are received

I don't see the problem? Can you give a concrete example?

metcoder95:
Sure, take a look at the following test; it is designed to handle the scenario of a bad-etag while the request failed after receiving the headers (socket closed), and for instance no `body` was ever received:

```js
test('Should handle 206 partial content - bad-etag', async t => {
  t = tspl(t, { plan: 3 })

  // Took from: https://github.com/nxtedition/nxt-lib/blob/4b001ebc2f22cf735a398f35ff800dd553fe5933/test/undici/retry.js#L47
  let x = 0
    if (x === 0) {
      t.ok(true, 'pass')
      res.setHeader('etag', 'asd')
      res.write('abc') // we do not end the response body correctly.
      setTimeout(() => {
        res.destroy()
      }, 1e2)
    } else if (x === 1) {
      t.deepStrictEqual(req.headers.range, 'bytes=3-')
      res.setHeader('content-range', 'bytes 3-6/6')
      res.setHeader('etag', 'erwsd')
      res.statusCode = 206
      res.end('def')
    }
    x++
  })

  const requestOptions = {
    method: 'GET',
    path: '/',
    headers: {
      'content-type': 'application/json'
    },
    retryOptions: {
      retry: (err, { state, opts }, done) => {
        if (err.message.includes('other side closed')) {
          setTimeout(done, 100)
          return
        }

        return done(err)
      }
    }
  }

  server.listen(0)

  const client = new Client(
    `http://localhost:${server.address().port}`
  ).compose(retry())

  after(async () => {
    await client.close()
    server.close()

    await once(server, 'close')
  })

  try {
    const response = await client.request(requestOptions)
    // await response.body.text() // <- if we uncomment this section, the flow will move and eventually go out of the stack; though the retry handler hasn't finished its job.
  } catch (error) {
    t.strict(error, {
      message: 'ETag mismatch',
      code: 'UND_ERR_REQ_RETRY',
      name: 'RequestRetryError'
    })
  }
})
```

Haven't tried adding a timeout, but just by the fact that we return the control to the caller at that point, even tho we have not received the body yet cause the `retry` handler continue its job but have not way to inform back to the caller about the updated status and the full body.

ronag:
I'll have a look but I still don't see any conceptual problem. Sounds more like an implementations issue to me.

metcoder95:
Have similar thought, but still not sure, as in fact we already received the headers but we haven't received the body at that point; I'll wait for your feedback, in the meanwhile moving with the rest of interceptors 👍

metcoder95:
I'll come with the TS types and documentation within the week, the implementation should be ready for review

lib/dispatcher/dispatcher.js
```diff
@@ -14,6 +13,30 @@ class Dispatcher extends EventEmitter {
...
+        throw new Error('invalid interceptor')
+      }
+
+      const newDispatch = interceptor(this)
+
+      if (newDispatch == null || typeof newDispatch !== 'function' || newDispatch.length !== 2) {
+        throw new Error('invalid interceptor')
+      }
+
+      this.dispatch = newDispatch
```
ronag, line 35: Return a new dispatcher instead of mutating.

lib/interceptor/proxy.js
```diff
@@ -0,0 +1,11 @@
...
+const ProxyAgent = require('../dispatcher/proxy-agent')
+
+module.exports = opts => {
+  const agent = new ProxyAgent(opts)
+  return () => {
+    return function proxyInterceptor (opts, handler) {
+      return agent.dispatch(opts, handler)
+    }
+  }
+}
```
ronag, line 11: This isn't an interceptor?

lib/interceptor/redirect.js
```diff
@@ -0,0 +1,21 @@
+'use strict'
+const RedirectHandler = require('../handler/redirect-handler')
+
+module.exports = opts => {
+  const globalMaxRedirections = opts?.maxRedirections
+  return dispatcher => {
+    const dispatch = dispatcher.dispatch.bind(dispatcher)
+
+    return function redirectInterceptor (opts, handler) {
```
ronag, line 9: Shouldn't an interceptor take dispatch, opts, handler?

ronag (changes requested):

lib/interceptor/proxy.js
ronag, line 11:
```suggestion
module.exports = opts => {
  return (dispatch) => {
    const agent = new ProxyAgent(dispatch, opts)
    dispatch = agent.dispatch.bind(agent)
    return function proxyInterceptor (opts, handler) {
      return agent.dispatch(opts, handler)
    }
  }
}
```

metcoder95: That's what I was referring to when I suggested that `ProxyAgent` works differently 😅 
It requires to hope over the whole dispatch, as it forwards the request to the proxy instead of the intended dispatch.
It does not respects the previous dispatcher, so if proxy is used, it should be placed at the very beginning of the chain.

I couldn't think of any other way without re-working the `ProxyAgent`, but totally open to suggestions

lib/dispatcher/dispatcher.js
```diff
@@ -14,6 +13,32 @@ class Dispatcher extends EventEmitter {
...
+      }
+
+      dispatch = interceptor(dispatch)
+
+      if (dispatch == null || typeof dispatch !== 'function' || dispatch.length !== 2) {
+        throw new Error('invalid interceptor')
+      }
+    }
+
+    this.dispatch = dispatch
```
ronag, line 38: Create a new dispatcher instead of mutating existing one

ronag (changes requested):

lib/dispatcher/dispatcher.js
```diff
@@ -14,6 +13,32 @@ class Dispatcher extends EventEmitter {
...
+    // So we handle [interceptor1, interceptor2] or interceptor1, interceptor2, ...
+    const interceptors = Array.isArray(args[0]) ? args[0] : args
+    let dispatch = this.dispatch.bind(this)
+
+    for (const interceptor of interceptors) {
+      if (interceptor == null) {
+        continue
+      }
+
+      if (typeof interceptor !== 'function') {
```
ronag, line 27: I still think we should allow return dispatcher shape?

ronag (changes requested):

lib/interceptor/proxy.js
ronag: I will have to investigate this

lib/interceptor/retry.js
```diff
@@ -0,0 +1,18 @@
+'use strict'
+const RetryHandler = require('../handler/retry-handler')
+
+module.exports = globalOpts => {
+  return dispatch => {
+    return function retryInterceptor (opts, handler) {
+      opts.retryOptions = { ...globalOpts, ...opts.retryOptions }
```
ronag, line 7: I don't think we need to mutate opts here. Just make a rw copy to pass to the handler.

ronag (changes requested):

lib/dispatcher/dispatcher.js
metcoder95: Hmm, not sure, but maybe I'm not seeing the full picture; what would be the benefit of exposing the whole dispatcher shape?

ronag: Not sure, mostly if the interceptor needs some kind of disposing.

lib/dispatcher/dispatcher.js
metcoder95: Done 👍

lib/dispatcher/dispatcher.js
metcoder95: How it should work?
Meaning, if I return a `dispatcher`-like, how the dispatcher should handle the rest of the properties, given that it will just allow to _patch_ the `dispatch` function 🤔

lib/dispatcher/dispatcher.js
```diff
@@ -25,19 +25,17 @@ class Dispatcher extends EventEmitter {
...
       if (dispatch == null || typeof dispatch !== 'function' || dispatch.length !== 2) {
-        throw new Error('invalid interceptor')
+        throw new TypeError('invalid interceptor')
       }
     }
 
-    this.dispatch = dispatch
-
-    return this
+    return Object.setPrototypeOf({ dispatch }, this)
```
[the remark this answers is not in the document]
ronag: Also, I don't think this will properly forward close/destroy? Should we have some kind of reference count?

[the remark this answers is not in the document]
metcoder95: >This is a bit weird? Why not just make a WrappedDispatcher class?

Hmm, that can be slightly hard, as each Dispatcher will have its own custom prototype implementation, although everyone extends from `Dispatcher`. Preserving the `prototype` is the most important thing, how do you propose to use the `WrappedDispatcher`?


>Also, I don't think this will properly forward close/destroy? Should we have some kind of reference count?

It seems like not, missed that, tests are failing most likely due to that; I'll check that out.
But what do you have in mind with reference count?
Make the `Dispatcher` aware of the number of instances wrapped?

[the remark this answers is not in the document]
ronag:
```
class WrappedDispatcher extends Dispatcher {
  #dispatch
  #dispatcher

  constructor(dispatch, dispatcher) {
    this.#dispatch = dispatch
  }

  dispatch(...args) {
    this.#dispatch(...args)
  }

  close (...args) {
    return this.#dispatcher.close(...args)
  }

  destroy (...args) {
    return this.#dispatcher.destroy(...args)
  }
}
```

mcollina (commented):
Good work! types are missing

docs/docs/api/Dispatcher.md
```diff
@@ -817,6 +817,160 @@ try {
 }
 ```
 
+### `Dispatcher.compose(interceptors[, interceptor])`
```
mcollina, line 820: are those interceptors the same signature of the interceptor options?

metcoder95: Yes, they should follow the same signature

lib/interceptor/redirect.js
```diff
@@ -0,0 +1,24 @@
...
+      const redirectHandler = new RedirectHandler(
+        dispatch,
+        maxRedirections,
+        opts,
+        handler
+      )
+      opts = { ...opts, maxRedirections: 0 } // Stop sub dispatcher from also redirecting.
+      return dispatch(opts, redirectHandler)
+    }
+  }
```
ronag, line 23:
```suggestion
      const { maxRedirections = globalMaxRedirections, ...opts2 } = opts

      if (!maxRedirections) {
        return dispatch(opts, handler)
      }

      const redirectHandler = new RedirectHandler(
        dispatch,
        maxRedirections,
        opts,
        handler
      )
      return dispatch(opts2, redirectHandler)
    }
  }
```

lib/dispatcher/dispatcher.js
```diff
@@ -14,6 +13,53 @@ class Dispatcher extends EventEmitter {
   destroy () {
     throw new Error('not implemented')
   }
+
+  compose (...args) {
+    // So we handle [interceptor1, interceptor2] or interceptor1, interceptor2, ...
+    const interceptors = Array.isArray(args[0]) ? args[0] : args
```
ronag, line 19:
```suggestion
    const interceptors = args.flat()
```

ronag (approved):

docs/docs/api/Dispatcher.md
```diff
@@ -817,6 +817,160 @@ try {
...
+Compose a new dispatcher from the current dispatcher and the given interceptors.
+
+> _Notes_:
+> - The order of the interceptors matters. The first interceptor will be the first to be called.
+> - It is important to note that the `interceptor` function should return a function that follows the `Dispatcher.dispatch` signature.
+> - Any fork of the chain of `interceptors` can lead to unexpected results.
+
+Arguments:
+
+* **interceptors** `Interceptor[interceptor[]]`: It is an array of `Interceptor` functions passed as only argument, or several interceptors passed as separate arguments. It can also support sevearl interceptors passed as positional arguments
```
metcoder95, line 831:
```suggestion
* **interceptors** `Interceptor[interceptor[]]`: It is an array of `Interceptor` functions passed as only argument, or several interceptors passed as separate arguments.
```

docs/docs/api/Dispatcher.md
```diff
@@ -817,6 +817,160 @@ try {
...
+
+Returns: `Dispatcher`.
+
+#### Parameter: `Interceptor`
+
+A function that takes a `dispatch` instance and returns a `dispatch`-like function.
```
metcoder95, line 837:
```suggestion
A function that takes a `dispatch` method and returns a `dispatch`-like function.
```

docs/docs/api/Dispatcher.md
```diff
@@ -817,6 +817,160 @@ try {
...
+        return dispatch(opts, handler)
+      }
+
+      const redirectHandler = new RedirectHandler(
+        dispatch,
+        maxRedirections,
+        opts,
+        handler
+      )
+      opts = { ...opts, maxRedirections: 0 } // Stop sub dispatcher from also redirecting.
```
metcoder95, line 888:
```suggestion
      opts = { ...opts, maxRedirections: 0 }
```

lib/interceptor/proxy.js
ronag: It sounds to me like this should not be a interceptor.

ronag (changes requested):
LGTM. Other than it doesn't make sense for proxy to be an interceptor.

ronag:
@mcollina PTAL

ronag:
Actually `ProxyAgent` might make sense as an interceptor *if* `[kClient]` is the outer dispatcher.

metcoder95:
> Actually `ProxyAgent` might make sense as an interceptor _if_ `[kClient]` is the outer dispatcher.

Your initial suggestion was doing that somehow, but that will mean break the current `ProxyAgent`; wdyt if we drop the `ProxyInterceptor` and we migrate the `ProxyAgent` to also be an interceptor in the next major?

ronag:
> > Actually `ProxyAgent` might make sense as an interceptor _if_ `[kClient]` is the outer dispatcher.
> 
> Your initial suggestion was doing that somehow, but that will mean break the current `ProxyAgent`; wdyt if we drop the `ProxyInterceptor` and we migrate the `ProxyAgent` to also be an interceptor in the next major?

It doesn't necessarily break it if the interceptor just passed a client factory that return the outer dispatcher?

metcoder95:
> > > Actually `ProxyAgent` might make sense as an interceptor _if_ `[kClient]` is the outer dispatcher.
> > 
> > 
> > Your initial suggestion was doing that somehow, but that will mean break the current `ProxyAgent`; wdyt if we drop the `ProxyInterceptor` and we migrate the `ProxyAgent` to also be an interceptor in the next major?
> 
> It doesn't necessarily break it if the interceptor just passed a client factory that return the outer dispatcher?

Right, that's true; but also we use the `client.connect` to obtain the socket from the Proxy and pass it down to the connector for doing an upgrade if necessary.
We would need to provide a wrapper around that as well; but its true might be doable within `minor`

ronag:
We can also just skip making it an interceptor. I'm fine with that.

lib/interceptor/proxy.js
ronag: Let's skip this for now so we can land.

ronag (approved):

metcoder95:
Dropped the `proxy` interceptor 👍

Issue #2828, opened 2024-02-23
Bind dispatcher to Request Object


matthieusieben:
## This would solve...

From the whatwg spec we can [read](https://github.com/nodejs/undici/blob/7623f106e2ff9eb2cd9c38a80e0e7be0a78f0f61/lib/web/fetch/index.js#L130-L133) `Let requestObject be the result of invoking the initial value of Request as constructor with input and init as arguments.`

This makes it seem like the two following are exactly equivalent, no matter what `input` and `init` are:

```ts
fetch(input, init)
fetch(new Request(input, init))
```

This is mostly true, with the exception of `init.dispatcher` (I know, not defined by the spec):

https://github.com/nodejs/undici/blob/7623f106e2ff9eb2cd9c38a80e0e7be0a78f0f61/lib/web/fetch/index.js#L242

I believe it would make more sense to be able to attach a dispatcher directly to the request object so that the behavior defined by the spec also applies to the custom(s) options added by `undici`.

## The implementation should look like...

Similar to the `body` Request attribute, `dispatcher` would be set as an attribute of the request that can be overridden using `init`.

## I have also considered...

Status quo implies that if we are only working with Request object as `input`, we still need to carry over the `init` for the sole purpose of being able to pass a custom `dispatcher` to undici.

## Additional context

I created a bunch of utilities to work with `fetch()`. Being able to work with a single value makes a lot of things simpler. Consider for example the following wrapper:

```ts
const logWrap = fetch => async (input, init) => {
  const request = new Request(input, init) // Should be fine per Whatwg spec
  console.log(request.method, request.uri)
  const response = await fetch(request)
  console.log(response.status)
  return response
}
```

Being able to use `new Request` to be able to compute the merge of `input` and `init` is super useful. This causes however an issue when working with `undici` because of the additionnal `dispatcher` option.

Fixing this requires to do the following:

```ts
const logWrap = fetch => async (input, init) => {
  const request = new Request(input, init)
  console.log(request.method, request.uri)
  const response = await fetch(request, { dispatcher: init?.dispatcher }) // undici specific
  console.log(response.status)
  return response
}
```

This implementation is not agnostic of the actual `fetch` implementation used under the hood, which breaks separation of concerns.

matthieusieben:
Thanks guys ❤️

PR #2831, opened 2024-02-23
add dispatcher option to Request

KhafraDev:
Fixes https://github.com/nodejs/undici/issues/2828

It also makes the dispatcher option in `fetching` optional

### DIFF
--- a/lib/web/fetch/request.js
+++ b/lib/web/fetch/request.js
@@ -24,7 +24,7 @@ const {
   requestDuplex
 } = require('./constants')
 const { kEnumerableProperty } = util
-const { kHeaders, kSignal, kState, kGuard, kRealm } = require('./symbols')
+const { kHeaders, kSignal, kState, kGuard, kRealm, kDispatcher } = require('./symbols')
 const { webidl } = require('./webidl')
 const { getGlobalOrigin } = require('./global')
 const { URLSerializer } = require('./dataURL')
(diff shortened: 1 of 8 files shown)


ronag (approved):

KhafraDev:
It's already in the types :)

KhafraDev:
https://github.com/nodejs/undici/blob/669a9241238fb7599c85ccaad524db511f4709f5/types/fetch.d.ts#L117

mcollina (approved):
lgtm

lib/web/fetch/request.js
```diff
@@ -101,6 +103,8 @@ class Request {
       // 5. Set fallbackMode to "cors".
       fallbackMode = 'cors'
     } else {
+      this[kDispatcher] = input[kDispatcher]
```
matthieusieben, line 106: Shouldn't this like look like this? Allowing to override the dispatcher when creating a new request from an existing one.

```suggestion
      this[kDispatcher] = init.dispatcher || input[kDispatcher]
```

matthieusieben (commented):
@KhafraDev it seems that with your changes, it is no longer possible to override the dispatcher when creating a request from an existing one. Is this a desired behavior?

lib/web/fetch/request.js
Uzlopak: Can you Provide a Pr and test?

matthieusieben: https://github.com/nodejs/undici/pull/2928

PR #2834, opened 2024-02-24
refactor: remove own sort logic

tsctx:
On second thought, large arrays do not need their own sorting logic. Mostly because they are small headers and benefit from fast-path.


mcollina (approved):
lgtm

PR #2842, opened 2024-02-25
remove signal validation from request

KhafraDev:
webidl handles this for us already


ronag:
CI fails

KhafraDev:
right, we allow third party abortcontrollers in fetch which seems to contradict the node-fetch tests we run... is that something we could drop in next?

PR #2843, opened 2024-02-25
unite webidl stringification

KhafraDev:
Before:
```js

webidl.converters.BufferSource(Object.create(null))
// TypeError: Cannot convert object to primitive value

webidl.converters.BufferSource(Symbol('t'))
// TypeError: Cannot convert a Symbol value to a string

webidl.converters.BufferSource('')
// TypeError: Could not convert  to a BufferSource.
// convert what?

```

After:
```js

webidl.converters.BufferSource(Object.create(null))
// TypeError: Could not convert {} to a BufferSource.

webidl.converters.BufferSource(Symbol('t'))
// TypeError: Could not convert Symbol(t) to a BufferSource.

webidl.converters.BufferSource('')
// TypeError: Could not convert "" to a BufferSource.

### DIFF
--- a/lib/web/fetch/webidl.js
+++ b/lib/web/fetch/webidl.js
@@ -339,7 +339,7 @@ webidl.interfaceConverter = function (i) {
     if (opts.strict !== false && !(V instanceof i)) {
       throw webidl.errors.exception({
         header: i.name,
-        message: `Expected ${JSON.stringify(V)} to be an instance of ${i.name}.`
+        message: `Expected ${webidl.util.Stringify(V)} to be an instance of ${i.name}.`
       })
     }
 
--- a/lib/web/fetch/webidl.js
+++ b/lib/web/fetch/webidl.js
@@ -224,6 +224,8 @@ webidl.util.Stringify = function (V) {
       return `Symbol(${V.description})`
     case 'Object':
       return JSON.stringify(V)
+    case 'String':
+      return V.length ? V : '""'
     default:
       return `${V}`
   }
--- a/lib/web/fetch/webidl.js
+++ b/lib/web/fetch/webidl.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const { types } = require('node:util')
+const { types, inspect } = require('node:util')
 const { toUSVString } = require('../../core/util')
 
 /** @type {import('../../../types/webidl').Webidl} */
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by KhafraDev** [comment]
this was done for everything that stringifies a value, I just used BufferSource as an example

File: lib/web/fetch/webidl.js [hunk 1]
```diff
@@ -216,6 +216,21 @@ webidl.util.IntegerPart = function (n) {
...
+webidl.util.Stringify = function (V) {
+  const type = webidl.util.Type(V)
+
+  switch (type) {
+    case 'Symbol':
+      return `Symbol(${V.description})`
+    case 'Object':
+      return JSON.stringify(V)
+    case 'String':
+      return V.length ? V : '""'
```
Uzlopak, line 228: So if V has a Value it wont be having double quotes but if it is empty it has double quotes. Not that nice.

lib/web/fetch/webidl.js
```diff
@@ -216,6 +216,21 @@ webidl.util.IntegerPart = function (n) {
...
 }
 
+webidl.util.Stringify = function (V) {
+  const type = webidl.util.Type(V)
+
+  switch (type) {
+    case 'Symbol':
+      return `Symbol(${V.description})`
+    case 'Object':
+      return JSON.stringify(V)
```
Uzlopak, line 226: Actually... should be checking if it is a plain object and then use JSON.stringify? What if we have circulation? It would error.

test/webidl/converters.js
```diff
@@ -183,3 +183,22 @@ test('ByteString', () => {
...
+    [{ a: 'b' }, '{"a":"b"}'],
+    [Symbol('sym'), 'Symbol(sym)'],
+    [Symbol.iterator, 'Symbol(Symbol.iterator)'], // well-known symbol
+    [true, 'true'],
+    [0, '0'],
+    ['hello', 'hello'],
+    ['', '""'],
+    [null, 'null'],
+    [undefined, 'undefined']
+  ]
```
Uzlopak, line 199: And "undefined" will also be 'undefined'

lib/web/fetch/webidl.js
KhafraDev: [object Object]

lib/web/fetch/webidl.js
```diff
@@ -1,6 +1,6 @@
 'use strict'
 
-const { types } = require('node:util')
+const { types, inspect } = require('node:util')
```
Uzlopak, line 3: I dont want to be nagging, but inspect is imho slow. will this have perf impact?

KhafraDev: it's only used for error cases

Uzlopak (approved):
LGTM

PR #2845, opened 2024-02-25
websocket server only needs to reply with a single subprotocol

### LINKED ISSUE #2844: WebSocket: wrong subprotocol choose check (response Sec-WebSocket-Protocol header)
## Bug Description

1. client sends `Sec-WebSocket-Protocol` header with protocol **list** (separated by ',')
2. server responds with **one** protocol from this list
3. WebSocket client compares request header  (protocol **list**) with response header (**one** protocol) value and fails with the error `Protocol was not set in the opening handshake`

https://github.com/nodejs/undici/blob/c49058b956bda999bc26a1c43c66f1bede2ebc18/lib/web/websocket/connection.js#L179

## Reproducible By

```js
let ws = new WebSocket('ws://server...', ['msgpack', 'json'])
ws.addEventListener('error', e => console.log(e.error))
// server choses json and responds with `Sec-WebSocket-Protocol: json`
// client failed
```

## Expected Behavior

Correct subprotocol check

## Environment

- Node.js v21.6.2
- `--experimental-websocket` flag

KhafraDev:
Fixes #2844

### DIFF
--- a/lib/web/websocket/connection.js
+++ b/lib/web/websocket/connection.js
@@ -13,6 +13,7 @@ const { CloseEvent } = require('./events')
 const { makeRequest } = require('../fetch/request')
 const { fetching } = require('../fetch/index')
 const { Headers } = require('../fetch/headers')
+const { getDecodeSplit } = require('../fetch/util')
 const { kHeadersList } = require('../../core/symbols')
 
 /** @type {import('crypto')} */
@@ -176,9 +177,19 @@ function establishWebSocketConnection (url, protocols, ws, onEstablish, options)
       //    the WebSocket Connection_.
       const secProtocol = response.headersList.get('Sec-WebSocket-Protocol')
 
-      if (secProtocol !== null && secProtocol !== request.headersList.get('Sec-WebSocket-Protocol')) {
-        failWebsocketConnection(ws, 'Protocol was not set in the opening handshake.')
-        return
+      if (secProtocol !== null) {
+        const requestProtocols = getDecodeSplit('sec-websocket-protocol', request.headersList)
+        const responseProtocols = getDecodeSplit('sec-websocket-protocol', response.headersList)
+
+        // The client can request that the server use a specific subprotocol by
+        // including the |Sec-WebSocket-Protocol| field in its handshake.  If it
+        // is specified, the server needs to include the same field and one of
+        // the selected subprotocol values in its response for the connection to
+        // be established.
+        if (!requestProtocols.some((subProtocol) => responseProtocols.includes(subProtocol))) {
+          failWebsocketConnection(ws, 'Protocol was not set in the opening handshake.')
+          return
+        }
       }
 
       response.socket.on('data', onSocketData)
--- a/lib/web/websocket/connection.js
+++ b/lib/web/websocket/connection.js
@@ -179,14 +179,13 @@ function establishWebSocketConnection (url, protocols, ws, onEstablish, options)
 
       if (secProtocol !== null) {
         const requestProtocols = getDecodeSplit('sec-websocket-protocol', request.headersList)
-        const responseProtocols = getDecodeSplit('sec-websocket-protocol', response.headersList)
 
         // The client can request that the server use a specific subprotocol by
         // including the |Sec-WebSocket-Protocol| field in its handshake.  If it
         // is specified, the server needs to include the same field and one of
         // the selected subprotocol values in its response for the connection to
         // be established.
-        if (!requestProtocols.some((subProtocol) => responseProtocols.includes(subProtocol))) {
+        if (!requestProtocols.includes(secProtocol)) {
           failWebsocketConnection(ws, 'Protocol was not set in the opening handshake.')
           return
         }
--- a/lib/web/websocket/connection.js
+++ b/lib/web/websocket/connection.js
@@ -189,6 +189,11 @@ function establishWebSocketConnection (url, protocols, ws, onEstablish, options)
           failWebsocketConnection(ws, 'Protocol was not set in the opening handshake.')
           return
         }
+      } else if (request.headersList.contains('sec-websocket-protocol', true)) {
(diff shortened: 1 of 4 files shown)


KhafraDev:
@lpinca I tried testing this with ws but it seemed to work, is this expected? The server isn't responding with one of the subprotocols the client requested:

<details>

<summary>example</summary>

```mjs
import { once } from 'events'
import { WebSocketServer, WebSocket } from 'ws'

const wss = new WebSocketServer({
  handleProtocols: () => 'idk',
  port: 0
})

wss.on('connection', function connection (ws) {
  ws.on('error', console.error)

  ws.on('message', function message (data) {
    console.log('received: %s', data)
  })

  ws.send('something')
})

await once(wss, 'listening')

const ws = new WebSocket(`ws://localhost:${wss.address().port}`, {
  protocols: ['msgpack', 'json']
})

ws.onerror = (e) => console.log('error', e.error)
ws.onmessage = (e) => console.log('message:', e.data)
```
</details>

Thanks!

test/websocket/issue-2844.js
```diff
@@ -0,0 +1,41 @@
...
+
+  wss.on('connection', (ws) => {
+    ws.on('error', fail)
+    ws.send('something')
+  })
+
+  await once(wss, 'listening')
+
+  const ws = new WebSocket(`ws://localhost:${wss.address().port}`, {
+    protocols: ['msgpack', 'json']
```
farwayer, line 29: The second constructor parameter must be subprotocol string or array of subprotocol strings, not object

lib/web/websocket/connection.js
```diff
@@ -176,9 +177,19 @@ function establishWebSocketConnection (url, protocols, ws, onEstablish, options)
       //    the WebSocket Connection_.
       const secProtocol = response.headersList.get('Sec-WebSocket-Protocol')
 
-      if (secProtocol !== null && secProtocol !== request.headersList.get('Sec-WebSocket-Protocol')) {
-        failWebsocketConnection(ws, 'Protocol was not set in the opening handshake.')
-        return
+      if (secProtocol !== null) {
+        const requestProtocols = getDecodeSplit('sec-websocket-protocol', request.headersList)
+        const responseProtocols = getDecodeSplit('sec-websocket-protocol', response.headersList)
```
farwayer, line 182: server must respond with a single value or with null (equivalent to no header), not with a list

farwayer (commented):
thank you for such a quick response! there are a couple of comments

test/websocket/issue-2844.js
KhafraDev: we allow objects too https://github.com/nodejs/undici/blob/main/docs/docs/api/WebSocket.md#new-websocketurl-protocol

lib/web/websocket/connection.js
```diff
@@ -176,9 +177,18 @@ function establishWebSocketConnection (url, protocols, ws, onEstablish, options)
       //    the WebSocket Connection_.
       const secProtocol = response.headersList.get('Sec-WebSocket-Protocol')
 
-      if (secProtocol !== null && secProtocol !== request.headersList.get('Sec-WebSocket-Protocol')) {
-        failWebsocketConnection(ws, 'Protocol was not set in the opening handshake.')
-        return
+      if (secProtocol !== null) {
```
farwayer, line 180: if we sent at least one subprotocol, but the server responded with a null header or did not send a header, then according to the specification we should fail too

KhafraDev: the spec also says

```
6.  If the response includes a |Sec-WebSocket-Protocol| header field
       and this header field indicates the use of a subprotocol that was
       not present in the client's handshake (the server has indicated a
       subprotocol not requested by the client), the client MUST _Fail
       the WebSocket Connection_.
```

the "if the response includes" part seems to contradict what it later says?

KhafraDev: I see, protocols are optional so the server doesn't need to send (if none are sent from the client). Definitely could be worded better...

mcollina (approved):
lgtm

lib/web/websocket/connection.js
farwayer: Yes, protocols are optional. But, if the client specifies them, then this becomes a mandatory requirement for establishing a connection. And if the server does not support any of the client protocols, then it responds with a null header or does not specify one to let the client know that it is not ready for further interaction

KhafraDev: I completely misunderstood what the rfc was stating.

lib/web/websocket/connection.js
```diff
@@ -189,11 +189,6 @@ function establishWebSocketConnection (url, protocols, ws, onEstablish, options)
           failWebsocketConnection(ws, 'Protocol was not set in the opening handshake.')
           return
         }
-      } else if (request.headersList.contains('sec-websocket-protocol', true)) {
```
farwayer: I think this check should be left though. 

In spec:
> In this version of the protocol,
>    the main option field is |Sec-WebSocket-Protocol|, which indicates
>    the subprotocol that the server has selected.  WebSocket clients
>    verify that the server included one of the values that was specified
>    in the WebSocket client's handshake.

Logically, this is an exchange of information about capabilities between the client and server. And if the server doesn't support what the client wants, then the connection should be impossible.

I checked how it works in the browser. And it's true
```js
ws = new WebSocket("wss://echo.websocket.org", ['rpc.msgpack', 'rpc.json'])
```

KhafraDev: https://github.com/nodejs/undici/blob/c49058b956bda999bc26a1c43c66f1bede2ebc18/lib/web/websocket/connection.js#L111-L118

farwayer: Sorry, I didn't notice that this check happens earlier

farwayer (approved):

Uzlopak (approved):
LGTM

PR #2851, opened 2024-02-26
fix: deflake connect-timeout test

Uzlopak:
it will probably not deflake connect-timeout but improves it in that way, that it avoids the error, that the test maybe results in some unhandled promise rejection warning

https://github.com/nodejs/undici/actions/runs/8017722218/job/21902182355?pr=2816
image


mcollina (approved):
lgtm

Issue #2854, opened 2024-02-26
[Bug] When body response is empty, body.json() returns error


ValchanOficial:
## Bug Description

I call some API, it returns or not a body
I always call body.json after a request
If body is empty, it throws: Unexpected end of JSON input at JSON
**Line 306: resolve(JSON.parse(chunksDecode(body, length)))**

## Expected Behavior

If body is empty, just return empty too

ronag:
Sorry. This works as expected.

ValchanOficial:
Try: https://developer.mozilla.org/pt-BR/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse
console.log(JSON.parse(''));

![307901357-f27add81-6c2c-4a1a-88f8-0bf691a0a29a](https://github.com/nodejs/undici/assets/16228014/594b338e-77c3-4e4c-8c63-f3b876dbefa6)

ValchanOficial:
https://github.com/ValchanOficial/example

Uzlopak:
I investigated this and this is definetly working as expected.

Chrome is also throwing an error when requesting a resource with no body and trying to call .json().

PR #2855, opened 2024-02-26
Update chunksDecode

ValchanOficial:
From this bug: [2854](https://github.com/nodejs/undici/issues/2854)

### DIFF
--- a/lib/api/readable.js
+++ b/lib/api/readable.js
@@ -281,7 +281,7 @@ function consumeStart (consume) {
  */
 function chunksDecode (chunks, length) {
   if (chunks.length === 0 || length === 0) {
-    return ''
+    return JSON.stringify('')
   }
   const buffer = chunks.length === 1 ? chunks[0] : Buffer.concat(chunks, length)
 


ronag:
This looks wrong to me. But maybe someone else disagrees.

Uzlopak:
This does not make any sense. Also why would we need to call JSON.stringify all the time as `JSON.stringify('')` is equal to `'""'`.

So why do we need here two double quotes?

Uzlopak (changes requested):
Makes no sense

ValchanOficial:
image

ValchanOficial:
If I add return JSON.stringify('')
image

ronag:
Who? The server? That's out of our control...

ValchanOficial:
If the person calls an external API they do not have control of the body.
Why not make the lives of those who use the lib easier?

Uzlopak:
The problem is that you are not describing the error properly. 

Your bug report is actually:

An endpoint returns an empty string as payload. Fetch against this endpoint is successful. Calling body.text() returns an empty string as expected. Calling body.json() returns in an error, because an empty string can not be parsed successfully.

The question is, is throwing an error correct? 
What does e.g. fetch return in chrome? Does it throw? Does it return null or undefined?
Do we have to encapsulate the JSON parse error and throw a fetch specific error?

Uzlopak:
I investigated this and this is definetly working as expected.

ValchanOficial:
> Do we have to encapsulate the JSON parse error and throw a fetch specific error?

Maybe just return body without parse it

Uzlopak:
No.

ronag:
We follow the body mixin spec on this. Which is what our users expect.

https://developer.mozilla.org/en-US/docs/Web/API/Request/json

PR #2863, opened 2024-02-27
perf: reuse TextDecoder instance

Uzlopak:
Was there a reason not to reuse the textdecoder instance?

### DIFF
--- a/lib/web/websocket/receiver.js
+++ b/lib/web/websocket/receiver.js
@@ -12,6 +12,8 @@ const { WebsocketFrameSend } = require('./frame')
 // Copyright (c) 2013 Arnout Kazemier and contributors
 // Copyright (c) 2016 Luigi Pinca and contributors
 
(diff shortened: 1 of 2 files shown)


KhafraDev (approved):

Ethan-Arrowood (approved):

Uzlopak:
@KhafraDev 

I wished I would be admin on this repo to directly check what is wrong with it...

kibertoad:
@ronag Is this a safe change? I remember you pointing out previously that TextDecoder is stateful

ronag:
> @ronag Is this a safe change? I remember you pointing out previously that TextDecoder is stateful

As long as you don't use `stream: true`, it's safe.

Uzlopak:
Also TextDecoderStream is unsafe.;)

panva:
FYI: https://github.com/nodejs/node/pull/52175#issuecomment-2020665559

> "fatal" option is not supported on Node.js compiled without ICU

cc @nodejs/undici

PR #2864, opened 2024-02-27
chore: restructure benchmarks, use kebab-case

Uzlopak:
this is just about restructuring the benchmarks folder, also use kebab-case for the file names


Ethan-Arrowood (approved):

Uzlopak:
lets see if the codecoverage is now fixed

PR #2865, opened 2024-02-27
websocket: improve .close()

Uzlopak:
Imho the close method was not right. I realized that after using isClosed and isClosing helper functions. thats why the last else black was dead code.

### DIFF
--- a/lib/web/websocket/websocket.js
+++ b/lib/web/websocket/websocket.js
@@ -3,7 +3,7 @@
 const { webidl } = require('../fetch/webidl')
 const { URLSerializer } = require('../fetch/dataURL')
 const { getGlobalOrigin } = require('../fetch/global')
-const { staticPropertyDescriptors, states, opcodes, emptyBuffer } = require('./constants')
+const { staticPropertyDescriptors, states, sentCloseFrameState, opcodes, emptyBuffer } = require('./constants')
 const {
   kWebSocketURL,
   kReadyState,
@@ -140,6 +140,8 @@ class WebSocket extends EventTarget {
     // be CONNECTING (0).
     this[kReadyState] = WebSocket.CONNECTING
 
+    this[kSentClose] = sentCloseFrameState.NOT_SENT
+
     // The extensions attribute must initially return the empty string.
 
     // The protocol attribute must initially return the empty string.
@@ -201,7 +203,7 @@ class WebSocket extends EventTarget {
       // to CLOSING (2).
       failWebsocketConnection(this, 'Connection was closed before it was established.')
       this[kReadyState] = WebSocket.CLOSING
-    } else if (!isClosing(this)) {
+    } else if (this[kSentClose] === sentCloseFrameState.NOT_SENT) {
       // If the WebSocket closing handshake has not yet been started
       // Start the WebSocket closing handshake and set this's ready
       // state to CLOSING (2).
@@ -211,6 +213,8 @@ class WebSocket extends EventTarget {
       //   WebSocket Close message must be the integer given by code.
       // - If reason is also present, then reasonBytes must be
       //   provided in the Close message after the status code.
+
+      this[kSentClose] = sentCloseFrameState.PROCESSING
 
       const frame = new WebsocketFrameSend()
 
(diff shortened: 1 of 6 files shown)


KhafraDev (changes requested):
```js
  // Upon either sending or receiving a Close control frame, it is said
  // that _The WebSocket Closing Handshake is Started_ and that the
  // WebSocket connection is in the CLOSING state.
```

(explained better below)

KhafraDev:
kSentClose is sent after we send the packet, but the closing handshake starts once we sent the handshake, if that makes sense.

KhafraDev:
@Uzlopak once the one thing is reverted this can land

Uzlopak:
Its not that i want to pressure my change. I am still investigating how we can be on spec. Because now it is not possible to achieve 100% test coverage because the last else case can never be hit. 

I will revert the mentioned change, so that we can merge the undisputed improvement and open an issue, were we can discuss the disputed problem ;).

lib/web/websocket/websocket.js
```diff
@@ -184,10 +192,10 @@ class WebSocket extends EventTarget {
     }
 
     // 3. Run the first matching steps from the following list:
-    if (this[kReadyState] === WebSocket.CLOSING || this[kReadyState] === WebSocket.CLOSED) {
+    if (isClosing(this) || isClosed(this)) {
       // If this's ready state is CLOSING (2) or CLOSED (3)
       // Do nothing.
-    } else if (!isEstablished(this)) {
+    } else if (isConnecting(this)) {
```
KhafraDev, line 198: "      // If the WebSocket connection is not yet established"

KhafraDev (changes requested):

lib/web/websocket/websocket.js
Uzlopak: I am currently investigating this. Thats why is set it to draft.

We have 4 states. Connecting, Open, Closing, Closed. !isEstablished equals Connecting, Closing and Closed. Just because we checked before the cases for closing and closed we can use !isEstablished, as it reduces !isEstablished to Connecting. 

Tbh. feels super strange to not check explicitly for the right state.

Uzlopak:
@KhafraDev 

What do you think now with sentCloseFrameState?

lib/web/websocket/websocket.js
KhafraDev: but isConnecting is

// If the WebSocket connection is not yet established, and the connection
  // is not yet closed, then the WebSocket connection is in the CONNECTING state.

KhafraDev: I prefer the name of isEstablished because it's what the spec uses. If isEstablished = isConnecting that's fine

Uzlopak: In this case

KhafraDev (approved):

KhafraDev:
> What do you think now with sentCloseFrameState?

It seems correct

Uzlopak:
@KhafraDev 

please review again. i added two tests to increase the coverage ;)

test/websocket/close.js
```diff
@@ -128,4 +130,53 @@ describe('Close', () => {
...
+    const ws = new WebSocket(`ws://localhost:${server.address().port}`)
+    ws.addEventListener('open', () => {
+      ws.close(1000)
+      ws.close(1000)
+    })
+
+    await t.completed
+  })
+
+  test('calling close when connection is open but close handshake has started does not result in sending the handshake again', async (t) => {
```
KhafraDev, line 156: this test doesn't serve a purpose

Uzlopak: why not? This test covers the last else case

https://app.codecov.io/gh/nodejs/undici/pull/2865/blob/lib/web/websocket/websocket.js#L253

KhafraDev: because the test isn't 'natural', you can't mutate internal state normally. If a test could cover that with what websocket publicly exposes, it would be useful

Uzlopak: There is no way to simulate this with the used websocket server of the npm ws server.

Uzlopak:
@KhafraDev 
PTAL

KhafraDev (approved):

mcollina (approved):
lgtm

PR #2871, opened 2024-02-28
chore: remove no-simd wasm

Uzlopak:
This PR ensures, that we also test the no simd wasm file.


mcollina:
I think the no-simd version can be removed in v6.

Uzlopak:
@mcollina 
changed PR

mcollina (approved):
lgtm

PR #2882, opened 2024-02-28
perf: dump immediatly if known size exceeds limit

ronag:
## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready


mcollina (changes requested):

lib/api/readable.js
```diff
@@ -160,6 +163,10 @@ module.exports = class BodyReadable extends Readable {
     }
 
     return await new Promise((resolve, reject) => {
+      if (this[kContentLength] > limit) {
+        this.destroy(new AbortError())
```
Uzlopak, line 167: Can we give theAbortError a message like content length exceeded?

metcoder95 (approved):
LGTM with @mcollina comments

lib/api/readable.js
Uzlopak: actually... dont you need to reject?

something like


  ```js
  return await new Promise((resolve, reject) => {
      if (this[kContentLength]) {
        const contentLength = +this[kContentLength]
        if (
          Number.isFinite(contentLength) &&
          contentLength > limit
        ) {
          const abortError = new AbortError('content length exceeded limit')
          this.destroy(abortError)
          reject(abortError)
          return
        }
      }

```

ronag: Not strictly necessary.

metcoder95 (approved):

PR #2888, opened 2024-02-29
cookies: fix validateMaxAge allow negative numbers

Uzlopak:
Imho the spec rfc 6265 is not correctly implemented. 

Either only positive integers excluding 0 are allowed (which the current implementation does not match) for MaxAge or all integer values are allowed (this PR)

I think the problem arises because imho https://www.rfc-editor.org/rfc/rfc6265#section-4.1.1 talks about how the data in the useragent is stored while https://www.rfc-editor.org/rfc/rfc6265#section-5.2.2 talks about the data which the useragent receives and how it should be processed.

So a server can send negative values for max-age, while a client has to drop all cookies with negative values immediatly. If the maxage is set to 0, then it would actually mean to drop the cookie according to the rfc, but it seems most useragents keep the cookie till the end of the session.

Anyhow, this means that we must allow negative integers for validateMaxAge.

### DIFF
--- a/benchmarks/cookies/validate-max-age.mjs
+++ b/benchmarks/cookies/validate-max-age.mjs
@@ -1,0 +1,12 @@
+import { bench, group, run } from 'mitata'
+import { validateCookieMaxAge } from '../../lib/web/cookies/util.js'
+
+const valid = 2000
+
+group('validateCookieMaxAge', () => {
+  bench(`valid: ${valid}`, () => {
+    return validateCookieMaxAge(valid)
+  })
+})
+
+await run()
--- a/lib/web/cookies/util.js
+++ b/lib/web/cookies/util.js
@@ -197,10 +197,21 @@ function toIMFDate (date) {
                        ; In practice, both expires-av and max-age-av
                        ; are limited to dates representable by the
                        ; user agent.
+ *
+ * @see https://www.rfc-editor.org/rfc/rfc6265#section-4.1.1
+ *
+ * but:
+ * If delta-seconds is less than or equal to zero (0), let expiry-time be the
+ * earliest representable date and time. Otherwise, let the expiry-time be the
+ * current date and time plus delta-seconds seconds.
+ * @see https://www.rfc-editor.org/rfc/rfc6265#section-5.2.2
+ *
+ * Any integer value is valid.
+ *
  * @param {number} maxAge
  */
 function validateCookieMaxAge (maxAge) {
-  if (maxAge < 0) {
+  if (Number.isInteger(maxAge) === false) {
     throw new Error('Invalid cookie max-age')
   }
 }
@@ -301,6 +312,7 @@ module.exports = {
   validateCookieName,
   validateCookiePath,
   validateCookieValue,
+  validateCookieMaxAge,
   toIMFDate,
   stringify,
   getHeadersList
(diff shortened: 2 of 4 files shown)


mcollina (approved):
lgtm

KhafraDev (changes requested):
the test comes from Deno

https://github.com/denoland/deno_std/blob/45f7a6b46018e9f4d330943918299c68ae80f1bb/http/cookie_test.ts#L238-L251

Uzlopak:
@KhafraDev 

Just because the test comes from deno, doesnt mean that it makes sense or is even correct. I argued with the spec and imho the spec clearly states that sending a negative value means removes the cookie from the user agent. 

Do I have to first upstream the change on deno, before i can propose the change on undici?

KhafraDev:
The logic is sound on deno's end. A cookie cannot expire on the server, therefore a negative value is invalid.

> ; In practice, both expires-av and max-age-av
                       ; are limited to dates representable by the
                       ; user agent.

Uzlopak:
First of all, the logic on deno is not following the spec if we follow your argument. according to the spec 0 is not allowed but deno allows it. second of all, not allowed negative numbers makes it impossible to delete immediatly cookies in the browser. 

Following case:

1. browser with no cookies
2. login into site, site sends cookie via Set-Cookie header, saying e.g. max-age=864000
3. you send now on every call to the page the cookie 
4. Now you log out. Server deletes your session and wants to tell the browser that the person is logged out. It sends the Set-Cookie header with ...
4a. max-age 1
5a. You make the next request in less than a second and again the cookie, as it is still not deleted as it needs to stay for a second, so you get a potential 401 error.
4b. max-age 0
5b. You make any request and again the cookie, as it is still not deleted as 0 means keep cookie till end of the session, so you get still 401 error. You need to close and open again your browser to drop the cookie
4c- max-age -1
5c. You make any requests and dont send any cookies, because any negative number means to drop the cookie.

This behaviour is described in the spec:
https://www.rfc-editor.org/rfc/rfc6265#section-5.2.2

Uzlopak:
This does not convince me. :(

metcoder95 (commented):
Giving my two cents here, I do actually see the point @Uzlopak is making, given the arguments of the spec, but the point from @KhafraDev is pretty valid.

I do believe that the use case that handling `-1` gives over `max-age` on the `set-cookie` header does not work in an environment that does not do automatic cookie management like Node, everything goes upon the implementation to decide how to manage it, it is of no use case for the Node at all (given how the `set-cookie` is designed for).

On the other hand, surfacing this implementation by allowing this might not be of value for the implementer as `cookies` are often of no importance in service-side communications.

My take is to keep it is as is, but won't block the PR from being merged

KhafraDev:
I don't have a strong opinion on this, but at the minimum an issue should be reported in the denostd repo.

Uzlopak:
@KhafraDev 
https://github.com/denoland/deno_std/issues/4426

Uzlopak:
@KhafraDev 

Seems like deno wants to restrict the maxAge to be a positive number

Uzlopak:
@KhafraDev 

I checked again deno and they purposely changed it because of the same reason I am argueing to allow negative numbers.

Uzlopak:
@KhafraDev 
According to deno team they want it to allow only positive integers... 
https://github.com/denoland/deno_std/pull/4459

Will adapt this PR if deno merges my corresponding PR ;)

PR #2889, opened 2024-02-29
perf: optimize check invalid field-vchar

tsctx:
```
benchmark              time (avg)             (min … max)       p75       p99      p999
--------------------------------------------------------- -----------------------------
• isValidHeaderChar# text/html
--------------------------------------------------------- -----------------------------
regexp.test            34 ns/iter     (28.27 ns … 303 ns)  32.71 ns  72.85 ns    133 ns
regexp.exec         40.45 ns/iter     (34.18 ns … 218 ns)  41.16 ns  83.69 ns    135 ns
charCodeAt          30.31 ns/iter     (26.07 ns … 217 ns)   29.1 ns  61.91 ns    123 ns
isValidHeaderChar   35.55 ns/iter      (29.3 ns … 236 ns)  35.45 ns  86.28 ns    126 ns

summary for isValidHeaderChar# text/html
  charCodeAt
   1.12x faster than regexp.test
   1.17x faster than isValidHeaderChar
   1.33x faster than regexp.exec

• isValidHeaderChar application/json; charset=UTF-8
--------------------------------------------------------- -----------------------------
regexp.test         65.64 ns/iter     (56.88 ns … 426 ns)  63.33 ns    152 ns    237 ns
regexp.exec         73.79 ns/iter     (63.48 ns … 283 ns)  72.61 ns    153 ns    208 ns
charCodeAt            108 ns/iter     (95.21 ns … 416 ns)    102 ns    238 ns    353 ns
isValidHeaderChar   66.89 ns/iter     (57.03 ns … 254 ns)  63.48 ns    156 ns    207 ns

summary for isValidHeaderChar application/json; charset=UTF-8
  regexp.test
   1.02x faster than isValidHeaderChar
   1.12x faster than regexp.exec
   1.64x faster than charCodeAt
```

### DIFF
--- a/benchmarks/core/is-valid-header-char.mjs
+++ b/benchmarks/core/is-valid-header-char.mjs
@@ -39,7 +39,7 @@ group(`isValidHeaderChar# ${html}`, () => {
   })
 })
 
-group(`isValidHeaderChar ${json}`, () => {
+group(`isValidHeaderChar# ${json}`, () => {
   bench('regexp.test', () => {
     return !headerCharRegex.test(json)
   })
--- a/benchmarks/core/is-valid-header-char.mjs
+++ b/benchmarks/core/is-valid-header-char.mjs
@@ -1,0 +1,57 @@
+import { bench, group, run } from 'mitata'
+import { isValidHeaderChar } from '../../lib/core/util.js'
+
+const html = 'text/html'
+const json = 'application/json; charset=UTF-8'
+
+const headerCharRegex = /[^\t\x20-\x7e\x80-\xff]/
+
+/**
+ * @param {string} characters
+ */
+function charCodeAtApproach (characters) {
+  // Validate if characters is a valid field-vchar.
+  //  field-value    = *( field-content / obs-fold )
+  //  field-content  = field-vchar [ 1*( SP / HTAB ) field-vchar ]
+  //  field-vchar    = VCHAR / obs-text
+  for (let i = 0; i < characters.length; ++i) {
+    const code = characters.charCodeAt(i)
+    // not \x20-\x7e, \t and \x80-\xff
+    if ((code < 0x20 && code !== 0x09) || code === 0x7f || code > 0xff) {
+      return false
+    }
+  }
+  return true
+}
+
+group(`isValidHeaderChar# ${html}`, () => {
+  bench('regexp.test', () => {
+    return !headerCharRegex.test(html)
+  })
+  bench('regexp.exec', () => {
+    return headerCharRegex.exec(html) === null
+  })
+  bench('charCodeAt', () => {
+    return charCodeAtApproach(html) === true
+  })
+  bench('isValidHeaderChar', () => {
+    return isValidHeaderChar(html) === true
+  })
+})
+
+group(`isValidHeaderChar ${json}`, () => {
+  bench('regexp.test', () => {
+    return !headerCharRegex.test(json)
+  })
+  bench('regexp.exec', () => {
+    return headerCharRegex.exec(json) === null
+  })
+  bench('charCodeAt', () => {
+    return charCodeAtApproach(json)
+  })
+  bench('isValidHeaderChar', () => {
+    return isValidHeaderChar(json)
+  })
+})
+
+await run()
--- a/benchmarks/core/is-valid-header-char.mjs
+++ b/benchmarks/core/is-valid-header-char.mjs
@@ -32,10 +32,10 @@ group(`isValidHeaderChar# ${html}`, () => {
     return headerCharRegex.exec(html) === null
   })
   bench('charCodeAt', () => {
-    return charCodeAtApproach(html) === true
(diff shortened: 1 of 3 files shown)


lib/core/request.js
```diff
@@ -335,7 +323,7 @@ function processHeader (request, key, val, skipAppend = false) {
     const arr = []
     for (let i = 0; i < val.length; i++) {
       if (typeof val[i] === 'string') {
-        if (headerCharRegex.exec(val[i]) !== null) {
+        if (!util.isValidHeaderChar(val[i])) {
```
Uzlopak, line 326: can we have named imports instead of default import?

KhafraDev: it'll be slower on longer headers and faster on shorter, a benchmark for every utility function isn't needed

KhafraDev (approved):

lib/core/request.js
KhafraDev: although it would be nice to see if this actually does improve performance, and if it does, maybe upstream it to core

Uzlopak: a benchmark would help to verify the proposal. e.g. using exec means instantiating an object to return. So why was the regex not using .test(), which always returns true and false?

tsctx: > So why was the regex not using .test(), which always returns true and false?

I think this is because exec is faster when primordials are used.

benchmarks/core/is-valid-header-char.mjs
```diff
@@ -0,0 +1,57 @@
...
+  })
+  bench('charCodeAt', () => {
+    return charCodeAtApproach(html)
+  })
+  bench('isValidHeaderChar', () => {
+    return isValidHeaderChar(html)
+  })
+})
+
+group(`isValidHeaderChar ${json}`, () => {
```
tsctx, line 42:
```suggestion
group(`isValidHeaderChar# ${json}`, () => {
```

lib/core/util.js
```diff
@@ -495,6 +495,24 @@ function isValidHTTPToken (characters) {
...
+ *  field-vchar    = VCHAR / obs-text
+ */
+const headerCharRegex = /[^\t\x20-\x7e\x80-\xff]/
+
+/**
+ * @param {string} characters
+ */
+function isValidHeaderChar (characters) {
+  return !headerCharRegex.test(characters)
+}
```
Uzlopak, line 514: interesting.

i wonder how the performance is if we flip it and do it like this

const isInvalidHeaderChar = headerCharRegex.test.bind(headerCharRegex)

Uzlopak (commented):
image

lib/core/util.js
tsctx: There is no change.
```

benchmark                     time (avg)             (min … max)       p75       p99      p999
---------------------------------------------------------------- -----------------------------
• isValidHeaderChar# text/html
---------------------------------------------------------------- -----------------------------
regexp.test                33.69 ns/iter     (28.27 ns … 297 ns)  31.45 ns  79.35 ns    143 ns
regexp.exec                35.91 ns/iter     (31.64 ns … 273 ns)  34.62 ns  80.37 ns    153 ns
charCodeAt                 27.58 ns/iter        (25 ns … 532 ns)  25.88 ns  62.16 ns    133 ns
isValidHeaderChar          32.33 ns/iter      (29.1 ns … 452 ns)  30.62 ns  78.03 ns    163 ns
isInvalidHeaderCharRegex   30.87 ns/iter     (28.22 ns … 195 ns)  30.03 ns  66.36 ns    133 ns

summary for isValidHeaderChar# text/html
  charCodeAt
   1.12x faster than isInvalidHeaderCharRegex
   1.17x faster than isValidHeaderChar
   1.22x faster than regexp.test
   1.3x faster than regexp.exec

• isValidHeaderChar# application/json; charset=UTF-8
---------------------------------------------------------------- -----------------------------
regexp.test                59.13 ns/iter      (54.2 ns … 443 ns)   56.4 ns    120 ns    229 ns
regexp.exec                64.15 ns/iter     (58.35 ns … 268 ns)  60.64 ns    134 ns    212 ns
charCodeAt                 97.63 ns/iter     (89.45 ns … 355 ns)  98.97 ns    180 ns    303 ns
isValidHeaderChar           74.7 ns/iter     (57.71 ns … 357 ns)     75 ns    170 ns    277 ns
isInvalidHeaderCharRegex   59.08 ns/iter     (54.25 ns … 280 ns)  56.25 ns    147 ns    245 ns

summary for isValidHeaderChar# application/json; charset=UTF-8
  isInvalidHeaderCharRegex
   1x faster than regexp.test
   1.09x faster than regexp.exec
   1.26x faster than isValidHeaderChar
   1.65x faster than charCodeAt
```

Uzlopak: thank you for the investigation

Uzlopak (approved):
LGTM

PR #2892, opened 2024-02-29
doc deprecate bodymixin.formData

KhafraDev:
## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -60,6 +60,8 @@ console.log('trailers', trailers)
 
 The `body` mixins are the most common way to format the request/response body. Mixins include:
 
+- [`.arrayBuffer()`](https://fetch.spec.whatwg.org/#dom-body-arraybuffer)
+- [`.blob()`](https://fetch.spec.whatwg.org/#dom-body-blob)
 - [`.formData()`](https://fetch.spec.whatwg.org/#dom-body-formdata)
 - [`.json()`](https://fetch.spec.whatwg.org/#dom-body-json)
 - [`.text()`](https://fetch.spec.whatwg.org/#dom-body-text)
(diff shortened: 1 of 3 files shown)


KhafraDev:
cc @Ethan-Arrowood 

Since @types/node relies on undici-types, people will notice that .formData() is "deprecated" if they use typescript. I feel very strongly that we should prevent people from using it, even if we did everything correctly, because it's an API that is totally unsuitable for any use case I can think of.

mcollina (approved):
lgtm

(unfortunately)

ronag (approved):

polaroidkidd:
hi @KhafraDev 

I'm one of those devs and I've tried to find a good explanation of _why_ it's deprecated but came up empty handed. I admit the form-data API isn't exactly the friendliest, but it feels like overkill to mark it as deprecated. I'm no node-js expert, so I'm assuming there's some good arguments here, I just can't seem to find any. 

All examples I could find revolve around JSON compatible APIs, but I'm sending a registration form (name, email & avatar). 

  * I don't want to convert the image to base64
  * I can't mix in binary with standard JSON in a single request
  * I don't want to separate the data endpoint and image endpoint

Using a formPost seemed like the obvious choice for the frontend. The backend can "simply" do the `const formData = await request.formData()` call. No extra deps, fairly easy to use and it's baked into most environments.

Do you happen to have a resource handy? I'd really like to understand the reasoning behind this before I go ahead and add a library to a web worker.

KhafraDev:
It wasn't deprecated for semantics, but because of how it's implemented. It's implemented the same in every other environment FWIW: the body is buffered in memory and then the "form data" body is parsed. The biggest issue therefore, and the reason for the deprecation, is that multipart/form-data is oftentimes used to send large files, which leads to potential DOSing on the backend.

Semantically it sucks. It's called "formData" but parses both multipart/form-data and x-www-form-urlencoded bodies. You cannot limit the body size nor do you have any control over how the body is parsed (busboy and @fastify/busboy both give fine grained control). The spec does not include steps to parse form data bodies, which means most implementations have quirks that do/don't work in other environments.

My recommendation is to use either busboy or @fastify/busboy (0 deps) and pipe a Response's body to it.

```js

const busboy = ...
const response = new Response(...)

if (response.body !== null) {
    Readable.fromWeb(response.body).pipe(busboy)
}
```

There are other libraries such as https://github.com/mjackson/remix-the-web/tree/main/packages/multipart-parser but I have never used it.

polaroidkidd:
@KhafraDev thank you. You have been my "TIL" moment! Going to switch to `@fastify/busboy` per recommendation.

ddtt786:
Considering that mitigating DoS from large payloads is a server-side responsibility and the API is meant for simple convenience rather than fine-grained control, deprecating .formData() doesn't seem to be the right course of action.

Issue #2898, opened 2024-03-01
Fetch fails for responses with http status 421 when the request has a body


bdemann:
## Bug Description

If the client sends a request with a body of length greater than 0 and the server responds with a 421 status then fetch will fail.

## Reproducible By

With an express server (for example) with an endpoint like this
```javascript
app.post('/421', (req, res) => {
  res.status(421).send('This is a 421 Misdirected Request response.');
});
```
Being hit with fetch like this
```javascript
async function fetch421Endpoint() {
  try {
    const response = await fetch('http://localhost:3000/421', {
      method: 'POST',
      body: new Uint8Array(Buffer.from('helloworld', 'utf-8'))
    });
    if (response.status === 421) {
      const data = await response.text();
      console.log('421 endpoint response:', data);
    } else {
      console.log('421 endpoint didn\'t return with 421 status');
      console.error('Error:', response.status, response.statusText);
    }
  } catch (error) {
    console.log('421 endpoint failed');
    console.error('Error:', error);
  }
}
```

## Expected Behavior

In the above example I would expect it to print out "421 endpoint response: This is a 421 Misdirected Request response." But instead it throws and it lands in the catch block. cURL, FireFox's and Chrome's implementation of fetch both have the expected behavior.

Because it only happens with request bodies of length 1 or greater, GET requests always work as expected

## Logs & Screenshots

```
Error: TypeError: fetch failed
    at node:internal/deps/undici/undici:12344:11
    at async fetch421Endpoint (fourZeroSeven/fetch.js:38:22) {
  cause: RequestContentLengthMismatchError: Request body length does not match content-length header
      at AsyncWriter.end (node:internal/deps/undici/undici:9742:19)
      at writeIterable (node:internal/deps/undici/undici:9646:16)
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
    code: 'UND_ERR_REQ_CONTENT_LENGTH_MISMATCH'
  }
}
```

## Environment

I have tried on the following node version: v18.18.0 and v21.6.1
Ubuntu 23.10

### Additional context

Additional information and a full server and client reproduction can be found here https://github.com/demergent-labs/fourZeroSeven

PR #2900, opened 2024-03-02
fix issue 2898

KhafraDev:
This took me an hour to solve. Fuck me.

Fixes https://github.com/nodejs/undici/issues/2898

### DIFF
--- a/lib/web/fetch/body.js
+++ b/lib/web/fetch/body.js
@@ -21,6 +21,7 @@ const { isArrayBuffer } = require('node:util/types')
 const { File: UndiciFile } = require('./file')
 const { serializeAMimeType } = require('./data-url')
 const { Readable } = require('node:stream')
+const { platform } = require('node:os')
 
 /** @type {globalThis['File']} */
 const File = NativeFile ?? UndiciFile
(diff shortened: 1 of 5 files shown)


Uzlopak (approved):

Uzlopak:
windows tests fail

KhafraDev:
wanna help figure out why

Uzlopak:
sure, will boot into windows

KhafraDev:
a windows only issue that's either with the testrunner or with webstreams. a truly fun issue. I can't get it to repro w/o the test runner

KhafraDev:
@Uzlopak minimal example, without using fetch 😒 

```mjs

function cloneBody (body) {
  // To clone a body body, run these steps:

  // https://fetch.spec.whatwg.org/#concept-body-clone

  // 1. Let « out1, out2 » be the result of teeing body’s stream.
  const [out1, out2] = body.stream.tee()
  const out2Clone = structuredClone(out2, { transfer: [out2] })
  // This, for whatever reasons, unrefs out2Clone which allows
  // the process to exit by itself.
  const [, finalClone] = out2Clone.tee()

  // 2. Set body’s stream to out1.
  body.stream = out1

  // 3. Return a body whose stream is out2 and other members are copied from body.
  return {
    stream: finalClone,
    length: body.length,
    source: body.source
  }
}

test('test', async (t) => {
  const rs = new ReadableStream({
    pull (controller) {
      controller.byobRequest?.respond(0)
    },
    type: 'bytes'
  })

  cloneBody({ stream: rs })
})

```

node --test test.js

test hangs

Uzlopak:
If I undo your commit it works locally. If I apply your commit it hangs. Maybe the original request is not finishing. Maybe we need to keep some reference between the clones and the original and if a clone closes/finishes/ends/errors whatever, it also destroys the original request?

KhafraDev:
its a bug in webstreams, we're basically replacing one bug for another.

KhafraDev:
We'll just fail the tests, every other platform does anyways last I checked. We shouldn't be bothering with workarounds for bugs in node core, this issue is more important to fix.

Uzlopak:
We patched it for wpt and that resulted in a bug in the actual tests? So we revert the change in the code so that wpt fails but the actual tests not?

KhafraDev:
yeah

Uzlopak (approved):

ronag (approved):

PR #2901, opened 2024-03-02
tests: ignore catch block when requiring crypto module

Uzlopak:
Just ignore the catch block to improve test coverage


mcollina (approved):
lgtm

PR #2903, opened 2024-03-02
perf(headers): a single set-cookie

tsctx:
**benchmarks/fetch/headers-length32.mjs**
- main
```
benchmark              time (avg)             (min … max)       p75       p99      p999
--------------------------------------------------------- -----------------------------
Headers@@iterator   5'731 ns/iter     (4'300 ns … 516 µs)  5'300 ns 11'600 ns    115 µs
```

- this patch
```
benchmark              time (avg)             (min … max)       p75       p99      p999
--------------------------------------------------------- -----------------------------
Headers@@iterator   5'151 ns/iter     (4'100 ns … 689 µs)  4'800 ns 11'100 ns    109 µs
```

### DIFF
--- a/benchmarks/fetch/headers-length32.mjs
+++ b/benchmarks/fetch/headers-length32.mjs
@@ -19,7 +19,7 @@ const headers = new Headers(
     'Width',
     'Accept-CH',
     'Via',
-    'Refresh',
+    'Set-Cookie',
     'Server',
     'Sec-Fetch-Dest',
     'Sec-CH-UA-Model',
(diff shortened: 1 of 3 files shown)


benchmarks/fetch/headers-length32.mjs
```diff
@@ -19,7 +19,7 @@ const headers = new Headers(
     'Width',
     'Accept-CH',
     'Via',
-    'Refresh',
```
mcollina, line 22: why was this removed?

tsctx: This is because I replaced it with set-cookie.
The length, which is a fast-path condition, must be 32.

KhafraDev (approved):

Uzlopak (approved):
I added a unit test to ensure it works as expected.

mcollina (approved):
lgtm

PR #2908, opened 2024-03-03
chore: remove import-fresh as devDependency

Uzlopak:
we dont need import-fresh

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,7 +104,6 @@
     "formdata-node": "^6.0.3",
     "https-pem": "^3.0.0",
     "husky": "^9.0.7",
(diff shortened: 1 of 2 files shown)


KhafraDev (commented):
It seems like we do need import-fresh since the functionality is being reimplemented?

Uzlopak:
@KhafraDev 
Its actually just a trivial two liner. I added only the asserts to make the review easier as it shows, that the two liner is a valid replacement.

ronag (approved):

mcollina (approved):
lgtm

PR #2912, opened 2024-03-03
chore: increase test coverage to 100% for /lib/api/api-request.js

Uzlopak:
adding some test cases

wondering if we actually should actually also check if hwm is a finite number so we can exclude NaN and Infinity as invalid cases.


Uzlopak:
Lol, why does it reduce the test coverage, when i am adding tests?

KhafraDev (approved):

PR #2921, opened 2024-03-04
perf: optimize getResolveErrorBodyCallback

tsctx:
I understand that this is not a hot path, but this function is slow.


Uzlopak:
Do you have any bencchmarks?

tsctx:
> Do you have any bencchmarks?

Benchmarking is difficult because it is an internal function 😓.

Uzlopak:
ok, makes sense ;)

Uzlopak (approved):
LGTM

Uzlopak:
@tsctx can you fix the merge conflict?

metcoder95 (approved):

metcoder95 (approved):

lib/api/util.js
```diff
@@ -26,13 +28,13 @@ async function getResolveErrorBodyCallback ({ callback, body, contentType, statu
 
   try {
     if (contentType.startsWith('application/json')) {
-      const payload = JSON.parse(toUSVString(Buffer.concat(chunks)))
+      const payload = JSON.parse(chunksDecode(chunks, length))
```
Uzlopak, line 31: I think this is false. If I understand it correctly chunksDecode second parameter is about the amount of array items and not the total length of the whole Buffer.

Uzlopak (changes requested):
probably wrong use of chunksDecode

lib/api/util.js
tsctx: No. The length of the buffer.

https://nodejs.org/docs/latest/api/buffer.html#static-method-bufferconcatlist-totallength

Uzlopak (approved):
LGTM

PR #2922, opened 2024-03-05
fix: chunksDecode cuts off 3 characters at the end if having BOM

Uzlopak:
I was writing some unit tests, when I realized that chunksDecode is cutting of characters when BOM is provided. We currently had no unit tests for chunks with BOM to detect this.... so I provide this PR to fix it.


lib/api/readable.js
```diff
@@ -284,16 +284,17 @@ function chunksDecode (chunks, length) {
...
-     ? 3
-     : 0
-  return buffer.utf8Slice(start, buffer.length - start)
+    bufferLength > 2 &&
+    buffer[0] === 0xef &&
+    buffer[1] === 0xbb &&
+    buffer[2] === 0xbf
+      ? 3
+      : 0
+  return buffer.utf8Slice(start, bufferLength)
```
Uzlopak, line 297: buffer length - start was wrong. needed to slice till the buffer end ;).

KhafraDev (approved):

ronag (approved):

PR #2924, opened 2024-03-05
add busboy tests

KhafraDev:


KhafraDev:
I removed all the tests that relied on busboy's limits, partial parsing, and some other things. Once I add `info` to the actual parsing we should be able to see what issues in the parsing there are (if any, I hope).

KhafraDev:
Surprisingly not many issues left, seemingly only edge cases that the RFCs aren't too open about.
- does the value of "name" have to be surrounded in quotes?
- supporting filename* params (we don't have to, but we should)
- empty `content-disposition` header ("Each part MUST contain a Content-Disposition header field [[RFC2183](https://www.rfc-editor.org/rfc/rfc2183)] where the disposition type is "form-data".")
- folded header value ("Header fields are lines beginning with a field name, followed by a colon (":"), followed by a field body, and terminated by CRLF.").
- no content-disposition (same reason as before)
- empty part (I think the only actual bug here, a form-data could start with \r\n)

@Uzlopak do you know why busboy/fastify's busboy supports some of these cases that go against the spec?

Uzlopak:
We had cases where busboy would throw errors and crashing servers. I think you are referring to those cases, where we handled them gracefully to avoid server crashes.

Also we had issues like these https://github.com/fastify/busboy/pull/71 where the implementation into busboy is not trivial.

KhafraDev:
It's in busboy, not @fastify/busboy fwiw. I just thought the libraries were pretty similar.

Uzlopak:
Cant we just take the cases and write a node:test based runner?

KhafraDev:
yeah, I'm more interested in fixing the two valid cases first before doing so. I'm thinking that busboy is clearly wrong to handle the other cases so we shouldn't.

KhafraDev:
@Uzlopak ptal

mcollina (approved):
lgtm

PR #2928, opened 2024-03-06
override request dispatcher from init

matthieusieben:
## This relates to...

The following comment: https://github.com/nodejs/undici/pull/2831/files#r1501662661

The following logic should result in `CustomAgentB` agent being used. It is currently not the case.

```ts
  const request = new Request(`http://localhost`, {
    dispatcher: new CustomAgentA()
  })

  await fetch(request, {
    dispatcher: new CustomAgentB()
  })
```

## Rationale

in both `Request` and `fetch()`, `init` is meant to override the values from the input. This is currently true for most fields but gets ignored in case of the `dispacher` field when `input` is a request instance.

## Changes

`init.dispatcher` takes precedence over `input[kDispatcher]` when creating a request from a request.

### Features

N/A

### Bug Fixes

- https://github.com/nodejs/undici/pull/2831/files#r1501662661

- If an implementation relied on the wrong dispatcher being used, this change might cause an issue.

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/web/fetch/request.js
+++ b/lib/web/fetch/request.js
@@ -103,7 +103,7 @@ class Request {
       // 5. Set fallbackMode to "cors".
       fallbackMode = 'cors'
     } else {
-      this[kDispatcher] = input[kDispatcher]
+      this[kDispatcher] = init.dispatcher || input[kDispatcher]
 
       // 6. Otherwise:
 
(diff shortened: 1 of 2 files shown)


KhafraDev (approved):

Uzlopak (approved):
Thank you

PR #2941, opened 2024-03-10
ci: fix broken ci on windows and node v21 because of libuv bug

Uzlopak:
Actually all the fixtures terminate the servers after the first request. So we just wont call child.kill() and it should not error anymore on windows.

image


KhafraDev (approved):

ronag:
Does libuv know about the bug? Is there an issue we can reference?

Uzlopak:
@ronag 

Dont worry. It is already taken care of in libuv. ;)

https://github.com/nodejs/node/issues/51766#issuecomment-1979574755

PR #2956, opened 2024-03-13
types: align RequestInit.body type with lib.dom.ts

jdufresne:
## This relates to...

#1172

## Rationale

The `RequestInit.body` property should allow `null` to match the TypeScript DOM types.

Expected to help avoid typing errors such as:

```
Argument of type 'RequestInit | undefined' is not assignable to parameter of type 'import(".../node_modules/undici/types/fetch").RequestInit | undefined'.
  Type 'RequestInit' is not assignable to type 'import(".../node_modules/undici/types/fetch").RequestInit'.
    Types of property 'body' are incompatible.
      Type 'BodyInit | null | undefined' is not assignable to type 'BodyInit | undefined'.

     const fetchRequest = new Request(fetchUrl, options);
                                                ~~~~~~~
```

## Changes

The `RequestInit.body` property now allows `null` in addition to `undefined`.

### Features

N/A

### Bug Fixes

N/A

N/A

## Status

- [x] Tested
- [s] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/test/types/fetch.test-d.ts
+++ b/test/types/fetch.test-d.ts
@@ -31,6 +31,7 @@ const requestInit2: RequestInit = {
 const requestInit3: RequestInit = {}
 // Test assignment. See https://github.com/whatwg/fetch/issues/1445
 requestInit3.credentials = 'include'
+const requestInit4: RequestInit = { body: null }
 
 declare const request: Request
 declare const headers: Headers
--- a/types/fetch.d.ts
+++ b/types/fetch.d.ts
@@ -122,7 +122,7 @@ export interface RequestInit {
   method?: string
   keepalive?: boolean
   headers?: HeadersInit
-  body?: BodyInit
+  body?: BodyInit | null
   redirect?: RequestRedirect
   integrity?: string
   signal?: AbortSignal | null


KhafraDev:
I'd be more inclined to think that typescript's types are wrong; `body` can technically be quite a few values, but the types should portray what values are expected, not what is allowed. For example, the `RequestInit` type doesn't include `Function` or `Regexp`, even though they technically are allowed.

KhafraDev:
Maybe @Ethan-Arrowood could give an opinion. Is it better to align ourselves with the DOM types or should they be aligned with the spec?

jdufresne:
@KhafraDev 

The standards suggests that `null` is allowed and supported:

https://fetch.spec.whatwg.org/#ref-for-dom-request%E2%91%A0

> [body](https://fetch.spec.whatwg.org/#dom-requestinit-body)
>
> A [BodyInit](https://fetch.spec.whatwg.org/#bodyinit) object **or null** to set request’s [body](https://fetch.spec.whatwg.org/#concept-request-body).

Emphasis added.

Ethan-Arrowood:
Both our types and TypeScript's should align to the spec. If either are incorrect we should contribute the necessary fix to make sure everyone is on the same page. If anything -- I'd expect TypeScript's DOM types to perfectly match spec. And ours can deviate where appropriate, but in this instance if `null` is an allowed value then we should update accordingly

KhafraDev:
So there's no mention of null being a special case other than the developer part and webidl definition. 

Typescript Dom types are usually pretty bad so I wouldn't rely on them as the penultimate source of truth. In this case they do seem to be right.

KhafraDev (approved):

Ethan-Arrowood:
Yeah absolutely not rely on TS types. Spec is only truth lol (and I guess how browsers actually implement things)

PR #2959, opened 2024-03-14
Cleanup format

KhafraDev:
FormData is better now, made Response/Request start with `Response {` or `Request {`, etc.

```js
> var fd = new FormData()
> fd.set('a', 'b')
> fd.append('a', new Blob(['hello']), 'hello.txt')
> fd.set('file', new Blob(['file']))
> fd

FormData {
  a: [
    'b',
    File { size: 5, type: '', name: 'hello.txt', lastModified: 1710432489080 }
  ],
  file: File { size: 4, type: '', name: 'blob', lastModified: 1710432492961 }
}
```


lib/web/fetch/formdata.js
```diff
@@ -157,12 +157,24 @@ class FormData {
...
+          a[b.name].push(b.value)
+        } else {
+          a[b.name] = [a[b.name], b.value]
+        }
+      } else {
+        a[b.name] = b.value
+      }
+
+      return a
+    }, {})
```
Uzlopak, line 172: What about prototype pollution?

KhafraDev: doesn't matter, there are ten thousand areas vulnerable to prototype pollution

Uzlopak: You joking, right?

KhafraDev: no, we don't use primordials anywhere in the codebase (nor should we). it's a non-issue.

KhafraDev: oh I see what you're talking about... I thought you meant more generally

mcollina (approved):
lgtm

Uzlopak (approved):
Lgtm

PR #2960, opened 2024-03-14
test: fix flakyness of issue-803 test

Uzlopak:
will this reduce the flakyness in macos? we will see.

Improving he performance and increasing the timeout did the trick.


mcollina (approved):
lgtm

Issue #2961, opened 2024-03-14
Option mismatch in RetryHandler


acommodari:
## Bug Description

There seems to be an option mismatch in the RetryHandler.

## Reproducible By

```javascript
const client = new Client("http://localhost:3000").compose(
  retry({
    retry: function (err, context, callback) {
      console.log("retrying", {
        err: err.code,
        state: context.state,
        opts: context.opts.retryOptions,
      });
      if (context.state.counter >= context.opts.retryOptions.maxRetries) {
        callback(err);
        return;
      }
      // Both the docs and types made me thing that context.opts.retryOptions.minTimeout 
      // should exist as an option but it in reality is context.opts.retryOptions.timeout
      setTimeout(() => callback(null), context.opts.retryOptions.timeout);
    },
  }),
  minTimeout: 400, // here it is called minTimeout
);

```

## Expected Behavior

Either the property in the code should be changed from `timeout` to `minTimeout` or the docs + types should be updated to reflect the difference.

## Logs & Screenshots

in `retry-handler.js` of `undici` v6.9.0 on line 40

```javascript
class RetryHandler {
  constructor (opts, handlers) {
    const { retryOptions, ...dispatchOpts } = opts
    const {
      // Retry scoped
      retry: retryFn,
      maxRetries,
      maxTimeout,
      minTimeout,
      timeoutFactor,
      // Response scoped
      methods,
      errorCodes,
      retryAfter,
      statusCodes
    } = retryOptions ?? {}

    this.dispatch = handlers.dispatch
    this.handler = handlers.handler
    this.opts = dispatchOpts
    this.abort = null
    this.aborted = false
    this.retryOpts = {
      retry: retryFn ?? RetryHandler[kRetryHandlerDefaultRetry],
      retryAfter: retryAfter ?? true,
      maxTimeout: maxTimeout ?? 30 * 1000, // 30s,
      timeout: minTimeout ?? 500, // .5s. // <=== here we are setting the value to timeout
      timeoutFactor: timeoutFactor ?? 2,
      maxRetries: maxRetries ?? 5,

...
```

## Environment

macOS Ventura 13.2.1
node 20.11.1
npm 10.2.4

### Additional context

I am willing to open a PR myself if needed.

metcoder95:
Thanks for the report! A PR is always welcomed 👍

acommodari:
PR opened 🙏

PR #2962, opened 2024-03-14
fix: fix retry handler option

acommodari:
## This relates to...

#2961 

## Rationale

To fix the issue #2961 

## Changes

- add `use strict` to `retry-handler.js`
- change `timeout` option in `retry-handler.js` to `minTimeout` to match docs and types.

### Features

N/A

### Bug Fixes

- fix docs/types mismatch

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/handler/retry-handler.js
+++ b/lib/handler/retry-handler.js
@@ -1,3 +1,4 @@
+'use strict'
 const assert = require('node:assert')
 
 const { kRetryHandlerDefaultRetry } = require('../core/symbols')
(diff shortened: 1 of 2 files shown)


ronag (commented):
can we have a test?

acommodari:
> can we have a test?

sure! in this case what would you like the test to cover? Would this be a test for types?

Uzlopak:
write a test where the minTimeout is set

metcoder95 (approved):

PR #2967, opened 2024-03-15
feat(TS): add types for composed dispatchers

metcoder95:
## This relates to...

Adding missing types for `Dispatcher#compose` and `ComposeDispatcher`

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/types/dispatcher.d.ts
+++ b/types/dispatcher.d.ts
@@ -96,7 +96,7 @@ declare class Dispatcher extends EventEmitter {
 }
 
 declare namespace Dispatcher {
-  export interface ComposedDispatcher {
+  export interface ComposedDispatcher extends Dispatcher {
     close: Dispatcher['close'];
     destroy: Dispatcher['destroy'];
     dispatch: Dispatcher['dispatch'];
--- a/types/dispatcher.d.ts
+++ b/types/dispatcher.d.ts
@@ -18,6 +18,9 @@ declare class Dispatcher extends EventEmitter {
   /** Starts two-way communications with the requested resource. */
   connect(options: Dispatcher.ConnectOptions): Promise<Dispatcher.ConnectData>;
   connect(options: Dispatcher.ConnectOptions, callback: (err: Error | null, data: Dispatcher.ConnectData) => void): void;
+  /** Compose a chain of dispatchers */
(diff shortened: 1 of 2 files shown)


types/dispatcher.d.ts
```diff
@@ -93,6 +96,12 @@ declare class Dispatcher extends EventEmitter {
 }
 
 declare namespace Dispatcher {
+  export interface ComposedDispatcher {
```
mcollina, line 99:
```suggestion
  export interface ComposedDispatcher extends Dispatcher {
```

mcollina (changes requested):

clovis-guillemot:
Hello guys, is that could be merged asap please, I need it to avoid this [warn](https://github.com/nodejs/undici/blob/dcab693198f3d3aea0651bdc94ee975855013302/lib/dispatcher/client.js#L216) _'UNDICI-CLIENT-INTERCEPTOR-DEPRECATED'_ 
regards

types/dispatcher.d.ts
```diff
@@ -93,6 +96,12 @@ declare class Dispatcher extends EventEmitter {
...
+    close: Dispatcher['close'];
+    destroy: Dispatcher['destroy'];
+    dispatch: Dispatcher['dispatch'];
+  }
```
KhafraDev, line 103: this could probably be

```suggestion
  export type ComposedDispatcher = Pick<Dispatcher, 'close' | 'destroy' | 'dispatch'>
```

types/dispatcher.d.ts
metcoder95: I was thinking about doing the same, but not sure; the `ComposedDispatcher` just has a subset of the `Dispatcher` methods, are we ok with that?

mcollina: `ComposedDispatcher` extends `Dispatcher`, so it has all the methods: https://github.com/nodejs/undici/blob/9a56796b3e61713f7f2e44ea8c9622b158697c45/lib/dispatcher/dispatcher.js#L42

metcoder95: 😓 

Thanks, I overlooked it!

PR #2971, opened 2024-03-18
fix(fetch): properly redirect non-ascii location header url

Xvezda:
## Rationale

The fetch function behaves differently from major web browsers(chrome, safari, firefox etc), and even CURL (with `-L` option).

Here is a minimal reproducible URL I made: https://non-ascii-location-header.sys.workers.dev/redirect

<details>
<summary>Worker code</summary>

```ts
export default {
	async fetch(request: Request, env: Env, ctx: ExecutionContext): Promise<Response> {
		try {
			const url = new URL(request.url);
			if (url.pathname === '/redirect') {
				return new Response('', {
					status: 302,
					headers: {
						Location: '/안녕',
					},
				});
			}
			return new Response(`pathname ${url.pathname}`);
		} catch (e) {
			return new Response(`error: ${(e as Error).message}`, { status: 500 });
		}
	},
};
```

</details>

By accessing the address at https://non-ascii-location-header.sys.workers.dev/redirect directly 
or https://non-ascii-location-header.sys.workers.dev/ first and then making a fetch request to the /redirect path using the developer tools leads to reaching https://non-ascii-location-header.sys.workers.dev/%EC%95%88%EB%85%95  
which shows `pathname /%EC%95%88%EB%85%95` as a result.

However, using undici fetch function does not behave in the same way.

```
> r = await fetch('https://non-ascii-location-header.sys.workers.dev/redirect')
... await r.text();
pathname /%C3%AC%C2%95%C2%88%C3%AB%C2%85%C2%95
```

This issue can also be observed at chrome web store: https://chromewebstore.google.com/detail/gbgmenmdglilmbmemagekpeaodajbeei

```
Welcome to Node.js v21.6.2.
> await fetch('https://chromewebstore.google.com/detail/gbgmenmdglilmbmemagekpeaodajbeei')
Uncaught TypeError: fetch failed
    at node:internal/deps/undici/undici:12443:11
    at async REPL1:1:33 {
  cause: Error: redirect count exceeded
      at makeNetworkError (node:internal/deps/undici/undici:5675:35)
      at httpRedirectFetch (node:internal/deps/undici/undici:10696:32)
      at httpFetch (node:internal/deps/undici/undici:10668:28)
      at async node:internal/deps/undici/undici:10440:20
      at async mainFetch (node:internal/deps/undici/undici:10430:20)
      at async httpFetch (node:internal/deps/undici/undici:10668:22)
      at async node:internal/deps/undici/undici:10440:20
      at async mainFetch (node:internal/deps/undici/undici:10430:20)
      at async httpFetch (node:internal/deps/undici/undici:10668:22)
}
```

I do not have an in-depth understanding of the fetch standard, so if this PR request violates the standard in any way, please feel free to close it.

Thank you.

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/web/fetch/util.js
+++ b/lib/web/fetch/util.js
@@ -44,7 +44,8 @@ function responseLocationURL (response, requestFragment) {
   // 3. If location is a header value, then set location to the result of
   //    parsing location with response’s URL.
   if (location !== null && isValidHeaderValue(location)) {
-    location = new URL(location, responseURL(response))
+    // Make sure location is properly encoded.
+    location = new URL(Buffer.from(location, 'ascii').toString('utf8'), responseURL(response))
   }
 
   // 4. If location is a URL whose fragment is null, then set location’s
--- a/lib/web/fetch/util.js
+++ b/lib/web/fetch/util.js
@@ -44,8 +44,10 @@ function responseLocationURL (response, requestFragment) {
   // 3. If location is a header value, then set location to the result of
   //    parsing location with response’s URL.
   if (location !== null && isValidHeaderValue(location)) {
-    // Make sure location is properly encoded.
-    location = new URL(Buffer.from(location, 'ascii').toString('utf8'), responseURL(response))
+    // Some websites respond location header in binary form without encoding them
+    // and major browsers redirect them to correctly UTF-8 encoded addresses.
+    // Here, we handle that behavior in the same way.
+    location = new URL(decodeBinaryStringToUtf8(location), responseURL(response))
   }
 
   // 4. If location is a URL whose fragment is null, then set location’s
@@ -56,6 +58,14 @@ function responseLocationURL (response, requestFragment) {
 
   // 5. Return location.
   return location
+}
+
+/**
+ * @param {string} value
+ * @returns {string}
+ */
+function decodeBinaryStringToUtf8 (value) {
+  return Buffer.from(value, 'binary').toString('utf8')
 }
 
 /** @returns {URL} */
--- a/lib/web/fetch/util.js
+++ b/lib/web/fetch/util.js
@@ -47,7 +47,7 @@ function responseLocationURL (response, requestFragment) {
     // Some websites respond location header in binary form without encoding them
     // and major browsers redirect them to correctly UTF-8 encoded addresses.
     // Here, we handle that behavior in the same way.
-    location = new URL(decodeBinaryStringToUtf8(location), responseURL(response))
+    location = new URL(normalizeBinaryStringToUtf8(location), responseURL(response))
   }
 
   // 4. If location is a URL whose fragment is null, then set location’s
@@ -61,10 +61,12 @@ function responseLocationURL (response, requestFragment) {
 }
 
 /**
+ * If string contains non-ASCII characters, assumes it's UTF-8 encoded and decodes it.
+ * Since UTF-8 is a superset of ASCII, this will work for ASCII strings as well.
  * @param {string} value
  * @returns {string}
  */
-function decodeBinaryStringToUtf8 (value) {
+function normalizeBinaryStringToUtf8 (value) {
   return Buffer.from(value, 'binary').toString('utf8')
 }
 
--- a/lib/web/fetch/util.js
+++ b/lib/web/fetch/util.js
@@ -45,7 +45,7 @@ function responseLocationURL (response, requestFragment) {
   //    parsing location with response’s URL.
   if (location !== null && isValidHeaderValue(location)) {
     if (!isValidEncodedURL(location)) {
-      // Some websites respond location header in binary form without encoding them
+      // Some websites respond location header in UTF-8 form without encoding them as ASCII
       // and major browsers redirect them to correctly UTF-8 encoded addresses.
       // Here, we handle that behavior in the same way.
       location = normalizeBinaryStringToUtf8(location)
--- a/lib/web/fetch/util.js
+++ b/lib/web/fetch/util.js
@@ -72,7 +72,7 @@ function isValidEncodedURL (url) {
   for (const c of url) {
     const code = c.charCodeAt(0)
     // Not used in US-ASCII
-    if (code >= 0x80 && code <= 0xFF) {
+    if (code >= 0x80) {
       return false
     }
     // Control characters
--- a/lib/web/fetch/util.js
+++ b/lib/web/fetch/util.js
@@ -44,10 +44,13 @@ function responseLocationURL (response, requestFragment) {
   // 3. If location is a header value, then set location to the result of
   //    parsing location with response’s URL.
   if (location !== null && isValidHeaderValue(location)) {
-    // Some websites respond location header in binary form without encoding them
-    // and major browsers redirect them to correctly UTF-8 encoded addresses.
-    // Here, we handle that behavior in the same way.
-    location = new URL(normalizeBinaryStringToUtf8(location), responseURL(response))
+    if (!isValidEncodedURL(location)) {
+      // Some websites respond location header in binary form without encoding them
+      // and major browsers redirect them to correctly UTF-8 encoded addresses.
+      // Here, we handle that behavior in the same way.
+      location = normalizeBinaryStringToUtf8(location)
+    }
+    location = new URL(location, responseURL(response))
   }
 
   // 4. If location is a URL whose fragment is null, then set location’s
@@ -58,6 +61,26 @@ function responseLocationURL (response, requestFragment) {
 
   // 5. Return location.
   return location
+}
+
+/**
+ * @see https://www.rfc-editor.org/rfc/rfc1738#section-2.2
+ * @param {string} url
+ * @returns {boolean}
+ */
+function isValidEncodedURL (url) {
+  for (const c of url) {
+    const code = c.charCodeAt(0)
+    // Not used in US-ASCII
+    if (code >= 0x80 && code <= 0xFF) {
+      return false
+    }
+    // Control characters
+    if ((code >= 0x00 && code <= 0x1F) || code === 0x7F) {
+      return false
+    }
+  }
+  return true
 }
 
 /**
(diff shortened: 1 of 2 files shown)


KhafraDev (approved):
this is correct spec-wise

KhafraDev:
@anonrig do you know why this is needed?

anonrig:
cc @lemire

anonrig:
Please don't merge it until I have some time to review this tomorrow.

KhafraDev:
Seems like a bug in URL?

```
> base = `http://127.0.0.1:51619/redirect`
'http://127.0.0.1:51619/redirect'
> location = `/ì\x95\x88ë\x85\x95`
'/ì\x95\x88ë\x85\x95'
> new URL(location, base).toString()
'http://127.0.0.1:51619/%C3%AC%C2%95%C2%88%C3%AB%C2%85%C2%95'
```

anonrig (changes requested):
@KhafraDev This is intended, and the output is correct. Look at the example below:

```
> let u = new URL('http://127.0.0.1:51619/redirect')
undefined
> u.pathname = '/ì\x95\x88ë\x85\x95'
'/ì\x95\x88ë\x85\x95'
> u
URL {
  href: 'http://127.0.0.1:51619/%C3%AC%C2%95%C2%88%C3%AB%C2%85%C2%95',
  origin: 'http://127.0.0.1:51619',
  protocol: 'http:',
  username: '',
  password: '',
  host: '127.0.0.1:51619',
  hostname: '127.0.0.1',
  port: '51619',
  pathname: '/%C3%AC%C2%95%C2%88%C3%AB%C2%85%C2%95',
  search: '',
  hash: ''
}
```

URL specification path state (https://url.spec.whatwg.org/#path-state) says that:

[UTF-8 percent-encode](https://url.spec.whatwg.org/#utf-8-percent-encode) [c](https://url.spec.whatwg.org/#c) using the [path percent-encode set](https://url.spec.whatwg.org/#path-percent-encode-set) and append the result to buffer.

anonrig:
You can replicate the same behavior on Safari as well.

lib/web/fetch/util.js
```diff
@@ -44,7 +44,8 @@ function responseLocationURL (response, requestFragment) {
   // 3. If location is a header value, then set location to the result of
   //    parsing location with response’s URL.
   if (location !== null && isValidHeaderValue(location)) {
-    location = new URL(location, responseURL(response))
+    // Make sure location is properly encoded.
+    location = new URL(Buffer.from(location, 'ascii').toString('utf8'), responseURL(response))
```
lemire, line 48: This seems strange to me: `Buffer.from(location, 'ascii').toString('utf8')`. I suggest at least some comments to explain why the intention is.

lemire: At glance, what this seems to do is replace any non-ASCII input by the replacement character? I don't understand. The fetch spec seems to say that you have to call the pathname setter:
Screenshot 2024-03-18 at 7 06 00 PM

lemire:
In Node 21, I am getting...

```
> var u = new URL('https://non-ascii-location-header.sys.workers.dev/redirect')
undefined
> u.pathname = "/안녕"
'/안녕'
> u
URL {
  href: 'https://non-ascii-location-header.sys.workers.dev/%EC%95%88%EB%85%95',
  origin: 'https://non-ascii-location-header.sys.workers.dev',
  protocol: 'https:',
  username: '',
  password: '',
  host: 'non-ascii-location-header.sys.workers.dev',
  hostname: 'non-ascii-location-header.sys.workers.dev',
  port: '',
  pathname: '/%EC%95%88%EB%85%95',
  search: '',
  hash: ''
}
```

This is correct, right?

lib/web/fetch/util.js
lemire: What does "parsing location with response URL" mean precisely? I interpret it is a pathname setter...

KhafraDev: > What does "parsing location with response URL" mean precisely? I interpret it is a pathname setter...

The spec links "parsing" to https://url.spec.whatwg.org/#concept-url-parser

KhafraDev:
> This is intended, and the output is correct. Look at the example below:

This seems like quite a bad issue if it's causing pages not to load. How does Safari handle loading the google webstore in this case ([example](https://chromewebstore.google.com/detail/gbgmenmdglilmbmemagekpeaodajbeei))?

lib/web/fetch/util.js
Xvezda: @lemire 
This issue arises when the location header's address is not encoded as with `encodeURI`, but is responded in binary form. 
For example, while the result of `encodeURIComponent('안녕')` is `%EC%95%88%EB%85%95`, in binary form, this appears as `\xEC\x95\x88\xEB\x85\x95`, similar to the outcome of `Buffer.from('안녕', 'utf8').toString('binary')`. 
What I am attempting in this code is to normalize characters that exist outside the ASCII range into UTF-8 form and URL encode them, to ensure behavior consistent with major web browsers.

Xvezda:
I've added comments to make the code clearer and revised it to be more concise. Please take a look.

lib/web/fetch/util.js
```diff
@@ -57,6 +60,16 @@ function responseLocationURL (response, requestFragment) {
...
 
+/**
+ * If string contains non-ASCII characters, assumes it's UTF-8 encoded and decodes it.
+ * Since UTF-8 is a superset of ASCII, this will work for ASCII strings as well.
+ * @param {string} value
+ * @returns {string}
+ */
+function normalizeBinaryStringToUtf8 (value) {
+  return Buffer.from(value, 'binary').toString('utf8')
+}
```
lemire, line 71:
```

```C++
  let b = Buffer.from(value, 'binary');
  if(buffer.isAscii(b)) { return value; } // skip new string alloc.
  return b.toString('utf8'); // must alloc a new string
```

mcollina (approved):
lgtm

lib/web/fetch/util.js
anonrig: No need to call `Buffer.from()`. `Buffer.isAscii(b)` is a static function.

lib/web/fetch/util.js
```diff
@@ -44,7 +44,10 @@ function responseLocationURL (response, requestFragment) {
   // 3. If location is a header value, then set location to the result of
   //    parsing location with response’s URL.
   if (location !== null && isValidHeaderValue(location)) {
-    location = new URL(location, responseURL(response))
+    // Some websites respond location header in binary form without encoding them
```
lemire:
```suggestion
    // Some websites respond location header in non-ASCII (UTF-8) form without encoding them into ASCII
```

lib/web/fetch/util.js
```diff
@@ -47,7 +47,7 @@ function responseLocationURL (response, requestFragment) {
     // Some websites respond location header in binary form without encoding them
     // and major browsers redirect them to correctly UTF-8 encoded addresses.
     // Here, we handle that behavior in the same way.
-    location = new URL(decodeBinaryStringToUtf8(location), responseURL(response))
+    location = new URL(normalizeBinaryStringToUtf8(location), responseURL(response))
```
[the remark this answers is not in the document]
Xvezda: @lemire Would this be better? efccdde

lib/web/fetch/util.js
lemire:
```suggestion
    // Some websites respond location header in UTF-8 form without encoding them as ASCII
```

lib/web/fetch/util.js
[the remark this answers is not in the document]
lemire: The benefit of Buffer.isAscii is that you calling on C++ which might be faster, but your code is nice too. I think it clarifies the issue as well.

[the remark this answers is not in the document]
Xvezda: @lemire 
Could `Buffer.from` potentially incur significant costs? Since `buffer.isAscii` [does not accept a string input](https://nodejs.org/docs/latest-v20.x/api/buffer.html#bufferisasciiinput), I wrote it this way.

lib/web/fetch/util.js
lemire: @anonrig We get at the core of the issue here. The `value` is a string not a buffer. At some point it was wrongly interpreted as latin1 (maybe in `http`) and so we must now convert it to a buffer and then back to a string.


I think that this function should receive a Buffer and then the Buffer could be interpreted as UTF-8 and we'd be done.

lemire: But this is far beyond the scope of this PR.

lemire: The irony is that we seem to try to purposefully (in http) prevent people from sending UTF-8 content as UTF-8... and then we go back and say "no, it is really UTF-8".

Much technical debt.

lib/web/fetch/util.js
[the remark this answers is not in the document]
lemire: @Xvezda What you did is good. I am not arguing. It is probably even better than my proposal because you sidestep entirely the creation of a Buffer.

KhafraDev (approved):

lib/web/fetch/util.js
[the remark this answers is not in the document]
lemire: In my view, the new code is easier to understand. It is absolutely clear now that it has to do with non-ASCII content.

KhafraDev (approved):

lib/web/fetch/util.js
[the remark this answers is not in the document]
Xvezda: I was just curious if the approach using Buffer.from might have more advantages. This should be sufficient then, thank you.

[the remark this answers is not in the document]
lemire: Yes. I definitively think your code is fine.

anonrig (approved):

PR #2984, opened 2024-03-22
perf: Micro-optimizations in isIterable, isAsyncIterable and parseHeaders core util

peterver:
## This relates to...

Micro optimizations regarding performance of core utils

## Rationale

I found two double negations that could be omitted as unnecessary, the map vs for...of adjustment is related to the simple fact that map operations are significantly slower compared to for...of. 

## Changes

- perf: lib/core/util@isIterable: Removed unnecessary double negation
- perf: lib/core/util@isAsyncIterable: Removed unnecessary double negation
- perf: lib/core/util@parseHeaders: Swapped out usage of map in favor of for...of with pre-assigned array

### Features

N/A

### Bug Fixes

N/A

N/A

## Status

- [x] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/core/util.js
+++ b/lib/core/util.js
@@ -158,11 +158,11 @@ function deepClone (obj) {
 }
 
 function isAsyncIterable (obj) {
-  return !!(obj != null && typeof obj[Symbol.asyncIterator] === 'function')
+  return obj != null && typeof obj[Symbol.asyncIterator] === 'function'
 }
 
 function isIterable (obj) {
-  return !!(obj != null && (typeof obj[Symbol.iterator] === 'function' || typeof obj[Symbol.asyncIterator] === 'function'))
+  return obj != null && (typeof obj[Symbol.iterator] === 'function' || typeof obj[Symbol.asyncIterator] === 'function')
 }
 
 function bodyLength (body) {
--- a/lib/core/util.js
+++ b/lib/core/util.js
@@ -264,8 +264,15 @@ function parseHeaders (headers, obj) {
       const headersValue = headers[i + 1]
       if (typeof headersValue === 'string') {
         obj[key] = headersValue
+      } else if (Array.isArray(headersValue)) {
+        const len = headersValue.length
+        const headersValueAsUTF8 = new Array(len)
+        for (let y = 0; y < len; y++) {
+          headersValueAsUTF8[y] = headersValue[y].toString('utf8')
+        }
+        obj[key] = headersValueAsUTF8
       } else {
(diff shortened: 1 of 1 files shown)


ronag (changes requested):
I don't think this does anything meaningful. Happy to be proven wrong with a microbenchmark.

peterver:
@ronag was just going off of benchmarks such as:
- https://leanylabs.com/blog/js-forEach-map-reduce-vs-for-for_of/
- https://github.com/dg92/Performance-Analysis-JS?ref=hackernoon.com.

Have seen some healthy performance improvements in personal libs I maintain when swapping out declarative behavior such as map/reduce/filter.

In hindsight I see that a large chunk of undici favors declarative, as such will close PR and refrain from opening up future PRs of this nature.

Uzlopak:
@ronag 

I think the change in parseHeaders was valid

peterver:
> @ronag
> 
> I think the change in parseHeaders was valid

In hindsight the order should probably be switched though for further micro-performance, instead of:

```
if (...) {
    ...
} else if (Array.isArray(...)) {
    ...
} else {
    ...
}
```

it should be:
```
if (...) {
    ...
} else if (!Array.isArray(...)) {
    ...
} else {
    ...
}
```

as I believe the notion of an array of values there has less occurrence than a string value

peterver:
@Uzlopak There's other places where this kind of micro-tuning could be beneficial, places like https://github.com/nodejs/undici/blob/8fce214517711203199421d2c00e96f64fff6f54/lib/web/fetch/formdata.js#L99 are currently doing more than O(n), something like the below would run in O(n):

```
const result = [];
let cursor;
for (let i = 0; i < this[kState].length; i++) {
    cursor = kState[i];
    if (cursor.name !== name) continue;
    result.push(cursor.value);
}
return result;
```

I saw several of these spots in core behaviors where performance is usually preferred.

Uzlopak:
What we need are supporting benchmarks. I am also pro improving cold paths, but e.g. ronag is a improve hot paths guy. So if you show that a change is improving the overall performance, than ronag is convincable to adapt the change.

peterver:
@Uzlopak I consider myself a performance junkie (and am also a first-time contributor to undici, having followed its development from afar for a couple years now), in what form would you wish to see the benchmarks? :nerd_face:

ronag:
I'm fine with cold path micro optimizations as far as it doens't impact readability/maintainability. I don't want to risk introducing bugs for this kind of stuff.

Issue #2808, opened 2024-02-22
`assert(!this.aborted)` triggers sporadically when `allowH2` is set to `true`


steveluscher:
## Bug Description

Under load, when used with an `Agent` dispatcher having `allowH2` set to `true`, fetch will sporadically throw an `ERR_ASSERTION`.

## Reproducible By

Click ‘fork and run’ here: https://replit.com/@StevenLuscher/DarkredIroncladMethods

## Expected Behavior

If the problem is that the underlying request is failing, I would expect it to trigger the `catch` block. Otherwise, we may be looking at a bug with the http/2 implementation itself.

## Logs & Screenshots

```
AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:

  assert(!this.aborted)

    at Request.onHeaders (/home/runner/DarkredIroncladMethods/node_modules/undici/lib/core/request.js:245:5)
    at ClientHttp2Stream.<anonymous> (/home/runner/DarkredIroncladMethods/node_modules/undici/lib/client.js:1793:17)
    at Object.onceWrapper (node:events:629:26)
    at ClientHttp2Stream.emit (node:events:514:28)
    at emit (node:internal/http2/core:331:3)
    at process.processTicksAndRejections (node:internal/process/task_queues:85:22) {
  generatedMessage: true,
  code: 'ERR_ASSERTION',
  actual: false,
  expected: true,
  operator: '=='
}
```

## Environment

* Node.js v20.10.0
* Linux baa638bf460f 6.5.0-1013-gcp #&#x2060;13~22.04.1-Ubuntu SMP Wed Jan 24 23:39:40 UTC 2024 x86_64 GNU/Linux
* undici 6.2.2

metcoder95:
Having an early assessment it seems this might have to be with the streams handling for H2; I'll try to dig into later Today or Tomorrow, have an idea where to look at

ktaekwon000:
Hi @metcoder95, thanks a lot for your work on the repo. Have you found the cause of the issue? This issue is affecting my workloads as well.

I also wanted to chip in that the trigger of this bug reminds me of https://github.com/nodejs/node/issues/47130 which affected Keep-Alive http/1.1 requests. (There is a workaround in that thread involving undici with Keep-Alive hints, but from my personal experience if the server kills the TCP connection without providing the Keep-Alive hints, undici is vulnerable to the same bug as node-fetch)

metcoder95:
👋 

[...]

PR #2985, opened 2024-03-22
fix(H2 Client): bind stream 'data' listener only after received 'response' event

St3ffGv4:
## This relates to...

https://github.com/nodejs/undici/issues/2808

## Rationale

Sporadically when doing massive h2 requests for a certain period of time a breaking assertion is triggered "assert(!this.aborted)" in onHeaders Method [undici/lib/core/request.js] causing application crash

## Changes

 - lib/dispatcher/client-h2.js

### Features

N/A

### Bug Fixes

- Sometimes happens that H2 Stream "data" event is triggered some milliseconds before the "response" event, this leads to an uninitialized body. To avoid this breaking error if the onData is called before onHeaders, return false so the stream is paused in client-h2.js

## Status

- [x] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

[cert]: https://github.com/nodejs/undici/blob/main/CONTRIBUTING.md #

### DIFF
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -253,7 +253,6 @@ function writeH2 (client, request) {
 
   /** @type {import('node:http2').ClientHttp2Stream} */
   let stream
-  let isStreamResponseStarted = false
 
   const { hostname, port } = client[kUrl]
 
@@ -389,7 +388,6 @@ function writeH2 (client, request) {
   ++session[kOpenStreams]
 
   stream.once('response', headers => {
-    isStreamResponseStarted = true
     const { [HTTP2_HEADER_STATUS]: statusCode, ...realHeaders } = headers
     request.onResponseStarted()
 
@@ -421,7 +419,7 @@ function writeH2 (client, request) {
   })
 
   stream.on('data', (chunk) => {
-    if (!isStreamResponseStarted || request.onData(chunk) === false) {
+    if (request.onData(chunk) === false) {
       stream.pause()
     }
   })
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -253,6 +253,7 @@ function writeH2 (client, request) {
 
   /** @type {import('node:http2').ClientHttp2Stream} */
   let stream
+  let isStreamResponseStarted = false
 
   const { hostname, port } = client[kUrl]
 
@@ -388,6 +389,7 @@ function writeH2 (client, request) {
   ++session[kOpenStreams]
 
   stream.once('response', headers => {
+    isStreamResponseStarted = true
     const { [HTTP2_HEADER_STATUS]: statusCode, ...realHeaders } = headers
     request.onResponseStarted()
 
@@ -419,7 +421,7 @@ function writeH2 (client, request) {
   })
 
   stream.on('data', (chunk) => {
-    if (request.onData(chunk) === false) {
+    if (!isStreamResponseStarted || request.onData(chunk) === false) {
       stream.pause()
     }
   })
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -394,6 +394,12 @@ function writeH2 (client, request) {
     if (request.onHeaders(Number(statusCode), realHeaders, stream.resume.bind(stream), '') === false) {
       stream.pause()
     }
+
+    stream.on('data', (chunk) => {
+      if (request.onData(chunk) === false) {
+        stream.pause()
+      }
+    })
   })
 
   stream.once('end', () => {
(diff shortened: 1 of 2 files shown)


ronag (changes requested):
This need to be fixed somewhere else. Can we defer the onData in the h2 wrapper?

St3ffGv4:
@ronag like doing this?

in client-h2.js

```javascript
let isResponseStarted = false

stream.once('response', headers => {
    isResponseStarted = true;
    const { [HTTP2_HEADER_STATUS]: statusCode, ...realHeaders } = headers
    request.onResponseStarted()

    if (request.onHeaders(Number(statusCode), realHeaders, stream.resume.bind(stream), '') === false) {
      stream.pause()
    }
  })
  
  stream.on('data', (chunk) => {
    if (!isResponseStarted || request.onData(chunk) === false) {
      stream.pause()
    }
  })
```

mcollina (changes requested):

St3ffGv4:
@mcollina this would maybe be a long term test, because the error can happen at any time, immediately or not, it would be like using the script reported in the issue and wait indefinitely if the scenario occurs

lib/dispatcher/client-h2.js
```diff
@@ -419,7 +421,7 @@ function writeH2 (client, request) {
   })
 
   stream.on('data', (chunk) => {
-    if (request.onData(chunk) === false) {
+    if (!isStreamResponseStarted || request.onData(chunk) === false) {
```
ronag, line 424: this will throw away the first chunk?

St3ffGv4: @ronag thaks, good point, I was focused on seeing that the error no longer occurs.

Is it ok to declare a temp chunks array and if isStreamResponseStarted is false in 'data' event push the chunk and then in 'response' event if the temp chunks array has length > 0 call the request.onData method after the request.onHeaders?

ronag: This does kind of feel like an error in node core http, it doesn't make sense to me that data is emitted before response. Do you think you could report it in the node core repo in addition to this workaround PR.

St3ffGv4: @ronag before report,
i was checking nodejs documentation about http2

there's one example here: https://nodejs.org/api/http2.html#clienthttp2sessionrequestheaders-options
```javascript
const http2 = require('node:http2');
const clientSession = http2.connect('https://localhost:1234');
const {
  HTTP2_HEADER_PATH,
  HTTP2_HEADER_STATUS,
} = http2.constants;

const req = clientSession.request({ [HTTP2_HEADER_PATH]: '/' });
req.on('response', (headers) => {
  console.log(headers[HTTP2_HEADER_STATUS]);
  req.on('data', (chunk) => { /* .. */ });
  req.on('end', () => { /* .. */ });
});
```

the data event listener is  binded inside the response listener callback.
So i have moved client-h2.js stream on data event inside stream on response (at the and of the callback) and it seems to be ok now

```javascript
 stream.once('response', headers => {
    const { [HTTP2_HEADER_STATUS]: statusCode, ...realHeaders } = headers
    request.onResponseStarted()

    if (request.onHeaders(Number(statusCode), realHeaders, stream.resume.bind(stream), '') === false) {
      stream.pause()
    }

    stream.on('data', (chunk) => {
      if (request.onData(chunk) === false) {
        stream.pause()
      }
    })
  })
``` 

my test script is:
```javascript
const Agent = require("./lib/dispatcher/agent");
const { fetch } = require("./lib/web/fetch");

const dispatcher = new Agent({ allowH2: true });

(async () => {
  while (true) {
    //console.log("fetching 1000 requests");
    //console.time();
    await Promise.all(
      Array.from({ length: 1000 }).map(async () => {
        try {
          const response = await fetch("https://http2.pro/api/v1", {
            dispatcher,
          });
          const text = await response.text();
          //console.log(text);
          if(!text) throw new Error("NOT GOOD")
        } catch (e) {
          console.error(e);
        }
      }),
    );
    //console.timeEnd();
  }
})();
``` 

This make more sense to you?
Thanks

ronag: Make sense, but yea, please open an issue because this seems wrong.

St3ffGv4: Issue opened

ronag (approved):

lib/dispatcher/client-h2.js
```diff
@@ -394,6 +394,12 @@ function writeH2 (client, request) {
     if (request.onHeaders(Number(statusCode), realHeaders, stream.resume.bind(stream), '') === false) {
       stream.pause()
     }
+
+    stream.on('data', (chunk) => {
```
metcoder95, line 398: Can we add the link to the issue as comment?

metcoder95 (approved):

mcollina (approved):
lgtm

PR #2991, opened 2024-03-25
docs: ✏️ fixed benchmark links

benhalverson:
## Changes
Fixed the broken links for benchmarks.js and post-benchmark.js

- [ ] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@ npm i undici
 
 ## Benchmarks
 
-The benchmark is a simple getting data [example](benchmarks/benchmark.js) using a
+The benchmark is a simple getting data [example](https://github.com/nodejs/undici/blob/main/benchmarks/benchmark.js) using a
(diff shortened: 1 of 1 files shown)


KhafraDev:
can you use relative links?

benhalverson:
> can you use relative links?

I don't think so because benchmark.js and post-benchmark.js are outside the docs directory. 
I tried `[example](../benchmarks/benchmark.js)` and the original code was `[example](.benchmarks/benchmark.js)` both return 404
I also tried this  https://docsify.js.org/#/helpers?id=ignore-to-compile-link

it might be possible with a plugin for docsify but I've never used this package and nothing sticks out in this plugin list.
https://docsify.js.org/#/awesome?id=plugins

If benchmark.js was in the docs directory would you want to load just the js or would it be in a markdown file like the existing examples?

metcoder95 (approved):

PR #2994, opened 2024-03-25
Added EnvHttpProxyAgent to support HTTP_PROXY

### LINKED ISSUE #1650: Support HTTP_PROXY and NO_PROXY env var
A somewhat standard env var used in proxying applications is the `HTTP_PROXY` and `NO_PROXY` variables.

Undici should support automatically setting a global dispatcher if HTTP_PROXY env var is used. Currently Undici supports using `setGlobalDispatcher()` method to set a specific proxy endpoint.

However there is a sister environment variable to HTTP_PROXY that supports contextually non-proxying domains: `NO_PROXY`

NO_PROXY requires a list of domain names to be NOT proxied and should make a request normally.

If any of the above cannot be supported, then Undici (and ultimately Node's Fetch-Undici) should allow per-request proxy settings or function hook to allow anyone to develop the same effect as HTTP_PROXY and NO_PROXY env var in their application that uses Undici/Node Fetch.

Thoughts?

10xLaCroixDrinker:
## This relates to...

Closes #1650

## Changes

Added EnvHttpProxyAgent and made it the default global dispatcher. The addition to index-fetch.js was required to keep those tests passing after changing the global dispatcher to a ProxyAgent

### Features

Automatically detect/support HTTP_PROXY, HTTPS_PROXY, & NO_PROXY

### Bug Fixes

N/A

Using these env vars by default could result in unexpected behavior

## Status

- [x] Tested
- [x] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready


lib/dispatcher/env-http-proxy-agent.js
```diff
@@ -0,0 +1,108 @@
...
+  #shouldProxy (hostname, port) {
+    const NO_PROXY = process.env.NO_PROXY || process.env.no_proxy
+    if (!NO_PROXY) {
+      return true // Always proxy if NO_PROXY is not set.
+    }
+    if (NO_PROXY === '*') {
+      return false // Never proxy if wildcard is set.
+    }
+
+    return NO_PROXY.split(/[,\s]/).filter((entry) => !!entry.length).every(function (entry) {
```
Uzlopak, line 85: Why would you always split NO_PROXY every time. You should pre compute it.

metcoder95 (commented):
Why not just detect the env var and set a `ProxyAgent` as `globalDispatcher`? ref: https://github.com/nodejs/undici/blob/ef25209858817e697fd7607850413357c2b3b7e2/lib/global.js#L10

I'm not really in sync with the addition of the new Agent as does not seems needed to enable this feature, we can keep it simple and just use the `ProxyAgent` if the feature is enabled or disabled otherwise.

Given the nature of the feature, I believe this might qualify as `major` as can cause unexpected side-effects if there's usage of `HTTP_PROXY` or similar already; unless we provide an option to ignore to force ignore the env var.

10xLaCroixDrinker:
The three separate agents are needed to fully support the HTTP_PROXY, HTTPS_PROXY, NO_PROXY use case. 

When only HTTP_PROXY is set, all requests go through it. When HTTPS_PROXY is set, https requests go through that one. And when NO_PROXY is set, requests that match are not proxied.

Creating a new agent is what appeared agreed upon in #1650

https://github.com/nodejs/undici/issues/1650#issuecomment-1348120245

KhafraDev:
I think it shouldn't be set as the default agent, making it a semver-minor change.

lib/dispatcher/env-http-proxy-agent.js
KhafraDev: env vars can be mutated.

```js
await fetch('...')
process.env.NO_PROXY = '*'
await fetch('...')
```

For performance reasons it should be possible to compare old/new values and only parse when a new one is set.

KhafraDev: this should also probably just be a for loop instead of .filter().every()

KhafraDev: I don't see a reason that it shouldn't update when changed

index-fetch.js
```diff
@@ -1,5 +1,9 @@
 'use strict'
 
+const Dispatcher = require('./lib/dispatcher/dispatcher')
+const api = require('./lib/api')
+Object.assign(Dispatcher.prototype, api)
+
```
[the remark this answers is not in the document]
10xLaCroixDrinker: It's required if it's the default dispatcher. Any ProxyAgent as the default dispatcher causes many tests to fail without this.

I'll go ahead and remove though since it won't be default.

mcollina:
I would prefer just adding this first, then we can turn it on by default. It seems the safer option.

lib/dispatcher/env-http-proxy-agent.js
```diff
@@ -0,0 +1,116 @@
...
+
+  #shouldProxy (hostname, port) {
+    if (this.#alwaysProxy) {
+      return true // Always proxy if NO_PROXY is not set or empty.
+    }
+    if (this.#neverProxy) {
+      return false // Never proxy if wildcard is set.
+    }
+
+    return this.#noProxyEntries.every(function (entry) {
```
mcollina, line 101: this is a hot path, please avoid these helper methos and just use a for loop.

lib/dispatcher/env-http-proxy-agent.js
```diff
@@ -0,0 +1,116 @@
...
+  #neverProxy = false
+  #alwaysProxy = false
+  #noProxyEntries = []
+
+  constructor (opts) {
+    super()
+
+    this[kNoProxyAgent] = new Agent(opts)
+
+    const HTTP_PROXY = process.env.HTTP_PROXY || process.env.http_proxy
```
mcollina, line 23: it would be good to also allow an option

lib/dispatcher/env-http-proxy-agent.js
```diff
@@ -0,0 +1,116 @@
...
+    if (HTTP_PROXY) {
+      this[kHttpProxyAgent] = new ProxyAgent({ ...opts, uri: HTTP_PROXY })
+    } else {
+      this[kHttpProxyAgent] = this[kNoProxyAgent]
+    }
+
+    const HTTPS_PROXY = process.env.HTTPS_PROXY || process.env.https_proxy
```
mcollina, line 30: it would be good to also allow an option

lib/dispatcher/env-http-proxy-agent.js
```diff
@@ -0,0 +1,116 @@
...
+    if (HTTPS_PROXY) {
+      this[kHttpsProxyAgent] = new ProxyAgent({ ...opts, uri: HTTPS_PROXY })
+    } else {
+      this[kHttpsProxyAgent] = this[kHttpProxyAgent]
+    }
+
+    const NO_PROXY = process.env.NO_PROXY || process.env.no_proxy || ''
```
mcollina, line 37: same

mcollina (changes requested):
Good job! Could you also add some docs?

10xLaCroixDrinker:
I've just got docs and refactoring tests away from mocks remaining. I believe all the other feedback is in now.

metcoder95 (commented):
Let's not forget about the `types` and documentation

mcollina:
@10xLaCroixDrinker I think it would be great to ship this.

10xLaCroixDrinker:
I can push those up today or tomorrow, but I haven't had time to work on rewriting the tests.

lib/dispatcher/env-http-proxy-agent.js
```diff
@@ -0,0 +1,151 @@
...
+  'http:': 80,
+  'https:': 443
+}
+
+class EnvHttpProxyAgent extends DispatcherBase {
+  #noProxyValue = null
+  #noProxyEntries = null
+  #opts = null
+
+  constructor (opts = {}) {
```
KhafraDev, line 18: It should be marked as experimental

KhafraDev (changes requested):
I don't necessarily like the name since it also handles https proxies, but I don't have any alternative suggestions. Good work on this, I never got around to implementing it. 😄

lib/dispatcher/env-http-proxy-agent.js
10xLaCroixDrinker: Like so?

```diff
+let experimentalWarned = false
+
 class EnvHttpProxyAgent extends DispatcherBase {
   #noProxyValue = null
   #noProxyEntries = null
@@ -19,6 +21,13 @@ class EnvHttpProxyAgent extends DispatcherBase {
     super()
     this.#opts = opts

+    if (!experimentalWarned) {
+      experimentalWarned = true
+      process.emitWarning('EnvHttpProxyAgent is experimental, expect them to change at any time.', {
+        code: 'UNDICI-EHPA'
+      })
+    }
```

metcoder95 (approved):
LGTM, nice 👍

mcollina (approved):
Lgtm

mcollina:
The tests are failing on Windows

mcollina:
@KhafraDev are you blocking this somehow? Can it land?

KhafraDev (approved):

Uzlopak (approved):

docs/docs/api/EnvHttpProxyAgent.md
```diff
@@ -0,0 +1,162 @@
...
+
+Stability: Experimental.
+
+Extends: `undici.Dispatcher`
+
+EnvHttpProxyAgent automatically reads the proxy configuration from the environment variables `HTTP_PROXY`, `HTTPS_PROXY`, and `NO_PROXY` and sets up the proxy agents accordingly. When `HTTP_PROXY` and `HTTPS_PROXY` are set, `HTTP_PROXY` is used for HTTP requests and `HTTPS_PROXY` is used for HTTPS requests. If only `HTTP_PROXY` is set, `HTTP_PROXY` is used for both HTTP and HTTPS requests. If only `HTTPS_PROXY` is set, it is only used for HTTPS requests.
+
+`NO_PROXY` is a comma or space-separated list of hostnames that should not be proxied. The list may contain leading wildcard characters (`*`). If `NO_PROXY` is set, the EnvHttpProxyAgent will bypass the proxy for requests to hosts that match the list. If `NO_PROXY` is set to `"*"`, the EnvHttpProxyAgent will bypass the proxy for all requests.
+
+Lower case environment variables are also supported: `http_proxy`, `https_proxy`, and `no_proxy`. However, if both the lower case and upper case environment variables are set, the lower case environment variables will be ignored.
```
silverwind, line 11: I'm late to the party, but I strongly recommend to reconsider order preference this because as per the [proxy var "spec"](https://about.gitlab.com/blog/2021/01/27/we-need-to-talk-no-proxy/#http_proxy-and-https_proxy-1), lowercase form should be preferred:

> Use lowercase form. HTTP_PROXY is not always supported or recommended.
> If you absolutely must use the uppercase form as well, be sure they share the same value.

metcoder95: cc @10xLaCroixDrinker

10xLaCroixDrinker: I can make the change some time in the next couple days.

10xLaCroixDrinker: Done: #3152

PR #2999, opened 2024-03-27
Revert "perf: reuse TextDecoder instance (#2863)"

panva:
This reverts commit 2091dcfa1a9864dab7675be7c6bff34dbde06051.

See
- https://github.com/nodejs/node/pull/52175#issuecomment-2020665559
- https://github.com/nodejs/undici/pull/2863#issuecomment-2020746531

### DIFF
--- a/lib/web/websocket/receiver.js
+++ b/lib/web/websocket/receiver.js
@@ -12,8 +12,6 @@ const { WebsocketFrameSend } = require('./frame')
 // Copyright (c) 2013 Arnout Kazemier and contributors
 // Copyright (c) 2016 Luigi Pinca and contributors
 
-const textDecoder = new TextDecoder('utf-8', { fatal: true })
-
 class ByteParser extends Writable {
   #buffers = []
   #byteOffset = 0
@@ -316,7 +314,8 @@ class ByteParser extends Writable {
     }
 
     try {
-      reason = textDecoder.decode(reason)
+      // TODO: optimize this
+      reason = new TextDecoder('utf-8', { fatal: true }).decode(reason)
     } catch {
       return null
     }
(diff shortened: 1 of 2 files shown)


ronag (approved):

ronag (changes requested):
ACtually doesn't this just move the issue to when websocket is used? Should we maybe not pass `fatal`? Does the spec require that? If it does, then maybe we should disable websocket for builds without ICU?

panva:
> ACtually doesn't this just move the issue to when websocket is used? Should we maybe not pass `fatal`? Does the spec require that? If it does, then maybe we should disable websocket for builds without ICU?

I don't particularly care how this moves forward but the revert is needed to unblock a Node.js version bump.

Uzlopak (changes requested):
It actually revealed that there is an issue with no-icu cases.

panva:
> It actually revealed that there is an issue with no-icu cases.

Which, now that we know about it, we can take all the time in the world to contemplate a fix for. In the meantime, node's bundled fetch is lagging behind on fixes which is why this revert is proposed.

metcoder95 (approved):

Uzlopak:
I would atleast expect that you open an issue here, so that it keeps tracked.

Uzlopak (approved):
Approving but somebody should create an issue.

panva:
@mcollina can you do a release? Node nightly is currently lagging on 6.6.2

ronag:
6.10.2 published

KhafraDev:
> Does the spec require that?

yes, we're meant to close the connection if we receive invalid utf-8

> If it does, then maybe we should disable websocket for builds without ICU?

> It actually revealed that there is an issue with no-icu cases.

It's already essentially disabled:
- Open a WebSocket connection (on a build without icu)
- The connection closes if it receives a message.
- The close frame has no reason if we receive a close frame.

We might be able to support builds without icu by using [Buffer.isUTF8](https://nodejs.org/api/buffer.html#bufferisutf8input), but it should only be used when we can't use TextDecoder for perf. I'm not really sure if it's available in builds without ICU and I don't have the time to test myself.

PR #2938, opened 2024-03-09
fetch: improve util.inspect output for web specifications

mertcanaltin:
Issue: web specs better util.inspect output #2917

Issue Description:
The issue revolves around enhancing the util.inspect output for the Headers class. The current output exposes too many internal implementation details and does not provide much useful information. 

Solution:
I have improved the util.inspect output to better reflect the contents of the Headers object for enhanced clarity.

PR #3007, opened 2024-03-28
fix: node:util instead of util

mertcanaltin:
@anonrig after your suggestion, I updated these
https://github.com/nodejs/undici/pull/2938#discussion_r1529551347

### DIFF
--- a/test/fetch/request-inspect-custom.js
+++ b/test/fetch/request-inspect-custom.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const { describe, it } = require('node:test')
-const assert = require('assert')
+const assert = require('node:assert')
 const util = require('node:util')
 const { Request } = require('../../')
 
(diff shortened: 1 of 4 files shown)


ronag (approved):

test/fetch/request-inspect-custom.js
```diff
@@ -2,7 +2,7 @@
 
 const { describe, it } = require('node:test')
 const assert = require('assert')
```
Uzlopak, line 4:
```suggestion
const assert = require('node:assert')
```

test/fetch/response-inspect-custom.js
```diff
@@ -2,7 +2,7 @@
 
 const { describe, it } = require('node:test')
 const assert = require('assert')
```
Uzlopak, line 4:
```suggestion
const assert = require('node:assert')
```

mertcanaltin: thankyou

Uzlopak (approved):
Lgtm

Issue #3002, opened 2024-03-27
Undici cannot be required in --without-intl Node.js


panva:
## Bug Description

Undici cannot be required in --without-intl Node.js

## Reproducible By

```
# in nodejs/node
./configure --node-builtin-modules-path $(pwd) --without-intl --ninja
make

# in a project with undici
cat package.json | jq '.dependencies.undici'
"^6.10.1"
../node/node -e "require('undici')"

node:internal/encoding:487
          throw new ERR_NO_ICU('"fatal" option');
          ^

TypeError [ERR_NO_ICU]: "fatal" option is not supported on Node.js compiled without ICU
    at new TextDecoder (node:internal/encoding:487:17)
    at Object.<anonymous> (/repo/node_modules/undici/lib/web/websocket/util.js:71:21)
    at Module._compile (node:internal/modules/cjs/loader:1421:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1499:10)
    at Module.load (node:internal/modules/cjs/loader:1232:32)
    at Module._load (node:internal/modules/cjs/loader:1048:12)
    at Module.require (node:internal/modules/cjs/loader:1257:19)
    at require (node:internal/modules/helpers:188:18)
    at Object.<anonymous> (/repo/node_modules/undici/lib/web/websocket/websocket.js:24:5)
    at Module._compile (node:internal/modules/cjs/loader:1421:14) {
  code: 'ERR_NO_ICU'
}
```

## Expected Behavior

Undici CI also tests on Node.js without ICU so that releases that cannot be merged to Node.js don't happen

## See

- https://github.com/nodejs/node/pull/52175#issuecomment-2020665559
- https://github.com/nodejs/undici/pull/2863#issuecomment-2020746531

KhafraDev:
this is a recurring issue, we should run tests on builds without icu

cc @mweberxyz @Uzlopak

mweberxyz:
👀 

@KhafraDev Do you happen to know of any other repos within nodejs/ building binaries `--without-intl`? I can work on a workflow that builds (and caches) a non-icu build just for undici's use, but better to not have to reinvent the wheel if it is already being done.

KhafraDev:
I don't sorry, I thought it'd be similar to the nightly workflow 😅

mweberxyz:
I also asked around in the #nodejs-build channel in OpenJS foundation. My intuition is no, I am going to have to build Node in this repo's actions.

[...]

Issue #3012, opened 2024-03-28
OpenSSF scorecard "Token-permissions" score zero


fraxken:
Hello there :wave:,

Today in the Node.js security meeting we discussed that the undici project score had dropped due to a change in the workflows. This seems to be related to the following [commit](https://github.com/nodejs/undici/commit/3be7ec9bcc4679a827199a2aea3f0f1bfbdaf325).

![image](https://github.com/nodejs/undici/assets/4438263/ec1d6e60-4240-44f9-bfc2-8bd993b10aad)

[Full report here](https://kooltheba.github.io/openssf-scorecard-api-visualizer/#/projects/github.com/nodejs/undici/compare/dae31826560112ccf1244172e398fc7d2fce075d/7485cd9b4cf9a86cb76b1597df527eba15755bfc)

1. [nightly.yml](https://github.com/nodejs/undici/blob/main/.github/workflows/nightly.yml) and [test.yml](https://github.com/nodejs/undici/blob/main/.github/workflows/test.yml) seems to lack global permissions

```yml
permissions:
  contents: read
```

2. In **nodejs.yml** we have warnings to reviews the following permission.

https://github.com/nodejs/undici/blob/d3d24e2722046733f4f78a273423f10d0107e74f/.github/workflows/nodejs.yml#L102-L105

[Github Action Merge Dependabot](https://github.com/fastify/github-action-merge-dependabot?tab=readme-ov-file#usage) require `pull-requests` and `contents` (not sure if `actions` is required). These warnings are **perhaps to be ignored**.

@nodejs/security-wg

metcoder95:
#3013 closes this issue?

fraxken:
> #3013 closes this issue?

Not yet. I'm working on nodejs workflow. If everything ok I will close myself (else I will open another PR).

mweberxyz:
My bad on that change. The GHA documentation is a bit bare on permissions, by not setting the global permissions does it essentially issue a full read-write token to the workflow?

Also just to confirm: Are job-level permissions additive to workflow-level?

For example, in nightly at the workflow level we now have:
```
permissions:
  contents: read
```
then the report-failure job within that workflow:
```
  report-failure:
    permissions:
      issues: write
``` 
So there's no need to explicitly set `contents: read` in the report-failure?

fraxken:
> My bad on that change. The GHA documentation is a bit bare on permissions, by not setting the global permissions does it essentially issue a full read-write token to the workflow?

[...]

PR #2728, opened 2024-02-09
remove FileReader

KhafraDev:

PR #3015, opened 2024-03-29
chore: add automated CI testing with —no-intl node

mweberxyz:
## This relates to...

- https://github.com/nodejs/undici/issues/3002
- https://github.com/nodejs/undici/pull/2999
- https://github.com/nodejs/node/pull/52175

## Rationale

This PR extends the Node CI workflow to run undici tests against Node 20 and 21 built with `--without-intl` flag, which would have prevented the regression fixed by https://github.com/nodejs/undici/pull/2999 before the commit even merged to undici/main.

The full test suite currently failing against `main`, so problematic tests have been fixed where possible, otherwise excluded from running in the w/o intl build. If possible, these problematic tests/functionality should be fixed to run in w/o intl builds and the tests added to `test:javascript:withoutintl` (or even remove it entirely and both the w/o intl and regular builds can use `test:javascript`).

Note that the initial run, and occasional runs thereafter, of the workflow will take a long time (around 1 hour for the clean Node 21 build) -- but subsequent runs will be much faster due to ccache caching.

## Changes
- Workflows:
  - add `test-without-intl` job to workflows/nodejs.yml
- package.json:
  - add `test:javascript:withoutintl` script to run tests known good with an w/o intl node build
  - add `test:eventsource:nobuild`, `test:fetch:nobuild` scripts to pull test running logic out of `test:eventsource` and `test:fetch` because undici is built _prior_ to running w/o intl tests
  - add `test:wpt:withoutintl` script to run wpt tests except those problematic for w/o intl, namely:
    - `test/wpt/start-websockets.mjs`, and
    - `test/wpt/start-FileAPI.mjs`
 - various test tweaks to keep existing tests running without disabling for w/o intl

### Features

N/A

### Bug Fixes

N/A

N/A

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready


mweberxyz:
I started looking into the actual test failures, the problems might not be all that arduous to fix and get the tests passing.

First, there's a failing test which is simply a misaligned console.table assertion.

Then, some tests run `npm run build:node` which fails, but maybe that doesn't maybe doesn't need to run in this configuration because the package can be built with a regular node?

I've gotten down into the `test:wpt` which is failing on steps that are clearly unicode related:
CleanShot 2024-03-28 at 21 56 56@2x

Will update with more findings and if I get the tests to fully pass.

KhafraDev:
that failure will be ignored once #2728 lands cc @ronag @mcollina

mweberxyz:
With most recent commits, w/o intl build tests are green, PR description has been updated, and ready for review.

As a example, I reintroduced the commit which caused nodejs CI to fail its w/o intl build, and the check failed as expected -- see https://github.com/mweberxyz/nodejs-undici/pull/6

test/node-test/debug.js
```diff
@@ -9,6 +9,9 @@ const { tspl } = require('@matteo.collina/tspl')
 const removeEscapeColorsRE = /[\u001b\u009b][[()#;?]*(?:[0-9]{1,4}(?:;[0-9]{0,4})*)?[0-9A-ORZcf-nqry=><]/g
 
 test('debug#websocket', async t => {
+  if (!process.versions.icu) {
+    return
+  }
```
metcoder95, line 14:
```suggestion
test('debug#websocket', { skip: !process.versions.icu }, async t => {
```

lib/mock/pending-interceptors-formatter.js
```diff
@@ -7,6 +7,9 @@ const { Console } = require('node:console')
  * Gets the output of `console.table(…)` as a string.
  */
 module.exports = class PendingInterceptorsFormatter {
+  static PERSISTENT = process.versions.icu ? '✅' : 'Y '
+  static NOT_PERSISTENT = process.versions.icu ? '❌' : 'N '
+
```
Uzlopak, line 12: Do we need to add these static properties?
Cant we define them outside the class?

test/mock-interceptor-unused-assertions.js
```diff
@@ -10,6 +10,10 @@ const util = require('../lib/core/util')
 // https://github.com/nodejs/node/pull/50135
 const tableRowsAlignedToLeft = util.nodeMajor >= 21 || (util.nodeMajor === 20 && util.nodeMinor >= 11)
 
+// `console.table` treats emoji as two character widths for cell width determination
+const Y = PendingInterceptorsFormatter.PERSISTENT
+const N = PendingInterceptorsFormatter.NOT_PERSISTENT
```
Uzlopak, line 15:
```suggestion
const Y = process.versions.icu ? '✅' : 'Y '
const N = process.versions.icu ? '❌' : 'N '
```

package.json
```diff
@@ -69,10 +69,13 @@
     "lint:fix": "standard --fix | snazzy",
     "test": "npm run test:javascript && cross-env NODE_V8_COVERAGE=  npm run test:typescript",
     "test:javascript": "node scripts/generate-pem && npm run test:unit && npm run test:node-fetch && npm run test:fetch && npm run test:cookies && npm run test:eventsource && npm run test:wpt && npm run test:websocket && npm run test:node-test && npm run test:jest",
+    "test:javascript:withoutintl": "node scripts/generate-pem && npm run test:unit && npm run test:node-fetch && npm run test:fetch:nobuild && npm run test:cookies && npm run test:eventsource:nobuild && npm run test:wpt:withoutintl && npm run test:node-test",
     "test:cookies": "borp -p \"test/cookie/*.js\"",
     "test:node-fetch": "borp -p \"test/node-fetch/**/*.js\"",
-    "test:eventsource": "npm run build:node && borp --expose-gc -p \"test/eventsource/*.js\"",
-    "test:fetch": "npm run build:node && borp --expose-gc -p \"test/fetch/*.js\" && borp -p \"test/webidl/*.js\" && borp -p \"test/busboy/*.js\"",
+    "test:eventsource": "npm run build:node && npm run test:eventsource:nobuild",
```
Uzlopak, line 75: why is building bad?

mweberxyz: Build fails with icu-less node build:

CleanShot 2024-03-29 at 14 25 05@2x

But we shouldn't to support _building_ undici with icu-less node, only running it.

This is why the workflow [installs regular node then builds undici](https://github.com/nodejs/undici/pull/3015/files#diff-e98936aa52a6dd7416e4296e9628456227d834f7245967383fd9ff80fd985dadR88) prior to switching to the icu-less build for running tests.

I tried to minimize changes to package.json, so as to make people not need to think about keeping the `test:javascript:withoutintl` script up-to-date with test strategy changes.

lib/mock/pending-interceptors-formatter.js
mweberxyz: I imported them from test file, and given that the default export is a class, static members was cleanest. Seeing your next comment, I'll pull them up out of the class here, and re-define in test.

mweberxyz:
One other omission I realize I haven't mentioned: `jest` doesn't work with icu-less node:

CleanShot 2024-03-29 at 14 55 14@2x

mweberxyz:
@fraxken Does anything in here conflict with your auditing re: #3012? I tried to be more mindful of security implications this time around (ie- I reviewed the code for [this newly called action](https://github.com/nodejs/undici/pull/3015/files#diff-e98936aa52a6dd7416e4296e9628456227d834f7245967383fd9ff80fd985dadR118) and made sure to pin it), but would appreciate an extra set of eyes.

fraxken:
@mweberxyz seems ok to me 😉

mcollina (approved):
lgtm

mcollina:
Could this be landed?

metcoder95 (approved):

PR #3024, opened 2024-03-31
fix: don't leak internal class

ronag:
Don't pass the internal request object, create a "safe" representation. Also don't allow external mutation through `addHeaders`.

### DIFF
--- a/types/diagnostics-channel.d.ts
+++ b/types/diagnostics-channel.d.ts
@@ -9,8 +9,7 @@ declare namespace DiagnosticsChannel {
     completed: boolean;
     method?: Dispatcher.HttpMethod;
     path: string;
-    headers: string;
-    addHeader(key: string, value: string): Request;
+    headers: Array<Buffer>;
   }
   interface Response {
     statusCode: number;
--- a/types/diagnostics-channel.d.ts
+++ b/types/diagnostics-channel.d.ts
@@ -9,7 +9,7 @@ declare namespace DiagnosticsChannel {
     completed: boolean;
     method?: Dispatcher.HttpMethod;
     path: string;
-    headers: string;
+    headers: any;
   }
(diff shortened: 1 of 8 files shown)


ronag:
The messages here are weird, `:headers` isn't actually headers it's response

metcoder95:
Shall we point to `next`?

types/diagnostics-channel.d.ts
```diff
@@ -9,8 +9,7 @@ declare namespace DiagnosticsChannel {
     completed: boolean;
     method?: Dispatcher.HttpMethod;
     path: string;
-    headers: string;
-    addHeader(key: string, value: string): Request;
+    headers: Array<Buffer>;
```
ronag, line 12:
```suggestion
    headers: string;
```

types/diagnostics-channel.d.ts
```diff
@@ -9,7 +9,7 @@ declare namespace DiagnosticsChannel {
     completed: boolean;
     method?: Dispatcher.HttpMethod;
     path: string;
-    headers: any;
+    headers: string;
```
ronag, line 12:
```suggestion
    headers: any;
```

mcollina (changes requested):
The goal is to leak the internal class, so it could be used as key in maps etc.

ronag:
That's a bad idea in my opinion and can lead to all kinds of undefined behaviors and breaking. The internal class has no guarantees to its interface.

ronag:
> The goal is to leak the internal class, so it could be used as key in maps etc.

I don't see how this PR would change it. You can still use the passed object as a key.

mcollina (approved):
lgtm

I missed that

metcoder95 (approved):

PR #3028, opened 2024-03-31
feat: onConnect w/ controller

ronag:
Part of a larger overhauld of hooks. This on is not breaking and can therefore land earlier.

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready


test/node-test/client-dispatch.js
```diff
@@ -938,8 +939,9 @@ test('dispatches in expected order for http2', async (t) => {
       onBodySent () {
         dispatches.push('onBodySent')
       },
-      onResponseStarted () {
-        dispatches.push('onResponseStarted')
```
mcollina, line 942: Can you add a test that keeps showing this behavior? Or add a new test for onStart?

mcollina (changes requested):

lib/core/request.js
```diff
@@ -251,8 +251,23 @@ class Request {
     }
   }
 
-  onResponseStarted () {
-    return this[kHandler].onResponseStarted?.()
+  onStart (resume) {
+    try {
+      let pause = false
+
+      // Compatibility with the old API.
```
metcoder95, line 258: Shall we add it to the major checklist?

metcoder95 (approved):
LGTM, shall we document it and add types for it?

lib/core/request.js
ronag: Already on the list to remove.

ronag:
Updated according to feedback from @BridgeAR

PR #3031, opened 2024-03-31
feat: add onField and onStatus hooks

ronag:
## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -231,6 +231,32 @@ class Request {
     return this[kHandler].onResponseStarted?.()
   }
 
+  onField (name, value) {
+    assert(!this.aborted)
+    assert(!this.completed)
+
+    if (this[kHandler].onField) {
+      try {
+        return this[kHandler].onField(name, value)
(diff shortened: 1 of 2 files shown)


lib/core/request.js
```diff
@@ -231,6 +231,32 @@ class Request {
     return this[kHandler].onResponseStarted?.()
   }
 
+  onField (name, value) {
+    assert(!this.aborted)
+    assert(!this.completed)
+
+    if (this[kHandler].onField) {
+      try {
+        return this[kHandler].onField(name, value)
```
metcoder95, line 240: Shouldn't this be equivalent, or are you looking to avoid create the `try/catch` scope?

```suggestion
        return this[kHandler].onField?.(name, value)
```

ronag: Avoiding the try catch

ronag:
Not sure if this is a good idea.

PR #3054, opened 2024-04-04
feat: new hooks API

ronag:
While this is not breaking per se. It's pretty much unusable with existing handlers...

### DIFF
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -2,9 +2,11 @@
 
 const {
   InvalidArgumentError,
-  NotSupportedError
+  NotSupportedError,
+  AbortError
 } = require('./errors')
 const assert = require('node:assert')
+const { parseHeaders } = require('./util')
 const {
   isValidHTTPToken,
   isValidHeaderChar,
@@ -25,6 +27,44 @@ const { headerNameLowerCasedRecord } = require('./constants')
 const invalidPathRegex = /[^\u0021-\u00ff]/
 
 const kHandler = Symbol('handler')
+
+class Controller {
+  #paused = false
+  #reason = null
+  #abort
+  #resume
+
+  constructor (abort, resume) {
+    this.#abort = abort
+    this.#resume = resume
+  }
+
+  pause () {
+    this.#paused = true
+  }
+
+  resume () {
+    this.#paused = false
+    this.#resume?.()
+  }
+
+  abort (reason) {
+    this.#reason = reason ?? new AbortError()
+    this.#abort?.(this.#reason)
+  }
+
+  get paused () {
+    return this.#paused
+  }
+
+  get aborted () {
+    return this.#reason !== null
+  }
+
+  get reason () {
+    return this.#reason
+  }
+}
 
 class Request {
   constructor (origin, {
@@ -90,6 +130,8 @@ class Request {
     this.method = method
 
     this.abort = null
+
+    this.controller = null
 
     if (body == null) {
       this.body = null
@@ -192,12 +234,11 @@ class Request {
   }
 
   onBodySent (chunk) {
-    if (this[kHandler].onBodySent) {
-      try {
-        return this[kHandler].onBodySent(chunk)
-      } catch (err) {
-        this.abort(err)
-      }
+    try {
+      this[kHandler].onRequestData?.(chunk)
+      this[kHandler].onBodySent?.(chunk)
+    } catch (err) {
+      this.abort(err)
     }
   }
 
@@ -206,41 +247,61 @@ class Request {
       channels.bodySent.publish({ request: this })
     }
 
-    if (this[kHandler].onRequestSent) {
-      try {
-        return this[kHandler].onRequestSent()
-      } catch (err) {
-        this.abort(err)
-      }
+    try {
+      this[kHandler].onRequestEnd?.()
+      this[kHandler].onRequestSent?.()
+    } catch (err) {
+      this.abort(err)
     }
   }
 
   onConnect (abort) {
     assert(!this.aborted)
     assert(!this.completed)
+    assert(!this.controller)
 
     if (this.error) {
       abort(this.error)
     } else {
       this.abort = abort
-      return this[kHandler].onConnect(abort)
+      return this[kHandler].onConnect?.(abort)
     }
   }
 
-  onResponseStarted () {
-    return this[kHandler].onResponseStarted?.()
-  }
-
-  onHeaders (statusCode, headers, resume, statusText) {
+  onResponseStarted (resume) {
     assert(!this.aborted)
     assert(!this.completed)
+
+    this.controller = new Controller(this.abort, resume)
+    try {
+      this[kHandler].onResponseStart?.(this.controller)
+
+      this[kHandler].onResponseStarted?.()
+
+      return !this.controller.paused
+    } catch (err) {
+      this.abort(err)
+    }
+  }
+
+  onHeaders (headers, statusCode, statusText) {
+    assert(!this.aborted)
+    assert(!this.completed)
+    assert(this.controller)
 
     if (channels.headers.hasSubscribers) {
       channels.headers.publish({ request: this, response: { statusCode, headers, statusText } })
     }
 
     try {
-      return this[kHandler].onHeaders(statusCode, headers, resume, statusText)
+      this[kHandler].onResponseHeaders?.(parseHeaders(headers), statusCode, statusText)
+
+      const controller = this.controller
+      if (this[kHandler].onHeaders?.(statusCode, headers, () => controller.resume(), statusText || '') === false) {
+        this.controller.pause()
+      }
+
+      return !this.controller.paused
     } catch (err) {
       this.abort(err)
     }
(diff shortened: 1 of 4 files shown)


types/dispatcher.d.ts
```diff
@@ -232,6 +241,28 @@ declare namespace Dispatcher {
...
+    onRequestStart?(/* controller: Controller */): void;
+    /** Invoked after headers data is sent */
+    // onRequestHeaders?(headers: Record<string, string>): void;
+    /** Invoked after payload data is sent. */
+    onRequestData?(chunk: Buffer | string): void;
+    /** Invoked after request has finished sending */
+    onRequestEnd?(): void;
+
+    /** Invoked after response is starting to be processed */
+    onResponseStart?(controller: Controller): void;
```
ShogunPanda, line 257: I think you're not passing a controller here.
Also, in the `onRequestStart` above, you don't pass it so I would remove the comment at all.

ShogunPanda (approved):
One nit, rest LGTM!

lib/core/request.js
```diff
@@ -206,41 +247,63 @@ class Request {
...
     assert(!this.completed)
+    assert(this.controller)
 
     if (channels.headers.hasSubscribers) {
       channels.headers.publish({ request: this, response: { statusCode, headers, statusText } })
     }
 
     try {
-      return this[kHandler].onHeaders(statusCode, headers, resume, statusText)
+      this[kHandler].onResponseHeaders?.(parseHeaders(headers), statusCode, statusText)
```
ronag, line 299: Here we are no longer paso g raw headers. We parse them immediately. As far as I can tell this is what most users do and it's unfortunate to force middleware to work on raw headers.

lib/core/request.js
```diff
@@ -206,41 +247,63 @@ class Request {
...
+    assert(!this.aborted)
+    assert(!this.completed)
+
+    this.controller = new Controller(this.abort, resume)
+    try {
+      this[kHandler].onResponseStart?.(this.controller)
+
+      if (this[kHandler].onResponseStarted?.() === false) {
+        this.controller.pause()
+      }
```
metcoder95, line 281: Question, this is to keep the backwards compatibility, or what's the use case of one or the other?

lib/core/request.js
```diff
@@ -206,41 +247,63 @@ class Request {
...
     try {
-      return this[kHandler].onHeaders(statusCode, headers, resume, statusText)
+      this[kHandler].onResponseHeaders?.(parseHeaders(headers), statusCode, statusText)
+
+      const controller = this.controller
+      if (this[kHandler].onHeaders?.(statusCode, headers, () => controller.resume(), statusText || '') === false) {
+        this.controller.pause()
+      }
+
+      return !this.controller.paused
```
metcoder95, line 306: The `onResponseHeaders` is meant to also return a `boolean`, or how does it decide wether to stop or not the `controller`?

types/dispatcher.d.ts
```diff
@@ -232,6 +241,28 @@ declare namespace Dispatcher {
...
+    /** Invoked after headers data has been received */
+    onResponseHeaders?(headers: Record<string, string>, statusCode: number, statusText?: string): void;
+    /** Invoked after response payload data is received. */
+    onResponseData?(chunk: Buffer | string): void;
```
metcoder95, line 261: Curious, how are we planning for the hook to hint the controller about backpressure, as it seems it does not receives the controller?

metcoder95 (commented):
LGTM, nice work; I can see the potential simplification of coordinating the flow control through the hooks thanks to the controller.

Just have some open questions left.

lib/core/request.js
ronag: Compat

lib/core/request.js
ronag: Not sure I understand the question.

types/dispatcher.d.ts
ronag: I was thinking the user saves a reference to the controller from the start hook. By I guess we could always send in the controller?

types/dispatcher.d.ts
ronag: So we have 2 different controllers, one for the request and one for the response, they may run in parallel.

ronag:
@mcollina @ShogunPanda @metcoder95 

- Should we have separate controllers for request and response or just one global? In theory the request and response can propagate independently hence it might make sense to be able to pause/resume them separately, on the other hand it's not a usual case and maybe just adds unnecessary complexity?
- Should it be possible to pause/resume the request? Or should we always leave it to the server to apply back-pressure? One use case would to be able to throttle the client upload.

lib/core/request.js
metcoder95: nvm, found my the answer, I overlooked the `return` statement 👍

ShogunPanda:
> @mcollina @ShogunPanda @metcoder95
> 
> * Should we have separate controllers for request and response or just one for both or just the response? In theory the request and response can propagate independently hence it might make sense to be able to pause/resume them separately, on the other hand it's not a usual case and maybe just adds unnecessary complexity?

I would keep one controller which acts on both. `pause`, `resume` and `abort` can act on both the request and response and the client can only be in either "sending request" or "receiving response" state at a time so no room for ambiguity.

> * Should it be possible to pause/resume the request? Or should we always leave it to the server to apply back-pressure? One use case would to be able to throttle the client upload.

I definitely would allow to also control the request flow.

types/dispatcher.d.ts
metcoder95: That's fair, we can keep it simple and keep the controller for that

metcoder95:
> > @mcollina @ShogunPanda @metcoder95
> > 
> > * Should we have separate controllers for request and response or just one for both or just the response? In theory the request and response can propagate independently hence it might make sense to be able to pause/resume them separately, on the other hand it's not a usual case and maybe just adds unnecessary complexity?
> 
> I would keep one controller which acts on both. `pause`, `resume` and `abort` can act on both the request and response and the client can only be in either "sending request" or "receiving response" state at a time so no room for ambiguity.

Agree with @ShogunPanda towards a single controller, though I believe that most of the time implementers will look to control the stream flow over the response rather than the request; by providing a controller that allows them to also alter the flow of the request we might enable scenarios where users puts themselves into a weird spot and `undici` not being fully able to do its work accordingly.

The last being said, I'd prefer to keep the controller only for the response and we can evaluate later if there's a need for one over the response.

> > * Should it be possible to pause/resume the request? Or should we always leave it to the server to apply back-pressure? One use case would to be able to throttle the client upload.
> 
> I definitely would allow to also control the request flow.

Couldn't we achieve that already with `Dispatch#compose`?
They can intercept the `dispatch` and throttle the body as they need

ronag:
> the client can only be in either "sending request" or "receiving response" state at a time so no room for ambiguity.

Actually, you can be sending and receiving at the same time. Though I'm not sure how common that is.

> Couldn't we achieve that already with Dispatch#compose? They can intercept the dispatch and throttle the body as they need

Yes, that's also possible. Though less "nice".

@metcoder95 @ShogunPanda So in theory we could totally skip the request hooks (at least for now). However, there is one problem. We need to be able to abort the request **before** the response has been received. That's the only place where we actually need to interact with the request.

i.e.

`onRequestStart(controller: { abort: Function })`

But I guess we can just implement that and leave the rest for the future. Or can you think of a nicer way? Preferably without any `onRequestXXX` hook.

ShogunPanda:
> > the client can only be in either "sending request" or "receiving response" state at a time so no room for ambiguity.
> 
> Actually, you can be sending and receiving at the same time. Though I'm not sure how common that is.
> 

Gosh, you are right. Why do I always forget about pipelining? XD

> > Couldn't we achieve that already with Dispatch#compose? They can intercept the dispatch and throttle the body as they need
> 
> Yes, that's also possible. Though less "nice".
> 
> @metcoder95 @ShogunPanda So in theory we could totally skip the request hooks (at least for now). However, there is one problem. We need to be able to abort the request **before** the response has been received (`onResponseStart`). That's the only place where we actually need to interact with the request.
> 
> i.e.
> 
> `onRequestStart(controller: { abort: Function })`
> 
> But I guess we can just implement that and leave the rest for the future. Or can you think of a nicer way? Preferably without any `onRequestXXX` hook.
> 
> One alternative is to keep `onConnect` (I'd prefer not).

Let's skip it for now. I would just say let's code it in a way we can introduce it later in a semver-minor.

metcoder95:
>@metcoder95 @ShogunPanda So in theory we could totally skip the >request hooks (at least for now). However, there is one problem. We >need to be able to abort the request before the response has been >received (onResponseStart). That's the only place where we actually >need to interact with the request.
>
>i.e.
>
>onRequestStart(controller: { abort: Function })
>
>But I guess we can just implement that and leave the rest for the future.

Agree and in sync with @ShogunPanda, let's do that 👍

lib/core/request.js
ronag: @ShogunPanda @metcoder95 @mcollina any thoughts here? Is this a mistake (i.e. performance blocker) or just being pragmatic?

ShogunPanda: The latter. Let's be pragmatic.

namka279:
8ff1792

lib/core/request.js
metcoder95: +1 on pragmatism

PR #3060, opened 2024-04-07
feat: modernize fuzzing

Uzlopak:
Resolves https://github.com/nodejs/undici/security/code-scanning/25

This PR modernizes the fuzzing code.

1 month ago I proposed changes to jsfuzz, https://gitlab.com/gitlab-org/security-products/analyzers/fuzzers/jsfuzz/-/merge_requests but even though it is managed by gitlab, it seems they can not even review typing only changes (which does not change the code at runtime)
Also latest jsfuzz is on a separate gitlab owned artifact server. the jsfuzz version on npm is obsolete.

So this PR introduces fast-check and removes jsfuzz. fast-check is suggested by the scorecard scan workflow (or whatever you call it).

Fuzzing will now run with every CI run. 
Removing the special fuzzing workflow.
fuzzing will take on local 10 seconds per case (total 30 secs)
fuzzing will take in the CI 1 minute per case (total 3 minutes)

Maybe this will motivate other maintainers to also write more fuzzing tests for other pieces of undici as it is directly integrated and does not need to run it via a special CLI command.


ronag (approved):

metcoder95 (approved):

mcollina (approved):
lgtm

PR #3062, opened 2024-04-07
mock: improve mock interceptor

Uzlopak:
I actually wanted to improve the test coverage. I could have just removed the data field. but actually passing objects is better.

wdyt?

### DIFF
--- a/lib/mock/mock-interceptor.js
+++ b/lib/mock/mock-interceptor.js
@@ -90,7 +90,7 @@ class MockInterceptor {
     this[kContentLength] = false
   }
 
-  createMockScopeDispatchData (statusCode, data, responseOptions = {}) {
+  createMockScopeDispatchData ({ statusCode, data, responseOptions }) {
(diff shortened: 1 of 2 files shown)


lib/mock/mock-interceptor.js
```diff
@@ -99,43 +99,40 @@ class MockInterceptor {
...
       // this function will properly resolve the data from the callback
       // when invoked.
       const wrappedDefaultsCallback = (opts) => {
         // Our reply options callback contains the parameter for statusCode, data and options.
-        const resolvedData = replyData(opts)
+        const resolvedData = replyOptionsCallbackOrStatusCode(opts)
 
         // Check if it is in the right format
-        if (typeof resolvedData !== 'object') {
+        if (typeof resolvedData !== 'object' || resolvedData === null) {
```
Uzlopak, line 126: if we dont do this, than null would result in destructuring issues in the original code

metcoder95 (approved):

PR #3064, opened 2024-04-07
test: client, set body to null if bigger than CHUNK_LIMIT

Uzlopak:


ronag (approved):
I don't quite understand this test

ronag (approved):

metcoder95 (approved):
I believe aims for this part that gets triggered on possible errors: https://github.com/nodejs/undici/blob/5575749dbe310ef6b97bb0984c4b02ba0d427d95/lib/api/util.js#L9

PR #3068, opened 2024-04-07
test: add test for llhttp wasm

Uzlopak:
I wanted to increase the coverage for the wasm files, but without having to run the full test suite with disabled simd fnctionality. 

So it basically I added a sanity check.


mcollina (approved):
lgtm

PR #3079, opened 2024-04-10
improve DecoratorHandler

Uzlopak:
I think this will make the DecoratorHandler more resilient. 

- Added typecheck for the provided handler
- added onResponseStarted
- made handler attribute private

### DIFF
--- a/lib/handler/decorator-handler.js
+++ b/lib/handler/decorator-handler.js
@@ -11,15 +11,15 @@ module.exports = class DecoratorHandler {
   }
 
   onConnect (...args) {
-    return this.#handler.onConnect(...args)
+    return this.#handler.onConnect?.(...args)
   }
 
   onError (...args) {
-    return this.#handler.onError(...args)
+    return this.#handler.onError?.(...args)
   }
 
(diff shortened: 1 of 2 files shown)


ronag (approved):

lib/handler/decorator-handler.js
```diff
@@ -1,35 +1,44 @@
...
   }
 
   onComplete (...args) {
-    return this.handler.onComplete(...args)
+    return this.#handler.onComplete(...args)
   }
 
   onBodySent (...args) {
-    return this.handler.onBodySent(...args)
+    return this.#handler.onBodySent(...args)
```
ronag, line 42:
```suggestion
    return this.#handler.onBodySent?.(...args)
```

I would also automatically skip un-implemented hooks.

mcollina (approved):
lgtm

Uzlopak:
@ronag PTAL

ronag (approved):

ronag:
@Uzlopak can you maybe make it so all our own handler implementations extend from this?

PR #3080, opened 2024-04-10
websocket: fix ByteParser.run and parseCloseBody

Uzlopak:
ParseCloseBody can return null. In that case this.#info.closeInfo would be null and the following line would result in a crash.

https://github.com/nodejs/undici/blob/5d5454380b72889706146a9c4d5f65225a99640f/lib/web/websocket/receiver.js#L111

Also reduces the amount of Buffers created.

### DIFF
--- a/lib/web/websocket/receiver.js
+++ b/lib/web/websocket/receiver.js
@@ -108,7 +108,7 @@ class ByteParser extends Writable {
             // sending a Close frame in response, the endpoint typically echos the
             // status code it received.)
             let body = emptyBuffer
-            if (this.#info.closeInfo.code) {
+            if (this.#info.closeInfo?.code) {
               body = Buffer.allocUnsafe(2)
               body.writeUInt16BE(this.#info.closeInfo.code, 0)
             }
@@ -298,26 +298,30 @@ class ByteParser extends Writable {
       // defined as the status code (Section 7.4) contained in the first Close
       // control frame received by the application
       code = data.readUInt16BE(0)
+      if (!isValidStatusCode(code)) {
+        return null
+      }
+    } else {
+      return null
     }
 
     // https://datatracker.ietf.org/doc/html/rfc6455#section-7.1.6
-    /** @type {Buffer} */
-    let reason = data.subarray(2)
+    /** @type {number} */
+    let reasonOffset = 2
 
-    // Remove BOM
-    if (reason[0] === 0xEF && reason[1] === 0xBB && reason[2] === 0xBF) {
-      reason = reason.subarray(3)
+    // Detect BOM and ignore it
+    if (data.length >= 5 && data[2] === 0xEF && data[3] === 0xBB && data[4] === 0xBF) {
+      reasonOffset = 5
     }
 
-    if (code !== undefined && !isValidStatusCode(code)) {
-      return null
+    if (data.length - reasonOffset < 1) {
+      return { code, reason: '' }
     }
 
+    let reason = ''
     try {
-      reason = utf8Decode(reason)
-    } catch {
-      return null
-    }
+      reason = utf8Decode(data.subarray(reasonOffset))
+    } catch { }
 
     return { code, reason }
   }
--- a/test/websocket/close.js
+++ b/test/websocket/close.js
@@ -151,4 +151,27 @@ describe('Close', () => {
 
     await t.completed
   })
+
+  test('Close with code and invalid encoded reason', () => {
+    return new Promise((resolve) => {
+      const server = new WebSocketServer({ port: 0 })
+
+      server.on('connection', (ws) => {
+        ws.close(1000, Buffer.from([0xFF, 0xFE]))
+
+        ws.on('close', (code, reason) => {
+          assert.equal(code, 1000)
+          assert.equal(reason.length, 0)
+        })
+      })
+
+      const ws = new WebSocket(`ws://localhost:${server.address().port}`)
+      ws.onclose = (ev) => {
+        assert.equal(ev.code, 1000)
+        assert.equal(ev.reason, '')
+        server.close()
+        resolve()
+      }
+    })
+  })
 })
--- a/test/websocket/receive.js
+++ b/test/websocket/receive.js
@@ -1,11 +1,13 @@
 'use strict'
 
+const { tspl } = require('@matteo.collina/tspl')
 const { test } = require('node:test')
-const assert = require('node:assert')
 const { WebSocketServer } = require('ws')
 const { WebSocket } = require('../..')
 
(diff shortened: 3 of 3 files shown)


KhafraDev:
tests are rightfully failing, the PR assumes that having no closing code is invalid when it's optional

mcollina (approved):
lgtm

Uzlopak:
I think there is a bug in the original logic. 
`parseCloseBody` is returning in two places null. This will actually result in code breaking.

Uzlopak:
@KhafraDev 

What do you think now about this?

lib/web/websocket/receiver.js
```diff
@@ -298,26 +298,28 @@ class ByteParser extends Writable {
       // defined as the status code (Section 7.4) contained in the first Close
       // control frame received by the application
       code = data.readUInt16BE(0)
+      if (!isValidStatusCode(code)) {
+        return null
+      }
+    } else {
+      return null
```
KhafraDev, line 305: A close frame doesn't need to contain a reason or code, `data` (the body) can be empty here. You could assert the data length isn't 1 here, but we do check for its length elsewhere so it'd just be for safety.

lib/web/websocket/receiver.js
```diff
@@ -298,26 +298,28 @@ class ByteParser extends Writable {
...
+      return { code, reason: '' }
     }
 
-    try {
-      reason = utf8Decode(reason)
-    } catch {
-      return null
-    }
+    /** @type {string} */
+    const reason = utf8DecodeSafe(data.subarray(reasonOffset))
```
KhafraDev, line 322: This is wrong. If we receive an invalid utf-8 reason we are meant to fail the websocket connection.

KhafraDev (changes requested):
There are some major issues with this PR

lib/web/websocket/receiver.js
KhafraDev: The reason we don't fail it in that spot is because it's already closing, but the logic is completely wrong here, and will lead to issues if we do check the return type.

lib/web/websocket/receiver.js
Uzlopak: The thing is: 
https://app.codecov.io/gh/nodejs/undici/pull/3080/blob/lib/web/websocket/receiver.js

This line is covered by test. So it is possible that the size of the data is 0.

KhafraDev: I know, I said that. The buffer here is not the entire frame (which has to have a minimum length of 2), but the body of a frame. If there is no code and no reason, it is still a valid close frame, and will be empty here. Returning `null` means that the frame body is invalid, but that's not correct.

lib/web/websocket/receiver.js
```diff
@@ -298,26 +298,30 @@ class ByteParser extends Writable {
...
+      reasonOffset = 5
     }
 
-    if (code !== undefined && !isValidStatusCode(code)) {
-      return null
+    if (data.length - reasonOffset < 1) {
+      return { code, reason: '' }
     }
 
+    let reason = ''
```
KhafraDev, line 321: There is a misunderstanding here. Returning `null` means that the close frame's body is invalid. If we can't parse the reason, due to invalid utf-8, we **need** to return null here.

KhafraDev (changes requested):

lib/web/websocket/receiver.js
KhafraDev: On the other hand, the body can be empty or 2+ bytes, but it cannot be 1 byte, which is why I mentioned earlier that for safety we could assert that the body isn't a single byte. We already check elsewhere, but as I mentioned earlier, would be for safety.

Uzlopak: Well but it is happening:

image

KhafraDev: I don't understand what that has to do with what I said?

KhafraDev:
I'm going to go ahead and close this. The main change is creating a singular less Buffer in a very cold case. As lpinca said in my original PR implementing websocket (paraphrasing): "the performance of the opening handshake doesn't matter because it rarely happens". It's the same for closing the connection.

Uzlopak:
I disagree. At first it was about a performance improvement. Then I found a bug when I implemented this PR, which could lead to a crash.

lib/web/websocket/receiver.js
Uzlopak: This means, that your claim, that we already handled all cases with data.length < 2 is invalid. Either data.length === 0 or data.length === 1 or both are not covered by the calling function.

lib/web/websocket/receiver.js
Uzlopak: This seems to be a false conclusion. We got a valid code, so we wont need to set the code 1005 or 1006. The message is somewhat corrupt. So. We need to prefer the valid code and discard the message.

KhafraDev:
There isn't a bug being fixed without a test to cover it. The test that you added passes in main. The logic is mostly flawed as I pointed out.

Uzlopak:
well, lets see if i can write a rogue websocket server.

Uzlopak:
@KhafraDev 
added unit test ;)

KhafraDev:
This diff makes that test pass. There is quite a lot of extraneous stuff here.

```diff
diff --git a/lib/web/websocket/receiver.js b/lib/web/websocket/receiver.js
index 4b35ceb5..b1ef4720 100644
--- a/lib/web/websocket/receiver.js
+++ b/lib/web/websocket/receiver.js
@@ -316,7 +316,7 @@ class ByteParser extends Writable {
     try {
       reason = utf8Decode(reason)
     } catch {
-      return null
+      reason = undefined
--- a/lib/web/websocket/receiver.js
+++ b/lib/web/websocket/receiver.js
@@ -316,7 +316,7 @@ class ByteParser extends Writable {
     try {
       reason = utf8Decode(reason)
     } catch {
-      return null
+      reason = undefined
     }

     return { code, reason }
```

KhafraDev:
I'm not sure if ignoring the reason is correct though. Maybe failing the connection (basically, force closing it) is a better idea. In that case, nothing would need to be changed in parseCloseBody.

Uzlopak:
same test against main:
image

KhafraDev:
I understand, but the way this PR is fixing this is wrong.

KhafraDev:
```
1007

      1007 indicates that an endpoint is terminating the connection
      because it has received data within a message that was not
      consistent with the type of the message (e.g., non-UTF-8 [[RFC3629]]
      data within a text message).
```

PR #3084, opened 2024-04-10
chore: removed unused escapeFormDataName

mcollina:


Uzlopak:
I guess this is only relevant when sending FormData? Also validateBoundary is also unused, but maybe we can use it somewhere?

Uzlopak (approved):
LGTM

PR #3102, opened 2024-04-12
core: improve parseURL

Uzlopak:
There were some issues hidden.

E.g. if fields are undefined it will serialize them to undefined

```js
    describe('common', () => {
      test('returns valid URL object', () => {
        const url = parseURL({ protocol: 'http:' })
        strictEqual(url instanceof URL, true)
        strictEqual(url.href, 'http://undefined/')
      })
    })
```

So it should be now hardened and if the provided object has not sufficient information it will be replaced with empty string and not with undefined.

### DIFF
--- a/benchmarks/core/is-valid-port.mjs
+++ b/benchmarks/core/is-valid-port.mjs
@@ -1,0 +1,16 @@
+import { bench, group, run } from 'mitata'
+import { isValidPort } from '../../lib/core/util.js'
+
+const string = '1234'
+const number = 1234
+
+group('isValidPort', () => {
+  bench('string', () => {
+    return isValidPort(string)
+  })
+  bench('number', () => {
+    return isValidPort(number)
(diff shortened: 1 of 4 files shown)


KhafraDev:
how is parsing `parseURL({ protocol: 'http:' })` as `http:///` better?

Uzlopak:
@KhafraDev 
What are you referring to?

KhafraDev:
the example test you gave

Uzlopak:
Ah now i get it.

Yes it is better, because it will break 

```js
new URL('http://undefined/') // no problem
new URL('http:///') // TypeError: Failed to construct 'URL': Invalid URL
```

Uzlopak:
Ok. I think this PR is ready. The function is having full test coverage and behaves now properly imho.

metcoder95 (approved):

PR #3107, opened 2024-04-13
perf: optimization of request instantiation

tsctx:
1. Reuse fetch optimization logic in core.
2.  Avoid lowercasing.


mcollina (approved):
lgtm

lib/core/request.js
```diff
@@ -51,13 +52,13 @@ class Request {
...
       throw new InvalidArgumentError('path must be an absolute URL or start with a slash')
-    } else if (invalidPathRegex.exec(path) !== null) {
+    } else if (invalidPathRegex.test(path)) {
       throw new InvalidArgumentError('invalid request path')
     }
 
     if (typeof method !== 'string') {
       throw new InvalidArgumentError('method must be a string')
-    } else if (!isValidHTTPToken(method)) {
+    } else if (normalizeMethodRecord[method] === undefined && !isValidHTTPToken(method)) {
```
Uzlopak, line 61: Is there a perf difference for doing 
```suggestion
    } else if (method in normalizeMethodRecord && !isValidHTTPToken(method)) {
```

lib/core/util.js
```diff
@@ -562,6 +562,26 @@ function errorRequest (client, request, err) {
 const kEnumerableProperty = Object.create(null)
 kEnumerableProperty.enumerable = true
 
+const normalizeMethodRecord = {
```
Uzlopak, line 565: I dont like the name as the first part is a verb, indicating a function
Rename to normaledMethodRecords?

lib/core/request.js
tsctx: Unfortunately, `in operation` is slow to look up properties that do not exist.

Uzlopak: k, thanks

KhafraDev (commented):
Is there a reason to combine method validation and normalization into a single method? This instantly breaks the second we need one and not the other.

tsctx:
> Is there a reason to combine method validation and normalization into a single method? 

The reason for wanting to normalize and verify at the same time is that the case used in the verification can be reused.

> This instantly breaks the second we need one and not the other.

Yes, it is. The function was not changed, but a new one was created.

metcoder95 (commented):
Can you share the bench results?

tsctx:
> Can you share the bench results?

<details>

```js
import Request from "../../lib/core/request.js";
import DecoratorHandler from "../../lib/handler/decorator-handler.js";

import { bench, group, run } from 'mitata'

const handler = new DecoratorHandler({})

bench('new Request()', () => {
    return new Request('https://localhost', { path: "/", method: "get", body: null }, handler);
})

await run()
```

</details>

- main

```
benchmark          time (avg)             (min … max)       p75       p99      p999
----------------------------------------------------- -----------------------------
new Request()   90.41 ns/iter   (70.85 ns … 1'729 ns)    106 ns    205 ns    717 ns
```

- this patch

```
benchmark          time (avg)             (min … max)       p75       p99      p999
----------------------------------------------------- -----------------------------
new Request()   90.41 ns/iter   (70.85 ns … 1'729 ns)    106 ns    205 ns    717 ns
```

metcoder95 (approved):

Uzlopak:
Thank you for the benchmark. Can you add it to the benchmark folder please?

lib/web/fetch/request.js
```diff
@@ -39,6 +37,21 @@ const requestFinalizer = new FinalizationRegistry(({ signal, abort }) => {
...
+
+  const upperCase = method.toUpperCase()
+
+  if (forbiddenMethodsSet.has(upperCase)) {
+    throw new TypeError(`'${method}' HTTP method is unsupported.`)
+  }
+
+  // Note: must be in uppercase
+  return normalizedMethodRecordsBase[upperCase] ?? method
+}
```
Uzlopak, line 53: Why dont you move it to util?
cam you write some test only for this method?

tsctx: This is covered by wpt.

lib/core/util.js
```diff
@@ -592,6 +592,31 @@ function errorRequest (client, request, err) {
...
+const normalizedMethodRecords = {
+  ...normalizedMethodRecordsBase,
+  patch: 'patch',
+  PATCH: 'PATCH'
+}
+
+// Note: object prototypes should not be able to be referenced. e.g. `Object#hasOwnProperty`.
+Object.setPrototypeOf(normalizedMethodRecordsBase, null)
+Object.setPrototypeOf(normalizedMethodRecords, null)
+
```
DarkGL, line 619: As this object shouldn't change overtime, maybe we can add also

`
Object.freeze(normalizedMethodRecordsBase)
Object.freeze(normalizedMethodRecords)
`

DarkGL: fyi: https://github.com/DarkGL/object-freeze-seal-performance

mcollina (approved):
lgtm

lib/web/fetch/request.js
```diff
@@ -38,6 +36,21 @@ const requestFinalizer = new FinalizationRegistry(({ signal, abort }) => {
   signal.removeEventListener('abort', abort)
 })
 
+function validateAndNormalizeMethod (method) {
```
tsctx, line 39:
```suggestion
/**
 * @see https://fetch.spec.whatwg.org/#concept-method-normalize
 */
function validateAndNormalizeMethod (method) {
```

lib/core/util.js
tsctx: This is not a real-world benchmark; it is actually slower one.

lib/web/fetch/request.js
```diff
@@ -347,30 +363,15 @@ class Request {
     // 25. If init["method"] exists, then:
     if (init.method !== undefined) {
       // 1. Let method be init["method"].
-      let method = init.method
+      const method = init.method
 
-      const mayBeNormalized = normalizeMethodRecord[method]
+      // 2. If method is not a method or method is a forbidden method, then
```
tsctx, line 368:
```suggestion
      const mayBeNormalized = normalizeMethodRecord[method]
      // 2. If method is not a method or method is a forbidden method, then
```

lib/web/fetch/request.js
```diff
@@ -347,30 +363,15 @@ class Request {
...
+      // throw a TypeError.
```
tsctx, line 370:
```suggestion

      if (mayBeNormalized !== undefined) {
        // Note: Bypass validation DELETE, GET, HEAD, OPTIONS, POST, PUT, PATCH and these lowercase ones
        request.method = mayBeNormalized
      } else {
        // 2. If method is not a method or method is a forbidden method, then
        // throw a TypeError.
        if (!isValidHTTPToken(method)) {
          throw new TypeError(`'${method}' is not a valid HTTP method.`)
        }
```

lib/web/fetch/request.js
```diff
@@ -347,30 +363,15 @@ class Request {
...
-          throw new TypeError(`'${method}' HTTP method is unsupported.`)
-        }
-
-        // 3. Normalize method.
-        method = normalizeMethod(method)
-
-        // 4. Set request’s method to method.
-        request.method = method
-      }
+      // 4. Set request’s method to method.
```
tsctx, line 373:
```suggestion
        if (forbiddenMethodsSet.has(method.toUpperCase())) {
          throw new TypeError(`'${method}' HTTP method is unsupported.`)
        }

        // 3. Normalize method.
        method = normalizedMethodRecords[method] ?? validateAndNormalizeMethod(method)

        // 4. Set request’s method to method.
        request.method = method
      }
      // 4. Set request’s method to method.
```

lib/web/fetch/request.js
```diff
@@ -347,30 +363,15 @@ class Request {
...
-      if (mayBeNormalized !== undefined) {
-        // Note: Bypass validation DELETE, GET, HEAD, OPTIONS, POST, PUT, PATCH and these lowercase ones
-        request.method = mayBeNormalized
-      } else {
-        // 2. If method is not a method or method is a forbidden method, then
-        // throw a TypeError.
-        if (!isValidHTTPToken(method)) {
-          throw new TypeError(`'${method}' is not a valid HTTP method.`)
-        }
+      // 3. Normalize method.
```
tsctx, line 371:
```suggestion
```

lib/web/fetch/request.js
```diff
@@ -347,9 +363,11 @@ class Request {
     // 25. If init["method"] exists, then:
     if (init.method !== undefined) {
       // 1. Let method be init["method"].
-      let method = init.method
+      const method = init.method
```
tsctx, line 366:
```suggestion
      let method = init.method
```

lib/web/fetch/request.js
```diff
@@ -347,9 +363,11 @@ class Request {
...
 
-      const mayBeNormalized = normalizeMethodRecord[method]
+      const mayBeNormalized = normalizedMethodRecords[method]
+      // 2. If method is not a method or method is a forbidden method, then
+      // throw a TypeError.
```
tsctx, line 370:
```suggestion
```

lib/web/fetch/request.js
```diff
@@ -366,11 +384,13 @@ class Request {
...
 
         // 3. Normalize method.
-        method = normalizeMethod(method)
+        method = normalizedMethodRecords[method] ?? validateAndNormalizeMethod(method)
 
         // 4. Set request’s method to method.
         request.method = method
       }
+      // 4. Set request’s method to method.
+      request.method = normalizedMethodRecords[method] ?? validateAndNormalizeMethod(method)
```
tsctx, line 393:
```suggestion
```

lib/web/fetch/request.js
```diff
@@ -366,7 +382,7 @@ class Request {
         }
 
         // 3. Normalize method.
-        method = normalizeMethod(method)
+        method = normalizedMethodRecords[method] ?? validateAndNormalizeMethod(method)
```
tsctx, line 385:
```suggestion
        method = normalizedMethodRecords[method] ?? method
```

lib/web/fetch/util.js
```diff
@@ -6,7 +6,7 @@ const { redirectStatusSet, referrerPolicySet: referrerPolicyTokens, badPortsSet
 const { getGlobalOrigin } = require('./global')
 const { collectASequenceOfCodePoints, collectAnHTTPQuotedString, removeChars, parseMIMEType } = require('./data-url')
 const { performance } = require('node:perf_hooks')
-const { isBlobLike, ReadableStreamFrom, isValidHTTPToken } = require('../../core/util')
+const { isBlobLike, ReadableStreamFrom, isValidHTTPToken, normalizedMethodRecordsBase } = require('../../core/util')
```
tsctx, line 9:
```suggestion
const { isBlobLike, ReadableStreamFrom, isValidHTTPToken } = require('../../core/util')
```

lib/web/fetch/request.js
```diff
@@ -361,12 +359,15 @@ class Request {
...
-        if (forbiddenMethodsSet.has(method.toUpperCase())) {
+        const upperCase = method.toUpperCase()
+
+        if (forbiddenMethodsSet.has(upperCase)) {
           throw new TypeError(`'${method}' HTTP method is unsupported.`)
         }
 
         // 3. Normalize method.
-        method = normalizeMethod(method)
+        // Note: must be in uppercase
```
tsctx, line 369:
```suggestion
        // https://fetch.spec.whatwg.org/#concept-method-normalize
        // Note: must be in uppercase
```

luanmuniz:
This is just a comment from someone who doesn't know how this repo works, so feel free to ignore it:

I see optimization, and I just want to say that I found Map() to be more performant than plain objects. It could be a slight performance improvement to have `normalizedMethodRecordsBase` as Map() instead of a plain object.

KhafraDev (approved):

Issue #2164, opened 2023-06-18
fetch: limit web streams usage in body mixin methods


KhafraDev:
The fetch spec requires us to create a new web Readable stream for every `Response` and `Request` object, which can likely be avoided in most cases.

Given the following examples:
```js
new Response('abc')
new Response(new TextEncoder().encode('abc'))
new Response(new Blob(['abc']))
// etc.
```

A web stream is created and then converted *back* to a string! The following diff improves `.text()` performance when the body passed is a string by **10%**, as a simple demo (we could further limit web stream usage for more performance gains!!!).

```diff
diff --git a/lib/fetch/response.js b/lib/fetch/response.js
index 1029dbef..c2d2bd3d 100644
--- a/lib/fetch/response.js
+++ b/lib/fetch/response.js
@@ -154,7 +154,7 @@ class Response {
     // 4. If body is non-null, then set bodyWithType to the result of extracting body.
     if (body != null) {
       const [extractedBody, type] = extractBody(body)
-      bodyWithType = { body: extractedBody, type }
+      bodyWithType = { body: extractedBody, type, source: body }
     }

     // 5. Perform initialize a response given this, init, and bodyWithType.
diff --git a/lib/fetch/util.js b/lib/fetch/util.js
index 400687ba..9e4302a9 100644
--- a/lib/fetch/util.js
+++ b/lib/fetch/util.js
@@ -824,6 +824,11 @@ function fullyReadBody (body, processBody, processBodyError) {
   //    with taskDestination.
   const errorSteps = (error) => queueMicrotask(() => processBodyError(error))

+  if (typeof body.source === 'string') {
+    successSteps(new TextEncoder().encode(body.source))
+    return
+  }
+
   // 4. Let reader be the result of getting a reader for body’s stream.
   //    If that threw an exception, then run errorSteps with that
   //    exception and return.
```

cc @anonrig

KhafraDev:
benchmarks:
```js

ronag:
I'm -1 for anything that goes against the spec without significant improvements.

KhafraDev:
I understand, but 10% performance improvement is significant for the small change I made. We still extract the body (convert it to a readable web stream), but instead of reading the bytes from that stream, we use the [body's source](https://fetch.spec.whatwg.org/#concept-body-source). The body is still parsed in the same way, the only thing missing is to set the body to used when consumed, since the stream won't be consumed we'd have to add a flag to handle that.

anonrig:
[...]

PR #3113, opened 2024-04-13
fetch: limit web streams usage in body mixin methods

mertcanaltin:
This PR aims to improve performance by optimizing the processing of webflows in the fetch module.
This experiment was already done by @KhafraDev , I just opened a pr

I thought there would be discussion about it and it might benefit fetch 🚀 
#2164
here is a benchmark made by @KhafraDev 🙏 
benchmarks:
```js
import { Response as UndiciResponse } from './index.js'
import { cronometro } from 'cronometro'

await cronometro({
  async 'undici Response' () {
    await new UndiciResponse('abc').text()
  },
  async 'global Response' () {
    await new Response('abc').text()
  }
})

```

```
╔═════════════════╤═════════╤═════════════════╤═══════════╗
║ Slower tests    │ Samples │          Result │ Tolerance ║
╟─────────────────┼─────────┼─────────────────┼───────────╢
║ global Response │   10000 │ 40169.19 op/sec │  ± 1.59 % ║
╟─────────────────┼─────────┼─────────────────┼───────────╢
║ Fastest test    │ Samples │          Result │ Tolerance ║
╟─────────────────┼─────────┼─────────────────┼───────────╢
║ undici Response │   10000 │ 44024.31 op/sec │  ± 2.12 % ║
╚═════════════════╧═════════╧═════════════════╧═══════════╝
```

### DIFF
--- a/lib/web/fetch/util.js
+++ b/lib/web/fetch/util.js
@@ -1056,7 +1056,8 @@ async function fullyReadBody (body, processBody, processBodyError) {
   const errorSteps = processBodyError
 
   if (typeof body.source === 'string') {
-    successSteps(new TextEncoder().encode(body.source))
+    const buffer = new TextEncoder().encode(body.source)
+    successSteps(Buffer.from(buffer.buffer, buffer.byteOffset, buffer.byteLength))
     return
   }
 
--- a/lib/web/fetch/util.js
+++ b/lib/web/fetch/util.js
@@ -1055,6 +1055,11 @@ async function fullyReadBody (body, processBody, processBodyError) {
   //    with taskDestination.
   const errorSteps = processBodyError
 
+  if (typeof body.source === 'string') {
+    successSteps(new TextEncoder().encode(body.source))
+    return
+  }
+
   // 4. Let reader be the result of getting a reader for body’s stream.
   //    If that threw an exception, then run errorSteps with that
   //    exception and return.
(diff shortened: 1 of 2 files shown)


mertcanaltin:
I'm running tests, I'll fix them 🧪

lib/web/fetch/util.js
```diff
@@ -1055,6 +1055,11 @@ async function fullyReadBody (body, processBody, processBodyError) {
   //    with taskDestination.
   const errorSteps = processBodyError
 
+  if (typeof body.source === 'string') {
+    successSteps(new TextEncoder().encode(body.source))
```
Uzlopak, line 1059: I did not check your code and if this makes sense. But I think this wil break without-intl builds . check how we implemented utf8Decode and look if you can implement something for utf8Encode?!

mertcanaltin: thank you very much for your suggestion I will try this

tsctx, line 1059: In this way, test failures can be corrected.
```suggestion
    const buffer = new TextEncoder().encode(body.source)
    successSteps(Buffer.from(buffer.buffer, buffer.byteOffset, buffer.byteLength))
```

mertcanaltin: thank you for suggestion 🙏

mertcanaltin: @tsctx I wonder how you caught this

tsctx: Because I had the same error.

KhafraDev:
There are related failures, but I don't necessarily think this should land as-is even if those failures were fixed.

KhafraDev (changes requested):
(making sure to block this)

mertcanaltin:
I wonder if you have a suggestion, if it's healthy to keep it as it is, I'll close this place @KhafraDev

Issue #2835, opened 2024-02-25
undici - next


ronag:
TODO:

- [x] remove anti-pattern dispatcher hooks
- [x] remove old interceptor pattern (https://github.com/nodejs/undici/pull/2754)
- [x] new interceptor pattern (https://github.com/nodejs/undici/pull/2795)
- [ ] change so that dispatcher hooks take an options object instead of multiple arguments
- [ ] remove pipelining retries from core
- [ ] implement "idempotent" retry interceptor for pipeline resets
- [ ] implement "decompress" interceptor - #3255 
- [x] separate out http2
- [ ] #3275
- [ ] #3276
- [x] make dump investigate content-length to see if it's worth trying to read
- [x] move dump to interceptor - #3118 
- [ ] fix `Readable.setEncoding` issue (https://github.com/nodejs/undici/issues/1125#issuecomment-987724800)
- ~~[x] non-keep alive connections are not data safe~~
- ~~[ ] Should we split `onHeaders` into more fine-grained hooks? i.e. `onHeader(key, val)` and `onStatus(code, message)`?~~

Refs: https://github.com/nodejs/undici/issues/2722

ronag:
@metcoder95

metcoder95:
With `dispatcher hooks` do you refer to interceptors?

ronag:
> With dispatcher hooks do you refer to interceptors?

It's methods on `Dispatcher`, i.e. `onHeaders`  etc...

ronag:
Some of the above we can probably land on main.

metcoder95:
>Should we split onHeaders into more fine-grained hooks? i.e. onHeader(key, val) and onStatus(code, message)?

This seems interesting, in theory, the HTTP response is split into three parts, the status line, headers and body; maybe we can follow this line?
See potential use cases, specially around stop request processing if the status is not the one expected

ronag:
> This seems interesting, in theory, the HTTP response is split into three parts, the status line, headers and body; maybe we can follow this line?

it's actually headers first, then status and then body.

metcoder95:
I'm working on the dump to interceptor, but I'm curious what do you have in mind for that API?

Also removing the legacy interceptor concepts from `next`, but is out of sync, should I open aim the branch anyways in this state?

metcoder95:
cc: @ronag

ronag:
I will sort the next branch sometime in the next two weeks. Just open or against current state.

ronag:
[...]

PR #3118, opened 2024-04-14
feat: dump interceptor

metcoder95:
## Rationale

Relates to #2835 

## Changes

- **feat: add dump interceptor**
- **fix(types): interceptor definitions**

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready


lib/interceptor/dump.js
```diff
@@ -0,0 +1,109 @@
...
+  }
+
+  onData (chunk) {
+    this.#size = this.#size + chunk.length
+
+    if (this.#size < this.maxSize) {
+      return true
+    }
+
+    // TODO: shall we forward the rest of the data to the handler or better to abort?
```
metcoder95, line 87: Suggestions? I might be had it wrong here 🤔

lib/interceptor/dump.js
```diff
@@ -0,0 +1,109 @@
+'use strict'
+
+const util = require('../core/util')
+const { InvalidArgumentError, RequestAbortedError } = require('../core/errors')
+
+class DumpHandler {
+  res = null
+  maxSize = 1024 * 1024
```
ronag, line 8: Why are these not private?

ronag (changes requested):
The point of the dump interceptor is that it will dump *on abort*.

docs/docs/api/Dispatcher.md
```diff
@@ -952,6 +952,38 @@ const client = new Client("http://example.com").compose(
 );
 ```
 
+##### `dump`
+
+The `dump` interceptor enables you to dump the response body from a request upon a given limit.
+
+**Options**
+- `maxSize` - The maximum size (in bytes) of the response body to dump. Default: `1048576`.
```
ronag, line 960:
```suggestion
- `maxSize` - The maximum size (in bytes) of the response body to dump. If the size exceeds this value then the connection will be closed. Default: `1048576`.
```

lib/interceptor/dump.js
metcoder95: Leftover

lib/interceptor/dump.js
```diff
@@ -0,0 +1,109 @@
...
+function createDumpInterceptor (
+  { maxSize: defaultMaxSize } = { maxSize: 1024 * 1024 }
+) {
+  return dispatch => {
+    return function Intercept (opts, handler) {
+      const { dumpMaxSize = defaultMaxSize } = opts
+
+      const redirectHandler = new DumpHandler({ maxSize: dumpMaxSize }, handler)
+
+      return dispatch(opts, redirectHandler)
```
metcoder95, line 104:
```suggestion
      const dumpHandler = new DumpHandler({ maxSize: dumpMaxSize }, handler)

      return dispatch(opts, dumpHandler)
```

docs/docs/api/Dispatcher.md
```diff
@@ -952,6 +952,38 @@ const client = new Client("http://example.com").compose(
 );
 ```
 
+##### `dump`
+
+The `dump` interceptor enables you to dump the response body from a request upon a given limit.
+
+**Options**
+- `maxSize` - The maximum size (in bytes) of the response body to dump. If the size exceeds this value then the connection will be closed. Default: `1048576`.
```
metcoder95, line 960:
```suggestion
- `maxSize` - The maximum size (in bytes) of the response body to dump. If the size of the request's body exceeds this value then the connection will be closed. Default: `1048576`.
```

metcoder95:
> Should the dump interceptor should also dump _on abort_?

Do you mean to not directly abort the connection, but rather let it flow? Or how so?

docs/docs/api/Dispatcher.md
```diff
@@ -952,6 +952,38 @@ const client = new Client("http://example.com").compose(
 );
 ```
 
+##### `dump`
```
mcollina, line 955: Why did you call it `dump`? I would call it `head`, as a reference to the UNIX head command. `curl http://nodejs.org | head`.

lib/interceptor/dump.js
mcollina: you _must_ keep the data flowing from the request or error/abort it. The latter will destroy the connection. It should be a user choice, but I would error/abort the socket by default.

docs/docs/api/Dispatcher.md
Uzlopak: `dump` not a very "sexy" name but is related to the linux term [core dump](https://en.wikipedia.org/wiki/Core_dump). The proposed "head" is not self-explaining, especially it is about snapshotting the body of the response and not the head of the reponse.

mcollina:
@ronag ptal

lib/interceptor/dump.js
```diff
@@ -0,0 +1,170 @@
+'use strict'
+
+const util = require('../core/util')
+const { InvalidArgumentError, RequestAbortedError } = require('../core/errors')
+
+class DumpHandler {
```
ronag, line 6:
```suggestion
class DumpHandler extends DecoratorHandler {
```

lib/interceptor/dump.js
```diff
@@ -0,0 +1,170 @@
...
+
+      this.#abort(this.#reason)
+      return
+    }
+
+    if (this.#waitForTrailers) {
+      this.#hasTrailers = headers.trailer != null
+    }
+
+    this.#handler.onHeaders(statusCode, rawHeaders, resume, statusMessage)
```
ronag, line 90:
```suggestion
    return this.#handler.onHeaders(statusCode, rawHeaders, resume, statusMessage)
```

ronag (changes requested):
I have more opinions. Will try to have a better look as soon as I can.

lib/interceptor/dump.js
```diff
@@ -0,0 +1,170 @@
...
+      if (this.#abort != null) {
+        this.#abort(reason)
+      } else {
+        this.#reason = reason
+      }
+    })
+  }
+
+  onConnect (...args) {
+    const [abort] = args
```
ronag, line 49:
```suggestion
  onConnect (abort) {
```

lib/interceptor/dump.js
```diff
@@ -0,0 +1,170 @@
...
+
+    // Handle possible onConnect duplication
+    this.#handler.onConnect(reason => {
+      this.#aborted = true
+      if (this.#abort != null) {
+        this.#abort(reason)
+      } else {
+        this.#reason = reason
+      }
+    })
```
ronag, line 45: This is not neccassary? The handler is idempotent.

lib/interceptor/dump.js
```diff
@@ -0,0 +1,170 @@
...
+    this.#handler.onResponseStarted?.()
+  }
+
+  onBodySent () {
+    this.#handler.onBodySent?.()
+  }
+
+  onUpgrade (statusCode, headers, socket) {
+    this.#handler.onUpgrade?.(statusCode, headers, socket)
+  }
```
ronag, line 68:
```suggestion
```

lib/interceptor/dump.js
```diff
@@ -0,0 +1,170 @@
...
+    const contentLength = headers['content-length']
+
+    if (contentLength != null && contentLength > this.#maxSize) {
+      this.#reason = new RequestAbortedError(
+        `Response size (${contentLength}) larger than maxSize (${
+          this.#maxSize
+        })`
+      )
+
+      this.#abort(this.#reason)
```
ronag, line 82:
```suggestion
      this.#abort(new RequestAbortedError(
        `Response size (${contentLength}) larger than maxSize (${
          this.#maxSize
        })`
      ))
```

lib/interceptor/dump.js
```diff
@@ -0,0 +1,170 @@
...
+  onComplete (trailers) {
+    this.#completed = true
+    this.#handler.onComplete(trailers)
+
+    if (this.#dumped && this.#abortOnDumped) {
+      this.#reason = new RequestAbortedError(
+        `Response dumped (${this.#size}) for max size (${this.#maxSize})`
+      )
+
+      this.#abort(this.#reason)
```
ronag, line 136:
```suggestion
      this.#reason = new RequestAbortedError(
        `Response dumped (${this.#size}) for max size (${this.#maxSize})`
      )

      this.#handler.onError(this.#reason)
```

ronag: abort doesn't do anything inside of onComplete

ronag (changes requested):
I would remove waitForTrailers. Just makes things complicated for an edge case.

docs/docs/api/Dispatcher.md
```diff
@@ -952,6 +952,44 @@ const client = new Client("http://example.com").compose(
...
+
+The `dump` interceptor enables you to dump the response body from a request upon a given limit.
+
+**Options**
+- `maxSize` - The maximum size (in bytes) of the response body to dump. If the size of the request's body exceeds this value then the connection will be closed. Default: `1048576`.
+- `abortOnDumped` - States whether or not abort the request after the response's body being dumped. Default: `true`.
+- `waitForTrailers` - Hints the dispatcher to wait for trailers if the response's body has been dumped. Default: `false`.
```
ronag, line 962: I would skip both of these. Also not sure what abortOnDumped does is needed for?

ronag: Actually, what would make sense is to have a `dumpOnAbort` abort option.

metcoder95: For `abortOnDumped` I was thinking for use cases where you do not want to abort the request but dump its body, in that way things calls to e.g. `client.request` does not reject the promise but rather returns you an empty `''` as body.

`waitForTrailers` in the case they are interested in the trailers, not really common so happy to drop it 👍 

For `dumpOnAbort`, still not sure, do you mean intercept the abort call and do not abort the request but rather consume it?

ronag: >  I was thinking for use cases where you do not want to abort the request but dump its body, in that way things calls to e.g

I'm confused. When would you not want that?

> do you mean intercept the abort call and do not abort the request but rather consume it?

So if you call abort in a sub handler, that will not kill the connection, but try to consume it instead, in order to re-use the connection.

lib/interceptor/dump.js
```diff
@@ -0,0 +1,154 @@
...
+    this.#handler = handler
+  }
+
+  onConnect (abort) {
+    if (this.#aborted) {
+      abort(this.#reason)
+      return
+    }
+
+    this.#abort = abort
```
ronag, line 47: You still need to call handler.onConnect

docs/docs/api/Dispatcher.md
metcoder95: >I'm confused. When would you not want that?

I can honestly just think in situations where you are just interested in dump it but do not want the request to be rejected (which currently happens if the request is aborted and try to read the body).

>So if you call abort in a sub handler, that will not kill the connection, but try to consume the body instead, that way one can re-use the connection of an aborted request.

Got it, that was the part that I was missing; then its not that far from what I've had in mind. Let me draft it quickly

metcoder95: >So if you call abort in a sub handler, that will not kill the connection, but try to consume the body instead, that way one can re-use the connection of an aborted request.

Thinking twice, this was the original situation of `abortOnDumped`, meaning that if enabled we'll abort the request and drop the connection; if disabled, we consume the body and keep the connection alive.

wdyt if we go with `dumpOnAbort` and we make naturally the handle do not abort the request so the connection can be reused?
Basically what you suggested and by nature we make the handler do not drop connections, otherwise it kind of feel pointless the interceptor as we are dropping the connection on every dump

metcoder95:
@ronag can you take a look; I'll fix the remaining test later today

lib/interceptor/dump.js
```diff
@@ -93,58 +91,33 @@ class DumpHandler extends DecoratorHandler {
...
-        (!this.#waitForTrailers || !this.#hasTrailers)
-      ) {
-        this.#reason = new RequestAbortedError(
-          `Response dumped (${this.#size}) for max size (${this.#maxSize})`
-        )
-
-        this.#abort(this.#reason)
-        return false
-      }
+      this.#handler.onData('')
```
ronag, line 94: This is strange?

lib/interceptor/dump.js
```diff
@@ -0,0 +1,129 @@
...
+    if (
+      !(err instanceof RequestAbortedError) &&
+      (!this.#dumped || this.#aborted)
+    ) {
+      this.#handler.onError(err)
+      return
+    }
+
+    if (this.#dumped) {
+      this.#handler.onComplete([])
```
ronag, line 85: This will call onComplete twice? I think you mean !this.#dumped

ronag (commented):
Just to be clear for me here are two cases:

1. You want to dump the response regardless
2. You want to dump the response, *if* aborting and avoid closing the connection

I guess the follow configuration apply at the moment:

1. { maxSize: 16 * 1024, dumpOnAbort: true }
2.  { max Size: 9999999999, dumpOnAbort: true }

lib/interceptor/dump.js
metcoder95: Otherwise the request keeps hanging, tried a workaround without luck; find it quite counterintuitive but most likely is something that might need to be addressed at the `request` level, but not sure the scope of this PR

metcoder95:
> Just to be clear for me here are two cases:
> 
> 1. You want to dump the response regardless
> 2. You want to dump the response, _if_ aborting and avoid closing the connection
> 
> I guess the follow configuration apply at the moment:
> 
> 1. { maxSize: 16 * 1024, dumpOnAbort: true }
> 2. { max Size: 9999999999, dumpOnAbort: true }

These are inclusive.
If:

>1. You want to dump the response regardless

This usually means you want to keep the connection alive, just dump the response's body, with the current setup is possible, even with the defaults.

>2. You want to dump on abort

This is also already covered, with the defaults.

>3. You want both, dump regardless, and dump on abort.

The default does that

>4. Want to dump on **abort only** but not regardless

This is not currently covered as regardless of the combination, you'll still be dumping the response even if not aborted; do we want to cover this case?
I imagined that if you want to use the dump is because you don't care about the response's body.

>5. Want to dump regardless but not on abort

Already covered by disabling `dumpOnAbort: false`

Anything else that I might be missing?

lib/interceptor/dump.js
ronag: I'd like to address this before landing. Do you have minimal a repro? I can take a look.

lib/interceptor/dump.js
ronag: @metcoder95 ping

metcoder95: pong

Sorry, leftover, forgot to delete it!

lib/interceptor/dump.js
metcoder95: nvm, I found the issue; it was on me, the request was in the aborted state and couldn't continue because already aborted but I never called the `onError` hook from the original handler. That made the work

metcoder95:
It seems only fails on windows but cannot understand fully why 🤔

ronag (approved):

metcoder95:
I disabled the tests for Windows temporary, I'll try to debug this later on.
cc: @ronag @mcollina

docs/docs/api/Dispatcher.md
```diff
@@ -952,6 +952,44 @@ const client = new Client("http://example.com").compose(
...
+);
+
+// or 
+client.dispatch(
+  {
+    path: "/",
+    method: "GET",
+    dumpMaxSize: 1024,
+    abortOnDumped: true,
+    waitForTrailers: false,
```
ronag, line 987:
```suggestion
```

docs/docs/api/Dispatcher.md
```diff
@@ -952,6 +952,44 @@ const client = new Client("http://example.com").compose(
...
+
+```js
+const { Client, interceptors } = require("undici");
+const { dump } = interceptors;
+
+const client = new Client("http://example.com").compose(
+  dump({
+    maxSize: 1024,
+    abortOnDumped: true,
+    waitForTrailers: false,
```
ronag, line 976:
```suggestion
```

docs/docs/api/Dispatcher.md
metcoder95, line 962:
```suggestion
```

types/interceptors.d.ts
```diff
@@ -1,5 +1,11 @@
 import Dispatcher from "./dispatcher";
+import RetryHandler from "./retry-handler";
 
-type RedirectInterceptorOpts = { maxRedirections?: number }
+export type DumpInterceptorOpts = { maxSize?: number, abortOnDumped?: boolean, waitForTrailers?: boolean}
```
metcoder95, line 4:
```suggestion
export type DumpInterceptorOpts = { maxSize?: number, abortOnDumped?: boolean }
```

lib/interceptor/dump.js
```diff
@@ -0,0 +1,132 @@
...
+    )
+  }
+
+  #customAbort (reason) {
+    this.#aborted = true
+    this.#reason = reason
+  }
+
+  // TODO: will require adjustment after new hooks are out
+  onHeaders (statusCode, rawHeaders, resume, statusMessage) {
```
ronag, line 52:
```suggestion
  onHeaders (statusCode, rawHeaders, resume, statusMessage) {
    if (this.#aborted) {
      return true
    }
```

lib/interceptor/dump.js
```diff
@@ -0,0 +1,132 @@
...
+      if (this.#aborted) {
+        this.#handler.onError(this.reason)
+        return
+      }
+
+      this.#handler.onError(err)
+    }
+  }
+
+  onData (chunk) {
```
ronag, line 83:
```suggestion
  onData (chunk) {
    if (this.#aborted) {
      return true
    }
```

lib/interceptor/dump.js
```diff
@@ -0,0 +1,132 @@
...
+  onComplete (trailers) {
+    if (!this.#dumped) {
+      if (this.#aborted) {
+        this.#handler.onError(this.reason)
+        return
+      }
+
+      this.#handler.onComplete(trailers)
+    }
+  }
```
ronag, line 108:
```suggestion
    if (this.#dumped) {
      return
    }
    
    if (this.#aborted) {
      this.#handler.onError(this.reason)
    } else {
      this.#handler.onComplete(trailers)
    }
  }
```

lib/interceptor/dump.js
```diff
@@ -0,0 +1,132 @@
...
+    if (maxSize != null && (!Number.isFinite(maxSize) || maxSize < 1)) {
+      throw new InvalidArgumentError('maxSize must be a number greater than 0')
+    }
+
+    if (dumpOnAbort != null && typeof dumpOnAbort !== 'boolean') {
+      throw new InvalidArgumentError('dumpOnAbort must be a boolean')
+    }
+
+    this.#maxSize = maxSize ?? this.#maxSize
+    this.#dumpOnAbort = dumpOnAbort ?? this.#dumpOnAbort
```
ronag, line 29: How about not having this as an option and just have it always enabled? We might also want to have a dump timeout in addition to a maxSize?

metcoder95: I believe this goes in line with your suggestions for the PR.
Don't have strong opinion on this one, I believe there might be situations where you don't want to automatically dump on abort, but if we add it, it might never get used. 

So let's simplify and follow your suggestions, if we see use cases or issues we can always add it 👍

metcoder95: Regarding the timeout, shouldn't that already be covered by the `bodyTimeout`?
It might look redundant, and to have effect it might require to always set it lower to the `bodyTimeout` set for the Client/Agent

lib/interceptor/dump.js
```diff
@@ -0,0 +1,130 @@
...
+  }
+
+  onData (chunk) {
+    this.#size = this.#size + chunk.length
+
+    if (this.#size >= this.#maxSize) {
+      this.#dumped = true
+
+      if (this.#aborted) {
+        this.#handler.onError(this.reason)
```
ronag, line 86:
```suggestion
        this.#handler.onError(this.#reason)
```

ronag (changes requested):

docs/docs/api/Dispatcher.md
```diff
@@ -952,6 +952,40 @@ const client = new Client("http://example.com").compose(
...
+  })
+);
+
+// or 
+client.dispatch(
+  {
+    path: "/",
+    method: "GET",
+    dumpMaxSize: 1024,
+    abortOnDumped: true,
```
ronag, line 983:
```suggestion
```

ronag (changes requested):
You haven't implemented the timeout? Remove?

docs/docs/api/Dispatcher.md
```diff
@@ -952,6 +952,40 @@ const client = new Client("http://example.com").compose(
...
+**Example - Basic Dump Interceptor**
+
+```js
+const { Client, interceptors } = require("undici");
+const { dump } = interceptors;
+
+const client = new Client("http://example.com").compose(
+  dump({
+    maxSize: 1024,
+    abortOnDumped: true,
```
metcoder95, line 973:
```suggestion
```

lib/interceptor/dump.js
```diff
@@ -0,0 +1,130 @@
...
+      this.#handler.onError(this.reason)
+      return
+    }
+
+    this.#handler.onComplete(trailers)
+  }
+}
+
+function createDumpInterceptor (
+  { maxSize: defaultMaxSize, dumpOnAbort: defaultDumpOnAbort } = {
```
metcoder95, line 110:
```suggestion
  { maxSize: defaultMaxSize } = {
```

lib/interceptor/dump.js
```diff
@@ -0,0 +1,130 @@
...
+    maxSize: 1024 * 1024,
+    timeout: 30e3 // 30s
```
metcoder95, line 112:
```suggestion
```

lib/interceptor/dump.js
```diff
@@ -0,0 +1,130 @@
...
+    if (maxSize != null && (!Number.isFinite(maxSize) || maxSize < 1)) {
+      throw new InvalidArgumentError('maxSize must be a number greater than 0')
+    }
+
+    if (timeout != null && (!Number.isFinite(timeout) || timeout < 1)) {
+      throw new InvalidArgumentError('timeout must be a number greater than 0')
+    }
+
+    this.#maxSize = maxSize ?? this.#maxSize
+    this.#timeout = timeout ?? this.#timeout
```
metcoder95, line 29:
```suggestion
```

lib/interceptor/dump.js
```diff
@@ -0,0 +1,130 @@
+'use strict'
+
+const util = require('../core/util')
+const { InvalidArgumentError, RequestAbortedError } = require('../core/errors')
+const DecoratorHandler = require('../handler/decorator-handler')
+
+class DumpHandler extends DecoratorHandler {
+  #maxSize = 1024 * 1024
+  #timeout = 30e3 // 30s
```
metcoder95, line 9:
```suggestion
```

lib/interceptor/dump.js
```diff
@@ -0,0 +1,127 @@
...
+  constructor ({ maxSize, timeout }, handler) {
+    super(handler)
+
+    if (maxSize != null && (!Number.isFinite(maxSize) || maxSize < 1)) {
+      throw new InvalidArgumentError('maxSize must be a number greater than 0')
+    }
+
+    if (timeout != null && (!Number.isFinite(timeout) || timeout < 1)) {
+      throw new InvalidArgumentError('timeout must be a number greater than 0')
+    }
```
ronag, line 25:
```suggestion
  constructor ({ maxSize }, handler) {
    super(handler)

    if (maxSize != null && (!Number.isFinite(maxSize) || maxSize < 1)) {
      throw new InvalidArgumentError('maxSize must be a number greater than 0')
    }
```

ronag (approved):

mcollina (approved):
lgtm

Issue #3123, opened 2024-04-15
`undici.fetch` breaks on Node.js <=18.13.0 in undici@6.13.0


trentm:
## Bug Description

Since undici@6.13.0, `undici.fetch` breaks on Node.js 18.13.0 and earlier.
"package.json" includes:

```
  "engines": {
    "node": ">=18.0"
  },
```

but I'm not sure if anything less than v18.17.0 is really supported anymore -- the move to borp as part of https://github.com/nodejs/undici/issues/2267 means that the test suite cannot be run with Node.js earlier than v18.17.0.

## Reproducible By

Using this "playfetch.js":

```js
% cat playfetch.js
const undici = require('./');
async function main() {
  const res = await undici.fetch('http://www.google.com/');
  console.log('res:', res.status);
  const text = await res.text();
  console.log('text: %s...', text.slice(0, 70));
}
main();
```

In a git clone at commit 8a07bbd7 (from #3095):

```
% git remote -v
origin	git@github.com:nodejs/undici.git (fetch)
origin	git@github.com:nodejs/undici.git (push)

% git log1 -1
* 8a07bbd7 - (HEAD) fetch: avoid creation of an intermediary ReadableStream (#3095) (4 days ago) <Khafra>

% node --version
v18.13.0

% node playfetch.js
res: 200
/Users/trentm/src/undici/lib/web/fetch/index.js:2034
        fetchParams.controller.controller.error(new TypeError('terminated', {
                                                ^

TypeError: terminated
    at Fetch.onAborted (/Users/trentm/src/undici/lib/web/fetch/index.js:2034:49)
    at Fetch.emit (node:events:513:28)
    at Fetch.terminate (/Users/trentm/src/undici/lib/web/fetch/index.js:93:10)
    at /Users/trentm/src/undici/lib/web/fetch/index.js:506:30
    at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  [cause]: TypeError [ERR_INVALID_ARG_TYPE]: The "stream" argument must be an instance of Stream. Received an instance of ReadableStream
      at new NodeError (node:internal/errors:400:5)
      at eos (node:internal/streams/end-of-stream:66:11)
      at fetchFinale (/Users/trentm/src/undici/lib/web/fetch/index.js:1093:5)
      at mainFetch (/Users/trentm/src/undici/lib/web/fetch/index.js:760:5)
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
[...]

PR #3125, opened 2024-04-15
drop node support for < v18.17.0

KhafraDev:
Refs: #3123

Since the switch to borp, our test suite no longer functions on anything below v18.17.0 (according to the commenter). If we drop this, we can remove our custom File implementation with no breaking changes.


mcollina (approved):
lgtm

Uzlopak (approved):
LGTM

PR #3127, opened 2024-04-16
fix: retry handler

ronag:
Clean up and fix various issues...

### DIFF
--- a/lib/handler/retry-handler.js
+++ b/lib/handler/retry-handler.js
@@ -1,19 +1,30 @@
 'use strict'
 const assert = require('node:assert')
 
-const { kRetryHandlerDefaultRetry } = require('../core/symbols')
-const { RequestRetryError } = require('../core/errors')
+const DecoratorHandler = require('./decorator-handler')
 const { isDisturbed, parseHeaders, parseRangeHeader } = require('../core/util')
+const { ResponseStatusCodeError } = require('../core/errors')
 
-function calculateRetryAfterHeader (retryAfter) {
-  const current = Date.now()
-  const diff = new Date(retryAfter).getTime() - current
+class RetryHandler extends DecoratorHandler {
+  #dispatch
+  #handler
+  #opts
+  #abort
 
-  return diff
-}
+  #retryOpts
+  #retryCount = -1
+  #retryCountCheckpoint = 0
 
-class RetryHandler {
-  constructor (opts, handlers) {
+  #headersSent = false
+  #aborted = false
+  #reason = null
+  #pos
+  #end
+  #etag
+
+  constructor (opts, { dispatch, handler }) {
+    super(handler)
+
     const { retryOptions, ...dispatchOpts } = opts
     const {
       // Retry scoped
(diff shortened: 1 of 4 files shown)


lib/handler/retry-handler.js
```diff
@@ -29,13 +40,13 @@ class RetryHandler {
...
-    this.dispatch = handlers.dispatch
-    this.handler = handlers.handler
-    this.opts = dispatchOpts
-    this.abort = null
-    this.aborted = false
-    this.retryOpts = {
-      retry: retryFn ?? RetryHandler[kRetryHandlerDefaultRetry],
+    this.#opts = dispatchOpts // TODO (fix): This assumes that opts are not mutated...
+    this.#dispatch = dispatch
+    this.#handler = handler
```
mcollina, line 45: I would recommend not doing this, as functions could be in there too. Is there a need?

mcollina:
What issues did you fix? Could you add a test for them?

lib/handler/retry-handler.js
ronag: Sorry, I don't understand.

ronag:
> What issues did you fix? Could you add a test for them?

No bandwidth. Maybe in future.

ronag:
@metcoder95 feel free to have a look if you see something you are interested in

Uzlopak:
Can you atleast name the bugs?

lib/handler/retry-handler.js
```diff
@@ -29,13 +40,13 @@ class RetryHandler {
...
+    this.#abort = null
+    this.#aborted = false
+    this.#retryOpts = {
+      retry: retryFn ?? defaultRetry,
```
metcoder95, line 49: This one I'm pretty sure is needed, leftover from initial ideas of it

lib/handler/retry-handler.js
```diff
@@ -59,292 +70,227 @@ class RetryHandler {
...
-
-    // Checkpoint for resume from where we left it
-    if (this.resume != null) {
-      this.resume = null
-
-      if (statusCode !== 206) {
-        return true
+      if (this.#retryOpts.statusCodes.includes(statusCode)) {
+        // TODO (fix): Remove this path and delegate to different interceptor which will read the error body?
+        const message = `Response status code ${statusCode}${statusMessage ? `: ${statusMessage}` : ''}`
```
metcoder95, line 100: Curious about this one, what did you have in mind @ronag?

lib/handler/retry-handler.js
```diff
@@ -59,292 +70,227 @@ class RetryHandler {
...
+        throw new ResponseStatusCodeError(message, statusCode, headers)
```
metcoder95, line 101: Why throw instead of abort?

lib/handler/retry-handler.js
```diff
@@ -59,292 +70,227 @@ class RetryHandler {
...
+      return this.#handler.onHeaders(statusCode, rawHeaders, resume, statusMessage)
+    } else if (this.#reason == null) {
+      assert(this.#etag = null)
+      assert(this.#pos == null)
+      assert(this.#end == null)
+      assert(this.#headersSent === false)
+
+      if (headers.trailer) {
+        this.#headersSent = true
+        return this.#handler.onHeaders(statusCode, rawHeaders, resume, statusMessage)
```
metcoder95, line 114: Never thought about the `trailers`, basically if trailers are hinted we just forward do not retry?

Shall we add some flag for implementers to maybe bypass it?
I see the benefit on this

metcoder95 (commented):
I'll definitely take the private properties from the PR 👍 
Following, it seems that you found a bug where the `onHeaders` were called twice? Is that correct?
If also, did you find something else and a way to reproduce it?

Thanks for the hints, I'll definitely take some improvements from here and add it maybe later this week 👍

lib/handler/retry-handler.js
ronag: easier to follow the flow

lib/handler/retry-handler.js
ronag: Should propagate similar to throwOnError

ronag:
Sorry, I'm super stressed to fix this in production.

PR #3138, opened 2024-04-19
perf: use class instead of object literals with getters

tsctx:
Slow initialization of object literals with getters.

- main
```
benchmark               time (avg)             (min … max)       p75       p99      p999
---------------------------------------------------------- -----------------------------
new Request(input)  12'398 ns/iter   (8'100 ns … 4'134 µs) 11'000 ns 34'100 ns    147 µs
```

- this patch

```
benchmark               time (avg)             (min … max)       p75       p99      p999
---------------------------------------------------------- -----------------------------
new Request(input)  10'997 ns/iter   (7'000 ns … 3'704 µs)  9'700 ns 34'000 ns    159 µs
```

### DIFF
--- a/lib/web/fetch/request.js
+++ b/lib/web/fetch/request.js
@@ -41,6 +41,21 @@ const requestFinalizer = new FinalizationRegistry(({ signal, abort }) => {
 
 let patchMethodWarning = false
 
+class RequestRealmSettingsObject {
+  baseUrl = getGlobalOrigin()
+
+  get origin () {
+    return this.baseUrl?.origin
+  }
+
+  policyContainer = makePolicyContainer()
+
+  static {
+    delete this.prototype.constructor
+    Object.freeze(this.prototype)
+  }
+}
+
 // https://fetch.spec.whatwg.org/#request-class
 class Request {
   // https://fetch.spec.whatwg.org/#dom-request
@@ -56,13 +71,9 @@ class Request {
 
     // https://html.spec.whatwg.org/multipage/webappapis.html#environment-settings-object
     this[kRealm] = {
-      settingsObject: {
-        baseUrl: getGlobalOrigin(),
-        get origin () {
-          return this.baseUrl?.origin
-        },
-        policyContainer: makePolicyContainer()
-      }
+      // Note: Slow initialization of object literals with getters.
+      // TODO: Remove workaround.
+      settingsObject: new RequestRealmSettingsObject()
     }
 
     // 1. Let request be null.
--- a/lib/web/fetch/request.js
+++ b/lib/web/fetch/request.js
@@ -373,7 +373,6 @@ class Request {
     // (https://dom.spec.whatwg.org/#dom-abortsignal-any)
     const ac = new AbortController()
     this[kSignal] = ac.signal
-    this[kSignal][kRealm] = this[kRealm]
 
     // 29. If signal is not null, then make this’s signal follow signal.
     if (signal != null) {
@@ -443,7 +442,6 @@ class Request {
     this[kHeaders] = new Headers(kConstruct)
     this[kHeaders][kHeadersList] = request.headersList
     this[kHeaders][kGuard] = 'request'
-    this[kHeaders][kRealm] = this[kRealm]
 
     // 31. If this’s request’s mode is "no-cors", then:
     if (mode === 'no-cors') {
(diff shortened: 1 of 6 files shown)


mcollina (approved):
lgtm

Uzlopak:
Did you also see that makePolicyContainer() seemed to be super slow?

KhafraDev (changes requested):
This should be done for all classes (there is no such thing as a "Request Realm Settings Object", there *is* an "Environment Settings Object").

KhafraDev:
> Did you also see that makePolicyContainer() seemed to be super slow?

It returns an object with a single property. If you're running creating a hundred thousand Requests everything will seem to be "super slow".

mcollina:
There is nothing absolutely “slow” or “fast”, as those are relative concepts. The only useful measure when doing optimization work is to identify bottlenecks and remove them.

(A tiny fast function called a bazillion times might be causing a bottleneck as much as a huge slow function seldomly called).

KhafraDev:
not in this context, it's called once per Request and is needed for spec compatibility

tsctx:
> not in this context, it's called once per Request and is needed for spec compatibility

What do you mean by that?  I've already created a new context for each call.

KhafraDev:
I was responding to Matteo's comment, nothing to do with this PR

KhafraDev (changes requested):
Response has a realm also, which is an Environment Settings Object.

We're also adding a realm to Headers in some cases but it's unused, should probably be removed.

tsctx:
@KhafraDev ptal

lib/web/fetch/request.js
```diff
@@ -55,16 +69,9 @@ class Request {
...
-        baseUrl: getGlobalOrigin(),
-        get origin () {
-          return this.baseUrl?.origin
-        },
-        policyContainer: makePolicyContainer()
-      }
-    }
-
+    // Note: Slow initialization of object literals with getters.
+    // TODO: Remove workaround.
```
tsctx, line 73:
```suggestion
```

KhafraDev:
EnvironmentSettingsObject should be shared between Response and Request, it's meant to be the same object. It's fine to lazily create the properties needed for each.

tsctx:
> I guess that can lead to further perf improvements, the `realm` should be the exact same object and it can't change because they're non-transferable and cannot be cloned. The issue there being that the global origin may be mutated, but that can be a getter while policyContainer and whatever else is constant?

If we use getter, there should be no problem.

KhafraDev (approved):

KhafraDev (changes requested):
Sorry, it seems like baseUrl is only set once. It can be set whenever so it should update appropriately.

lib/web/fetch/util.js
```diff
@@ -1569,6 +1569,28 @@ function utf8DecodeBytes (buffer) {
...
+class EnvironmentSettingsObjectBase {
+  #baseUrl
+
+  get baseUrl () {
+    return (this.#baseUrl ??= getGlobalOrigin())
+  }
+
+  set baseUrl (baseUrl) {
+    this.#baseUrl = baseUrl
+  }
```
tsctx, line 1581:
```suggestion
  baseUrl =  getGlobalOrigin()
```

lib/web/fetch/util.js
```diff
@@ -1569,6 +1569,20 @@ function utf8DecodeBytes (buffer) {
   return output
 }
 
+class EnvironmentSettingsObjectBase {
+  baseUrl =  getGlobalOrigin()
```
tsctx, line 1573:
```suggestion
  baseUrl = getGlobalOrigin()
```

KhafraDev:
I don't think that those changes will solve the issue, consider the following:

```mjs
import { setGlobalOrigin, Response } from 'undici'

new Request('/').url // http://localhost:3000

setGlobalOrigin('https://www.google.com')
new Request('/').url // https://www.google.com
```

If this works, please add a test so we don't regress in the future.

tsctx:
> I don't think that those changes will solve the issue, consider the following:
> 
> ```js
> import { setGlobalOrigin, Request } from 'undici'
> 
> setGlobalOrigin('http://localhost:3000')
> new Request('/').url // http://localhost:3000
> 
> setGlobalOrigin('https://www.google.com')
> new Request('/').url // https://www.google.com
> ```
> 
> If this works, please add a test so we don't regress in the future.

It works correctly before and after the change.

KhafraDev (approved):
I see why, didn't realize a new object was being created every time. We should be able to reuse a single instance of `EnvironmentSettingsObject` instead of creating a new realm per Request/Response. This PR is fine to land as-is, I've already made you do enough...

PR #3139, opened 2024-04-19
feat: add lookup option to Client, Pool and Agent (#2440)

hjr3:
We add an optional custom lookup function to that has the same interface as the lookup option in TCP/UDP socket.connect. This allows users of this library to implement their own DNS caching implementation.

This does not preclude the future possibility of including a DNS caching implementation by default.

## This relates to...

- https://github.com/nodejs/undici/issues/2440

## Rationale

PR https://github.com/nodejs/undici/pull/2552 is attempting to add a default dns caching implementation to undici. Instead of trying to first add a default implementation, which is a lot of work, this PR adds the standard `lookup` option which allows a user to bring their own DNS lookup implementation.

I believe this is a better place to start from because:

- avoids committing to a default DNS caching implementation
   - there is no one-size DNS caching implementation that fits all use cases
- users may have other DNS implementation requirement beyond caching
- this is a smaller change to the undici public interface
- this follows existing Node.js convention of providing a `lookup` option

Adding DNS caching is now as easy as:

```js
import CacheableLookup from 'cacheable-lookup';

const cacheable = new CacheableLookup();

new Agent({
  lookup: cacheable.lookup
});
```

### Features

- adds an optional `lookup` option to `Client`, `Pool` and `Agent`

### Bug Fixes

N/A

N/A

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/docs/docs/api/Client.md
+++ b/docs/docs/api/Client.md
@@ -42,6 +42,7 @@ Every Tls option, see [here](https://nodejs.org/api/tls.html#tls_tls_connect_opt
 Furthermore, the following options can be passed:
 
 * **socketPath** `string | null` (optional) - Default: `null` - An IPC endpoint, either Unix domain socket or Windows named pipe.
+* **lookup** `Function` (optional) - Custom lookup function. Default: [dns.lookup()](https://nodejs.org/api/dns.html#dnslookuphostname-options-callback).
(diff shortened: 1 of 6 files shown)


ronag (changes requested):
This is already possible. Pass lookup in the connect option.

hjr3:
For anyone else wading through the docs:

```js
import * as dns from 'dns';


const agent = new Agent({
  connect: {
    lookup: dns.lookup,
  },
});
```

PR #3142, opened 2024-04-21
reuse realm for Request/Response

KhafraDev:
This change has a slight speedup, but this is mostly for correctness. Having a new "realm" (which is meant to be a distinct global environment) or settings object per Request/Response doesn't make sense.


metcoder95 (approved):

mcollina:
BTW, I think we can remove the property altogether and save a few bytes per object.

PR #3144, opened 2024-04-21
feat: add support for if-match on retry handler

### LINKED ISSUE #3128: Retry handler should send if-match on retry request

metcoder95:
## This relates to...

Closes #3128 

## Changes

- **feat: add support for ifmatch over retry-handler**
- **feat: differentiate between weak and strong etags**

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready


lib/handler/retry-handler.js
```diff
@@ -36,6 +35,7 @@ class RetryHandler {
     this.aborted = false
     this.retryOpts = {
       retry: retryFn ?? RetryHandler[kRetryHandlerDefaultRetry],
+      ifMatch: ifMatch ?? true,
```
ronag, line 38: Why option?

ronag (commented):
Why even bother saving etag if it's weak? Also why differentiate? When are weak etags not good enough?

metcoder95:
> Why even bother saving etag if it's weak? Also why differentiate? When are weak etags not good enough?

Overall, weak etags are not meant to be cheaper to generate, meaning that the content can be considered similar but not equal on a byte-per-byte basis.

> From MDN
>The comparison with the stored [ETag](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag) uses the strong comparison algorithm, meaning two files are considered identical byte by byte only. If a listed ETag has the W/ prefix indicating a weak entity tag, this comparison algorithm will never match it.

Within an `if-match` header, if the header is applied accordingly to the spec, a weak etag might never match failing the pre-condition.

I do get your point of dropping the `etag` if weak, specially if not meant for stating a byte-per-byte match. Happy to do that 👍

lib/handler/retry-handler.js
metcoder95: To allow disable the `if-match` if not interested into using it; just to give the opportunity to disable the request conditional.

mcollina (approved):
lgtm

docs/docs/api/RetryHandler.md
```diff
@@ -29,6 +29,7 @@ Extends: [`Dispatch.DispatchOptions`](Dispatcher.md#parameter-dispatchoptions).
 - **methods** `string[]` (optional) - Array of HTTP methods to retry. Default: `['GET', 'PUT', 'HEAD', 'OPTIONS', 'DELETE']`
 - **statusCodes** `number[]` (optional) - Array of HTTP status codes to retry. Default: `[429, 500, 502, 503, 504]`
 - **errorCodes** `string[]` (optional) - Array of Error codes to retry. Default: `['ECONNRESET', 'ECONNREFUSED', 'ENOTFOUND', 'ENETDOWN','ENETUNREACH', 'EHOSTDOWN', 'UND_ERR_SOCKET']`
+- **ifMatch** `boolean` (optional) - It enables the send of `if-match` pre-condition header if `etag` is present in the first response. Default: `true`
```
ronag, line 32: Why would ever *not* send this?

metcoder95: I'd say if you are not interested in a `pre-condition` check; but, cannot think of a use-case right now. 

Yeah, let's drop it and reconsider if a use case comes in

ronag (approved):

Issue #3141, opened 2024-04-19
Race condition at-or-near keep-alive expiration


mweberxyz:
## Bug Description

I've managed to (fairly) reliably replicate an error when a request is made right around the time a connection is approaching it's keep alive timeout. I have seen it happen in both:

- Node 20
- Node 21

using each of the following clients:

- Node-native fetch
- Undici 6.13.0
- Node http.get

## Reproducible By

Run server.js:
```
const { createServer } = require("node:http");

const server = createServer(
  { keepAlive: true, keepAliveTimeout: 5000 },
  (req, res) => {
    res.writeHead(200, { "Content-Type": "application/json" });
    res.end(
      JSON.stringify({
        data: "Hello World!",
      })
    );
  }
);

server.listen(3000, async () => {
  const endpoint = `http://localhost:${server.address().port}`;
  console.log(endpoint);
});
```

Then run `client-undici.js`:
```
const undici = require("undici");

let timeout = 4993;
const start = async () => {
  timeout += 0.5;
  console.log({ timeout });

  const res = await undici.fetch("http://127.0.0.1:3000");
  await res.text();
  setTimeout(start, timeout);
};

start();
```
OR `client-http.js`:
```
const http = require("http");
let timeout = 4995;
const start = async () => {
  timeout += 0.5;
  console.log({ timeout });

  http.get("http://127.0.0.1:3000", (res) => {
    res.on("data", () => {});
    res.on("end", () => {
      setTimeout(start, timeout);
    });
  });
};

start();
```

A few runs might need to be tried, and the starting `let timeout = ...` might need to be adjusted a bit.. but give it a few tries and eventually you'll see a crash like:

```
      Error.captureStackTrace(err, this)
            ^

TypeError: fetch failed
    at Object.fetch (/Users/matt/code/replicate-undici-3133/node_modules/undici/index.js:109:13)
    at async Timeout.start [as _onTimeout] (/Users/matt/code/replicate-undici-3133/client.js:8:15) {
  [cause]: Error: read ECONNRESET
      at TCP.onStreamRead (node:internal/stream_base_commons:217:20) {
    errno: -54,
    code: 'ECONNRESET',
    syscall: 'read'
  }
}

Node.js v21.7.3
```

To speed it finding the sweet spot, run with `NODE_DEBUG=net`, if you see the socket being torn down, then the connection being re-established successfully, then the timeout needs to be reduced.

## Expected Behavior

Node does not crash when making a request near keep alive expiration.

[...]

Issue #3133, opened 2024-04-17
Observations regarding instability of native fetch in AWS Lambda x86_64 nodejs20.x


jandppw:
We observe consistent intermittent errors when using native `fetch` in an _AWS Lambda `x86_64` `nodejs20.x`_. We were
unable to find the particular place in `undici` code that is the culprit of this behavior.

The intention of this submission is to report the issues and to record our observations as accurately as possible, to enable
people with better detailed knowledge to isolate and possibly fix the issues, if they are not already fixed (`fetch` is still
experimental in Node 20).

## Stable precursor

Versions of this service have been in production since 2022-05-19 on _AWS Lambda x86_64 nodejs16.x_ using
[`node-fetch@2.6.7`](https://www.npmjs.com/package/node-fetch) without many issues. Notably, in the last 6 months of
2023 exactly 1 communication error calling the remote API was logged. It handles ~700 invocations per day.

## Where are the issues observed?

The functionality of the service requires it to do ±25 remote API calls per invocation. The remote API is hosted on a
Windows VM through IIS 8, and is accessed through a Cloudfront distribution, setup to use `http2and3`.

We observe this issue, or these issues, only in _AWS Lambda x86_64 nodejs20.x_, and have not observed the issue during
development on macOS (This does not guarantee that the issues are limited to _AWS Lambda x86_64 nodejs20.x_, but that it
is _possible_ that the issues are limited to _AWS Lambda x86_64 nodejs20.x_).

A new version of the code was released at 2024-03-25, but this did not have any impact on the issues. It seems _AWS
Lambda x86_64 nodejs20.x_ changed Node version at 2024-04-03. We cannot report on the exact Node version (and thus
`undici` version) because sadly we do not log that. Given that the latest LTS version of node uses `undici@5.28.4`, this
reports on versions `undici@<=5.28.4`. The issues might have been fixed already in later releases, although we don’t see
any relevant mentioning of it in the [release notes](https://github.com/nodejs/undici/releases) since v5.25.1. Between
2024-02-06 and 2024-04-03 the errors are reported on `node:internal/deps/undici/undici:12344:11`. Between 2024-04-03 and
2024-04-15 the errors are reported on `node:internal/deps/undici/undici:12345:11`.

## ~1% of invocations fail with 4 ≠ types of failures

The version of this service that shows the issues has been in production since 2024-02-06. It is deployed on _AWS Lambda
[...]

PR #3145, opened 2024-04-21
fix: tweak keep-alive timeout implementation

mweberxyz:
Closes nodejs/undici#3141

## This relates to...

- https://github.com/nodejs/undici/issues/3141

## Rationale

The goal of the util/timers.js is to provide setTimeout-compatible timers with 1 second granularity, as opposed to native setTimeout's 1ms. Because the timers were scheduled, then executed in two separate "ticks" of the fast timeout, this meant timeouts are called best case 1000ms after intended, and worst case 2000ms. Combined with the previous 1000ms default for `keepAliveTimeoutThreshold`, the end result is an effective keepAliveTimeoutThreshold of 0 -- so unluckily timed requests would attempt to use a connection which has already been closed by the remote.

## Changes

Multi-pronged approach here to prioritize safety and avoid attempting to re-use sockets that have already been closed:

1. Increase default keepAliveTimeoutThreshold from 1000 to 2000
2. Determine timeout times immediately upon timeout registration, instead of in the next tick of fast timer
3. Double fast timer tick rate
4. Timer tests added: see test file comments for approach details

Note to @ronag I went with change 2 instead of the proposed subtraction of the tick time from the duration because after writing the tests, that approach was more flaky.

### Features

N/A

### Bug Fixes

Some of the reports in https://github.com/nodejs/undici/issues/3133 _may_ be resolved by this change, though it would need to be backported to 5.x, included in a Node v20 release, and then picked up by Amazon before we would see any improvement. ¯\\\_(ツ)_/¯

The more conservative default keepAliveTimeoutThreshold value should not be a breaking change.

## Benchmarks
Within run-to-run variance:
 
PR:
```
┌─────────┬─────────────────────┬─────────┬────────────────────┬────────────┬─────────────────────────┐
│ (index) │ Tests               │ Samples │ Result             │ Tolerance  │ Difference with slowest │
├─────────┼─────────────────────┼─────────┼────────────────────┼────────────┼─────────────────────────┤
│ 0       │ 'undici - fetch'    │ 45      │ '16454.53 req/sec' │ '± 2.74 %' │ '-'                     │
│ 1       │ 'undici - pipeline' │ 25      │ '23976.66 req/sec' │ '± 2.61 %' │ '+ 45.71 %'             │
│ 2       │ 'undici - stream'   │ 15      │ '31539.92 req/sec' │ '± 2.94 %' │ '+ 91.68 %'             │
│ 3       │ 'undici - request'  │ 40      │ '31627.55 req/sec' │ '± 2.98 %' │ '+ 92.21 %'             │
│ 4       │ 'undici - dispatch' │ 10      │ '36186.91 req/sec' │ '± 2.54 %' │ '+ 119.92 %'            │
└─────────┴─────────────────────┴─────────┴────────────────────┴────────────┴─────────────────────────┘
```
`main`:
```
┌─────────┬─────────────────────┬─────────┬────────────────────┬────────────┬─────────────────────────┐
│ (index) │ Tests               │ Samples │ Result             │ Tolerance  │ Difference with slowest │
├─────────┼─────────────────────┼─────────┼────────────────────┼────────────┼─────────────────────────┤
│ 0       │ 'undici - fetch'    │ 40      │ '17433.63 req/sec' │ '± 2.88 %' │ '-'                     │
│ 1       │ 'undici - pipeline' │ 30      │ '23934.65 req/sec' │ '± 3.00 %' │ '+ 37.29 %'             │
│ 2       │ 'undici - request'  │ 101     │ '29557.97 req/sec' │ '± 3.88 %' │ '+ 69.55 %'             │
│ 3       │ 'undici - dispatch' │ 20      │ '32711.63 req/sec' │ '± 2.60 %' │ '+ 87.64 %'             │
│ 4       │ 'undici - stream'   │ 15      │ '34008.68 req/sec' │ '± 2.35 %' │ '+ 95.08 %'             │ 
└─────────┴─────────────────────┴─────────┴────────────────────┴────────────┴─────────────────────────┘
```

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready


mweberxyz:
The results of the replication used in tracking down the issue, after this PR changes:

```
% node test-latency.js
target = 1000, actual = 1003
target = 1500, actual = 1520
target = 1990, actual = 2022
target = 1999, actual = 2022
target = 2000, actual = 2022
target = 2001, actual = 2022
target = 2500, actual = 2524
target = 3000, actual = 3026
target = 2990, actual = 3026
target = 2999, actual = 3028
target = 4000, actual = 4030
target = 5000, actual = 5032
10s elapsed, exiting
```

lib/util/timers.js
```diff
@@ -73,7 +73,7 @@ class Timeout {
       }
     }
 
-    this.state = 0
+    this.state = Date.now() + this.delay
```
ronag, line 76: Can we avoid this?

ronag (changes requested):

ronag (changes requested):
LGTM other than calling `Date.now()` on every registration.

mcollina:
CI is really red.

I don't think there is literally anything we can do for #3133 without the intervention of AWS. Let's fix the race condition anyway.

undici v6 will land on v20 soonish, so let's get this done ;).

ronag (changes requested):
I think the only thing you need to change is the interval to 500 ms. Maybe the 2s default also makes sense.

mweberxyz:
Tests should go green - that test was impacted by the change to the default.

mcollina (approved):
lgtm

mweberxyz:
@ronag ptal. I tried this strategy at first, but it lead to a slight amount of non-determinism when a timer is requested that is an exact multiple of the tick rate (such as: `test/request-timeout.js:31`). Setting TICK_MS to 499 appears to be enough to make test runs pass consistently.

lib/util/timers.js
mweberxyz: Yes, see newest commit.

mweberxyz:
Ah, here is an example of the loss of precision (even with the TICK_MS set to 499), two consecutive runs of the replication from the issue:

```
% node test-latency.js
target = 1000, actual = 1001
target = 1500, actual = 1516
target = 1990, actual = 2016
target = 1999, actual = 2017
target = 2000, actual = 2017
target = 2001, actual = 2017
target = 2500, actual = 2517
target = 3000, actual = 3017
target = 2990, actual = 3018
target = 2999, actual = 3018
target = 4000, actual = 4018
target = 5000, actual = 5019
```

followed by
```
node test-latency.js
target = 1000, actual = 1501
target = 1500, actual = 1519
target = 1990, actual = 2020
target = 1999, actual = 2020
target = 2000, actual = 2020
target = 2001, actual = 2020
target = 2500, actual = 2521
target = 3000, actual = 3021
target = 2990, actual = 3022
target = 2999, actual = 3022
target = 4000, actual = 4022
target = 5000, actual = 5022
```

Note the 1000ms timer fired much later the second run.

To help with very-short timeouts, I could keep the fallback to regular setTimeout at 1000ms as it was previously? Or do nothing here. Let me know :)

ronag (approved):

mweberxyz:
@mcollina @ronag I would be more comfortable increasing the minimum timeout to use the fast timers implementation to >1000ms, okay to push that small tweak?

ronag (approved):

PR #3146, opened 2024-04-22
don't store realm on Request/Response

KhafraDev:
Refs: https://github.com/nodejs/undici/pull/3142#issuecomment-2068066669


KhafraDev:
@Uzlopak I cleaned up some things in EventSource, but I mostly split settings from state. PTAL

mcollina (approved):
lgtm

PR #3150, opened 2024-04-22
do not reinitalize HeadersList

mcollina:


anonrig (approved):
Nice stream. LGTM.

ronag (changes requested):
I think the reason this is done to make sure to have a unique copy that the caller can't mutate after passing it.

KhafraDev (commented):
@ronag the underlying map gets copied anyways

but I'm wondering what the difference is between what we're already doing? I couldn't stay for the whole stream, sorry. https://github.com/nodejs/undici/blob/9458ffd4f1e5116cdcf8b421a3171d3aded41d46/lib/web/fetch/headers.js#L133-L139

mcollina:
> I think the reason this is done to make sure to have a unique copy that the caller can't mutate after passing it.

@ronag I think we are the caller. No tests or WPT are failing with this change.

Uzlopak:
Funny, i have a similar change stashed locally.
Instead of instanceof we should use a symbolattribute and check for it?

ronag (approved):

tsctx:
Looks like it needs to be copied. 
https://github.com/nodejs/undici/blob/a8551ad17ead3f35ad9757be090725860f054171/lib/web/fetch/request.js#L140-L141

KhafraDev:
@tsctx good catch, I had to dig very far into the spec to figure out what "A copy of" meant here. That line was added/modified nearly 6 years ago in [this commit](https://github.com/whatwg/fetch/commit/819d8c9d6617986a831ecd9cf21c34ba9589a890), but comments as to why the headerslist were removed since then:

```html
  <p class=note>The headers are sanitised as they might contain headers that are not allowed by this
  mode. Otherwise, they were previously sanitised or are unmodified since creation by a privileged
  API.
```

It appears that when, or if, the internal request mode is modified, some previously allowed headers may be forbidden. Not only do we not implement forbidden headers, but the current code wouldn't satisfy this requirement regardless.

A more correct fix should be what I mentioned previously: https://github.com/nodejs/undici/pull/3150#issuecomment-2070875768, but if that can't be done I don't think there's anything wrong with landing this.

tsctx:
This is also...

```js
import { Request } from "./index.js";

const request = new Request("https://localhost", {
  headers: {
    a: "a",
    b: "b",
  },
});

const childRequest = new Request(request);

[...childRequest.headers.keys()].forEach((key) => childRequest.headers.delete(key));

console.log([...request.headers], [...childRequest.headers]); 
// expected: [ [ 'a', 'a' ], [ 'b', 'b' ] ] []
// actual: [] []
```

KhafraDev:
Makes sense, I'm surprised the tests didn't catch that.

tsctx (changes requested):
No copying can be avoided.

KhafraDev:
@tsctx could you make a PR that tests this so we don't regress?

mcollina:
Closing for now, will revisit later.

(if anybody beats me to it.. go for it!)

PR #3158, opened 2024-04-24
Do not leak signal listeners

### LINKED ISSUE #3157: Long-lived AbortSignals and undici cause MaxListenersExceededWarnings
### Version

v20.11.0

### Platform

Darwin MacBook-Pro-194.localdomain 23.4.0 Darwin Kernel Version 23.4.0: Fri Mar 15 00:10:42 PDT 2024; root:xnu-10063.101.17~1/RELEASE_ARM64_T6000 arm64

### Subsystem

undici

Making multiple `fetch` requests with the same `AbortSignal` causes `MaxListenersExceededWarning`s to appear in the console with a high number of listeners.

If `undici` is adding event listeners without removing them, this is a memory leak.

```js
import http from 'http'

const PORT = 49823

  res.writeHead(200, {'Content-Type': 'text/plain'})
  res.write('Hello World!')
  res.end()
}).listen(PORT)

const controller = new AbortController()

for (let i = 0; i < 200; i++) {
  // make request
  const res = await fetch(`http://localhost:${PORT}`, {
    signal: controller.signal
  })

  // drain body
  await res.text()

  // in theory any added event listeners added by undici should/could be removed
  // now as the request has finished so there's nothing left to abort.
}
```

There is a repro repo here: https://github.com/achingbrain/node-fetch-maxlistenersexceededwarning

Every time.  No required condition.

Undici should remove any added event listeners when there is nothing left to abort (e.g. the request has ended and so either errored or the body has been consumed/cancelled).

### What do you see instead?

```
(node:37396) MaxListenersExceededWarning: Possible EventTarget memory leak detected. 101 abort listeners added to [AbortSignal]. Use events.setMaxListeners() to increase limit
(Use `node --trace-warnings ...` to show where the warning was created)
(node:37396) MaxListenersExceededWarning: Possible EventTarget memory leak detected. 102 abort listeners added to [AbortSignal]. Use events.setMaxListeners() to increase limit
(node:37396) MaxListenersExceededWarning: Possible EventTarget memory leak detected. 103 abort listeners added to [AbortSignal]. Use events.setMaxListeners() to increase limit
...etc
```

### Additional information

A workaround appears to be to create a request-specific signal and abort it if the long lived signal aborts, taking care to remove any added listeners on the long-lived signal when we are done with the request.

See: https://github.com/achingbrain/node-fetch-maxlistenersexceededwarning/blob/main/workaround.js

This may be a duplicate of https://github.com/nodejs/node/issues/52203 - please close this issue if you consider it so, but the workaround above may be of use to someone if there's no way to get `undici` to clean up after itself.

mcollina:
Fixes #3157

### DIFF
--- a/lib/web/fetch/request.js
+++ b/lib/web/fetch/request.js
@@ -38,6 +38,29 @@ const requestFinalizer = new FinalizationRegistry(({ signal, abort }) => {
   signal.removeEventListener('abort', abort)
 })
 
+function buildAbort (acRef) {
+  return abort
+
+  function abort () {
+    const ac = acRef.deref()
+    if (ac !== undefined) {
+      // Currently, there is a problem with FinalizationRegistry.
+      // https://github.com/nodejs/node/issues/49344
+      // https://github.com/nodejs/node/issues/47748
+      // In the case of abort, the first step is to unregister from it.
+      // If the controller can refer to it, it is still registered.
+      // It will be removed in the future.
+      requestFinalizer.unregister(abort)
+
+      // Unsubscribe a listener.
+      // FinalizationRegistry will no longer be called, so this must be done.
+      this.removeEventListener('abort', abort)
+
+      ac.abort(this.reason)
+    }
+  }
+}
+
 let patchMethodWarning = false
 
 // https://fetch.spec.whatwg.org/#request-class
@@ -377,24 +400,7 @@ class Request {
         this[kAbortController] = ac
 
         const acRef = new WeakRef(ac)
-        const abort = function () {
-          const ac = acRef.deref()
-          if (ac !== undefined) {
-            // Currently, there is a problem with FinalizationRegistry.
-            // https://github.com/nodejs/node/issues/49344
-            // https://github.com/nodejs/node/issues/47748
-            // In the case of abort, the first step is to unregister from it.
-            // If the controller can refer to it, it is still registered.
-            // It will be removed in the future.
-            requestFinalizer.unregister(abort)
-
-            // Unsubscribe a listener.
-            // FinalizationRegistry will no longer be called, so this must be done.
-            this.removeEventListener('abort', abort)
-
-            ac.abort(this.reason)
-          }
-        }
+        const abort = buildAbort(acRef)
 
         // Third-party AbortControllers may not work with these.
         // See, https://github.com/nodejs/undici/pull/1910#issuecomment-1464495619.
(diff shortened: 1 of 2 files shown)


lib/web/fetch/request.js
```diff
@@ -377,34 +400,17 @@ class Request {
...
+        const abort = buildAbort(acRef)
 
         // Third-party AbortControllers may not work with these.
         // See, https://github.com/nodejs/undici/pull/1910#issuecomment-1464495619.
         try {
           // If the max amount of listeners is equal to the default, increase it
           // This is only available in node >= v19.9.0
           if (typeof getMaxListeners === 'function' && getMaxListeners(signal) === defaultMaxListeners) {
-            setMaxListeners(100, signal)
+            setMaxListeners(800, signal)
```
Uzlopak, line 411: why do you increse too 800?

lib/web/fetch/request.js
```diff
@@ -377,34 +400,17 @@ class Request {
...
-            requestFinalizer.unregister(abort)
-
-            // Unsubscribe a listener.
-            // FinalizationRegistry will no longer be called, so this must be done.
-            this.removeEventListener('abort', abort)
-
-            ac.abort(this.reason)
-          }
-        }
+        const abort = buildAbort(acRef)
```
Uzlopak, line 403: why not put the weakref wrapping into buildAbort?

Uzlopak: like this:
```js
function buildAbort (ac) {
  const acRef = new WeakRef(ac)

  return abort

  function abort () {
    const ac = acRef.deref()
    if (ac !== undefined) {
      // Currently, there is a problem with FinalizationRegistry.
      // https://github.com/nodejs/node/issues/49344
      // https://github.com/nodejs/node/issues/47748
      // In the case of abort, the first step is to unregister from it.
      // If the controller can refer to it, it is still registered.
      // It will be removed in the future.
      requestFinalizer.unregister(abort)

      // Unsubscribe a listener.
      // FinalizationRegistry will no longer be called, so this must be done.
      this.removeEventListener('abort', abort)

      ac.abort(this.reason)
    }
  }
}
```

lib/web/fetch/request.js
mcollina: It's a magic number. Basically "how many can stack up on my system before the gc kicks in and clean them up"

lib/web/fetch/request.js
Uzlopak: k

tsctx, line 403: Why did you make this place a function?

mcollina: because it would keep the ac inside the closure.

ronag (approved):

PR #3160, opened 2024-04-24
fix casing issue when cloning Headers object

KhafraDev:
Refs: https://github.com/nodejs/undici/pull/3159#issuecomment-2075537226

## Status

- [ ] I have read and agreed to the [Developer's Certificate of Origin][cert]
- [ ] Tested
- [ ] Benchmarked (**optional**)
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

[cert]: https://github.com/nodejs/undici/blob/main/CONTRIBUTING.md

PR #3161, opened 2024-04-24
Revert "fix casing issue when cloning Headers object (#3160)"

tsctx:
Refs: https://github.com/nodejs/undici/pull/3160#issuecomment-2075797019


KhafraDev (changes requested):
This should be easily fixable by fixing .entries https://github.com/nodejs/undici/blob/034e3f08009120cdc8e811f7e9d0154062147ccb/lib/web/fetch/headers.js#L250 ?

tsctx:
It's impossible because it's an object. It must be an array.

KhafraDev:
We could probably use kHeadersSortedMap (how the iterators run) instead. The webidl part can return an array too.

KhafraDev:
I'll take a look at it in a few hours, fixing it from my point of view doesn't seem too difficult.

tsctx:
This seems not the best solution.

PR #3166, opened 2024-04-25
dup-keys-non-strings

JavaLavaMT:
## Rationale

now throws an error if the headers key is not a string or duplicated 

## Changes

- added Tests
- added a guard for duplicate headers 
- added a guard for non string header keys

Not sure, but this will block ALL requests that have headers which are not of type 'String' or are duplicated.

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -323,6 +323,10 @@ function processHeader (request, key, val) {
     return
   }
 
+  if (typeof key !== 'string') {
+    throw new InvalidArgumentError('invalid header key, needs to be a "String"')
+  }
+
   let headerName = headerNameLowerCasedRecord[key]
 
   if (headerName === undefined) {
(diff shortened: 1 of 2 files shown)


ronag:
I think duplicate headers are allowed by the spec

ronag:
Actually not sure what is correct here.

James1x0:
@ronag I think duplicates are only list values?
https://www.rfc-editor.org/rfc/rfc9110.html#section-5.2
https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2

JavaLavaMT:
based on https://www.rfc-editor.org/rfc/rfc9110.html#name-field-lines-and-combined-fi

seems like keys CAN be duplicates as long as the values are an array. If i am understanding this correct, commit # [698e64e](https://github.com/nodejs/undici/pull/3166/commits/698e64e50ea4a120562367a07a6e72a0399320bc) should fix this.

metcoder95:
The actual commit still seems to have the implementation that throws on duplicate, can we verify we change it?

PR #3173, opened 2024-04-27
ci: ensure wasm integrity

Uzlopak:
Ensure that the wasm files are not tampered with.

If the llhttp source files or llhttp wasm files are touched, we check also if the wasm files are correct or if they have some tampered binary code. Basically trying to remove the potential attack vector like it was built in into xz. 

We can then close 

https://github.com/nodejs/undici/security/code-scanning/251
https://github.com/nodejs/undici/security/code-scanning/252

If the code is tampered with, you will see it 
https://github.com/nodejs/undici/actions/runs/8915422913/job/24485005620

@mweberxyz
Any Ideas how to improve it?


.github/workflows/nodejs.yml
```diff
@@ -181,6 +181,10 @@ jobs:
       - name: Run typings tests
         run: npm run test:typescript
 
+  test-llhttp-integrity:
+    name: Ensure integrity of llhttp wasm files
+    uses: uzlopak/undici/.github/workflows/llhttp-wasm-integrity.yml@llhttp-integrity
```
Uzlopak, line 186: this has to be modified to use undici main,

.github/workflows/llhttp-wasm-integrity.yml
```diff
@@ -0,0 +1,59 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 2
+      - name: check modified files
+        id: check_files
+        run: |
+          echo "=============== list modified files ==============="
+          git diff --name-only HEAD^ HEAD
```
Uzlopak, line 24: @mweberxyz  can we maybe get somehow the parent commit?

mcollina (approved):
lgtm

metcoder95 (approved):

.github/workflows/nodejs.yml
Uzlopak: before merging, this has to be appied:

```suggestion
    uses: nodejs/undici/.github/workflows/llhttp-wasm-integrity.yml@main
```

Uzlopak:
Please dont merge yet. I really think the solution is not proper yet. `git diff --name-only HEAD^ HEAD` will only process one commit. But I need to compare it with the last common commit (hash).

ronag:
@Uzlopak please make a review with "request changes" to avoid this being merged

Uzlopak:
@mcollina 
@metcoder95 
@mweberxyz 

PTAL. Should be correct now.
Maybe @mweberxyz can have a more strict review.

.github/workflows/nodejs.yml
```diff
@@ -181,6 +181,10 @@ jobs:
       - name: Run typings tests
         run: npm run test:typescript
 
+  test-llhttp-integrity:
+    name: Ensure integrity of llhttp wasm files
+    uses: nodejs/undici/.github/workflows/llhttp-wasm-integrity.yml@main
```
mcollina, line 186: I'm not sure I would run this for every PR. Maybe only if those files were changed?

mcollina (changes requested):

.github/workflows/nodejs.yml
Uzlopak: It is a reusable workflow. It has two steps. First check if wasm related files were changed, if so run the integrity check. Maybe the first step can be further optimized by using https://github.com/tj-actions/changed-files so that we can also take care of cases where we directly push into main. Also makes sense to obligatory run the integrity check in the automated release workflow.

mcollina: Maybe we can just run it in the release workflow then.

Uzlopak: @mcollina

Imho we should ensure that the repo has the correct content at any time. So if somebody tries to upstream malicious code into the repo, then it should be imho detected early and not as the last step of the release workflow. 

The only issue i see, is that people could force push into main and bypass the ci.

mcollina: we can limit that now that we have automated releases, open a separate issue.

Issue #3131, opened 2024-04-16
Memory issues / many instances of MaxListenersExceededWarning when using Undici via Elasticsearch


daveyarwood:
## Bug Description

This is a follow-up to #1711. @KhafraDev asked that I file a separate issue because Elasticsearch does not appear to be using `fetch`, which is what was fixed in that issue (via #2823).

I've been running into what I think is the same or similar issue to what was reported in #1711 - lots of MaxListenersExceededWarnings being logged, and my service eventually runs out of memory and dies. When I run with `--trace-warnings`, the stacktrace points to Undici:

```
(node:14480) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 101 abort listeners added to [EventEmitter]. Use emitter.setMaxListeners() to increase limit
    at _addListener (node:events:588:17)
    at EventEmitter.addListener (node:events:606:10)
    at addAbortListener (/Users/daveyarwood/code/spark/node_modules/undici/lib/core/util.js:443:10)
    at addSignal (/Users/daveyarwood/code/spark/node_modules/undici/lib/api/abort-signal.js:36:3)
    at new RequestHandler (/Users/daveyarwood/code/spark/node_modules/undici/lib/api/api-request.js:66:5)
    at Pool.request (/Users/daveyarwood/code/spark/node_modules/undici/lib/api/api-request.js:171:25)
    at /Users/daveyarwood/code/spark/node_modules/undici/lib/api/api-request.js:164:15
    at new Promise (<anonymous>)
    at Pool.request (/Users/daveyarwood/code/spark/node_modules/undici/lib/api/api-request.js:163:12)
```

This is with Undici 6.13.0, which is a transitive dependency of `@elastic/transport` 8.5.0.

## Reproducible By

I would be happy to provide a minimal repro case if I had any idea how to do that! :)

Please let me know if you have any ideas of things I can test in my setup.

## Expected Behavior

* No MaxListenersExceededWarning messages should be logged.
* In a deployed environment, the memory graphs should not show available memory declining significantly, and the machine shouldn't run out of memory and die.

### Local

2021 Macbook Pro M1 Max
macOS Sonoma 14.11
Node 20.9.0

I'm seeing the MaxListenersExceededWarnings when I run my service locally.

### Deployed

[...]

PR #3174, opened 2024-04-28
fix: don't add listeners to the same signal more than once

tsctx:


ronag:
What is this solving?

tsctx:
#3131

Uzlopak:
But isnt it a bug in the product too, that it tries to add again and again the same listeners? Not saying we should not solve it here. But now it hints to a bug, but after we merge this you dont get a hint that something is wrong.

ronag:
I don't think we should merge this. It doesn't solve an actual bug. It just bypasses the warning.

ronag (changes requested):
Se above

tsctx:
> But isnt it a bug in the elasticsearch client too, that it tries to add again and again the same listeners? 

elasticsearch client is not the cause of the problem, it is undici-specific. In fact, it does not leak.

> Not saying we should not solve it here. But now it hints to a bug, but after we merge this you dont get a hint that something is wrong.

When a signal was given and a large number of requests were sent at once (easy to see in the test), listeners were added more than the limit and a warning was generated.

metcoder95:
> > Not saying we should not solve it here. But now it hints to a bug, but after we merge this you dont get a hint that something is wrong.
> 
> When a signal was given and a large number of requests were sent at once (easy to see in the test), listeners were added more than the limit and a warning was generated.

I am not sure if we should consider that an `undici` problem. This issue is potentially everywhere and it seems we are trying to protect the users from themselves when they should be aware of how they handle their signals and others.
No strong opinion on this, but I'd advise considering if worth the complexity given this is more a problem of how is `AbortSignal` used by the user rather than by `undici`.

tsctx:
Oh, you are right.

PR #3175, opened 2024-04-28
webidl changes

KhafraDev:
- In 'singular' converters, we now default the options to undefined.
- defaultValue (for dictionary converters) is now a function
- We now attach context (a header like 'Headers.has' and the argument name) to each webidl call for better errors.


tsctx (approved):

ronag (approved):

KhafraDev:
Still a little more to go regarding the errors, but they're generally much better & consistent now. Errors regarding sequences are definitely the worst right now.

PR #3185, opened 2024-05-01
fix 3 mimesniff tests

KhafraDev:
I noticed that these tests were failing in node but were passing in Deno. 
https://wpt.fyi/results/mimesniff/mime-types/parsing.any.html?label=master&label=experimental&product=node.js&product=deno&product=chrome&product=firefox&view=subtest

Of course it was an issue with regex.

We now pass 1031/1898 of the tests while Deno "only" passes 1030/1898!


Uzlopak (approved):
LGTM

PR #3200, opened 2024-05-03
test: avoid tempfile generation in request-timeout.js

Uzlopak:
When I run the tests and manually abort them, then sometimes `request-timeout.js.10mb.txt` is left as an artifact. So this PR avoids the generation of this file.

### DIFF
--- a/test/request-timeout.js
+++ b/test/request-timeout.js
@@ -2,7 +2,6 @@
 
 const { tspl } = require('@matteo.collina/tspl')
 const { test, after } = require('node:test')
-const { createReadStream, writeFileSync, unlinkSync } = require('node:fs')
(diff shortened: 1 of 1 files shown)


metcoder95 (commented):
Code LGTM, but I imagine the intention of writing the file was to put the tests cloes-to-real situation where we are reading the file up to  (opening fd, simulating the read between byte ranges and what not).

Why not just remove the file after the tests are done?

Uzlopak:
@metcoder95 

Doesnt seem so, see #1515 ;)

mcollina (changes requested):
Part of the test is to verify that a body coming in chunks would work. You would need to allocate several buffers and add push them
into the stream with a setImmediate().

PR #3204, opened 2024-05-05
websocket: 200x faster generate mask

tsctx:
The disadvantage is increased memory usage.
```
benchmark                  time (avg)             (min … max)       p75       p99      p999
------------------------------------------------------------- -----------------------------
• generate
------------------------------------------------------------- -----------------------------
generateMask            15.69 ns/iter      (9.77 ns … 332 ns)  15.14 ns  72.66 ns    180 ns
crypto.randomBytes(4)   3'874 ns/iter   (2'971 ns … 6'260 ns)  4'528 ns  6'086 ns  6'260 ns

summary for generate
  generateMask
   246.99x faster than crypto.randomBytes(4)
```

### DIFF
--- a/lib/web/websocket/frame.js
+++ b/lib/web/websocket/frame.js
@@ -6,7 +6,7 @@ const BUFFER_SIZE = 16386
 
 /** @type {import('crypto')} */
 let crypto
-const buffer = Buffer.allocUnsafe(BUFFER_SIZE)
+const buffer = null
 let bufIdx = BUFFER_SIZE
 
 try {
--- a/lib/web/websocket/frame.js
+++ b/lib/web/websocket/frame.js
@@ -2,13 +2,29 @@
 
 const { maxUnsigned16Bit } = require('./constants')
 
+const BUFFER_SIZE = 16386
+
 /** @type {import('crypto')} */
 let crypto
+let buffer = null
+let bufIdx = 0
+
 try {
   crypto = require('node:crypto')
 /* c8 ignore next 3 */
 } catch {
 
+}
+
+function generateMask () {
+  if (buffer === null) {
+    buffer = crypto.randomFillSync(Buffer.allocUnsafe(BUFFER_SIZE), 0, BUFFER_SIZE)
+  }
+  if (bufIdx + 4 > BUFFER_SIZE) {
+    bufIdx = 0
+    crypto.randomFillSync(buffer, 0, BUFFER_SIZE)
+  }
+  return [buffer[bufIdx++], buffer[bufIdx++], buffer[bufIdx++], buffer[bufIdx++]]
 }
 
 class WebsocketFrameSend {
@@ -17,11 +33,12 @@ class WebsocketFrameSend {
    */
   constructor (data) {
     this.frameData = data
-    this.maskKey = crypto.randomBytes(4)
   }
 
   createFrame (opcode) {
-    const bodyLength = this.frameData?.byteLength ?? 0
+    const frameData = this.frameData
+    const maskKey = generateMask()
+    const bodyLength = frameData?.byteLength ?? 0
 
     /** @type {number} */
     let payloadLength = bodyLength // 0-125
@@ -43,10 +60,10 @@ class WebsocketFrameSend {
     buffer[0] = (buffer[0] & 0xF0) + opcode // opcode
 
     /*! ws. MIT License. Einar Otto Stangvik <einaros@gmail.com> */
-    buffer[offset - 4] = this.maskKey[0]
-    buffer[offset - 3] = this.maskKey[1]
-    buffer[offset - 2] = this.maskKey[2]
-    buffer[offset - 1] = this.maskKey[3]
+    buffer[offset - 4] = maskKey[0]
+    buffer[offset - 3] = maskKey[1]
+    buffer[offset - 2] = maskKey[2]
+    buffer[offset - 1] = maskKey[3]
 
     buffer[1] = payloadLength
 
@@ -61,8 +78,8 @@ class WebsocketFrameSend {
     buffer[1] |= 0x80 // MASK
 
     // mask body
-    for (let i = 0; i < bodyLength; i++) {
-      buffer[offset + i] = this.frameData[i] ^ this.maskKey[i % 4]
+    for (let i = 0; i < bodyLength; ++i) {
+      buffer[offset + i] = frameData[i] ^ maskKey[i & 3]
     }
 
     return buffer
--- a/lib/web/websocket/frame.js
+++ b/lib/web/websocket/frame.js
@@ -19,7 +19,7 @@ try {
 function generateMask () {
   if (bufIdx === BUFFER_SIZE) {
     bufIdx = 0
-    crypto.randomFillSync(buffer, 0, BUFFER_SIZE)
+    crypto.randomFillSync((buffer ??= Buffer.allocUnsafe(BUFFER_SIZE)), 0, BUFFER_SIZE)
   }
   return [buffer[bufIdx++], buffer[bufIdx++], buffer[bufIdx++], buffer[bufIdx++]]
 }
--- a/lib/web/websocket/frame.js
+++ b/lib/web/websocket/frame.js
@@ -13,7 +13,14 @@ try {
   crypto = require('node:crypto')
 /* c8 ignore next 3 */
 } catch {
-
+  crypto = {
+    // not full compatibility, but minimum.
+    randomFillSync: function randomFillSync (buffer, _offset, _size) {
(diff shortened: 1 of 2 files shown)


metcoder95 (approved):

lib/web/websocket/frame.js
```diff
@@ -2,26 +2,43 @@
...
   crypto = require('node:crypto')
 /* c8 ignore next 3 */
 } catch {
 
 }
 
+function generateMask () {
+  if (buffer === null) {
+    buffer = crypto.randomFillSync(Buffer.allocUnsafe(BUFFER_SIZE), 0, BUFFER_SIZE)
+  }
```
Uzlopak, line 22: you can avoid this if condition, when initilalizing bufIdx with BUFFER_SIZE

lib/web/websocket/frame.js
```diff
@@ -2,26 +2,43 @@
...
+  if (bufIdx + 4 > BUFFER_SIZE) {
```
Uzlopak, line 23: why though?

why not (if bufIdx === BUFFER_SIZE) { ?

lib/web/websocket/frame.js
tsctx: Okey, thanks

lib/web/websocket/frame.js
tsctx: Thanks!

ronag:
You don't actually need the mask. You can just use a 0 mask. It's only necessary for browsers.

ronag:
I open end an issue for this over at ws some time ago.

KhafraDev (approved):

KhafraDev:
> You don't actually need the mask. You can just use a 0 mask. It's only necessary for browsers.

WebSocket is a browser spec, but it's also part of the RFC ;)

```
   The masking key is a 32-bit value chosen at random by the client.
   When preparing a masked frame, the client MUST pick a fresh masking
   key from the set of allowed 32-bit values.  The masking key needs to
   be unpredictable
```

ronag:
Just saying. That's one place you could deviate and gain significant performance with no practical side effects. A zero mask is a noop when applied.

KhafraDev:
I don't disagree that masking is pretty much useless, but I'd prefer to follow the spec at least until WebSocket is mature. Even then I'd prefer if it was opt-in behavior, similar to ws iirc, maybe via a WebSocketDispatcher or similar.

Uzlopak (approved):
LGTM

lib/web/websocket/frame.js
```diff
@@ -2,26 +2,40 @@
 
 const { maxUnsigned16Bit } = require('./constants')
 
+const BUFFER_SIZE = 16386
+
 /** @type {import('crypto')} */
 let crypto
+const buffer = Buffer.allocUnsafe(BUFFER_SIZE)
```
mcollina, line 9: I would rather not allocate 16KB even if WebSocket is never used.

mcollina (changes requested):

lib/web/websocket/frame.js
Uzlopak: @tsctx 

You dont need to allocate here. this can be initialized with null anyway, because if bufIdx === BUFFER_SIZE you will anyway randomFill it in generateMask. 

```suggestion
const buffer = null
```

KhafraDev: @Uzlopak that doesn't work because randomFillSync expects a buffer (etc.) as its first argument.

lib/web/websocket/frame.js
```diff
@@ -2,26 +2,40 @@
...
   crypto = require('node:crypto')
 /* c8 ignore next 3 */
 } catch {
 
 }
 
+function generateMask () {
+  if (bufIdx === BUFFER_SIZE) {
+    bufIdx = 0
+    crypto.randomFillSync(buffer, 0, BUFFER_SIZE)
```
tsctx, line 22:
```suggestion
    crypto.randomFillSync((buffer ??= Buffer.allocUnsafe(BUFFER_SIZE)), 0, BUFFER_SIZE)
```

tsctx:
@mcollina ptal

lib/web/websocket/frame.js
```diff
@@ -2,26 +2,40 @@
 
 const { maxUnsigned16Bit } = require('./constants')
 
+const BUFFER_SIZE = 16386
+
 /** @type {import('crypto')} */
 let crypto
+const buffer = null
```
tsctx, line 9:
```suggestion
let buffer = null
```

KhafraDev (approved):

lib/web/websocket/frame.js
```diff
@@ -2,26 +2,40 @@
...
   crypto = require('node:crypto')
 /* c8 ignore next 3 */
 } catch {
 
 }
 
+function generateMask () {
+  if (bufIdx === BUFFER_SIZE) {
+    bufIdx = 0
+    crypto.randomFillSync((buffer ??= Buffer.allocUnsafe(BUFFER_SIZE)), 0, BUFFER_SIZE)
```
mcollina, line 22: This will throw if there is no crypto.

mcollina (changes requested):

lib/web/websocket/frame.js
tsctx: Why? It is not called when there is no crypto.

tsctx: @KhafraDev Which choice would you prefer?

KhafraDev: I'd probably just use Math.random. We can take a look at disabling masking at another time.

lib/web/websocket/frame.js
```diff
@@ -2,26 +2,40 @@
...
+
 /** @type {import('crypto')} */
 let crypto
+let buffer = null
+let bufIdx = BUFFER_SIZE
+
 try {
   crypto = require('node:crypto')
 /* c8 ignore next 3 */
 } catch {
```
tsctx, line 16:
```suggestion
  crypto = {
    // not full compatibility, but minimum.
    randomFillSync: function randomFillSync (buffer, _offset, _size) {
      for (let i = 0; i < buffer.length; ++i) {
        buffer[i] = Math.floor(Math.random() * 255)
      }
    }
  }
```

lib/web/websocket/frame.js
```diff
@@ -2,13 +2,33 @@
...
+  crypto = {
+    // not full compatibility, but minimum.
+    randomFillSync: function randomFillSync (buffer, _offset, _size) {
+      for (let i = 0; i < buffer.length; ++i) {
+        buffer[i] = Math.floor(Math.random() * 255)
+      }
```
tsctx, line 21:
```suggestion
      }
      return buffer
```

tsctx:
@mcollina ptal

lib/web/websocket/frame.js
```diff
@@ -13,7 +13,14 @@ try {
   crypto = require('node:crypto')
 /* c8 ignore next 3 */
 } catch {
-
+  crypto = {
+    // not full compatibility, but minimum.
+    randomFillSync: function randomFillSync (buffer, _offset, _size) {
+      for (let i = 0; i < buffer.length; ++i) {
+        buffer[i] = Math.floor(Math.random() * 255)
```
Uzlopak, line 20:
```suggestion
        buffer[i] = Math.random() * 255 | 0
```

mcollina (approved):
lgtm

lpinca:
I think that using a non-cryptographically strong random number generator by default is not a wise decision.

1. It violates the specification (the RFC, not the WHATWG spec) that all implementations should follow.
1. It completely defeats the point of masking. Using a 0 mask is more or less the same.
1. It is true that the user has full control over the bytes that are sent over the wire when using the client programmatically, but the client (Node.js) might be embedded in applications that run in the user devices and that work exactly like a browser.

`ws` moved in the opposite direction a long time ago (https://github.com/websockets/ws/pull/832).

Be prepared to receive security disclosures about this or add documentation to prevent them.

tsctx:
Non-cryptographic random number generation is only available in the absence of the crypto module, which by default performs cryptographic and secure random number generation

tsctx:
Non-encrypted random number generation is for fallback when there is no cryptographic module.
By default, encrypted secure random number generation is performed.

lpinca:
Oh, sorry I did not notice that. Out of curiosity, when is the `crypto` module not used?

tsctx:
See Document https://nodejs.org/api/crypto.html#determining-if-crypto-support-is-unavailable

lpinca:
Ok, it's from the Node.js build. I thought `undici` did something special.

PR #3206, opened 2024-05-05
fix websocket receiving an invalid utf-8 in close frame

KhafraDev:
Closes #3080

### DIFF
--- a/lib/web/websocket/connection.js
+++ b/lib/web/websocket/connection.js
@@ -308,7 +308,7 @@ function onSocketClose () {
   if (result) {
     code = result.code ?? 1005
     reason = result.reason
-  } else if (!ws[kReceivedClose]) {
+  } else if (isClosed(ws) && !ws[kReceivedClose]) {
     // If _The WebSocket
     // Connection is Closed_ and no Close control frame was received by the
     // endpoint (such as could occur if the underlying transport connection
--- a/lib/web/websocket/connection.js
+++ b/lib/web/websocket/connection.js
@@ -305,7 +305,7 @@ function onSocketClose () {
 
   const result = ws[kByteParser].closingInfo
 
-  if (result) {
+  if (result && !result.error) {
     code = result.code ?? 1005
     reason = result.reason
   } else if (!ws[kReceivedClose]) {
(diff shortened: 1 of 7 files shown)


Uzlopak:
But shouldnt 1007 be sent by the server if the message sent by the client is not valid?

KhafraDev:
that's for the server to handle, this is the client

Uzlopak:
So?

The spec says:

> 1007
> 
>       1007 indicates that an endpoint is terminating the connection
>       because it has received data within a message that was not
>       consistent with the type of the message (e.g., non-UTF-8 [[RFC3629](https://datatracker.ietf.org/doc/html/rfc3629)]
>       data within a text message).

So, what happens is, that the websocket server is terminating/closing the websocket conenction, because thes server received invalid data. Where does the spec say, that the client is receiving an invalid frame, and thus returns 1007?

KhafraDev:
I don't totally understand what you are asking:

- The server closes the connection for whatever reason. The close frame it sends has invalid utf-8 as its body.
- The client receives the frame, parses the body, sees it is invalid utf-8.
- The client emits an error.
- The client terminates the connection with the appropriate status code.
- Since the server sent a close frame, it will no longer send data to the client.
- Since the client never receives a response, and the server closed the connection, the status code defaults to `1006`.

That is why the server receives a 1007 status code, but the client receives a 1006 status code. The reason we send a 1007 status code is precisely for that section you linked.

KhafraDev:
I tested with Deno (which has a bug in its implementation) and ws. This matches ws' behavior and the logic makes sense to me.

mcollina (approved):
lgtm

lib/web/websocket/connection.js
```diff
@@ -308,7 +308,7 @@ function onSocketClose () {
   if (result) {
     code = result.code ?? 1005
     reason = result.reason
-  } else if (isClosed(ws) && !ws[kReceivedClose]) {
```
KhafraDev: onSocketClose is called once the socket is closed. By the RFC that means we are in the CLOSED state already, we just haven't set it yet.

KhafraDev:
@Uzlopak PTAL

Uzlopak (approved):
After the last commit I love it.

PR #3215, opened 2024-05-07
websocket: move codeblock in parseCloseBody

Uzlopak:
I know, I know... we wont improve the performance of cold branches... but I really think it makes sense to move these lines higher. Checking the code for validity after we set the code makes more sense then checking the validity after we instantiate a buffer for reason.

### DIFF
--- a/lib/web/websocket/receiver.js
+++ b/lib/web/websocket/receiver.js
@@ -314,6 +314,10 @@ class ByteParser extends Writable {
       code = data.readUInt16BE(0)
     }
 
+    if (code !== undefined && !isValidStatusCode(code)) {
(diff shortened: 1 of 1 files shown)


lib/web/websocket/receiver.js
```diff
@@ -314,6 +314,10 @@ class ByteParser extends Writable {
       code = data.readUInt16BE(0)
     }
 
+    if (code !== undefined && !isValidStatusCode(code)) {
+      return { code: 1002, reason: 'Invalid status code', error: true }
+    }
```
Uzlopak, line 319: We could actually also integrate this into the if block above, but probably you consider it to invasive.

KhafraDev (approved):

PR #3219, opened 2024-05-08
Add test to verify if the connection is correctly aborted on cancel

mcollina:
Adds a test to verify `res.body.cancel()` works as expected.

### DIFF
--- a/test/fetch/exiting.js
+++ b/test/fetch/exiting.js
@@ -1,0 +1,42 @@
+'use strict'
+
+const { test } = require('node:test')
+const { fetch } = require('../..')
+const { createServer } = require('node:http')
+const { closeServerAsPromise } = require('../utils/node-http')
+const tspl = require('@matteo.collina/tspl')
+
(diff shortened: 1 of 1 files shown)


metcoder95 (commented):
LGTM but tests seems failing

metcoder95 (approved):
LGTM

test/fetch/exiting.js
```diff
@@ -0,0 +1,42 @@
...
+  await new Promise((resolve) => {
+    server.listen(0, resolve)
+  })
+
+  const url = new URL(`http://127.0.0.1:${server.address().port}`)
+
+  const response = await fetch(url, {
+    // if added both scripts will also exit on 21 and 22.0
+    // dispatcher: new Agent({ bodyTimeout: 1000 })
+  })
```
elf-pavlik, line 30:
```suggestion
  const response = await fetch(url)
```

KhafraDev (approved):

PR #3222, opened 2024-05-08
fix: ignore content-length when dumping HEAD

ronag:
## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/api/api-request.js
+++ b/lib/api/api-request.js
@@ -44,6 +44,7 @@ class RequestHandler extends AsyncResource {
       throw err
     }
 
+    this.method = method
     this.responseHeaders = responseHeaders || null
     this.opaque = opaque || null
     this.callback = callback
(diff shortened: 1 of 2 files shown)


lib/api/api-request.js
```diff
@@ -111,7 +112,13 @@ class RequestHandler extends AsyncResource {
...
     const contentType = parsedHeaders['content-type']
     const contentLength = parsedHeaders['content-length']
-    const body = new Readable({ resume, abort, contentType, contentLength, highWaterMark })
+    const body = new Readable({
+      resume,
+      abort,
+      contentType,
+      contentLength: this.method === 'HEAD' ? null : contentLength,
+      highWaterMark
+    })
```
Uzlopak, line 121: why not the following?
```js
    const contentLength = method === 'HEAD' ? null : parsedHeaders['content-length'] 
    const body = new Readable({
      resume,
      abort,
      contentType,
      contentLength,
      highWaterMark
    })
```

lib/api/api-request.js
```diff
@@ -111,12 +111,13 @@ class RequestHandler extends AsyncResource {
...
     const parsedHeaders = responseHeaders === 'raw' ? util.parseHeaders(rawHeaders) : headers
     const contentType = parsedHeaders['content-type']
-    const contentLength = parsedHeaders['content-length']
     const body = new Readable({
       resume,
       abort,
       contentType,
-      contentLength: this.method === 'HEAD' ? null : contentLength,
+      contentLength: this.method !== 'HEAD' && parsedHeaders['content-length']
+        ? Number(parsedHeaders['content-length'])
```
Uzlopak, line 119: is it ensured, that it is a positive integer?

ronag: No, but does it matter?

mcollina (changes requested):
+1 to the change. Can you please add a unit test?

lib/api/api-request.js
Uzlopak: what happens if it is NaN?

metcoder95: If the server is fully compliant, it should never sent any non-positive value; if a non-positive value is sent, the parser will fail with a non-compliant error

metcoder95 (approved):

PR #3223, opened 2024-05-08
fetch: enable wpt test request-referrer.any.js

Uzlopak:
It seems, that the test is passing. I looked and all the asserts are getting "hit".


KhafraDev (approved):

KhafraDev:
I used to have the runner throw if the failed tests didn't match the expectations. I don't recall why I removed that.

Uzlopak:
Imho the wpt tests are not well written. I manually check if the test is really passing and not giving a false negative or false positive.

PR #3227, opened 2024-05-08
fetch: fix captureStackTrace

### LINKED ISSUE #2869: avoid containing undici internals in error stacktrace in web api
I find it highly confusing, that e.g. when I call .body() on a fetch response and it errors I get all the internal functions in the stacktrace.

This is maybe more about convenience but it makes more sense when using fetch in nodejs... there i want to have e.g. less information about the internals.

Uzlopak:
Closes #2869 

This is an odd bug. 

Obviously `this` is `undefined`, because nobody will run fetch with the `new` operator.

So currently if fetch throws, we create a second time the same stacktrace?

Maybe it is wrong and we should set fetchImpl instead of fetch as reference point?

Maybe we can improve the perf, by storing the stackTraceLimit and set it to 0 and after the fetch we reset it. But problem is, that aggregated errors in cause will have no stack trace then?

### DIFF
--- a/index-fetch.js
+++ b/index-fetch.js
@@ -7,7 +7,7 @@ const fetchImpl = require('./lib/web/fetch').fetch
 module.exports.fetch = function fetch (resource, init = undefined) {
   return fetchImpl(resource, init).catch((err) => {
     if (err && typeof err === 'object') {
-      Error.captureStackTrace(err, this)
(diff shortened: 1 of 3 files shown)


Uzlopak:
well, fetch and fetchImpl are not valid values, as it seems. ;)

mcollina (approved):
lgtm

PR #3228, opened 2024-05-08
add pipe support for wpt server

KhafraDev:
Many tests will communicate with the server and change request behaviors by using "pipes". A "pipe" is a list of commands that modifies the request in different ways. For example:

```js
...?pipe=status(206)|header(Content-Type, text/plain)
```

will cause the server to change the status code to 206 and add a `content-type: text/plain` header.

### DIFF
--- a/test/wpt/status/fetch.status.json
+++ b/test/wpt/status/fetch.status.json
@@ -446,6 +446,8 @@
   },
   "security": {
     "1xx-response.any.js": {
+      "note": "TODO(@KhafraDev): investigate timeout",
+      "skip": true,
       "fail": [
         "Status(100) should be ignored.",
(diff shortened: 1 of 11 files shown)


test/wpt/status/fetch.status.json
```diff
@@ -446,6 +446,8 @@
   },
   "security": {
     "1xx-response.any.js": {
+      "note": "TODO(@KhafraDev): investigate timeout",
```
KhafraDev, line 449: @Uzlopak here are some more WPT failures that should be fixable ;)

mcollina (approved):
lgtm

Uzlopak (approved):

PR #3233, opened 2024-05-09
fetch: Fix wpt test request-upload.any.js

Uzlopak:
See the first commit for the fix, the second for the refactoring of the file, which makes it more like the python file.


Uzlopak:
wait a second, this is wrong... nevermind.

PR #3234, opened 2024-05-09
fetch: fix wpt test request-upload.any.js

Uzlopak:
Strange... it seems that the python code in network-partition-key.py is invalid anyway?

### DIFF
--- a/test/wpt/server/lockedresource.mjs
+++ b/test/wpt/server/lockedresource.mjs
@@ -1,0 +1,31 @@
+export class LockedResource {
+  constructor (resource) {
+    this.locked = false
+    this.waitingQueue = []
+    this.resource = resource
+  }
+
+  async acquire () {
+    return new Promise(resolve => {
+      if (!this.locked) {
+        // If the lock is not already acquired, acquire it immediately
+        this.locked = true
+        resolve(this.resource)
+      } else {
+        // If the lock is already acquired, queue the resolve function
+        this.waitingQueue.push(resolve)
+      }
+    })
+  }
+
+  release () {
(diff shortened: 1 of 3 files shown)


Uzlopak:
Tbh i dont understand why the corresponding network-partition-key.py code is [not failing on other platforms](https://wpt.fyi/results/fetch/api/basic/request-upload.any.serviceworker.html?label=master&label=experimental&aligned). 

I checked our fetch implementation, and our code is working accordingly. We create two socket connections, and we do two requests.  

So I patched the code to adapt it to our needs. So we keep track of the sockets and when a new socket is used, we know that there was a new connection. 

If you dislike this change, I can impelment a minimal version of this PR.

KhafraDev:
> Strange... it seems that the python code in network-partition-key.py is invalid anyway?

Could you expand on this?

> tbh i dont understand why the corresponding network-partition-key.py code is [not failing on other platforms](https://wpt.fyi/results/fetch/api/basic/request-upload.any.serviceworker.html?label=master&label=experimental&aligned).

same here

> If you dislike this change, I can impelment a minimal version of this PR.

Why would I dislike the change?

mcollina (approved):
lgtm

Uzlopak:
I am not a python developer. It makes not much sense to me. Maybe for some reason connection_count and request_count are some kind of globals in python, which is not the case in javascript. 

Well, anyway. This implementation deviates from the original by using the sockets and also it is commented properly. 

If you agree, we can merge it ;).

KhafraDev:
For future reference, I don't care about following what the WPT server does. It's usually easier to copy verbatim, but I don't expect everything from python to translate perfectly to js.

PR #3235, opened 2024-05-09
fetch: don't set an invalid origin header

KhafraDev:
refs: https://github.com/DIYgod/RSSHub/issues/15532

### DIFF
--- a/lib/web/fetch/index.js
+++ b/lib/web/fetch/index.js
@@ -444,7 +444,7 @@ function fetching ({
   // 9. If request’s origin is "client", then set request’s origin to request’s
   // client’s origin.
   if (request.origin === 'client') {
-    request.origin = request.client?.origin ?? request.origin
+    request.origin = request.client.origin
   }
 
   // 10. If all of the following conditions are true:
--- a/lib/web/fetch/index.js
+++ b/lib/web/fetch/index.js
@@ -444,8 +444,7 @@ function fetching ({
   // 9. If request’s origin is "client", then set request’s origin to request’s
   // client’s origin.
   if (request.origin === 'client') {
-    // TODO: What if request.client is null?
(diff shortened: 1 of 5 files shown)


KhafraDev:
this wasn't caught in the WPTs because we set a global origin which doesn't trigger the error case

lib/web/fetch/index.js
```diff
@@ -444,8 +444,7 @@ function fetching ({
   // 9. If request’s origin is "client", then set request’s origin to request’s
   // client’s origin.
   if (request.origin === 'client') {
```
Uzlopak, line 446:
```
  if (request.origin === 'client' && request.client?.origin) {
    request.origin = request.client.origin
  }
```

Uzlopak: atleast a comment would be nice

Uzlopak (approved):
Makes sense

lib/web/fetch/util.js
```diff
@@ -255,16 +255,23 @@ function appendFetchMetadata (httpRequest) {
 
 // https://fetch.spec.whatwg.org/#append-a-request-origin-header
 function appendRequestOriginHeader (request) {
-  // 1. Let serializedOrigin be the result of byte-serializing a request origin with request.
+  // 1. Let serializedOrigin be the result of byte-serializing a request origin
+  //    with request.
+  // TODO: implement "byte-serializing a request origin"
```
Uzlopak, line 260: btw. you could use isomorphicEncode from lib/web/fetch/util.js

KhafraDev: byte serializing a request origin is much more than isomorphic encoding

Uzlopak: right

Uzlopak (approved):

Issue #2859, opened 2024-02-27
WebSocket: message handler sometimes called with null data (potentially issue in frame processing)


farwayer:
## Bug Description

I’ll say right away that the bug is floating and I don’t have a ready-made recipe for reproducing it. The bug usually appeared during debugging. Apparently, timings and the sequence of data arrival are important. The server should also send pings to get this effect.

From the screenshot you can see that the `websocketMessageReceived` function is called with `this.#info.originalOpcode=9` (PING frame type), which should not happen.

## Logs & Screenshots

![2024_02_27T05 33 03](https://github.com/nodejs/undici/assets/2343101/80279f6f-fb33-42a9-a909-a4bab15091ee)

## Environment

node v21.6.2

KhafraDev:
does this only occur during debugging?

farwayer:
Yes, I only encountered this error in debugging. But, I assume that the problem is not related to it, but to the timings. Before sending null to the message handler, two breakpoints are triggered. But if you send a ping from the server with a little delay (calling `process.hrtime()` before sending is enough), then everything goes fine.

KhafraDev:
Maybe, I think there's definitely room for improvement in parsing close frames (frames that can't be fragmented). I'll look into it but no promises.

LubosD:
FWIW, I definitely encountered this bug outside of debugging. The WebSocket would receive messages correctly for a while and then it would always give me `null`s.

PR #3239, opened 2024-05-10
websocket: fix parsing of control frames

tai-kun:
## Rationale

If a PING frame is received before binary/text frames is received, those data will be dropped.

Message data is decoded to UTF-8 text if the `opcode` is TEXT, or converted to Blob or ArrayBuffer if it is BINARY. However, the receiver (`ByteParser`) determines that the `opcode` of a frame that cannot be fragmented, such as a PING, is the `opcode` of a potentially fragmented message. And it will not be overturned until a message handler is called. Therefore, by restricting the type of `opcode` that is set in `this.#info.originalOpcode`, we ensure that messages are processed accordingly with the appropriate `opcode`.

## Changes

- Restrict the frame type of fragmented messages to be binary or text only.

### Features

N/A

### Bug Fixes

- #2859

N/A

## Status

- [x] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/web/websocket/receiver.js
+++ b/lib/web/websocket/receiver.js
@@ -65,7 +65,9 @@ class ByteParser extends Writable {
 
         // If we receive a fragmented message, we use the type of the first
         // frame to parse the full message as binary/text, when it's terminated
-        this.#info.originalOpcode ??= this.#info.opcode
+        if (this.#info.opcode === opcodes.BINARY || this.#info.opcode === opcodes.TEXT) {
+          this.#info.originalOpcode ??= this.#info.opcode
+        }
 
         this.#info.fragmented = !this.#info.fin && this.#info.opcode !== opcodes.CONTINUATION
 
--- a/test/websocket/issue-2859.js
+++ b/test/websocket/issue-2859.js
@@ -1,0 +1,38 @@
+'use strict'
+
+const { test } = require('node:test')
+const { WebSocketServer } = require('ws')
+const { WebSocket } = require('../..')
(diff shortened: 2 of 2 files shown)


Uzlopak:
Ok, existing tests are not breaking but can you add a unit test, please?

tai-kun:
yes. I'll write tests as soon as possible, even though I won't be able to write any code for a while.

KhafraDev:
I believe my PR should fix this also, but we'd need a test.

tai-kun:
I have added a test here:
`test/websocket/issue-2859.js`

Added test is based on `test/websocket/fragments.js`. In this test, a TEXT frame is sent before sending a PING, so message data was never dropped. Added test reproduces the problem this time by sending a PING first.

Before fixing control frames parsing, the test `issue-2859.js` fails with an error similar to the following:
```
✖ Fragmented frame with a ping frame in the first of it (42.77125ms)
  AssertionError [ERR_ASSERTION]: Expected values to be strictly equal:
  
  null !== 'Hello'
  
      at <...trace> {
    generatedMessage: true,
    code: 'ERR_ASSERTION',
    actual: null,
    expected: 'Hello',
    operator: 'strictEqual'
  }
```

mcollina (approved):
lgtm

KhafraDev (changes requested):
This is fixed in #3241 as well, which is a more reliable fix. Thank you for the test, I added it there as well..

PR #3241, opened 2024-05-10
refactor websocket control frame handling

### LINKED ISSUE #2859: WebSocket: message handler sometimes called with null data (potentially issue in frame processing)

KhafraDev:
Closes #3239 
Closes #2859 

Control frames cannot be fragmented which guarantees that they will be handled in full in a single message. Since this is the case, there's no reason to assign values to `this.#info` (which is the internal state/settings of a frame) which should avoid race conditions.

### DIFF
--- a/lib/web/websocket/receiver.js
+++ b/lib/web/websocket/receiver.js
@@ -1,10 +1,11 @@
 'use strict'
 
 const { Writable } = require('node:stream')
+const assert = require('node:assert')
 const { parserStates, opcodes, states, emptyBuffer, sentCloseFrameState } = require('./constants')
 const { kReadyState, kSentClose, kResponse, kReceivedClose } = require('./symbols')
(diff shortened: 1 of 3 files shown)


KhafraDev:
@Uzlopak @tsctx

KhafraDev:
@nodejs/undici

tsctx (approved):

KhafraDev:
thank you

PR #3244, opened 2024-05-10
fix: parseHeaders should take Record<string, ...>

ronag:
## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/core/util.js
+++ b/lib/core/util.js
@@ -273,28 +273,75 @@ function bufferToLowerCasedHeaderName (value) {
 }
 
 /**
- * @param {Record<string, string | string[]> | (Buffer | string | (Buffer | string)[])[]} headers
+ * @param {Record<string, string | string[] | null | undefined> | (Buffer | string | (Buffer | string)[])[]} headers
  * @param {Record<string, string | string[]>} [obj]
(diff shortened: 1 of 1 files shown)


mcollina (commented):
Can you add a test? What is this fixing?

lib/core/util.js
```diff
@@ -278,23 +278,50 @@ function bufferToLowerCasedHeaderName (value) {
...
+      if (val) {
+        if (typeof val === 'string') {
+          val = [val]
+          obj[key] = val
+        }
+        assert(!Array.isArray(val2))
+        val.push(val2.toString())
+      } else {
+        // TODO (fix): assert type...
+        obj[key] = Array.isArray(val2) ? val2.map(x => x.toString()) : val2.toString()
```
Uzlopak, line 306: This works? What if val2 is null? Is this possible?

ronag:
> Can you add a test? What is this fixing?

It's fixing so that the function does what the types say. Since we export this function it's important that it's correct.

ronag:
Added in nxt-undici for now. If someone wants to makes tests feel free to take over.

PR #3246, opened 2024-05-11
Implement Service Worker API

kettanaito:
## Rationale

Implementing the [ServiceWorker API](https://w3c.github.io/ServiceWorker) in Node.js can enable request interception and caching capabilities. This becomes more and more common as JavaScript frameworks expand onto the server, with concepts like React Server Components. Those frameworks rely on custom patches of globals to implement features like request deduplication and caching, whereas they could use the standard API designed just for that. 

The API itself is meant to be W3C specification-compliant to the best of Node.js abilities. Naturally, things like `.openWindow()` and other browser-related methods or properties will not be implemented (I advocate for leaving them there for compatibility but skip the implementation as it's usually a side effect to perform an action in the browser). 

This implementation may also deviate from the specification where appropriate. For example, representing the `client.type` as `window` or `frame` is irrelevant and confusing in Node.js. Instead, I'd propose to either coerce the parent/child relationship of the main/worker threads onto `window` or `worker`, or introduce _custom client types_ like `main-thread` and `worker`. 

## Changes

Implements the classes needed for the Service Worker API (please see the "files" for more detailed overview). 

This change does not impose any breaking changes as this is a new API being added to Undici. 

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

## Implementation

> [!WARNING]
> This is a _work-in-progress implementation_. I would like to get some code review on it before continuing. It still needs tests, for once. Thank you. 

## Roadmap

- [ ] `ServiceWorkerContainer.getRegistration()`
- [ ] `ServiceWorkerContainer.getRegistrations()`
- [ ] `ServiceWorkerGlobalScope.registration`
- [ ] `importScripts()` in the worker context (needs discussion).  
- [ ] `Client.type` (needs discussion)
- [ ] `Client.frameType` (needs discussion)

### DIFF
--- a/lib/web/service-worker/index.js
+++ b/lib/web/service-worker/index.js
@@ -1,0 +1,3 @@
+import { ServiceWorkerContainer } from './service-worker-container.js'
+
+export const serviceWorker = new ServiceWorkerContainer()
--- a/lib/web/service-worker/interceptor.js
+++ b/lib/web/service-worker/interceptor.js
@@ -1,0 +1,13 @@
+import { BatchInterceptor } from '@mswjs/interceptors'
+import { ClientRequestInterceptor } from '@mswjs/interceptors/ClientRequest'
+import { FetchInterceptor } from '@mswjs/interceptors/fetch'
+import { XMLHttpRequestInterceptor } from '@mswjs/interceptors/XMLHttpRequest'
+
+export const interceptor = new BatchInterceptor({
+  name: 'node-service-worker',
+  interceptors: [
+    new ClientRequestInterceptor(),
+    new FetchInterceptor(),
+    new XMLHttpRequestInterceptor()
+  ]
+})
--- a/lib/web/service-worker/service-worker-container.js
+++ b/lib/web/service-worker/service-worker-container.js
@@ -1,0 +1,201 @@
+import { Worker, MessageChannel } from 'node:worker_threads'
+import { DeferredPromise } from '@open-draft/deferred-promise'
+import { ServiceWorkerRegistration } from './service-worker-registration.js'
+import { ServiceWorker } from './service-worker.js'
+import { parseModuleUrlFromStackTrace } from './utils/parseModuleUrlFromStackTrace.js'
+import { interceptor } from './interceptor.js'
+
+/**
+ * @typedef {Object} WorkerData
+ * @property {string} scriptUrl
+ * @property {unknown} options @todo
+ * @property {import('./client.js').SerializedClient} clientInfo
+ * @property {MessagePort} clientMessagePort
+ * @property {MessagePort} interceptorMessagePort
+ */
+
+const clientMessageChannel = new MessageChannel()
+const interceptorMessageChannel = new MessageChannel()
+
+/**
+ * @see https://w3c.github.io/ServiceWorker/#serviceworkercontainer-interface
+ */
+export class ServiceWorkerContainer {
+  /** @type {ServiceWorkerRegistration | undefined} */
+  #registration
+  /** @type {DeferredPromise<ServiceWorkerRegistration>} */
+  #ready
+
+  /** @todo Event handlers (oncontrollerchange, onmessage, onerror) */
+
+  constructor () {
+    this.#ready = new DeferredPromise()
+  }
+
+  /**
+   * @returns {Promise<ServiceWorkerRegistration>}
+   *
+   * @see https://w3c.github.io/ServiceWorker/#navigator-service-worker-ready
+   */
+  get ready () {
+    return this.#ready
+  }
+
+  /**
+   * @returns {ServiceWorker | null}
+   *
+   * @see https://w3c.github.io/ServiceWorker/#navigator-service-worker-controller
+   */
+  get controller () {
+    return this.#registration?.active || null
+  }
+
+  /**
+   * @param {string} scriptUrl
+   * @param {ServiceWorkerRegistrationOptions} [options]
+   * @returns {Promise<ServiceWorkerRegistration>}
+   *
+   * @see https://w3c.github.io/ServiceWorker/#navigator-service-worker-register
+   */
+  async register (scriptUrl, options = {}) {
+    /** @type {SerializedClient} */
+    const clientInfo = {
+      id: process.pid.toString(),
+      url: parseModuleUrlFromStackTrace(new Error()),
+      type: 'worker',
+      frameType: '???' /** @todo */
+    }
+
+    const worker = new Worker(new URL('./worker.ts', import.meta.url), {
+      name: `[worker ${scriptUrl}]`,
+      workerData: {
+        scriptUrl,
+        options,
+        clientInfo,
+        clientMessagePort: clientMessageChannel.port2,
+        interceptorMessagePort: interceptorMessageChannel.port2
+      },
+      transferList: [
+        clientMessageChannel.port2,
+        interceptorMessageChannel.port2
+      ]
+    })
+
+    const serviceWorker = this.#createServiceWorker(scriptUrl, worker)
+    const registration = new ServiceWorkerRegistration(serviceWorker)
+    this.#registration = registration
+
+    serviceWorker.addEventListener('statechange', () => {
+      if (serviceWorker.state === 'activating') {
+        this.#ready.resolve(registration)
+      }
+    })
+
+    this.#enableRequestInterception(interceptorMessageChannel)
+
+    return registration
+  }
+
+  /**
+   * @param {string} clientUrl
+   * @returns {Promise<ServiceWorkerRegistration | undefined>}
+   *
+   * @see https://w3c.github.io/ServiceWorker/#navigator-service-worker-getRegistration
+   */
+  getRegistration (clientUrl) {
+    throw new Error('Not implemented')
+  }
+
+  /**
+   * @returns {Promise<Array<ServiceWorkerRegistration>>}
+   *
+   * @see https://w3c.github.io/ServiceWorker/#navigator-service-worker-getRegistrations
+   */
+  getRegistrations () {
+    throw new Error('Not implemented')
+  }
+
+  /**
+   * @param {string} scriptUrl
+   * @param {Worker} worker
+   * @returns {ServiceWorker}
+   */
+  #createServiceWorker (scriptUrl, worker) {
+    const serviceWorker = new ServiceWorker(
+      scriptUrl,
+      worker.postMessage.bind(worker)
+    )
+
+    // Listen to the Service Worker signaling its events
+    // and update the main thread Service Worker instance accordingly.
+    worker.addListener('message', (message) => {
+      switch (message.type) {
+        case 'worker/statechange': {
+          serviceWorker.state = message.state
+          break
+        }
+        case 'worker/error': {
+          serviceWorker.dispatchEvent(new Event('error'))
+          break
+        }
+      }
+    })
+
+    // Forward the messages sent via `client.postMessage()` in the worker
+    // directly to the Service Worker interface.
+    clientMessageChannel.port1.addListener('message', (data) => {
+      serviceWorker.dispatchEvent(new MessageEvent('message', { data }))
+    })
+
+    return serviceWorker
+  }
+
+  /**
+   * @param {MessageChannel} channel
+   * @returns {void}
+   */
+  #enableRequestInterception (channel) {
+    interceptor.apply()
+
+    interceptor.on('request', async ({ requestId, request }) => {
+      const requestBody = await request.arrayBuffer()
+      channel.port1.postMessage(
+        {
+          type: 'request',
+          requestId,
+          request: {
+            method: request.method,
+            url: request.url,
+            headers: Object.fromEntries(request.headers.entries()),
+            body:
+              request.method === 'HEAD' || request.method === 'GET'
+                ? null
+                : requestBody
+          }
+        },
+        [requestBody]
+      )
+
+      const responsePromise = new DeferredPromise()
+
+      const responseListener = (data) => {
+        if (requestId === data.requestId) {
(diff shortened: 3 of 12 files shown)


lib/web/service-worker/service-worker-global-scope.js
```diff
@@ -0,0 +1,84 @@
+import { parentPort } from 'node:worker_threads'
+import { Clients, kAddClient } from './clients.js'
+import { ServiceWorker } from './service-worker.js'
+import { Client } from './client.js'
+import { CacheStorage } from '../cache/cachestorage.js'
```
kettanaito, line 5: TIL Undici has its own `CacheStorage` implementation. I'm reusing it for the `ServiceWorkerGlobalScope` as-is.

lib/web/service-worker/worker.js
```diff
@@ -0,0 +1,106 @@
...
+
+parentPort.postMessage({
+  type: 'worker/statechange',
+  state: 'parsed'
+})
+
+// Run the worker script within the controller global scope.
+const script = new vm.Script(content)
+
+script.runInNewContext({
```
kettanaito, line 36: This is the heart of the implementation: running the actual worker script in a VM _inside_ a `worker.js` context. See `service-worker-container.js` for more details on how `worker.js` is spawned.

lib/web/service-worker/service-worker-container.js
```diff
@@ -0,0 +1,201 @@
...
+  async register (scriptUrl, options = {}) {
+    /** @type {SerializedClient} */
+    const clientInfo = {
+      id: process.pid.toString(),
+      url: parseModuleUrlFromStackTrace(new Error()),
+      type: 'worker',
+      frameType: '???' /** @todo */
+    }
+
+    const worker = new Worker(new URL('./worker.ts', import.meta.url), {
```
kettanaito, line 69: The container (the `serviceWorker` object) spawns a new worker of `worker.js` to provide scope isolation akin to how a Service Worker operates in the browser. I suggest we keep that isolation for the worker code not to leak into the main thread.

lib/web/service-worker/service-worker-container.js
```diff
@@ -0,0 +1,201 @@
...
+    return registration
+  }
+
+  /**
+   * @param {string} clientUrl
+   * @returns {Promise<ServiceWorkerRegistration | undefined>}
+   *
+   * @see https://w3c.github.io/ServiceWorker/#navigator-service-worker-getRegistration
+   */
+  getRegistration (clientUrl) {
```
kettanaito, line 105: ## Question

Should we abide by the rule that only _one_ Service Worker can control the same scope, like in the browser? Or should we allow an arbitrary number of Service Workers controlling the main thread? 

I lean toward the first option, would like to hear your thoughts on this as well.

lib/web/service-worker/index.js
```diff
@@ -0,0 +1,3 @@
+import { ServiceWorkerContainer } from './service-worker-container.js'
+
+export const serviceWorker = new ServiceWorkerContainer()
```
kettanaito, line 3: The API is meant to be consumed via the `serviceWorker` object:

```js
// main.js
import { serviceWorker } from 'undici'

await serviceWorker.register('./my-worker.js')
```

lib/web/service-worker/interceptor.js
```diff
@@ -0,0 +1,13 @@
+import { BatchInterceptor } from '@mswjs/interceptors'
+import { ClientRequestInterceptor } from '@mswjs/interceptors/ClientRequest'
+import { FetchInterceptor } from '@mswjs/interceptors/fetch'
+import { XMLHttpRequestInterceptor } from '@mswjs/interceptors/XMLHttpRequest'
+
+export const interceptor = new BatchInterceptor({
```
kettanaito, line 6: ## Question

Which interception algorithm do you see more fitting?

I was considering between the `MockAgent` of Undici and Interceptors from MSW. There are reasons to use either. 

- `MockAgent` is in-line with Undici but covers _only_ Undici's implementation of `fetch`. I'd expect the Service Worker API to intercept _any_ HTTP requests, just like it does in the browser.
- Interceptors intercepts any HTTP requests but is a third-party library. If you are willing to add it as a dependency, I have nothing against using it. Interceptors is also slowly becoming the backbone for request interceptions for other libraries too, with Nock [slowly migrating](https://github.com/nock/nock/pull/2517) to use it instead. Using it in Undici may solidify it as a well-established request interception API in Node.js.

lib/web/service-worker/interceptor.js
```diff
@@ -0,0 +1,13 @@
...
+  name: 'node-service-worker',
+  interceptors: [
+    new ClientRequestInterceptor(),
+    new FetchInterceptor(),
+    new XMLHttpRequestInterceptor()
```
kettanaito, line 11: ## Todo

Need to remove this. There are no XHR in Node.js. We won't be tailoring for browser-like environments like JSDOM.

lib/web/service-worker/service-worker-global-scope.js
```diff
@@ -0,0 +1,84 @@
...
+ * This acts as `self` in the global scope of the Service Worjer.
+ */
+export class ServiceWorkerGlobalScope extends EventTarget {
+  /** @type {import('./service-worker-container.js').WorkerData} */
+  #parentData
+
+  /**
+   * @param {import('./service-worker-container.js').WorkerData} parentData
+   */
+  constructor (parentData) {
```
kettanaito, line 18: ## Todo

Implement the `registration` representation on the global scope. Something similar to how the `self.serviceWorker` is implemented. 

Those will be different objects in the main thread and in the worker thread but connected via messaging.

kettanaito:
I'd like to improve the spec-compliance because I spent two last days implementing `Cache` and `CacheStorage` verbatim only to find out you've already done so in Undici 😆 I will go through the classes and make them follow the spec as much as possible. 

I still expect some internal deviations and perhaps even public deviations (see the PR description), both of which should not disrupt the usage of the API.

kettanaito:
@mcollina, I'm curious how much the Service Worker API falls under the direction of Undici. I'm mainly choosing to land this in Undici since I get an impression it's the frontliner for trying out the experimental things in Node.js. Let me know what you think about where this API belongs to.

mcollina:
Can you add an example or (better) a test to use this?

KhafraDev:
I'm not sure if undici is the right place to implement this

kettanaito:
@mcollina, will do. A basic integration test is the next on my list. 

@KhafraDev, my thinking as well. Can we please agree on that first, since this PR is a manually edited fork of my repo, which is in TypeScript. I'd love to keep maintaining just one implementation. If it doesn't belong in Undici, I will simply make my repo public and publish this on NPM. 

We can also create a repo under the Node.js org if that's not running ahead of the bus, of course. I leave this to your best judgment, just please let me know. Thanks.

rluvaton:
I don't think it's the correct place as well, and also, regarding tests, due it being behind a web standard spec there are already tests to it part of WPT, we should enable that for them

kettanaito:
Unless you are giving me a hard pass, I'd ship this as a third-party package, gave it some time, improve it, and then eventually land it into Node.js, or any more suiting repos. 

@rluvaton, yeah, that was my thinking as well. Just need to do the test orchestration for that.

KhafraDev:
There are a number of pieces that are missing from node before this can be done, right? You'd need globalThis to extend EventTarget, you'd need to add `self` as an alias for globalThis (otherwise mdn tutorials wouldn't work), web workers, and probably a bit more (I'm not familiar to the spec and its prerequisites).

I'm sure it's possible to get it running without these, but then the api is fundamentally different and at that point I see little reason to implement a w3c spec. Same for the differences you mentioned - if the api isn't suitable to node I see little reason to implement it.

Even then, the benefits you mentioned (caching and intercepting) are both already possible to do in node/undici, and a majority of the use cases that the spec directly outlines are irrelevant to node. Could you outline why service workers should be implemented?

mcollina:
Personally I think that having `ServiceWorkers` in Node.js would be _useful_, as a lot of people are using them to intercept `fetch()` calls.

The only question I have is that if we need other things before implementing this here (HTTP caching support, globalThis, etc).
Then, I think there is a good demand on adding them to Node.js.

As for landing this here, I think it's the right place mostly because we have a better webidl implementation and generically better support for WPTs.

I'm going to ping the @nodejs/tsc before we give you an answer. What I want to avoid is to have another situation similar to what happened to node-fetch when we shipped `fetch()` in node.

kettanaito:
> You'd need globalThis to extend EventTarget, you'd need to add self as an alias for globalThis (otherwise mdn tutorials wouldn't work), web workers, and probably a bit more (I'm not familiar to the spec and its prerequisites).

@KhafraDev, this is already done, perhaps partially. I'm using VM context to declare my own `global`, `globalThis`, and so forth for the worker's scope. It worker precisely as MDN and the spec describe, with things like `self.addEventListener` and `addEventListener` as well (take a look at `worker.js`). That's how the Service Worker scope was meant to be implemented from day one—explicitly exposing globals onto it. 

> Could you outline why service workers should be implemented?

While it's possible to achieve caching in Undici, that is a non-standard behavior specific to Undici. What I'm proposing here is a standard API designated for caching to work _together_ with Undici and beyond, as Undici still has a vastly smaller usage compared to `http.request` (we'll get there, I know, but we are not there just yet). 

@mcollina, thank you. I'd love to have this answered so I can continue working on this.

mcollina:
I personally won't recommend to touch http.request() for anything. Even bugfixes causes so many regression that is playing whack-a-mole. That's the reason undici exists.

I would recommend focusing on fetch().

targos:
> I'm going to ping the @nodejs/tsc before we give you an answer. What I want to avoid is to have another situation similar to what happened to node-fetch when we shipped fetch() in node.

What happened to node-fetch?

mcollina:
> What happened to node-fetch?

As far as I recall the maintainers were relatively frustrated when we shipped `fetch()` because it would have taken market share (and possibly sponsor money) from them.  I don't have time to collect references for now.

(The same happened with other libraries when we shipped them in core).

I want to avoid this situation if possible.

KhafraDev:
> As for landing this here, I think it's the right place mostly because we have a better webidl implementation and generically better support for WPTs.

True, but this is also an http library. Everything web related revolves around fetch (except FileReader, which was largely a mistake that is planned on being removed). I can't see the justification for undici to start implementing web workers and service workers.

It is also possible to use undici's webidl outside of undici; `require('undici/lib/web/fetch/webidl.js')`

> While it's possible to achieve caching in Undici, that is a non-standard behavior specific to Undici. What I'm proposing here is a standard API designated for caching to work together with Undici and beyond

undici also has Cache/CacheStorage.

---

> A service worker is a type of [web worker](https://html.spec.whatwg.org/multipage/workers.html#workers).

One of the first lines of the spec. We can't have service workers without web workers but we're skipping past the requirements? It would be like implementing fetch without web streams or Blob; there's a reason fetch came into existence once all of the pieces were in-place.

KhafraDev:
One more thing, please read what the Deno maintainers have to say about service workers, I think it's largely relevant here:
https://github.com/denoland/deno/issues/5957/#issuecomment-987375430

> because we do not see enough user value to justify the significant time and architecture complexity investment.

> incoming request interceptors - also known as HTTP servers

Server environments can already "intercept" requests, as mentioned previously, by using servers. In an environment that does not have this convenience, a service worker may be needed, but node has the tooling to handle this far better than the service workers api provides.

benjamingr:
>. What happened to node-fetch?

> As far as I recall the maintainers were relatively frustrated when we shipped fetch() because it would have taken market share (and possibly sponsor money) from them. I don't have time to collect references for now.

The original author (bitinn) was helpful and engaged in a lot of discussions and provided valuable feedback, do you mean string formattting by any chance? https://github.com/nodejs/node/issues/43382

kettanaito:
> I would recommend focusing on fetch().

> We can't have service workers without web workers but we're skipping past the requirements?

Two more reason to keep this a third-party package and evolve it over time. I have no time to dive into a full web worker implementation but the package can already provide value. The requirement bar is simply different here, and I have no problem writing a mention in the README of the lib's limitations. Landing it in Node.js, on other hand, is a far more serious commitment. 

> One more thing, please read what the Deno maintainers have to say about service workers, I think it's largely relevant here:
https://github.com/denoland/deno/issues/5957

Read through that, thanks. I share some of those concerns and it's yet another reason to push this to the userland. I don't mind it being incomplete or partially spec-compliant. I also want the Service Worker API to work with any HTTP request clients, just like it does in the browser, not just fetch. Since Interceptors already provide an extensive interception for `http.request` (and a grand rewrite is on the way), I am getting that support for free. 

**I am moving this to userland until Node.js voices the need to have the Service Worker API**. It will allow the API to evolve faster while also keeping commitment lower on my end, as well as lower the contribution threshold to the project. 

Thank you everyone for the discussion!

rluvaton:
Thank you for your PR even though it was not merged, we know you put your time into it and we appreciate it

if you have a repo I would be happy to take a look to give feedback if you want

kettanaito:
@rluvaton, thank you for the encouragement! I'd be grateful if you looked at what I have right now. As I am still working on making it specification-compliant, one area I'd love to hear thoughts on is my overall architecture of this API (the usage of worker_threads + VM) and whether it's a good approach in general. 

Here's the repo: https://github.com/kettanaito/service-worker. The bits I'd love to have reviewed:

- https://github.com/kettanaito/service-worker/blob/b1d14f58377438a8a83e3d11cf68e2e98a0577f5/src/ServiceWorkerContainer.ts#L55
- https://github.com/kettanaito/service-worker/blob/b1d14f58377438a8a83e3d11cf68e2e98a0577f5/src/worker.ts#L36

Thank you! ❤️

PR #3247, opened 2024-05-12
fix parsing continuation frames in websocket

KhafraDev:
Fixes a bug I discovered when testing continuation frames, but this is more of a general refactor.

The old way didn't work at all and sucked pretty hard.


tsctx (approved):

KhafraDev:
thank you for the quick review!

Issue #3256, opened 2024-05-13
Implement Request/Response `.bytes()` method


bakkot:
See https://github.com/whatwg/fetch/issues/1732.

Tests at https://github.com/web-platform-tests/wpt/pull/46198.

The above spec PR adds a `bytes()` method to get a Uint8Array to Body, instead of requiring users to get an ArrayBuffer with `arrayBuffer()` and then wrap it.

PR #3262, opened 2024-05-14
add bodymixin bytes

KhafraDev:
closes: #3256 
refs: https://github.com/whatwg/fetch/pull/1753

this PR also moves away from using git sparse-checkout for pulling WPTs and instead makes use of node's own node core utils, which is where 99% of the changes come from.


lib/web/fetch/body.js
```diff
@@ -385,6 +385,15 @@ function bodyMixinMethods (instance) {
...
         )
       }, instance, false)
+    },
+
+    bytes () {
+      // The bytes() method steps are to return the result of running consume body
+      // with this and the following step given a byte sequence bytes: return the
+      // result of creating a Uint8Array from bytes in this’s relevant realm.
+      return consumeBody(this, (bytes) => {
+        return new Uint8Array(bytes)
```
KhafraDev, line 395: cc @tsctx if `fullyReadBody`, `readAllBytes`, and `consumeBody` returned a Uint8Array instead of a Buffer we could avoid this clone

tsctx: I will work on it in my free time :)

tsctx (approved):
lgtm

lib/web/fetch/body.js
tsctx, line 395: This will avoid cloning.
```suggestion
        return new Uint8Array(bytes.buffer, 0, bytes.byteLength)
```

PR #3269, opened 2024-05-16
use private properties in Headers

### LINKED ISSUE #3267: bug: spreading `Headers` object has breaking behavior in versions between Node 20.12 and 20.13
## Reproducible By

```js
// Comment out this comment and have the latest undici installed just to verify it's still
// present in latest Undici, but for the test you can use the global Undici bundled in Node
// import { fetch } from "undici";

{
  const baseHeaders = { "x-foo": "bar" };

  const requestHeaders = new Headers({ "Content-Type": "application/json" });
  const headers = {
    ...baseHeaders,
    ...requestHeaders,
  };
  console.log("Result", headers);

  await fetch("https://google.com", {
    method: "POST",
    headers,
  });
}
```

Run the code in Node 20.12 and 20.13. 20.12 works fine, while 20.13 panics with `TypeError: Could not convert argument of type symbol to string`.

## Expected Behavior

It to work like it did in 20.12 - afterall I didn't bump any major version, so if this is an intentional breaking change on Undici's side it shouldn't have been included in a minor version of Node.js.

## Logs & Screenshots

CleanShot 2024-05-16 at 12 00 10

## Environment

- Latest MacOS
- Node 20.13

### Additional context

A separate issue for Node perhaps, but spreading the Headers object probably shouldn't even include the Symbol(headers list) to begin with, this is how it looks in the browser for example:

CleanShot 2024-05-16 at 12 04 29

KhafraDev:
fixes #3267

### DIFF
--- a/benchmarks/fetch/headers-length32.mjs
+++ b/benchmarks/fetch/headers-length32.mjs
@@ -1,5 +1,5 @@
 import { bench, run } from 'mitata'
-import { Headers } from '../../lib/web/fetch/headers.js'
+import { Headers, getHeadersList } from '../../lib/web/fetch/headers.js'
 
 const headers = new Headers(
   [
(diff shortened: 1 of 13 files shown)


ronag (approved):

mcollina (approved):
lgtm

Issue #3255, opened 2024-05-13
Interceptors: add response decompress interceptor


metcoder95:
## This would solve...

Decompression of compressed responses based on `content-encoding` header.

## The implementation should look like...

Implemented as interceptor

## Additional context

Ref: https://github.com/nodejs/undici/discussions/3253#discussion-6660897

metcoder95:
cc: @ronag

adrianfalleiro:
Will you accept a PR for this? I would like to implement this

tsctx:
PR welcome

adrianfalleiro:
Ok made a first pass at this. Please take a look https://github.com/nodejs/undici/pull/3274

metcoder95:
Done by #4316

PR #3274, opened 2024-05-18
feat: add decompress interceptor

adrianfalleiro:
Added a new interceptor to decompress the response body. I lifted some of the implementation from the way decompression is handled in the [fetch client](https://github.com/nodejs/undici/blob/main/lib/web/fetch/index.js#L2139-L2161)

## This relates to...

- https://github.com/nodejs/undici/issues/3255

## Rationale

- See discussion on https://github.com/nodejs/undici/discussions/3253#discussion-6660897

## Changes

- Adds a response interceptor to decompress gzip, brotli and deflate responses

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready


Uzlopak:
Should we also handle rawdeflate?

adrianfalleiro:
> Should we also handle rawdeflate?

Yep, I'm using the existing `createInflate()` util from the fetch client which handles both deflate and deflateRaw so it should handle both already. I'll add a test.

test/interceptors/decompress.js
```diff
@@ -0,0 +1,321 @@
...
+  const client = new Client(
+    `http://localhost:${server.address().port}`
+  ).compose(decompress())
+
+  after(async () => {
+    await client.close()
+
+    server.close()
+    await once(server, 'close')
+  })
```
tsctx, line 77: use 
‎closeClientAndServerAsPromise‎.

https://github.com/nodejs/undici/blob/main/test/utils/node-http.js

lib/interceptor/decompress.js
```diff
@@ -0,0 +1,114 @@
...
+
+  #inputStream = null
+  #trailers = null
+
+  constructor (opts, handler) {
+    super(handler)
+    this.#handler = handler
+    this.#opts = opts
+  }
+
```
ronag, line 23:
```suggestion
onConnect(abort) {
  this.#inputStream = null
  this.#trailers = null
  return this.#handler.onConnect(abort)
}
```

lib/interceptor/decompress.js
```diff
@@ -0,0 +1,114 @@
...
+                this.#handler.onError(err)
+              } else {
+                this.#handler.onComplete(this.#trailers)
+              }
+            }
+          ).on('data', (chunk) => this.#handler.onData(chunk))
+        } else {
+          this.#inputStream.on('data', (chunk) => this.#handler.onData(chunk))
+          this.#inputStream.on('end', () => this.#handler.onComplete(this.#trailers))
+          this.#inputStream.on('error', (err) => this.#handler.onError(err))
```
ronag, line 75:
```suggestion
onError should not be called after onComplete, which is possible atm
```

test/interceptors/decompress.js
adrianfalleiro: Thanks, fixed

lib/interceptor/decompress.js
adrianfalleiro: Ah good catch. I refactored it as follows

```js
#onDecompressStreamFinished (err) {
  if (err) {
    this.#handler.onError(err)
  } else {
    this.#handler.onComplete(this.#trailers)
  }
}

if (decoders.length !== 0) {
  const [firstDecoder, ...restDecoders] = decoders
  this.#inputStream = firstDecoder

  if (restDecoders.length !== 0) {
    pipeline(
      this.#inputStream,
      ...restDecoders,
      this.#onDecompressStreamFinished.bind(this)
    ).on('data', (chunk) => this.#handler.onData(chunk))
  } else {
    finished(
      this.#inputStream.on('data', (chunk) => this.#handler.onData(chunk)),
      this.#onDecompressStreamFinished.bind(this)
    )
  }
}
```

lib/interceptor/decompress.js
```diff
@@ -0,0 +1,123 @@
...
+      if (decoders.length !== 0) {
+        const [firstDecoder, ...restDecoders] = decoders
+        this.#inputStream = firstDecoder
+
+        if (restDecoders.length !== 0) {
+          pipeline(
+            this.#inputStream,
+            ...restDecoders,
+            this.#onDecompressStreamFinished.bind(this)
+          ).on('data', (chunk) => this.#handler.onData(chunk))
```
ronag, line 79: Missing backpressure

lib/interceptor/decompress.js
```diff
@@ -0,0 +1,123 @@
...
+      this.#trailers = trailers
+      this.#inputStream.end()
+      return
+    }
+
+    return this.#handler.onComplete(trailers)
+  }
+}
+
+function createDecompressionInterceptor () {
```
metcoder95, line 115: We might need to support add the hinting of supported encoding in case implementers wants to limit this (I can only think for performance reasons).
They can pass a set of encodings supported, and the decompress decorator hints it using [`accept-encoding`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding)

metcoder95 (commented):
Documentation seems lost

lib/interceptor/decompress.js
```diff
@@ -0,0 +1,123 @@
...
+
+  constructor (opts, handler) {
+    super(handler)
+    this.#handler = handler
+    this.#opts = opts
+  }
+
+  onConnect (abort) {
+    this.#inputStream = null
+    this.#trailers = null
```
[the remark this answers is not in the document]
ronag: They are needed. OnConnect may be called multiple times and the handler should reset I that case

lib/interceptor/decompress.js
```diff
@@ -0,0 +1,123 @@
...
+            finishFlush: zlib.constants.Z_SYNC_FLUSH
+          }))
+        } else if (encoding === 'deflate') {
+          decoders.push(createInflate())
+        } else if (encoding === 'br') {
+          decoders.push(zlib.createBrotliDecompress())
+        } else {
+          decoders.length = 0
+          break
+        }
```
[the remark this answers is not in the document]
tsctx: break statement is present and cannot be converted.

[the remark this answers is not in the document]
metcoder95: Oh wow, it really seems that the days off payoff

lib/interceptor/decompress.js
```diff
@@ -0,0 +1,131 @@
...
+          finished(
+            this.#inputStream,
+            this.#onDecompressStreamFinished.bind(this)
+          )
+        }
+
+        outputStream.on('data', (chunk) => {
+          if (!this.#handler.onData(chunk)) {
+            this.#inputStream.pause()
+            this.#inputStream.once('drain', () => this.#inputStream.resume())
```
ronag, line 91: Those is wrong. Drain is signaled through the resume function sent to onHeaders

ronag (changes requested):

lib/interceptor/decompress.js
```diff
@@ -0,0 +1,131 @@
...
+    const parsedHeaders = util.parseHeaders(rawHeaders)
+    const contentEncoding = parsedHeaders['content-encoding']
+    const requestEncodings = contentEncoding ? contentEncoding.split(',').map(e => e.trim().toLowerCase()) : []
+
+    const { method } = this.#opts
+
+    // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding
+    if (requestEncodings.length !== 0 && method !== 'HEAD' && method !== 'CONNECT' && !nullBodyStatus.includes(statusCode)) {
+      const decoders = []
+      for (let i = 0; i < requestEncodings.length; ++i) {
```
tsctx, line 48:
```suggestion
      for (let i = requestEncodings.length - 1; i >=0; --i) {
```

Uzlopak: The other way round, right?

tsctx: Yeah, I made the same changes #3343.

adrianfalleiro: Thanks, I'll come back to this PR this week

lib/interceptor/decompress.js
```diff
@@ -0,0 +1,131 @@
...
+        outputStream.on('data', (chunk) => {
+          if (!this.#handler.onData(chunk)) {
+            this.#inputStream.pause()
+            this.#inputStream.once('drain', () => this.#inputStream.resume())
+          }
+        })
+      }
+    }
+
+    return this.#handler.onHeaders(
```
tjhiggins, line 97: Thanks for putting up this pr! I added this interceptor into my project. 

This interceptor could also be used with fetch if you remove the `content-encoding`. My use case is I have a LoggingInterceptor that sends the response body to GCP logging. This worked great except one API sends a gzipped response. Now I have the following flow:

Decompress -> Logging -> Fetch

```javascript
    delete parsedHeaders['content-encoding'];
    const newRawHeaders = Object.entries(parsedHeaders)
      .flat()
      .map((e) => Buffer.from(e));

    return this.#handler.onHeaders(statusCode, newRawHeaders, resume, statusMessage);
```

tjhiggins: Welp - I ended up having issues with the interceptor. Haven't been able to dive in yet, but I feel like my flow should work. Will try again once this is reviewed/merged.

tjhiggins:
Last piece for this to work for me with fetch. Fetch does some sort of back pressuring. 

```ts
const wrappedResume = () => {
  if (this.#inputStream) {
    this.#inputStream.resume();
  }
  return resume();
};

return this.#handler.onHeaders!(statusCode, newRawHeaders, wrappedResume, statusText);
```

Full onHeaders
```ts
onHeaders(statusCode: number, rawHeaders: Buffer[], resume: () => void, statusText: string) {
  const parsedHeaders = util.parseHeaders(rawHeaders);
  const contentEncoding = parsedHeaders['content-encoding'] as string | undefined;
  const requestEncodings = contentEncoding ? contentEncoding.split(',').map((e) => e.trim().toLowerCase()) : [];

  const { method } = this.#opts;

  // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding
  if (
    requestEncodings.length !== 0 &&
    method !== 'HEAD' &&
    method !== 'CONNECT' &&
    !nullBodyStatus.includes(statusCode)
  ) {
    const decoders: Transform[] = [];
    for (let i = requestEncodings.length - 1; i >= 0; --i) {
      const requestEncoding = requestEncodings[i];
      // https://www.rfc-editor.org/rfc/rfc9112.html#section-7.2
      if (requestEncoding === 'x-gzip' || requestEncoding === 'gzip') {
        decoders.push(
          zlib.createGunzip({
            // Be less strict when decoding compressed responses, since sometimes
            // servers send slightly invalid responses that are still accepted
            // by common browsers.
            // Always using Z_SYNC_FLUSH is what cURL does.
            flush: zlib.constants.Z_SYNC_FLUSH,
            finishFlush: zlib.constants.Z_SYNC_FLUSH,
          }),
        );
      } else if (requestEncoding === 'deflate') {
        throw new NotImplementedException('deflate is not supported');
      } else if (requestEncoding === 'br') {
        decoders.push(zlib.createBrotliDecompress());
      } else {
        decoders.length = 0;
        break;
      }
    }

    if (decoders.length !== 0) {
      const [firstDecoder, ...restDecoders] = decoders;
      this.#inputStream = firstDecoder;
      this.#inputStream.on('drain', () => {
        if (this.#inputStream) {
          this.#inputStream.resume();
        }
      });
      let outputStream = firstDecoder;

      if (restDecoders.length !== 0) {
        // eslint-disable-next-line @typescript-eslint/ban-ts-comment
        // @ts-expect-error
        outputStream = pipeline(this.#inputStream, ...restDecoders, this.#onDecompressStreamFinished.bind(this));
      } else {
        finished(this.#inputStream, this.#onDecompressStreamFinished.bind(this));
      }

      outputStream.on('data', (chunk) => {
        if (!this.#handler.onData!(chunk)) {
          if (this.#inputStream) {
            this.#inputStream.pause();
          }
        }
      });
    }
  }

  delete parsedHeaders['content-encoding'];
  const newRawHeaders = Object.entries(parsedHeaders)
    .map(([key, value]) => {
      if (Array.isArray(value)) {
        return value.map((v) => [key, v]).flat();
      } else {
        return [key, value];
      }
    })
    .flat()
    .map((v) => Buffer.from(v));

  const wrappedResume = () => {
    if (this.#inputStream) {
      this.#inputStream.resume();
    }
    return resume();
  };

  return this.#handler.onHeaders!(statusCode, newRawHeaders, wrappedResume, statusText);
}
```

adrianfalleiro:
@tjhiggins Do you want to collaborate on this PR with me? I can give you commit access to the base branch. 

I think the interceptor interface has changed a little since I opened this PR. I can work on ensuring it works with the latest version of Undici. 

Once I've done that you, do you want to add the changes above for onHeaders and backpressure? Sound good?

tjhiggins:
> @tjhiggins Do you want to collaborate on this PR with me? I can give you commit access to the base branch.
> 
> I think the interceptor interface has changed a little since I opened this PR. I can work on ensuring it works with the latest version of Undici.
> 
> Once I've done that you, do you want to add the changes above for onHeaders and backpressure? Sound good?

I only just saw the new hooks for v7. Fetch does decompress, but after the interceptors in v6. Maybe with the new hook `onResponseData` I don't even need an interceptor. I'll wait for you to update before making any changes. Not in a huge hurry to spend more time on this - now that I have it working for v6.

Uzlopak:
I basically forgot this PR1

We have now decompressInterceptor, but we have to tackle the streaming properly.

PR #3279, opened 2024-05-20
fix: Fix server closing in tests.

ShogunPanda:
## This relates to...

https://github.com/nodejs/node/pull/53028

## Rationale

When running in 18.20.3, the test suite fails to stop the test as the HTTP server does not close if some connections are still pending.

## Changes

This PR changes the test utils to close pending connection first before closing the server. This is already the default behavior in Node 20+.

### Features

N/A

### Bug Fixes

N/A

N/A
## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/test/utils/node-http.js
+++ b/test/utils/node-http.js
@@ -3,7 +3,10 @@
 const util = require('node:util')
 
 function closeServerAsPromise (server) {
-  return () => util.promisify(server.close.bind(server))()
+  return () => {
+    server.closeIdleConnections()
+    return util.promisify(server.close.bind(server))()
+  }
 }
 
 function closeClientAndServerAsPromise (client, server) {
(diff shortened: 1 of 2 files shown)


test/utils/node-http.js
```diff
@@ -3,7 +3,10 @@
 const util = require('node:util')
 
 function closeServerAsPromise (server) {
-  return () => util.promisify(server.close.bind(server))()
+  return () => {
+    server.closeIdleConnections()
```
mcollina, line 7: Is this needed?

ShogunPanda: I tried to change the order of the `t.after` in a test but there a couple others exhibiting the same behavior. I therefore chose to act at the source rather than each specific case.

PR #3283, opened 2024-05-21
websocket: use FixedQueue instead of Set

tsctx:
@KhafraDev

### DIFF
--- a/lib/web/websocket/sender.js
+++ b/lib/web/websocket/sender.js
@@ -2,13 +2,13 @@
 
 const { WebsocketFrameSend } = require('./frame')
 const { opcodes, sendHints } = require('./constants')
+const FixedQueue = require('../../dispatcher/fixed-queue')
 
 /** @type {typeof Uint8Array} */
 const FastBuffer = Buffer[Symbol.species]
 
 /**
  * @typedef {object} SendQueueNode
- * @property {SendQueueNode | null} next
  * @property {Promise<void> | null} promise
  * @property {((...args: any[]) => any)} callback
  * @property {Buffer | null} frame
@@ -16,13 +16,9 @@ const FastBuffer = Buffer[Symbol.species]
 
 class SendQueue {
   /**
-   * @type {SendQueueNode | null}
+   * @type {FixedQueue | null}
    */
-  #head = null
-  /**
-   * @type {SendQueueNode | null}
-   */
-  #tail = null
+  #queue = null
 
   /**
    * @type {boolean}
@@ -45,22 +41,20 @@ class SendQueue {
       } else {
         /** @type {SendQueueNode} */
         const node = {
-          next: null,
           promise: null,
           callback: cb,
           frame
         }
-        if (this.#tail !== null) {
-          this.#tail.next = node
+        if (this.#queue === null) {
+          this.#queue = new FixedQueue()
         }
-        this.#tail = node
+        this.#queue.push(node)
       }
       return
     }
 
     /** @type {SendQueueNode} */
     const node = {
-      next: null,
       promise: item.arrayBuffer().then((ab) => {
         node.promise = null
         node.frame = createFrame(ab, hint)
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

tsctx:
> Why not `fixed_queue` from node core?

There is no particular reason for this.

ronag:
> In this use case, I prefer the simple linkedlist; Initialization of the array of 2048 entries takes time :)

I think that is amortized quite quickly...

tsctx:
> do you have benchmarks?

What about?

tsctx:
If it was about sending data, it would be 1.2 times faster since we changed to zero-copy.

KhafraDev:
Yeah, sending messages. Do you have the data to backup the 1.2x faster claim?

tsctx:
okey, look this

```
benchmark      time (avg)             (min … max)       p75       p99      p999
• send
undici      1'467 µs/iter   (1'187 µs … 6'165 µs)  1'442 µs  4'022 µs  6'165 µs <- before
undici      1'658 µs/iter   (1'305 µs … 4'145 µs)  1'673 µs  3'043 µs  4'145 µs <- after
```

tsctx:
> > In this use case, I prefer the simple linkedlist; Initialization of the array of 2048 entries takes time :)
> 
> I think that is amortized quite quickly...

Rewrite with fixed_queue!

KhafraDev (approved):
why is queue being lazily created?

KhafraDev (changes requested):
tests are failing

tsctx:
I don't have time right now, so I will rewrite it in another PR.

lib/web/websocket/sender.js
```diff
@@ -2,15 +2,30 @@
...
+ * @property {Buffer | null} frame
+ */
+
 class SendQueue {
-  #queued = new Set()
-  #size = 0
+  /**
+   * @type {FixedQueue | null}
+   */
+  #queue = null
```
KhafraDev, line 21:
```suggestion
  /**
   * @type {FixedQueue}
   */
  #queue = new FixedQueue()
```

lib/web/websocket/sender.js
```diff
@@ -19,66 +34,78 @@ class SendQueue {
...
-        this.#run()
+        /** @type {SendQueueNode} */
+        const node = {
+          promise: null,
+          callback: cb,
+          frame
+        }
+        if (this.#queue === null) {
+          this.#queue = new FixedQueue()
+        }
```
KhafraDev, line 50:
```suggestion
```

lib/web/websocket/sender.js
```diff
@@ -19,66 +34,78 @@ class SendQueue {
...
+        node.frame = createFrame(ab, hint)
+      }),
+      callback: cb,
+      frame: null
+    }
 
-      if (!done) return
+    if (this.#queue === null) {
+      this.#queue = new FixedQueue()
+    }
```
KhafraDev, line 68:
```suggestion
```

lib/web/websocket/sender.js
```diff
@@ -19,66 +34,78 @@ class SendQueue {
...
-    const frame = new WebsocketFrameSend()
-    const opcode = hint === sendHints.string ? opcodes.TEXT : opcodes.BINARY
-
-    frame.frameData = data
-    const buffer = frame.createFrame(opcode)
-
-    this.#socket.write(buffer, cb)
+  async #run () {
+    this.#running = true
+    /** @type {FixedQueue} */
```
tsctx, line 79:
```suggestion
```

lib/web/websocket/sender.js
```diff
@@ -19,66 +34,71 @@ class SendQueue {
...
+    const node = {
+      promise: item.arrayBuffer().then((ab) => {
+        node.promise = null
+        node.frame = createFrame(ab, hint)
+      }),
+      callback: cb,
+      frame: null
+    }
 
-      if (!done) return
```
tsctx, line 63:
```suggestion
```

KhafraDev:
Thanks, WebSocket is bidirectional so it doesn't make sense to only be receiving messages (which is the only benefit I can see in lazily creating the queue?). We also have EventSource for that use case.

tsctx:
When I received the message, I mistakenly thought I was generating a queue.

So there is no advantage.

KhafraDev (approved):

PR #3284, opened 2024-05-21
websocket: pre-calculated length

tsctx:


lib/web/websocket/util.js
```diff
@@ -294,6 +294,29 @@ const utf8Decode = hasIntl
     throw new TypeError('Invalid utf-8 received.')
   }
 
+function bufferConcat (buffers, length) {
```
tsctx, line 297: Avoid instance assertion in Buffer.concat.

Issue #3289, opened 2024-05-21
autobahn workflow is always 1 commit behind


KhafraDev:
https://github.com/nodejs/undici/blob/main/.github/workflows/autobahn.yml

@Uzlopak

Uzlopak:
I will implement it in few hours.

PR #3290, opened 2024-05-21
revert: "websocket: pre-calculated length"

KhafraDev:
This reverts 13523fd7fe66a4238f9ee79326b64cac4fc47d15 which broke permessage-deflate support.

See https://github.com/nodejs/undici/issues/3289 for why the tests did not catch this.

### DIFF
--- a/lib/web/websocket/receiver.js
+++ b/lib/web/websocket/receiver.js
@@ -240,8 +240,8 @@ class ByteParser extends Writable {
 
               this.#loop = true
               this.#state = parserStates.INFO
-              this.run(callback)
               this.#fragments.length = 0
+              this.run(callback)
             })
 
             this.#loop = false
--- a/lib/web/websocket/receiver.js
+++ b/lib/web/websocket/receiver.js
@@ -11,7 +11,6 @@ const {
   failWebsocketConnection,
   websocketMessageReceived,
   utf8Decode,
-  bufferConcat,
   isControlFrame,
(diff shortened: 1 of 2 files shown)


mcollina (dismissed):
Can you add a test to avoid further regressions?

KhafraDev:
The tests catch it already, it's just that the Autobahn workflow doesn't quite work as expected and it uses a previous commit rather than the current one. Once the issue linked is resolved it'll be fixed.

KhafraDev:
The tests in https://github.com/nodejs/undici/pull/3283 are failing because of this, or you can see the first comment that the bot made before the edit. There is a lot broken right now I'd appreciate if we could land this in the meantime.

Uzlopak:
@mcollina 
Please remove your blocking review.

Uzlopak:
@KhafraDev 
Is it normal that the autobahn workflow is so slow?

KhafraDev:
Yes, you have to understand that these tests send GB of frames over that need to parsed and mirrored to the server. The longer the tests, the more likely the tests are to succeed because the server will instantly fail/close the connection if it receives something unexpected.

Uzlopak:
Ah, ok. ;)

Uzlopak (approved):

KhafraDev:
cc @tsctx I'm pretty positive that the issue came with the custom Buffer.concat.

PR #3291, opened 2024-05-21
ci: fix autobahn workflow

Uzlopak:
@khafraDev

### DIFF
--- a/.github/workflows/autobahn.yml
+++ b/.github/workflows/autobahn.yml
@@ -2,7 +2,7 @@ name: Autobahn
 on:
   workflow_dispatch:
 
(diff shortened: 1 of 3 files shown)


KhafraDev:
@Uzlopak there's no markdown on the comment anymore

Uzlopak:
Yeah, removing the markdown stuff. We dont make any pr comment, which was always a pita to get it right. So the simple solution is to do the normal pull_request event.

Uzlopak:
Yes, but I think #3290 should be merged first.

KhafraDev:
I'll rebase to make sure it works

PR #3294, opened 2024-05-22
fix: retry on body support

### LINKED ISSUE #3288: Retry Handler Fails When Request Has a Body
## Bug Description

When a body is provided to a request, such as in a `PUT` request, the retry handler fails with `UND_ERR_REQ_CONTENT_LENGTH_MISMATCH`.

## Reproducible By

repro.ts:

```
import http from 'http';
import { Agent, RetryAgent } from 'undici';

const myDispatcher = new RetryAgent(new Agent());

async function failListener(req, res) {
  res.setHeader('content-type', 'application/json');
  res.writeHead(500);
  res.end(`{"message": "failure"}`);
}

async function makeRetryRequest() {
  const result = await fetch('http://localhost:3333/fail', {
    method: 'PUT',
    body: JSON.stringify({ foo: 'bar', baz: 'bat' }),
    headers: {
      'content-type': 'application/json',
    },
    dispatcher: myDispatcher,
  });
  console.log(result);
}

async function main() {
  http.createServer(failListener).listen(3333, makeRetryRequest);
}

main();
```

```
npx tsx repro.ts
```

## Expected Behavior

The request should be retried and should not fail because `undici` fails to handle re-use of the `body` correctly.

## Logs & Screenshots

```
> npx tsx repro.ts
node:internal/deps/undici/undici:11372
    Error.captureStackTrace(err, this);
          ^

TypeError: fetch failed
    at Object.fetch (node:internal/deps/undici/undici:11372:11)
    at Server.makeRetryRequest REDACTED/repro.ts:13:18) {
  cause: RequestContentLengthMismatchError: Request body length does not match content-length header
      at AsyncWriter.end REDACTED/node_modules/undici/lib/dispatcher/client-h1.js:1319:15)
      at writeIterable (REDACTED/node_modules/undici/lib/dispatcher/client-h1.js:1201:12) {
    code: 'UND_ERR_REQ_CONTENT_LENGTH_MISMATCH'
  }
}

Node.js v20.8.1
```

## Environment

* macOS 14.4.1
* nodejs v20.8.1
* undici 6.18.0

### Additional context

None

metcoder95:
## This relates to...

Closes #3288

## Status

- [x] Tested
- [x] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/core/util.js
+++ b/lib/core/util.js
@@ -1,18 +1,71 @@
 'use strict'
 
 const assert = require('node:assert')
-const { kDestroyed, kBodyUsed, kListeners } = require('./symbols')
+const { kDestroyed, kBodyUsed, kListeners, kBody } = require('./symbols')
 const { IncomingMessage } = require('node:http')
 const stream = require('node:stream')
 const net = require('node:net')
-const { InvalidArgumentError } = require('./errors')
 const { Blob } = require('node:buffer')
 const nodeUtil = require('node:util')
 const { stringify } = require('node:querystring')
+const { EventEmitter: EE } = require('node:events')
+const { InvalidArgumentError } = require('./errors')
 const { headerNameLowerCasedRecord } = require('./constants')
 const { tree } = require('./tree')
 
 const [nodeMajor, nodeMinor] = process.versions.node.split('.').map(v => Number(v))
+
+class BodyAsyncIterable {
+  constructor (body) {
+    this[kBody] = body
+    this[kBodyUsed] = false
+  }
+
+  async * [Symbol.asyncIterator] () {
+    assert(!this[kBodyUsed], 'disturbed')
+    this[kBodyUsed] = true
+    yield * this[kBody]
+  }
+}
+
+function wrapRequestBody (body) {
+  if (isStream(body)) {
+    // TODO (fix): Provide some way for the user to cache the file to e.g. /tmp
+    // so that it can be dispatched again?
+    // TODO (fix): Do we need 100-expect support to provide a way to do this properly?
+    if (bodyLength(body) === 0) {
+      body
+        .on('data', function () {
+          assert(false)
+        })
+    }
+
+    if (typeof body.readableDidRead !== 'boolean') {
+      body[kBodyUsed] = false
+      EE.prototype.on.call(body, 'data', function () {
+        this[kBodyUsed] = true
+      })
+    }
+
+    return body
+  } else if (body && typeof body.pipeTo === 'function') {
+    // TODO (fix): We can't access ReadableStream internal state
+    // to determine whether or not it has been disturbed. This is just
+    // a workaround.
+    return new BodyAsyncIterable(body)
+  } else if (
+    body &&
+    typeof body !== 'string' &&
+    !ArrayBuffer.isView(body) &&
+    isIterable(body)
+  ) {
+    // TODO: Should we allow re-using iterable if !this.opts.idempotent
+    // or through some other flag?
+    return new BodyAsyncIterable(body)
+  } else {
+    return body
+  }
+}
 
 function nop () {}
 
@@ -634,5 +687,6 @@ module.exports = {
   isHttpOrHttpsPrefixed,
   nodeMajor,
   nodeMinor,
-  safeHTTPMethods: ['GET', 'HEAD', 'OPTIONS', 'TRACE']
+  safeHTTPMethods: ['GET', 'HEAD', 'OPTIONS', 'TRACE'],
+  wrapRequestBody
 }
--- a/lib/handler/retry-handler.js
+++ b/lib/handler/retry-handler.js
@@ -3,7 +3,14 @@ const assert = require('node:assert')
 
 const { kRetryHandlerDefaultRetry } = require('../core/symbols')
 const { RequestRetryError } = require('../core/errors')
-const { isDisturbed, parseHeaders, parseRangeHeader } = require('../core/util')
+const {
+  isStream,
+  isDisturbed,
+  isIterable,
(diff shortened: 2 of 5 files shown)


metcoder95:
I'd need to adjust testing and also scope properly the validation for the body; will try to do it Today's evening

ronag (commented):
isDisturbed already handles this. not sure what this is trying to fix?

metcoder95:
Yeah, the `isDisturbed` handles the stream correctly, but the `AsyncIterator` or `Iterator` does not behaves accordingly; we should also throw when receiving these as body as they have states and when consumed, they are not longer usable.

metcoder95:
Saw what you mean, the issue is that the handler was not wrapping the body as done for redirect handler e.g.

I made an abstraction and will open other PRs to remove the duplication and use the abstraction instead

lib/core/util.js
```diff
@@ -1,19 +1,72 @@
...
+    // a workaround.
+    return new BodyAsyncIterable(body)
+  } else if (
+    body &&
+    typeof body !== 'string' &&
+    !ArrayBuffer.isView(body) &&
+    isIterable(body)
+  ) {
+    // TODO: Should we allow re-using iterable if !this.opts.idempotent
+    // or through some other flag?
```
ronag, line 63: An iterable cannot be re-used, ever.

mcollina (changes requested):
I'm -1 to this implementation.

I don't think POST, PUT etc should be retried. If we want to retry them, we need to provide an on-disk store and write the content down there, so it can be resumed. The user should enable this manually.

lib/core/util.js
ronag: Ah, unless you never read from it I guess

lib/handler/retry-handler.js
```diff
@@ -63,6 +68,7 @@ class RetryHandler {
     this.end = null
     this.etag = null
     this.resume = null
+    this.opts.body = wrapRequestBody(this.opts.body)
```
ronag, line 71: Don't. Mutate the users object

ronag (changes requested):

metcoder95:
> I'm -1 to this implementation.
> 
> I don't think POST, PUT etc should be retried. If we want to retry them, we need to provide an on-disk store and write the content down there, so it can be resumed. The user should enable this manually.

So you suggest to just remove them from the `methods` defaults?
That's fine by me, I can take the other way around and only retry `safe-methods` instead of `idempotent` ones.
I'd still try to abstract that class as is used across two implementations

lib/core/util.js
metcoder95: Yeah, that's the only possible scenario; tho maybe not something break our heads on this rn but for the future. I might try to decuple that from this PR so we can discuss it better 👍

ronag:
PUT should be idempotent so I don't see why not retry. Either way we should retry if the body was never sent.

ronag:
But I don't think that has anything to do with this PR?

ronag:
Part of the problem is that you are using fetch? Retries are not really part of the standard.

sdelano:
> Part of the problem is that you are using fetch? Retries are not really part of the standard.

Okay, that's fair I guess. I'm using the `node` implementation of `fetch`, which is powered by `undici`, and both of these projects are under the `nodejs` organization in GitHub. There appears to be a tight coupling here, and I was hoping to be able to use some of the more advanced features of `undici` to power some libraries that are built on `fetch`, the standard.

I took what you said above literally, in that "this problem doesn't exist if you don't use `fetch`", and you're right. If I replace `fetch` with `undici.request` in my original bug report, I no longer get the same failure as before.

At this point I've worked around this issue by using `fetch-retry` since I can't actually use `undici` directly in a NextJS project as there is a bug in their compiler. Maybe at some point in the future I'll dig into why exactly `undici` isn't handling the closing of the connection on the other end gracefully here.

mcollina:
@metcoder95 PUT POST etc are not idempotent ;).

Note that this does not fix the issue. That fix should be to not convert to a stream things that are not.

metcoder95:
> @metcoder95 PUT POST etc are not idempotent ;).
> 
> Note that this does not fix the issue. That fix should be to not convert to a stream things that are not.

Oh sure, I'm not implying that `POST` is idempotent 😅 
Rather, I was implying about that decision we made about the [default methods we support](https://github.com/nodejs/undici/blob/a4879b4663f1cbe190bcd6e62879ab4f994a0409/lib/handler/retry-handler.js#L43) to retry, which are idempotent.

As @sdelano exemplifies, the `RetryHandler` enables implementations to give it a try to request at their need, so even if we have defaults, they can always provide a callback that bypasses the defaults and execute retries on custom scenarios, the state assurance should be assume by the user and is something we cannot control. I'd say that that's our scope is to provide the mechanisms to ensure correctness as much as we can.

The biggest issue comes with stateful bodies (i.e. `streams` and `iterators`) that they should be backed with on-disk strategies to allow retries, which is something I'm not fully convinced to support (maybe just not yet).

The PR just attempts to identify and **not retry** on stateful bodies (buffers and strings can be retried "_safely_" up to the user).

I can also extend documentation to highlight this.

Are we ok moving this way or we prefer to take another direction?

mcollina (approved):
LGTM

lib/handler/retry-handler.js
ronag, line 71:
```suggestion
    this.opts = { ...opts, body: wrapRequestBody(this.opts.body) }
```

ronag (approved):

ronag (approved):

PR #3286, opened 2024-05-21
fix: compatibility for global headers

tsctx:
Fixes #3285

cc: @mcollina

PR #3295, opened 2024-05-22
don't use internal header state for cookies

KhafraDev:
refs: https://github.com/nodejs/undici/pull/3286

I have no idea why we didn't do this before 🤷


metcoder95 (approved):

tsctx (approved):
lgtm

PR #3302, opened 2024-05-29
perf: convert object to params

DarkGL:
Convert passing object to passing params to `writeBuffer` `writeIterable` `writeBlob` `writeStream`

## Rationale

Generally speaking constructing new object and passing it to function is more costly than just passing params.

Also using [deoptexplorer-vscode](https://github.com/microsoft/deoptexplorer-vscode) I have found that writeBuffer was deoptimized probably because off passing different shape of object.

## Changes

- Convert passing object to passing params to `writeBuffer` `writeIterable` `writeBlob` `writeStream`

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -479,72 +479,70 @@ function writeH2 (client, request) {
     if (!body || contentLength === 0) {
       writeBuffer(
         abort,
-        client,
-        request,
-        contentLength,
-        expectsPayload,
         stream,
         null,
-        client[kSocket]
+        client,
+        request,
+        client[kSocket],
+        contentLength,
+        expectsPayload
       )
     } else if (util.isBuffer(body)) {
       writeBuffer(
         abort,
+        stream,
+        expectsPayload,
         client,
         request,
+        client[kSocket],
         contentLength,
-        body,
-        expectsPayload,
-        stream,
-        client[kSocket]
+        body
       )
     } else if (util.isBlobLike(body)) {
       if (typeof body.stream === 'function') {
         writeIterable(
           abort,
-          client,
-          request,
-          contentLength,
-          expectsPayload,
           stream,
           body.stream(),
-          client[kSocket]
+          client,
+          request,
+          client[kSocket],
+          contentLength,
+          expectsPayload
         )
       } else {
         writeBlob(
           abort,
+          stream,
           body,
           client,
           request,
+          client[kSocket],
           contentLength,
-          expectsPayload,
-          stream,
-          client[kSocket]
+          expectsPayload
         )
       }
     } else if (util.isStream(body)) {
       writeStream(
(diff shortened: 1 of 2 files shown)


DarkGL:
Tests failing , should I progress with it ?

metcoder95:
It shouldn't be easier to just fill the gaps of the object to sync its shape?

What ware the comparison results?

tsctx:
do you have any benchmark results? I don't think this impromving performance.

tsctx:
do you have any benchmark results? I don't think improving performance.

ronag:
I don't think this is meaningful

DarkGL:
> do you have any benchmark results? I don't think improving performance.

I currently don't have access to pc but

https://x.com/tmikov/status/1763790902458560658 this is from the author of static hermes

mcollina:
> I don't think this is meaningful

It was the last time I checked. Converting some huge params to objects in Fastify had a slowdown.

metcoder95:
I'm interested into the deoptimization piece, that's for sure something we can dig into it and maybe start from there

DarkGL:
Simple test

```javascript

createServer((req, res) => {
    res.end('hello world');
}).listen(3000);
```

```javascript

console.time('fetch');
for (let i = 0; i < 100000; i++) {
    const req = await request('http://localhost:3000');
    await req.body.text();
}
console.timeEnd('fetch');
```

image

image

image

lib/dispatcher/client-h2.js
metcoder95, line 1: If moving this path, you'll need to adjust the functions's body to the new signature

metcoder95:
The benchmarks shares ~6% difference with `main`. I'm a little bit over the fence due to maintainability of the positional args, but if we are ok with that, it LGTM

lib/dispatcher/client-h2.js
DarkGL: What do you mean exactly ? I have corrected the order of passed params.

DarkGL:
> The benchmarks shares ~6% difference with `main`. I'm a little bit over the fence due to maintainability of the positional args, but if we are ok with that, it LGTM

Thanks, I will try to fix tests in upcoming days

lib/dispatcher/client-h2.js
metcoder95: The refactor of the positional args will require to double-check you are receiving the right args

DarkGL: Ah yes sure, I already tried to correct it, probably will catch any mistake when fixing tests :)

DarkGL:
Can we rerun Fuzzing ? It's working fine on my pc

metcoder95 (approved):

mcollina (approved):
lgtm

Issue #3312, opened 2024-06-05
publish autobahn test results on the docs site


KhafraDev:
The autobahn report generates a couple hundred html files as a report, having this published without having to run the suite for oneself could be useful IMO.

KhafraDev:
cc @Ethan-Arrowood

mcollina:
I'm +1. How can we generate it? Is there an npm script?

KhafraDev:
1. run the server with `docker run -it --rm     -v "${PWD}/config:/config"     -v "${PWD}/reports:/reports"     -p 9001:9001     --name fuzzingserver     crossbario/autobahn-testsuite`
2. `node test/autobahn/client.js`

mcollina:
How to get the report then?

KhafraDev:
after the node script finishes it generates the report automatically in test/autobahn/reports/clients/

KhafraDev:
I've published the results onto my own .github.io site https://khafradev.github.io/autobahn/clients/index.html

PR #3317, opened 2024-06-08
remove autobahn reports from gitignore

KhafraDev:
part of #3312


tsctx:
json may not be necessary

tsctx (approved):
lgtm

KhafraDev:
it's not necessary for the report but could be useful for detecting test result changes

tsctx (approved):
lgtm

mcollina (changes requested):
I'm -1. These are over 1000 files with a total of 353468 lines. If we kept updating them, they would engulf our github repository. We need a different strategy for this.

I didn't expect the autobahn report to be this big. Is there a summary possible?

KhafraDev:
They won't need to be updated much - the tests are already passing. If we remove all of the html/json files other than index.html we lose out on any benefit of publishing it.

mcollina:
@ronag @Ethan-Arrowood have you got a solution for this?

Issue #3308, opened 2024-06-04
'interceptors' export doesn't work like in the docs


jrr:
## Bug Description

I'm trying to use the built-in retry interceptor, like in this [docs example](https://undici.nodejs.org/#/docs/api/Dispatcher?id=retry):

```ts
const { Client, interceptors } = require("undici");
const { retry } = interceptors;
```

```
error TS2614: Module '"undici"' has no exported member 'interceptors'. Did you mean to use 'import interceptors from "undici"' instead?

import { Client, interceptors } from "undici";
                 ~~~~~~~~~~~~
```

Following TypeScript's suggestion, this seems to work:

```ts
import { Client } from "undici";
import undici from "undici";
const { interceptors } = undici;
const { retry } = interceptors;
```

That's a decent workaround for now, but it'd be great if the export could work as the doc describes.

Using undici v6.18.2, the latest as of this writing.

PR #3318, opened 2024-06-10
export interceptor to type def file

jakecastelli:
Fixes: https://github.com/nodejs/undici/issues/3308

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/test/imports/undici-import.ts
+++ b/test/imports/undici-import.ts
@@ -1,5 +1,7 @@
 import { request } from '../../'
+import { interceptors } from '../../'
 
 async function exampleCode() {
+  const ri = interceptors.createRedirectInterceptor({})
   await request('http://localhost:3000/foo')
 }
--- a/test/imports/undici-import.ts
+++ b/test/imports/undici-import.ts
@@ -2,6 +2,9 @@ import { request } from '../../'
 import { interceptors } from '../../'
 
 async function exampleCode() {
-  const ri = interceptors.createRedirectInterceptor({})
+  const retry = interceptors.retry()
+  const rd = interceptors.redirect()
+  const dump = interceptors.dump()
+
   await request('http://localhost:3000/foo')
 }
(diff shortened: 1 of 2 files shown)


mcollina (changes requested):
Thanks for opening a PR! Can you please add a unit test? we use tsd for typescript

jakecastelli:
Sure, thanks for letting me know! Added an unit test to verify the behaviour.

<details><summary>Test would fail without the export to def file</summary>

```
> undici@6.18.2 test:typescript
> tsd && tsc --skipLibCheck test/imports/undici-import.ts

test/imports/undici-import.ts:2:10 - error TS2614: Module '"../../"' has no exported member 'interceptors'. Did you mean to use 'import interceptors from "../../"' instead?

2 import { interceptors } from '../../'
           ~~~~~~~~~~~~


Found 1 error in test/imports/undici-import.ts:2
```

</details>

I added a dummy call `interceptors.createRedirectInterceptor({})` just to ensure member can be accessed without any type error, not sure if its necessary.

mcollina (approved):
lgtm

test/imports/undici-import.ts
```diff
@@ -1,5 +1,7 @@
 import { request } from '../../'
+import { interceptors } from '../../'
 
 async function exampleCode() {
+  const ri = interceptors.createRedirectInterceptor({})
```
metcoder95, line 5: Can you add instead the `retry`, `redirect` and `dump` interceptors; the current `createRedirectInterceptor` will be dropped in next major

jakecastelli: sure, thanks for the headsup!

mcollina:
@metcoder95 ptal

metcoder95 (approved):

PR #3332, opened 2024-06-17
fix: interceptors.d.ts has no default export

### LINKED ISSUE #3330: undici/types/index.d.ts:21:8: Type error TS1192: Module '"undici@6.19.0/node_modules/undici/types/interceptors"' has no default export
## Bug Description

After upgrading to v6.19.0. When running the `check-dts` program to verify the types. The following error occurs:

```
undici/types/index.d.ts:21:8: Type error TS1192
│   Module '"undici@6.19.0/node_modules/undici/types/interceptors"' has no default export

```

## Reproducible By

Running `check-dts`

## Expected Behavior

No Type error

## Environment

Linux 6.9.4-arch1-1 x86_64 unknown

Node v22.3.0

### Additional context

Here is the line causing the error: https://github.com/nodejs/undici/blob/main/types/index.d.ts#L21

Came from https://github.com/nodejs/undici/pull/3318

Uzlopak:
Closes #3330 

We need to remove skipLibCheck to throw the error. Now we can detect the bug.

### DIFF
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
     "test:node-test": "borp -p \"test/node-test/**/*.js\"",
     "test:tdd": "borp --expose-gc -p \"test/*.js\"",
     "test:tdd:node-test": "borp -p \"test/node-test/**/*.js\" -w",
-    "test:typescript": "tsd && tsc --skipLibCheck test/imports/undici-import.ts",
+    "test:typescript": "tsd && tsc test/imports/undici-import.ts --typeRoots ./types && tsc ./types/*.d.ts --noEmit --typeRoots ./types",
     "test:webidl": "borp -p \"test/webidl/*.js\"",
     "test:websocket": "borp -p \"test/websocket/*.js\"",
     "test:websocket:autobahn": "node test/autobahn/client.js",
--- a/types/formdata.d.ts
+++ b/types/formdata.d.ts
@@ -2,7 +2,7 @@
 /// <reference types="node" />
 
 import { File } from './file'
-import { SpecIterator, SpecIterableIterator } from './fetch'
+import { SpecIterableIterator } from './fetch'
 
 /**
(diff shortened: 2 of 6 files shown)


types/formdata.d.ts
```diff
@@ -2,7 +2,7 @@
 /// <reference types="node" />
 
 import { File } from './file'
-import { SpecIterator, SpecIterableIterator } from './fetch'
+import { SpecIterableIterator } from './fetch'
```
Uzlopak, line 5: was unused import

types/webidl.d.ts
```diff
@@ -55,9 +55,7 @@ interface WebidlUtil {
     V: unknown,
     bitLength: number,
     signedness: 'signed' | 'unsigned',
-    opts?: ConvertToIntOpts,
-    prefix: string,
-    argument: string
+    opts?: ConvertToIntOpts
```
Uzlopak, line 58: tsc reported that after optional paramaters there can not be mandatory parameters.

Compared with implementation and the following arguments are not existing.

package.json
```diff
@@ -85,7 +85,7 @@
     "test:node-test": "borp -p \"test/node-test/**/*.js\"",
     "test:tdd": "borp --expose-gc -p \"test/*.js\"",
     "test:tdd:node-test": "borp -p \"test/node-test/**/*.js\" -w",
-    "test:typescript": "tsd && tsc --skipLibCheck test/imports/undici-import.ts",
+    "test:typescript": "tsd && tsc test/imports/undici-import.ts --typeRoots ./types && tsc ./types/*.d.ts --noEmit --typeRoots ./types",
```
Uzlopak, line 88: eslint is throwing typings errors. to avoid it we simulate an skipLibCheck by pointing to the types folder. this is not nice but eslint sucks. So this is a low level solution.

mcollina (approved):
lgtm

PR #3338, opened 2024-06-17
fix #3337

### LINKED ISSUE #3328: Failing JSON parsing in Node 22
> EDIT
>I managed to reproduce the error locally and changed the bug description, and reproduction steps accordingly

## Bug Description

When making a fetch call the JSON parsing fails and spits out some broken chars. This only happens in Node 22.3.0, **does not** happen in node 22.1.0

## Reproducible By

Fetching data from an API (in my case Wordpress) and parsing the body as json. Currently this happens in a Next.js project. When reaching my homepage I make a call to my cms

```bash
nvm use 22.3.0
```

```javascript
async function call(url) {
  const response = await fetch(`${WP_DEFAULT_URL}${url}`, {});
  return response.json();
}
```

## Expected Behavior

The JSON being parsed correctly. This behavior came up circa 3 days ago, I figure it has something to do with the latest [Node 22.3.0](https://github.com/nodejs/node/releases/tag/v22.3.0) release

## Logs & Screenshots

I added a trycatch to log the query if failing to see more clearly in the github logs, since I firstly bumped into the problem during a github action build
```
#9 28.25 SyntaxError: Unexpected non-whitespace character after JSON at position 2908 (line 1 column 2909)
#9 28.25     at JSON.parse (<anonymous>)
#9 28.25     at parseJSONFromBytes (node:internal/deps/undici/undici:5477:19)
#9 28.25     at successSteps (node:internal/deps/undici/undici:5459:27)
#9 28.25     at fullyReadBody (node:internal/deps/undici/undici:4378:9)
#9 28.25     at async consumeBody (node:internal/deps/undici/undici:5468:7)
#9 28.25     at async Object.getItem (/usr/src/app/.next/server/app/cerca/page.js:1:8661)
#9 28.25     at async o (/usr/src/app/.next/server/app/[lang]/(pages)/cerca/page.js:1:4250)
{... some other json from my cms...},
[{"name":"wp","href":"https:\/\/api.w.org\/{rel}","templated":true}]}}]���!*� �!*�ȉ!*�P�!*�x�!*���!*�.(�!*�@�!*�~!*���y�@�!*�b���!*���!*���L�c*��c*�`�*���*�^��!*���!*�c �!*� �!*��Ȋ!*�(�)*���c*��c*�(�)*� d*���L
#9 28.25 �

#9 28.25 ,���y����y���L
#9 28.25 �

#9 28.25 ,
#9 28.25 D�t0	
#9 28.25 p���y����y�g%8��y�8��y��3��L
#9 28.25 �

#9 28.25 ,
#9 28.25 D�t0	
#9 28.25 p 	H8	���
#9 28.25 ��8��y�8��y��!*�?X��y�X��y�\�c*��c*�(�)*� d*�xd*��d*�xe*��e*�`�*���*�0�*���*��Ȋ!*��d*�T��L
#9 28.25 �

#9 28.25 ,
#9 28.25 D�t0	
#9 28.25 p 	H8	���
#9 28.25 ��H0	
#9 28.25 �$��%�
�����	�� 
#9 28.25 �!
```

## Environment

Github worker using the latest Node 22.3.0 Docker image, but it also happens locally using the latest Node 22.3.0

### Additional context

My application code **did not** change in such a timeframe, nor its configs. I managed to work around the issue and correctly publish my repository's images by downgrading to Node 20. 

Since the stack trace refers to files inside /undici I figured issuing in this repository would have fit better rather than Node itself

### LINKED ISSUE #3337: fetch arrayBuffer() leaks random data from process memory
### Version

22.3.0

### Platform

Linux

### Subsystem

_No response_

Here is a test script that will reproduce the bug.
I've been unable to reproduce it with a local web server, even if I replicate the cloudinary response.

```js
async function test() {
  const url = process.argv[2];
  const f = await fetch(url);
  const body = await f.arrayBuffer();

  console.log('Content-Length:', f.headers.get('content-length'), 'byteLength:',  body.byteLength);

  if (body.byteLength !== 8192) {
    console.log('Bug not reproduced. Please try again.');
    return;
  }

  const decoder = new TextDecoder();
  const str = decoder.decode(body);

  console.log('Reproduced successfully');
  console.log('####################### Arraybuffer');
  console.log(str);
  console.log('#######################');
}

test();
```

`$ node test.js https://res.cloudinary.com/demo/image/upload/h_50,w_50/docs/camera-640.webp`

If the script says that it was not able to reproduce it then wait for a bit and try again. Sometimes it happens 10 times in a row and sometimes not for 30 times in a row.

_No response_

### What do you see instead?

As you can see below the buffer returned by arrayBuffer is fixed at 8192 bytes and the webp image received from the actual HTTP request will start at `RIFF`.
The data before the RIFF header is something random nearby in memory. I had this bug in a production environment and got server logs in there instead of the random HTTP text. In my node instance where I first detected this, I'm serving lots of request and the data leaked was different every time.
I also noticed that the HTTP request data might end up at a random offset, with more leaked data at the end of the blob.

```
Content-Length: 980, byteLength: 8192
Reproduced successfully
####################### Arraybuffer
/
HTTP/1.1 400 Bad Request
Connection: close

HTTP/1.1 408 Request Timeout
Connection: close

HTTP/1.1 431 Request Header Fields Too Large
Connection: close

HTTP/1.1 413 Payload Too Large
Connection: close

form-data; name="; filename----
�http/1.1RIFF�WEBPVP8 �P�*22>�:�G��!�.0�        d�\�d�3~;V��oޏ�p[\��e��O�Z�q�����{���L�K�
                                                                                         x\
                                                                                           �MdM�j�7�Y[2m�Y�>o6���䂫��/LE4�.�vGo�����S/L����(��� ��19��f����T�����!s�※�W4~M��e��VGb�[�H�n��`�]�V������E�އ����(Ba�f5ت��=�S�%c���j���e��ˎ��?�M���X
                                                                                ��<S�>�Ͳ!��W    P9�6����m�
-���    �~�I�����̈́*�XK��:U9&���O�����0�7Էc����"�+��z��                                                     R�~�H��0p�La��K��l��
���F?��籁���UV�<I~炷N
                     �[M)֚��N��TlK>�e������e��!�Y�Ĳ��f`5om$ݎ�
                                                            ���1)�
Xl�#�YAu�;��D����ə��a�s��!�o��x�Ċy�Z����_�5Q8��S�3V��<2�fVh�1�5��K�Y�H/
o�SG����c'M>&qb[�q��/�`�E1��J�۽/[J�w!�{���2�:;p"������J���2q���y��>���鬵������R/R�]��W�X��R�|�$����ɽ�d��v�`Qd*G\�NL��������u2e�
                                                                                                                               ���1{�7>c��~v���aV�;�    7��>�@����������P�9����F��TQ��08P��Y���6�B�Fw�H��wPu^�oy������I�5H�d0���=�0R���
�1p:���j��50J���GC=��'��Η?t෨ŋ}g�/�Vx/�,�G��*�U`��
#######################
```

### Additional information

Reproducible on Node 22.3 and NOT on 22.2.

KhafraDev:
fixes #3337 (confirmed with the repro op gave, couldn't repro locally)
might fix #3328 but not sure

Effectively reverts f979ec3204ca489abf30e7d20e9fee9ea7711d36

Adding this to the slight-spec-deviation-that-ends-up-fucking-up-things list

### DIFF
--- a/lib/web/fetch/body.js
+++ b/lib/web/fetch/body.js
@@ -309,7 +309,7 @@ function bodyMixinMethods (instance) {
         // Return a Blob whose contents are bytes and type attribute
         // is mimeType.
         return new Blob([bytes], { type: mimeType })
-      }, instance, false)
+      }, instance)
     },
 
     arrayBuffer () {
@@ -318,21 +318,20 @@ function bodyMixinMethods (instance) {
       // given a byte sequence bytes: return a new ArrayBuffer
       // whose contents are bytes.
       return consumeBody(this, (bytes) => {
-        // Note: arrayBuffer already cloned.
-        return bytes.buffer
-      }, instance, true)
+        return new Uint8Array(bytes).buffer
+      }, instance)
     },
 
     text () {
       // The text() method steps are to return the result of running
       // consume body with this and UTF-8 decode.
-      return consumeBody(this, utf8DecodeBytes, instance, false)
+      return consumeBody(this, utf8DecodeBytes, instance)
     },
 
     json () {
       // The json() method steps are to return the result of running
       // consume body with this and parse JSON from bytes.
-      return consumeBody(this, parseJSONFromBytes, instance, false)
+      return consumeBody(this, parseJSONFromBytes, instance)
     },
 
     formData () {
@@ -384,7 +383,7 @@ function bodyMixinMethods (instance) {
         throw new TypeError(
           'Content-Type was not one of "multipart/form-data" or "application/x-www-form-urlencoded".'
         )
-      }, instance, false)
+      }, instance)
     },
 
     bytes () {
@@ -393,7 +392,7 @@ function bodyMixinMethods (instance) {
       // result of creating a Uint8Array from bytes in this’s relevant realm.
       return consumeBody(this, (bytes) => {
         return new Uint8Array(bytes.buffer, 0, bytes.byteLength)
-      }, instance, true)
+      }, instance)
     }
   }
 
@@ -409,9 +408,8 @@ function mixinBody (prototype) {
  * @param {Response|Request} object
  * @param {(value: unknown) => unknown} convertBytesToJSValue
  * @param {Response|Request} instance
- * @param {boolean} [shouldClone]
  */
-async function consumeBody (object, convertBytesToJSValue, instance, shouldClone) {
+async function consumeBody (object, convertBytesToJSValue, instance) {
   webidl.brandCheck(object, instance)
 
   // 1. If object is unusable, then return a promise rejected
@@ -449,7 +447,7 @@ async function consumeBody (object, convertBytesToJSValue, instance, shouldClone
 
   // 6. Otherwise, fully read object’s body given successSteps,
   //    errorSteps, and object’s relevant global object.
-  await fullyReadBody(object[kState].body, successSteps, errorSteps, shouldClone)
+  await fullyReadBody(object[kState].body, successSteps, errorSteps)
 
   // 7. Return promise.
   return promise.promise
--- a/lib/web/fetch/body.js
+++ b/lib/web/fetch/body.js
@@ -391,7 +391,7 @@ function bodyMixinMethods (instance) {
       // with this and the following step given a byte sequence bytes: return the
       // result of creating a Uint8Array from bytes in this’s relevant realm.
       return consumeBody(this, (bytes) => {
-        return new Uint8Array(bytes.buffer, 0, bytes.byteLength)
+        return new Uint8Array(bytes)
       }, instance)
     }
   }
--- a/lib/web/fetch/util.js
+++ b/lib/web/fetch/util.js
@@ -1026,7 +1026,7 @@ function iteratorMixin (name, object, kInternalIterator, keyIndex = 0, valueInde
 /**
  * @see https://fetch.spec.whatwg.org/#body-fully-read
  */
-async function fullyReadBody (body, processBody, processBodyError, shouldClone) {
+async function fullyReadBody (body, processBody, processBodyError) {
   // 1. If taskDestination is null, then set taskDestination to
   //    the result of starting a new parallel queue.
 
@@ -1052,7 +1052,7 @@ async function fullyReadBody (body, processBody, processBodyError, shouldClone) 
 
   // 5. Read all bytes from reader, given successSteps and errorSteps.
   try {
-    successSteps(await readAllBytes(reader, shouldClone))
+    successSteps(await readAllBytes(reader))
   } catch (e) {
     errorSteps(e)
   }
@@ -1100,9 +1100,8 @@ function isomorphicEncode (input) {
  * @see https://streams.spec.whatwg.org/#readablestreamdefaultreader-read-all-bytes
  * @see https://streams.spec.whatwg.org/#read-loop
  * @param {ReadableStreamDefaultReader} reader
- * @param {boolean} [shouldClone]
  */
-async function readAllBytes (reader, shouldClone) {
+async function readAllBytes (reader) {
   const bytes = []
   let byteLength = 0
 
@@ -1111,13 +1110,6 @@ async function readAllBytes (reader, shouldClone) {
 
     if (done) {
       // 1. Call successSteps with bytes.
-      if (bytes.length === 1) {
-        const { buffer, byteOffset, byteLength } = bytes[0]
-        if (shouldClone === false) {
-          return Buffer.from(buffer, byteOffset, byteLength)
-        }
-        return Buffer.from(buffer.slice(byteOffset, byteOffset + byteLength), 0, byteLength)
-      }
       return Buffer.concat(bytes, byteLength)
     }
 


Uzlopak:
@KhafraDev 
Tests fail?

lib/web/fetch/body.js
```diff
@@ -318,21 +318,20 @@ function bodyMixinMethods (instance) {
       // given a byte sequence bytes: return a new ArrayBuffer
       // whose contents are bytes.
       return consumeBody(this, (bytes) => {
-        // Note: arrayBuffer already cloned.
-        return bytes.buffer
-      }, instance, true)
+        return new Uint8Array(bytes).buffer
```
tsctx, line 321:
```suggestion
        return bytes.buffer
```

tsctx, line 321: The change here seems to be the cause.

lib/web/fetch/body.js
```diff
@@ -393,7 +392,7 @@ function bodyMixinMethods (instance) {
       // result of creating a Uint8Array from bytes in this’s relevant realm.
       return consumeBody(this, (bytes) => {
         return new Uint8Array(bytes.buffer, 0, bytes.byteLength)
```
tsctx, line 394:
```suggestion
        return new Uint8Array(bytes)
```

lib/web/fetch/util.js
```diff
@@ -1111,13 +1110,6 @@ async function readAllBytes (reader, shouldClone) {
...
     if (done) {
       // 1. Call successSteps with bytes.
-      if (bytes.length === 1) {
-        const { buffer, byteOffset, byteLength } = bytes[0]
-        if (shouldClone === false) {
-          return Buffer.from(buffer, byteOffset, byteLength)
-        }
-        return Buffer.from(buffer.slice(byteOffset, byteOffset + byteLength), 0, byteLength)
-      }
       return Buffer.concat(bytes, byteLength)
```
tsctx, line 1113:
```suggestion
      if (bytes.length === 1) {
        const { buffer, byteOffset, byteLength } = bytes[0]
        return Buffer.from(buffer, byteOffset, byteLength)
      }
      return Buffer.concat(bytes, byteLength)
```

tsctx (approved):

KhafraDev:
We do need a test for this still, but considering the severity of this I think it was appropriate to land. Note that this is only a revert and now the reverted sections are verbatim with the spec.

From the linked issue: "Reproducible on Node 22.3 and NOT on 22.2."
- node v22.2 uses undici v16.3.0
- node v22.3 uses undici v16.8.0

I'm assuming that this is *not* an issue in the undici dispatcher api, otherwise it would have likely been noticed much earlier. The only change done with fetch's body between those two versions was the commit I linked (https://github.com/nodejs/undici/commit/f979ec3204ca489abf30e7d20e9fee9ea7711d36).

Furthermore, I will be blocking any change done to the specs we implement if they attempt to skip or modify the spec in some way. I think we should accept the limitations of these specs, instead of attempting to "node-ify" them.

KhafraDev:
I've created a [CSV](https://gist.github.com/KhafraDev/9d24c5f5f3f68edfcdd8bd54a2f9944c) of pretty much every bug fix in fetch over the last 4 months (as far back as github would take me, since I proved my point I didn't bother going back further).

Out of 13 bug fixes:
- 9 of them were caused by deviating from the spec.
- 2 of them are legitimate bugs that occurred despite us following the spec.
- 1 of them was effectively a typo.

There is a chance I missed some, but there are a disproportionate amount of bug reports and bug fixes revolving around deviations from the spec.

mcollina:
@KhafraDev I don't think this specific test was meant to break the spec, and it didn't, otherwise our WPTs would have caught it, right?

I think we can all agree on these 3 goals:
1. improve the performance of fetch()
2. don't break our users
3. break the spec only as a last resort

If the current set of tests cannot catch regressions, then the problem is in the test coverage.

KhafraDev:
I don't agree, fetch is a web spec and we should do our best to implement the spec. There are better suited apis for people who need better performance. Otherwise, I can't help but agree with what ljharb said initially about fetch (paraphrasing) "if it's not compliant with the spec, why not rename it to something else?".

KhafraDev:
I would like to mention that I wholeheartedly do agree w.r.t test coverage. Our current tests are not enough to safely deviate from the spec, so we should refrain, from all means possible, from doing so.

mcollina:
> I don't agree, fetch is a web spec and we should do our best to implement the spec.

I think we are in agreement on this. 

>  There are better suited apis for people who need better performance. Otherwise, I can't help but agree with what ljharb said initially about fetch (paraphrasing) "if it's not compliant with the spec, why not rename it to something else?".

We disagree here. `fetch()` is extremely popular, and we should improve its performance compared to the rest of the libraries. The two goals are not at odds, and I don't understand why you claim they are.

> I would like to mention that I wholeheartedly do agree w.r.t test coverage. Our current tests are not enough to safely deviate from the spec, so we should refrain, from all means possible, from doing so.

I think our tests are not enough to make many changes to `fetch()`, independently of the spec. The given PR was not "violating/breaking" the spec in any deliberate way.
It didn't break any wpt, and so far I haven't been able to reproduce the bug at all.

lib/web/fetch/util.js
ChALkeR, line 1113: This case wasn't cloned, as `Buffer.concat` returns a pooled buffer

KhafraDev:
> The given PR was not "violating/breaking" the spec in any deliberate way.

It wasn't a deliberate change, but it absolutely did violate the spec in any sense of the definition. If the spec tells us to clone the buffer, and we do not clone it, we have issues like this. The PR willingly skipped the clone, knowing full well that the spec tells us to clone the buffer.

KhafraDev:
> The two goals are not at odds, and I don't understand why you claim they are.

Recent changes have shown that following the spec - as stringently as we once did - comes second to performance. In another issue you said that support node.js use cases comes before following the spec - "IMHO the only deviation from the spec we should be doing is to support fetch() use cases that our users expect to be able to do in Node.js and they are blocked in Browsers". I am assuming that performance is one of these use cases (... and diagnostics channel support...).

> I think our tests are not enough to make many changes to fetch(), independently of the spec

I agree. The WPTs don't cover everything, not that I expect them to, but what they do cover is the text in the spec. 
For example, before submitting a PR to a WHATWG spec repo, they ask if you've 1) filed implementation issues, and 2) if you've opened a PR in the wpt repo that adds tests. That is why it is so important to follow the spec, instead of skipping steps for whatever reasons.

PR #3343, opened 2024-06-18
fetch: fix content-encoding order

tsctx:
When decoding, it must be processed from the reverse. We didn't catch this because the test is wrong.


mcollina (approved):
lgtm

metcoder95 (approved):

KhafraDev:
At some point c415fbbb59e2b898c5db6a681265cf3da865d02c got reverted... why?

KhafraDev (approved):

climba03003:
@KhafraDev see https://github.com/nodejs/undici/pull/2061 for the reason why it being reverted.

KhafraDev:
It seems like I mentioned this in https://github.com/nodejs/undici/pull/2061/files#r1305403864 and then never followed up on.

bobis19:
> lgtm

/

PR #3344, opened 2024-06-19
Fixed  ReadableStream reference error

tas33n:
Dunno the exact reason but it gives error ReadableStream reference error when using with @distube/ytdl-core

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready


mcollina:
Thanks for reporting. You are using an old version of node that do not have ReadableStream. Update your Node.js installation.

Issue #3353, opened 2024-06-21
Undici.Request and AbortController doesn't work well


fawazahmed0:
## Bug Description

Sometimes Aborting Undici.Request causes whole program to stop, even though everything is wrapped around try catch.

## Reproducible By

~~I don't have minimal reproducible example, as my code is larger in size.~~ But Undici.Fetch doesn't have this issue.

Here is minimal reproducible example:
```js
const undici = require('undici')

async function testUndici() {
    const controller = new AbortController()
    const {statusCode,headers,trailers,body} = await undici.request("https://github.com/fawazahmed0/tiger/releases/download/v2/pipetocsv.7z", { signal: controller.signal }).catch(console.error)
    setTimeout(()=>controller.abort(), 500)
}
testUndici()
```

## Expected Behavior

Undici.Request and AbortController should work

## Logs & Screenshots

```
node:events:498
      throw er; // Unhandled 'error' event
      ^
DOMException [AbortError]: This operation was aborted
    at new DOMException (node:internal/per_context/domexception:53:5)
    at AbortController.abort (node:internal/abort_controller:395:18)
    at Timeout._onTimeout (T:\xx\other scripts\undici-debug.js:7:31)
    at listOnTimeout (node:internal/timers:573:17)
    at process.processTimers (node:internal/timers:514:7)
Emitted 'error' event on BodyReadable instance at:
    at emitErrorNT (node:internal/streams/destroy:170:8)
    at emitErrorCloseNT (node:internal/streams/destroy:129:3)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21)

Node.js v22.3.0
```

## Environment

Ubuntu 24.04 & Windows 11
Node v22.3.0
Undici 6.19.1

Gigioliva:
I am having the same problem. The bug was introduced in `6.16.0`. I think the problem is in this #3209. In case of  [onComplete](https://github.com/nodejs/undici/blob/9134ee653af1ac933df975d76883fefd92bc23c2/lib/api/api-request.js#L156) we don't remove the listener from the signal

Gigioliva:
Adding this in `api-request.js` fixes the issue

<img width="385" alt="Screenshot 2024-06-21 alle 21 38 51" src="https://github.com/nodejs/undici/assets/26747867/270e4895-288a-4012-b2f9-6ad6f70242cf">

mcollina:
[...]

PR #3354, opened 2024-06-21
fix: post request signal

Gigioliva:
## Fix post request abort signal 

Fix issue #3353 

## Rationale

Request API doesn't remove the signal listener after complete the request

## Changes

Remove abort listener after success

## Status

- [x] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/test/request-signal.js
+++ b/test/request-signal.js
@@ -36,9 +36,11 @@ test('post abort signal', async (t) => {
 
   server.listen(0, async () => {
     const ac = new AbortController()
-    const ures = await request(`http://0.0.0.0:${server.address().port}`, { signal: ac.signal })
+    const uresPromise = request(`http://0.0.0.0:${server.address().port}`, { signal: ac.signal })
     ac.abort()
+
     try {
+      const ures = await uresPromise
       /* eslint-disable-next-line no-unused-vars */
       for await (const chunk of ures.body) {
         // Do nothing...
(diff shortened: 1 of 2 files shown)


test/request-signal.js
```diff
@@ -36,9 +36,11 @@ test('post abort signal', async (t) => {
 
   server.listen(0, async () => {
     const ac = new AbortController()
-    const ures = await request(`http://0.0.0.0:${server.address().port}`, { signal: ac.signal })
+    const uresPromise = request(`http://0.0.0.0:${server.address().port}`, { signal: ac.signal })
```
Gigioliva, line 39: ℹ️ this test was wrong because `request` api terminates when the stream has been completely consumed, so the connection has been closed. If we postpone the `await` we can throw the abort before dispatch starts consuming the stream, so the request is actually aborted

test/request-signal.js
```diff
@@ -74,3 +78,20 @@ test('post abort signal w/ reason', async (t) => {
...
+  const server = createServer((req, res) => {
+    res.end('asd')
+  })
+  after(() => server.close())
+
+  server.listen(0, async () => {
+    const ac = new AbortController()
+    const ures = await request(`http://0.0.0.0:${server.address().port}`, { signal: ac.signal })
+    ac.abort()
+    t.equal(await ures.body.text(), 'asd')
```
Gigioliva, line 94: ℹ️ in this case the abort signal is ignored since the request is completed

mcollina (approved):
lgtm

Uzlopak (approved):
makes kind of sense

RSLGTM

test/request-signal.js
KhafraDev: Since these are modeled after the fetch body mixins, I'm wondering if they should throw. In the fetch spec, if a request was aborted and you attempt to read the body, it'll throw an error. @ronag wdyt?

ronag (approved):

uriva:
Hey, how can I add this fix to my nodejs? (I'm running latest and it's not there iiuc)

PR #3357, opened 2024-06-22
websocket: don't use pooled buffer in mask pool

tsctx:
@mcollina

### DIFF
--- a/lib/web/websocket/frame.js
+++ b/lib/web/websocket/frame.js
@@ -27,7 +27,7 @@ try {
 function generateMask () {
   if (bufIdx === BUFFER_SIZE) {
     bufIdx = 0
-    crypto.randomFillSync((buffer ??= Buffer.allocUnsafe(BUFFER_SIZE)), 0, BUFFER_SIZE)
+    crypto.randomFillSync((buffer ??= Buffer.allocUnsafeSlow(BUFFER_SIZE)), 0, BUFFER_SIZE)
   }
   return [buffer[bufIdx++], buffer[bufIdx++], buffer[bufIdx++], buffer[bufIdx++]]
 }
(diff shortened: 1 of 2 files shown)


lib/web/websocket/frame.js
```diff
@@ -27,7 +27,7 @@ try {
 function generateMask () {
   if (bufIdx === BUFFER_SIZE) {
     bufIdx = 0
-    crypto.randomFillSync((buffer ??= Buffer.allocUnsafe(BUFFER_SIZE)), 0, BUFFER_SIZE)
+    crypto.randomFillSync((buffer ??= Buffer.allocUnsafeSlow(BUFFER_SIZE)), 0, BUFFER_SIZE)
```
tsctx, line 30: No problem, but with the current code, I have some concerns.

ronag (approved):

lib/web/websocket/frame.js
mcollina: I don't see them, unless we acess buffer.buffer anywhere

tsctx: If they access it, they can rewrite the mask, right?

tsctx: example:
```js
import { WebSocket } from './index.js'

Buffer.poolSize = 16386 * 4

// any server url
const u = new WebSocket('ws://localhost:3000')

u.onmessage = (data) => {
  // The mask can be rewritten here.
  new Uint8Array(Buffer.allocUnsafe(1).buffer).fill(0)
  u.send('Hi')
}
u.onopen = () => {
  u.send('Hi')
}

```

KhafraDev (approved):
Is it possible to add a test?

tsctx:
> Is it possible to add a test?

Unfortunately, I can write a test, but I can't catch this completely.

tsctx:
Unfortunately, I can write a test, but this can't catch using pooled buffer completely.

KhafraDev:
better something than nothing

tsctx:
test was placed incorrectly. this can catch it until now!

ronag (approved):

PR #3358, opened 2024-06-22
test: fix tests

ronag:
Reverts test changes from https://github.com/nodejs/undici/commit/a5eac88b6c9cb760b30367993f65f330782f2283

Now the tests fails as they should. A proper fix needs to be developed.


ronag:
@mcollina any opinion? revert or land failing tests and work on a proper fix.

mcollina:
revert and land a proper fix

PR #3360, opened 2024-06-22
fix: consider bytes read when dumping

ronag:
## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/api/readable.js
+++ b/lib/api/readable.js
@@ -14,6 +14,7 @@ const kBody = Symbol('kBody')
 const kAbort = Symbol('kAbort')
 const kContentType = Symbol('kContentType')
 const kContentLength = Symbol('kContentLength')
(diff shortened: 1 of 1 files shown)


lib/api/readable.js
```diff
@@ -165,7 +169,7 @@ class BodyReadable extends Readable {
     }
 
     return await new Promise((resolve, reject) => {
-      if (this[kContentLength] > limit) {
+      if (this[kContentLength] > limit || this[kBytesRead] > limit) {
```
metcoder95, line 172: Shall we consider this a breaking change?

metcoder95 (approved):

lib/api/readable.js
ronag: I don't think so.

PR #3365, opened 2024-06-24
fix: retry handler provide invalid body with fetch

### LINKED ISSUE #3356: The RetryHandler receives a duplicate body when the server does not support Range requests.
```
import { createServer } from "node:http"
import { once } from "node:events"
import { RetryAgent, Agent, fetch } from "undici";
import asserts from "node:assert";

let retry = true;
  res.writeHead(200, { "Content-Type": "text/plain" });
	if (retry) {
		retry = false;
		res.flushHeaders();
		res.write('H');
		setTimeout(() => {
			res.end('ello World');
		}, 1000);
	} else {
		res.end('Hello World');
	}
})

server.listen();

const agent = new RetryAgent(new Agent({bodyTimeout: 500}), {
	errorCodes: ['UND_ERR_BODY_TIMEOUT'],
});
const res = await fetch(`http://localhost:${port}`, {
    dispatcher: agent,
});
asserts.equal(await res.text(), 'Hello World');

server.close();
```

output:
```
AssertionError [ERR_ASSERTION]: 'HHello World' == 'Hello World'
```

climba03003:
## This relates to...

Fixes #3356 

## Rationale

`RetryHandler` should buffer the body instead of streaming to the next handler.
The reason behind is that the body maybe used by the next handler immediately and leads to stacking multiple invalid response.

## Changes

Buffer the body until the response is success or error.
It should result on all status code that is not `206` to retain the current behavior.

I though it may increase in memory usage for non-fetch people, because the payload can be streamed before but not now.
But in order to retain the correct behavior for retry, buffering should happen.

## Status

- [x] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/handler/retry-handler.js
+++ b/lib/handler/retry-handler.js
@@ -319,6 +319,11 @@ class RetryHandler {
 
   onError (err) {
     if (this.aborted || isDisturbed(this.opts.body)) {
+      // when every ended with error we should provide
+      // the non-complete body
+      for (const chunk of this.body) {
+        this.handler.onData(chunk)
+      }
       return this.handler.onError(err)
     }
 
@@ -344,6 +349,11 @@ class RetryHandler {
 
     function onRetry (err) {
       if (err != null || this.aborted || isDisturbed(this.opts.body)) {
+        // when every ended with error we should provide
+        // the non-complete body
+        for (const chunk of this.body) {
+          this.handler.onData(chunk)
+        }
         return this.handler.onError(err)
       }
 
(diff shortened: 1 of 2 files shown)


climba03003:
Test failure seems unrelated to the PR, it failed to download `Node.js`.

ronag (changes requested):
We should not be buffering in memory.

metcoder95 (commented):
Let's better throw, is safer and imposes less overhead, wdyt?

mcollina:
Yes

climba03003:
See #3389

PR #3368, opened 2024-06-25
fix: forward dispatch return value

ronag:
Refs: https://github.com/nodejs/undici/discussions/3373
Refs: https://github.com/nodejs/undici/issues/3370

### DIFF
--- a/lib/dispatcher/dispatcher.js
+++ b/lib/dispatcher/dispatcher.js
@@ -59,6 +59,7 @@ class ComposedDispatcher extends Dispatcher {
     this[kOnDisconnect] = (...args) => this.emit('disconnect', ...args)
     this[kOnConnectionError] = (...args) => this.emit('connectionError', ...args)
 
+    // TODO (fix): These must be weak references...
     this.#dispatcher
       .on('drain', this[kOnDrain])
       .on('connect', this[kOnConnect])
--- a/lib/dispatcher/dispatcher.js
+++ b/lib/dispatcher/dispatcher.js
@@ -39,18 +39,35 @@ class Dispatcher extends EventEmitter {
   }
 }
 
+const kOnDrain = Symbol('onDrain')
+const kOnConnect = Symbol('onConnect')
+const kOnDisconnect = Symbol('onDisconnect')
+const kOnConnectionError = Symbol('onConnectionError')
+
 class ComposedDispatcher extends Dispatcher {
-  #dispatcher = null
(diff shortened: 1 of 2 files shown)


ronag:
Things this doesn't solve.

- async dispatch 
- drain event parameters (which you could argue are useless)

metcoder95 (commented):
It has sense; I just feel that this can cause the false sensation of having a fully fledge `dispatcher`.

It might be possible that we end with max listeners problems.

Shall we tests the events are emitted as we expect?

ronag:
I feel this is just a hack until we have a better solution.

ronag:
We would need to have some form of weak event listeners for this to work well.

ronag:
@benjamingr Do you have any input on how we can have weak event listeners here that when the `ComposedDispatcher` is gc:d will be freed?

ronag:
This is still not great though... until GC the event listeners are still registered.

ronag:
The alternative is that we have some kind of ref/unref API. But that is going to be a constant pain of people doing it wrong.

ronag:
Also noticed that `Agent.dispatch` doesn't forward return value.

benjamingr:
> @benjamingr Do you have any input on how we can have weak event listeners here that when the ComposedDispatcher is gc:d will be freed?

We don't currently expose it but you can use kWeakHandler in the version bundled with Node, that lets you pass a listener and an object that the listener's lifetime is bound to - that's EventTarget though.

So we would need to build kWeakHandler (or similar) to EventEmitter and expose that to undici somehow - it has to be baked into the abstraction (EventEmitter) itself - otherwise it leaks the WeakRef itself (as you probably noticed).

ronag:
@metcoder95 @mcollina wdyt of returning a `Proxy`?

metcoder95:
>So we would need to build kWeakHandler (or similar) to EventEmitter and expose that to undici somehow

This sounds great, but personally I'm afraid we will be attaching `undici` to Node internals and userland usage might suffer from a workaround that does not behaves the same.
It would be nice to have something we can use in the form of a public API for that

>wdyt of returning a Proxy?

It simplifies a lot certainly, how does the gc behave (out of curiosity mostly)?

ronag:
> how does the gc behave (out of curiosity mostly)?

We cannot say anything about gc behavior other than it probably runs every now and then.

metcoder95 (approved):

PR #3117, opened 2024-04-14
perf: avoid unnecessary clone

tsctx:
Deps: #3109

Optimize cases where the body is a string, BufferSource, URLSearchParams, or stream (only one chunk).

Avoid cloning if unnecessary.

- #3109

```
benchmark                 time (avg)             (min … max)       p75       p99      p999
------------------------------------------------------------ -----------------------------
• small (length 512)
------------------------------------------------------------ -----------------------------
Response#arrayBuffer  22'113 ns/iter  (11'200 ns … 3'439 µs) 19'100 ns    116 µs    441 µs
Response#text         16'603 ns/iter    (11'700 ns … 673 µs) 15'100 ns 51'600 ns    268 µs

summary for small (length 512)
  Response#text
   1.33x faster than Response#arrayBuffer

• middle (length 8192)
------------------------------------------------------------ -----------------------------
Response#arrayBuffer  25'506 ns/iter  (12'500 ns … 3'582 µs) 21'600 ns    113 µs    337 µs
Response#text         27'878 ns/iter    (14'000 ns … 885 µs) 25'900 ns    115 µs    321 µs

summary for middle (length 8192)
  Response#arrayBuffer
   1.09x faster than Response#text

• long (length 131072)
------------------------------------------------------------ -----------------------------
Response#arrayBuffer     233 µs/iter  (40'300 ns … 3'191 µs)    218 µs  1'369 µs  2'948 µs
Response#text            374 µs/iter     (137 µs … 3'573 µs)    350 µs  2'034 µs  3'573 µs

summary for long (length 131072)
  Response#arrayBuffer
   1.61x faster than Response#text
```

- this patch

```
benchmark                 time (avg)             (min … max)       p75       p99      p999
------------------------------------------------------------ -----------------------------
• small (length 512)
------------------------------------------------------------ -----------------------------
Response#arrayBuffer  22'036 ns/iter  (11'300 ns … 2'884 µs) 21'100 ns    106 µs    453 µs
Response#text         13'975 ns/iter  (10'200 ns … 1'246 µs) 13'200 ns 37'900 ns    235 µs

summary for small (length 512)
  Response#text
   1.58x faster than Response#arrayBuffer

• middle (length 8192)
------------------------------------------------------------ -----------------------------
Response#arrayBuffer  25'928 ns/iter  (12'600 ns … 4'707 µs) 22'400 ns    111 µs    351 µs
[...]

PR #3377, opened 2024-06-26
fetch: don't use pooled buffer in body mixin

### LINKED ISSUE #3337: fetch arrayBuffer() leaks random data from process memory

tsctx:
completely fix #3337

### DIFF
--- a/lib/web/fetch/util.js
+++ b/lib/web/fetch/util.js
@@ -1099,6 +1099,41 @@ function isomorphicEncode (input) {
   return input
 }
 
+/** @type {Uint8ArrayConstructor} */
+const FastBuffer = Buffer[Symbol.species]
+
+/**
+ * Buffer is always cloned and does not use buffer pools.
+ * @param {Uint8Array[]} buffers
+ * @param {number} [length]
+ * @returns {Buffer}
+ */
+function safeBufferConcat (buffers, length) {
+  if (buffers.length === 0 || length === 0) {
+    return new FastBuffer(0)
+  }
+  if (buffers.length === 1) {
+    return new FastBuffer(buffers[0])
+  }
+  let offset = 0
+  if (length === undefined) {
+    length = 0
+    for (let i = 0; i < buffers.length; ++i) {
+      length += buffers[i].length
+    }
+  }
+  const output = Buffer.allocUnsafeSlow(length)
+  for (let i = 0; i < buffers.length; ++i) {
+    const buffer = buffers[i]
+    output.set(buffer, offset)
+    offset += buffer.length
+  }
+  if (offset < length) {
+    output.fill(0, offset, length)
+  }
+  return output
+}
+
 /**
  * @see https://streams.spec.whatwg.org/#readablestreamdefaultreader-read-all-bytes
  * @see https://streams.spec.whatwg.org/#read-loop
@@ -1113,7 +1148,7 @@ async function readAllBytes (reader) {
 
     if (done) {
       // 1. Call successSteps with bytes.
-      return Buffer.concat(bytes, byteLength)
+      return safeBufferConcat(bytes, byteLength)
     }
 
     // 1. If chunk is not a Uint8Array object, call failureSteps
--- a/test/fetch/body-mixin-no-pooled-buffer.js
+++ b/test/fetch/body-mixin-no-pooled-buffer.js
@@ -10,7 +10,7 @@ test('Do not use pooled buffer in body mixin', async () => {
   try {
     let counter = 0
     Buffer.allocUnsafe = function (...args) {
-      counter++a
+      counter++
       return allocUnsafe(...args)
     }
     // Do not use Buffer.allocUnsafe as it exposes the body to the pooled buffer.
--- a/test/fetch/body-mixin-no-pooled-buffer.js
+++ b/test/fetch/body-mixin-no-pooled-buffer.js
@@ -10,10 +10,13 @@ test('Do not use pooled buffer in body mixin', async () => {
   try {
     let counter = 0
     Buffer.allocUnsafe = function (...args) {
-      counter++
+      counter++a
       return allocUnsafe(...args)
     }
-    await new Response(new Uint8Array(1)).text()
+    // Do not use Buffer.allocUnsafe as it exposes the body to the pooled buffer.
+    await new Response("...any body").text()
+    // Body will be printed included.
+    // console.log(new TextDecoder().decode(Buffer.allocUnsafe(1).buffer.slice(304)))
     assert.strictEqual(counter, 0)
   } finally {
     Buffer.allocUnsafe = allocUnsafe
--- a/test/fetch/body-mixin-no-pooled-buffer.js
+++ b/test/fetch/body-mixin-no-pooled-buffer.js
@@ -1,0 +1,21 @@
+'use strict'
+
+const { Response } = require('../../')
+const assert = require('node:assert')
+const { test } = require('node:test')
+
+test('Don not use pooled buffer in body mixin', async () => {
+  const allocUnsafe = Buffer.allocUnsafe
+
+  try {
+    let counter = 0
+    Buffer.allocUnsafe = function (...args) {
+      counter++
+      return allocUnsafe(...args)
+    }
+    await new Response([new Uint8Array(1)]).text()
+    assert.strictEqual(counter, 0)
+  } finally {
+    Buffer.allocUnsafe = allocUnsafe
+  }
+})
--- a/test/fetch/body-mixin-no-pooled-buffer.js
+++ b/test/fetch/body-mixin-no-pooled-buffer.js
@@ -14,8 +14,7 @@ test('Do not use pooled buffer in body mixin', async () => {
       return allocUnsafe(...args)
     }
     // Do not use Buffer.allocUnsafe as it exposes the body to the pooled buffer.
-    await new Response('...any body')
-).text()
+    await new Response('...any body').text()
     // Body will be printed included.
     // console.log(new TextDecoder().decode(Buffer.allocUnsafe(1).buffer.slice(304)))
     assert.strictEqual(counter, 0)
--- a/test/fetch/body-mixin-no-pooled-buffer.js
+++ b/test/fetch/body-mixin-no-pooled-buffer.js
@@ -16,7 +16,7 @@ test('Do not use pooled buffer in body mixin', async () => {
     // Do not use Buffer.allocUnsafe as it exposes the body to the pooled buffer.
     await new Response('...any body').text()
     // Body will be printed included.
-    // console.log(new TextDecoder().decode(Buffer.allocUnsafe(1).buffer.slice(304)))
+    // console.log(new TextDecoder().decode(Buffer.allocUnsafe(1).buffer))
     assert.strictEqual(counter, 0)
   } finally {
     Buffer.allocUnsafe = allocUnsafe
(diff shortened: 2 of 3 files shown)


test/fetch/body-mixin-no-pooled-buffer.js
```diff
@@ -0,0 +1,21 @@
+'use strict'
+
+const { Response } = require('../../')
+const assert = require('node:assert')
+const { test } = require('node:test')
+
+test('Don not use pooled buffer in body mixin', async () => {
```
tsctx, line 7:
```suggestion
test('Do not use pooled buffer in body mixin', async () => {
```

ronag (approved):

tsctx:
@mcollina 
there are many such cases, should they be fixed? it seems that there are some things that should not be leaked...

tsctx:
@mcollina 
there are many such cases, should they be fixed?

test/fetch/body-mixin-no-pooled-buffer.js
```diff
@@ -0,0 +1,21 @@
...
+test('Do not use pooled buffer in body mixin', async () => {
+  const allocUnsafe = Buffer.allocUnsafe
+
+  try {
+    let counter = 0
+    Buffer.allocUnsafe = function (...args) {
+      counter++
+      return allocUnsafe(...args)
+    }
+    await new Response([new Uint8Array(1)]).text()
```
tsctx, line 16:
```suggestion
    await new Response(new Uint8Array(1)).text()
```

lib/web/fetch/util.js
```diff
@@ -1099,6 +1099,41 @@ function isomorphicEncode (input) {
...
+/** @type {Uint8ArrayConstructor} */
+const FastBuffer = Buffer[Symbol.species]
+
+/**
+ * Buffer is always cloned and does not use buffer pools.
+ * @param {Uint8Array[]} buffers
+ * @param {number} [length]
+ * @returns {Buffer}
+ */
+function safeBufferConcat (buffers, length) {
```
KhafraDev, line 1111: was this taken from node's implementation?

tsctx: I referred to it but it is different.

1. It doesn't check whether the Uint8Array (already checked).
2. It is safe without using pooled buffers and doesn't expose body.

KhafraDev: wdym by "doesn't expose body"?

Technically this doesn't do anything, the issue only occurs when we don't clone the body and then access .buffer. The spec, for the body mixins, tells us to clone the body. As long as we continue to follow the spec, this won't be an issue, and I'm wondering if taking a performance hit (?) is worth it?

tsctx: > wdym by "doesn't expose body"?

It means that the body is not leaked into the pooled buffer.

Revert #3117 and it seems to have been resolved. But, it can still be reproduced via `Buffer.allocUnsafe(1).buffer.slice(308)`.

Although I don't know, but if NodeJS's security team considers #3337 to be a security threat, it has not actually been resolved.

> I'm wondering if taking a performance hit (?) is worth it?

Basically, they are equivalent.
Slower than current implementation if multiple chunks are smaller than the pool size.

tsctx: If you are concerned about performance, you can do the same optimisation as #3117 again, as `safeBufferConcat` will always generate a new arraybuffer.

KhafraDev: I am concerned about trading performance for no benefit. We clone the buffer when the spec tells us to which makes this unnecessary. Am I missing something?

KhafraDev: My general sentiment is spec > performance, but we are already compliant with the spec, from what I see this PR is unnecessarily cloning the buffer that we already clone in the `consumeBody` callback.

tsctx: > My general sentiment is spec > performance

I agreed with you.

> We clone the buffer when the spec tells us to which makes this unnecessary. Am I missing something?

The only difference is that it does not use pooled buffers

It prevents the following:
```js
// node v22.2.0
// import { Response } from "./index.js";

await new Response(new TextEncoder().encode("private-key")).text();

// print private-key
console.log(new TextDecoder().decode(Buffer.allocUnsafe(1).buffer.slice(304)));
```

> I see this PR is unnecessarily cloning the buffer that we already clone in the consumeBody callback.

...I don't have time right now so I will answer later.

KhafraDev: Would you mind replacing the test you added with that? Much clearer on why we're avoiding allocUnsafe. I'm still not really convinced that it's worth fixing, since it's not a security related bug, but I don't mind this landing.

tsctx: > Would you mind replacing the test you added with that

The test catches this perfectly. What makes you think that way?

KhafraDev: Ensuring we don't use allocUnsafe is very different from *why* we don't use it.

tsctx: Thanks, I understood what you want to say.

test/fetch/body-mixin-no-pooled-buffer.js
```diff
@@ -0,0 +1,24 @@
...
+  const allocUnsafe = Buffer.allocUnsafe
+
+  try {
+    let counter = 0
+    Buffer.allocUnsafe = function (...args) {
+      counter++a
+      return allocUnsafe(...args)
+    }
+    // Do not use Buffer.allocUnsafe as it exposes the body to the pooled buffer.
+    await new Response("...any body").text()
```
tsctx, line 17:
```suggestion
    await new Response(new TextEncoder().encode('...any body')).text()
```

test/fetch/body-mixin-no-pooled-buffer.js
```diff
@@ -0,0 +1,24 @@
...
+const assert = require('node:assert')
+const { test } = require('node:test')
+
+test('Do not use pooled buffer in body mixin', async () => {
+  const allocUnsafe = Buffer.allocUnsafe
+
+  try {
+    let counter = 0
+    Buffer.allocUnsafe = function (...args) {
+      counter++a
```
tsctx, line 13:
```suggestion
      counter++
```

test/fetch/body-mixin-no-pooled-buffer.js
```diff
@@ -0,0 +1,24 @@
...
+      return allocUnsafe(...args)
+    }
+    // Do not use Buffer.allocUnsafe as it exposes the body to the pooled buffer.
+    await new Response(new TextEncoder().encode('...any body')).text()
```
tsctx, line 17:
```suggestion
    await new Response('...any body')
).text()
```

test/fetch/body-mixin-no-pooled-buffer.js
```diff
@@ -0,0 +1,25 @@
...
+
+  try {
+    let counter = 0
+    Buffer.allocUnsafe = function (...args) {
+      counter++
+      return allocUnsafe(...args)
+    }
+    // Do not use Buffer.allocUnsafe as it exposes the body to the pooled buffer.
+    await new Response('...any body')
+).text()
```
tsctx, line 18:
```suggestion
    await new Response('...any body').text()
```

test/fetch/body-mixin-no-pooled-buffer.js
```diff
@@ -0,0 +1,24 @@
...
+  try {
+    let counter = 0
+    Buffer.allocUnsafe = function (...args) {
+      counter++
+      return allocUnsafe(...args)
+    }
+    // Do not use Buffer.allocUnsafe as it exposes the body to the pooled buffer.
+    await new Response('...any body').text()
+    // Body will be printed included.
+    // console.log(new TextDecoder().decode(Buffer.allocUnsafe(1).buffer.slice(304)))
```
tsctx, line 19:
```suggestion
    // console.log(new TextDecoder().decode(Buffer.allocUnsafe(1).buffer))
```

KhafraDev:
I think there's still some questions left to answer here. Mostly: is this something we need? If not (and I am assuming we don't), what benefit is there in adding it?

tsctx:
> is this something we need? If not (and I am assuming we don't), what benefit is there in adding it?

I also have my doubts. but, double cloning can be avoided, as in the commit just added.

In fact, it is not skipping the spec, but creating a new array at `safeBufferConcat` according to this, which should not be a violation of the spec.

KhafraDev (changes requested):
I'm not interested in any solution that changes around the steps in the spec

tsctx:
Hmm... If I was to say it, It is completely fix #3337, but we may not need to worry about pooled buffer.

PR #3385, opened 2024-07-02
Add generic type for opaque object

### LINKED ISSUE #3378: Add generic type for opaque object for stream() and pipeline()
## This would solve...

Currently, when using `stream()` or `pipeline()`, `opaque` is typed as `unknown` in the callback, e.g:

```javascript
const bufs = [];

undici.stream("https://example.com", {
    method: 'GET',
    opaque: { bufs }
}, ({ statusCode, headers, opaque }) => {
    // TypeError here: Property 'bufs' does not exist on type 'unknown'.ts(2339)
    const { bufs } = opaque;
    return new Writable({
        write(chunk, encoding, callback) {
            bufs.push(chunk)
            callback()
        }
    })
});
```

Checking this code with TypeScript will generate an error: `Property 'bufs' does not exist on type 'unknown'.ts(2339)`

If `opaque` was defined as generic, TypeScript could infer the type from the second argument to `stream()` and provide the correct type inside the callback.

Add a generic type parameter (`<TOpaque = null>`) to the `stream()` and `pipeline()` type declarations as well as to `RequestOptions` and a few other interfaces. For example:

```TypeScript
declare function stream<TOpaque = null>(
  options: { dispatcher?: Dispatcher } & Omit<Dispatcher.RequestOptions<TOpaque>, 'origin' | 'path'>,
  factory: Dispatcher.StreamFactory<TOpaque>
): Promise<Dispatcher.StreamData>;
```

## I have also considered...

The alternative is to keep the current type declarations and require users to always cast `opaque` to a certain type before using it.

## Additional context

@Ethan-Arrowood , who originally added the type declarations, said [in his pull request](https://github.com/nodejs/undici/pull/375#issuecomment-722063227):

> For simplicity, I'm keeping this as unknown. Using a generic to pass through a type is no safer than type casting an unknown value.

Maybe I'm missing something here, but it seems to me that the generic approach is safer because it enforces that the `opaque` object that was passed in the request options has the same type as the one received in the callback. For example, the following would currently only cause a runtime error, but with the generic approach would give a type error:

```javascript
const bufs  = [];

undici.stream("https://example.com", {
    method: 'GET',
    // forgot to pass opaque
}, ({ statusCode, headers, opaque }) => {
    // error: opaque is null here
    const { bufs } = opaque;
    return new Writable({
        write(chunk, encoding, callback) {
            bufs.push(chunk)
            callback()
        }
    })
});
```

jfhr:
## This relates to...

This closes #3378 

## Changes

This adds a `TOpaque` generic type parameter to the type definitions for request(), connect(), stream(), and pipeline(). The type parameter defaults to null, which is the default value of the opaque property. If an opaque value is passed in the options, its type can usually be inferred automatically, such that no explicit type declaration is necessary. This commit also adds tsd tests to make sure the type definitions work as expected.

### Features

N/A

### Bug Fixes

N/A

No breaking change, unless I've missed something in the explanation below:

Previously, the type of `opaque` was `unknown`, which means it needed to be either type-checked or casted to another type before anything could be done with it. Such code should not be broken by this commit, although some type checks or assertions might become redundant. Code that disabled type checks (e.g. by casting to `any` or using `@ts-ignore` should be unaffected. Code that does not use typescript at all is also unaffected.

## Status

- [x] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/test/types/api.test-d.ts
+++ b/test/types/api.test-d.ts
@@ -1,5 +1,5 @@
 import { Duplex, Readable, Writable } from 'stream'
-import { expectAssignable } from 'tsd'
+import { expectAssignable, expectType } from 'tsd'
 import { Dispatcher, request, stream, pipeline, connect, upgrade } from '../..'
 
 // request
@@ -10,12 +10,22 @@ expectAssignable<Promise<Dispatcher.ResponseData>>(request('', { method: 'GET', 
 // stream
 expectAssignable<Promise<Dispatcher.StreamData>>(stream('', { method: 'GET' }, data => {
   expectAssignable<Dispatcher.StreamFactoryData>(data)
+  expectType<null>(data.opaque)
+  return new Writable()
+}))
+expectAssignable<Promise<Dispatcher.StreamData<{ example: string }>>>(stream('', { method: 'GET', opaque: { example: '' } }, data => {
+  expectType<{ example: string }>(data.opaque)
   return new Writable()
 }))
 
 // pipeline
 expectAssignable<Duplex>(pipeline('', { method: 'GET' }, data => {
   expectAssignable<Dispatcher.PipelineHandlerData>(data)
+  expectType<null>(data.opaque)
+  return new Readable()
+}))
+expectAssignable<Duplex>(pipeline('', { method: 'GET', opaque: { example: '' } }, data => {
+  expectType<{ example: string }>(data.opaque)
   return new Readable()
 }))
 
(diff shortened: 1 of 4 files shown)


ronag (commented):
RSLGTM

tsctx (approved):

mcollina (approved):
lgtm

PR #3389, opened 2024-07-03
fix: throw on retry when payload is consume by downstream

### LINKED ISSUE #3356: The RetryHandler receives a duplicate body when the server does not support Range requests.

climba03003:
Close #3365 Fixes #3356

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/handler/retry-handler.js
+++ b/lib/handler/retry-handler.js
@@ -68,7 +68,6 @@ class RetryHandler {
     this.end = null
     this.etag = null
     this.resume = null
-    this.consumed = false
 
     // Handle possible onConnect duplication
     this.handler.onConnect(reason => {
@@ -194,12 +193,13 @@ class RetryHandler {
       this.resume = null
 
       if (statusCode !== 206) {
-        // Abort when status code is not partial content
-        // and payload is already consumed because downstream
-        // will concatenate response from two request wrongly.
-        if (this.consumed) {
+        // Only Partial Content 206 supposed to provide Content-Range,
+        // any other status code that partially consumed the payload
+        // should not be retry because it would result in downstream
+        // wrongly concatanete multiple responses.
+        if (this.start > 0) {
           this.abort(
-            new RequestRetryError('unsupported retry when payload is consumed', statusCode, {
+            new RequestRetryError('server does not support the range header and the payload was partially consumed', statusCode, {
               headers,
               data: { count: this.retryCount }
             })
@@ -308,7 +308,6 @@ class RetryHandler {
   }
 
   onData (chunk) {
-    this.consumed = true
     this.start += chunk.length
 
     return this.handler.onData(chunk)
--- a/lib/handler/retry-handler.js
+++ b/lib/handler/retry-handler.js
@@ -68,6 +68,7 @@ class RetryHandler {
     this.end = null
     this.etag = null
     this.resume = null
+    this.consumed = false
 
     // Handle possible onConnect duplication
     this.handler.onConnect(reason => {
@@ -193,6 +194,19 @@ class RetryHandler {
       this.resume = null
 
       if (statusCode !== 206) {
+        // Abort when status code is not partial content
+        // and payload is already consumed because downstream
+        // will concatenate response from two request wrongly.
+        if (this.consumed) {
+          this.abort(
+            new RequestRetryError('unsupported retry when payload is consumed', statusCode, {
+              headers,
+              data: { count: this.retryCount }
+            })
+          )
+          return false
+        }
+
         return true
       }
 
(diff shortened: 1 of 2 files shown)


lib/handler/retry-handler.js
```diff
@@ -296,7 +310,10 @@ class RetryHandler {
   onData (chunk) {
     this.start += chunk.length
 
-    return this.handler.onData(chunk)
+    const written = this.handler.onData(chunk)
```
climba03003, line 313: I use the return of `onData` to indicate the data is written to downstream.
When it return `false` from the first write, we still have chance to retry.

ronag: You can't do that. It only indicated back pressure. Not whether the data was sent or not.

climba03003: Doesn't `false` means the data didn't push to the buffer?
If yes, then it is what we need for this case.

climba03003: Nevermind, I found the answer. It is no way back.

lib/handler/retry-handler.js
```diff
@@ -193,6 +194,19 @@ class RetryHandler {
       this.resume = null
 
       if (statusCode !== 206) {
+        // Abort when status code is not partial content
+        // and payload is already consumed because downstream
+        // will concatenate response from two request wrongly.
```
mcollina, line 199: Can you please expand the comment, explaining that anything that is not a 206 implies that a range request cannot be made?

lib/handler/retry-handler.js
```diff
@@ -294,6 +308,7 @@ class RetryHandler {
   }
 
   onData (chunk) {
+    this.consumed = true
```
mcollina, line 311: I don't think we need a new prop, we could just use .start

lib/handler/retry-handler.js
```diff
@@ -193,6 +194,19 @@ class RetryHandler {
       this.resume = null
 
       if (statusCode !== 206) {
+        // Abort when status code is not partial content
+        // and payload is already consumed because downstream
+        // will concatenate response from two request wrongly.
+        if (this.consumed) {
+          this.abort(
+            new RequestRetryError('unsupported retry when payload is consumed', statusCode, {
```
mcollina, line 202: The error message is not very actionable. I would rephrase to something like
```suggestion
            new RequestRetryError('server does not support the range header and the payload was partially consumed', statusCode, {
```

lib/handler/retry-handler.js
```diff
@@ -193,6 +194,19 @@ class RetryHandler {
       this.resume = null
 
       if (statusCode !== 206) {
```
ronag, line 196:
```suggestion
      if (statusCode !== 206 && (this.start > 0 || statusCode !== 200) {
```

lib/handler/retry-handler.js
```diff
@@ -192,8 +192,18 @@ class RetryHandler {
...
+      // should not be retry because it would result in downstream
+      // wrongly concatanete multiple responses.
+      if (statusCode !== 206 && (this.start > 0 || statusCode !== 200)) {
+        this.abort(
+          new RequestRetryError('server does not support the range header and the payload was partially consumed', statusCode, {
+            headers,
+            data: { count: this.retryCount }
+          })
+        )
+        return false
```
ronag, line 206:
```suggestion
        throw new RequestRetryError('server does not support the range header and the payload was partially consumed', statusCode, {
            headers,
            data: { count: this.retryCount }
          })
        )
```

ronag (approved):

lib/handler/retry-handler.js
climba03003: Are you sure it should throw instead of similar to the others error?

ronag: I think all errors should throw

ronag: it's a nit, you can leave it if you want

mcollina (approved):
lgtm

PR #3391, opened 2024-07-03
feat: implement `BodyReadable.bytes`

tsctx:
Refs: #3262


ronag (approved):

mcollina (changes requested):
Can you add it to the docs?

KhafraDev (approved):
as a side note we should probably add a test similar to 9134ee653af1ac933df975d76883fefd92bc23c2 to be safe

tsctx:
@mcollina ptal

mcollina (approved):
lgtm

PR #3407, opened 2024-07-17
fix: don't call onConnect automatically

ronag:
In some edge cases this causes issues. Leave it to the user to handle if necessary.

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/handler/decorator-handler.js
+++ b/lib/handler/decorator-handler.js
@@ -1,11 +1,9 @@
 'use strict'
 
 const assert = require('node:assert')
-const noop = () => {}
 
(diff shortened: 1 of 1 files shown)


metcoder95 (commented):
Like which scenarios?

ronag:
I actually don't quite remember. I think it was related to an error case in redirect handler.

metcoder95 (approved):
LGTM

Let me know if you can remember the scenario, so we can add a test for it later on

PR #3408, opened 2024-07-17
fix: always parse headers

ronag:
Changes onHeaders and onComplete headers/trailers to always be a pre-parsed headers record.

### DIFF
--- a/lib/core/util.js
+++ b/lib/core/util.js
@@ -342,13 +342,13 @@ function parseHeaders (headers, obj) {
         val = [val]
         obj[key] = val
       }
-      val.push(headers[i + 1].toString('utf8'))
+      val.push(headers[i + 1].toString('latin1'))
     } else {
       const headersValue = headers[i + 1]
       if (typeof headersValue === 'string') {
         obj[key] = headersValue
       } else {
-        obj[key] = Array.isArray(headersValue) ? headersValue.map(x => x.toString('utf8')) : headersValue.toString('utf8')
+        obj[key] = Array.isArray(headersValue) ? headersValue.map(x => x.toString('latin1')) : headersValue.toString('latin1')
       }
     }
   }
--- a/lib/core/util.js
+++ b/lib/core/util.js
@@ -332,6 +332,7 @@ function bufferToLowerCasedHeaderName (value) {
  */
 function parseHeaders (headers, obj) {
   if (obj === undefined) obj = {}
+  if (headers == null) return
   for (let i = 0; i < headers.length; i += 2) {
     const key = headerNameToString(headers[i])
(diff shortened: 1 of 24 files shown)


ronag:
@metcoder95 @mcollina I have some http2 and mock test failures I haven't been able to figure out. Would appreciate some help.

metcoder95 (commented):
Shall we update the docs to state the change?

LGTM, I'm having a hectic morning; I'll see the H2 issues throughout the day and see if I can submit a PR or hint you the issue.

ronag:
> Shall we update the docs to state the change?

Where how? I tried updating docs. Did I miss something?

metcoder95:
This [dispatcher handlers](https://undici.nodejs.org/#/docs/api/Dispatcher?id=parameter-dispatchhandler) will start receiving the `headers` already parsed instead of in a Buffer format, isn't it?

metcoder95 (approved):

metcoder95:
Will check the H2 issues throughout the day

ronag:
I think I managed to fix the failures

mertcanaltin:
> I think I managed to fix the failures

greetings some test errors appear I wonder if these could be flakky?

metcoder95:
I don't believe they are flaky as they are pretty consistent across jobs: https://github.com/nodejs/undici/actions/runs/9980661941/job/27582582584?pr=3408#step:8:2242

Seems to be related with `fetch` and parsing non-Latin1 chars on headers.

lib/core/util.js
```diff
@@ -341,13 +342,13 @@ function parseHeaders (headers, obj) {
...
       }
-      val.push(headers[i + 1].toString('utf8'))
+      val.push(headers[i + 1].toString('latin1'))
     } else {
       const headersValue = headers[i + 1]
       if (typeof headersValue === 'string') {
         obj[key] = headersValue
       } else {
-        obj[key] = Array.isArray(headersValue) ? headersValue.map(x => x.toString('utf8')) : headersValue.toString('utf8')
+        obj[key] = Array.isArray(headersValue) ? headersValue.map(x => x.toString('latin1')) : headersValue.toString('latin1')
```
ronag, line 351: @metcoder95 @mcollina PTAL

metcoder95: Adjusted the tests to align with the change

metcoder95 (approved):

metcoder95:
It seems CI is still not happy...

mcollina (changes requested):
We should update the global dispatcher API version number.

Ideally it would be best if we could provide a polyfill for the v1 API, so that using latest undici on old node would still work.

PR #3418, opened 2024-07-24
fix: Improve timeout handling during event loop lag

ronag:
Make timeouts behave more closer that what is expected during event loop lag.

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready


lib/dispatcher/client-h1.js
```diff
@@ -182,10 +185,17 @@ class Parser {
...
         this.timeout.refresh()
+        this.timeoutDeadline = this.timeoutValue ? Date.now() + this.timeoutValue : 0
       }
     }
   }
 
+  updateTimeout () {
+    if (this.timeoutDeadline && this.timeoutDeadline < Date.now()) {
+      onParserTimeout(this)
+    }
```
ronag, line 196: This will call `Date.now()` quite often... which is not optimal performance-wise.

mcollina:
Have you got a reproduction of this? I'd like to tinker with this problem too.

ronag:
Nope. Had an 8 hour drive last night and was thinking about it. We've had a few possibly related issues here and there.

mcollina:
Yes, I'm aware, I just don't know what's the best way to solve this.

ronag:
> Yes, I'm aware, I just don't know what's the best way to solve this.

What are your thoughts on my proposal in this PR?

ronag:
I think we have 2 different problems:

- For the body and headers timeout: event loop lag can cause a false positive by including the lag in the timeout hence making the elapsed duration too long.
- For the idle timeout: event loop lag can cause a false negative by delaying the timeout detection hance making the elapsed duration too short.

i.e. basically two problems which an inverse of eachother.

metcoder95:
Just out of curiosity, even with this, it might still fall under the problem of high event loop lag, isn't it? Or I get it wrong?

ronag:
>  it might still fall under the problem of high event loop lag, isn't it? Or I get it wrong?

Not that I can think of?

QuiiBz:
👋 Hi! Is there any way we can help/contribute to push this PR forward? We (Vercel) are getting [more and more reports](https://github.com/vercel/vercel/issues/11692), which we believe are related to this specific issue.

I'm personally not super familiar with Undici's codebase, but please let me know if there is any way we can help.

ronag:
Not high on my list. If you want more urgent help feel free to mail me.

metcoder95:
@ronag are we moving on this or do you need support on this front?

ronag:
I don't have time to work on it ATM. Feel free to take over.

metcoder95:
In the issue @Uzlopak said he's working on it. If you need support or something, let me know, I can work on it later this week

PR #3423, opened 2024-07-26
Add "decompress" response utility

### LINKED ISSUE #3412: Expose "Content-Encoding" handling publicly
## This would solve...

Hi! In [Interceptors](https://github.com/mswjs/interceptors), we are looking at improving the compatibility of response handling for mocked responses. As a part of that effort, we'd like to produce an identical behavior that Undici exhibits with compressed response bodies. 

If the following handling logic is a tad abstracted and exposed as a standalone function, it could help us a lot to stay consistent and make countless developers' tests more reliable:

https://github.com/nodejs/undici/blob/c1f7d2a9a6556e5a34fdf5ed38c77ce636f1ff02/lib/web/fetch/index.js#L2139-L2161

## I have also considered...

We are considering writing the `Content-Encoding` handling by hand but I'd like to join efforts on this one. I don't wish to repeat things. I like using the platform. I like Undici. 

## Additional context

I'm interested in opening a pull request and seeing this through, given it's something you agree on. Let me know.

kettanaito:
## This relates to...

- Fixes #3412

## Rationale

Exposing the response decompression logic (i.e. the handling of the `Content-Encoding` header) will allow other Node.js libraries to reuse it, resulting in a consistent experience for everyone. 

> See more detailed reasoning in the referenced issue. 

## Changes

- Adds a new `decompress` utility, abstracting the `Content-Type` handling from the existing `lib/fetch/index.js`.
- [ ] Uses the `decompress` utility in the `onHeaders` callback.

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/web/fetch/decompress.js
+++ b/lib/web/fetch/decompress.js
@@ -1,0 +1,80 @@
+const { pipeline } = require('node:stream')
+const zlib = require('node:zlib')
+const { redirectStatusSet, nullBodyStatus } = require('./constants')
+const { createInflate } = require('./util')
+
+/**
+ * Decompress the given response body.
+ * @param {Request} request
+ * @param {Response} response
+ * @returns {ReadableStream<Uint8Array> | null}
+ */
+function decompress (request, response) {
+  const contentEncoding = response.headers.get('content-encoding')
+
+  if (!contentEncoding) {
+    return response.body
+  }
+
+  // https://www.rfc-editor.org/rfc/rfc7231#section-3.1.2.1
+  // "All content-coding values are case-insensitive..."
+  const codings = contentEncoding
+    .toLowerCase()
+    .split(',')
+    .map((coding) => coding.trim())
+
+  if (codings.length === 0) {
+    return response.body
+  }
+
+  const willFollow =
+    response.headers.get('location') &&
+    request.redirect === 'follow' &&
+    redirectStatusSet.has(response.status)
+
+  if (
+    request.method === 'HEAD' ||
+    request.method === 'CONNECT' ||
+    nullBodyStatus.includes(response.status) ||
+    willFollow
+  ) {
+    return response.body
+  }
+
+  const decoders = []
+
+  for (let i = codings.length - 1; i >= 0; --i) {
+    const coding = codings[i]
+
+    // https://www.rfc-editor.org/rfc/rfc9112.html#section-7.2
+    if (coding === 'x-gzip' || coding === 'gzip') {
+      decoders.push(
+        zlib.createGunzip({
+          // Be less strict when decoding compressed responses, since sometimes
+          // servers send slightly invalid responses that are still accepted
+          // by common browsers.
+          // Always using Z_SYNC_FLUSH is what cURL does.
+          flush: zlib.constants.Z_SYNC_FLUSH,
+          finishFlush: zlib.constants.Z_SYNC_FLUSH
+        })
+      )
+    } else if (coding === 'deflate') {
+      decoders.push(createInflate())
+    } else if (coding === 'br') {
+      decoders.push(zlib.createBrotliDecompress())
+    } else {
+      decoders.length = 0
+      break
+    }
+  }
+
+  if (decoders.length === 0) {
+    return response.body
+  }
+
+  return pipeline(response.body, ...decoders)
+}
+
+module.exports = {
+  decompress
+}


KhafraDev:
it'd be nice if this worked with the rest of undici too, not just fetch

kettanaito:
@KhafraDev, can you point me to other usages where this should work?

KhafraDev:
> can you point me to other usages where this should work?

undici.request, stream, etc. https://github.com/nodejs/undici/discussions/1155

kettanaito:
Added a basic set of unit tests for the `decompress` function. 

```sh
✔ ignores responses without the "Content-Encoding" header (32.383292ms)
✔ ignores responses with empty "Content-Encoding" header (0.780042ms)
✔ ignores redirect responses (0.439708ms)
✔ ignores HEAD requests (0.456958ms)
﹣ ignores CONNECT requests (0.0775ms) # SKIP
✔ ignores responses with unsupported encoding (0.219292ms)
✔ decompresses responses with "gzip" encoding (15.465084ms)
```

Still need to cover multiple `Content-Encoding` values, like `gzip, br`, etc.

kettanaito:
@KhafraDev, thinking of tailoring this to a generic body decompression purposes, I can imagine it being used this way: 

```ts
function createDecompressionStream(args): TransformStream 

// Then, usage:
anyCompressedBodyStream.pipe(createDecompressionStream(args))
```

The most challenging part is figuring out the `args` here.

kettanaito:
Any thoughts about this?

```js
/**
 * Create a decompression stream for the given `Content-Encoding` codings.
 * @param {string[]} codings
 * @returns {WritableStream}
 */
function createDecompressionStream(codings) {
  const decoders = [];

  for (let i = codings.length - 1; i >= 0; --i) {
    const coding = codings[i];

    // https://www.rfc-editor.org/rfc/rfc9112.html#section-7.2
    if (coding === "x-gzip" || coding === "gzip") {
      decoders.push(
        zlib.createGunzip({
          // Be less strict when decoding compressed responses, since sometimes
          // servers send slightly invalid responses that are still accepted
          // by common browsers.
          // Always using Z_SYNC_FLUSH is what cURL does.
          flush: zlib.constants.Z_SYNC_FLUSH,
          finishFlush: zlib.constants.Z_SYNC_FLUSH,
        })
      );
    } else if (coding === "deflate") {
      decoders.push(createInflate());
    } else if (coding === "br") {
      decoders.push(zlib.createBrotliDecompress());
    } else {
      decoders.length = 0;
      break;
    }
  }

  return pipeline(...decoders, () => {});
}
```

KhafraDev:
I think for arguments a stream and a header value would be fine

kettanaito:
Got it. Added a [`decompressStream` utility](https://github.com/nodejs/undici/pull/3423/files#diff-7a181e92a0fcee24b8bf2c92d7161131e08db363521b3d356d76a1d3ea9f55f5R42), reused it in the `decompress()`.

lib/web/fetch/decompress.js
```diff
@@ -0,0 +1,95 @@
...
+    response.headers.get('content-encoding')
+  )
+}
+
+/**
+ * Decompress the given stream based on the "Content-Encoding" response header.
+ * @param {ReadableStream<Uint8Array>} stream
+ * @param {string | null} contentEncoding The value of the "Content-Encoding" response header.
+ */
+function decompressStream (stream, contentEncoding) {
```
metcoder95, line 42: Wondering if it might be useful to allow customize the options for each one of the decompressions implementation.

metcoder95 (commented):
So far, it LGTM;

I believe the `decompressStream` can be used within an interceptor if we want to aim for that in the other PR as well.

lib/web/fetch/decompress.js
```diff
@@ -0,0 +1,95 @@
...
+          finishFlush: zlib.constants.Z_SYNC_FLUSH
+        })
+      )
+    } else if (coding === 'deflate') {
+      decoders.push(createInflate())
+    } else if (coding === 'br') {
+      decoders.push(zlib.createBrotliDecompress())
+    } else {
+      decoders.length = 0
+      break
```
ronag, line 81:
```suggestion
      return null
```

Maybe return null or otherwise there is no way to know if it "failed".

ronag: Also `decompressStream` should probably not live under `/web/`

lib/web/fetch/decompress.js
kettanaito: @metcoder95, do we have a use case for that? I'd love to see it, think how we can support additional parameters. Perhaps the way forward would be through more granular utilities, if any, then making one function with complex input.

lib/web/fetch/decompress.js
kettanaito: @ronag, can it fail though? The intention was that, if there's no known compression provided, the stream is returned as-is. 

You can still handle failures as you normally would:

```js
decompressStream(myStream).on('error', callback)
```

KhafraDev: > if there's no known compression provided

The logic seems to be that if an unknown coding is sent, the body isn't decompressed at all. I'm not sure if that sounds right either.

`Content-Encoding: gzip, zstd, br` -> ignored (zstd is not supported)
`Content-Encoding: gzip, br` -> decompressed

kettanaito: As a consumer, I expect the body stream to be returned as-is if:

1. It has no `Content-Encoding` set (no `codings` provided as an argument);
2. It has unknown `Content-Encoding` set or an unknown `coding(s)` is provided as an argument. 

I wouldn't expect it to throw because that's a disruptive behavior that implies I need to add a guard _before_ trying to call `decompress`/`decompressStream`. At least, not at a level this low. 

If the API is potentially disruptive, I need to have an extra check in place to prevent scenarios where decompression is impossible. This puts extra work on me as a consumer but also results in a more frail API since the list of supported encodings is baked in Undici and may change across releases. 

I strongly suggest to return the body stream as-is in those two cases I described above. `decompressString()` must never error, save for the streams errors itself (and those are irrelevant in the context of this function).

kettanaito: We have to outline whether we are after a runtime behavior or a DX improvement. I shared my opinion on this in the context of runtime behaviors. 

If we are talking about a DX improvements, as in not silently ignore unknown compression algorithms, allowing bugs to slip in, then printing a warning may be enough. 

Alternatively, `decompressString()` may trigger error on the given stream if an unknown encoding is encountered. I still think that's too disruptive. Would love to have a more detailed, use case-driven discussion.

lib/web/fetch/decompress.js
metcoder95: Hmm, that's a good point; let's leave it for later then. I don't have a specific use-case in mind but just if you want to add extra params to e.g. `brotli` decompress, or strategy to `gunzip`.

But unless strong request, we can keep it simple as you suggest 👍

lib/web/fetch/decompress.js
metcoder95: What about making the behaviour optional?

I see the benefit of throwing an error if the decompression is not possible, at the same time to return the decompressed stream and move forward.
Maybe using a flag whereas the default behaviour is to not throw, and throw otherwise (or the other way around) can provide the enough ergonomics to enable these use cases?

kettanaito: Would erroring the _returned_ stream be a semantically correct thing to do instead of throwing an error? 

That way, the `decompressStream` function doesn't have to become more complex. It can always error the returned stream if the decompression is impossible. The consumer can then decide how to handle those errors in the `error` listener.

I emphasize the returned stream because the input stream must remain intact so it can be operated with by other agent regardless of the decompression results. I also propose this on the assumption that `pipeline()` returns a _new_ stream, which I believe it does. We can error _that_.

```js
function decompressStream(stream, codings) {
  // ...
  const result = pipeline(...)

  if (noDecompressionPossible) {
    result.destroy(new Error('Unknown compression algorithm')
  }

  return result
}
```

One problem with this approach is that it doesn't fail fast enough. We know if one of the `codings` is not supported when iterating over them, even before the pipeline is created. 

I see two options to solve this:

1. Error the pipeline but it's not fail-fast.
2. Accept additional argument but it makes the function more complex:

```js
function decompressStream(stream, codings, throwOnUnknownCoding) {}
```

kettanaito: I can also argue that the function can fail _always_, and the consumer can always make it safe with this:

```js
import { decompress } from 'undici'

function decompressOrReturnAsIs(request, response) {
  try {
    let r = decompress(request, response)
    return r
  } catch { 
    return response.body
  }
}
```

If you think throwing will benefit most consumers, I'm not against it.

kettanaito: @metcoder95 @KhafraDev do you agree with my reasoning in the post above?

kettanaito: Hah, so there's an existing expectation that decompression is _skipped_ if encountered unknown encoding:

```
 ✖ should skip decompression if unsupported codings
```

I suppose that answers the question.

kettanaito:
I've added unit tests for `x-gzip` and `gzip, br` encodings. Both seem to pass 🎉

lib/web/fetch/decompress.js
KhafraDev: throwing an error will make fetch slower

kettanaito:
The CI is failing with some issues download Node.js binary, it seems. Doesn't appear to be related to the changes.

lib/web/fetch/decompress.js
kettanaito: Regarding the decoders customization discussion, I don't think it justifies the cost of making `decompressStream` more complex. The only thing you need is the list of decodings, which you can get using a one-liner:

```js
 const codings = contentEncoding
    .toLowerCase()
    .split(',')
    .map((coding) => coding.trim())
```

The purpose of `decompressStream` is to grab the _exact_ behavior Undici has under the hood. If you need a different behavior, you should (1) parse the `Content-Encoding` header; (2) map encodings to decompression streams with custom options. 

I can see how you'd want to _extend_ or override certain things from the default decompression, and for that we, perhaps, can consider exporting those options separately? 

```js
module.exports = {
  gzipDecompressionOptions: {
    flush: zlib.constants.Z_SYNC_FLUSH,
    finishFlush: zlib.constants.Z_SYNC_FLUSH
  }
}
```

metcoder95: Regarding erroring or throwing; As you suggest @kettanaito, I'd like to fail faster and inform the implementer the encoding is not supported without even disturbing the input stream, so fallbacks can be applied if they want to.

For `fetch`, as @KhafraDev points out, the `throw`ing might be slower, so we can disable by default, and not throw at all, but rather ignore the non-supported encodings.

The issue I'd like to avoid, is putting the implementer in the spot of all or nothing.

>Even then, that's all the options Undici uses right now. To me this looks like a nice thing to have without substantial use case argumentation. I lean toward iterating on this when we have more of the latter.

Agree, let's do that 👍

kettanaito:
## Update

I've added remaining unit tests for the `decompressStream` utility, all are passing locally. 

Based on the [conclusion](https://github.com/nodejs/undici/pull/3423#discussion_r1699608433) of the discussion around unknown encodings, and taking into the consideration that there's an explicit existing test suite that unknown encodings must be ignored when decompressing the `fetch()` response body, looks like the way forward is to leave the implementation as-is. Please let me know if I misunderstood the conclusion. 

I suggest adding the throwing logic to `decompressStream` once we discover a use case for that (decompression in the interceptor may be that use case). From then, we can open a new pull request to extend the `decompressStream` utility with `options` or anything else we find proper.

lib/web/fetch/decompress.js
```diff
@@ -0,0 +1,95 @@
...
+const { redirectStatusSet, nullBodyStatus } = require('./constants')
+const { createInflate } = require('./util')
+
+/**
+ * Decompress the given response body.
+ * @param {Request} request
+ * @param {Response} response
+ * @returns {ReadableStream<Uint8Array> | null}
+ */
+function decompress (request, response) {
```
tsctx, line 12: I think decompressStream is sufficient

lib/web/fetch/index.js
```diff
@@ -2111,62 +2110,24 @@ async function httpNetworkFetch (
...
             status,
             statusText,
             headersList,
-            body: decoders.length
-              ? pipeline(this.body, ...decoders, () => { })
-              : this.body.on('error', () => { })
+            body: decompress(request, {
+              status,
+              statusText,
+              headers: headersList,
```
tsctx, line 2128: headersList is not an instance of Headers

index.js
```diff
@@ -138,9 +138,12 @@ module.exports.getSetCookies = getSetCookies
 module.exports.setCookie = setCookie
 
 const { parseMIMEType, serializeAMimeType } = require('./lib/web/fetch/data-url')
+const { decompress, decompressStream } = require('./lib/web/fetch/decompress')
 
 module.exports.parseMIMEType = parseMIMEType
 module.exports.serializeAMimeType = serializeAMimeType
+module.exports.decompress = decompress
+module.exports.decompressStream = decompressStream
```
tsctx, line 146: I think it is better to use the namespace `util` https://github.com/nodejs/undici/blob/main/docs/docs/api/Util.md

lib/web/fetch/decompress.js
kettanaito: I'd like to have a designated utility that performs the decompression _including_ the handling of redirect responses, etc. `decompressStream` implies you handle that yourself.

lib/web/fetch/index.js
kettanaito: Technically, not, but its APIs are compatible, from what I can see. At least, the entire test suite hasn't proven me wrong. 

I can construct `Headers` instance out of `headersList` but it may have performance implications. Would you advise me to do that?

metcoder95 (commented):
LGTM, just having the considerations of @tsctx about the util namespace and the following thread: https://github.com/nodejs/undici/pull/3423/files#r1694215822

mcollina:
Can you please add the types for this and the related type tests?

lib/web/fetch/decompress.js
tsctx: It looks to me like you are targeting the Fetch api response.
In this case, have you already been redirected?

kettanaito:
On a related subject, I've recently learned about [`DecompressionStream`](https://developer.mozilla.org/en-US/docs/Web/API/DecompressionStream) global API, which is also available in Node.js as of now. Can it be utilize to the same effect as the API being proposed here? 

```js
response.body.pipeTo(new DecompressionStream('gzip'))
```

One practical downside of this is that there's no connection between the response content encoding and the decompression used. You can also provide only _one_ decompression format (gzip, deflate, deflate-raw), so in case of response streams encoded with multiple codings, you'd have to create a pipe of decompression streams. 

Does anybody know if there's any difference in `DecompressionStream` and the content encoding handling Undici has as of now? 

I'd still much like to be consistent with Unidici here, but knowing more would be good.

lib/web/fetch/decompress.js
kettanaito: @tsctx, I'm moving the existing content encoding logic to this utility function. In the existing logic, Undici does accept the response, which is the redirect response, not the redirected response. This is correct. This utility must also decide if decompression is unnecessary if the response is a redirect response.

index.js
kettanaito: Moved it under `core/util.js`.

index.js
```diff
@@ -130,19 +141,31 @@ const { kConstruct } = require('./lib/web/cache/symbols')
 // in an older version of Node, it doesn't have any use without fetch.
 module.exports.caches = new CacheStorage(kConstruct)
 
-const { deleteCookie, getCookies, getSetCookies, setCookie } = require('./lib/web/cookies')
+const {
```
kettanaito, line 144: This is the result of running `npm run lint:fix` on this branch. You may have something misconfigured if it reports diffs.

kettanaito: Or this branch is behind and main has applied this formatting.

kettanaito:
Moved the tests to the root level of `test`, and now random tests are failing:

```js
✖ should include encodedBodySize in performance entry (1.82966ms)
  TypeError [Error]: webidl.converters.USVString is not a function
      at fetch (/Users/kettanaito/Projects/contrib/undici/index.js:121:13)
      at async Server.<anonymous> (/Users/kettanaito/Projects/contrib/undici/test/fetch/resource-timing.js:68:18)
```

The tests are in their own modules, no existing tests were edited. The errors themselves don't seem to be related to the changes. 

Does anybody know what's causing this?

index.js
KhafraDev: please undo them

mcollina:
@kettanaito are you still considering finishing this?

kettanaito:
@mcollina, yes. I've fallen off lately, doing other things. This is still on my todo list, and I'd love to see this merged. Still got those test shenanigans, no idea what causing seemingly unrelated tests to fail. I would appreciate your patience with me here, it may take me time to get back to this.

kettanaito:
I've rebased the branch against the latest `main`, and have one question:

Is the rightmost `err` stream for the body necessary just to have 3+ streams here, or does it serve a different purpose? I suspect stream errors will already propagate as the `error` event on the stream itself. 

```diff
-body: decoders.length
-  ? pipeline(this.body, ...decoders, (err) => {
-    if (err) {
-      this.onError(err)
-    }
-  }).on('error', onError)
-  : this.body.on('error', onError)

+body: decompress(request, {
+  status,
+  statusText,
+  headers: headersList,
+  body: this.body
+}).on('error', onError)
```

Basically, my rebase removed the `(err) => {}` part of the pipeline. Please let me know if that's okay. 

Also, since `decompress()` now returns the body stream as-is if there are no transformations found to decompress that stream, there's no need to have a conditional body assignment.

kettanaito:
## On code formatting

I undid the code formatting to keep the diff sane, but I highly encourage you to investigate your Prettier setup. Running `npm run lint:fix` produces a lot of formatting diff. That's an indication that either it's not set up properly, or the contributors are ignoring that and pushing changes with their own formatter settings instead of using those set up in the project.

index.js
kettanaito: Done. https://github.com/nodejs/undici/pull/3423#issuecomment-2410888383

kettanaito:
Can I please get some help with these failing tests? They don't seem to be related to the changes I'm introducing. Looks more like I need to build something or update something...

```
✖ Buffer will be returned as string via text() (0.754584ms)
  TypeError [Error]: webidl.converters.USVString is not a function
      at webidl.converters.RequestInfo (/Users/kettanaito/Projects/contrib/undici/lib/web/fetch/request.js:996:30)
      at new Request (/Users/kettanaito/Projects/contrib/undici/lib/web/fetch/request.js:104:31)
      at fetch (/Users/kettanaito/Projects/contrib/undici/lib/web/fetch/index.js:145:21)
      at TestContext.<anonymous> (/Users/kettanaito/Projects/contrib/undici/test/mock-interceptor.js:376:30)
      at Test.runInAsyncScope (node:async_hooks:206:9)
      at Test.run (node:internal/test_runner/test:631:25)
      at Suite.processPendingSubtests (node:internal/test_runner/test:374:18)
      at Test.postRun (node:internal/test_runner/test:715:19)
      at Test.run (node:internal/test_runner/test:673:12)
```

- [Test run](https://github.com/nodejs/undici/actions/runs/11326308560/job/31494990473?pr=3423)

Why is `webidl.converters.USVString` suddenly became undefined? Would appreciate assistance on this so I could proceed with adding type tests and having this merged. 

cc @KhafraDev @mcollina

lib/core/util.js
```diff
@@ -895,6 +894,13 @@ function getCodings (contentEncoding) {
   return codings
 }
 
+let createInflate
+
+function lazyCreateInflate (zlibOptions) {
+  createInflate ??= require('../web/fetch/util').createInflate
```
kettanaito, line 900: @KhafraDev, wouldn't it make more sense to lift `createInflate` to `../core/util` in this case? If multiple modules depend on it, perhaps it's a sign to make it a generic utility.

KhafraDev: I went for the path of least resistance, I have no objections in either case.

KhafraDev (approved):

test/decompress.js
```diff
@@ -0,0 +1,141 @@
...
+      }
+    })
+  )
+  assert.equal(await new Response(body).text(), 'hello world')
+})
+
+/**
+ * @note Undici does not support the "CONNECT" request method.
+ */
+test.skip('ignores CONNECT requests', async () => {
```
metcoder95, line 59: We should, or what do you mean exactly?

mcollina (commented):
Should this utility get publicly exposed?

kettanaito:
@mcollina, yes, that's the intention. We would like to reuse it in MSW to have a consistent body decompression across the board. As suggested by prior reviewers, we can expose this from utils, but I suppose that still means the root export of the package.

kettanaito:
@KhafraDev, thank you so much for your support on this one! 

I believe there are still type tests remaining on my end, and making sure we export this correctly (perhaps @mcollina has some thoughts on this).

test/decompress.js
Uzlopak: CONNECT is a special method which opens a tunnel.

mcollina:
@kettanaito add the export from the root package and related docs.

kettanaito:
Will do! I should have some time tomorrow to finalize this.

KhafraDev:
Is this needed after https://github.com/mswjs/interceptors/pull/661/? I think there are some semver issues to consider along with the general weirdness of exposing internals for public use

kettanaito:
@KhafraDev, that's a good question. I've come to realize that we wouldn't be able to reuse this compression straight away, and would have to either ship `undici` alongside Interceptors, or inline its decompression logic anyway. 

I would rather not ship `undici` because it creates two separate universes of the bundled Undici in the user's Node, and explicit Undici they install with MSW. Not a good thing. 

I was surprised to learn about the Compression Streams API. Sadly, it doesn't support Brotli, and I'm researching the possible ways to move forward here. 

That being said, I still find the `decompress` logic being exposed as beneficial to Undici and its consumers. I believe there were issues that linked to its need as well.

Uzlopak:
I think that moving the code from fetch to core is wrong. decompress expects Request and Response from fetch and not the core Request and Response instances. So the decompression needs to be moved back to the fetch folder. Or do I understand that wrong?

Uzlopak:
Closing due to inactivity.

PR #3453, opened 2024-08-13
test: streamline test scripts in regard of without-intl and run more tests for without-intl case

Uzlopak:
streamline the test scripts and also run e.g. websocket tests.


Uzlopak:
I fixed it already locally but forgot to push..

PR #3421, opened 2024-07-26
fix: restore externalized Node.js dep compatibility

richardlau:
## This relates to...

Fixes: https://github.com/nodejs/undici/issues/3345
Refs: https://github.com/nodejs/undici/pull/2643

## Rationale

Changes were previously accepted to allow Undici to be built to work with Node.js' `--shared-builtin-undici/undici-path` `configure` option. This allows Undici to exist on disk instead of being compiled into the Node.js binary. This is used by downstream Node.js packagers (e.g. Linux distributions, such as Fedora) that require the ability to build everything from source (including all dependencies and tooling). 

This functionality was regressed by https://github.com/nodejs/undici/pull/3074 and was not noticed as nothing in this repository broke.

## Changes

Restore the ability to build Undici compatible with Node.js' `configure --shared-builtin-undici/undici-path ...` build option and adds a workflow to test this configuration.

Scopes the `hasApk` conditional to only cover the part that requires `apk`. This allows `build/wasm.js` to be run on a non-Alpine based Linux distribution.

Makes the WASM optimizer (binaryen) optional to allow building on Linux distributions that do not package `binaryen` and must be able to rebuild everything (including tooling) from source.

### Features

N/A

### Bug Fixes

- Restore the ability to build Undici usable with Node.js built with `configure --shared-builtin-undici/undici-path ...`.

### Breaking Changes and Deprecations

N/A

## Status

- [x] I have read and agreed to the [Developer's Certificate of Origin][cert]
- [x] Tested
- [ ] Benchmarked (**optional**)
- [x] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

[cert]: https://github.com/nodejs/undici/blob/main/CONTRIBUTING.md

PR #3454, opened 2024-08-13
build: parametrize the location of wasm-opt

khardix:
## This relates to...

Enhances: https://github.com/nodejs/undici/pull/3421

## Rationale

The wasm-opt binary may be available in different place than the local directory (`./wasm-opt`) – for example, in `/usr/bin/wasm-opt`.

## Changes

Similarly to the other parametrized WASM build options, use WASM_OPT environment variable to identify the wanted binary, with fallback to the previous value (`./wasm-opt`).

Even with the new environment variable, the `hasOptimizer` check is kept for the cases where no optimizer is available anywhere.

### Features

N/A

### Bug Fixes

N/A

N/A

## Status

- [x] Tested
- [x] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready


Uzlopak (approved):
LGTM

PR #3457, opened 2024-08-14
types: allow non strict HTTPMethod

### LINKED ISSUE #3168: Type 'string' is not assignable to type 'HttpMethod'.ts(2345)
Can we change the type of `method` to just regular string? I keep getting this ts warning.

Uzlopak:
fixes #3168 

I still think, that it reduces the necessary strictness, as now also lowercased methods like "get" are allowed via types, but are invalid values. 

So at transpile time we get no warning if the value is wrong. But atleast we keep the autocomplete capability. 

I  think that #3168 should be closed as on wont fix, but here is the solution.

For more inforation for the solution see https://github.com/fastify/fastify/pull/5046


mcollina (approved):
lgtm

PR #3462, opened 2024-08-15
add autocompletable header types

KhafraDev:
the header keys were taken from [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers) and the values for the content-type header were taken from [hono](https://github.com/honojs/hono/commit/ac069d1696536e6d8f43e9b360cbee389035e15f)


KhafraDev:
The types only work for the object param of the constructor. It would be possible to add support for the other methods, but would require a generic which would interfere with the DOM types.

types/fetch.d.ts
```diff
@@ -67,7 +67,7 @@ export interface SpecIterable<T> {
   [Symbol.iterator](): SpecIterator<T>;
 }
 
-export type HeadersInit = string[][] | Record<string, string | ReadonlyArray<string>> | Headers
+export type HeadersInit = [string, string][] | HeaderRecord | Headers
```
KhafraDev, line 70: ReadonlyArray<string> is just not correct. If we were to follow the logic in the PR that added it, the type should be any.

Any array pair that doesn't contain exactly 2 strings throws an error too, so string[][] is not strict enough.

mcollina (approved):
lgtm

types/header.d.ts
```diff
@@ -2,3 +2,155 @@
  * The header type declaration of `undici`.
  */
 export type IncomingHttpHeaders = Record<string, string | string[] | undefined>;
+
+type HeaderNames =
```
Uzlopak, line 6: I think they should be lowercased?

types/header.d.ts
```diff
@@ -2,3 +2,155 @@
...
+  | 'Upgrade-Insecure-Requests'
+  | 'User-Agent'
+  | 'Vary'
+  | 'Via'
+  | 'WWW-Authenticate'
+  | 'X-Content-Type-Options'
+  | 'X-Frame-Options'
+  | (string & {})
+
+type KnownHeaderValues = {
```
Uzlopak, line 98: Why not a type ContentType ?

types/header.d.ts
```diff
@@ -2,3 +2,155 @@
...
+    | 'application/zip'
+    | 'video/3gpp'
+    | 'video/3gpp2'
+    | 'model/gltf+json'
+    | 'model/gltf-binary'
+    | (string & {})
+}
+
+export type HeaderRecord = {
+  [K in HeaderNames | Lowercase<HeaderNames>]?: Lowercase<K> extends keyof KnownHeaderValues
```
Uzlopak, line 153: Why do we lowercase them via a utiliy type?

types/header.d.ts
```diff
@@ -2,3 +2,155 @@
...
+  | 'Transfer-Encoding'
+  | 'Upgrade'
+  | 'Upgrade-Insecure-Requests'
+  | 'User-Agent'
+  | 'Vary'
+  | 'Via'
+  | 'WWW-Authenticate'
+  | 'X-Content-Type-Options'
+  | 'X-Frame-Options'
+  | (string & {})
```
Uzlopak, line 96: Maybe create a utility.d.ts if not exisiting already and create a Autocomplete type?

Uzlopak (changes requested):
We should discuss this a little bit before merging.

types/header.d.ts
Ethan-Arrowood: I believe this expands to allow the user to specify either the uppercased (`Accept`) or lowercased (`accept`) option depending on their own personal preference. 

Per spec Headers can be specified either way and are normalized to be the same thing.

Ethan-Arrowood (approved):
lgtm

types/header.d.ts
KhafraDev: they are in the map, so it auto completes both the cased and lower cased values

types/header.d.ts
KhafraDev: I assumed that there would be more headers that we'd want to add types to

types/header.d.ts
KhafraDev: so we have intellisense for both 'Content-Type' and 'content-type'

PR #3463, opened 2024-08-15
fix: do validation first before actual business logic, like super()

Uzlopak:
Highly opinionated change. Or is there some hidden logic, which says, that super() calls are optimized in v8?


ronag (approved):
🤷‍♂️

mcollina (approved):
lgtm

PR #3466, opened 2024-08-16
perf: use isIPv6 for checking if hostname is isIPv6

Uzlopak:
trivial change. isIP will first run isIPv4 and then isIPv6. Also both are basically calling regexes on node. So by calling isIPv6 directly we improve the performance.

### DIFF
--- a/lib/dispatcher/client.js
+++ b/lib/dispatcher/client.js
@@ -374,7 +374,7 @@ async function connect (client) {
     assert(idx !== -1)
     const ip = hostname.substring(1, idx)
 
-    assert(net.isIP(ip))
+    assert(net.isIPv6(ip))
     hostname = ip
   }
 


ronag:
But this will fail on ipv4 addresses now?

Uzlopak:
Nope. It can not be a IP v4. This is basically "unpacking" a IPv6 from the bracket notation. Like `http://[::1]:3000`. hostname is then `[::1]` but needs to be `::1`. So this is detecting the wrapping with brackets by checking the if it is starting with a bracket etc..

An IPv4 would not be valid. 

See [RFC 3986, section 3.2.2.](https://www.ietf.org/rfc/rfc3986.txt)

```
   A host identified by an Internet Protocol literal address, version 6
   [RFC3513] or later, is distinguished by enclosing the IP literal
   within square brackets ("[" and "]").  This is the only place where
   square bracket characters are allowed in the URI syntax.  In
   anticipation of future, as-yet-undefined IP literal address formats,
   an implementation may use an optional version flag to indicate such a
   format explicitly rather than rely on heuristic determination.
```

ronag (approved):

PR #3467, opened 2024-08-16
fix: remove unnecessary parameters in USVString calls

Uzlopak:
USVString only expects a value parameter.

### DIFF
--- a/lib/web/eventsource/eventsource.js
+++ b/lib/web/eventsource/eventsource.js
@@ -115,7 +115,7 @@ class EventSource extends EventTarget {
       })
     }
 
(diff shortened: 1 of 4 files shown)


KhafraDev (commented):
I think errors coming from toUSVString should match the rest of the webidl impl. Can toWellFormed throw?

Uzlopak:
I dont think that toWellFormed can throw. Its basically a put any garbage in and get a valid/well formed string back.

Wdyt?

KhafraDev:
It technically can throw, but not in any way undici is using it.

String.prototype.toWellFormed.call(Symbol() / null / undefined)

KhafraDev (approved):

PR #3472, opened 2024-08-16
fix: instantiation of ResponseError, pass headers and data correctly

Uzlopak:
It is disputable. Maybe body should not be passed as data?


mertcanaltin (approved):
lgtm

PR #3475, opened 2024-08-16
fix: fix codesmells in retry-handler

Uzlopak:
triivial changes


ronag (approved):

KhafraDev:
I don't think we should consider string.startsWith a code smell.

PR #3477, opened 2024-08-16
fix: set default value for disableColors to false in PendingInterceptorsFormatter

Uzlopak:
trivial change, reduces a codesmell.


metcoder95 (approved):

ronag (changes requested):
This is kind of breaking... if one passes null

PR #3483, opened 2024-08-17
Update WPT

Uzlopak:
Automated update of the WPT


KhafraDev:
So the action can't create an issue but it can comment on them? We can wait for permissions to get changed, there isn't anything important (I don't see a single new test being run).

mcollina:
> 
So the action can't create an issue but it can comment on them? We can wait for permissions to get changed, there isn't anything important (I don't see a single new test being run).


This should be fixed, try again.

PR #3485, opened 2024-08-18
chore: replace standard and snazzy with neostandard

Uzlopak:
Migrate from standard and snazzyy to neostandard by @voxpelli . Also starting to lint typescript files.

To make it easier to review:

- First commit is just setting up eslint with neostandard instead of standard and snazzy, + configure for our project
- Second commit is with automatic fixing via `--fix` flag
- Third commit is manually fixing linting errors


eslint.config.js
```diff
@@ -0,0 +1,32 @@
+'use strict'
+
+const neo = require('neostandard')
+
+module.exports = [
+  {
+    ignores: [
```
voxpelli, line 7: You don't need these separate ignores if you provide them to neostandard as well, the neostandard ones are set as global ignores 🙂

Uzlopak: done :)

mertcanaltin (approved):
LGTM 🚀

ronag (commented):
Why? If we are going this way I'd rather go with eslint + prettier.

eslint.config.js
```diff
@@ -0,0 +1,29 @@
...
+
+const neo = require('neostandard')
+
+module.exports = [
+  ...neo({
+    env: ['jest'],
+    ignores: [
+      'lib/llhttp',
+      'test/fixtures/wpt',
+      'node_modules',
```
voxpelli, line 11: Not needed, this is already part of ESLint's default ignore: https://github.com/eslint/eslint/blob/21d3766c3f4efd981d3cc294c2c82c8014815e6e/lib/config/default-config.js#L59

eslint.config.js
```diff
@@ -0,0 +1,29 @@
...
+      'undici-fetch.js'
+    ],
+    ts: true
+  }),
+  {
+    rules: {
+      '@stylistic/comma-dangle': ['error', {
```
voxpelli, line 18: Would love to get feedback on https://github.com/neostandard/neostandard/issues/79 to make this unnecessary 🙏

eslint.config.js
```diff
@@ -0,0 +1,29 @@
...
+        arrays: 'never',
+        objects: 'never',
+        imports: 'never',
+        exports: 'never',
+        functions: 'never'
+      }],
+      '@typescript-eslint/no-redeclare': 'off',
```
voxpelli, line 25: Was this a particular issue with TS-files? Maybe it's something we should tweak in neostandard itself?

eslint.config.js
```diff
@@ -0,0 +1,29 @@
...
+      'no-lone-blocks': 'off'
```
voxpelli, line 26: Was this a regression since standard? Or what prompted the addition?

eslint.config.js
Uzlopak: this basically happens if you have something like


```ts
export interface CacheStorage {
  match (request: RequestInfo, options?: MultiCacheQueryOptions): Promise<Response | undefined>,
  has (cacheName: string): Promise<boolean>,
  open (cacheName: string): Promise<Cache>,
  delete (cacheName: string): Promise<boolean>,
  keys (): Promise<string[]>
}

declare const CacheStorage: {
  prototype: CacheStorage
  new(): CacheStorage
}
```

Because the interface CacheStorage exists declaring the const CacheStorage results in a no-redeclare linting error

eslint.config.js
Uzlopak: If you have 

```
{
  new Date()
}
```

you get no-lone-blocks linting error but if you do 

```
{
  const date = new Date()
}
```
You wont.

Not a regression because we didnt lint ts files before.

Removed those blocks

mcollina (approved):
lgtm

mcollina:
> Why? If we are going this way I'd rather go with eslint + prettier.

prettier is incompatible with standard so it would be impossible to do backports. neostandard is just an eslint config and backward compatible with standard.

Uzlopak:
I worked now together with @voxpelli regarding this PR, because the linting was quite slow (~10s). Disabling jsx to avoid running react related linting saved 2 s. Also activated eslint cache makes it run significantly faster. 

So the first run is slow but consecutive runs will have the same perf as before.

Uzlopak:
Well. I think the PR is now mature enough for the merge.

So if @ronag 's is ok with it, then we could merge it :).

metcoder95 (commented):
lgtm

eslint.config.js
voxpelli: If you want, you can update to [`0.11.3`](https://github.com/neostandard/neostandard/releases/tag/v0.11.3) now and this override will no longer be needed (though having it will enforce no dangling commas if that's desireable)

KhafraDev (approved):

Issue #3493, opened 2024-08-23
`UND_ERR_HEADERS_TIMEOUT` when changing system time during fetch


IboKhaled:
## Bug Description

When the system time is changed forward by more than 5 minutes during a `fetch()` call, the operation terminates with `UND_ERR_HEADERS_TIMEOUT`.

As far as I can tell, this is due to how the timeout mechanism is implemented in `lib/util/timers.js`.
Timers above 1000 ms are realized with a custom, low resolution timer implementation, which relies on `Date.now()`. Therefore the timers are influenced by the system time. (I guess this would also mean the timeout would fire late/never when time is set back during an active `fetch()` operation too.)
In my testing, JS inbuilt `setTimeout()` is not impacted by system time changes.

Is this expected behavior?
Are there valid reasons why the timeout should be relative to the wall clock?
If this is a bug, could monotonic time functions like `performance.now()` or `process.hrtime` be used here instead?

## Reproducible By

1. Perform a long fetch operation
2. Set system time forward by more than 5 minutes or `headersTimeout`
3. Observe timeout error

Here is an unit test that verifies the early timeout for `test/util.js`:

```JS
test("do not timeout early on forward system time change", async (ctx) => {
  let t = tspl(ctx, { plan: 3 });

  let dateMock = ctx.mock.method(Date, "now");
  dateMock.mock.mockImplementation(() =>
    new Date("2024-08-22T10:00:00.000Z").valueOf()
  );

  const start = process.hrtime.bigint();

  timers.setTimeout(() => {
    const delta = getDelta(start, 2000);
    t.ok(delta >= -1n, "timer fired early after system time change");
    t.ok(
      delta < ACCEPTABLE_DELTA,
      "timer fired late after system time change"
    );
  }, 2000);

  // Simulate user setting time forwards with active timer
  setTimeout(
    () =>
      dateMock.mock.mockImplementation(() =>
        new Date("2024-08-22T11:00:00.000Z").valueOf()
      ),
    750
  );

  setTimeout(() => t.ok(true), 2500);
  await t.completed;
});
```

## Expected Behavior

A `fetch()` calls timeout should not be influenced by system time.

## Environment

Observed in Node v18.14.2, unit test written for undici on current main (69cfd975).

Uzlopak:
Could you please explain, in which cases you change your system time?

IboKhaled:
I start the fetch operation, which triggers a change in the system time. This fetch operation times out because of the change in system time it itself triggered.

[...]

PR #3494, opened 2024-08-23
fix: tick timer instead of using Date.now

ronag:
Partially fixes some of the timeout issues we have been facing. Again needs work with tests.

Refs: https://github.com/nodejs/undici/issues/3493

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/util/timers.js
+++ b/lib/util/timers.js
@@ -2,13 +2,13 @@
 
 const TICK_MS = 499
 
-let fastNow = Date.now()
+let fastNow = 0
 let fastNowTimeout
 
 const fastTimers = []
 
 function onTimeout () {
-  fastNow = Date.now()
+  fastNow += TICK_MS
 
   let len = fastTimers.length
   let idx = 0


mcollina:
I'm generically ok with the change, but I would be happier with tests. Maybe one from @Uzlopak's PR?

ronag:
I suspect that or includes this change somewhere

Uzlopak:
This PR is replacing the Date.now() call with saying that every time setTimeout is triggered that 499 ms are passed. This means, that if the event loop blocks for e.g. 900ms, we still "think" that just 499ms passed.  So the fasttimer will be artificially delayed.

Uzlopak:
We could change it in my PR to just increment the fastNow value instead of calling performance.now. But somehow it feels wrong.

ronag:
The whole point is that we don't want to include event loop delay

Uzlopak:
adapted
https://github.com/nodejs/undici/pull/3495/commits/ad339a2feddf08267e287799332cd5ed48ba799a

PR #3495, opened 2024-08-23
fix: refactor fast timers, fix UND_ERR_CONNECT_TIMEOUT on event loop blocking

### LINKED ISSUE #3410: `UND_ERR_CONNECT_TIMEOUT` errors thrown when there is CPU intensive code on the event loop
## Bug Description

Work on the event loop can interrupt the Undici lifecycle for making requests, causing errors to be thrown even when there is no problem with the underlying connection. For example, if a fetch request is started and then work on the event loop takes more than 10 seconds (default connect timeout), Undici will throw a `UND_ERR_CONNECT_TIMEOUT` error even if the connection could be established very quickly.

I believe what is happening is:

1. When the fetch request is started, Undici starts the work to make a connection. Undici calls `setTimeout` with the value of the `connectTimeoutMs` to throw an error and cancel the connection if it takes too long (https://github.com/nodejs/undici/blob/main/lib/core/connect.js). It makes a call to `GetAddrInfoReqWrap` (https://github.com/nodejs/node/blob/main/lib/dns.js#L221), but this is asynchronous and processing of the callback will be delayed until the next event loop.
2. User tasks block the event loop for a long period of time.
3. The `onConnectTimeout` timer is run because the previous task took longer than the timeout. `onConnectTimeout` calls `setImmediate` with a function to destroy the socket and throw the error. https://github.com/nodejs/undici/blob/main/lib/core/connect.js
4. The `GetAddrInfoReq` lookup callback (`emitLookup` in `node:net`) is run. This code begins the TCP connection (`internalConnect` is called in https://github.com/nodejs/node/blob/main/lib/net.js#L1032) but that is also asynchronous, so it won't finish in this round of the event loop.
5. The `setImmediate` function is run in the next phase which destroys the socket and throws the `UND_ERR_CONNECT_TIMEOUT` error.
6. Undici never gets a chance to handle the TCP connection response.

Internally at Vercel, we have been seeing a high number of these `UND_ERR_CONNECT_TIMEOUT` issues while pre-rendering pages in our Next.js application. I can't run this task on my local machine so it's harder to debug, but it's a CPU intensive task and moving fetch requests to a worker thread eliminated the Undici errors. We tried other suggestions (like `--dns-result-order=ipv4first` and verified that we were not seeing any packet loss) that did not resolve the issue. Increasing the connect timeout resolves the issue in the reproduction but _not_ the issue in our Next.js build (which I can't explain).

## Reproducible By

A minimal reproduction is available at https://github.com/mknichel/undici-connect-timeout-errors.

We can reproduce the behavior on Node 18.x and 20.x and with the `5.24.0` and the latest version of Undici (`6.19.2`)

## Expected Behavior

The Undici request lifecycle could operate on a separate thread that does not get blocked by user code. By separating it out from the user code, this would remove impact of any user code on requests.

To test this theory, we created a dispatcher that proxied the fetch request to a dedicated worker thread (`new Worker` from `worker_threads`). This eliminated all the Undici errors that we were seeing in our Next.js build.

## Logs & Screenshots

In the minimal reproduction, the error is:

```
TypeError: fetch failed
    at fetch (/Users/mknichel/code/tmp/undici-connect-timeout-errors/node_modules/.pnpm/undici@6.19.2/node_modules/undici/index.js:112:13)
    at fetchExample (/Users/mknichel/code/tmp/undici-connect-timeout-errors/index.ts:21:20)
    at main (/Users/mknichel/code/tmp/undici-connect-timeout-errors/index.ts:66:3) {
  [cause]: ConnectTimeoutError: Connect Timeout Error
      at onConnectTimeout (/Users/mknichel/code/tmp/undici-connect-timeout-errors/node_modules/.pnpm/undici@6.19.2/node_modules/undici/lib/core/connect.js:190:24)
      at /Users/mknichel/code/tmp/undici-connect-timeout-errors/node_modules/.pnpm/undici@6.19.2/node_modules/undici/lib/core/connect.js:133:46
      at Immediate._onImmediate (/Users/mknichel/code/tmp/undici-connect-timeout-errors/node_modules/.pnpm/undici@6.19.2/node_modules/undici/lib/core/connect.js:174:9)
      at process.processImmediate (node:internal/timers:478:21) {
    code: 'UND_ERR_CONNECT_TIMEOUT'
```

In our Next.js builds, the error is:

```
TypeError: fetch failed
    at node:internal/deps/undici/undici:12618:11
    at async s (elided path)
    at async elided path {
  cause: ConnectTimeoutError: Connect Timeout Error
      at onConnectTimeout (node:internal/deps/undici/undici:7760:28)
      at node:internal/deps/undici/undici:7716:50
      at Immediate._onImmediate (node:internal/deps/undici/undici:7748:13)
      at process.processImmediate (node:internal/timers:478:21)
      at process.callbackTrampoline (node:internal/async_hooks:130:17) {
    code: 'UND_ERR_CONNECT_TIMEOUT'
    }
  }
```

## Environment

The reproduction repo was erroring for me on Mac OS 14.4, while internally we are seeing issues on AWS EC2 Intel machines.

### Additional context

Vercel/Next.js users have reported `UND_ERR_CONNECT_TIMEOUT` issues to us:

- https://github.com/vercel/vercel/issues/11692
- https://github.com/vercel/next.js/issues/66373

### LINKED ISSUE #3493: `UND_ERR_HEADERS_TIMEOUT` when changing system time during fetch

Uzlopak:
superseeds #3494 
superseeds #3418 
fixes #3493
fixes #3410

-  an internal "clock" is now used instead of `Date.now()` - ensuring that changes of the system clock wont effect the timers
- `timers.js` exports a `now` method to retrieve the value of the internal clock
- fasttimers clearTimeout expects a FastTimer instance or a native Timer. Before it accepted also undefined or null values
- connect timeouts are now using fasttimers

Based on the work of @ronag i refactored the fast timers and hopefully fixed also the connection timeout bug when the event loop is blocked.


ronag (commented):
This is a huge PR... not sure I have time to review properly in forseeable future.

thisislvca:
This is a big issue over at Vercel, know it's a big PR but it's also quite important for lots of people :)

ronag:
There are simpler PRs that solve the issue. I think the problem here is that we are combining a significant refactoring with bug fixes.

mcollina (approved):
LGTM

can you run the benchmark before/after to see if this impacted us at all?

mcollina:
After a quick skim it seems fine, and all tests are passing. If it introduces some regresisons, then I'm pretty sure some users of Next.js/Vercel would report this quickly find out.

ronag:
I'm +-0

Uzlopak:
I want to add a small explaination of this refactoring. 

I wanted to modify the fasttimer implementation. Yes it is kind of simple while you are implementing it and everything makes sense. But when working on it the first time, it is kind of hard to grok. So I started to comment the code and e.g. also replace the magic numbers with constants, and renamed the atrtibutes to be more expressive.

While I was investigating how the native timers are structured, I realized I should use their data structure/naming of the attributes. Thats why I renamed e.g. delay with _idleTimeout or callback with _onTimeout.

image

Also I tried to rename variables. when a different name was making more sense. E.g. in client-h1.js `value` to `delay`.

So in my opinion to review the timers.js file the first step is to just read the comments and check if it matches with the code it describes. And if it is good commented you should have a good idea on how the fast timers actually work. 

The rest should be then easy to review.

ronag:
IMHO all the comments and explanations make it harder to grasp and read. But if everyone else is happy with it I won't block. Good job.

ronag:
Could you maybe summarize what behavior changes this introduces?

Uzlopak:
Of course. I will in about 30 min, when i am back from groceries shopping ;)

Uzlopak:
Well, I was looking into the code while writing what changes I implemented. I made a coverage check and realized, that the changes in client-h1.js seem to be not covered by a unit test. So I am currently implementing a unit test for that, just as a sanity check.

Uzlopak:
I reverted the changes regarding timeoutDeadline in client-h1,js as those changes are not covered by a unit test. As soon as I find a test case and those lines fix the test then I would add them back.

thisislvca:
Is there a chance to push out the Vercel-related errror fix first (so the most urgent) and then open a separate PR for everything else?

metcoder95 (approved):
lgtm

ronag (approved):

trivikr:
Should this be backported to 6.x?

A backport was requested in follow-up PR, but it failed because of merge conflicts https://github.com/nodejs/undici/pull/3552#issuecomment-2389079958

metcoder95:
I believe it should, but will need manual backport

cc: @Uzlopak @mcollina

trivikr:
The backport fails to 6.x, as reducing memory usage in client-h1 from https://github.com/nodejs/undici/pull/3510 is not backported.

mcollina:
@trivikr if you want to open a Pr to backport those commits down, please be our guest!

Uzlopak:
@mcollina 

Thanks to the comment by @trivikr I understand now why the backporting failed. Anyhow, I think only #3675 needs to be merged?!

trivikr:
> if you want to open a Pr to backport those commits down, please be our guest!

I followed the instructions posted by the bot in https://github.com/nodejs/undici/pull/3495#issuecomment-2390645589, and then found out backporting failed as the previous PR wasn't backported. I thus posted https://github.com/nodejs/undici/pull/3495#issuecomment-2390680166 to inform maintainers to take appropriate actions.

I'll provide console output in future, so that there's clarification about the comment.

PR #3497, opened 2024-08-23
fix(types): MockAgent accepts ProxyAgent, EnvHttpProxyAgent and RetryAgent for agent option

### LINKED ISSUE #3444: MockAgent not compatible with ProxyAgent,EnvHttpProxyAgent,RetryAgent
## Bug Description

`ProxyAgent`, `EnvHttpProxyAgent`, `RetryAgent` can't be passed to `new MockAgent({ agent })`, because they're missing `closed` and `destroyed` props.

## Environment

Win 10, Node v20.13.1

### Additional context

Not sure if I am approching this correctly, but I would like to handle non-itercepted requests exactly the same way as they would be by the app. So my assumption is that I am supposed to pass an agent used by my application to `MockAgent` and it will be used for non-intercepted requests.

Uzlopak:
Fixes #3444

### DIFF
--- a/test/types/mock-agent.test-d.ts
+++ b/test/types/mock-agent.test-d.ts
@@ -1,6 +1,10 @@
 import { expectAssignable, expectType } from 'tsd'
 import { Agent, Dispatcher, MockAgent, MockClient, MockPool, setGlobalDispatcher } from '../..'
 import { MockInterceptor } from '../../types/mock-interceptor'
+import ProxyAgent from '../../types/proxy-agent'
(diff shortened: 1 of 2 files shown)


KhafraDev:
The issue sounds like a runtime issue, not a type issue.

Uzlopak:
I checked and MockAgent doesnt use the closed and destroyed attributes of the agent. Atleast I dont see it in the code. And the the here added typings tests fail in main, atleast you see the swiggly line in vscode

KhafraDev (approved):

PR #3502, opened 2024-08-24
remove third party everything support in fetch

KhafraDev:
Third party Blob, FormData, and AbortController in fetch should never have landed. Even after it did, there was never any documentation on it - the only evidence that this was a feature was in tests and a few comments on issues.

Support is so bad, currently, in fact, that webidl skips validation entirely. Take `webidl.converters.Blob` for example: it's an interface converter (ie. validating that an object is an instance of a class), but when passed the `strict: false` option, it wouldn't perform any validation.

I understand that support was added for performance reasons, but it's something that we should have let node core fix, rather than "node-ifying" the spec. For use cases that are not possible with fetch, we should start recommending undici directly.


ronag (approved):

KhafraDev:
Node v22.7.0 broke some of the autobahn tests, confirmed it locally. When running in v22.6.0, all the tests pass. It wasn't caught because the autobahn suite only runs when changes to `lib/web/websocket` are made. @Uzlopak

KhafraDev:
It's probably the buffer bugs, thinking about it.

Uzlopak:
@KhafraDev 
can you look at #3503  to avoid something similar in the future?

lib/web/fetch/request.js
```diff
@@ -400,16 +400,6 @@ class Request {
 
     // 29. If signal is not null, then make this’s signal follow signal.
     if (signal != null) {
-      if (
```
KhafraDev, line 403: this is no longer needed because we are actually validating signal in webidl now

Uzlopak (approved):
LGTM

PR #3504, opened 2024-08-24
test: improve gc detection

Uzlopak:
Now it will throw a helpful error, when gc is not exposed

image


KhafraDev:
not running the tests is worse than throwing an error imo

mcollina (approved):
lgtm

mcollina (changes requested):
Actually, the test should fail

Uzlopak:
@mcollina 
@KhafraDev 
PTAL

KhafraDev (approved):

PR #3505, opened 2024-08-25
fix: setEncoding should not throw on body #1125

### LINKED ISSUE #1125: client.request throws Error with setEncoding
## Bug Description

`client.request` throws an error when `body.json()` & `body.text()` are called with `setEncoding`.

## Reproducible By

```javascript
import { once } from 'events'
import assert from 'assert'

const data = 'a'.repeat(100)

const server = createServer((request, response) => {
  response.end(JSON.stringify({ data }))
}).listen()

try {
  const { body, headers, statusCode } = await client.request({
    path: '/',
    method: 'GET'
  })
  console.log(`response received ${statusCode}`)
  body.setEncoding('utf8')
  const json = await body.json()
  assert.strictEqual(json.data, data, 'multi byte')
} catch (error) {
  console.error('error!!:', error)
} finally {
  client.close()
  server.close()
}
```

```bash
$ node index.mjs
error!!: TypeError [ERR_INVALID_ARG_TYPE]: The "list[0]" argument must be an instance of Buffer or Uint8Array. Received type string ('{"data":"aaaaaaaaaaaaaaa...)
    at new NodeError (node:internal/errors:371:5)
    at Function.concat (node:buffer:559:13)
    at consumeEnd (/Users/kohtaito/dev/tmp/test_undici/node_modules/undici/lib/api/readable.js:238:33)
    at BodyReadable.<anonymous> (/Users/kohtaito/dev/tmp/test_undici/node_modules/undici/lib/api/readable.js:220:7)
    at BodyReadable.emit (node:events:390:28)
    at BodyReadable.emit (/Users/kohtaito/dev/tmp/test_undici/node_modules/undici/lib/api/readable.js:66:18)
    at endReadableNT (node:internal/streams/readable:1343:12)
    at processTicksAndRejections (node:internal/process/task_queues:83:21) {
  code: 'ERR_INVALID_ARG_TYPE'
}
```

## Environment

Node.js v16.13.1
undici v4.11.0

### Additional context

`chunk` is converted to a string by StringDecoder only when consumeStart is executed.

```diff
$ git diff
diff --git a/lib/api/readable.js b/lib/api/readable.js
index 3bb454e..8ea758e 100644
--- a/lib/api/readable.js
+++ b/lib/api/readable.js
@@ -94,6 +94,7 @@ module.exports = class BodyReadable extends Readable {

   push (chunk) {
     if (this[kConsume] && chunk !== null && !this[kReading]) {
+      console.log('BodyReadable: push', typeof chunk, this.readableEncoding)
       consumePush(this[kConsume], chunk)
       return true
     } else {
@@ -196,6 +197,7 @@ function consumeStart (consume) {
   const { _readableState: state } = consume.stream

   for (const chunk of state.buffer) {
+    console.log('consumeStart', typeof chunk, consume.stream.readableEncoding)
     consumePush(consume, chunk)
   }
```

```bash
$ node index.mjs
response received 200
headers {
  date: 'Mon, 06 Dec 2021 09:08:48 GMT',
  connection: 'keep-alive',
  'keep-alive': 'timeout=5',
  'content-length': '10000011'
}
consumeStart string utf8
BodyReadable: push object utf8
BodyReadable: push object utf8
BodyReadable: push object utf8
BodyReadable: push object utf8
BodyReadable: push object utf8
```

I think that this code needs to use `consume.stream._readableState.decoder.end()` in `consumeEnd` instead of `Buffer.concat(body)` if the value is in `readableEncoding`.
Is this the right way to fix it?

I also considered a method to overwrite StringDecoder so that it is not created in setEncoding.
However, it seems that it will be difficult to upgrade because it depends too much on the structure of Stream.

Ref: https://github.com/nodejs/undici/issues/1119

CC: @mcollina

Uzlopak:
Fixes #1125 

Super old bug #1125 is the result of some premature encoding of already buffered content in node core.

https://github.com/nodejs/node/blob/4e68b541fd7ccd9fe1328d5b4dc93a3689aa505d/lib/internal/streams/readable.js#L585

This PR overwrites the setEncoding function, with a custom solution which is not encoding the buffered content.

### DIFF
--- a/lib/api/readable.js
+++ b/lib/api/readable.js
@@ -201,6 +201,17 @@ class BodyReadable extends Readable {
         .resume()
     })
   }
+
+  /**
+   * @param {BufferEncoding} encoding
+   * @returns {BodyReadable}
+   */
+  setEncoding (encoding) {
+    if (Buffer.isEncoding(encoding)) {
+      this._readableState.encoding = encoding
+    }
+    return this
+  }
 }
 
 // https://streams.spec.whatwg.org/#readablestream-locked
@@ -278,10 +289,10 @@ function consumeStart (consume) {
   }
 
   if (state.endEmitted) {
-    consumeEnd(this[kConsume])
+    consumeEnd(this[kConsume], this._readableState.encoding)
   } else {
     consume.stream.on('end', function () {
-      consumeEnd(this[kConsume])
+      consumeEnd(this[kConsume], this._readableState.encoding)
     })
   }
 
@@ -295,8 +306,10 @@ function consumeStart (consume) {
 /**
  * @param {Buffer[]} chunks
  * @param {number} length
+ * @param {BufferEncoding} encoding
+ * @returns {string}
  */
-function chunksDecode (chunks, length) {
+function chunksDecode (chunks, length, encoding) {
   if (chunks.length === 0 || length === 0) {
     return ''
   }
@@ -311,7 +324,11 @@ function chunksDecode (chunks, length) {
     buffer[2] === 0xbf
       ? 3
       : 0
-  return buffer.utf8Slice(start, bufferLength)
+  if (!encoding || encoding === 'utf8' || encoding === 'utf-8') {
+    return buffer.utf8Slice(start, bufferLength)
+  } else {
+    return buffer.subarray(start, bufferLength).toString(encoding)
+  }
 }
 
 /**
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

ronag (approved):

PR #788, opened 2021-05-05
fix: detached buffer error

ronag:
Somehow the error `Cannot perform %TypedArray%.prototype.set on a detached ArrayBuffer` occurs when using a saved buffer view.

```js
[1620202083995] FATAL (query): error caused exit
    err: {
      "type": "TypeError",
      "message": "Cannot perform %TypedArray%.prototype.set on a detached ArrayBuffer",
      "stack":
          TypeError: Cannot perform %TypedArray%.prototype.set on a detached ArrayBuffer
              at Uint8Array.set (<anonymous>)
              at Parser.execute (/usr/src/app/node_modules/undici/lib/client.js:513:23)
              at Socket.onSocketData (/usr/src/app/node_modules/undici/lib/client.js:892:10)
              at Socket.emit (node:events:369:20)
              at addChunk (node:internal/streams/readable:313:12)
              at readableAddChunk (node:internal/streams/readable:288:9)
              at Socket.Readable.push (node:internal/streams/readable:227:10)
              at TCP.onStreamRead (node:internal/stream_base_commons:190:23)
    }
```

PR #3514, opened 2024-08-26
perf: reduce amount of used Uin8Arrays in client-h1.js

Uzlopak:
instead of instantiating an Uint8Array on every llhttp execution, we keep a view on the ArrayBuffer and reuse it.

### DIFF
--- a/lib/dispatcher/client-h1.js
+++ b/lib/dispatcher/client-h1.js
@@ -57,6 +57,7 @@ const EMPTY_BUF = Buffer.alloc(0)
 const FastBuffer = Buffer[Symbol.species]
 const addListener = util.addListener
 const removeAllListeners = util.removeAllListeners
+const utf8Slice = Buffer.prototype.utf8Slice
 
 let extractBody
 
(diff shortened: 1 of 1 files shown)


mcollina (commented):
I have a rough memory that we tried this before but we had to revert it, because there were multiple http responses being written into the same buffer in parallel, and data got corrupted.

@ronag wdyt?

lib/dispatcher/client-h1.js
```diff
@@ -230,9 +236,10 @@ class Parser {
       }
       currentBufferSize = Math.ceil(data.length / 4096) * 4096
       currentBufferPtr = llhttp.malloc(currentBufferSize)
+      currentBufferView = new Uint8Array(llhttp.memory.buffer, 0, llhttp.memory.buffer.byteLength)
```
ronag, line 239:
```suggestion
      currentBufferView = new Uint8Array(llhttp.memory.buffer)
```

ronag (approved):
I don't remember if this could cause problems. LGTM with the change i proposed.

ronag (approved):

metcoder95 (approved):

metcoder95 (approved):

ronag (changes requested):
https://github.com/nodejs/undici/pull/788

PR #3518, opened 2024-08-27
populate defaultValues in webidl dict. converter when passing null or undefined

KhafraDev:
… undefined

The test explains the commit purpose better.

`benchmarks/fetch/request-creation.mjs` was consistently 10-15% faster. Not sure how @Uzlopak @tsctx, was this a fluke?

### DIFF
--- a/lib/web/fetch/webidl.js
+++ b/lib/web/fetch/webidl.js
@@ -354,12 +354,9 @@ webidl.interfaceConverter = function (i) {
 
 webidl.dictionaryConverter = function (converters) {
   return (dictionary, prefix, argument) => {
-    const type = webidl.util.Type(dictionary)
     const dict = {}
 
-    if (type === 'Null' || type === 'Undefined') {
-      return dict
-    } else if (type !== 'Object') {
+    if (dictionary != null && typeof dictionary !== 'object') {
       throw webidl.errors.exception({
         header: prefix,
         message: `Expected ${dictionary} to be one of: Null, Undefined, Object.`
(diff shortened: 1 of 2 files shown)


lib/web/fetch/webidl.js
```diff
@@ -354,12 +354,9 @@ webidl.interfaceConverter = function (i) {
 
 webidl.dictionaryConverter = function (converters) {
   return (dictionary, prefix, argument) => {
-    const type = webidl.util.Type(dictionary)
     const dict = {}
 
-    if (type === 'Null' || type === 'Undefined') {
-      return dict
-    } else if (type !== 'Object') {
+    if (dictionary != null && typeof dictionary !== 'object') {
```
tsctx, line 359:
```suggestion
    if (dictionary != null && webidl.util.Type(dictionary) !== 'Object') {
```

The 'function' is also an 'Object' but it errors.

lib/web/fetch/webidl.js
```diff
@@ -378,13 +375,13 @@ webidl.dictionaryConverter = function (converters) {
...
+      let value = dictionary?.[key]
+      const hasDefault = defaultValue !== undefined
 
       // Only use defaultValue if value is undefined and
       // a defaultValue options was provided.
-      if (hasDefault && value !== null) {
-        value ??= defaultValue()
+      if (hasDefault && value === undefined) {
+        value = defaultValue()
       }
```
tsctx, line 385:
```suggestion
      if (hasDefault && value === undefined) {
        value = defaultValue()
      } else if (required === true && value === undefined) {
        throw webidl.errors.exception({
            header: prefix,
            message: `Missing required key "${key}".`
          })
      }
```

KhafraDev:
I'm still confused how Request creation (with no RequestInit) is 10-15% faster.

tsctx (approved):
The constructor of Request has an empty object by default, so passing undefined to RequestInit will result in default value.
The performance increase is probably due to the removal of Object.hasOwn.

KhafraDev:
I thought it could be Object.hasOwn, but the extra "overhead" from now having to loop over all of the converters made me think it would be slower. That's why I bothered benchmarking it in the first place.

PR #3521, opened 2024-08-29
set default argument values to undefined instead of {}

KhafraDev:
now that c108287b85e63bf7c5ef0613c20bd0ee40474897 has landed, this can be done. 

a micro-optimization, but a harmless change


tsctx (approved):
lgtm

PR #3525, opened 2024-08-30
ci: fix nightly workflow

Uzlopak:
## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/.github/workflows/autobahn.yml
+++ b/.github/workflows/autobahn.yml
@@ -14,7 +14,6 @@ on:
       - 'test/autobahn/**'
 permissions:
   contents: read
-  pull-requests: write
 
 jobs:
   autobahn:


mcollina (approved):
lgtm

.github/workflows/autobahn.yml
```diff
@@ -14,7 +14,6 @@ on:
       - 'test/autobahn/**'
 permissions:
   contents: read
-  pull-requests: write
```
mcollina, line 17: why does this fix the nightly workflow?

I think we should just skip autobahn for nightly instead.

Uzlopak: We needed pull-requests write permission when we created a comment in the PR. This was unreliable. So we removed the comment creation but did not remove the permission. Now in a scheduled run, where we call this workflow as reusable workflow, it can not have more perissions than the calling workflow.

We should not skip the autobahn tests. If we had them in the first place, we would have seen the issues with the regressions in node core regarding buffers earlier.

PR #3529, opened 2024-08-31
fix formdata arg validation

KhafraDev:
```js
var fd = new FormData()

fd.set('key', new Blob([]), undefined)
fd.get('key').name // should be 'blob'

fd.append('key1', new Blob([]), undefined)
fd.get('key1').name // should be 'blob'
```

I painfully found out that the wpts are not extensive

### DIFF
--- a/lib/web/fetch/formdata.js
+++ b/lib/web/fetch/formdata.js
@@ -30,21 +30,19 @@ class FormData {
     const prefix = 'FormData.append'
     webidl.argumentLengthCheck(arguments, 2, prefix)
 
-    if (arguments.length === 3 && !(value instanceof Blob)) {
-      throw new TypeError(
-        "Failed to execute 'append' on 'FormData': parameter 2 is not of type 'Blob'"
-      )
+    name = webidl.converters.USVString(name)
+
+    if (arguments.length === 3 || value instanceof Blob) {
+      value = webidl.converters.Blob(value, prefix, 'value')
+
+      if (filename !== undefined) {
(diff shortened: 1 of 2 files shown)


lib/web/fetch/formdata.js
```diff
@@ -123,25 +121,23 @@ class FormData {
...
+    name = webidl.converters.USVString(name)
+
+    if (arguments.length === 3 || value instanceof Blob) {
+      value = webidl.converters.Blob(value, prefix, 'value')
+
+      if (filename !== undefined) {
+        filename = webidl.converters.USVString(filename)
+      }
+    } else {
+      value = webidl.converters.USVString(value)
```
Uzlopak, line 133: cant this be at line 140?

KhafraDev: No. The webidl stuff occurs before the steps, but the step is also not needed because js doesn't have method overrides (hence the weird webidl here).

> The set(name, value) and set(name, blobValue, filename) method steps are:

The first step is telling us to set value to "blobValue" if there were 3 arguments passed, otherwise do nothing. "value" is only ever "value" here.

mcollina (approved):
lgtm

Uzlopak (approved):

Uzlopak:
@KhafraDev 
I think this is worth to get backported

osvein:
Is there a plan to backport this into Node.js v22 LTS? It's still not present in v6.12.3

PR #3367, opened 2024-06-24
Remove file

KhafraDev:
depends on #3366

File has been a wrapper for node's global File for a little bit, no reason to keep it.

PR #3531, opened 2024-09-01
Remove patched dom types (`v6.x` branch)

### LINKED ISSUE #3524: Unable to access base `Event` properties from `CloseEvent`, `ErrorEvent` & `MessageEvent` for WebSocket
## Bug Description

[`MessageEvent`](https://developer.mozilla.org/en-US/docs/Web/API/MessageEvent) inherits [`Event`](https://developer.mozilla.org/en-US/docs/Web/API/Event) and is supposed to provide the properties of `Event` such as `Event.type` and `Event.timeStamp` to `MessageEvent`. However, any class inheriting `Event` (`CloseEvent`, `ErrorEvent` & `MessageEvent`) fails to make any of the base properties accessible and causes the typescript compiler to fail unless `MessageEvent` type gets replaced by `any` when trying to access the base event properties.

## Reproducible By

The following code is successfully compiled by tsc compiler.

```
const ws = new WebSocket('ws://localhost:8080');

ws.onmessage = (ev) => {
  const e = ev as any;
  console.log(e.type);
  console.log(e.defaultPrevented);
  console.log(e.cancelable);
  console.log(e.timeStamp);
};
```

The following code will throw an error while trying to compile with tsc due to `Property '' does not exist on type MessageEvent<any>`.

```
const ws = new WebSocket('ws://localhost:8080');

ws.onmessage = (e) => {
  console.log(e.type);
  console.log(e.defaultPrevented);
  console.log(e.cancelable);
  console.log(e.timeStamp);
};
```

## Expected Behavior

Base properties of `Event` class accessible from inherited classes.

## Logs & Screenshots

_None_

## Environment

macOS 15.1 Beta (24B5035e), node v22.6.0, undici v6.19.8

eXhumer:
## This relates to...

Fixes #3524

## Rationale

Several missing DOM types from `@types/node` were added as part of `patch.d.ts`. `Event` & `EventTarget` exported DOM types in global were overwritten with `patch.d.ts`.

## Changes

* Remove patched `Event` & `EventTarget` type and references to it from other declaration files.
* Lock version for `@types/node` to `~18.17.19`.

### Features

N/A

### Bug Fixes

* Fixes typing bug related to DOM `Event` type.

N/A

## Status

- [x] Tested
- [S] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### Notes

This only applies to branch `v6.x`, which is the latest stable version available currently. This will also need to be ported for the `main` branch, where the major version was bumped to v7.

`@types/node` version was updated to reflect the minimum supported engine version in `package.json`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -105,7 +105,7 @@
     "@fastify/busboy": "2.1.1",
     "@matteo.collina/tspl": "^0.1.1",
     "@sinonjs/fake-timers": "^11.1.0",
-    "@types/node": "^18.0.3",
+    "@types/node": "~18.17.19",
     "abort-controller": "^3.0.0",
     "borp": "^0.15.0",
     "c8": "^10.0.0",
--- a/types/eventsource.d.ts
+++ b/types/eventsource.d.ts
@@ -2,8 +2,6 @@ import { MessageEvent, ErrorEvent } from './websocket'
 import Dispatcher from './dispatcher'
 
 import {
-  EventTarget,
-  Event,
   EventListenerOptions,
   AddEventListenerOptions,
   EventListenerOrEventListenerObject
--- a/types/filereader.d.ts
+++ b/types/filereader.d.ts
@@ -1,7 +1,7 @@
 /// <reference types="node" />
 
 import { Blob } from 'buffer'
-import { DOMException, Event, EventInit, EventTarget } from './patch'
+import { DOMException, EventInit } from './patch'
 
 export declare class FileReader {
   __proto__: EventTarget & FileReader
--- a/types/patch.d.ts
+++ b/types/patch.d.ts
@@ -6,44 +6,6 @@ export type DOMException = typeof globalThis extends { DOMException: infer T }
  ? T
  : any
 
-export type EventTarget = typeof globalThis extends { EventTarget: infer T }
-  ? T
-  : {
-    addEventListener(
-      type: string,
-      listener: any,
-      options?: any,
-    ): void
-    dispatchEvent(event: Event): boolean
-    removeEventListener(
-      type: string,
-      listener: any,
-      options?: any | boolean,
-    ): void
-  }
-
-export type Event = typeof globalThis extends { Event: infer T }
-  ? T
-  : {
-    readonly bubbles: boolean
-    cancelBubble: () => void
-    readonly cancelable: boolean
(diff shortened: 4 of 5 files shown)


package.json
```diff
@@ -105,7 +105,7 @@
     "@fastify/busboy": "2.1.1",
     "@matteo.collina/tspl": "^0.1.1",
     "@sinonjs/fake-timers": "^11.1.0",
-    "@types/node": "^18.0.3",
+    "@types/node": "~18.17.19",
```
mcollina, line 108: why?

mcollina (approved):
lgtm

package.json
eXhumer: This just provides type for node 18.17 specifically, which is the minimum required based on the engines version. The listed version `^18.0.3` doesn't contain the `Event` and `EventTarget`. And the version that ends up being installed installed typing related to node 18.19 (18.19.47) instead of the minimum supported 18.17

eXhumer: I can revert back this change if necessary, but I strongly recommend locking the `@types/node` to the minimum supported version mentioned in `engines.node` in `package.json`.

eXhumer:
Should I create a separate PR for the `main` branch or should I close this PR and create a PR targeting the `main` branch and then back port the patch with `backport.yml` workflow?

mcollina:
If possible target the main branch and we use the backport action.

mcollina:
@eXhumer PTAL, I think we'd need a manual backport after all.

eXhumer:
Correct. This requires manual backport to `v6.x` as [File API and related APIs get replaced for node's buffer module](https://github.com/nodejs/undici/pull/3367/files#diff-4f90466f7013a8e803d057b92e6403b972831a48eac2943a45996812370a6b37) for `main` branch (in development for v7). In `v6.x`, `DOMException` is used within `types/filereader.d.ts`. Also, in `v6.x` branch, we have to replace the patched `Event` & `EventTarget` within `types/filereader.d.ts` as well.

Uzlopak:
I guess #3533 must be first merged before merging this.

mcollina:
@eXhumer @Uzlopak is this ready to land?

eXhumer:
Yes, LGTM

PR #3541, opened 2024-09-03
fix: run asserts first if possible

Uzlopak:
this PR tries to run various asserts before instantiating variables. also use assert instead of assert.strictEqual because passing boolean values is cheaper than passing primitive and non-primitive values to the assert.strictEqual function.


ronag (changes requested):
Sorry. I don't think this is an improvement.

ronag (approved):

PR #3548, opened 2024-09-04
fix: fire close on fail for WebSocket

### LINKED ISSUE #3546: WebSockets do not fire 'close' event if the connection failed to be established
Consider the following code:
```js
const webSocket = new undici.WebSocket("wss://invalid-domain.example.com/");
webSocket.onopen = () => { console.log("open"); };
webSocket.onclose = () => { console.log("close"); };
webSocket.onerror = () => { console.log("error"); };
```
It outputs:
```
error
```

However, a standard-compliant WebSocket implementation would output:
```
error
close
```

This is because [establish a WebSocket connection](https://websockets.spec.whatwg.org/#concept-websocket-establish) algorithm would invoke [fail the WebSocket Connection](https://datatracker.ietf.org/doc/html/rfc6455#section-7.1.7) algorithm in step 11.1, which would [Close the WebSocket Connection](https://datatracker.ietf.org/doc/html/rfc6455#section-7.1.1). The WebSocket standard states:

> When [the WebSocket connection is closed](https://datatracker.ietf.org/doc/html/rfc6455#section-7.1.4), possibly cleanly, the user agent must [queue a task](https://html.spec.whatwg.org/multipage/webappapis.html#queue-a-task) to run the following substeps:
> ...
> 3. [Fire an event](https://dom.spec.whatwg.org/#concept-event-fire) named close at the [WebSocket](https://websockets.spec.whatwg.org/#websocket) object.

This bug is reproducible on commit [4b8958a](https://github.com/nodejs/undici/commit/4b8958a228aae55366dfba1dc70f03f08bc2d48d).

eXhumer:
## This relates to...

Fixes #3546

## Rationale

`close` event is never fired after a WebSocket connection is failed to establish.

## Changes

* change `WebSocket.#onSocketClose` when trying to access `closingInfo`. `WebSocket.#parser` is `undefined` when connection fails to establish.
* run `WebSocket.#onSocketClose` after firing `error` event, which in turn fires the `close` event.

### Features

_N/A_

### Bug Fixes

* fix `close` event not firing on failed connection.

_N/A_

## Status

- [x] Tested
- [S] Documented
- [x] Review ready
- [x] In review
- [ ] Merge ready

### DIFF
--- a/lib/web/websocket/websocket.js
+++ b/lib/web/websocket/websocket.js
@@ -531,6 +531,8 @@ class WebSocket extends EventTarget {
         message: reason
       })
     }
+
+    this.#onSocketClose()
   }
 
   #onMessage (type, data) {
@@ -666,7 +668,7 @@ class WebSocket extends EventTarget {
     let code = 1005
     let reason = ''
 
-    const result = this.#parser.closingInfo
+    const result = this.#parser?.closingInfo
 
     if (result && !result.error) {
       code = result.code ?? 1005
--- a/lib/web/websocket/websocket.js
+++ b/lib/web/websocket/websocket.js
@@ -515,7 +515,7 @@ class WebSocket extends EventTarget {
     fireEvent('open', this)
   }
 
-  #onFail (reason) {
+  #onFail (reason = '') {
     this.#controller.abort()
 
     if (this.#handler.socket && !this.#handler.socket.destroyed) {
@@ -524,15 +524,29 @@ class WebSocket extends EventTarget {
 
     this.#handler.readyState = states.CLOSED
 
-    if (reason) {
-      // TODO: process.nextTick
-      fireEvent('error', this, (type, init) => new ErrorEvent(type, init), {
-        error: new Error(reason),
-        message: reason
-      })
+    // TODO: process.nextTick
+    fireEvent('error', this, (type, init) => new ErrorEvent(type, init), {
+      error: new Error(reason),
+      message: reason
+    })
+
+    const result = this.#parser?.closingInfo
+    let code
+    if (result && !result.error) {
+      code = result.code ?? 1005
+    } else if (!this.#handler.receivedClose) {
+      // If _The WebSocket
+      // Connection is Closed_ and no Close control frame was received by the
+      // endpoint (such as could occur if the underlying transport connection
+      // is lost), _The WebSocket Connection Close Code_ is considered to be
+      // 1006.
+      code = 1006
     }
 
-    this.#onSocketClose()
+    // TODO: process.nextTick
+    fireEvent('close', this, (type, init) => new CloseEvent(type, init), {
+      wasClean: false, code, reason
+    })
   }
 
   #onMessage (type, data) {
--- a/test/websocket/issue-3546.js
+++ b/test/websocket/issue-3546.js
@@ -1,0 +1,24 @@
+'use strict'
+
+const { test } = require('node:test')
+const { WebSocket } = require('../..')
+const { tspl } = require('@matteo.collina/tspl')
+
+test('first error than close event is fired on failed connection', async (t) => {
+  const { completed, strictEqual } = tspl(t, { plan: 4 })
+  const ws = new WebSocket('ws://localhost:1')
+
+  let orderOfEvents = 0
+
+  ws.addEventListener('error', () => {
+    strictEqual(orderOfEvents++, 0)
+    strictEqual(ws.readyState, WebSocket.CLOSED)
+  })
+
+  ws.addEventListener('close', () => {
+    strictEqual(orderOfEvents++, 1)
+    strictEqual(ws.readyState, WebSocket.CLOSED)
+  })
+
+  await completed
+})


marximimus:
LGTM

KhafraDev (changes requested):
please add a test

lib/web/websocket/websocket.js
```diff
@@ -666,7 +668,7 @@ class WebSocket extends EventTarget {
     let code = 1005
     let reason = ''
 
-    const result = this.#parser.closingInfo
+    const result = this.#parser?.closingInfo
```
Uzlopak, line 671: why this change?

eXhumer: `WebSocket.#socketClose` will thrown an error trying to access properties on `undefined` as `WebSocket.#parser` is `undefined` when the connection never establishes.

Uzlopak: seems odd

eXhumer:
Some of the autobahn test keep failing locally due to tests timing out. If there is any actual issue unrelated to timeout, please let me know.

lib/web/websocket/websocket.js
eXhumer: https://github.com/nodejs/undici/blob/89a46dd54f7d7db7513c435fac8042769ee9e9b5/lib/web/websocket/websocket.js#L481-L490

`WebSocket.#parser` is set inside `WebSocket.#onConnectionEstablished`, which is never reached since an actual connection is never established.

test/websocket/issue-3546.js
```diff
@@ -0,0 +1,16 @@
...
+test('close is fired on failed connection', async (t) => {
+  const { completed, deepStrictEqual } = tspl(t, { plan: 1 })
+  const ws = new WebSocket('ws://localhost:1')
+
+  ws.addEventListener('close', () => {
+    deepStrictEqual(ws.readyState, WebSocket.CLOSED)
+  })
+
+  await completed
+})
```
Uzlopak, line 16: imho check the right order too

```suggestion
test('first error than close event is fired on failed connection', async (t) => {
  const { completed, strictEqual } = tspl(t, { plan: 4 })
  const ws = new WebSocket('ws://localhost:1')

  let orderOfEvent = 0

  ws.addEventListener('error', () => {
    strictEqual(orderOfEvent++, 0)
    strictEqual(ws.readyState, WebSocket.CLOSED)
  })

  ws.addEventListener('close', () => {
    strictEqual(orderOfEvent++, 1)
    strictEqual(ws.readyState, WebSocket.CLOSED)
  })

  await completed
})
```

Uzlopak (changes requested):
I think the test can test a little bit more

lib/web/websocket/websocket.js
Uzlopak: i checked it and it seems ok.

Uzlopak (approved):
LGTM

KhafraDev (changes requested):
a bunch of autobahn tests are failing

Uzlopak:
We can remove line 525, where we set the connection status, because we call #onSocketClose.

And the change in #onSocketClose with optional chaining brakes the autobahn. Because it changes the error code to 1006 instead of 1005.

Probably need to make #onSocketClose accept a parameter. If we pass a symbol kFail than we handle the onFail case and code stays at 1005. Or maybe use the wasClean variable?

Uzlopak (changes requested):
.

KhafraDev:
The issue is that onSocketClose is not appropriate to call here. You need to handle setting the state to CLOSED and emitting a close event in a separate function that both onFail and onSocketClose can use.

eXhumer:
Sorry for not getting to back to this earlier. After KhafkaDev said there were autobahn tests failing, I wanted to properly setup the tests locally to make sure they were working as expected. Although the test seem to be running fine on CI, it seems to fail test cases between 12.4.3 and 12.4.18 locally, while passing on CI. Does anyone have any insight on why this may be the case? And in some cases, re-running the same test locally seem to produce different fails. I have tried to do the autobahn test on two different machines (macOS 15.1 & Windows 11), both seem to fail these 16 tests locally.

```
│ 12.4.3  │ 'FAILED'        │ 'UNCLEAN'       │ 399      │ null            │
│ 12.4.4  │ 'FAILED'        │ 'UNCLEAN'       │ 135      │ null            │
│ 12.4.5  │ 'FAILED'        │ 'UNCLEAN'       │ 91       │ null            │
│ 12.4.6  │ 'FAILED'        │ 'UNCLEAN'       │ 54       │ null            │
│ 12.4.7  │ 'FAILED'        │ 'UNCLEAN'       │ 181      │ null            │
│ 12.4.8  │ 'FAILED'        │ 'UNCLEAN'       │ 351      │ null            │
│ 12.4.9  │ 'FAILED'        │ 'UNCLEAN'       │ 630      │ null            │
│ 12.4.10 │ 'FAILED'        │ 'UNCLEAN'       │ 1022     │ null            │
│ 12.4.11 │ 'FAILED'        │ 'UNCLEAN'       │ 61       │ null            │
│ 12.4.12 │ 'FAILED'        │ 'UNCLEAN'       │ 209      │ null            │
│ 12.4.13 │ 'FAILED'        │ 'UNCLEAN'       │ 419      │ null            │
│ 12.4.14 │ 'FAILED'        │ 'UNCLEAN'       │ 722      │ null            │
│ 12.4.15 │ 'FAILED'        │ 'UNCLEAN'       │ 1147     │ null            │
│ 12.4.16 │ 'FAILED'        │ 'UNCLEAN'       │ 1066     │ null            │
│ 12.4.17 │ 'FAILED'        │ 'UNCLEAN'       │ 1052     │ null            │
│ 12.4.18 │ 'FAILED'        │ 'UNCLEAN'       │ 1042     │ null            │
```

I was unable to do any testing locally as a result.

PR #3549, opened 2024-09-04
Remove unnecessary async in connectH2

DarkGL:
## Rationale

No await is used inside `connectH2`

## Changes

- Remove unnecessary async in connectH2

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -73,7 +73,7 @@ function parseH2Headers (headers) {
   return result
 }
 
-async function connectH2 (client, socket) {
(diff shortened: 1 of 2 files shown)


Uzlopak (commented):
Somehow I doubt that we should do this.

metcoder95 (commented):
To keep parity with `connectH1`, let’s still return a Promise, maybe instead of wrapping everything as Promise, a `Promise.resolve`.

But we should try to not break this parity.

PR #3550, opened 2024-09-04
fix: add jsdoc and do minor changes in utils.js

Uzlopak:
add jsdoc to the utils.js file to improve typings in other places.

also made minor changes, which i will comment

### DIFF
--- a/lib/core/util.js
+++ b/lib/core/util.js
@@ -28,6 +28,10 @@ class BodyAsyncIterable {
   }
 }
 
+/**
+ * @param {*} body
+ * @returns {*}
+ */
 function wrapRequestBody (body) {
   if (isStream(body)) {
     // TODO (fix): Provide some way for the user to cache the file to e.g. /tmp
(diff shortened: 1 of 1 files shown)


lib/core/util.js
```diff
@@ -303,8 +369,12 @@ function destroy (stream, err) {
 }
 
 const KEEPALIVE_TIMEOUT_EXPR = /timeout=(\d+)/
+/**
+ * @param {string} val
+ * @returns {number | null}
+ */
 function parseKeepAliveTimeout (val) {
-  const m = val.toString().match(KEEPALIVE_TIMEOUT_EXPR)
+  const m = val.match(KEEPALIVE_TIMEOUT_EXPR)
```
Uzlopak, line 377: parseKeepAliveTimeout gets according to my analysis always a string, so we can avoid the toString() call

lib/core/util.js
```diff
@@ -363,17 +434,24 @@ function parseHeaders (headers, obj) {
...
 }
 
+/**
+ * @param {Buffer[]} headers
+ * @returns {string[]}
+ */
 function parseRawHeaders (headers) {
-  const len = headers.length
-  const ret = new Array(len)
+  const headersLength = headers.length
```
Uzlopak, line 442: we can reuse the headersLength variable for the for loop

lib/core/util.js
```diff
@@ -594,17 +700,24 @@ const headerCharRegex = /[^\t\x20-\x7e\x80-\xff]/
...
+ * Parse accordingly to RFC 9110
+ * @see https://www.rfc-editor.org/rfc/rfc9110#field.content-range
+ * @param {string} [range]
+ * @returns
+ */
 function parseRangeHeader (range) {
   if (range == null || range === '') return { start: 0, end: null, size: null }
 
-  const m = range ? range.match(/^bytes (\d+)-(\d+)\/(\d+)?$/) : null
+  const m = range ? range.match(rangeHeaderRegex) : null
```
Uzlopak, line 720: extracted the range regex for better perf

lib/core/util.js
```diff
@@ -614,20 +727,35 @@ function parseRangeHeader (range) {
...
   return obj
 }
 
+/**
+ * @param {Record<string|symbol, any>} obj
+ */
 function removeAllListeners (obj) {
-  for (const [name, listener] of obj[kListeners] ?? []) {
-    obj.removeListener(name, listener)
+  if (obj[kListeners] != null) {
```
Uzlopak, line 746: if obj[kListeners] is not null or undefined, then we know it is an array, so only do the steps if we have somthing set.

mcollina (approved):
lgtm

metcoder95 (approved):

PR #3552, opened 2024-09-05
fix: use fasttimers for all connection timeouts

### LINKED ISSUE #3410: `UND_ERR_CONNECT_TIMEOUT` errors thrown when there is CPU intensive code on the event loop

### LINKED ISSUE #3553: TIMEOUT_IDLE should include event loop lag
Tee `TIMEOUT_IDLE` state used by the keep alive timeout should include the event loop lag since it's tracking a remote servers time.

I think the recent fast timers event loop lag fixes has broken this. We should not use fast timers for this state.

Uzlopak:
Fixes #3410
Fixes #3553

currently WIP!

Assessment:

The PR #3495 was imho the right step to fix the event loop issues. But Connection issues were still reported. Maybe the issue still existing because timeouts smaller than 1000ms will be handled by native Timers. Maybe those have to handled by the low resolution fast timer logic too, so that event loop blockers are less effecting the processing of requests.

Maybe we should increase the resolution of the timers to 200ms?! It should be still more performant than having a gazillion of native timers? 

Maybe we should change this else if to an if and maybe check if the timer missed a tick (with the side effect that we need to call `performance.now()`).
https://github.com/nodejs/undici/blob/89a46dd54f7d7db7513c435fac8042769ee9e9b5/lib/util/timers.js#L159

So if we skipped a tick, we immediatly run the timer, because we are just too late?

But on the other hand, this could actually trigger timeouts too early?!

So yeah, I would prefer to increase the resolution.


KhafraDev:
Are the fast timers even needed anymore after https://github.com/nodejs/node/pull/46579/?

ronag:
Yea. I don't think that PR makes much difference.

ronag:
> Are the fast timers even needed anymore after [nodejs/node#46579](https://github.com/nodejs/node/pull/46579)?

Fast timers also adjust for event loop lag.

Uzlopak:
Ok, then i change this PR to rip out fast timers

ronag:
No, don't remove fast timers.

ronag (changes requested):
Do not remove fast timers.

Uzlopak:
@ronag 

Oh! good that i had a backup :). reverted

lib/util/timers.js
```diff
@@ -362,6 +362,31 @@ module.exports = {
...
+   * expires.
+   * @param {number} delay The time, in milliseconds that the timer should
+   * wait before the specified function or code is executed.
+   * @param {*} [arg] An optional argument to be passed to the callback function
+   * when the timer expires.
+   * @returns {FastTimer}
+   */
+  setFastTimeout (callback, delay, arg) {
+    // If the delay is less than or equal to the RESOLUTION_MS value return a
+    // native Node.js Timer instance.
```
Uzlopak, line 378:
```suggestion
```

ronag (approved):

Uzlopak:
@ronag 

but one thing maybe for interest

The test run for ripped out fast timers passes the tests

https://github.com/nodejs/undici/actions/runs/10714949877

This indicates that event loop lags are also handled properly by native timers.

But I guess we want to keep our fast timers solution due to the fact, that it is still less resource hungry.

ronag:
> This indicates that event loop lags are also handled properly by native timers.

This can't be true as it's based on the uv loop....

metcoder95 (approved):

Uzlopak:
let me fix the request-timeout tests.

Uzlopak:
@ronag 

Should we also use setFastTimeout here?

https://github.com/nodejs/undici/blob/61a4b699c9f851d8c3d5b2a6f51f8c477249b921/lib/core/connect.js#L169

ronag:
Yes

Uzlopak:
adapted so far, but not all tests are reactivated. so please not merge yet

lib/core/connect.js
```diff
@@ -116,11 +120,14 @@ function buildConnector ({ allowH2, maxCachedSessions, socketPath, timeout, sess
         })
     } else {
       assert(!httpSocket, 'httpSocket can only be sent on TLS update')
+
+      port = port || 80
```
ronag, line 124: why?

lib/core/connect.js
```diff
@@ -96,6 +98,8 @@ function buildConnector ({ allowH2, maxCachedSessions, socketPath, timeout, sess
 
       const session = customSession || sessionCache.get(sessionKey) || null
 
+      port = port || 443
```
ronag, line 101: why?

Uzlopak: I want to see where the socket error is originating.

Uzlopak:
@ronag 

Because you asked, why i change the port. The thing is, that I want to use the port for the connection timeout to have in the logs the chance to determine atleast which server/connection resulted in a connection timeout.

ronag (approved):

trivikr:
The backport fails to 6.x, as the refactor of fast timers in https://github.com/nodejs/undici/pull/3495 is not backported to 6.x

PR #3554, opened 2024-09-05
fix: use native timeouts for TIMEOUT_IDLE, rename TIMEOUT_IDLE to TIMEOUT_KEEP_ALIVE

### LINKED ISSUE #3553: TIMEOUT_IDLE should include event loop lag

Uzlopak:
Fixes #3553 

This PR targets the fast-timer-connect branch of PR #3552 

@ronag 

Is this to your liking?

### DIFF
--- a/lib/dispatcher/client-h1.js
+++ b/lib/dispatcher/client-h1.js
@@ -176,7 +176,7 @@ class Parser {
     // timer and set a new one.
     if (
       delay !== this.timeoutValue ||
-      (type & USE_FAST_TIMER) !== (this.timeoutType & USE_FAST_TIMER)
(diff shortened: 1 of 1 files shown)


ronag (approved):

Uzlopak:
@ronag 

PTAL

I think now it is ensured that  that fast timers and native timers are set properly.

Uzlopak:
Will merge it and make it part of #3552

PR #3560, opened 2024-09-06
feat: implement WebSocketStream

KhafraDev:
With better-organized internals, this is now much easier and almost completely detached from WebSocket. I could rewrite it to be completely independent of WebSocket, which would allow us to remove it at any time (the only shared internals are a new `getURLRecord` function and the closing logic).

```mjs
const stream = new WebSocketStream('https://echo.websocket.org/')
const { readable, writable } = await stream.opened

;(async () => {
  /** @type {ReadableStreamReader} */

  while (true) {
    const { done, value } = await reader.read()
    if (done) break
    console.log('received message', { value })
  }
})();

setInterval(() => {
  console.log('sending message.')
  /** @type {WritableStreamDefaultWriter} */
  const writer = writable.getWriter()

  writer.write('Hello, world!')

  writer.releaseLock()
}, 5000)
```


KhafraDev:
@nodejs/undici This passes the same subset of WebSocketStream tests that Deno passes. It's missing types and docs now. Looking for code reviews before doing the trivial stuff.

lib/web/websocket/connection.js
```diff
@@ -214,13 +216,80 @@ function establishWebSocketConnection (url, protocols, client, handler, options)
...
 
 /**
- * @param {import('./websocket').Handler} handler
- * @param {number} code
- * @param {any} reason
- * @param {number} reasonByteLength
+ * @see https://whatpr.org/websockets/48.html#close-the-websocket
+ * @param {import('./websocket').Handler} object
+ * @param {number} [code]
+ * @param {string} [reason]
```
Uzlopak, line 222:
```suggestion
 * @param {string} [reason='']
 * @param {boolean} [validate=false]
```

lib/web/websocket/connection.js
```diff
@@ -214,13 +216,80 @@ function establishWebSocketConnection (url, protocols, client, handler, options)
 }
 
 /**
- * @param {import('./websocket').Handler} handler
- * @param {number} code
- * @param {any} reason
- * @param {number} reasonByteLength
+ * @see https://whatpr.org/websockets/48.html#close-the-websocket
+ * @param {import('./websocket').Handler} object
+ * @param {number} [code]
```
Uzlopak, line 221:
```suggestion
 * @param {number} [code=null]
```

lib/web/websocket/connection.js
```diff
@@ -1,12 +1,14 @@
...
-const { failWebsocketConnection, parseExtensions } = require('./util')
+const { uid, states, sentCloseFrameState, emptyBuffer, opcodes } = require('./constants')
+const { failWebsocketConnection, parseExtensions, isClosed, isClosing, isEstablished, validateCloseCodeAndReason } = require('./util')
 const { channels } = require('../../core/diagnostics')
 const { makeRequest } = require('../fetch/request')
 const { fetching } = require('../fetch/index')
 const { Headers, getHeadersList } = require('../fetch/headers')
 const { getDecodeSplit } = require('../fetch/util')
+const { WebsocketFrameSend } = require('./frame')
+const assert = require('node:assert')
```
Uzlopak, line 11: nit: can you move this require to the top please?

lib/web/websocket/connection.js
```diff
@@ -214,13 +216,80 @@ function establishWebSocketConnection (url, protocols, client, handler, options)
...
+    // WebSocket Close message must be the integer given by code.
+    // If code is null and reason is the empty string, the WebSocket Close frame must not have a body.
+    // If reason is non-empty but code is null, then set code to 1000 ("Normal Closure").
+    if (reason.length !== 0 && code === null) {
+      code = 1000
+    }
+
+    // If code is set, then the status code to use in the WebSocket Close frame must be the integer given by code.
+    assert(code === null || Number.isInteger(code))
+
```
Uzlopak, line 264: why assert if we had before set code to 1000?

```suggestion
    } else {
      // If code is set, then the status code to use in the WebSocket Close frame must be the integer given by code.
      assert(code === null || Number.isInteger(code))
    }
```

lib/web/websocket/stream/websocketerror.js
```diff
@@ -0,0 +1,69 @@
...
+class WebSocketError extends DOMException {
+  #closeCode
+  #reason
+
+  constructor (message = '', init = undefined) {
+    message = webidl.converters.DOMString(message, 'WebSocketError', 'message')
+
+    // 1. Set this 's name to " WebSocketError ".
+    // 2. Set this 's message to message .
+    super(message, 'WebSocketError')
```
Uzlopak, line 16: The message attribute of DomException is not enumerable or so

image

I think we want to have that field to be serializable?

```suggestion
    super(message, 'WebSocketError')
    this.message = message
```

lib/web/websocket/stream/websocketerror.js
```diff
@@ -0,0 +1,69 @@
...
+  get reason () {
+    return this.#reason
+  }
+
+  /**
+   * @param {string} message
+   * @param {number|null} code
+   * @param {string} reason
+   */
+  static createUnvalidatedWebSocketError (message, code, reason) {
```
Uzlopak, line 58: unvalidated? do you mean invalidated?

lib/web/websocket/stream/websocketerror.js
```diff
@@ -0,0 +1,69 @@
...
+  }
+
+  get reason () {
+    return this.#reason
+  }
+
+  /**
+   * @param {string} message
+   * @param {number|null} code
+   * @param {string} reason
```
Uzlopak, line 56:
```suggestion
   * @param {string} reason
   * @returns {WebSocketError}
```

lib/web/websocket/util.js
```diff
@@ -247,6 +247,68 @@ function isValidClientWindowBits (value) {
   return true
 }
 
+/**
+ * @see https://whatpr.org/websockets/48/7b748d3...d5570f3.html#get-a-url-record
+ * @param {string} url
+ * @param {string|undefined} baseURL
```
Uzlopak, line 253:
```suggestion
 * @param {string} [baseURL]
 * @returns {URL}
```

lib/web/websocket/util.js
```diff
@@ -247,6 +247,68 @@ function isValidClientWindowBits (value) {
...
+  // If urlRecord ’s fragment is non-null, then throw a " SyntaxError " DOMException .
+  if (urlRecord.hash.length || urlRecord.href.endsWith('#')) {
+    throw new DOMException('hash', 'SyntaxError')
+  }
+
+  // Return urlRecord .
+  return urlRecord
+}
+
+// https://whatpr.org/websockets/48.html#validate-close-code-and-reason
```
Uzlopak, line 288:
```suggestion

/**
 * @see https://whatpr.org/websockets/48.html#validate-close-code-and-reason
 * @param {number | null} code
 * @param {string | null} reason
 * @throws {DOMException}
 */
```

lib/web/websocket/stream/websocketstream.js
```diff
@@ -0,0 +1,494 @@
...
+      let string
+
+      try {
+        string = webidl.converters.USVString(chunk)
+      } catch (e) {
+        return Promise.reject(e)
+      }
+
+      // 5.2. Set data to the result of UTF-8 encoding string .
+      data = new TextEncoder().encode(string)
```
Uzlopak, line 234: Can we have one global TextEncoder instead of instantiating potentially one on each write?

lib/web/websocket/stream/websocketstream.js
```diff
@@ -0,0 +1,494 @@
...
+      // 5. Otherwise,
+
+      // 5.1. Let string be the result of converting chunk to an IDL USVString .
+      //    If this throws an exception, return a promise rejected with the exception.
+      let string
+
+      try {
+        string = webidl.converters.USVString(chunk)
+      } catch (e) {
+        return Promise.reject(e)
```
Uzlopak, line 230: couldnt we reuse the already created deferred promise?

`return promise.reject(e)`
?

lib/web/websocket/stream/websocketstream.js
```diff
@@ -0,0 +1,494 @@
...
+  }
+
+  // The close( closeInfo ) method steps are:
+  close (closeInfo = undefined) {
+    if (closeInfo !== null) {
+      closeInfo = webidl.converters.WebSocketCloseInfo(closeInfo)
+    }
+
+    // 1. Let code be closeInfo [" closeCode "] if present, or null otherwise.
+    const code = closeInfo.closeCode ?? null
```
Uzlopak, line 194: This looks strange

Maybe `closeInfo != null` to handle also undefined cases?

lib/web/websocket/stream/websocketstream.js
```diff
@@ -0,0 +1,494 @@
...
+    // 2. Let data be null.
+    let data = null
+
+    // 3. Let opcode be null.
+    let opcode = null
+
+    // 4. If chunk is a BufferSource ,
+    if (ArrayBuffer.isView(chunk) || types.isAnyArrayBuffer(chunk)) {
+      // 4.1. Set data to a copy of the bytes given chunk .
+      data = Buffer.from(chunk)
```
tsctx, line 216:
```suggestion
      data = new Uint8Array(ArrayBuffer.isView(chunk) ? new Uint8Array(chunk.buffer, chunk.byteOffset, chunk.byteLength) : chunk)
```

Handling of non-Uint8Array

lib/web/websocket/stream/websocketerror.js
Uzlopak: Ah you mean an error which is not validated.

lib/web/websocket/stream/websocketstream.js
```diff
@@ -0,0 +1,494 @@
...
+    // 6. In parallel,
+    // 6.1. Wait until there is sufficient buffer space in stream to send the message.
+
+    // 6.2. If the closing handshake has not yet started , Send a WebSocket Message to stream comprised of data using opcode .
+    if (this.#handler.closeState === sentCloseFrameState.NOT_SENT) {
+      const frame = new WebsocketFrameSend(data)
+
+      this.#handler.socket.write(frame.createFrame(opcode), () => {
+        promise.resolve(undefined)
+      })
```
tsctx, line 249: Why don't you use SendQueue?

lib/web/websocket/connection.js
KhafraDev: we only set it to 1000 if code is null and reason is non-empty

lib/web/websocket/stream/websocketerror.js
KhafraDev: then we wouldn't be extending DOMException entirely

lib/web/websocket/stream/websocketstream.js
```diff
@@ -0,0 +1,494 @@
...
+    const promise = createDeferredPromise()
+
+    // 2. Let data be null.
+    let data = null
+
+    // 3. Let opcode be null.
+    let opcode = null
+
+    // 4. If chunk is a BufferSource ,
+    if (ArrayBuffer.isView(chunk) || types.isAnyArrayBuffer(chunk)) {
```
tsctx, line 214:
```suggestion
    if (ArrayBuffer.isView(chunk) || types.isArrayBuffer(chunk)) {
```

omit SharedArrayBuffer

lib/web/websocket/stream/websocketstream.js
KhafraDev: No need to, you can only write BufferSources and strings

Uzlopak:
I wanted to test the WebSocketError thing i mentioned above and found this.

```js
const { WebSocketError } = require('./lib/web/websocket/stream/websocketerror');

console.log(Object.keys(new WebSocketError('asd')));
```

It is not possible to instantiate WebSocketError by requiring the file alone.

image

lib/web/websocket/stream/websocketerror.js
Uzlopak: Ok, seems to be consistent with chrome

lib/web/websocket/stream/websocketstream.js
KhafraDev: https://github.com/nodejs/node/blob/c3a7b29e56a5ada6327ebb622ba746d022685742/lib/internal/encoding.js#L339-L342

lib/web/websocket/stream/websocketerror.js
KhafraDev: lsp can infer the return type

KhafraDev:
@tsctx @Uzlopak PTAL.

KhafraDev:
> I wanted to test the WebSocketError thing i mentioned above and found this.

A limitation of cjs - there isn't much we can do about shared webidl declarations. If we moved them to webidl.js, we'd have circular imports, if we move them to another file, we have to import another file. You can `const { WebSocketError } = require('.')` instead.

lib/web/websocket/stream/websocketstream.js
```diff
@@ -0,0 +1,511 @@
...
+
+    if (type === opcodes.TEXT) {
+      try {
+        chunk = utf8Decode(data)
+      } catch {
+        failWebsocketConnection(this.#handler, 'Received invalid UTF-8 in text frame.')
+        return
+      }
+    } else if (type === opcodes.BINARY) {
+      chunk = new Uint8Array(data.buffer, data.byteOffset, data.byteLength)
```
tsctx, line 346:
```suggestion
      chunk = data.byteLength === data.buffer.byteLength ? new Uint8Array(data.buffer, data.byteOffset, data.byteLength) : new Uint8Array(data)
```

Copy buffer if needed.

tsctx (changes requested):

KhafraDev:
please approve so I can land... nits can wait, this is experimental

I want to make websocket PRs but I don't want to manage multiple branches at once. ty

anonrig (approved):

PR #3565, opened 2024-09-07
handle websocket closed correctly

KhafraDev:
Fixes #3546 
Closes #3548

PR #3507, opened 2024-08-25
websocket: set websocket readyState on fail

KhafraDev:
Fixes https://github.com/nodejs/undici/issues/3506

## Status

- [ ] I have read and agreed to the [Developer's Certificate of Origin][cert]
- [ ] Tested
- [ ] Benchmarked (**optional**)
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

[cert]: https://github.com/nodejs/undici/blob/main/CONTRIBUTING.md

PR #3566, opened 2024-09-08
fix: fire `close` on failed WebSocket connection

### LINKED ISSUE #3546: WebSockets do not fire 'close' event if the connection failed to be established

eXhumer:
## This relates to...

See nodejs/undici#3546, nodejs/undici#3548 & nodejs/undici#3565

CC: @Uzlopak @KhafraDev

## Rationale

#3565 attempted to fix #3546, however the test for #3546 made in #3548 was modified in #3565 and [skipped the `close` event check](https://github.com/nodejs/undici/pull/3565/files#r1749099808), which ultimately ended up with not actually fixing the original issue. This PR will check the `close` event correctly and correctly fixes the original issue.

Additionally, [this commit has been already tested to make sure it passes the autobahn testsuite](https://github.com/eXhumer/undici/actions/runs/10758006235/job/29832745599)

## Changes

* correctly fire `close` event if there is no successful connection to server.
* fix test to actually check the `close` event correctly firing.

### Features

_N/A_

### Bug Fixes

* correctly fire `close` event on unsuccessful connection

_N/A_

## Status

- [x] Tested
- [S] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/web/websocket/websocket.js
+++ b/lib/web/websocket/websocket.js
@@ -537,6 +537,14 @@ class WebSocket extends EventTarget {
         message: reason
       })
     }
+
+    if (!this.#parser && !this.#handler.receivedClose) {
+      fireEvent('close', this, (type, init) => new CloseEvent(type, init), {
+        wasClean: false,
+        code: 1006,
+        reason
+      })
+    }
   }
 
   #onMessage (type, data) {
--- a/test/websocket/issue-3546.js
+++ b/test/websocket/issue-3546.js
@@ -5,7 +5,7 @@ const { WebSocket } = require('../..')
 const { tspl } = require('@matteo.collina/tspl')
 
 test('first error than close event is fired on failed connection', async (t) => {
-  const { completed, strictEqual } = tspl(t, { plan: 2 })
+  const { completed, strictEqual } = tspl(t, { plan: 4 })
   const ws = new WebSocket('ws://localhost:1')
 
   let orderOfEvents = 0


eXhumer:
Not sure why the `Node CI / test (20, macos-latest)` test failed, but it should work on re-run. [Had a similar issue before](https://github.com/nodejs/undici/actions/runs/10652241249/job/29526382309).

eXhumer:
Some details on what is actually the issue.

The issue revolves around why the event is not being fired on unsuccessful connection. It is because the function the does the event firing is linked to an object that does not exist before a connection is actually successfully established. `WebSocket.#parser` is only initialized in `WebSocket#onConnectionEstablished`.

https://github.com/nodejs/undici/blob/6818acc06830a3119720bc8e97ae286ee0ebebb1/lib/web/websocket/websocket.js#L481-L490

eXhumer:
Thoughts? Reviews? Any change requested?

The current implementation of WebSocket is not firing the `close` event at all on connection establish fail in any branch (this PR attempts to fix that case in `main`) and the `error` event is not being fired correctly on connection establish fail with the `WebSocket.readyState` being set to [`WebSocket.CONNECTING`](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/readyState#websocket.connecting) instead of [`WebSocket.CLOSED`](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/readyState#websocket.closed) in latest release (`v6.x` branch) (fixed in `main` with #3507, not backported yet to `v6.x`). Currently implemented test at `test/websocket/issue-3546.js` bypasses the `close` event related checks added with #3565 by setting the number of plans to 2 instead of 4.

https://github.com/nodejs/undici/blob/990df2c7e37cbe5bb44fe2f576dddeaeb5916590/test/websocket/issue-3546.js#L8

KhafraDev:
I do not like this patch but I haven't had time to look into a real fix.
- The condition is too "smart" by relying on the order of parameters being initialized. Why can't it rely on handler.readyState being OPEN?
- It relies on the connection being failed, but a websocket can be closed without it failing.

eXhumer:
> * The condition is too "smart" by relying on the order of parameters being initialized. Why can't it rely on handler.readyState being OPEN?

AKA why not `!isEstablished(this.#handler.readyState)` instead.

It is because it fails autobahn test suite, see https://github.com/eXhumer/undici/actions/runs/10837039351/job/30072121866

> * It relies on the connection being failed, but a websocket can be closed without it failing.

Correct, but in a case where connection is closed without failing, it is instead handled by `WebSocket.#onSocketClose` instead, **which is only executed when a connection is actually successfully established and `WebSocket.#parser` has been initialized**

eXhumer:
Currently, `close` event is only fired if `WebSocket.#onSocketClose` is executed.

And `handler.onSocketClose` (which then executes `WebSocket.#onSocketClose`) is attached to the socket only if a connection has been successfully established as the lines before are never reached if a connection is failed due to protocol error or network error.

https://github.com/nodejs/undici/blob/618005eb539a410ef7e855c78cbbecbe916a4234/lib/web/websocket/connection.js#L197-L199

Also, should the close code be 1002 if there is a connection establish error? Should it not be 1006 instead as the connection is not considered clean and a close frame is never received?

Uzlopak:
I atleast reopened the issue.

Uzlopak (approved):
LGTM

Issue #2760, opened 2024-02-14
fetch: caching


Ethan-Arrowood:
Based on the numerous `todo: cache` comments (such as: https://github.com/nodejs/undici/blob/eb884c5f792345e4f13d0e555a68ebb06020b796/lib/fetch/index.js#L1394) I assume our fetch implementation does not support any kind of caching? 

I was reading this mdn documentation: https://developer.mozilla.org/en-US/docs/Web/API/Request/cache and trying to demo the default cache behavior with a test script like:

```js
import http from 'node:http';
import events from 'node:events';

const server = http.createServer((req, res) => {
  console.log('hello from server');
  res.writeHead(200, { 'Content-Type': 'application/json', 'Cache-Control': 'max-age=604800' });
  res.end(JSON.stringify({
    data: 'Hello, World!'
  }));
});

server.listen();

await events.once(server, 'listening');

const address = server.address();

console.log(address);

const r = await fetch(`http://localhost:${address.port}`);
await r.json();

console.log(`status ${r.status}`)

const r2 = await fetch(`http://localhost:${address.port}`);
await r2.json();

console.log(`status ${r2.status}`)

server.close();

```

But this logs `hello from server` twice.

Ethan-Arrowood:
Related: https://github.com/nodejs/undici/issues/1146

KhafraDev:
this is already possible with CacheStorage, it's just not automatic (and it *is* an in-memory cache right now).

adapted from mdn:

```js
import { caches, Request, fetch } from './index.js'

async function cacheFetch (url, init) {
  const request = new Request(url, init)

  return caches.match(request).then((response) => {
    if (response !== undefined) {
      return response
    } else {
      return fetch(request)
        .then((response) => {
          // response may be used only once
          // we need to save clone to put one copy in cache
          // and serve second one
          const responseClone = response.clone()

          caches.open('v1').then((cache) => {
            cache.put(request, responseClone)
          })

          return response
        })
        .catch(() => /* handle error case */ {})
    }
  })
}

```

KhafraDev:
I don't think automatic caching is a good idea, it's possible via the dispatcher API in a way that is extendable and configurable (you can use whatever storage method you desire). From the issue linked, there are already quite a few ideas on how to handle caching. I don't think it's something we can do in a way that would be acceptable for most people.

KhafraDev:
We are spec compliant, caching is completely optional. `httpCache` can be null and therefore caching is ignored:

[...]

PR #3569, opened 2024-09-09
fix(types): add missing `cache` prop to RequestInit

rindeal:
Also sort the props.

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [x] In review
- [ ] Merge ready

### DIFF
--- a/test/types/fetch.test-d.ts
+++ b/test/types/fetch.test-d.ts
@@ -38,17 +38,18 @@ declare const request: Request
 declare const headers: Headers
 declare const response: Response
 
-expectType<string | undefined>(requestInit.method)
-expectType<boolean | undefined>(requestInit.keepalive)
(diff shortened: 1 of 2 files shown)


eXhumer:
Lint issues, 4 spaces instead of 2.

I don't think `undici.fetch` supports caching yet (See #2760).

If you want to implement some sort of in-memory caching, see https://github.com/nodejs/undici/issues/2760#issuecomment-1944119404.

mcollina (changes requested):
Can you add some type tests for this? We use `tsd`.

Uzlopak (changes requested):
And tomorrow somebody argues, that because the cache field is in the types, we need to implement asap caching.

Uzlopak (approved):
@KhafraDev 

image

mcollina (approved):
lgtm

PR #3578, opened 2024-09-10
chore: remove lib/api/util.js

Uzlopak:
This PR adds jdoc to lib/api/util.js... but lib/api/util.js seem to be not be used anymore. dead code.

Any idea where it was used originally?


Ethan-Arrowood:
I recommend checking git blame to see when it was introduced and what it was used for. But anyways, if the code is actually dead and unused, feel free to just delete! It'll always exist in git history if anyone wants to refer back to it in the future.

Uzlopak:
Yeah, i know that ;). I just thought that maybe somebody comes with a super surprising reason why we should keep that file.

Removing this file.
Can you please review?

@Ethan-Arrowood

Ethan-Arrowood (approved):

PR #3579, opened 2024-09-10
Generate types dynamically

Uzlopak:
POC


KhafraDev:
I don't think this is a good direction to go in.

Uzlopak:
As long as typescript doesnt allow an export of typedefs it doesnt make any sense to try to autogenerate types from jsdoc

PR #3586, opened 2024-09-12
chore: remove pluralizer

Uzlopak:
We dont use the pluralizer that excessive, even though it is in the codebase for more than 2 years. I think this means that it did not had the desired impact.


mcollina (approved):
lgtm

PR #3600, opened 2024-09-13
diagnostics-channel: use not deprecated subscribe fn

Uzlopak:
In engines we declare that we support node >= 18.17. and the new subscribe notation is existitng since 18.7. 

https://nodejs.org/api/diagnostics_channel.html#diagnostics_channelsubscribename-onmessage


mcollina (approved):
lgtm

Issue #3486, opened 2024-08-19
socketPath option is not passed to the connect function


zdm:
## Bug Description

In `Pool` dispatcher `socketPath` option is not passed to the `connect` if `connect` is a function.

## Reproducible By

```
const dispatcher = new Agent ( {
    socketPath: "/var/run/socket",
    connect ( options, callback ) {
        console.log( options );
        process.exit();
    }
} )

await fetch( "http://local/", {
    dispatcher
} );
```

## Expected Behavior

Expecting to see `socketPath` in options, passed to the `connect` function.

When `connect` is specified as object - it works, but for function you just forget to pass `socketPath`.

metcoder95:
Good catch, would you like to send a PR?

Uzlopak:
@metcoder95 
Actually i doubt that this is a bug. If I read the code it looks like it was on purpose not allowed to manipulate the socketPath.

zdm:
Hmm, bit if we specify socketPath, connect function should receive if, or how it should work?

metcoder95:
Well, in theory it should, we are just not passing it to the `connect` factory .

We have documented that we support that to be passed down to the [connector factory](https://undici.nodejs.org/#/docs/api/Client?id=parameter-connectoptions). I'd also expect that they are passed down to the custom factory that its submitted.

zdm:
Ok. what is your decision?
I cam make PR it this is a bug.

zdm:
I think it should be fixed.
I user do not want to change socketPath in the future - he can use closure, but by default this parameter should pass.

mcollina:
I think a PR with the change would help us spot why we did not pass it. If it's that's the case there is likely a test to cover it.

metcoder95:
+1

Uzlopak:
What happens if you specify some file like `/etc/passwd` as socketPath?

zdm:
> What happens if you specify some file like `/etc/passwd` as socketPath?

Error, because this is not a socket.

zdm:
So, I carefully looked at the code.
`sockerPath` should behave like `localAddress` option.
It should be removed from `buildConnector` options and added to the `connect` call options.
This will be a breaking change because type script interfaces will be changed.

zdm:
`timeout` also not passed to `connect` function.
I think all options, which are passed to the `buildConnector` should be passed to the `connect` function without exclusions.

metcoder95:
[...]

PR #3602, opened 2024-09-13
feat: dispatcher _buildConnector() method

zdm:
## This relates to...

https://github.com/nodejs/undici/issues/3486

## Rationale

This patch allows to access buildConnector options in the DispatcherBase subclasses.
This will allows to create custom `connect` function in subclasses using `buiildConnector` options.

## Changes

1. `_buildConnector()` protected method added to `DispatcherBase`.
2. `buildConnector` calls changed from direct use to use `this._buildConnector()` method.
3. helper function `configureSocket()` added. It sets socket `no delay` and `keep alive` based on passed options.

### Features

It is too hard to make `connect` callback receive full list of connect options. 
At least I have no idea how to implement it on current code base without deep refactoring.
This patch helps to solve this problem in dispatcher subclasses, that will be enough for most of developers.

Use case:
```
import Client from 'undici/dispatcher/client'
import { configureSocket } from 'undici/core/connect'

export default class extends Client {

  _buildConnector (buildOptions) {

    // create custom connector
    if (this.proxy) {
      return async (options, callback) => {
        const socket = await this._createSocket(buildOptions, options)

        callback(configureSocket(socket))
      }

    // or fallback to the default behavior
    } else {
      return super._buildConnector(buildOptions)
    }
  }
}
```

### Bug Fixes

https://github.com/nodejs/undici/issues/3486
## Status

- [x ] I have read and agreed to the [Developer's Certificate of Origin][cert]
- [x] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/core/connect.js
+++ b/lib/core/connect.js
@@ -132,15 +132,9 @@ function buildConnector ({ allowH2, maxCachedSessions, socketPath, timeout, sess
       })
     }
 
-    // Set TCP keep alive options on the socket here instead of in connect() for the case of assigning the socket
-    if (options.keepAlive == null || options.keepAlive) {
-      const keepAliveInitialDelay = options.keepAliveInitialDelay === undefined ? 60e3 : options.keepAliveInitialDelay
-      socket.setKeepAlive(true, keepAliveInitialDelay)
-    }
-
     const clearConnectTimeout = setupConnectTimeout(new WeakRef(socket), { timeout, hostname, port })
 
-    socket
+    configureSocket(socket, options)
       .setNoDelay(true)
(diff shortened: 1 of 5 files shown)


mcollina:
why is this needed?

zdm:
This is needed if I want to create custom dispatcher and get access to buildConnector options (with resolved default values), please look at related issue.

ronag (changes requested):
I don't think this is a good idea...

zdm:
Why nor? Do you have arguments?
"buildConnector" is used in subclasses constructor, why not provide this standard functionality as private method?
I think this is obvious

PR #3603, opened 2024-09-13
fetch: make fullyReadBody sync

Uzlopak:
closer to the spec and avoiding an async call.

### DIFF
--- a/lib/web/fetch/util.js
+++ b/lib/web/fetch/util.js
@@ -1029,7 +1029,7 @@ function iteratorMixin (name, object, kInternalIterator, keyIndex = 0, valueInde
 /**
  * @see https://fetch.spec.whatwg.org/#body-fully-read
  */
-async function fullyReadBody (body, processBody, processBodyError) {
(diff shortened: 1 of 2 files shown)


lib/web/fetch/util.js
```diff
@@ -1096,30 +1092,39 @@ function isomorphicEncode (input) {
...
-async function readAllBytes (reader) {
+async function readAllBytes (reader, successSteps, failureSteps) {
   const bytes = []
   let byteLength = 0
 
-  while (true) {
-    const { done, value: chunk } = await reader.read()
+  try {
+    do {
+      const { done, value: chunk } = await reader.read()
```
Uzlopak, line 1104: can reader.read() throw?

KhafraDev: yes

Uzlopak: thx 

than putting it into a try catch block makes sense.

mcollina (approved):
lgtm

metcoder95 (approved):

PR #3605, opened 2024-09-13
fetch: avoid async function in mainFetch to generate response

Uzlopak:
We can avoid some promises here.


mcollina (approved):
lgtm

mcollina:
@KhafraDev could you take a look?

PR #3606, opened 2024-09-13
fetch: make consumeBody sync

Uzlopak:
consumeBody does not need to be async.

### DIFF
--- a/lib/web/fetch/body.js
+++ b/lib/web/fetch/body.js
@@ -298,12 +298,6 @@ function cloneBody (instance, body) {
   }
 }
 
-function throwIfAborted (state) {
-  if (state.aborted) {
-    throw new DOMException('The operation was aborted.', 'AbortError')
-  }
-}
-
 function bodyMixinMethods (instance) {
   const methods = {
     blob () {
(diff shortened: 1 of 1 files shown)


lib/web/fetch/body.js
```diff
@@ -424,16 +418,18 @@ function mixinBody (prototype) {
...
   //    with a TypeError.
   if (bodyUnusable(object)) {
-    throw new TypeError('Body is unusable: Body has already been read')
+    return Promise.reject(new TypeError('Body is unusable: Body has already been read'))
   }
 
-  throwIfAborted(object[kState])
+  if (object[kState].aborted) {
+    return Promise.reject(new DOMException('The operation was aborted.', 'AbortError'))
+  }
```
Uzlopak, line 432: Actually I thin aborted is a subtype of the object being unusable. Does it have to be a DOMException?

KhafraDev (changes requested):
this type of change makes maintaining these sections annoying and prone to errors

Uzlopak:
Despite saving one promise i also dislike the try catch. Retractng.

PR #3608, opened 2024-09-14
fetch: optimize .bytes() mixin

Uzlopak:
looks faster... :-P

Needs a Benchmark like the one for arraybuffers made by @tsctx

### DIFF
--- a/lib/web/fetch/body.js
+++ b/lib/web/fetch/body.js
@@ -305,6 +305,8 @@ function throwIfAborted (state) {
 }
 
 function bodyMixinMethods (instance) {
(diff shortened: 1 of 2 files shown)


tsctx:
I think it will be slower.
*`Uint8Array.from`* -> To create a new buffer, loop through %Symbol.iterator%, set it in the buffer. [spec](https://tc39.es/ecma262/multipage/indexed-collections.html#sec-%typedarray%.from)
*`new Uint8Array`* -> Using an internal slot. [spec](https://tc39.es/ecma262/multipage/indexed-collections.html#sec-typedarray)

Uzlopak:
no perf differnce

PR #3609, opened 2024-09-14
fetch: optimize .arrayBuffer() mixin

Uzlopak:
avoid instantiating an Uint8Array just to get the arraybuffer by directly slicing the necessary arrayBuffer.


Uzlopak:
The Benchmark by @tsctx does not show an improvement. Hmm

Uzlopak:
no perf difference

PR #3620, opened 2024-09-18
fix: dont crash on invalid brotli payload

### LINKED ISSUE #3616: Empty fetch() response with 'Content-Encoding: br' header causes a crash due to unhandled error event in BrotliDecompress
### Version

22.6.0, 20.9.0, 20.17.0

### Platform

_No response_

### Subsystem

_No response_

An empty response with `Content-Encoding: br` results in an error if the response is canceled, not consumed, consumed by a stream. 

This can be reproduced with a server:
```

createServer((req, res) => {
  res.writeHead(200, {
    "Content-Length": "0",
    "Connection": "close",
    "Content-Encoding": "br"
  });
  res.end();
}).listen(3010);
```

and then a client calling:
```
async function main() {
  const url = "http://localhost:3010/"
  try {
    const r = await fetch(url);
    console.log("done");
  } catch (e) {
    console.log(e);
  }
}

main();
```

Although `done` will be printed, an exception will be thrown afterwards:
```
done
node:events:498
      ^

Error: unexpected end of file
    at genericNodeError (node:internal/errors:983:15)
    at wrappedFn (node:internal/errors:537:14)
    at BrotliDecoder.zlibOnError [as onerror] (node:zlib:191:17)
Emitted 'error' event on BrotliDecompress instance at:
    at emitErrorNT (node:internal/streams/destroy:170:8)
    at emitErrorCloseNT (node:internal/streams/destroy:129:3)
  errno: -5,
  code: 'Z_BUF_ERROR'
}
```

This happens whenever the stream is not immediately consumed

The following will cause a crash:

```
const r = await fetch("http://localhost:3010/");
console.log("done");
```

```
const r = await fetch(url);
await r.body.cancel();
console.log("done");
```

```
const r = await fetch(url);
const reader = r.body.getReader()
console.log("done");
```

The following will NOT cause a crash, because the body is immediately consumed:
```
const r = await fetch(url);
await r.text();
console.log("done");
```

```
const r = await fetch(url);
const reader = r.body.getReader();
while (true) {
  const { value, done } = await reader.read();
  if (done) {
    break;
  }
}
```

fetch() should not a cause an unhandled error that crashes node, even if the response body is never consumed, especially since it is empty. This is the behavior with other compression types, such as `Content-Encoding: gzip`.

### What do you see instead?

fetch() causes node to crash due to an unhandled error event from BrotliDecompress

### Additional information

_No response_

Uzlopak:
Fixes #3616 by setting the corresponding flags for brotli like we did for gzip in #2126

https://nodejs.org/api/zlib.html#compressing-http-requests-and-responses

image

But why does deflate not need those flags?

### DIFF
--- a/lib/web/fetch/index.js
+++ b/lib/web/fetch/index.js
@@ -2060,7 +2060,7 @@ async function httpNetworkFetch (
 
   function dispatch ({ body }) {
     const url = requestCurrentURL(request)
-    /** @type {import('../..').Agent} */
+    /** @type {import('../../..').Agent} */
     const agent = fetchParams.controller.dispatcher
 
     return new Promise((resolve, reject) => agent.dispatch(
@@ -2074,6 +2074,9 @@ async function httpNetworkFetch (
         upgrade: request.mode === 'websocket' ? 'websocket' : undefined
       },
       {
+        /**
+         * @type {import('node:stream').Readable|null}
+         */
         body: null,
         abort: null,
 
@@ -2114,35 +2117,37 @@ async function httpNetworkFetch (
 
           /** @type {string[]} */
           let codings = []
-          let location = ''
 
           const headersList = new HeadersList()
 
           for (let i = 0; i < rawHeaders.length; i += 2) {
             headersList.append(bufferToLowerCasedHeaderName(rawHeaders[i]), rawHeaders[i + 1].toString('latin1'), true)
           }
+
           const contentEncoding = headersList.get('content-encoding', true)
           if (contentEncoding) {
             // https://www.rfc-editor.org/rfc/rfc7231#section-3.1.2.1
             // "All content-coding values are case-insensitive..."
             codings = contentEncoding.toLowerCase().split(',').map((x) => x.trim())
           }
-          location = headersList.get('location', true)
 
           this.body = new Readable({ read: resume })
 
-          const decoders = []
+          /** @type {[src: import('node:stream').Readable, ...target: import('node:stream').Writable[]]} */
+          let streams
 
-          const willFollow = location && request.redirect === 'follow' &&
-            redirectStatusSet.has(status)
+          const willFollow = request.redirect === 'follow' &&
+            redirectStatusSet.has(status) &&
+            headersList.get('location', true)
 
           // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding
           if (codings.length !== 0 && request.method !== 'HEAD' && request.method !== 'CONNECT' && !nullBodyStatus.includes(status) && !willFollow) {
+            streams = [this.body]
             for (let i = codings.length - 1; i >= 0; --i) {
               const coding = codings[i]
               // https://www.rfc-editor.org/rfc/rfc9112.html#section-7.2
               if (coding === 'x-gzip' || coding === 'gzip') {
-                decoders.push(zlib.createGunzip({
+                streams.push(zlib.createGunzip({
                   // Be less strict when decoding compressed responses, since sometimes
                   // servers send slightly invalid responses that are still accepted
                   // by common browsers.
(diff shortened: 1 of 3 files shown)


KhafraDev (changes requested):
an error listener should still be added onto pipeline(...)... this only fixes the one case
https://github.com/nodejs/undici/issues/3616#issuecomment-2356674965

Uzlopak:
@KhafraDev better?

lib/web/fetch/index.js
```diff
@@ -2165,9 +2178,9 @@ async function httpNetworkFetch (
             status,
             statusText,
             headersList,
-            body: decoders.length
-              ? pipeline(this.body, ...decoders, noop)
-              : this.body.on('error', noop)
+            body: streams === undefined
+              ? this.body.on('error', noop)
+              : pipeline(streams, noop).on('error', noop)
```
ronag, line 2183: this is just hiding a bug somewhere else

ronag (changes requested):

lib/web/fetch/index.js
Uzlopak: how about now?

Uzlopak:
@KhafraDev 
@ronag ptal

lib/web/fetch/index.js
```diff
@@ -2114,35 +2115,37 @@ async function httpNetworkFetch (
...
             // "All content-coding values are case-insensitive..."
             codings = contentEncoding.toLowerCase().split(',').map((x) => x.trim())
           }
-          location = headersList.get('location', true)
 
           this.body = new Readable({ read: resume })
 
-          const decoders = []
+          /** @type {[src: import('node:stream').Readable, ...target: import('node:stream').Writable[]]} */
+          let streams
```
KhafraDev, line 2135:
```suggestion
          const streams = []
```

lib/web/fetch/index.js
```diff
@@ -2114,35 +2115,37 @@ async function httpNetworkFetch (
...
 
-          const willFollow = location && request.redirect === 'follow' &&
-            redirectStatusSet.has(status)
+          const willFollow = request.redirect === 'follow' &&
+            redirectStatusSet.has(status) &&
+            headersList.get('location', true)
 
           // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding
           if (codings.length !== 0 && request.method !== 'HEAD' && request.method !== 'CONNECT' && !nullBodyStatus.includes(status) && !willFollow) {
+            streams = [this.body]
```
KhafraDev, line 2143:
```suggestion
```

lib/web/fetch/index.js
```diff
@@ -2151,23 +2154,38 @@ async function httpNetworkFetch (
...
-                decoders.push(zlib.createBrotliDecompress())
+                streams.push(zlib.createBrotliDecompress({
+                  flush: zlib.constants.BROTLI_OPERATION_FLUSH,
+                  finishFlush: zlib.constants.BROTLI_OPERATION_FLUSH
+                }))
               } else {
-                decoders.length = 0
+                // If the server sends the payload with a coding which his not
+                // supported, the body will be passed through without decoding.
+                streams = undefined
```
KhafraDev, line 2169:
```suggestion
                streams.length = 0
```

lib/web/fetch/index.js
```diff
@@ -2151,23 +2154,38 @@ async function httpNetworkFetch (
...
+          }
+
           resolve({
             status,
             statusText,
             headersList,
-            body: decoders.length
-              ? pipeline(this.body, ...decoders, noop)
-              : this.body.on('error', noop)
+            body: streams === undefined
```
KhafraDev, line 2186:
```suggestion
            body: streams.length === 0
```

lib/web/fetch/index.js
```diff
@@ -2151,23 +2154,38 @@ async function httpNetworkFetch (
...
+              ? this.body.on('error', onError)
+              : pipeline(streams, callback).on('error', onError)
```
KhafraDev, line 2188:
```suggestion
              : pipeline(this.body, ...streams, callback).on('error', onError)
```

KhafraDev (commented):
there are quite a few unnecessary changes that make this section harder to read

Uzlopak:
Tbh i dont like the suggested changes. So I retract this PR. No bad feelings ;).

KhafraDev:
I will work on it tomorrow.

PR #3632, opened 2024-09-20
handle body errors

### LINKED ISSUE #3616: Empty fetch() response with 'Content-Encoding: br' header causes a crash due to unhandled error event in BrotliDecompress

KhafraDev:
fixes #3616

### DIFF
--- a/lib/web/fetch/index.js
+++ b/lib/web/fetch/index.js
@@ -65,8 +65,6 @@ const { webidl } = require('./webidl')
 const { STATUS_CODES } = require('node:http')
 const GET_OR_HEAD = ['GET', 'HEAD']
 
-const noop = () => {}
-
 const defaultUserAgent = typeof __UNDICI_IS_NODE__ !== 'undefined' || typeof esbuildDetection !== 'undefined'
   ? 'node'
   : 'undici'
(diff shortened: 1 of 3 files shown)


ronag (approved):

Uzlopak:
Very good code ;)

PR #3628, opened 2024-09-20
fix: fire close on failed WebSocket connection

KhafraDev:
probably relies on #3627 landing first

the way we previously handled closing made no sense; either side could send/receive a close frame, but the closing handshake is only completed \_cleanly\_ when the said side sends & receives a close frame. We were only tracking a single state, and then we had a `receivedClose` boolean property... which is very confusing

now we know if a closing handshake
- has started (sent OR received close frame)
- has completed (sent AND received close frame)
- was closed cleanly (same logic as the bullet point above)

I have no idea if this will pass the autobahn tests, but it's too late for me to be running them (they take forever)

PR #3635, opened 2024-09-21
fix(websocket): use binary operations instead of a Set for closeStatus

Uzlopak:
Follow up to #3628

### DIFF
--- a/lib/web/websocket/websocket.js
+++ b/lib/web/websocket/websocket.js
@@ -39,7 +39,7 @@ const { channels } = require('../../core/diagnostics')
  *
  * @property {number} readyState
  * @property {import('stream').Duplex} socket
- * @property {Set<number>} closeState
+ * @property {number} closeState
  */
 
 // https://websockets.spec.whatwg.org/#interface-definition
@@ -84,7 +84,7 @@ class WebSocket extends EventTarget {
 
     readyState: states.CONNECTING,
     socket: null,
-    closeState: new Set()
+    closeState: sentCloseFrameState.NOT_SENT
   }
 
   #url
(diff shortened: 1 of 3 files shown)


KhafraDev (changes requested):
> It's possible to avoid the set *but nothing here is a hot path - websockets are not closed often.*

It's not worth the mental overhead. The websocket closing handshake is an extremely cold path.

Uzlopak:
Using a Set is an unconventional and to honest very ugly Implementation in our codebase. I strongly recommend to avoid ES6 Sets to just store 2 bit information...

lib/web/websocket/websocket.js
```diff
@@ -671,9 +671,7 @@ class WebSocket extends EventTarget {
     // If the TCP connection was closed after the
     // WebSocket closing handshake was completed, the WebSocket connection
     // is said to have been closed _cleanly_.
-    const wasClean =
-      this.#handler.closeState.has(sentCloseFrameState.SENT) &&
-      this.#handler.closeState.has(sentCloseFrameState.RECEIVED)
+    const wasClean = this.#handler.closeState === (sentCloseFrameState.SENT & sentCloseFrameState.RECEIVED)
```
Uzlopak, line 674:
```suggestion
    const wasClean = this.#handler.closeState === (sentCloseFrameState.SENT | sentCloseFrameState.RECEIVED)
```

KhafraDev:
I don't agree with any of that. The implementation is easy to understand and not in a critical path. Those paths are triggered when
- manually closing (`WebSocket.prototype.close`)
- receiving a close frame

You can only close a websocket once per instance, and websockets are rarely closed by nature. Not only that, `NOT_SENT` is an illogical name for a state because the closing handshake starts when sending **or** receiving a close frame.

Uzlopak:
@KhafraDev 

What if rename it to NONE?

PR #3636, opened 2024-09-21
feat: add new hooks api

ronag:
## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/handler/decorator-handler.js
+++ b/lib/handler/decorator-handler.js
@@ -1,11 +1,23 @@
 'use strict'
 
 const assert = require('node:assert')
+const util = require('../core/util')
+
+function toRawHeaders (headers) {
+  const rawHeaders = []
+  if (headers != null) {
+    for (const [key, value] of Object.entries(headers)) {
+      rawHeaders.push(Buffer.from(key), Buffer.from(value))
+    }
+  }
+  return rawHeaders
+}
 
 module.exports = class DecoratorHandler {
   #handler
   #onCompleteCalled = false
   #onErrorCalled = false
+  #resume = null
 
   constructor (handler) {
     if (typeof handler !== 'object' || handler === null) {
@@ -14,13 +26,96 @@ module.exports = class DecoratorHandler {
     this.#handler = handler
   }
 
+  // New API
+
+  onRequestStart (reserved, abort) {
+    if (this.#handler.onRequestStart) {
+      this.#handler.onRequestStart(null, abort)
+    }
+  }
+
+  onResponseStart (resume) {
+    this.#resume = resume
+
+    if (this.#handler.onResponseStart) {
+      return this.#handler.onResponseStart(resume)
+    }
+
+    return true
+  }
+
+  onResponseHeaders (statusCode, headers) {
+    if (this.#handler.onResponseHeaders) {
+      this.#handler.onResponseHeaders(headers, statusCode)
+    }
+
+    if (this.#handler.onConnect) {
+      this.#handler.onConnect(statusCode, toRawHeaders(headers), this.#resume)
+    }
+
+    return true
+  }
+
+  onResponseData (data) {
+    if (this.#handler.onResponseData) {
+      this.#handler.onResponseData(data)
+    }
+
+    if (this.#handler.onData) {
+      this.#handler.onData(data)
+    }
+
+    return true
+  }
+
+  onResponseEnd (trailers) {
+    if (this.#handler.onResponseEnd) {
+      this.#handler.onResponseEnd(trailers)
+    }
+
+    if (this.#handler.onComplete) {
+      this.#handler.onComplete(toRawHeaders(trailers))
+    }
+  }
+
+  onResponseError (err) {
+    if (this.#handler.onResponseError) {
+      this.#handler.onResponseError(err)
+    }
+
+    if (this.#handler.onError) {
+      this.#handler.onError(err)
+    }
+  }
+
+  // Old API
+
+  onRequestSent (...args) {
+    if (this.#handler.onRequestSent) {
+      this.#handler.onRequestSent(...args)
+    }
+  }
+
   onConnect (...args) {
-    return this.#handler.onConnect?.(...args)
+    if (this.#handler.onRequestStart) {
+      this.#handler.onRequestStart(null, args[0])
+    }
+
+    if (this.#handler.onConnect) {
+      return this.#handler.onConnect(...args)
+    }
   }
 
   onError (...args) {
     this.#onErrorCalled = true
-    return this.#handler.onError?.(...args)
+
+    if (this.#handler.onResponseError) {
+      this.#handler.onResponseError(args[0])
+    }
+
+    if (this.#handler.onError) {
+      return this.#handler.onError(...args)
+    }
   }
 
   onUpgrade (...args) {
@@ -32,7 +127,7 @@ module.exports = class DecoratorHandler {
 
   onResponseStarted (...args) {
     assert(!this.#onCompleteCalled)
-    assert(!this.#onErrorCalled)
+    // assert(!this.#onErrorCalled)
 
     return this.#handler.onResponseStarted?.(...args)
   }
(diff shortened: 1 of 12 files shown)


ronag:
@mcollina @KhafraDev @metcoder95 @ShogunPanda PTAL

ronag:
```
diff --git a/lib/cache/memory-cache-store.js b/lib/cache/memory-cache-store.js
index 9baa30b9..a523fb52 100644
--- a/lib/cache/memory-cache-store.js
+++ b/lib/cache/memory-cache-store.js
@@ -74,66 +74,42 @@ class MemoryCacheStore {
 
     const value = this.#findValue(req, values)
 
-    return value ? new MemoryStoreReadableStream(value) : undefined
+    if (!value) {
+      return undefined
+    }
+
+    return new Readable.from(value.body)
   }
 
-  createWriteStream (req, value) {
+  createWriteStream (req, opts) {
     if (typeof req !== 'object') {
       throw new TypeError(`expected req to be object, got ${typeof req}`)
     }
-    if (typeof value !== 'object') {
-      throw new TypeError(`expected value to be object, got ${typeof value}`)
+    if (typeof opts !== 'object') {
+      throw new TypeError(`expected value to be object, got ${typeof opts}`)
     }
 
     if (this.isFull) {
       return undefined
     }
 
-    const values = this.#getValuesForRequest(req, true)
-    let storedValue = this.#findValue(req, values)
-    if (!storedValue) {
-      this.#entryCount++
+    // TODO (fix): Handle options
 
-      // TODO better name for this
-      storedValue = {
-        complete: false,
-        value,
-        emitter: new EventEmitter()
-      }
-
-      if (
-        values.length === 0 ||
-        value.deleteAt < values[values.length - 1].deleteAt
-      ) {
-        values.push(storedValue)
-      } else if (value.deleteAt >= values[0].deleteAt) {
-        values.unshift(storedValue)
-      } else {
-        let startIndex = 0
-        let endIndex = values.length
-        while (true) {
-          if (startIndex === endIndex) {
-            values.splice(startIndex, 0, storedValue)
-            break
-          }
-
-          const middleIndex = Math.floor((startIndex + endIndex) / 2)
-          const middleValue = values[middleIndex]
-          if (value.deleteAt === middleIndex) {
-            values.splice(middleIndex, 0, storedValue)
-            break
-          } else if (value.deleteAt > middleValue.value.deleteAt) {
-            endIndex = middleIndex
-            continue
-          } else {
-            startIndex = middleIndex
-            continue
-          }
+    const body = []
+    return new Writable({
+      write (chunk, encoding, callback) {
+        if (body.length < this.#maxEntrySize) {
+          body.push(typeof chunk === 'string' ? Buffer.from(chunk, encoding) : chunk)
         }
+        callback(null)
+      },
+      final(callback) {
+        if (body.length < this.#maxEntrySize) {
+          // Store body
+        }
+        callback(null)
       }
-    }
-
-    return new MemoryStoreWritableStream(storedValue)
+    })
   }
 
   /**
```

ShogunPanda (approved):
LGTM!

lib/handler/decorator-handler.js
```diff
@@ -32,7 +127,7 @@ module.exports = class DecoratorHandler {
 
   onResponseStarted (...args) {
     assert(!this.#onCompleteCalled)
-    assert(!this.#onErrorCalled)
+    // assert(!this.#onErrorCalled)
```
mcollina, line 130: why?

mcollina (changes requested):
My understanding is that this is fully backward compatible?
I think we should bump the version on the global dispatcher symbol:

https://github.com/nodejs/undici/blob/90e2e1318852f1d4a2ce5c0571a91912cb7e519e/lib/global.js#L5

And then add an adapter to the "1" API, so that fetch keeps working.

Also... tests are missing.

lib/handler/decorator-handler.js
ronag: We have a bug in the h2 client that wasn't caught before.

ronag:
> My understanding is that this is fully backward compatible?

Yes

> I think we should bump the version on the global dispatcher symbol:

Any specific reason why you think we should bump it?

> Also... tests are missing.

👓

lib/handler/decorator-handler.js
ronag: @metcoder95: client-h2. calls onResponseStarted after onError

ronag:
@Uzlopak do you think you could help with tests?

ronag:
This needs more work but I think it's mostly semver minor.

lib/handler/decorator-handler.js
metcoder95: Hmm, interesting; I imagine this happens if something failed before dispatching the request, or how do you find it?

ronag: I applied the assertions to the core request handling:

https://github.com/nodejs/undici/pull/3636/files#diff-ea7696549c3a0b60a4a7e07cc79b6d4e950c7cb1068d47e368a510967d77e7e5R49-R51

lib/handler/decorator-handler.js
```diff
@@ -14,12 +26,88 @@ module.exports = class DecoratorHandler {
...
+    }
+
+    return ret
+  }
+
+  onResponseData (data) {
+    let ret = true
+
+    if (this.#handler.onResponseData?.(data) === false) {
+      ret = false
```
metcoder95, line 69: Same as my previous comment

lib/handler/decorator-handler.js
```diff
@@ -14,12 +26,88 @@ module.exports = class DecoratorHandler {
...
+    if (this.#handler.onResponseHeaders?.(headers, statusCode) === false) {
+      ret = false
+    }
+
+    if (this.#handler.onHeaders) {
+      const rawHeaders = toRawHeaders(headers)
+      if (this.#handler.onHeaders(statusCode, rawHeaders, this.#resume) === false) {
+        ret = false
+      }
+    }
```
metcoder95, line 60: Here we will have consistency conditions, isn't it?
Because if one of them returns `false`, and other `true`, we will continue reading chunks, causing that possibly one of them gets corrupted.

Shall we better allow one or the other but nut the two of them?

metcoder95 (commented):
The new APIs lgtm, left some minor comments 👍 

Seems some tests will need adjustment as well documentation. I will check the h2 bug later this week

lib/handler/decorator-handler.js
ronag: Not sure I follow. If one of them returns false then the return value is false.

metcoder95: Sure. But then what would happen if e.g. `onResponseHeaders` returns `false` but `onHeaders` returns `true`?

I think we are allowing them to pass both APIs at the same time? Wouldn't that maybe cause inconsistencies?

ronag: No that's fine, if one returns false then it's paused.

lib/handler/decorator-handler.js
```diff
@@ -14,12 +26,88 @@ module.exports = class DecoratorHandler {
...
+  }
+
+  onResponseHeaders (headers, statusCode) {
+    let ret = true
+
+    if (this.#handler.onResponseHeaders?.(headers, statusCode) === false) {
+      ret = false
+    }
+
+    if (this.#handler.onHeaders) {
```
ShogunPanda, line 55: Shall we prevent to have both the old and new API?

ronag: How would we do that?

metcoder95: Maybe we can check at [`dispatcher-base`](https://github.com/nodejs/undici/blob/e6e87c13dbf8788fe158fd60631cee78eb84f198/lib/dispatcher/dispatcher-base.js#L128) when calling `dispatch`. We already validate the `handler`, we can take advantage of validating the method as well if we want to avoid receiving both functions/method at the same time.

PR #3638, opened 2024-09-22
cleanup web symbol removal

KhafraDev:
- removes kHeadersSortedMap from Headers' prototype
- fix wpt runner (we were `import`ing from a file which was removed


tsctx (approved):

mcollina (approved):
lgtm

PR #3650, opened 2024-09-26
chore: integrate consume functions into BodyReadable

Uzlopak:
All the consume functions are actually processing the internal state of the BodyReadable indicating that they should be private methods and not standalone functions.


ronag (changes requested):
I don't think this is an improvement.

Uzlopak:
Actually it improves also the stack traces.

PR #3654, opened 2024-09-30
feat: check maxHeadersSize on client instantiation and not on Parser instantion

### LINKED ISSUE #3653: Instructions on proper cleanup when using fetch()
## This would solve...

Undici documentation could be more exact on the cleanup of resources allocated by `fetch()`. The documentation mentions [Specification Compliance regarding garbage collection](https://undici.nodejs.org/#/?id=garbage-collection) that Undici expects the response body to be consumed but no mentioning on how to best consume them in different cases. This leaves uncertainty on how to best cleanup body and if there might be other resources to be cleaned up (and how) in case of 1) successful responses, 2) HTTP errors (e.g. responses with non-200 status codes) and 3) in cases fetch throws an error. 

A simple snippet on proper cleanup of resources would build confidence that we have done the cleanup properly.

## Additional context

In our cases, specifically, connected services may return garbled responses with fully valid status codes. Therefore we would not like to consume the body at all, but cancel it with `await response.body?.cancel()`. This seems to work alright, but it still leaves us questions whether some cleanup would be needed in case fetch throws an error. There might be some other cases, as well, that we are unaware of.

Uzlopak:
<s>I guess this would fix 3652 .</s> I am less inclined to download NW.js and figure out how it works. Imho validation of maxHeadersSize should not be done in Parser construction but in the client instantiation.


metcoder95 (commented):
Adding the extra step of validation on `dispatcher/client` lgmt, but I believe the issue will prevail.

The problem roots to the defaults. If no `maxHeaderSize` passed and if the runtime doesn't populate `[(node:http).maxHeaderSize`](https://nodejs.org/api/http.html#httpmaxheadersize), there won't be value to check against it.

I believe the check on `client-h1` aimed to avoid that exact scenario.

Uzlopak:
Yeah, I guess we can consider this an improvement and not a fix. Unfortunately i can somehow not remove the link to that issue. If you approve I would merge it and then reopen the linked issue.

lib/dispatcher/client-h1.js
```diff
@@ -201,8 +201,6 @@ class Parser {
      * @param {*} llhttp
      */
   constructor (client, socket, { exports }) {
-    assert(Number.isFinite(client[kMaxHeadersSize]) && client[kMaxHeadersSize] > 0)
```
metcoder95, line 204: I'd suggest to keep this one because of my previous point about the defaults, the rest LGTM

Uzlopak:
@metcoder95 

I still think that we dont need that check in the parser. Also why should the parser check if the attribute on the client is valid? Actually hints it to be a validation, that has to happen in the Client API. But you are right, we need to be stricter.

PTAL

Uzlopak:
Or wait... hmm, wont it mean, that you can set maxHeaderSize to 0 meaning any header size?

metcoder95 (approved):
LGTM,

I was mostly meaning the way we get the default max header size; which you already addressed within the `lib/dispatcher/client`

metcoder95:
Just tests seems not happy

lib/dispatcher/client-h1.js
Uzlopak: we dont need to have this check because we set it already properly on client.

Uzlopak:
@metcoder95 

Tests should now be happy. :D

Also I think that the check in parser is not necessary anymore, because it is handled properly in client already.

metcoder95 (approved):
lgtm

PR #3656, opened 2024-10-01
test: less flaky timers acceptance test, rework fast timer tests to pass them faster

Uzlopak:
using sinon fake timers to run the tests faster


Uzlopak:
Thank you!

PR #3562, opened 2024-09-07
feat: http caching

flakey5:
Implements bare-bones opt-in http caching as per [rfc9111](https://www.rfc-editor.org/rfc/rfc9111.html). Bare-bones in this case means what's required by the spec and a few extra bits, with more coming in future prs.

~~Opening as a draft since there's still some more work to be done (mostly tests, but a bit more functionality-wise)~~

No request cache directives are supported at this time, this will come later.

Response caching directives supported:
 * `public`
 * `private`
 * `s-maxage`
 * `max-age`
 * `Expires` header
 * `no-cache`
 * `no-store`
 * `stale-while-revalidate`

## This relates to...

Closes #3231
Closes #2760
Closes #2256
Closes #1146

### Features

 * Opt-in http caching
 * In-memory default cache store

### Bug Fixes

n/a

### Breaking Changes and Deprecations

n/a

## Status

- [x] I have read and agreed to the [Developer's Certificate of Origin][cert]
- [x] Tested
- [ ] Benchmarked (**optional**)
- [x] Documented
- [x] Review ready
- [x] In review
- [ ] Merge ready

[cert]: https://github.com/nodejs/undici/blob/main/CONTRIBUTING.md

PR #3657, opened 2024-10-01
feat: sqlite cache store

flakey5:
~~**Note**: this is a draft since #3562 isn't landed. Until it is landed, this will be based off of that pr's branch. For the actual diff see https://github.com/flakey5/undici/compare/flakey5/3231...flakey5:undici:flakey5/20240910/sqlite-cache-store~~

## This relates to...

Adding client side http caching (#3562)

## Rationale

Adds a sqlite cache store option given the `--experimental-sqlite` flag is provided

### Features

 * Cache store using sqlite as a backend

### Bug Fixes

n/a

n/a

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [x] In review
- [ ] Merge ready

cc @mcollina @ronag

### DIFF
--- a/docs/docs/api/CacheStore.md
+++ b/docs/docs/api/CacheStore.md
@@ -13,8 +13,20 @@ The `MemoryCacheStore` stores the responses in-memory.
 
 **Options**
 
-- `maxEntries` - The maximum amount of responses to store. Default `Infinity`.
+- `maxCount` - The maximum amount of responses to store. Default `Infinity`.
 - `maxEntrySize` - The maximum size in bytes that a response's body can be. If a response's body is greater than or equal to this, the response will not be cached.
+
+### `SqliteCacheStore`
+
+The `SqliteCacheStore` stores the responses in a SQLite database.
+Under the hood, it uses Node.js' [`node:sqlite`](https://nodejs.org/api/sqlite.html) api.
+The `SqliteCacheStore` is only exposed if the `node:sqlite` api is present.
+
+**Options**
+
+- `location` - The location of the SQLite database to use. Default `:memory:`.
+- `maxCount` - The maximum number of entries to store in the database. Default `Infinity`.
+- `maxEntrySize` - The maximum size in bytes that a resposne's body can be. If a response's body is greater than or equal to this, the response will not be cached. Default `Infinity`.
 
 ## Defining a Custom Cache Store
 
--- a/index.js
+++ b/index.js
@@ -48,6 +48,15 @@ module.exports.cacheStores = {
   MemoryCacheStore: require('./lib/cache/memory-cache-store')
 }
 
+try {
+  const SqliteCacheStore = require('./lib/cache/sqlite-cache-store')
+  module.exports.cacheStores.SqliteCacheStore = SqliteCacheStore
+} catch (err) {
+  if (err.code !== 'ERR_UNKNOWN_BUILTIN_MODULE') {
+    throw err
+  }
+}
+
 module.exports.buildConnector = buildConnector
 module.exports.errors = errors
 module.exports.util = {
--- a/lib/cache/sqlite-cache-store.js
+++ b/lib/cache/sqlite-cache-store.js
@@ -1,0 +1,450 @@
+'use strict'
+
+const { DatabaseSync } = require('node:sqlite')
+const { Writable } = require('stream')
+const { assertCacheKey, assertCacheValue } = require('../util/cache.js')
+
+/**
+ * @typedef {import('../../types/cache-interceptor.d.ts').default.CacheStore} CacheStore
+ * @implements {CacheStore}
+ *
+ * @typedef {{
+ *  id: Readonly<number>
+ *  rawHeaders?: string
+ *  vary?: string | object
+ *  body: string
+ * } & import('../../types/cache-interceptor.d.ts').default.CacheValue} SqliteStoreValue
+ */
+class SqliteCacheStore {
+  #maxEntrySize = Infinity
+  #maxCount = Infinity
+
+  /**
+   * @type {import('node:sqlite').DatabaseSync}
+   */
+  #db
+
+  /**
+   * @type {import('node:sqlite').StatementSync}
+   */
+  #getValuesQuery
+
+  /**
+   * @type {import('node:sqlite').StatementSync}
+   */
+  #updateValueQuery
+
+  /**
+   * @type {import('node:sqlite').StatementSync}
+   */
+  #insertValueQuery
+
+  /**
+   * @type {import('node:sqlite').StatementSync}
+   */
+  #deleteExpiredValuesQuery
+
+  /**
+   * @type {import('node:sqlite').StatementSync}
+   */
+  #deleteByUrlQuery
+
+  /**
+   * @type {import('node:sqlite').StatementSync}
+   */
+  #countEntriesQuery
+
+  /**
+   * @type {import('node:sqlite').StatementSync}
+   */
+  #deleteOldValuesQuery
+
+  /**
+   * @param {import('../../types/cache-interceptor.d.ts').default.SqliteCacheStoreOpts | undefined} opts
+   */
+  constructor (opts) {
+    if (opts) {
+      if (typeof opts !== 'object') {
+        throw new TypeError('SqliteCacheStore options must be an object')
+      }
+
+      if (opts.maxEntrySize !== undefined) {
+        if (
+          typeof opts.maxEntrySize !== 'number' ||
+          !Number.isInteger(opts.maxEntrySize) ||
+          opts.maxEntrySize < 0
+        ) {
+          throw new TypeError('SqliteCacheStore options.maxEntrySize must be a non-negative integer')
+        }
+        this.#maxEntrySize = opts.maxEntrySize
+      }
+
+      if (opts.maxCount !== undefined) {
+        if (
+          typeof opts.maxCount !== 'number' ||
+          !Number.isInteger(opts.maxCount) ||
+          opts.maxCount < 0
+        ) {
+          throw new TypeError('SqliteCacheStore options.maxCount must be a non-negative integer')
+        }
+        this.#maxCount = opts.maxCount
+      }
+    }
+
+    this.#db = new DatabaseSync(opts?.location ?? ':memory:')
+
+    this.#db.exec(`
+      CREATE TABLE IF NOT EXISTS cacheInterceptorV1 (
+        -- Data specific to us
+        id INTEGER PRIMARY KEY AUTOINCREMENT,
+        url TEXT NOT NULL,
+        method TEXT NOT NULL,
+
+        -- Data returned to the interceptor
+        body TEXT NULL,
+        deleteAt INTEGER NOT NULL,
+        statusCode INTEGER NOT NULL,
+        statusMessage TEXT NOT NULL,
+        rawHeaders TEXT NULL,
+        etag TEXT NULL,
+        vary TEXT NULL,
+        cachedAt INTEGER NOT NULL,
+        staleAt INTEGER NOT NULL
+      );
+
+      CREATE INDEX IF NOT EXISTS idx_cacheInterceptorV1_url ON cacheInterceptorV1(url);
+      CREATE INDEX IF NOT EXISTS idx_cacheInterceptorV1_method ON cacheInterceptorV1(method);
+      CREATE INDEX IF NOT EXISTS idx_cacheInterceptorV1_deleteAt ON cacheInterceptorV1(deleteAt);
+    `)
+
+    this.#getValuesQuery = this.#db.prepare(`
+      SELECT
+        id,
+        body,
+        deleteAt,
+        statusCode,
+        statusMessage,
+        rawHeaders,
+        etag,
+        vary,
+        cachedAt,
+        staleAt
+      FROM cacheInterceptorV1
+      WHERE
+        url = ?
+        AND method = ?
+      ORDER BY
+        deleteAt ASC
+    `)
+
+    this.#updateValueQuery = this.#db.prepare(`
+      UPDATE cacheInterceptorV1 SET
+        body = ?,
+        deleteAt = ?,
+        statusCode = ?,
+        statusMessage = ?,
+        rawHeaders = ?,
+        etag = ?,
+        cachedAt = ?,
+        staleAt = ?,
+        deleteAt = ?
+      WHERE
+        id = ?
+    `)
+
+    this.#insertValueQuery = this.#db.prepare(`
+      INSERT INTO cacheInterceptorV1 (
+        url,
+        method,
+        body,
+        deleteAt,
+        statusCode,
+        statusMessage,
+        rawHeaders,
+        etag,
+        vary,
+        cachedAt,
+        staleAt,
+        deleteAt
+      ) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+    `)
+
+    this.#deleteExpiredValuesQuery = this.#db.prepare(
+      'DELETE FROM cacheInterceptorV1 WHERE deleteAt <= ?'
+    )
+
+    this.#deleteByUrlQuery = this.#db.prepare(
+      'DELETE FROM cacheInterceptorV1 WHERE url = ?'
+    )
+
+    this.#countEntriesQuery = this.#db.prepare(
+      'SELECT COUNT(*) AS total FROM cacheInterceptorV1'
+    )
+
+    const pruneLimit = this.#maxCount === Infinity
+      ? 20
+      : Math.max(Math.floor(this.#maxCount * 0.1), 1)
+
+    this.#deleteOldValuesQuery = this.#db.prepare(`
+      DELETE FROM cacheInterceptorV1
+      WHERE id IN (
+        SELECT
+          id
+        FROM cacheInterceptorV1
+        ORDER BY cachedAt DESC
+        LIMIT ${pruneLimit}
+      )
+    `)
+  }
+
+  close () {
+    this.#db.close()
+  }
+
+  /**
+   * @param {import('../../types/cache-interceptor.d.ts').default.CacheKey} key
+   * @returns {import('../../types/cache-interceptor.d.ts').default.GetResult | undefined}
+   */
+  get (key) {
+    assertCacheKey(key)
+
+    const value = this.#findValue(key)
+
+    if (!value) {
+      return undefined
+    }
+
+    /**
+     * @type {import('../../types/cache-interceptor.d.ts').default.GetResult}
+     */
+    const result = {
+      body: value.body ? parseBufferArray(JSON.parse(value.body)) : null,
+      statusCode: value.statusCode,
+      statusMessage: value.statusMessage,
+      rawHeaders: value.rawHeaders ? parseBufferArray(JSON.parse(value.rawHeaders)) : undefined,
+      etag: value.etag ? value.etag : undefined,
+      cachedAt: value.cachedAt,
+      staleAt: value.staleAt,
+      deleteAt: value.deleteAt
+    }
+
+    return result
+  }
+
+  /**
+   * @param {import('../../types/cache-interceptor.d.ts').default.CacheKey} key
+   * @param {import('../../types/cache-interceptor.d.ts').default.CacheValue} value
+   * @returns {Writable | undefined}
+   */
+  createWriteStream (key, value) {
+    assertCacheKey(key)
+    assertCacheValue(value)
+
+    const url = this.#makeValueUrl(key)
+    let currentSize = 0
+    /**
+     * @type {Buffer[] | null}
+     */
+    let body = key.method !== 'HEAD' ? [] : null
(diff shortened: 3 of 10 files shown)


test/cache-interceptor/cache-stores.js
```diff
@@ -1,12 +1,20 @@
...
+const { env, execArgv } = require('node:process')
 const { describe, test } = require('node:test')
 const { deepStrictEqual, notEqual, equal } = require('node:assert')
 const { once } = require('node:events')
 const MemoryCacheStore = require('../../lib/cache/memory-cache-store')
 
 cacheStoreTests(MemoryCacheStore)
 
+if (
+  (env.NODE_OPTIONS && env.NODE_OPTIONS.match(/experimental(-|_)sqlite/)) ||
```
flakey5, line 12: since this requires the experimental flag, what do we want to do for ci?

lib/cache/sqlite-cache-store.js
```diff
@@ -0,0 +1,597 @@
...
+        throw new TypeError('MemoryCacheStore options must be an object')
+      }
+
+      if (opts.maxEntries !== undefined) {
+        if (
+          typeof opts.maxEntries !== 'number' ||
+          !Number.isInteger(opts.maxEntries) ||
+          opts.maxEntries < 0
+        ) {
+          throw new TypeError('MemoryCacheStore options.maxEntries must be a non-negative integer')
```
SukkaW, line 98:
```suggestion
          throw new TypeError('SqliteCacheStore options.maxEntries must be a non-negative integer')
```

Any many below.

test/cache-interceptor/cache-stores.js
SukkaW: What about adding a new matrix for `NODE_OPTIONS`, and using `matrix.include` and `matrix.exclude` to guard the experimental flag on specific Node.js versions?

GitHub docs happen to have a matrix example with `NODE_OPTIONS`: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/running-variations-of-jobs-in-a-workflow#example-using-a-multi-dimension-matrix

metcoder95: I'd say we can create a new matrix that focuses on features that requires experimental functionalities from Node (similar to what @SukkaW suggests).

mcollina (changes requested):
Can you add some docs?

lib/cache/sqlite-cache-store.js
```diff
@@ -0,0 +1,597 @@
...
+
+  #deleteExpiredValuesStatement
+
+  #deleteByOriginStatement
+
+  /**
+   * Ids of rows currently being written too and so can't be read from
+   * @type {Set<number>}
+   */
+  #readLocks = new Set()
```
mcollina, line 74: I don't think you can do this, because it wouldn't work across processes

lib/cache/sqlite-cache-store.js
```diff
@@ -0,0 +1,597 @@
...
+
+  /**
+   * Denotes a row can't be written to because there's readables reading from it
+   * Row id:number of readables open for that value
+   * @type {Map<number, number>}
+   */
+  #writeLocks = new Map()
```
mcollina, line 81: I don't think you can do this, because it wouldn't work across processes

lib/cache/sqlite-cache-store.js
```diff
@@ -0,0 +1,597 @@
...
+        rawTrailers TEXT NULL,
+        vary TEXT NULL,
+        cachedAt INTEGER NOT NULL,
+        staleAt INTEGER NOT NULL
+      );
+
+      CREATE INDEX IF NOT EXISTS idx_cacheInterceptor_origin ON cacheInterceptor(origin);
+      CREATE INDEX IF NOT EXISTS idx_cacheInterceptor_method ON cacheInterceptor(method);
+      CREATE INDEX IF NOT EXISTS idx_cacheInterceptor_deleteAt ON cacheInterceptor(deleteAt);
+    `)
```
mcollina, line 147: I think we need to add a tiny bit of migrations for this.

mcollina (changes requested):

lib/cache/sqlite-cache-store.js
flakey5: Wdym?

mcollina: I think you need to query SQLite for that specific table, ask if the fields you need are there, and if not remove it and create it from scratch. It's a cache, it's ok ti wipe. Also check what Node.js is doing for handling localStorage.

metcoder95: Agree with @mcollina, something similar to `localStorage` can be beneficial

SukkaW: > Agree with @mcollina, something similar to `localStorage` can be beneficial

IMHO we should not do that. `SQLiteCacheStore` stores multi-dimensions data, it is just bad to use a KV-like `localStorage` w/ serialized cache.

flakey5: > IMHO, we should not do that. We are storing multi-dimensional data. It is just bad to use a KV-like `localStorage` with serialized cache keys. It is also difficult to implement cache expiration and purge with a KV-like `localStorage`.
> 
> IMHO, just create another cache provider with/ `localStorage` then.

I think they were referencing the implementation of `localStorage`, not about using it. The implementation in node uses sqlite, https://github.com/nodejs/node/blob/7ddd2c2282968cdb17b541bb8fe07f039c429897/src/node_webstorage.cc#L103

mcollina (approved):
lgtm

mcollina (approved):
lgtm

lib/cache/sqlite-cache-store.js
```diff
@@ -0,0 +1,514 @@
+'use strict'
+
+const { env, execArgv } = require('node:process')
+if (
+  !(env.NODE_OPTIONS && env.NODE_OPTIONS.match(/experimental(-|_)sqlite/)) &&
+  !execArgv.some(argv => argv.match(/experimental(-|_)sqlite/))
+) {
+  throw new Error('SqliteCacheStore needs the --experimental-sqlite flag enabled')
+}
```
mcollina, line 9: I don't think this is needed here

index.js
```diff
@@ -48,6 +49,13 @@ module.exports.cacheStores = {
   MemoryCacheStore: require('./lib/cache/memory-cache-store')
 }
 
+if (
+  (env.NODE_OPTIONS && env.NODE_OPTIONS.match(/experimental(-|_)sqlite/)) ||
+  execArgv.some(argv => argv.match(/experimental(-|_)sqlite/))
+) {
+  module.exports.cacheStores.SqliteCacheStore = require('./lib/cache/sqlite-cache-store')
+}
```
mcollina, line 57: I would actually try requiring `node:sqlite` instead, and catch that error

test/cache-interceptor/cache-stores.js
```diff
@@ -1,21 +1,94 @@
...
+const { rm } = require('node:fs/promises')
 const MemoryCacheStore = require('../../lib/cache/memory-cache-store')
 
 cacheStoreTests(MemoryCacheStore)
 
+if (
+  (env.NODE_OPTIONS && env.NODE_OPTIONS.match(/experimental(-|_)sqlite/)) ||
+  execArgv.some(argv => argv.match(/experimental(-|_)sqlite/))
+) {
+  const SqliteCacheStore = require('../../lib/cache/sqlite-cache-store.js')
```
mcollina, line 17: can you split this test in another file?

ronag (changes requested):
- Unnecessarily complicated by locking.
- How does it "prune" i.e. free space?
- Why errorCallback?

ronag:
IMHO I would think you could do this quite similar to the "simplified" memory store.

lib/cache/sqlite-cache-store.js
```diff
@@ -0,0 +1,506 @@
...
+    }
+
+    this.#deleteByUrlQuery.run(this.#makeValueUrl(key))
+  }
+
+  /**
+   * Checks the database's sqlite_master table to ensure we're working with the
+   *  correct schema
+   */
+  #verifyTableSql () {
```
ronag, line 354: This seems a bit paranoid? @mcollina?

index.js
```diff
@@ -48,6 +48,13 @@ module.exports.cacheStores = {
   MemoryCacheStore: require('./lib/cache/memory-cache-store')
 }
 
+try {
+  const SqliteCacheStore = require('./lib/cache/sqlite-cache-store')
+  module.exports.cacheStores.SqliteCacheStore = SqliteCacheStore
+} catch (_) {
+  // Do nothing
```
ronag, line 55: I think we should check for expect error

flakey5: Wdym?

mcollina: Basically verify that this errors with a certain error code, and if not rethrow the error

flakey5:
> Why errorCallback?

Goes back to the discussion we were having in the original caching pr for how to handle errors, we decided to handle them silently so the request still goes through.

Here it's just being used for handling json parse errors, since those _shouldn't_ ever happen I would be in favor of just throwing outright though

> How does it "prune" i.e. free space?

It prunes the entire database of expired values when we come across one when looking up a key, we can make this more stricter though (i.e. run the purge query on every write), wdyt?

flakey5:
I omitted maxCount here since it would break if there were multiple processes using a db for cache

If we do want it, I _think_ we might be able to get away with it being stored somewhere in the db and being modified only with a procedure.

ronag:
> If we do want it, I think we might be able to get away with it being stored somewhere in the db and being modified only with a procedure.

I believe sqlite has meta commands for that, @mcollina ?

lib/cache/sqlite-cache-store.js
```diff
@@ -0,0 +1,399 @@
...
+          //  requirements no matter what, let's return early
+          return undefined
+        }
+
+        try {
+          value.vary = JSON.parse(value.vary)
+        } catch (err) {
+          if (this.#errorCallback !== undefined) {
+            this.#errorCallback(err)
+          }
```
ronag, line 351: IMHO If the error callback is just for this I would rather skip it. Would even argue we should crash here if this happens (it shouldn't). @mcollina?

lib/cache/sqlite-cache-store.js
```diff
@@ -0,0 +1,399 @@
...
+            value.deleteAt,
+            value.statusCode,
+            value.statusMessage,
+            JSON.stringify(stringifyBufferArray(value.rawHeaders)),
+            value.vary ? JSON.stringify(value.vary) : null,
+            value.cachedAt,
+            value.staleAt,
+            value.deleteAt
+          )
+        }
```
ronag, line 284: Can't you just always just overwrite? Does it update and insert need to be separate ops?

flakey5: I think so but I'm not sure how messy it would be compared to just having two queries

ronag: Does update fail if the value doesn't exist? Doesn't sql have an insert or update command?

flakey5: > Does update fail if the value doesn't exist

It fails in the sense that it noops.

> Doesn't sql have an insert or update command?

I'm not sure how well insert or update would work here based on how we're checking if there's a already cached value though. The only reliable and fast way we have to compare two cached values is with their ids. Otherwise, we need to compare their url, method, and vary headers to see if they were the same. The url and method comparisons are simple and can be done in sql, but comparing the vary headers is more difficult since it's stringified json. We can't compare it in sql since we don't know if the properties are in the same order, meaning we need to parse then compare (like what `#findValues` does).

I was experimenting with hashing the vary headers (sort properties alphabetically -> stringify -> crypto.createHash), however, it was a lot more complicated than just having two separate queries. It also would've only benefited writes, for reads we would still need to fetch each response and compare the vary headers like we already do

lib/cache/sqlite-cache-store.js
```diff
@@ -0,0 +1,406 @@
...
+            value.statusCode,
+            value.statusMessage,
+            value.rawHeaders ? JSON.stringify(stringifyBufferArray(value.rawHeaders)) : null,
+            value.etag ? value.etag : null,
+            value.vary ? JSON.stringify(value.vary) : null,
+            value.cachedAt,
+            value.staleAt,
+            value.deleteAt
+          )
+        }
```
ronag, line 291: IMHO this should always delete and then insert.

lib/cache/sqlite-cache-store.js
```diff
@@ -0,0 +1,406 @@
...
+      }
+
+      let matches = true
+
+      if (value.vary) {
+        if (!key.headers) {
+          // Request doesn't have headers so it can't fulfill the vary
+          //  requirements no matter what, let's return early
+          return undefined
+        }
```
ronag, line 351: Should we also check `deletedAt` here?

lib/cache/sqlite-cache-store.js
```diff
@@ -0,0 +1,406 @@
...
+ *
+ * @typedef {{
+ *  id: Readonly<number>
+ *  rawHeaders?: string
+ *  vary?: string | object
+ *  body: string
+ * } & import('../../types/cache-interceptor.d.ts').default.CacheValue} SqliteStoreValue
+ */
+class SqliteCacheStore {
+  #maxSize = Infinity
```
ronag, line 19: this needs maxEntrySize as well?

lib/cache/sqlite-cache-store.js
```diff
@@ -0,0 +1,406 @@
...
+
+    if (values.length === 0) {
+      // No responses, let's just return early
+      return undefined
+    }
+
+    const now = nowAbsolute()
+    for (const value of values) {
+      if (now >= value.deleteAt && !canBeExpired) {
+        this.#deleteExpiredValuesQuery.run(now)
```
ronag, line 340: Can this be a slow query? Should you just do a delete with id?

types/cache-interceptor.d.ts
```diff
@@ -56,6 +57,11 @@ declare namespace CacheHandler {
    * Underlying storage provider for cached responses
    */
   export interface CacheStore {
+    /**
+     * Whether or not the cache is full and can not store any more responses
+     */
+    isFull?: Readonly<boolean>
```
ronag, line 63: I still don't see the point of this. The cache is never full. If it's full then the older values should be removed instead of not caching the most recent value.

docs/docs/api/CacheStore.md
```diff
@@ -14,7 +14,15 @@ The `MemoryCacheStore` stores the responses in-memory.
 **Options**
 
 - `maxEntries` - The maximum amount of responses to store. Default `Infinity`.
-- `maxEntrySize` - The maximum size in bytes that a response's body can be. If a response's body is greater than or equal to this, the response will not be cached.
+- `maxSize` - The maximum size in bytes that a response's body can be. If a response's body is greater than or equal to this, the response will not be cached.
+
+### `SqliteCacheStore`
+
+The `SqliteCacheStore` stores the responses in a SQLite database.
```
metcoder95, line 21: Shall we add a note for the `node:sqlite` dependency in case someone attempts to use it within Node.js <= 22?

lib/cache/sqlite-cache-store.js
```diff
@@ -0,0 +1,406 @@
...
+  close () {
+    this.#db.close()
+  }
+
+  /**
+   * @param {import('../../types/cache-interceptor.d.ts').default.CacheKey} key
+   * @returns {import('../../types/cache-interceptor.d.ts').default.GetResult | undefined}
+   */
+  get (key) {
+    if (typeof key !== 'object') {
```
metcoder95, line 180: do we also support arrays?

lib/cache/sqlite-cache-store.js
```diff
@@ -0,0 +1,406 @@
...
+    return result
+  }
+
+  /**
+   * @param {import('../../types/cache-interceptor.d.ts').default.CacheKey} key
+   * @param {import('../../types/cache-interceptor.d.ts').default.CacheValue} value
+   * @returns {Writable | undefined}
+   */
+  createWriteStream (key, value) {
+    if (typeof key !== 'object') {
```
metcoder95, line 213: We should check that at least the minimum required properties are there; otherwise it can lead to confusing errorrs

lib/cache/sqlite-cache-store.js
metcoder95: Depends, although indexed, as we are doing a range-like query it might hurt on the index lookup.

Have we considered maybe something like an LRU instead?

ronag (changes requested):
Why not make the database primary key a hash of url + method + vary headers?

lib/cache/sqlite-cache-store.js
flakey5: Of cache keys? Currently no

flakey5:
> Why not make the database primary key a hash of url + method + vary headers?

That would still lead to the same state that I talked about here https://github.com/nodejs/undici/pull/3657#discussion_r1849032745

> I was experimenting with hashing the vary headers (sort properties alphabetically -> stringify -> crypto.createHash), however, it was a lot more complicated than just having two separate queries. It also would've only benefited writes, for reads we would still need to fetch each response and compare the vary headers like we already do

mcollina:
CI is very red :(

mcollina:
>> If we do want it, I think we might be able to get away with it being stored somewhere in the db and being modified only with a procedure.

> I believe sqlite has meta commands for that, @mcollina ?

I _think_ this would do it:

```sql
PRAGMA page_count;
PRAGMA page_size;
```

Then do `page_count * page_size`.

mcollina:
> Why not make the database primary key a hash of url + method + vary headers?

@ronag I think this is possible, but I would likely investigate it after this lands. I think we would need to bring in https://www.npmjs.com/package/safe-stable-stringify to be able to have "stable" JSONs.

ronag:
> > Why not make the database primary key a hash of url + method + vary headers?
> 
> @ronag I think this is possible, but I would likely investigate it after this lands. I think we would need to bring in https://www.npmjs.com/package/safe-stable-stringify to be able to have "stable" JSONs.



> > Why not make the database primary key a hash of url + method + vary headers?
> 
> @ronag I think this is possible, but I would likely investigate it after this lands. I think we would need to bring in https://www.npmjs.com/package/safe-stable-stringify to be able to have "stable" JSONs.

What's wrong with e.g:

```
function makeKey({ origin, path, vary }) {
  const hasher = crypto.createHash('sha256')
  hasher.update(origin.toLowerCase())
  hasher.update(path)
  hasher.update(vary.map(x => x.toLowerCase()).sort().join(','))
  return hasher.digest('hex')
}
```

mcollina:
That would be very slow unfortunately :(. Unless we really need it.

lib/cache/sqlite-cache-store.js
```diff
@@ -0,0 +1,380 @@
...
+            value.statusCode,
+            value.statusMessage,
+            value.rawHeaders ? JSON.stringify(stringifyBufferArray(value.rawHeaders)) : null,
+            value.etag ? value.etag : null,
+            value.vary ? JSON.stringify(value.vary) : null,
+            value.cachedAt,
+            value.staleAt,
+            value.deleteAt
+          )
+        }
```
ronag, line 272: What about https://www.sqlite.org/lang_upsert.html?

lib/cache/sqlite-cache-store.js
```diff
@@ -0,0 +1,449 @@
...
+  prune () {
+    const total = this.size()
+
+    if (total <= this.#maxDbEntries) {
+      return
+    }
+
+    const oldestCachedAt = this.#getOldCacheAtQuery.get(total - 1)?.cachedAt
+
+    const res = this.#deleteOldCacheAtQuery.run(oldestCachedAt)
```
ronag, line 348: Can't you do this in a single query?

```delete from XXX where id IN (SELECT id from XXX order by cachedAt DESC limit 20)```

Comment (ronag) [reply lib/cache/sqlite-cache-store.js:348]: Also I would prune at least 10% of elements to we do not have to run this so often...

Comment (ronag) [reply lib/cache/sqlite-cache-store.js:348]: I would also move `this.#deleteExpiredValuesQuery.run(now)` to here and have it first try that and see if it's sufficient and then do a drop oldest.

File: lib/cache/sqlite-cache-store.js [hunk 23]
```diff
@@ -0,0 +1,449 @@
...
+    this.#deleteByUrlQuery.run(this.#makeValueUrl(key))
+  }
+
+  /**
+   * This method is called to prune the cache when it exceeds the maximum number
+   * of entries. It removes half the entries in the cache, ordering them the oldest.
+   *
+   * @returns {Number} The number of entries removed
+   */
+  prune () {
```
ronag, line 339:
```suggestion
  #prune () {
```

lib/cache/sqlite-cache-store.js
flakey5: I'm not sure how we could incorporate that since the only unique key we have is the id

ronag (approved):

PR #3682, opened 2024-10-05
ci: pr-title-check

Uzlopak:
This adds a workflow, which we also use in fastify. Benefit is, that pull requests can be implicitly marked as fix or fix! and thus makes it easier to decide if we should backport the bug fix or not. It feels like that some PRs could be put into the v6 branch but we didnt because we dont kept track of which PRs would make sense.


KhafraDev (changes requested):
no thanks

PR #3684, opened 2024-10-05
test: use globalThis.Headers and skip if is missing

Uzlopak:
If you run those test cases with ` --no-experimental-fetch`  they will fail. They use on purpose the global HEaders and this PR makes it clear that they use globalThis.Headers and not by accident undici own Headers.


mcollina (approved):
Lgtm

PR #3686, opened 2024-10-05
test: split wpt test runs, use flags to ensure no globals are used

Uzlopak:
hopefully self explaining


Uzlopak:
On second thought probably not that important

PR #3691, opened 2024-10-06
test: make fetch test independent from internet connection

Uzlopak:
missed this test when wanting to make test run without internet


mcollina (approved):
lgtm

PR #3692, opened 2024-10-07
replace instanceof in brand checks with isPrototypeOf

KhafraDev:
As a side note, but not majorly important: isPrototypeOf is also resilient against Symbol.hasInstance shenanigans. The real benefit is needing less imports to perform brand checks now that it is unified.

### DIFF
--- a/lib/web/cookies/index.js
+++ b/lib/web/cookies/index.js
@@ -5,9 +5,7 @@ const { stringify } = require('./util')
 const { webidl } = require('../fetch/webidl')
 const { Headers } = require('../fetch/headers')
 
-const headersToBrandCheck = globalThis.Headers
-  ? [Headers, globalThis.Headers]
-  : [Headers]
+const brandChecks = webidl.brandCheckMultiple([Headers, globalThis.Headers].filter(Boolean))
 
 /**
  * @typedef {Object} Cookie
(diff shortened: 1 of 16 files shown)


lib/web/cookies/index.js
```diff
@@ -5,9 +5,7 @@ const { stringify } = require('./util')
 const { webidl } = require('../fetch/webidl')
 const { Headers } = require('../fetch/headers')
 
-const headersToBrandCheck = globalThis.Headers
-  ? [Headers, globalThis.Headers]
-  : [Headers]
+const brandChecks = webidl.brandCheckMultiple([Headers, globalThis.Headers].filter(Boolean))
```
Uzlopak, line 8: What if you run node with --no-experimental-fetch? Then globalThis.Headers will be undefined. Thus breaking the code.

Uzlopak (changes requested):
See comment

lib/web/cookies/index.js
KhafraDev: that's what the .filter(Boolean) is for

mcollina (approved):
lgtm

metcoder95 (approved):

lib/web/cookies/index.js
Uzlopak: Ah right

KhafraDev:
@mcollina I don't think this is semver-major?

PR #3696, opened 2024-10-07
fetch: add docs on why forbidden header names are not supported

trivikr:
## This relates to...

* Comment https://github.com/nodejs/undici/pull/3695#pullrequestreview-2350953699
* Fetch spec https://fetch.spec.whatwg.org/#concept-headers-append

## Changes

Documents why undici fetch does not support forbidden header names

### Features

N/A

### Bug Fixes

N/A

N/A

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/web/fetch/headers.js
+++ b/lib/web/fetch/headers.js
@@ -89,8 +89,9 @@ function appendHeader (headers, name, value) {
   // 1. Normalize value.
   value = headerValueNormalize(value)
 
-  // 2. If name is not a header name or value is not a
-  //    header value, then throw a TypeError.
+  // 2. If validating (name, value) for headers returns false, then return.
+  //    1. If name is not a header name or value is not a header value,
+  //    then throw a TypeError.
   if (!isValidHeaderName(name)) {
     throw webidl.errors.invalidArgument({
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

Uzlopak (changes requested):
Changes in comments, @KhafraDev ptal

KhafraDev:
I find the comments harder to read considering it merges two different spec functions (I understand they're out of date). I don't see any value in documenting why we don't implement forbidden headers either as it's well-known and [documented](https://github.com/nodejs/undici?tab=readme-ov-file#forbidden-and-safelisted-header-names) already.

Uzlopak:
Why do we deviate from the spec?

trivikr:
The source code is for https://fetch.spec.whatwg.org/#concept-headers-append
It's provided in "relates to" section, as well as existing documentation, i.e. line 83, in the source code.

The screenshot provided is for https://fetch.spec.whatwg.org/#headers-class

KhafraDev (changes requested):
This should not land per my comment

PR #3699, opened 2024-10-07
fix: restructure determineRequestsReferrer to match better spec

Uzlopak:
While I investigated if there were significant changes in esbuild, I found the following change:

Screenshot from 2024-10-08 01-30-40

So the comments were added because they were of  `/** */`  notation and not of `//` notation. While I fixed, I saw that the implementation is not matching 1:1 the spec text, like the order of the cases. So I adapted it accordingly.

### DIFF
--- a/lib/web/fetch/util.js
+++ b/lib/web/fetch/util.js
@@ -418,18 +418,37 @@ function determineRequestsReferrer (request) {
     referrerURL = referrerOrigin
   }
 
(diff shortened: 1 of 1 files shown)


KhafraDev (approved):
Is it obvious that I didn't originally implement this? 😜

Uzlopak:
Well atleast I get once i a while an approval for a fetch PR  😜

PR #3709, opened 2024-10-10
web: mark as uncloneable when possible

jazelly:
## Rationale

This tells node that the marked instances from undici are not cloneable, so that attempts to cloning those throw `DataCloneError`.

This is a follow-up work from https://github.com/nodejs/node/pull/55234

## Changes

Marked the following as uncloneable.
```
'FormData', 'Headers', 'Request', 'Response', 'MessageEvent', 'CloseEvent',  'EventSource', 'WebSocket'
```

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [x] In review
- [ ] Merge ready


jazelly:
I am hesitant to write a test here, as it would be like testing the behaviour of `structuredClone`, which should be placed at Node. I will do a follow up once this is bumped into node.

mcollina (approved):
lgtm

mcollina:
I really think having a test would be useful, add it.

ronag:
Why is Headers unclonable?

jazelly:
AFAIK, any cloneable that wants to be cloned in node runtime needs to 
1. implement `kClone`
2. implement `kDeserialize`
3. `markTransferMode()` as "cloneable"

One example you can find in node is https://github.com/nodejs/node/blob/00b2f07f9ddeb8ffd2fb2108b0ed9ffa81ea000d/lib/internal/webstreams/transfer.js#L49

I don't think the above are exposed, so undici classes are not cloneable in node from that sense.

ronag:
I didn't mean to close this. Sorry.

Uzlopak:
How can we access those symbols from userland?

KhafraDev:
> Why is Headers unclonable?

@ronag it's because webidl platform objects cannot be cloned:

> A JavaScript value value is a platform object if [Type](https://tc39.es/ecma262/#sec-ecmascript-data-types-and-values)(value) is Object and if value has a [[PrimaryInterface]] internal slot.

> [Platform objects](https://webidl.spec.whatwg.org/#dfn-platform-object) can be [serializable objects](https://html.spec.whatwg.org/multipage/structured-data.html#serializable-objects) if their [primary interface](https://webidl.spec.whatwg.org/#dfn-primary-interface) is decorated with the [Serializable] IDL [extended attribute](https://webidl.spec.whatwg.org/#dfn-extended-attribute). Such interfaces must also define the following algorithms:

Notice that the webidl for Headers does not include serializable:
```bikeshed
[[Exposed](https://webidl.spec.whatwg.org/#Exposed)=(Window,Worker)]
interface Headers {
  ...
}
````

File: lib/core/util.js [hunk 1]
```diff
@@ -816,6 +817,12 @@ function errorRequest (client, request, err) {
   }
 }
 
+function maybeMarkAsUncloneable (target) {
+  if (markAsUncloneable !== undefined) {
+    markAsUncloneable(target)
+  }
+}
```
KhafraDev, line 824: This should be added to `lib/web/fetch/webidl.js`

KhafraDev (changes requested):

jazelly:
I should've mentioned, this is more to tell Node to treat undici instances as a platform object, as this api will attach a private attribute that Node can recognize. Without it, Node will try to find the `kTransfer` method. If it's not present, it will return `{}`, which is why we had [the original issue](https://github.com/nodejs/node/issues/55120) on node side.

```
structuredClone(new Response()) // {}
```

I think whether/how we make some classes cloneable in undici should be a separate and broader discussion, since it might involve whether/how Node exposes its internal symbols.

jazelly:
I also added the test. It works fine on my local but won't be effective on current CI, as the API hasn't been released. The API is on `v22.10` release line, so I am happy to hold this PR till then if that's more applicable.

lib/core/util.js
jazelly: Should it stay as a separate function? Feels weird if I attach it to `webidl.util` as it's not part of the WebIDL spec.

KhafraDev: I think it is, at least in part, because the webidl declaration defines if an object can be serialized. It also defines platform objects. webidl.util.markAsUncloneable or webidl.markAsUncloneable are both fine.

lib/web/fetch/webidl.js
```diff
@@ -119,6 +120,12 @@ webidl.util.TypeValueToString = function (o) {
   }
 }
 
+webidl.util.maybeMarkAsUncloneable = function (target) {
+  if (markAsUncloneable !== undefined) {
+    markAsUncloneable(target)
+  }
+}
+
```
Uzlopak, line 128:
```suggestion
webidl.util.markAsUncloneable = markAsUncloneable
  ? markAsUncloneable
  : () => {}

```

Uzlopak (changes requested):
why "maybeMarkAsUncloneable"? If in the future all supported platforms have markAsUncloneable, then do we need to rename this  function and all the places where it is called?

More likely that we forget that in 3 years.

jazelly:
> More likely that we forget that in 3 years.

Good point. I updated the naming.

Uzlopak (approved):
RSLGTM

btw. cloneable or clonable, is the spelling right?

jazelly:
Yeah, it's correct. In node it's always `cloneable`.

KhafraDev (commented):
Does this also need to be done for EventSource, WebSocket, WebsocketStream, CloseEvent, ErrorEvent, MessageEvent, Cache, CacheStorage, and possibly others?

jazelly:
I think we should. I was focusing on the classes that are exposed to node only.

KhafraDev:
WebSocket, CloseEvent, and MessageEvent are exposed in node

jazelly:
Yes, they are updated within this PR. I think any class under `lib/web` should be marked so they can be treated as platform objects.

KhafraDev (approved):

mcollina (approved):
lgtm

mcollina:
Can you please fix linting?

KhafraDev (approved):

mcollina:
Tagging as backportable

mcollina:
@jazelly would you mind sending a backport commit with the instructions above?

jazelly:
I am happy to do it.

PR #3717, opened 2024-10-10
fix: data-url set extractValue of collectAnHTTPQuotedString by default to false

Uzlopak:


metcoder95 (approved):
CI doesn't seems happy?

PR #3719, opened 2024-10-11
feat: improve testing of http2

Uzlopak:
I am too lazy to build with without-ssl flag locally. I think the values we destructure when importing http2 are not critical. more critical is, that http2.connect is not replaced with some function which throws an Internal Error which would indicate that http2 is not supported. Probably we get only a "connect is not defined" error. when calling it.

### DIFF
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -141,6 +141,77 @@ jobs:
       - name: Run tests
         run: npm run test:javascript:without-intl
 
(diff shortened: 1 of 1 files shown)


mcollina:
seems CI is failing

Uzlopak:
I am still checking if this makes any sense. maybe a proxyquire would be more useful instead of using this approach.

mcollina:
I think we should just skip a bunch of tests if ssl is not available.

PR #3722, opened 2024-10-12
http2: extract listenHandlers and one bugfix

Uzlopak:

PR #3725, opened 2024-10-12
http: extract listeners from client-h1

Uzlopak:
analogue to #3722 this PR extracts the functions from the client function.


mcollina (approved):
lgtm

PR #3729, opened 2024-10-14
fix: dns interceptor error handling

ronag:
## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/interceptor/dns.js
+++ b/lib/interceptor/dns.js
@@ -220,37 +220,20 @@ class DNSDispatchHandler extends DecoratorHandler {
   }
 
   onError (err) {
-    switch (err.code) {
-      case 'ETIMEDOUT':
-      case 'ECONNREFUSED': {
-        if (this.#state.dualStack) {
-          // We delete the record and retry
-          this.#state.runLookup(this.#origin, this.#opts, (err, newOrigin) => {
-            if (err) {
-              return this.#handler.onError(err)
-            }
+    this.#state.deleteRecord(this.#origin)
 
-            const dispatchOpts = {
-              ...this.#opts,
-              origin: newOrigin
-            }
-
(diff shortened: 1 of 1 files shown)


mcollina (commented):
Can you fix the test?

lib/interceptor/dns.js
```diff
@@ -220,37 +220,20 @@ class DNSDispatchHandler extends DecoratorHandler {
...
-        this.#handler.onError(err)
-        break
+    if (this.#state.dualStack && (err.code === 'ETIMEDOUT' || err.code === 'ECONNREFUSED')) {
+      // We delete the record and retry
+      this.#state.runLookup(this.#origin, this.#opts, (err, newOrigin) => {
+        if (err) {
+          this.#handler.onError(err)
+        } else {
+          // TODO (fix): How to avoid infinite loop?
+          this.#dispatch({ ...this.#opts, origin: newOrigin }, this)
```
metcoder95, line 232: If `dualStack` I’d say we need to verify that we already attempted both families; if no `dualStack` we error after first attempt (as we do already)

metcoder95 (approved):

lib/interceptor/dns.js
ronag: @metcoder95 do you think you can follow up on that?

metcoder95: Sure 👍

ronag:
@metcoder95 there is no deleteRecord?

ronag:
@metcoder95 I have more opinions on this but unfortunately little time. If you are interested in more feedback on this interceptor; what do you think we do a 15 min call?

ronag:
https://github.com/nxtedition/nxt-undici/pull/5/files

PR #3733, opened 2024-10-14
[http-cache] follow up

Uzlopak:
follow up of http-cache pr

### DIFF
--- a/lib/handler/cache-handler.js
+++ b/lib/handler/cache-handler.js
@@ -6,7 +6,7 @@ const {
   assertCacheStoreType,
   parseCacheControlHeader,
   parseVaryHeader,
-  UNSAFE_METHODS
+  assertCacheMethods
 } = require('../util/cache')
 
 /**
@@ -17,6 +17,11 @@ class CacheHandler extends DecoratorHandler {
    * @type {import('../../types/cache-interceptor.d.ts').default.CacheStore}
    */
   #store
+
+  /**
+   * @type {import('../../types/cache-interceptor.d.ts').default.CacheMethods}
+   */
+  #methods
 
   /**
    * @type {import('../../types/dispatcher.d.ts').default.RequestOptions}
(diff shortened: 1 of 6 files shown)


lib/handler/cache-handler.js
```diff
@@ -75,14 +92,14 @@ class CacheHandler extends DecoratorHandler {
     )
 
     if (
-      UNSAFE_METHODS.includes(this.#req.method) &&
+      !this.#methods.includes(this.#requestOptions.method) &&
       statusCode >= 200 &&
       statusCode <= 399
     ) {
```
Uzlopak, line 98: This whole part seems strange. We do need to check

lib/handler/cache-handler.js
```diff
@@ -75,14 +75,14 @@ class CacheHandler extends DecoratorHandler {
     )
 
     if (
-      UNSAFE_METHODS.includes(this.#req.method) &&
+      !this.#methods.includes(this.#requestOptions.method) &&
```
Uzlopak, line 78: WHy did we check for unsafe methods? this whole if condition seems actually kind of dubious.

Uzlopak: I would have expected 

```js
    if (
      !this.#methods.includes(this.#requestOptions.method) ||
      (
        statusCode < 200 &&
        statusCode >= 400
      )
    ) {
```

Uzlopak: @flakey5 

The invalidation logic is kind of strange. Also what is the point of passing methods, when they are ignored anyway? So I guess actually methods would have been an array of safe methods?

Uzlopak: > A cache MUST invalidate the target URI ([Section 7.1](https://www.rfc-editor.org/rfc/rfc9110#section-7.1) of [[HTTP](https://www.rfc-editor.org/rfc/rfc9111.html#HTTP)]) when it receives a non-error status code in response to an unsafe request method (including methods whose safety is unknown).

So I guess my proposed change should be right, as the passed methods will be considered safe methods. Especially in regard of the last sentence in the spec, that methods whose safety is unknown is also covered.

lib/interceptor/cache.js
```diff
@@ -4,48 +4,40 @@ const util = require('../core/util')
...
-    }
-  } else {
-    globalOpts.methods = ['GET']
+  const globalOpts = {
+    store,
+    methods
   }
 
-  // Safe methods the user wants and unsafe methods
-  const methods = [...globalOpts.methods, ...UNSAFE_METHODS]
```
Uzlopak, line 39: why did me "mixed" with unsafe methods?

this would result in caching unsafe requests

Uzlopak:
@flakey5 PTAL

lib/interceptor/cache.js
flakey5: this allows the unsafe methods to get to the cache handler, where they'll cause a cache purge if the response is a success

metcoder95 (approved):

PR #3735, opened 2024-10-15
[http-cache]: handle cache invalidation of location and content-location headers

Uzlopak:
this is a follow up to the follow up.

### DIFF
--- a/lib/cache/memory-cache-store.js
+++ b/lib/cache/memory-cache-store.js
@@ -214,6 +214,20 @@ class MemoryCacheStore {
   }
 
   /**
+   * @param {string} origin
+   * @param {string} path
+   */
+  deleteByOriginAndPath (origin, path) {
+    const cachedPaths = this.#data.get(origin)
+    for (const key of cachedPaths.keys()) {
+      // not good to use startsWith here, but it's fine for now
+      if (key.startsWith(path)) {
+        cachedPaths.delete(key)
+      }
+    }
+  }
+
+  /**
    * Gets all of the requests of the same origin, path, and method. Does not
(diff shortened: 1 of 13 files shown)


ivan-tymoshenko:
@Uzlopak What do you think about making this method more generic for calling it outside of the undici package. I have two suggestions:
- pass an array of paths instead of one path. This might be useful for another cache store types if user wants to delete mutiple paths. These storages might perform a batch delete operation much faster than deleting it one by one.
- pass also an http method and not only the path. For most cases users would want to cache only GET/HEAD requests, but still someone might want to cache other methods, so it would be nice to have an ability to delete routes from the cache independenty

lib/cache/memory-cache-store.js
```diff
@@ -213,6 +213,20 @@ class MemoryCacheStore {
...
 
+  /**
+   * @param {string} origin
+   * @param {string} path
+   */
+  deleteByOriginAndPath (origin, path) {
+    const cachedPaths = this.#data.get(origin)
+    for (const key of cachedPaths.keys()) {
+      // not good to use startsWith here, but it's fine for now
+      if (key.startsWith(path)) {
```
ivan-tymoshenko, line 224: I would split the `key` here by the `:` because the `MemoryCacheStore` indeed stores keys in a format `path:method`, but this might be not true for other types of stores. So if someone will rely on this it might be a breaking change in the future.

Uzlopak:
@ivan-tymoshenko 
@ronag 

I guess this can be closed, because you took care of ideas?

mcollina:
I think so

PR #3739, opened 2024-10-16
fix: unsafe methods not causing cache purge

flakey5:
Fixes bug introduced in #3733 where unsafe methods no longer make it to the cache handler and cause a cache purge for an origin.

Also removes a duplicate test file.

cc @mcollina @Uzlopak


mcollina (approved):
lgtm

Uzlopak (changes requested):
have to review

lib/interceptor/cache.js
```diff
@@ -32,7 +32,7 @@ module.exports = (opts = {}) => {
 
   return dispatch => {
     return (opts, handler) => {
-      if (!opts.origin || !methods.includes(opts.method)) {
+      if (!opts.origin || (opts.method in util.safeHTTPMethods && !methods.includes(opts.method))) {
```
metcoder95, line 35: Why not `includes`?

Uzlopak:
I think there is a bigger issue here.

lib/interceptor/cache.js
```diff
@@ -32,7 +32,7 @@ module.exports = (opts = {}) => {
 
   return dispatch => {
     return (opts, handler) => {
-      if (!opts.origin || !methods.includes(opts.method)) {
+      if (!opts.origin || (util.safeHTTPMethods.includes(opts.method) && !methods.includes(opts.method))) {
```
Uzlopak, line 35: just delete this if block and your test will pass.

Uzlopak:
According to my assessment the whole logic is flawed.

Uzlopak:
provided an alternative PR #3739

Uzlopak:
What is the point of "methods" if you ignore it basically?

flakey5:
> What is the point of "methods" if you ignore it basically?

It isn't being ignored, [the check at the beginning of each request](https://github.com/flakey5/undici/blob/ab2d8ddbe979b6142e076ddbfded86e072b4fa21/lib/interceptor/cache.js#L35) is all it's needed for

Uzlopak:
I thought about it and I think the methods option makes not much sense.

lets say you have array safeMethods with values [A,B] and methods with values [B]

So method is 

A => [A,B].includes(A) && !([B].includes(A)) => true && true => true
B => [A,B].includes(B) && !([B].includes(B)) => true && false => false
C => [A,B].includes(C) && !([B].includes(C)) => false && true => false

But what if I want to add in the future e.g. PROPFIND, which should be in theory a safe method (also ignore the fact that we would throw now a TypeError)

I add methods to PROPFIND but it wont work, because it is not in the safeMethods Array. I would purge the cache even if I added PROPFIND to the methods array.

Uzlopak:
I think you partially convinced me.

test/cache-interceptor/interceptor.js
```diff
@@ -1,240 +0,0 @@
-'use strict'
```
Uzlopak, line 1: Why did you delete this test file?

Uzlopak:
We could safe the two includes by generating ONE array

const safeMethodsToNotCache = util.safeHTTPMethods.filter(method => methods.includes(method) === false)

safeMethods would be [A,B], methods would be [B], so the resulting safeMethodToNotCache would be [A]

then the table would look like:

A => [A].includes(A) => true
B => [A].includes(B) => false
C => [A].includes(C) => false

test/interceptors/cache.js
```diff
@@ -250,4 +250,55 @@ describe('Cache Interceptor', () => {
...
+    })
+
+    strictEqual(deleteByOriginCalled, false)
+
+    // Make sure unsafe methods cause a cache purge
+    await client.request({
+      origin: 'localhost',
+      method: 'DELETE',
+      path: '/'
+    })
```
Ethan-Arrowood, line 300: For coverage purposes, I'd like to see this repeated for all of the unsafe methods

Ethan-Arrowood (approved):
not sure why the `cache-interceptor/interceptor.js` was deleted. I'd prefer slightly more test coverage, but wont block on that. lgtm otherwise

flakey5:
@Uzlopak @Ethan-Arrowood the deleted test file was a duplicate - it still exists under `test/interceptors/cache.js`

Uzlopak:
please adapt the code regarding the one array includes check.

flakey5:
> But what if I want to add in the future e.g. PROPFIND

I do think something like the `safeMethods` option you were talking about would be valuable for this, but I'm not sure how well that behavior falls into the scope of this pr. This pr (and the original caching pr) was mainly targeting the methods specified as safe in the http semantics spec ([9.2.1](https://www.rfc-editor.org/rfc/rfc9110#section-9.2.1-3))

I do think it'd be a good idea to make an issue or pr to talk about supporting methods that were defined in extension specs though

Uzlopak (approved):
LGTM

PR #3742, opened 2024-10-16
fix: unsafe methods should result in cache purge

Uzlopak:
the problem is, that the cache invalidation logic is done in cache-handler.js. So it makes no sense to avoid calling the cache logic, if the method is unsafe. All requests with origin have to go through the cache logic.

@flakey5

### DIFF
--- a/lib/interceptor/cache.js
+++ b/lib/interceptor/cache.js
@@ -32,7 +32,7 @@ module.exports = (opts = {}) => {
 
   return dispatch => {
     return (opts, handler) => {
-      if (!opts.origin || !methods.includes(opts.method)) {
+      if (!opts.origin) {
         // Not a method we want to cache or we don't have the origin, skip
         return dispatch(opts, handler)
       }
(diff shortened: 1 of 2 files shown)


flakey5 (changes requested):
This allows safe methods to purge the cache if they aren't included in `methods`. While this isn't stated in the spec, only unsafe methods should cause cache purges imo

i.e. add
```javascript
await client.request({
  origin: 'localhost',
  method: 'HEAD',
  path: '/'
})

strictEqual(deleteByOriginCalled, false)
```
to the test _before_ the `DELETE` request and the test will fail

Uzlopak:
You implemented the interface and added the methods attribute. If this attribute is an artifact, than we should remove it.

Uzlopak:
I think it works as expected. Maybe methods should be renamed to safeMethods?

flakey5:
> Maybe methods should be renamed to safeMethods?

That still wouldn't fix the issue of safe methods being able to purge the cache?

Uzlopak:
closing for now

PR #3743, opened 2024-10-16
chore(docs): add request() example for conditionally reading the body

styfle:
## This relates to...

Documentation for dispatcher.request() 

## Rationale

It is very easy to cause a memory leak or even crash the process if you forget to handle the response body.

## Changes

A new example showing how to conditionally reading the body

### Features

N/A

### Bug Fixes

N/A

N/A

## Status

- [ ] Tested
- [x] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/docs/docs/api/Dispatcher.md
+++ b/docs/docs/api/Dispatcher.md
@@ -641,18 +641,13 @@ const { body, statusCode } = await client.request({
   method: 'GET'
 })
 
-body.on('error', console.error) // prevent process from crashing on error
-
 if (statusCode === 200) {
-  const buffer = await body.arrayBuffer()
-  return buffer
+  return await body.arrayBuffer()
 }
 
-for await (const _chunk of body) {
-  // force consumption of body to avoid memory leak
-}
+await body.dump()
+
 return null
-```
 
 ### `Dispatcher.stream(options, factory[, callback])`
 
--- a/docs/docs/api/Dispatcher.md
+++ b/docs/docs/api/Dispatcher.md
@@ -527,6 +527,7 @@ try {
   console.log('headers', headers)
   body.setEncoding('utf8')
   body.on('data', console.log)
+  body.on('error', console.error)
   body.on('end', () => {
     console.log('trailers', trailers)
   })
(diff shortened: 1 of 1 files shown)


docs/docs/api/Dispatcher.md
```diff
@@ -630,6 +631,29 @@ try {
...
+if (statusCode === 200) {
+  const buffer = await body.arrayBuffer()
+  return buffer
+}
+
+for await (const _chunk of body) {
+  // force consumption of body to avoid memory leak
+}
+return null
+```
```
ronag, line 655:
```suggestion
```js
const { body, statusCode } = await client.request({
  path: '/',
  method: 'GET'
})

if (statusCode === 200) {
  return await body.arrayBuffer()
}

await body.dump()

return null
```
```

mcollina (approved):
lgtm with @ronag fix

docs/docs/api/Dispatcher.md
styfle: @ronag @mcollina I think its important to keep the `body.on('error', console.error)` handler because any code added between the `client.request()` and the `await body` calls could throw and then it would be uncatchable.

Since this example is about conditional code, I think its best practice to keep it in there to avoid this footgun.

styfle: Furthermore, I don't think `await body.dump()` is sufficient.

It says it defaults to 131072 but what if the body is larger than that?

https://github.com/nodejs/undici/blob/8e025d1cf5a453ff158cb1bd40ff6929e1b27bb1/lib/api/readable.js#L254

Ethan-Arrowood (changes requested):
lgtm otherwise

docs/docs/api/Dispatcher.md
styfle: Friendly ping @mcollina @ronag

Ethan-Arrowood:
What about @styfle 's comment / question here: https://github.com/nodejs/undici/pull/3743#discussion_r1804918142 

Is dump sufficient if the body is larger than its limit?

Ethan-Arrowood (approved):
lgtm - but there is still the outstanding question about `.dump()` limits

docs/docs/api/Dispatcher.md
styfle: I created PR https://github.com/nodejs/undici/pull/3788

ronag: > handler because any code added between the client.request() and the await body calls could throw and then it would be uncatchable.

That would be incorrect code anyway...

> It says it defaults to 131072 but what if the body is larger than that?

It will close the connection if the body is larger.

PR #3747, opened 2024-10-18
feat: add optional methods to the cache store

ivan-tymoshenko:
The idea of this PR is to add some optional methods to the cache store that might not be used by the interceptor, but are useful for managing the cache. Right now it's hard to invalidate the cache from the outside of the package, because there is no way to know what values have been cached. 

Any ideas about the intercase for these methods are welcome. I'm opening it's as a draft because it's missing a docs at least.

### DIFF
--- a/lib/cache/memory-cache-store.js
+++ b/lib/cache/memory-cache-store.js
@@ -76,6 +76,29 @@ class MemoryCacheStore {
   }
 
   /**
+   * @returns {string[]}
+   */
+  getOrigins () {
+    return Array.from(this.#data.keys())
+  }
+
+  /**
+   * @param {string} origin
+   * @returns {{ method: string, path: string }[]}
+   */
+  getRoutesByOrigin (origin) {
+    const paths = this.#data.get(origin)
+    if (!paths) return []
+
+    let cachedRoutes = []
+    for (const cachedValue of paths.keys()) {
+      const [path, method] = cachedValue.split(':')
+      cachedRoutes.push({ method, path })
+    }
+    return cachedRoutes
+  }
+
+  /**
    * @param {import('../../types/dispatcher.d.ts').default.RequestOptions} req
    * @returns {import('../../types/cache-interceptor.d.ts').default.CacheStoreReadable | undefined}
    */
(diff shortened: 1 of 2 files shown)


lib/cache/memory-cache-store.js
```diff
@@ -75,6 +75,29 @@ class MemoryCacheStore {
     return this.#entryCount >= this.#maxEntries
   }
 
+  /**
+   * @returns {string[]}
+   */
+  getOrigins () {
```
mcollina, line 81: This should be async

lib/cache/memory-cache-store.js
```diff
@@ -75,6 +75,29 @@ class MemoryCacheStore {
...
+    return Array.from(this.#data.keys())
+  }
+
+  /**
+   * @param {string} origin
+   * @returns {{ method: string, path: string }[]}
+   */
+  getRoutesByOrigin (origin) {
```
mcollina, line 89: This should be async

mcollina (commented):
Looks good to me,

Please add types too.

types/cache-interceptor.d.ts
```diff
@@ -88,6 +92,10 @@ declare namespace CacheHandler {
 
     get isFull (): boolean
 
+    getOrigins (): Promise<string[]>
+
+    getRoutesByOrigin (origin: string): Promise<{ method: string, path: string }[]>
```
flakey5, line 97: nit but would prefer if the `{ method: string, path: string }` type was defined outside of the fn declaration so it can be referenced elsewhere if needed (incl user space)

i.e.
```typescript
export interface RouteByOrigin {
  method: string
  path: string
}

export interface CacheStore {
  getRoutesByOrigin(origin: string): Promise<RouteByOrigin[]>
  // ...
}

flakey5 (changes requested):
can you add type checks for the new methods in `assertCacheStore` please https://github.com/nodejs/undici/blob/30305063937a87e70561d2aeb5795788ed628ca2/lib/util/cache.js#L184

ivan-tymoshenko:
> can you add type checks for the new methods in `assertCacheStore` please
> 
> https://github.com/nodejs/undici/blob/30305063937a87e70561d2aeb5795788ed628ca2/lib/util/cache.js#L184

I'm not sure about that as these methods are optional, because they are not used by the undici package.

flakey5 (approved):

mcollina:
Ci does not look green

ivan-tymoshenko:
> Ci does not look green

It fails in the main with the same error.

ivan-tymoshenko:
I make it a draft for now. Not sure about the interface.

PR #3749, opened 2024-10-19
lib: add `nowAbsolute` to fast timers

flakey5:
This adds a `nowAbsolute` function to the fast timers lib that allows for fetching the approximate absolute timestamp. This is useful in places like the cache interceptor and cache stores, where we're currently calling `Date.now()`.

image

cc @mcollina

### DIFF
--- a/benchmarks/timers/now-absolute.mjs
+++ b/benchmarks/timers/now-absolute.mjs
@@ -1,0 +1,13 @@
+import { bench, group, run } from 'mitata'
+import timers from '../../lib/util/timers.js'
+
+group(() => {
+  bench('Date.now()', () => {
+    Date.now()
+  })
+  bench('nowAbsolute()', () => {
+    timers.nowAbsolute()
+  })
+})
+
+await run()
--- a/lib/util/timers.js
+++ b/lib/util/timers.js
@@ -22,6 +22,13 @@
 let fastNow = 0
 
 /**
+ * The fastNowAbsolute variable contains the rough result of Date.now()
+ *
+ * @type {number}
+ */
+let fastNowAbsolute = Date.now()
+
+/**
  * RESOLUTION_MS represents the target resolution time in milliseconds.
  *
  * @type {number}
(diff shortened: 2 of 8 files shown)


test/utils/timers.js
```diff
@@ -0,0 +1,9 @@
+const { test, describe } = require('node:test')
```
flakey5, line 1: are there no tests for the fast timers api already?

test/utils/timers.js
```diff
@@ -0,0 +1,9 @@
...
+const { nowAbsolute } = require('../../lib/util/timers')
+const { equal } = require('assert')
+
+describe('Fast Timers', () => {
+  test('nowAbsolute', () => {
+    equal(Date.now(), nowAbsolute())
```
KhafraDev, line 7: this looks like it'll be extremely flaky

ronag (changes requested):
This can't use the no stall timer.

ronag:
Add another variable called fastNowAbsolute which is updated to date.now in every refresh

lib/util/timers.js
```diff
@@ -21,6 +21,13 @@
  */
 let fastNow = 0
 
+/**
+ * The fastNowAbsolute variable contains the rough result of Date.now()
+ *
+ * @type {number}
+ */
+let fastNowAbsolute = 0
```
ronag, line 29:
```suggestion
let fastNowAbsolute = Date.now()
```

benchmarks/timers/now-absolute.mjs
```diff
@@ -0,0 +1,13 @@
+import { bench, group, run } from 'mitata'
```
Uzlopak, line 1: kind of trivial benchmark.

mcollina:
CI is very red

mcollina:
Still very red

ronag (approved):

mcollina:
This now conflicts

PR #2823, opened 2024-02-23
remove all fetchParam event handlers

KhafraDev:
Actually fixes #1711

we could also probably have fetchParams stop relying on EventEmitter if we want to microoptimize 🤷

PR #3754, opened 2024-10-21
fix aborting Streams

epistemancering:
## This relates to...
https://github.com/nodejs/undici/issues/1711
## Rationale
The simple fix to the memory issue caused a worse issue. Common patterns that involve catching AbortErrors, such as retrying slow requests, silently fail if the signal unluckily happens to be aborted during the read. A proper fix to the memory issue will likely be more involved.
## Changes
revert https://github.com/nodejs/undici/pull/2823/commits/49e33a87411b003705e377a8d5ea6d2d1acef108
### Bug Fixes
Aborting a Stream doesn't throw.
```javascript
const abortController = new AbortController

;(async () => {
    const readStream = (await fetch("https://www.googleapis.com/customsearch/v1", { signal: abortController.signal })).arrayBuffer()

    setTimeout(() => {
        abortController.abort()
    })

    await readStream
})()
```
## Status
- [x] Tested
- [x] Review ready


KhafraDev (changes requested):
This is missing a test and is replacing one issue for another. I will take a look in a few days.

KhafraDev:
I need a test case that doesn't use a third party endpoint to work on it. The case you provided doesn't abort in node or Deno, and I cannot test in browsers because of CSP.

The correct fix, afaict, is to remove the *previous* listener if > 1 exists (such that 1 listener always exists) or to only add a listener if there are none.

epistemancering:
```javascript
import * as http from "http"

const abortStream = async () => {
    const abortController = new AbortController
    const readStream = (await fetch("http://localhost", { signal: abortController.signal })).arrayBuffer()
    abortController.abort()
    console.log(await readStream)
}

http.createServer((request, response) => {
    if (request.url?.[1]) {
        response.setHeader("content-type", "text/html")

        response.end(`<!doctypehtml>
            <html>
                <body>
                    <script>
                        (${abortStream})()
                    </script>
                </body>
            </html>
        `)
    } else {
        response.end(new Uint8Array(20000))
    }
}).listen(80, abortStream)
```

KhafraDev (approved):

KhafraDev:
Thank you!

PR #3758, opened 2024-10-21
feat: cache etag support

### LINKED ISSUE #3718: Easier way to attach `ETag` to `If-None-Match` after #3562
## This would solve...

#3562 introduces HTTP Caching (RFC9110) to undici, which also implements HTTP 304 with `If-Modified-Since`. But it seems that `ETag` is missing.

IMHO #3562 should not be delayed anymore, this could be implemented after #3562 is merged.

The `undici` could check the stored `ETag` header from the cache store and use that to negotiate with the `If-None-Match` in future requests. If the server returns HTTP 304, undici would then return the cached (transparent) HTTP 200 response to the client (the same behavior of `fetch` in browsers).

## Additional context

#3562 

https://github.com/JakeChampion/fetch/issues/241

`ETag` is described in RFC9110 here: https://httpwg.org/specs/rfc9110.html#field.etag
`If-None-Match` is described in RFC9110 here: https://httpwg.org/specs/rfc9110.html#field.if-none-match
HTTP 304 is described in RFC9110 here: https://httpwg.org/specs/rfc9110.html#status.304
The handling of HTTP 304 is described in RFC9110 here: https://httpwg.org/specs/rfc9111.html#freshening.responses

flakey5:
Re http caching, adds support for etags for when we need to revalidate a cached response.

No breaking changes to the public api. the `etag` property is nullable so a cache store _could_ just not implement it and it will be fine. However, I did add it to the tests as if it was required just so we can ensure the builtin cache stores fully support the behavior.

This only checks if the etag exists in the response's headers and ignores it if it's in the trailers for simplicity sakes (as allowed by the spec in [8.8.3](https://www.rfc-editor.org/rfc/rfc9110.html#section-8.8.3-8))

Closes #3718

### DIFF
--- a/lib/handler/cache-handler.js
+++ b/lib/handler/cache-handler.js
@@ -4,7 +4,8 @@ const util = require('../core/util')
 const DecoratorHandler = require('../handler/decorator-handler')
 const {
   parseCacheControlHeader,
-  parseVaryHeader
+  parseVaryHeader,
+  isEtagUsable
 } = require('../util/cache')
 const { nowAbsolute } = require('../util/timers.js')
 
@@ -136,7 +137,10 @@ class CacheHandler extends DecoratorHandler {
         cacheControlDirectives
       )
 
-      this.#writeStream = this.#store.createWriteStream(this.#cacheKey, {
+      /**
+       * @type {import('../../types/cache-interceptor.d.ts').default.CacheValue}
+       */
+      const value = {
         statusCode,
         statusMessage,
         rawHeaders: strippedHeaders,
@@ -144,7 +148,13 @@ class CacheHandler extends DecoratorHandler {
         cachedAt: now,
         staleAt,
         deleteAt
-      })
+      }
+
+      if (typeof headers.etag === 'string' && isEtagUsable(headers.etag)) {
+        value.etag = headers.etag
+      }
+
+      this.#writeStream = this.#store.createWriteStream(this.#cacheKey, value)
 
       if (this.#writeStream) {
         const handler = this
--- a/lib/util/cache.js
+++ b/lib/util/cache.js
@@ -202,6 +202,48 @@ function parseVaryHeader (varyHeader, headers) {
 }
 
 /**
+ * Note: this deviates from the spec a little. Empty etags ("", W/"") are valid,
+ *  however, including them in cached resposnes serves little to no purpose.
+ *
+ * @see https://www.rfc-editor.org/rfc/rfc9110.html#name-etag
+ *
+ * @param {string} etag
+ * @returns {boolean}
+ */
+function isEtagUsable (etag) {
+  if (etag.length <= 2) {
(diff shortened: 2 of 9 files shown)


lib/handler/cache-handler.js
```diff
@@ -126,15 +126,25 @@ class CacheHandler extends DecoratorHandler {
...
         vary: varyDirectives,
         cachedAt: now,
         staleAt,
         deleteAt
-      })
+      }
+
+      // Check for the response's etag header, ignore if there's multiple
+      if (typeof headers.etag === 'string') {
+        value.etag = headers.etag
```
KhafraDev, line 144: Shouldn't etag be validated here? https://httpwg.org/specs/rfc9110.html#field.etag

metcoder95: and also account for weak etags

mcollina:
I think this is a bit too light on tests, and I'm not sure if this implementation is correct.

Can you add some tests where the response is a 304 thanks to if-none-match?

SukkaW:
> I think this is a bit too light on tests, and I'm not sure if this implementation is correct.
> 
> Can you add some tests where the response is a 304 thanks to if-none-match?

Also, we could test how undici responds to different server implementations. I myself have run into situations when locally stored strong ETag matches the server-sent one yet the server is returning 200 with the same ETag, instead of 304.

ronag (approved):

lib/util/cache.js
```diff
@@ -201,6 +201,48 @@ function parseVaryHeader (varyHeader, headers) {
...
+
+  if (etag[0] === '"' && etag[etag.length - 1] === '"') {
+    if (etag[1] === '"' || etag.startsWith('"W/')) {
+      // ETag: ""asd123"" or ETag: "W/"asd123"", kinda undefined behavior in the
+      //  spec. Some servers will accept these while others don't.
+      return false
+    }
+
+    // ETag: "asd123"
+    return true
```
metcoder95, line 228: `nit`:
```suggestion
    // ETag: ""asd123"" or ETag: "W/"asd123"", kinda undefined behavior in the
    //  spec. Some servers will accept these while others don't.
    // ETag: "asd123"
    return !(etag[1] === '"' || etag.startsWith('"W/'))
```

lib/util/cache.js
```diff
@@ -201,6 +201,48 @@ function parseVaryHeader (varyHeader, headers) {
...
+  }
+
+  if (etag.startsWith('W/"') && etag[etag.length - 1] === '"') {
```
metcoder95, line 231: `nit`: Similar suggestion as per my previous comment

metcoder95 (approved):

mcollina (approved):
lgtm

mcollina:
@metcoder95 I'm landing this as it has been around for a while.

mcollina:
@flakey5 this now conflicts

mcollina (approved):
lgtm

PR #3761, opened 2024-10-23
fix: http2 queueing

### LINKED ISSUE #3753: HTTP/2 GOAWAY event crashes application (Undici v6.20.1)
## Bug Description

In release 6.20.1, there appears to be a regression in http/2 support which crashes Node with a TypeError on handling HTTP/2 GoAway events. It appears to be an error introduced by [this change](https://github.com/nodejs/undici/commit/d6c44f3b49949dbdd66d9c56e729d503e6b09ddc) (specifically [here](https://github.com/nodejs/undici/blob/30305063937a87e70561d2aeb5795788ed628ca2/lib/dispatcher/client-h2.js#L201)), which adds a call to `util.errorRequest(client, request, err)` when `request` is undefined. To be honest, I'm not sure if the root of the problem is that `request` is undefined and it shouldn't be, or if that's an allowed scenario which just needs an undefined check.

The problem does not appear in 6.20.0, and I added a simple check for undefined before calling `util.errorRequest(client, request, err)` in 6.20.1 which prevents the app from crashing but I'm not familiar enough with the internals here to know if this is an acceptable solution or a deeper investigation is required to understand why `request` is undefined.

## Reproducible By

This is reproducible in my project/environment by simply waiting 6-7 minutes after issuing some http/2 requests. A 'goaway' event is eventually received and the app crashes. Unfortunately this is a large project in a private repo which I can't share. If it's not obvious to the maintainers what the problem is based on the description/screenshots here, I will be happy to try and create a simple shareable repro. 

## Expected Behavior

Receiving a goaway event on an http/2 session should not crash the application.

## Logs & Screenshots

Screenshot 2024-10-21 at 1 53 36 PM

## Environment

Node v22.10.0, Docker node:22-alpine image

metcoder95:
## This relates to...

Fixes #3753 

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -192,6 +192,7 @@ function onHttp2SessionGoAway (errorCode) {
   client[kHTTPContext] = null
 
   if (this[kHTTP2Session] !== null) {
+    this[kHTTP2Session].close()
     this[kHTTP2Session].destroy(err)
     this[kHTTP2Session] = null
   }
@@ -324,7 +325,9 @@ function writeH2 (client, request) {
       stream.close()
       // We delay the destroy to allow the stream to send the RST_STREAM frame
       queueMicrotask(() => util.destroy(stream, err))
+      // We move the running index to the next request
       client[kQueue][client[kRunningIdx]++] = null
+      client[kPendingIdx] = client[kRunningIdx]
       client[kResume]()
     }
 
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -303,7 +303,7 @@ function writeH2 (client, request) {
   }
 
   /** @type {import('node:http2').ClientHttp2Stream} */
-  let stream
+  let stream = null
 
   const { hostname, port } = client[kUrl]
 
@@ -321,13 +321,13 @@ function writeH2 (client, request) {
 
     if (stream != null) {
       util.destroy(stream, err)
+      client[kQueue][client[kRunningIdx]++] = null
+      client[kResume]()
     }
 
     // We do not destroy the socket as we can continue using the session
     // the stream gets destroyed and the session remains to create new streams
     util.destroy(body, err)
-    client[kQueue][client[kRunningIdx]++] = null
-    client[kResume]()
   }
 
   try {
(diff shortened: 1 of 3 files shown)


ronag (approved):

test/http2.js
```diff
@@ -1261,10 +1268,130 @@ test('#2364 - Concurrent aborts', async t => {
...
+
+test('#2364 - Concurrent aborts (2nd variant)', async t => {
+  const server = createSecureServer(pem)
+  let counter = 0
+
+  server.on('stream', (stream, headers, _flags, rawHeaders) => {
+    counter++
+
+    // TODO: fix this
+    // While destroyed, the stream is not handling correctly the goaway
```
metcoder95, line 1353:
```suggestion
```

lib/dispatcher/client-h2.js
```diff
@@ -318,14 +321,19 @@ function writeH2 (client, request) {
     util.errorRequest(client, request, err)
 
     if (stream != null) {
-      util.destroy(stream, err)
+      // On Abort, we close the stream to send RST_STREAM frame
+      stream.close()
+      // We delay the destroy to allow the stream to send the RST_STREAM frame
+      queueMicrotask(() => util.destroy(stream, err))
```
ronag, line 327:
```suggestion
      setImmediate(() => util.destroy(stream, err))
```

lib/dispatcher/client-h2.js
```diff
@@ -318,14 +321,19 @@ function writeH2 (client, request) {
     util.errorRequest(client, request, err)
 
     if (stream != null) {
-      util.destroy(stream, err)
+      // On Abort, we close the stream to send RST_STREAM frame
+      stream.close()
+      // We delay the destroy to allow the stream to send the RST_STREAM frame
+      setImmediate(() => util.destroy(stream, err))
```
metcoder95, line 327:
```suggestion
      queueMicrotask(() => util.destroy(stream, err))
```

metcoder95:
Can I have some support with the merge? For some reason, even if approved, I cannot merge it

Uzlopak:
Here my approval but without having reviewed it. Just be responsible with it :P

PR #3057, opened 2024-04-05
fix(H2): handle goaway properly

metcoder95:
## Changes

- **refactor: small consolidation of functions**
- **fix: handle goaway frame properly**
- **docs: add documentation about GOAWAY**

## Status

- [ ] I have read and agreed to the [Developer's Certificate of Origin][cert]
- [ ] Tested
- [ ] Benchmarked (**optional**)
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

[cert]: https://github.com/nodejs/undici/blob/main/CONTRIBUTING.md

Issue #3011, opened 2024-03-28
Cannot read properties of null (reading 'ref') using fetch with HTTP/2 enabled and assertion error


st3ffgv4:
## Bug Description

While making continuous HTTP/2 requests after certain amount of time (minutes),  happens that session goaway event is triggered, but meanwhile the close event is not fired and when a new request is pushed this error is thrown

<details><summary>Details</summary>
<p>

> 1711624898513 START
> (node:1108) [UNDICI-H2] Warning: H2 support is experimental, expect them to change at any time.
> (Use `node --trace-warnings ...` to show where the warning was created)
> 1711625507059 TypeError: fetch failed
>     at fetch (/home/runner/UNdici-http2-test/node_modules/undici/index.js:109:13)
>     at process.processTicksAndRejections (node:internal/process/task_queues:95:5)
>     at async Timeout._onTimeout (file:///home/runner/UNdici-http2-test/index.js:10:19) {
>   [cause]: TypeError: Cannot read properties of null (reading 'ref')
>       at writeH2 (/home/runner/UNdici-http2-test/node_modules/undici/lib/dispatcher/client-h2.js:371:11)
>       at Object.write (/home/runner/UNdici-http2-test/node_modules/undici/lib/dispatcher/client-h2.js:129:7)
>       at _resume (/home/runner/UNdici-http2-test/node_modules/undici/lib/dispatcher/client.js:604:50)
>       at resume (/home/runner/UNdici-http2-test/node_modules/undici/lib/dispatcher/client.js:529:3)
>       at Client.<computed> (/home/runner/UNdici-http2-test/node_modules/undici/lib/dispatcher/client.js:259:31)
>       at [dispatch] (/home/runner/UNdici-http2-test/node_modules/undici/lib/dispatcher/client.js:314:20)
>       at Client.Intercept (/home/runner/UNdici-http2-test/node_modules/undici/lib/interceptor/redirect-interceptor.js:11:16)
>       at Client.dispatch (/home/runner/UNdici-http2-test/node_modules/undici/lib/dispatcher/dispatcher-base.js:179:40)
>       at [dispatch] (/home/runner/UNdici-http2-test/node_modules/undici/lib/dispatcher/pool-base.js:143:28)
>       at Pool.dispatch (/home/runner/UNdici-http2-test/node_modules/undici/lib/dispatcher/dispatcher-base.js:179:40)
> }
> node:assert:399
>     throw err;
>     ^
> 
[...]

PR #3763, opened 2024-10-23
fix: emitting unknown error event

### LINKED ISSUE #3753: HTTP/2 GOAWAY event crashes application (Undici v6.20.1)

iiAku:
## This relates to...
Might fix #3753
related to #3011 as well, since the fix of that one added `errorRequest` through #3057 and fix another issue

## Rationale

This should fixes the following error, but I'm not sure if it's fine console.error this error or if we expect something else. Also I wonder wether we should check in the first place if request exist before accessing it, or this is expected to be catched in that case.

Let me know if we would like to handle that another way.

We are using undici dispatcher with pool client and this error occur.

This is occurring because when the `onError` from `errorRequest` in utils is called, then if request is undefined it got caught by the try catch and then we try to emit an error, but it seems emit does not expect such event.

```
function errorRequest (client, request, err) {
  try {
    request.onError(err)
    assert(request.aborted)
  } catch (err) {
    client.emit('error', err)
  }
}
```

image

```
node:events:496
      ^
TypeError: Cannot read properties of undefined (reading 'onError')
    at Object.errorRequest (/usr/src/app/node_modules/undici/lib/core/util.js:638:13)
    at ClientHttp2Session.onHTTP2GoAway (/usr/src/app/node_modules/undici/lib/dispatcher/client-h2.js:245:8)
    at ClientHttp2Session.emit (node:events:518:28)
    at Http2Session.onGoawayData (node:internal/http2/core:678:11)
    at Http2Session.callbackTrampoline (node:internal/async_hooks:130:17)
Emitted 'error' event on Client instance at:
    at Object.errorRequest (/usr/src/app/node_modules/undici/lib/core/util.js:641:12)
    at ClientHttp2Session.onHTTP2GoAway (/usr/src/app/node_modules/undici/lib/dispatcher/client-h2.js:245:8)
    [... lines matching original stack trace ...]
    at Http2Session.callbackTrampoline (node:internal/async_hooks:130:17)
Node.js v20.11.0
```

Undici v6.20.1

### Bug Fixes

should close #3753

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/core/util.js
+++ b/lib/core/util.js
@@ -850,7 +850,7 @@ function errorRequest (client, request, err) {
     request.onError(err)
     assert(request.aborted)
   } catch (err) {
-    client.emit('error', err)
+    console.error(err)
   }
 }
 


ronag (changes requested):
This doesn't fix the root issue. Just hides it.

iiAku:
> This doesn't fix the root issue. Just hides it.

I agree and I believe https://github.com/nodejs/undici/pull/3761 that one is addressing the root cause.

However, as shown in the screen are we purposely expecting to `client.emit('error', err)` as the signature does not seems to let doing that. If this is expected in this case, we can close that MR.

PR #3770, opened 2024-10-25
fix: dns interceptor ip ttl

luddd3:
## This relates to...

How long time to cache the dns records from the dns interceptor.

## Rationale

For how long time the dns records from the dns interceptor should live.

## Changes

Set timestamp once on each resolved record and compare the ttl to that timestamp. If the record is too old, remove it.

### Bug Fixes

N/A

## Status

- [x] Tested
- [S] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/interceptor/dns.js
+++ b/lib/interceptor/dns.js
@@ -55,7 +55,7 @@ class DNSInstance {
           return
         }
 
-        this.setRecords(origin, newOpts, addresses)
+        this.setRecords(origin, addresses)
         const records = this.#records.get(origin.hostname)
 
         const ip = this.pick(
@@ -176,12 +176,17 @@ class DNSInstance {
     return ip
   }
 
-  setRecords (origin, opts, addresses) {
+  setRecords (origin, addresses) {
     const timestamp = Date.now()
     const records = { records: { 4: null, 6: null } }
     for (const record of addresses) {
       record.timestamp = timestamp
-      record.ttl = opts.maxTTL
+      if (typeof record.ttl === 'number') {
+        // The record TTL is expected to be in ms
+        record.ttl = Math.min(record.ttl, this.#maxTTL)
+      } else {
+        record.ttl = this.#maxTTL
+      }
 
       const familyRecords = records.records[record.family] ?? { ips: [] }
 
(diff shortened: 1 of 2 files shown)


lib/interceptor/dns.js
```diff
@@ -171,24 +165,24 @@ class DNSInstance {
...
     }
 
-    ip.timestamp = timestamp
-
     this.lastIpFamily = newIpFamily
     return ip
   }
 
-  setRecords (origin, addresses) {
+  setRecords (origin, opts, addresses) {
```
metcoder95, line 179: No need to pass the opts, the options are already attached to instance properties

lib/interceptor/dns.js
```diff
@@ -171,24 +165,24 @@ class DNSInstance {
...
+    const timestamp = Date.now()
     const records = { records: { 4: null, 6: null } }
     for (const record of addresses) {
+      record.timestamp = timestamp
+      record.ttl = opts.maxTTL
```
metcoder95, line 184: If the record already contains a `ttl` let's preserve it, otherwise default to `maxTTL`.

On top, let's use the min from both values.

luddd3: I have added it.

How do you feel about expecting the record result to be in ms? `dns.lookup` doesn't usually provide ttl. `dns.resolve` does, but it's ttl, and other time properties, are in seconds.
https://nodejs.org/api/dns.html#dnsresolveanyhostname-callback

test/interceptors/dns.js
```diff
@@ -546,31 +550,13 @@ test('Should we handle TTL (4)', async t => {
...
-                address: addr.address,
-                ttl: opts.maxTTL,
-                family: addr.family
-              }
-
-              results.set(`${record.address}:${record.family}`, record)
-            }
-
-            cb(null, results.values())
-          }
```
mcollina, line 573: why was this deleted?

luddd3: I didn't think it adds any value to the test. Does it?

mcollina (approved):
lgtm

metcoder95 (approved):

PR #3771, opened 2024-10-25
fix: assume blocking unless HEAD

ronag:
This is a safer default.

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -143,7 +143,7 @@ class Request {
       ? method === 'HEAD' || method === 'GET'
       : idempotent
 
-    this.blocking = blocking == null ? false : blocking
+    this.blocking = blocking ?? this.method !== 'HEAD'
 
     this.reset = reset == null ? null : reset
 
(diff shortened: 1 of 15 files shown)


mcollina (commented):
Can you adjust the tests?

mcollina:
Does this affect our benchmarks?

ronag:
> Does this affect our benchmarks?

Yes, if they are using pipelining.

ronag:
Fixed benchmarks as well.

lib/core/request.js
```diff
@@ -143,7 +143,7 @@ class Request {
       ? method === 'HEAD' || method === 'GET'
       : idempotent
 
-    this.blocking = blocking == null ? false : blocking
+    this.blocking = blocking ?? this.method !== 'HEAD'
```
ronag, line 146: We could also check range request headers and/or range and/or content-length response header to dynamically determine this.

ronag:
With this change we could probably turn pipelining on by default?

ronag:
Failing tests seem to be flaky and unrelated

metcoder95 (approved):

ronag:
@mcollina ping

Issue #3773, opened 2024-10-25
RedirectHandler cleans headers too aggressively


joelschou:
## Bug Description

Undici explicitly supports [headers](https://undici.nodejs.org/#/docs/api/Dispatcher?id=parameter-undiciheaders) passed as a [Headers](https://developer.mozilla.org/en-US/docs/Web/API/Headers) object, but the `RedirectHandler` does not account for this object type in its internal [cleanRequestHeaders](https://github.com/nodejs/undici/blob/main/lib/handler/redirect-handler.js#L221) function.

## Reproducible By

1. Configure global dispatcher to use redirect interceptor
2. Use `undici.request` to send a request with headers constructed via `new Headers(...)`

   ```js
   import { Headers, request } from 'undici';

   await request( '/local/test/url', { headers: new Headers({ 'x-my-header': true }) } );
   ```

3. Your local test server should see `'x-my-header': 'true'` in the request headers
4. Your local test server should respond with a `3xx` statusCode and a `location: '/local/test/alt-url'` header
5. Undici should follow the redirect and make a request to the new location
6. Your local test server **WILL NOT** see `'x-my-header': 'true'` in the request headers

Compare to:

1. same as above
2. Send the request this way instead:

   ```js
   await request( '/local/test/url', { headers: { 'x-my-header': true } } );
   ```

3. same as above
4. same as above
5. same as above
6. Your local test server **WILL** see `'x-my-header': 'true'` in the request headers

## Expected Behavior

1. Headers on the initial request should be repeated on the redirect request(s) for all supported ways to pass headers

## Logs & Screenshots

I believe I've tracked this down to this logic inside `cleanRequestHeaders`:

```js
// ...
} else if (headers && typeof headers === 'object') {
    for (const key of Object.keys(headers)) {
      if (!shouldRemoveHeader(key, removeContent, unknownOrigin)) {
        ret.push(key, headers[key])
      }
    }
  }
// ...
```

For `headers` constructed with `new Headers(...)`, the `typeof headers === 'object'` condition will return `true`, but `Object.keys(...)` does not behave as expected on a `Headers` object. As a result, this for loop does not push any keys/values into `ret`.

## Environment

* macOS 15.0.1 (current)
* Node v20.16.0 (LTS, but two minors behind)

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

PR #3777, opened 2024-10-27
fix: handle Headers in RedirectHandler

iiAku:
## This relates to...

This should close https://github.com/nodejs/undici/issues/3773

## Status

- [x] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready


metcoder95 (changes requested):
Can you add tests for it?

lib/handler/redirect-handler.js
```diff
@@ -239,3 +240,4 @@ function cleanRequestHeaders (headers, removeContent, unknownOrigin) {
 }
 
 module.exports = RedirectHandler
+module.exports.cleanRequestHeaders = cleanRequestHeaders
```
metcoder95, line 243: Let's not do that if we are just exporting it to testing it out.

You can attempt to test the same on the [`redirect-request`](https://github.com/nodejs/undici/blob/main/test/redirect-request.js) test suite where you submit a `Headers` instance instead of the array/object way.

lib/handler/redirect-handler.js
```diff
@@ -227,9 +227,10 @@ function cleanRequestHeaders (headers, removeContent, unknownOrigin) {
       }
     }
   } else if (headers && typeof headers === 'object') {
-    for (const key of Object.keys(headers)) {
+    const entries = headers instanceof Headers ? headers.entries() : Object.entries(headers)
```
metcoder95, line 230:
```suggestion
    const entries = Object.getPrototypeOf(headers) === Headers ? headers.entries() : Object.entries(headers)
```

iiAku: Don't exactly understand why but that option does not work, so I stuck with the instanceof let me know if it's good to you.

lib/handler/redirect-handler.js
iiAku: Added tests in appropriate file

lib/handler/redirect-handler.js
metcoder95: Hmm, I think i know why but let's keep it as is then 👍

metcoder95 (approved):

lib/handler/redirect-handler.js
KhafraDev, line 230: this won't work for Maps, only Headers

ronag: maybe better to check for `Symbol.Iterable`

ronag: @iiAku can you create a follow up PR?

```js
const entries = utils.isIterable(headers) ? headers : Object.entries(headers)
```

KhafraDev: this also only works with the global Headers, not Headers imported from undici

iiAku: > @iiAku can you create a follow up PR?
> 
> ```js
> const entries = utils.isIterable(headers) ? headers : Object.entries(headers)
> ```

Yes I can look into this, thanks for reviewing raising that point @KhafraDev

PR #3778, opened 2024-10-28
fix: dns interceptor affinity

luddd3:
## This relates to...

DNS interceptor

## Rationale

* error was thrown when a single ip address was resolved with dual-stack: "Cannot read properties of null (reading 'offset')"
* `pick` always started on the second ip address if multiple were resolved
* affinity was ignored with dual-stack

## Changes

Always start pick with the first ip address when multiple are resolved (ipv4 before ipv6)

### Features

Prefer affinity family with dual-stack, fallback to other family

### Bug Fixes

Handle single ip address resolved with dual-stack

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/interceptor/dns.js
+++ b/lib/interceptor/dns.js
@@ -13,7 +13,6 @@ class DNSInstance {
   affinity = null
   lookup = null
   pick = null
-  lastIpFamily = null
 
   constructor (opts) {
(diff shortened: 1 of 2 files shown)


metcoder95 (approved):
Good catch!

PR #3779, opened 2024-10-28
use AbortSignal.any if possible

### LINKED ISSUE #3731: test/fetch/long-lived-abort-controller.js is very flaky
it has been failing consistently, ie https://github.com/nodejs/undici/actions/runs/11329027907/job/31503604247?pr=3423

tsctx:
[*AbortSignal.any*](https://nodejs.org/docs/latest/api/globals.html#static-method-abortsignalanysignals) exists from the [lowest version](https://github.com/nodejs/undici/blob/2de0f341f44ab33fa342fe7c43669cf458e93a2a/package.json#L132) currently supported, which simplifies the handling of AbortSignal. 
~Also, tests that are no longer needed as a result of this have been removed.~

### DIFF
--- a/lib/web/fetch/request.js
+++ b/lib/web/fetch/request.js
@@ -29,6 +29,9 @@ const { kConstruct } = require('../../core/symbols')
 const assert = require('node:assert')
 const { getMaxListeners, setMaxListeners, defaultMaxListeners } = require('node:events')
 
+// https://github.com/nodejs/node/commit/d4736060404726a24d4e52647b8c9b88914b8ddf
+const isFixedOrderAbortSignalAny = typeof AbortSignal.any === 'function' && util.nodeMajor >= 23
+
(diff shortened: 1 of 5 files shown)


KhafraDev (changes requested):
I believe there are a number of memory leaks related to AbortSignal.any.

edit: Yeah, there are.
https://github.com/nodejs/node/issues/54614
https://github.com/nodejs/node/issues/55328
https://github.com/nodejs/node/pull/55354 (which has only made it into v23.1.0 so far)

tsctx:
Let's wait until the fix patch lands on LTS.

KhafraDev:
There are two issues without patches yet and then we'll have to manage both the AbortSignal.any path and our own. Is there any benefit in using AbortSignal.any? I see that the issues mentioned that this PR would fix were removed.

PR #3788, opened 2024-10-30
feat!: change default dump maxSize/limit to infinity

styfle:
## This relates to...

- https://github.com/nodejs/undici/pull/3743#discussion_r1804929992

## Rationale

I think `dump()` should be able to dump the entire response body, not limited to a max size, you often don't know how big the response body will be.

## Changes

N/A

### Features

N/A

### Bug Fixes

N/A

- Change default maxSize from `1048576` (1024 * 1024) to `null` (Infinity)
- Change default limit from `131072` (128 * 1024) to `null` (Infinity)

## Status

- [ ] Tested
- [x] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/docs/docs/api/Dispatcher.md
+++ b/docs/docs/api/Dispatcher.md
@@ -978,7 +978,7 @@ const client = new Client("http://example.com").compose(
 The `dump` interceptor enables you to dump the response body from a request upon a given limit.
 
 **Options**
-- `maxSize` - The maximum size (in bytes) of the response body to dump. If the size of the request's body exceeds this value then the connection will be closed. Default: `1048576`.
+- `maxSize` - The maximum size (in bytes) of the response body to dump. If the size of the response's body exceeds this value then the connection will be closed. Default: `null`.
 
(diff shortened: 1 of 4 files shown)


ronag (changes requested):
If the response is too big it can cause significant delay before the response is dumped. IMHO it's better to just close the connection in those cases.

styfle:
@ronag Feel free to make any changes to this PR. 

I created this PR since there was no discussion happening on the original docs PR and a DM to @mcollina suggested creating a PR to change the value to infinity.

mcollina:
> If the response is too big it can cause significant delay before the response is dumped. IMHO it's better to just close the connection in those cases.

I agree, but then why do we have a `.dump()` method to begin with?

Calling `.dump()` with a limit is _dangerous_, as users would still need to consume the rest of stream or risk crashing their process.

ronag:
> > If the response is too big it can cause significant delay before the response is dumped. IMHO it's better to just close the connection in those cases.
> 
> I agree, but then why do we have a `.dump()` method to begin with?

We have it so that if a user aborts a request with a small response we consume it and re-use the connection rather than killing it and starting a new one for the next request.

mcollina:
Ah I misunderstood this. Then we can close this.

Issue #3714, opened 2024-10-10
Move to using common container for building WASM


mhdawson:
## This would solve...

The security wg has been looking at auditing build dependencies for Node.js. https://github.com/nodejs/security-wg/issues/1236

One area we thought that we could improve is how we build WASM.  Today it's done differently, in some cases with different tools, different versions of tools and the versions used may or may not be known/captured in a way that would allow us to be sure we could reproduce a build.

I've looked at creating a common container that would be shared across the dependencies that build WASM and which captures the versions of tools etc. In addition in using a shared container that we store in the project we know we can build with the same tools because they are in the container stored within the registry.

The approach is based on how undici was build, with the main change being to use a pre-existing container. I've already talked to maintainers of amaro and cjs-module-lexer which are the other deps that build WASM blobs and they seem willing to adopt the approach.

I still have a bit of work in terms of how the container would be published (as I believe GitHub has a migration from ghcr of some sort) and the repo mhdawson/wasm-builder would need to move under the nodejs org.

The PR to adopt in undici would look like: https://github.com/nodejs/undici/compare/main...mhdawson:undici:wasm-build-experiment

You can see what it looks like for the other deps from the links in - https://github.com/nodejs/security-wg/issues/1236#issuecomment-2378148384

The reason for this issue is to ask if undici would be willing to adopt the approach as well.

Uzlopak:
Why do we need to use your wasm-builder image?
Would the repo for this image be added to nodejs org?

mcollina:
I'm +1 but the image needs to be hosted in the project.

mhdawson:
> needs to be hosted in the org.

absolutely. I would move mhdawson/wasm-builder to nodejs/wasm-builder

mhdawson:
> Why do we need to use your wasm-builder image?

There are a number of benefits on using a common approach/container across the dependencies. 
* Fewer tools and different versions of tools used to build Node.js ( ie smaller supply chain)
* It will be easier to ensure that builds capture all information related to tools and versions used and that we do things in the way we believe it good for supply chain security and repeatability.
  * For example I had PR'd in a change to generate a file with the tools/versions into undici but later changes broke that
* By using a container that we store instead of building one on the fly we will be sure we have the exact environment available if we need to rebuild at a later time. For example, in the current undici build process if the external link to install a particular version of BINARYEN no longer exists we would not be able to rebuild in the same way.

[...]

Issue #3787, opened 2024-10-29
pretty much everything is flaky


KhafraDev:
https://github.com/nodejs/undici/actions/runs/11581098341/

6 runs failed despite none of the fails being related to the changes in the PR

- https://github.com/nodejs/undici/actions/runs/11581098341/job/32244649295#step:7:2342 (still... #3731)
- https://github.com/nodejs/undici/actions/runs/11581098341/job/32244649573#step:8:1709 (dns interceptor, @metcoder95)
- https://github.com/nodejs/undici/actions/runs/11581098341/job/32244650191#step:8:5278 (@Uzlopak you edited this last, fails on node v23)

Uzlopak:
--no-experimental-fetch is not allowed in node23. if i make that conditional, than the diagnostic channel will be set up twice. once by native fetch and once by undici. unresolvable

KhafraDev:
So those tests should be skipped in node v23?

metcoder95:
Yeah, Windows has been bringing troubles here and there; I'll check them out down the week

trivikr:
There are new flaky tests in undici in 2026, but this issue can be closed since we don't test on node 23 anymore.

PR #3791, opened 2024-10-31
chore: use common WASM builder

mhdawson:
## This relates to...

See discussion in https://github.com/nodejs/undici/issues/3714

Note that with the updated common/container the same WASM seems to be generated so there should be
no functional changes.

The container itself is build using an approach that was based off of what undici was doing.

## Rationale

 https://github.com/nodejs/undici/issues/3714 includes the rational and discussion agreeing that I should submit a PR.

## Changes

Updates the build process to use the Node.js common wasm-builder container.

## Status

- [X] Tested
- [ ] Documented
- [X] Review ready
- [ ] In review
- [ ] Merge ready


mhdawson:
Rebased

mhdawson:
Since the WASM did not change I don't think any of the failures are related to this PR.

This issue seems to say there is a lot of flakiness and probably confirms that - https://github.com/nodejs/undici/issues/3787

build/wasm.js
```diff
@@ -1,13 +1,14 @@
 'use strict'
 
+const WASM_BUILDER_CONTAINER = 'ghcr.io/nodejs/wasm-builder@sha256:975f391d907e42a75b8c72eb77c782181e941608687d4d8694c3e9df415a0970' // v0.0.9
```
metcoder95, line 3: Are there plans to provide regular updates to it?

mhdawson: I have been thinking a bit about that, I think we want to update regularly but not necessarily too often. At least a month before every new Node.js major but I could see more often making sense as well. 

The repo where it is maintained has release please and all of the automation to generate a new vesion so it's as simple as a PR to update the dockerfile used to build the container and once that lands the rest is pretty much automated.

@metcoder95 how often do you think an update would make sense ? How often was undici updating?

ronag (approved):

build/wasm.js
metcoder95: I cannot develop a cadence, as I'd say that should be up to the things added to the new version. If it contains security fixes, maybe faster than usual, otherwise we can make a check every time we prepare a new `undici`'s release.

@mcollina @ronag wdyt?

metcoder95 (approved):

PR #3805, opened 2024-11-06
fix: response error interceptor broken

luddd3:
## This relates to...

The response-error interceptor

## Rationale

The interceptor is currently broken since Request no longer accepts `throwOnError` as an option.

## Changes

* Expose the interceptor in the same manner as the others.
* Remove `throwOnError` as it is considered an invalid argument.
* Removed broken/incomplete tests

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready


metcoder95 (approved):

lib/interceptor/response-error.js
```diff
@@ -84,6 +84,10 @@ class Handler extends DecoratorHandler {
...
 
-module.exports = (dispatch) => (opts, handler) => opts.throwOnError
-  ? dispatch(opts, new Handler(opts, { handler }))
-  : dispatch(opts, handler)
+module.exports = () => {
+  return (dispatch) => {
+    return function Intercept (opts, handler) {
+      return dispatch(opts, new ResponseErrorHandler(opts, { handler }))
+    }
+  }
```
Uzlopak, line 92: Hmm, this makes me having stomach ache

Uzlopak:
@metcoder95 

When we are already touching this file massively, should we extract the handler into a new file and put it into the handler folder?

lib/interceptor/response-error.js
luddd3: Care to elaborate or give a suggestion?

metcoder95:
It depends on what we do with that handler.
If the handler remains used within the interceptor only, let's keep it close to it; otherwise, let's add it into a separate handler file for easier management.

mcollina (approved):
lgtm

PR #3814, opened 2024-11-08
feat!(H2): enable H2 by default

metcoder95:
## This relates to...

Enable HTTP/2 support by default.

## Rationale

After a chat with @mcollina, we have noticed that the most important pieces for HTTP/2 to work within Undici has been already implemented.
The number of issues and bugs has been reduced since the implementation of the feature, and now the HTTP/2 client is fully onboarded into the Undici's queue system.

Given that the adoption seems good and stable, I propose to enable the H2 by default for the upcoming major release.

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready


ronag (changes requested):
I'm ok with it being enabled by default for fetch. Though I don't think it provides any meaningful advantage outside of fetch.

metcoder95:
My thoughts on this is that it benefits mostly the use-cases that requires usually network calls to the internet (basically in the wild), as its where most of the HTTP/2 interactions happen.

Fetch sadly does not really benefits from it as is not totally capable of handling H2 headers accordingly (and based on its semantics it seems mostly modeled around H1 in mind).

The benefit I see on enabling it is for the uses cases previously mentioned, but it would beup to the server to decide which protocol to use.

On top, this will only have visible effects on TLS connections, as we do not support h2c; for instance in a good majority of cases H1 will prevail.

Maybe worth to document more of its intricacies just to put users aware?

ronag (approved):

mcollina:
> Fetch sadly does not really benefits from it as is not totally capable of handling H2 headers accordingly (and based on its semantics it seems mostly modeled around H1 in mind).

How? Are you saying that a `fetch()` call will fail if we enable this?

metcoder95:
> > Fetch sadly does not really benefits from it as is not totally capable of handling H2 headers accordingly (and based on its semantics it seems mostly modeled around H1 in mind).
> 
> How? Are you saying that a `fetch()` call will fail if we enable this?

Nop, fetch works with H2 regardless. What I am referring to with incompatibilities is that fetch does not handle several concepts of H2 (e.g. [pseudo headers](https://httpwg.org/specs/rfc7540.html#PseudoHeaderFields) handling, does not fail but we do not parse these headers for the Headers map)

mcollina:
> Nop, fetch works with H2 regardless. What I am referring to with incompatibilities is that fetch does not handle several concepts of H2 (e.g. [pseudo headers](https://httpwg.org/specs/rfc7540.html#PseudoHeaderFields) handling, does not fail but we do not parse these headers for the Headers map)

THis seems a bit of a blocker of enabling this by default. Can you make some example? And possibly open PRs to handle those.

metcoder95:
I remember we have a chat about that on Fetch some time back, along with @KhafraDev regarding the handling of the pseudo headers; we didn't do it back then due to the spec, but if we are willing to explore that, happy to address it

metcoder95:
Although, worth mentioning that often than not the pseudo headers are of not much usage for the client; but still have seen some interests in having them available

mcollina:
> I remember we have a chat about that on Fetch some time back, along with @KhafraDev regarding the handling of the pseudo headers; we didn't do it back then due to the spec, but if we are willing to explore that, happy to address it

Can you make an example? I _think_ that `fetch()` in the browser is capable of selecting HTTP/2 and having all the relevant headers parsed.

mcollina:
@metcoder95 could you implement these https://github.com/nodejs/undici/blob/2252554f43ceefb67089ec85db265c96c77bb14d/lib/dispatcher/client-h2.js#L535-L545? As for `push`, you should just disable it when connecting (https://nodejs.org/api/http2.html#settings-object).

metcoder95:
Sure, let me check that; I'll go back on these over the week, still catching up after holidays.

metcoder95:
Disregard my last point about `fetch`, it works as is 👍 

Opened #3854 for support of timeout and trailers. Will add some docs to this PR for what we discussed here

mcollina (approved):
lgtm

ronag (approved):

mcollina:
I've spoken with a few people, and they would not recommend using HTTP/2 in production and stick to HTTP/1.1.

I'm inclined to just flag it as stable for now, and reconsider flipping it on the next major.

metcoder95:
That SGTM, closing then

PR #3816, opened 2024-11-08
lib: more cache fixes

### LINKED ISSUE #3806: The cache handler should not invalidate all requests to an origin
When the cache-handler recevies a succesful response from one of "unsafe" http methods such as POST/PUT/DELETE... it invalidates _all_ other cached responses to this origin. 

https://github.com/nodejs/undici/blob/main/lib/handler/cache-handler.js#L79

RFC says:

`
A cache MUST invalidate the target URI ([Section 7.1](https://www.rfc-editor.org/rfc/rfc9110#section-7.1) of [[HTTP](https://www.rfc-editor.org/rfc/rfc9111.html#HTTP)]) when it receives a non-error status code in response to an unsafe request method (including methods whose safety is unknown).
`

By "target URI" here I understand the the full URL and not just an origin (correct me here if I;m wrong). For example if a client receives a succesful response for the `DELETE http://example.com/foo/bar` the cache hanlder can delete the `GET http://example.com/foo/bar`, but not all other cached responses to the `http://example.com`.

IMHO With a current behaviour the cache seems useless, because a one regulary called POST requests keeps your cache almost always empty.

flakey5:
Closes #3806

- Removes `CacheStoreReadable` and `CacheStoreWritable`, replaces w/ just `Readable` and `Writable`
  - Reasoning: api simplicity
- Renames `CacheStoreValue` type to `CachedResponse`
- Removes `CacheStore#createReadStream` function (see one below on reasoning)
- Adds `CacheStore#get` function that acts similarly to `createReadStream`
  - Reasoning: api simplicity & niceness. Returns a `GetResult` which separates the `CacheResponse` and `Readable` for the body
- Simplifies `MemoryCacheStore`
  - Inline readable & writables
  - `readLock` and `writeLock` condensed into just `locked` that only stops value from being read
- Introduces `CacheKey` type that we pass into the cache handler, cache stores
  - Replaces us passing in the full request options
  - Reasoning: cache stores don't need the full thing, guides them to be more spec-compliant
- Various internal type fixes
- Fixes regarding a header being present multiple times
  - Stripping headers would result in them being re-encoded in the wrong format
  - Cache control & vary header parsings
- Makes `isFull` optional
  - Reasoning: some cache stores can't get full / don't have a max. This only exists as a very small optimization
- Removes caching of trailers
  - Reasoning: simplicity sakes, ideally this will be added back later
- Replace `deleteByOrigin` with `deleteByKey`
  - Reasoning: #3806
- Adds `CachedValue` type
  - Separates the `body` stream and the properties in `CachedResponse`
  - Allows the stream to be undefined for HEAD requests since there's no body

cc @mcollina @ronag

### DIFF
--- a/docs/docs/api/CacheStore.md
+++ b/docs/docs/api/CacheStore.md
@@ -25,13 +25,14 @@ The store must implement the following functions:
 Optional. This tells the cache interceptor if the store is full or not. If this is true,
 the cache interceptor will not attempt to cache the response.
 
-### Function: `getValueByKey`
+### Function: `get`
 
 Parameters:
 
 * **req** `Dispatcher.RequestOptions` - Incoming request
 
-Returns: `GetValueByKeyResult | Promise<GetValueByKeyResult | undefined> | undefined` - If the request is cached, a readable for the body is returned. Otherwise, `undefined` is returned.
+Returns: `GetResult | Promise<GetResult | undefined> | undefined` - If the request is cached, the cached response is returned. If the request's method is anything other than HEAD, the response is also returned.
+If the request isn't cached, `undefined` is returned.
 
 Response properties:
 
--- a/docs/docs/api/CacheStore.md
+++ b/docs/docs/api/CacheStore.md
@@ -2,7 +2,8 @@
 
 A Cache Store is responsible for storing and retrieving cached responses.
 It is also responsible for deciding which specific response to use based off of
-a response's `Vary` header (if present).
+a response's `Vary` header (if present). It is expected to be compliant with
+[RFC-9111](https://www.rfc-editor.org/rfc/rfc9111.html).
 
 ## Pre-built Cache Stores
 
@@ -21,27 +22,32 @@ The store must implement the following functions:
 
 ### Getter: `isFull`
 
-This tells the cache interceptor if the store is full or not. If this is true,
+Optional. This tells the cache interceptor if the store is full or not. If this is true,
 the cache interceptor will not attempt to cache the response.
 
-### Function: `createReadStream`
+### Function: `getValueByKey`
 
 Parameters:
 
 * **req** `Dispatcher.RequestOptions` - Incoming request
 
-Returns: `CacheStoreReadable | Promise<CacheStoreReadable | undefined> | undefined` - If the request is cached, a readable for the body is returned. Otherwise, `undefined` is returned.
+Returns: `GetValueByKeyResult | Promise<GetValueByKeyResult | undefined> | undefined` - If the request is cached, a readable for the body is returned. Otherwise, `undefined` is returned.
+
+Response properties:
+
+* **response** `CachedResponse` - The cached response data.
+* **body** `Readable | undefined` - The response's body.
 
 ### Function: `createWriteStream`
 
 Parameters:
 
 * **req** `Dispatcher.RequestOptions` - Incoming request
-* **value** `CacheStoreValue` - Response to store
+* **value** `CachedResponse` - Response to store
 
-Returns: `CacheStoreWriteable | undefined` - If the store is full, return `undefined`. Otherwise, return a writable so that the cache interceptor can stream the body and trailers to the store.
+Returns: `Writable | undefined` - If the store is full, return `undefined`. Otherwise, return a writable so that the cache interceptor can stream the body and trailers to the store.
 
-## `CacheStoreValue`
+## `CachedResponse`
 
 This is an interface containing the majority of a response's data (minus the body).
 
@@ -55,15 +61,11 @@ This is an interface containing the majority of a response's data (minus the bod
 
 ### Property `rawHeaders`
 
-`(Buffer | Buffer[])[]` - The response's headers.
-
-### Property `rawTrailers`
-
-`string[] | undefined` - The response's trailers.
+`Buffer[]` - The response's headers.
 
 ### Property `vary`
 
-`Record<string, string> | undefined` - The headers defined by the response's `Vary` header
+`Record<string, string | string[]> | undefined` - The headers defined by the response's `Vary` header
 and their respective values for later comparison
 
 For example, for a response like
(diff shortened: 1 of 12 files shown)


docs/docs/api/CacheStore.md
```diff
@@ -21,27 +22,32 @@ The store must implement the following functions:
 
 ### Getter: `isFull`
 
-This tells the cache interceptor if the store is full or not. If this is true,
+Optional. This tells the cache interceptor if the store is full or not. If this is true,
 the cache interceptor will not attempt to cache the response.
 
-### Function: `createReadStream`
+### Function: `getValueByKey`
```
ronag, line 28: Any reason it can't be just `get`?

lib/handler/cache-handler.js
```diff
@@ -176,7 +175,7 @@ class CacheHandler extends DecoratorHandler {
   onData (chunk) {
     let paused = false
 
-    if (this.#writeStream) {
+    if (this.#writeStream && this.#cacheKey.method !== 'HEAD') {
```
ronag, line 178: If HEAD then shouldn't writeStream just be null?

lib/handler/cache-handler.js
```diff
@@ -295,7 +297,7 @@ function determineStaleAt (now, headers, cacheControlDirectives) {
     return now + (maxAge * 1000)
   }
 
-  if (headers.expire) {
+  if (headers.expire && typeof headers.expire === 'string') {
     // https://www.rfc-editor.org/rfc/rfc9111.html#section-5.3
     return now + (Date.now() - new Date(headers.expire).getTime())
```
ronag, line 302: Might want to make sure expire is a valid date?

lib/interceptor/cache.js
```diff
@@ -47,27 +47,30 @@ module.exports = (opts = {}) => {
...
+       * @param {import('../../types/cache-interceptor.d.ts').default.CachedResponse} value
        */
       const respondWithCachedValue = (stream, value) => {
-        assert(!stream.destroyed, 'stream should not be destroyed')
-        assert(!stream.readableDidRead, 'stream should not be readableDidRead')
-
+        assert(!stream || !stream.destroyed, 'stream should not be destroyed')
+        assert(!stream || !stream.readableDidRead, 'stream should not be readableDidRead')
         try {
           stream
```
ronag, line 72: Can stream be nully? When?

lib/interceptor/cache.js
```diff
@@ -76,21 +79,22 @@ module.exports = (opts = {}) => {
...
+                handler.onComplete([])
               }
             })
 
           if (typeof handler.onConnect === 'function') {
             handler.onConnect((err) => {
-              stream.destroy(err)
+              if (stream) {
+                stream.destroy(err)
+              }
```
ronag, line 94:
```suggestion
              stream?.destroy(err)
```

lib/interceptor/cache.js
```diff
@@ -103,13 +107,15 @@ module.exports = (opts = {}) => {
...
+            if (handler.onHeaders(value.statusCode, rawHeaders, () => stream?.resume(), value.statusMessage) === false) {
+              stream?.pause()
             }
           }
 
           if (opts.method === 'HEAD') {
-            stream.destroy()
+            if (typeof handler.onComplete === 'function') {
+              handler.onComplete([])
+            }
```
ronag, line 118: either destroy stream or `assert(!stream)`

lib/interceptor/cache.js
```diff
@@ -162,22 +168,23 @@ module.exports = (opts = {}) => {
...
+      if (typeof result.then === 'function') {
+        result.then((result) => {
+          if (!result) {
+            // Request isn't cached
+            return dispatch(opts, new CacheHandler(globalOpts, cacheKey, handler))
           }
-        })
+
+          handleStream(result)
+        }).catch(handler.onError)
```
ronag, line 185:
```suggestion
        }).catch(err => handler.onError(err))
```

ronag: If handleStream throws we should crash, that should never throw...

lib/util/cache.js
```diff
@@ -4,6 +4,31 @@ const {
...
+   */
+  const cacheKey = {
+    origin: opts.origin.toString(),
+    method: opts.method,
+    path: opts.path
+  }
+
+  if (opts.headers) {
+    cacheKey.headers = opts.headers
+  }
```
ronag, line 27:
```suggestion
  const cacheKey = {
    origin: opts.origin.toString(),
    method: opts.method,
    path: opts.path,
    headers: opts.headers
  }
```

lib/core/util.js
```diff
@@ -516,6 +478,24 @@ function parseRawHeaders (headers) {
...
+    throw new TypeError('expected headers to be an array')
+  }
+
+  const rawHeaders = new Array(headers.length)
+
+  for (let i = 0; i < headers.length; i++) {
+    rawHeaders[i] = Buffer.from(headers[i])
+  }
+
+  return rawHeaders
```
ronag, line 496:
```suggestion
  return headers.map(x => Buffer.from(x))
```

lib/interceptor/cache.js
flakey5: On HEAD requests

docs/docs/api/CacheStore.md
flakey5: Personal preference, but either or work for me

lib/interceptor/cache.js
KhafraDev: the asserts already catch it

docs/docs/api/CacheStore.md
ronag: Seem unnecessarily verbose to me...

lib/interceptor/cache.js
ronag: This is broken though like this... onConnect cannot abort the request

DTrombett:
Asking here as it seems related, but why is the `public` directive required? Shouldn't it be required only with authenticated requests?
https://github.com/nodejs/undici/blob/c94fda14b0c58cd1083c6efb7a49868bab4d835f/lib/handler/cache-handler.js#L238-L245

flakey5:
> Asking here as it seems related, but why is the public directive required? Shouldn't it be required only with authenticated requests?

Will fix, thanks!

lib/interceptor/cache.js
flakey5: Wdym?

ronag (approved):

mcollina (approved):
lgtm

PR #3819, opened 2024-11-10
fix: handle undici Headers and Maps in redirect-handler

iiAku:
## This relates to...

#3777

This is a follow up MR addressing that: https://github.com/nodejs/undici/pull/3777#discussion_r1833340766

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/handler/redirect-handler.js
+++ b/lib/handler/redirect-handler.js
@@ -228,7 +228,7 @@ function cleanRequestHeaders (headers, removeContent, unknownOrigin) {
       }
     }
   } else if (headers && typeof headers === 'object') {
-    const entries = util.isIterable(headers) ? headers : Object.entries(headers)
+    const entries = typeof headers[Symbol.iterator] === 'function' ? headers : Object.entries(headers)
     for (const [key, value] of entries) {
       if (!shouldRemoveHeader(key, removeContent, unknownOrigin)) {
         ret.push(key, value)
--- a/lib/handler/redirect-handler.js
+++ b/lib/handler/redirect-handler.js
@@ -228,7 +228,8 @@ function cleanRequestHeaders (headers, removeContent, unknownOrigin) {
       }
     }
   } else if (headers && typeof headers === 'object') {
-    const entries = headers instanceof Headers ? headers.entries() : Object.entries(headers)
+    const iterableHeaderEntries = util.isIterable(headers) ? headers : Object.entries(headers)
+    const entries = headers instanceof Headers ? headers.entries() : iterableHeaderEntries
     for (const [key, value] of entries) {
       if (!shouldRemoveHeader(key, removeContent, unknownOrigin)) {
         ret.push(key, value)
--- a/lib/handler/redirect-handler.js
+++ b/lib/handler/redirect-handler.js
@@ -228,8 +228,7 @@ function cleanRequestHeaders (headers, removeContent, unknownOrigin) {
       }
     }
   } else if (headers && typeof headers === 'object') {
-    const iterableHeaderEntries = util.isIterable(headers) ? headers : Object.entries(headers)
-    const entries = headers instanceof Headers ? headers.entries() : iterableHeaderEntries
+    const entries = util.isIterable(headers) ? headers.entries() : Object.entries(headers)
     for (const [key, value] of entries) {
       if (!shouldRemoveHeader(key, removeContent, unknownOrigin)) {
         ret.push(key, value)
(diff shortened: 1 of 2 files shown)


lib/handler/redirect-handler.js
```diff
@@ -228,7 +228,8 @@ function cleanRequestHeaders (headers, removeContent, unknownOrigin) {
       }
     }
   } else if (headers && typeof headers === 'object') {
-    const entries = headers instanceof Headers ? headers.entries() : Object.entries(headers)
+    const iterableHeaderEntries = util.isIterable(headers) ? headers : Object.entries(headers)
+    const entries = headers instanceof Headers ? headers.entries() : iterableHeaderEntries
```
ronag, line 232:
```suggestion
    const entries = util.isIterable(headers) ? headers : Object.entries(headers)
```

iiAku: I'm afraid it won't work as native Headers is iterable as well.

image

Meanwhile let me know if you want me to decompose that logic in something like that instead of nested ternary.


```
if (headers instanceof Headers) {
  entries = headers.entries();
} else if (util.isIterable(headers)) {
  entries = headers;
} else {
  entries = Object.entries(headers);
}

```

ronag: Yes, native headers iterate entries.

ronag:
```js
> Array.from(new Headers({ a: 1 }))
[ [ 'a', '1' ] ]
> Object.entries({ a: 1 })
[ [ 'a', 1 ] ]
```

iiAku: Ended up doing that
```
const entries = util.isIterable(headers) ? headers.entries() : Object.entries(headers)
```

instead of

```
const entries = util.isIterable(headers) ? headers : Object.entries(headers)
```

otherwise we can't destructure. That way I believe we are fine :)

tsctx: my opinion:

```js
const entries = typeof headers[Symbol.iterator] === 'function' ? headers[Symbol.iterator]() : Object.entries(headers)
```

ronag: but that's not correct, just because it's iterable doesn't not neccessarily mean it has `.entries()`. I don't understand what's from with my suggestion?

lib/handler/redirect-handler.js
```diff
@@ -228,7 +228,7 @@ function cleanRequestHeaders (headers, removeContent, unknownOrigin) {
       }
     }
   } else if (headers && typeof headers === 'object') {
-    const entries = headers instanceof Headers ? headers.entries() : Object.entries(headers)
+    const entries = util.isIterable(headers) ? headers.entries() : Object.entries(headers)
```
ronag, line 231:
```suggestion
    const entries = util.isIterable(headers) ? headers : Object.entries(headers)
```

ronag (changes requested):

lib/handler/redirect-handler.js
ronag: @tsctx it's the same as `const entries = util.isIterable(headers) ? headers : Object.entries(headers)`

mcollina (approved):
lgtm

mcollina:
@ronag PTAL

ronag (approved):

lib/handler/redirect-handler.js
tsctx: it's different. `util.isIterable` will catch `util.isIterable` .

lib/handler/redirect-handler.js
```diff
@@ -228,7 +228,7 @@ function cleanRequestHeaders (headers, removeContent, unknownOrigin) {
       }
     }
   } else if (headers && typeof headers === 'object') {
-    const entries = headers instanceof Headers ? headers.entries() : Object.entries(headers)
+    const entries = util.isIterable(headers) ? headers : Object.entries(headers)
```
tsctx, line 231:
```suggestion
    const entries = typeof headers[Symbol.iterator] === 'function' ? headers : Object.entries(headers)
```

`util.isIterable` will catch `Symbol.asyncIterator` .

lib/handler/redirect-handler.js
iiAku: sorry I mix and match @ronag lgtm

lib/handler/redirect-handler.js
ronag: yea, that does seem wrong

PR #3833, opened 2024-11-13
Headers webidl errors

### LINKED ISSUE #3829: Error message for invalid headers could be improved
## Bug Description

I'm getting the error message:

> TypeError: Headers contructor: init is a symbol, which cannot be converted to a DOMString.

## Reproducible By

```
const headers = new Headers({'x-fake-header': Symbol('why is this here?')})
```

## Expected Behavior

`contructor` should not be typo'd

"init is a symbol" is incorrect. It _contains_ a symbol, but isn't one

## Logs & Screenshots

```
TypeError: Headers contructor: init is a symbol, which cannot be converted to a DOMString.
    at webidl.errors.exception (node:internal/deps/undici/undici:3384:14)
    at webidl.converters.DOMString (node:internal/deps/undici/undici:3650:29)
    at webidl.converters.ByteString (node:internal/deps/undici/undici:3658:35)
    at Object.record<ByteString, ByteString> (node:internal/deps/undici/undici:3568:32)
    at webidl.converters.HeadersInit (node:internal/deps/undici/undici:8695:67)
```

## Environment

Tested on Node.js v20.17.0 and v22.11.0

### Additional context

Users were running into this in SvelteKit projects. I'm not quite sure how/why yet, but I think maybe `connect` (which is used by `vite`) has something like this in its headers and the headers are being copied from there

KhafraDev:
fixes #3829

### DIFF
--- a/lib/web/fetch/headers.js
+++ b/lib/web/fetch/headers.js
@@ -451,7 +451,7 @@ class Headers {
 
     // 2. If init is given, then fill this with init.
     if (init !== undefined) {
-      init = webidl.converters.HeadersInit(init, 'Headers contructor', 'init')
+      init = webidl.converters.HeadersInit(init, 'Headers constructor', 'init')
       fill(this, init)
     }
   }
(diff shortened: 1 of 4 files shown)


KhafraDev:
everything that uses `webidl.recordConverter` will also benefit from these changes.

tsctx (approved):

PR #3834, opened 2024-11-13
fix: memory store

ronag:
Simplify and fix memory leak

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/test/cache-interceptor/cache-stores.js
+++ b/test/cache-interceptor/cache-stores.js
@@ -15,7 +15,6 @@ function cacheStoreTests (CacheStore) {
   describe(CacheStore.prototype.constructor.name, () => {
     test('matches interface', async () => {
       const store = new CacheStore()
-      equal(typeof store.isFull, 'boolean')
       equal(typeof store.get, 'function')
       equal(typeof store.createWriteStream, 'function')
       equal(typeof store.delete, 'function')
@@ -195,8 +194,9 @@ function cacheStoreTests (CacheStore) {
 
       const readStream = store.get(structuredClone(request))
       notEqual(readStream, undefined)
+      const { vary, ...responseValue } = requestValue
       deepStrictEqual(await readResponse(readStream), {
-        ...requestValue,
+        ...responseValue,
         body: requestBody
       })
 
--- a/test/cache-interceptor/cache-stores.js
+++ b/test/cache-interceptor/cache-stores.js
@@ -213,46 +213,6 @@ function cacheStoreTests (CacheStore) {
   })
 }
 
-test('MemoryCacheStore locks values properly', async () => {
-  const store = new MemoryCacheStore()
-
-  const request = {
-    origin: 'localhost',
-    path: '/',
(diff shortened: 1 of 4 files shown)


ronag:
@flakey5

flakey5:
The test for locking is failing

ronag:
I dont' see why we need any locking here... I might need some help fixing the tests though

test/cache-interceptor/cache-stores.js
```diff
@@ -213,46 +213,6 @@ function cacheStoreTests (CacheStore) {
...
-
-  // Stream is now closed, let's lock any new write streams
-  const result = store.get(request)
-  notEqual(result, undefined)
-
-  // Consume & close the result, this should lift the write lock
-  await readResponse(result)
-
-  notEqual(store.createWriteStream(request, requestValue), undefined)
-})
```
mcollina, line 254: Why this test was removed? Is this functionality not needed anymore?

ronag: IMHO. We don't need locking. The feature is removed.

mcollina (approved):
lgtm

mcollina:
cc @flakey5

flakey5:
With how we have the cache store currently set up, locking is needed so we don't read from a response that's currently being written to

For example if a request is made to some server, it's possible for another request to be made to the same resource before the first request finishes. If that happens currently (with locking), the cache store will see that the first request is still in flight since it's locked and return undefined.

Without locking, the cache store will see that the first request has some cached details (from the `CachedResponse` object) but won't know that it's still in flight. The cache store would then return the incomplete response to the user as if it was completed. The body would also be missing since we only set it in the value once the write stream finishes.

Or in other words what could happen w/o locking,
 - request 1 is made to `http://localhost`
   - cache store writes the data of `CachedResponse` into memory (status code, headers, ...)
   - write stream continues
 - request 2 is made to `http://localhost` before request 1 finishes
   - cache store sees the contents from `CachedResponse`, thinks the result is complete, and returns it
   - client doesn't get the full response (only gets things set in `CachedResponse`: status code, headers, ...)
 - client waiting on request 1 eventually gets the full response

Another smaller thing is that w/o locking we could end up w/ two write streams for the same resource. Currently, because we return `undefined` from `get` if the resource is locked, the cache handler ultimately does try to make another write stream for a locked resource. However, we again check if the value is locked in `createWriteStream` and if it is, we return undefined. (This is a smaller thing though that just concerns resource consumption. The different write streams would race to write to `this.#values` but that operation is about as atomic as we can make it)

Optionally, we could just not write the response into `this.#values` until we have the full thing (which is what was done here), however, the way it's set up currently will help us out when we figure out how we wanna add back request deduplication in the future. This also doesn't solve the smaller issue of multiple write streams in the cache store for the same resource.

I think we should stick with the locking, however, either or works

PR #3658, opened 2024-10-01
feat: support request cache control directives

flakey5:
~~**Note**: this is a draft since https://github.com/nodejs/undici/pull/3562 isn't landed. Until it is landed, this will be based off of that pr's branch. For the actual diff see https://github.com/flakey5/undici/compare/flakey5/3231...flakey5:undici:flakey5/20240924/cli-cache-control~~

## This relates to...

Adding client side http caching (https://github.com/nodejs/undici/pull/3562)

### Features

RFC9111 allows the request to specify a cache-control header with directives that can affect how we handle the request in the cache interceptor (re [5.2.1](https://www.rfc-editor.org/rfc/rfc9111.html#name-request-directives)). This adds support for them.

## Status

- [x] I have read and agreed to the [Developer's Certificate of Origin][cert]
- [x] Tested
- [ ] Benchmarked (**optional**)
- [ ] Documented
- [x] Review ready
- [x] In review
- [ ] Merge ready

[cert]: https://github.com/nodejs/undici/blob/main/CONTRIBUTING.md

cc @mcollina @ronag

PR #3842, opened 2024-11-18
test: add cache testing suite

### LINKED ISSUE #3852: Add cache support for responses without `cache-control: public`
According to https://www.rfc-editor.org/rfc/rfc9111.html#response.cacheability, it's possible to cache responses if `Cache-Control: public` is not present.

Going through that list (non exhaustively):

- [ ] Support caching data with `Cache-Control: private`, behind an option (disabled by default)
- [ ] Support `Expires` header
- [ ] Caching responses that are not explicitly uncacheable, behind an option (disabled by default)

On that latter point:

> Note that, in normal operation, some caches will not store a response that has neither a cache validator nor an explicit expiration time, as such responses are not usually useful to store. However, caches are not prohibited from storing such responses.

This might be useful for some application.

### LINKED ISSUE #3869: sql cache should always have a entry size limit
We can't store more than 2GB.

flakey5:
This adds the test suite from https://cache-tests.fyi to test the cache interceptor.

Note these aren't compliance tests, we don't need to pass 100% of these. Arguably we shouldn't since some of the tests aren't even applicable to Undici (i.e. cdn-specific, we can skip these though). However, I do think we should try to rely on these in CI.

Current issues:
* [x] #3658 needs to be merged
* [x] Lots of tests failing due to network errors (`TypeError: fetch failed`, mostly socket & http protocol issues)
* [ ] Some test failures

Closes #3852 
Closes #3869 

cc @mcollina @ronag


test/cache-interceptor/cache-tests.mjs
```diff
@@ -0,0 +1,111 @@
...
+import { Client, interceptors, setGlobalDispatcher } from '../../index.js'
+import { getResults, runTests } from '../fixtures/cache-tests/test-engine/client/runner.mjs'
+import tests from '../fixtures/cache-tests/tests/index.mjs'
+import { once } from 'node:events'
+import { determineTestResult, testLookup } from '../fixtures/cache-tests/test-engine/lib/results.mjs'
+
+const PROTOCOL = 'http'
+const PORT = 8000
+const BASE_URL = `${PROTOCOL}://localhost:${PORT}`
+const PIDFILE = '/tmp/http-cache-test-server.pid'
```
KhafraDev, line 15: this won't work on windows

test/cache-interceptor/cache-tests.mjs
```diff
@@ -0,0 +1,111 @@
...
+
+// Wait for the server to start
+await once(worker, 'message')
+
+// Expose cache globally
+const client = new Client(BASE_URL).compose(interceptors.cache())
+setGlobalDispatcher(client)
+
+// Run the tests
+await runTests(tests, globalThis.fetch, false, BASE_URL).catch(err => {
```
KhafraDev, line 35: this should use undici's fetch, not the global

KhafraDev (commented):
> Note these aren't compliance tests, we don't need to pass 100% of these. Arguably we shouldn't since some of the tests aren't even applicable to Undici.

If these tests are targeting fetch, and fetch is a browser spec, why wouldn't some of the tests be applicable?

flakey5:
> If these tests are targeting fetch

They aren't targeting fetch. It has tests for every part of the stack that can implement caching. They're written in js w/ fetch so that they can be ran in browsers, but browsers aren't the only thing the tests cover. A lot of the behaviors do overlap, however, there are still things specific to servers or cdns. 

There are also tests for things that rfc9111 obsoletes (such as using the `pragma` header for caching directives)

KhafraDev:
> They aren't targeting fetch

What I mean is that the tests explicitly use fetch and it tests browser caching behavior (every http request made through browsers is done through fetch, at least in the specs). Anything that doesn't work would seem like a bug to me.

> A lot of the behaviors do overlap, however, there are still things specific to servers or cdns.

Is there anything specific?

> There are also tests for things that rfc9111 obsoletes (such as using the pragma header for caching directives)

Is lack of pragma support intentional, because it's obsolete? Is there no plan on adding support for it?

flakey5:
> Is there anything specific?

They're mostly tests for `CDN-Cache-Control`

> Is lack of pragma support intentional, because it's obsolete? Is there no plan on adding support for it?

It was intentionally left out, right now there aren't any plans to add it.

flakey5:
So it does look like we can run tests that only apply to browsers, which should skip the cdn tests

mcollina:
Why did you lift the entirety of the source code of that project? What's the long term maintainability strategy?

Either:

- we can use the npm module they publish, so just add a dependency
- we vendor all their repo, but you also add a way to automatically update it
- we only lift the test cases, essentially making future updates on these harder

Anyhow, I would recommend _not_ to use `fetch()` for this, because it'd make things harder long term. Use `undici.request()`.

KhafraDev:
> we can use the npm module they publish, so just add a dependency

We can't - it hasn't been updated in 4 years and there seem to be relevant tests added and fixed in that time.

> Anyhow, I would recommend not to use fetch() for this, because it'd make things harder long term. Use undici.request().

afaict the tests are specifically vendored for use in browsers and the `runTests` function they export expects fetch. Using anything other than fetch would likely end up being harder to maintain.

flakey5:
> Why did you lift the entirety of the source code of that project

In addition to what @KhafraDev said, there's a fix I needed to make to the source (re http-tests/cache-tests#139)

> afaict the tests are specifically vendored for use in browsers and the runTests function they export expects fetch. Using anything other than fetch would likely end up being harder to maintain.

+1

ronag (approved):

test/cache-interceptor/cache-tests.mjs
```diff
@@ -0,0 +1,346 @@
...
+          reject(new Error('worker thread startup timed out'))
+        }, 5000)
+      })
+    ])
+
+    const client = new Agent().compose(interceptors.cache(environment.opts))
+    setGlobalDispatcher(client)
+
+    // Run the tests
+    await runTestSuite(tests, fetch, true, BASE_URL)
```
KhafraDev, line 214: the test suite is still using the global fetch, not the one that undici exports.

test/cache-interceptor/cache-tests.mjs
```diff
@@ -0,0 +1,346 @@
...
+
+  try {
+    await Promise.race([
+      once(worker, 'message'),
+      new Promise((_resolve, reject) => {
+        setTimeout(() => {
+          reject(new Error('worker thread startup timed out'))
+        }, 5000)
+      })
+    ])
```
KhafraDev, line 208:
```suggestion
    await once(worker, 'message', { signal: AbortSignal.timeout(5000) })
```

nit

KhafraDev (approved):

mcollina (approved):
lgtm

PR #3844, opened 2024-11-19
Support globalThis.FormData

mcollina:
Since https://github.com/nodejs/undici/pull/3502, we only support our own classes, and that is the right move. However, we need to relax this condition to allow for `globalThis.FormData`, as that is a _different_ copy of undici.

In essence, this change make the following fork

```js
// import { fetch, FormData } from 'undici'
import fastify from 'fastify'
import fastifyMultipart from '@fastify/multipart'

const app = fastify()

app.register(fastifyMultipart, {
  attachFieldsToBody: true
})

app.post('/upload', async (req, reply) => {
  console.log('received', req.body.file.filename)
  return 'ok'
})

await app.listen({ port: 3000 })

const body = new FormData()
body.append('file', new Blob(['hello world'], { type: 'text/plain' }), 'hello.txt')

const res = await fetch('http://localhost:3000/upload', {
  method: 'POST',
  body
})

console.log(res.status)
console.log(await res.text())
```

### DIFF
--- a/lib/web/fetch/body.js
+++ b/lib/web/fetch/body.js
@@ -109,7 +109,7 @@ function extractBody (object, keepalive = false) {
 
     // Set source to a copy of the bytes held by object.
     source = new Uint8Array(object.buffer.slice(object.byteOffset, object.byteOffset + object.byteLength))
-  } else if (webidl.is.FormData(object)) {
+  } else if (webidl.is.FormData(object) || (globalThis.FormData && object instanceof FormData)) {
     const boundary = `----formdata-undici-0${`${Math.floor(Math.random() * 1e11)}`.padStart(11, '0')}`
     const prefix = `--${boundary}\r\nContent-Disposition: form-data`
 
--- a/lib/web/fetch/formdata.js
+++ b/lib/web/fetch/formdata.js
@@ -258,6 +258,13 @@ function makeEntry (name, value, filename) {
   return { name, value }
 }
 
-webidl.is.FormData = webidl.util.MakeTypeAssertion(FormData)
+const _isFormData = webidl.util.MakeTypeAssertion(FormData)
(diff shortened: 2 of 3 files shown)


lib/web/fetch/body.js
```diff
@@ -109,7 +109,7 @@ function extractBody (object, keepalive = false) {
 
     // Set source to a copy of the bytes held by object.
     source = new Uint8Array(object.buffer.slice(object.byteOffset, object.byteOffset + object.byteLength))
-  } else if (webidl.is.FormData(object)) {
+  } else if (webidl.is.FormData(object) || (globalThis.FormData && object instanceof FormData)) {
```
tsctx, line 112:
```suggestion
  } else if (webidl.is.FormData(object)) {
```

lib/web/fetch/formdata.js
```diff
@@ -258,6 +258,13 @@ function makeEntry (name, value, filename) {
   return { name, value }
 }
 
-webidl.is.FormData = webidl.util.MakeTypeAssertion(FormData)
+const _isFormData = webidl.util.MakeTypeAssertion(FormData)
+webidl.is.FormData = function (V) {
```
tsctx, line 262: How about something like this?
```js
// lib/web/fetch/webidl.js
webidl.util.MakeTypeAssertionMultiple = function (List) {
  return (O) => List.some((I) => FunctionPrototypeSymbolHasInstance(I, O))
}
```

Usage: 
```js
webidl.is.FormData = webidl.util.MakeTypeAssertionMultiple([FormData, gloablThis.FormData].filter(Boolean))
```

KhafraDev:
I don't understand the need for this change.

mcollina:
Ok, let's close this. We would need to document this a bit better, as I suspect a flurry of issues.

PR #3846, opened 2024-11-19
Set pipelining to 4 by default

mcollina:
After we landed https://github.com/nodejs/undici/pull/3771, we could easily set a `pipelining` value greater than zero, because all request will queue up until headers of the previous response have been received.


mcollina:
This _seemed_ a good idea, but it breaks `fetch()` badly.

ronag (approved):

mcollina:
> How? Fetch should always set blocking true.

The wpts expects no pipelining for it. I don't really understand if it's a _strong requirement_, or not.

ronag (approved):

metcoder95 (approved):

PR #3854, opened 2024-11-20
feat: add support for Timeout and Trailers

metcoder95:
## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -27,7 +27,8 @@ const {
   kResume,
   kSize,
   kHTTPContext,
-  kClosed
+  kClosed,
+  kBodyTimeout
 } = require('../core/symbols.js')
 
 const kOpenStreams = Symbol('open streams')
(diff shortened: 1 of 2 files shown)


lib/dispatcher/client-h2.js
```diff
@@ -278,6 +283,7 @@ function shouldSendContentLength (method) {
 }
 
 function writeH2 (client, request) {
+  const requestTimeout = request.bodyTimeout ?? client[kBodyTimeout]
```
metcoder95, line 286: I was quite unsure if using the `bodyTimeout` for this, and adding a new one. But given the nature of the timeout, using the body's one should be enough.

lib/dispatcher/client-h2.js
```diff
@@ -375,6 +381,7 @@ function writeH2 (client, request) {
       session[kOpenStreams] -= 1
       if (session[kOpenStreams] === 0) session.unref()
     })
+    stream.setTimeout(requestTimeout)
```
mcollina, line 384: I would not call this if there isn't a requestTimeout.

metcoder95: Do you mean that if no `bodyTimeout` passed at a request level, we just ignore it, discarding the one set for the client itself? Or do you have something else in mind?

mcollina (approved):
lgtm once windows passes

mcollina (approved):
lgtm

PR #3858, opened 2024-11-21
fix: pass down context in onConnect

DTrombett:
## This relates to...

The retry and cache handlers do not pass down the context of the `onConnect` method. This means that using the redirect handler with one of them will not pass the context to the main handler:

```js
request("http://github.com", {
    dispatcher: new Agent().compose(
        interceptors.redirect({ maxRedirections: 1 }),
        interceptors.retry(),
    ),
}).then((res) => console.log(res.context)); // undefined

request("http://github.com", {
    dispatcher: new Agent().compose(
        interceptors.redirect({ maxRedirections: 1 }),
    ),
}).then((res) => console.log(res.context)) // { history: [URL, URL] }
```

## Changes

- Cache handler: used `...args` to pass all the args (like in the decorator handler)
- Retry handler: `onConnect` is now called on the first connection instead that in the constructor, allowing to pass the context

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [x] In review
- [ ] Merge ready

### DIFF
--- a/lib/handler/cache-handler.js
+++ b/lib/handler/cache-handler.js
@@ -49,14 +49,14 @@ class CacheHandler extends DecoratorHandler {
     this.#handler = handler
   }
 
-  onConnect (abort) {
+  onConnect (...args) {
     if (this.#writeStream) {
       this.#writeStream.destroy()
       this.#writeStream = undefined
     }
 
     if (typeof this.#handler.onConnect === 'function') {
-      this.#handler.onConnect(abort)
+      this.#handler.onConnect(...args)
     }
   }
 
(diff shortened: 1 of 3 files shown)


mcollina (changes requested):

ronag (approved):

Uzlopak:
...args is imho not that nice. i think it does not cover in the types or so, that it can have two arguments. only abort.

DTrombett:
I noticed that `...args` is what is used in the decorator handler too, so I assumed it would be ok

mcollina (approved):
lgtm

PR #3862, opened 2024-11-21
fix: 301 and 302 change method to GET

DTrombett:
## This relates to...

Didn't open an issue before this PR so sorry if this is expected but I could not find any discussion about it.
The redirect handler doesn't change the method to `GET` when the status code is 301/302 and the original method is `POST` despite being the standard in fetch spec and browser behavior. Also the HTTP standard mentions this too:
> Note: For historical reasons, a user agent MAY change the request method from POST to GET for the subsequent request. If this behavior is undesired, the 307 (Temporary Redirect) status code can be used instead.

## Changes

Method is changed to `GET` and body is removed when a 301/302 response code is received, aligning with fetch and browser behavior
The tests were updated to reflect this change and the ones specifically about request body/method were updated to use `PUT` instead

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [x] In review
- [ ] Merge ready

### DIFF
--- a/lib/handler/redirect-handler.js
+++ b/lib/handler/redirect-handler.js
@@ -135,7 +135,7 @@ class RedirectHandler {
     // https://tools.ietf.org/html/rfc7231#section-6.4.2
     // https://fetch.spec.whatwg.org/#http-redirect-fetch
     // In case of HTTP 301 or 302 with POST, change the method to GET
-    if ((statusCode === 302 || statusCode === 301) && this.opts.method === 'POST') {
+    if ((statusCode === 301 || statusCode === 302) && this.opts.method === 'POST') {
       this.opts.method = 'GET'
       this.opts.body = null
     }
--- a/lib/handler/redirect-handler.js
+++ b/lib/handler/redirect-handler.js
@@ -137,6 +137,7 @@ class RedirectHandler {
     // In case of HTTP 301 or 302 with POST, change the method to GET
     if ((statusCode === 301 || statusCode === 302) && this.opts.method === 'POST') {
       this.opts.method = 'GET'
+      util.destroy(this.opts.body)
       this.opts.body = null
     }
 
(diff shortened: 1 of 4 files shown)


ronag (approved):

lib/handler/redirect-handler.js
```diff
@@ -132,6 +132,14 @@ class RedirectHandler {
     this.opts.maxRedirections = 0
     this.opts.query = null
 
+    // https://tools.ietf.org/html/rfc7231#section-6.4.2
+    // https://fetch.spec.whatwg.org/#http-redirect-fetch
+    // In case of HTTP 301 or 302 with POST, change the method to GET
+    if ((statusCode === 301 || statusCode === 302) && this.opts.method === 'POST') {
+      this.opts.method = 'GET'
+      this.opts.body = null
```
ronag, line 140: If body is a stream it needs to be destroyed.

ronag (changes requested):

lib/handler/redirect-handler.js
DTrombett: Do we need to do update the HTTP 303 case as well?

lib/handler/redirect-handler.js
```diff
@@ -132,10 +132,20 @@ class RedirectHandler {
     this.opts.maxRedirections = 0
     this.opts.query = null
 
+    // https://tools.ietf.org/html/rfc7231#section-6.4.2
+    // https://fetch.spec.whatwg.org/#http-redirect-fetch
+    // In case of HTTP 301 or 302 with POST, change the method to GET
+    if ((statusCode === 301 || statusCode === 302) && this.opts.method === 'POST') {
+      this.opts.method = 'GET'
+      util.destroy(this.opts.body)
```
ronag, line 140:
```suggestion
      util.destroy(this.opts.body.on('error', () => {})
```

ronag (approved):

lib/handler/redirect-handler.js
DTrombett: Maybe is better to use `util.isStream()`?

Uzlopak:
semver major?

PR #3871, opened 2024-11-22
fix: cache fixes

ronag:
## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/types/cache-interceptor.d.ts
+++ b/types/cache-interceptor.d.ts
@@ -6,7 +6,7 @@ declare namespace CacheHandler {
   export type CacheMethods = 'GET' | 'HEAD' | 'OPTIONS' | 'TRACE'
 
   export interface CacheOptions {
-    store?: CacheStore
+    store: CacheStore
 
     /**
      * The methods to cache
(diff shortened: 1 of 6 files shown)


types/cache-interceptor.d.ts
```diff
@@ -6,7 +6,7 @@ declare namespace CacheHandler {
   export type CacheMethods = 'GET' | 'HEAD' | 'OPTIONS' | 'TRACE'
 
   export interface CacheOptions {
-    store?: CacheStore
+    store: CacheStore
```
mcollina, line 9: This is incorrect, we default to memory.

ronag: Not according the cache handler. Then we have a different type issue

mcollina: https://github.com/nodejs/undici/blob/a78147362135d7e5ea1617cef093fe5512a27ad9/lib/interceptor/cache.js#L90-L93 doesn't look that way.

ronag, line 9:
```suggestion
    store?: CacheStore
```

mcollina: This was not committed

ronag: This is different. It's `CacheHandlerOptions` you are thinking of `CacheOptions`

mcollina: Ah!

mcollina (approved):
lgtm

PR #3878, opened 2024-11-24
feat: new hooks

ronag:


ronag:
I'm unsure about `context` in `onRequestStart`. 

A. Is it necessary?
B. Should it live on the controller instead?

test/mock-agent.js
```diff
@@ -142,89 +142,6 @@ describe('MockAgent - dispatch', () => {
...
-      method: 'GET'
-    }, {
-      onError: (err) => { throw err },
-      onConnect: () => {},
-      onBodySent: () => {},
-      onHeaders: () => {},
-      onData: () => {},
-      onComplete: 'INVALID'
-    }), new InvalidArgumentError('invalid onComplete method'))
-  })
```
mcollina, line 227: why was this removed?

mcollina:
This looks good to me. Can you update the docs?

mcollina (approved):
lgtm

test/mock-agent.js
ronag: The redirect interceptor has some kind of assertion of the handler that does not exist anywhere else and is difficult to have compat with through the wrap/unwrap.

tim1uphealth:
Nit: could update the docs example, it's currently using the deprecated hooks: https://github.com/nodejs/undici/blob/6bb527e9d768c5ca5b98765c28cf403ad8b250fb/docs/docs/api/Dispatcher.md?plain=1#L239

metcoder95:
Would you like to send a PR to update documentation?

PR #3879, opened 2024-11-25
refactor: port redirect handler to new hooks

ronag:
## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/core/util.js
+++ b/lib/core/util.js
@@ -436,46 +436,6 @@ function parseHeaders (headers, obj) {
   return obj
 }
 
-function normalizeHeaders (headers) {
-  const ret = {}
-  if (headers == null) {
-    // Do nothing...
-  } else if (Array.isArray(headers)) {
-    for (let i = 0; i < headers.length; i += 2) {
-      if (Array.isArray(headers[i]) && headers[i].length === 2) {
-        const key = headerNameToString(headers[i][0])
-        const val = ret[key]
-        if (val == null) {
-          ret[key] = String(headers[i][1])
-        } else if (typeof val === 'string') {
-          ret[key] = [val, String(headers[i][1])]
-        } else {
-          val.push(String(headers[i][1]))
-        }
-      } else {
-        const key = headerNameToString(headers[i])
-        const val = ret[key]
-        if (val == null) {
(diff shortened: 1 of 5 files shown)


lib/core/util.js
```diff
@@ -436,6 +436,46 @@ function parseHeaders (headers, obj) {
...
+    for (const [key, val] of headers) {
+      ret[headerNameToString(key)] = Array.isArray(val) ? val.map(x => String(x)) : String(val)
+    }
+  } else {
+    for (const [key, val] of Object.entries(headers)) {
+      ret[headerNameToString(key)] = Array.isArray(val) ? val.map(x => String(x)) : String(val)
+    }
+  }
+  return ret
+}
```
mcollina, line 477: This logic seems very complex. why is it needed?

ronag: To normalize different headers shapes?

lib/core/util.js
```diff
@@ -436,6 +436,46 @@ function parseHeaders (headers, obj) {
...
+    }
+  } else if (typeof headers[Symbol.iterator] === 'function') {
+    for (const [key, val] of headers) {
+      ret[headerNameToString(key)] = Array.isArray(val) ? val.map(x => String(x)) : String(val)
+    }
+  } else {
+    for (const [key, val] of Object.entries(headers)) {
+      ret[headerNameToString(key)] = Array.isArray(val) ? val.map(x => String(x)) : String(val)
+    }
+  }
```
DTrombett, line 475: Maybe here we can avoid the code duplication?
```suggestion
  } else {
    if (typeof headers[Symbol.iterator] !== 'function') {
      headers = Object.entries(headers)
    }
    for (const [key, val] of headers) {
      ret[headerNameToString(key)] = Array.isArray(val) ? val.map(x => String(x)) : String(val)
    }
  }
```
Although I think it is better to leave the headers as-is if it is an object so that if we need to normalize headers in more than one place (e.g. another interceptor) we do not need to loop again through all the headers (also considering that the only thing we are doing is casting to a string)

mcollina (approved):
lgtm

Issue #3386, opened 2024-07-03
v7


ronag:
- [x] llhttp (@metcoder95 merges next branch) - https://github.com/nodejs/undici/pull/3388
- [x] remove deprecated interceptor api (@metcoder95 merges next branch) #3399 
- [x] dispatcher API issues (@mcollina, @ronag, @metcoder95 )
- [x] remove polyfills (@KhafraDev has already prepared PRs)
- [x] add alternative to onHeaders with pre-parsed headers (@ronag )
- [ ] diagnostics channels (@mcollina )

We start with merging the next branch into main (@metcoder95) and then start working actively on v7 with a deadline relative to Node v23.

metcoder95:
Shall we consider the exposure of the lenient flags from `lhttp` as part of this?

mcollina:
> Shall we consider the exposure of the lenient flags from lhttp as part of this?

Yes, I think so.

Uzlopak:
How can i support?

KhafraDev:
> * diagnostics channels

is this something we're absolutely landing?

Also hopefully https://github.com/nodejs/undici/pull/2727 is included in the fetch stuff, it's slightly older so I'm hoping it won't get missed.

metcoder95:
> How can i support?

Would you like to support adding the lenient flags for llhttp @Uzlopak?

metcoder95:
I was kicking off the work for the lenient flags, but I just noticed that most of them as marked as opening a security vector if enabled; do we want to enable that within `undici`?

By default they are disabled sure, but do we want to expose the users to that possibility?
I'd consider to not do so, cannot imagine under which circumstances (beyond legacy servers) you might want to enable this

b2whats:
Why was dropped interceptors ? What should i use instead ?

metcoder95:
Interceptors hasn't been dropped, but rather refactored into a new API.
Check: https://undici.nodejs.org/#/docs/api/Dispatcher?id=dispatchercomposeinterceptors-interceptor

Uzlopak:
can the npm package `next` point to v7?

Uzlopak:
![image](https://github.com/user-attachments/assets/88b8c172-d722-44c7-975e-313a810b3d61)

mcollina:
That's done.

---

On a different topic, I think we might want to release v7 _before_ Node.js v23, to be included there.

metcoder95:
Backported some of the work done for interceptors and smaller fixes at core.

[...]

PR #3880, opened 2024-11-25
Drop node v18

mcollina:


mcollina:
cc @trivikr

KhafraDev:
I haven't seen any reason to drop v18. Nothing in fetch or other web specs needs it gone, and it's in maintenance mode for another 5 months. Is there any reason other than precedence? If precedence is the main concern, there were genuine issues with previous versions of node that required us to drop them - I can't think of any similar issues in v18.

trivikr:
@KhafraDev I'd provided estimates on how long unidici maintainers have to support Node.js 18.x in https://github.com/nodejs/undici/issues/3386#issuecomment-2489199082 after it's end-of-life.
In summary, it needs to be supported for at least 6 months under the assumption that undici@v8 is released in Dec 2025.

Matteo also provided how dropping Node.js 18.x will ease maintenance of undici@v7 in  https://github.com/nodejs/undici/issues/3386#issuecomment-2498533727

trivikr (approved):
LGTM, but this needs approval from a maintainer.

ronag (approved):

package.json
```diff
@@ -128,7 +128,7 @@
     "ws": "^8.11.0"
   },
   "engines": {
-    "node": ">=18.17"
+    "node": ">=20.18.1"
```
RobinTail, line 131: what was the reason for this particular version?

mcollina: Because it's the latest version that was available at the time of the release. You should not be using _anything else_.

mcollina: > The latest Android version is 15, so the all developers should do applications only for Android 15?

This is not a good analogy. A good analogy is slightly different: should Google recommend the use of _anything_ but the latest patch available for that given version? As we speak, this is 15.0.0_r5. Why would Google recommend the use of a version that contains known vulnerabilities? You can find the details related to Android CVEs here: https://source.android.com/docs/security/bulletin/2024-11-01#spl-vulnerability-details.

AlttiRi: [`engines.node`](https://docs.npmjs.com/cli/v6/configuring-npm/package-json#engines) field is not about "recommendations". 

The purpose of this field is to show which versions of Node.js the library will work on and which will not.

Wrong use of this field feels like not a recommendation, but a force to update Node.js.

PR #3891, opened 2024-11-27
Make the responseError interceptor backward compatible, add a body property

mcollina:
This makes the `responseError` interceptor backward compatible.
`throwOnError` added a `err.body` property, while the interceptor adds a `err.data` property. I've renamed `data` to `body`.


mertcanaltin (approved):
LGTM

PR #3900, opened 2024-11-29
feat: Adjust allowed error codes for detecting node:sqlite

### LINKED ISSUE #3899: V7 unexpected sqlite error when using PM2
## Bug Description

I am using pm2 for managing my application and something about the expermental node:sqlite detection doesn't work correctly/is incompatible.

## Reproducible By

test.js:
```javascript
const { fetch } = require("undici");
console.log("hello");
```

This works fine of course:
```
node test.js
```

This does not:

```
pm2 start test.js
```

and errors with this

```
pm2 logs
```

```
Error: Cannot find module 'node:sqlite'
Require stack:
- /path/src/archive/test/node_modules/undici/lib/cache/sqlite-cache-store.js
- /path/src/archive/test/node_modules/undici/index.js
- /path/src/archive/test/test.js
    at Function._resolveFilename (node:internal/modules/cjs/loader:1249:15)
    at Hook._require.Module.require (/path/.local/share/nvm/v22.11.0/lib/node_modules/pm2/node_modules/require-in-the-middle/index.js:81:25)
    at require (node:internal/modules/helpers:141:16)
    at Object.<anonymous> (/path/src/archive/test/node_modules/undici/lib/cache/sqlite-cache-store.js:3:26)
    at Module._compile (node:internal/modules/cjs/loader:1546:14)
    at Object..js (node:internal/modules/cjs/loader:1689:10)
    at Module.load (node:internal/modules/cjs/loader:1318:32)
    at Function._load (node:internal/modules/cjs/loader:1128:12)
    at TracingChannel.traceSync (node:diagnostics_channel:315:14)
    at wrapModuleLoad (node:internal/modules/cjs/loader:218:24) {
  code: 'MODULE_NOT_FOUND',
  requireStack: [
    '/path/src/archive/test/node_modules/undici/lib/cache/sqlite-cache-store.js',
    '/path/src/archive/test/node_modules/undici/index.js',
    '/path/src/archive/test/test.js'
  ]
}
```

I can get around it by explicitly setting the flag like this

```
pm2 start test.js --node-args="--experimental-sqlite"
```

## Expected Behavior

node:sqlite is detected correctly (as missing or not) when running through pm2

## Environment

Node v22
Undici v7.0.0

xconverge:
I am not really sure how to write a test for this...

## This relates to...

N/A

## Rationale

This was a bit too restrictive for how big of a failure it can lead to

### Features

N/A

### Bug Fixes

Fixes #3899

N/A

## Status

- [x] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/index.js
+++ b/index.js
@@ -53,7 +53,7 @@ try {
   const SqliteCacheStore = require('./lib/cache/sqlite-cache-store')
   module.exports.cacheStores.SqliteCacheStore = SqliteCacheStore
 } catch (err) {
-  if (err.code !== 'ERR_UNKNOWN_BUILTIN_MODULE') {
+  if (err.code !== 'ERR_UNKNOWN_BUILTIN_MODULE' && err.code !== 'MODULE_NOT_FOUND') {
     throw err
   }
 }
--- a/index.js
+++ b/index.js
@@ -53,9 +53,10 @@ try {
   const SqliteCacheStore = require('./lib/cache/sqlite-cache-store')
   module.exports.cacheStores.SqliteCacheStore = SqliteCacheStore
 } catch (err) {
-  if (err.code !== 'ERR_UNKNOWN_BUILTIN_MODULE' && err.code !== 'MODULE_NOT_FOUND') {
-    throw err
-  }
+  // Most likely node:sqlite was not present, since SqliteCacheStore is
+  // optional, don't throw. Don't check specific error codes here because while
+  // ERR_UNKNOWN_BUILTIN_MODULE is expected, users have seen other codes like
+  // MODULE_NOT_FOUND
 }
 
 module.exports.buildConnector = buildConnector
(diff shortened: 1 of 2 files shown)


metcoder95 (approved):

metcoder95:
cc: @flakey5

mcollina (changes requested):
I have absolutely no clue, because that require/import _always_ throw that error code from Node.js core. Something, something is 100% off on your end due to some monkeypatching one of your dependencies is making.

I recommend we drop the if-check completely.

xconverge:
@mcollina in the stack trace for the issue you can see that https://www.npmjs.com/package/require-in-the-middle is used by pm2, so yes there is definitely some monkey patching going on. I agree the if check should be dropped completely.

@metcoder95 see my message above, I do realize that this is related to pm2 specifically, but I don't think undici should be failing catastropically here for an optional/experimental feature check when we could just skip it and keep going. I am not even sure I agree that an issue needs to be opened with pm2 and/or require-in-the-middle, I think that this check should be less strict for greater compatibility

flakey5 (approved):

mcollina (approved):
lgtm

PR #3906, opened 2024-11-30
docs: add advanced usage examples for ProxyAgent

mertcanaltin:
I have added additional examples to the ProxyAgent documentation, including how to use ProxyAgent with fetching, custom proxy server setup and HTTPS tunneling.

I would be grateful if you have any suggestions 🚀

### DIFF
--- a/docs/docs/api/ProxyAgent.md
+++ b/docs/docs/api/ProxyAgent.md
@@ -200,3 +200,19 @@ const response = await fetch('https://secure.endpoint.com/api/data', {
 console.log('Response status:', response.status);
 console.log('Response data:', await response.json());
 ```
+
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

mcollina (commented):
Good work, you should add one example that configures it as a global dispatcher, so that it can be used without passing it.

mertcanaltin:
> Good work, you should add one example that configures it as a global dispatcher, so that it can be used without passing it.

I Added, thanks a lot for the suggestion

metcoder95 (approved):

mcollina (approved):
lgtm

PR #3916, opened 2024-12-02
Add example for `request` + "Garbage Collection"

WTCT-TOP:
## This relates to...

See #3914

## Changes

Added an example for `request` + "Garbage Collection".

## Status

- [ ] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/README.md
+++ b/README.md
@@ -299,9 +299,7 @@ The same applies for `request` too:
 // Do
 const response = await request(url)
   .then(res => {
-    response.body.dump().catch(err => {
-      console.error('Error while dumping the body');
-    });
+    response.body.dump();
     return res.headers;
   });
 
--- a/README.md
+++ b/README.md
@@ -293,8 +293,8 @@ const { headers } = await fetch(url);
 The same applies for `request` too:
 ```js
 // Do
-const { body, headers } = await request(url)
-await res.body.dump();
+const { body, headers } = await request(url);
+await res.body.dump(); // force consumption of body
 
 // Do not
 const { headers } = await request(url);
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -294,6 +294,22 @@ const headers = await fetch(url)
   .then(res => res.headers)
 ```
 
+The same applies for `request` too:
+```js
+// Do
+const response = await request(url)
+  .then(res => {
+    response.body.dump().catch(err => {
+      console.error('Error while dumping the body');
```
WTCT-TOP, line 303: Is there anything to do here? Will the body be cleared in this case?

ronag: dump can't throw

KhafraDev:
It was added under the fetch specification compliance section. It should be moved elsewhere.

mcollina (commented):
Can you remove the unnecessary promises?

WTCT-TOP:
@KhafraDev do you think I should move the whole GC section or just the new example for request?
@mcollina what do you mean by "unnecessary promises"? use `async/await` instead of `.then/.catch`? I followed the existing example..

Also, is there a reason to `await res.body.dump()` if I don't care about the body? isn't `res.body.dump()` (without await) good enough?

mcollina:
> @mcollina what do you mean by "unnecessary promises"? use async/await instead of .then/.catch? 

yes exactly. Currently this allocates one promise too much ;).

> Also, is there a reason to await res.body.dump() if I don't care about the body? isn't res.body.dump() (without await) good enough?

If you don't await dump, you'll still have asynchronous activity happening in the background, as the body will not be dumped immediately. Not awaiting it defeats the purpose of the example, as without it there will be lingering memory and activity running.

WTCT-TOP:
I've updated both examples to use async await, let me know what you think..

mcollina (approved):
lgtm

PR #3920, opened 2024-12-03
fix(engines): Define proper `node` version

rozzilla:
## This relates to...
When installing the last `undici` version (`v7.0.0`) on `node` version `20.18.1`:
```shell
node:internal/modules/cjs/loader:1225
  const err = new Error(message);
              ^

Error: Cannot find module 'node:sqlite'
```

## Rationale
The `node` version defined in the `engines` of the `package.json` should be updated, considering that the brand [new SQLite version](https://nodejs.org/docs/latest-v22.x/api/sqlite.html) is available only on `node >= 22.x`

## Changes
Updated the `package.json`

### Features
N/A

### Bug Fixes
N/A

I think we should communicate that `undici` version `7` will NOT work on `node` version `20.x`

## Status

- [x] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/package.json
+++ b/package.json
@@ -129,7 +129,7 @@
     "ws": "^8.11.0"
   },
   "engines": {
-    "node": ">=20.18.1"
+    "node": ">=22.11.0"
   },
   "tsd": {
     "directory": "test/types",


mcollina:
This loads correctly on Node.js v20, so maybe you have something odd in your setup.

rozzilla:
I confirm this is now fixed by version `7.1.0` ✅

PR #3924, opened 2024-12-04
fix mimetype parser wrong operator

tsctx:
The current code does not include *past the end of input*, so add equal sign.

### DIFF
--- a/lib/web/fetch/data-url.js
+++ b/lib/web/fetch/data-url.js
@@ -283,7 +283,7 @@ function parseMIMEType (input) {
 
   // 5. If position is past the end of input, then return
   // failure
-  if (position.position > input.length) {
+  if (position.position >= input.length) {
     return 'failure'
   }
 
@@ -364,7 +364,7 @@ function parseMIMEType (input) {
     }
 
     // 6. If position is past the end of input, then break.
-    if (position.position > input.length) {
+    if (position.position >= input.length) {
       break
     }
 


mcollina:
Can you include a test? Or maybe there is a failing WPT this is fixing?

KhafraDev:
Tests passing on main: 1031
Tests passing here: 1031

tsctx:
This step is there to skip the rest if it is not needed, so all tests pass without this change.

**position** is never greater than the length of **input**, so it is a dead code.

For some reason, this is the only place where the correct greater-than or equal was used, while other places used the correct greater-than or equal.

KhafraDev:
I'm not arguing with your logic (and I agree with it), but if no test failures change I would assume this code is still dead code. Should these be replaced with assertions instead?

tsctx:
> I'm not arguing with your logic (and I agree with it), but if no test failures change I would assume this code is still dead code. Should these be replaced with assertions instead?

As I mentioned before, this part of the process is to skip an extra step (at this point, an invalid entry or exit), so it is virtually untestable since it does not make any difference whether this part is there or not.

KhafraDev:
Both conditions can be caught with tests?
1. Find an input where `parseMIMEType` returns failure on that condition.
2. Find an input with an invalid parameter that triggers the if and make sure it isn't set on the returned object's parameters.

tsctx:
> Both conditions can be caught with tests?
> 
> 1. Find an input where `parseMIMEType` returns failure on that condition.
> 2. Find an input with an invalid parameter that triggers the if and make sure it isn't set on the returned object's parameters.

I have checked wpt mimesniff tests, these cases are covered.

KhafraDev (approved):

KhafraDev:
It would be nice to get codecov working again, so we could really see what is and isn't being tested without doing it manually.

Issue #3928, opened 2024-12-07
`interceptors.responseError` doesn't work with `MockAgent`


Gigioliva:
## `interceptors.responseError` doesn't work with `MockAgent`

In version `7.0.0`, the `throwOnError` option was removed.
[Here](https://blog.platformatic.dev/undici-v7-is-here#heading-removed-throwonerror-option) is indicated to use the interceptor `interceptors.responseError` to get the same behavior as before. This solution seems to work when calling a real endpoint. However, this does not work in the test environment using the MockAgent.

## Reproducible By

Here a code example (an example extracted from a test case):

```ts
import { MockAgent, Pool, interceptors } from "undici";

const URL = "https://www.foo.com";

async function main() {
  const mockAgent = new MockAgent({ connections: 1 });
  mockAgent.disableNetConnect();
  const mockPool = mockAgent.get(URL);
  mockPool.intercept({ method: "GET", path: "/foo" }).reply(400).persist();

  const client = new Pool(URL, { factory: () => mockPool });
  try {
    const response = await client
      .compose(interceptors.responseError({ throwOnError: true }))
      .request({
        method: "GET",
        path: "/foo",
      });

    console.log(response.statusCode);
  } catch (e) {
    console.log(e);
  }
}

if (require.main === module) {
  void main();
}
```

## Expected Behavior

The expected behavior is that an exception is raised, however this does not occur (it prints `undefined` as statusCode).

## Environment

- node: `20.18`
- undici: `7.1.0`

Gigioliva:
This is due to an `undefined` error [here](https://github.com/nodejs/undici/blob/cac18e12a794800c6405ed633006cff44ca6f664/lib/interceptor/response-error.js#L19)

Gigioliva:
https://github.com/nodejs/undici/pull/3931

PR #3931, opened 2024-12-07
fix: handle case no content type

Gigioliva:
## Rationale

If the server does not return the value for `content-type`, the interceptor raises an exception since it tries to use `indexOf` of undefined.

## Changes

Add a fallback in case `#contentType` is undefined.

### Bug Fixes

Fix [this issue](https://github.com/nodejs/undici/issues/3928)

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready


mcollina:
Can you please rebase/update this branch?

lib/interceptor/response-error.js
```diff
@@ -16,7 +16,7 @@ class ResponseErrorHandler extends DecoratorHandler {
   }
 
   #checkContentType (contentType) {
-    return this.#contentType.indexOf(contentType) === 0
+    return (this.#contentType ?? '').indexOf(contentType) === 0
```
metcoder95, line 19: This looks fishy; more like a side-effect of not handling responses without `content-type` correctly.

Shall we allow them to be cached if we do not know the type?

cc: @ronag @flakey5 @mcollina

mcollina: I don't think there is any limitation of content types in caching.

Gigioliva:
> Can you please rebase/update this branch?

Done... but still there are some flaky tests

mcollina:
This now conflicts, could you update? Unfortunately that's a known flake.

mcollina:
This now conflicts, could you update? Unfortunately that's a known flake.

metcoder95 (approved):

Gigioliva:
All good? or something is missing?

mcollina (approved):
lgtm

PR #3940, opened 2024-12-10
fix(#3937): respect correct host header

### LINKED ISSUE #3937: dns interceptor doesn't set port in host header
https://github.com/nodejs/undici/blob/503ad7679118530d8ecdf0f202aad298aab4ec47/lib/interceptor/dns.js#L361

I believe this should be using `host` instead of `hostname`... otherwise we will strip the port...

metcoder95:
## This relates to...

Closes #3937

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/test/interceptors/dns.js
+++ b/test/interceptors/dns.js
@@ -124,7 +124,7 @@ test('Should respect DNS origin hostname for SNI on TLS', async t => {
   }
 
   server.on('request', (req, res) => {
-    t.equal(req.headers.host, 'localhost')
+    t.equal(req.headers.host, `localhost:${server.address().port}`)
     res.writeHead(200, { 'content-type': 'text/plain' })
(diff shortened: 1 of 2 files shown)


test/interceptors/dns.js
```diff
@@ -1731,3 +1731,86 @@ test('Should handle max cached items', async t => {
   t.equal(response3.statusCode, 200)
   t.equal(await response3.body.text(), 'hello world! (x2)')
 })
+
+test('#3937 - Handle host correctly', { only: true }, async t => {
```
mcollina, line 1735:
```suggestion
test('#3937 - Handle host correctly', async t => {
```

mcollina (approved):
lgtm with a nit

PR #3945, opened 2024-12-13
docs: enhance documentation

### LINKED ISSUE #3944: ProxyAgent ignores connect timeout options and lacks documentation for signal parameter
## Bug Description

The `connect` options passed to `ProxyAgent` are being ignored, despite the type definitions suggesting they should work. This appears to be both a potential bug and a documentation issue.

## Reproducible By

```javascript
const { ProxyAgent } = require('undici');

const agent = new ProxyAgent({
  uri: 'http://8.8.8.8:8000',
  connect: {
    signal: AbortSignal.timeout(100),
    timeout: 100,
  },
});

  dispatcher: agent,
})
  .then((res) => res.text())
  .then(console.log)
  .catch(console.error);
```

## Current Behavior

1. The `timeout` and `signal` parameters in `connect` options are not being processed when passed to `ProxyAgent`
2. The TypeScript definitions (.d.ts) show that `signal` can be passed optionally, but this is not documented
3. The only working way to add a timeout for `ProxyAgent` is through the `proxyTls` parameter, which is only mentioned in GitHub issues and not in the official documentation

## Expected Behavior

According to the documentation:

- `ProxyAgentOptions` extends `AgentOptions`
- `AgentOptions` extends `PoolOptions`
- `PoolOptions` extends `ClientOptions`
- `ClientOptions` includes `connect` parameter which accepts `ConnectOptions`

Therefore, the `timeout` and `signal` options should work when passed through the `connect` parameter.

## Documentation Issues

1. The optional `signal` parameter is present in TypeScript definitions but missing from documentation
2. The working `proxyTls` parameter solution is not documented
3. The inheritance chain of options interfaces could be more clearly documented

## Relevant Documentation Links

- [ConnectOptions Documentation](https://undici.nodejs.org/#/docs/api/Client?id=parameter-connectoptions)
- [ClientOptions Documentation](https://undici.nodejs.org/#/docs/api/Client?id=parameter-clientoptions)

## Proposed Solutions

1. Fix the handling of `connect` options in `ProxyAgent`
2. Update documentation to include:
   - Information about the `signal` parameter
   - Documentation for the `proxyTls` parameter and its usage
   - Clear explanation of how options inheritance works

metcoder95:
## This relates to...

Closes #3944 

## Rationale

We might want to adjust several options from the ProxyAgent for better scoping; but that might require a major bump on it.

## Status

- [ ] Tested
- [x] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/docs/docs/api/ProxyAgent.md
+++ b/docs/docs/api/ProxyAgent.md
@@ -15,6 +15,7 @@ Returns: `ProxyAgent`
 ### Parameter: `ProxyAgentOptions`
 
 Extends: [`AgentOptions`](/docs/docs/api/Agent.md#parameter-agentoptions)
+> It ommits `AgentOptions#connect`.
 
 * **uri** `string | URL` (required) - The URI of the proxy server.  This can be provided as a string, as an instance of the URL class, or as an object with a `uri` property of type string.
 If the `uri` is provided as a string or `uri` is an object with an `uri` property of type string, then it will be parsed into a `URL` object according to the [WHATWG URL Specification](https://url.spec.whatwg.org).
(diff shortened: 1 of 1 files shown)


Uzlopak (commented):
RSLGTM

Uzlopak (approved):

mcollina (approved):
lgtm

PR #3955, opened 2024-12-17
More cache fixes

flakey5:
- Spec compliant http date parsing
  - Also fixes invalid (NaN) values in dates just being defaulted (ex/ for `new Date('Wed, aa Dec 2024 23:20:57 GMT')`, the day defaults to 1 instead of it being marked as an invalid date)
  - Performance is about the same, still might be some room for optimization
- Heuristic caching based on `last-modified`
- Rewrote the cache test suite runner
  - Runs each environment in a child process now because the suite doesn't really like being ran multiple times in the same process
  - These are also now ran in CI
- Adds CI job to update the cache tests automatically
- Strips the cache test suite down to just what we need to run them + tell if a test is broken

With this, we're passing 65% of the tests in the suite. The two main behaviors that are left are invalidating the urls given in the `Location` header and some behaviors regarding status codes other than 200 and 307 (note these are still optional)

cc @mcollina


ronag (approved):

.github/workflows/update-cache-tests.yml
```diff
@@ -0,0 +1,45 @@
...
+    steps:
+      - name: Git Config
+        run: |
+          git config --global user.email "41898282+github-actions[bot]@users.noreply.github.com"
+          git config --global user.name "github-actions[bot]"
+
+      - name: Checkout Repository
+        uses: actions/checkout@9bb56186c3b09b4f86b1c65136769dd318469633 # v4.1.2
+
+      - name: Update WPT
```
mcollina, line 25: those are not wpt

mcollina (commented):
The cache tests are not run automatically yet as part of CI, right? Can you add them (and maybe flag the failing ones as expected failures, or just implement those behaviors)?

lib/handler/cache-handler.js
```diff
@@ -87,97 +99,109 @@ class CacheHandler {
...
-      let baseTime = now
-      if (headers['date']) {
-        const parsedDate = parseInt(headers['date'])
-        const date = new Date(isNaN(parsedDate) ? headers['date'] : parsedDate)
-        if (date instanceof Date && !isNaN(date)) {
-          baseTime = date.getTime()
-        }
-      }
+    const resAge = resHeaders.age ? getAge(resHeaders.age) : undefined
+    if (resAge && resAge >= 2147483647000) {
```
metcoder95, line 132: Shall we store this as constant for reference?

lib/handler/cache-handler.js
```diff
@@ -87,97 +99,109 @@ class CacheHandler {
...
-              handler.#writeStream = undefined
-            }
-
-            // TODO (fix): Should we resume even if was paused downstream?
-            controller.resume()
-          })
-      }
-    }
+        // TODO (fix): Should we resume even if was paused downstream?
+        controller.resume()
```
metcoder95, line 203: Let's aim to keep it in sync with the controller as possible; at the end it is what allow to apply back pressure while consuming the body

lib/handler/cache-handler.js
```diff
@@ -87,97 +99,109 @@ class CacheHandler {
...
+    }
 
-      if (typeof headers.etag === 'string' && isEtagUsable(headers.etag)) {
-        value.etag = headers.etag
-      }
+    const handler = this
+    this.#writeStream
+      .on('drain', () => controller.resume())
+      .on('error', function () {
+        // TODO (fix): Make error somehow observable?
```
metcoder95, line 194: If we cannot store the response's body, best will be to just handover the body to the caller and fail silently; I'm thinking on some sort of on `error` event here?

lib/util/date.js
```diff
@@ -0,0 +1,261 @@
...
+
+  date = date.toLowerCase()
+
+  if (date[3] === ',') {
+    return parseImfDate(date)
+  } else if (date[3] === ' ') {
+    return parseAscTimeDate(date)
+  } else {
+    return parseRfc850Date(date, now)
+  }
```
metcoder95, line 32: Maybe a switch here?
```suggestion
  switch (date[3]) {
    case ',': return parseImfDate(date)
    case ' ': return parseAscTimeDate(date)
    default: return parseRfc850Date(date, now)
  }
```

lib/handler/cache-handler.js
flakey5: > I'm thinking on some sort of on error event here?

Where would this be?

mcollina (approved):
lgtm

lib/handler/cache-handler.js
metcoder95: Good question; maybe the `cache-handler` itself? I believe we do not have reference to the dispatcher at this point but the dispatch method.

metcoder95: Let's explore it in further PR

metcoder95 (approved):

PR #3961, opened 2024-12-18
fix: records of pick callback in dns interceptor is a nested object

EAGzzyCSL:
## This relates to...

fix d.ts of dns interceptor['pick']

## Rationale

the parameter `records` pass to pick is from `this.#records`, ref [code](https://github.com/nodejs/undici/blob/65695850297550105e1034a2fdd8927ab2cb1d70/lib/interceptor/dns.js#L58C1-L64C10)

`ths.#records` is a map with type `Record<string, {records: {4: { ips: DNSInterceptorRecord[] } , 6: ...}}>`, ref [code](https://github.com/nodejs/undici/blob/65695850297550105e1034a2fdd8927ab2cb1d70/lib/interceptor/dns.js#L197C1-L213C48)

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/types/interceptors.d.ts
+++ b/types/interceptors.d.ts
@@ -20,7 +20,7 @@ declare namespace Interceptors {
     maxTTL?: number
     maxItems?: number
     lookup?: (hostname: string, options: LookupOptions, callback: (err: NodeJS.ErrnoException | null, addresses: DNSInterceptorRecord[]) => void) => void
-    pick?: (origin: URL, records: DNSInterceptorOriginRecords, affinity: 4 | 6) => DNSInterceptorRecord
+    pick?: (origin: URL, records: { records: DNSInterceptorOriginRecords }, affinity: 4 | 6) => DNSInterceptorRecord
     dualStack?: boolean
     affinity?: 4 | 6
   }


types/interceptors.d.ts
```diff
@@ -20,7 +20,7 @@ declare namespace Interceptors {
     maxTTL?: number
     maxItems?: number
     lookup?: (hostname: string, options: LookupOptions, callback: (err: NodeJS.ErrnoException | null, addresses: DNSInterceptorRecord[]) => void) => void
-    pick?: (origin: URL, records: DNSInterceptorOriginRecords, affinity: 4 | 6) => DNSInterceptorRecord
+    pick?: (origin: URL, records: { records: DNSInterceptorOriginRecords }, affinity: 4 | 6) => DNSInterceptorRecord
```
ronag, line 23: This looks like a bug in the implementation rather than the types..

EAGzzyCSL: oh no…，what should we do? Change implementation will cause a breaking change but fix types is odd.

PR #3970, opened 2024-12-23
increase timeouts for test/issue-3346.js

### LINKED ISSUE #3969: `test/issue-3356.js` very flaky when run with test runner
Reproduce with `npm run test:unit`. It almost fails every time on my machine (MacBook M2 Pro).
This makes `undici` fail CITGM runs.

mcollina:
Fixes #3969

### DIFF
--- a/test/issue-3356.js
+++ b/test/issue-3356.js
@@ -45,7 +45,7 @@ test('https://github.com/nodejs/undici/issues/3356', async (t) => {
 
   fastTimersTick()
 
-  await setTimeoutPromise(1500)
+  await setTimeoutPromise(2000)
   try {
     t.assert.equal(response.status, 200)
     // consume response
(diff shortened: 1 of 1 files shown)


mcollina:
cc @targos

targos:
It fails because the expected error doesn't happen (`await response.text()` resolves).

targos:
It works if I only increase the last timeout. I pushed some changes, but I don't know if the test is still valid that way.
My commit includes a refactoring to remove the use of tspl so that it doesn't wait for 30 seconds before failing the test in case there's a mismatch between assertions and plan.

targos:
There's something wrong with the test, I think. Even with 2 seconds, it fails (less often, but still)

targos:
Note that the timeout before the try/catch can be removed and the test will still pass if run in isolation.

mcollina:
Then let's remove that.

Timing tests are flaky by nature unfortunately :(. We can 100% skip this in CITGM and other envs.

metcoder95 (approved):

mcollina:
CLosing for now, this was skipped in citgm

PR #3977, opened 2024-12-31
fix(#3975): do not unref timeout

### LINKED ISSUE #3975: RetryAgent doesn't keep the event loop alive
## Bug Description

When using RetryAgent, Node exits while there are pending retries in flight. 

## Reproducible By

I've been playing around in a brand new node project with only `unidici` installed and trying to run the following script. This is mostly copy/pasted from the docs - I don't think I'm doing anything unorthodox here.

```js
import { request, Agent, RetryAgent } from "undici";

const resp = await request("http://httpbin.org/status/418", {
  dispatcher: new RetryAgent(new Agent(), {
    statusCodes: [418],
  }),
  headers: { "User-Agent": "teapot" },
});

console.log("status:", resp.statusCode);
console.log("headers:", resp.headers);
```

## Expected Behavior

I'd expect to get either an HTTP response or a retry error.

## Logs & Screenshots

```
$ node example.js; echo "exit code: $status"
Warning: Detected unsettled top-level await at file:///[redacted]/example.js:3
const resp = await request("http://httpbin.org/status/418", {
             ^



exit code: 13
```

### Additional context

```
$ node --version
v23.5.0
$ grep undici package.json
    "undici": "^7.2.0"
```

metcoder95:
## This relates to...

Closes #3975

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/handler/retry-handler.js
+++ b/lib/handler/retry-handler.js
@@ -133,7 +133,7 @@ class RetryHandler {
         ? Math.min(retryAfterHeader, maxTimeout)
         : Math.min(minTimeout * timeoutFactor ** (counter - 1), maxTimeout)
 
-    setTimeout(() => cb(null), retryTimeout).unref()
+    setTimeout(() => cb(null), retryTimeout)
   }
 
   onResponseStart (controller, statusCode, headers, statusMessage) {


ronag (changes requested):
It should inred if the request is aborted

metcoder95:
> It should inred if the request is aborted

Do you mean to unref if the request can be aborted?
If the request was aborted, it will exit early and not reach the timeout 🤔

ronag:
Consider the following case:

The retry timeout is active for another 60s. We wan't close the application by aborting all pending requests. With this change the application will continue hanging for 60s for no purpose.

ronag (approved):
Not perfect but probably better than current state of things...

metcoder95:
Ah ok yes, see now.
Let me see if I can set a workaround here

metcoder95:
It will require further thinking; I've tried to make it bare basic but it seems we are prone to several race conditions, so keeping it within the handler class might not sort out all the situations; let me spend more time on that later this week.

In the meanwhile, let's merge this one to unblock the issue and will continue further

blinsay:
Appreciate the quick fix!

> The retry timeout is active for another 60s. We wan't close the application by aborting all pending requests. With this change the application will continue hanging for 60s for no purpose.

Is `destroy` guaranteed to get called on the dispatcher in that scenario? Naively, it seems like stashing the timeout on the `RetryHandler` and deleting it on `destroy` seems appropriate. Not sure how hard that is though.

PR #3990, opened 2025-01-05
cache: make vary headers case-insensitive

flakey5:


metcoder95 (commented):
CI seems to not like it that much

ronag (commented):
resHeaders should already always be lowercase?

ronag (changes requested):
I think you are solving the wrong problem here. The header names are already lower case. What you want to do is `resHeaders.vary.toLowerCase()`

flakey5:
Closing in favor of #4031

PR #3992, opened 2025-01-08
fix: retry flaky test

metcoder95:
## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready


ronag:
Doesn't this kind of remove what the test is testing?

metcoder95:
Don't believe so, the scenario the testing covers is that the event loops keeps ticking and throws accordingly to a custom configuration.

The rule of thumb can be that if we change the custom configuration to something else, we continue allowing the event loop to tick and throw accordingly

ronag (approved):

PR #4001, opened 2025-01-13
fix: dual-stack retries infinite loop

luddd3:
## This relates to...

DNS interceptor

## Rationale

The DNS interceptor can be stuck in an infinite loop if dual-stack is enabled and a lookup results in a network error. It is also possible for a request to retry the same family multiple times if another request is made in-between.

## Changes

Retry a failed request **once** with dual-stack and force it to use the other ip family.
If no valid ip address is found, e.g. due to ttl timeout, no retry should be made.

## Status

- [x] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/interceptor/dns.js
+++ b/lib/interceptor/dns.js
@@ -32,7 +32,7 @@ class DNSInstance {
 
     // If full, we just return the origin
     if (ips == null && this.full) {
-      cb(null, origin.origin)
+      cb(null, origin)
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

metcoder95 (approved):

PR #4002, opened 2025-01-14
ReadableStreamFrom pull until cannot on empty enqueu

KhafraDev:
refs: https://github.com/nodejs/node/issues/56474

ReadableByteStreamController.enqueue handles empty typed arrays differently than the normal ReadableStreamController: it throws an error when receiving one. To combat this, we didn't enqueue anything and did nothing if the buffer was empty. For whatever reason, this caused the promise to never resolve (I didn't dig into the spec long enough to figure out why, sorry). The fix is to simply continue pulling until we respond to the pull in some manner.

I am not a webstreams expert please review with caution.

### DIFF
--- a/lib/core/util.js
+++ b/lib/core/util.js
@@ -600,20 +600,25 @@ function ReadableStreamFrom (iterable) {
       async start () {
         iterator = iterable[Symbol.asyncIterator]()
       },
-      async pull (controller) {
-        const { done, value } = await iterator.next()
-        if (done) {
-          queueMicrotask(() => {
-            controller.close()
-            controller.byobRequest?.respond(0)
-          })
-        } else {
-          const buf = Buffer.isBuffer(value) ? value : Buffer.from(value)
-          if (buf.byteLength) {
-            controller.enqueue(new Uint8Array(buf))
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

mcollina (changes requested):
This `controller.desiredSize > 0` is what makes backpressure possible and what makes sure we do not have a memory leak. It's critical.

KhafraDev:
pull ignores the return value which is what made me confused.

https://github.com/nodejs/node/blob/5770972dc6c0458af5458b8056b16d70905ea9d8/lib/internal/webstreams/readablestream.js#L3144-L3153

KhafraDev:
```
This method, also defined by the developer, will be called repeatedly when the stream's internal queue of chunks is not full, up
until it reaches its high water mark. If pull() returns a promise, then it won't be called again until that promise fulfills
```
https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream/ReadableStream

mcollina (approved):
lgtm

metcoder95 (approved):

PR #4004, opened 2025-01-14
fetch: make readAllBytes "sync"

Uzlopak:
Well, i just played around. I actually think that the original implementation is better. 

But maybe we can discuss this.


KhafraDev:
> I actually think that the original implementation is better.

I agree

PR #4018, opened 2025-01-20
feat: sqlite add set and minor cleanup

ronag:
set is a more performant variant of createWriteStream for small responses...

### DIFF
--- a/types/cache-interceptor.d.ts
+++ b/types/cache-interceptor.d.ts
@@ -90,7 +90,7 @@ declare namespace CacheHandler {
     headers: Record<string, string | string[]>
     vary?: Record<string, string | string[]>
     etag?: string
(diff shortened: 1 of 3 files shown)


types/cache-interceptor.d.ts
```diff
@@ -90,7 +90,7 @@ declare namespace CacheHandler {
     headers: Record<string, string | string[]>
     vary?: Record<string, string | string[]>
     etag?: string
-    body: null | Readable | Iterable<Buffer> | AsyncIterable<Buffer> | Buffer | Iterable<string> | AsyncIterable<string> | string
+    body?: Readable | Iterable<Buffer> | AsyncIterable<Buffer> | Buffer | Iterable<string> | AsyncIterable<string> | string
```
metcoder95, line 93: This wouldn't be breaking if we drop `null`?

metcoder95 (approved):
lgtm

mcollina:
CI is not happy

Issue #3959, opened 2024-12-17
Response headers `vary: Accept-Encoding` is stopping undici from caching the http request


SukkaW:
## Bug Description

I was trying to try out undici cache with the following code snippets:

```ts
import undici, {
  interceptors,
  Agent,
  setGlobalDispatcher
} from 'undici';

setGlobalDispatcher(agent.compose(
  interceptors.cache({
    store: new undici.cacheStores.MemoryCacheStore()
  })
));

(async () => {
  console.time('fetch');
  await undici.request('https://easylist.to/easylist/easylist.txt');
  console.timeEnd('fetch');

  console.time('fetch 2');
  await undici.request('https://easylist.to/easylist/easylist.txt');
  console.timeEnd('fetch 2');
})();
```

However, I noticed that the request is never cached.

## Reproducible By

See code snippet above.

## Expected Behavior

The request should be cached and the `fetch 2` should happen "instantly".

## Environment

Node.js v22.11.0 on Darwin

### Additional context

I noticed that `https://easylist.to/easylist/easylist.txt` is returning the `Vary: Accept-Encoding` response header regardless of whether `Accept-Encoding` is included in the request header. This causes undici to panic because undici doesn't expect `Accept-Encoding` to be included in the `cacheKey`:

https://github.com/nodejs/undici/blob/2c3644247a53132267de2cf09cd3dbe3600c21d3/lib/handler/cache-handler.js#L133-L137

cc @flakey5

SukkaW:
It seems that Fastly CDN will always returns the "Vary: Accept-Encoding" header whether the HTTP client is sending `Accept-Encoding` request header or not. This is not an issue for browsers because they always send `Accept-Encoding` request header, but this becomes a problem for undici.

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

SukkaW:
> Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

I am not sure of the proper way to fix this:

We could simply ignore any missing request headers (those undici are not sending) when parsing the `Vary` response header, but I am afraid that later requests with those request headers presented will result in cache poisoning.

Or we could use a placeholder for those missing request headers, but the placeholder couldn't be a symbol because it can't be properly serialized for storing.

flakey5:
[...]

PR #4031, opened 2025-01-27
fix: handle missing vary header values

gurgunday:
fixes: https://github.com/nodejs/undici/issues/3959

### DIFF
--- a/lib/util/cache.js
+++ b/lib/util/cache.js
@@ -26,10 +26,14 @@ function makeCacheKey (opts) {
       if (typeof key !== 'string' || typeof val !== 'string') {
         throw new Error('opts.headers is not a valid header map')
       }
-      headers[key] = val
+      headers[key.toLowerCase()] = val
     }
   } else if (typeof opts.headers === 'object') {
-    headers = opts.headers
+    headers = {}
+
+    for (const key of Object.keys(opts.headers)) {
+      headers[key.toLowerCase()] = opts.headers[key]
+    }
   } else {
     throw new Error('opts.headers is not an object')
   }
@@ -260,19 +264,16 @@ function parseVaryHeader (varyHeader, headers) {
     return headers
   }
 
-  const output = /** @type {Record<string, string | string[]>} */ ({})
+  const output = /** @type {Record<string, string | string[] | null>} */ ({})
 
   const varyingHeaders = typeof varyHeader === 'string'
     ? varyHeader.split(',')
     : varyHeader
+
   for (const header of varyingHeaders) {
     const trimmedHeader = header.trim().toLowerCase()
 
-    if (headers[trimmedHeader]) {
-      output[trimmedHeader] = headers[trimmedHeader]
-    } else {
-      return undefined
-    }
+    output[trimmedHeader] = headers[trimmedHeader] ?? null
   }
 
   return output
--- a/test/issue-3959.js
+++ b/test/issue-3959.js
@@ -1,0 +1,65 @@
+const { describe, test, after } = require('node:test')
+const assert = require('node:assert')
+const { createServer } = require('node:http')
+const MemoryCacheStore = require('../lib/cache/memory-cache-store.js')
+const SqliteCacheStore = require('../lib/cache/sqlite-cache-store.js')
+const { request, Agent, setGlobalDispatcher } = require('..')
+const { interceptors } = require('..')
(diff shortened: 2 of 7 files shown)


gurgunday:
I'd honestly prefer using `undefined` instead of `null` to indicate the absence of a string, is anyone against that?

lib/util/cache.js
```diff
@@ -26,10 +26,14 @@ function makeCacheKey (opts) {
...
-      headers[key] = val
+      headers[key.toLowerCase()] = val
     }
   } else if (typeof opts.headers === 'object') {
-    headers = opts.headers
+    headers = {}
+
+    for (const key of Object.keys(opts.headers)) {
+      headers[key.toLowerCase()] = opts.headers[key]
+    }
```
gurgunday, line 36: If we don't do this, we lose case insensitivity in the cache unfortunately and the unit test fails

test/issue-3959.js
```diff
@@ -0,0 +1,55 @@
+const { describe, test } = require('node:test')
```
mcollina, line 1:
```suggestion
'use strict'

const { describe, test } = require('node:test')
```

test/utils/cache.js
```diff
@@ -0,0 +1,39 @@
+const { describe, test } = require('node:test')
```
mcollina, line 1:
```suggestion
'use strict'

const { describe, test } = require('node:test')
```

lib/util/cache.js
mcollina: why? I thought headers name were already lowercased elsewhere?

gurgunday: The values in the cache are lowercase but opts.headers keys are not

In the last request of the unit test we add Accept-Encoding and it's value in the opts object is still Accept-Encoding, so the cache thinks the third request still doesn't have an accept-encoding property and uses the cache

gurgunday: In other words, this request uses the cache when it shouldn't because it trues to check the cached accept-encoding key with the passed Accept-Encoding

https://github.com/nodejs/undici/pull/4031/files#diff-f8735499820cf76c057aaee30fe0717b6b9ed5407d6028e3695fbb111beddf0dR48

metcoder95: Yeah, the response headers are already lowercased once parsed, but we do not handle the ones passed by the call to `dispatch`; those can be potentially upper cased

test/issue-3959.js
```diff
@@ -0,0 +1,57 @@
...
+    const res3 = await request(url, {
+      headers: {
+        'Accept-Encoding': 'gzip'
+      }
+    })
+    const body3 = await res3.body.text()
+    assert.strictEqual(body3, 'Request count: 2')
+    assert.strictEqual(requestCount, 2)
+
+    await new Promise(resolve => server.close(resolve))
```
metcoder95, line 55: `nit`: you can use `after` for this

metcoder95 (approved):

gurgunday:
@KhafraDev can you take a look as well?

This issue recently bit me too

gurgunday:
Cc @flakey5, just saw your PR but leaving it here if you want to merge this here

It also both the undercase problem and the cache issue

metcoder95 (approved):

test/utils/cache.js
```diff
@@ -0,0 +1,41 @@
+'use strict'
+
+const { describe, test } = require('node:test')
+const assert = require('node:assert')
+const { parseVaryHeader } = require('../../lib/util/cache.js')
+
+describe('parseVaryHeader', () => {
```
flakey5, line 7: nit but could you move these to where the other tests are for this function? https://github.com/nodejs/undici/blob/a9176c9ec67b3efa1732799d18162c3a4307164e/test/cache-interceptor/utils.js#L179

flakey5 (changes requested):
The issue looks to be still present when using the sqlite cache store?
image

gurgunday:
Yeah had to go back to null, it works better 😁

flakey5 (approved):

metcoder95 (approved):

gurgunday:
Rebased

mertcanaltin (approved):
lgtm

mcollina (approved):
lgtm

PR #4048, opened 2025-02-07
address bad usage of TST

ywave620:
## Rationale

Inserting keys to TST in order yields the worst performance of this data structure. It is similar to inserting keys to a binary search tree in order makes it degenerate to a link list.

Randomize the insertion order simply works, but we do not want any potential instable performance, so I take a snapshot of one randomized insertion order that works well

WeChatWorkScreenshot_ea6ff026-5e0d-44b5-8dce-4da5858291c9

## Changes

Randomize the insertion order of the well known header keys

### Features

N/A

### Bug Fixes

N/A

N/A

## Status

- [x] Tested
- [S] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready


lib/core/tree.js
```diff
@@ -149,8 +149,38 @@ class TernarySearchTree {
...
+const randomIndexOfWellknownHeaderNames = [
+  33, 61, 78, 55, 28, 2, 65, 68, 62, 74, 91, 71,
+  31, 85, 58, 41, 43, 70, 34, 57, 66, 29, 3, 50,
+  8, 39, 47, 54, 80, 64, 93, 26, 75, 63, 22, 20,
+  90, 88, 17, 19, 92, 18, 1, 77, 67, 79, 6, 42,
+  76, 23, 7, 84, 72, 89, 81, 37, 27, 69, 4, 87,
+  94, 86, 15, 9, 56, 25, 38, 49, 12, 73, 24, 60,
+  46, 11, 14, 83, 40, 52, 13, 21, 48, 30, 45, 10,
+  53, 5, 59, 44, 35, 16, 36, 32, 0, 51, 82
+]
```
mcollina, line 175: Can you move it to a separate file, or can it be easily generated by a script?

lib/core/tree.js
```diff
@@ -149,8 +149,38 @@ class TernarySearchTree {
...
+//     const j = Math.floor(Math.random() * (i + 1));
+//     [array[i], array[j]] = [array[j], array[i]]
+//   }
+//   return array
+// }
+// const indexes = []
+// wellknownHeaderNames.forEach((key, idx) => {
+//   indexes.push(idx)
+// })
+// randomIndexOfWellknownHeaderNames = shuffleArray(indexes)
```
mcollina, line 163: why is this commented out? Can you nmove this to a script or another js file instead?

ywave620: I think we do not want any potential instable performance, right? (even the chance is low :) ) so I take a snapshot of one randomized insertion order that works well

ywave620: Maybe I just remove them from codebase outright and put it the PR

lib/core/tree.js
```diff
@@ -149,8 +149,38 @@ class TernarySearchTree {
 
 const tree = new TernarySearchTree()
 
-for (let i = 0; i < wellknownHeaderNames.length; ++i) {
-  const key = headerNameLowerCasedRecord[wellknownHeaderNames[i]]
+// function shuffleArray (array) {
```
metcoder95, line 152: The plan is to shuffle them upon first script evaluation or how exactly?

ywave620: My plan is to generate statically(before any script evaluation), and I include how to generate it in a file that will not be loaded. Is this OK?

metcoder95: So, it will need to be generated before actually publishing the library?

I'm ok either way, I believe I find generating the array a single time upon script valuation a bit easier to maintain, tho don't have a strong preference.

ywave620:
@metcoder95 could you help move this forward

tsctx:
If not properly shuffled to balance the left and right sides, it will cause a bad case and further degrade performance. This is similar to the problem with binary trees.

ywave620:
> If not properly shuffled to balance the left and right sides, it will cause a bad case and further degrade performance. This is similar to the problem with binary trees.

Given the array is sorted, randomly retrieving item from it and inserting to a TST/Binary search tree should yield a balanced tree. And the bottom line is that we definitely don't want to insert items to a TST in order because this is the worst case of this data structure.

Maybe I can print out the TST and do a comparison?  I did a non formal one before and it turned out that when searching for `Via` we actually saved a lot of node traversing

metcoder95:
Seems CI is not happy, can you try to rebase with `main` and adjust?

mcollina (approved):
lgtm

mcollina:
However I would prefer if:

1. a script to generate a "working well" snapshot would be placed in `scripts`
2. a test that verifies that the snapshots contains all keys is added

PR #4049, opened 2025-02-07
feat: mark `EnvHttpProxyAgent` as stable

aduh95:
## Rationale

Looking at the issue tracker and the file history, it looks like the implementation has been quite stable since it was introduced, so I suggest we make the docs reflect that.

## Changes

Remove the "Experimental" mention in the docs.
Remove the experimental warning

### Features

N/A

### Bug Fixes

N/A

N/A

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/docs/docs/api/EnvHttpProxyAgent.md
+++ b/docs/docs/api/EnvHttpProxyAgent.md
@@ -1,7 +1,5 @@
 # Class: EnvHttpProxyAgent
 
-Stability: Experimental.
-
 Extends: `undici.Dispatcher`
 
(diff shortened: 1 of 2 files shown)


metcoder95 (approved):

mcollina (approved):
lgtm

@aduh95 do you think we should make this the default in Node?

aduh95:
> do you think we should make this the default in Node?

I don't think I have an informed enough opinion 🤔 maybe it's something we could mention during a Collaboration Summit

PR #4064, opened 2025-02-21
fix: fix EnvHttpProxyAgent for the Node.js bundle

joyeecheung:
The Dispatcher needs some methods from lib/api for EnvHttpProxyAgent, otherwise it's incomplete.

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/index-fetch.js
+++ b/index-fetch.js
@@ -26,6 +26,9 @@ module.exports.createFastMessageEvent = createFastMessageEvent
 
 module.exports.EventSource = require('./lib/web/eventsource/eventsource').EventSource
 
+const api = require('./lib/api')
+const Dispatcher = require('./lib/dispatcher/dispatcher')
+Object.assign(Dispatcher.prototype, api)
 // Expose the fetch implementation to be enabled in Node.js core via a flag
 module.exports.EnvHttpProxyAgent = EnvHttpProxyAgent
 module.exports.getGlobalDispatcher = getGlobalDispatcher


index-fetch.js
```diff
@@ -26,6 +26,9 @@ module.exports.createFastMessageEvent = createFastMessageEvent
 
 module.exports.EventSource = require('./lib/web/eventsource/eventsource').EventSource
 
+const api = require('./lib/api')
```
joyeecheung, line 29: From my local testing, it seems only api-connect is missing. But my testing isn't very thorough, so I am not entirely sure if other methods may also be needed in other paths, in any case it seems a bit of a footgun to leave the Dispatcher partially complete in the Node.js bundle, so I included them all. I am not sure if it's against any principle to include the `lib/api` parts in the Node.js, bundle though. Locally it increases the bundle size from 610KB+ to 650KB+

joyeecheung: Also for context, this is done for the public bundle https://github.com/nodejs/undici/blob/ff18d8c9b213e1ea95f7e9917afcf8fc056b8be7/index.js#L25 (and funny enough if seems corepack once used internal paths instead of the bundle, so there were issues like https://github.com/nodejs/corepack/issues/417

mcollina (approved):
lgtm

mcollina:
Can you add a test for it?

metcoder95 (approved):

mertcanaltin (approved):
LGTM

joyeecheung:
Added a test - I didn't use mocks but spin up two servers to try to exercise more paths, in case the mocks hide any real issues in the paths that would be otherwise exercised if used for real. Also I just wrote the test to use them together since https://github.com/nodejs/node/pull/57165 doesn't need them exposed individually.

mcollina (approved):
lgtm

joyeecheung:
Depends on https://github.com/nodejs/node/pull/57236

PR #4070, opened 2025-02-28
Allow disabling autoSelectFamily in an Agent

hitsthings:
Currently the options are only passed if they enable autoSelectFamily. However, if autoSelectFamily is the default, we want to be able to disable it too.

Noticed this while debugging https://github.com/nodejs/node/issues/47822#issuecomment-2689454370

## This relates to...

https://github.com/nodejs/node/issues/47822#issuecomment-2689454370

## Rationale

Currently this claims to accept a boolean but really only accepts `true`.

## Changes

React to the autoSelectFamily option if it is passed in.

### Features

Agent's can disable autoSelectFamily

None?

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/types/client.d.ts
+++ b/types/client.d.ts
@@ -70,7 +70,7 @@ export declare namespace Client {
     /** TODO */
     maxRedirections?: number;
     /** TODO */
-    connect?: buildConnector.BuildOptions | buildConnector.connector;
+    connect?: Partial<buildConnector.BuildOptions> | buildConnector.connector;
     /** TODO */
     maxRequestsPerClient?: number;
     /** TODO */
--- a/lib/dispatcher/client.js
+++ b/lib/dispatcher/client.js
@@ -207,7 +207,7 @@ class Client extends DispatcherBase {
         allowH2,
         socketPath,
         timeout: connectTimeout,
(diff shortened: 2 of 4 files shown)


types/client.d.ts
```diff
@@ -70,7 +70,7 @@ export declare namespace Client {
     /** TODO */
     maxRedirections?: number;
     /** TODO */
-    connect?: buildConnector.BuildOptions | buildConnector.connector;
+    connect?: Partial<buildConnector.BuildOptions> | buildConnector.connector;
```
hitsthings, line 73: `new Agent({ connect: { autoSelectFamily: true } }) is not actually considered valid currently because it doesn't specify required properties of `buildConnector.BuildOptions`. But the rest of the options come from elsewhere (hence why `connect` itself is not a required property) so you don't need to specify everything.

metcoder95 (approved):

hitsthings:
Let me know if I need to do something about those tests. They seem unrelated to this afaict.

mcollina (approved):
lgtm

PR #4084, opened 2025-03-06
fix: Fix retry-handler.js when retry-after header is a Date

fgiova:
## Rationale

Fix retry-handler when retry-after header is a Date

## Changes

Fix using original header retry-after for calculate timeout instead header casted to Number

### Bug Fixes

Fix retry-handler when retry-after header is a Date
Fix test related using longer timeout 

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/handler/retry-handler.js
+++ b/lib/handler/retry-handler.js
@@ -11,8 +11,8 @@ const {
 } = require('../core/util')
 
 function calculateRetryAfterHeader (retryAfter) {
-  const current = Date.now()
-  return new Date(retryAfter).getTime() - current
+  const retryTime = new Date(retryAfter).getTime()
+  return isNaN(retryTime) ? 0 : retryTime - Date.now()
 }
 
 class RetryHandler {
@@ -124,7 +124,7 @@ class RetryHandler {
     if (retryAfterHeader) {
       retryAfterHeader = Number(retryAfterHeader)
       retryAfterHeader = Number.isNaN(retryAfterHeader)
-        ? calculateRetryAfterHeader(retryAfterHeader)
+        ? calculateRetryAfterHeader(headers['retry-after'])
         : retryAfterHeader * 1e3 // Retry-After is in seconds
     }
 
--- a/test/interceptors/retry.js
+++ b/test/interceptors/retry.js
@@ -191,19 +191,19 @@ test('Should use retry-after header for retries (date)', async t => {
   server.on('request', (req, res) => {
     switch (counter) {
       case 0:
+        checkpoint = Date.now()
         res.writeHead(429, {
           'retry-after': new Date(
-            new Date().setSeconds(new Date().getSeconds() + 1)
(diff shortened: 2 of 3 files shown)


lib/handler/retry-handler.js
```diff
@@ -124,7 +124,7 @@ class RetryHandler {
     if (retryAfterHeader) {
       retryAfterHeader = Number(retryAfterHeader)
       retryAfterHeader = Number.isNaN(retryAfterHeader)
-        ? calculateRetryAfterHeader(retryAfterHeader)
+        ? calculateRetryAfterHeader(headers['retry-after'])
```
metcoder95, line 127: Let's add a proper check and parsing for it

metcoder95 (commented):
Can you add a testing for it?

lib/handler/retry-handler.js
fgiova: Into tests?

fgiova:
> Can you add a testing for?

I also fixed the related test; it now waits only 1ms, even though the date parser has failed. In fact Number(iso-date) has become NaN on function to evaluate retry-after header.

lib/handler/retry-handler.js
metcoder95: I was meaning here.

The `retryAfterHeader` is being inferred in L123, it does some checks, and later it pass it down in L127 where you are replacing it with a direct access to the header.

My suggestion is, when is not a number, validate is a valid Date string, and parse it into seconds for usage.

fgiova: OK, now I understand, I'm going to do a check according to [RFC 7231](https://datatracker.ietf.org/doc/html/rfc7231#section-7.1.1.1). 
Is it acceptable in your opinion to include this check in the calculateRetryAfterHeader function?

Shall I also add an explicit failure test?

fgiova: A doubt:
The current form of calculateRetryAfterHeader returns a negative integer if the date is not well formatted.
This suggestion returns a zero value.
Perhaps something like:
```javascript
function calculateRetryAfterHeader (retryAfter) {
  const rfcHttpDateRegex = /^(Mon|Tue|Wed|Thu|Fri|Sat|Sun), ([0-3][0-9]) (Jan|Feb|Mar|Apr|May|Jun|Jul|Aug|Sep|Oct|Nov|Dec) ([0-9]{4}) ([01][0-9]|2[0-3])(:[0-5][0-9]){2} GMT$/
  return rfcHttpDateRegex.test(retryAfter) ? new Date(retryAfter).getTime() - Date.now() : 0
}
```
Or leave the function as it is?

fgiova: updated pull request

metcoder95 (approved):

mcollina (approved):
lgtm

Issue #4120, opened 2025-03-26
Update DNS caching example?


GeoffreyBooth:
Despite the outcome of https://github.com/nodejs/undici/issues/2440, it seems like DNS caching is already part of Undici, via https://undici.nodejs.org/#/docs/api/Dispatcher?id=dns:

> The `dns` interceptor enables you to cache DNS lookups for a given duration, per origin.

Assuming I understand this correctly, the example at https://undici.nodejs.org/#/examples/?id=cacheable-dns-lookup should probably be updated to use the built-in `dns` interceptor instead of `cacheable-lookup`.

Here is an example from my code; improvement suggestions welcome:

```js
import { Agent, interceptors, setGlobalDispatcher } from 'undici'

// Interceptors to add response caching, DNS caching and retrying to the dispatcher
const { cache, dns, retry } = interceptors

const defaultDispatcher = new Agent({
	connections: 100, // https://github.com/nodejs/undici/issues/3221
	headersTimeout: 10_000,
	bodyTimeout: 10_000,
}).compose(cache(), dns(), retry())

setGlobalDispatcher(defaultDispatcher)
```

Perhaps the example in the docs should be updated to something similar to this? And maybe rather than showing only DNS caching, it might be a better example to show multiple interceptors for increasing robustness; or more generally, I would find it useful for the docs to include a recommendation for how to configure the default dispatcher for high-traffic production applications.

GeoffreyBooth:
> Would you like to send a PR?

Sure. Is what I wrote above correct? The `dns` interceptor does what I think it does?

And is there a better dispatcher configuration that you would recommend for production apps?

mcollina:
No, that's pretty similar to what we run/recommend.

GeoffreyBooth:
If the example in the docs includes the options I chose here, `connections` and `headersTimeout` / `bodyTimeout`, the docs should include some explanation for why they should be specified. The timeout ones seem easy enough to explain; just take a guess as to what the reasonable upper bound would be for the latency of the servers you’re calling. For `connections`, though, my understanding from https://github.com/nodejs/undici/issues/3221 is a bit hazy and seems to boil down to just “pick the highest value that seems to work for your hardware” which feels like a bit hand-wavy of an explanation. Is there a better guideline you would recommend, or should I just leave `connections` out of the example?

mcollina:
>  Is there a better guideline you would recommend, or should I just leave connections out of the example?

For most systems, 100 is a good number.

PR #4127, opened 2025-03-29
Update DNS caching example to include other interceptors, production configuration

GeoffreyBooth:
Resolves https://github.com/nodejs/undici/issues/4120#issuecomment-2753713036. Updates the DNS caching example to use Undici’s own DNS interceptor rather than recommending an external library; and adds the retry and response caching interceptors to make the example more representative of a typical production configuration. cc @mcollina

### DIFF
--- a/docs/examples/README.md
+++ b/docs/examples/README.md
@@ -126,17 +126,21 @@ async function deleteRequest (port = 3001) {
 }
 ```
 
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

metcoder95 (approved):

PR #4128, opened 2025-03-30
feat(#4086): proxy keep alive

### LINKED ISSUE #4086: Support proxy keep-alive
As a side-effect of the investigation of @joyeecheung in #4083, it seems we should be doing proxy keep-alive, while we are not.

Undici:

```
----- Received Request -----
host: example.com
connection: close
```

Curl:

```
----- Received Request -----
host: example.com:443
user-agent: curl/8.7.1
proxy-connection: Keep-Alive 
```

metcoder95:
## This relates to...

Closes #4086

## Rationale

Do we see a potential harm of this? I know the original idea was to tell the queue not to re-use the socket so it can be used by the caller instead.

From tests, everything seems normal, but wondering if there's something possibly that I'm not seeing to not do it.

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready


mcollina (changes requested):
this is not enough to close that, as one header is missing. A test should also be added for that specific case.

mcollina (approved):
lgtm

Issue #4124, opened 2025-03-28
database is locked


ronag:
```
Error: database is locked
    at get size (file:///usr/src/app/node_modules/@nxtedition/nxt-undici/lib/sqlite-cache-store.js:298:47)
    at #prune (file:///usr/src/app/node_modules/@nxtedition/nxt-undici/lib/sqlite-cache-store.js:270:14)
    at SqliteCacheStore.set (file:///usr/src/app/node_modules/@nxtedition/nxt-undici/lib/sqlite-cache-store.js:238:16)
    at CacheHandler.onComplete (file:///usr/src/app/node_modules/@nxtedition/nxt-undici/lib/interceptor/cache.js:159:19)
    at Handler.onComplete (file:///usr/src/app/node_modules/@nxtedition/nxt-undici/lib/utils.js:263:40)
    at Handler.onComplete (file:///usr/src/app/node_modules/@nxtedition/nxt-undici/lib/utils.js:263:40)
    at Handler.onComplete (file:///usr/src/app/node_modules/@nxtedition/nxt-undici/lib/interceptor/response-retry.js:202:20)
    at Handler.onComplete (file:///usr/src/app/node_modules/@nxtedition/nxt-undici/lib/utils.js:263:40)
    at Handler.onComplete (file:///usr/src/app/node_modules/@nxtedition/nxt-undici/lib/interceptor/dns.js:22:11)
    at Handler.onComplete (file:///usr/src/app/node_modules/@nxtedition/nxt-undici/lib/utils.js:263:40)
```

Something with the SQL database? I guess prune should at least catch and ignore the error?

ronag:
Not sure when/how this can occur.

mcollina:
Might it be possible that you are running multiple servers connecting to the same SQLite instance? See https://www2.sqlite.org/cvstrac/wiki?p=DatabaseIsLocked for examples.

ronag:
Isn't the idea that the sql cache should work cross process? We should probably handle errors like these?

mcollina:
>  I thought sqlite was designed not to have this issue?

Unfortunately no, you can have this issue in SQLite as the link above describe. However, this shouldn't really be possible with how _we_ are using SQLite in the cache store.

ronag:
>  However, this shouldn't really be possible with how we are using SQLite in the cache store.

Question is how I manage to trigger this then :D.

metcoder95:
[...]

PR #4139, opened 2025-04-02
cache: more efficient pruning

ronag:
Reduce transaction lock times.

I suspect there is a non-trivial overhead to `countEntriesQuery`.

Refs: https://github.com/nodejs/undici/issues/4124

### DIFF
--- a/lib/cache/sqlite-cache-store.js
+++ b/lib/cache/sqlite-cache-store.js
@@ -63,14 +63,11 @@ module.exports = class SqliteCacheStore {
   #deleteByUrlQuery
 
   /**
-   * @type {import('node:sqlite').StatementSync}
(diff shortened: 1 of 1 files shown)


ronag:
Hmmm... offset is not a good solution either...

lib/cache/sqlite-cache-store.js
```diff
@@ -213,12 +206,17 @@ module.exports = class SqliteCacheStore {
             id
           FROM cacheInterceptorV${VERSION}
           ORDER BY cachedAt DESC
-          LIMIT ?
+          OFFSET ${this.#maxCount}
```
ronag, line 209: This is also problematic...

PR #4142, opened 2025-04-03
cache: more efficient sqlite indices

ronag:
- Drop unnecessary single-column indices (url, method, start alone), as the composite index effectively covers them.

- Avoid redundant indices that can slow down inserts without additional query benefits.


mcollina (approved):
lgtm

Should we bump the version?

mcollina:
There are some failing wpts, which seems odd.

mcollina (approved):
lgtm

mcollina:
https://github.com/nodejs/undici/pull/4145 should cover for the failing tests

Issue #4146, opened 2025-04-04
Bug: MockAgent `intercept`s don't handle params with square brackets


dario-piotrowicz:
## Bug Description

`MockAgent#intercept` doesn't seem to support query params with square brackets for requests such as:
`https://example.com/qux-c?arraykey[]=a&arraykey[]=b`

## Reproducible By

You can fine a minimal reproduction of this issue here: https://github.com/dario-piotrowicz/undici-square-params-repro

## Expected Behavior

Such parameters should be supported, meaning that requests to `https://example.com/qux-c?arraykey=a&arraykey=b` and 
`https://example.com/qux-c?arraykey[]=a&arraykey[]=b` should work practically in the same exact way

## Environment

Mac OS Sequoia 15.3.2 

### Additional context

It seems like there is no consensus on the syntax for array search params ([medium article](https://medium.com/raml-api/arrays-in-query-params-33189628fa68)) but that this syntax with square brackets is generally supported. (Someone also told me that there are/where some php frameworks that would only accept this syntax).

I've encountered this when trying to unit test code using the [`stripe-node` package](https://github.com/stripe/stripe-node), it seems like the stripe platform enforces such syntax ([stripe docs reference](https://docs.stripe.com/api/expanding_objects)) so this is not really work-aroundable as far as I can tell.

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

dario-piotrowicz:
> Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

Hey @mcollina thanks for the prompt reply, sure I'm happy to give it a shot 😄

PR #4148, opened 2025-04-04
feat: add new `acceptNonStandardSearchParameters` MockAgent option

dario-piotrowicz:
## Rationale

 
This PR adds a new `acceptNonStandardSearchParameters` option to make
instances of `MockAgent` accept search parameters specified
using non-standard syntaxes such as multi-value items specified with `[]`
(e.g. `param[]=1&param[]=2&param[]=3`) and multi-value items which values
are comma separated (e.g. `param=1,2,3`) 

With these changes `MockAgent` instances can be enabled to intercept requests
such as `https://example.com/qux-c?arraykey[]=a&arraykey[]=b`, etc...

## Changes

MockAgents created with the `acceptNonStandardSearchParameters` "normalize" requests such as `https://example.com/qux-c?arraykey[]=a&arraykey[]=b` into ones like
`https://example.com/qux-c?arraykey=a&arraykey=b` so that they can then be handled normally

### Features

N/A

### Bug Fixes

Fixes https://github.com/nodejs/undici/issues/4146

None

## Status

- [x] Tested
- [S] Documented
- [x] Review ready
- [x] In review
- [ ] Merge ready

___

> [!Note]
> My implementation supports the case of array values having `[]` (e.g. `a[]=1,a[]=2,a[]=3`) and also different array values separated by commas (e.g. `a=1,2,3`) and that's pretty much it, it could potentially support other syntaxes like the JSON API and Bitbucket API mentioned in https://medium.com/raml-api/arrays-in-query-params-33189628fa68, but that feels to me like quite an overkill right now. I am thinking that probably the cases I am adding support for now should cover most use cases anyways and if the extra ones will be needed in the future (someone opening an issue, etc...) they can always be incrementally added as needed on top of my changes :thinking:

### DIFF
--- a/lib/mock/mock-utils.js
+++ b/lib/mock/mock-utils.js
@@ -92,13 +92,42 @@ function matchHeaders (mockDispatch, headers) {
   return true
 }
 
+function normalizeSearchParams (query) {
+  if (typeof query !== 'string') {
+    return query
+  }
+
+  const originalQp = new URLSearchParams(query)
+  const normalizedQp = new URLSearchParams()
+
+  for (let [key, value] of originalQp.entries()) {
+    key = key.replace('[]', '')
+
+    const valueRepresentsString = /^(['"]).*\1$/.test(value)
+    if (valueRepresentsString) {
+      normalizedQp.append(key, value)
+      continue
+    }
+
+    if (value.includes(',')) {
+      const values = value.split(',')
+      for (const v of values) {
+        normalizedQp.append(key, v)
+      }
+      continue
+    }
+
+    normalizedQp.append(key, value)
+  }
+
+  return normalizedQp
+}
+
 function safeUrl (path) {
   if (typeof path !== 'string') {
     return path
   }
-
   const pathSegments = path.split('?', 3)
-
   if (pathSegments.length !== 2) {
     return path
   }
@@ -376,6 +405,10 @@ function buildAndValidateMockOptions (opts) {
       throw new InvalidArgumentError('options.enableCallHistory must to be a boolean')
     }
 
+    if ('acceptNonStandardSearchParameters' in mockOptions && typeof mockOptions.acceptNonStandardSearchParameters !== 'boolean') {
+      throw new InvalidArgumentError('options.acceptNonStandardSearchParameters must to be a boolean')
+    }
+
     return mockOptions
   }
 }
@@ -395,5 +428,6 @@ module.exports = {
   checkNetConnect,
   buildAndValidateMockOptions,
   getHeaderByName,
-  buildHeadersFromArray
+  buildHeadersFromArray,
+  normalizeSearchParams
 }
--- a/test/mock-utils.js
+++ b/test/mock-utils.js
@@ -9,7 +9,8 @@ const {
   getResponseData,
   getStatusText,
   getHeaderByName,
-  buildHeadersFromArray
+  buildHeadersFromArray,
+  normalizeSearchParams
 } = require('../lib/mock/mock-utils')
 
 test('deleteMockDispatch - should do nothing if not able to find mock dispatch', (t) => {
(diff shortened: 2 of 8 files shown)


test/mock-utils.js
```diff
@@ -242,3 +243,33 @@ describe('buildHeadersFromArray', () => {
     t.strictEqual(headers.key, 'value')
   })
 })
+
+describe.only('normalizeQueryParams', () => {
```
metcoder95, line 247:
```suggestion
describe('normalizeQueryParams', () => {
```

dario-piotrowicz: oops... sorry, thanks for noticing it :slightly_smiling_face: :pray:

metcoder95 (approved):

lib/mock/mock-utils.js
```diff
@@ -103,8 +135,7 @@ function safeUrl (path) {
     return path
   }
 
-  const qp = new URLSearchParams(pathSegments.pop())
-  qp.sort()
+  const qp = normalizeQueryParams(pathSegments.pop())
```
mcollina, line 138: Can you put this behind an option? This behavior is non-standard.

dario-piotrowicz: of course, I will look into doing that :slightly_smiling_face::+1: 

my reasoning was that this is more of additive support that allows the other syntaxes without taking anything away from people not caring about/using them, but I guess that it'd be cleaner not to include them by default more from a spec-compliant/correctness point of view? :thinking:

dario-piotrowicz: I've put the logic in `MockAgent` under a new `acceptNonStandardSearchParameters` option, please have a look and let me know how this looks :pray: 

(PS: I still need to add this to the docs, but before doing that I wanted to check with you to see if I am heading to the right direction :slightly_smiling_face:)

lib/mock/mock-utils.js
```diff
@@ -92,13 +92,43 @@ function matchHeaders (mockDispatch, headers) {
...
+      for (const v of values) {
+        normalizedQp.append(key, v)
+      }
+      continue
+    }
+
+    normalizedQp.append(key, value)
+  }
+
+  normalizedQp.sort()
```
metcoder95, line 123: The sorting is necessary?

metcoder95: I've the impression that might cause unexpected results if not documented

dario-piotrowicz: No it's not necessary :slightly_smiling_face: , thanks for catching it, it's simply a leftover of my previous iteration

removed :+1:

metcoder95 (approved):

types/mock-agent.d.ts
```diff
@@ -59,6 +59,9 @@ declare namespace MockAgent {
     /** Ignore trailing slashes in the path */
     ignoreTrailingSlash?: boolean;
 
+    /** Accept URLs with search parameters using non standard syntaxes. default false */
+    acceptNonStandardSearchParameters?: boolean;
```
mcollina, line 63: Can you add a test for this type? We use tsd.

dario-piotrowicz: Sorry, what do you mean? are these not enough?

https://github.com/nodejs/undici/blob/716fb9447ec29095215b4e566b1e9120b1b64fd7/test/mock-agent.js#L2892-L2982

dario-piotrowicz: sure, sorry I missed that 🙂👍

dario-piotrowicz: Done :+1: 

is this ok? :slightly_smiling_face:

https://github.com/nodejs/undici/pull/4148/commits/cf7daab4e54a5a8b0fffbaba63a0fb02eae9faf1

mcollina (approved):
lgtm

mcollina:
Actually, can you add docs for this?

dario-piotrowicz:
thanks @mcollina 🫶 🚀 

Of course I'll add the docs right away, I was waiting to hear your thoughts on the API before doing that (as I mentioned [above](https://github.com/nodejs/undici/pull/4148#discussion_r2032092672)) 👍

dario-piotrowicz:
ok docs added 🙂 https://github.com/nodejs/undici/pull/4148/commits/c8abf9e99e200ba704c19cabed6affac837ff6f1

It was much quicker than I thought 😄, I was initially thinking of adding an example for this but it feels like a bit of an overkill to me right now 🤔, please let me know what you think 🙂

mcollina (approved):
lgtm

PR #4158, opened 2025-04-13
feat: use provided maxConcurrentStreams as multiplexing factor

### LINKED ISSUE #4143: Default pipelining to maxConcurrentStreams with allowH2
## Bug Description
The recent H2CClient work (#4118) does it well. If `pipelining` is not set, it defaults to `maxConcurrentStreams` which makes sense. This however, is not a case when a regular Client (or its descendent Agent) is in use.

Being on a HTTP/2 connection is still limited to either the default (1) or requires manual setup.
Manual setup is inconvenient and confusing. Pipelining is a HTTP/1.1 concept.

Docs say nothing about adjust pipelining when using HTTP/2 **_multiplexing_** and actually link to the [HTTP/1.1 RFC](https://datatracker.ietf.org/doc/html/rfc7230#section-6.3.2) under pipelining.

## Reproducible By
<details>
  <summary>sever.js</summary>
  
  ```javascript
import { fetch, Agent, setGlobalDispatcher, interceptors } from "undici";

setGlobalDispatcher(
  new Agent({
    allowH2: true,
    connections: 10,
    //pipelining: 100,
  }).compose([interceptors.dns(), interceptors.retry()]),
);

  fetch("https://my-https2-server.com", {
    signal: AbortSignal.timeout(1_000),
  })
    .then((result) => {
      return Array.fromAsync(result.body);
    })
    .then(() => {
      res.end("");
    })
    .catch((err) => {
      res.writeHead(500);
      res.end("internal server error");
    });
}).listen("3000");

console.log("server started on port 3000");
  ```
  
</details>

Then run the server:
```
node index.js
```

And load test it. Run the same with and without `pipelining` set.
```sh
hey -n 15000 -c 100 -q 300 -t "1" http://localhost:3000
```

## Expected Behavior
Same performance with or without pipelining set.

## Environment
Node 23, macOS 15.3, but it's not platform related, it seems.

### Additional context

Should #2750 make HTTP/2 on by default, users will not benefit from HTTP/2 at all - it might behave pretty much the same as HTTP/1.1 connections with keep-alive + some TLS overhead...

slagiewka:
## This relates to...

Closes #4143.

## Rationale
For H2 sessions pipelining is misleading and is a concept of HTTP/1.1.

This makes use of `maxConcurrentStreams` parameter to be equivalent of what H2 calls multiplexing. It makes sure that clients do not have to change `pipelining` in order to achieve multiplexing over a H2 stream.

### Features

- Allow controlling multiplexing over a stream without increasing `pipelining`.

- None, unless users defined `maxConcurrentStreams` themselves.

## Status

- [ ] Tested
- [x] Benchmarked (**optional**) [^1]
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

[^1]:This makes undici (non-fetch) results way closer to "native - http2" under a single connection. It is worth noting that any benchmarks running on 2+ connections are unfair to "native" as it has a single connection open.


benchmarks/benchmark-http2.js
```diff
@@ -15,7 +15,7 @@ const servername = 'agent1'
 const iterations = (parseInt(process.env.SAMPLES, 10) || 10) + 1
 const errorThreshold = parseInt(process.env.ERROR_THRESHOLD, 10) || 3
 const connections = parseInt(process.env.CONNECTIONS, 10) || 50
-const pipelining = parseInt(process.env.PIPELINING, 10) || 10
+const maxConcurrentStreams = parseInt(process.env.MULTIPLEXING, 10) || 10
```
slagiewka, line 18: Either this or:
```suggestion
const maxConcurrentStreams = parseInt(process.env.MAX_CONCURRENT_STREAMS, 10) || 10
```

Unless you would prefer re-using the same semantic across benchmarks and keep "pipelining" here.

lib/dispatcher/client.js
```diff
@@ -68,6 +68,10 @@ const getDefaultNodeMaxHeaderSize = http &&
 const noop = () => {}
 
 function getPipelining (client) {
+  if (client[kHTTPContext]?.version === 'h2') {
+    return client[kMaxConcurrentStreams] ?? client[kHTTPContext]?.defaultPipelining ?? 100
+  }
```
slagiewka, line 73: What remains unclear to me is under which circumstances `getPipelining` can be called with a client that has no `kHTTPContext` present.

In any case, this tries to figure out whether we're using H2 and short circuit on that decision. 

The statement below acts as a fallback but I'm almost 100% sure that `client[kPipelining]` always wins (is never nullish) since a client is always created with on in its constructor. However, the setter for pipelining COULD make it nullish (for some reason) and only then it would fall to the context's setting.

slagiewka (commented):
I didn't touch the `H2CClient` yet, but it might be addressed as well. Granted, this has lesser priority as it already defaults `pipelining` to be equal `maxConcurrentStreams`.

metcoder95 (changes requested):
We are not yet preparing a new major version, so we will need to make it backwards compatible with a warning that states the change;

We can follow the path of putting `maxCurrentStreams` in higher order of priority compared to `pipelining` for `h2`, but still honor `pipelining` if `maxConcurrentStreams` is not passed so we keep do not introduce a breaking change (just yet)

mcollina (changes requested):
Good work! Can you fix the tests?

lib/dispatcher/client.js
```diff
@@ -68,6 +69,10 @@ const getDefaultNodeMaxHeaderSize = http &&
 const noop = () => {}
 
 function getPipelining (client) {
+  if (client[kHTTPContext]?.version === 'h2' && client[kMaxConcurrentStreamsManual]) {
```
metcoder95, line 72: Let's just add a deprecation warning if `maxConcurrentStreams` is not set, and we should be good to go

metcoder95 (commented):
lgtm, just minor comment

lib/dispatcher/client.js
slagiewka: When would that happen? The symbol (which name I don't like yet :see_no_evil: ) that I'm using is only set if `maxConcurrentStreams` is set (line 240).

Or your suggestion is to notify everyone using `allowH2` but not setting it - therefore missing the benefit of multiplexing?

I'm also not sure WHEN that deprecation warning would happen. I assume that somewhere in Client constructor. Otherwise the noise would be significant.

slagiewka:
One comment from me: with the [original change](https://github.com/nodejs/undici/pull/4158/commits/810d7c6d841f8afd1745d07b067b2b511c9b30ca) enabling 100 concurrent streams for the entire codebase might have uncovered some issues when multiplexing is in use. Tests might be indication of that.
I'm not yet versed well enough with the internals, but that's a yellow flag to me.

lib/dispatcher/client.js
metcoder95: I was thinking to add it only when the `kMaxConcurrentStreamsManual` is set to `false` but `pipeline` was altered instead; so the warning states that in further major `pipeline` will not take the effect of `maxConcurrentStreams`

metcoder95:
> One comment from me: with the [original change](https://github.com/nodejs/undici/pull/4158/commits/810d7c6d841f8afd1745d07b067b2b511c9b30ca) enabling 100 concurrent streams for the entire codebase might have uncovered some issues when multiplexing is in use. Tests might be indication of that. I'm not yet versed well enough with the internals, but that's a yellow flag to me.

Haven't run the benchmarks yet (might do it Today); but what's have you see?

slagiewka:
> Haven't run the benchmarks yet (might do it Today); but what's have you see

It's actually not in benchmarks but in tests:
```
test at test/http2.js:1247:1
✖ #2364 - Concurrent aborts (111.166097ms)
  Error [ERR_HTTP2_INVALID_STREAM]: The stream has been destroyed
      at ServerHttp2Stream.respond (node:internal/http2/core:2786:13)
      at Timeout._onTimeout (/home/runner/work/undici/undici/test/http2.js:1252:14)
      at listOnTimeout (node:internal/timers:581:17)
      at process.processTimers (node:internal/timers:519:7) {
    code: 'ERR_HTTP2_INVALID_STREAM'
  }

test at test/http2.js:1:1
```

Which is 100% reproducible for me when `pipelining == default maxConcurrentStreams` (original commit) and does not show up otherwise (current state and on main).

At first I thought that maybe there's something wrong with how "destroy" happens on a stream once request gets aborted and instead the session gets hit. But the error happens on the mock server in this test (trying to respond) - haven't looked yet into why that may happen with many streams vs a single stream.

metcoder95:
I've been running benchmarks with different combinations and I'm finding a strong correlation between the `pipelining` (parallel requests send through the same `session` within `undici`) and the number of max concurrent streams available; having a negative impact on performance if both increased in the same direction (both increased and close to each other).

See following table
- `maxConcurrentStreams` = 100
- `pipelining` = 50

image

in contrast with
- `maxConcurrentStreams` = 100
- `pipelining` = 100

image

Given the results and findings, I'll suggest to re-evaluate the change and, instead of keeping them in sync, maintain `pipelining` as a way to handle concurrent requests within `undici` while `maxConcurrentStreams` maintains keeping the stream management independent.

If we want to be explicit, we can add a further option property (e.g. `multiplexing`) that states the max number of concurrent requests while `maxConcurentStreams` remains applying to the same concept (although I believe `pipelining` can keep it with just extra documentation added).

mcollina (changes requested):
tests are needed

slagiewka:
> The results are consistent between executions, feel free to play with the combination, but the closer the pipelining is to maxConcurrentStreams, the worst the performance becomes.

Hey @metcoder95

I've found multiple problems with these benchmarks in general:
* Pipelining is never used with this change. I've manually hooked into all the places that would set or get the value and it's never used. Therefore I'm baffled why it would impact benchmark results in any way. It doesn't for me.
* `maxConcurrentStreams` is effectively no-op:
  * test server does not use any explicit settings -> its `maxConcurrentStreams` defaults to 2^32  -1
  * undici is instructed to [always overwrite](https://github.com/nodejs/undici/blob/a4eaed1124ec084665d215af6395c70439a7dd5a/lib/dispatcher/client-h2.js#L244) it with `remoteSettings` (granted, this seems to have been [added in the meantime](https://github.com/nodejs/undici/commit/92715645444a41d3d6b3be0a458630cf32150adf)).
* when I patch the server to limit itself the the same value (from benchmark env) we're actually using limited streams
* however, I could find no reasonable setting that would tell me what is the optimal connections + streams config that gives the best performance
* in no configuration I could say that HTTP/2 performance is on par with HTTP/1.1. Maybe some profiling is in order to figure it out.

As a bonus, the "native" pass will be in a disadvantage since it's always 1 connection (session). Maybe sticking to 1 connection/session for all of them and controlling number of streams would give every pass a fair playing ground.

> If we want to be explicit, we can add a further option property (e.g. multiplexing) that states the max number of concurrent requests while maxConcurentStreams remains applying to the same concept (although I believe pipelining can keep it with just extra documentation added).

I'm having doubts about such an explicit approach.
1. The config is already quite loaded
2. As a user I would be having hard time trying to figure out the difference between `multiplexing`, `maxConcurrentStreams` and `pipelining`.
3. `pipelining` controlling multiplexing is what I wanted to avoid in the first place. My goal here was to allow HTTP/2 use concurrency allowed by the design while not allowing HTTP/1.1 to "queue" more than 1 request per connection. I might be mistaken here, but my understanding is that users need to be careful with `pipelining` > 1 over HTTP/1.1.

metcoder95:
Thanks a lot for the feeadback, it is really valuable; I cannot remember the setup I've had for the server back then, but I'm pretty sure i just used the defaults and didn't alter, on purpose, the server settings, causing `undici` to just use the remote settings, overriding the local ones as you stated.

Nonetheless, the `pipelining` is mostly applied to the internal queueing system of `undici` that hopes into `llhttp`. `h2` is a different story, as it adapts to the queuing system of `undici` to provide similar experience as per `h1`.

I will definitely need to revisit this to ensure testing the right things, will take your feedback in consideration.

In regards of the suggestion; I see we agree on the following things:

1. `pipelining` is not a concept of `h2`
2. Too many options can be cumbersome and complicated to users to understand

The last being said, I believe we can suggest the following:

1. `h2` should be, somehow (albeit complex), decoupled from the `pipelining` concept of `undici`. This is complicated as this is manages the queuing system within `undici` and decoupling it will require find another way to hope into that system (more research might need to be done, although even an explicit document change can be enough).
	- Also needs to be clarified than, in comparison with undici's `h1`, the number of concurrent request per stream is managed by the underlying `node:http2` module, so we are limited by that constant.
2. `maxConcurrentStreams` needs to stay as this manage the number of concurrent streams per session (although can be altered by the remote settings frame)
3. `multiplexing` can be considered unnecessary and we can disregard it

slagiewka:
Looks like #5362 did solve what I was trying to achieve here - especially in my first iteration. Super happy to see this, closing this then :wink:

PR #4175, opened 2025-04-23
Add "clientLifetime" option to close and remove connections from the pool after a specified time.

dhalbrook:
## This relates to...

This feature allows clients to be closed and removed from the pool after a specified amount of time.  

## Rationale

This feature can be helpful to avoid GOAWAY connection errors when dealing with HTTP2 connections where the amount of time they are valid for is known beforehand.

## Changes

This adds a "clientLifetime" option to the Pool options allowing specification of a fixed lifetime for clients in mils

### Bug Fixes

N/A 

## Status

- [ x] I have read and agreed to the [Developer's Certificate of Origin][cert]
- [x ] Tested
- [ x] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready


lib/dispatcher/pool.js
```diff
@@ -71,6 +72,21 @@ class Pool extends PoolBase {
       : undefined
     this[kFactory] = factory
 
+    this.on('connect', (origin, targets) => {
+      if (clientLifetime != null && clientLifetime > 0) {
+        setTimeout(() => {
```
dhalbrook, line 77: I am a little concerned that creating N timers isn't the most efficient way of handling this.  Another option would be to mark the client with a create time and then validate it before operations to make sure it hasn't expired.  If it has, we perform the logic here.

metcoder95: I'd prefer to explore that option; under high level of clients, this can create unnecessary overhead

metcoder95:
I'm under the impression that this can be already achieved through [`PoolOptions#factory`](https://undici.nodejs.org/#/docs/api/Pool?id=parameter-pooloptions); I'm happy either way, but it we want to land it in core I'd suggest to assign a more explicit name like `ttl` or something.

On top, consider the possibility of verify/preempt the client upon a request, so we reduce the surface for possible race conditions between removal and new upcoming request.

dhalbrook:
> I'm under the impression that this can be already achieved through [`PoolOptions#factory`](https://undici.nodejs.org/#/docs/api/Pool?id=parameter-pooloptions);
> 

I would have gone down that route but i'm not sure how to remove a client from the pool.  That doesn't seem to be exposed anywhere in `Pool`.

dhalbrook:
I updated the logic to remove clients in the `getDispatcher` call when the TTL has passed.  This should not add any significant overhead.

metcoder95 (approved):

dhalbrook:
Ugh, I was trying to rebase my fork to fix the test issues and screwed something up.

dhalbrook:
@metcoder95 Any idea why those tests might be failing?  They don't seem related to may changes as far as I can tell.

metcoder95:
Yeah, these are flaky tests that needs to be fixed; but are not related to your changes.

dhalbrook:
Is there a way to re-run them until they pass?

metcoder95:
They will fail constantly. I hope I can have a fix soon

dhalbrook:
Hi @metcoder95 , I'm a bit bummed this didn't make it into the 7.9.0 release.  What are the steps forward?

mcollina (approved):
lgtm

PR #4177, opened 2025-04-23
cache-control: no-cache: use quoted-string form

alxndrsn:
Update comment example of qualified no-cache directives (`no-cache=...`) to use the quoted-string form instead of the token form.

Per https://datatracker.ietf.org/doc/html/rfc9111#name-no-cache-2:

> This directive uses the quoted-string form of the argument syntax. A sender SHOULD NOT generate the token form (even if quoting appears not to be needed for single-entry lists).

## This relates to...

`cache-control` response header's qualified `no-cache` directive.

## Rationale

Seems more correct to follow spec recommendations in documentation.

Support for token form seems to be deliberate:

https://github.com/nodejs/undici/blob/0daba937990d92f670b858216e4715ec9f92da44/test/cache-interceptor/utils.js#L68-L96

## Changes

Update comment and test examples of qualified no-cache directives (`no-cache=...`) to use the quoted-string form instead of the token form.

### Features

N/A

### Bug Fixes

N/A

Should not break anything, as no code changed.

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/util/cache.js
+++ b/lib/util/cache.js
@@ -180,8 +180,8 @@ function parseCacheControlHeader (header) {
         if (value) {
           // The private and no-cache directives can be unqualified (aka just
           //  `private` or `no-cache`) or qualified (w/ a value). When they're
-          //  qualified, it's a list of headers like `no-cache="header1"`,
-          //  `no-cache="header1, header2"`
+          //  qualified, it's a list of headers like `no-cache=header1`,
+          //  `no-cache="header1"`, or `no-cache="header1, header2"`
           // If we're given multiple headers, the comma messes us up since
           //  we split the full header by commas. So, let's loop through the
           //  remaining parts in front of us until we find one that ends in a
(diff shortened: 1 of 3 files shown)


alxndrsn:
The `no-cache=...` parsing code seems to go out of its way to support various different token or mixed token/quoted-string formats:

https://github.com/nodejs/undici/blob/0daba937990d92f670b858216e4715ec9f92da44/test/cache-interceptor/utils.js#L68-L96

Would it be better to only allow quote-string format?

mcollina:
> Would it be better to only allow quote-string format?

@flakey5 was this due to the compatibility tests?

flakey5:
> @flakey5 was this due to the compatibility tests?

Yes. If this isn't wanted behavior, the test cases for it in the caching suite will need to be added to the `ignoredTests` list in `test/cache-interceptor/cache-tests.mjs`

Uzlopak (approved):
Well, I guess it isnt harming

PR #4180, opened 2025-04-24
feat(ProxyAgent): match Curl behavior in HTTP->HTTP Proxy connections

### LINKED ISSUE #4083: ProxyAgent has different behavior from curl -x
## Bug Description

When testing https://github.com/nodejs/node/pull/57165 I noticed that there are some behavior differences between ProxyAgent and curl which may be interesting. Most notably, ProxyAgent always tunnels with CONNECT even for pure HTTP traffic.

## Reproducible By

Create a proxy server, like this:

```js
// server.js
'use strict';

const http = require('http');
const net = require('net');

function log(req) {
  console.log('----- Received Request -----');
  console.log(`${req.method} ${req.url} HTTP/${req.httpVersion}`);
  for (const header in req.headers) {
    console.log(`${header}: ${req.headers[header]}`);
  }
}

  log(req);
  const [hostname, port] = req.headers.host.split(':');
  const targetPort = port || 80;

  const options = {
    hostname: hostname,
    port: targetPort,
    path: req.url,
    method: req.method,
    headers: req.headers
  };

  const proxyReq = http.request(options, proxyRes => {
    res.writeHead(proxyRes.statusCode, proxyRes.headers);
    proxyRes.pipe(res, { end: true });
  });

  proxyReq.on('error', err => {
    console.error('Proxy request error: ' + err.message);
    res.writeHead(500);
    res.end('Proxy error: ' + err.message);
  });

  req.pipe(proxyReq, { end: true });
});

server.on('connect', function (req, clientSocket, head) {
  log(req);
  const [hostname, port] = req.url.split(':');

  const serverSocket = net.connect(port, hostname, () => {
    clientSocket.write(
      'HTTP/1.1 200 Connection Established\r\n' +
      'Proxy-agent: Node.js-Proxy\r\n' +
      '\r\n'
    );
    serverSocket.write(head);
    clientSocket.pipe(serverSocket);
    serverSocket.pipe(clientSocket);
  });

  serverSocket.on('error', (err) => {
    console.error('Error on CONNECT tunnel:', err.message);
    clientSocket.write('HTTP/1.1 500 Connection Error\r\n\r\n');
    clientSocket.end();
  });
});

server.listen(8000, '127.0.0.1', () => {
  const { address, port } = server.address();
  console.log(`Proxy server listening on http://${address}:${port}`);
});
```

And send a request to `http://example.com` like this:

```js
// proxy
const { fetch, ProxyAgent } = require('undici');

(async () => {
  const res = await fetch('http://example.com', {
    dispatcher: new ProxyAgent('http://127.0.0.1:8000')
  });
  console.log(await res.text());
})();
```

## Expected Behavior

If I run `curl -x http://127.0.0.1:8000 http://example.com`, this is the output from the proxy server:

```
GET http://example.com/ HTTP/1.1
host: example.com
user-agent: curl/8.7.1
accept: */*
proxy-connection: Keep-Alive
```

## Logs & Screenshots

Using the client with undici, the proxy server always get a CONNECT:

```
CONNECT example.com:80 HTTP/1.1
host: example.com
connection: close
```

This is somewhat similar to the one I get for sending requests to `https://example.com` using curl - which does use CONNECT.

Output from `curl -x http://127.0.0.1:8000 https://example.com`

```
host: example.com:443
user-agent: curl/8.7.1
proxy-connection: Keep-Alive
```

Output from client with undici ProxyAgent (change the `http://example.com` to `https://example.com` in the snippet above)

```
CONNECT example.com:80 HTTP/1.1
host: example.com
connection: close
```

## Environment

macOS though I am sure this is irrelevant.

### Additional context

I am not sure this really counts as a bug or not. But from my impression, not all proxy servers in the wild supports tunneling on non-443 ports. It may be safer to follow what curl does, because the setup of proxy server and target servers are not always in the control of users. If they end up having to work with a proxy server that does not support tunnelling on non-443 ports, and also having to connect to a server on 80 port or others with the proxy, there would be no way for them to work around it.

caitp:
By default, Curl does not send a CONNECT request to a non-secure Proxy with a non-secure endpoint. This can be overridden using the --proxytunnel or -p parameter.

This change modifies ProxyAgent's constructor to accept a `tunnelProxy` option, sends a CONNECT if either `tunnelProxy` is true, or either the Proxy or endpoint use a non-'http:' protocol.

Disabling tunneling for HTTP->HTTP by default would be a breaking change, so currently, the tunneling behavior requires an opt-out. This may change depending on feedback during code review.

This adds a new test case which explicitly disables tunneling for an HTTP->HTTP connection, and asserts that no CONNECT message is sent to the server or proxy, and that the expected HTTP request is sent to the proxy.

Closes #4083

## This relates to...

#4083

## Rationale

Closer alignment with widely used industry standard libraries (specifically Curl)

## Changes

- A new `tunnelProxy` option is added to the ProxyAgent dispatcher which requests whether or not CONNECT requests be sent to an HTTP->HTTP Proxy
- A hackish `rawSocket` option is added to Request. This is awkward, advice on a cleaner/better way to do this would be appreciated. This conveys to a dispatcher several layers down that it should avoid proceeding with a CONNECT request and instead fulfill the connect() call with just the socket.
- A new test case showcasing the altered behavior when `opts.tunnelProxy===false`

### Features

The new `tunnelProxy` option, which permits matching Curl (however, currently this sort of works in the opposite way to Curl, as tunneling occurs by default and requires opt-out rather than opt-in)

### Bug Fixes

#4083

It would technically be a breaking change to fully match Curl by opting out of tunneling by default. That is (in the first iteration of this patch) not done.

## Status

- [x] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/dispatcher/proxy-agent.js
+++ b/lib/dispatcher/proxy-agent.js
@@ -14,6 +14,7 @@ const kProxyHeaders = Symbol('proxy headers')
 const kRequestTls = Symbol('request tls settings')
 const kProxyTls = Symbol('proxy tls settings')
 const kConnectEndpoint = Symbol('connect endpoint function')
+const kTunnelProxy = Symbol('tunnel proxy')
 
 function defaultProtocolPort (protocol) {
   return protocol === 'https:' ? 443 : 80
@@ -35,6 +36,8 @@ class ProxyAgent extends DispatcherBase {
     if (typeof clientFactory !== 'function') {
       throw new InvalidArgumentError('Proxy opts.clientFactory must be a function.')
     }
+
+    const { tunnelProxy = true } = opts
 
     super()
 
@@ -60,6 +63,7 @@ class ProxyAgent extends DispatcherBase {
     const connect = buildConnector({ ...opts.proxyTls })
     this[kConnectEndpoint] = buildConnector({ ...opts.requestTls })
     this[kClient] = clientFactory(url, { connect })
+    this[kTunnelProxy] = tunnelProxy
     this[kAgent] = new Agent({
       ...opts,
       connect: async (opts, callback) => {
@@ -67,6 +71,7 @@ class ProxyAgent extends DispatcherBase {
         if (!opts.port) {
           requestedPath += `:${defaultProtocolPort(opts.protocol)}`
         }
+        const shouldConnect = this[kTunnelProxy] || protocol !== 'http:' || opts.protocol !== 'http:'
         try {
           const { socket, statusCode } = await this[kClient].connect({
             origin,
@@ -77,6 +82,7 @@ class ProxyAgent extends DispatcherBase {
               ...this[kProxyHeaders],
               host: opts.host
             },
+            rawSocket: !shouldConnect,
             servername: this[kProxyTls]?.servername || proxyHostname
           })
           if (statusCode !== 200) {
@@ -113,6 +119,12 @@ class ProxyAgent extends DispatcherBase {
     if (headers && !('host' in headers) && !('Host' in headers)) {
       const { host } = new URL(opts.origin)
       headers.host = host
+    }
+
+    // If we have a non-secure, non-tunneling proxy, the server URL is prepended to the
+    // path to ensure that the Proxy can handle the request appropriately.
+    if (!this[kTunnelProxy] && opts.path && opts.origin) {
+      opts.path = `${opts.origin}${opts.path}`
     }
 
     return this[kAgent].dispatch(
--- a/lib/dispatcher/proxy-agent.js
+++ b/lib/dispatcher/proxy-agent.js
@@ -1,12 +1,13 @@
 'use strict'
 
-const { kProxy, kClose, kDestroy } = require('../core/symbols')
+const { kProxy, kClose, kDestroy, kDispatch, kConnector } = require('../core/symbols')
 const { URL } = require('node:url')
 const Agent = require('./agent')
 const Pool = require('./pool')
 const DispatcherBase = require('./dispatcher-base')
 const { InvalidArgumentError, RequestAbortedError, SecureProxyConnectionError } = require('../core/errors')
 const buildConnector = require('../core/connect')
+const Client = require('./client')
 
 const kAgent = Symbol('proxy agent')
 const kClient = Symbol('proxy client')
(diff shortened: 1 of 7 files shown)


lib/dispatcher/proxy-agent.js
metcoder95, line 1: I'm also not in favor of the small hack around the `rawSocket` concept. I believe you can achieve the same by creating the connection and returning the resulted `socket` (similar to what's done for the [`h2c-client`](https://github.com/nodejs/undici/blob/main/lib/dispatcher/h2c-client.js)

mcollina:
cc @joyeecheung

lib/dispatcher/proxy-agent.js
caitp: I've had a go at doing something like this, not sure if this is quite what you had in mind, but it does seem to work

types/dispatcher.d.ts
```diff
@@ -140,6 +140,8 @@ declare namespace Dispatcher {
     redirectionLimitReached?: boolean;
     /** Default: `null` */
     responseHeaders?: 'raw' | null;
+    /** Default: false */
+    rawSocket?: boolean;
```
metcoder95, line 144: Shall we remove it?

types/proxy-agent.d.ts
```diff
@@ -24,5 +24,6 @@ declare namespace ProxyAgent {
     requestTls?: buildConnector.BuildOptions;
     proxyTls?: buildConnector.BuildOptions;
     clientFactory?(origin: URL, opts: object): Dispatcher;
+    tunnelProxy?: boolean;
```
metcoder95, line 27: Ditto

metcoder95 (commented):
lgtm, just documentation left

types/proxy-agent.d.ts
caitp: this is technically still used

metcoder95: Oh sorry, I misread it, disregard my comment

caitp:
I may have accidentaly messed up the merge, so replacing that with a force push

metcoder95:
tests does not seems happy

caitp:
> tests does not seems happy

seems to be from new stuff imported by dcf82a729c34430188b9dec3ca3add7c0186ec24, the same test fails with this whole PR reverted

mcollina (approved):
lgtm

joyeecheung:
Was trying to update the test expectations in Node.js now that this gets fixed in https://github.com/nodejs/node/pull/58859, but I see that it still gets a CONNECT for the HTTP test. I think the problem is that `tunnelProxy: false` is an agent-wide setting, but what we want is for it to be automatic - false for HTTP traffic, true for HTTPS, I think that's the default behavior of curl as well.

metcoder95:
cc: @caitp

caitp:
I think this is right there in the PR description :P "Disabling tunneling for HTTP->HTTP by default would be a breaking change, so currently, the tunneling behavior requires an opt-out. This may change depending on feedback during code review."

There is already logic to tunnel if an endpoint or the proxy itself uses a secure protocol -- it should be the same logic as in Curl --- the only thing needed to change this to be the default is just flipping the default flag value to `false`, and then tunneling for unsecure protocols would become opt-in.

metcoder95:
If that's the case, we can keep this until further major where the flag is turned `true` by default

caitp:
I should add, there do seem to be some issues with it, some of which have been raised by Joyee in the past, so I am working on an update to fix it up a bit -- maybe flipping the flag by default makes sense after that

Issue #3268, opened 2024-05-16
Limit the maximum numbers of open sockets


mcollina:
One of the first issues with "fetch failed" is that we don't limit the number of outgoing sockets, and the process runs out of them, triggering a timeout error.

I think we should add a limit of 127 per pool by default (or any other magic number)

metcoder95:
The global dispatcher or the overall default Client/Agent settings?

mcollina:
Both are good ideas.

JoshMock:
The Elasticsearch client library (of which I am the current maintainer) [currently has an open request](https://github.com/elastic/elasticsearch-js/issues/1740) from the Kibana team to support limiting the maximum open sockets and max idle sockets when keepalive is enabled. If this functionality is added to Undici, it would be helpful if it exposed options to limit the max open and max total sockets, both globally and per upstream target, as well as max free sockets for keepalive use cases, to provide parity with `http.Agent`'s options:

- `maxFreeSockets`: max free sockets when keepalive is enabled
- `maxSockets`: max concurrent sockets per origin
- `maxTotalSockets`: max concurrent sockets globally

I've not contributed to this library before, but would love to help add this functionality.

metcoder95:
Overall, I think this should be already doable (unless I'm overlooking at something).

By using `Pool` you can limit the number of connections (`Client` instance) that the pool can have. As a `Client` is mapped to a single Socket.

>Ref: https://undici.nodejs.org/#/docs/api/Pool?id=parameter-pooloptions

The options listed should be ported to the `Agent`, as it can manage these options at a higher level, at the end, depending on the number of connections passed, can be a Pool of connections or a single connection per origin.

`maxFreeSockets` -  maybe `maxIdleConnections`? Here the `Pool` and/or the `Agent` can be preemptive and destroy idle connections before waiting for more request to come.

`maxTotalSockets` - these should be ported to `Agent` I believe, as the `Pool` already has an option for that.

Would you like to send a PR for it?

mcollina:
@JoshMock this is already possible at Pool level.

tdeekens:
I struggle to read from the code but what happens if a pool reaches its connection limit. I can see that [here](https://github.com/nodejs/undici/blob/main/lib/dispatcher/pool.js#L96) we won't create a dispatcher. What are the implications of this. Will the request be dropped or queued?

gurgunday:
The request might fail with if it runs of file descriptors AFAIK

[...]

PR #4192, opened 2025-05-02
WIP: Max total connections support for BalancedPool

JoshMock:
## This relates to...

https://github.com/nodejs/undici/issues/3268#issuecomment-2833338941

## Rationale

Work in progress on being able to enforce maximum total connections across all origins in a `BalancedPool` and `Agent`.

### Features

Exposes `maxTotalConnections` option on `BalancedPool` and `Agent` that ensures the total number of open connections across all origins does not exceed the configured maximum.

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/dispatcher/balanced-pool.js
+++ b/lib/dispatcher/balanced-pool.js
@@ -10,10 +10,11 @@ const {
   kNeedDrain,
   kAddClient,
   kRemoveClient,
-  kGetDispatcher
+  kGetDispatcher,
+  kQueue
 } = require('./pool-base')
 const Pool = require('./pool')
-const { kUrl } = require('../core/symbols')
+const { kUrl, kConnected, kDispatch, kQueued, kFree } = require('../core/symbols')
 const { parseOrigin } = require('../core/util')
 const kFactory = Symbol('factory')
 
@@ -24,6 +25,8 @@ const kIndex = Symbol('kIndex')
 const kWeight = Symbol('kWeight')
 const kMaxWeightPerServer = Symbol('kMaxWeightPerServer')
 const kErrorPenalty = Symbol('kErrorPenalty')
+const kMaxTotalConnections = Symbol('kMaxTotalConnections')
+const kQueueResume = Symbol('kQueueResume')
 
 /**
  * Calculate the greatest common divisor of two numbers by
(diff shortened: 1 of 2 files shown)


lib/dispatcher/balanced-pool.js
```diff
@@ -146,6 +154,50 @@ class BalancedPool extends PoolBase {
...
 
+  get [kFree] () {
+    return this[kClients].map(pool => pool[kFree]).reduce((a, b) => a + b, 0)
+  }
+
+  get [kConnected] () {
+    return this[kClients].map(pool => pool[kConnected]).reduce((a, b) => a + b, 0)
+  }
+
+  [kQueueResume] () {
```
JoshMock, line 165: @metcoder95 You asked elsewhere why I overrode other queue resume code. For one, I couldn't see how other extensions of `PoolBase` or `DispatcherBase` were triggering a queue resume event; if you can point me to where/how that happens that would be fantastic.

But also, this implementation specifically optimizes for finding an available `Pool` with a matching origin.

Thanks for your patience, as this is my first time digging through the Undici codebase, so I'm learning the internals as I go.

lib/dispatcher/balanced-pool.js
```diff
@@ -146,6 +154,50 @@ class BalancedPool extends PoolBase {
...
+      }
+    }
+  }
+
+  [kDispatch] (opts, handler) {
+    // wrap handler.onComplete so it triggers kQueueResume after each request completes
+    const { onComplete } = handler
+    handler.onComplete = (...args) => {
+      setTimeout(this[kQueueResume](), 0)
+      if (onComplete) onComplete(...args)
```
JoshMock, line 195: I mentioned this in https://github.com/nodejs/undici/issues/3268, but to put it in context: wrapping a handler's `onComplete` function appears to happen every time a request is attempted, so if a request is taken from the queue `x` times before finally being sent, this call to `kQueueResume` happens `x` times as well. Testing this locally I noted that it caused `kQueueResume` calls to increase exponentially over time, which is not at all ideal.

I tried adding a private symbol `handler[kResumable] = true` when it had been wrapped, to avoid re-wrapping, but it seems that `handler` is a brand new object every time `kDispatch` is called.

lib/dispatcher/balanced-pool.js
```diff
@@ -49,16 +52,21 @@ function defaultFactory (origin, opts) {
 }
 
 class BalancedPool extends PoolBase {
-  constructor (upstreams = [], { factory = defaultFactory, ...opts } = {}) {
+  constructor (upstreams = [], { factory = defaultFactory, maxTotalConnections = 0, ...opts } = {}) {
     if (typeof factory !== 'function') {
       throw new InvalidArgumentError('factory must be a function.')
     }
 
+    if (maxTotalConnections != null && (!Number.isFinite(maxTotalConnections) || maxTotalConnections < 0)) {
```
metcoder95, line 60: For backwards compatibility, shouldn't be better to default it to `Infinity`?

lib/dispatcher/balanced-pool.js
metcoder95: But is there a specific reason to override it? Or why was required? (maybe I missed something before)

PR #4198, opened 2025-05-08
Restore cache tests & fix max-age behavior

mcollina:
The cache tests were not being run. Restore them and add a check to prevent this situation to happen again.

Moreover, it fixes a bug in the memory storage that prevented the max-age and expiry directives to work properly.

fixes https://github.com/nodejs/undici/issues/4203


flakey5 (approved):

ShogunPanda (approved):
LGTM!

metcoder95 (approved):

Issue #4185, opened 2025-04-24
Add ability to detect when `MemoryCacheStore` reaches max size


OliverJAsh:
### What is the problem this feature will solve?

Currently it's possible to configure a `maxSize` with `MemoryCacheStore`, but there doesn't seem to be any way to detect when this max size has been exceeded.

```ts
new Undici.cacheStores.MemoryCacheStore({
  maxSize: 100_000_000,
})
```

### What is the feature you are proposing to solve the problem?

Add a mechanism to detect when the max size has been exceeded.

metcoder95:
An event (e.g. `full`) or a property (e.g. `size` getter) can be helpful here; but of curiosity, what use case are you looking exactly?

metcoder95:
Got it, would you like to send a PR for that?

FelixVaughan:
I'd like to work on this. I'm thinking of adding a `size` getter, and an `isFull` method. Thoughts?

OliverJAsh:
Would it make sense to have an event for when the max size is reached?

FelixVaughan:
Yes, I believe this makes sense. Perhaps all three would be useful.

PR #4224, opened 2025-05-15
Add ability to detect when MemoryCacheStore reaches max size

FelixVaughan:
## This relates to...

This is addressing #4185

## Rationale

Currently it's possible to configure a maxSize with MemoryCacheStore, but there isn't a way to detect when this max size has been exceeded.

## Changes

Added methods to get size of a memory cache, and check whether it is overflowing as well as an event which is emitted when the cache is full. Also added tests and updated docs.

### Features

1. Added `size` getter on MemoryCacheStore
2. Added `isFull` public method on MemoryCacheStore
3. Added `maxSizeExceeded` event on MemoryCacheStore

### Bug Fixes

N/A

N/A

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready


test/memory-cache-store-size.js
```diff
@@ -0,0 +1,123 @@
+'use strict'
```
mcollina, line 1: can you please move this test together to the others of the cache?

FelixVaughan: Yes, I moved them to `memory-cache-store-tests.js`. Let me know if you had somewhere else in mind.

mcollina (approved):
lgtm

PR #4236, opened 2025-05-23
fix(debug): remove extra forward slash in logs

aidant:
## Rationale

When fixing an issue in a unrelated third-party package that was incorrectly adding an extra forward slash. The node debug logs still contained an extra forward slash which lead me to believe the issue had not been patched in the unrelated third-party package. 

### Bug Fixes

When enabling the fetch debug logs, the request contains two forward slashes.

## Status

- [ ] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready


Uzlopak (approved):
LGTM

but why are the tests failing?

PR #2665, opened 2024-01-29
remove timers/promises import

KhafraDev:
Fixes https://github.com/nodejs/undici/issues/2664

## Status

- [ ] I have read and agreed to the [Developer's Certificate of Origin][cert]
- [ ] Tested
- [ ] Benchmarked (**optional**)
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

[cert]: https://github.com/nodejs/undici/blob/main/CONTRIBUTING.md

PR #4247, opened 2025-05-30
fix: eventsource does not reconnect on network error

Uzlopak:
To be honest: I am pretty much confused, why we had this regression. In #2665 the setTimeout got replaced with the non promise one, and thus broke the reconnection logic. It seems like setTimeout from timers/promise without reference, still keeps the process running, but the unref`d normal setTimeout results in letting the process die, even though we are still in reconning state.

Also I dont understand anymore why I wrote the close logic into `processEventSourceEndOfBody`.  

So take this PR with a grain of salt.

### DIFF
--- a/test/eventsource/eventsource-connect.js
+++ b/test/eventsource/eventsource-connect.js
@@ -5,6 +5,7 @@ const events = require('node:events')
 const http = require('node:http')
 const { test, describe } = require('node:test')
 const { EventSource } = require('../../lib/web/eventsource/eventsource')
+const { randomInt } = require('node:crypto')
 
 describe('EventSource - sending correct request headers', () => {
   test('should send request with connection keep-alive', async () => {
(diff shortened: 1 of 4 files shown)


test/eventsource/eventsource-connect.js
```diff
@@ -181,4 +182,21 @@ describe('EventSource - received response must have content-type to be text/even
...
+    const onerrorCalls = []
+    eventSourceInstance.onerror = (error) => {
+      onerrorCalls.push(error)
+    }
+
+    await new Promise(resolve => setTimeout(resolve, 8000))
+
+    eventSourceInstance.close()
+
+    assert.strictEqual(onerrorCalls.length, 3)
```
KhafraDev, line 200: 3 errors?

Uzlopak: Yes, defaultReconnectTime is 3000. So 1. Is about after 50-100 ms, next is 3050-3100ms and the third is at 6050-6100ms. A fourth would be after 9050-9100 ms, but we just do 7000ms (or 8000?) and so we have only 3.

PR #4251, opened 2025-06-01
docs: fix interceptor order description in compose method

### LINKED ISSUE #4241: First interceptor in compose is not the first to be called
## Bug Description

The doc states "The order of the interceptors matters. The first interceptor will be the first to be called."

https://github.com/nodejs/undici/blob/4ab315e2e123f4e6ebcb7d859cb8530e7c76c622/docs/docs/api/Dispatcher.md?plain=1#L844

## Reproducible By

```js
    const a = new Agent().compose([
      (dispatch) => {
        return (opts, handler) => {
          console.log('interceptor 1');
          return dispatch(opts, handler);
        };
      },
      (dispatch) => {
        return (opts, handler) => {
          console.log('interceptor 2');
          return dispatch(opts, handler);
        };
      },
    ]);

    a.request({
      origin: '127.0.0.1',
      path: '/',
      method: 'GET',
    });
```

## Expected Behavior

It should print

```
interceptor 1
interceptor 2
```

but actually it prints

```
interceptor 2
interceptor 1
```

mcollina:
The documentation incorrectly stated that the first interceptor would be called first, but due to function composition, the last interceptor in the array is actually called first.

Fixes #4241

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/docs/docs/api/Dispatcher.md
+++ b/docs/docs/api/Dispatcher.md
@@ -841,9 +841,28 @@ try {
 Compose a new dispatcher from the current dispatcher and the given interceptors.
 
 > _Notes_:
-> - The order of the interceptors matters. The first interceptor will be the first to be called.
+> - The order of the interceptors matters. The last interceptor will be the first to be called.
 > - It is important to note that the `interceptor` function should return a function that follows the `Dispatcher.dispatch` signature.
 > - Any fork of the chain of `interceptors` can lead to unexpected results.
+>
+> **Interceptor Stack Visualization:**
+> ```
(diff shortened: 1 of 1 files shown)


ronag:
The request goes bottom => top and the response goes top => bottom. Would be nice to have a visualization.

ronag (approved):

mcollina:
@ronag I've asked Claude to draw it. I guess it's better than nothing!

PR #4258, opened 2025-06-02
feat: Add Symbol.dispose and Symbol.asyncDispose support to DispatcherBase

PandaWorker:
This PR implements the new ECMAScript Explicit Resource Management proposal by adding disposal functionality to the `DispatcherBase` class. The changes include:

1. Added `[Symbol.asyncDispose]()` method that asynchronously closes the dispatcher
2. Added `[Symbol.dispose]()` method that synchronously closes the dispatcher
3. Both methods reuse the existing `close()` functionality
4. The synchronous version uses the `noop` function as a callback

These changes enable modern resource management patterns:
```js
// Synchronous disposal
using dispatcher = new Agent();

// Async disposal
await using asyncDispatcher = new Client();
```

The implementation follows the TC39 proposal currently in Stage 3 (https://github.com/tc39/proposal-explicit-resource-management)

### DIFF
--- a/lib/dispatcher/dispatcher-base.js
+++ b/lib/dispatcher/dispatcher-base.js
@@ -11,8 +11,6 @@ const { kDestroy, kClose, kClosed, kDestroyed, kDispatch } = require('../core/sy
 const kOnDestroyed = Symbol('onDestroyed')
 const kOnClosed = Symbol('onClosed')
 
-function noop () {}
-
 class DispatcherBase extends Dispatcher {
   constructor () {
     super()


metcoder95 (commented):
Can you add tests to it?

mcollina:
I think this is a good start. You should make sure tha the result of .compose() can be disposed.

Also, docs are missing.

PandaWorker:
I'm also confused about what 's after .the close() agent no longer accepts requests, I think we need to completely not accept requests only at .destroy(), or add a method .closeIdleConnections(), which will simply wait for requests to complete and close all connections.

mcollina:
> I'm also confused about what 's after .the close() agent no longer accepts requests

I agree, I think it should error at this point but not drop any current requests.

lib/dispatcher/dispatcher-base.js
```diff
@@ -156,6 +157,14 @@ class DispatcherBase extends Dispatcher {
...
     }
   }
+
+  async [Symbol.asyncDispose] () {
+    await this.close()
+  }
+
+  [Symbol.dispose] () {
+    this.close(noop)
+  }
```
Renegade334:
```suggestion
```

Queueing a callback does not equate to a synchronous task. The semantics of explicit resource management are such that by the time the disposer returns (or the async disposer's Promise is resolved), the resource disposal should be complete, to maintain the expected order of operations; this is not the case here.

metcoder95: I don’t believe there’s much else we can do for this; the `close` method is async by default, we cannot ensure sync guarantees.

Renegade334: Aye, hence only the asyncDispose method makes sense here.

lib/dispatcher/dispatcher-base.js
```diff
@@ -8,10 +8,11 @@ const {
   InvalidArgumentError
 } = require('../core/errors')
 const { kDestroy, kClose, kClosed, kDestroyed, kDispatch } = require('../core/symbols')
-
 const kOnDestroyed = Symbol('onDestroyed')
 const kOnClosed = Symbol('onClosed')
 
+function noop () {}
+
```
Uzlopak:
```suggestion
```

PR #4259, opened 2025-06-03
Extract webidl

Uzlopak:
Move webidl into its own web folder, write some tests


lib/web/webidl/index.js
```diff
@@ -531,11 +574,26 @@ webidl.converters.ByteString = function (V, prefix, argument) {
 
 // https://webidl.spec.whatwg.org/#es-USVString
 // TODO: rewrite this so we can control the errors thrown
-webidl.converters.USVString = toUSVString
+webidl.converters.USVString = (() => {
```
KhafraDev, line 577: toWellFormed was added in node v20. We don't need the fallback anymore

mcollina (approved):
lgtm

lib/web/webidl/index.js
Uzlopak: I will provide this in a follow up

PR #4260, opened 2025-06-04
feat: EventSource can be configured with reconnectionTime

Uzlopak:
Tests can be implemented if #4247  is merged

### DIFF
--- a/test/eventsource/eventsource-connect.js
+++ b/test/eventsource/eventsource-connect.js
@@ -4,7 +4,7 @@ const assert = require('node:assert')
 const events = require('node:events')
 const http = require('node:http')
 const { test, describe } = require('node:test')
-const { EventSource } = require('../../lib/web/eventsource/eventsource')
+const { EventSource, defaultReconnectionTime } = require('../../lib/web/eventsource/eventsource')
 const { randomInt } = require('node:crypto')
 
 describe('EventSource - sending correct request headers', () => {
@@ -184,6 +184,7 @@ describe('EventSource - received response must have content-type to be text/even
   })
 
   test('should try to connect again if server is unreachable', async () => {
+    const reconnectionTime = defaultReconnectionTime
     const domain = 'bad.n' + randomInt(1e10).toString(36) + '.proxy'
 
     const eventSourceInstance = new EventSource(`http://${domain}`)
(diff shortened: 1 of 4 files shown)


Uzlopak:
The dispatch option is also not spec compliant

mcollina:
I don't think we should add these kind of features.

Uzlopak:
In smee-client we use the ` eventsource` package. And with that we patch directly the reconnectionTime to 0 for immediate reconnect. I cant set the reconnectionTime with undici. Any suggestions?

KhafraDev:
After reading through https://github.com/whatwg/html/issues/2177, I am a +1 on adding this in a different capacity. Like WebSocket, browsers (mostly Chrome) have seemingly given up on EventSource and have held back improvements on it.

I recommend the option being put under an `undici`/`node` object, something that [cannot be mistaken](https://github.com/nodejs/undici/pull/4260#pullrequestreview-2895099684) as anything other than a deviation from the spec. Then we could add in a dispatcher/headers object later if needed.

This also would needs docs and tests.

```ts
new EventSource('https://my.event.source/', { node: { reconnectionTime: 0 } })
```

KhafraDev:
> The dispatch option is also not spec compliant

I consider adding this to fetch/WebSocket a mistake. It should have been namespaced, like Cloudflare does with custom options. https://developers.cloudflare.com/workers/runtime-apis/request/#the-cf-property-requestinitcfproperties

mcollina:
We could add the namespaced one and deprecate the previous one in the next release (maybe major?).

KhafraDev:
That sounds great to me. I think it drastically reduces confusion and allows us to add whatever we want in the future, without having to regard conflicts with upcoming spec additions or other environments (ie. Deno has a "client" option which functions similarly to our "dispatcher" option, yet neither are compatible). WebSocket is a slightly different issue as passing a plain object to the constructor is a deviation from the spec in and of itself already.

We could recommend the following, which would function in all environments, but is an objectively terrible solution.

```js
const options = ['protocols']
options.headers = { 'Authorization': 'Bearer 123456' }
options.dispatcher = ...

// with the new namespaced options:
options.node = {
  headers: { 'Authorization': 'Bearer 123456' },
  dispatcher: ...
}

const ws = new WebSocket('ws://localhost:1', options)
```

Currently we allow

```js
const ws = new WebSocket('ws://localhost:1', {
  protocols: ['protocols'],
  headers:  { 'Authorization': 'Bearer 123456' },
  dispatcher: ...
})
```

but I think I am getting off topic now, so I digress. I spared you my rant about Chrome holding up the websocket spec, you're welcome. 😂

Uzlopak:
@mcollina 
@KhafraDev 

Is this to your liking? :)

KhafraDev:
Yes but there seem to be related test failures

Uzlopak:
@KhafraDev 

Now runs through.

test/eventsource/eventsource-connect.js
```diff
@@ -199,4 +199,25 @@ describe('EventSource - received response must have content-type to be text/even
...
+        reconnectionTime: 1000
+      }
+    })
+
+    const onerrorCalls = []
+    eventSourceInstance.onerror = (error) => {
+      onerrorCalls.push(error)
+    }
+
+    await new Promise(resolve => setTimeout(resolve, 5000))
```
KhafraDev: This test looks like it will be very flaky

Uzlopak: We already have one test similar to this in the codebase for weeks and it was not very flaky in the ci. Or was it?

Uzlopak: See the test above

KhafraDev: I don't understand that test - it gives 8 seconds for 3 connection errors despite the default reconnection time being  3 seconds? I would expect anything that relies on timing to be flaky, especially when the timing is strict, such as with this test. You give exactly 5 seconds for 5 reconnection errors with a 1 second reconnection time.

Uzlopak:
@mcollina PTAL
@KhafraDev PTAL

test/eventsource/eventsource-connect.js
Uzlopak: Hmm. Yeah, I could make it more resilient.

Uzlopak:
@mcollina 
@KhafraDev 

PTAL. I think the test is now reasonable.

mcollina (approved):
lgtm

KhafraDev (approved):

Issue #4114, opened 2025-03-21
Record and play back requests


GeoffreyBooth:
## This would solve...

For tests, it's useful to be able to mock the responses to outbound network requests so that tests are not reliant on external resources. This is already possible via `MockAgent`, but all the examples at https://github.com/nodejs/undici/blob/main/docs/docs/api/MockAgent.md involve writing code to explicitly define each desired mock response. For an app that makes many outbound network requests, it becomes tedious if not impractical to define every response by hand. I would like to somehow record the outbound network requests and their responses, so that later I can attach them to `MockAgent` to be played back as mocks.

## The implementation should look like...

I think this should be achievable using existing APIs, though it isn’t obvious how to do so. I found an implementation by @aduh95 in [this comment](https://github.com/nock/nock/issues/2397#issuecomment-1924424258) from several years ago, and [a version for `fetch`](https://github.com/nodejs/corepack/blob/main/tests/recordRequests.js) is in Corepack. At minimum I think there should be an example in the Undici docs of how to record requests and then use those recordings with `MockAgent`, assuming this is all possible using existing APIs; and if the example has excessive boilerplate, perhaps a new API such as `RecordingAgent` could be considered.

## I have also considered...

The tool [`nock`](https://github.com/nock/nock/issues/2397#issuecomment-1924424258) has APIs for [recording](https://github.com/nock/nock#recording) mocks, and the latest version of `nock` supports Node global `fetch`, but as far as I can tell `nock` doesn’t support Undici `request`. I would like to use `request` for its increased performance, and because I am using `RetryAgent` and `cacheable-lookup` with `request`.

## Additional context

I’ve been searching online for examples of intercepting requests in Undici and besides the mentioned snippets above, the article https://blog.platformatic.dev/http-fundamentals-understanding-undici-and-its-working-mechanism has some interesting examples.

metcoder95:
Really like this idea; currently the recently introduced the [MockCallHistory](https://undici.nodejs.org/#/docs/api/MockCallHistory?id=class-mockcallhistory) can be of help here.

I imagine an interceptor (or even an Agent) that actively records the requests and returns a `MockCallHistory` instance that can be used to return and reply the recorded requests.

Happy to support this feature

blephy:
Hi,

If i understand well, you want to be able to run your tests suite with real external endpoint and register all the response according to a request configuration.
So then subsequence test suite run, you can mock the registered response for each request.

[...]

PR #4270, opened 2025-06-08
feat: add SnapshotAgent for HTTP request recording and playback

mcollina:
## Summary

This PR implements a new `SnapshotAgent` that extends `MockAgent` to provide automatic recording and playback of HTTP requests for testing purposes. This addresses the feature request in #4114 for "record and play back requests" functionality.

## Features

- **Record Mode**: Captures real HTTP requests and responses to snapshot files
- **Playback Mode**: Replays recorded interactions without making real network calls  
- **Update Mode**: Uses existing snapshots when available, records new ones when missing
- **Full undici compatibility**: Works with `request()`, `fetch()`, `stream()`, `pipeline()`, etc.
- **Base64 response storage**: Consistent serialization of all response body types
- **TypeScript support**: Complete type definitions and comprehensive tests

### Recording Real Requests
```javascript
import { SnapshotAgent, setGlobalDispatcher } from 'undici'

const agent = new SnapshotAgent({ 
  mode: 'record',
  snapshotPath: './test/snapshots/api-calls.json'
})
setGlobalDispatcher(agent)

// Makes real requests and records them
await fetch('https://api.example.com/users')
await agent.saveSnapshots()
```

### Replaying in Tests
```javascript
const agent = new SnapshotAgent({
  mode: 'playback',
  snapshotPath: './test/snapshots/api-calls.json'
})
setGlobalDispatcher(agent)

// Uses recorded response instead of real request
const response = await fetch('https://api.example.com/users')
```

## Test plan

- [x] Comprehensive unit tests for `SnapshotRecorder` utility class
- [x] Integration tests for `SnapshotAgent` in all three modes (record/playback/update)
- [x] TypeScript definition tests using `tsd`
- [x] Example code with real API integration patterns
- [x] POST request handling with request bodies
- [x] Error handling for missing snapshots
- [x] File format validation and persistence


mcollina:
cc @GeoffreyBooth let me know what you think

GeoffreyBooth:
This is a great addition, thank you for doing this! A few things come to mind reading the documentation:

- How can the matching be customized? For example, to match on some headers but not all, to ignore authentication tokens for example. In my app I had subsequent POST calls to the same API where everything matched except the body, so I needed to include that as part of the matching criteria.
- How are subsequent identical calls handled? One [feature of nock](https://github.com/nock/nock#read-this---about-interceptors) is that you can define responses like “the first time, return X, the second time, return Y”.
- How can I update existing mocks? I see the “update mode” example but that seems to be an additive operation to add new mocks to an existing set; is there a way to fully replace an existing set?

Maybe some of these are appropriate to handle as follow-up enhancements. I’ll try to get something together soon that’s a minimal reproduction of what I’ve been working with.

lib/mock/snapshot-agent.js
```diff
@@ -0,0 +1,238 @@
...
+        return this._recordAndReplay(opts, handler)
+      } else {
+        // Playback mode but no snapshot found
+        throw new UndiciError(`No snapshot found for ${opts.method || 'GET'} ${opts.path}`)
+      }
+    } else if (mode === 'record') {
+      // Record mode - make real request and save response
+      return this._recordAndReplay(opts, handler)
+    } else {
+      throw new InvalidArgumentError(`Invalid snapshot mode: ${mode}. Must be 'record', 'playback', or 'update'`)
```
metcoder95, line 113: Shouldn't we add this to the constructor as part of the opts validation?

lib/mock/snapshot-agent.js
```diff
@@ -0,0 +1,238 @@
...
+   * Saves snapshots to file
+   */
+  async saveSnapshots (filePath) {
+    return this[kSnapshotRecorder].saveSnapshots(filePath || this[kSnapshotPath])
+  }
+
+  /**
+   * Sets up MockAgent interceptors based on recorded snapshots
+   */
+  _setupMockInterceptors () {
```
metcoder95, line 253: What's for exactly?

lib/mock/snapshot-recorder.js
```diff
@@ -0,0 +1,182 @@
...
+    const responseData = {
+      statusCode: response.statusCode,
+      headers: normalizeHeaders(response.headers),
+      body: Buffer.isBuffer(response.body) 
+        ? response.body.toString('base64') 
+        : Buffer.from(String(response.body || '')).toString('base64'),
+      trailers: response.trailers
+    }
+
+    this.snapshots.set(hash, {
```
metcoder95: It might be interesting to allow customize the max number of items (as enhancement), just to avoid go out of bounds

metcoder95: It can play with some sort of automatic `flush` that saves it automatically into disk if configured to

metcoder95:
>How can the matching be customized? For example, to match on some headers but not all, to ignore authentication tokens for example. In my app I had subsequent POST calls to the same API where everything matched except the body, so I needed to include that as part of the matching criteria.

This can be a pretty interesting enhancement. Especially for request that requires some level of security, often those ones are good candidates to be excluded from the snapshot itself.

Having a way to state wether or not record a given request/response (either at the dispatch level of agent instantiation) will bring these control benefits.

From my end (and as possible enhancement) I'd like to explore the integration with the `Mocks` feature. As replaying these request under testing environments can be beneficial, especially if custom mocks are already set. I don't believe they should be mutually exclusive but compatible between each other.

lib/mock/snapshot-agent.js
mcollina: This configures the mockPools.

lib/mock/snapshot-recorder.js
```diff
@@ -0,0 +1,182 @@
...
+ * Normalizes headers for consistent comparison
+ */
+function normalizeHeaders (headers) {
+  if (!headers) return {}
+  
+  const normalized = {}
+  
+  // Handle array format (undici internal format: [name, value, name, value, ...])
+  if (Array.isArray(headers)) {
+    for (let i = 0; i < headers.length; i += 2) {
```
FelixVaughan, line 113: Might want to make sure the length of the headers array is even

mcollina:
@Uzlopak @metcoder95 could you take another look?

lib/mock/snapshot-agent.js
```diff
@@ -0,0 +1,318 @@
+'use strict'
+
+const Agent = require('../dispatcher/agent')
+const MockAgent = require('./mock-agent')
+const { SnapshotRecorder } = require('./snapshot-recorder')
+const { InvalidArgumentError, UndiciError } = require('../core/errors')
+// Removed unused import: const { kClients } = require('../core/symbols')
```
metcoder95: leftover?

docs/docs/api/SnapshotAgent.md
```diff
@@ -0,0 +1,405 @@
...
+
+## Constructor
+
+```javascript
+new SnapshotAgent([options])
+```
+
+### Parameters
+
+- **options** `Object` (optional)
```
metcoder95, line 20: aren't we missing the other options available? ([ref](https://github.com/nodejs/undici/pull/4270/files#diff-609e36802b05af005f5a29d44af71da6e1629dac3f2f0645bd66c5d92c4e607eR43-R58))

lib/mock/snapshot-agent.js
```diff
@@ -0,0 +1,318 @@
...
+
+    // Auto-load snapshots in playback/update mode
+    if ((this[kSnapshotMode] === 'playback' || this[kSnapshotMode] === 'update') && this[kSnapshotPath]) {
+      this.loadSnapshots().catch(() => {
+        // Ignore load errors - file might not exist yet
+      })
+    }
+  }
+
+  async dispatch (opts, handler) {
```
metcoder95: Shall we keep it sync whereas possible?

Just `loadSnapshots` are async; unless `playback` or `update` maybe we can delay the `dispatch`, and keep it normal for the rest?

lib/mock/snapshot-recorder.js
```diff
@@ -0,0 +1,490 @@
...
+
+    // Skip if in exclude list (for security)
+    if (excludeSet.has(headerKey)) continue
+
+    // Skip if in ignore list (for matching)
+    if (ignoreSet.has(headerKey)) continue
+
+    // If matchHeaders is specified, only include those headers
+    if (matchHeaders && Array.isArray(matchHeaders)) {
+      const matchSet = new Set(matchHeaders.map(h => caseSensitive ? h : h.toLowerCase()))
```
metcoder95: Can we possibly simplify this? Especially as it seems will do the exact same mapping and conversion on every iteration

test/snapshot-recorder.js
```diff
@@ -0,0 +1,366 @@
...
+const { SnapshotRecorder, formatRequestKey, createRequestHash, filterHeadersForMatching, filterHeadersForStorage } = require('../lib/mock/snapshot-recorder')
+
+test('SnapshotRecorder - basic recording and retrieval', (t) => {
+  const recorder = new SnapshotRecorder()
+
+  const requestOpts = {
+    origin: 'https://api.example.com',
+    path: '/users/123',
+    method: 'GET',
+    headers: { authorization: 'Bearer token' }
```
metcoder95, line 17: Shall we apply heuristics for these situations or better just document it? Especially as these kind of requests can be written into disk

mcollina: What do you mean?

metcoder95: I was referring to the security considerations you added in [4c1d77c](https://github.com/nodejs/undici/pull/4270/commits/4c1d77c525c1d441fa40dda7dadf0eccc0e0ba2a#diff-6727035b6aaada8eff1c42fa39e7b4839914985944f3cc50d504d396f82e8b28R428-R490)

mcollina:
@metcoder95 I've found a bug on the compose/interceptor handling, I'm working on it.

mcollina:
@metcoder95 it’s done.

lib/mock/snapshot-agent.js
```diff
@@ -0,0 +1,334 @@
...
+        // Record the interaction using captured 'self' context (fire and forget)
+        const responseBody = Buffer.concat(responseData.body)
+        self[kSnapshotRecorder].record(opts, {
+          statusCode: responseData.statusCode,
+          headers: responseData.headers,
+          body: responseBody,
+          trailers: responseData.trailers
+        }).catch((error) => {
+          // Log recording errors for debugging
+          console.error('Recording error:', error)
```
metcoder95: is this intentional or should we find another way of warn this?

metcoder95 (approved):

mcollina:
@metcoder95 removed.

mcollina:
@GeoffreyBooth ptal

metcoder95 (approved):

PR #4282, opened 2025-06-16
feat(fetch): introduce `HttpMethod` type and update `RequestInit` method typing

plvo:
## This relates to...

This PR enhances the developer experience by adding proper TypeScript typing for HTTP methods and missing RequestInit properties to match the standard Fetch API specification. https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Methods

Added a new `HttpMethod` type to define valid HTTP methods and updated the method property in `RequestInit` to use this type. Updated tests to reflect these changes.

### Enhanced TypeScript Definitions

- Added `HttpMethod` type: Union of standard HTTP methods + extensible string type using `(string & {})` pattern
- Updated `RequestInit.method`: Now typed as `HttpMethod` instead of generic `string`

### Test Coverage

- Added comprehensive TypeScript tests for all new properties and types
- Enhanced method testing to verify `HttpMethod` typing works correctly
- Maintained backward compatibility - all existing code continues to work

### Features

- HTTP Method Autocompletion: IntelliSense now suggests standard HTTP methods (GET, POST, PUT, DELETE, HEAD, OPTIONS, PATCH, TRACE, CONNECT)
- Extensible Method Typing: Custom HTTP methods still work via `(string & {})` pattern
- Improved Developer Experience: Better type safety and IDE support

None - This is a purely additive change that maintains full backward compatibility. All existing code will continue to work without modification.

## Status

- [x] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/types/fetch.d.ts
+++ b/types/fetch.d.ts
@@ -119,6 +119,8 @@ type RequestDestination =
   | 'worker'
   | 'xslt'
 
+export type HttpMethod = 'GET' | 'HEAD' | 'POST' | 'PUT' | 'DELETE' | 'CONNECT' | 'OPTIONS' | 'TRACE' | 'PATCH' | (string & {})
+
 export interface RequestInit {
   body?: BodyInit | null
   cache?: RequestCache
@@ -128,7 +130,7 @@ export interface RequestInit {
   headers?: HeadersInit
   integrity?: string
   keepalive?: boolean
-  method?: string
+  method?: HttpMethod;
   mode?: RequestMode
   redirect?: RequestRedirect
   referrer?: string
--- a/test/types/fetch.test-d.ts
+++ b/test/types/fetch.test-d.ts
@@ -21,7 +21,8 @@ import {
   ResponseInit,
   ResponseType,
   ReferrerPolicy,
-  Dispatcher
+  Dispatcher,
+  HttpMethod
 } from '../..'
 
 const requestInit: RequestInit = {}
@@ -33,6 +34,12 @@ const requestInit3: RequestInit = {}
 // Test assignment. See https://github.com/whatwg/fetch/issues/1445
 requestInit3.credentials = 'include'
 const requestInit4: RequestInit = { body: null }
+const requestInit5: RequestInit = {
+  method: 'GET'
+}
+const requestInit6: RequestInit = {
+  method: 'POST'
+}
 
 declare const request: Request
 declare const headers: Headers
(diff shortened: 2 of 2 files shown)


Ethan-Arrowood (approved):

tsctx:
> Breaking Changes and Deprecations
> None - This is a purely additive change that maintains full backward compatibility. All existing code will continue to work without modification.
> 

Isn't this a breaking change? It won't be possible to merge of declarations of the following `declare module 'undici' {interface RequestInit {method?: string;}}` .

plvo:
@tsctx You're right about the issue. I've found this is a fundamental TS limitation after testing:

- `method?: string` → Module augmentation works but no autocompletion  
- `method?: HttpMethod` → Autocompletion works but breaks module augmentation

But it works like this 
```ts
import "undici"
import type { HttpMethod } from "undici";

declare module "undici" {
  interface RequestInit {
    method?: HttpMethod;
  }
}
```

types/fetch.d.ts
```diff
@@ -119,6 +119,8 @@ type RequestDestination =
   | 'worker'
   | 'xslt'
 
+export type HttpMethod = 'GET' | 'HEAD' | 'POST' | 'PUT' | 'DELETE' | 'CONNECT' | 'OPTIONS' | 'TRACE' | 'PATCH' | (string & {})
```
Uzlopak, line 122: Maybe do it like we do it in fastify?

https://github.com/fastify/fastify/blob/main/types/utils.d.ts

Also I realize now, you could force uppercase strings for HttpMethod.

https://www.typescriptlang.org/play/?#code/KYDwDg9gTgLgBDAnmYcASMZgLLBgCwgBM4BeOAcgHEBRAFQrgB9K0aBBAEUZYoAUA8gGUGzSnwCqo3pxoAZejR6UAwgIBy6miumUBfOgEkNQ5RToAldiqVj+7OirTKAFBLAooAYwCGAZ2AAHj8YKABLADsAcwA+OAAyOAtgL2giQIjgADdgKAAaOEycqBiASgBYACgq1IiQuAAjABsfAC50TBw8QiIAbQBdMjheigAhOQlRigKKfyIKfqA

plvo: @Uzlopak good idea to use `Autocomplete`, the code is cleaner that way :
```ts
export type HTTPMethods =
  | 'GET'  | 'HEAD'  | 'POST'  | 'PUT'  | 'DELETE'  | 'CONNECT'  | 'OPTIONS'  | 'TRACE'  | 'PATCH'

export type HTTPMethod = Autocomplete<HTTPMethods | Lowercase<HTTPMethods>>

export interface RequestInit {
  method?: HTTPMethod
  // ...
}
```
however, there's still the problem of module augmentation with `method? : string`

Uzlopak: Can you be more specific, please? What do you mean withmodule augmentation?

plvo: I was referring to this message when I was talking about the **module augmentation**

https://github.com/nodejs/undici/pull/4282#issuecomment-2976620722

types/fetch.d.ts
GustavoOS, line 122: could you please add "QUERY" ? See [RFC 1008](https://datatracker.ietf.org/doc/rfc10008/)

PR #4299, opened 2025-06-26
Clarify the type option of the cache interceptor

fredericDelaporte:
People not knowing well the HTTP cache semantic may misleadingly believe that `private` is a safer choice than `shared`. Add a bit of guidance about the effect of private.

## This relates to...

The `type` option of the cache interceptor, so, HTTP caching of fetches done by Undici.

## Rationale

The `private` `type` option of the cache interceptor causes the Undici HTTP cache to act as a private cache while in most circumstances a Node.js application will be shared between many users. This can lead to private data leakage. Furthermore people not understanding well the HTTP cache semantic may misleadingly think choosing `private` instead of `shared` is safer.

## Changes

Add some guidance about the `private` choice.

### Features

N/A

### Bug Fixes

N/A

None

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [x] In review
- [ ] Merge ready

### DIFF
--- a/docs/docs/api/Dispatcher.md
+++ b/docs/docs/api/Dispatcher.md
@@ -1104,7 +1104,7 @@ The `cache` interceptor implements client-side response caching as described in
 - `store` - The [`CacheStore`](/docs/docs/api/CacheStore.md) to store and retrieve responses from. Default is [`MemoryCacheStore`](/docs/docs/api/CacheStore.md#memorycachestore).
 - `methods` - The [**safe** HTTP methods](https://www.rfc-editor.org/rfc/rfc9110#section-9.2.1) to cache the response of.
 - `cacheByDefault` - The default expiration time to cache responses by if they don't have an explicit expiration. If this isn't present, responses without explicit expiration will not be cached. Default `undefined`.
-- `type` - The type of cache for Undici to act as. Can be `shared` or `private`. Default `shared`.
+- `type` - The type of cache for Undici to act as. Can be `shared` or `private`. Default `shared`. `private` means privately cacheable responses will be cached and potentially shared with other users of your application.
 
 ## Instance Events
 
--- a/docs/docs/api/Dispatcher.md
+++ b/docs/docs/api/Dispatcher.md
@@ -1104,7 +1104,7 @@ The `cache` interceptor implements client-side response caching as described in
 - `store` - The [`CacheStore`](/docs/docs/api/CacheStore.md) to store and retrieve responses from. Default is [`MemoryCacheStore`](/docs/docs/api/CacheStore.md#memorycachestore).
 - `methods` - The [**safe** HTTP methods](https://www.rfc-editor.org/rfc/rfc9110#section-9.2.1) to cache the response of.
 - `cacheByDefault` - The default expiration time to cache responses by if they don't have an explicit expiration. If this isn't present, responses without explicit expiration will not be cached. Default `undefined`.
-- `type` - The type of cache for Undici to act as. Can be `shared` or `private`. Default `shared`. `private` means privately cacheable responses will be cached and potentially shared with other users of your application.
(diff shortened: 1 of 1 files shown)


docs/docs/api/Dispatcher.md
```diff
@@ -1104,7 +1104,7 @@ The `cache` interceptor implements client-side response caching as described in
 - `store` - The [`CacheStore`](/docs/docs/api/CacheStore.md) to store and retrieve responses from. Default is [`MemoryCacheStore`](/docs/docs/api/CacheStore.md#memorycachestore).
 - `methods` - The [**safe** HTTP methods](https://www.rfc-editor.org/rfc/rfc9110#section-9.2.1) to cache the response of.
 - `cacheByDefault` - The default expiration time to cache responses by if they don't have an explicit expiration. If this isn't present, responses without explicit expiration will not be cached. Default `undefined`.
-- `type` - The type of cache for Undici to act as. Can be `shared` or `private`. Default `shared`.
+- `type` - The type of cache for Undici to act as. Can be `shared` or `private`. Default `shared`. `private` means privately cacheable responses will be cached and potentially shared with other users of your application.
```
metcoder95: It would be worth it to reference this to the spec or an article that describes the meaning of it

fredericDelaporte: The spec is more permanent than an article which is more likely to turn into a 404, so I would add a link to the RFC. Or MDN maybe, if I find a suitable page on it.

(I assume that you (the Node.js main contributors) wish the documentation to be concise and straight to the point mainly about Node.js itself, so, it should avoid explaining/documenting related concepts that are not Node.js specific. That is why I suggested an edit as concise as I could came up with.)

fredericDelaporte: In the RFC, the definition of shared vs private cache is in the introduction of the spec. Someone in "TLDR" mindset might overlook it. https://www.rfc-editor.org/rfc/rfc9111#name-introduction

I may instead link to the definition of `private` in the cache-directive, but that is less straightforward to infer from there what are shared and private caches. https://www.rfc-editor.org/rfc/rfc9111#name-private

There is also a dedicated section about the risk I see for people misunderstanding what are these types: https://www.rfc-editor.org/rfc/rfc9111#name-caching-of-sensitive-inform But that does not explain the types of cache. (And that is not the only risk. A private type cache on a server might also result in too much "pressure" on the cache storage for very few cache hits due to storing responses with highly specific keys, if their `vary` includes their authorization, by example.)

On MDN, there is a dedicated paragraph about the two types of cache, so, that looks best to me.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Caching#types_of_caches

docs/docs/api/Dispatcher.md
```diff
@@ -1104,7 +1104,7 @@ The `cache` interceptor implements client-side response caching as described in
 - `store` - The [`CacheStore`](/docs/docs/api/CacheStore.md) to store and retrieve responses from. Default is [`MemoryCacheStore`](/docs/docs/api/CacheStore.md#memorycachestore).
 - `methods` - The [**safe** HTTP methods](https://www.rfc-editor.org/rfc/rfc9110#section-9.2.1) to cache the response of.
 - `cacheByDefault` - The default expiration time to cache responses by if they don't have an explicit expiration. If this isn't present, responses without explicit expiration will not be cached. Default `undefined`.
-- `type` - The type of cache for Undici to act as. Can be `shared` or `private`. Default `shared`.
+- `type` - The [type of cache](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/Caching#types_of_caches) for Undici to act as. Can be `shared` or `private`. Default `shared`. `private` implies privately cacheable responses will be cached and potentially shared with other users of your application.
```
fredericDelaporte, line 1107: Example end result:
image

fredericDelaporte, line 1107: The addition of the link alone could be considered enough. Should I remove my kind of warning added sentence?

Ethan-Arrowood (approved):

docs/docs/api/Dispatcher.md
metcoder95: sgtm

mcollina (approved):
lgtm

metcoder95 (approved):

PR #4302, opened 2025-06-27
chore: jsdoc use `@returns` everywhere

Uzlopak:
Is it already hacktober? LOL

We have 200 places where we use `@returns` and only two places where we use `@return`


Uzlopak:
can someone approve this please? it is not changing code, but only jsdoc

PR #4303, opened 2025-06-27
fetch: remove await, add jsdoc for some body read functions

Uzlopak:
the function `fullyReadBody` is sync, so the await is unnecessary.


mcollina (approved):
lgtm

Issue #4295, opened 2025-06-26
Can't parse JSON response exceeding 512 MB


the-sun-will-rise-tomorrow:
## Bug Description

When using `fetch` and `.json()` to fetch and parse a response whose body length exceeds 512 MB, Node throws "Cannot create a string longer than 0x1fffffe8 characters" (`ERR_STRING_TOO_LONG`).

Note that this happens even when no individual string inside the JSON value is long - the limit applies to the size of the entire response body.

## Reproducible By

1. Create a 1GB JSON file:

   ```console
   $ jq -nc '[range(1024 * 1024) | [[range(102) | "abcdefghij"] | add]] | add' > 1gb.json
   ```

2. Start an HTTP server serving the file:

   ```console
   $ npx http-server
   ```

3. Try to `fetch` the file:

   ```js
   fetch("http://127.0.0.1:8080/1gb.json")
	   .then((response) => response.json())
	   .then((data) => {
		   console.log(data.length);
	   })
	   .catch((error) => {
		   console.error(error);
	   });
   ```

## Expected Behavior

Ideally, the JSON response should be parsed as it is received, without being collected into a single buffer or string first, thus avoiding any Node string limits.

## Logs & Screenshots

Above steps fail with:

```
Error: Cannot create a string longer than 0x1fffffe8 characters
    at TextDecoder.decode (node:internal/encoding:447:16)
    at utf8DecodeBytes (node:internal/deps/undici/undici:4863:34)
    at parseJSONFromBytes (node:internal/deps/undici/undici:5738:25)
    at successSteps (node:internal/deps/undici/undici:5719:27)
    at fullyReadBody (node:internal/deps/undici/undici:4609:9)
    at process.processTicksAndRejections (node:internal/process/task_queues:105:5)
    at async consumeBody (node:internal/deps/undici/undici:5728:7) {
  code: 'ERR_STRING_TOO_LONG'
}
```

## Environment

Node v22.14.0 on NixOS (Linux)

### Additional context

The error happens because the response is first collected into a buffer, then decoded into an UTF-8 string, and then parsed from this single string using `JSON.parse`:

https://github.com/nodejs/undici/blob/b1f45cfec28de07ec4950bbe84598b865d70d104/lib/web/fetch/body.js#L493-L499

[...]

PR #4305, opened 2025-06-27
poc: what are bytes sequences

Uzlopak:
regarding #4295 by @the-sun-will-rise-tomorrow reported issue with TextDecoder hitting the limit of 512MB i want to discuss the approach in this PR.

In the spec, I actually dont find a strict definition of what a byte sequence actually is. 

It just says: 

> A byte sequence is a sequence of [bytes](https://infra.spec.whatwg.org/#byte), represented as a space-separated sequence of bytes. Byte sequences with bytes in the range 0x20 (SP) to 0x7E (~), inclusive, can alternately be written as a string, but using backticks instead of quotation marks, to avoid confusion with an actual [string](https://infra.spec.whatwg.org/#string).

https://infra.spec.whatwg.org/#byte-sequence

So we have maybe some wiggle room and can define byte sequence matching that definition but have our own internal representation?

So instead of always passing around the a big Uint8Array we just pass around a Byte Sequence object and if needed transform it to a target representation form. 

As @mcollina pointed out the maximum length of a string in v8 can be 1 GB in x64 architecture. There seems to be a limitation in TextDecoder of only 512MB being allowed and not 1GB. So we have actually some kind of arbitrary restriction, despite that we could have  512 MB longer strings?

I did not test it with a 1GB string. But by avoiding to force the TextDecoder to process a huuuge Uint8Array, it should work also in that case. I expect that the chunks have a size of the corresponding specified highWaterMark, so I dont check if the chunks are above the limit and then split it on the fly. Only drawback is, that we have bigger an bigger growing string. 

Well, this is just a poc. Maybe you say, that this is useless. Or maybe it is anyway an anti pattern to load such huuuge payloads? But we should atleast discuss it.

I just wanted to get the tests pass as much as possible. Of course there are still optimizations possible. Potentially I oversaw some bugs for edge cases. 

E.g. A Byte Sequence Object with only one element doesnt need to be run with concat. If bytesLength is smaller than like e.g. 10 mb just use the old logic and reuse a pre instantiated TextDecoder-instance and just concat everything, instead of chunkWise decoding.

And of course this doesnt solve the limitation, that we dont create JSON in a stream, as we need the whole payload to pass it to JSON.

Now I go to lunch ;)

### DIFF
--- a/benchmarks/fetch/bytes-match.mjs
+++ b/benchmarks/fetch/bytes-match.mjs
@@ -1,24 +1,25 @@
 import { createHash } from 'node:crypto'
 import { bench, run } from 'mitata'
-import { bytesMatch } from '../../lib/web/fetch/util.js'
+import { bytesMatch, ByteSequence } from '../../lib/web/fetch/util.js'
 
-const body = Buffer.from('Hello world!')
-const validSha256Base64 = `sha256-${createHash('sha256').update(body).digest('base64')}`
-const invalidSha256Base64 = `sha256-${createHash('sha256').update(body).digest('base64')}`
-const validSha256Base64Url = `sha256-${createHash('sha256').update(body).digest('base64url')}`
-const invalidSha256Base64Url = `sha256-${createHash('sha256').update(body).digest('base64url')}`
+const buffer = Buffer.from('Hello world!')
+const bytes = new ByteSequence([buffer])
+const validSha256Base64 = `sha256-${createHash('sha256').update(buffer).digest('base64')}`
+const invalidSha256Base64 = `sha256-${createHash('sha256').update(buffer).digest('base64')}`
+const validSha256Base64Url = `sha256-${createHash('sha256').update(buffer).digest('base64url')}`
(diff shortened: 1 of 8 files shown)


KhafraDev:
> So we have maybe some wiggle room and can define byte sequence matching that definition but have our own internal representation?

Yes, this is something I struggled with before landing on either an array of bytes or a TypedArray (this *could* be anything holding a sequence of bytes - even a Blob in my mind).

> the maximum length of a string in v8 can be 1 GB in x64 architecture. There seems to be a limitation in TextDecoder of only 512MB being allowed and not 1GB. So we have actually some kind of arbitrary restriction, despite that we could have 512 MB longer strings?

The limit is [`buffer.constants.MAX_STRING_LENGTH`](https://nodejs.org/docs/latest/api/buffer.html#bufferconstantsmax_string_length). This will not solve the problem.

```mjs
import { constants } from 'node:buffer'

'a'.repeat(constants.MAX_STRING_LENGTH + 1) // RangeError: Invalid string length
```

PR #4307, opened 2025-06-28
feat: extract sri from fetch, upgrade to latest spec

Uzlopak:


Uzlopak:
I am kind of dissappointed of the latest spec.

lib/web/sri/sri.js
```diff
@@ -0,0 +1,302 @@
...
+    // 2. Let expectedValue be the item["val"].
+    const expectedValue = item.val
+
+    // See https://github.com/web-platform-tests/wpt/commit/e4c5cc7a5e48093220528dfdd1c4012dc3837a0e
+    // "be liberal with padding". This is annoying, and it's not even in the spec.
+
+    // 3. Let actualValue be the result of applying algorithm to bytes .
+    const actualValue = applyAlgorithmToBytes(algorithm, bytes)
+
+    // 4. If actualValue is a case-sensitive match for expectedValue,
```
KhafraDev: The spec reads as though "case-sensitive match" is just an ===. It feels like we're missing something if we have to implement our own "base64 === base64url" function.

lib/web/sri/sri.js
```diff
@@ -0,0 +1,302 @@
...
+    const expressionAndOptions = item.split('?', 1)
+
+    // 2. Let algorithm-expression be expression-and-options[0].
+    const algorithmExpression = expressionAndOptions[0]
+
+    // 3. Let base64-value be the empty string.
+    let base64Value = ''
+
+    // 4. Let algorithm-and-value be the result of splitting algorithm-expression on U+002D (-).
+    const algorithmAndValue = [algorithmExpression.slice(0, 6), algorithmExpression.slice(7)]
```
KhafraDev: I know it's equivalent for now, but I would prefer if this used .split('-') instead.

lib/web/sri/sri.js
```diff
@@ -0,0 +1,302 @@
...
+ * Applies the specified hash algorithm to the given bytes
+ *
+ * @typedef {(algorithm: SRIHashAlgorithm, bytes: Uint8Array) => string} ApplyAlgorithmToBytes
+ * @param {SRIHashAlgorithm} algorithm
+ * @param {Uint8Array} bytes
+ * @returns {string}
+ */
+const applyAlgorithmToBytes = crypto.hash
+  ? /** @type {ApplyAlgorithmToBytes} */ (algorithm, bytes) => {
+      return crypto.hash(algorithm, bytes, 'base64')
```
KhafraDev: Will this cause an ExperimentalWarning on node v20? If not, crypto.hash is available in all versions of node undici supports. If it does, we should only use createHash for now.

KhafraDev (commented):
Very good

lib/web/sri/sri.js
Uzlopak: sure, but base64url has `-` as a character. breaking the logic totally

test/sri/get-strongest-metadata.js
```diff
@@ -0,0 +1,75 @@
...
+    const result = getStrongestMetadata([
+      { alg: 'sha256', val: 'sha256-abc' },
+      { alg: 'sha384', val: 'sha384-def' }
+    ])
+    assert.deepEqual(result, [
+      { alg: 'sha384', val: 'sha384-def' }
+    ])
+  })
+
+  test('should return strongest sha256', () => {
```
mozfreddyb: Nit: should return *both* strongest *sha384* (not sha256).

lib/web/sri/sri.js
```diff
@@ -0,0 +1,301 @@
...
+  // 4. For each item in metadata:
+  for (const item of metadata) {
+    // 1. Let algorithm be the item["alg"].
+    const algorithm = item.alg
+
+    // 2. Let expectedValue be the item["val"].
+    const expectedValue = item.val
+
+    // See https://github.com/web-platform-tests/wpt/commit/e4c5cc7a5e48093220528dfdd1c4012dc3837a0e
+    // "be liberal with padding". This is annoying, and it's not even in the spec.
```
mozfreddyb: Can you file an issue? I agree we want that tackled. It's very annoying, but Firefox was perceived as "incompatible" because the implementation in Chrome was more permissive and websites didn't work outside of Chrome. They couldn't easily go through an deprecation, so we had to budge.

The way we do this (iirc), is to decode all of the values liberally (base64, base64url with/without padding) and compare raw bytes

Uzlopak: Whoa! That is of course smarter?!

test/sri/get-strongest-metadata.js
Uzlopak:
```suggestion
  test('should return both strongest sha384', () => {
```

Uzlopak: thanks

lib/web/sri/sri.js
Uzlopak: I didnt see any ExperimentalWarning on node 20.19.2

lib/web/sri/sri.js
Uzlopak: not possible

Uzlopak:
I think we can review and merge it. The problems with base64url are well known.

lib/web/sri/sri.js
KhafraDev: crypto.hash will be available on all supported versions of node then, we don't need the fallback

KhafraDev (approved):

test/fetch/integrity.js
```diff
@@ -149,6 +149,23 @@ test('request with sha512 hash', { skip: !supportedHashes.includes('sha512') },
   }))
 })
 
+test('request with sha512 hash', { skip: !supportedHashAlgorithms.includes('sha512') }, async (t) => {
```
KhafraDev:
```suggestion
test('request with sha512 hash', { skip: !supportedHashAlgorithms.includes('sha512') || !supportedHashAlgorithms.includes('sha384') }, async (t) => {
```

lib/web/subresource-integrity/subresource-integrity.js
```diff
@@ -0,0 +1,308 @@
...
+ * @returns {MetadataList}
+ *
+ * @see https://w3c.github.io/webappsec-subresource-integrity/#parse-metadata
+ */
+function parseMetadata (metadata) {
+  // 1. Let result be the empty set.
+  /** @type {MetadataList} */
+  const result = []
+
+  // 2. For each item returned by splitting metadata on spaces:
```
KhafraDev, line 200: the totally unambiguous wording used in no other spec (not even on the next line) is why I love this spec so much

lib/web/subresource-integrity/subresource-integrity.js
```diff
@@ -0,0 +1,308 @@
...
+ * @type {SRIHashAlgorithm[]}
+ *
+ * The valid SRI hash algorithm token set is the ordered set « "sha256",
+ * "sha384", "sha512" » (corresponding to SHA-256, SHA-384, and SHA-512
+ * respectively). The ordering of this set is meaningful, with stronger
+ * algorithms appearing later in the set.
+ *
+ * @see https://w3c.github.io/webappsec-subresource-integrity/#valid-sri-hash-algorithm-token-set
+ */
+const validSRIHashAlgorithmTokenSet = ['sha256', 'sha384', 'sha512']
```
KhafraDev: nit: it'd be easier and cleaner to replace this with a set, then replace the new Set on line 76 with this one

Uzlopak: @KhafraDev 

The problem is the word "validSRIHashAlgorithmTokenSet". I think the following patch would have been cleaner, but because of "Set" suffix, would be confusing.

```patch
diff --git a/lib/web/subresource-integrity/subresource-integrity.js b/lib/web/subresource-integrity/subresource-integrity.js
index f0702c2c..1390a6cd 100644
--- a/lib/web/subresource-integrity/subresource-integrity.js
+++ b/lib/web/subresource-integrity/subresource-integrity.js
@@ -17,7 +17,7 @@ const assert = require('node:assert')
  */
 
 /**
- * @type {SRIHashAlgorithm[]}
+ * @type {Map<SRIHashAlgorithm, number>}
  *
  * The valid SRI hash algorithm token set is the ordered set « "sha256",
  * "sha384", "sha512" » (corresponding to SHA-256, SHA-384, and SHA-512
@@ -26,7 +26,7 @@ const assert = require('node:assert')
  *
  * @see https://w3c.github.io/webappsec-subresource-integrity/#valid-sri-hash-algorithm-token-set
  */
-const validSRIHashAlgorithmTokenSet = ['sha256', 'sha384', 'sha512']
+const validSRIHashAlgorithmTokenSet = new Map([['sha256', 0], ['sha384', 1], ['sha512', 2]])
 
 // https://nodejs.org/api/crypto.html#determining-if-crypto-support-is-unavailable
 /** @type {import('crypto')} */
@@ -37,21 +37,19 @@ try {
 
   // If no hashes are available, we cannot support SRI.
   if (cryptoHashes.length === 0) {
-    validSRIHashAlgorithmTokenSet.length = 0
+    validSRIHashAlgorithmTokenSet.clear()
   }
 
-  for (let i = 0; i < validSRIHashAlgorithmTokenSet.length; i++) {
-    const algorithm = validSRIHashAlgorithmTokenSet[i]
+  for (const algorithm of validSRIHashAlgorithmTokenSet.keys()) {
     // If the algorithm is not supported, remove it from the list.
     if (cryptoHashes.includes(algorithm) === false) {
-      validSRIHashAlgorithmTokenSet.splice(i, 1)
-      i--
+      validSRIHashAlgorithmTokenSet.delete(algorithm)
     }
   }
   /* c8 ignore next 4 */
 } catch {
   // If crypto is not available, we cannot support SRI.
-  validSRIHashAlgorithmTokenSet.length = 0
+  validSRIHashAlgorithmTokenSet.clear()
 }
 
 /**
@@ -63,7 +61,7 @@ try {
  */
 
 const getSRIHashAlgorithmIndex = /** @type {GetSRIHashAlgorithmIndex} */ (Map.prototype.get.bind(
-  new Map(validSRIHashAlgorithmTokenSet.map((alg, index) => [alg, index]))))
+  validSRIHashAlgorithmTokenSet))
 
 /**
  * @typedef IsValidSRIHashAlgorithm
@@ -73,7 +71,7 @@ const getSRIHashAlgorithmIndex = /** @type {GetSRIHashAlgorithmIndex} */ (Map.pr
  */
 
 const isValidSRIHashAlgorithm = /** @type {IsValidSRIHashAlgorithm} */ (
-  Set.prototype.has.bind(new Set(validSRIHashAlgorithmTokenSet))
+  Map.prototype.has.bind(validSRIHashAlgorithmTokenSet)
 )
 
 /**
```

lib/web/subresource-integrity/subresource-integrity.js
Uzlopak: hmm, yeah seems like it was actually about splitting on various whitespace characters.

lib/web/subresource-integrity/subresource-integrity.js
KhafraDev: That works as well.

KhafraDev (approved):

Issue #4316, opened 2025-07-02
Decompression Interceptor


FelixVaughan:
# Add decompression interceptor for automatic response decompression

On #3253, a future [proposal](https://github.com/nodejs/undici/discussions/3253#discussioncomment-9422210) was made for the addition of a decompression interceptor and I think this would be a worthwhile feature.

## Benefits

Currently, `fetch()` automatically handles compression while `request()` requires manual decompression. This interceptor bridges that gap, making `request()` behavior consistent with `fetch()` and reduces boilerplate for handling compressed responses.

## Implementation Details

- Supports **gzip**, **deflate**, and **brotli** compression
- **Streaming decompression** (memory efficient)
- **Proper header cleanup** (removes content-encoding/length)
- **Error handling** with proper cleanup
- **Comprehensive test coverage**

## Usage Example

```javascript
const { Client } = require('undici')
const createDecompressInterceptor = require('./decompress')

const client = new Client('https://api.example.com')
  .compose(createDecompressInterceptor())

// Now all responses are automatically decompressed
const response = await client.request({ path: '/data' })
```

## Implementation

I've implemented a solution (will link draft PR) with tests and would be interested in getting some feedback. Below is what the interceptor would look like:

```javascript
'use strict'

const { createInflate, createGunzip, createBrotliDecompress } = require('node:zlib')
const DecoratorHandler = require('../handler/decorator-handler')

class DecompressHandler extends DecoratorHandler {
  #decompressor = null

  onResponseStart (controller, statusCode, headers, statusMessage) {
    const contentEncoding = headers['content-encoding']

    const shouldSkip = !contentEncoding || statusCode < 200 || statusCode >= 400 || [204, 304].includes(statusCode)

    if (shouldSkip) {
      return super.onResponseStart(controller, statusCode, headers, statusMessage)
    }

    // Remove content-encoding header since we're decompressing
    const { 'content-encoding': _, 'content-length': __, ...newHeaders } = headers

    const supportedEncodings = {
      gzip: createGunzip,
      deflate: createInflate,
      br: createBrotliDecompress
    }

    const createDecompressor = supportedEncodings[contentEncoding.toLowerCase()]
    if (!createDecompressor) {
      return super.onResponseStart(controller, statusCode, headers, statusMessage)
    }

    this.#decompressor = createDecompressor()

    // Set up decompressor event handlers
    this.#decompressor.on('data', (chunk) => {
      super.onResponseData(controller, chunk)
    })

[...]

PR #4317, opened 2025-07-02
Decompression Interceptor

FelixVaughan:
## This relates to...
Address proposal brought up [here](https://github.com/nodejs/undici/discussions/3253#discussioncomment-9422210)  
and is related to #4316 

## Rationale

This PR implements a decompression interceptor for Undici, as discussed in the linked proposal and issue. The goal is to provide automatic response decompression for `request()` (matching the behavior of `fetch()`), reducing boilerplate and improving developer experience when working with compressed HTTP responses.

## Changes

- Adds a new decompression interceptor for Undici clients
- Supports gzip, deflate, and brotli content-encoding
- Skips decompression for 4xx/5xx, 204, and 304 responses (per spec and browser behavior)
- Removes `content-encoding` and `content-length` headers when decompressing
- Streams decompressed data for memory efficiency
- Additional tests for supported encodings and edge cases

### Features

- Automatic decompression for gzip, deflate, brotli, and zstd responses
- Transparent header cleanup
- Skips decompression for error and no-content responses
- Streaming implementation for large responses
- Case-insensitive encoding support
- Pass-through for unsupported encodings

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [x] In review
- [ ] Merge ready


lib/interceptor/decompress.js
```diff
@@ -0,0 +1,86 @@
...
+const { createInflate, createGunzip, createBrotliDecompress } = require('node:zlib')
+const DecoratorHandler = require('../handler/decorator-handler')
+
+class DecompressHandler extends DecoratorHandler {
+  #decompressor = null
+
+  onResponseStart (controller, statusCode, headers, statusMessage) {
+    const contentEncoding = headers['content-encoding']
+
+    const shouldSkip = !contentEncoding || statusCode < 200 || statusCode >= 400 || [204, 304].includes(statusCode)
```
metcoder95: Although is not common, `5xx` responses can also be compressed; I'd suggest to add an option that allows implementer to customize when to skip decompression while keeping these defaults.

lib/interceptor/decompress.js
```diff
@@ -0,0 +1,86 @@
...
+    })
+
+    return super.onResponseStart(controller, statusCode, newHeaders, statusMessage)
+  }
+
+  onResponseData (controller, chunk) {
+    // If we have a decompressor, pipe data through it
+    if (this.#decompressor) {
+      this.#decompressor.write(chunk)
+      return true
```
metcoder95:
```suggestion
      return this.#decompressor.write(chunk)
```

See: https://nodejs.org/api/stream.html#writablewritechunk-encoding-callback

metcoder95 (commented):
Nice job!

I'd just suggest to implement the decompressor in accordance to [RFC-9110](https://www.rfc-editor.org/rfc/rfc9110.html#name-content-encoding), in that way decompressor can work with different combination in according to standards.

I believe the only missing part might be the support for multiple encodings on a response

FelixVaughan:
@metcoder95 Thanks! All the above have been implemented along with documentation. Let me know if anything else is required

lib/interceptor/decompress.js
```diff
@@ -0,0 +1,146 @@
...
+
+      decompressor.on('error', (error) => {
+        super.onResponseError(controller, error)
+      })
+
+      decompressor.on('finish', () => {
+        super.onResponseEnd(controller, {})
+      })
+    } else {
+      const lastDecompressor = this.#decompressors[this.#decompressors.length - 1]
```
metcoder95: Maybe using [`pipeline`](https://nodejs.org/docs/latest/api/stream.html#streampipelinestreams-callback) can simplify?

lib/interceptor/decompress.js
```diff
@@ -0,0 +1,146 @@
...
+
+      for (let i = 0; i < this.#decompressors.length - 1; i++) {
+        this.#decompressors[i].pipe(this.#decompressors[i + 1])
+      }
+
+      lastDecompressor.on('data', (chunk) => {
+        super.onResponseData(controller, chunk)
```
metcoder95: We also need to account for backpressure here, in case the reader needs to stop the stream to drain buffer

metcoder95 (commented):
Nice job! I left few commenets.

Do not forget about documentation and TS types

lib/interceptor/decompress.js
FelixVaughan: Accounted for backpressure and included stream pipelining for decompression. Documentation and TS types should also be up to date to the best of my knowledge. Also added zstd support

lib/interceptor/decompress.js
```diff
@@ -0,0 +1,157 @@
...
+      })
+
+      lastDecompressor.on('drain', () => {
+        controller.resume()
+      })
+
+      lastDecompressor.on('finish', () => {
+        super.onResponseEnd(controller, {})
+      })
```
metcoder95: you can use the same `pipeline` callback to close up the response stream as it will be called once the pipeline has been done or errored

lib/interceptor/decompress.js
```diff
@@ -0,0 +1,157 @@
...
+
+    this.#decompressors = decompressors
+
+    const { 'content-encoding': _, 'content-length': __, ...newHeaders } = headers
+
+    if (this.#decompressors.length === 1) {
+      const decompressor = this.#decompressors[0]
+
+      decompressor.on('data', (chunk) => {
+        super.onResponseData(controller, chunk)
```
metcoder95: if `onResponseData` returns `false`, we need to also pause the `decompressor` stream until the handler calls `controller.resume` again.

lib/interceptor/decompress.js
```diff
@@ -0,0 +1,157 @@
...
+    }
+
+    return super.onResponseStart(controller, statusCode, newHeaders, statusMessage)
+  }
+
+  onResponseData (controller, chunk) {
+    if (this.#decompressors.length > 0) {
+      const writeResult = this.#decompressors[0].write(chunk)
+      if (writeResult === false) {
+        console.log('pause')
```
metcoder95:
```suggestion
```

lib/interceptor/decompress.js
```diff
@@ -0,0 +1,160 @@
+'use strict'
+
+const { createInflate, createGunzip, createBrotliDecompress, createZstdDecompress } = require('node:zlib')
+const { pipeline } = require('node:stream')
+const DecoratorHandler = require('../handler/decorator-handler')
+
+class DecompressHandler extends DecoratorHandler {
```
metcoder95, line 32: The whole is looking great, you aligned the the backpressure from each chunk processing of the handler when handling the decompressed response, as well as to align the decompressor stream state with the chunk processing one, tho there's a step still missed.

Within the handler there's a controller that also states the state of the chunk processing with the handler, if the controller get's called with `pause`, or `resume` the decompressing stream should react to that. See: https://github.com/nodejs/undici/blob/a36e299d544863c5ade17d4090181be894366024/lib/handler/unwrap-handler.js#L20-L29

If we can add testing around it will be amazing.

FelixVaughan: Right on! What're your thoughts on something like this in the decompressor's `onResponseStart`?

```
const superPause = controller.pause.bind(controller)
const superResume = controller.resume.bind(controller)

controller.pause = () => {
  const result = superPause()
  if (this.#decompressors.length > 0) {
    this.#decompressors.forEach(d => {
      if (!d.readableEnded && !d.destroyed) {
        d.pause()
      }
    })
  }
  return result
}

controller.resume = () => {
  const result = superResume()
  if (this.#decompressors.length > 0) {
    this.#decompressors.forEach(d => {
      if (!d.readableEnded && !d.destroyed) {
        d.resume()
      }
    })
  }
  return result
}
```

Happy to add additional tests, however I'm unsure how to simulate pauses (and backpressure) in the handlers without exposing internals. Open to suggestions!

lib/interceptor/decompress.js
```diff
@@ -0,0 +1,160 @@
...
+        if (result === false) {
+          lastDecompressor.pause()
+        }
+      })
+
+      lastDecompressor.on('drain', () => {
+        controller.resume()
+      })
+
+      pipeline(this.#decompressors, (err) => {
```
metcoder95: Pipeline returns a `stream` which you can subscribe for the chunks of data

lib/interceptor/decompress.js
metcoder95: LGTM, although you can just pause the head of the compressors, and the rest will stop processing any more chunks as they will be waiting for the head to continue pass more chunks.

Additionally you can also pause the resulting stream of `pipeline` if available when the controller gets called (especially as the other streams will continue processing their buffered data).

For the tests, it should be enough by calling `client.dispatch` and set a custom request handler. As you receive the controller during `onRequsestStart`, you should have the chance to pause and resume as needed.

metcoder95 (commented):
LGTM, just tests seems failing

FelixVaughan:
Failures were due to lack of `createZstdCompress` support prior to v22.15.0. Added a skip condition to handle this.

If everything else looks good, I'd like to get out a final commit to clean up some code

lib/interceptor/decompress.js
```diff
@@ -0,0 +1,188 @@
...
+    if (this.#skipErrorResponses && statusCode >= 400) return true
+    return false
+  }
+
+  #createDecompressor (encoding) {
+    const supportedEncodings = {
+      gzip: createGunzip,
+      'x-gzip': createGunzip,
+      br: createBrotliDecompress,
+      zstd: createZstdDecompress,
```
metcoder95: If this is only supported on `>v22`, let's conditionally import it; otherwise can lead to problems if used on v20

FelixVaughan: All set, also added the additional commit for tidying up.

I seem to be getting some flaky failures on unrelated tests though. Is there a way to trigger CI/CD without making changes/empty commits?

mcollina:
Can you add a test on how this would interact with Fetch?

FelixVaughan:
Added tests for direct comparison and global dispatcher scenarios

lib/interceptor/decompress.js
metcoder95: I can trigger flows manually, no problerm

metcoder95 (approved):
lgtm

FelixVaughan:
Can we land this?

lib/interceptor/decompress.js
```diff
@@ -0,0 +1,217 @@
...
+
+  // Creates a chain of decompressors for multiple content encodings
+  #createDecompressionChain (encodings) {
+    const encodingList = encodings.split(',').map(e => e.trim()).filter(Boolean)
+
+    const decompressors = encodingList
+      .reverse() // Apply decompression in reverse order of encoding
+      .map(encoding => this.#createDecompressor(encoding))
+
+    return decompressors.some(d => !d) ? null : decompressors
```
mcollina: it's better to avoid map/some/forEach as they are slow, and this is relatively an hot path.

lib/interceptor/decompress.js
```diff
@@ -0,0 +1,217 @@
...
+      this.#pauseDecompressors()
+      return result
+    }
+
+    controller.resume = () => {
+      const result = superResume()
+      this.#resumeDecompressors()
+      return result
+    }
+  }
```
mcollina, line 86: why is this needed? It shouldn't be

mcollina (changes requested):

lib/interceptor/decompress.js
FelixVaughan: @mcollina This handles explicit pause/resume cycles and bidirectional backpressure control. We pause/resume the controller based on decompressors' state, and need to align decompressors with controller state. 

Without it, decompressor streams stay paused after controller.resume() and the test `should handle multiple pause/resume cycles during decompression` fails. 

Would you recommend an alternative approach?

FelixVaughan:
Any chance we can progress this?

Uzlopak:
@FelixVaughan

Please merge latest main into this PR

FelixVaughan:
@Uzlopak Done

metcoder95 (approved):

Uzlopak:
@FelixVaughan 

Can you evaluate these changes, please?

the most important ones:
- extracted the supportedEncodings outside of the createDecompressor method, should be more performant
- onResponeStart passes a lowercased contentEncoding parameter to createDecompressionChain, thus avoiding that we call multiple times toLowerCase in createDecompressionChain
- avoid .forEach in onResponseError
- cleanupDecompressors, just do `this.#decompressors.length = 0` instead of `this.#decompressors = []`
- added alot of jsdoc

```patch
diff --git a/lib/interceptor/decompress.js b/lib/interceptor/decompress.js
index 87ebe29e..fde4df39 100644
--- a/lib/interceptor/decompress.js
+++ b/lib/interceptor/decompress.js
@@ -4,19 +4,51 @@ const { createInflate, createGunzip, createBrotliDecompress, createZstdDecompres
 const { pipeline } = require('node:stream')
 const DecoratorHandler = require('../handler/decorator-handler')
 
+/** @typedef {import('node:stream').Transform} Transform */
+/** @typedef {import('node:stream').Transform} Controller */
+/** @typedef {Transform&import('node:zlib').Zlib} DecompressorStream */
+
+/** @type {Map<string, () => DecompressorStream>} */
+const supportedEncodings = new Map(Object.entries({
+  gzip: createGunzip,
+  'x-gzip': createGunzip,
+  br: createBrotliDecompress,
+  deflate: createInflate,
+  compress: createInflate,
+  'x-compress': createInflate,
+  ...(createZstdDecompress ? { zstd: createZstdDecompress } : {})
+}))
+
+const defaultSkipStatusCodes = /** @type {const} */ ([204, 304])
+
+/**
+ * @typedef {Object} DecompressHandlerOptions
+ * @property {number[]|Readonly<number[]>} [skipStatusCodes=[204, 304]] - List of status codes to skip decompression for
+ * @property {boolean} [skipErrorResponses] - Whether to skip decompression for error responses (status codes >= 400)
+ */
+
 class DecompressHandler extends DecoratorHandler {
+  /** @type {Transform[]} */
   #decompressors = []
-  #pipelineStream = null
+  /** @type {NodeJS.WritableStream&NodeJS.ReadableStream|null} */
+  #pipelineStream
+  /** @type {Readonly<number[]>} */
   #skipStatusCodes
+  /** @type {boolean} */
   #skipErrorResponses
 
-  constructor (handler, { skipStatusCodes = [204, 304], skipErrorResponses = true } = {}) {
+  constructor (handler, { skipStatusCodes = defaultSkipStatusCodes, skipErrorResponses = true } = {}) {
     super(handler)
     this.#skipStatusCodes = skipStatusCodes
     this.#skipErrorResponses = skipErrorResponses
   }
 
-  // Determines if decompression should be skipped based on encoding and status code
+  /**
+   * Determines if decompression should be skipped based on encoding and status code
+   * @param {string} contentEncoding - Content-Encoding header value
+   * @param {number} statusCode - HTTP status code of the response
+   * @returns {boolean} - True if decompression should be skipped
+   */
   #shouldSkipDecompression (contentEncoding, statusCode) {
     if (!contentEncoding || statusCode < 200) return true
     if (this.#skipStatusCodes.includes(statusCode)) return true
@@ -24,41 +56,37 @@ class DecompressHandler extends DecoratorHandler {
     return false
   }
 
-  // Creates a decompressor stream for the given encoding
-  #createDecompressor (encoding) {
-    const supportedEncodings = {
-      gzip: createGunzip,
-      'x-gzip': createGunzip,
-      br: createBrotliDecompress,
-      deflate: createInflate,
-      compress: createInflate,
-      'x-compress': createInflate
-    }
-    if (createZstdDecompress) supportedEncodings.zstd = createZstdDecompress
-
-    const create = supportedEncodings[encoding.toLowerCase()]
-    return create ? create() : null
-  }
-
-  // Creates a chain of decompressors for multiple content encodings
+  /**
+   * Creates a chain of decompressors for multiple content encodings
+   * 
+   * @param {string} encodings - Comma-separated list of content encodings
+   * @returns {Array<DecompressorStream>} - Array of decompressor streams
+   */
   #createDecompressionChain (encodings) {
     const parts = encodings.split(',')
+
+    /** @type {DecompressorStream[]} */
     const decompressors = []
 
     for (let i = parts.length - 1; i >= 0; i--) {
       const encoding = parts[i].trim()
       if (!encoding) continue
 
-      const decompressor = this.#createDecompressor(encoding)
-      if (!decompressor) return null
+      if (supportedEncodings.has(encoding) === false) {
+        decompressors.length = 0 // Clear if unsupported encoding
+        return decompressors // Unsupported encoding
+      }
 
-      decompressors.push(decompressor)
+      decompressors.push(supportedEncodings.get(encoding)())
     }
 
     return decompressors
   }
 
-  // Pauses all decompressor streams to handle backpressure
+  /**
+   * Pauses all decompressor streams to handle backpressure
+   * @returns {void}
+   */
   #pauseDecompressors () {
     if (this.#decompressors.length > 0) {
       const head = this.#decompressors[0]
@@ -71,7 +99,10 @@ class DecompressHandler extends DecoratorHandler {
     }
   }
 
-  // Resumes all decompressor streams after backpressure is resolved
+  /**
+   * Resumes all decompressor streams after backpressure is resolved
+   * @returns {void}
+   */
   #resumeDecompressors () {
     if (this.#decompressors.length > 0) {
       const head = this.#decompressors[0]
@@ -84,7 +115,11 @@ class DecompressHandler extends DecoratorHandler {
     }
   }
 
-  // Overrides controller pause/resume methods to coordinate with decompressor streams
+  /**
+   * Overrides controller pause/resume methods to coordinate with decompressor streams
+   * @param {Controller} controller - The controller to override
+   * @returns {void}
+   */
   #overrideControllerMethods (controller) {
     const superPause = controller.pause.bind(controller)
     const superResume = controller.resume.bind(controller)
@@ -103,7 +138,12 @@ class DecompressHandler extends DecoratorHandler {
     }
   }
 
-  // Sets up event handlers for a decompressor stream
+  /**
+   * Sets up event handlers for a decompressor stream
+   * @param {Transform} decompressor - The decompressor stream to set up
+   * @param {Controller} controller - The controller to handle events
+   * @returns {void}
+   */
   #setupDecompressorEvents (decompressor, controller) {
     decompressor.on('data', (chunk) => {
       const result = super.onResponseData(controller, chunk)
@@ -121,7 +161,11 @@ class DecompressHandler extends DecoratorHandler {
     })
   }
 
-  // Sets up event handling for a single decompressor
+  /**
+   * Sets up event handling for a single decompressor
+   * @param {Controller} controller - The controller to handle events
+   * @returns {void}
+   */
   #setupSingleDecompressor (controller) {
     const decompressor = this.#decompressors[0]
     this.#setupDecompressorEvents(decompressor, controller)
@@ -131,7 +175,11 @@ class DecompressHandler extends DecoratorHandler {
     })
   }
 
-  // Sets up event handling for multiple chained decompressors using pipeline
+  /**
+   * Sets up event handling for multiple chained decompressors using pipeline
+   * @param {Controller} controller - The controller to handle events
+   * @returns {void}
+   */
   #setupMultipleDecompressors (controller) {
     const lastDecompressor = this.#decompressors[this.#decompressors.length - 1]
     this.#setupDecompressorEvents(lastDecompressor, controller)
@@ -145,12 +193,23 @@ class DecompressHandler extends DecoratorHandler {
     })
   }
 
-  // Cleans up decompressor references to prevent memory leaks
+  /**
+   * Cleans up decompressor references to prevent memory leaks
+   * @returns {void}
+   */
   #cleanupDecompressors () {
-    this.#decompressors = []
+    this.#decompressors.length = 0
     this.#pipelineStream = null
   }
 
+  /**
+   * 
+   * @param {Controller} controller 
+   * @param {number} statusCode 
+   * @param {Record<string, string>} headers 
+   * @param {string} statusMessage 
+   * @returns 
+   */
   onResponseStart (controller, statusCode, headers, statusMessage) {
     const contentEncoding = headers['content-encoding']
 
@@ -159,9 +218,9 @@ class DecompressHandler extends DecoratorHandler {
       return super.onResponseStart(controller, statusCode, headers, statusMessage)
     }
 
-    const decompressors = this.#createDecompressionChain(contentEncoding)
+    const decompressors = this.#createDecompressionChain(contentEncoding.toLowerCase())
 
-    if (!decompressors) {
+    if (decompressors.length === 0) {
       this.#cleanupDecompressors()
       return super.onResponseStart(controller, statusCode, headers, statusMessage)
     }
@@ -192,6 +251,10 @@ class DecompressHandler extends DecoratorHandler {
     return super.onResponseData(controller, chunk)
   }
 
+  /**
+   * @param {Controller} controller 
+   * @param {Record<string, string>} trailers
+   */
   onResponseEnd (controller, trailers) {
     if (this.#decompressors.length > 0) {
       this.#decompressors[0].end()
@@ -201,17 +264,26 @@ class DecompressHandler extends DecoratorHandler {
     return super.onResponseEnd(controller, trailers)
   }
 
+  /**
+   * @param {Controller} controller 
+   * @param {Error} err 
+   */
   onResponseError (controller, err) {
     if (this.#decompressors.length > 0) {
-      this.#decompressors.forEach(d => {
-        d.destroy(err)
-      })
+      for (const decompressor of this.#decompressors) {
+        decompressor.destroy(err)
+      }
       this.#cleanupDecompressors()
     }
     return super.onResponseError(controller, err)
   }
 }
 
+/**
+ * Creates a decompression interceptor for HTTP responses
+ * @param {DecompressHandlerOptions} [options] - Options for the interceptor
+ * @returns {Function} - Interceptor function
+ */
 function createDecompressInterceptor (options = {}) {
   return (dispatch) => {
     return (opts, handler) => {
```

FelixVaughan:
@Uzlopak these optimizations are sound and the docs are a nice touch! 

I noticed annotations are missing on onResponseData, and onResponseStart, onResponseEnd and onResponeError dont have return types on their annotations but lgtm otherwise

Uzlopak:
@FelixVaughan 

I mean, can you integrate them into this PR please? I dont know the types for those methods, and did not want to dig deeper. LOL.

Uzlopak (approved):
Applied my recommended changes.
It adds the said performance optimizations.

Also I added jsdoc which resulted to basically understand the full PR and despite the before existing performance issues, the code is as far as I can see correct.

@metcoder95 
can you please review my changes and approve and merge?

metcoder95 (approved):

Uzlopak (changes requested):
After talk with Matteo, need some experimental warning or use different approach, readable stream instead of data-event

metcoder95:
We can start with an experimental warning and take it from there

Uzlopak:
Exactly. Can you take care of it?

FelixVaughan:
@Uzlopak I looked at `dispatcher.md` and it looks like the handler methods return void in all cases. 

Do you think its worth making changes to the jsdocs and code at this stage? 

Everything still functions as expected

Uzlopak:
We did not merge yet, and if you can make the correct jsdoc, then please add them. 

Also we need to add an experimentalwarning OR you need to change it to a readable stream.

FelixVaughan:
Sounds good, I updated the jsdocs (some code as well) and added the experimental flag

docs/docs/api/Dispatcher.md
```diff
@@ -1094,6 +1094,63 @@ await client.request({
 });
 ```
 
+##### `decompress`
```
metcoder95, line 1097: Let's add a note that this is experimental and can change any time

metcoder95 (approved):
Small not added for documentation, rest lgtm

docs/docs/api/Dispatcher.md
FelixVaughan: Sweet, all set

lib/interceptor/decompress.js
```diff
@@ -0,0 +1,313 @@
...
+/** @type {Map<string, () => DecompressorStream>} */
+const supportedEncodings = new Map(Object.entries({
+  gzip: createGunzip,
+  'x-gzip': createGunzip,
+  br: createBrotliDecompress,
+  deflate: createInflate,
+  compress: createInflate,
+  'x-compress': createInflate,
+  ...(createZstdDecompress ? { zstd: createZstdDecompress } : {})
+}))
```
mcollina: It would be better to use a normal object instead, this is just initialization overhead.

lib/interceptor/decompress.js
mcollina: You should use the `readable` event.

mcollina (approved):
lgtm with a nit.

mcollina (approved):
lgtm

metcoder95 (approved):

gurgunday (approved):
lgtm as well

Uzlopak (approved):
Just for documentation purposes:


After a chat with Matteo I understand the issue, he sees with this PR.


While the interceptor is working, the majority of its code is about managing the streams. All that logic on when to pause and when to resume is a potential beesnest for errors. 


Unfortunately there is no good example in the project on how to do it properly. 


Other interceptors have the handler logic extracted in the handlers folder. This one is an amalgam of interceptor and handler. 


I am personally not well versed in these parts of the project. I really wished it was somehow more intuitive or atleast there was a handler and interceptor which was the gold standard for handlers and interceptors. Then we could use that as a blueprint to standardize handlers and/or interceptors.


E.g. i had a look at the dump interceptor and it seemed that it is more like we need it. But it is storing the payload into the memory. Not the logic on how we can stream the payload out.


IIRC The cache interceptor seems to be better, but it is too complicated to grok it in short time.


Anyhow. There was for sure some thought when interceptors and handlers were designed, but it is not easy to grok. 


So we have onResponseStart, onResponseData and onResponseEnd. So they are like events from a classical stream. But there is nothing intuitive in it, that I could just code a simple pipelining of the incoming response?!


I guess for all the other parts than the body of http requests/response the Interceptors are good?! But for handling the body it feels clunky.


Anyhow… i will make this week a deep dive into the interceptor stuff.

metcoder95:
>So we have onResponseStart, onResponseData and onResponseEnd. So they are like events from a classical stream. But there is nothing intuitive in it, that I could just code a simple pipelining of the incoming response?!

Yeah, but this is aside of the interceptors, is more about the Dispatch handler.

They are intertwined in some way, but the Request events are the ones that manages the data flow, not necessarily the interceptors (the interceptors are just abstractions on top of the Dispatcher handlers).

I can see the potential issue with the body pipelining into the decompressors (same problem I saw when I was playing at the beginning when the issue was made). Everything boils down to the body processing with the decompressor streams, it might be a bit noisy and potentially have some issues, but cannot think on a better way (right now).

I can revisit and adjust later on and based on issues, but not see another way currently.

FelixVaughan:
Happy to revisit and improve upon this as well. In the meantime, it would be great if we could merge this

GeoffreyBooth:
> We can start with an experimental warning and take it from there

@metcoder95 It’s been over a year. Is there a reason that this is still experimental?

metcoder95:
Hasn’t seen any breaking change recently, mostly bug fixes. I believe we can mark it as stable unless strong argument to don’t do it

GeoffreyBooth:
Doing so would remove an experimental warning from my app 😀

It seems stable to me.

PR #4318, opened 2025-07-03
cache: allow caching heuristically cacheable error status codes

### LINKED ISSUE #4293: Cache cacheable response with an error status code
## This would solve...

#3231 introduced HTTP caching (great job!), but in my tests it does not cache cacheable responses with an error status code (specifically, 404 in my case, but all cacheable responses ~~should~~ could be cached, included 204, 400, ...), and responses with error status codes are cacheable according to [RFC 9111](https://datatracker.ietf.org/doc/html/rfc9111).

If there is actually a restriction on cacheable status codes in Undici going beyond what the RFC mandates:
- maybe realign with the spec, (Edit: I realized later it was still aligned since the spec allow to be more restrictive.)
- or introduce an option in the [cache interceptor](https://undici.nodejs.org/#/docs/api/Dispatcher?id=cache-interceptor) to relax this constraint.

## I have also considered...

I have checked my test request gets a proper cache-control header in its 404 response (`public,max-age=60` in my case), and repeatedly calling it from the node server does always actually call the external API. Doing the same test with a request, replied with a 200 and the same cache directive, does not always actually call the external API. Doing the same tests with these requests from a browser does actually cache both cases, 404 and 200 response cases.

## Additional context

This is a nice to have in my opinion, because most gains are already there with the caching of 200 responses, which are usually much more common than other responses. Again, great job with this feature which I really appreciate, and many thanks for it.

These tests were done with Undici 7.10.0 on a Node 22.15.1, just overriding the global dispatcher, still using the default fetch (through ofetch actually, used through Nuxt useFetch). Caching was activated along with HTTP/2 in a [Nitro plugin](https://nitro.build/guide/plugins):

```
import { setGlobalDispatcher, Agent, interceptors } from 'undici'

export default defineNitroPlugin(() => {
  setGlobalDispatcher((new Agent({
    allowH2: true
  })).compose(
    interceptors.cache()
  ))
})
```

mcollina:
## Summary

Enable caching for error responses with heuristically cacheable status codes (204, 404, 410, etc.) when they include explicit cache directives, following RFC 9111 HTTP caching standards.

## Changes

- Modified `canCacheResponse` function in `lib/handler/cache-handler.js` to allow caching responses with status codes from `HEURISTICALLY_CACHEABLE_STATUS_CODES` array when they have explicit cache directives
- Added comprehensive tests for caching 404, 204, and 410 responses with cache headers
- Added test to verify non-heuristically cacheable status codes (418) are not cached

## Background

Previously, only 200 and 307 responses could be cached regardless of cache headers. This was overly restrictive compared to HTTP caching standards, which allow many error status codes to be cached when they have appropriate cache directives.

The existing `HEURISTICALLY_CACHEABLE_STATUS_CODES` array already contained the appropriate status codes (200, 203, 204, 206, 300, 301, 308, 404, 405, 410, 414, 501) but they weren't being used in the main caching decision logic.

## Test plan

- [x] All existing cache tests pass
- [x] New tests verify 404 responses with `Cache-Control: public, max-age=60` are cached
- [x] New tests verify 204 and 410 responses with cache headers are cached  
- [x] New test verifies non-heuristically cacheable status codes (418) are not cached
- [x] Linting passes

Fixes #4293

### DIFF
--- a/test/interceptors/cache.js
+++ b/test/interceptors/cache.js
@@ -1437,137 +1437,57 @@ describe('Cache Interceptor', () => {
     })
   })
 
-  test('caches 404 response with cache headers', async () => {
-    let requestsToOrigin = 0
-    const server = createServer({ joinDuplicateHeaders: true }, (_, res) => {
-      requestsToOrigin++
-      res.statusCode = 404
-      res.setHeader('cache-control', 'public, max-age=60')
-      res.end('Not Found')
-    }).listen(0)
+  const cacheableStatusCodes = [
+    { code: 204, body: '' },
+    { code: 404, body: 'Not Found' },
+    { code: 410, body: 'Gone' }
+  ]
 
-    const client = new Client(`http://localhost:${server.address().port}`)
-      .compose(interceptors.cache())
+  for (const { code, body } of cacheableStatusCodes) {
+    test(`caches ${code} response with cache headers`, async () => {
+      let requestsToOrigin = 0
+      const server = createServer({ joinDuplicateHeaders: true }, (_, res) => {
+        requestsToOrigin++
+        res.statusCode = code
+        res.setHeader('cache-control', 'public, max-age=60')
+        res.end(body)
+      }).listen(0)
 
-    after(async () => {
-      server.close()
-      await client.close()
+      const client = new Client(`http://localhost:${server.address().port}`)
+        .compose(interceptors.cache())
+
+      after(async () => {
+        server.close()
+        await client.close()
+      })
+
+      await once(server, 'listening')
+
+      equal(requestsToOrigin, 0)
+
+      const request = {
+        origin: 'localhost',
+        method: 'GET',
+        path: '/'
+      }
+
+      // First request should hit the origin
+      {
+        const res = await client.request(request)
+        equal(requestsToOrigin, 1)
+        equal(res.statusCode, code)
+        strictEqual(await res.body.text(), body)
+      }
+
+      // Second request should be served from cache
+      {
+        const res = await client.request(request)
+        equal(requestsToOrigin, 1) // Should still be 1 (cached)
+        equal(res.statusCode, code)
+        strictEqual(await res.body.text(), body)
+      }
     })
-
-    await once(server, 'listening')
-
-    equal(requestsToOrigin, 0)
-
-    const request = {
-      origin: 'localhost',
-      method: 'GET',
-      path: '/'
-    }
-
-    // First request should hit the origin
-    {
-      const res = await client.request(request)
-      equal(requestsToOrigin, 1)
-      equal(res.statusCode, 404)
-      strictEqual(await res.body.text(), 'Not Found')
-    }
-
-    // Second request should be served from cache
-    {
-      const res = await client.request(request)
-      equal(requestsToOrigin, 1) // Should still be 1 (cached)
-      equal(res.statusCode, 404)
-      strictEqual(await res.body.text(), 'Not Found')
-    }
-  })
-
-  test('caches 204 response with cache headers', async () => {
-    let requestsToOrigin = 0
-    const server = createServer({ joinDuplicateHeaders: true }, (_, res) => {
-      requestsToOrigin++
-      res.statusCode = 204
-      res.setHeader('cache-control', 'public, max-age=60')
-      res.end()
-    }).listen(0)
-
-    const client = new Client(`http://localhost:${server.address().port}`)
-      .compose(interceptors.cache())
-
-    after(async () => {
-      server.close()
-      await client.close()
-    })
-
-    await once(server, 'listening')
-
-    equal(requestsToOrigin, 0)
-
-    const request = {
-      origin: 'localhost',
-      method: 'GET',
-      path: '/'
-    }
-
-    // First request should hit the origin
-    {
-      const res = await client.request(request)
-      equal(requestsToOrigin, 1)
-      equal(res.statusCode, 204)
-      strictEqual(await res.body.text(), '')
-    }
-
-    // Second request should be served from cache
-    {
-      const res = await client.request(request)
-      equal(requestsToOrigin, 1) // Should still be 1 (cached)
-      equal(res.statusCode, 204)
-      strictEqual(await res.body.text(), '')
-    }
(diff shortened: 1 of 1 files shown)


metcoder95 (approved):

fredericDelaporte:
Sorry to be late on this, but just to let you know if that matters to you:

> allow caching responses with status codes from HEURISTICALLY_CACHEABLE_STATUS_CODES array when they have explicit cache directives

Is still more restrictive than the [RFC](https://datatracker.ietf.org/doc/html/rfc9111#section-3-2.7.1).

> [Storing Responses in Caches](https://datatracker.ietf.org/doc/html/rfc9111#name-storing-responses-in-caches)
> A cache MUST NOT store a response to a request unless:
> ...
> - the response status code is final (see [Section 15](https://www.rfc-editor.org/rfc/rfc9110#section-15) of [[HTTP](https://datatracker.ietf.org/doc/html/rfc9110)]);
> - if the response status code is 206 or 304, or the must-understand cache directive (see [Section 5.2.2.3](https://datatracker.ietf.org/doc/html/rfc9111#cache-response-directive.must-understand)) is present: the cache understands the response status code;
> ...
> - the response contains at least one of the following:
>   - a public response directive (see [Section 5.2.2.9](https://datatracker.ietf.org/doc/html/rfc9111#cache-response-directive.public));
>   - a private response directive, if the cache is not shared (see [Section 5.2.2.7](https://datatracker.ietf.org/doc/html/rfc9111#cache-response-directive.private));
>   - an [Expires](https://datatracker.ietf.org/doc/html/rfc9111#field.expires) header field (see [Section 5.3](https://datatracker.ietf.org/doc/html/rfc9111#field.expires));
>   - a max-age response directive (see [Section 5.2.2.1](https://datatracker.ietf.org/doc/html/rfc9111#cache-response-directive.max-age));
>   - if the cache is shared: an s-maxage response directive (see [Section 5.2.2.10](https://datatracker.ietf.org/doc/html/rfc9111#cache-response-directive.s-maxage));
>   - a cache extension that allows it to be cached (see [Section 5.2.3](https://datatracker.ietf.org/doc/html/rfc9111#cache.control.extensions)); or
>   - a status code that is defined as heuristically cacheable (see [Section 4.2.2](https://datatracker.ietf.org/doc/html/rfc9111#heuristic.freshness)).

The status code being heuristically cacheable is not required for caching, unless none of the previous alternatives are met. Which means, a response containing a cache directive may be cached event if its status code is not heuristically cacheable.

Other paragraphs ~~add further restrictions but also~~ clarify this.

> [5.2.2.9. ](https://datatracker.ietf.org/doc/html/rfc9111#section-5.2.2.9)[public](https://datatracker.ietf.org/doc/html/rfc9111#name-public)
> The public response directive indicates that a cache MAY store the response even if it would otherwise be prohibited, subject to the constraints defined in [Section 3](https://datatracker.ietf.org/doc/html/rfc9111#response.cacheability). In other words, public explicitly marks the response as cacheable. For example, public permits a shared cache to reuse a response to a request containing an Authorization header field ([Section 3.5](https://datatracker.ietf.org/doc/html/rfc9111#caching.authenticated.responses)).
>
> Note that it is unnecessary to add the public directive to a response that is already cacheable according to [Section 3](https://datatracker.ietf.org/doc/html/rfc9111#response.cacheability).
>
> If a response with the public directive has no explicit freshness information, it is heuristically cacheable ([Section 4.2.2](https://datatracker.ietf.org/doc/html/rfc9111#heuristic.freshness)).

~~According to the last sentence the heuristically cacheable test has to be done in the case no expiry is provided in the cache directive.~~ Nothing mandates the response to be heuristically cacheable for explicitly cacheable responses ~~(with expiry for the shared cache case, the private one does not have that constraint of having an explicit expiry)~~.

For example it means 302 and 307 are cacheable too, if there is an explicit caching directive.

But implementing the RFC in order to cache in all circumstances allowed by the RFC looks like it would requires a convoluted logic, which seems undesirable to me. Especially implementing heuristic caching (caching of responses without explicit caching directive) looks cumbersome.

Anyway, your implementation is compliant and aligned with the RFC, just not as permissive as the RFC. It does not mandate caching, but states instead "may cache".

So, why not leaving it as it is currently.

Edit: I think I have miss-understood the sentence:

> If a response with the public directive has no explicit freshness information, it is heuristically cacheable ([Section 4.2.2](https://datatracker.ietf.org/doc/html/rfc9111#heuristic.freshness)).

I have read it as meaning it should have an heuristically cacheable status code. But I now think it just means an heuristic cache duration can be applied to cache response with `public` directive but no expiry.

lib/handler/cache-handler.js
```diff
@@ -241,7 +241,10 @@ class CacheHandler {
  * @param {import('../../types/cache-interceptor.d.ts').default.CacheControlDirectives} cacheControlDirectives
  */
 function canCacheResponse (cacheType, statusCode, resHeaders, cacheControlDirectives) {
-  if (statusCode !== 200 && statusCode !== 307) {
+  // Allow caching for status codes 200 and 307 (original behavior)
+  // Also allow caching for other status codes that are heuristically cacheable
+  // when they have explicit cache directives
+  if (statusCode !== 200 && statusCode !== 307 && !HEURISTICALLY_CACHEABLE_STATUS_CODES.includes(statusCode)) {
```
fredericDelaporte, line 247: The comment seems to imply an intent to cache heuristically cacheable status only if an explicit caching directive is provided, but I have not seen any code checking there is an explicit caching directive. The following code only checks there is not a caching directive forbidding to cache.

So, here, by still allowing 307 caching without checking there is actually an explicit caching directive, undue caching of some cases of 307 responses may happen (307 not being heuristically cacheable, but being cacheable if a directive tells to cache it).

It looks to me the test here should be changed to only check the status is final, and if not, shortcircuit returning false.

Then test if either the status code is not heuristically cacheable and there is neither a public nor a private directive, and in such case, return false.

All that is a bit nit-picking, I do not think it matters much due to theses cases being infrequent, but I prefer let you know in case you think otherwise.

mcollina:
@fredericDelaporte further PRs that improve things are always very welcome! If you want to open a tracking issue.. go ahead!

Issue #4081, opened 2025-03-04
WebSocket diagnostic channels lack request identities


legendecas:
## This would solve...

The diagnostic channels for WebSocket connections have no request identities to chain up diagnostic information. Listeners of these channels can not associate each diagnostic events with a certain request with the limited information.

For example, [`undici:websocket:socket_error`](https://github.com/nodejs/undici/blob/main/docs/docs/api/DiagnosticsChannel.md#undiciwebsocketsocket_error) only provides an `error` object, without a hint of which websocket request caused the error. So do channels like `undici:websocket:open`, `undici:websocket:ping`, and `undici:websocket:pong`.

## The implementation should look like...

The WebSocket diagnostic channel should also provide a consistent request identity (object), like the `WebSocket` instance.

## Additional context

- https://github.com/nodejs/node/issues/53946

PR #4321, opened 2025-07-05
add websocket to websocket diagnostic channels

KhafraDev:
Fixes https://github.com/nodejs/undici/issues/4081

This might be a breaking change? `undici:websocket:open` is now published after the open event is emitted (similar to the current close channel behavior), before it was published before. Also in the opening event, `extensions` and `protocol` are now equivalent to their WebSocket.prototype versions, which means that they are now '' instead of null when empty.


KhafraDev:
If it is a semver-major change, I'll throw in the changes to socket_error too. I was hoping it wouldn't be breaking, but looking at it now it probably is.

mcollina (approved):
why is it breaking? The object only had an additive change. Is there a major change in behavior?

KhafraDev:
`extensions` and `protocol` were both already in the opening channel, but now when not present, they will be an empty string rather than null. It does also change the order of when the opening event is fired, but this isn't as important, but still potentially noticeable if you had something along the lines of:

```js

const events = []
const ws = new WebSocket('ws://localhost')

diagnosticsChannel.subscribe('undici:websocket:open', () => events.push('dc open'))
ws.addEventListener('open', () => events.push('ws open'))

console.log(events) 
// before: ['dc open', 'ws open']
// after: ['ws open', 'dc open']
```

mcollina:
I'd say we ship and revert if we broke anyone.

PR #4322, opened 2025-07-06
websocketstream: close readablestream properly

KhafraDev:
https://github.com/ricea/websocketstream-explainer/issues/25

In the test added, `reader.closed` would hang forever because readablestream.cancel() would always throw due to the stream being locked from the getReader().

### DIFF
--- a/lib/web/websocket/stream/websocketstream.js
+++ b/lib/web/websocket/stream/websocketstream.js
@@ -388,7 +388,7 @@ class WebSocketStream {
     // 6. If the connection was closed cleanly ,
     if (wasClean) {
       // 6.1. Close stream ’s readable stream .
-      this.#readableStream.cancel().catch(() => {})
+      this.#readableStreamController.close()
 
       // 6.2. Error stream ’s writable stream with an " InvalidStateError " DOMException indicating that a closed WebSocketStream cannot be written to.
       if (!this.#writableStream.locked) {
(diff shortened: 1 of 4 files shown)


lib/web/websocket/stream/websocketstream.js
```diff
@@ -388,7 +388,7 @@ class WebSocketStream {
     // 6. If the connection was closed cleanly ,
     if (wasClean) {
       // 6.1. Close stream ’s readable stream .
-      this.#readableStream.cancel().catch(() => {})
+      this.#readableStreamController.close()
```
tsctx, line 391: We should probably handle the rejection of this return value, shouldn't we?
```suggestion
      this.#readableStreamController.close().catch(() => {})
```
Ref: https://streams.spec.whatwg.org/#ws-close

KhafraDev: That is for a WritableStream, we have a ReadableStreamDefaultController, which returns void and can't throw an error afaict https://github.com/nodejs/node/blob/1effb262b64defdaa8e7d41a08b9cc5e9d0c075d/lib/internal/webstreams/readablestream.js#L1066

KhafraDev: It could throw if the stream is already closed which either isn't possible since it only gets closed once the websocket connection closes (can only happen once) or our tests don't cover it.

tsctx: I apologize, I seem to have misunderstood something.

tsctx (approved):

PR #4325, opened 2025-07-07
add ping(websocket, payload) util

KhafraDev:


mcollina (approved):
lgtm

Clever trick!

lib/web/websocket/websocket.js
```diff
@@ -586,8 +588,32 @@ class WebSocket extends EventTarget {
       })
     }
   }
+
+  /**
+   * @param {WebSocket} ws
+   * @param {Buffer|undefined} buffer
+   */
+  static ping (ws, buffer) {
```
tsctx, line 596: The documentation and type definitions suggest that `undefined` is accepted, but the actual implementation doesn't allow it.

KhafraDev: it does?

tsctx: It appears `Buffer.isBuffer(buffer)` doesn't strictly adhere to the documentation, as it filters out anything that isn't a `Buffer`.

Issue #4331, opened 2025-07-11
Pool does not respect keep alive headers and closes connections once queue is empty


Ethan-Arrowood:
This bug was tested on both v6 and v7

The gist is that a `Pool` is not respecting keep-alive timeouts and closes connections once the request queue is empty.  For example, if you create a pool and execute a batch of requests, then execute another batch of requests after some timeout less than either the configured `keepAliveTimeout` option or the response `keep-alive` directive, the connections are **not** reused from the first batch.

I've created a test case in PR #4332 demonstrating the issue.

Ethan-Arrowood:
Discussion is ongoing in the PR thread 👍

Ethan-Arrowood:
Not a bug, I was missing the appropriate `keepAliveTimeout` value in the Node.js server I was using to test. 🤦‍♂️

PR #4332, opened 2025-07-11
Pool does not respect keep-alive bug

Ethan-Arrowood:
Ref #4331

### DIFF
--- a/test/issue-pool-keep-alive.js
+++ b/test/issue-pool-keep-alive.js
@@ -1,0 +1,60 @@
+'use strict'
+
+const { tspl } = require('@matteo.collina/tspl')
+const { test, after } = require('node:test')
+const { Pool } = require('..')
+const { createServer } = require('node:http')
+
+test('pool respects keep-alive', async t => {
+  t = tspl(t, { plan: 201 })
+
+  const server = createServer({ joinDuplicateHeaders: true }, (req, res) => {
+    // Server always responds with a 20s keep-alive header
+    res.writeHead(200, { 'Content-Type': 'text/plain', Connection: 'keep-alive', 'Keep-Alive': 'timeout=20' })
+    res.end('foo')
+  })
+  let connections = 0
+  server.on('connection', () => {
+    connections++
+  })
+  after(() => server.close())
+  const expectedConnections = 10
+  const keepAliveTimeout = 20 * 1000 // 20 seconds
+  server.listen(0, () => {
+    // Pool is set to keep connections alive for 20 seconds too
+    const pool = new Pool(`http://localhost:${server.address().port}`, { connections: expectedConnections, keepAliveTimeout })
+    after(() => pool.close())
+
+    // Execute 100 requests
+    const batchSize = 100
+    let batch1Completed = 0
+    for (let i = 0; i < batchSize; i++) {
+      pool.request({ path: '/', method: 'GET' }, (err, res) => {
+        t.ifError(err)
+        res.body.on('end', () => {
+          batch1Completed++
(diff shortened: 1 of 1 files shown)


ronag:
Any chance to have a simpler test?

Ethan-Arrowood:
Not sure how to simplify this any further

Ethan-Arrowood:
I can try to reduce some of the numbers like connections and total requests if thats what you mean.

Ethan-Arrowood:
Okay it looks like it reproduces with a single connection and only a couple requests per batch. Are you looking for something different?

metcoder95:
Does this happens only on batches or can be reproduced by attempting a single request before the keep alive timeout reaches?

Have you tried attaching a listener to pool for `connectionError` to see if the socket times out in between?

Will try to spend some time over the week on this

Ethan-Arrowood:
Okay so it works fine with just `Client`. It still fails with a singular request (just change batchsize to 1), and I'm not seeing any connectionErrors get emitted. I did try the disconnect event and that fires (obviously) and all its telling me is that after the first request, it gets an "other side closed" error. I don't know how this could be the server's fault since it works fine for a regular `Client`. Is pool not maintaining the client connections the same way?

~It seems like it simply boils down to `Pool` does not respect keep alive~

Its not so simple it turns out... after more testing it seems like a Pool configured to a single connection, with two requests going through it, the second firing half the timeout after the first, **sometimes** it reuses the connection, sometimes it doesn't. 

For very short values like 2s it is flaky. For values from 4s to 10s it seems to consistently respect the keep-alive, but after 20s it seems to always fail. 

I'm always setting the max timeout to double the timeout value. 

Client always respects keep alive no matter the timing. 

---

I debugged and at least this part: https://github.com/nodejs/undici/blob/bug-pool-keep-alive/lib/dispatcher/pool.js#L99 is **always returning the existing client for the second request**. But yet the connection is not being reused.

I added some log lines around the parser timeout (https://github.com/nodejs/undici/blob/bug-pool-keep-alive/lib/dispatcher/client-h1.js#L744) and I don't see it timing out. So its gotta be something else. 

Does anyone else have a better idea? I'm going to keep trying here but would love second set of eyes.

Ethan-Arrowood:
I may have a fix - i think this commit may have broken things: https://github.com/nodejs/undici/commit/faca33ae2a5f92889a69b74e1f72ef244da2875f

Ethan-Arrowood:
I debugged and at least this part: https://github.com/nodejs/undici/blob/bug-pool-keep-alive/lib/dispatcher/pool.js#L99 is returning the existing client for the second request. But yet the connection is not being reused. Trying to debug further and see where its disconnecting.

Ethan-Arrowood:
If I make the timeout longer, like `20 seconds` the client passes pretty consistently, but pool starts failing again 😵

Ethan-Arrowood:
I've been chasing ghosts. And I feel pretty dumb about it. I was missing the fact that Node.js HTTP server has a 5s default keep alive timeout (`keepAliveTimeout` property here https://nodejs.org/api/http.html#httpcreateserveroptions-requestlistener). So it was the Node.js server all along that was shutting down connections. 

My test code verifying `Client` was invalid (i had the second request happening outside the setTimeout and didn't realize until further review). 

So there actually was **no** difference between Client and Server, and the specific timings I was seeing had to do with that 5s default on the server side.

All in all Pool does respect keep alive just like Client does, just make sure the server respects it as well!

Ethan-Arrowood:
Apologies for the noise.

Issue #4334, opened 2025-07-14
Excessive caching of some response cases, and lack of caching of some others


fredericDelaporte:
## Bug Description

The cache interceptor may cache responses it should not:
 - Temporary redirect responses (status 307) without explicit caching directive, if the `cacheByDefault` option is set to a positive number. 307 is not a heuristically cacheable status code, it should not be cached without an explicit response directive telling to do so.
 - 206 responses (Partial Content) can be cached since #4318, but they should be cached only if the cache "understand" their semantic, which does not look to be the case. In short, it should at least include the `Range` request header in the cache key, but I have not found code which would do that. And my test shows the trouble. See [Storing Incomplete Responses](https://datatracker.ietf.org/doc/html/rfc9111#name-storing-incomplete-response).

Some cacheable case remains not cached, while I see no harm in caching them:
 - There are still some cases of cacheable response statuses that are not cached despite having an explicit caching directive, like 302 Found. In short, all non heuristically cacheable statuses excepted 307 are never cached, although final ones can be cached if there is an explicit caching directive in the response.
(This point is not an implementation bug as of the RFC9111, since caching them is only optional, as all caching. But why not caching them?)
Doing this point should take care of not caching 304 responses, unless handling their semantic, meaning, including if-none-match/if-modified-since request headers in their cache key.

## Reproducible By

- Setup a cache interceptor with the `cacheByDefault` set to 60, for example.
Query twice an endpoint replying a 307 without an explicit caching directive in less than 60 seconds.
Server is hit only once.
- Query with a range request an endpoint replying the partial content with a caching directive allowing to cache the response.
Query it again but with a different range before the cache expires.
The partial content for the first request will be served to the second call.
- Query an endpoint replying a 302 with an explicit caching directive allowing to cache it.
Query it again before the cache expires.
The server will be hit twice instead of only once.

## Expected Behavior

- Do not cache non heuristically cacheable responses without an explicit cache directive.
- Do not cache cacheable responses if their semantic is not properly handled by the cache (206, 304).
- Cache other cacheable responses.

## Environment

Windows 11, Node 22.16, testing directly from Undici sources.

### Additional context

See also https://github.com/nodejs/undici/pull/4318#issuecomment-3037129167.

PR #4335, opened 2025-07-14
cache: fix excessive caching and some other lack of caching

fredericDelaporte:
307 responses must not be cached if they do not have an explicit caching directive.

Final responses with an explicit caching directive can be cached even if their status code is not heuristically cacheable status codes.

## This relates to...

fix part of #4334.
I have yet to check there is the troubles I expect about caching 206 and 304 responses without properly handling their semantic at the cache level.

## Changes

- Cease caching 307 responses without caching directive.
- Cache responses with explicit caching directive regardless of their status being heuristically cacheable.

### Features

N/A

### Bug Fixes

- 307 Temporary Redirect without explicit caching directive should not be cached.

N/A

## Status

- [x] Tested
- [S] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/handler/cache-handler.js
+++ b/lib/handler/cache-handler.js
@@ -241,10 +241,13 @@ class CacheHandler {
  * @param {import('../../types/cache-interceptor.d.ts').default.CacheControlDirectives} cacheControlDirectives
  */
 function canCacheResponse (cacheType, statusCode, resHeaders, cacheControlDirectives) {
-  // Allow caching for status codes 200 and 307 (original behavior)
-  // Also allow caching for other status codes that are heuristically cacheable
-  // when they have explicit cache directives
-  if (statusCode !== 200 && statusCode !== 307 && !HEURISTICALLY_CACHEABLE_STATUS_CODES.includes(statusCode)) {
+  // Status code must be final.
+  if (statusCode < 200) {
+    return false
+  }
+  // Responses with neither status codes that are heuristically cacheable, nor explicit caching directives,
+  // are not cacheable.
+  if (!HEURISTICALLY_CACHEABLE_STATUS_CODES.includes(statusCode) && !cacheControlDirectives && !resHeaders['expires']) {
     return false
   }
 
--- a/test/interceptors/cache.js
+++ b/test/interceptors/cache.js
@@ -1437,8 +1437,11 @@ describe('Cache Interceptor', () => {
     })
   })
 
+  // Partial list.
   const cacheableStatusCodes = [
     { code: 204, body: '' },
+    { code: 302, body: 'Found' },
+    { code: 307, body: 'Temporary Redirect' },
     { code: 404, body: 'Not Found' },
     { code: 410, body: 'Gone' }
   ]
@@ -1453,7 +1456,7 @@ describe('Cache Interceptor', () => {
         res.end(body)
       }).listen(0)
 
-      const client = new Client(`http://localhost:${server.address().port}`)
+      const client = new Client(`http://localhost:${server.address().port}`, { maxRedirections: 0 })
         .compose(interceptors.cache())
 
       after(async () => {
@@ -1489,47 +1492,55 @@ describe('Cache Interceptor', () => {
     })
   }
 
-  test('does not cache non-heuristically cacheable error status codes', async () => {
-    let requestsToOrigin = 0
-    const server = createServer({ joinDuplicateHeaders: true }, (_, res) => {
-      requestsToOrigin++
-      res.statusCode = 418 // I'm a teapot - not in heuristically cacheable list
-      res.setHeader('cache-control', 'public, max-age=60')
-      res.end('I am a teapot')
-    }).listen(0)
+  // Partial list.
+  const nonHeuristicallyCacheableStatusCodes = [
+    { code: 201, body: 'Created' },
+    { code: 307, body: 'Temporary Redirect' },
+    { code: 418, body: 'I am a teapot' }
+  ]
 
-    const client = new Client(`http://localhost:${server.address().port}`)
-      .compose(interceptors.cache())
+  for (const { code, body } of nonHeuristicallyCacheableStatusCodes) {
+    test(`does not cache non-heuristically cacheable status ${code} without explicit directive`, async () => {
+      let requestsToOrigin = 0
+      const server = createServer({ joinDuplicateHeaders: true }, (_, res) => {
+        requestsToOrigin++
+        res.statusCode = code
+        res.end(body)
+      }).listen(0)
 
-    after(async () => {
-      server.close()
-      await client.close()
+      const client = new Client(`http://localhost:${server.address().port}`, { maxRedirections: 0 })
+        .compose(interceptors.cache({ cacheByDefault: 60 }))
+
+      after(async () => {
+        server.close()
+        await client.close()
+      })
+
+      await once(server, 'listening')
+
+      equal(requestsToOrigin, 0)
+
+      const request = {
+        origin: 'localhost',
+        method: 'GET',
+        path: '/'
+      }
+
+      // First request should hit the origin
+      {
+        const res = await client.request(request)
+        equal(requestsToOrigin, 1)
+        equal(res.statusCode, code)
+        strictEqual(await res.body.text(), body)
+      }
+
+      // Second request should also hit the origin (not cached)
+      {
+        const res = await client.request(request)
+        equal(requestsToOrigin, 2) // Should be 2 (not cached)
+        equal(res.statusCode, code)
+        strictEqual(await res.body.text(), body)
+      }
     })
-
-    await once(server, 'listening')
-
-    equal(requestsToOrigin, 0)
-
-    const request = {
-      origin: 'localhost',
-      method: 'GET',
-      path: '/'
-    }
-
-    // First request should hit the origin
-    {
-      const res = await client.request(request)
-      equal(requestsToOrigin, 1)
-      equal(res.statusCode, 418)
-      strictEqual(await res.body.text(), 'I am a teapot')
-    }
-
-    // Second request should also hit the origin (not cached)
-    {
-      const res = await client.request(request)
-      equal(requestsToOrigin, 2) // Should be 2 (not cached)
-      equal(res.statusCode, 418)
-      strictEqual(await res.body.text(), 'I am a teapot')
-    }
-  })
+  }
 })


lib/handler/cache-handler.js
```diff
@@ -241,10 +241,13 @@ class CacheHandler {
  * @param {import('../../types/cache-interceptor.d.ts').default.CacheControlDirectives} cacheControlDirectives
  */
 function canCacheResponse (cacheType, statusCode, resHeaders, cacheControlDirectives) {
-  // Allow caching for status codes 200 and 307 (original behavior)
-  // Also allow caching for other status codes that are heuristically cacheable
-  // when they have explicit cache directives
-  if (statusCode !== 200 && statusCode !== 307 && !HEURISTICALLY_CACHEABLE_STATUS_CODES.includes(statusCode)) {
```
fredericDelaporte, line 247: Having a heuristically cacheable status code matters only for responses without an explicit caching directive. It is not a requirement for responses with an explicit caching directive, see [Storing Responses in Caches](https://datatracker.ietf.org/doc/html/rfc9111#name-storing-responses-in-caches).

This previous code also allows to cache 307 responses without an explicit caching directive if the `cacheByDefault` option is used. But 307 is not heuristically cacheable, and so, it should not be cached if its response lacks explicit caching directive.

test/interceptors/cache.js
```diff
@@ -1453,7 +1456,7 @@ describe('Cache Interceptor', () => {
         res.end(body)
       }).listen(0)
 
-      const client = new Client(`http://localhost:${server.address().port}`)
+      const client = new Client(`http://localhost:${server.address().port}`, { maxRedirections: 0 })
```
fredericDelaporte: Since the test includes redirect responses (furthermore, invalid ones), explicitly disable following redirections.

test/interceptors/cache.js
```diff
@@ -1489,47 +1492,55 @@ describe('Cache Interceptor', () => {
     })
   }
 
-  test('does not cache non-heuristically cacheable error status codes', async () => {
```
fredericDelaporte, line 1492: The RFC allows to cache non-heuristically cacheable error responses if they have an explicit caching directive.

So, I change this test to check we do not cache non-heuristically cacheable response not having a caching directive.

lib/handler/cache-handler.js
```diff
@@ -15,6 +15,15 @@ const HEURISTICALLY_CACHEABLE_STATUS_CODES = [
   200, 203, 204, 206, 300, 301, 308, 404, 405, 410, 414, 501
 ]
 
+// Status codes which semantic is not handled by the cache
+// https://datatracker.ietf.org/doc/html/rfc9111#section-3
+// This list should not grow beyond 206 and 304 unless the RFC is updated
+// by a newer one including more. Please introduce another list if
+// implementing caching of responses with the 'must-understand' directive.
```
fredericDelaporte, line 22: Currently responses with the `must-understand` directive may be cached without further checks by the current implementation, if they do not follow the RFC recommendation which states they should also includes a `no-store` directive (that has to be ignored by caches supporting `must-understand`).

So, I think the current implementation is good enough for now, about that directive. (The current implementation appears to me as just reading this directive and flagging it, but not doing anything of it beyond that.)

test/interceptors/cache.js
```diff
@@ -1489,13 +1492,67 @@ describe('Cache Interceptor', () => {
...
+      {
+        const res = await client.request(request)
+        equal(requestsToOrigin, 2) // Should be 2 (not cached)
+        equal(res.statusCode, code)
+        strictEqual(await res.body.text(), body)
+      }
+    })
+  }
+
+  test('discriminates caching of range requests, or does not cache them', async () => {
```
fredericDelaporte, line 1551: Prior the second commit of this PR, this new test was failing.

test/interceptors/cache.js
```diff
@@ -1513,23 +1570,87 @@ describe('Cache Interceptor', () => {
...
-      equal(requestsToOrigin, 2) // Should be 2 (not cached)
-      equal(res.statusCode, 418)
-      strictEqual(await res.body.text(), 'I am a teapot')
+      equal(requestsToOrigin, 2)
+      equal(res.statusCode, code)
+      strictEqual(await res.body.text(), body)
+    }
+  })
+
+  test('discriminates caching of conditionnal requests, or does not cache them', async () => {
```
fredericDelaporte: Prior the second commit of this PR but after the first one, this new test was failing too.

(Prior the first one, it was not, since non heuristically cacheable statuses were not cached, and 304 is not heuristically cacheable.)

test/interceptors/cache.js
```diff
@@ -1513,23 +1570,87 @@ describe('Cache Interceptor', () => {
...
+    request.headers['if-none-match'] = 'another-etag'
+    {
+      const res = await client.request(request)
+      equal(requestsToOrigin, 2)
+      equal(res.statusCode, code)
+      strictEqual(await res.body.text(), body)
+    }
+
+    // Third request with different since should hit the origin too
+    request.headers['if-modified-since'] = new Date(0).toUTCString()
```
fredericDelaporte, line 1692: I have checked this third request case before the fix by moving it as the second request, and it was failing too.

mcollina:
@fredericDelaporte could you check if there is any docs that needs updating?

fredericDelaporte:
I have put a "skipped" in the "Documented" item because I considered no documentation update was required. (Nothing documented about cache has changed, and the change is bug fixes and a slight improvement. I have also checked no mention of 307 being specifically cached were there.)

Maybe I should have instead ticked it for meaning I have done these checks?

test/interceptors/cache.js
fredericDelaporte: This test needs to be split in two, because `if-none-match`, if supported by the cache, causes `if-modified-since` to be ignored. So, its third assert could wrongly fail.

lib/handler/cache-handler.js
```diff
@@ -241,10 +250,13 @@ class CacheHandler {
...
-  // Also allow caching for other status codes that are heuristically cacheable
-  // when they have explicit cache directives
-  if (statusCode !== 200 && statusCode !== 307 && !HEURISTICALLY_CACHEABLE_STATUS_CODES.includes(statusCode)) {
+  // Status code must be final and understood.
+  if (statusCode < 200 || NOT_UNDERSTOOD_STATUS_CODES.includes(statusCode)) {
+    return false
+  }
+  // Responses with neither status codes that are heuristically cacheable, nor explicit caching directives,
+  // are not cacheable.
+  if (!HEURISTICALLY_CACHEABLE_STATUS_CODES.includes(statusCode) && !cacheControlDirectives && !resHeaders['expires']) {
```
fredericDelaporte: `cacheControlDirectives` is never undefined or null but just `{ }` if there is no directive. So, that is a bug here.

But I have not seen yet why the tests still succeed on my setup.

Edit: On my setup the test response has a date which causes the `cacheByDefault` directive to be ignored for determining the freshness.

lib/handler/cache-handler.js
```diff
@@ -241,10 +250,19 @@ class CacheHandler {
...
+  }
+  // Responses with neither status codes that are heuristically cacheable, nor "explicit enough" caching
+  // directives, are not cacheable. "Explicit enough": see https://www.rfc-editor.org/rfc/rfc9111.html#section-3
+  if (!HEURISTICALLY_CACHEABLE_STATUS_CODES.includes(statusCode) && !resHeaders['expires'] &&
+    !cacheControlDirectives.public &&
+    cacheControlDirectives['max-age'] === undefined &&
+    // RFC 9111: a private response directive, if the cache is not shared
+    !(cacheControlDirectives.private && cacheType === 'private') &&
+    !(cacheControlDirectives['s-maxage'] !== undefined && cacheType === 'shared')
+  ) {
```
fredericDelaporte, line 265: Simply testing there is a `cache-control` directive is not enough. It is approximate, and thinking about it again, it may allow to cache cases that should not.

But the complete test is not straightforward.

Especially I thought writing ` !(cacheControlDirectives.private === true && cacheType === 'private')` instead of ` !(cacheControlDirectives.private && cacheType === 'private')`, because a qualified `private` directive allow caching by shared caches if they strip the headers listed by `private`.

But the RFC wording, "a private response directive, if the cache is not shared", can be understood as private directive, qualified or not, should allow caching (in the absence of the other alternatives) only for private caches.

lib/handler/cache-handler.js
```diff
@@ -239,15 +248,30 @@ class CacheHandler {
...
+    if (!HEURISTICALLY_CACHEABLE_STATUS_CODES.includes(statusCode) && !resHeaders['expires'] &&
+      !cacheControlDirectives.public &&
+      cacheControlDirectives['max-age'] === undefined &&
+      // RFC 9111: a private response directive, if the cache is not shared
+      !(cacheControlDirectives.private && cacheType === 'private') &&
+      !(cacheControlDirectives['s-maxage'] !== undefined && cacheType === 'shared')
+    ) {
+      return false
+    }
+  }
```
fredericDelaporte: When I considered the subject, I was not expecting it to turn that way because I was not aware of the `cacheByDefault` option.

Due to the code preceding the call to canCacheResponse, which actually does already some cacheability checks, we need that big test only when cacheByDefault is truthy.

See https://github.com/nodejs/undici/blob/4608ef157cc75d5ce3d2802e6949cb865d73146c/lib/handler/cache-handler.js#L115C1-L125C6

Maybe that code should be moved into `canCacheResponse` instead of staying out of it, allowing to have a better picture of how the cache interceptor checks cacheability.

fredericDelaporte:
The remaining CI failures seem to be a flaky websockets test. (There is a flaky cache test too which has blown on my setup once, name "revalidates the request when the response is stale".)

lib/handler/cache-handler.js
```diff
@@ -239,15 +248,31 @@ class CacheHandler {
...
+function canCacheResponse (cacheType, statusCode, resHeaders, cacheControlDirectives, cacheByDefault) {
+  // Status code must be final and understood.
+  if (statusCode < 200 || NOT_UNDERSTOOD_STATUS_CODES.includes(statusCode)) {
     return false
   }
 
+  if (cacheByDefault) {
+    // When cacheByDefault is truthy, the cache interceptor does no more mandate an explicit expiry or
+    // heuristically cacheability prior to reaching here. We should then check the full condition allowing
+    // to consider caching responses.
```
fredericDelaporte: That is wrong, even when cacheByDefault is not defined or falsy, an explicit expiry is not mandated by the current implementation. (Also see #4338.) This shortcut has to be removed.

test/interceptors/cache.js
metcoder95: Max redirections is not supported anymore, you'd require to use the interceptor instead.

fredericDelaporte: If I understand it well, it means that redirection are not followed by default, an interceptor has to be added for that. Fine for the test then.

fredericDelaporte: I have removed this option.

fredericDelaporte:
Is there is something more I should do for this PR?

PR #4338, opened 2025-07-16
Fix misleading cacheByDefault documentation

fredericDelaporte:
## This relates to...

The `cacheByDefault` states:

> The default expiration time to cache responses by if they don't have an explicit expiration. If this isn't present, responses without explicit expiration will not be cached.

That is wrong. If the response has a `last-modified` header, an heuristic expiry will be used by current Undici code base instead of `cacheByDefault`. If the response is heuristically cacheable, it will be cached even if `cacheByDefault` is undefined.

See

https://github.com/nodejs/undici/blob/4608ef157cc75d5ce3d2802e6949cb865d73146c/lib/handler/cache-handler.js#L346C26-L358

https://github.com/nodejs/undici/blob/4608ef157cc75d5ce3d2802e6949cb865d73146c/lib/handler/cache-handler.js#L115C1-L125C6

## Changes

 - Update `cacheByDefault` about heuristic caching.

### Features

N/A

### Bug Fixes

 - Inaccurate/outdated documentation.

N/A

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/docs/docs/api/Dispatcher.md
+++ b/docs/docs/api/Dispatcher.md
@@ -1103,7 +1103,7 @@ The `cache` interceptor implements client-side response caching as described in
 
 - `store` - The [`CacheStore`](/docs/docs/api/CacheStore.md) to store and retrieve responses from. Default is [`MemoryCacheStore`](/docs/docs/api/CacheStore.md#memorycachestore).
 - `methods` - The [**safe** HTTP methods](https://www.rfc-editor.org/rfc/rfc9110#section-9.2.1) to cache the response of.
-- `cacheByDefault` - The default expiration time to cache responses by if they don't have an explicit expiration. If this isn't present, responses without explicit expiration will not be cached. Default `undefined`.
+- `cacheByDefault` - The default expiration time to cache responses by if they don't have an explicit expiration and cannot have an heuristic expiry computed. If this isn't present, responses neither with an explicit expiration nor heuristically cacheable will not be cached. Default `undefined`.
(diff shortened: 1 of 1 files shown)


docs/docs/api/Dispatcher.md
```diff
@@ -1103,7 +1103,7 @@ The `cache` interceptor implements client-side response caching as described in
 
 - `store` - The [`CacheStore`](/docs/docs/api/CacheStore.md) to store and retrieve responses from. Default is [`MemoryCacheStore`](/docs/docs/api/CacheStore.md#memorycachestore).
 - `methods` - The [**safe** HTTP methods](https://www.rfc-editor.org/rfc/rfc9110#section-9.2.1) to cache the response of.
-- `cacheByDefault` - The default expiration time to cache responses by if they don't have an explicit expiration. If this isn't present, responses without explicit expiration will not be cached. Default `undefined`.
+- `cacheByDefault` - The default expiration time to cache responses by if they don't have an explicit expiration and cannot have an heuristic expiry computed. If this isn't present, responses neither with an explicit expiration nor heuristically cacheable will not be cached. Default `undefined`.
```
fredericDelaporte, line 1106: Any suggestion of a better wording is welcome. I do not find this easy to understand. But my attempts to make it more straightforward resulted in a lot more sentences, which ends up no better.

fredericDelaporte:
Actually I find it surprising to have heuristic caching enabled by default in a shared cache (shared in most case at least). I was not expecting to have heuristic caching at all in the cache interceptor. And I wonder if that is really a desirable default feature of it.

Also see #4335. Correctly supporting heuristic caching while not disabling more desirable caching (in my opinion at least) like caching of responses with explicit caching directive (but non heuristically cacheable status) is a bit complicated.

PR #4340, opened 2025-07-16
feat(ProxyAgent) improve Curl-y behavior in HTTP->HTTP Proxy connections (#4180)

caitp:
This refactors the way the legacy unsecured behaviour is implemented, by wrapping the Proxy client in a wrapper which rewrites requests, and handles errors. This will also insert authentication headers in each request.

CURRENTLY:

For nonsecure channels, clients allocated by the pool all delegate to the proxy client, and reuse the same HttpContext. This is probably not a good thing, but at least in cases where only one request is performed at a time, it seems harmless enough. In order to make this ready to land, this probably needs to handle simultaneous concurrent requests and have a test for that.

## Status

- [x] Tested
- [x] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/dispatcher/proxy-agent.js
+++ b/lib/dispatcher/proxy-agent.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const { kProxy, kClose, kDestroy, kDispatch, kConnector } = require('../core/symbols')
+const { kProxy, kClose, kDestroy, kDispatch } = require('../core/symbols')
 const { URL } = require('node:url')
 const Agent = require('./agent')
 const Pool = require('./pool')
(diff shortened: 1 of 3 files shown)


lib/dispatcher/proxy-agent.js
```diff
@@ -26,62 +25,74 @@ function defaultFactory (origin, opts) {
...
   }
 
   async [kClose] () {
-    await this.#client.close()
+    // FIXME: remove wrapper from the pool
   }
 
   async [kDestroy] () {
-    await this.#client.destroy()
+    // FIXME: remove wrapper from the pool
```
metcoder95: why not close the `this.#client`?

caitp: as noted in the PR description, `this.#client` is currently the original Proxy client -- which I think is actually a huge problem, but I haven't written a test yet to demonstrate that it is, which is probably next up. Maybe it's actually fine to reuse the same HttpContext for concurrent requests, but I suspect it isn't

But given that it is the Proxy server client, closing it breaks the ProxyAgent

caitp:
I've attempted to add a test for this reuse of the HttpContext, and haven't been able to reproduce any issues yet (although the test case is pretty small) -- so maybe it actually does work?

but still need to look into the other CI failures.

caitp:
summary of the changes:

- `factory` can now be passed to ProxyAgent again, which was needed to get tests using MockAgent etc working. By default, it will use something lifted from Agent.js's defaultFactory, which takes `options.connections` into account to decide whether to creates a Pool or a Client. 
- Http1ProxyWrapper no longer reuses the existing connection to the Proxy, but instead creates a new one, and retains it for the life time of the ProxyWrapper. The wrapper is responsible for rewriting requests in the format of HTTP1 non-tunneling proxy requests
- proxyTunnel now defaults to true again, so the non-tunneling becomes opt-in again (was already the case in the last update) -- however now, tests have been updated to explicitly specify the proxyTunnel value, so that it's less of a change needed when flipping the flag.

metcoder95 (approved):

PR #4343, opened 2025-07-17
speed up flaky websocket test

KhafraDev:
Locally, this makes the test go from >2 seconds to <50ms

I saw it time out in a few places:
https://github.com/nodejs/undici/actions/runs/16346630382/job/46181950819
https://github.com/nodejs/undici/actions/runs/16346630382/job/46181950577


mcollina (approved):
lgtm

PR #4349, opened 2025-07-18
fetch: make readAllBytes sync

Uzlopak:
Because we cant access the ReadableStreamDefaultReader logic inside node core to provide our own reader with the specified logic as mentioned in the spec. Before readAllBytes is async, which is confusing because we provide two callbacks to it. Now it reads more natural for me. 

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready


KhafraDev (changes requested):
There is no reason to change it

mcollina (commented):
This allocates way less promises and I expect to improve performance. 

I don't really understand how this work tho. What if part of the body is not available?

Uzlopak:
If body is null or invalid then it would break already in an earlier stage in the function which calls readAllBytes, because we call ` getReader()`  on the body which can throw an error. We expect in readAllFiles the generated Reader.

The only place we can error is when the ` reader` is locked, so that ` read` will throw. Other than that i dont see any code path possible to result in an error.

Uzlopak:
LOL, I actually found a bug in the sri code. I was wondering why setting the rejection handler in readLoop is not working.

mcollina:
What about if a large file is transmitted? Is fetch already buffering all of that in memory?

Uzlopak:
Only if you use one of those methods from the bodyMixing, like  `.text()``,.json()` and so on. If you know that you expect a huge file you should stream it anyway.

For streams you have that logic directly in mainFetch called incrementalReadBody or so. There we use alot of magic like async generators.

gurgunday:
Can you bench it somehow? I think it's a pretty good change if there's a significant improvement

Uzlopak:
@gurgunday 

What I need is a way to count the generated Promises.

gurgunday:
I use https://github.com/bengl/count-promises for that

gurgunday (approved):
lgtm

mcollina (approved):
lgtm

mcollina:
@Uzlopak this now conflicts. can you use https://github.com/nodejs/undici/pull/4349#issuecomment-3194195829 to count the promises?

@KhafraDev this seems to allocate way less promises than async/await combo.

Uzlopak:
My change reduces according to the "test" that the amount of promises is reduced by 1 per fetch call

KhafraDev:
1 less promise?

Uzlopak:
yes, 1 less promise. We create about 50 promises for a fetch call. 

See the other PRs regarding removing async/promises. Maybe we can remove overall about 10 promises.

test/fetch/count-promises.js
```diff
@@ -0,0 +1,37 @@
...
+const { once } = require('node:events')
+const { createServer } = require('node:http')
+const { test } = require('node:test')
+
+const countPromises = require('count-promises')
+const { fetch } = require('../..')
+
+const { closeServerAsPromise } = require('../utils/node-http')
+
+const body = 'abcdefg'.repeat(1e6)
```
Uzlopak: Make one small body to reduce the noice by the fact that of course big payloads get chunked, thus increasing the promises. First goal should be reducing the amount of promises for a simple fetch call 

```suggestion
const body = 'abcdefgh'
```

gurgunday (approved):
Just my 50c but I think this and other PRs that remove unnecessary promises are ultimately beneficial and they add up

mcollina (approved):
lgtm

mcollina:
@KhafraDev can you take another look? I don't see how this is problematic

trivikr:
This needs a rebase, so that latest tests can run

PR #4357, opened 2025-07-22
fix: avoid setup global dispatcher in advance

himself65:
## Changes

lazy setup dispatcher so that lazyLoadUndici in node.js core module won't affect the `globalThis`

### Bug Fixes

Fixes: https://github.com/nodejs/node/issues/59012

## Status

- [x] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/global.js
+++ b/lib/global.js
@@ -6,20 +6,34 @@ const globalDispatcher = Symbol.for('undici.globalDispatcher.1')
 const { InvalidArgumentError } = require('./core/errors')
 const Agent = require('./dispatcher/agent')
 
-if (getGlobalDispatcher() === undefined) {
-  setGlobalDispatcher(new Agent())
+let currentDispatcher = null
+
+if (Object.hasOwn(globalThis, globalDispatcher)) {
+  const existing = Reflect.get(globalThis, globalDispatcher)
+  if (existing && typeof existing.dispatch === 'function') {
+    currentDispatcher = existing
+  }
+} else {
+  Object.defineProperty(globalThis, globalDispatcher, {
+    get () {
+      if (currentDispatcher === null) {
+        currentDispatcher = new Agent()
+      }
+      return currentDispatcher
+    },
+    set (agent) {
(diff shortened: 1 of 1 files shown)


mcollina:
CI doesn't seem happy

himself65:
Yeah I see, trying to figure out

himself65:
@mcollina updated the code

metcoder95 (approved):

gurgunday (approved):

lib/global.js
```diff
@@ -6,22 +6,36 @@ const globalDispatcher = Symbol.for('undici.globalDispatcher.1')
...
-    writable: true,
+    get () {
+      if (currentDispatcher === null) {
+        currentDispatcher = new Agent()
+      }
+      return currentDispatcher
+    },
+    set (agent) {
+      setGlobalDispatcher(agent)
+    },
```
mcollina, line 26: Unfortunately this implementation is not equivalent. After this change, whoever is reading the dispatcher would invoke a function, which would generate a slight overhead. The solution is on first access to remove all this getter business.

lib/global.js
```diff
@@ -6,22 +6,36 @@ const globalDispatcher = Symbol.for('undici.globalDispatcher.1')
...
     configurable: false
   })
 }
 
+function setGlobalDispatcher (agent) {
+  if (!agent || typeof agent.dispatch !== 'function') {
+    throw new InvalidArgumentError('Argument agent must implement Agent')
+  }
+  currentDispatcher = agent
+}
```
mcollina, line 37: I'm relatively certain this function is already available somewhere

mcollina (changes requested):

KhafraDev:
Would it make more sense for undici to always overwrite the global dispatcher (if it's not from undici already)? Importing `node:http` might not cause issues anymore, but you'd still have these issues the second you used any of the undici-provided globals, right?

PR #4359, opened 2025-07-24
perf: use Pool instead of Client in ProxyAgent for HTTP proxy connections

mcollina:
## Summary
- Changed ProxyClient to use Pool instead of Client for HTTP proxy connections
- Updated factory function to use Pool for non-tunneling scenarios
- Added proper connector handling for CONNECT methods when tunneling is disabled

## Benefits
- Better connection management through connection pooling
- Improved performance for HTTP proxy scenarios
- Maintains all existing functionality and backward compatibility

## Changes
- `ProxyClient` constructor now uses `new Pool()` instead of `new Client()`
- Added `#connector` field to store raw connector for CONNECT method handling
- Updated factory function fallback to also use Pool
- Cleaned up unused imports (`kConnector`, `Client`)

## Test Plan
- [x] All existing proxy-agent tests pass
- [x] Linting passes
- [x] Both tunneling and non-tunneling scenarios work correctly

### DIFF
--- a/lib/dispatcher/proxy-agent.js
+++ b/lib/dispatcher/proxy-agent.js
@@ -1,13 +1,12 @@
 'use strict'
 
-const { kProxy, kClose, kDestroy, kDispatch, kConnector } = require('../core/symbols')
+const { kProxy, kClose, kDestroy, kDispatch } = require('../core/symbols')
 const { URL } = require('node:url')
 const Agent = require('./agent')
 const Pool = require('./pool')
 const DispatcherBase = require('./dispatcher-base')
 const { InvalidArgumentError, RequestAbortedError, SecureProxyConnectionError } = require('../core/errors')
 const buildConnector = require('../core/connect')
-const Client = require('./client')
 
 const kAgent = Symbol('proxy agent')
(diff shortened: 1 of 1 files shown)


mcollina:
cc @caitp (I know this now conflicts)

lib/dispatcher/proxy-agent.js
```diff
@@ -40,7 +40,9 @@ class ProxyClient extends DispatcherBase {
 
     super()
 
-    this.#client = new Client(origin, opts)
+    // Store the connector for direct socket connections
+    this.#connector = opts?.connect || buildConnector(opts)
+    this.#client = new Pool(origin, opts)
```
caitp, line 45: the new Http1ProxyWrapper does this, although looking at the diff it's kind of clumbsy how it does it -- but the default factory it uses will make a Pool unless opts.connections === 1, which is pulled from the regular Agent factory.

I did switch to using a user supplied factory in order to get MockAgent tests passing, though. maybe it is a good simplification to just always create a Pool, though

mcollina: Then this PR is not needed

PR #4361, opened 2025-07-25
fix: detect and prevent redirect loops with Client/Pool

### LINKED ISSUE #4360: Redirect interceptor fails to redirect
## Bug Description

When creating my own undici Client, 3xx responses with a Location header don't  get followed properly. 

I don't _always_ see this 🤔 Some redirects seem to work just fine. I have a little unit test here where it works great on 5.x. I think it's something about redirecting with the same path but a different origin. OR maybe something with absolute URLs.

It's like the client is using the wrong origin when it calls the `Location` URL.

It'll _try_ to redirect, but it ends up calling the original URL and getting the same redirect header back. It'll loop for as long as `maxRedirects` will let it.

On undici 5.x I get this behaviour out-of-the-box from `client.request`. On 7.x I'm getting it when I add my own redirect interceptor.

## Reproducible By

Here's a little demo script: two basic HTTP servers are set up. One redirects to the other, but the redirect is never followed.

```
import http from "http";
import assert from "assert";
import { Client, interceptors } from "undici";

const a = http.createServer((req, res) => {
  console.log(req.url);
  res.writeHead(301, {
    Location: `http://localhost:8081${req.url}`,
  });
  res.end();
});
a.listen(8080);

const b = http.createServer((req, res) => {
  res.writeHead(200);
  res.end();
});
b.listen(8081);

const client = new Client("http://localhost:8080").compose(
  interceptors.redirect({
    maxRedirections: 5,
  })
);
const r = await client.request({
  method: "GET",
  path: "/abc",
});

console.log(r.statusCode);
assert.equal(r.statusCode, 200);
```

## Expected Behavior

I mean the URL in the `Location` header should be getting called by the redirect interceptor.

My test code above should not throw.

## Environment

Node 22.13.0
Seen on Undici 7.12 (went back a few minors and still seeing it) and undici 5.x (where we caught this in prod)

### Additional context

We found this issue because undici was failing to  follow a redirect on a particular URL. The redirect is actually just http -> https. I was trying to repro this in my test script locally but redirecting to two localhost servers seemed to repro the problem.

I don't want to share the server here (it's in development) but I think I can share privately. The URL will redirect properly with `curl`, but I'm seeing this in my local logs:
```
UNDICI 122454: connecting to <private> using http:undefined
UNDICI 122454: connecting to <private> using http:undefined
UNDICI 122454: connected to <private> using http:h1
UNDICI 122454: connected to <private> using http:h1
UNDICI 122454: sending request to GET http:/ <private>/openmrs/ws/rest/v1/location
UNDICI 122454: sending request to GET http:/ <private>//openmrs/ws/rest/v1/location
UNDICI 122454: received response to GET http:/ <private>/openmrs/ws/rest/v1/location - HTTP 308
UNDICI 122454: received response to GET http:/ <private>//openmrs/ws/rest/v1/location - HTTP 308
UNDICI 122454: trailers received from GET http:/ <private>/openmrs/ws/rest/v1/location
UNDICI 122454: trailers received from GET http:/ <private>//openmrs/ws/rest/v1/location
UNDICI 122454: sending request to GET https:/ <private>/openmrs/ws/rest/v1/location
UNDICI 122454: sending request to GET https:/ <private>//openmrs/ws/rest/v1/location
UNDICI 122454: received response to GET https:/ <private>/openmrs/ws/rest/v1/location - HTTP 308
UNDICI 122454: received response to GET https:/ <private>//openmrs/ws/rest/v1/location - HTTP 308
UNDICI 122454: trailers received from GET https:/ <private>/openmrs/ws/rest/v1/location
UNDICI 122454: trailers received from GET https:/ <private>//openmrs/ws/rest/v1/location
UNDICI 122454: sending request to GET https:/ <private>/openmrs/ws/rest/v1/location
UNDICI 122454: sending request to GET https:/ <private>//openmrs/ws/rest/v1/location
UNDICI 122454: received response to GET https:/ <private>/openmrs/ws/rest/v1/location - HTTP 308
UNDICI 122454: received response to GET https:/ <private>//openmrs/ws/rest/v1/location - HTTP 308
UNDICI 122454: trailers received from GET https:/ <private>/openmrs/ws/rest/v1/location
UNDICI 122454: trailers received from GET https:/ <private>//openmrs/ws/rest/v1/location
```

`fetch()` seems to work absolutely fine.

mcollina:
## Summary
Fixes redirect interceptor failing to redirect with Client and Pool by detecting redirect loops that occur when single-host dispatchers attempt cross-origin redirects.

## Problem
Issue #4360 reported that when using Client or Pool with the redirect interceptor, cross-origin redirects would fail silently and loop infinitely instead of following the redirect properly. This happened because:

1. Client and Pool are single-host dispatchers bound to specific origins
2. They cannot handle cross-origin redirects properly
3. Instead of failing cleanly, they would loop trying to redirect to the same URL repeatedly

## Solution  
- Added redirect loop detection in `RedirectHandler`
- When a redirect loop is detected, throw a clear error message instead of infinite looping
- The error message guides users to use Agent for cross-origin redirects
- Agent continues to work correctly for cross-origin redirects

## Test Plan
- Added comprehensive tests covering Client, Pool, and Agent behavior
- ✅ Client throws clear error for cross-origin redirects  
- ✅ Pool throws clear error for cross-origin redirects
- ✅ Agent successfully follows cross-origin redirects
- ✅ Same-origin redirects continue to work with Agent

## Error Message
Before: Silent infinite loops until maxRedirections reached
After: `Redirect loop detected. Cannot redirect to ${origin}. This typically happens when using a Client or Pool with cross-origin redirects. Use an Agent for cross-origin redirects.`

Fixes #4360

### DIFF
--- a/lib/handler/redirect-handler.js
+++ b/lib/handler/redirect-handler.js
@@ -139,7 +139,7 @@ class RedirectHandler {
     const redirectUrlString = `${origin}${path}`
     for (const historyUrl of this.history) {
       if (historyUrl.toString() === redirectUrlString) {
-        throw new Error(`Redirect loop detected. Cannot redirect to ${origin}. This typically happens when using a Client or Pool with cross-origin redirects. Use an Agent for cross-origin redirects.`)
+        throw new InvalidArgumentError(`Redirect loop detected. Cannot redirect to ${origin}. This typically happens when using a Client or Pool with cross-origin redirects. Use an Agent for cross-origin redirects.`)
       }
     }
 
--- a/lib/handler/redirect-handler.js
+++ b/lib/handler/redirect-handler.js
@@ -133,6 +133,16 @@ class RedirectHandler {
     const { origin, pathname, search } = util.parseURL(new URL(this.location, this.opts.origin && new URL(this.opts.path, this.opts.origin)))
     const path = search ? `${pathname}${search}` : pathname
 
+    // Check for redirect loops by seeing if we've already visited this URL in our history
+    // This catches the case where Client/Pool try to handle cross-origin redirects but fail
+    // and keep redirecting to the same URL in an infinite loop
+    const redirectUrlString = `${origin}${path}`
+    for (const historyUrl of this.history) {
+      if (historyUrl.toString() === redirectUrlString) {
+        throw new Error(`Redirect loop detected. Cannot redirect to ${origin}. This typically happens when using a Client or Pool with cross-origin redirects. Use an Agent for cross-origin redirects.`)
+      }
+    }
+
     // Remove headers referring to the original URL.
     // By default it is Host only, unless it's a 303 (see below), which removes also all Content-* headers.
     // https://tools.ietf.org/html/rfc7231#section-6.4
--- a/test/interceptors/redirect-cross-origin-fix.js
+++ b/test/interceptors/redirect-cross-origin-fix.js
@@ -1,0 +1,123 @@
+'use strict'
+
+const { test, after } = require('node:test')
+const { createServer } = require('node:http')
+const { once } = require('node:events')
+const { tspl } = require('@matteo.collina/tspl')
+const undici = require('../..')
+
+const {
+  interceptors: { redirect }
+} = undici
+
+test('Client should throw redirect loop error for cross-origin redirect', async (t) => {
+  t = tspl(t, { plan: 2 })
+
+  const serverA = createServer((req, res) => {
+    res.writeHead(301, {
+      Location: 'http://localhost:9999/target' // Different port = cross-origin
+    })
+    res.end()
+  })
+
+  serverA.listen(0)
+  after(() => serverA.close())
+  await once(serverA, 'listening')
+
+  const client = new undici.Client(`http://localhost:${serverA.address().port}`).compose(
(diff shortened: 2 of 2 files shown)


lib/handler/redirect-handler.js
```diff
@@ -133,6 +133,16 @@ class RedirectHandler {
     const { origin, pathname, search } = util.parseURL(new URL(this.location, this.opts.origin && new URL(this.opts.path, this.opts.origin)))
     const path = search ? `${pathname}${search}` : pathname
 
+    // Check for redirect loops by seeing if we've already visited this URL in our history
+    // This catches the case where Client/Pool try to handle cross-origin redirects but fail
+    // and keep redirecting to the same URL in an infinite loop
+    const redirectUrlString = `${origin}${path}`
+    for (const historyUrl of this.history) {
+      if (historyUrl.toString() === redirectUrlString) {
+        throw new Error(`Redirect loop detected. Cannot redirect to ${origin}. This typically happens when using a Client or Pool with cross-origin redirects. Use an Agent for cross-origin redirects.`)
```
metcoder95: Shall we use an `undici`'s error?

metcoder95 (approved):

metcoder95 (approved):

PR #4362, opened 2025-07-28
remove node:buffer imports

KhafraDev:
Blob has been available since node [v18](https://nodejs.org/docs/latest/api/globals.html#class-blob)
File has been available since node [v20](https://nodejs.org/docs/latest/api/globals.html#class-file)


mcollina (approved):
lgtm

PR #4364, opened 2025-07-30
Amend the GC warning in the README to clarify that consuming bodies is recommended but not required

mcollina:
For the curious mind, this was implemented in https://github.com/nodejs/undici/pull/3199.


Ethan-Arrowood:
Its also not just about garbage collection, right? If you keep around the body without consuming it, then the connection could also stay open. At least in certain circumstances.

Ethan-Arrowood (approved):
I see we already document about connection usage. LGTM 👍

Issue #4156, opened 2025-04-11
`error:0A00018F:SSL routines::ee key too small` in tests using modern openssl


LiviaMedeiros:
## Bug Description

`test/http2.js`, `test/h2c-client.js`, and `test/connect-pre-shared-session.js` are failing with the `error:0A00018F:SSL routines::ee key too small` error on modern systems.

## Reproducible By

```console
$ openssl --version
OpenSSL 3.4.1 11 Feb 2025 (Library: OpenSSL 3.4.1 11 Feb 2025)
$ npm i && npm run test
```

## Expected Behavior

Tests not failing and being future-proof.

## Logs & Screenshots

Example of error:
```yaml
test at test/http2.js:1345:1
✖ #2364 - Concurrent aborts (2nd variant) (1.548565ms)
  Error: error:0A00018F:SSL routines::ee key too small
      at node:internal/tls/secure-context:70:13
      at Array.forEach (<anonymous>)
      at setCerts (node:internal/tls/secure-context:68:3)
      at configSecureContext (node:internal/tls/secure-context:191:5)
      at Object.createSecureContext (node:_tls_common:113:3)
      at Server.setSecureContext (node:_tls_wrap:1490:27)
      at new Server (node:_tls_wrap:1354:8)
      at new Http2SecureServer (node:internal/http2/core:3174:5)
      at createSecureServer (node:internal/http2/core:3385:10)
      at TestContext.<anonymous> (/home/livia/zram2/undici/test/http2.js:1346:18) {
    library: 'SSL routines',
    reason: 'ee key too small',
    code: 'ERR_SSL_EE_KEY_TOO_SMALL'
  }
```

## Environment

`dev-libs/openssl-3.4.1` built without `weak-ssl-ciphers` USE flag (it's disabled by default).

### Additional context

The culprits are likely to be `http-pem` and `selfsigned` packages. The former unconditionally calls `selfsigned.generate()` without any parameters and the latter has key of size `1024` by default.

metcoder95:
This started hitting tests recently; opened and issue time ago to make it customizable (see https://github.com/watson/https-pem/issues/9), but I believe I'll make a fork out of it to fix it.

Will try to have a look later this week

himself65:
I can see it on my PR https://github.com/nodejs/undici/pull/4357

PR #4365, opened 2025-07-30
Support for capping the total number of connections across all origins

JoshMock:
Addresses https://github.com/nodejs/undici/issues/3268, but currently only in `Agent`. If maintainers believe the approach is sound, I will happily add support to `BalancedPool` as well.

The general approach:

1. Expose a `maxGlobalConnections` option in `Agent`
2. Add a `maxGlobalConnectionsReached()` function that can be passed down to `Client`
3. If a `Client` attempts to create a new connection when the max has been reached, add the request to the `Client` queue and clean up the socket
4. If an `Agent` attempts to dispatch when the max has been reached, add the request to the (new!) `Agent` queue
5. When `Agent` receives a `disconnect` event, drain requests from its queue until `maxGlobalConnections` is reached or a dispatch returns `false`

### DIFF
--- a/docs/docs/api/Agent.md
+++ b/docs/docs/api/Agent.md
@@ -19,6 +19,7 @@ Returns: `Agent`
 Extends: [`PoolOptions`](/docs/docs/api/Pool.md#parameter-pooloptions)
 
 * **factory** `(origin: URL, opts: Object) => Dispatcher` - Default: `(origin, opts) => new Pool(origin, opts)`
+* **maxGlobalConnections** `number | null` - Default: `null` - Limits the total number of connections that can be open at one time across all origins. If `null`, no limit is enforced.
 
 ## Instance Properties
 
--- a/lib/dispatcher/agent.js
+++ b/lib/dispatcher/agent.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const { InvalidArgumentError } = require('../core/errors')
-const { kClients, kRunning, kClose, kDestroy, kDispatch, kUrl } = require('../core/symbols')
+const { kClients, kConnected, kRunning, kClose, kDestroy, kDispatch, kUrl, kQueue, kResume } = require('../core/symbols')
 const DispatcherBase = require('./dispatcher-base')
 const Pool = require('./pool')
 const Client = require('./client')
@@ -15,7 +15,7 @@ const kFactory = Symbol('factory')
 const kOptions = Symbol('options')
 
 function defaultFactory (origin, opts) {
-  return opts && opts.connections === 1
+  return opts && (opts.connections === 1 || opts.maxGlobalConnections === 1)
     ? new Client(origin, opts)
     : new Pool(origin, opts)
 }
@@ -39,34 +39,71 @@ class Agent extends DispatcherBase {
     this[kOptions] = { ...util.deepClone(options), connect }
     this[kFactory] = factory
     this[kClients] = new Map()
+    this[kQueue] = []
+
+    if (options.maxGlobalConnections != null) {
+      if (!Number.isFinite(options.maxGlobalConnections) || options.maxGlobalConnections <= 0) {
+        throw new InvalidArgumentError('maxGlobalConnections must be a number greater than 0')
+      }
+
+      if (options.connections != null && options.maxGlobalConnections < options.connections) {
+        throw new InvalidArgumentError('maxGlobalConnections must be greater than or equal to connections')
+      }
+    }
+
+    this[kOptions].maxGlobalConnectionsReached = () => {
+      if (options.maxGlobalConnections == null) return false
+      return (this[kConnected] >= options.maxGlobalConnections)
+    }
 
     this[kOnDrain] = (origin, targets) => {
-      this.emit('drain', origin, [this, ...targets])
+      this.emit('drain', origin.origin, [this, ...targets])
     }
 
     this[kOnConnect] = (origin, targets) => {
-      const result = this[kClients].get(origin)
+      const key = origin.origin
+      const result = this[kClients].get(key)
       if (result) {
         result.count += 1
       }
-      this.emit('connect', origin, [this, ...targets])
+      this.emit('connect', key, [this, ...targets])
     }
 
     this[kOnDisconnect] = (origin, targets, err) => {
-      const result = this[kClients].get(origin)
+      const key = origin.origin
+      const result = this[kClients].get(key)
       if (result) {
         result.count -= 1
-        if (result.count <= 0) {
-          this[kClients].delete(origin)
+        if (result.count <= 0 && err?.code !== 'UND_ERR_INFO') {
+          this[kClients].delete(key)
           result.dispatcher.destroy()
         }
       }
-      this.emit('disconnect', origin, [this, ...targets], err)
+      this.emit('disconnect', key, [this, ...targets], err)
+      this[kResume]()
     }
 
     this[kOnConnectionError] = (origin, targets, err) => {
       // TODO: should this decrement result.count here?
-      this.emit('connectionError', origin, [this, ...targets], err)
+      this.emit('connectionError', origin.origin, [this, ...targets], err)
+    }
+  }
+
+  get [kConnected] () {
+    return this[kClients].values().reduce((acc, { count = 0 }) => acc + count, 0)
+  }
+
+  [kResume] () {
+    for (const { dispatcher } of this[kClients].values()) {
+      dispatcher[kResume](true)
+    }
+
+    while (this[kQueue].length > 0 && !this[kOptions].maxGlobalConnectionsReached()) {
+      const { opts, handler } = this[kQueue].shift()
+      const result = this[kDispatch](opts, handler)
+      if (!result) {
+        break
+      }
     }
   }
 
@@ -84,6 +121,11 @@ class Agent extends DispatcherBase {
       key = String(opts.origin)
     } else {
       throw new InvalidArgumentError('opts.origin must be a non-empty string or URL.')
+    }
+
+    if (this[kOptions].maxGlobalConnectionsReached()) {
+      this[kQueue].push({ opts, handler })
+      return false
     }
 
     const result = this[kClients].get(key)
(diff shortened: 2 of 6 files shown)


lib/dispatcher/agent.js
```diff
@@ -86,6 +123,11 @@ class Agent extends DispatcherBase {
       throw new InvalidArgumentError('opts.origin must be a non-empty string or URL.')
     }
 
+    if (this[kOptions].maxGlobalConnectionsReached()) {
+      this[kQueue].push({ opts, handler })
+      return false
+    }
```
metcoder95: Lets implement something similar as per what `PoolBase` has to enqueue requests when no more connections available.

If not taken with care will grow out of bounds, not to mention that what happens if no `connection` gets dropped with any origin?

docs/docs/api/Agent.md
```diff
@@ -19,6 +19,7 @@ Returns: `Agent`
 Extends: [`PoolOptions`](/docs/docs/api/Pool.md#parameter-pooloptions)
 
 * **factory** `(origin: URL, opts: Object) => Dispatcher` - Default: `(origin, opts) => new Pool(origin, opts)`
+* **maxGlobalConnections** `number | null` - Default: `null` - Limits the total number of connections that can be open at one time across all origins. If `null`, no limit is enforced.
```
metcoder95: I'd also add a similar feature to `Pool`, and possible also forward it in cascade to `Pool` (although maybe is better to use another option for that).

lib/dispatcher/client.js
```diff
@@ -440,6 +447,12 @@ async function connect (client) {
       return
     }
 
+    if (client[kMaxGlobalConnectionsReached] && client[kMaxGlobalConnectionsReached]()) {
```
metcoder95: Let's not add it to the `Client`. `Client` maps to a single socket; in all cases `Pool` and `Agent` are the ones that requires this property.

lib/dispatcher/agent.js
```diff
@@ -39,34 +39,71 @@ class Agent extends DispatcherBase {
...
 
     this[kOnConnectionError] = (origin, targets, err) => {
       // TODO: should this decrement result.count here?
-      this.emit('connectionError', origin, [this, ...targets], err)
+      this.emit('connectionError', origin.origin, [this, ...targets], err)
+    }
+  }
+
+  get [kConnected] () {
+    return this[kClients].values().reduce((acc, { count = 0 }) => acc + count, 0)
```
mcollina: try avoiding a reduce(), it's super slow. Also the continuous conversion of `kClients` to values will slow things down significantly. We probably need to maintain two data structures for them.

mcollina (commented):
Good work! It's a good start! I wonder if it would be better to maintain a counter in the Agent instead of sum it up at every request.

lib/dispatcher/client.js
JoshMock: Just making sure I understand: you're suggesting moving this logic upstream to `Pool` so that `Client` would be unaware of the limit? :+1:

JoshMock:
> For `Agent` the situation is a bit more complicated.
> As it manages pool/client per origin, might be better to re-shape the approach for it by limiting the number of `origins` if desired (defaulting to `Inifinity`).

This approach makes sense to me, as controlling max connections would just become a downstream effect of limiting `origins`, with max connections effectively becoming `origins * connections`.

> If no more room for adding more origins, requests should be dropped

When you say it's dropped, you mean that Agent shouldn't have a queue at all? If so, would it throw an exception like `AgentMaxOriginsReached`?

docs/docs/api/Agent.md
JoshMock: Since I've updated this to use `origins` instead, it no longer makes sense to cascade anything to `Pool` since it only manages one upstream origin.

JoshMock:
Okay, I've simplified it to be an `origins` restriction in `Agent`, which simplified things quite a bit. `Agent[kOrigins]` tracks unique origins in its own `Set` now.

I kept the queue in `Agent` until I get clarification from @metcoder95 on [my question](https://github.com/nodejs/undici/pull/4365#issuecomment-3140555790) about whether to queue or throw an error. If we do keep the queue, should I switch it from an array to a `FixedQueue`, to align it more closely with `PoolBase` as you suggested [here](https://github.com/nodejs/undici/pull/4365#discussion_r2244696882)?

lib/dispatcher/client.js
metcoder95: Exactly 👍

metcoder95:
>When you say it's dropped, you mean that Agent shouldn't have a queue at all? If so, would it throw an exception like AgentMaxOriginsReached?

Exactly; if a max limit of `origin`s has been reached, I'm doubtful new connections to origins will be dropped in a timely manner (staling the request), as is complicated to ensure when a connection will be dropped to leave space for a new one.

docs/docs/api/Agent.md
```diff
@@ -19,6 +19,7 @@ Returns: `Agent`
 Extends: [`PoolOptions`](/docs/docs/api/Pool.md#parameter-pooloptions)
 
 * **factory** `(origin: URL, opts: Object) => Dispatcher` - Default: `(origin, opts) => new Pool(origin, opts)`
+* **origins** `number` (optional) - Default: `Infinity` - Limits the total number of origins that can receive requests at a time. If `Infinity`, no limit is enforced.
```
metcoder95: Maybe just align with other properties that has a similar purpose? i.e. `maxOrigins`

lib/dispatcher/agent.js
```diff
@@ -13,15 +13,16 @@ const kOnConnectionError = Symbol('onConnectionError')
 const kOnDrain = Symbol('onDrain')
 const kFactory = Symbol('factory')
 const kOptions = Symbol('options')
+const kOrigins = Symbol('origins')
 
 function defaultFactory (origin, opts) {
-  return opts && opts.connections === 1
+  return opts && (opts.connections === 1 || opts.origins === 1)
```
metcoder95: Let's keep `connections` handling the decision of wether or not use a `Pool/Client`. Overall, is possible they want a single origin with multiple connections or a single connection with multiple origins (which makes me doubtful if the `or` will ever reach the origins branch)

lib/dispatcher/pool.js
```diff
@@ -96,6 +96,12 @@ class Pool extends PoolBase {
     })
   }
 
+  [kResume] (sync) {
+    for (const client of this[kClients]) {
+      client[kResume](sync)
+    }
+  }
```
metcoder95: Why this is needed?

JoshMock: It was needed to ensure `Client[kResume]` was triggered when `Agent[kResume]` was also triggered. I've removed it now that `Agent` throws an exception instead of maintaining its own queue.

JoshMock:
@metcoder95 Agent now throws an error on dispatch when `maxOrigins` is reached, rather than managing its own queue. Please take a look!

types/agent.d.ts
metcoder95, line 1: Can you add a test for it?

JoshMock: Added in b09fb0a.

metcoder95 (approved):

JoshMock:
@metcoder95 can you reapprove the workflows? I had some failing tests so I merged main in. Most of the failures seem unrelated, and appear to be failing similarly on other PRs.

metcoder95:
H2 tests are failing due to #4156, feel free to ignore them; will try to work on it later this week

JoshMock:
@metcoder95 could you trigger the tests again? I tracked down and fixed the cause of the one failing test that wasn't related to #4156.

JoshMock:
That last test fix caused another failure somewhere else. :laughing: I believe I've got them all passing now. :pray:

JoshMock:
Merged `main` in to account for the conflict introduced by #4425, and updated the description to use the PR template.

@metcoder95 this is ready to merge when you approve the latest merge commit.

lib/core/errors.js
```diff
@@ -217,6 +217,15 @@ class SecureProxyConnectionError extends UndiciError {
   }
 }
 
+class AgentMaxOriginsReached extends UndiciError {
```
Uzlopak: rename to `AgentMaxOriginsReachedError`

Uzlopak:
I am actually now a little bit hesitant. Should we really call it AgentMaxOriginsReachedError? Or should it be just MaxOriginsReachedError?

@metcoder95 
@mcollina 

Could this be also be used in Pools or Clients? If so, we should remove the Prefix "Agent".

lib/core/errors.js
```diff
@@ -217,6 +217,15 @@ class SecureProxyConnectionError extends UndiciError {
   }
 }
 
+class AgentMaxOriginsReachedError extends UndiciError {
+  constructor (message) {
+    super(message)
+    this.name = 'AgentMaxOriginsReachedError'
+    this.message = message || 'Agent has reached the maximum allowed origins'
+    this.code = 'UND_ERR_MAX_ORIGINS'
```
Uzlopak: I guess this shows also that it is maybe a more common error?

Maybe rename it to `MaxOriginsReachedError` and the code 

```suggestion
    this.code = 'UND_ERR_MAX_ORIGINS_REACHED'
```

Uzlopak:
@JoshMock 

Dont change yet the code. Lets clarify first what the error is "doing", before you do extra work.

JoshMock:
> Lets clarify first what the error is "doing"

I'm not sure I understand what you mean. :thinking: Can you clarify?

I tend to agree with @Uzlopak that a more generic `MaxOriginsReachedError` is probably more useful, especially since it would be reusable if I add the same functionality to `BalancedPool` as I've planned. If we wanted to customize the error for each use case, the error message could be overridden rather than having a different `Error` class for each.

It's a quick change, so let me know if that's how we want to proceed.

Uzlopak:
If you consider to use it in various places, than you can generalize it of course. Please proceed ;)

JoshMock:
@Uzlopak updated in 5994344ea8a3ca6f4a36b2c8808412fd902b375c. :pray:

JoshMock:
@Uzlopak @metcoder95 Any update on when we can merge this?

metcoder95 (approved):
lgtm

mcollina (approved):
lgtm

Uzlopak (approved):
We changed the Error class thing with instanceof and symbols.

LGTM

Uzlopak:
@metcoder95 
@mcollina 

can someone of you reapprove?

metcoder95:
Failures are unrelated to the PR itself

Issue #4376, opened 2025-07-31
throwOnMaxRedirects is documented, but misnamed, and not in type definitions


williamhaley:
## Bug Description

[Documentation on `main` references](https://github.com/nodejs/undici/blob/ca1855491082cb3bacd726afbf7a8e348fc58db4/docs/docs/api/Dispatcher.md?plain=1#L965) a `throwOnMaxRedirects` field that can be set on the `RedirectInterceptorOpts` to throw an error if the maximum number of redirects are reached. I do not see `throwOnMaxRedirects` in the [undici code](https://github.com/nodejs/undici) aside from where it's mentioned in that documentation. That field [seems to not be defined on the `RedirectInterceptorOpts`](https://github.com/nodejs/undici/blob/ca1855491082cb3bacd726afbf7a8e348fc58db4/types/interceptors.d.ts#L11).

I do however see `throwOnMaxRedirect` (no trailing `s`) that is [actually used to trigger exceptions](https://github.com/nodejs/undici/blob/ca1855491082cb3bacd726afbf7a8e348fc58db4/lib/handler/redirect-handler.js#L95). Just like `throwOnMaxRedirects`, `throwOnMaxRedirect` is not defined in the `interceptor.d.ts` type declarations/definitions for the pre-defined redirect interceptor opts.

## Reproducible By

```
$ git clone https://github.com/nodejs/undici
$ grep -wrn 'throwOnMaxRedirects' ./undici
./undici/docs/docs/api/Dispatcher.md:965:  redirect({ maxRedirections: 3, throwOnMaxRedirects: true })
$ grep -wrn 'throwOnMaxRedirect' ./undici
./undici/test/interceptors/redirect.js:456:        throwOnMaxRedirect: true
./undici/test/redirect-request.js:416:        throwOnMaxRedirect: true
./undici/lib/handler/redirect-handler.js:95:    if (this.opts.throwOnMaxRedirect && this.history.length >= this.maxRedirections) {
```

## Expected Behavior

1. Documentation specifies how someone can properly trigger an exception using either `throwOnMaxRedirect` or `throwOnMaxRedirects` (not sure which one is correct, but personal preference is `throwOnMaxRedirects`

2. `RedirectInterceptorOpts` are updated to accept an optional `throwOnMaxRedirects` (or `throwOnMaxRedirect` if that's the preference) boolean for typescript users

## Logs & Screenshots

[...]

PR #4377, opened 2025-08-01
fix: update redirect handler options handling, docs, tests

williamhaley:
## This relates to...

https://github.com/nodejs/undici/issues/4376

## Rationale

I discovered what I believe are a few issues while addressing https://github.com/nodejs/undici/issues/4376:

Let me know if I'm wrong and these are working as expected.

* The documentation referenced a `throwOnMaxRedirects` option that doesn't exist
* The typescript type for `RedirectInterceptorOpts` didn't include the `throwOnMaxRedirects` option for typescript users
* The variable that is actually used for throwing an exception when the redirect limit is reached, `throwOnMaxRedirect`, doesn't seem to be documented anywhere
* `redirectionLimitReached` seems to be unused

## Rename `throwOnMaxRedirect` to `throwOnMaxRedirections`

I think there's some risk here. It's possible some users may have been using the `throwOnMaxRedirect` despite the lack of documentation. I'm not sure if that _was_ documented in the past and people could reasonably expect it to work.

The docs were updated to reflect the new name and tests were updated as well.

I went with `throwOnMaxRedirections` since the existing var `maxRedirections` aligns with that.

## Added validation to `throwOnMaxRedirections`

Validation already existed for `maxRedirections`, but not for `throwOnMaxRedirect`. Seemed worthwhile to add the validations and tests around that while renaming this.

## Remove `redirectionLimitReached`

While updating tests I found that `redirectionLimitReached` seems to be unused. I left the type declaration in place though in case callers might have used it in the past (despite it seemingly doing nothing to my eyes).

### Features

N/A

### Bug Fixes

N/A

* `redirectionLimitReached` was marked as deprecated, which I think is low-risk since it seems to be unused
* `throwOnMaxRedirect` was renamed, but also, it's not documented. I found that a Google search for `"undici" "throwOnMaxRedirect"` has almost no meaningful results so I would hope the impacted audience here is low

## Status

- [X] Tested
- [X] Documented
- [X] Review ready
- [ ] In review
- [ ] Merge ready


lib/interceptor/redirect.js
```diff
@@ -2,17 +2,17 @@
...
-      const { maxRedirections = defaultMaxRedirections, ...rest } = opts
+      const { maxRedirections = defaultMaxRedirections, throwOnMaxRedirections = defaultThrowOnMaxRedirections, ...rest } = opts
 
       if (maxRedirections == null || maxRedirections === 0) {
         return dispatch(opts, handler)
       }
 
       const dispatchOpts = { ...rest } // Stop sub dispatcher from also redirecting.
-      const redirectHandler = new RedirectHandler(dispatch, maxRedirections, dispatchOpts, handler)
+      const redirectHandler = new RedirectHandler(dispatch, maxRedirections, { throwOnMaxRedirections, ...dispatchOpts }, handler)
```
williamhaley, line 15: Maybe a little confusing that `maxRedirections` is a positional argument, but `throwOnMaxRedirections` is part of the `opts` object. That seemed to be the original intention based on existing code, as far as I could tell, so I kept with that.

test/interceptors/redirect.js
```diff
@@ -23,23 +23,23 @@ const {
...
+    new undici.Client(`http://${server}`, dispatchOpts).compose(
+      redirect({ maxRedirections: dispatchOpts?.maxRedirections, throwOnMaxRedirections: dispatchOpts?.throwOnMaxRedirections })
     )
 ]) {
-  const request = (t, server, opts, ...args) => {
-    const dispatcher = factory(server, opts)
+  const request = (t, server, dispatchOpts, url, requestOpts) => {
+    const dispatcher = factory(server, dispatchOpts)
     after(() => dispatcher.close())
-    return undici.request(args[0], { ...args[1], dispatcher }, args[2])
```
williamhaley, line 42: `args[2]` seemed to never be used. I named the spread args to try and clarify the intention. Seems like the third positional arg was the opts that are always sent to the `factory` dispatchers, the fourth positional arg was always the `url`, and the fifth positional arg was the `undici.request()` opts.

test/interceptors/redirect.js
```diff
@@ -23,23 +23,23 @@ const {
 } = undici
 
 for (const factory of [
-  (server, opts) =>
-    new undici.Agent(opts).compose(
-      redirect({ maxRedirections: opts?.maxRedirections })
+  (server, dispatchOpts) =>
+    new undici.Agent(dispatchOpts).compose(
+      redirect({ maxRedirections: dispatchOpts?.maxRedirections, throwOnMaxRedirections: dispatchOpts?.throwOnMaxRedirections })
```
williamhaley, line 28: I think there could even be another arg, `intercepterOpts`, passed to the `factory()` so that it can be used as a distinct arg from the default opts we use when instantiating each Dispatcher (Agent, Pool, Client) with its default args

types/interceptors.d.ts
```diff
@@ -8,7 +8,7 @@ export default Interceptors
 declare namespace Interceptors {
   export type DumpInterceptorOpts = { maxSize?: number }
   export type RetryInterceptorOpts = RetryHandler.RetryOptions
-  export type RedirectInterceptorOpts = { maxRedirections?: number }
+  export type RedirectInterceptorOpts = { maxRedirections?: number, throwOnMaxRedirections?: boolean }
```
williamhaley, line 11: Probably worth defining `RedirectHandler.RedirectOptions` like exists for the `RetryHandler`, but that seemed a bit daunting to me [based on all the type definitions for retry-handler.d.ts](https://github.com/nodejs/undici/blob/ca1855491082cb3bacd726afbf7a8e348fc58db4/types/retry-handler.d.ts#L37) so I just left this as a simple one-off type declaration

types/dispatcher.d.ts
```diff
@@ -135,7 +135,7 @@ declare namespace Dispatcher {
     signal?: AbortSignal | EventEmitter | null;
     /** This argument parameter is passed through to `ConnectData` */
     opaque?: TOpaque;
-    /** Default: false */
+    /** @deprecated. Default: false */
```
williamhaley: Not 100% sure if this `@deprecated` syntax is valid/meaningful 🤔 Maybe just better to remove these unused `redirectionLimitReached` options completely

types/dispatcher.d.ts
```diff
@@ -168,7 +168,7 @@ declare namespace Dispatcher {
     protocol?: string;
     /** Default: `null` */
     signal?: AbortSignal | EventEmitter | null;
-    /** Default: false */
+    /** @deprecated. Default: false */
```
metcoder95: We can delete that as it doesn't seem to be used

types/dispatcher.d.ts
```diff
@@ -145,7 +145,7 @@ declare namespace Dispatcher {
     opaque?: TOpaque;
     /** Default: `null` */
     signal?: AbortSignal | EventEmitter | null;
-    /** Default: false */
+    /** @deprecated .Default: false */
```
metcoder95: ditto

types/dispatcher.d.ts
metcoder95: Let's remove it 👍

docs/docs/api/Dispatcher.md
```diff
@@ -962,7 +962,7 @@ const { Client, interceptors } = require("undici");
 const { redirect } = interceptors;
 
 const client = new Client("http://example.com").compose(
-  redirect({ maxRedirections: 3, throwOnMaxRedirects: true })
+  redirect({ maxRedirections: 3, throwOnMaxRedirections: true })
```
mcollina, line 965: Please keep both name around but document only the new one

types/dispatcher.d.ts
williamhaley: Sounds good! Removed in b74f1a1a572fd9d56512c77b32c15cd793f38220

docs/docs/api/Dispatcher.md
williamhaley: Thanks @mcollina. That makes sense to me. Restored `throwOnMaxRedirect` and its tests in 971cc6eb

lib/interceptor/redirect.js
```diff
@@ -2,17 +2,17 @@
 
 const RedirectHandler = require('../handler/redirect-handler')
 
-function createRedirectInterceptor ({ maxRedirections: defaultMaxRedirections } = {}) {
+function createRedirectInterceptor ({ maxRedirections: defaultMaxRedirections, throwOnMaxRedirections: defaultThrowOnMaxRedirections } = {}) {
```
mcollina, line 5: This would also need to handle throwOnMaxRedirects.

williamhaley: Do you mean `throwOnMaxRedirect` (no `s`)? `throwOnMaxRedirects` was [in the documentation](https://github.com/nodejs/undici/blob/431dfdc693df0a15f2d5a5a3933e421c8459a74d/docs/docs/api/Dispatcher.md?plain=1#L965), but it [wasn't used in the functional code](https://github.com/nodejs/undici/blob/431dfdc693df0a15f2d5a5a3933e421c8459a74d/lib/handler/redirect-handler.js#L95) on `main` so I don't think that value could have worked. `throwOnMaxRedirect` was used in the functional code though.

 I can add `throwOnMaxRedirect` here since that used to work in the functional code. My rationale for not including it here originally was that `throwOnMaxRedirect` was never documented so I wasn't sure if it was worth advertising it now in Typescript or better to just consolidate on `throwOnMaxRedirections`.  `throwOnMaxRedirections` is now in the `Dispatcher.md` docs as well as the functional code and the naming aligns with `maxRedirections`.

metcoder95 (approved):

test/interceptors/redirect.js
Ethan-Arrowood: i like this idea, but likely should go in a separate PR 👍

Ethan-Arrowood (approved):
lgtm

williamhaley:
I'm happy to catch this work up from upstream, but still waiting on an approval to handle the `requested changes` on the PR.

Uzlopak:
@williamhaley 
what happened?

williamhaley:
@Uzlopak I was waiting on final approval for several weeks. There were requested changes and I made some updates, but this PR was still stuck due to the requested changes.

This went stale in the meantime so I decided to close it. I'll re-open and catch-up from the latest if I can get feedback on https://github.com/nodejs/undici/pull/4377#discussion_r2248385811. I think that issue may have been the blocker for my last approval

lib/interceptor/redirect.js
samuel871211: @williamhaley 

Current status, as I understand it:

1. `throwOnMaxRedirects` has now been cleaned up in ( #4924 )
2. `redirectionLimitReached` has now been cleaned up in ( #4933 )
3. `throwOnMaxRedirect` appears to be the option introduced by ( #2563 ), but it still seems under-documented and its behavior/supported surface is not very clear.

With those pieces out of the way, I think this PR may be worth reopening.

PR #4378, opened 2025-08-01
Actually flush the file in lib/mock/snapshot-recorder.js

mcollina:
Wrong method signature :(. I autocompleted too much.


mcollina:
Landing without approval to fix CI.

PR #4384, opened 2025-08-03
feat: add install() function to type definitions

jsaguet:
## Rationale

The install function is not available in undici types so when trying to use it in a TS project, we have to add @ts-ignore or similar to be able to use it.

## Changes

Adds install function to d.ts

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/test/types/index.test-d.ts
+++ b/test/types/index.test-d.ts
@@ -1,5 +1,18 @@
-import { expectAssignable } from 'tsd'
-import Undici, { Pool, Client, errors, fetch, Interceptable, RedirectHandler, Headers, Response, Request, FormData, SnapshotAgent } from '../..'
+import { expectAssignable, expectType } from 'tsd'
+import Undici, {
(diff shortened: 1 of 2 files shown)


Uzlopak:
Do we really need a separate install.d.ts for a trivial function?

mcollina (changes requested):
Can you plese integrate this this in `types/index.d.ts` directly? Could you also add a test for this type?

jsaguet:
I implemented the changes.
Since I'm now in the same file, I had to declare the function with a different name to be able to add it to the Undici namespace as `install`. Let me know if you know another way to achieve this.

Ethan-Arrowood (approved):

mcollina (approved):
lgtm

gurgunday (approved):
lgtm!

PR #4390, opened 2025-08-06
fix: correctly handle multi-value headers in fetch when `rawHeaders` contains array

### LINKED ISSUE #4389: Set-Cookie header gets concatenated in fetch if an interceptor is in place
## Bug Description

If the dispatcher contains an interceptor, even a dummy one, the `Set-Cookie` headers will be concatenated together.

## Reproducible By

```js
const http = require('node:http')
const {Agent} = require('undici');

const dispatcher = new Agent().compose((dispatch) => dispatch)

    res.writeHead(200, [['set-cookie', 'a=1'], ['set-cookie', 'b=2'], ['set-cookie', 'c=3']])
    res.end("ok")
})
server.listen({port: 3002}, () => {
    Promise.all([fetch('http://localhost:3002'), fetch('http://localhost:3002', {dispatcher})])
        .then(([r1, r2]) => {
            console.log(r1.headers.getSetCookie())
            console.log(r2.headers.getSetCookie())
        })
        .then(() => server.close())
});

```

The output will be:
```
[ 'a=1', 'b=2', 'c=3' ]
[ 'a=1,b=2,c=3' ]
```

## Expected Behavior
The two outputs should be identical.

## Environment
Nodejs v24.4.1

hexchain:
## This relates to...

#4389 

## Changes

In `httpNetworkFetch->dispatch->onHeaders`, append multi-value header values one by one and let `HeadersList` handle concatenation.

### Features

N/A

### Bug Fixes

#4389 

N/A

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/web/fetch/index.js
+++ b/lib/web/fetch/index.js
@@ -2110,7 +2110,15 @@ async function httpNetworkFetch (
           const headersList = new HeadersList()
 
           for (let i = 0; i < rawHeaders.length; i += 2) {
-            headersList.append(bufferToLowerCasedHeaderName(rawHeaders[i]), rawHeaders[i + 1].toString('latin1'), true)
+            const nameStr = bufferToLowerCasedHeaderName(rawHeaders[i])
+            const value = rawHeaders[i + 1]
+            if (Array.isArray(value) && !Buffer.isBuffer(rawHeaders[i + 1])) {
+              for (const val of value) {
+                headersList.append(nameStr, val.toString('latin1'), true)
+              }
+            } else {
+              headersList.append(nameStr, value.toString('latin1'), true)
+            }
           }
           const contentEncoding = headersList.get('content-encoding', true)
           if (contentEncoding) {
--- a/test/issue-4389.js
+++ b/test/issue-4389.js
@@ -1,0 +1,45 @@
+const { describe, test } = require('node:test')
+const assert = require('node:assert')
+const { once } = require('node:events')
+const { createServer: createHttpServer } = require('node:http')
+const { createServer: createSocketServer } = require('node:net')
+const { fetch, Agent } = require('..')
+
+describe('Interceptor should not break multi-value headers', () => {
+  test('Interceptor should not break Set-Cookie', async (t) => {
+    const server = createHttpServer({ joinDuplicateHeaders: true }, async (_req, res) => {
+      const headers = [
+        ['set-cookie', 'a=1'],
+        ['set-cookie', 'b=2'],
+        ['set-cookie', 'c=3']
+      ]
+      res.writeHead(200, headers)
+      res.end()
+    }).listen(0)
+
+    await once(server, 'listening')
+    t.after(() => server.close())
+
+    const dispatcher = new Agent().compose((dispatch) => dispatch)
+
+    const { headers } = await fetch(`http://localhost:${server.address().port}`, { dispatcher })
+    assert.deepStrictEqual(headers.getSetCookie(), ['a=1', 'b=2', 'c=3'])
+  })
+
+  test('Interceptor should not break other multi-value header', async (t) => {
+    const server = createSocketServer((socket) => {
+      socket.write('HTTP/1.0 204 No Content\r\n')
+      socket.write('X-Test-Header: 1\r\n')
+      socket.write('X-Test-Header: 2\r\n')
+      socket.write('X-Test-Header: 3\r\n')
+      socket.end('\r\n\r\n')
+    }).listen(0)
+    t.after(() => server.close())
+    await once(server, 'listening')
+
+    const dispatcher = new Agent().compose((dispatch) => dispatch)
+
+    const { headers } = await fetch(`http://localhost:${server.address().port}`, { dispatcher })
+    assert.deepStrictEqual(headers.get('x-test-header'), '1, 2, 3')
+  })
+})


lib/handler/wrap-handler.js
```diff
@@ -62,7 +62,13 @@ module.exports = class WrapHandler {
   onResponseStart (controller, statusCode, headers, statusMessage) {
     const rawHeaders = []
     for (const [key, val] of Object.entries(headers)) {
-      rawHeaders.push(Buffer.from(key), Array.isArray(val) ? val.map(v => Buffer.from(v)) : Buffer.from(val))
+      if (Array.isArray(val)) {
```
Uzlopak: i dont think so. You have to check if the key is set-cookie. See the correspongin code in nodejs incomingmessage

hexchain: I've changed this to check for `set-cookie` but don't think it's sufficient.

According to https://github.com/nodejs/node/blob/0309806a74f8596059061525ba56fd9c82c56a8f/lib/_http_incoming.js#L276 there are many other headers without the "no duplicate" flag, which means they can have more than one instance. Should these also be checked? That list does not seem to be exhaustive either.

Also, this change here is supposed to undo the effect of `parseHeaders` in `UnwrapHandler::onHeaders`, which doesn't check if the header can be duplicated at all.

Another fix would be to change [`httpNetworkFetch->dispatch->onHeaders`](https://github.com/nodejs/undici/blob/c3f559154b061d73e9ce4be7e540ae69c8f26fcb/lib/web/fetch/index.js#L2113) to support such case where the header value is a list.

Uzlopak: I actually took this PR to investigate the behavior in node core. No duplicate means only to keep the first seen header value and drop following headers with same fieldname. 

Node core and undici use llhttp, so basically we stream the http request/response, so we have to concat values and stuff. 

Your claims are interesting. Maybe we have some blind spot in our logic? Definetely needs further investigation.

But when I saw your initial commit, set-cookie came immediatly in my mind... That header is always s... special.... ;)

lib/handler/wrap-handler.js
```diff
@@ -62,7 +62,13 @@ module.exports = class WrapHandler {
   onResponseStart (controller, statusCode, headers, statusMessage) {
     const rawHeaders = []
     for (const [key, val] of Object.entries(headers)) {
-      rawHeaders.push(Buffer.from(key), Array.isArray(val) ? val.map(v => Buffer.from(v)) : Buffer.from(val))
+      if (Array.isArray(val) && key === 'set-cookie') {
```
metcoder95: I've the impression we can make this general and possibly implement a lever that disables the behavior (similar to [`joinDuplicateHeaders`](https://nodejs.org/api/http.html#httpcreateserveroptions-requestlistener)).

I'm over the fence if we should do it by default as it kinda imposes a breaking change, but maybe it doesn't.

lib/handler/wrap-handler.js
hexchain: Oops, I might have read the comments wrongly. Sorry.

In the current Node core, only `Set-Cookie` is allowed to be an array (indeed the most special one), others will simply be joined with commas (or semicolons if it's `Cookie`) or dropped if duplicates are not allowed, and thus won't have multiple instances.

I guess the better way of handling this would be to leave them as arrays here, and join them in `fetch` then. That would make its behavior closer to a browser: in Blink, it stores fetch headers in a multimap and only join the values with commas when getting one[1].

[1] https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/fetch/fetch_header_list.cc;drc=29830f68ddfceee2b180d8795583b2d7dd58a949;l=82

hexchain:
I've pushed a different fix as discussed in https://github.com/nodejs/undici/pull/4390#discussion_r2259309353. Please take a look, thanks!

lib/web/fetch/index.js
```diff
@@ -2110,7 +2110,15 @@ async function httpNetworkFetch (
           const headersList = new HeadersList()
 
           for (let i = 0; i < rawHeaders.length; i += 2) {
-            headersList.append(bufferToLowerCasedHeaderName(rawHeaders[i]), rawHeaders[i + 1].toString('latin1'), true)
+            const nameStr = bufferToLowerCasedHeaderName(rawHeaders[i])
```
metcoder95, line 2113: As this only breaks given an interceptor, I'd prefer that we fix it within `undici`'s rather than fetch

metcoder95 (commented):
code wise, lgtm; but defer to @KhafraDev

lib/web/fetch/index.js
hexchain: My idea is that fetch should properly support arrays of header values in `rawHeaders`, if such case is valid.

lib/handler/wrap-handler.js
KhafraDev: > it stores fetch headers in a multimap and only join the values with commas when getting one[1].

It is more efficient not to do this, so we don't

hexchain: Yes, but in undici the `HeadersList` joins when appending, so it won't be a problem.

hexchain: I'm going to resolve this conversation and the one below to avoid confusion, as they target outdated changes.

lib/web/fetch/index.js
metcoder95: Sure, but I'd prefer to cross check with the spec to see what it has to say regarding duplicated headers, if the implementation already aligns, most likely is an `undici` issue.

hexchain: IIUC, [RFC 7230 Section 3.2.2](https://datatracker.ietf.org/doc/html/rfc7230#section-3.2.2) means that:
- A sender must not send multiple headers with the same name unless it is `Set-Cookie`,
- A receiver may combine values of multiple headers with the same name into a comma-separated list, except for `Set-Cookie`.

I'd say it is fine to have them separated in Undici/interceptor APIs, and only join them in fetch.

hexchain:
Gentle ping. May I have some reviews again?

The latest change only touches `fetch` and not (Un)wrapHandlers, so no changes to existing interceptor/handler behaviors. Duplicate headers are processed mostly in the same way as before, except when interceptors exist, in which case it is more correct than before - it uses `HeadersList.append` now instead of `Array<Buffer>.toString`.

lib/web/fetch/index.js
Uzlopak: You refer to RFC7230 but not the fetch spec

hexchain: It seems to me that [HTTP-network fetch](https://fetch.spec.whatwg.org/#http-network-fetch) doesn't really specify how headers should be treated while being received? It merely says:

> Follow the relevant requirements from HTTP. [[HTTP]](https://fetch.spec.whatwg.org/#biblio-http) [[HTTP-CACHING]](https://fetch.spec.whatwg.org/#biblio-http-caching)

> Wait until all the HTTP response headers are transmitted.

in which "[HTTP]" refers to [RFC9110](https://httpwg.org/specs/rfc9110.html). In that document, section 5.2 and 5.3 are basically a rephrase of RFC7230 section 3.2.2.

In `undici`, the relevant code is in `httpNetworkFetch->dispatch->onHeaders`, and I believe is an implementation detail. IIUC, changes to this part of the code should count as fixing the issue in undici.

mcollina (approved):
lgtm

metcoder95 (commented):
I'd still suggest to alter the interceptor rather than fetch itself

hexchain:
@metcoder95 By "interceptor" do you mean Wrap/UnwrapHandler? There is (almost) no interceptor involved in the original reproducer.

metcoder95:
Yes, exactly; sorry for the confusion, was referring to the wrapper

mcollina:
@metcoder95 can you do an alternative PR with that fix? I'm not sure it'd be possible to actually do that.

hexchain:
I believe that was also the very first iteration of this PR.

I'll try to summarize the situation and hopefully uncover a clear path forward.

The issue happens because the original dispatch handler (which is "old"-style, according to #3878) gets wrapped and unwrapped, so the headers argument of `onHeaders`/`onResponseStart` changed, from a `Buffer[]` to a `Record<string, string | string[]>` and then (incorrectly) to an `Array<Buffer | Buffer[]>`, which ultimately breaks the dispatch handler in `fetch`. So far, I see two approaches:

1. The first iteration: fix the header conversion in the "wrapping" process: expand header values of `string[]` back to multiple key-value `Buffer` pairs - basically undo `UnwrapHandler::onHeaders`.
  This was the first iteration, however in [this review][ref-review1] I was asked to add a check to see whether the header allows multi-value before expanding. It feels a bit awkward to do that because:
   - The same kind of check also happens somewhere in `node:http` and (to some extent) `HeadersList` but neither is reusable.
   - `WrapHandler::onResponseStart` isn't the best place to modify/sanitize the data, when it is only supposed to restore the headers to what it looked like before "unwrapping". Even if ensuring the header being spec-compliant is a general concern, it should still happen somewhere else, either in the dispatcher or higher-level APIs.
2. Current: fix the dispatch handler in `fetch` to also accept `Array<Buffer | Buffer[]>`, in addition to `Buffer[]`. For now, the type of `headers` in the signature of (the deprecated) `onHeaders` is still `Buffer[]`, so this also feels like a hack.

Please enlighten me if there are better ways. Personally, I think the first idea is more sound but I'm really unsure about adding the header name check.

[ref-review1]: https://github.com/nodejs/undici/pull/4390#pullrequestreview-3091933012

PR #4392, opened 2025-08-07
fix: shell command built from environment values 

ptrgits:
fix the problem, we should avoid constructing shell commands as strings with interpolated variables and instead use the `execFileSync` function, which takes the command and its arguments as separate parameters. This ensures that arguments are passed directly to the executable and are not interpreted by the shell, preventing issues with spaces or special characters in paths.

Specifically, for the line:
```js
execSync(`${WASM_OPT} ${WASM_OPT_FLAGS} --enable-simd -o ${join(WASM_OUT, 'llhttp_simd.wasm')} ${join(WASM_OUT, 'llhttp_simd.wasm')}`, { stdio: 'inherit' })
```
We should:
- Parse `WASM_OPT_FLAGS` into an array of arguments (if it is a string).
- Build an array of arguments: `[...WASM_OPT_FLAGS, '--enable-simd', '-o', join(WASM_OUT, 'llhttp_simd.wasm'), join(WASM_OUT, 'llhttp_simd.wasm')]`
- Use `execFileSync(WASM_OPT, args, { stdio: 'inherit' })` instead of `execSync`.

We may need to add a helper function to split flags safely (e.g., using a simple split on spaces, or a more robust shell-words parser if available, but for this fix, a simple split is acceptable given the context).

## Status

- [x] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [x] Merge ready


mcollina:
Great find! how did you spot this?

metcoder95 (commented):
lgtm, just linting

gurgunday (approved):

Ethan-Arrowood (approved):
lgtm but needs update to pass linter

Uzlopak:
@nodejs/undici PTAL

Uzlopak:
@metcoder95 
@Ethan-Arrowood 
@mcollina 

Can someone of you approve it? I cant merge it as I fixed the linting.

metcoder95 (approved):

PR #4400, opened 2025-08-09
ci: fix node24 ci

Uzlopak:


gurgunday (approved):

himself65 (approved):

himself65:
Fixes: https://github.com/nodejs/undici/issues/4156

himself65 (commented):
Should we probably uninstall `https-pem`?

mcollina:
I prefer generating them, as including the certs will generate additional work here.

Uzlopak:
lets get the ci green again

richardlau:
> 
> @richardlau is there a key in node core, that we could use?

Node.js has https://github.com/nodejs/node/tree/main/test/fixtures/keys which are generated by https://github.com/nodejs/node/blob/main/test/fixtures/keys/Makefile. 
There were updates to some of the keys there for OpenSSL 3.2, https://github.com/nodejs/node/pull/54599.

PR #4408, opened 2025-08-12
cache: change normaliseHeaders to normalizeHeaders

Uzlopak:
Its more like a chore Pr. It feels unnatural, that we used here BE. Everywhere else we use AE


KhafraDev:
these files come from https://cache-tests.fyi/

Uzlopak:
i changed only the code in lib now

PR #4409, opened 2025-08-12
sri: better algorithms for sri

Uzlopak:
I think we should deviate from the spec, as we can implement better and faster code than the algorithms defined in the spec. @annevk states, that:

["Generally we don't write algorithms in specifications with specific memory or performance targets in mind. The main goal is clarity. You can read more about this at https://infra.spec.whatwg.org/#algorithm-conformance."
](https://github.com/w3c/webappsec-subresource-integrity/pull/151)

The bytesMatch function is basically calculating the actualValue over and over. I addressed it here: https://github.com/w3c/webappsec-subresource-integrity/pull/153 

getStrongestMetadata can reuse the provided array. There is no point to create a new array and assign the items to it. Also there is no point in getting the index of the currentAlgorithm again and again. We just need to store the algorithmIndex in the strongest variable.

Also @annevk refers to [infra conformance ](https://infra.spec.whatwg.org/#algorithm-conformance)

> Conformance requirements phrased as algorithms or specific steps may be implemented in any manner, so long as the end result is equivalent. (In particular, the algorithms are intended to be easy to follow, and not intended to be performant.)

So we just need to ensure, that the end result is equivalent to the end result of the spec. 

If you agree on this, I would improve the comments in this PR, so that we have a better implementation of the sri

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/web/sri/sri.js
+++ b/lib/web/sri/sri.js
@@ -1,7 +1,5 @@
 'use strict'
 
-const assert = require('node:assert')
-
 /**
  * @typedef {object} Metadata
  * @property {SRIHashAlgorithm} alg - The algorithm used for the hash.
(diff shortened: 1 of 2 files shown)


KhafraDev:
> So we just need to ensure, that the end result is equivalent to the end result of the spec.

No one is questioning our ability to implement the spec in our own manner, but even that wasn't done (see commented out test).

> "Generally we don't write algorithms in specifications with specific memory or performance targets in mind. The main goal is clarity."

Yes, this is why we follow the spec. It makes it easier to maintain in the future, especially when other implementations deviate from it (like Chrome, with base64url) since we can float similar patches. 

Since this is opt-in (only runs when passing an integrity option to RequestInit), conformance with the spec and other browsers is paramount to removing an array allocation or replacing a length check.

Uzlopak:
Yeah you are right. I hope the spec people will change some of these slow performance issues.

PR #4410, opened 2025-08-12
test: fix snapshot testing flaky test

Uzlopak:
@mcollina 

Was there a reason to make _replaySnapshot async? Because it was asyncy it resulted in the flaky behaviour. Basically returning the snapshot was async so it was not ensured which response Body needed to be returned.


mcollina (approved):
lgtm

PR #4411, opened 2025-08-13
fix formdata constructor args mark optional

tsctx:
The [FormData](https://xhr.spec.whatwg.org/#dom-formdata) constructor's arguments are all marked as optional in the specification, so let's make them optional.

### DIFF
--- a/test/fetch/formdata.js
+++ b/test/fetch/formdata.js
@@ -341,7 +341,7 @@ test('FormData should be compatible with third-party libraries', (t) => {
 })
 
 test('arguments', () => {
-  assert.strictEqual(FormData.prototype.constructor.length, 0)
+  assert.strictEqual(FormData.length, 0)
   assert.strictEqual(FormData.prototype.append.length, 2)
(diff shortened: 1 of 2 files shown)


test/fetch/formdata.js
```diff
@@ -341,7 +341,7 @@ test('FormData should be compatible with third-party libraries', (t) => {
 })
 
 test('arguments', () => {
-  assert.strictEqual(FormData.constructor.length, 1)
+  assert.strictEqual(FormData.prototype.constructor.length, 0)
```
tsctx:
```
Welcome to Node.js v24.5.0.
> `${FormData.constructor}`
'function Function() { [native code] }'
```

tsctx:
I'm not sure what "other implementations" you're referring to, but browsers (Chrome, Firefox) also have a length of 0, just like this one.

test/fetch/formdata.js
KhafraDev:
```suggestion
  assert.strictEqual(FormData.length, 0)
```

PR #4413, opened 2025-08-15
refactor: snapshot-recorder

Uzlopak:
I used jsdoc to make a static code analysis. I found some redundant calculations etc.. "Simplified" it.

@mcollina


Uzlopak:
I think I maybe fixed that flaky test regarding auto flush.

PR #3458, opened 2024-08-14
use FinalizationRegistry for cloned response body

KhafraDev:
running test with `node --expose-gc --test test/fetch/fire-and-forget.js`

<details>
<summary>Before</summary>

```
RSS 66 MB after 50 fetch() requests
RSS 68 MB after 100 fetch() requests
RSS 76 MB after 150 fetch() requests
RSS 84 MB after 200 fetch() requests
RSS 93 MB after 250 fetch() requests
RSS 101 MB after 300 fetch() requests
RSS 110 MB after 350 fetch() requests
RSS 117 MB after 400 fetch() requests
RSS 125 MB after 450 fetch() requests
RSS 133 MB after 500 fetch() requests
RSS 141 MB after 550 fetch() requests
RSS 134 MB after 600 fetch() requests
RSS 143 MB after 650 fetch() requests
RSS 149 MB after 700 fetch() requests
RSS 155 MB after 750 fetch() requests
RSS 163 MB after 800 fetch() requests
RSS 171 MB after 850 fetch() requests
RSS 178 MB after 900 fetch() requests
RSS 185 MB after 950 fetch() requests
RSS 183 MB after 1000 fetch() requests
RSS 196 MB after 1050 fetch() requests
RSS 205 MB after 1100 fetch() requests
RSS 212 MB after 1150 fetch() requests
RSS 219 MB after 1200 fetch() requests
RSS 226 MB after 1250 fetch() requests
RSS 233 MB after 1300 fetch() requests
RSS 241 MB after 1350 fetch() requests
RSS 249 MB after 1400 fetch() requests
RSS 257 MB after 1450 fetch() requests
RSS 264 MB after 1500 fetch() requests
RSS 240 MB after 1550 fetch() requests
RSS 244 MB after 1600 fetch() requests
RSS 248 MB after 1650 fetch() requests
RSS 257 MB after 1700 fetch() requests
RSS 263 MB after 1750 fetch() requests
RSS 269 MB after 1800 fetch() requests
RSS 275 MB after 1850 fetch() requests
RSS 282 MB after 1900 fetch() requests
RSS 290 MB after 1950 fetch() requests
RSS 299 MB after 2000 fetch() requests
RSS 307 MB after 2050 fetch() requests
RSS 284 MB after 2100 fetch() requests
RSS 287 MB after 2150 fetch() requests
RSS 294 MB after 2200 fetch() requests
RSS 300 MB after 2250 fetch() requests
RSS 305 MB after 2300 fetch() requests
[...]

PR #4415, opened 2025-08-15
fix: fix wrong stream canceled up after cloning (v7)

snyamathi:
Pretty much the same as https://github.com/nodejs/undici/pull/4414 but for v7

## This relates to...

- https://github.com/nodejs/undici/pull/3458
- https://github.com/nodejs/undici/issues/4150

## Rationale

The request/stream registered with the finalization registry is mixed up between the clone and the original request.

## Changes

This PR just changes which of the two tee'd streams is associated

### Features

N/A 

### Bug Fixes

N/A 

N/A 

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/web/fetch/response.js
+++ b/lib/web/fetch/response.js
@@ -354,7 +354,7 @@ function cloneResponse (response) {
   if (response.body != null) {
     newResponse.body = cloneBody(response.body)
 
-    streamRegistry.register(newResponse, new WeakRef(response.body.stream))
+    streamRegistry.register(newResponse, new WeakRef(newResponse.body.stream))
   }
 
   // 4. Return newResponse.
--- a/test/fetch/response.js
+++ b/test/fetch/response.js
@@ -297,3 +297,30 @@ test('clone body garbage collection', async () => {
   const cloneBody = ref.deref()
   assert.equal(cloneBody, undefined, 'clone body was not garbage collected')
 })
+
+test('clone body garbage collection should not affect original res', async () => {
+  let ref
+
+  const doFetch = async () => {
+    const res = new Response('hello world')
+
+    // Clone res and consume its body
+    const clone = res.clone()
+    await clone.text()
+    ref = new WeakRef(clone)
+
+    return res
+  }
+
+  const res = await doFetch()
+
+  // wait for garbage collection
+  while (true) {
+    await setImmediate()
+    global.gc()
+    if (!ref.deref()) break
+  }
+
+  await setImmediate()
+  assert.doesNotThrow(() => res.clone(), 'Body has already been consumed')
+})


lib/web/fetch/body.js
```diff
@@ -29,12 +29,17 @@ try {
...
-const streamRegistry = new FinalizationRegistry((weakRef) => {
-  const stream = weakRef.deref()
-  if (stream && !stream.locked && !isDisturbed(stream) && !isErrored(stream)) {
-    stream.cancel('Response object has been garbage collected').catch(noop)
-  }
-})
+const hasFinalizationRegistry = globalThis.FinalizationRegistry
+let streamRegistry
+
+if (hasFinalizationRegistry) {
```
Uzlopak: Why? We support only node 20+. So we always have FinalizationRegistry. Is this some kind of artifact from undici 6?

Uzlopak (changes requested):
seems wrong.

lib/web/fetch/response.js
```diff
@@ -352,9 +352,7 @@ function cloneResponse (response) {
   // 3. If response’s body is non-null, then set newResponse’s body to the
   // result of cloning response’s body.
   if (response.body != null) {
-    newResponse.body = cloneBody(response.body)
-
-    streamRegistry.register(newResponse, new WeakRef(response.body.stream))
```
Uzlopak, line 357: I guess, it needed here to be?

`streamRegistry.register(newResponse, new WeakRef(newResponse.body.stream))`

lib/web/fetch/body.js
snyamathi: I simply reverted https://github.com/nodejs/undici/commit/c5e74641f40ee12b9f0b101235e792f476a62c20 so this is what was in the codebase prior to https://github.com/nodejs/undici/pull/4320

I can make any changes as needed, but thought that it would be easier to review this way

snyamathi:
## This is all out of date - ignore this comment

~@Uzlopak this is re-adding the same code from https://github.com/nodejs/undici/pull/3458 which you had reviewed.~

~There were some modifications before it was reverted in https://github.com/nodejs/undici/pull/4320 to address the issue https://github.com/nodejs/undici/issues/4150~

~I think it is still of value to clean up the response body if a request is cloned, especially since the docs say:~

> ~The [Fetch Standard](https://fetch.spec.whatwg.org/) allows users to skip consuming the response body by relying on [garbage collection](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Memory_Management#garbage_collection) to release connection resources.~

~Although they do go on to say~

> ~it is important to always either consume or cancel the response body anyway.~

~This being a clone vs a fetch makes it unclear if the body will be garbage collected or not, and some frameworks like Next.JS clone every request by default to provide deduping of feches from a single inbound request~

lib/web/fetch/body.js
```diff
@@ -301,13 +306,17 @@ function safelyExtractBody (object, keepalive = false) {
...
   // To clone a body body, run these steps:
 
   // https://fetch.spec.whatwg.org/#concept-body-clone
 
   // 1. Let « out1, out2 » be the result of teeing body’s stream.
-  const { 0: out1, 1: out2 } = body.stream.tee()
+  const [out1, out2] = body.stream.tee()
+
+  if (hasFinalizationRegistry) {
+    streamRegistry.register(instance, new WeakRef(out2))
```
snyamathi: The rest of the code is just a reversion of a prior commit - this is the only change here where we fix the `out1` vs `out2` mixup. I've also added a unit test to assert this behavior

lib/web/fetch/response.js
snyamathi: Yes, that's correct - I've updated the PR to drop all the reversion business and just make this one change + add a test

test/fetch/response.js
```diff
@@ -297,3 +297,30 @@ test('clone body garbage collection', async () => {
...
+  // wait for garbage collection
+  while (true) {
+    await setImmediate()
+    global.gc()
+    if (!ref.deref()) break
+  }
+
+  await setImmediate()
+  assert.doesNotThrow(() => res.clone(), 'Body has already been consumed')
+})
```
snyamathi, line 326: This fails on `main`

```
✖ clone body garbage collection should not affect original res (6.698336ms)
  AssertionError [ERR_ASSERTION]: Got unwanted exception: Body has already been consumed
```

test/fetch/response.js
```diff
@@ -297,3 +297,30 @@ test('clone body garbage collection', async () => {
...
+
+    // Clone res and consume its body
+    const clone = res.clone()
+    await clone.text()
+    ref = new WeakRef(clone)
+
+    return res
+  }
+
+  const res = await doFetch()
```
snyamathi, line 315: `res` has never had its body read, but ends up being canceled in the finalization registry b/c we register the wrong stream in `lib/web/fetch/response.js`

test/fetch/response.js
```diff
@@ -297,3 +297,30 @@ test('clone body garbage collection', async () => {
...
+
+  // wait for garbage collection
+  while (true) {
+    await setImmediate()
+    global.gc()
+    if (!ref.deref()) break
+  }
```
snyamathi, line 322: When the clone is garbage collected, the original res body stream should not be canceled

mcollina:
Why did you revert https://github.com/nodejs/undici/pull/4320?

snyamathi:
Sorry, I ended up not needing to revert it. I'll update the PR description, but I've squashed that change out.

The only change now is updating which tee'd stream is put into the finalization registry

snyamathi:
https://github.com/snyamathi/nextjs-mem-leak/blob/main/index.mjs Here's a minimal reproduction of the issue. I ended up finding this while tracing down a memory leak in NextJS

```js
/**
 * This is a minimal reproduction of an issue in
 * https://github.com/nodejs/undici/pull/3458
 *
 * When cloning a response, the wrong stream is
 * registered with undici's finalization registry
 * causing the original stream to be reclaimed, not
 * the cloned stream.
 */

import { setImmediate } from "node:timers/promises";

const { promise, resolve } = Promise.withResolvers();
const registry = new FinalizationRegistry(resolve);

const doFetch = async () => {
  const res = await fetch(
    "https://microsoftedge.github.io/Demos/json-dummy-data/512KB-min.json"
  );

  // Clone res and consume its body
  const clone = res.clone();
  await clone.json();

  // Use finalization registry to know when the *clone* has been reclaimed
  registry.register(clone, {});

  return res;
};

const res = await doFetch();

global.gc();
await promise;
await setImmediate();

// TypeError: Response.clone: Body has already been consumed.
res.clone();
```

tsctx:
You've caused a regression of the memory leak issue in the cloned response.

```
> node --expose-gc ./test/fetch/fire-and-forget.js
✔ test finalizer cloned request (103.7647ms)
RSS 81 MB after 50 fetch() requests
RSS 84 MB after 100 fetch() requests
...
RSS 536 MB after 4900 fetch() requests
RSS 537 MB after 4950 fetch() requests
RSS 541 MB after 5000 fetch() requests
✔ does not need the body to be consumed to continue (10846.5894ms)
ℹ tests 2
ℹ suites 0
ℹ pass 2
ℹ fail 0
ℹ cancelled 0
ℹ skipped 0
ℹ todo 0
ℹ duration_ms 10973.2636
```

Uzlopak:
I actually wonder what actually happens when you tee. Maybe the issue is, that we expect that both streams get consumed? And our expectations are wrong?

mcollina:
> Maybe the issue is, that we expect that both streams get consumed?

No, each stream could be consumed independently.

Uzlopak:
@mcollina 
Sure, both streams could be consumed independently. But if it isnt, how do we know that they are done and can be gc'ed?

mcollina:
The finalizationregistry/weakref business should make sure of it.

tsctx:
I wrote a better solution in #4419 's PR.

Uzlopak:
I merged the other PR. Please verify and if you like please backport for v6

snyamathi:
@Uzlopak  I've updated https://github.com/nodejs/undici/pull/4414 to backport the changes for v6

PR #4425, opened 2025-08-18
fix: memory leak in Agent

### LINKED ISSUE #4424: Memory leak in Agent
## Bug Description

There is still a memory leak in `Agent`.

## Reproducible By

```javascript
const { Agent, fetch, setGlobalDispatcher } = require('undici');

const agent = new Agent({
  keepAliveMaxTimeout: 1000,
});
setGlobalDispatcher(agent);

globalThis.agent = agent;

async function main() {
  const rs = await Promise.all([
    fetch('http://www.example.com/123'),
    fetch('http://example.com/234'),
    fetch('https://www.example.com/345'),
    fetch('https://example.com/456'),
  ]);

  console.log('done')
  setTimeout(() => undefined, 1000000)
}

void main();
```

Run the script with `--inspect`. After 'done', inspect `agent[kClients]`. There will be 4 items in the map.

## Expected Behavior

All clients should have been destroyed.

## Logs & Screenshots

Image

## Environment

Undici 7.14.0, Nodejs 22.16.0.

### Additional context

In `Agent::[kDispatch]` the `key` of `this[kClients]` is a string, but in `Agent::[kOnConnect]` and `Agent::[kOnDisconnect]` the `origin` is a URL object. Even if it is coerced into a string, there will be a trailing slash which makes it different from the original key.

hexchain:
## This relates to...

#4424

## Changes

Instead of using `origin` from the `connect` and `disconnect` events, track client using the original key.

### Features

N/A

### Bug Fixes

#4424

N/A

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/dispatcher/agent.js
+++ b/lib/dispatcher/agent.js
@@ -45,27 +45,14 @@ class Agent extends DispatcherBase {
     }
 
     this[kOnConnect] = (origin, targets) => {
-      const result = this[kClients].get(origin)
-      if (result) {
-        result.count += 1
-      }
       this.emit('connect', origin, [this, ...targets])
     }
 
     this[kOnDisconnect] = (origin, targets, err) => {
-      const result = this[kClients].get(origin)
-      if (result) {
-        result.count -= 1
-        if (result.count <= 0) {
-          this[kClients].delete(origin)
-          result.dispatcher.destroy()
-        }
-      }
       this.emit('disconnect', origin, [this, ...targets], err)
     }
 
     this[kOnConnectionError] = (origin, targets, err) => {
-      // TODO: should this decrement result.count here?
       this.emit('connectionError', origin, [this, ...targets], err)
     }
   }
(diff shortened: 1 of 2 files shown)


mcollina:
Can you add a unit test?

hexchain:
Added a unit test for the referenced issue.

lib/dispatcher/agent.js
```diff
@@ -91,7 +79,23 @@ class Agent extends DispatcherBase {
     if (!dispatcher) {
       dispatcher = this[kFactory](opts.origin, this[kOptions])
         .on('drain', this[kOnDrain])
+        .on('connect', () => {
```
metcoder95: Why not reuse the same event handler?

hexchain: There was no particular reason. I've updated to just register one event handler, and call the original handler in the closure.

mcollina:
There are relevant regressions

hexchain:
I've pushed a new version to fix the tests. Some other changes worth noting:

- In `Agent`, `close` is now used instead of `destroy` after untracking a client
- `connectionError` events are also handled now

mcollina (approved):
lgtm on green CI

Issue #4430, opened 2025-08-20
v6.x Backporting


joyeecheung:
For backporting https://github.com/nodejs/node/pull/57165 to Node.js 22, we need to backport https://github.com/nodejs/undici/pull/4064 to undici v6.x

It seems to be planned to backport to v6.21.3 in https://github.com/nodejs/undici/pull/4211 but I took a look at https://github.com/nodejs/undici/compare/v6.21.2...v6.21.3 and it doesn't seem to be there?

Commits needed:

- https://github.com/nodejs/undici/commit/a217002369c403f5a84a870520294435632ef6eb
- https://github.com/nodejs/undici/commit/95fd9d3feebcdede11a4e6b66f57aac34abf482f
- https://github.com/nodejs/undici/commit/b7513d4df62e9d1c1ecc34c3a418bd402e3c8432

cc @mcollina

metcoder95:
Let me see if they can be automated, otherwise will try to backport them manually

Uzlopak:
@metcoder95 

Are you done with backporting?

metcoder95:
Yes, done already

joyeecheung:
Thanks for the backports! Is it ready for another release?

metcoder95:
cc: @mcollina

curiosity26:
Hi guys. Any update on this issue? I know LTS 24 is coming out next month, but we are hoping that this backport happens quickly as we have a customer who needs it. Thanks

mcollina:
Shipping now.

metcoder95:
Can we close this issue now?

PR #4433, opened 2025-08-21
feat(ProxyAgent): match Curl behavior in HTTP->HTTP Proxy connections (#4180)

### LINKED ISSUE #4083: ProxyAgent has different behavior from curl -x

metcoder95:
Backports #4180  - Relates to #4430

### DIFF
--- a/docs/docs/api/ProxyAgent.md
+++ b/docs/docs/api/ProxyAgent.md
@@ -22,8 +22,9 @@ For detailed information on the parsing process and potential validation errors,
 * **token** `string` (optional) - It can be passed by a string of token for authentication.
 * **auth** `string` (**deprecated**) - Use token.
 * **clientFactory** `(origin: URL, opts: Object) => Dispatcher` (optional) - Default: `(origin, opts) => new Pool(origin, opts)`
-* **requestTls** `BuildOptions` (optional) - Options object passed when creating the underlying socket via the connector builder for the request. See [TLS](https://nodejs.org/api/tls.html#tlsconnectoptions-callback).
-* **proxyTls** `BuildOptions` (optional) - Options object passed when creating the underlying socket via the connector builder for the proxy server. See [TLS](https://nodejs.org/api/tls.html#tlsconnectoptions-callback).
+* **requestTls** `BuildOptions` (optional) - Options object passed when creating the underlying socket via the connector builder for the request. It extends from [`Client#ConnectOptions`](/docs/docs/api/Client.md#parameter-connectoptions).
+* **proxyTls** `BuildOptions` (optional) - Options object passed when creating the underlying socket via the connector builder for the proxy server. It extends from [`Client#ConnectOptions`](/docs/docs/api/Client.md#parameter-connectoptions).
+* **proxyTunnel** `boolean` (optional) - By default, ProxyAgent will request that the Proxy facilitate a tunnel between the endpoint and the agent. Setting `proxyTunnel` to false avoids issuing a CONNECT extension, and includes the endpoint domain and path in each request.
 
 Examples:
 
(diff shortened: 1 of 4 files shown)


Uzlopak (approved):
LGTM

PR #4439, opened 2025-08-21
fix (websocket): ErrorEvent constructor should convert arguments before passing to `super()`

Uzlopak:
Now the order of operations is comparable to CloseEvent and MessageEvent

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready


KhafraDev (changes requested):
This doesn't fix anything. If it does fix something, we need a test.

PR #4443, opened 2025-08-21
fix: catch synchronous errors in request callbacks

mcollina:
## Summary

- Fix synchronous errors thrown in request callbacks being swallowed when underlying stream is destroyed
- Add proper error handling in RequestHandler.onHeaders() to catch, cleanup, and re-throw errors
- Ensure errors are propagated through existing error handling chain via Request.onHeaders()

## Rationale

Previously, when a synchronous error was thrown in a request callback after the underlying stream was destroyed, the error would be swallowed and become an uncaught exception. This happened because the `RequestHandler.onHeaders()` method called user callbacks via `runInAsyncScope()` without wrapping it in a try-catch block.

### Bug Fixes

- Added try-catch block around callback invocation in `RequestHandler.onHeaders()`
- Properly destroy response stream when callback throws synchronously  
- Re-throw error to be handled by existing error handling in `Request.onHeaders()`
- Added comprehensive test coverage for synchronous error scenarios

### Implementation Details

The fix ensures that:
1. Synchronous errors in callbacks are caught immediately
2. Response streams are properly cleaned up before propagating errors
3. Errors flow through the existing error handling chain (`Request.onHeaders` → `this.abort(err)`)
4. No uncaught exceptions occur due to swallowed callback errors

## Test Plan

- [x] Added tests for synchronous errors in successful request callbacks
- [x] Added tests for immediate errors thrown in callbacks  
- [x] Added tests for error handling in failure scenarios
- [x] Verified existing tests continue to pass
- [x] Confirmed no regressions in error handling behavior

### DIFF
--- a/lib/api/api-request.js
+++ b/lib/api/api-request.js
@@ -118,14 +118,23 @@ class RequestHandler extends AsyncResource {
     this.callback = null
     this.res = res
     if (callback !== null) {
-      this.runInAsyncScope(callback, null, null, {
-        statusCode,
-        headers,
-        trailers: this.trailers,
-        opaque,
-        body: res,
-        context
-      })
+      try {
+        this.runInAsyncScope(callback, null, null, {
+          statusCode,
+          headers,
+          trailers: this.trailers,
+          opaque,
+          body: res,
+          context
+        })
+      } catch (err) {
+        // If the callback throws synchronously, we need to handle it
+        // Destroy the response stream and propagate the error immediately
+        this.res = null
+        util.destroy(res.on('error', noop), err)
+        // Re-throw the error so it gets caught in Request.onHeaders which will call abort
+        throw err
+      }
     }
   }
 
--- a/test/sync-error-in-callback.js
+++ b/test/sync-error-in-callback.js
@@ -1,0 +1,95 @@
+'use strict'
+
+const { tspl } = require('@matteo.collina/tspl')
+const { test, after } = require('node:test')
+const { Client } = require('..')
+const { createServer } = require('node:http')
+
+test('synchronous error in request callback should be caught', async (t) => {
+  const p = tspl(t, { plan: 2 })
+
+  const server = createServer((req, res) => {
+    res.end('hello')
+  })
+  after(() => server.close())
+
+  server.listen(0, () => {
+    const client = new Client(`http://localhost:${server.address().port}`)
+    after(() => client.close())
+
+    const testError = new Error('sync error in callback')
+
+    client.request({
+      path: '/',
+      method: 'GET'
+    }, (err, data) => {
+      p.ifError(err)
+      p.strictEqual(data.statusCode, 200)
+
+      // Destroy the stream to simulate the described scenario
+      data.body.destroy()
+
+      // This synchronous error should be caught and not become an uncaught exception
+      throw testError
+    })
+  })
+
+  await p.completed
+})
+
+test('synchronous error thrown immediately in request callback', async (t) => {
+  const p = tspl(t, { plan: 2 })
+
+  const server = createServer((req, res) => {
+    res.end('hello')
+  })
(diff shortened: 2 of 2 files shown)


Uzlopak:
@mcollina 

Do we need to check all occurences of runInAsyncScope in regard of being wrapped in try catch blocks?

mcollina:
That might be the case.

lib/api/api-request.js
```diff
@@ -118,14 +118,23 @@ class RequestHandler extends AsyncResource {
...
+          body: res,
+          context
+        })
+      } catch (err) {
+        // If the callback throws synchronously, we need to handle it
+        // Destroy the response stream and propagate the error immediately
+        this.res = null
+        util.destroy(res.on('error', noop), err)
+        // Re-throw the error so it gets caught in Request.onHeaders which will call abort
+        throw err
```
Uzlopak: too much condensed for my taste and not easy to grok. I did not test following code, but it is this what you want to achieve (see comments)

```suggestion
        // If the callback throws synchronously, we need to handle it
        
        // Remove reference to res to allow res being garbage collected
        this.res = null
        
        // handle error-events with a noop to avoid thrown errors
        // by the stream logic
        res.on('error', noop)

        // Destroy the response stream and propagate the error immediately
        util.destroy(res, err)

        // Re-throw the error so it gets caught in Request.onHeaders which will call abort
        throw err
```

mcollina:
I think this has a bug still, I need to look at it again.

lib/api/api-request.js
```diff
@@ -118,14 +118,31 @@ class RequestHandler extends AsyncResource {
...
+          statusCode,
+          headers,
+          trailers: this.trailers,
+          opaque,
+          body: res,
+          context
+        })
+      } catch (err) {
+        // If the callback throws synchronously, we need to handle it
+
```
Uzlopak:
```suggestion

```

Uzlopak:
```suggestion

```

lib/api/api-request.js
```diff
@@ -118,14 +118,31 @@ class RequestHandler extends AsyncResource {
...
+          opaque,
+          body: res,
+          context
+        })
+      } catch (err) {
+        // If the callback throws synchronously, we need to handle it
+        
+        // Remove reference to res to allow res being garbage collected
+        this.res = null
+
```
Uzlopak:
```suggestion

```

Uzlopak:
Fixed the linting

lib/api/api-request.js
```diff
@@ -118,14 +118,31 @@ class RequestHandler extends AsyncResource {
...
+        this.res = null
+        // handle error-events with a noop to avoid thrown errors
+        // by the stream logic
+        res.on('error', noop)
+
+        // Destroy the response stream
+        util.destroy(res, err)
+
+        // Use process.nextTick to re-throw the error so it reaches uncaughtException
+        process.nextTick(() => {
```
Uzlopak:
```suggestion
        // Use queueMicrotask to re-throw the error so it reaches uncaughtException
        queueMicrotask(() => {
```

Uzlopak: We use no where in the code process.nextTick. I guess we should use here queueMicrotask?!

Uzlopak: But to behonest,  i dont understand why we want an uncaught exception

mcollina: we don't want exceptions to be swallowed, otherwise they are impossible to track. This is default behavior of Node.js.

Uzlopak:
@mcollina 
Oh you mean, it was totally "swallowed" no even throwing an unhandledPromiseRejection?

mcollina:
Yes

Uzlopak (approved):
LGTM

PR #4449, opened 2025-08-22
fetch: make consumeBody sync

Uzlopak:
It reduces the amount of Promises created by a fetch call by 2.


lib/web/fetch/body.js
```diff
@@ -443,24 +437,30 @@ function mixinBody (prototype, getInternalState) {
...
+    return Promise.reject(e)
+  }
 
   const state = getInternalState(object)
 
   // 1. If object is unusable, then return a promise rejected
   //    with a TypeError.
   if (bodyUnusable(state)) {
-    throw new TypeError('Body is unusable: Body has already been read')
+    return Promise.reject(new TypeError('Body is unusable: Body has already been read'))
```
Uzlopak, line 452: THis makes actually alot of sense, because it is returning a promise rejected.

lib/web/fetch/body.js
```diff
@@ -443,24 +437,30 @@ function mixinBody (prototype, getInternalState) {
...
  * @param {(target: any) => any} getInternalState
  */
-async function consumeBody (object, convertBytesToJSValue, instance, getInternalState) {
-  webidl.brandCheck(object, instance)
+function consumeBody (object, convertBytesToJSValue, instance, getInternalState) {
+  try {
+    webidl.brandCheck(object, instance)
+  } catch (e) {
+    return Promise.reject(e)
+  }
```
Uzlopak, line 445: this on the other hand is kind of stupid

PR #4454, opened 2025-08-24
fix(h2): adjust :scheme on h2 requests

metcoder95:
## Rationale

Adjust the pseudo header `:scheme` in accordance to the actual protocol used (support for h2c client).

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -141,7 +141,7 @@ class Request {
 
     this.path = query ? serializePathWithQuery(path, query) : path
 
-    // TODO: shall we maybe standardise it to an URL object?
+    // TODO: shall we maybe standardize it to an URL object?
     this.origin = origin
 
     this.protocol = new URL(origin).protocol
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -17,7 +17,8 @@ const {
   serializePathWithQuery,
   assertRequestHandler,
   getServerName,
-  normalizedMethodRecords
+  normalizedMethodRecords,
+  getProtocolFromUrlString
 } = require('./util')
 const { channels } = require('./diagnostics.js')
 const { headerNameLowerCasedRecord } = require('./constants')
@@ -144,7 +145,7 @@ class Request {
     // TODO: shall we maybe standardize it to an URL object?
     this.origin = origin
 
-    this.protocol = new URL(origin).protocol
+    this.protocol = getProtocolFromUrlString(origin)
 
     this.idempotent = idempotent == null
       ? method === 'HEAD' || method === 'GET'
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -141,7 +141,10 @@ class Request {
 
     this.path = query ? serializePathWithQuery(path, query) : path
 
+    // TODO: shall we maybe standardise it to an URL object?
     this.origin = origin
+
+    this.protocol = new URL(origin).protocol
 
     this.idempotent = idempotent == null
       ? method === 'HEAD' || method === 'GET'
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -279,7 +279,7 @@ function shouldSendContentLength (method) {
 function writeH2 (client, request) {
   const requestTimeout = request.bodyTimeout ?? client[kBodyTimeout]
   const session = client[kHTTP2Session]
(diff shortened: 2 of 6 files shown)


lib/dispatcher/client.js
```diff
@@ -296,8 +296,7 @@ class Client extends DispatcherBase {
   }
 
   [kDispatch] (opts, handler) {
-    const origin = opts.origin || this[kUrl].origin
```
metcoder95, line 299: I'm not sure why did we have this set here if clients maps to single origin (single socket); it was causing to request have different origin than the actual client creating the request.

Tests does not seems to show a problem removing it.

lib/core/request.js
```diff
@@ -141,8 +141,11 @@ class Request {
 
     this.path = query ? serializePathWithQuery(path, query) : path
 
+    // TODO: shall we maybe standardise it to an URL object?
```
Uzlopak: nit:

```suggestion
    // TODO: shall we maybe standardize it to an URL object?
```

lib/dispatcher/client-h2.js
```diff
@@ -387,7 +387,7 @@ function writeH2 (client, request) {
   // :path and :scheme headers must be omitted when sending CONNECT
 
   headers[HTTP2_HEADER_PATH] = path
-  headers[HTTP2_HEADER_SCHEME] = 'https'
+  headers[HTTP2_HEADER_SCHEME] = protocol.slice(0, protocol.length - 1)
```
Uzlopak: can it be http and https only?

```suggestion
  headers[HTTP2_HEADER_SCHEME] = procotol === 'http:' ? 'http' : 'https'
```

lib/core/request.js
```diff
@@ -141,8 +141,11 @@ class Request {
...
     this.origin = origin
 
+    this.protocol = new URL(origin).protocol
```
Uzlopak: Why tho? Either the origin starts with 'https:' or with 'http://'. We can add a util function to `/lib/core/util.js` next to `isHttpOrHttpsPrefixed`

```js
function getProtocolFromUrlString (urlString) {
  if (
    urlString[0] === 'h' &&
    urlString[1] === 't' &&
    urlString[2] === 't' &&
    urlString[3] === 'p'
    ) {
      switch (urlString[4]) {
        case ':':
          return 'http:'
        case 's':
          if (urlString[5] === ':') {
            return 'https:'
          }
      }
   }
  // fallback if none of the usual suspects
  return urlString.slice(0, urlString.indexOf(':') + 1)
} 
```

lib/dispatcher/client.js
Uzlopak: what does git blame say?

lib/core/request.js
metcoder95: Not strong opinion on that, was just the easiest way to do that; but we can adjust it

lib/dispatcher/client.js
metcoder95: Haven checked, let me verify

lib/dispatcher/client-h2.js
```diff
@@ -387,7 +387,7 @@ function writeH2 (client, request) {
   // :path and :scheme headers must be omitted when sending CONNECT
 
   headers[HTTP2_HEADER_PATH] = path
-  headers[HTTP2_HEADER_SCHEME] = 'https'
+  headers[HTTP2_HEADER_SCHEME] = procotol === 'http:' ? 'http' : 'https'
```
metcoder95:
```suggestion
  headers[HTTP2_HEADER_SCHEME] = protocol === 'http:' ? 'http' : 'https'
```

lib/dispatcher/client.js
metcoder95: It should be good to go, was implemented here #1053 as safe guard. We can get rid of it

Uzlopak (approved):
We can take care of performance issues some other time.

lib/core/request.js
mcollina: Using new URL() is going to significantly regress performance.

mcollina (changes requested):
no new URL in the hot path. It got incredibly faster, but it's still a very complex algorithm.

Issue #4391, opened 2025-08-07
Add support for resolving wildcard subdomains *.localhost


yonran:
## This would solve...

In popular browsers, subdomains of localhost (e.g. `http://myserver.localhost`) all resolve to loopback [::1, 127.0.0.1]. This is also also specified in the fetch spec.

* Chrome: 2015-03 http://crrev.com/322452 (linked from [comment](https://issues.chromium.org/issues/41175806#comment6))
* Firefox: 2020-01 https://bugzilla.mozilla.org/show_bug.cgi?id=1220810
* WebKit: not implemented yet https://bugs.webkit.org/show_bug.cgi?id=160504

References:
* https://fetch.spec.whatwg.org/#resolve-an-origin: “If origin’s [host](https://html.spec.whatwg.org/multipage/browsers.html#concept-origin-host)’s [public suffix](https://url.spec.whatwg.org/#host-public-suffix) is "localhost" or "localhost.", then return « ::1, 127.0.0.1 ».”
* https://datatracker.ietf.org/doc/html/rfc6761#section-6.3 (2013): “ The domain "localhost." and any names falling within ".localhost." are special”… “Name resolution APIs and libraries SHOULD recognize localhost names as special and SHOULD always return the IP loopback address for address queries”

## The implementation should look like...

Extract the public suffix, compare it to `localhost` and `localhost.` before calling dns.lookup.

## Additional context

This could also be a workaround for a bug resolving non-wildcard localhost itself #1602, since glibc `getaddrinfo` mistakenly filters out the localhost loopback ipv6 address when you pass in the `ADDRCONFIG` flag (https://sourceware.org/bugzilla/show_bug.cgi?id=14969).

metcoder95:
cc: @tsctx @KhafraDev

mcollina:
This can easily be added as interceptor.

Personally, I think this is better implemented in Node.js itself given the linked RFC.

KhafraDev:
I think there is something to be fixed in fetch as well, since we skip over https://fetch.spec.whatwg.org/#resolve-an-origin and https://fetch.spec.whatwg.org/#concept-connection-obtain

FelixVaughan:
Took a stab at the first part of this, suggestions and feedback welcome

PR #4456, opened 2025-08-24
Resolve .localhost to loopback per Fetch spec

FelixVaughan:
## This relates to...
#4391

## Rationale
Align localhost public-suffix resolution with the Fetch spec (Resolving domains -> “resolve an origin”): any `localhost` or `*.localhost` host resolves to loopback.  
Spec: [Fetch 2.5 resolve an origin](https://fetch.spec.whatwg.org/#resolve-an-origin)

## Changes/Fixes
- Add `isLocalhostPublicSuffix(hostname)`
- Inject custom [`lookup`](https://nodejs.org/api/dns.html#dnslookuphostname-options-callback) (`localhostLookup`) for both net and TLS paths to short‑circuit DNS for `.localhost`
- No change for non‑`localhost` hosts or literal IPs

## Status
- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/core/connect.js
+++ b/lib/core/connect.js
@@ -60,6 +60,7 @@ function isLocalhostPublicSuffix (hostname) {
   return h === 'localhost' || h.endsWith('.localhost')
 }
 
+// https://nodejs.org/download/release/v9.10.1/docs/api/dns.html#dns_dns_lookup_hostname_options_callback
 // Custom DNS lookup that resolves any *.localhost name to loopback addresses
 // Returns both ::1 and 127.0.0.1 when opts.all is true, otherwise
 // respects requested family, defaulting to IPv4 for maximal compatibility.
@@ -71,10 +72,12 @@ function localhostLookup (_host, opts, cb) {
 
   const all = !!(opts && opts.all)
   const family = opts && opts.family
+  const ipv4 = '127.0.0.1'
+  const ipv6 = '::1'
 
   const results = [
-    { address: '::1', family: 6 },
-    { address: '127.0.0.1', family: 4 }
+    { address: ipv6, family: 6 },
+    { address: ipv4, family: 4 }
   ]
 
   if (all) {
@@ -83,16 +86,16 @@ function localhostLookup (_host, opts, cb) {
   }
 
   if (family === 6) {
-    cb(null, '::1', 6)
+    cb(null, ipv6, 6)
     return
   }
   if (family === 4) {
-    cb(null, '127.0.0.1', 4)
+    cb(null, ipv4, 4)
     return
   }
 
   // Default to IPv4 to avoid environments without IPv6
-  cb(null, '127.0.0.1', 4)
+  cb(null, ipv4, 4)
 }
 
 function buildConnector ({ allowH2, maxCachedSessions, socketPath, timeout, session: customSession, ...opts }) {
--- a/lib/core/connect.js
+++ b/lib/core/connect.js
@@ -60,7 +60,7 @@ function isLocalhostPublicSuffix (hostname) {
   return h === 'localhost' || h.endsWith('.localhost')
 }
 
-// https://nodejs.org/download/release/v9.10.1/docs/api/dns.html#dns_dns_lookup_hostname_options_callback
+// https://nodejs.org/download/release/v24.6.0/docs/api/dns.html#dns_dns_lookup_hostname_options_callback
 // Custom DNS lookup that resolves any *.localhost name to loopback addresses
 // Returns both ::1 and 127.0.0.1 when opts.all is true, otherwise
 // respects requested family, defaulting to IPv4 for maximal compatibility.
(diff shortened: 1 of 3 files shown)


lib/core/connect.js
```diff
@@ -43,6 +43,61 @@ const SessionCache = class WeakSessionCache {
   }
 }
 
+// Implements the relevant part of "resolve an origin" for localhost public suffix
+// https://fetch.spec.whatwg.org/#resolve-an-origin
```
Uzlopak, line 47: This is fetch specific code and should not be in core

lib/core/connect.js
```diff
@@ -43,6 +43,61 @@ const SessionCache = class WeakSessionCache {
...
+  }
+
+  // Normalize trailing dot and casing
+  let h = hostname.toLowerCase()
+  if (h[h.length - 1] === '.') {
+    h = h.slice(0, -1)
+  }
+
+  // Exact localhost or any subdomain of .localhost
+  return h === 'localhost' || h.endsWith('.localhost')
```
Uzlopak, line 60: I am sure this can be more performant implemented

lib/core/connect.js
```diff
@@ -43,6 +43,61 @@ const SessionCache = class WeakSessionCache {
...
+    cb = opts
+    opts = {}
+  }
+
+  const all = !!(opts && opts.all)
+  const family = opts && opts.family
+  const ipv4 = '127.0.0.1'
+  const ipv6 = '::1'
+
+  const results = [
```
Uzlopak, line 78: Why do you instantiate this array on every call?

lib/core/connect.js
```diff
@@ -43,6 +43,61 @@ const SessionCache = class WeakSessionCache {
   }
 }
 
+// Implements the relevant part of "resolve an origin" for localhost public suffix
```
Uzlopak, line 46: Only the relevant part? What parts are messing?

lib/core/connect.js
```diff
@@ -77,7 +132,14 @@ function buildConnector ({ allowH2, maxCachedSessions, socketPath, timeout, sess
         socket: httpSocket, // upgrade socket connection
         port,
         host: hostname
-      })
+      }
+
+      // Spec: If public suffix is localhost, resolve to loopback
+      if (isLocalhostPublicSuffix(hostname)) {
+        tlsOptions.lookup = localhostLookup
```
Uzlopak, line 139: Have to check what this lookup attribute is doing

Uzlopak (changes requested):

Uzlopak:
@KhafraDev

lib/core/connect.js
FelixVaughan: I'll find a new home for it. Any suggestions?

lib/core/connect.js
FelixVaughan: I'll see what I can do

lib/core/connect.js
FelixVaughan: In the specs there are 4 different cases for origin resolution; only one has to do with localhost

> If origin’s [host](https://html.spec.whatwg.org/multipage/browsers.html#concept-origin-host)’s [public suffix](https://url.spec.whatwg.org/#host-public-suffix) is "localhost" or "localhost.", then return « ::1, 127.0.0.1 ».

Thats the one this addresses

lib/core/connect.js
FelixVaughan: Its purpose is to resolve a hostname. By default it performs a dns lookup but here we override it to short-circuit the lookup. Here's a [link](https://nodejs.org/api/dns.html#dnslookuphostname-options-callback) to the docs

lib/core/connect.js
FelixVaughan: I could hoist out of the function but I think that would sacrifice readability for minor optimizations

FelixVaughan:
> This isn't "per Fetch spec". The implementation is noticeably different.

@KhafraDev could point me in the right direction?

KhafraDev:
There is nothing I can say here that wasn't already said in the issue.

3 possibilities:
- Implement as an interceptor
- Implement in node core
- For fetch only, update `httpNetworkFetch` to match the spec (easier said than done)

This also has other issues like overriding tlsoptions.lookup which iirc can be set by users and following the fetch spec for some reason. You link [this](https://fetch.spec.whatwg.org/#resolve-an-origin) but none of the steps are actually implemented, and even if they were, undici hardly seems like the correct place to implement this. As mentioned in the issue, given the RFC, it seems like something that should be fixed in node core (it seems like `cares` does it already?).

FelixVaughan:
Good to know I wasn't on the right path. I agree that ideally the main fix should be in core and I could open an issue there but I wanted to see if there's anything patchable here. I would be interested in hearing more about the `httpNetworkFetch` approach as well

Also not sure what `cares` means in this context

Uzlopak:
https://github.com/c-ares/c-ares

KhafraDev:
https://github.com/search?q=repo%3Anodejs%2Fnode%20rfc6761&type=code

PR #4472, opened 2025-08-26
feat: make UndiciErrors reliable to instanceof

### LINKED ISSUE #3987: errors from node:internal leak
## Bug Description

When importing `undici` (from node_modules) after using `fetch` (nodejs built-in), then `globalDispatcher` is filled by built-in undici.
Therefore using `instanceof errors.UndiciError` is unreliable

## Reproducible By

```
Promise.resolve().then(async () => {
  fetch('http://httpstat.us/200') // if fetch is commented out, then it works properly

  require('net').createServer((c) => {
    setTimeout(() => c.destroy(), 100)
  }).listen(8888)

  const { request, errors } = require('undici')
  try {
    await request('http://127.0.0.1:8888/')
  } catch (err) {
    console.info(err instanceof errors.UndiciError)
  }
  process.exit(0)
})
```

## Expected Behavior

`true` would be printed, as error should be instance of `SocketError` that inherits `UndiciError`

## Logs & Screenshots

Image

## Environment

Checked on Node v20.13.1 and v23.5.0 using undici 7.2.0 and 5.28.4, on ubuntu 22.04 and in node:20.13-alpine3.19 container

## Context

It seems natural to use `instanceof`, because error classes are exported. `npm` allows to have multiple versions of `undici` exported, so it's easy to have built-in undici, `node_modules/undici` and undici nested inside node_modules libraries.

To make it worse: ResponseStatusCodeError (which is easiest to test) is thrown outside dispatcher, so when I checked it, `instanceof` worked properly.

Uzlopak:
fixes #3987

This is basically the same approach we use in @fastify/errors

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/core/errors.js
+++ b/lib/core/errors.js
@@ -1,13 +1,21 @@
 'use strict'
 
+const kUndiciError = Symbol.for('undici.error.UND_ERR')
 class UndiciError extends Error {
   constructor (message, options) {
     super(message, options)
     this.name = 'UndiciError'
     this.code = 'UND_ERR'
   }
+
+  static [Symbol.hasInstance] (instance) {
+    return instance && instance[kUndiciError] === true
+  }
+
+  [kUndiciError] = true
 }
 
+const kConnectTimeoutError = Symbol.for('undici.error.UND_ERR_CONNECT_TIMEOUT')
 class ConnectTimeoutError extends UndiciError {
   constructor (message) {
     super(message)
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

mcollina:
this should be backporrted to v6

metcoder95:
@Uzlopak can you try to backport it?

Uzlopak:
i will try to backport it. Currently doing some other stuff locally. ;)

PR #4478, opened 2025-08-27
lint: avoid unintented use of globals  in code and tests, improve test for installing/overwriting globals

Uzlopak:
This PR is using the eslint rule `no-restricted-globals` so that we avoid using accidently globals, while we want to use or test our own implementation. Also rewrote the test for the `install()` function.

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/test/env-http-proxy-agent-nodejs-bundle.js
+++ b/test/env-http-proxy-agent-nodejs-bundle.js
@@ -20,7 +20,7 @@ describe('EnvHttpProxyAgent and setGlobalDispatcher', () => {
     process.env = { ...env }
   })
 
-  test('should work together from the Node.js bundle', async (t) => {
+  test('should work global fetch from undici bundled with Node.js', async (t) => {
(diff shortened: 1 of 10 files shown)


test/env-http-proxy-agent-nodejs-bundle.js
```diff
@@ -20,7 +20,7 @@ describe('EnvHttpProxyAgent and setGlobalDispatcher', () => {
     process.env = { ...env }
   })
 
-  test('should work together from the Node.js bundle', async (t) => {
+  test('should work global fetch from undici bundled with Node.js', async (t) => {
```
Uzlopak:
```suggestion
  test('should work with global fetch from undici bundled with Node.js', async (t) => {
```

metcoder95 (approved):

PR #4491, opened 2025-08-31
fix: deflake cache-fastimers-fix.js

Uzlopak:
This fixes the flakyness of test/interceptors/cache-fastimers-fix.js and renames the filename to a more descriptive name cache-revalidate-statel.js. Also "fastimers" is a clear typo.

The flakyness was due to the fact, that the http-server by nodejs is sending the date header, which is not overridable by modifying the global Date and/or Timer Logic. I guess node core also caches the date and refreshes it every second and not on every http request?! That resulted in the flakyness, we saw.

I basically deactivated sending the date header via the http server and send my own mocked date header t. Thus making it fully deterministic.

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/test/interceptors/cache-revalidate-stale.js
+++ b/test/interceptors/cache-revalidate-stale.js
@@ -4,12 +4,19 @@ const { test, after } = require('node:test')
 const { strictEqual } = require('node:assert')
 const { createServer } = require('node:http')
 const { once } = require('node:events')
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

PR #4492, opened 2025-08-31
fix: implement proper stale-while-revalidate behavior per RFC 5861

### LINKED ISSUE #4471: Cache: Stale-while-revalidate not fully working as intended by RFC5861
## Description

The main benefit of using stale-while-revalidate is that we can issue a cached but stale response while revalidating in the background. This decreases latency at the trade-off of serving potentially outdated data.

Currently, stale-while-revalidate is not implemented this way. Undici cache does not return the cached value during the stale-while-revalidate window immediately, but revalidates synchronously. The only benefit stale-while-revalidate provides at this moment is the reduction in data retrieval and processing when the endpoint replies with a 304 _not modified_.

Although this is acceptable according to [RFC 5861 section 3](https://datatracker.ietf.org/doc/html/rfc5861#section-3), it is quite unoptimal and, in my opinion, unexpected behavior.

## Reproducible By

I've made a small [repo](https://github.com/daan944/undici-stale-while-revalidate) to test this, see test.ts within.

## Expected Behavior

Upon requesting a cached yet stale entry within stale-while-revalidate window, Undici should:
- return the cached yet stale entry immediately
- re-validate the stale entry (and update cache accordingly) in the background
See [RFC 5861 section 3](https://datatracker.ietf.org/doc/html/rfc5861#section-3)

## Environment

Node 24.6.0, Undici 7.15.0, OSX (although I don't think that applies).

### Additional context

I've tried to update [the cache interceptor](https://github.com/daan944/undici/blob/main/lib/interceptor/cache.js#L240) with a bit of code to check whether we're in the stale-while-revalidate window and return the cached result immediately. That seems to work partially: it does return the stale item immediately, it does perform a revalidate request, but it does **not** update the cache if the server responds with new data. I might continue on this later.

I've filed this as a bug, as, in my opinion, this is unexpected behavior.

mcollina:
## Summary

This PR fixes the `stale-while-revalidate` cache directive implementation to comply with RFC 5861. Previously, it was performing synchronous revalidation, which defeated the primary purpose of reducing latency.

Fixes #4471

## Changes

- **Added proper stale-while-revalidate detection**: New `withinStaleWhileRevalidateWindow()` helper function determines if a stale response is within the revalidate window
- **Implemented background revalidation**: When within the window, the cache now:
  - Returns stale content immediately (reducing latency)
  - Triggers background revalidation asynchronously using `setImmediate()`
  - Updates the cache silently when new content is received
- **Added comprehensive tests**: Two new tests verify RFC 5861 compliance
- **Updated existing tests**: Modified tests that expected the old (incorrect) synchronous behavior

## Test Plan

- [x] Added test: `stale-while-revalidate returns stale immediately and revalidates in background (RFC 5861)`
- [x] Added test: `stale-while-revalidate updates cache after background revalidation`
- [x] Updated existing tests to expect correct behavior
- [x] All cache interceptor tests pass
- [x] Linting passes

### Before
When a cached response was stale but within the `stale-while-revalidate` window:
- Client waits for revalidation request to complete
- Response time = network latency to origin server
- Defeats the purpose of stale-while-revalidate

### After  
When a cached response is stale but within the `stale-while-revalidate` window:
- Client receives stale content immediately (< 10ms)
- Background revalidation updates cache for future requests
- Achieves the intended latency reduction benefit

## RFC Compliance

This implementation follows [RFC 5861 Section 3](https://datatracker.ietf.org/doc/html/rfc5861#section-3) which states:
> When present in an HTTP response, the stale-while-revalidate Cache-Control extension indicates that caches MAY serve the response in which it appears after it becomes stale, up to the indicated number of seconds... while they revalidate it in the background.

### DIFF
--- a/lib/interceptor/cache.js
+++ b/lib/interceptor/cache.js
@@ -254,7 +254,7 @@ function handleResult (
       sendCachedValue(handler, opts, result, age, null, true)
 
       // Start background revalidation (fire-and-forget)
-      process.nextTick(() => {
+      queueMicrotask(() => {
         let headers = {
           ...opts.headers,
           'if-modified-since': new Date(result.cachedAt).toUTCString()


lib/interceptor/cache.js
```diff
@@ -231,6 +247,51 @@ function handleResult (
...
     }
 
+    // RFC 5861: If we're within stale-while-revalidate window, serve stale immediately
+    // and revalidate in background
+    if (withinStaleWhileRevalidateWindow(result)) {
+      // Serve stale response immediately
+      sendCachedValue(handler, opts, result, age, null, true)
+
+      // Start background revalidation (fire-and-forget)
+      process.nextTick(() => {
```
Uzlopak:
```suggestion
      queueMicrotask(() => {
```

ShogunPanda (approved):

metcoder95 (approved):
lgtm

Uzlopak (approved):
RSLGTM

Issue #4444, opened 2025-08-22
DNS interceptor breaks headers of type `[string, string][]`


peterhirn:
## Bug Description

Unable to use headers of type `[string, string][]`, using `Record<string, string>` works (see below).

## Reproducible By

```js
import http from "node:http";
import { Agent, interceptors, request } from "undici";

const port = process.env.PORT ?? 3001;

const { cache, dns } = interceptors;

const good = new Agent().compose(cache());

const fail = new Agent().compose(cache(), dns());

http
  .createServer((req, res) => {
    console.log(req.headers);
    res.end();
  })
  .listen(port, async () => {
    const headers = [["foo", "bar"]];

    await request(`http://localhost:${port}`, {
      dispatcher: good,
      headers,
    });

    await request(`http://localhost:${port}`, {
      dispatcher: fail,
      headers,
    });

    const headersAsRecord = { foo: "bar" };
    await request(`http://localhost:${port}`, {
      dispatcher: fail,
      headers: headersAsRecord,
    });
  });
```

```
node index.js
{ host: 'localhost:3001', connection: 'keep-alive', foo: 'bar' }
{ '0': 'foo, bar', host: 'localhost:3001', connection: 'keep-alive' }
{ host: 'localhost:3001', connection: 'keep-alive', foo: 'bar' }
```

peterhirn:
The problem seems to be here: https://github.com/nodejs/undici/blob/main/lib/interceptor/dns.js#L416

Maybe I should switch all my code to use [IncomingHttpHeaders](https://github.com/nodejs/undici/blob/main/types/header.d.ts#L6) (which still allows multiple values)

```ts
export type IncomingHttpHeaders = Record<string, string | string[] | undefined>
```

Which type do you recommend for headers?

Uzlopak:
I agree with you.. there should be proper handling of headers. I think we have some functions which takes care of headers being processed properly. we just need to use them in the interceptor.

PandaWorker:
> I agree with you.. there should be proper handling of headers. I think we have some functions which takes care of headers being processed properly. we just need to use them in the interceptor.

Or just normalize the headers into one normal view and use this structure everywhere after the request, in each interceptor to check the type and format of the headers so-so

[https://github.com/nodejs/undici/issues/4336](url)

Uzlopak:
We dont normalize headers for performance reasons.

PandaWorker:
In any case, the headers are normalized when sending requests. 

[...]

PR #4498, opened 2025-09-01
fix(dns): preserve header shape and ensure Host

FelixVaughan:
## This relates to...
#4444

## Rationale
Preserve header shape and correct Host after DNS resolution to avoid routing issues/header corruption

## Changes
Added `addHostHeader` to dns interceptor and parameterized tests for different header types.

### Bug Fixes
Fixed DNS interceptor turning [string, string][]/iterables into numeric-key headers

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready


lib/interceptor/dns.js
```diff
@@ -5,6 +5,43 @@ const DecoratorHandler = require('../handler/decorator-handler')
 const { InvalidArgumentError, InformationalError } = require('../core/errors')
 const maxInt = Math.pow(2, 31) - 1
 
+function addHostHeader (headers, host) {
```
metcoder95: Hmm, you don't need to parse the headers by yourself, this is done by undici automatically, we just need to standardise the way that these are forward to undici.

If Array, push `host` to the headers array, if an object, attach `host` as a new property

FelixVaughan: I've simplified the code!

A header case I'm slightly worried about though is non-array iterables. 

`request({ origin, path: '/', method: 'GET', headers: new Set([['f0', 'b0'], ['f1', 'b1']]) })`. 

These headers are valid under `UndiciHeaders` and a request like this (though rare) is possible; but the proposed array/object check would break this.

Another issue I've noticed is [string,string][] headers always raise an error due to the way header processing is handled in `Request.js`. Say headers =  [['f0', 'b0']] or [['f0', 'b0'], ['f1', 'b1']]

```
if (Array.isArray(headers)) {
      if (headers.length % 2 !== 0) {
        throw new InvalidArgumentError('headers array must be even')
      }
        processHeader(this, headers[i], headers[i + 1])
      }
}
```

metcoder95: >These headers are valid under UndiciHeaders and a request like this (though rare) is possible; but the proposed array/object check would break this.

Usually these are iterables, and we iterate over them, but not use utility functions; if that's the case, we can try to also check for `Symbol.iterator` to decide what to do with them.

>Say headers = [['f0', 'b0']] or [['f0', 'b0'], ['f1', 'b1']]

That's somehow expected, we only expect them in the format `[string, string]`, anything outside should throw

lib/interceptor/dns.js
```diff
@@ -5,6 +5,17 @@ const DecoratorHandler = require('../handler/decorator-handler')
 const { InvalidArgumentError, InformationalError } = require('../core/errors')
 const maxInt = Math.pow(2, 31) - 1
 
+function addHostHeader (headers = {}, host) {
+  if (Array.isArray(headers)) {
+    const header = ['host', host]
+    if (Array.isArray(headers[0])) {
```
metcoder95: Do we really support them? They should throw if `request` receives a multi-dimensional array.

lib/interceptor/dns.js
FelixVaughan: Handling iterators might mean we can't avoid some normalization. However this can be streamlined with the `normalizeHeaders` function also used in `cache.js`

https://github.com/nodejs/undici/blob/ad4d54791ffb667b1c5e5426a02e6d3e15ef71c1/lib/util/cache.js#L36

```patch
diff --git a/lib/interceptor/dns.js b/lib/interceptor/dns.js
index 38287607..a878f39f 100644
--- a/lib/interceptor/dns.js
+++ b/lib/interceptor/dns.js
@@ -4,6 +4,15 @@ const { lookup } = require('node:dns')
 const DecoratorHandler = require('../handler/decorator-handler')
+const { normalizeHeaders } = require('../util/cache.js')
+
+function addHostHeader (opts, host) {
+  if (Array.isArray(opts.headers)) {
+    return ['host', host, ...opts.headers]
+  }
+  const headers = normalizeHeaders(opts)
+  return { host, ...headers }
+}
 
 class DNSInstance {
   #maxTTL = 0
@@ -411,10 +420,7 @@ module.exports = interceptorOpts => {
           ...origDispatchOpts,
           servername: origin.hostname, // For SNI on TLS
           origin: newOrigin.origin,
-          headers: {
-            host: origin.host,
-            ...origDispatchOpts.headers
-          }
+          headers: addHostHeader(origDispatchOpts, origin.host)
         }
 
         dispatch(
```

lib/interceptor/dns.js
FelixVaughan: They do throw for arrays, just not other iterables

lib/interceptor/dns.js
metcoder95: Hmm... that's a good catch; I'd suggest to move the normalize headers to `core/utils` and implement it within the DNS interceptor first, and we can check if we wanna do the same for the core request in a further issue

lib/interceptor/dns.js
metcoder95: Got it, let's do this: https://github.com/nodejs/undici/pull/4498#discussion_r2317932477

lib/interceptor/dns.js
gurgunday: > Hmm... that's a good catch; I'd suggest to move the normalize headers to `core/utils` and implement it within the DNS interceptor first, and we can check if we wanna do the same for the core request in a further issue

This makes sense to me

lib/interceptor/dns.js
```diff
@@ -4,15 +4,13 @@ const { lookup } = require('node:dns')
...
 
-function addHostHeader (headers = {}, host) {
-  if (Array.isArray(headers)) {
-    const header = ['host', host]
-    if (Array.isArray(headers[0])) {
-      return [header, ...headers]
-    }
-    return [...header, ...headers]
+function addHostHeader (opts, host) {
+  if (Array.isArray(opts.headers)) {
```
metcoder95: The `normalizeHeaders` will normalize it to a plain JS object, feel free to append it once normalized.

FelixVaughan: Not for flat arrays for some reason. Refactored to accept types of `[k1, v1, k2, v2, ...]`

lib/core/util.js
```diff
@@ -904,6 +904,44 @@ const normalizedMethodRecords = {
...
+    for (let i = 0; i < src.length; i += 2) {
+      headers[src[i]] = src[i + 1]
+    }
+    return headers
+  }
+
+  if (typeof src[Symbol.iterator] === 'function') {
+    const msg = 'opts.headers is not a valid header map'
+    for (const s of src) {
+      if (!Array.isArray(s)) throw new Error(msg)
```
metcoder95: If there's an Undici error that is usable, let's make usage of it

metcoder95 (approved):

lib/core/util.js
```diff
@@ -904,6 +904,44 @@ const normalizedMethodRecords = {
...
+      if (typeof key !== 'string' || typeof val !== 'string') {
+        throw new InvalidArgumentError(errMsg)
+      }
+      headers[key.toLowerCase()] = val
+    }
+    return headers
+  }
+
+  if (typeof src === 'object') {
+    for (const key of Object.keys(src)) headers[key.toLowerCase()] = src[key]
```
Uzlopak: dont we have somehwere a lookup function to potentially avoid the lowercase?

lib/core/util.js
```diff
@@ -904,6 +904,44 @@ const normalizedMethodRecords = {
 Object.setPrototypeOf(normalizedMethodRecordsBase, null)
 Object.setPrototypeOf(normalizedMethodRecords, null)
 
+/**
+ * @param {Record<string, string | string[]> | Iterable<[string, string]> | string[]}
+ * @returns {Record<string, string[] | string>}
+ */
+function normalizeHeaders (opts) {
+  const headers = {}
+  const src = opts.headers
```
Uzlopak: is the jsdoc fake news :D?

lib/interceptor/dns.js
```diff
@@ -411,10 +412,7 @@ module.exports = interceptorOpts => {
           ...origDispatchOpts,
           servername: origin.hostname, // For SNI on TLS
           origin: newOrigin.origin,
-          headers: {
-            host: origin.host,
-            ...origDispatchOpts.headers
-          }
+          headers: { host: origin.host, ...normalizeHeaders(origDispatchOpts) }
```
Uzlopak: Why would you pass origDispatchOpts if your function is named normalizeHeaders?!

```suggestion
          headers: { host: origin.host, ...normalizeHeaders(origDispatchOpts.headers) }
```

lib/core/util.js
```diff
@@ -904,6 +904,44 @@ const normalizedMethodRecords = {
 Object.setPrototypeOf(normalizedMethodRecordsBase, null)
 Object.setPrototypeOf(normalizedMethodRecords, null)
 
+/**
+ * @param {Record<string, string | string[]> | Iterable<[string, string]> | string[]}
+ * @returns {Record<string, string[] | string>}
+ */
+function normalizeHeaders (opts) {
```
Uzlopak:
```suggestion
function normalizeHeaders (src) {
```

lib/core/util.js
```diff
@@ -904,6 +904,44 @@ const normalizedMethodRecords = {
 Object.setPrototypeOf(normalizedMethodRecordsBase, null)
 Object.setPrototypeOf(normalizedMethodRecords, null)
 
+/**
+ * @param {Record<string, string | string[]> | Iterable<[string, string]> | string[]}
```
Uzlopak:
```suggestion
 * @param {Record<string, string | string[]> | Iterable<[string, string]> | string[]} src
```

lib/core/util.js
```diff
@@ -904,6 +904,44 @@ const normalizedMethodRecords = {
...
+ * @returns {Record<string, string[] | string>}
+ */
+function normalizeHeaders (opts) {
+  const headers = {}
+  const src = opts.headers
+
+  if (src == null) return headers
```
Uzlopak:
```suggestion
  if (typeof src !== 'object' || src === null) return headers
```

lib/core/util.js
```diff
@@ -904,6 +904,44 @@ const normalizedMethodRecords = {
...
+      const [key, val] = s
+      if (typeof key !== 'string' || typeof val !== 'string') {
+        throw new InvalidArgumentError(errMsg)
+      }
+      headers[key.toLowerCase()] = val
+    }
+    return headers
+  }
+
+  if (typeof src === 'object') {
```
Uzlopak:
```suggestion
```

lib/core/util.js
```diff
@@ -904,6 +904,44 @@ const normalizedMethodRecords = {
...
+    for (const key of Object.keys(src)) headers[key.toLowerCase()] = src[key]
+    return headers
+  }
```
Uzlopak, line 965:
```suggestion
```

also needs to run through the linter

lib/core/util.js
```diff
@@ -904,6 +904,44 @@ const normalizedMethodRecords = {
...
+
+  throw new InvalidArgumentError('opts.headers is not an object')
```
Uzlopak: is this really a problem?

Uzlopak (changes requested):
-1

metcoder95:
-1 on the algorithm or the feature itself?

FelixVaughan:
@Uzlopak the normalization function was moved out of `lib/util/cache.js` to `lib/core/util.js`,  with some minor refactors
https://github.com/nodejs/undici/blob/95d835cd9db685e4f7baaea8e10aee8ce77e7989/lib/util/cache.js#L36

Many suggestions are referring to things that were present in the initial implementation which were left unchanged to avoid introducing reversions.

metcoder95:
cc @Uzlopak

Uzlopak:
I know it is maybe a little bit frustrating situation right now. I am still thinking if this is the right thing to do. Also I want to evaluate it further and see what the rest of the code does after passing further down the line. Didnt have the muse to do it till now.

It creates alot of new intermediary  objects and makes alot of toLowerCase() calls. And that just for the sake of normalizing the header. 

What if it is an array? why not push key and value? Dont we have a logic, which overwrites the value if a key is already set? So why do we need to transform it to an object in that moment?

What if it is an Object? Afaik an Object.keys does not return the keys alphabetically but in the order of being assigned to the object. Should we rely on this behavior?  

Should we not use wellknown header logic (see /lib/core/constants.js)  to avoid potential `toLowerCase()` calls. Maybe it was already a bad design, when cache did not use that weelknownheader logic. Doesnt mean that we should ignore this perf regression

Or should we not just actually create a Header Instance?

FelixVaughan:
All great questions, some of which I don’t have immediate answers for either, but in general I think we should be utilizing the lowercase wellKnownHeader functionality already in place wherever possible. 

For the arrays and objects, we normalize to avoid the case where the passed in header is neither an object in the traditional key/value sense or an array, but an iterable like a set; where a traditional push would lead to undefined behavior. We might be able to shortcircuit this by a straight append in the case of an array as you pointed out but objects and iterators are slightly more complex.

Also, though I think it makes sense to potentially tackle it in this pr, due to the pre-existing nature of the problem, should this perf concern be handled as a separate issue?

Uzlopak:
We should not transplant one bad implementation to other places and widely use it.

Right now only one person or few people complained regarding passing arrays into the interceptor. 

Who is gonna fix that bad implementation later? 

"A permanent solution is a temporary solutation that works." or other variation: "Temporary solutions often become permanent problems."

Also: We have multiple implementations of normalizeHeaders. One in snapshot. Also it is dangerous, because normalizeHeader is actually just normalizeRequestHeader.

Uzlopak (changes requested):
Per explaination.

lib/core/util.js
FelixVaughan: not anymore hah

FelixVaughan:
I understand your concern. I'd work on it if we move it to a separate issue. My main worry is potential scope creep; case in point the duplicate normalization in snapshot.

I've made the changes to address the above, but if we decide to tackle the rest here, we should lay out what all we want to change and I can handle it.

Cheers!

FelixVaughan:
@metcoder95 can we progress or close this?

metcoder95:
@Uzlopak can you take a look?

Codewise, lgtm, but I understand the concerns about performance implications; I'd like suggest fix the issue at hand a find a midterm of optimization/bug-fixing.

mcollina (approved):
lgtm

mcollina:
@Uzlopak can you take another look? What would be your recommendation to fix this bug then?

PR #4504, opened 2025-09-01
fix writing to websocketstream with SharedArrayBuffer/SharedArrayBuff…

KhafraDev:
…er view

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/web/websocket/stream/websocketstream.js
+++ b/lib/web/websocket/stream/websocketstream.js
@@ -199,6 +199,9 @@ class WebSocketStream {
   }
 
   #write (chunk) {
+    // See /websockets/stream/tentative/write.any.html
+    chunk = webidl.converters.WebSocketStreamWrite(chunk)
+
     // 1. Let promise be a new promise created in stream ’s relevant realm .
     const promise = createDeferredPromise()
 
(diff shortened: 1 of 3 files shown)


lib/web/websocket/stream/websocketstream.js
```diff
@@ -199,6 +199,9 @@ class WebSocketStream {
   }
 
   #write (chunk) {
+    // See /websockets/stream/tentative/write.any.html
```
Uzlopak, line 202: I would prefer not having references to the wpt test suite...

lib/web/websocket/stream/websocketstream.js
```diff
@@ -484,4 +487,12 @@ webidl.converters.WebSocketCloseInfo = webidl.dictionaryConverter([
   }
 ])
 
+webidl.converters.WebSocketStreamWrite = function (V) {
+  if (typeof V === 'string') {
+    return webidl.converters.USVString(V)
+  }
+
+  return webidl.converters.BufferSource(V)
```
Renegade334, line 495: The only minor thing with this approach is that the failure message on passing an invalid type will only list ArrayBuffer and ArrayBufferView as valid types, not DOMString.

lib/web/websocket/stream/websocketstream.js
KhafraDev: As the person maintaining websocketstream, it helps tremendously to know why this was added. It's not mentioned in the spec, this isn't how other specs work (typically SharedArrayBuffer/view would fallback to being stringified), and the only reference is this singular test file.

PR #4529, opened 2025-09-07
Revert "chore: update llhttp"

Uzlopak:
Reverts nodejs/undici#4431


metcoder95 (approved):

mcollina:
Can you please add a regression test?

Uzlopak:
I am not sure regarding regressions. I wanted to put this up for discussion. I was partially hoping that it maybe fixes the macos ci issues. Anyhow. My plan would be actually to create a PR against nodejs with the contents of this PR, and see if nodejs is still broken on those other architectures.

mcollina:
Try getting the test working, I would prefer to _not_ land this unless we have a regression - and if we do have a regression test, we can get @ShogunPanda fix it in llhttp and avoid reverting it completely.

Uzlopak:
@mcollina 

THe problem is, that I dont have a computer with AIX or ppc64.

mcollina:
Wait, is the problem only on AIX or PPC?

Uzlopak:
@mcollina 

That what I am seeing. 
image

https://ci.nodejs.org/job/node-test-pull-request/68984/

Uzlopak:
Assuming that this is basically obsolete due to #4530

PR #4538, opened 2025-09-11
chore: enforce explicit Buffer import

shivarm:
Explicitly import Buffer using `const { Buffer } = require('node:buffer')` for clarity and It would be good to keep consistency.

cc: @mcollina


Uzlopak:
Dont get me wrong, but we should be consistent. 

What comes next?
requiring `setTImeout` and `clearTimeout` from `node:timers` ?

This would also make sense, because `setTimeout` from `node:timers` has definetly `refresh()` while setTimeout from global can be without `refresh()` in deno. 

Where do we draw the line? 

Also if you want to be consistent, than we should modify our eslint config too, so that globals should also fail linting if not used from the corresponding module.

So either we say, no globals for clarity and keeping consistency or use globals. But please not a case by case decision.

shivarm:
> Also if you want to be consistent, than we should modify our eslint config too, so that globals should also fail linting if not used from the corresponding module.

Yes, I like to do this.

> So either we say, no globals for clarity and keeping consistency or use globals. But please not a case by case decision.

We will target some modules like `node:buffer` that are very often used and very important because node.js team still making features on top of the globals and deprecating API for the module. But if you want the same for others then we can do  it.

KhafraDev:
> clarity and It would be good to keep consistency.

How does this improve clarity and consistency with what?

shivarm:
> How does this improve clarity and consistency with what?

- By explicitly importing Buffer from `node:buffer`, it’s clear to every reader (especially new contributors) what module Buffer comes from. 
- Consistent import style also helps during refactoring or if Node.js ever changes how globals work
- In Node.js, Buffer is globally available, but explicit imports make the dependency clear and self-documenting.

KhafraDev:
Someone contributing to undici wouldn't know that Buffer is a global? Seriously? Also Buffer being a global will never change.

shivarm:
> Someone contributing to undici wouldn't know that Buffer is a global? Seriously? Also Buffer being a global will never change.

Yeah, you are right but as we know our project based on node.js as javascript runtime that's why I want to integrate this change so that we can have best practice of node.js. 

Also, is there any problem with reading Node.js docs for someone? which is have well structure and clear information not like traditional one.

@mcollina @KhafraDev @Uzlopak WDYT?

Uzlopak:
Its like saying that you wear a raincoat because it is raining, but you are not telling us if you are inside or outside a house. If you are outside a house, it would be considered reasonable. But if you are inside a house, it would be considered as unreasonable.

Argueing in the name of best practice, without defining what best practice is, makes it just a hollow catch word. 

If we get to this level of discussion, then I can simply claim that it is best practice to use globals. 

What now? Who is right? Who is wrong? 

Till now i didnt read here any valid Argument. Please come with arguments and not with catch-words.

shivarm:
> Its like saying that you wear a raincoat because it is raining, but you are not telling us if you are inside or outside a house. If you are outside a house, it would be considered reasonable. But if you are inside a house, it would be considered as unreasonable.

It is recommended by node.js docs itself check - https://nodejs.org/docs/latest/api/buffer.html#:~:text=While%20the,require%20statement

> Argueing in the name of best practice, without defining what best practice is, makes it just a hollow catch word.

Explicit imports make it clear where a symbol comes from, simplifying dependency analysis and automated refactoring. Tools like `ESLint, TypeScript, and IDEs` can track `imports` more reliably than `globals`.

> If we get to this level of discussion, then I can simply claim that it is best practice to use globals.

The broader JS ecosystem is moving away from globals and toward explicit imports (ES Modules, strict mode, etc.). Node’s own documentation is trending toward this style, and many popular libraries (e.g., Express, Fastify) prefer explicit imports for clarity.

KhafraDev:
It was changed in https://github.com/nodejs/node/pull/39043 with no justification or reasoning. Yet, the previous statement and PR description both conclude the same thing:

> The `Buffer` class is within the global scope, making it unlikely that one
would need to ever use `require('buffer').Buffer`.

> It does not deprecate or say that the global forms are deprecated though.

The documentation needs an update.

>  and many popular libraries (e.g., Express, Fastify)

Express only includes this because you made a similar PR, which was originally blocked on similar grounds (it seems they got tired of arguing over a minor, unnecessary change).

Fastify doesn't or it isn't enforced. https://github.com/fastify/fastify/blob/39f35f9a0596537326b960183b1aa64aca796632/lib/reply.js

Uzlopak:
It was only added when they changed the docs for esm. I dont even know if there was a big consensus, as only @aduh95 approved that PR.

https://github.com/nodejs/node/commit/f4d0a6a07bf72c556c5ba8e98b3b23327c273e80#diff-2aeb90ba4258a1154f8558dc18c7ad46778ac0bb479f564fe96853d35303bca0

Anyhow, we are not bound to this "recommendation" in the nodejs docs. We also dont rewrite everything to esm, despite that is maybe now the "recommendation".

Also you claim, that e.g. fastify is "importing" Buffer and not using the global. I checked now the sourcecode of fastify, and I cant find a proof for your claim. 

So right now, the first potential argument would be that you claim that "automated refactoring tools" can track them better, which i dont see any benefit for?! What is the benefit of doing a "find all references" on Buffer, which I cant do with a search in the IDE? When do I need to use those "automated refactoring tools"?

Right now there is only a proposal for adding alot of require calls and the argument: "the others do, so we should do it too". 

Also on a sidenote, your arguments seem to be AI generated. Thats why there is the hallucination that fastify would import the buffer from a module or the "automated refactoring tools" could determine where a "symbol" (biggest smell for AI) comes from.

Uzlopak:
I am really tending to close this as it seems to be a waste of time when argueing against an AI.

shivarm:
@mcollina @KhafraDev @Uzlopak oh, then I think, I missguided by node.js docs. it is waste of time. 

Closing this!

Uzlopak:
Thank you for your understanding.

PR #4540, opened 2025-09-11
feat(#2458): WebSocket through HTTP/2

### LINKED ISSUE #2458: websocket over http2
Refs: https://github.com/nodejs/undici/pull/2061#issuecomment-1685395999

metcoder95:
## This relates to...

Closes #2458 

## Rationale

Implements WebSocket over HTTP/2

> [!NOTE]
> Experimental warning is emitted on first usage.

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/test/http2-dispatcher.js
+++ b/test/http2-dispatcher.js
@@ -265,7 +265,6 @@ test('Dispatcher#Upgrade', async t => {
     stream.end()
   })
 
-  after(() => server.close())
   await once(server.listen(0), 'listening')
 
   const client = new Client(`https://localhost:${server.address().port}`, {
@@ -274,7 +273,7 @@ test('Dispatcher#Upgrade', async t => {
     },
     allowH2: true
   })
-  after(() => client.close())
+  after(() => client.close().then(() => { server.close() }))
 
   const { socket } = await client.upgrade({ path: '/', protocol: 'websocket' })
 


KhafraDev:
I started looking into it too... perfect timing

test/websocket/opening-handshake.js
```diff
@@ -49,6 +51,76 @@ test('Open event is emitted', () => {
   })
 })
 
+// TODO: ws does not supports HTTP2; will need to potentially write a custom server for this
```
lpinca: See https://github.com/websockets/ws/issues/1458#issuecomment-1856886384 for a simple/limited wrapper.

metcoder95:
I found a bit of a wall here; for `fetch`, it seems the header expected is `upgrade` in order to understand the connection should be upgraded to a `websocket`. Nonetheless, this is not a valid H2 header, so we can just `ack` if we identify is an H2 connection or some other idea that we might have.

> Ref: https://github.com/nodejs/undici/blob/be390d8cdd621a19da87ea65e0110f932394d72f/lib/web/websocket/connection.js

Happy to hear your thoughts @KhafraDev

KhafraDev:
We check for the header because it's needed to complete the http upgrade and opening handshake (RFC 6455). For http2, we should follow RFC 8441.

lib/dispatcher/client-h2.js
```diff
@@ -364,13 +406,60 @@ function writeH2 (client, request) {
...
+        util.errorRequest(client, request, new InformationalError('HTTP/2: Extended CONNECT protocol not supported by server'))
+        session.unref()
+        return false
+      }
+
+      // We force the method to CONNECT
+      // as per RFC-8441
+      // https://datatracker.ietf.org/doc/html/rfc8441#section-4
+      headers[HTTP2_HEADER_METHOD] = 'CONNECT'
+      headers[':protocol'] = upgrade
```
KhafraDev: nit: http2.constants.HTTP2_HEADER_PROTOCOL for consistency

lib/dispatcher/client-h2.js
```diff
@@ -364,13 +406,60 @@ function writeH2 (client, request) {
     return false
   }
 
-  if (method === 'CONNECT') {
+  if (upgrade || method === 'CONNECT') {
     session.ref()
-    // We are already connected, streams are pending, first request
+
+    // We cannot upgrade to websocket if extended CONNECT protocol is not supported
+    if (upgrade === 'websocket') {
```
KhafraDev, line 451: nit: upgrade can only be 'websocket' at this point

lib/dispatcher/client-h2.js
```diff
@@ -364,13 +406,60 @@ function writeH2 (client, request) {
...
+        const { [HTTP2_HEADER_STATUS]: statusCode, ...realHeaders } = headers
+
+        request.onUpgrade(statusCode, parseH2Headers(realHeaders), stream)
+
+        ++session[kOpenStreams]
+        client[kQueue][client[kRunningIdx]++] = null
+      })
+
+      stream.on('error', () => {
+        if (stream.rstCode === 7 || stream.rstCode === 8) {
```
KhafraDev: nit: http2.constants.NGHTTP2_REFUSED_STREAM and http2.constants.NGHTTP2_CANCEL

lib/web/fetch/index.js
```diff
@@ -2228,8 +2228,9 @@ async function httpNetworkFetch (
         },
 
         onUpgrade (status, rawHeaders, socket) {
-          if (status !== 101) {
-            return
+          // We need to support 200 for websocket over h2 as per RFC-8441
+          if (status !== 101 && status !== 200) {
```
KhafraDev: can we assert that it's http2 when status = 200?

lib/web/websocket/connection.js
```diff
@@ -97,11 +99,29 @@ function establishWebSocketConnection (url, protocols, client, handler, options)
...
       }
 
-      // 1. If response is a network error or its status is not 101,
+      // 1. If response is a network error or its status is not 101 (or 200 for H2),
       //    fail the WebSocket connection.
+      // if (response.type === 'error' || ((response.socket?.session != null && response.status !== 200) && response.status !== 101)) {
       if (response.type === 'error' || response.status !== 101) {
-        failWebsocketConnection(handler, 1002, 'Received network error or non-101 status code.', response.error)
-        return
+        console.log(response.socket?.session != null, response.status !== 200)
```
KhafraDev:
```suggestion
```

lib/web/websocket/connection.js
```diff
@@ -97,11 +99,29 @@ function establishWebSocketConnection (url, protocols, client, handler, options)
         handler.readyState = states.CLOSED
       }
 
-      // 1. If response is a network error or its status is not 101,
+      // 1. If response is a network error or its status is not 101 (or 200 for H2),
```
KhafraDev:
```suggestion
      // 1. If response is a network error or its status is not 101,
```

this comment comes from https://websockets.spec.whatwg.org/#concept-websocket-establish

lib/dispatcher/client-h2.js
metcoder95: So you suggest to only verify that `upgrade` is not `null` or `undefined`?

metcoder95:
@KhafraDev feel free to take another look

lib/dispatcher/client-h2.js
```diff
@@ -151,6 +185,17 @@ function resumeH2 (client) {
   }
 }
 
+// TODO: fix - if request attempting to upgrade connection and remote settings not received
+// we need to halt execution until we receive the remote settings
+// Other requests can continue as normal
+function onHttp2RemoteSettings (settings) {
```
KhafraDev, line 214: > A sender MUST NOT send a SETTINGS_ENABLE_CONNECT_PROTOCOL parameter
   with the value of 0 after previously sending a value of 1.

We should check that it hasn't been set to 1 already

lib/dispatcher/client-h2.js
```diff
@@ -364,13 +409,60 @@ function writeH2 (client, request) {
...
+        util.errorRequest(client, request, new InformationalError('HTTP/2: Extended CONNECT protocol not supported by server'))
+        session.unref()
+        return false
+      }
+
+      // We force the method to CONNECT
+      // as per RFC-8441
+      // https://datatracker.ietf.org/doc/html/rfc8441#section-4
+      headers[HTTP2_HEADER_METHOD] = 'CONNECT'
+      headers[HTTP2_HEADER_PROTOCOL] = upgrade
```
KhafraDev: > The :protocol pseudo-header field MUST be included in the CONNECT
   request, and it MUST have a value of "websocket"

We should set this to 'websocket' directly just for brevity

lib/dispatcher/client-h2.js
```diff
@@ -364,13 +409,60 @@ function writeH2 (client, request) {
     return false
   }
 
-  if (method === 'CONNECT') {
+  if (upgrade || method === 'CONNECT') {
```
KhafraDev, line 448: with the early return it's possible to return before setting :scheme and :path

> On requests that contain the :protocol pseudo-header field, the
      :scheme and :path pseudo-header fields of the target URI (see
      [Section 5](https://www.rfc-editor.org/rfc/rfc8441#section-5)) MUST also be included.

lib/dispatcher/client-h2.js
```diff
@@ -364,13 +409,60 @@ function writeH2 (client, request) {
...
     session.ref()
-    // We are already connected, streams are pending, first request
+
+    if (upgrade === 'websocket') {
+      // We cannot upgrade to websocket if extended CONNECT protocol is not supported
```
KhafraDev, line 452: I believe fetch will already rewrite wss -> https and ws -> http, but if someone were to use this outside of fetch/WebSocket we should set :scheme if we have a ws(s):// url

> The scheme of the target URI ([Section 5.1 of [RFC7230]](https://www.rfc-editor.org/rfc/rfc7230#section-5.1)) MUST be
   "https" for "wss"-schemed WebSockets and "http" for "ws"-schemed
   WebSockets.

lib/dispatcher/client-h2.js
KhafraDev: what I mean is `if (settings.enableConnectProtocol != null && !settings.enableConnectProtocol && this[kEnableConnectProtocol]) { // error }`

metcoder95: Oh, ok got it know; basically check that a remote settings frame does not changes values between requests. Sorry, missed that, thanks!

lib/dispatcher/client-h2.js
metcoder95: Good catch!

docs/docs/api/WebSocket.md
```diff
@@ -2,6 +2,8 @@
 
 Extends: [`EventTarget`](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget)
 
+> ⚠️ Warning: WebSocket over HTTP/2 is experimental, it is likely to change in the future.
```
KhafraDev: I think there should be a section on how to enable it and mention how we might enable it by default in another (major?) version

KhafraDev (approved):

metcoder95:
@KhafraDev feel free to have another look to the docs and let me know your thoughts 👍

docs/docs/api/WebSocket.md
```diff
@@ -34,6 +34,27 @@ import { WebSocket } from 'undici'
 const ws = new WebSocket('wss://echo.websocket.events', ['echo', 'chat'])
 ```
 
+### Example with HTTP/2:
+
+> ⚠️ Warning: WebSocket over HTTP/2 is experimental, it is likely to change in the future.
+
+> 🗒️ Note: WebSocket over HTTP/2 is possible to be enabled by default in further major versions,
+> this will happen by enabling HTTP/2 connections as the de`fault behavior of Undici's Agent as well the global dispatcher.
+> Stay tuned to the changelog for more information.
```
KhafraDev:
```suggestion
> 🗒️ Note: WebSocket over HTTP/2 may be enabled by default in a future version,
> this will happen by enabling HTTP/2 connections as the default behavior of Undici's Agent as well the global dispatcher.
> Stay tuned to the changelog for more information.
```

KhafraDev: Is the idea to enable http/2 by default outside of websocket? I was thinking of having it enabled only for websockets because I remember there was some push back last time you tried enabling it by default.

KhafraDev (commented):
There seems to be a relevant test failure in test/http2-dispatcher.js

docs/docs/api/WebSocket.md
metcoder95: Not really sure what could be the plan here; the thing with enabling it only for WebSocket is that it might require to use a custom agent if H2 not enabled through global dispatcher as default.

It will be better to sync them, but if we are ok deviating one from the other, that should be totally fine

metcoder95:
@KhafraDev can you check? Let me know if is ok to merge it in this way

KhafraDev:
There are some http2 test failures, I'm not sure if they're relevant.

metcoder95:
Seems that they are now fixed, the failures doesn't seem related to the PR's changes

PR #4545, opened 2025-09-14
feat: add runtime feature "detection"

Uzlopak:
Basically centralizes the detection of crypto and sqlite. Also I rewrote some code to avoid the loading of crypto and sqlite module to early.

Maybe it makes sense to have some smaller PRs, if you dont think runtimeFeatures instance/class is a  good solution. But with runtimeFeatures Instance, we could actually disable specific features and test it.

We could actually also have a little crypto ponyfill

```js
'use strict'

const { featureSet } = require('../../util/feature-detection.js')
/** @type {import('node:crypto')} */
const crypto = featureSet.has('crypto')
  ? require('node:crypto')
  : {
      getHashes: () => [],
      hash: (algorithm, data, outputEncoding) => {
        return outputEncoding === undefined || outputEncoding === 'buffer'
          ? Buffer.alloc(0)
          : ''
      },
      randomBytes: (size) => {
        const buf = Buffer.allocUnsafe(size)
        for (let i = 0; i < size; i++) {
          buf[i] = Math.floor(Math.random() * 256)
        }
        return buf
      },
      randomInt: function randomInt (max) {
        return Math.floor(Math.random() * max)
      },
      randomFillSync: function randomFillSync (buffer, _offset, _size) {
        for (let i = 0; i < buffer.length; ++i) {
          buffer[i] = Math.random() * 255 | 0
        }
        return buffer
      }
    }

module.exports = crypto

```

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready


lib/util/runtime-features.js
```diff
@@ -0,0 +1,79 @@
...
+  try {
+    require(moduleName)
+    return true
+  } catch (err) {
+    if (err.code !== 'ERR_UNKNOWN_BUILTIN_MODULE') {
+      throw err
+    }
+    return false
+  }
+}
```
metcoder95, line 21: Could it be better to return the module itself as well?

Uzlopak: I would actually try to avoid to have a second require cache or something like that. 

I actually think also about a check for zsdt or maybe markAsClonable. 

It could help have us to test branches which are supported by the engine.

Uzlopak:
@metcoder95 
I added more "features".

mcollina (approved):
lgtm

metcoder95 (approved):

PR #4546, opened 2025-09-15
types: use MessagePort instance type in MessageEvent

Renegade334:
These previously mistakenly used the constructor type. Doesn't appear to be any similar errors elsewhere.


Uzlopak:
I think I wrote that. It makes sense to not use the constructor.

PR #4555, opened 2025-09-20
fix: use `globalThis.require` to require dynamically

ShenHongFei:
## This relates to
Optimize the following pr
feat: add runtime feature "detection" https://github.com/nodejs/undici/pull/4545
@mcollina 

## Rationale

Use `globalThis.require` to dynamically require to avoid dynamic require warnings generated by bundlers such as webpack when packaging undici.  
The final product `bundle.js` does not actually require the corresponding node.js built-in module, resulting in the following error:
```log
WARNING in ./lib/util/runtime-features.js 13:4-23
Critical dependency: the request of a dependency is an expression

Error: Cannot find module 'node:worker_threads'
    webpackEmptyContext() @ internal/deps/undici/undici:10888:10
    detectRuntimeFeatureByExportedProperty() @ internal/deps/undici/undici:8762:43
    detectRuntimeFeature() @ internal/deps/undici/undici:8790:12
    #detectRuntimeFeature() @ internal/deps/undici/undici:8836:20
    RuntimeFeatures.has() @ internal/deps/undici/undici:8816:59
    2480() @ internal/deps/undici/undici:4890:49
    __webpack_require__() @ internal/deps/undici/undici:24322:41
    575() @ internal/deps/undici/undici:832:20
    __webpack_require__() @ internal/deps/undici/undici:24322:41
    8129() @ internal/deps/undici/undici:19837:5 {
  code: 'MODULE_NOT_FOUND'
}
```

caused by 
```js
function detectRuntimeFeatureByNodeModule (moduleName) {
  try {
    __webpack_require__(4839)(moduleName)
    return true
  } catch (err) {
    if (err.code !== 'ERR_UNKNOWN_BUILTIN_MODULE') {
      throw err
    }
    return false
  }
}


/***/ 4839:
/***/ ((module) => {
function webpackEmptyContext(req) {
	var e = new Error("Cannot find module '" + req + "'");
	e.code = 'MODULE_NOT_FOUND';
	throw e;
}
webpackEmptyContext.keys = () => ([]);
webpackEmptyContext.resolve = webpackEmptyContext;
webpackEmptyContext.id = 4839;
module.exports = webpackEmptyContext;
/***/ })
```

## Changes

require -> for static import, processed by bundlers
globalThis.require -> for dynamic import

### Features

N/A

### Bug Fixes

N/A

N/A

## Status

- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/util/runtime-features.js
+++ b/lib/util/runtime-features.js
@@ -10,7 +10,7 @@
  */
 function detectRuntimeFeatureByNodeModule (moduleName) {
   try {
-    require(moduleName)
+    globalThis.require(moduleName)
     return true
   } catch (err) {
     if (err.code !== 'ERR_UNKNOWN_BUILTIN_MODULE') {
@@ -26,7 +26,7 @@ function detectRuntimeFeatureByNodeModule (moduleName) {
  * @returns {boolean}
  */
 function detectRuntimeFeatureByExportedProperty (moduleName, property) {
-  const module = require(moduleName)
+  const module = globalThis.require(moduleName)
   return typeof module[property] !== 'undefined'
 }
 


ShenHongFei:
@Uzlopak Can you take a look at this?

Uzlopak:
To be honest, i dont understand why this should be in this case an issue. 

E.g. in client-h2 we lazily require `node:tls`. Why is this not resulting in that error? 

Oh wait...

Well, that could be hmm... I am now passing an arbitrary string to require... Hmm

So if we would do something like 

const lazyCrypto = () => require('node:crypto')

than your issue should be also solved. 

Can you rewrite it to work with lazy$moduleName functions? Without doing globalThis.require

Something like

/** @type {Record<Feature, Function> */
const lazyRequireLookup = {
  __proto__: null,
  crypto: () => require('node:crypto'),
  ...
}

And then in detectRuntimeByNodeModule instead of require('moduleName') you do lazyRequireLookup[moduleName]()

Wrote this comment with my mobile.
 Cant use proper markdown. And did not test it. Probably function names and stuff is wrong. But you should get the gist of it.

ShenHongFei:
I think the following should be able to be bundled normally.  
`const lazyCrypto = () => require('node:crypto')`

But using lazy... function is not very general. You need to add a lazy... function for each additional module.  

If there are many modules, the code will be redundant.

Uzlopak:
So let it be redundant. 

I dislike the idea to use globalThis.require

Uzlopak:
I dislike use of globalThis.require because i think that bundlers, not only webpack, can handle the "normal" require better.

I worry that later somebody comes and says that globalThis.require is making issues.

Probably one aspect which i did not consider is also that maybe the security aspect that with this function any module could be required. 

So lets do it with lazyLoaders.

mcollina:
CI is _very red_, and it shows that this solution is not working as expected. I'm not exactly sure what globalThis.require is, but it's not something of "stock" Node.js.

@ShenHongFei why don't you work in creating a test case to reproduce the problem you are trying to fix?

Uzlopak (changes requested):
What is going on? I will be at my Computer in an hour and will in investigate this.

ShenHongFei:
OK, if you have time, please submit a new PR and do as you said above. I am going to close this one. @Uzlopak

PR #4557, opened 2025-09-20
fix: use lazyloaders for runtime feature detection

Uzlopak:
as discussed #4555 

use lazyLoader to avoid having a require fn which dynamically loads "any" module.

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/util/runtime-features.js
+++ b/lib/util/runtime-features.js
@@ -1,8 +1,15 @@
-// @ts-check
-
 'use strict'
 
 /** @typedef {`node:${string}`} NodeModuleName */
+
+/** @type {Record<NodeModuleName, () => any>} */
+const lazyLoaders = {
+  __proto__: null,
+  'node:crypto': () => require('node:crypto'),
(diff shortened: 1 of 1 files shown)


lib/util/runtime-features.js
```diff
@@ -1,16 +1,23 @@
...
  * @param {NodeModuleName} moduleName
  * @returns {boolean}
  */
 function detectRuntimeFeatureByNodeModule (moduleName) {
   try {
-    require(moduleName)
+    lazyLoaders[moduleName]()
     return true
   } catch (err) {
     if (err.code !== 'ERR_UNKNOWN_BUILTIN_MODULE') {
```
ShenHongFei, line 23: Is this err.code still the same after being packaged by the bundler?

Uzlopak: Yes of course.

Uzlopak: On second thought this actually irrelevant.

PR #4565, opened 2025-09-22
feat(interceptors/dns): cache option

SuperOleg39:
## Rationale

Custom `cache` for DNS interceptor allows:
- add custom metrics for DNS cache hit rate monitoring
- use any cache library (`lru-cache`, etc.)

Reference - https://github.com/szmarczak/cacheable-lookup?tab=readme-ov-file#cache

If these changes are acceptable, I will add unit test and validation for this parameter.

## Changes

Added `cache` parameter for `DNSInstance`

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/interceptor/dns.js
+++ b/lib/interceptor/dns.js
@@ -8,11 +8,11 @@ const maxInt = Math.pow(2, 31) - 1
 class DNSInstance {
   #maxTTL = 0
   #maxItems = 0
-  #records = new Map()
   dualStack = true
   affinity = null
   lookup = null
   pick = null
+  cache = null
 
   constructor (opts) {
     this.#maxTTL = opts.maxTTL
@@ -21,14 +21,15 @@ class DNSInstance {
     this.affinity = opts.affinity
     this.lookup = opts.lookup ?? this.#defaultLookup
     this.pick = opts.pick ?? this.#defaultPick
+    this.cache = opts.cache ?? new Map()
   }
 
   get full () {
-    return this.#records.size === this.#maxItems
+    return this.cache.size === this.#maxItems
   }
 
   runLookup (origin, opts, cb) {
-    const ips = this.#records.get(origin.hostname)
+    const ips = this.cache.get(origin.hostname)
 
     // If full, we just return the origin
     if (ips == null && this.full) {
@@ -55,7 +56,7 @@ class DNSInstance {
         }
 
         this.setRecords(origin, addresses)
-        const records = this.#records.get(origin.hostname)
+        const records = this.cache.get(origin.hostname)
 
         const ip = this.pick(
           origin,
@@ -89,7 +90,7 @@ class DNSInstance {
 
       // If no IPs we lookup - deleting old records
       if (ip == null) {
-        this.#records.delete(origin.hostname)
+        this.cache.delete(origin.hostname)
         this.runLookup(origin, opts, cb)
         return
       }
@@ -193,7 +194,7 @@ class DNSInstance {
   }
 
   pickFamily (origin, ipFamily) {
-    const records = this.#records.get(origin.hostname)?.records
+    const records = this.cache.get(origin.hostname)?.records
     if (!records) {
       return null
     }
@@ -242,11 +243,11 @@ class DNSInstance {
       records.records[record.family] = familyRecords
     }
 
-    this.#records.set(origin.hostname, records)
+    this.cache.set(origin.hostname, records)
   }
 
   deleteRecords (origin) {
-    this.#records.delete(origin.hostname)
+    this.cache.delete(origin.hostname)
   }
 
   getHandler (meta, opts) {


SuperOleg39:
Hi! Thanks for the quick response.

> I have questions: Why is cache now publicly accessible? Did you do this on purpose? Please explain.

I made `cache` public for a few reasons, but it's not that important:
- simpler debugging and monkeypatching if necessary
- fast access to cache if we need to call dump/load or other lru-cache methods - useful is dns interceptor instance will be saved as DI provider for example

> What happens if i pass an lru cache which handles ttl and max items by itself. How is `full` coming into play with this if the LRU cache is configured with other values?

At first, I think that cache ttl and size is a bit out of scope of interceptor, and in the future this logic can be incapsulated in some default cache implementation.

But for now, if `lru-cache` used and `cache` property will be added, I think `cache` need to be provided with `Infinity` ttls and max items.

> In runLookup we just ignore new entries if the Map is full. Isnt the point of having an LRU, to have the last recently used element be in the cache and not the first x elements till hitting the maxItems limit?

Yes, and because of that LRU cache can be more effective. But it is just an example, I'm not sure that applications have so many hosts for resolution that you need to worry about cache size.

DNS cache is more about balance between requests speed and failure rate because of outdated ips. So custom cache metrics can be really useful here to fine tuning.

> There is no definition of the shape of the cache. What if a `.get()` call returns `undefined` and not `null`?

Can we do here something without TypeScript? Maybe I can define interface with JSdoc

Uzlopak:
I think the interceptor was planned with integrated cache. Infinity as default values is not acceptable, because it would be a source for a memory leak.

I really think, that you need to either extract the caching logic properly or it is imho not acceptable.

metcoder95:
Agree with @Uzlopak in the term of extracting the `cache` implementation.

We need to set an explicit contract the passed `cache` should follow, and the `cache` should account for the TTL given in the DNS record (at least pass that information, and let the cache decide what to do with it).

Abstracting the current `cache` implementation will help you shape the contract and the interaction between the interceptor and the cache semantics.

> **Note**: what I mean with extracting, is to create a `Cache` abstraction and pass it as default cache if no cache is passed.

Uzlopak:
Maybe we should also avoid the word Cache. Maybe we should use the word Store? Or cache-store? :/

SuperOleg39:
100% agreed with cache extraction, will be back soon

SuperOleg39:
And another idea - we can just integrate diagnostic channel for monitoring purposes

SuperOleg39:
Demo with extracted cache - https://github.com/nodejs/undici/pull/4565/commits/66057186a564acb393316e3be1cb4be4d8eceb27

Extracted cache have so much logic to handle TTL at IP records level, and looking at it, I no longer think the decision is such a good idea.
Because possible `lru-cache` adapter, if we want the same behaviour, will duplicate this logic.

Maybe we need just a simple Map-like API for cache and use this cache exclusively as storage.

But this brings us back to decide how this will be combined with ttl at the lru-cache level - I think `cacheable-lookup` have exact the same problem, and they have [example](https://github.com/sindresorhus/quick-lru?tab=readme-ov-file#maxage) with QuickLRU library where default maxAge is [Infinity](https://github.com/sindresorhus/quick-lru?tab=readme-ov-file#maxage).

Also, `cacheable-lookup` can't prevent lookup if cache is full, and I can't solve this case and stay with Map-like API.

lib/interceptor/dns.js
```diff
@@ -5,6 +5,112 @@ const DecoratorHandler = require('../handler/decorator-handler')
...
+    this.#maxTTL = opts.maxTTL
+    this.#maxItems = opts.maxItems
+  }
+
+  get size () {
+    return this.#records.size
+  }
+
+  // TODO: it will require to write adapter for different caches, look for a better ideas
+  get full () {
```
SuperOleg39, line 23: Incompatible with js Map interface

lib/interceptor/dns.js
```diff
@@ -5,14 +5,120 @@ const DecoratorHandler = require('../handler/decorator-handler')
 const { InvalidArgumentError, InformationalError } = require('../core/errors')
 const maxInt = Math.pow(2, 31) - 1
 
-class DNSInstance {
+export class DNSCache {
```
metcoder95, line 8:
```suggestion
export class DNSStore {
```

As @Uzlopak suggestion 😛

SuperOleg39:
> > Maybe we need just a simple Map-like API for cache and use this cache exclusively as storage.
> 
> This was the exact suggestion from @Uzlopak, let's follow that path; and if seeking for compatibility with `lru-cache`-like solutions, we can hint the recommendations for making it compatible with the `DNSStorage` interface.

Thanks, I get it too late)

Is this solution will be more appropriate?

https://github.com/SuperOleg39/undici/commit/2c156e16a72a2deb0f2a3a955dd80e93ca116b56

lib/interceptor/dns.js
```diff
@@ -5,14 +5,120 @@ const DecoratorHandler = require('../handler/decorator-handler')
...
+  constructor (opts) {
+    this.#maxTTL = opts.maxTTL
+    this.#maxItems = opts.maxItems
+  }
+
+  get size () {
+    return this.#records.size
+  }
+
+  // TODO: it will require to write adapter for different caches, look for a better ideas
```
metcoder95, line 22: Shall we remove the `TODO`?

lib/interceptor/dns.js
```diff
@@ -5,14 +5,120 @@ const DecoratorHandler = require('../handler/decorator-handler')
...
+        if (typeof ip.ttl === 'number') {
+          // The record TTL is expected to be in ms
+          ip.ttl = Math.min(ip.ttl, this.#maxTTL)
+        } else {
+          ip.ttl = this.#maxTTL
+        }
+      }
+    }
+
+    // TODO: deduplicate logic
```
metcoder95, line 90: Shall we abstract it then?

lib/interceptor/dns.js
```diff
@@ -5,14 +5,120 @@ const DecoratorHandler = require('../handler/decorator-handler')
...
+      records.records[4].ips = ips.filter(ip => {
+        return Date.now() - ip.timestamp <= ip.ttl
+      })
+
+      if (records.records[4].ips.length === 0) {
+        records.records[4] = null
+      }
+    }
+
+    // TODO: deduplicate logic
```
metcoder95, line 51: ditto

lib/interceptor/dns.js
SuperOleg39: Before changes, can you please help me to choose between DNSStorage implementations:
- current implementation with incapsulated TTL logic inside storage (have a serious doubts against it as described in https://github.com/nodejs/undici/pull/4565#issuecomment-3328800144)
- simple implementation, only storage logic extracted - https://github.com/nodejs/undici/pull/4589/files

metcoder95: >simple implementation, only storage logic extracted - [#4589 (files)](https://github.com/nodejs/undici/pull/4589/files)

I'd prefer this one, is simpler and with the example you shared can be enough for implementers to adapt it to their needs

SuperOleg39:
Closed in favour of https://github.com/nodejs/undici/pull/4589

lib/interceptor/dns.js
SuperOleg39: Thans, https://github.com/nodejs/undici/pull/4589 is ready to review!

PR #4589, opened 2025-09-26
Feat dns interceptor storage

SuperOleg39:
## Rationale

Alternative implementation for https://github.com/nodejs/undici/pull/4565

Custom `cache` for DNS interceptor allows:
- add custom metrics for DNS cache hit rate monitoring
- use any cache library (`lru-cache`, etc.)

Reference - https://github.com/szmarczak/cacheable-lookup?tab=readme-ov-file#cache

## Changes

Added `cache` parameter for `DNSInstance`

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/docs/docs/api/Dispatcher.md
+++ b/docs/docs/api/Dispatcher.md
@@ -1043,7 +1043,7 @@ The `dns` interceptor enables you to cache DNS lookups for a given duration, per
   - The function should return a single record from the records array.
   - By default a simplified version of Round Robin is used.
   - The `records` property can be mutated to store the state of the balancing algorithm.
-- `storage: { size: number, get(origin: string): DNSInterceptorOriginRecords | null, set(origin: string, records: DNSInterceptorOriginRecords | null, options?: { ttl: number }): void, delete(origin: string): void, full(): boolean }` - Custom storage for resolved DNS records
+- `storage: DNSStorage` - Custom storage for resolved DNS records
 
 > The `Dispatcher#options` also gets extended with the options `dns.affinity`, `dns.dualStack`, `dns.lookup` and `dns.pick` which can be used to configure the interceptor at a request-per-request basis.
 
@@ -1057,6 +1057,15 @@ It represents a DNS record.
 It represents a map of DNS IP addresses records for a single origin.
 - `4.ips` - (`DNSInterceptorRecord[] | null`) The IPv4 addresses.
 - `6.ips` - (`DNSInterceptorRecord[] | null`) The IPv6 addresses.
+
+**DNSStorage**
+It represents a storage object for resolved DNS records.
+- `size` - (`number`) current size of the storage.
+- `get` - (`(origin: string) => DNSInterceptorOriginRecords | null`) method to get the records for a given origin.
+- `set` - (`(origin: string, records: DNSInterceptorOriginRecords | null, options: { ttl: number }) => void`) method to set the records for a given origin.
+- `delete` - (`(origin: string) => void`) method to delete records for a given origin.
+- `full` - (`() => boolean`) method to check if the storage is full, if returns `true`, DNS lookup will be skipped in this interceptor and new records will not be stored.
+```ts
 
 **Example - Basic DNS Interceptor**
 
--- a/docs/docs/api/Dispatcher.md
+++ b/docs/docs/api/Dispatcher.md
@@ -1043,6 +1043,7 @@ The `dns` interceptor enables you to cache DNS lookups for a given duration, per
   - The function should return a single record from the records array.
   - By default a simplified version of Round Robin is used.
   - The `records` property can be mutated to store the state of the balancing algorithm.
+- `storage: { size: number, get(origin: string): DNSInterceptorOriginRecords | null, set(origin: string, records: DNSInterceptorOriginRecords | null, options?: { ttl: number }): void, delete(origin: string): void, full(): boolean }` - Custom storage for resolved DNS records
 
 > The `Dispatcher#options` also gets extended with the options `dns.affinity`, `dns.dualStack`, `dns.lookup` and `dns.pick` which can be used to configure the interceptor at a request-per-request basis.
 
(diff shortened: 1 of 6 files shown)


lib/interceptor/dns.js
```diff
@@ -5,14 +5,44 @@ const DecoratorHandler = require('../handler/decorator-handler')
...
+
+  set (hostname, records) {
+    this.#records.set(hostname, records)
+  }
+
+  delete (hostname) {
+    this.#records.delete(hostname)
+  }
+
+  // Delegate to storage decide can we do more lookups or not
```
SuperOleg39, line 32: For example, lru-cache adapter can always return `false` here

types/interceptors.d.ts
```diff
@@ -19,14 +19,22 @@ declare namespace Interceptors {
 
   // DNS interceptor
   export type DNSInterceptorRecord = { address: string, ttl: number, family: 4 | 6 }
-  export type DNSInterceptorOriginRecords = { 4: { ips: DNSInterceptorRecord[] } | null, 6: { ips: DNSInterceptorRecord[] } | null }
+  export type DNSInterceptorOriginRecords = { records: { 4: { ips: DNSInterceptorRecord[] } | null, 6: { ips: DNSInterceptorRecord[] } | null } }
```
SuperOleg39, line 22: Here was incorrect typings - cached records shape is `{ records: ... }`

types/interceptors.d.ts
```diff
@@ -19,14 +19,22 @@ declare namespace Interceptors {
...
+    get(origin: string): DNSInterceptorOriginRecords | null
+    set(origin: string, records: DNSInterceptorOriginRecords | null, options?: { ttl: number }): void
+    delete(origin: string): void
+    full(): boolean
+  }
   export type DNSInterceptorOpts = {
     maxTTL?: number
     maxItems?: number
-    lookup?: (hostname: string, options: LookupOptions, callback: (err: NodeJS.ErrnoException | null, addresses: DNSInterceptorRecord[]) => void) => void
+    lookup?: (origin: URL, options: LookupOptions, callback: (err: NodeJS.ErrnoException | null, addresses: DNSInterceptorRecord[]) => void) => void
```
SuperOleg39, line 33: Here also was incorrect typings - before `pick` method calls, string origin is converted to URL

lib/interceptor/dns.js
```diff
@@ -5,14 +5,44 @@ const DecoratorHandler = require('../handler/decorator-handler')
 const { InvalidArgumentError, InformationalError } = require('../core/errors')
 const maxInt = Math.pow(2, 31) - 1
 
+class DNSStorage {
```
SuperOleg39, line 8: `lru-cache` adapter example:

```ts
const cache = new LRU({ max: maxItems })

const adapter = {
  set: (hostname: string, records: any, opts: { ttl: number }): void => {
    cache.set(hostname, records, opts);
  },
  get: (hostname: string): any => {
    return cache.get(hostname);
  },
  delete: (hostname: string) => {
    cache.delete(hostname);
  },
  full: (): boolean => {
    return false;
  },
};

const interceptor = interceptors.dns({ maxItems: Infinity, maxTTL: 60000, cache: adapter })
```

SuperOleg39: Also, if you will be interesting about lru-cache with integrated prometheus metrics, here is our implementation:
- adapter - https://github.com/tramvaijs/tramvai/blob/main/packages/modules/common/src/cache/cacheWithMetricsProxy.ts
- metrics - https://github.com/tramvaijs/tramvai/blob/main/packages/modules/common/src/cache/cacheMetrics.ts#L22

SuperOleg39: Grafana visualization:

Image

SuperOleg39: Cache metrics was a veeeery useful feature for us. But mostly for HTTP requests ofc)

lib/interceptor/dns.js
```diff
@@ -227,11 +254,13 @@ class DNSInstance {
   setRecords (origin, addresses) {
     const timestamp = Date.now()
     const records = { records: { 4: null, 6: null } }
+    let minTTL = this.#maxTTL
     for (const record of addresses) {
       record.timestamp = timestamp
       if (typeof record.ttl === 'number') {
         // The record TTL is expected to be in ms
         record.ttl = Math.min(record.ttl, this.#maxTTL)
+        minTTL = Math.min(minTTL, record.ttl)
```
SuperOleg39, line 263: For LRU or any other caches, which will not be able to respect ttl per record, just pass the minimum value from records.

Can't find better solution because `records` is a mutable object and we need to store the same for every origin,

docs/docs/api/Dispatcher.md
```diff
@@ -1043,6 +1043,7 @@ The `dns` interceptor enables you to cache DNS lookups for a given duration, per
   - The function should return a single record from the records array.
   - By default a simplified version of Round Robin is used.
   - The `records` property can be mutated to store the state of the balancing algorithm.
+- `storage: { size: number, get(origin: string): DNSInterceptorOriginRecords | null, set(origin: string, records: DNSInterceptorOriginRecords | null, options?: { ttl: number }): void, delete(origin: string): void, full(): boolean }` - Custom storage for resolved DNS records
```
metcoder95: Let's decompose it into its own section within the `DNS` interceptor itself

lib/interceptor/dns.js
metcoder95: It should be fine to keep it as is, this just duplicates it.

lib/interceptor/dns.js
```diff
@@ -242,11 +271,12 @@ class DNSInstance {
       records.records[record.family] = familyRecords
     }
 
-    this.#records.set(origin.hostname, records)
+    // We provide a default TTL if external storage will be used without TTL per record-level support
+    this.storage.set(origin.hostname, records, { ttl: minTTL })
```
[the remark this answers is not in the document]
metcoder95: The third argument is not used isn't it?

lib/interceptor/dns.js
SuperOleg39: The idea is to find minimum TTL for all resolved records for origin, and pass it to external storage, I don't quite understand about duplication

lib/interceptor/dns.js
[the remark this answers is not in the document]
SuperOleg39: It can be used in external storages, if lru-cache or quick-lru will be used under the hood

mcollina (approved):
lgtm

mcollina (commented):
Can you add test for the types?

SuperOleg39:
> Can you add test for the types?

Yep, ready

lib/interceptor/dns.js
metcoder95: ah ok see it now, missed that bit

metcoder95 (approved):

SuperOleg39:
Hello! I wanted to clarify when we can expect the release of these changes?

metcoder95:
cc. @mcollina

Issue #4588, opened 2025-09-25
Listen to Crawl-Delay by default


TechnologyClassroom:
## Bug Description

I see undici in my server logs. It check the robots.txt which is good, but then ignores the `Crawl-delay: 4`  line entirely. Can you fix this?

## Reproducible By

Looking at logs.

## Expected Behavior

I would expect a crawler written in good faith to follow the CrawlDelay by default.

## Logs & Screenshots

Here you can see undici requesting three times in one second twice in three seconds from one IP.

```
REDACTEDIP - - [24/Sep/2025:17:18:50 -0400] "GET /home.de.html HTTP/1.1" 200 11620 "-" "undici"
REDACTEDIP - - [24/Sep/2025:17:18:50 -0400] "GET /home.el.html HTTP/1.1" 200 13001 "-" "undici"
REDACTEDIP - - [24/Sep/2025:17:18:50 -0400] "GET /home.es.html HTTP/1.1" 200 11568 "-" "undici"
REDACTEDIP - - [24/Sep/2025:17:18:51 -0400] "GET /home.fa.html HTTP/1.1" 200 12060 "-" "undici"
REDACTEDIP - - [24/Sep/2025:17:18:52 -0400] "GET /home.fr.html HTTP/1.1" 200 11517 "-" "undici"
REDACTEDIP - - [24/Sep/2025:17:18:52 -0400] "GET /home.it.html HTTP/1.1" 200 11336 "-" "undici"
REDACTEDIP - - [24/Sep/2025:17:18:52 -0400] "GET /home.ja.html HTTP/1.1" 200 11927 "-" "undici"
```

## Environment

Server logs

### Additional context

This is not ok. I'm blocking this behavior with [reaction](https://framagit.org/ppom/reaction/).

KhafraDev:
undici is not a crawler

TechnologyClassroom:
Ok. Are you going to fix it?

Uzlopak:
No, because undici is not a crawler.

TechnologyClassroom:
I keep asking this question to a number of tools and I keep getting the same set of answers. We're just a tool. It isn't in RFC 9309.

Take some responsibility. There is no reason that I should be seeing those logs coming from the default behavior of your tool.

KhafraDev:
Maybe you keep seeing the same response because your reasoning is flawed.

mplscorwin:
[...]

PR #4590, opened 2025-09-26
docs: add crawling best practices

Uzlopak:
In the fallout of issue #4588 I was in the FSF IRC channel and discussed the issue and we discussed about potential solutions. cURL`s user agent is also blocked by many servers. The reason is, that bad implemented crawlers can be identified by it. 

I offered, that we can add a page in to the documentation regarding crawling and that we recommend to change the user agent to some random value, but that would not solve potential bots which ignore robots.txt.

It was further discussed, that doing a curl request against 'bgp.tools' `curl https://bgp.tools` results in following response:

```sh
Requests from default user agents are not allowed, please set a descriptive user-agent. For example: 'acmeco bgp.tools - contact@acme.co', This is so that if your program gets out of control I can contact you.

Setting a dishonest user agent will result in bans.
```

I used this information to write a reasonable documentation page regarding crawling. 

I used chatGPT and claude to refine the text.

I just added the test to ensure that my example in the doc is not a fantasy, lol.

Realistically seen, the default user agent can always be banned. We should anyway recommend to use a custom user agent. And instead of giving some recommendation like "generate something random as user agent" we give them some reasonable guidance. A malicious crawler will anyway use some fake user agent and a valid crawler will set a reasonable user agent.

### DIFF
--- a/docs/docs/best-practices/crawling.md
+++ b/docs/docs/best-practices/crawling.md
@@ -1,0 +1,56 @@
+# Crawling
+
+[RFC 9309](https://datatracker.ietf.org/doc/html/rfc9309) defines crawlers as automated clients.
+
+Some web servers may reject requests that omit the `User-Agent` header or that use common defaults such as `'curl/7.79.1'`.
+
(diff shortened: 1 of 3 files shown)


TechnologyClassroom:
I think this documentation is well written and good advice for people building crawlers.

metcoder95 (approved):

PR #4592, opened 2025-09-27
fix: set kBodyUsed only once

Uzlopak:
@metcoder95 wdyt? 
#3294

### DIFF
--- a/lib/core/util.js
+++ b/lib/core/util.js
@@ -6,7 +6,6 @@ const { IncomingMessage } = require('node:http')
 const stream = require('node:stream')
 const net = require('node:net')
 const { stringify } = require('node:querystring')
-const { EventEmitter: EE } = require('node:events')
 const timers = require('../util/timers')
 const { InvalidArgumentError, ConnectTimeoutError } = require('./errors')
 const { headerNameLowerCasedRecord } = require('./constants')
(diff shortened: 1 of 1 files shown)


Uzlopak:
@metcoder95 

There is no code coverage. And i guess we can remove this code, because https://nodejs.org/api/stream.html#readablereadabledidread is existing since v14.

lib/core/util.js
```diff
@@ -43,12 +42,10 @@ function wrapRequestBody (body) {
...
           assert(false)
         })
-    }
-
-    if (typeof body.readableDidRead !== 'boolean') {
+    } else if (typeof body.readableDidRead !== 'boolean') {
       body[kBodyUsed] = false
-      EE.prototype.on.call(body, 'data', function () {
-        this[kBodyUsed] = true
+      body.once('data', () => {
```
metcoder95, line 47: I believe the initial intention was to avoid tampering, but given that every stream is EE based, it shouldn’t be a problem (bad stream, bad implementation, up to caller).

I’m just more into keep it with `this` instead of creating another closure; the rest LGTM

ronag: This is wrong. It's not because of tampering but because otherwise the stream will start when registering a data listener.

ronag:
Streams will start the stream if on is called with `'data'` that's why the EE prototype is used.

Uzlopak:
Ok, but the thing is, that the code is not covered by unit tests.

ronag:
Would be nice to have that fixed.

gdanigarcia06-creator (changes requested):
Ayuda

PR #4612, opened 2025-10-05
Add tests for empty POST and content-length

fredericDelaporte:
## This relates to...

A discrepancy in behavior about empty POST request, when enabling or not enabling H2.
See #4594.

## Rationale

This adds two tests:
- [Fetch] Empty POST without h2 has Content-Length
- [Fetch] Empty POST with h2 has no Content-Length

According to [RFC 9110](https://httpwg.org/specs/rfc9110.html#field.content-length):

> A user agent SHOULD send Content-Length in a request when the method defines a meaning for enclosed content and it is not sending [Transfer-Encoding](https://httpwg.org/specs/rfc9112.html#field.transfer-encoding). For example, a user agent normally sends Content-Length in a POST request even when the value is 0 (indicating empty content).

So, the behavior of Undici sending a `0` `Content-Length` on empty POST request when H2 is not enabled is recommended by the RFC, while its behavior of sending no Content-Length on such requests when H2 is enabled does not match the recommendation. (It is still compliant since that is a "SHOULD", not a "MUST".)

Whether this should be changed or not, I do not know. Is there a good reason for this behavior with H2?

For your information, some servers cannot receive empty POST requests without a Content-Length. That is the case of Internet Information Server (IIS), the default web server of Windows OSes. Windows OSes underlying `http.sys` [rejects such requests](https://stackoverflow.com/a/61627100/1178314) with a [411 Length required](https://httpwg.org/specs/rfc9110.html#status.411) response.

## Changes

Add two tests to demonstrate the current behavior of Undici.

### Features

N/A

### Bug Fixes

N/A

None

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/test/fetch/http2.js
+++ b/test/fetch/http2.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const { createSecureServer } = require('node:http2')
+const { createServer } = require('node:http')
 const { createReadStream, readFileSync } = require('node:fs')
 const { once } = require('node:events')
 const { Readable } = require('node:stream')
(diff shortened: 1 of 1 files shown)


fredericDelaporte:
I have not checked other verbs which semantics imply a content, like PUT, PATCH. If changing this behavior about POST, they should likely be checked too.

fredericDelaporte:
Replaced by #4613. (Thanks @mcollina.)

PR #4613, opened 2025-10-05
fix: ensure HTTP/2 sends Content-Length for empty POST requests

mcollina:
## This relates to...

Fixes #4594
Related to #4612

## Rationale

This PR addresses the discrepancy in behavior between HTTP/1 and HTTP/2 when handling empty POST requests.

According to [RFC 9110](https://httpwg.org/specs/rfc9110.html#field.content-length):

> A user agent SHOULD send Content-Length in a request when the method defines a meaning for enclosed content and it is not sending Transfer-Encoding. For example, a user agent normally sends Content-Length in a POST request even when the value is 0 (indicating empty content).

Previously:
- **HTTP/1**: Empty POST requests correctly included `Content-Length: 0`
- **HTTP/2**: Empty POST requests omitted the `Content-Length` header entirely

This inconsistency causes compatibility issues with servers like Internet Information Server (IIS), which reject empty POST requests without a Content-Length header, returning `411 Length Required`.

### Implementation (`lib/dispatcher/client-h2.js`)
- Modified the logic to only omit `Content-Length` when both `contentLength === 0` AND the method doesn't expect a payload (like GET, HEAD)
- For methods that expect payloads (POST, PUT, PATCH, etc.), `Content-Length: 0` is now sent even when the body is empty
- Updated the assertion to allow `contentLength === 0` without a body

### Tests (`test/fetch/http2.js`)
- Updated the test to verify that empty POST requests with HTTP/2 now include `Content-Length: 0`
- Updated test name and comments to reflect the corrected behavior
- Both HTTP/1 and HTTP/2 tests now expect consistent behavior

## Test Results

All HTTP/2 tests pass (51 tests total):
- ✅ `[Fetch] Empty POST without h2 has Content-Length`
- ✅ `[Fetch] Empty POST with h2 has Content-Length`

None. This is a bug fix that aligns behavior with RFC 9110 recommendations and HTTP/1 implementation.

## Status

- [x] Tested
- [x] Documented
- [x] Review ready

### DIFF
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -429,15 +429,12 @@ function writeH2 (client, request) {
     contentLength = request.contentLength
   }
 
-  if (contentLength === 0 && !expectsPayload) {
+  if (!expectsPayload) {
     // https://tools.ietf.org/html/rfc7230#section-3.3.2
     // A user agent SHOULD NOT send a Content-Length header field when
     // the request message does not contain a payload body and the method
     // semantics do not anticipate such a body.
-
-    contentLength = null
-  } else if (!expectsPayload) {
-    // For methods that don't expect a payload, omit Content-Length
+    // And for methods that don't expect a payload, omit Content-Length.
     contentLength = null
   }
 


lib/dispatcher/client-h2.js
```diff
@@ -429,12 +429,15 @@ function writeH2 (client, request) {
...
     // https://tools.ietf.org/html/rfc7230#section-3.3.2
     // A user agent SHOULD NOT send a Content-Length header field when
     // the request message does not contain a payload body and the method
     // semantics do not anticipate such a body.
 
+    contentLength = null
+  } else if (!expectsPayload) {
+    // For methods that don't expect a payload, omit Content-Length
     contentLength = null
   }
```
fredericDelaporte: It seems to me it is equivalent to:

```suggestion
  if (!expectsPayload) {
    // https://tools.ietf.org/html/rfc7230#section-3.3.2
    // A user agent SHOULD NOT send a Content-Length header field when
    // the request message does not contain a payload body and the method
    // semantics do not anticipate such a body.
    // And for methods that don't expect a payload, omit Content-Length.
    contentLength = null
  }
```

But I wonder, is-it really adequate to remove Content-Length even when there is some content on methods not expected to have some? Though changing that would be an additional change.

Some tools like [elastic-search do use GET request with a body](https://www.elastic.co/docs/reference/elasticsearch/rest-apis/api-conventions#get-requests)...

So, it seems safer to do only this:

```suggestion
  if (contentLength === 0 && !expectsPayload) {
    // https://tools.ietf.org/html/rfc7230#section-3.3.2
    // A user agent SHOULD NOT send a Content-Length header field when
    // the request message does not contain a payload body and the method
    // semantics do not anticipate such a body.
    contentLength = null
  }
```

metcoder95: We should not support body for method that does not have defined semantics for servers to have bodies (e.g. `GET`).

The best is attach to the parts of the spec that are clear so we can get a standardized behaviour

metcoder95 (approved):

metcoder95:
CI failures are unrelated

lib/dispatcher/client-h2.js
fredericDelaporte: If you are waiting for further feedback on my side for merging, well, I just do not understand the later reply.

The spec seems clear enough to me, as the extract put in comment is:

> A user agent SHOULD NOT send a Content-Length header field when the request message does not contain a payload body and the method semantics do not anticipate such a body.

From that extract, we deduce we should send the `Content-Length` header when the method semantics do anticipate a body, even if there is no body. That is further backed by an example given right after in the spec.

But the complete "boolean inversion" of the sentence is, by the logic rules:

> A user agent SHOULD send a Content-Length header field when the request message has a payload body or the method semantics do anticipate such a body.

So why actively removing that header when there is a payload but on a method not expecting one? That deviates from the client spec, from my viewpoint. The fact that a payload on methods not expecting one should not have any meaning on the server side is a server subject, not a client one.

If willing to actively prevent support of a payload in a GET or similar request, the client should ignore the entire payload, not just sending no content-length header even if there is a payload. (Or maybe even throw an error, rather than silently doing some stuff in order to not support it.)

But anyway, I have currently no interest in the ability to send a payload on GET or the likes. It felt just curious to me to do something neither mandated nor advised by the spec and seeming rendering the request inconsistent (still having a payload, but without a content-length anymore). I have not checked what was the actual behavior for this case, with or without H2, with or without this change, and I will not check.

I just wanted to make sure this was deliberate. So, even if I do not understand your rationale here, or course, do as you see fit.

mcollina:
@metcoder95 can I get a fresh approval?

Issue #4596, opened 2025-09-30
Stale-while-revalidate background revalidation does not lead to updated cache entries upon receiving 304 not modified.


daan944:
## Bug Description
Although async behavior was added in https://github.com/nodejs/undici/pull/4492 (thank you for that 👍 ), the CacheHandler does not know what to do with `304 not modified` responses and ignores them. This causes the next request to be hit the backend again, instead of the updated cache.

See interceptor/cache.js#259 and handler/cache-handler.js#254 (`NOT_UNDERSTOOD_STATUS_CODES` includes 304).

## Reproducible By

1. Configure undici with cache interceptor.
2. Have a backend that replies with a 304 when `if-none-match` and/or `if-modified-since` headers are present. The reply must include cache headers, e.g. `max-age=5, stale-while-revalidate=3600`.
3. Perform a request with empty cache: see 200 OK and cache entry is inserted.
4. Wait 5s (so response is stale) and perform a new request.
5. Cache handler is responding immediately with stale data (good!) and starts revalidation. 
6. Revalidation is being answered by backend with `304 not modified`.
7. Perform a new request
8. See request hitting backend server, not using cache (I think). Backend server replies with 200 OK again and it is being cached again.

In my Varnish logs I see requests answered with 304 and upon next request requested again and answered with 200 (which implies this is not a revalidation request).

## Expected Behavior

When revalidation request is being answered with a `304 not modified` and cache headers, we expect the cache expiry to be updated for this request.

### Additional context

I think that we should use the CacheRevalidationHandler in the cache interceptor (instead of the CacheHandler) and update the cache upon succesful revalidation. Similar to what is happening when we're within the `stale-if-error` period, see interceptor/cache.js#325.

metcoder95:
Thanks for the report! Would you like to send a PR for it?

daan944:
I see 2 routes to fix this: 

1. re-write the cache revalidation handler. Upon reading the code, I think there's more amiss there. I suspect a 304 from the backend within stale-if-error period will also result in a stale warning. And I don't think there's a cache update at that moment. I haven't had the time to doublecheck these suspicions. I think this is the preferred approach, but the most time consuming.
[...]

PR #4617, opened 2025-10-08
fix: `304 not modified` reply upon revalidation did not update cache.

daan944:
## This relates to...
https://github.com/nodejs/undici/issues/4596

## Rationale
When a 304 was returned during revalidation, the cache was not updated to reflect this. Instead, the cached value was kept as-is, eventually expiring. This is particularly a problem for anyone using stale-while-revalidate, as in the entire stale-while-revalidate window there'll be no updates while loading the backend with (async) revalidations.

To handle a 304, the cached body and headers will be re-used and re-inserted into cache. This is not the most optimal way, but the only way I feel is possible without breaking changes. A better way would be to require support for an `update` method in the cacheStore to only update the headers/metadata and keep body as-is. I didn't want to go there without discussing it first.

## Changes
1. `304 not modified` replies will update the cache with new headers, cache times etc., but will re-use original statuscode/message, etag and body. (see changes in lib/handler/cache-handler.js)
2. Manual re-validation possible by performing a request with `if-modified-since` or `if-none-match`. These will trigger a synchronous revalidation flow.
3. Removed two tests (`stale responses are revalidated before deleteAt (if-modified-since)` and `stale responses are revalidated before deleteAt (if-none-match)` from `test/interceptors/cache.js` as these did not contain the actual expected behavior. Instead I've added `revalidates the request, handles 304s during stale-while-revalidate` in `test/interceptors/cache-revalidate-stale.js`.

### Bug Fixes

See https://github.com/nodejs/undici/issues/4596

None

## Status

- [x] Tested
- [s] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/handler/cache-handler.js
+++ b/lib/handler/cache-handler.js
@@ -17,11 +17,11 @@ const HEURISTICALLY_CACHEABLE_STATUS_CODES = [
 
 // Status codes which semantic is not handled by the cache
 // https://datatracker.ietf.org/doc/html/rfc9111#section-3
-// This list should not grow beyond 206 and 304 unless the RFC is updated
+// This list should not grow beyond 206 unless the RFC is updated
 // by a newer one including more. Please introduce another list if
 // implementing caching of responses with the 'must-understand' directive.
 const NOT_UNDERSTOOD_STATUS_CODES = [
-  206, 304
+  206
 ]
 
 const MAX_RESPONSE_AGE = 2147483647000
@@ -104,6 +104,7 @@ class CacheHandler {
         resHeaders,
         statusMessage
       )
+    const handler = this
 
     if (
       !util.safeHTTPMethods.includes(this.#cacheKey.method) &&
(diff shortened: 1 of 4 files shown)


test/interceptors/cache.js
```diff
@@ -1888,4 +1609,81 @@ describe('Cache Interceptor', () => {
       equal(requestsToOrigin, 1) // Still only one origin request
     }
   })
+
+  // this test describes the expected behavior, but it currently fails.
+  test.skip('stale-while-revalidate updates cache after background revalidation', async () => {
```
mcollina: I'm a bit lost on this, can you articulate?

lib/handler/cache-handler.js
```diff
@@ -189,36 +190,94 @@ class CacheHandler {
...
+      const streamCachedBody = () => {
+        const { value, done } = bodyIterator.next()
+
+        if (!done) {
+          const full = this.#writeStream.write(value) === false
+          this.#handler.onResponseData?.(controller, value)
+          // when stream is full stop writing until we get a 'drain' event
+          if (!full) {
+            streamCachedBody()
+          }
```
metcoder95: I'd prefer to seek for a loop rather than recursion

test/interceptors/cache.js
daan944: Sorry! This was part of a previous attempt that escaped my cleanup before opening this PR. I'll remove it.

lib/handler/cache-handler.js
daan944: You're right, that's cleaner. Fixed.

test/interceptors/cache.js
```diff
@@ -387,285 +387,6 @@ describe('Cache Interceptor', () => {
     }
   })
 
-  test('stale responses are revalidated before deleteAt (if-modified-since)', async () => {
```
mcollina, line 390: Why was this deleted?

daan944: I feel this is covered by the other tests already (incl the ones I've added).

Also, on line 491 and 627 there's conditional requests that behave different now: they update the cache too. The scenario in these tests are thus now not entirely valid anymore.

If you prefer I could modify and reinstate them.

I do feel the organisation of the cache/revalidation tests can be improved, as we have both these tests and the ones in test/interceptors/cache-revalidate-stale.js - shouldn't we merge these files?

mcollina:
@metcoder95 ptal

Issue #4627, opened 2025-10-17
SSE connection may leak when fetch init.redirect = 'error'


liheng314:
## Bug Description

Fetch abort may not take effect when fetch init.redirect = 'error' and this causes SSE connection leak

## Reproducible By
SSE Server:
```
const express = require('express');
const app = express();
const PORT = 3000;

class SSEServer {
  constructor() {
    this.clients = new Set();
  }

  // 添加客户端
  addClient(res) {
    this.clients.add(res);
    
    // 设置 SSE 头
    res.writeHead(200, {
      'Content-Type': 'text/event-stream',
      'Cache-Control': 'no-cache',
      'Connection': 'keep-alive',
      'Access-Control-Allow-Origin': '*'
    });

    // 发送欢迎消息
    this.sendToClient(res, 'connected', { message: 'Welcome to SSE server' });

    // 客户端断开时清理
    res.on('close', () => {
      this.clients.delete(res);
      console.log('Client disconnected. Total clients:', this.clients.size);
    });
  }

  // 向单个客户端发送消息
  sendToClient(res, event, data) {
    res.write(`event: ${event}\n`);
    res.write(`data: ${JSON.stringify(data)}\n\n`);
  }

  // 向所有客户端广播消息
  broadcast(event, data) {
    this.clients.forEach(client => {
      this.sendToClient(client, event, data);
    });
  }

  // 获取客户端数量
  getClientCount() {
    return this.clients.size;
  }
}

// 创建 SSE 服务器实例
const sseServer = new SSEServer();

// 中间件
app.use(express.json());
app.use(express.static('public'));

// SSE 端点
app.get('/sse', (req, res) => {
  sseServer.addClient(res);
  console.log('New client connected. Total clients:', sseServer.getClientCount());
});

// 手动触发事件端点（可选）
app.post('/send-message', (req, res) => {
  const { message } = req.body;
  sseServer.broadcast('custom-message', {
    message: message || 'Manual broadcast',
    timestamp: new Date().toISOString()
  });
  res.json({ success: true, clients: sseServer.getClientCount() });
});

// 启动定时广播
let eventCount = 1;
setInterval(() => {
  const eventData = {
    id: eventCount,
    timestamp: new Date().toISOString(),
    message: `Automatic event #${eventCount}`,
    random: Math.random().toFixed(4),
    clients: sseServer.getClientCount()
  };
  
  sseServer.broadcast('message', eventData);
[...]

PR #4629, opened 2025-10-18
fix: prevent connection leak when redirect='error'

mcollina:
Fixes issue where SSE and other long-lived connections could leak when using fetch with redirect: 'error' option. The spec optimization of reusing fetchParams when window is 'no-window' and redirect is 'error' was causing connection lifecycle issues.

In Node.js, request.window is always set to 'no-window', so this optimization was being applied in all cases where redirect: 'error' was used. This caused problems with connection cleanup, particularly for EventSource and SSE use cases where the connection needs to be properly aborted.

The fix disables this optimization and always clones the request and copies fetchParams to ensure proper connection lifecycle management.

Fixes: #4627

### DIFF
--- a/lib/web/fetch/index.js
+++ b/lib/web/fetch/index.js
@@ -1380,24 +1380,26 @@ async function httpNetworkOrCacheFetch (
 
   // 8. Run these steps, but abort when the ongoing fetch is terminated:
 
-  //    1. If request’s window is "no-window" and request’s redirect mode is
+  //    1. If request's window is "no-window" and request's redirect mode is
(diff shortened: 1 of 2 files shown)


KhafraDev (changes requested):
I find it highly unlikely that this is an issue in undici. The `eventsource` package on npm creates its own signal and doesn't seem to work with a custom one (when providing your own fetch function), hence it's never aborted. I'd need a smaller repro to be certain. Also, if it is a spec issue, we should upstream a fix, or at the very least a note.

PR #4630, opened 2025-10-18
fix: prevent duplicate debug logs when multiple undici instances exist

### LINKED ISSUE #4615: bug: debug logs duplicated
## Bug Description

When launching debug mode, all events are written to logs 2-3 times.
Moreover, if you subscribe to the channel yourself, you will receive one event.

I understood that this is due to the fact that we subscribe to the event twice if we use undici, this is in undici inside the node and in undici as a dependency.

## Reproducible By

```bash
NODE_DEBUG=undici node ./debug.js
```

```ts
import { request, Agent } from 'undici';

const dispatcher = new Agent({ connections: 1 });

async function execute() {
  const resp = await request('https://cdn.datatables.net/2.3.4/js/dataTables.min.js', { dispatcher });
  await resp.body.dump();
}

execute();
```

```ts
async function undici() {
  const { fetch} = await import('undici');
  const resp = await fetch('https://cdn.datatables.net/2.3.4/js/dataTables.min.js');
  await resp.arrayBuffer();
}

async function nodeUndici(){
  const resp = await fetch('https://cdn.datatables.net/2.3.4/js/dataTables.min.js');
  await resp.arrayBuffer();
}

// one debug logs
nodeUndici();

// duplicated debug logs
// undici();
```

## Expected Behavior

Logs for a single query should not be duplicated.

## Logs & Screenshots

Image

## Environment

node - v24.9.0 and v22.19.0
undici - 7.16.0

mcollina:
## Summary

Fixes #4615 - Prevents duplicate debug logs when both Node.js built-in undici and undici as a dependency are present.

## Problem

When using `NODE_DEBUG=undici`, debug logs were being duplicated 2-3 times because both Node.js's built-in undici and undici installed as a dependency would subscribe to the same diagnostics channels. Each module instance would independently subscribe, causing the same event to be logged multiple times.

## Solution

Added `hasSubscribers` checks in the diagnostics channel tracking functions before subscribing:
- `trackClientEvents()` - checks client-related channels
- `trackRequestEvents()` - checks request-related channels  
- `trackWebSocketEvents()` - checks websocket-related channels

This ensures that even when multiple module instances exist, only the first instance subscribes to the diagnostics channels, preventing duplicate logs.

## Test Plan

- Added regression test `debug#undici no duplicates` that reproduces the issue with multiple module loads
- All existing debug tests pass (4/4)
- All diagnostics channel tests pass (6/6)
- All websocket diagnostics tests pass (3/3)
- Linting passes

### DIFF
--- a/lib/core/diagnostics.js
+++ b/lib/core/diagnostics.js
@@ -36,6 +36,14 @@ function trackClientEvents (debugLog = undiciDebugLog) {
     return
   }
 
+  // Check if any of the channels already have subscribers to prevent duplicate subscriptions
+  // This can happen when both Node.js built-in undici and undici as a dependency are present
+  if (channels.beforeConnect.hasSubscribers || channels.connected.hasSubscribers ||
+      channels.connectError.hasSubscribers || channels.sendHeaders.hasSubscribers) {
+    isTrackingClientEvents = true
+    return
+  }
+
   isTrackingClientEvents = true
 
   diagnosticsChannel.subscribe('undici:client:beforeConnect',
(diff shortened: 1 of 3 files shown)


Uzlopak (commented):
I dont like this solution, as it is not possible to determine if the listener is actually a debug listener or some other listener.

I already thought about the problem multiple times I think the actual solution has to be something like using a `Symbol.for('undici,
-diagnostic-channel-debug')` or what i prefer even more finegrained `Symbol.for('diagnostic-channel-debug-websocket')` and using it as a key on globalThis and persist if we already activated specified channels.

metcoder95 (approved):

PR #4636, opened 2025-10-21
fix: prevent abort listener leak in fetch with long-lived AbortController

mcollina:
## Summary

Fixes the flaky test `test/fetch/long-lived-abort-controller.js` by ensuring abort listeners are cleaned up deterministically instead of relying on garbage collection.

## Problem

The test was failing ~60% of the time with `MaxListenersExceededWarning` because:
- Abort listeners added in both `fetch()` and the `Request` constructor were not being removed explicitly
- The code relied on `FinalizationRegistry` to clean up listeners via garbage collection
- When making 2000+ requests rapidly with the same `AbortController`, listeners accumulated faster than GC could clean them up

## Solution

- Store the cleanup function returned by `addAbortListener()` in both locations
- Call cleanup functions immediately when fetch completes in `processResponseEndOfBody` 
- Also clean up on error paths in `processResponse`

## Testing

- Verified the flaky test now passes 100/100 consecutive runs (previously failed ~20/30)
- All fetch-related tests pass including `abort.js`, `issue-1711.js`, and `request.js`
- No regressions in existing functionality

### DIFF
--- a/test/http2-dispatcher.js
+++ b/test/http2-dispatcher.js
@@ -156,11 +156,7 @@ test('Dispatcher#Connect', async t => {
       })
 
       stream.respond({ ':status': 200, 'x-my-header': response.headers['x-my-header'] })
-      pipeline(response.body, stream, (err) => {
-        if (err) {
-          stream.destroy(err)
-        }
-      })
+      pipeline(response.body, stream, () => {})
     } catch (err) {
       stream.destroy(err)
     }


mcollina:
@KhafraDev this is needed because I saw that kind of problem in multiple CI runs and locally. It's very sporadic on my machine but shows reliably on CI. https://github.com/nodejs/undici/actions/runs/18676765035/job/53248303527#step:12:2644

I don't think there is another way vs cleaning it up.

lib/web/fetch/request.js
```diff
@@ -436,7 +437,7 @@ class Request {
           setMaxListeners(1500, signal)
         }
 
-        util.addAbortListener(signal, abort)
+        this[kRemoveAbortListener] = util.addAbortListener(signal, abort)
```
KhafraDev, line 440: We should avoid using symbols.

KhafraDev (changes requested):
Is it a leak or is the gc just not running in a loop? The test doesn't really seem valid to me. There is also a test failing in WPT.

mcollina:
Ahum, let me verify if the gc is not running on Mac anymore for whatever reason.

The problem is that that test start failing on every CI run.

mcollina:
There are no test changes. I just want CI to pass.

KhafraDev:
> There are no test changes. I just want CI to pass.

WPTs don't fail if expectations change (we cover too many node versions and flakiness combined with my lack of free time). There is a test failing here that doesn't fail on main.

PR #4638, opened 2025-10-21
Run the gc() in long-lived-abort-controller test

mcollina:
Alternative to https://github.com/nodejs/undici/pull/4636.


KhafraDev (approved):
This makes more sense to me.

PR #4639, opened 2025-10-22
fix(wasm): shell command built from environment values

odaysec:
The issue occurs because the shell command is constructed as a string containing environment-derived values (`WASM_SRC`, etc.) and then executed via `execSync`, which invokes the shell and interprets special characters. To fix the issue, we should avoid shell interpretation of dynamic arguments:

- Replace the use of `execSync` with `execFileSync`. This API directly executes binaries and takes command-line arguments as an array, avoiding shell interpolation and interpretation of special characters.
- Build the command and its arguments as an array, separating fixed command strings from variable paths and flags to ensure no shell metacharacters can interfere.
- Apply this fix to the block starting from line 99 where the problematic command is built and executed.
- Default `WASM_CC` may be a compiler (`clang`), so the command should be formatted as:  
  `WASM_CC [WASM_CFLAGS components..., "-msimd128", ...WASM_LDFLAGS components..., src files, "-Iinclude", "-o", output wasm filename, ...WASM_LDLIBS components...]`
- We will need to split environment variable flags into arrays instead of interpolating strings (unless they're static).
- All environment-derived values should be split and sanitized: e.g., handle `WASM_CFLAGS`, `WASM_LDFLAGS`, `WASM_LDLIBS` as arrays (splitting on whitespace).
- No new imports are needed; only additional variable handling and replacement of the exec call.

## Status

- [ ] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [x] Merge ready

### DIFF
--- a/build/wasm.js
+++ b/build/wasm.js
@@ -96,11 +96,19 @@ if (hasOptimizer) {
 writeWasmChunk('llhttp.wasm', 'llhttp-wasm.js')
 
 // Build wasm simd binary
-execSync(`${WASM_CC} ${WASM_CFLAGS} -msimd128 ${WASM_LDFLAGS} \
-${join(WASM_SRC, 'src')}/*.c \
--I${join(WASM_SRC, 'include')} \
--o ${join(WASM_OUT, 'llhttp_simd.wasm')} \
-${WASM_LDLIBS}`, { stdio: 'inherit' })
+// Split flags into arrays for safe argument passing
+const wasmCcArgsSimd = [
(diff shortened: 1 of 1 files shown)


Uzlopak (changes requested):
Claims to solve an issue, but it is not clear what the issue is. This script is alone "worthless" and is only  used to build the wasm files when we update llhttp every 1-2 years. 

I have a proposed PR which ensures the integrity of the wasm files, which is more relevant than ensuring   that you sanitize the env variables in this script.

Uzlopak:
Closing due to inactivity

PR #4640, opened 2025-10-22
Fix http2-dispatcher test

mcollina:


mcollina:
@Uzlopak @metcoder95 this is ready

Issue #3648, opened 2024-09-26
Pool distributes load unevenly across connections.


JustinTRoss:
## Bug Description

Pool distributes load unevenly across connections, prioritizing them by creation time. See [here](https://github.com/nodejs/undici/blob/e6e87c13dbf8788fe158fd60631cee78eb84f198/lib/dispatcher/pool.js#L75-L87).

Where your connections are load-balanced across instances (pods/containers), this results in overuse of some instances and underuse of others.

## Reproducible By

1. Create a service, `ServiceA`, with an endpoint that takes 100ms - 200ms to respond.
2. Using Kubernetes, deploy 5 pods with instances of `ServiceA`.
3. In another service, `ServiceB`, create a Pool of connections to `ServiceA`, referencing it by Kubernetes Service address _(https://service-name)_, such that we'll be load-balancing connections through kube-proxy.
4. Send 1 request to `ServiceB` every 50ms for 1 minute.
5. Observe balanced connections across `ServiceA` pods, but imbalanced request distribution and resource usage across `ServiceA` pods.

Thinking through, this may be otherwise observable plainly through internal diagnostics around amount of requests processed by each Client or something like that.

## Expected Behavior

Roughly balanced request distribution and resource usage across Clients and `ServiceA` pods.

## Environment

MacOS, Node v18

### Additional context

There are a number of ways we could resolve this intelligently. Most simply/crudely though, we could track index of last used client and pick back up next scan from that index to roughly round-robin.

Something like:

```
...
const totalClients = clients.length;
let clientIndex = (this.lastUsedIndex + 1) % totalClients;

for (let i = 0; i < totalClients; i++) {
  const client = clients[clientIndex];

  if (!client[kNeedDrain]) {
    this.lastUsedIndex = clientIndex;
    return client
  }

  clientIndex = (clientIndex + 1) % totalClients;
}
```

Uzlopak:
Imho duplicate of #310

mcollina:
The problem stems from the use of Keep-Alive connections. Essentially, we prioritize "alive"  connections as much as possible because they are the ones we know are lively and less likely to timeout/error due to network disconnections. This is suboptimal in case of low traffic but minimizes errors. In a high-traffic scenario, this is equivalent to round-robin.

Have you tried using the `BalancedPool`? It should be implementing a better algorithm that would work better for your use case.

JustinTRoss:
Thanks for the quick response.

[...]

PR #4650, opened 2025-11-01
Feat/round robin pool

FelixVaughan:
## Rationale
Addresses #3648

Pool distributes load unevenly across connections when used with load-balanced endpoints (e.g., Kubernetes Services). It always selects the first available connection, causing some backend servers to be overused while others remain idle.

RoundRobinPool solves this by cycling through connections in a round-robin fashion, ensuring even distribution of requests across all backend servers.

## Changes

`Pool.js` distribution
pool

`RoundRobinPool.js` distribution
rr-pool

### Features

  - New `RoundRobinPool` class(`lib/dispatcher/round-robin-pool.js`)
  - Documentation(`docs/docs/api/RoundRobinPool.md`)
  - Tests
  - TS definitions
 
## Status

- [x] Tested (10 comprehensive tests including distribution verification)
- [x] Benchmarked (optional) - Maintains same performance characteristics as Pool
- [x] Documented (API docs, examples, and use case guidance)
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/.github/workflows/autobahn.yml
+++ b/.github/workflows/autobahn.yml
@@ -44,7 +44,7 @@ jobs:
           args: docker restart --time 0 --signal=SIGKILL fuzzingserver
 
       - name: Setup Node.js@${{ inputs.node-version }}
-        uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # v5.0.0
+        uses: actions/setup-node@2028fbc5c25fe9cf00d9f06a71cc4710d4507903 # v6.0.0
         with:
           node-version: ${{ inputs.node-version }}
 
--- a/.github/workflows/bench.yml
+++ b/.github/workflows/bench.yml
@@ -23,7 +23,7 @@ jobs:
           persist-credentials: false
           ref: ${{ github.base_ref }}
       - name: Setup Node
-        uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # v5.0.0
+        uses: actions/setup-node@2028fbc5c25fe9cf00d9f06a71cc4710d4507903 # v6.0.0
         with:
           node-version: 'lts/*'
       - name: Install Modules for undici
@@ -44,7 +44,7 @@ jobs:
         with:
           persist-credentials: false
       - name: Setup Node
-        uses: actions/setup-node@a0853c24544627f65ddf259abe73b1d18a591444 # v5.0.0
+        uses: actions/setup-node@2028fbc5c25fe9cf00d9f06a71cc4710d4507903 # v6.0.0
         with:
           node-version: 'lts/*'
       - name: Install Modules for undici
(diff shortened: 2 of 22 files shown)


test/round-robin-pool.js
```diff
@@ -0,0 +1,304 @@
...
+
+  try {
+    new RoundRobinPool('http://localhost', { factory: '' }) // eslint-disable-line
+  } catch (err) {
+    p.ok(err instanceof errors.InvalidArgumentError)
+    p.strictEqual(err.message, 'factory must be a function.')
+  }
+})
+
+test('basic get', async (t) => {
```
metcoder95, line 46: As this implementation relies on the RoundRobin algorithm implemented, I'd suggest that tests also accounts for that

lib/dispatcher/round-robin-pool.js
```diff
@@ -0,0 +1,137 @@
...
+const kOptions = Symbol('options')
+const kConnections = Symbol('connections')
+const kFactory = Symbol('factory')
+const kIndex = Symbol('index')
+
+function defaultFactory (origin, opts) {
+  return new Client(origin, opts)
+}
+
+class RoundRobinPool extends PoolBase {
```
metcoder95, line 28: I wonder if we can also provide a similar approach to `BalancedPool` to account for faulty upstreams (Clients) that are misbehaving and possibly re-balance distribution accordingly to implementer options

mcollina:
The reason the default is to always reuse the last use connection is to avoid the inherent risk of keep-alive race conditions.

IMHO, this should take into account the keep alive timeout of the individual connection, and just skip "old" ones that are most likely to cause an ECONNRESET. If a connection is over 80% of its keepalive, it's probably better to leave it alone unless if there is a different option. This could also be handled in the client itself as it would benefit all pools tbh. Without that change, users of this would see an increase of ECONNRESET errors.

I'm not sure how you guarantee that each connection actually goes to a different server. This seems unlikely, as it depends on the underneath load balancer algorithm.

metcoder95:
>This could also be handled in the client itself as it would benefit all pools tbh. 

Do you refer to the `PoolBase` isn't it?
I'm afraid that changing that behavior can break contracts that callers already rely on.

>IMHO, this should take into account the keep alive timeout of the individual connection, and just skip "old" ones that are most likely to cause an ECONNRESET. If a connection is over 80% of its keepalive, it's probably better to leave it alone unless if there is a different option.

What option do you have in mind?

The proposal SGTM 👍

mcollina:
@metcoder95 nevermind, we already have that logic: https://github.com/nodejs/undici/blob/6301265a20868d077faae6d51f5f6cf57ac2ebfe/lib/dispatcher/client-h1.js#L592.

test/round-robin-pool.js
FelixVaughan: Added `verifies round-robin kGetDispatcher cycling algorithm` to complement the `round-robin wraps around correctly` test

lib/dispatcher/round-robin-pool.js
FelixVaughan: Plus one for this. Would be happy to revisit and see what we can do

FelixVaughan:
You're correct, we only guarantee client-side connection distribution. 

Backend distribution depends on how the load balancer assigns TCP connections to backends. I initially tested this with a homespun TCP LB which assigned each new connection to backends in round-robin fashion, but if the LB assigns using session affinity for example, there's not a whole lot we can do. 

I added documentation explaining the caveat. Also open to other suggestions.

metcoder95 (approved):
lgtm

PR #4656, opened 2025-11-03
don't freeze urlList for opaque filtered responses

### LINKED ISSUE #4647: `fetch()` hangs with redirec and `no-cors`
## Bug Description

The promise returned by `fetch(redirectedURL, { mode: 'no-cors' })` never resolves.

## Reproducible By

```js
import http from 'http'

const a = http.createServer((req, res) => {
    console.log(req.url)
    res.writeHead(200).end()
}).listen(8080)

const b = http.createServer((req, res) => {
    res.writeHead(301, { Location: `http://localhost:8080${req.url}` }).end()
}).listen(8081);

(async () => {
  // Remove `{ mode: 'no-cors' }` and the promise resolves
  const response = await fetch('http://localhost:8081/abc', { mode: 'no-cors' })
  console.log(response.status)
  a.close()
  b.close()
})()
```

## Expected Behavior

I expect the promise to resolve, assuming I have not missed anything in the Fetch specification that requires the current behavior.

## Environment

Debian v12.12 / Node v24.10 / undici v7.16.0

KhafraDev:
fixes #4647 

the offending line was added in 9e5316c8b04a7b35522d0d5b8de71f67fa2a3c34 but I don't see any rationale for why urlList was frozen, the spec doesn't mention freezing the list.

the issue was caused by freezing the list which would prevent us from pushing to it which would swallow the error

### DIFF
--- a/test/fetch/issue-4647.js
+++ b/test/fetch/issue-4647.js
@@ -1,0 +1,23 @@
+'use strict'
+
+const { createServer } = require('node:http')
+const { test } = require('node:test')
+const { fetch } = require('../..')
+
+// https://github.com/nodejs/undici/issues/4647
+test('fetch with mode: no-cors does not hang', async (t) => {
+  const a = createServer((req, res) => {
+    res.writeHead(200).end()
+  }).listen(0)
+
+  const b = createServer((req, res) => {
+    res.writeHead(301, { Location: `http://localhost:${a.address().port}${req.url}` }).end()
+  }).listen(0)
+
+  t.after(() => {
+    a.close()
+    b.close()
+  })
+
+  await fetch(`http://localhost:${b.address().port}/abc`, { mode: 'no-cors' })
+})


tsctx:
I believe it was wrong to classify status 0 as a network error, but I'm unsure which way is correct.

```diff
diff --git a/lib/web/fetch/index.js b/lib/web/fetch/index.js
index 0f0a94d3..da58a0dc 100644
--- a/lib/web/fetch/index.js
+++ b/lib/web/fetch/index.js
@@ -646,7 +646,7 @@ async function mainFetch (fetchParams, recursive) {
 
     // 13. If response is not a network error and response is not a filtered
     // response, then:
-    if (response.status !== 0 && !response.internalResponse) {
+    if (response.type !== 'error' && !response.internalResponse) {
       // If request’s response tainting is "cors", then:
       if (request.responseTainting === 'cors') {
         // 1. Let headerNames be the result of extracting header list values
@@ -676,8 +676,7 @@ async function mainFetch (fetchParams, recursive) {
 
     // 14. Let internalResponse be response, if response is a network error,
     // and response’s internal response otherwise.
-    let internalResponse =
-      response.status === 0 ? response : response.internalResponse
+    let internalResponse = response.type === 'error' ? response : (response.internalResponse ?? response)
 
     // 15. If internalResponse’s URL list is empty, then set it to a clone of
     // request’s URL list.
@@ -717,7 +716,7 @@ async function mainFetch (fetchParams, recursive) {
     // set internalResponse’s body to null and disregard any enqueuing toward
     // it (if any).
     if (
-      response.status !== 0 &&
+      response.type !== 'error' &&
       (request.method === 'HEAD' ||
         request.method === 'CONNECT' ||
         nullBodyStatus.includes(internalResponse.status))

```

KhafraDev:
both are correct

KhafraDev:
> In other words, an [opaque filtered response](https://fetch.spec.whatwg.org/#concept-filtered-response-opaque) and an [opaque-redirect filtered response](https://fetch.spec.whatwg.org/#concept-filtered-response-opaque-redirect) are nearly indistinguishable from a [network error](https://fetch.spec.whatwg.org/#concept-network-error). When introducing new APIs, do not use the [internal response](https://fetch.spec.whatwg.org/#concept-internal-response) for internal specification algorithms as that will leak information.

> A network error is a [response](https://fetch.spec.whatwg.org/#concept-response) whose [type](https://fetch.spec.whatwg.org/#concept-response-type) is "error", [status](https://fetch.spec.whatwg.org/#concept-response-status) is 0, [status message](https://fetch.spec.whatwg.org/#concept-response-status-message) is the empty byte sequence, [header list](https://fetch.spec.whatwg.org/#concept-response-header-list) is « », [body](https://fetch.spec.whatwg.org/#concept-response-body) is null, and [body info](https://fetch.spec.whatwg.org/#concept-response-body-info) is a new [response body info](https://fetch.spec.whatwg.org/#response-body-info).

I did notice what you are saying when working on it, but it's not relevant to the bug so I didn't bother. We are *technically* detecting network errors incorrectly, but if it's not causing bugs... 🤷

Issue #4661, opened 2025-11-07
Allow form-data with filename and filename*


FuJa0815:
## Bug Description

Currently this request cannot be parsed:

```
--1df6c75e-c5a7-486c-af47-67b632b19522
Content-Disposition: form-data; name="Abc"; filename="E 1962 029 1342-0003.JPG"; filename*=utf-8''E%201962%20029%201342-0003.JPG

Hello
--1df6c75e-c5a7-486c-af47-67b632b19522--
```

The error is `TypeError: expected CRLF`. Reading through the code passing a filename and a filename* is not supported. This is up to spec tho and something the C# implementation generates.

KhafraDev:
Now that it seems to be fixed in undici, this is an issue in the C# implementation. According to [RFC 7578](https://www.rfc-editor.org/rfc/rfc7578#section-4.2):

> NOTE: The encoding method described in [[RFC5987](https://www.rfc-editor.org/rfc/rfc5987)], which would add a
   "filename*" parameter to the Content-Disposition header field, MUST
   NOT be used.

While receivers (such as undici) should still parse it correctly, senders shouldn't be adding it.

FuJa0815:
I'm actually not sure on how to read this, this seems to be in conflict with https://datatracker.ietf.org/doc/html/rfc6266

Or is this a case of "well they must not do it but if somebody still does this should happen"?

KhafraDev:
RFC 6266, I believe, is referring to `content-disposition` in http responses, not multipart/form-data bodies. At least, that's how I'm interpreting it.

KhafraDev:
Before this closes, although from my limited testing it seems to be fixed, I would like to add a few tests.
- A test that explicitly ensures filename* is parsed correctly
- A test that ensures that without filename, ~~there is an error~~ parsing still succeeds
- A test that ensures filename* takes priority over filename when both are present

I've been busy and haven't gotten around to it

PR #4662, opened 2025-11-08
fix fd parsing unquoted attribute values

### LINKED ISSUE #4660: Allow form-data name without quotes
## Bug Description

Currently in a multipart form request the following cannot be parsed:

```
--7427efb8-6ce7-4740-9198-d90399842641
Content-Type: text/plain; charset=utf-8
Content-Disposition: form-data; name=test

abc
--7427efb8-6ce7-4740-9198-d90399842641--
```

with this error: `TypeError: expected form-data; name=" for content-disposition header`.

This is a valid request though that the C# MultipartFormDataContent also creates.

## Expected Behavior

This request should be handled

KhafraDev:
fixes #4660 

claude did ~half of this PR so be wary

### DIFF
--- a/lib/web/fetch/formdata-parser.js
+++ b/lib/web/fetch/formdata-parser.js
@@ -11,6 +11,8 @@ const { utf8DecodeBytes } = require('../../encoding')
 const dd = Buffer.from('--')
 const ddcrlf = Buffer.from('--\r\n')
 
+const decoder = new TextDecoder()
+
 /**
  * @param {string} chars
  */
@@ -277,7 +279,7 @@ function parseContentDispositionAttribute (input, position) {
     }
 
     // Skip utf-8'' and decode the rest
-    value = decodeURIComponent(new TextDecoder().decode(headerValue.subarray(7)))
+    value = decodeURIComponent(decoder.decode(headerValue.subarray(7)))
   } else if (input[position.position] === 0x22 /* " */) {
     // Quoted string
     position.position++ // skip opening quote
@@ -293,7 +295,7 @@ function parseContentDispositionAttribute (input, position) {
     }
     position.position++ // skip closing quote
 
-    value = new TextDecoder().decode(quotedValue)
+    value = decoder.decode(quotedValue)
       .replace(/%0A/ig, '\n')
       .replace(/%0D/ig, '\r')
       .replace(/%22/g, '"')
@@ -305,7 +307,7 @@ function parseContentDispositionAttribute (input, position) {
       position
     )
 
-    value = new TextDecoder().decode(tokenValue)
+    value = decoder.decode(tokenValue)
   }
 
   return { name: attrNameStr, value }


lib/web/fetch/formdata-parser.js
```diff
@@ -206,6 +204,113 @@ function multipartFormDataParser (input, mimeType) {
...
+      (headerValue[1] !== 0x74 && headerValue[1] !== 0x54) || // t or T
+      (headerValue[2] !== 0x66 && headerValue[2] !== 0x46) || // f or F
+      headerValue[3] !== 0x2d || // -
+      headerValue[4] !== 0x38 // 8
+    ) {
+      throw parsingError('unknown encoding, expected utf-8\'\'')
+    }
+
+    // Skip utf-8'' and decode the rest
+    value = decodeURIComponent(new TextDecoder().decode(headerValue.subarray(7)))
```
Uzlopak: cant we cache this TextDecoder?

lib/web/fetch/formdata-parser.js
```diff
@@ -206,6 +204,113 @@ function multipartFormDataParser (input, mimeType) {
...
+
+    if (input[position.position] !== 0x22) {
+      throw parsingError('Closing quote not found')
+    }
+    position.position++ // skip closing quote
+
+    value = new TextDecoder().decode(quotedValue)
+      .replace(/%0A/ig, '\n')
+      .replace(/%0D/ig, '\r')
+      .replace(/%22/g, '"')
```
Uzlopak: nasty regex stuff. Cant we cache the decoder and maybe check if the value contains a '%' before running those replacers?

lib/web/fetch/formdata-parser.js
```diff
@@ -493,6 +521,58 @@ function parsingError (cause) {
...
+    char === 0x3d    // +
+  )
+}
+
+/**
+ * token := 1*<any (US-ASCII) CHAR except SPACE, CTLs,
+ *          or tspecials>
+ * @param {number} char
+ */
+function isToken (char) {
```
Uzlopak, line 568: dont we have such a function already?

lib/web/fetch/formdata-parser.js
KhafraDev: this was copy and pasted from the old code

lib/web/fetch/formdata-parser.js
KhafraDev: not one that's 100% equivalent, if you're talking about the cookie parser one

lib/web/fetch/formdata-parser.js
```diff
@@ -8,11 +8,11 @@ const assert = require('node:assert')
 const { isomorphicDecode } = require('../infra')
 const { utf8DecodeBytes } = require('../../encoding')
 
-const formDataNameBuffer = Buffer.from('form-data; name="')
-const filenameBuffer = Buffer.from('filename')
 const dd = Buffer.from('--')
 const ddcrlf = Buffer.from('--\r\n')
 
+const decoder = new TextDecoder()
```
mcollina, line 14: Why this is global?

lib/web/fetch/formdata-parser.js
```diff
@@ -206,6 +206,113 @@ function multipartFormDataParser (input, mimeType) {
...
+
+    if (input[position.position] !== 0x22) {
+      throw parsingError('Closing quote not found')
+    }
+    position.position++ // skip closing quote
+
+    value = decoder.decode(quotedValue)
+      .replace(/%0A/ig, '\n')
+      .replace(/%0D/ig, '\r')
+      .replace(/%22/g, '"')
```
mcollina, line 301: Can you have Claude optimize this and just do a single loop?

lib/web/fetch/formdata-parser.js
KhafraDev: https://github.com/nodejs/undici/pull/4662#discussion_r2505948012

lib/web/fetch/formdata-parser.js
KhafraDev: this is the same code as before, copied from another function

KhafraDev:
Can we land this...? It's a bug fix. I don't know why I'm being asked to microoptimize everything.

KhafraDev:
this seems to have fixed #4661 as well, great

Issue #4663, opened 2025-11-10
SnapshotAgent in playback mode errors when it encounters an excluded url


LeoMcA:
## Bug Description

When excluding a url from a "record" run of SnapshotAgent, encountering that same url in "playback" mode throws an error: despite the url being ignored.

This means it's not possible to e.g. record requests from an app being tested on one run of an integration test, and play them back on another run, while excluding the requests to the app being tested (they'll raise an error on the playback run).

## Reproducible By

```js
import { SnapshotAgent, setGlobalDispatcher } from "undici";

const recordingAgent = new SnapshotAgent({
  mode: "record",
  snapshotPath: "./snapshot.json",
  excludeUrls: [`https://example.com/2`],
});
setGlobalDispatcher(recordingAgent);

await fetch("https://example.com/1");
await fetch("https://example.com/2");

await recordingAgent.saveSnapshots();

const playbackAgent = new SnapshotAgent({
  mode: "playback",
  snapshotPath: "./snapshot.json",
  excludeUrls: [`https://example.com/2`],
});
setGlobalDispatcher(playbackAgent);

await fetch("https://example.com/1");
await fetch("https://example.com/2");
```

## Expected Behavior

`SnapshotAgent` should allow the request to pass unmocked, without raising an error.

The issue seems to stem from `findSnapshot` returning `undefined` for both excluded/ignored requests *and* missing requests:

https://github.com/nodejs/undici/blob/f29471cf6460a53997f2e947c2f6b30536b682fe/lib/mock/snapshot-recorder.js#L342-L350

https://github.com/nodejs/undici/blob/f29471cf6460a53997f2e947c2f6b30536b682fe/lib/mock/snapshot-recorder.js#L356

So in `SnapshotAgent` it's not possible to distinguish between the two:

https://github.com/nodejs/undici/blob/f29471cf6460a53997f2e947c2f6b30536b682fe/lib/mock/snapshot-agent.js#L99-L106

## Logs & Screenshots

```
node:internal/deps/undici/undici:13510
      Error.captureStackTrace(err);
            ^

TypeError: fetch failed
    at node:internal/deps/undici/undici:13510:13
    at async file:///home/leo/Projects/mdn/undici-snapshot-bug/index.js:23:1 {
  [cause]: UndiciError: No snapshot found for GET /2
      at SnapshotAgent.dispatch (/home/leo/Projects/mdn/undici-snapshot-bug/node_modules/undici/lib/mock/snapshot-agent.js:101:23)
[...]

PR #4670, opened 2025-11-14
fix: snapshot url exclusion

FelixVaughan:
## Rationale

Addresses #4663

## Changes/Fixes

- Check URL exclusions before snapshot lookup in `SnapshotAgent.dispatch()`
- Instantiate real HTTP agent in playback mode when `excludeUrls` is configured
- Route excluded URLs to real agent instead of returning "No snapshot found" error
- Expose `SnapshotRecorder.isUrlExcluded()` for early URL exclusion checks
- Add test coverage for playback mode with excluded URLs

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [x] In review
- [ ] Merge ready


lib/mock/snapshot-agent.js
```diff
@@ -80,6 +82,20 @@ class SnapshotAgent extends MockAgent {
     handler = WrapHandler.wrap(handler)
     const mode = this[kSnapshotMode]
 
+    // Check if URL should be excluded (pass through without mocking/recording)
+    if (this[kSnapshotRecorder].isUrlExcluded(opts)) {
+      if (this[kRealAgent]) {
+        return this[kRealAgent].dispatch(opts, handler)
+      }
+      // If no real agent in playback mode, this is a configuration error
+      const error = new UndiciError('Cannot pass through excluded URL: no real agent available')
```
metcoder95: I'm not sure we should throw, maybe just go with no-op?

FelixVaughan: Sure, my thought was since this would likely be because of misconfiguration, we would want to fail loudly rather than have the request swallowed. Thoughts?

Yeah I noticed those as well. My thinking was these were unrelated but can we confirm?

metcoder95: For a configuration error, it will be better to throw at the moment the agent is being instantiated, as the configuration mismatch should be identified and advertised early, usually when at runtime is already late to take further action.

If we can fail early, the better, otherwise a warning will be better.

FelixVaughan: Added some constructor logic in an earlier commit ensuring an agent is created if needed so we should be good. Removed the error block

mcollina:
tests are still not happy

FelixVaughan:
@mcollina from what I can see, the remaining tests seem unrelated

mcollina (approved):
lgtm

PR #4672, opened 2025-11-16
Formdata ignore epilogue preamble

### LINKED ISSUE #4671: Non-compliant FormData parser: Doesn't ignore the "preamble" or the "epilogue"
### Version

v25.2.0

### Platform

```text
Darwin SilverMini 25.1.0 Darwin Kernel Version 25.1.0: Mon Oct 20 19:32:47 PDT 2025; root:xnu-12377.41.6~2/RELEASE_ARM64_T8103 arm64
```

### Subsystem

_No response_

Execute: `node formdata.js`
```js
// formdata.js
var request = new Request("https://example.com", {
  method: "POST",
  body: (function () {
    formData.append("a", "b");
    return formData;
  })(),
});
var contentType = request.headers.get("Content-Type");
console.log(contentType);

var bytes = await request.bytes();
test(bytes, 1);

bytes = new Uint8Array(bytes.buffer.transfer(bytes.length + 10));
test(bytes, 2);

bytes.set(bytes.subarray(0, -10), 10);
bytes.fill(0, 0, 8);
bytes[8] = 13;
bytes[9] = 10;
test(bytes, 3);

async function test(buffer, x) {
  await new Response(buffer, {
    headers: {
      "Content-Type": contentType,
    },
  }).formData();
  console.log(x);
}
```

Every time.

The code creates a FormData in the form of a Uint8Array then tests the FormData parser:
1. To see if it can reconstruct it unmutilated.
2. To see if it can reconstruct it with an "epilogue".
3. and, to see if it can reconstruct it with a "preamble".

The expected behaviour is that all three tests pass. According to [RFC2046 Section 5.1.1](https://datatracker.ietf.org/doc/html/rfc2046#section-5.1.1)
> [There is] room for additional information prior to the first boundary delimiter line and following the final boundary delimiter line. [...] implementations must ignore anything that appears before the first boundary delimiter line or after the last one.

### What do you see instead?

Tests 1 passes, but tests 2 and 3 fail. It doesn't seem to be able to handle a "preamble" or an "epilogue".

### Additional information

Chrome, Safari, Deno and Bun all seem to be able to handle a "preamble" and an "epilogue". Firefox seems to fail at handling a "preamble" but can handle an "epilogue".

KhafraDev:
fixes #4671 

Claude generated everything minus the tests

### DIFF
--- a/test/busboy/issue-4671.js
+++ b/test/busboy/issue-4671.js
@@ -16,7 +16,7 @@ test('preamble and epilogue is ignored', async (t) => {
 
   const contentType = request.headers.get('Content-Type')
   let bytes = await request.bytes()
-  bytes = new Uint8Array(bytes.buffer.transfer(bytes.length + 10))
+  bytes = new Uint8Array([...bytes, ...Array(10).fill(0)])
 
   await t.test('epilogue', async () => {
     await new Response(bytes, {


KhafraDev:
🦗

tsctx (approved):

Issue #4068, opened 2025-02-27
Cloned requests loose abort signal on GC


andreas-karlsson:
## Bug Description

We noticed that `AbortSignal.timeout()` stopped working in our library with longer timeouts (in our case around 9s). Debugging showed that cloned (or constructor copied) requests won't respect an abort signal if GC has happened in between.

## Reproducible By
Run with `node --expose-gc`
```js
import { createServer } from 'node:http';

createServer(() => {
  // server that just hangs
}).listen(3000);

const ac = new AbortController();

let req = new Request('http://localhost:3000', {
  signal: ac.signal,
});

// cloned or copied requests (req = new Request(req)) breaks abort after gc
req = req.clone();

setTimeout(() => {
  // comment out the gc to make abort work
  global.gc();
  ac.abort();
  console.log('aborted, expecting immediate exception from fetch...');
});

await fetch(req);
```
Commenting out the cloning and/or the gc will make the code work as expected, i.e. the fetch call rejecting after abort.

## Expected Behavior

A cloned request should respect it's abort signal even after a long time (or garbage collection).

## Environment
macOS 15.3.1 (arm64), Node v23.4.0, v22.14.0, v20.18.0

### Additional context

Debugging found that the difference before/after gc is that the "dependentControllerMap" entry for the signal is missing [here](https://github.com/nodejs/node/blob/269c8512409ce88e6af6f5f7dbaca0cb773bfbd7/deps/undici/src/lib/web/fetch/request.js#L60)

KhafraDev:
> Debugging found that the difference before/after gc is that the "dependentControllerMap" entry for the signal is missing [here](https://github.com/nodejs/node/blob/269c8512409ce88e6af6f5f7dbaca0cb773bfbd7/deps/undici/src/lib/web/fetch/request.js#L60)

This looks familiar to another issue with fetch/signals caused by the AbortSignal implementation in node core but I couldn't find the issue. Should this be posted in node core instead? Is there anything we can do to fix this in undici?

PR #4678, opened 2025-11-27
fix: ac lost on cloned request (#4068)

gyx1000:
## Changes

This PR fixes an issue where cloning a `Request` caused the cloned instance to lose its `AbortSignal` behavior under certain GC conditions.  
The underlying problem was that the `signal` reference could be garbage-collected prematurely, preventing the abort event from propagating to the cloned request. This resulted in inconsistent abort behavior on the very first request processed after a clone.

The fix ensures the `signal` and its abort listener remain properly retained across clones, restoring consistent abort propagation for cloned `Request` objects.

### Bug Fixes

- Fixed a bug where `Request.clone()` or `new Request(request)`, could drop the associated `AbortSignal` listener during garbage collection.
- Ensured cloned `Request` objects consistently propagate abort events, including during the initial request lifecycle.

### DIFF
--- a/lib/web/fetch/request.js
+++ b/lib/web/fetch/request.js
@@ -29,6 +29,7 @@ const assert = require('node:assert')
 const { getMaxListeners, setMaxListeners, defaultMaxListeners } = require('node:events')
 
 const kAbortController = Symbol('abortController')
+const kDependantAbortController = Symbol('dependantAbortController')
 
 const requestFinalizer = new FinalizationRegistry(({ signal, abort }) => {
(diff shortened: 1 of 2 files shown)


KhafraDev (changes requested):
My comment in the original issue makes it seem like this is not an issue in undici.

tsctx:
> My comment in the original issue makes it seem like this is not an issue in undici.

It's undici that is the problem, not node.

Once the parent signal is garbage collected, the child loses its reference and therefore cannot propagate the abort signal, even if the original signal is aborted.

PR #4679, opened 2025-11-28
feat: add deduplicate interceptor for request deduplication

mcollina:
## Summary

Adds a new `deduplicate` interceptor for request deduplication. When enabled, concurrent identical requests are deduplicated so only one request is sent to the origin server, and all waiting handlers receive the same response.

### Changes

- **New file**: `lib/interceptor/deduplicate.js` - Standalone interceptor for request deduplication
- **New file**: `lib/handler/deduplication-handler.js` - Handler that buffers responses for multiple waiting handlers
- **New file**: `test/interceptors/deduplicate.js` - Comprehensive test suite (16 tests)
- **Modified**: `lib/util/cache.js` - Added `makeDeduplicationKey()` function
- **Modified**: `index.js` - Export new `deduplicate` interceptor
- **Modified**: `types/interceptors.d.ts` - Added types for deduplicate interceptor
- **Modified**: `docs/docs/api/Dispatcher.md` - Documented the feature
- **Modified**: `docs/docs/api/DiagnosticsChannel.md` - Documented `undici:request:pending-requests` channel

### Usage

```js
const { Client, interceptors } = require("undici");
const { deduplicate, cache } = interceptors;

// Deduplicate only
  deduplicate()
);

// Deduplicate with caching
const clientWithCache = new Client("http://example.com").compose(
  deduplicate(),
  cache()
);
```

### Options

- `methods` - The safe HTTP methods to deduplicate. Default `['GET']`.

### Observability

A diagnostic channel `undici:request:pending-requests` is available for monitoring:

```js
const diagnosticsChannel = require('node:diagnostics_channel');

diagnosticsChannel.channel('undici:request:pending-requests').subscribe(({ type, size, key }) => {
  console.log(type);  // 'added' or 'removed'
  console.log(size);  // current number of pending requests
  console.log(key);   // the deduplication key
});
```

## Test plan

- [x] All existing cache tests pass (39 tests)
- [x] New deduplication tests pass (16 tests)
- [x] Tests cover: basic deduplication, header handling, different paths, error propagation, cache integration, chunked bodies, diagnostic channel events, Authorization/Cookie header differentiation
- [x] Lint passes

### DIFF
--- a/lib/util/cache.js
+++ b/lib/util/cache.js
@@ -368,9 +368,10 @@ function assertCacheMethods (methods, name = 'CacheMethods') {
  * Creates a string key for request deduplication purposes.
  * This key is used to identify in-flight requests that can be shared.
  * @param {import('../../types/cache-interceptor.d.ts').default.CacheKey} cacheKey
+ * @param {Set<string>} [excludeHeaders] Set of lowercase header names to exclude from the key
  * @returns {string}
  */
-function makeDeduplicationKey (cacheKey) {
+function makeDeduplicationKey (cacheKey, excludeHeaders) {
   // Create a deterministic string key from the cache key
   // Include origin, method, path, and sorted headers
   let key = `${cacheKey.origin}:${cacheKey.method}:${cacheKey.path}`
@@ -378,6 +379,10 @@ function makeDeduplicationKey (cacheKey) {
   if (cacheKey.headers) {
     const sortedHeaders = Object.keys(cacheKey.headers).sort()
     for (const header of sortedHeaders) {
+      // Skip excluded headers
+      if (excludeHeaders?.has(header.toLowerCase())) {
+        continue
+      }
       const value = cacheKey.headers[header]
       key += `:${header}=${Array.isArray(value) ? value.join(',') : value}`
     }
--- a/docs/docs/api/Dispatcher.md
+++ b/docs/docs/api/Dispatcher.md
@@ -1221,6 +1221,8 @@ The `deduplicate` interceptor deduplicates concurrent identical requests. When m
 **Options**
 
 - `methods` - The [**safe** HTTP methods](https://www.rfc-editor.org/rfc/rfc9110#section-9.2.1) to deduplicate. Default `['GET']`.
+- `skipHeaderNames` - Header names that, if present in a request, will cause the request to skip deduplication entirely. Useful for headers like `idempotency-key` where presence indicates unique processing. Header name matching is case-insensitive. Default `[]`.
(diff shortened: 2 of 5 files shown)


mcollina:
I had claude opus 4.5 prepare this. Looks good enough but a second opionion would be interesting.

I opted to not put the dedupe burden on the cache storage. This means that in a distributed cache scenario, you’ll
see n invocations, one per event loop.

lib/util/cache.js
```diff
@@ -364,6 +364,28 @@ function assertCacheMethods (methods, name = 'CacheMethods') {
...
+  // Include origin, method, path, and sorted headers
+  let key = `${cacheKey.origin}:${cacheKey.method}:${cacheKey.path}`
+
+  if (cacheKey.headers) {
+    const sortedHeaders = Object.keys(cacheKey.headers).sort()
+    for (const header of sortedHeaders) {
+      const value = cacheKey.headers[header]
+      key += `:${header}=${Array.isArray(value) ? value.join(',') : value}`
+    }
+  }
```
metcoder95, line 389: Cannot find the place where we strip sensitive headers (e.g. `authorization`), it might be good to remove it from plain text to be used for deduplication key.

lib/interceptor/cache.js
```diff
@@ -399,6 +444,12 @@ module.exports = (opts = {}) => {
 
   const safeMethodsToNotCache = util.safeHTTPMethods.filter(method => methods.includes(method) === false)
 
+  /**
+   * Map of pending requests for deduplication (null if deduplication is disabled)
+   * @type {Map<string, CacheDeduplicationHandler> | null}
+   */
+  const pendingRequests = deduplication ? new Map() : null
```
metcoder95: Shall we worried that it can go out of bounds?

mcollina:
I wonder if a better approach would be to have a completely separate interceptor instead.

metcoder95:
But doesn't this depends on the cache interceptor?

mcollina:
@metcoder95 not much tbh.

metcoder95:
Then yeah, better to have it as a separated interceptor

metcoder95 (commented):
lgtm, just left a comment about potentially using sensitive headers as part of the plain text key

lib/util/cache.js
mcollina: why would you remove them?

metcoder95: mostly to avoid keeping it in memory plain text, but happy to keep it as is if we feel no possible harm can be done

mcollina:
if we didn't include sensitive headers, then requests with different Authorization would be deduplicated, causing security incidents.

I'll add an option to not deduplicate requests with certain headers.

metcoder95 (approved):

gurgunday (approved):
lgtm

PR #4704, opened 2025-12-17
fix: update Content-Type header when body is re-extracted on redirect

### LINKED ISSUE #4065: Undici fails to update `Content-Type` header after following redirects `multipart/form-data` requests
## Bug Description

There is a problem when sending a `multipart/form-data` request that follows redirects. After the redirect, `undici` overrides the request body but does not update the `Content-Type` header, which causes an error when parsing the body. The updated content type just ignored [here](https://github.com/nodejs/undici/blob/608d5f6683b70576aa8dc0fb87a3514f370c28d5/lib/web/fetch/index.js#L1318).

## Reproducible By

```js
const { fetch, FormData } = require("undici");

createServer((req, res) => {
  if (req.method === "POST" && req.url === "/first") {
    res.writeHead(307, {
      Location: "http://localhost:3000/second",
    });
    res.end();
  } else if (req.method === "POST" && req.url === "/second") {
    console.log("Request headers:\n", req.headers);

    let body = "";
    req.on("data", (chunk) => {
      body += chunk;
    });

    req.on("end", () => {
      console.log("Request body:\n", body);
      res.writeHead(200);
      res.end("Request received");
    });
  }
}).listen(3000);

// Make a request to the first endpoint using undici
async function makeRequest() {
  formData.append("test", "data");

  const response = await fetch("http://localhost:3000/first", {
    method: "POST",
    body: formData,
    redirect: "follow",
  });

  console.log("Response status:\n", response.status);
}

makeRequest().catch(console.error);
```

## Expected Behavior

The `Content-Type` header should be updated along with the body after a redirect.

## Logs & Screenshots

```
Request headers:
 {
  host: 'localhost:3000',
  connection: 'keep-alive',
  'content-type': 'multipart/form-data; boundary=----formdata-undici-073413599880',
  accept: '*/*',
  'accept-language': '*',
  'sec-fetch-mode': 'cors',
  'user-agent': 'undici',
  'accept-encoding': 'gzip, deflate',
  'content-length': '127'
}
Request body:
 ------formdata-undici-045095077684
Content-Disposition: form-data; name="test"

data
------formdata-undici-045095077684--
```

brunorodmoreira:
## This relates to...

Fixes #4065

## Rationale

When a `multipart/form-data` POST request encounters a redirect (e.g., 307), the body is re-extracted via `safelyExtractBody()`, which generates a **new boundary**. The original code only captured the body `[0]` and discarded the new `Content-Type` `[1]`, leaving the header with the stale boundary.

This fix captures the Content-Type returned by `safelyExtractBody()` and updates the request headers accordingly, ensuring the boundary in the header matches the boundary in the body.

### Features

N/A

### Bug Fixes

- Update `Content-Type` header when body is re-extracted during redirect handling, ensuring the boundary in the header matches the boundary in the body.

N/A

## Status

- [x] Tested
- [S] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/web/fetch/index.js
+++ b/lib/web/fetch/index.js
@@ -1322,7 +1322,11 @@ function httpRedirectFetch (fetchParams, response) {
   // value of safely extracting request’s body’s source.
   if (request.body != null) {
     assert(request.body.source != null)
-    request.body = safelyExtractBody(request.body.source)[0]
+    const [body, contentType] = safelyExtractBody(request.body.source)
+    request.body = body
+    if (contentType) {
+      request.headersList.set('content-type', contentType, true)
+    }
   }
 
   // 15. Let timingInfo be fetchParams’s timing info.
--- a/test/fetch/redirect.js
+++ b/test/fetch/redirect.js
@@ -3,7 +3,7 @@
 const { test } = require('node:test')
 const { createServer } = require('node:http')
 const { once } = require('node:events')
-const { fetch } = require('../..')
+const { fetch, FormData } = require('../..')
 const { closeServerAsPromise } = require('../utils/node-http')
 
 // https://github.com/nodejs/undici/issues/1776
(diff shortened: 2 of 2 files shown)


KhafraDev (changes requested):
This is likely a patch for a larger, still unfixed bug. Otherwise I would expect other implementations to have similar issues.

https://fetch.spec.whatwg.org/#http-redirect-fetch

KhafraDev:
this is fixed in #5292

Issue #4710, opened 2025-12-23
match throws an error after external resource allocation occurs


fabiancook:
## Bug Description

When calling `.match(url)` for a cache, it is possible to get an error thrown that `Body has already been consumed`

## Reproducible By

I have been trying to reproduce this for a good while now, this here consistently has been popping up for our team. The reproduction takes hint from the original posted here but this consistently exits after seconds (or less)

Changing `ALLOCATION_SIZE_BETWEEN_USE` to a larger value makes the issue occur sooner.   

```typescript
import { caches, Response } from "undici";

const url = "https://example.com/";

const ALLOCATION_SIZE_BETWEEN_USE = 30000;
const cache = await caches.open("cache");
await cache.put(
  url,
  Response.json({
    initiatedAt: new Date().toISOString(),
    random: Math.random(),
    answer: 42,
  }),
);

for (let i = 0; i < 100; i += 1) {
  {
    void Array.from({ length: ALLOCATION_SIZE_BETWEEN_USE }, () => ({
      value: Math.random(),
    }));
    // Without the long wait, no problem occurs
    // 1-2ms timeout does not cause problem
    // >3ms timeout does cause problem inconsistently
    // >4ms timeout does cause problem consistently
    // queueMicrotask does not cause problem
    await new Promise((resolve) => setTimeout(resolve, 4));
  }

  const match = await cache.match(url);
  const result = await match.json();

  console.log(result);
}
```

In local environments where we are wanting to use in memory cache storage this makes any non trivial use non usable if we're doing things alongside. 

In tests it had always worked with small cases, but as we used this more and had other resources being allocated in memory, it seems we hit it faster, which can be seen by the change in object allocation causing issue. 

## Expected Behavior

The code to run (It does in a browser, below)

<img width="608" height="329" alt="Image" src="https://github.com/user-attachments/assets/6d627b0a-c2ef-45ce-bb4a-1768601b58a8" />

## Logs & Screenshots

```
{
  initiatedAt: '2025-12-23T01:39:11.929Z',
  random: 0.1796295151908207,
  answer: 42
}
/project/node_modules/undici/lib/web/webidl/index.js:34
  return new TypeError(`${message.header}: ${message.message}`)
         ^

TypeError: Response.clone: Body has already been consumed.
    at webidl.errors.exception (/project/node_modules/undici/lib/web/webidl/index.js:34:10)
    at Response.clone (/project/node_modules/undici/lib/web/fetch/response.js:236:27)
[...]

PR #4713, opened 2025-12-29
fix(cache): regenerate stream from source when cache.match is called after GC

mcollina:
## Summary

- Fixes cache.match() throwing "Body has already been consumed" after garbage collection runs between calls
- When a cached response's stream becomes unusable (cancelled by FinalizationRegistry), recreates the stream from the stored body source

## Root Cause

The bug occurs due to the interaction between the `FinalizationRegistry` (used to track Response objects) and the `cloneBody()` function which mutates the original body's stream in-place:

1. `cache.match()` calls `fromInnerResponse()` which creates a temporary Response and registers the cached stream
2. Then `response.clone()` is called, which tees the stream and **mutates** the cached stream to be one of the tee outputs
3. `clone()` re-registers this mutated stream with the temporary Response
4. When the temporary Response is GC'd, the FinalizationRegistry callback cancels the stream
5. Since the cancelled stream IS the cached stream (mutated in place), subsequent `cache.match()` calls fail

## Fix

The fix checks if the cached stream is unusable but `body.source` is available (which `cache.put()` always stores after reading the body bytes), and recreates the stream from source before creating the Response object.

Fixes: https://github.com/nodejs/undici/issues/4710

## Test plan

- [x] Added test that reproduces the issue with GC pressure
- [x] Test passes with the fix
- [x] Full cache test suite passes
- [x] Linter passes

### DIFF
--- a/lib/web/cache/cache.js
+++ b/lib/web/cache/cache.js
@@ -4,12 +4,12 @@ const assert = require('node:assert')
 
 const { kConstruct } = require('../../core/symbols')
 const { urlEquals, getFieldValues } = require('./util')
-const { kEnumerableProperty, isDisturbed, isBuffer } = require('../../core/util')
+const { kEnumerableProperty, isDisturbed } = require('../../core/util')
 const { webidl } = require('../webidl')
 const { cloneResponse, fromInnerResponse, getResponseState } = require('../fetch/response')
 const { Request, fromInnerRequest, getRequestState } = require('../fetch/request')
 const { fetching } = require('../fetch/index')
-const { urlIsHttpHttpsScheme, readAllBytes, readableStreamClose } = require('../fetch/util')
+const { urlIsHttpHttpsScheme, readAllBytes } = require('../fetch/util')
 const { createDeferredPromise } = require('../../util/promise')
 
 /**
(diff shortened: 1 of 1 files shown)


tsctx:
```diff
diff --git a/lib/web/cache/cache.js b/lib/web/cache/cache.js
index 70a3787a..10decbed 100644
--- a/lib/web/cache/cache.js
+++ b/lib/web/cache/cache.js
@@ -794,9 +794,9 @@ class Cache {
     // 5.5.2
     for (const response of responses) {
       // 5.5.2.1
-      const responseObject = fromInnerResponse(response, 'immutable')
+      const responseObject = fromInnerResponse(cloneResponse(response), 'immutable')
 
-      responseList.push(responseObject.clone())
+      responseList.push(responseObject)
 
       if (responseList.length >= maxResponses) {
         break
diff --git a/lib/web/fetch/response.js b/lib/web/fetch/response.js
index b3f942c3..6b954afa 100644
--- a/lib/web/fetch/response.js
+++ b/lib/web/fetch/response.js
@@ -555,7 +555,8 @@ function fromInnerResponse (innerResponse, guard) {
   setHeadersList(headers, innerResponse.headersList)
   setHeadersGuard(headers, guard)
 
-  if (innerResponse.body?.stream) {
+  // Note: If innerResponse's urlList contains a URL, it is a fetch response.
+  if (innerResponse.urlList.length !== 0 && innerResponse.body?.stream) {
     // If the target (response) is reclaimed, the cleanup callback may be called at some point with
     // the held value provided for it (innerResponse.body.stream). The held value can be any value:
     // a primitive or an object, even undefined. If the held value is an object, the registry keeps

```

@mcollina 
I just avoided `Response#clone()` and also improved it so that finalizers are not added to anything other than fetch responses. I think this is better, wdyt?

mcollina:
@tsctx ptal

lib/web/cache/cache.js
```diff
@@ -793,10 +793,38 @@ class Cache {
...
+              }
+
+              queueMicrotask(() => readableStreamClose(controller))
+            },
+            start () {},
+            type: 'bytes'
+          })
+        }
+      }
+
```
tsctx: We don't need this workaround anymore because the issue has been resolved.

mcollina:
@tsctx ptal

fabiancook:
Noting I've been making use of the iterations of changes locally using a patch against the package, has been working great for a larger amount of caches, time, & other used resources.

Issue #4703, opened 2025-12-17
MockAgent origin is case-sensetive


artur-ma:
## Bug Description

MockAgent comparse the origin in case-sensetive way

## Reproducible By

```js
    const url = 'http://myEndpoint';

    const mockAgent = new MockAgent();
    const mockPool = mockAgent.get(url);
    mockPool
      .intercept({
        path: '/path',
        method: 'GET',
      })
      .reply(
        200,
        { key: 'value' },
        {
          headers: { 'content-type': 'application/json' },
        },
      );

    await mockAgent.request({
      origin: new URL(url),
      method: 'GET',
      path: '/path',
    });
```
## Expected Behavior

The reuqest is interecepted and correct reply returned

## Logs & Screenshots

but actuin behavior is 

```
getaddrinfo ENOTFOUND myendpoint
```

metcoder95:
The host + tld of a domain should be case insensitive, for instance the current behavior is indeed a bug.

Would you like to send a PR for it?

PR #4731, opened 2026-01-07
Implement origin normalization in MockAgent for case-insensitivity and URL handling

SksOp:
## Changes
- adds a `normalizeOrigin` function in `lib/mock/mock-utils.js`

### Bug Fixes

MockAgent origin is case-sensetive [#4703](https://github.com/nodejs/undici/issues/4703)

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/test/mock-utils.js
+++ b/test/mock-utils.js
@@ -290,37 +290,11 @@ describe('normalizeOrigin', () => {
   })
 
   test('should preserve port numbers', (t) => {
-    t.plan(4)
+    t.plan(3)
 
     t.assert.strictEqual(normalizeOrigin('http://Example.com:8080'), 'http://example.com:8080')
-    // Note: url.origin omits default ports (443 for HTTPS, 80 for HTTP)
-    t.assert.strictEqual(normalizeOrigin('https://Api.Example.com:443'), 'https://api.example.com')
     t.assert.strictEqual(normalizeOrigin(new URL('http://Example.com:3000')), 'http://example.com:3000')
     t.assert.strictEqual(normalizeOrigin(new URL('https://Test.com:8443')), 'https://test.com:8443')
-  })
-
-  test('should handle default ports correctly', (t) => {
-    t.plan(2)
-
-    // Default ports should be omitted from origin
-    t.assert.strictEqual(normalizeOrigin('http://Example.com:80'), 'http://example.com')
-    t.assert.strictEqual(normalizeOrigin('https://Example.com:443'), 'https://example.com')
-  })
-
-  test('should remove trailing slash when ignoreTrailingSlash is true', (t) => {
-    t.plan(4)
-
-    t.assert.strictEqual(normalizeOrigin('http://example.com/'), 'http://example.com')
-    t.assert.strictEqual(normalizeOrigin('https://example.com/'), 'https://example.com')
-    t.assert.strictEqual(normalizeOrigin(new URL('http://example.com/')), 'http://example.com')
-    t.assert.strictEqual(normalizeOrigin('http://example.com'), 'http://example.com')
-  })
-
-  test('should not remove trailing slash when ignoreTrailingSlash is false', (t) => {
-    t.plan(2)
-
-    t.assert.strictEqual(normalizeOrigin('http://example.com/'), 'http://example.com')
-    t.assert.strictEqual(normalizeOrigin('http://example.com'), 'http://example.com')
   })
 
   test('should return RegExp matchers as-is', (t) => {


lib/mock/mock-agent.js
```diff
@@ -176,26 +179,31 @@ class MockAgent extends Dispatcher {
...
 
     // If we match, create a pool and assign the same dispatches
     for (const [keyMatcher, result] of Array.from(this[kClients])) {
-      if (result && typeof keyMatcher !== 'string' && matchValue(keyMatcher, origin)) {
-        const dispatcher = this[kFactory](origin)
-        this[kMockAgentSet](origin, dispatcher)
-        dispatcher[kDispatches] = result.dispatcher[kDispatches]
-        return dispatcher
+      if (result && typeof keyMatcher !== 'string') {
+        const normalizedKeyMatcher = normalizeOrigin(keyMatcher, this[kIgnoreTrailingSlash])
```
metcoder95: At this point, if stored correctly, the key matcher doe snot need new normalization, isn't it?

lib/mock/mock-agent.js
```diff
@@ -176,26 +179,31 @@ class MockAgent extends Dispatcher {
   }
 
   [kMockAgentGet] (origin) {
-    // First check if we can immediately find it
-    const result = this[kClients].get(origin)
+    const normalizedOrigin = normalizeOrigin(origin, this[kIgnoreTrailingSlash])
```
metcoder95: On L59 the origin is already normalized by the caller, we can save another normalization, wdyt?

lib/mock/mock-agent.js
```diff
@@ -165,7 +167,8 @@ class MockAgent extends Dispatcher {
   }
 
   [kMockAgentSet] (origin, dispatcher) {
-    this[kClients].set(origin, { count: 0, dispatcher })
+    const normalizedOrigin = normalizeOrigin(origin, this[kIgnoreTrailingSlash])
```
metcoder95: Same as per the getter one, unless this is called explicitly somewhere else and the origin is not normalized, we can save this call as well

lib/mock/mock-utils.js
```diff
@@ -396,6 +396,28 @@ function checkNetConnect (netConnect, origin) {
   return false
 }
 
+function normalizeOrigin (origin, ignoreTrailingSlash = false) {
+  if (typeof origin !== 'string' && !(origin instanceof URL)) {
+    return origin
+  }
+
+  let originString = typeof origin === 'string' ? origin : origin.toString()
```
metcoder95: maybe?

```suggestion
  let originString = typeof origin === 'string' ? origin : origin.origin
```

lib/mock/mock-utils.js
```diff
@@ -396,6 +396,28 @@ function checkNetConnect (netConnect, origin) {
...
+
+  try {
+    const url = new URL(originString)
```
metcoder95: If the passed arg is already an URL instance, maybe we can try to just infer the origin first?

lib/mock/mock-utils.js
```diff
@@ -396,6 +396,28 @@ function checkNetConnect (netConnect, origin) {
...
+    url.hostname = url.hostname.toLowerCase()
```
metcoder95: The `URL` already does the parsing into lower case as part of the spec

lib/mock/mock-utils.js
SksOp: Yahh true, since we are just dealing with origin and its an instance of URL

So maybe I should just check if its a URL instance i can just return origin

SksOp:
I removed the redundant normalisation in mock agent and kept the normalising logic to dispatch and get only. Also changed the logic of normalization

SksOp:
I think passed option ignoreTrailingSpace is not affecting the result. 
As the URL.origin always removes the trailing space. 
Even if a string is passed as origin, I am converting it to the URL instance and retreiving the origin, which again remove the trailing space.

In general, trailing space in the origin do not makes sense?
Should I just remove the option ?
WDYT?

mcollina:
Remove it!

SksOp:
Okay I will remove it.

SksOp:
Done

metcoder95:
Tests are not happy 😞

mcollina:
CI is red

SksOp:
I suspected this would happend, now the only other way I can think of is 
use `toLowerCase` for string type origin instead of `originString = new URL(origin).origin`

And move the logic of ignoreTrailingSlash back to the get as it was.

Locally tests looks happy

mcollina:
CI is red

SksOp:
I see there are some tests hanging in the `normalizeOrigin` function that I implemented which were ment to be for ignore trailing slash argument. Since I removed that param from the utility function. It needs a cleanup.
I will push new changes soon.

SksOp:
CI should be fine now, I tried in my local fork, [here](https://github.com/SksOp/undici/actions/runs/20918123696).

The one that are failing, are failing for all the PRs out there I think.

PR #4745, opened 2026-01-17
fix(websocket): add Authorization header and fix close event timing

### LINKED ISSUE #4741: WebSocket events fire synchronously during close() instead of asynchronously
## Bug Description

## Reproducible By

```js

let closeReturned = false;

});

  process.exit(0);
});

ws.close();
closeReturned = true;

setTimeout(() => {
  process.exit(1);
}, 5000);
```

## Expected Behavior

```
Calling close()...
```

## Logs & Screenshots

Actual behavior:

```
Calling close()...
```

## Environment

- Node v25.2.1
- undici 7.18.2

### LINKED ISSUE #4742: WebSocket transitions back from CLOSED to CLOSING
## Bug Description

When calling close() on a CONNECTING WebSocket, the readyState property is 3 (CLOSED) during the close event handler, but incorrectly reverts to 2 (CLOSING) after the handler returns. Once a WebSocket reaches the CLOSED state, it should never transition back to CLOSING.

## Reproducible By

```js

let capturedReadyState;

  capturedReadyState = ws.readyState;
  console.log('during close event, readyState =', ws.readyState);
});

ws.close();

console.log('after close(), readyState =', ws.readyState);

setTimeout(() => {
  console.log('captured readyState from inside handler =', capturedReadyState);
  console.log('after 10ms, readyState =', ws.readyState);
}, 10);

setTimeout(() => {
  console.log('after 100ms, readyState =', ws.readyState);

  // Expected: readyState should be 3 (CLOSED) after close event
  // Bug: readyState is 2 (CLOSING) after the event handler returns
  if (ws.readyState === 3) {
    console.log('PASS: readyState is CLOSED (3)');
  } else {
    console.log('FAIL: readyState is', ws.readyState, '(expected 3)');
  }
  process.exit(0);
}, 100);

setTimeout(() => {
  console.log('Timeout - something went wrong');
  process.exit(1);
}, 5000);
```

## Expected Behavior

```
after close(), readyState = 2
captured readyState from inside handler = undefined
after 10ms, readyState = 2
during close event, readyState = 3
after 100ms, readyState = 3
PASS: readyState is CLOSED (3)
```

See: https://jsbin.com/vugosuzeqa/edit?html,js,output

This follows from how the spec synchronously changes the state to CLOSING in https://websockets.spec.whatwg.org/#dom-websocket-close , and then after closing, [queues a task](https://websockets.spec.whatwg.org/#closeWebSocket:~:text=cleanly%2C-,the%20user%20agent%20must%20queue%20a%20task) to change the state to CLOSED and fire events.

So, this is probably the same root cause as #4741.

Relevant WPTs:

- https://github.com/web-platform-tests/wpt/blob/master/websockets/interfaces/WebSocket/close/close-connecting.html
- https://github.com/web-platform-tests/wpt/blob/master/websockets/interfaces/WebSocket/close/close-connecting-async.any.js

## Logs & Screenshots

Actual behavior:

```
during close event, readyState = 3
after close(), readyState = 2
captured readyState from inside handler = 3
after 10ms, readyState = 2
after 100ms, readyState = 2
FAIL: readyState is 2 (expected 3)
```

The misordering is again related to #4741.

## Environment

- Node v25.2.1
- undici 7.18.2

### LINKED ISSUE #4744: WebSocket (and probably fetch) does not support Authorization header from URL
## Bug Description

When I create a connection to a `WebSocket` URL such as `ws://foo:bar@localhost:1337/`, no `Authorization` header shows up.

## Reproducible By

```js
const { WebSocket } = require("undici");

const server = createServer();
server.on("upgrade", (req) => {
  console.log("Authorization:", req.headers.authorization);
  console.log("Expected:     ", "Basic " + Buffer.from("foo:bar").toString("base64"));
  process.exit();
});

server.listen(0, () => {
  new WebSocket(`ws://foo:bar@localhost:${server.address().port}/`);
});
```

## Expected Behavior

The header should be available.

Web platform test: https://github.com/web-platform-tests/wpt/blob/master/websockets/basic-auth.any.js

## Logs & Screenshots

```
Authorization: undefined
Expected:      Basic Zm9vOmJhcg==
```

## Environment

- Node v25.2.1
- undici 7.18.2

### Additional context

Claude claims this is caused by https://github.com/nodejs/undici/blob/250efc8eab72086384fe91b5ee3adbcd9d45dac0/lib/web/fetch/index.js#L1527-L1534 , and that seems plausible, although I haven't double-checked that the fetch code is used for the WebSocket connection.

The `ws` package gets this correct: https://github.com/websockets/ws/blob/ca533a53f338c4a40a3881dd6dff7e5867d06893/lib/websocket.js#L848-L856

mcollina:
## Summary
- Adds Authorization header from URL credentials when establishing WebSocket connections
- Respects custom Authorization headers set via `options.headers`
- Configures WPT test runner to trust the WPT server's CA certificate for TLS connections
- Fixes close event timing when `close()` is called in CONNECTING state:
  - Events now fire asynchronously after `close()` returns (instead of synchronously during)
  - readyState now transitions correctly: CONNECTING → CLOSING → CLOSED

## WPT Tests Now Passing
- `websockets/basic-auth.any.html?wss`
- `websockets/interfaces/WebSocket/close/close-connecting-async.any.html` (all variants)
- `websockets/interfaces/WebSocket/close/close-connecting.html` (all variants)
- `websockets/interfaces/WebSocket/close/close-multiple.html` (all variants)
- `websockets/interfaces/WebSocket/close/close-nested.html` (all variants)

Fixes #4744
Fixes #4741
Fixes #4742

### DIFF
--- a/lib/web/websocket/connection.js
+++ b/lib/web/websocket/connection.js
@@ -323,7 +323,8 @@ function failWebsocketConnection (handler, code, reason, cause) {
 
   if (isConnecting(handler.readyState)) {
     // If the connection was not established, we must still emit an 'error' and 'close' events
-    handler.onSocketClose()
+    // Queue via microtask so close() returns before events fire and readyState transitions correctly
+    queueMicrotask(() => handler.onSocketClose())
   } else if (handler.socket?.destroyed === false) {
     handler.socket.destroy()
   }
--- a/lib/web/websocket/websocket.js
+++ b/lib/web/websocket/websocket.js
@@ -554,6 +554,11 @@ class WebSocket extends EventTarget {
    * @see https://datatracker.ietf.org/doc/html/rfc6455#section-7.1.4
    */
   #onSocketClose () {
+    // Guard against duplicate calls
+    if (this.#handler.readyState === states.CLOSED) {
+      return
+    }
+
     // If the TCP connection was closed after the
     // WebSocket closing handshake was completed, the WebSocket connection
     // is said to have been closed _cleanly_.
--- a/test/web-platform-tests/expectation.json
+++ b/test/web-platform-tests/expectation.json
@@ -37734,8 +37734,7 @@
             "cases": [
               {
                 "name": "close event should be fired asynchronously when WebSocket is connecting",
-                "success": false,
-                "message": "assert_true: ws.close() should have returned expected true got false"
+                "success": true
               }
             ]
           },
@@ -37744,8 +37743,7 @@
             "cases": [
               {
                 "name": "close event should be fired asynchronously when WebSocket is connecting",
-                "success": false,
-                "message": "assert_true: ws.close() should have returned expected true got false"
+                "success": true
               }
             ]
           },
@@ -37754,8 +37752,7 @@
             "cases": [
               {
                 "name": "close event should be fired asynchronously when WebSocket is connecting",
-                "success": false,
-                "message": "assert_true: ws.close() should have returned expected true got false"
+                "success": true
               }
             ]
           },
@@ -37764,8 +37761,7 @@
             "cases": [
               {
                 "name": "close-connecting",
-                "success": false,
-                "message": "assert_unreached: Reached unreachable code"
+                "success": true
               }
             ]
           },
@@ -37774,8 +37770,7 @@
             "cases": [
               {
                 "name": "close-connecting",
-                "success": false,
-                "message": "assert_unreached: Reached unreachable code"
+                "success": true
               }
             ]
           },
@@ -37784,8 +37779,7 @@
             "cases": [
               {
                 "name": "close-multiple",
-                "success": false,
-                "message": "assert_equals: expected 1 but got 2"
+                "success": true
               }
             ]
           },
@@ -37794,8 +37788,7 @@
             "cases": [
               {
                 "name": "close-multiple",
-                "success": false,
-                "message": "assert_equals: expected 1 but got 2"
+                "success": true
               }
             ]
           },
@@ -37804,8 +37797,7 @@
             "cases": [
               {
                 "name": "close-nested",
-                "success": false,
-                "message": "assert_equals: expected 1 but got 2"
+                "success": true
               }
             ]
           },
@@ -37814,8 +37806,7 @@
             "cases": [
               {
                 "name": "close-nested",
-                "success": false,
-                "message": "assert_equals: expected 1 but got 2"
+                "success": true
               }
             ]
           },


lib/web/websocket/websocket.js
```diff
@@ -554,6 +554,11 @@ class WebSocket extends EventTarget {
    * @see https://datatracker.ietf.org/doc/html/rfc6455#section-7.1.4
    */
   #onSocketClose () {
+    // Guard against duplicate calls
+    if (this.#handler.readyState === states.CLOSED) {
+      return
+    }
```
KhafraDev: this is fine for now, but this shouldn't be called more than once 🤔

lib/web/websocket/connection.js
```diff
@@ -52,6 +52,14 @@ function establishWebSocketConnection (url, protocols, client, handler, options)
     request.headersList = headersList
   }
 
+  // If the URL has credentials, add Authorization header (unless already set by options.headers)
+  // @see https://fetch.spec.whatwg.org/#concept-basic-fetch (step 12.2.3)
+  if ((url.username || url.password) && !request.headersList.contains('authorization', true)) {
+    const credentials = `${url.username}:${url.password}`
+    const encoded = Buffer.from(credentials).toString('base64')
+    request.headersList.append('authorization', `Basic ${encoded}`, true)
+  }
```
KhafraDev, line 61: See my reasoning for why this should not be set: https://github.com/nodejs/undici/issues/4744#issuecomment-3764218474

KhafraDev (changes requested):

lib/web/websocket/connection.js
KhafraDev: > step 12.2.3

What is this referring to? scheme fetch doesn't have a 12.2.3

lib/web/websocket/websocket.js
KhafraDev, line 560: From local testing, this only seems to be called once. If we kept this in we'd be hiding other bugs.

domenic:
I'm concerned this fix isn't quite following the spec. It adds some extra state transition guards and a microtask. Whereas, the proper per-spec fix would be to add a real task (not microtask) at a higher level. E.g. instead of a microtask just for the `handler.onSocketClose()` part of `failWebsocketConnection`, the entirety `establishWebSocketConnection`, `closeWebSocketConnection`, and whatever undici's equivalent of ["when a WebSocket message has been received"](https://websockets.spec.whatwg.org/#:~:text=When%20a%20WebSocket%20message%20has%20been%20received%20with%20type%20type%20and%20data%20data%2C%20the%20user%20agent%20must%20queue%20a%20task) is.

I suspect if that sort of fix is implemented, then no specific state transition patches will be necessary, since the state transitions will always be queued properly in the task queue.

lib/web/websocket/websocket.js
mcollina: Sorry, I don't understand what you want to be done here

mcollina: The reference is actually wrong, I'll amend.

PR #4750, opened 2026-01-20
fix: prevent AbortController GC when redirect is 'error'

### LINKED ISSUE #4627: SSE connection may leak when fetch init.redirect = 'error'

mcollina:
## Summary

Fixes #4627

When `fetch()` is called with `redirect: 'error'`, the `requestObject` (the `Request` instance) could be garbage collected while the fetch was still in progress. The `requestObject` holds an internal `AbortController` via the `kAbortController` property. When `requestObject` was GC'd, the `AbortController` was also collected, triggering a `FinalizationRegistry` callback that removed the abort listener from the user's signal. This caused abort signals to have no effect, resulting in connection leaks for SSE streams.

## Changes

The fix stores a reference to `requestObject` in `fetchParams`, which stays alive for the entire duration of the fetch operation. This prevents premature garbage collection of the `requestObject` and its internal `AbortController`.

- Pass `requestObject` to the `fetching()` function
- Add `requestObject` parameter to `fetching()` function signature
- Include `requestObject` in `fetchParams` object

## Test Plan

- Added regression test `test/fetch/issue-4627.js` that requires `--expose-gc` flag
- Test creates an SSE-like streaming server, fetches with `redirect: 'error'`, triggers GC, aborts, and verifies no data is received after abort
- Both tests pass:
  - `abort should work with redirect: error` ✅
  - `abort should work without redirect: error (control test)` ✅

### DIFF
--- a/lib/web/fetch/index.js
+++ b/lib/web/fetch/index.js
@@ -436,7 +436,6 @@ function fetching ({
     taskDestination,
     crossOriginIsolatedCapability,
     // Keep requestObject alive to prevent its AbortController from being GC'd
-    // See https://github.com/nodejs/undici/issues/4627
     requestObject
   }
 


lib/web/fetch/index.js
```diff
@@ -430,7 +434,10 @@ function fetching ({
     processResponseConsumeBody,
     processResponseEndOfBody,
     taskDestination,
-    crossOriginIsolatedCapability
+    crossOriginIsolatedCapability,
+    // Keep requestObject alive to prevent its AbortController from being GC'd
+    // See https://github.com/nodejs/undici/issues/4627
```
Uzlopak:
```suggestion
```

I would remove this line as it is just a "diary entry". The corresponding test is having the link and is totally fine.

PR #4752, opened 2026-01-22
fix: WebSocket readyState should not revert from CLOSED to CLOSING

### LINKED ISSUE #4742: WebSocket transitions back from CLOSED to CLOSING

mcollina:
## Summary

- Fixes WebSocket `readyState` incorrectly reverting from CLOSED (3) to CLOSING (2) after close event fires
- Uses `queueMicrotask()` to defer `onSocketClose()` call, ensuring close event fires after state transition completes
- Adds guard to prevent duplicate `onSocketClose()` calls

## Test plan

- [x] Added test case for issue #4742
- [x] All 89 WebSocket tests pass
- [x] WPT `close-connecting` test now passes (was failing with "assert_unreached")
- [x] WPT `close-connecting-async` test passes

Fixes: https://github.com/nodejs/undici/issues/4742

### DIFF
--- a/lib/web/websocket/websocket.js
+++ b/lib/web/websocket/websocket.js
@@ -554,6 +554,12 @@ class WebSocket extends EventTarget {
    * @see https://datatracker.ietf.org/doc/html/rfc6455#section-7.1.4
    */
   #onSocketClose () {
+    // Guard against duplicate calls (e.g., from both the socket 'close' event
+    // and the queueMicrotask in failWebsocketConnection)
+    if (isClosed(this.#handler.readyState)) {
+      return
+    }
+
     // If the TCP connection was closed after the
     // WebSocket closing handshake was completed, the WebSocket connection
     // is said to have been closed _cleanly_.
--- a/lib/web/websocket/websocket.js
+++ b/lib/web/websocket/websocket.js
@@ -80,7 +80,9 @@ class WebSocket extends EventTarget {
 
       this.#handler.socket.destroy()
     },
-    onSocketClose: () => this.#onSocketClose(),
+    // Queue a task for the close event
+    // See: https://websockets.spec.whatwg.org/#closeWebSocket
+    onSocketClose: () => setImmediate(() => this.#onSocketClose()),
     onPing: (body) => {
       if (channels.ping.hasSubscribers) {
         channels.ping.publish({
@@ -554,8 +556,7 @@ class WebSocket extends EventTarget {
    * @see https://datatracker.ietf.org/doc/html/rfc6455#section-7.1.4
    */
   #onSocketClose () {
-    // Guard against duplicate calls (e.g., from both the socket 'close' event
-    // and the queueMicrotask in failWebsocketConnection)
+    // Guard against the close event firing twice
     if (isClosed(this.#handler.readyState)) {
       return
     }
--- a/lib/web/websocket/connection.js
+++ b/lib/web/websocket/connection.js
@@ -316,7 +316,9 @@ function failWebsocketConnection (handler, code, reason, cause) {
 
   if (isConnecting(handler.readyState)) {
     // If the connection was not established, we must still emit an 'error' and 'close' events
-    handler.onSocketClose()
+    // Use queueMicrotask to ensure the close event fires after readyState is set to CLOSING
+    // See: https://websockets.spec.whatwg.org/#closeWebSocket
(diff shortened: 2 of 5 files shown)


KhafraDev:
https://github.com/nodejs/undici/pull/4745#issuecomment-3764404322

mcollina:
@KhafraDev that does not link directly to anything on my machine.

KhafraDev:
Not sure why, this is what I was linking to (domenic's comment):

> I'm concerned this fix isn't quite following the spec. It adds some extra state transition guards and a microtask. Whereas, the proper per-spec fix would be to add a real task (not microtask) at a higher level. E.g. instead of a microtask just for the handler.onSocketClose() part of failWebsocketConnection, the entirety establishWebSocketConnection, closeWebSocketConnection, and whatever undici's equivalent of ["when a WebSocket message has been received"](https://websockets.spec.whatwg.org/#:~:text=When%20a%20WebSocket%20message%20has%20been%20received%20with%20type%20type%20and%20data%20data%2C%20the%20user%20agent%20must%20queue%20a%20task) is.

> I suspect if that sort of fix is implemented, then no specific state transition patches will be necessary, since the state transitions will always be queued properly in the task queue.

mcollina:
I'm not convinced by that argument, unless needed by the spec.

The problem is that adding macroticks everywhere will necessarily _slow things down_ and increase memory consumption.

cc @domenic

domenic:
It is needed by the spec? The spec is very explicit!

lib/web/websocket/websocket.js
```diff
@@ -554,6 +556,11 @@ class WebSocket extends EventTarget {
    * @see https://datatracker.ietf.org/doc/html/rfc6455#section-7.1.4
    */
   #onSocketClose () {
+    // Guard against the close event firing twice
+    if (isClosed(this.#handler.readyState)) {
+      return
+    }
+
```
KhafraDev, line 563:
```suggestion
```

This would be hiding a potential bug.

PR #4756, opened 2026-01-22
fix(fetch): prevent infinite retry loop on 401 responses

killagu:
`isTraversableNavigable()` was returning `true` unconditionally, which caused an infinite retry loop when the server responded with 401.

In Node.js environment, there is no traversable navigable that can prompt the user for credentials (unlike browsers). The 401 retry logic at step 14 of HTTP-network-or-cache fetch requires user interaction to provide credentials, which is not possible in Node.js.

Returning `false` disables the incomplete 401 retry logic that was causing the infinite loop, while preserving the URL credentials functionality added in #4747.

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/test/fetch/401-statuscode-no-infinite-loop.js
+++ b/test/fetch/401-statuscode-no-infinite-loop.js
@@ -1,0 +1,48 @@
+'use strict'
+
+const { fetch } = require('../..')
+const { createServer } = require('node:http')
+const { once } = require('node:events')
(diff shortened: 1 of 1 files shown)


silversf:
Stumbled independently upon the same issue so can confirm described issue is real, didn't confirm the fix though.

PR #4763, opened 2026-01-24
fix: properly close idle connections in encoding test

mcollina:
## Summary
- Fix flaky `test/fetch/encoding.js` test that fails intermittently on macOS with `ECONNRESET` errors
- Changed both `after()` hooks to call `server.closeIdleConnections()` before closing the server
- This ensures idle connections are properly drained rather than abruptly terminated

## Test plan
- [x] Ran encoding tests multiple times locally to verify stability
- [x] Linting passes


mcollina:
already fix

Issue #4767, opened 2026-01-24
Fetch 401 behavior nodejs


xconverge:
## Bug Description

I have been using the library for a long time, and up until v7.19.0, had no problems with unauthorized checks and 401 returns. I have a few different unit tests that use fetch against some unauthorized paths and check to make sure that 401s are correctly returned.

v7.19.0 introduced the 401 infinite loop issue and new authentication behavior and then v7.19.1 fixed/adjusted that in https://github.com/nodejs/undici/pull/4761

I am still a little confused, am I supposed to be able to actually get the 401 anymore, something still seems off to me?

In the return/error for [test/fetch/401-statuscode-no-infinite-loop.js](https://github.com/nodejs/undici/blob/main/test/fetch/401-statuscode-no-infinite-loop.js#L25) is there any way to get that it was a 401 (like the test cases in https://github.com/nodejs/undici/pull/4756/changes#diff-95de560a29bdff1c72b0c3af8fef710333db9464257e92a85784594b623cc68c)?

## Environment

Node v22 and v24

PR #4769, opened 2026-01-25
return response when receiving 401 instead of network error

KhafraDev:
fixes https://github.com/nodejs/undici/issues/4767

it didn't really make sense to throw an error when receiving a 401, did it...


xconverge:
Thanks, glad I wasn't crazy

PR #4775, opened 2026-01-27
fix: support Connection header with connection-specific header names per RFC 7230

### LINKED ISSUE #4774: Support sending Connection header with connection-specific header names per RFC 7230
## Summary

When using undici to make HTTP requests, attempting to send a `Connection` header with a value that lists other header names (per RFC 7230 Section 6.1) results in an `InvalidArgumentError: invalid connection header`.

## Background

RFC 7230 Section 6.1 specifies that the Connection header field can contain a comma-separated list of connection-option tokens, which correspond to header field names that should be removed by the first proxy or gateway that receives the message.

> The "Connection" header field allows the sender to indicate desired control options for the current connection. In order to avoid confusing downstream recipients, a proxy or gateway MUST remove or replace any received connection options before forwarding the message.

Example from the RFC:
```
Connection: close
Connection: keep-alive, Upgrade
Connection: X-Custom-Header
```

## Current Behavior

```javascript

await request('http://example.com', {
  headers: {
    'X-Custom-Header': 'some-value',
    'Connection': 'X-Custom-Header'  // RFC-compliant header value
  }
});
// Throws: InvalidArgumentError: invalid connection header
```

## Expected Behavior

Undici should allow Connection header values that contain comma-separated lists of header names, per RFC 7230 Section 6.1.

## Use Case

This is relevant for proxy implementations where a client sends a Connection header listing custom headers to strip. While undici/proxies should strip these headers before forwarding (as we now do in @fastify/reply-from), the current undici behavior prevents clients from sending RFC-compliant requests in the first place.

## References

- [RFC 7230 Section 6.1 - Connection](https://datatracker.ietf.org/doc/html/rfc7230#section-6.1)
- Related fix in @fastify/reply-from: https://github.com/fastify/fastify-reply-from/issues/449

mcollina:
## Summary

- Allows Connection header to contain comma-separated list of connection option tokens (header names) per RFC 7230 Section 6.1
- Previously only `close` or `keep-alive` were accepted, now any valid HTTP token is allowed
- Sets `request.reset = true` if `close` is present in the token list

## Test plan

- [x] Added tests for `close` and `keep-alive` values
- [x] Added test for custom header name as connection option
- [x] Added test for comma-separated list of connection options
- [x] Added test for rejecting invalid tokens
- [x] All existing tests pass

Fixes #4774

### DIFF
--- a/docs/docs/api/Dispatcher.md
+++ b/docs/docs/api/Dispatcher.md
@@ -476,6 +476,7 @@ The `RequestOptions.method` property should not be value `'CONNECT'`.
 #### Parameter: `ResponseData`
 
 * **statusCode** `number`
+* **statusText** `string` - The status message from the response (e.g., "OK", "Not Found").
 * **headers** `Record<string, string | string[]>` - Note that all header keys are lower-cased, e.g. `content-type`.
 * **body** `stream.Readable` which also implements [the body mixin from the Fetch Standard](https://fetch.spec.whatwg.org/#body-mixin).
 * **trailers** `Record<string, string>` - This object starts out
@@ -517,7 +518,7 @@ await once(server, 'listening')
 const client = new Client(`http://localhost:${server.address().port}`)
 
 try {
-  const { body, headers, statusCode, trailers } = await client.request({
+  const { body, headers, statusCode, statusText, trailers } = await client.request({
     path: '/',
     method: 'GET'
   })
--- a/docs/docs/api/MockPool.md
+++ b/docs/docs/api/MockPool.md
@@ -323,7 +323,8 @@ try {
     method: 'GET'
   })
 } catch (error) {
-  console.error(error) // Error: kaboom
+  console.error(error) // TypeError: fetch failed
+  console.error(error.cause) // Error: kaboom
 }
 ```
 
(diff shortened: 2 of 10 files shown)


lib/core/request.js
```diff
@@ -394,12 +394,21 @@ function processHeader (request, key, val) {
...
-    if (value !== 'close' && value !== 'keep-alive') {
+    // Per RFC 7230 Section 6.1, Connection header can contain
+    // a comma-separated list of connection option tokens (header names)
+    const value = typeof val === 'string' ? val : null
+    if (value === null) {
       throw new InvalidArgumentError('invalid connection header')
     }
 
-    if (value === 'close') {
+    const tokens = value.split(',').map(t => t.trim().toLowerCase())
```
metcoder95: We are iterating over them twice, can we do it at one pass?

Uzlopak: You could actually lowercase before split.

mcollina: Done! Refactored to lowercase before split and combined validation + close-check in a single loop.

mcollina:
@metcoder95 ping

PR #4786, opened 2026-01-30
Remove legacy handler wrappers

mcollina:
## Summary
- fix raw header handling in cache/decompress interceptors
- add shared toRawHeaders utility
- update snapshot replay raw headers
- bulk of the v8 version change

## Testing
- npm test


mcollina:
I used AI for this, so review with caution

ronag (approved):

metcoder95 (approved):

hexchain:
This should also fix #4389.

domenic:
Let me know if it would be helpful to test the `next` branch in jsdom, and if so, when a good point for that would be. (Presumably not right now since it's 18 commits behind `main`, but perhaps as you get closer to release.)

mcollina:
I'll ship an alpha soon.

PR #4788, opened 2026-01-31
test: reduce retry-after invalid date flake

mcollina:
## Summary
- use monotonic timing and tolerance for the invalid Retry-After date test
- avoids wall-clock jitter causing false failures

## Testing
- npx borp -p "test/retry-handler.js"


mertcanaltin (approved):
Lgtm

PR #4798, opened 2026-02-02
fix: use latin1 encoding for headers in WrapHandler

### LINKED ISSUE #4797: WrapHandler translates headers using utf-8 instead of latin1
## Bug Description

When `WrapHandler` gets involved to translate from the new handler API calls to the old handler API, it incorrectly translates headers using utf-8 instead of isomorphic encoding, due to calling `Buffer.from(x)` instead of `Buffer.from(x, 'latin1')`.

This affects all APIs that use `WrapHandler` such as `request()`.

## Reproducible By

```js
const WrapHandler = require("undici/lib/handler/wrap-handler.js");

const oldAPIHandler = {
  onHeaders(statusCode, rawHeaders) {
    const value = rawHeaders[1].toString("latin1");
    console.log("Expected:", JSON.stringify("\xE2\x80\xA6"));
    console.log("Got:     ", JSON.stringify(value));
  }
};

const wrapped = WrapHandler.wrap(oldAPIHandler);
wrapped.onResponseStart({ resume() {} }, 200, { "x-test": "\xE2\x80\xA6" }, "OK");
```

<details>
<summary>A repro case using request() around a custom dispatcher that uses WrapHandler</summary>

```js
const http = require("http");
const { Dispatcher, request } = require("undici");
const WrapHandler = require("undici/lib/handler/wrap-handler.js");

  res.writeHead(200, { "x-test": "\xE2\x80\xA6" });
  res.end();
});

// Dispatcher that does its own HTTP and calls onResponseStart with object headers
class MyDispatcher extends Dispatcher {
  dispatch(opts, handler) {
    const wrapped = WrapHandler.wrap(handler);

    http.get(opts.origin + opts.path, res => {
      wrapped.onRequestStart?.({ abort() {}, pause() {}, resume() {} }, {});
      wrapped.onResponseStart?.({ resume() {}, pause() {} }, res.statusCode, res.headers, res.statusMessage);
      res.on("data", chunk => wrapped.onResponseData?.({}, chunk));
      res.on("end", () => wrapped.onResponseEnd?.({}, {}));
    }).on("error", err => wrapped.onResponseError?.({}, err));

    return true;
  }
  close() { return Promise.resolve(); }
  destroy() { return Promise.resolve(); }
}

(async () => {
  await new Promise(r => server.listen(0, r));
  const url = `http://localhost:${server.address().port}`;

  const response = await request(url, { dispatcher: new MyDispatcher() });
  await response.body.dump();

  console.log("Expected:", JSON.stringify("\xE2\x80\xA6"));
  console.log("Got:     ", JSON.stringify(response.headers["x-test"]));

  server.close();
})();
```
</details>

## Expected Behavior

```
Expected: "â¦"
Got:      "Ã¢ÂÂ¦"
```

## Environment

- Node v25.3.0
- Undici 7.20.0

### Additional context

I don't believe https://github.com/nodejs/undici/pull/4785 will fix this.

https://github.com/nodejs/undici/pull/4786 might fix this.

mcollina:
## Summary

- Fix `WrapHandler` to use Latin1 encoding when converting header strings to Buffers
- Previously used `Buffer.from(val)` which defaults to UTF-8, causing header values with non-ASCII bytes to be incorrectly encoded
- Affects `onRequestUpgrade`, `onResponseStart`, and `onResponseEnd` methods

Fixes #4797

## Test plan

- [x] Verified both repro cases from the issue now work correctly
- [x] All existing unit tests pass (1221 tests)
- [x] Lint passes

### DIFF
--- a/lib/handler/wrap-handler.js
+++ b/lib/handler/wrap-handler.js
@@ -53,7 +53,7 @@ module.exports = class WrapHandler {
   onRequestUpgrade (controller, statusCode, headers, socket) {
     const rawHeaders = []
     for (const [key, val] of Object.entries(headers)) {
-      rawHeaders.push(Buffer.from(key), Array.isArray(val) ? val.map(v => Buffer.from(v)) : Buffer.from(val))
+      rawHeaders.push(Buffer.from(key, 'latin1'), Array.isArray(val) ? val.map(v => Buffer.from(v, 'latin1')) : Buffer.from(val, 'latin1'))
     }
 
     this.#handler.onUpgrade?.(statusCode, rawHeaders, socket)
@@ -62,7 +62,7 @@ module.exports = class WrapHandler {
   onResponseStart (controller, statusCode, headers, statusMessage) {
     const rawHeaders = []
     for (const [key, val] of Object.entries(headers)) {
-      rawHeaders.push(Buffer.from(key), Array.isArray(val) ? val.map(v => Buffer.from(v)) : Buffer.from(val))
+      rawHeaders.push(Buffer.from(key, 'latin1'), Array.isArray(val) ? val.map(v => Buffer.from(v, 'latin1')) : Buffer.from(val, 'latin1'))
     }
 
     if (this.#handler.onHeaders?.(statusCode, rawHeaders, () => controller.resume(), statusMessage) === false) {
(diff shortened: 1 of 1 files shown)


metcoder95 (commented):
Shall we add tests to it?

Issue #4799, opened 2026-02-02
Cloning Response change default aborting behavior


afoures:
## Bug Description

When using an AbortSignal with fetch, and aborting it right before reading the body via `text()`/`json()`/..., it throws an abort error, as expected.

But if the response was cloned, then it does not throw an abort error but a TypeError, like if the body was read by someone.

## Reproducible By

```js
async function test() {
  try {
    const controller = new AbortController();
    const response = await fetch("https://example.com/data", {
      signal: controller.signal,
    });
    const cloned_response = response.clone();
    controller.abort();
    // body is now used in the original response
    console.log("response.bodyUsed", response.bodyUsed);
    console.log("cloned_response.bodyUsed", cloned_response.bodyUsed);
    // throws "Body is unusable: Body has already been read"
    // but it should only throw an AbortError
    await response.text();
  } catch (error) {
    console.error(error);
  }

  console.log("--------------------------------");

  // this works as expected
  try {
    const controller = new AbortController();
    const response = await fetch("https://example.com/data", {
      signal: controller.signal,
    });
    // const cloned_response = response.clone();
    controller.abort();
    // throws AbortError as expected
    await response.text();
  } catch (error) {
    console.error(error);
  }
}

test();
```

```txt
response.bodyUsed true
cloned_response.bodyUsed false
TypeError: Body is unusable: Body has already been read
    at consumeBody (node:internal/deps/undici/undici:6873:31)
    at _Response.text (node:internal/deps/undici/undici:6821:18)
    at test (file:///.../reproduction.js:14:20)
    at process.processTicksAndRejections (node:internal/process/task_queues:103:5)
--------------------------------
DOMException [AbortError]: The operation was aborted.
    at new DOMException (node:internal/per_context/domexception:76:18)
    at consumeBody (node:internal/deps/undici/undici:6876:31)
    at _Response.text (node:internal/deps/undici/undici:6821:18)
    at test (file:///.../reproduction.js:31:20)
    at process.processTicksAndRejections (node:internal/process/task_queues:103:5)
```

## Expected Behavior

I expected it to throw an AbortError, exactly like the example without a cloned response.

## Environment

macOS 15.6.1
node 24.13.0

PR #4801, opened 2026-02-03
fix: return AbortError instead of TypeError when consuming aborted cloned response

mcollina:
When a Response was cloned and the AbortController was aborted before consuming the body, consuming the body threw a TypeError ('Body is unusable') instead of the expected AbortError.

## Changes
- Check for aborted state before checking if body is unusable in `consumeBody()`
- Set `response.aborted` flag in `abortFetch()` to properly track abort state

## Test
Includes test from #4800 (thanks @afoures)

Fixes: #4799
Refs: #4800

### DIFF
--- a/lib/web/fetch/body.js
+++ b/lib/web/fetch/body.js
@@ -440,12 +440,14 @@ function consumeBody (object, convertBytesToJSValue, instance, getInternalState)
 
   // 1. If object is unusable, then return a promise rejected
   //    with a TypeError.
-  if (bodyUnusable(state)) {
-    return Promise.reject(new TypeError('Body is unusable: Body has already been read'))
-  }
-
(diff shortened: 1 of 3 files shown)


KhafraDev:
What I've noticed from using Claude on the code in undici: it *sucks* by default. I've found ways to improve on its output while also speeding up my process for bug fixing; have it fetch the spec in its entirety and then ask it to 1) do not edit existing comments under any circumstance, and 2) implement changes verbatim to the spec text. Without the spec text, it often does things like we see here where it adds a non-existent step 6 with a nonsensical change.

I usually end up doing all of the coding manually, using Claude to track down bugs, but then implementing the fix myself to ensure Claude doesn't muck it up.

PR #4809, opened 2026-02-05
fix: handle schema-less HTTP_PROXY env vars

### LINKED ISSUE #4736: Schema-less HTTP_PROXY and HTTPS_PROXY env vars should maybe assume `http`
## This would solve...

M-YasirGhaffar:
## This relates to...

Fixes #4736

## Rationale

When HTTP_PROXY or HTTPS_PROXY is set without a URL scheme (e.g., localhost:8080 instead of http://localhost:8080), the EnvHttpProxyAgent fails. This PR adds automatic http:// prefix for schema-less proxy URLs, matching Go's httpproxy behavior.

## Changes

Added normalizeProxyUrl() helper function in lib/dispatcher/env-http-proxy-agent.js. Also added tests in test/env-http-proxy-agent.js.

### Features

N/A

### Bug Fixes

- Schema-less proxy URLs now work by assuming http:// protocol

None. This is additive behavior - existing URLs with explicit schemes are unchanged.

## Status

- [x] Tested
- [ ] Benchmarked (optional)
- [S] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/dispatcher/env-http-proxy-agent.js
+++ b/lib/dispatcher/env-http-proxy-agent.js
@@ -10,6 +10,14 @@ const DEFAULT_PORTS = {
   'https:': 443
 }
 
+// Prepend http:// to proxy URLs without a scheme (matches Go's httpproxy behavior)
+function normalizeProxyUrl (proxyUrl) {
+  if (proxyUrl && !proxyUrl.startsWith('http://') && !proxyUrl.startsWith('https://')) {
+    return 'http://' + proxyUrl
+  }
+  return proxyUrl
+}
+
 class EnvHttpProxyAgent extends DispatcherBase {
   #noProxyValue = null
   #noProxyEntries = null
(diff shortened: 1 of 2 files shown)


mcollina:
cc @joyeecheung can you take a look as well?

metcoder95:
Will this require adjustments in Node for `fetch` besides the Undici upgrade?

M-YasirGhaffar:
@metcoder95 This shouldn't require any changes for `fetch`. The normalization happens inside EnvHttpProxyAgent before the URL reaches ProxyAgent.

joyeecheung:
As a user I'd find this behavior
a bit surprising....is there any tool other than the go package doing this convention?

M-YasirGhaffar:
@joyeecheung curl does this. If you pass a proxy without a scheme it defaults to HTTP. git does too since it uses libcurl under the hood. Go's httpproxy package also prepends http:// when the scheme is missing.

I think if someone sets HTTP_PROXY=myproxy:8080, they'd expect it to work, especially since curl handles it fine. The change is also non-breaking, URLs with explicit schemes go through unchanged.

**Sources:**
- curl/libcurl: ["HTTP Proxy. Default when no scheme or proxy type is specified."](https://curl.se/libcurl/c/CURLOPT_PROXY.html)
- Go httpproxy: [`parseProxy()` prepends `http://` for schema-less values](https://tip.golang.org/src/vendor/golang.org/x/net/http/httpproxy/proxy.go) (line 153-158)
- git: uses libcurl for HTTP proxy handling, inherits the same behavior.

joyeecheung:
Thanks for the pointers. I agree converging with what curl does is generally the safest.

mcollina (approved):
lgtm

mcollina:
There are some related failures

M-YasirGhaffar:
@mcollina could you point me to the specific failures? If you mean the macOS ECONNRESET in test/fetch/encoding.js and the Node 24/25 shared-builtin compilation ones, those seem unrelated to this change. Happy to dig further.

NimaKeshavarzian:
wouldn't it be better to make this a **private method** of ```EnvHttpProxyAgent```?

Issue #4815, opened 2026-02-07
`includesCredentials()` silently drops Authorization header for username-only or password-only URLs on 401 retry


jackhax:
## Bug Description

The `includesCredentials()` function in `lib/web/fetch/util.js` uses the `&&` (AND) operator instead of `||` (OR) when checking if a URL includes credentials. This contradicts the [WHATWG URL Standard §4.4](https://url.spec.whatwg.org/#include-credentials):

> A URL **includes credentials** if its username is not the empty string **or** its password is not the empty string.

## Affected Code

`lib/web/fetch/util.js`, lines 1437-1440:

```javascript
function includesCredentials (url) {
  // A URL includes credentials if its username or password is not the empty string.
  return !!(url.username && url.password)
}
```

Note the comment correctly states "or" while the code implements "and."

## Impact

This causes the 401 authentication retry path in `httpNetworkOrCacheFetch` to skip constructing the `Authorization` header when a URL contains only a username or only a password:

```javascript
} else if (includesCredentials(requestCurrentURL(httpRequest)) && isAuthenticationFetch) {
  const { username, password } = requestCurrentURL(httpRequest);
  authorizationValue = `Basic ${Buffer.from(`${username}:${password}`).toString("base64")}`;
}
```

When `includesCredentials()` incorrectly returns `false`, `authorizationValue` remains `null` and no `Authorization` header is sent on the retry. This primarily affects WebSocket connections (which set `useURLCredentials: true`), causing silent authentication failure for token-as-username patterns like `ws://apitoken@host/path`.

## Reproduction

Tested on Node.js v25.6.0 (bundled undici):

```javascript
const net = require('net');

function runTest(port, url, label) {
  return new Promise((resolve) => {
    let connections = [];
    const server = net.createServer((socket) => {
      let data = '';
      socket.on('data', (chunk) => {
        data += chunk.toString();
        const authMatch = data.match(/Authorization: (.+)\r\n/);
        connections.push({
          hasAuth: !!authMatch,
          authValue: authMatch ? authMatch[1] : 'NONE',
        });
        socket.write(
          'HTTP/1.1 401 Unauthorized\r\nWWW-Authenticate: Basic realm="test"\r\n' +
          'Content-Length: 0\r\nConnection: close\r\n\r\n'
        );
        socket.end();
      });
    });

    server.listen(port, () => {
      const ws = new WebSocket(url);
      ws.addEventListener('error', () => {});
      ws.addEventListener('close', () => {
[...]

PR #4816, opened 2026-02-07
fix: use OR operator in includesCredentials per WHATWG URL Standard

jackhax:
## Summary

- Fix logical operator in `includesCredentials()` (`lib/web/fetch/util.js`): `&&` → `||`
- Add unit tests for `includesCredentials()` covering all credential combinations

## Problem

The `includesCredentials()` function uses AND (`&&`) instead of OR (`||`), contradicting both:
1. Its own comment: *"A URL includes credentials if its username **or** password is not the empty string."*
2. [WHATWG URL Standard §4.4](https://url.spec.whatwg.org/#include-credentials): *"A URL includes credentials if its username is not the empty string **or** its password is not the empty string."*

This causes the Authorization header to be silently omitted during authentication retries for URLs with only a username (e.g., `http://apitoken@host`) or only a password (e.g., `http://:secret@host`).

## Fix

```diff
 function includesCredentials (url) {
   // A URL includes credentials if its username or password is not the empty string.
-  return !!(url.username && url.password)
+  return !!(url.username || url.password)
 }
```

## Test plan

- [x] New test: URL with both username and password → `true`
- [x] New test: URL with only username → `true`
- [x] New test: URL with only password → `true`
- [x] New test: URL with no credentials → `false`
- [x] Existing fetch tests pass (no regressions)
- [x] ESLint passes

Fixes: https://github.com/nodejs/undici/issues/4815

Ref: [CWE-480: Use of Incorrect Operator](https://cwe.mitre.org/data/definitions/480.html)

### DIFF
--- a/lib/web/fetch/util.js
+++ b/lib/web/fetch/util.js
@@ -1439,7 +1439,7 @@ function hasAuthenticationEntry (request) {
  */
 function includesCredentials (url) {
   // A URL includes credentials if its username or password is not the empty string.
-  return !!(url.username && url.password)
+  return !!(url.username || url.password)
 }
 
 /**
--- a/test/fetch/includes-credentials.js
+++ b/test/fetch/includes-credentials.js
@@ -1,0 +1,24 @@
+'use strict'
+
+const { test } = require('node:test')
+const { includesCredentials } = require('../../lib/web/fetch/util')
+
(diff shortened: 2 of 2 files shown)


mcollina (approved):
lgtm

PR #4823, opened 2026-02-12
feat!: enable h2 by default

metcoder95:
## This relates to...

This enables HTTP/2 support by default in Undici.
## Rationale

Currently, users must explicitly opt-in to H2 support via `allowH2: true`. This change adjusts Undici's defaults to align with current web standards and improve out-of-the-box performance.

## Changes

This PR changes the default value of the `allowH2` option from `false` to `true` in the connection builder (`lib/core/connect.js`). When connecting to HTTPS servers, Undici will now use HTTP/2 if the server advertises it through ALPN negotiation, falling back to HTTP/1.1 if not supported.

### Features

- HTTP/2 is now enabled by default for connections to servers that support it via ALPN negotiation
- Users can still explicitly disable H2 by setting `allowH2: false` in client options

### Bug Fixes

N/A

**Breaking Change**: The default value of `allowH2` changes from `false` to `true`.

**Impact**: 
- Applications relying on HTTP/1.1-only behavior by default will now use HTTP/2 when connecting to H2-capable servers
- Users who need HTTP/1.1-only connections must now explicitly set `allowH2: false` in their client configuration
- The protocol upgrade is transparent and should not affect most applications, but may change connection behavior and performance characteristics

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready


mcollina:
ahum, didn't work as expected.

metcoder95:
I think it needs a rebase, let me give it a shot

metcoder95:
Closed in favor of #4828

PR #4824, opened 2026-02-12
fix: harden header iterable checks for prototype-pollution scenarios

mcollina:
## Summary

This PR hardens header processing against prototype-pollution side effects.

Undici previously used inherited `Symbol.iterator` lookups for some header inputs. If `Object.prototype[Symbol.iterator]` was polluted, plain header objects could be misclassified as iterables and silently lose headers.

This is **not a standalone vulnerability** in Undici; it requires a pre-existing prototype-pollution primitive. This change improves resilience in that scenario.

## Changes

Updated iterable detection in:

- `lib/core/request.js`
- `lib/handler/redirect-handler.js`
- `lib/util/cache.js`

New behavior:

- Treat headers as iterable only when:
  - `Symbol.iterator` is an **own** property, or
  - the object has a **non-plain prototype** and an iterator function (e.g. `Map`, `Headers`)
- Otherwise treat headers as plain objects via `Object.keys` / `Object.entries`.

## Tests

- `test/request.js`
  - plain object headers are preserved when `Object.prototype[Symbol.iterator]` is polluted
- `test/interceptors/redirect.js`
  - same-origin redirect preserves plain object headers under polluted iterator
- `test/cache-interceptor/cache-utils.js` (new)
  - `normalizeHeaders` works for polluted plain objects
  - `normalizeHeaders` still works for `Map` headers

## Verification

- `npx borp -p "test/request.js"`
- `npx borp -p "test/interceptors/redirect.js"`
- `npx borp -p "test/cache-interceptor/cache-utils.js"`
- `npx borp -p "test/interceptors/cache.js"`
- `npm run lint -- lib/core/request.js lib/util/cache.js lib/handler/redirect-handler.js test/request.js test/interceptors/redirect.js test/cache-interceptor/cache-utils.js`

### DIFF
--- a/lib/handler/redirect-handler.js
+++ b/lib/handler/redirect-handler.js
@@ -222,10 +222,7 @@ function cleanRequestHeaders (headers, removeContent, unknownOrigin) {
       }
     }
   } else if (headers && typeof headers === 'object') {
-    const prototype = Object.getPrototypeOf(headers)
-    const ownIterator = Object.prototype.hasOwnProperty.call(headers, Symbol.iterator)
-    const hasIterator = ownIterator || (prototype != null && prototype !== Object.prototype && typeof headers[Symbol.iterator] === 'function')
-    const entries = hasIterator ? headers : Object.entries(headers)
+    const entries = util.hasSafeIterator(headers) ? headers : Object.entries(headers)
 
     for (const [key, value] of entries) {
       if (!shouldRemoveHeader(key, removeContent, unknownOrigin)) {
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -13,6 +13,7 @@ const {
   isBuffer,
   isFormDataLike,
   isIterable,
+  hasSafeIterator,
   isBlobLike,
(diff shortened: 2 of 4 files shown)


lib/handler/redirect-handler.js
```diff
@@ -222,7 +222,11 @@ function cleanRequestHeaders (headers, removeContent, unknownOrigin) {
       }
     }
   } else if (headers && typeof headers === 'object') {
-    const entries = typeof headers[Symbol.iterator] === 'function' ? headers : Object.entries(headers)
+    const prototype = Object.getPrototypeOf(headers)
+    const ownIterator = Object.prototype.hasOwnProperty.call(headers, Symbol.iterator)
+    const hasIterator = ownIterator || (prototype != null && prototype !== Object.prototype && typeof headers[Symbol.iterator] === 'function')
+    const entries = hasIterator ? headers : Object.entries(headers)
```
metcoder95: Shall we maybe create an utility function to share it across?

PR #4825, opened 2026-02-12
feat: Option to restrict the number of free/idle sockets

JoshMock:
## This relates to...

Part of https://github.com/nodejs/undici/issues/3268. Follow-up to https://github.com/nodejs/undici/pull/4365.

## Rationale

Adds support to `Pool` (and, indirectly, `Agent`) for limiting the maximum number of idle sockets. Because `Agent` just passes `maxFree` through to `Pool`, it inherently works like [`Agent.maxFreeSockets`](https://nodejs.org/api/http.html#agentmaxfreesockets) in the `http` standard library, limiting the maximum on a per-origin basis.

## Changes

- Adds `maxFree` option to `Pool`
- Adds tests to verify it works properly for both `Pool` and `Agent`
- Updates docs

### Features

- Adds `maxFree` option to `Pool` and `Agent`

### Bug Fixes

N/A

N/A

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/dispatcher/pool-base.js
+++ b/lib/dispatcher/pool-base.js
@@ -43,7 +43,10 @@ class PoolBase extends DispatcherBase {
     if (this[kMaxIdle] < Infinity && this[kFree] > this[kMaxIdle]) {
       for (let i = 0; i < this[kClients].length; i++) {
         const client = this[kClients][i]
-        if (!client[kNeedDrain]) this[kRemoveClient](client)
+        if (!client[kNeedDrain]) {
+          if (client.ttl && client.ttl > Date.now()) continue
+          this[kRemoveClient](client)
+        }
         if (this[kFree] <= this[kMaxIdle]) break
       }
     }
--- a/lib/dispatcher/pool.js
+++ b/lib/dispatcher/pool.js
@@ -82,7 +82,7 @@ class Pool extends PoolBase {
     this.on('connect', (origin, targets) => {
       if (clientTtl != null && clientTtl > 0) {
         for (const target of targets) {
-          Object.assign(target, { ttl: Date.now() })
+          Object.assign(target, { ttl: Date.now() + clientTtl })
         }
       }
     })
@@ -106,7 +106,7 @@ class Pool extends PoolBase {
     const clientTtlOption = this[kOptions].clientTtl
     for (const client of this[kClients]) {
       // check ttl of client and if it's stale, remove it from the pool
-      if (clientTtlOption != null && clientTtlOption > 0 && client.ttl && ((Date.now() - client.ttl) > clientTtlOption)) {
+      if (clientTtlOption != null && clientTtlOption > 0 && client.ttl && (client.ttl <= Date.now())) {
         this[kRemoveClient](client)
       } else if (!client[kNeedDrain]) {
         return client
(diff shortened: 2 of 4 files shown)


docs/docs/api/Pool.md
```diff
@@ -20,6 +20,7 @@ Extends: [`ClientOptions`](/docs/docs/api/Client.md#parameter-clientoptions)
 * **factory** `(origin: URL, opts: Object) => Dispatcher` - Default: `(origin, opts) => new Client(origin, opts)`
 * **connections** `number | null` (optional) - Default: `null` - The number of `Client` instances to create. When set to `null`, the `Pool` instance will create an unlimited amount of `Client` instances.
 * **clientTtl** `number | null` (optional) - Default: `null` - The amount of time before a `Client` instance is removed from the `Pool` and closed.   When set to `null`, `Client` instances will not be removed or closed based on age.
+* **maxFree** `number` (optional) - Default: `Infinity` - Limits the number of open connections that do not have an active request.
```
metcoder95: I'd rename it to something like `maxIdle` or `maxIdleConnections` to avoid ambiguity

lib/dispatcher/pool-base.js
```diff
@@ -40,6 +40,14 @@ class PoolBase extends DispatcherBase {
       needDrain = !client.dispatch(item.opts, item.handler)
     }
 
+    if (this[kMaxFree] < Infinity && this[kFree] > this[kMaxFree]) {
```
metcoder95: I'm under the impression that, if offering max number of idle connections, it needs to come with some companion like an `idle` event that hints when a connection has become idle.

On top, I'm having a bit of doubts about preempting connections like that; on top, they also need to live in cooperation with `clientTtl` at some extent (they cannot/should not be preempted if the ttl hasn't been expired).

JoshMock: > it needs to come with some companion like an `idle` event that hints when a connection has become idle

Can you explain how this would differ from the `drain` event that is already emitted under the same conditions? The only place that `kNeedDrain` is set back to idle is [right before `drain` is emitted](https://github.com/nodejs/undici/blob/f3c5c614901e4ba29f2f60730c3d4f28e60a5186/lib/dispatcher/client.js#L550-L550).

> On top, I'm having a bit of doubts about preempting connections like that; on top, they also need to live in cooperation with `clientTtl` at some extent (they cannot/should not be preempted if the ttl hasn't been expired).

`clientTtl` should definitely be respected. I'll fix that. Is there anything else that would make the preempt logic better?

JoshMock: Support for `clientTtl` added in 60dfc59cd552b2d84dc3cf88308d3077d89d93ba.

metcoder95:
Missed the context completely from that issue, but I can see what you mean; the max number of sockets should be a global limit shared across `undici`.

Limiting, basically, the creation of new Clients if the mas sockets has been reached.

What's the upper limit we want to set by default, `infinity`?

mcollina:
that works

JoshMock:
Thanks, friends. I'll close this and open a new PR as soon as I'm able using the global approach.

PR #4834, opened 2026-02-23
fix: wrap kConnector call in try/catch to prevent client hang

### LINKED ISSUE #4697: Client.request will hang indefinitely if Client[kConnector] throws an error
## Bug Description

The removal of async from `Client.connect` also removed a try/catch around the call to `client[kConnector](...)`, which used to catch errors thrown by `socket = tls.connect(...)` in `lib/core/connect.js`. The catch in `Client.connect` would then perform clean up work, now handled by `handleConnectError` and a call to `client[kResume]()`. Without the try/catch, this no longer occurs, and as a result the Client gets into a bad state wherein additional calls to `Client.request` will hang indefinitely.

## Reproducible By

```javascript

const client = new Client('https://example.com', {
  connect: {
    key: 'bad key',
    cert: 'bad cert',
  }
})

try {
  await client.request({
    method: 'GET',
    path: '/',
  })
} catch (err) {
  console.log(
    'Got the thrown error bubbled up from tls.connect, but the client is now in a bad state...',
    err,
  )
}

try {
  await client.request({
    method: 'GET',
    path: '/',
  })
} catch {
  // irrelevant
}

console.log('We will never get here, because request will hang indefinitely...')

```

Which, when run, will yield:
```sh
Warning: Detected unsettled top-level await at test.mjs:23
  await client.request({
  ^
```

## Expected Behavior

The second call in the example above should throw the same error as the first call, with another (bad) `tls.connect(...)` attempt being made.

## Logs & Screenshots

n/a

## Environment

MacOS v15.7.1, Node v22.20.0

### Additional context

Wrapping the call to `client[kConnector](...)` in a try/catch with the same cleanup as the err handling in its callback seems to fix the issue, i.e.
```javascript
try {
  client[kConnector](...)
} catch (err) {
  handleConnectError(client, err, { host, hostname, protocol, port })
  client[kResume]()
}
```

This is the commit where the problem started occurring: https://github.com/nodejs/undici/commit/56d0cacec76557b9a8d6fea6fb8015517d939661

veeceey:
When `client[kConnector]()` throws synchronously (e.g. from `tls.connect` with a bad cert/key), the error goes unhandled and leaves the client in a bad state where all subsequent `client.request()` calls hang indefinitely.

This wraps the `client[kConnector](...)` invocation in a try/catch that performs the same cleanup as the error callback path — calling `handleConnectError` and `client[kResume]()` — so the client recovers properly and further requests can proceed.

**Repro from the issue:**
```js
const client = new Client('https://example.com', {
  connect: { key: 'bad key', cert: 'bad cert' }
})

// First request throws, but client gets stuck
await client.request({ method: 'GET', path: '/' }).catch(() => {})

// This hangs forever without the fix
await client.request({ method: 'GET', path: '/' })
```

Tested locally — both requests now throw the expected error and the client stays usable.

Fixes #4697

### DIFF
--- a/test/node-test/client-connector-throw.js
+++ b/test/node-test/client-connector-throw.js
@@ -1,0 +1,85 @@
+'use strict'
+
+const { test, after } = require('node:test')
+const { tspl } = require('@matteo.collina/tspl')
+const { Client } = require('../..')
+const { createServer } = require('node:http')
+
+const { closeServerAsPromise } = require('../utils/node-http')
+
+test('client does not hang when connector throws synchronously', async (t) => {
+  const p = tspl(t, { plan: 4 })
+
+  // We need a server just so the URL resolves, but we'll never actually
+  // connect because our custom connector throws before that.
+  const server = createServer({ joinDuplicateHeaders: true }, (req, res) => {
+    res.end()
+  })
+  after(closeServerAsPromise(server))
+
+  server.listen(0, () => {
+    const client = new Client(`http://localhost:${server.address().port}`, {
+      connect: () => {
+        throw new Error('connector boom')
+      }
+    })
+    after(() => client.destroy())
+
+    // First request should get the error from the throwing connector
+    client.request({ path: '/', method: 'GET' }, (err) => {
+      p.ok(err instanceof Error)
+      p.strictEqual(err.message, 'connector boom')
+
+      // Second request should also get an error and not hang
+      client.request({ path: '/', method: 'GET' }, (err) => {
+        p.ok(err instanceof Error)
+        p.strictEqual(err.message, 'connector boom')
+      })
+    })
+  })
+
+  await p.completed
+})
+
+test('client recovers after connector stops throwing', async (t) => {
+  const p = tspl(t, { plan: 4 })
+
+  const server = createServer({ joinDuplicateHeaders: true }, (req, res) => {
+    res.writeHead(200)
+    res.end('ok')
+  })
+  after(closeServerAsPromise(server))
+
+  server.listen(0, () => {
+    let shouldThrow = true
+    const client = new Client(`http://localhost:${server.address().port}`, {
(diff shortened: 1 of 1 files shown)


veeceey:
### Manual test results

Tested with the repro from the issue:

```
$ node test-repro.js
First request error (expected): error:0480006C:PEM routines::no start line
Second request error (expected): error:0480006C:PEM routines::no start line
PASS: Both requests completed without hanging
```

Before the fix, the second request would hang indefinitely. Now both requests throw the expected TLS error and the client stays in a usable state.

Existing test suites pass:
- `client-connect.js` — 1/1 pass
- `client-request.js` — 47/47 pass
- `client-reconnect.js` — 1/1 pass

veeceey:
bumping this one — is there anything on my end that needs updating before review?

veeceey:
gentle ping - would love to get some feedback on this when you get a chance

Invader444:
Have been unable to upgrade undici because of this and other fixes are piling up... can we please get this merged?

Uzlopak:
@Invader444 

Sure. Can you provide a unit test?

veeceey:
@Uzlopak @metcoder95 yep, I should've added a test from the start — that's on me. I'll put together a unit test that reproduces the hang scenario (bad TLS config causing kConnector to throw, second request hanging). will push it shortly.

veeceey:
pushed the tests — two cases covering the sync throw scenario:

1. `client does not hang when connector throws synchronously` — verifies both requests get the error instead of hanging
2. `client recovers after connector stops throwing` — confirms the client can still make successful requests after the connector is fixed

both use a custom `connect` function that throws instead of calling the callback.

Invader444:
How can we retrigger the build? Curious if that was a test timeout or related directly to this change 🤔 @veeceey can you push a no-op commit perhaps?

PR #4842, opened 2026-02-27
fix: recover BalancedPool after equal-weight outage state

### LINKED ISSUE #4839: BalancedPool deteriorates to a single point of failure after outages
## Bug Description

After the configured backends for a balanced pool have failed for a bit, they all end up at equal low-end weights. If a balanced pool is configured with 2 hosts and 1 connection each, at this point the current implementation will always pick the same backend for further requests instead of switching between the 2 options. This causes a disruption to expand unnecessarily until whichever Backend ended up on top of the heap becomes healthy again.

The mechanism seems to work fine when more 2 hosts are added, recovering soon after one host is available.

## Reproducible By

Set up a balanced pool with 2 hosts. Spam some requests while the hosts are down. Observe requests being sent. 

## Expected Behavior

Hosts of equal weight have equal chances of selection, leading to successful recovery when 1 of 2 hosts are available, regardless of ordering.

## Logs & Screenshots

Let me know if a particular log would help the investigation. I'm not familiar with this projects best practices.

## Environment
Undici 7.2.3 in project where I first observed the bug / current main branch for reproduction on private pc
Node 22, MacOS 26.3, Fedora 43

mcollina:
## Summary
- fix `BalancedPool` weighted round-robin recovery after outages when all upstream weights reach the minimum value
- avoid mutating scheduler state during dispatcher availability probes by adding a non-mutating `peek` mode to `kGetDispatcher` callers
- add a regression test that reproduces the 2-upstream outage/recovery scenario from #4839

## Details
When two upstreams both degrade to weight `1`, `BalancedPool` could repeatedly select the same failed upstream and never probe the healthy one during recovery.

This was caused by two interactions:
1. stateful dispatcher selection was being invoked for a "can dispatch?" probe in `PoolBase`, advancing scheduler state unexpectedly
2. weighted round-robin weight reset used `maxWeightPerServer` instead of the current max available upstream weight

## Testing
- `npx borp -p "test/node-test/balanced-pool.js"`
- `npx borp -p "test/pool.js" -p "test/round-robin-pool.js"`
- `npm run lint`

Closes: #4839

### DIFF
--- a/lib/dispatcher/balanced-pool.js
+++ b/lib/dispatcher/balanced-pool.js
@@ -156,7 +156,7 @@ class BalancedPool extends PoolBase {
       .map((p) => p[kUrl].origin)
   }
 
-  [kGetDispatcher] () {
+  [kGetDispatcher] (peek = false) {
     // We validate that pools is greater than 0,
     // otherwise we would have to wait until an upstream
     // is added, which might never happen.
@@ -174,6 +174,10 @@ class BalancedPool extends PoolBase {
       return
     }
 
+    if (peek) {
+      return dispatcher
+    }
+
     const allClientsBusy = this[kClients].map(pool => pool[kNeedDrain]).reduce((a, b) => a && b, true)
 
     if (allClientsBusy) {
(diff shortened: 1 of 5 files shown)


metcoder95 (commented):
tests doesn't seems happy

Issue #4843, opened 2026-02-27
HTTP/2: assert(!this.completed) crash in Request.onData


brunoleite:
## Bug Description

We're hitting assert(!this.completed) in Request.onData ([request.js:265](https://github.com/nodejs/undici/blob/v7.21.0/lib/core/request.js#L265)) in production with HTTP/2
enabled. The process crashes because onData() is called after onComplete() has already set this.completed = true.

Looking at client-h2.js, the [data handler (line 705)](https://github.com/nodejs/undici/blob/v7.21.0/lib/dispatcher/client-h2.js#L705-L709) calls request.onData(chunk) without
checking request.completed or request.aborted. The [end handler (line 711)](https://github.com/nodejs/undici/blob/v7.21.0/lib/dispatcher/client-h2.js#L711-L712) removes data 
listeners before calling onComplete, but the [trailers handler (line 765)](https://github.com/nodejs/undici/blob/v7.21.0/lib/dispatcher/client-h2.js#L765-
L770) calls request.onComplete(trailers) without removing data listeners first.

## Reproducible By

I couldn't reproduce it on localhost, but the code below triggers the assertion error if any data arrives after onComplete has been called.

```
const Request = require("undici/lib/core/request");

const request = new Request("http://localhost", {
  path: "/", method: "GET", headers: {},
}, {
  onConnect() {}, onHeaders() { return true; },
  onData() { return true; }, onComplete() {}, onError() {},
});

request.onConnect(() => {});
request.onHeaders(200, [], () => {}, "OK");
request.onData(Buffer.from("data"));
request.onComplete({});

// This simulates what happens when a late H2 DATA frame
// arrives after END_STREAM has been processed
request.onData(Buffer.from("late"));
```

```
AssertionError [ERR_ASSERTION]: The expression evaluated to a falsy value:

  assert(!this.completed)

    at Request.onData (/node_modules/undici/lib/core/request.js:265:5)
```

Our config:

```
new Agent({
  keepAliveTimeout: 20_000,
  keepAliveMaxTimeout: 60_000,
  bodyTimeout: 10_000,
  headersTimeout: 10_000,
  connections: 500,
  pipelining: 100,
  allowH2: true,
  maxConcurrentStreams: 100,
}).compose(interceptors.responseError())
```

GET requests against remote HTTPS servers. The crash happened at ~6 req/s.

## Expected Behavior

Late data events after stream completion should not crash the process.

## Logs & Screenshots

```
[...]

PR #4845, opened 2026-02-28
fix(h2): ignore late data frames after request completion

mcollina:
## Description

Fixes a crash in the HTTP/2 dispatcher when a late `data` event arrives after the request has already completed.

In `client-h2`, `stream.on('data')` called `request.onData()` unconditionally. If completion happened first (for example via `trailers`), `request.onData()` could hit the internal `assert(!this.completed)` and crash the process.

This change:

- Adds an early return in the H2 `data` handler when `request.aborted || request.completed`
- Removes `data` listeners before invoking `request.onComplete(trailers)` in the `trailers` path
- Adds a regression test that simulates `response -> trailers -> late data`

Fixes: #4843

## Testing

- `npx borp -p "test/http2-late-data.js"`
- `npx borp -p "test/http2-trailers.js"`
- `npx borp -p "test/http2-dispatcher.js"`
- `npm run lint`

### DIFF
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -703,6 +703,10 @@ function writeH2 (client, request) {
   })
 
   stream.on('data', (chunk) => {
+    if (request.aborted || request.completed) {
+      return
+    }
+
     if (request.onData(chunk) === false) {
       stream.pause()
     }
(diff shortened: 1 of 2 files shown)


lib/dispatcher/client-h2.js
hxinhan, line 695: Should we also check if request is complete here?

hxinhan:
Should we merge the PR now?

PR #4852, opened 2026-03-02
ci: disable coverage on Node.js 25

mcollina:
## Summary
- disable Codecov/coverage collection for the Node.js 25 CI matrix entry
- keep coverage enabled only for Node.js 24 on ubuntu-latest
- add an inline note referencing the upstream issue: https://github.com/nodejs/node/issues/61971

## Testing
- npm run lint


mcollina:
merging until a new version of core is out

PR #4860, opened 2026-03-04
fix(env-http-proxy-agent): assume http for schema-less proxy env vars

### LINKED ISSUE #4736: Schema-less HTTP_PROXY and HTTPS_PROXY env vars should maybe assume `http`

nthbotast:
## Summary
- assume an `http://` scheme when `HTTP_PROXY` / `HTTPS_PROXY` values are provided without a URL scheme
- keep existing behavior unchanged for already well-formed proxy URLs
- add coverage for schema-less proxy environment variables in `EnvHttpProxyAgent` tests

## Testing
- node --test test/env-http-proxy-agent.js

Closes #4736

### DIFF
--- a/lib/dispatcher/env-http-proxy-agent.js
+++ b/lib/dispatcher/env-http-proxy-agent.js
@@ -10,8 +10,8 @@ const DEFAULT_PORTS = {
   'https:': 443
 }
 
-function normalizeProxyURI (proxy) {
-  return /^[a-zA-Z][a-zA-Z\d+.-]*:\/\//.test(proxy) ? proxy : `http://${proxy}`
+function normalizeProxyURI (proxy, defaultScheme = 'http://') {
+  return /^[a-zA-Z][a-zA-Z\d+.-]*:\/\//.test(proxy) ? proxy : `${defaultScheme}${proxy}`
 }
 
 class EnvHttpProxyAgent extends DispatcherBase {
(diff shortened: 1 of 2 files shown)


lib/dispatcher/env-http-proxy-agent.js
```diff
@@ -25,14 +29,14 @@ class EnvHttpProxyAgent extends DispatcherBase {
...
-      this[kHttpProxyAgent] = new ProxyAgent({ ...agentOpts, uri: HTTP_PROXY })
+      this[kHttpProxyAgent] = new ProxyAgent({ ...agentOpts, uri: normalizeProxyURI(HTTP_PROXY) })
     } else {
       this[kHttpProxyAgent] = this[kNoProxyAgent]
     }
 
     const HTTPS_PROXY = httpsProxy ?? process.env.https_proxy ?? process.env.HTTPS_PROXY
     if (HTTPS_PROXY) {
-      this[kHttpsProxyAgent] = new ProxyAgent({ ...agentOpts, uri: HTTPS_PROXY })
+      this[kHttpsProxyAgent] = new ProxyAgent({ ...agentOpts, uri: normalizeProxyURI(HTTPS_PROXY) })
```
metcoder95: What's the rationale of normalizing https to http?
unless tunneling, that can cause connect rejection from proxy.

nthbotast:
Good question. The intent here is only to handle *scheme-less* values (e.g. `localhost:8080`) that currently throw on URL parsing. We are **not** rewriting explicit `https://...` proxies.

I defaulted to `http://` because for `HTTP_PROXY`/`HTTPS_PROXY`, the variable name usually indicates the target traffic, while the proxy transport is commonly plain HTTP (CONNECT for TLS upstream), and this is what many existing env setups rely on.

That said, I agree ambiguity exists for scheme-less `HTTPS_PROXY`. If you prefer, I can change this to default `HTTPS_PROXY` to `https://` (or split behavior by variable) and update tests accordingly.

metcoder95:
Let's do that.

I can see a bit of ambiguity if I try to set a schema less origin `proxy.example` that I know only accepts TLS traffic and starts attempting plain text connection.
On top, it seems is what also [curl](https://curl.se/docs/manpage.html#--proxy:~:text=HTTPS_PROXY%20%5Bprotocol%3A//%5D%3Chost%3E%5B%3Aport%5D) does.

nthbotast:
Done 👍\n\nI pushed commit d0235259 to apply the requested split behavior:\n- schemaless  => \n- schemaless  => \n\nI also updated the corresponding env-proxy test expectation to match.\n\nI couldn't run  in this ephemeral checkout because test deps are not installed yet ( missing).

nthbotast:
Follow-up (my previous comment was mangled by shell escaping):

I pushed commit d0235259 to apply the requested split behavior:
- schemaless `HTTP_PROXY` => `http://...`
- schemaless `HTTPS_PROXY` => `https://...`

I also updated the matching env-proxy test expectation.

I couldn't run `node --test test/env-http-proxy-agent.js` in this ephemeral checkout because test deps are not installed yet (`@matteo.collina/tspl` missing).

PR #4863, opened 2026-03-05
fix(interceptor): preserve tuple headers in dns interceptor

### LINKED ISSUE #4444: DNS interceptor breaks headers of type `[string, string][]`

theamodhshetty:
## Summary
- preserve non-object header shapes when the dns interceptor injects the host header
- normalize tuple/iterable header inputs into flat key/value arrays before redispatching
- apply the same host-header handling on DNS fallback retries (other IP family)

## Why
`interceptors.dns()` currently rewrites headers with object spread. For tuple-style headers (for example `[ ['foo', 'bar'] ]`) this can produce malformed headers (`'0': 'foo, bar'`) and break requests in composed interceptor flows.

## Tests
- added `#4444 - Should preserve tuple-style headers` in `test/interceptors/dns.js`
- `node --test test/interceptors/dns.js`
- `npm run lint`

Fixes #4444

### DIFF
--- a/lib/interceptor/dns.js
+++ b/lib/interceptor/dns.js
@@ -5,6 +5,12 @@ const DecoratorHandler = require('../handler/decorator-handler')
 const { InvalidArgumentError, InformationalError } = require('../core/errors')
 const maxInt = Math.pow(2, 31) - 1
 
+function hasSafeIterator (headers) {
+  const prototype = Object.getPrototypeOf(headers)
+  const ownIterator = Object.prototype.hasOwnProperty.call(headers, Symbol.iterator)
+  return ownIterator || (prototype != null && prototype !== Object.prototype && typeof headers[Symbol.iterator] === 'function')
+}
+
 function isHostHeader (key) {
   return typeof key === 'string' && key.toLowerCase() === 'host'
 }
@@ -19,6 +25,19 @@ function normalizeHeaders (headers) {
       return headers
     }
 
+    const normalized = []
+    for (const header of headers) {
+      if (Array.isArray(header) && header.length === 2) {
+        normalized.push(header[0], header[1])
+      } else {
+        normalized.push(header)
+      }
+    }
+
+    return normalized
+  }
+
+  if (typeof headers === 'object' && hasSafeIterator(headers)) {
     const normalized = []
     for (const header of headers) {
       if (Array.isArray(header) && header.length === 2) {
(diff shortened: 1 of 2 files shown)


lib/interceptor/dns.js
```diff
@@ -5,6 +5,100 @@ const DecoratorHandler = require('../handler/decorator-handler')
...
+        normalized.push(header[0], header[1])
+      } else {
+        normalized.push(header)
+      }
+    }
+
+    return normalized
+  }
+
+  if (typeof headers === 'object' && typeof headers[Symbol.iterator] === 'function') {
```
metcoder95: I'd suggest to wait a bit until the changes from this PR are added so we can enhance the iterator check: https://github.com/nodejs/undici/pull/4824/changes#diff-e0a4e92ad66667607d700021123bf26260711f87fda60b44f6867a3f22087410R338

theamodhshetty: Good point, thanks. I removed the generic iterable normalization from this PR and kept the fix scoped to array/object header shapes for now. Once #4824 lands, I can follow up with the safer iterator handling path.

mcollina: I've landed that

theamodhshetty: Picked this up in `bee4c1c`. I kept the tuple-header fix and extended the normalization path to treat safe iterables the same way, so `Map`/`Headers` style inputs survive the DNS host injection too. I also added a DNS interceptor regression for iterable headers and rechecked the behavior with a direct request round-trip locally.

PR #4867, opened 2026-03-05
fix(types): align Response with DOM fetch types

### LINKED ISSUE #3639: (types) undici.Response cannot be assigned to Response
## Bug Description

type undici.Response cannot be assigned to type Response so that module authors can use globally available `Response` as a type but users may use `undici` as the implementation to obtain one.

## Reproducible By

https://gist.github.com/panva/a350238b643aef7a77b29e6db395c3b9

```
git clone https://gist.github.com/panva/a350238b643aef7a77b29e6db395c3b9 response-type-reproduction
cd response-type-reproduction
npm i
npx tsc
```

## Expected Behavior

The types should be compatible.

## Logs & Screenshots

```
index.ts:4:3 - error TS2322: Type 'Promise<import("./node_modules/undici/types/fetch").Response>' is not assignable to type 'Promise<Response>'.
  Type 'import("./node_modules/undici/types/fetch").Response' is not assignable to type 'Response'.
    The types returned by 'headers.entries()' are incompatible between these types.
      Type 'SpecIterableIterator<[string, string]>' is missing the following properties from type 'HeadersIterator<[string, string]>': map, filter, take, drop, and 9 more.

4   return undici.fetch(url)
    ~~~~~~


Found 1 error in index.ts:4
```

theamodhshetty:
## Summary
- align `SpecIterableIterator` with the iterator-helper surface expected by modern DOM `HeadersIterator` types
- add the missing `File[Symbol.toStringTag]` declaration on `node:buffer` so `Response.formData()` stays structurally compatible
- add a `tsd` regression proving Undici `Response` is assignable to `globalThis.Response`

## Why
Newer TypeScript DOM libs model `Headers.entries()`/`keys()`/`values()` as `HeadersIterator`, which includes iterator helper methods. Undici's `SpecIterableIterator` was missing that surface, so `undici.Response` could not be used where a standard `Response` type was expected. Once that was fixed, `formData()` still diverged because Node's `buffer.File` typing was missing the `Symbol.toStringTag` property that exists at runtime.

## Tests
- `npm run test:typescript`
- `npm run lint`

Fixes #3639

### DIFF
--- a/test/types/fetch.test-d.ts
+++ b/test/types/fetch.test-d.ts
@@ -178,6 +178,8 @@ expectType<Promise<Uint8Array>>(response.bytes())
 expectType<Promise<unknown>>(response.json())
 expectType<Promise<string>>(response.text())
 expectType<Response>(response.clone())
+expectAssignable<globalThis.Response>(response)
+expectAssignable<Promise<globalThis.Response>>(fetch(request))
 
 expectType<Request>(new Request('https://example.com', { body: 'Hello, world', duplex: 'half' }))
 expectAssignable<RequestInit>({ duplex: 'half' })
--- a/types/fetch.d.ts
+++ b/types/fetch.d.ts
@@ -60,12 +60,32 @@ export interface SpecIterator<T, TReturn = any, TNext = undefined> {
   next(...args: [] | [TNext]): IteratorResult<T, TReturn>;
 }
 
-export interface SpecIterableIterator<T> extends SpecIterator<T> {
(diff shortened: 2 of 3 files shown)


mcollina (approved):
lgtm

mcollina (approved):
lgtm

Renegade334:
Folks, please please _please_ do not start externally monkeypatching @types/node from here, particularly high-exposure interfaces like this. Because @types/node loads undici's types for every single user, we now can't upgrade undici-types because doing so now causes assignability errors for a load of polyfills, and breaks lib.dom interoperability. We are going to have to pin to an LKG until this is removed.

Can we revert this aspect of the PR?

PR #4869, opened 2026-03-06
docs,test: use RFC 2606 reserved domains in examples

### LINKED ISSUE #4866: Use RFC 2606 TLDs for documentation and tests
https://datatracker.ietf.org/doc/html/rfc2606

>  To safely satisfy these needs, four domain names are reserved as
>    listed and described below.
> 
>                    .test
>                 .example
>                 .invalid
>               .localhost
> 
>       ".test" is recommended for use in testing of current or new DNS
>       related code.
> 
>       ".example" is recommended for use in documentation or as examples.
> 
>       ".invalid" is intended for use in online construction of domain
>       names that are sure to be invalid and which it is obvious at a
>       glance are invalid.
> 
>       The ".localhost" TLD has traditionally been statically defined in
>       host DNS implementations as having an A record pointing to the
>       loop back IP address and is reserved for such use.  Any other use
>       would conflict with widely deployed code which assumes this use.

Basically avoid using domains like example.com etc..

nthbotast:
## Summary
- replace non-reserved example domains in docs examples with RFC 2606-style reserved TLDs
- update node-test fixtures that previously used `google.com`
- update type tests that previously used `www.nodejs.org` as an upstream example

## Why
Closes #4866 by avoiding real/public domains in documentation and test examples.

## Validation
- `npm run test:node-test -- test/node-test/util.js test/node-test/client-dispatch.js`
- `npx tsd --files test/types/balanced-pool.test-d.ts`
- pre-commit lint hook (`npm run lint`)

### DIFF
--- a/test/node-test/rfc2606-doc-placeholders.js
+++ b/test/node-test/rfc2606-doc-placeholders.js
@@ -19,7 +19,8 @@ const disallowedHosts = [
   'example-2.com',
   'example-3.com',
   'mysite.com',
-  'secure.endpoint.com'
+  'secure.endpoint.com',
+  'example.example'
 ]
 
 test('docs examples use RFC2606-style placeholder domains', () => {
--- a/docs/docs/api/Dispatcher.md
+++ b/docs/docs/api/Dispatcher.md
@@ -962,7 +962,7 @@ It accepts the same arguments as the [`RedirectHandler` constructor](/docs/docs/
 const { Client, interceptors } = require("undici");
 const { redirect } = interceptors;
 
-const client = new Client("http://example.example").compose(
+const client = new Client("http://nodejs.example").compose(
   redirect({ maxRedirections: 3, throwOnMaxRedirects: true })
 );
 client.request({ path: "/" })
@@ -980,7 +980,7 @@ It accepts the same arguments as the [`RetryHandler` constructor](/docs/docs/api
 const { Client, interceptors } = require("undici");
 const { retry } = interceptors;
 
-const client = new Client("http://example.example").compose(
+const client = new Client("http://nodejs.example").compose(
   retry({
     maxRetries: 3,
     minTimeout: 1000,
@@ -1006,7 +1006,7 @@ The `dump` interceptor enables you to dump the response body from a request upon
 const { Client, interceptors } = require("undici");
 const { dump } = interceptors;
 
-const client = new Client("http://example.example").compose(
+const client = new Client("http://nodejs.example").compose(
   dump({
     maxSize: 1024,
   })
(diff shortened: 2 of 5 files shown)


mcollina:
did you find all of them? Seems too few.

Can you add some kind of automation on CI to avoid them in the future?

nthbotast:
Implemented, thanks for the review.\n\nI pushed commit  with two follow-ups:\n\n1. **Expanded placeholder-domain cleanup** in docs examples (replaced remaining non-RFC2606 placeholders in the touched API docs).\n2. **Added CI guard test**:  to fail if those non-RFC2606 placeholders are reintroduced in these docs.\n\nValidation run:  and .

nthbotast:
(Correction for formatting)

Implemented, thanks for the review.

I pushed commit 3fd8bd48 with two follow-ups:

1. Expanded placeholder-domain cleanup in docs examples (replaced remaining non-RFC2606 placeholders in the touched API docs).
2. Added CI guard test: test/node-test/rfc2606-doc-placeholders.js to fail if those non-RFC2606 placeholders are reintroduced in these docs.

Validation run: node --test test/node-test/rfc2606-doc-placeholders.js and git diff --check.

test/node-test/rfc2606-doc-placeholders.js
metcoder95, line 1: I believe this could be better if offloaded into a GHA, on top, maybe iterate over the docs and verify each link, attempting to identify if they are following the RFC-2606 standard

Uzlopak: Yeah, somehow gives false sense of security.

Uzlopak:
May we have some nicer domains?

It now looks kind strange, when i read `example.example` in the docs. Reads like we didnt where motivated to give some good domain name for our documentation, which is our "business card".

Lets use nodejs.example.and nodejs.test

nthbotast:
Good point — agreed on readability.\n\nI pushed commit 554aec2a to switch the touched examples from  to clearer reserved domains:\n-  (majority of examples)\n-  (crawling example)\n\nI also updated the placeholder guard test to forbid  in these files going forward.\n\nValidation run:\n- \n-

nthbotast:
(Correction for formatting)

Good point — agreed on readability.

I pushed commit 554aec2a to switch the touched examples from `example.example` to clearer reserved domains:
- `nodejs.example` (majority of examples)
- `nodejs.test` (crawling example)

I also updated the placeholder guard test to forbid `example.example` in these files going forward.

Validation run:
- `node --test test/node-test/rfc2606-doc-placeholders.js`
- `git diff --check`

Fdawgs:
@nthbotast is an AI bot btw, see the flood of PRs here and on the Fastify repo.

PR #4871, opened 2026-03-06
docs(fetch): document forbidden request headers

nthbotast:
## Summary\n- add a `Forbidden headers` section to the Fetch docs\n- explain why headers like `connection` are rejected\n- link to the Fetch spec list of forbidden request-header names\n\nFixes #1470\n\n## Validation\n- `git diff --check`\n- `npx --yes markdownlint-cli2 docs/docs/api/Fetch.md` *(fails on pre-existing issues in this file: MD013 at lines 3/11/50 and MD025 for existing second H1 `# Body Mixins`)*


KhafraDev:
If you can't bother to preview the PR before opening, why should we spend time reviewing it?

PR #4872, opened 2026-03-08
docs(eventsource): document connection lifecycle and close usage

### LINKED ISSUE #536: How to handle text/event-stream?
I think `text/event-stream` responses should be handled as an upgrade and the socket released from the undici client (like with websockets).

nthbotast:
## Summary
- add a dedicated section describing how `EventSource` manages long-lived `text/event-stream` connections
- document that `.close()` should be called to release connection resources when no longer needed
- add a small example with `onmessage`, `onerror`, and explicit shutdown

Closes #536

## Validation
- `git diff --check`

### DIFF
--- a/docs/docs/api/EventSource.md
+++ b/docs/docs/api/EventSource.md
@@ -19,6 +19,31 @@ eventSource.onmessage = (event) => {
 }
 ```
 
+## Connection lifecycle (`text/event-stream`)
+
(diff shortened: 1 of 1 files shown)


metcoder95 (commented):
Don't mind these kind of contributions as are not making much of a harm, but would appreciate to state that as disclosure of the PR description

Issue #4880, opened 2026-03-10
fix(h2): TypeError: Cannot read properties of null (reading 'push') in RequestHandler.onData under high H2 concurrency


hxinhan:
## Bug Description

TypeError: Cannot read properties of null (reading 'push')

## Reproducible By

Dispatcher config

```
new Agent({
  keepAliveTimeout: 20_000,
  keepAliveMaxTimeout: 60_000,
  bodyTimeout: 10_000,
  headersTimeout: 10_000,
  connections: 500,
  pipelining: 100,
  allowH2: true,
  maxConcurrentStreams: 100,
}).compose(interceptors.responseError())
```

## Expected Behavior

onData should not throw when a data event fires after a stream has been torn down. Chunks arriving after stream abort should be silently dropped, and all requests should either succeed or fail gracefully without crashing.

## Logs & Screenshots

```
"Cannot read properties of null (reading 'push')"
First error stack:
TypeError: Cannot read properties of null (reading 'push')
    at RequestHandler.onData (undici/lib/api/api-request.js:148:21)
    at WrapHandler.onResponseData (undici/lib/handler/wrap-handler.js:74:31)
    at ResponseErrorHandler.onResponseData (undici/lib/handler/decorator-handler.js:47:42)
    at ResponseErrorHandler.onResponseData (undici/lib/interceptor/response-error.js:48:20)
    at UnwrapHandler.onData (undici/lib/handler/unwrap-handler.js:81:35)
    at Request.onData (undici/lib/core/request.js:271:29)
    at ClientHttp2Stream.<anonymous> (undici/lib/dispatcher/client-h2.js:706:17)
```

## Environment

* OS Linux
* Node.js v20.19.0
* undici v7.21.0
* allowH2: true

### Additional context

Two changes work together to close the race:

#### client-h2.js — move data listener inside response handler
Previously the data listener was registered at stream creation time, meaning data events could fire before onHeaders had run and RequestHandler.res was set. The fix moves it inside the response callback so it's only registered after res is guaranteed to exist:
```
// Before
stream.once('response', headers => {
  // ...
  request.onHeaders(...)
})

stream.on('data', (chunk) => {          // ← registered too early
  if (request.onData(chunk) === false) {
    stream.pause()
  }
})

// After
stream.once('response', headers => {
  // ...
  request.onHeaders(...)

  // Only register after onHeaders has run and res is set
  stream.on('data', (chunk) => {
    if (request.onData(chunk) === false) {
      stream.pause()
[...]

PR #4881, opened 2026-03-10
fix(h2): TypeError: Cannot read properties of null (reading 'push') i…

hxinhan:
## This relates to...
https://github.com/nodejs/undici/issues/4880


hxinhan:
> Can you solve the TOCTOU without adding the ? guard. That sit in a hot path, and it's likely that it will add some overhead.

I'm not sure how to avoid a non-null check for `res`. I've removed that now.

Only with the change of moving `stream.on('data', ...)`, the issue is still not reproducible in the regression test issue-4880.js

mcollina (approved):
lgtm

PR #4891, opened 2026-03-13
remove maxDecompressedMessageSize

KhafraDev:
If you want to set the limit, use ws.

### DIFF
--- a/test/websocket/permessage-deflate-limit.js
+++ b/test/websocket/permessage-deflate-limit.js
@@ -2,60 +2,8 @@
 
 const { test } = require('node:test')
 const { once } = require('node:events')
-const http = require('node:http')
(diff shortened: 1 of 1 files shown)


mcollina:
Can we add this to the dispatcher or something similar?

mcollina:
Sorry for this. I fixed the vuln and didn’t realize this should have been hardcoded, we typically add security overrides when doing sec work.

mcollina (approved):
lgtm

KhafraDev:
> Can we add this to the dispatcher or something similar?

I think we could add a new WebSocketDispatcher or similar, just wanted to get this in before people started using it.

mcollina:
Yep. I got it out and it will not be in Node.js

PR #4906, opened 2026-03-17
docs: fix docsify sidebar links

DeoJin:
## Summary
- switch `docs/docsify/sidebar.md` from root-relative links to file-relative links so the sidebar works when viewed directly on GitHub
- fix the SOCKS5 sidebar entry to point at the existing `Socks5ProxyAgent.md` page

## Validation
- powershell link check to verify every sidebar target resolves on disk


mcollina:
have you verified this work in docsify too?

Issue #3741, opened 2024-10-16
Provide a more descriptive error on fetching "file" scheme


kettanaito:
## This would solve...

The current error is cryptic, to say the least:

https://github.com/nodejs/undici/blob/68107da3fdd6cd0397f4800a2bc9630c802eda58/lib/web/fetch/index.js#L945

## The implementation should look like...

```js
makeNetworkError('Failed to fetch the "file" URL scheme: not implemented')
```

kettanaito:
The comment above the network error also mentions that file fetching is left as an exercise for the reader. Is there a workaround for this, somehow? 

I'm stumbling upon this through another library so I have no control over that fetch request. Can I "teach" Undici to fetch files? Can I override this behavior in any way, not being in charge of that request-issuing library? (it loads a WASM)

mcollina:
Thanks for reporting! Would you like to send a Pull Request to address this issue? Remember to add unit tests.

KhafraDev:
> Can I "teach" Undici to fetch files? Can I override this behavior in any way, not being in charge of that request-issuing library?

We've decided in previous discussions not to support fetching file urls, it cannot be changed. Realistically, there isn't a good way for us to satisfy security requirements and user requirements.

PR #4907, opened 2026-03-18
feat(fetch): support file URLs via opt-in dispatcher interceptor

mcollina:
## Summary
- add a new `interceptors.file()` dispatcher interceptor for handling `file:` URLs
- keep policy app-defined and deny-by-default via `allow()` callback
- do **not** rely on Node process permissions
- wire `fetch()` file scheme path to dispatcher so custom dispatchers/interceptors can handle it

## Details
- New interceptor: `lib/interceptor/file.js`
- Exported via `interceptors.file` in `index.js`
- Added types in `types/interceptors.d.ts`
- Added docs in `docs/docs/api/Dispatcher.md`

## Tests
- `test/interceptors/file.js`
  - serves allowed files
  - blocks disallowed files
  - passes through non-file requests
- `test/fetch/file-url-interceptor.js`
  - default `fetch(file://...)` rejects
  - `fetch(file://...)` works with custom file interceptor
  - disallowed policy still rejects

### DIFF
--- a/docs/docs/api/Dispatcher.md
+++ b/docs/docs/api/Dispatcher.md
@@ -1275,6 +1275,60 @@ All deduplicated requests receive the complete response including status code, h
 
 For observability, request deduplication events are published to the `undici:request:pending-requests` [diagnostic channel](/docs/docs/api/DiagnosticsChannel.md#undicirequestpending-requests).
 
+##### `file`
+
+The `file` interceptor allows a dispatcher to serve `file:` URLs.
+
+By default, Undici `fetch()` does not read `file:` URLs. This interceptor is an explicit opt-in mechanism that lets applications define their own policy.
+
+**Security model**
+
+- Deny by default.
+- You must provide an `allow()` policy callback.
+- No reliance on Node's process permission model.
+
+**Options**
+
+- `allow({ path, url, method, opts })` - Policy callback. Must return `true` to allow access. Default denies all paths.
+- `resolvePath(url)` - Converts a `file:` URL into a filesystem path. Default: `node:url.fileURLToPath`.
+- `read(path)` - Reads file content. Default: `node:fs/promises.readFile`.
+- `contentType({ path, url, method, opts })` - Optional callback to set a `content-type` header.
+
+**Example - Enable file URLs for a specific directory with `fetch`**
+
+```js
+const { Agent, fetch, interceptors } = require('undici')
+
+const root = '/srv/static'
+
+const dispatcher = new Agent().compose(interceptors.file({
+  allow: ({ path }) => path.startsWith(root)
+}))
+
+const response = await fetch(new URL('file:///srv/static/readme.txt'), {
+  dispatcher
+})
+
+console.log(await response.text())
+```
+
+**Example - Custom content type resolver**
+
+```js
+const nodePath = require('node:path')
+const { Agent, request, interceptors } = require('undici')
+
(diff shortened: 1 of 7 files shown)


lib/web/fetch/index.js
```diff
@@ -951,9 +951,9 @@ function schemeFetch (fetchParams) {
       }))
     }
     case 'file:': {
-      // For now, unfortunate as it is, file URLs are left as an exercise for the reader.
```
KhafraDev, line 992: This comment comes from the spec and should be kept.

mcollina:
What do you think @KhafraDev, will this be a better model?

KhafraDev:
Yes, I have no problems with it. Is there an interest in this again, I saw my old PR getting activity too?

travisbreaks (commented):
Nicely designed interceptor. The deny-by-default security model and clean dispatch pattern are solid. A few observations:

**1. Symlink traversal**
The `allow()` callback receives the path from `resolvePath` (default: `fileURLToPath`), but `fileURLToPath` does not resolve symlinks. If a user's policy allows `/srv/static/` and `/srv/static/link` is a symlink to `/etc/passwd`, the path check passes but the read accesses sensitive data outside the allowed directory.

Consider documenting that consumers should use `fs.realpath` in their `allow()` callback if symlink traversal is a concern, or make `resolvePath` resolve symlinks by default:
```js
resolvePath = async (url) => {
  const { realpath } = require('node:fs/promises')
  return realpath(fileURLToPath(url))
}
```

**2. Unhandled rejection edge case**
The IIFE `(async () => { ... })()` returns a floating promise. If any synchronous code before the `try` block throws (unlikely in current code but possible in future edits), it becomes an unhandled rejection. Consider:
```js
;(async () => { ... })().catch(err => {
  handler.onResponseError?.(controller, err)
  handler.onError?.(err)
})
```

**3. Test coverage gaps**
Tests cover allow/deny/passthrough, which covers the core contract. A few more cases would strengthen confidence:
- Path traversal: `file:///srv/static/../../../etc/passwd` (after `fileURLToPath` normalization)
- Non-existent file (does the error message from `readFile` leak internal paths?)
- Custom `contentType` callback returning the value is reflected in response headers

**4. Minor: `toFileURL` origin check**
The `opts.origin instanceof URL` branch is a nice touch for supporting both string and URL origins.

Clean, well-scoped feature. The main thing I'd want addressed before merge is the symlink consideration, at minimum in documentation.

mcollina:
> Yes, I have no problems with it. Is there an interest in this again, I saw my old PR getting activity too?

it was marked as semver-major, and I took a look at it again before cutting undici 8.

kjsik11:
Without an interceptor, `fetch('file:///...')` on this branch rejects with `InvalidArgumentError: opts.origin must be a non-empty string or URL.` as the cause.

Related to #3741 — wouldn't it be better for the default case to return a clearer error here?

docs/docs/api/Dispatcher.md
```diff
@@ -1318,6 +1275,60 @@ All deduplicated requests receive the complete response including status code, h
...
+
+**Security model**
+
+- Deny by default.
+- You must provide an `allow()` policy callback.
+- No reliance on Node's process permission model.
+
+**Options**
+
+- `allow({ path, url, method, opts })` - Policy callback. Must return `true` to allow access. Default denies all paths.
```
metcoder95, line 1292: Shall we add `maxFileSize` option?

PR #4914, opened 2026-03-19
fix: assume http/https scheme for scheme-less proxy env vars

### LINKED ISSUE #4736: Schema-less HTTP_PROXY and HTTPS_PROXY env vars should maybe assume `http`

travisbreaks:
## Summary

Fixes #4736.

When `HTTP_PROXY` or `HTTPS_PROXY` environment variables (or the `httpProxy`/`httpsProxy` constructor options) are set without a URL scheme (e.g. `localhost:8080` instead of `http://localhost:8080`), `EnvHttpProxyAgent` now prepends the appropriate scheme before passing the value to `ProxyAgent`.

- `HTTP_PROXY` / `httpProxy` values default to `http://`
- `HTTPS_PROXY` / `httpsProxy` values default to `https://`
- Values that already include a scheme (e.g. `http://`, `https://`, `socks5://`) are unchanged

This matches the behavior of [curl](https://curl.se/libcurl/c/CURLOPT_PROXY.html) and [Go's httpproxy package](https://pkg.go.dev/golang.org/x/net/http/httpproxy), which both assume `http://` for scheme-less proxy values. The split behavior (`http://` for HTTP_PROXY, `https://` for HTTPS_PROXY) was requested by a maintainer in #4860 (comment) and aligns with what curl does.

## Changes

- Added `normalizeProxyUrl()` helper in `lib/dispatcher/env-http-proxy-agent.js`
- Applied normalization to both env var and constructor option code paths
- Added 6 new test cases covering scheme-less values for env vars, options, hostname-only values, and the no-op case for URLs that already have a scheme

## Test plan

- [x] All 40 tests in `test/env-http-proxy-agent.js` pass (34 existing + 6 new)
- [x] Lint passes

### DIFF
--- a/lib/handler/cache-handler.js
+++ b/lib/handler/cache-handler.js
@@ -493,8 +493,16 @@ function determineDeleteAt (now, cacheControlDirectives, staleAt) {
     staleIfError = staleAt + (cacheControlDirectives['stale-if-error'] * 1000)
   }
 
-  if (staleWhileRevalidate === -Infinity && staleIfError === -Infinity) {
+  if (cacheControlDirectives.immutable && staleWhileRevalidate === -Infinity && staleIfError === -Infinity) {
     immutable = now + 31536000000
+  }
+
+  // When no stale directives or immutable flag, add a revalidation buffer
+  // equal to the freshness lifetime so the entry survives past staleAt long
+  // enough to be revalidated instead of silently disappearing.
+  if (staleWhileRevalidate === -Infinity && staleIfError === -Infinity && immutable === -Infinity) {
+    const freshnessLifetime = staleAt - now
+    return staleAt + freshnessLifetime
   }
 
   return Math.max(staleAt, staleWhileRevalidate, staleIfError, immutable)
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "undici",
-  "version": "7.24.4",
+  "version": "7.24.5",
   "description": "An HTTP/1.1 client, written from scratch for Node.js",
   "homepage": "https://undici.nodejs.org",
   "bugs": {
--- a/test/interceptors/cache.js
+++ b/test/interceptors/cache.js
@@ -7,6 +7,7 @@ const { equal, strictEqual, notEqual, fail } = require('node:assert')
 const { setTimeout: sleep } = require('node:timers/promises')
 const FakeTimers = require('@sinonjs/fake-timers')
 const { Client, interceptors, cacheStores: { MemoryCacheStore } } = require('../../index')
+const { makeCacheKey } = require('../../lib/util/cache.js')
 
 describe('Cache Interceptor', () => {
   test('caches request', async () => {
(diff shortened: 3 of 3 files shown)


williammartin:
For what it's worth, I don't think curl treats schemaless `HTTPS_PROXY` values as `https://`, under the [`--proxy` flag](https://curl.se/docs/manpage.html#--proxy).

> The proxy string can be specified with a protocol:// prefix. No protocol specified or http:// it is treated as an HTTP proxy.

This is because `HTTPS_PROXY` specifies "which proxy to use for HTTPS requests", not "this proxy speaks HTTPS." As @metcoder95 called out in https://github.com/nodejs/undici/pull/4860#discussion_r2884828366, this might cause issues, this is beyond my knowledge, just clarifying what I think curl's behaviour is here.

I believe this slopcode script also demonstrates this behavior, by showing that there is no TLS handshake if the scheme is left out:

```
We set HTTPS_PROXY three ways and capture the raw bytes curl
sends to a local TCP listener. No real proxy needed.

━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
  HTTPS_PROXY=127.0.0.1:17730
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  → Plaintext HTTP CONNECT (curl guessed http://)

    Host: example.com:443
    User-Agent: curl/8.7.1
    Proxy-Connection: Keep-Alive


━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
  HTTPS_PROXY=http://127.0.0.1:17731
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  → Plaintext HTTP CONNECT (same as no-scheme)

    Host: example.com:443
    User-Agent: curl/8.7.1
    Proxy-Connection: Keep-Alive


━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━
  HTTPS_PROXY=https://127.0.0.1:17732
━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  → TLS ClientHello (0x16 0x03 = TLS record header)

    00000000: 1603 0101 2501 0001 2103 030a eb6d 9698  ....%...!....m..
    00000010: 544a f913 c62b dbfe 8793 d99b 11ea fe9f  TJ...+..........
    00000020: 3b46 8f10 409b 5ef8 49dc f620 ca69 8aef  ;F..@.^.I.. .i..
    00000030: 40a7 a85b d8bb ba3e a956 b6e4 423e 2ca3  @..[...>.V..B>,.
    ...
```

<details>

#!/usr/bin/env bash
#
# demo-proxy-scheme-guess.sh
#
# Demonstrates that curl guesses "http://" when HTTPS_PROXY has no scheme.
# Uses a tiny Python TCP listener to capture the raw bytes curl sends,
# proving the proxy type without needing a real proxy.
#
set -euo pipefail

PORT_BASE=17730  # arbitrary high port range
TMPDIR=$(mktemp -d)
trap 'rm -rf "$TMPDIR"' EXIT

# listen_once PORT OUTFILE — accept one connection, save first recv, exit
listen_once() {
  python3 -c "
import socket, sys
s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
s.setsockopt(socket.SOL_SOCKET, socket.SO_REUSEADDR, 1)
s.settimeout(5)
s.bind(('127.0.0.1', int(sys.argv[1])))
s.listen(1)
try:
    conn, _ = s.accept()
    conn.settimeout(1)
    open(sys.argv[2], 'wb').write(conn.recv(4096))
    conn.close()
except Exception:
    open(sys.argv[2], 'wb').write(b'(no data received)')
s.close()
" "$1" "$2" &
}

run_test() {
  local env_val="$1" port="$2" outfile="$3" mode="${4:-text}" description="$5"

  listen_once "$port" "$outfile"
  local listener_pid=$!
  sleep 0.3

  HTTPS_PROXY="$env_val" curl -s --connect-timeout 2 https://example.com 2>/dev/null &
  local curl_pid=$!

  wait "$listener_pid" 2>/dev/null || true
  wait "$curl_pid"     2>/dev/null || true

  echo "━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━"
  echo "  HTTPS_PROXY=$env_val"
  echo "━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━"
  echo ""
  if [[ "$mode" == "text" ]]; then
    echo "  → $description"
    echo ""
    sed 's/^/    /' "$outfile"
  else
    echo "  → $description"
    echo ""
    xxd "$outfile" | head -4 | sed 's/^/    /'
    echo "    ..."
  fi
  echo ""
}

cat <<'HEADER'

╔═══════════════════════════════════════════════════════════╗
║  Demo: What scheme does curl guess for a schemeless      ║
║  proxy in HTTPS_PROXY?                                   ║
║                                                          ║
║  Answer: http:// — it sends a plaintext HTTP CONNECT.    ║
╚═══════════════════════════════════════════════════════════╝

We set HTTPS_PROXY three ways and capture the raw bytes curl
sends to a local TCP listener. No real proxy needed.

HEADER

# ── Test 1: No scheme (the interesting case) ──────────────────────
p=$((PORT_BASE))
run_test "127.0.0.1:$p" "$p" "$TMPDIR/no_scheme.txt" text \
  "Plaintext HTTP CONNECT (curl guessed http://)"

# ── Test 2: Explicit http:// (should be identical to Test 1) ─────
p=$((PORT_BASE + 1))
run_test "http://127.0.0.1:$p" "$p" "$TMPDIR/http.txt" text \
  "Plaintext HTTP CONNECT (same as no-scheme)"

# ── Test 3: Explicit https:// (TLS handshake, not plaintext) ─────
p=$((PORT_BASE + 2))
run_test "https://127.0.0.1:$p" "$p" "$TMPDIR/https.bin" binary \
  "TLS ClientHello (0x16 0x03 = TLS record header)"

cat <<'FOOTER'
┌───────────────────────────────────────────────────────────┐
│  Conclusion                                               │
│                                                           │
│  Tests 1 and 2 produce identical plaintext HTTP CONNECT   │
│  requests — a schemeless proxy is treated as http://.     │
│                                                           │
│  Test 3 sends a TLS ClientHello instead, proving that     │
│  https:// makes curl encrypt its connection TO the proxy  │
│  before sending the CONNECT request inside the tunnel.    │
│                                                           │
│  This matters: without the https:// prefix, your proxy    │
│  credentials and the target hostname are sent in the      │
│  clear, even when the final destination is HTTPS.         │
│                                                           │
│  The guessing happens in guess_scheme() in lib/urlapi.c   │
│  — it defaults to "http" for any hostname that doesn't    │
│  start with ftp., dict., ldap., imap., smtp., or pop3.   │
└───────────────────────────────────────────────────────────┘
FOOTER

</details>

PR #4920, opened 2026-03-21
fix(fetch): accept global FormData in BodyInit

### LINKED ISSUE #4918: Issue with formData post request with latest 6.24.1 version
## Bug Description

Hi everyone,
I encountered an issue with multipart/form-data fetch requests after upgrading from `6.24.0` to `6.24.1`.
Could you please investigate.

## Reproducible By

From a quick look while debugging, I’m passing the same data to undici’s fetch in both versions. => 
```
fetchParams {
  method: 'POST',
  headers: {
    accept: 'application/json, application/problem+json',
    authorization: 'Bearer ......'
  },
  body: FormData {
    name: 'some',
    someOtherProperty: '8000'
  },
  redirect: 'follow',
  signal: AbortSignal { aborted: false }
}
```

But starting from `6.24.1` I noticed that my generic NestAPI receives `undefined` values for the same endpoint.
 
```
  ✗ POST /some-endpoint

    Request URL: http://localhost:4096/some
    Request Headers:
      accept: application/json, application/problem+json
      authorization: Bearer ********
    Request Body:
      {
        "name": "some",
        "someOtherProperty": "8000",
      } 
 

    Response status code: 400
    Response time: 2 ms
    Response Headers:
      access-control-allow-credentials: true
      connection: keep-alive
      content-length: 445
      content-type: application/problem+json; charset=utf-8
      date: Fri, 20 Mar 2026 12:58:31 GMT
      etag: W/"1bd-V4YJteIrvGmGoHeCtvLwBMlEjS0"
      keep-alive: timeout=5
      vary: Origin
      x-powered-by: Express
    Response Size: 445 bytes
    Response Body:
      {
        "type": "about:blank",
        "title": "Validation failed",
        "code": "BAD_REQUEST",
        "message": "Validation failed",
        "errors": [
          {
            "code": "invalid_type",
            "expected": "string",
            "received": "undefined",
            "path": [
              "name"
            ],
            "message": "Required"
          },
          {
            "code": "invalid_type",
            "expected": "string",
            "received": "undefined",
            "path": [
              "someOtherProperty"
            ],
            "message": "Required"
          }
        ]
      } 
 ```

## Expected Behavior

I would expect multipart/form-data to work the same way as in previous version of `undici`.

## Environment

NODE: v24.13.0
NPM: v11.10.1

### Additional context

I can see same issue on never version `7.24.5`.

mcollina:
## Summary

Fix fetch body handling so `globalThis.FormData` is treated as multipart form data instead of being coerced to `"[object FormData]"`.

## Problem

`fetch()`, `new Request()`, and `new Response()` were using a strict Undici `FormData` brand check in the fetch body conversion path. That rejected compatible global/native `FormData` instances and caused them to fall through to string coercion.

As a result, callers could observe requests being sent with:

- `content-type: text/plain;charset=UTF-8`
- body: `"[object FormData]"`

instead of a proper `multipart/form-data` payload.

## Fix

Replace the strict `webidl.is.FormData(...)` checks in the fetch body conversion path with `util.isFormDataLike(...)`.

This matches the rest of Undici’s cross-realm / third-party FormData handling better and allows compatible global/native `FormData` instances to be encoded as multipart bodies.

## Changes

- update fetch body extraction to accept FormData-like objects
- update `XMLHttpRequestBodyInit` conversion to accept FormData-like objects
- add regression tests covering:
  - `new Request(..., { body: new globalThis.FormData() })`
  - `new Response(new globalThis.FormData())`
  - `fetch(..., { body: new globalThis.FormData() })`

## Testing

Validated with targeted tests:

- `node --test --test-name-pattern="global FormData" test/fetch/request.js test/fetch/response.js test/fetch/client-fetch.js`

Also verified syntax with `node --check` on the touched files.

## Notes

I could reproduce the regression on current `main` / `7.x`. I could not reproduce it locally on `v6.24.1`, even though the issue report mentions that release as well.

Fixes #4918

### DIFF
--- a/lib/web/fetch/body.js
+++ b/lib/web/fetch/body.js
@@ -106,7 +106,7 @@ function extractBody (object, keepalive = false) {
   } else if (webidl.is.BufferSource(object)) {
     // Set source to a copy of the bytes held by object.
     source = webidl.util.getCopyOfBytesHeldByBufferSource(object)
-  } else if (webidl.is.FormData(object)) {
+  } else if (util.isFormDataLike(object)) {
     const boundary = `----formdata-undici-0${`${random(1e11)}`.padStart(11, '0')}`
     const prefix = `--${boundary}\r\nContent-Disposition: form-data`
 
--- a/lib/web/fetch/response.js
+++ b/lib/web/fetch/response.js
@@ -583,7 +583,7 @@ webidl.converters.XMLHttpRequestBodyInit = function (V, prefix, name) {
     return V
   }
 
-  if (webidl.is.FormData(V)) {
+  if (util.isFormDataLike(V)) {
     return V
   }
 
--- a/test/fetch/client-fetch.js
+++ b/test/fetch/client-fetch.js
@@ -510,6 +510,28 @@ test('post FormData with Blob', (t, done) => {
   })
 })
 
+test('post global FormData', (t, done) => {
+  t.plan(2)
+
+  const body = new globalThis.FormData()
+  body.append('field1', 'asd1')
+
+  const server = createServer({ joinDuplicateHeaders: true }, (req, res) => {
+    t.assert.ok(req.headers['content-type']?.startsWith('multipart/form-data; boundary='))
+    req.pipe(res)
+  })
+  t.after(closeServerAsPromise(server))
+
+  server.listen(0, async () => {
+    const res = await fetch(`http://localhost:${server.address().port}`, {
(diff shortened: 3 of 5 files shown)


tsctx:
We should avoid adding support for third-party or global FormData, as it would bypass validation. It's the user's responsibility to import and use FormData from undici.

It looks like we need to clarify this in the documentation.

PR #4925, opened 2026-03-23
fix(types): Fix clone method type declaration to be an instance method rather than instance property

mistval:
## This relates to...

The `Response` [type declaration](https://github.com/nodejs/undici/blob/b64e7e45cd805e245ccf1f2561a843fa431e2d49/types/fetch.d.ts#L206) currently declares `clone` as an instance property, but it's actually an [instance method](https://github.com/nodejs/undici/blob/main/lib/web/fetch/response.js#L230). This PR updates the type declaration to reflect that, and also makes the equivalent change to the `Request` type declaration which is in the same situation.

## Rationale

The distinction between a function instance property and an instance method rarely matters, but can impact certain use cases, including trying to extend Response and override the method. For example this code emits TS errors yet works fine:

```ts
class ExtendResponse extends Response {
  public override clone(): Response {
    return super.clone();
  }
}
```

TS errors:

> (Second line) Class 'Response' defines instance member property 'clone', but extended class 'ExtendResponse' defines it as instance member function.
> (Third line) Class field 'clone' defined by the parent class is not accessible in the child class via super.

This change also aligns with the built in [TS lib DOM](https://raw.githubusercontent.com/microsoft/TypeScript/refs/heads/main/src/lib/dom.generated.d.ts) which declares the method like so:

```ts
/**
 * The **`clone()`** method of the Response interface creates a clone of a response object, identical in every way, but stored in a different variable.
 *
 * [MDN Reference](https://developer.mozilla.org/docs/Web/API/Response/clone)
 */
clone(): Response;
```

This rationale applies equally to `Request.prototype.clone` too.

## Changes

* Changes the type declaration of `Response.clone()` from `readonly clone: () => Response` to `public clone (): Response`.
* Changes the type declaration of `Request.clone()` from `readonly clone: () => Request` to `public clone (): Request`.

### Features

N/A

### Bug Fixes

N/A

N/A

I can't think of a situation where this would break anything, as both before and after, `typeof Response.prototype.clone` is `() => Response`.

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [x] In review
- [x] Merge ready

### DIFF
--- a/test/types/fetch.test-d.ts
+++ b/test/types/fetch.test-d.ts
@@ -186,3 +186,15 @@ expectNotAssignable<RequestInit>({ duplex: 'not valid' })
 expectType<string[]>(headers.getSetCookie())
 
 expectType<Request>(new Request('https://example.com', request))
+
+expectAssignable<Response>(new (class extends Response {
+  override clone () {
+    return this
+  }
+})())
+
+expectAssignable<Request>(new (class extends Request {
+  override clone () {
+    return this
+  }
+})('https://example.com'))


mistval:
Updated to add a couple of tests to fetch.test-d.ts:

```ts
expectAssignable<Response>(new (class extends Response {
  override clone () {
    return this
  }
})());

expectAssignable<Request>(new (class extends Request {
   override clone () {
    return this
  }
})('https://example.com'));
```

The real test is whether the class declarations error with tsc errors or not. The `expectAssignable()`s are just so we're at least asserting something for form's sake. The `typeof` the clone method hasn't actually changed, so I found it a bit awkward to test this nicely with tsd, but I feel this is reasonable.

Without the fix to the clone method signatures in `fetch.d.ts`, `npx tsd` would now fail with:
```
test\types\fetch.test-d.ts:191:11
✖  191:11  Class Response defines instance member property clone, but extended class ExtendResponseOverrideClone defines it as instance member function.        
✖  199:11  Class Request defines instance member property clone, but extended class ExtendRequestOverrideClone defines it as instance member function.
```

PR #4941, opened 2026-03-29
fix: avoid 401 failures for stream-backed request bodies

### LINKED ISSUE #4910: Providing a request body when the response is 401 fails
## Bug Description

What seems to be from a change in https://github.com/nodejs/undici/pull/4747

Any (non websocket) request that includes a request body and gets a 401 response errors.

```
    TypeError: fetch failed

      at Object.processResponse (node_modules/.pnpm/wrangler@4.75.0/node_modules/wrangler/wrangler-dist/cli.js:31323:21)
      at node_modules/.pnpm/wrangler@4.75.0/node_modules/wrangler/wrangler-dist/cli.js:31709:23

    Cause:
    expected non-null body source

      at makeNetworkError (node_modules/.pnpm/wrangler@4.75.0/node_modules/wrangler/wrangler-dist/cli.js:30097:36)
      at httpNetworkOrCacheFetch (node_modules/.pnpm/wrangler@4.75.0/node_modules/wrangler/wrangler-dist/cli.js:31924:20)
      at httpFetch (node_modules/.pnpm/wrangler@4.75.0/node_modules/wrangler/wrangler-dist/cli.js:31733:37)
      at mainFetch (node_modules/.pnpm/wrangler@4.75.0/node_modules/wrangler/wrangler-dist/cli.js:31495:24)
```

## Reproducible By

https://github.com/beanow-at-crabnebula/undici-401-request-body

### Additional context

The reproduction there uses Cloudflare worker's wrangler.
Which bundles `undici@7.24.4`.

### LINKED ISSUE #4940: fetch() throws network error on POST 401 responses when body source is null
## Bug Description

`fetch()` throws `TypeError: fetch failed` for **POST requests** that receive a **401 response** when the request body is a `ReadableStream` (i.e. `body.source` is `null`). GET requests with 401 responses work correctly since #4769.

This is a remaining edge case from the `isTraversableNavigable` / 401 retry logic introduced in #4747 and partially fixed by #4761 and #4769.

### Root Cause

In [`httpNetworkOrCacheFetch`](https://github.com/nodejs/undici/blob/v7.24.4/lib/web/fetch/index.js#L1647-L1697), the 401 handler block has an early return for POST requests that fires **before** the `return response` fix from #4769:

```javascript
// Step 13 (L1647)
if (response.status === 401 && ... && isTraversableNavigable(...)) {
  // Step 13.2: body check
  if (request.body != null) {                     // POST → true
    if (request.body.source == null) {            // ReadableStream → source is null
      return makeNetworkError('expected non-null body source')  // ← throws here
    }
  }

  // ...

  if (request.useURLCredentials === undefined || isAuthenticationFetch) {
    // ...
    return response  // ← #4769 fix (never reached for POST with ReadableStream body)
  }
}
```

The flow:
1. `isTraversableNavigable()` returns `true` (the TODO stub) → enters the block
2. POST has a body → `request.body != null` is `true`
3. When body was constructed from a `ReadableStream` (e.g. `new Request(url, { body: stream, method: 'POST', duplex: 'half' })`), `body.source` is `null`
4. Returns `makeNetworkError('expected non-null body source')` → `TypeError: fetch failed`
5. The `return response` fix from #4769 at the bottom of the block is **never reached**

### Why GET 401 works

GET requests have no body, so `request.body != null` is `false`, skipping the body check entirely and reaching the `return response` fix from #4769.

## Reproducible Example

```javascript
import { once } from 'node:events';

  res.writeHead(401, { 'Content-Type': 'application/json' });
  res.end(JSON.stringify({ error: 'unauthorized' }));
});

server.listen(0);
await once(server, 'listening');

const url = `http://localhost:${server.address().port}`;

// ✅ GET 401 — works correctly (fixed by #4769)
const getRes = await fetch(url);
console.log('GET status:', getRes.status); // 401

// ❌ POST 401 — throws TypeError: fetch failed
try {
  const postRes = await fetch(url, {
    method: 'POST',
    headers: { 'Content-Type': 'application/json' },
    body: JSON.stringify({ data: 'test' }),
  });
  console.log('POST status:', postRes.status);
} catch (e) {
  console.error('POST threw:', e.message); // "fetch failed"
}

server.close();
```

## Suggested Fix

Two possible approaches:

### Option A: Return response early for null body source (minimal, consistent with #4769)

```diff
  if (request.body != null) {
    if (request.body.source == null) {
-     return makeNetworkError('expected non-null body source')
+     return response
    }
```

### Option B: Make `isTraversableNavigable` return `false` (addresses the root cause)

```diff
  function isTraversableNavigable (navigable) {
-   // TODO
-   return true
+   return false
  }
```

Option B prevents entering the entire 401 retry block, which makes sense since Node.js has no traversable navigable that can prompt users for credentials. This was the approach originally proposed in #4756 by @killagu.

## Impact

This affects any Node.js application using `fetch()` for HTTP APIs that return 401 on POST requests — a very common pattern for authentication endpoints. The workaround is to patch `isTraversableNavigable` to return `false`, but a proper fix in undici would be preferable.

Real-world reproduction: Cloudflare Workers local dev via `@cloudflare/vite-plugin` + miniflare (which uses undici internally). Any POST endpoint returning 401 crashes the dev server.

## Related

- #4747 — PR that introduced `isTraversableNavigable` returning `true`
- #4756 — @killagu's original fix proposal (`isTraversableNavigable → false`), closed in favor of #4761
- #4761 — @KhafraDev's fix for the infinite loop (added `return makeNetworkError()`)
- #4769 — @KhafraDev's follow-up fix (changed to `return response`), fixed GET 401 but missed POST
- #4767 — @xconverge's report that 401 responses always throw

## Environment

- undici: 7.24.4 (also reproduced on 7.24.5)
- Node.js: v24.13.1
- OS: macOS (Darwin 25.3.0)

mcollina:
## This relates to...

- Fixes #4940
- Fixes #4910

## Summary

This PR fixes a bug where `fetch()` throws `TypeError: fetch failed` for POST/PUT requests with ReadableStream bodies when the server responds with a 401 status code.

## Root Cause

The issue occurred in the 401 HTTP authentication handling code in `httpNetworkOrCacheFetch()`. When:
1. A request with a ReadableStream body receives a 401 response
2. `isTraversableNavigable()` returned `true` (a TODO stub)

The code would enter the 401 retry block and check `request.body.source == null`, which is true for ReadableStreams, causing it to throw a network error before the `return response` fix from #4769 could be reached.

### 1. Updated `isTraversableNavigable()` (lib/web/fetch/util.js)

Changed from always returning `true` to properly checking if we have an actual navigable object:

```javascript
function isTraversableNavigable (navigable) {
  // Returns true only if we have an actual traversable navigable object
  // that can prompt the user for credentials. In Node.js, this will always
  // be false since there's no Window object or navigable.
  return navigable != null && navigable !== 'client' && navigable !== 'no-traversable'
}
```

This ensures that in Node.js (which has no Window/navigable), the 401 retry block is never entered, and the response is returned directly.

### 2. Changed network error to return response (lib/web/fetch/index.js)

For the edge case where `isTraversableNavigable()` returns `true` (in browser environments with actual navigables), changed the body source check to return the response instead of a network error:

```javascript
if (request.body.source == null) {
  return response  // Previously: return makeNetworkError('expected non-null body source')
}
```

This is more spec-compliant as it aligns with the Fetch spec discussion in [whatwg/fetch#1132](https://github.com/whatwg/fetch/issues/1132), which allows implementations flexibility when credentials can't be obtained.

## Spec Compliance

The current undici behavior was **not** fully spec-compliant. The Fetch specification:

1. Allows implementations flexibility in handling 401 responses when credentials can't be obtained
2. Discusses returning opaque responses or the 401 response directly (see whatwg/fetch#1132)
3. Does not mandate throwing a network error for stream-backed bodies when not actually retrying

This fix brings undici closer to spec compliance by:
- Properly implementing `isTraversableNavigable()` to distinguish between browser and Node.js contexts
- Returning the 401 response directly when credentials can't be obtained (rather than throwing a network error)
- Only entering the retry logic when we actually have a traversable navigable

### Features

N/A

### Bug Fixes

- Returning 401 for stream-backed request bodies no longer throws `fetch failed: expected non-null body source`
- Properly implements `isTraversableNavigable()` to avoid entering 401 retry block in Node.js

### Breaking Changes

N/A

## Testing

Added test cases for:
1. GET requests with 401 responses (existing test)
2. PUT requests with ReadableStream bodies receiving 401
3. POST requests with JSON bodies receiving 401

All tests pass successfully.

### DIFF
--- a/lib/web/fetch/index.js
+++ b/lib/web/fetch/index.js
@@ -1644,7 +1644,14 @@ async function httpNetworkOrCacheFetch (
   // 14. If response’s status is 401, httpRequest’s response tainting is not "cors",
   //     includeCredentials is true, and request’s traversable for user prompts is
   //     a traversable navigable:
-  if (response.status === 401 && httpRequest.responseTainting !== 'cors' && includeCredentials && isTraversableNavigable(request.traversableForUserPrompts)) {
+  //
+  //     In Node.js there is no traversable navigable to prompt the user, but we
+  //     still need to handle URL-embedded credentials so authentication retries
+  //     for WebSocket handshakes continue to work.
+  if (response.status === 401 && httpRequest.responseTainting !== 'cors' && includeCredentials && (
+    request.useURLCredentials !== undefined ||
+    isTraversableNavigable(request.traversableForUserPrompts)
+  )) {
     // 2. If request’s body is non-null, then:
     if (request.body != null) {
       // 1. If request’s body’s source is null, then return a network error.


mcollina:
Somehow I garbled the file, will fix

mcollina:
@KhafraDev this should be correct for us

KhafraDev (approved):

mcollina:
@KhafraDev ptal

jamesopstad:
@mcollina @KhafraDev Is it possible to get this PR merged and released soon? A lot of Cloudflare users are affected because of Miniflare's dependency on Undici.

PR #4955, opened 2026-04-02
feat: add configurable maxPayloadSize for WebSocket

### LINKED ISSUE #4944: Allow configuring maxDecompressedMessageSize
In order to address the potential per-message deflate decompress bomb, the message size was limited to 4MB in https://github.com/nodejs/undici/commit/2ee00cb322c76b0bf56829462d7c1dc53d1cbe3d with a configuration option.

That configuration option was then removed in #4891 .

I understand the reasoning behind removing the configuration option, but it effectively limits any decompressed message to 4MB without providing the consumer the ability to set their own limit. AFAICT, the only way forward is to use a different library such as [ws](https://www.npmjs.com/package/ws).

The removal PR mentions the ability to provide configuration via a dispatcher. Creating this issue to track progress on that enhancement as we would like to continue using undici bundled in node, but need the ability process messages larger than 4MB.

mcollina:
Fixes #4944

## Changes
- Added dispatcher-level webSocketOptions.maxPayloadSize configuration
- Default: 128 MB for both compressed and uncompressed payloads
- Add estimated expansion check (10x ratio) for decompression bomb protection
- Add raw payload size check before accepting uncompressed data
- Fixed actual size tracking to use configured limit

## Usage
```javascript
import { Agent, WebSocket } from 'undici'

const agent = new Agent({
  webSocket: {
    maxPayloadSize: 128 * 1024 * 1024 // 128 MB
  }
})

const ws = new WebSocket('wss://example.com', { dispatcher: agent })
```

### DIFF
--- a/lib/web/websocket/permessage-deflate.js
+++ b/lib/web/websocket/permessage-deflate.js
@@ -8,36 +8,22 @@ const tail = Buffer.from([0x00, 0x00, 0xff, 0xff])
 const kBuffer = Symbol('kBuffer')
 const kLength = Symbol('kLength')
 
-// Default maximum decompressed message size: 128 MB
-const kDefaultMaxDecompressedSize = 128 * 1024 * 1024
-
 class PerMessageDeflate {
   /** @type {import('node:zlib').InflateRaw} */
   #inflate
 
   #options = {}
 
-  /** @type {number} */
-  #maxPayloadSize
-
-  /** @type {boolean} */
-  #aborted = false
-
-  /** @type {Function|null} */
-  #currentCallback = null
-
-  /** @type {number} */
-  #currentBufferedBytes = 0
+  #maxPayloadSize = 0
 
   /**
    * @param {Map<string, string>} extensions
-   * @param {{ maxPayloadSize?: number }} [options]
    */
-  constructor (extensions, options = {}) {
+  constructor (extensions, options) {
     this.#options.serverNoContextTakeover = extensions.has('server_no_context_takeover')
     this.#options.serverMaxWindowBits = extensions.get('server_max_window_bits')
-    // 0 disables the limit
-    this.#maxPayloadSize = options.maxPayloadSize ?? kDefaultMaxDecompressedSize
+
+    this.#maxPayloadSize = options.maxPayloadSize
   }
 
   /**
@@ -45,19 +31,12 @@ class PerMessageDeflate {
    * @param {Buffer} chunk Compressed data
    * @param {boolean} fin Final fragment flag
    * @param {Function} callback Callback function
-   * @param {number} [bufferedBytes] Bytes already accumulated for the current message
    */
-  decompress (chunk, fin, callback, bufferedBytes = 0) {
+  decompress (chunk, fin, callback) {
     // An endpoint uses the following algorithm to decompress a message.
     // 1.  Append 4 octets of 0x00 0x00 0xff 0xff to the tail end of the
     //     payload of the message.
     // 2.  Decompress the resulting data using DEFLATE.
-
-    if (this.#aborted) {
-      callback(new MessageSizeExceededError())
-      return
-    }
-
     if (!this.#inflate) {
       let windowBits = Z_DEFAULT_WINDOWBITS
 
@@ -80,24 +59,12 @@ class PerMessageDeflate {
       this.#inflate[kLength] = 0
 
       this.#inflate.on('data', (data) => {
-        if (this.#aborted) {
-          return
-        }
-
         this.#inflate[kLength] += data.length
 
-        // 0 disables the limit
-        if (this.#maxPayloadSize > 0 && this.#currentBufferedBytes + this.#inflate[kLength] > this.#maxPayloadSize) {
-          this.#aborted = true
+        if (this.#maxPayloadSize > 0 && this.#inflate[kLength] > this.#maxPayloadSize) {
+          callback(new MessageSizeExceededError())
           this.#inflate.removeAllListeners()
-          this.#inflate.destroy()
           this.#inflate = null
-
-          if (this.#currentCallback) {
-            const cb = this.#currentCallback
-            this.#currentCallback = null
-            cb(new MessageSizeExceededError())
-          }
           return
         }
 
@@ -110,15 +77,13 @@ class PerMessageDeflate {
       })
     }
 
-    this.#currentCallback = callback
-    this.#currentBufferedBytes = bufferedBytes
     this.#inflate.write(chunk)
     if (fin) {
       this.#inflate.write(tail)
     }
 
     this.#inflate.flush(() => {
-      if (this.#aborted || !this.#inflate) {
+      if (!this.#inflate) {
         return
       }
 
@@ -126,8 +91,6 @@ class PerMessageDeflate {
 
       this.#inflate[kBuffer].length = 0
       this.#inflate[kLength] = 0
-      this.#currentCallback = null
-      this.#currentBufferedBytes = 0
 
       callback(null, full)
     })
--- a/lib/web/websocket/receiver.js
+++ b/lib/web/websocket/receiver.js
@@ -72,20 +72,11 @@ class ByteParser extends Writable {
   }
 
   #validatePayloadLength () {
-    if (this.#maxPayloadSize <= 0 || isControlFrame(this.#info.opcode)) {
-      return true
-    }
-
-    if (this.#info.compressed) {
-      if (this.#info.payloadLength > this.#maxPayloadSize) {
-        failWebsocketConnection(this.#handler, 1009, 'Payload size exceeds maximum allowed size')
-        return false
-      }
-
-      return true
-    }
-
-    if (this.#fragmentsBytes + this.#info.payloadLength > this.#maxPayloadSize) {
+    if (
+      this.#maxPayloadSize > 0 &&
+      !isControlFrame(this.#info.opcode) &&
+      this.#info.payloadLength > this.#maxPayloadSize
+    ) {
       failWebsocketConnection(this.#handler, 1009, 'Payload size exceeds maximum allowed size')
       return false
     }
@@ -181,6 +172,10 @@ class ByteParser extends Writable {
         if (payloadLength <= 125) {
           this.#info.payloadLength = payloadLength
           this.#state = parserStates.READ_DATA
+
+          if (!this.#validatePayloadLength()) {
+            return
+          }
         } else if (payloadLength === 126) {
           this.#state = parserStates.PAYLOADLENGTH_16
         } else if (payloadLength === 127) {
@@ -196,10 +191,6 @@ class ByteParser extends Writable {
         this.#info.masked = masked
         this.#info.fin = fin
         this.#info.fragmented = fragmented
-
-        if (this.#state === parserStates.READ_DATA && !this.#validatePayloadLength()) {
-          return
-        }
       } else if (this.#state === parserStates.PAYLOADLENGTH_16) {
         if (this.#byteOffset < 2) {
           return callback()
@@ -268,13 +259,18 @@ class ByteParser extends Writable {
               this.#info.fin,
               (error, data) => {
                 if (error) {
-                  // Use 1009 (Message Too Big) for decompression size limit errors
                   const code = error instanceof MessageSizeExceededError ? 1009 : 1007
                   failWebsocketConnection(this.#handler, code, error.message)
                   return
                 }
 
                 this.writeFragments(data)
+
+                // Check cumulative fragment size
+                if (this.#maxPayloadSize > 0 && this.#fragmentsBytes > this.#maxPayloadSize) {
+                  failWebsocketConnection(this.#handler, 1009, new MessageSizeExceededError().message)
+                  return
+                }
 
                 if (!this.#info.fin) {
                   this.#state = parserStates.INFO


mcollina:
> My issues are still present. By default, code that works in ws, Bun, deno, firefox, chrome, Safari, and every other browser and environment will fail in node.

The security model of Firefox, Chrome and Safari is radically different from what we can offer in Node.js. The challenges are sometimes opposite.

Out-of-memory errors cannot be easily caught by Node.js and V8 (Chrome can just crash a tab) and they are handled as best-effort cases. The only safe way to prevent a crash is to provide a solution to avoid allocating that memory.

I will investigate Bun and Deno if they deal with this in some different way.

KhafraDev:
Different security models, but the same spec. `ws` added a maxPayloadSize option 2 weeks ago (possibly inspired by undici's CVE?) that is disabled by default as well.

mcollina:
I did a pass over what limits other WebSocket implementations actually expose, because RFC 6455 leaves this up to implementations.

A few things seem worth separating:

- **frame / message limits**
- **compressed vs decompressed limits**
- **documented public API limits vs implementation details**

## 1) What the spec says

RFC 6455 explicitly says this is implementation-specific. In §10.4:

> Implementations that have implementation- and/or platform-specific limitations regarding the frame size or total message size after reassembly from multiple frames MUST protect themselves against exceeding those limits ... Such an implementation SHOULD impose a limit on frame sizes and the total message size after reassembly from multiple frames.

Ref: https://www.rfc-editor.org/rfc/rfc6455#section-10.4

So the spec does **not** define one universal limit, but it does explicitly expect implementations to protect themselves.

## 2) Browser-style WebSocket APIs

For **Chrome / Firefox / Safari**, I couldn't find an official public doc that says “the maximum incoming WebSocket message size is N bytes”. What MDN *does* document for the standard browser `WebSocket` API is that it has **no backpressure**, so it can fill memory if messages arrive faster than the application can process them:

- MDN: https://developer.mozilla.org/en-US/docs/Web/API/WebSocket

Deno's browser-style `WebSocket` docs are similar: they expose the standard API, but I couldn't find a documented numeric incoming-message limit there either:

- Deno WebSocket docs: https://docs.deno.com/api/web/~/WebSocket

So for browser-style APIs, the public contract appears to be closer to **"implementation / memory dependent"** than to a stable documented numeric cap.

## 3) Server/runtime implementations that *do* expose explicit limits

These are the clearest documented limits I found:

- **ws**: `maxPayload`, default **100 MiB** (`104857600` bytes)
  - Docs: https://raw.githubusercontent.com/websockets/ws/master/doc/ws.md

- **Bun server**: `maxPayloadLength`, default **16 MB**
  - Bun docs say it closes the connection if it receives a message larger than that.
  - Docs: https://bun.sh/docs/api/websockets

- **uWebSockets.js**: `maxPayloadLength`, default **16 * 1024** (**16 KiB**)
  - If a client sends a larger message, the connection is immediately closed.
  - Docs: https://unetworking.github.io/uWebSockets.js/generated/interfaces/WebSocketBehavior.html

- **Python websockets**: `max_size`, default **1048576** (**1 MiB**)
  - `None` disables the limit.
  - Docs: https://websockets.readthedocs.io/en/stable/reference/asyncio/server.html

- **Socket.IO / Engine.IO**: `maxHttpBufferSize`, default **1e6** (**1 MB**)
  - Docs say this defines how many bytes a single message can be before closing the socket.
  - Docs also note that it matches the `ws` package’s `maxPayload` option.
  - Docs: https://socket.io/docs/v4/server-options/

- **Gorilla WebSocket**: explicit `SetReadLimit(limit int64)` API
  - I didn't find a documented default read limit.
  - If exceeded, docs say it sends a close frame and returns `ErrReadLimit`.
  - Docs: https://pkg.go.dev/github.com/gorilla/websocket

- **Deno server internals**: there is an open issue saying the server currently has an internal **128 MiB** message limit and requesting configurability.
  - This is from issue discussion, not the main public API docs, so I’d treat it as implementation detail rather than stable API contract.
  - Issue: https://github.com/denoland/deno/issues/15809

## 4) What this means for undici

The cross-implementation picture looks like this:

- The **spec** expects implementations to defend themselves.
- **Browsers** don’t seem to publish a stable numeric limit in their public WebSocket API docs.
- **Server/runtime libraries usually do expose an explicit configurable limit**.
- The defaults vary **a lot**: from **16 KiB** (uWS) to **100 MiB** (`ws`).

Also, most of the limits above are **message-size limits in general**, not specifically **decompressed-message-size** limits. In that sense, what this PR adds is actually *more targeted* than many existing APIs, because it directly addresses the `permessage-deflate` expansion case instead of only limiting wire payload size.

So IMO this PR is well aligned with the ecosystem:

1. RFC 6455 says implementations should protect themselves.
2. Server-side WebSocket stacks commonly expose a configurable limit.
3. There is no obvious single "everyone uses X" default across the ecosystem anyway.
4. A configurable **decompressed** size limit is a reasonable server/runtime safeguard, especially in Node where OOM is process-fatal rather than "just crash a tab".

KhafraDev:
Note that permessage-deflate is standardized in [RFC 7692](https://datatracker.ietf.org/doc/html/rfc7692) and not RFC 6455

- ws' maxPayload option: I believe 100 MB limit is server-only. For the client, it is disabled by default. (https://github.com/websockets/ws/blob/843925544e2f4cffe445e0179947f56d6c5b608f/lib/permessage-deflate.js#L42)
- Bun: websocket server
- uws: websocket server
- Deno: you mention the server again, we're talking about the client
- socket.io: server-options in the url
- gorilla/websocket: seems to be disabled by default
- python/websocket: server in the the url

mcollina:
I doesn't matter if it's server or client. If one is receiving data from untrusted sources, they are vulnerable to this attack (in Node.js).

I'll keep digging on more proof points.

mcollina:
A bit more data after testing other runtimes in the same setup (host `ws` server, `perMessageDeflate: true`, client in a `256MB` memory-capped Docker container):

- **Bun 1.3.11 client** negotiated `permessage-deflate` and I could reproduce the same broad failure mode as Node: `64 MiB` and `80 MiB` succeeded, but `96 MiB` and `128 MiB` caused the client process to be killed with **exit 137**. On the server side that showed up as close code **1006**.

- **Deno 2.7.11 client** behaved differently: it did **not** negotiate `permessage-deflate` at all in this repro, so it doesn't exercise the decompression path. Instead it has an explicit size limit and fails cleanly with **`Error: Frame too large`**.
  - Empirically, `67108863` bytes succeeds and `67108864` bytes fails.
  - That matches Deno's current implementation, which uses `fastwebsockets`; `fastwebsockets` defaults `max_message_size` to **`64 << 20`** and rejects when `payload_len >= self.max_message_size`.
  - It also documents that **permessage-deflate is not supported yet**.

References:
- Deno source: https://raw.githubusercontent.com/denoland/deno/main/ext/websocket/lib.rs
- fastwebsockets docs/source (default `64 MiB`, no PMD support): https://docs.rs/fastwebsockets/latest/src/fastwebsockets/lib.rs.html

Given that, one question: would it make sense to also consider a more general **`maxPayload` / `maxMessageSize`** style option, in addition to (or instead of) `maxDecompressedMessageSize`?

My intuition is that `maxDecompressedMessageSize` is the right targeted fix for the expansion problem behind this issue, but a generic payload/message limit may be a better *overall* protection story. At the same time, without compression-driven expansion it probably isn't in the same DoS/CVE class.

mcollina:
I'm also ok in raising the limit to 64MB, I was too conservative with 4MB

tsctx:
I'm inclined to think we should be cautious about introducing limits like 64MB that aren’t explicitly in the specification. Since the client can't currently restrict uncompressed messages, limiting only the compressed path might not provide full protection against OOM crashes; raw data could still potentially be streamed up to the 2GB limit.

Unless we can implement a unified maxPayloadSize for both paths, it might be more consistent to default to the protocol or platform maximum (approx. 2GB). It may be better to let developers configure specific restrictions themselves, rather than having the library impose caps that aren't standard in browser environments.

lpinca:
> `ws` added a maxPayloadSize option 2 weeks ago

No, it is there since [forever](https://github.com/websockets/ws/commit/61de51465b8c78062f269d201792690f97a84497), the [commit](https://github.com/websockets/ws/commit/3ee5349a0b1580f6e1f347b59ec3371011bd8481) you are referring to is just moving a positional argument to an option.

> ws' maxPayload option: I believe 100 MB limit is server-only. For the client, it is disabled by default.

No, it works in the same way for both the client and the server. 0 means no limit.

The extension is enabled by default on the client and is disabled by default on the server due to memory fragmentation issues, see https://github.com/websockets/ws?tab=readme-ov-file#websocket-compression. Anyway a lot has changed/improved since the defaults decision and it is now used in many production environments.

You can count the size of the chunks in the listener of the `'data'` event of the `zlib.InflateRaw` stream to prevent "zip bombs", see https://github.com/websockets/ws/blob/84392554/lib/permessage-deflate.js#L482-L506.

KhafraDev:
> No, it works in the same way for both the client and the server. 0 means no limit.

My bad, in jsdoc `[param=value]` means `value` is the default (https://jsdoc.app/tags-param#optional-parameters-and-default-values)

> You can count the size of the chunks in the listener of the 'data' event of the zlib.InflateRaw stream to prevent "zip bombs"

This is what the PR is doing.

lpinca:
> in jsdoc `[param=value]` means `value` is the default

Yes, but the class is not instantiated with that value (https://github.com/websockets/ws/blob/84392554/lib/websocket-server.js#L299, https://github.com/websockets/ws/blob/84392554/lib/websocket.js#L761).

mcollina:
@tsctx @KhafraDev PTAL

KhafraDev:
I would need a test that causes node to run out of memory without the limit (on main) before I would approve. Right now, the tests only check that messages > the configured payload size close the connection/error.

mcollina:
I'll send it to you privately, I prefer not to disclose publicly.

mcollina:
Shared. Note that it takes just 130KB to generate 128MB of data, it's the kind of expansion that is classified as DoS.

KhafraDev:
> I'll send it to you privately, I prefer not to disclose publicly.

Can we eventually add the repro in undici? I understand why it is private currently.

mcollina:
once the vulnerability is disclosed, yes.

https://hackerone.com/reports/3481206

There is probably enough here to understand the problem and recreate it tbh, but I prefer to not make things easy.

mcollina:
@KhafraDev are you ok for landing this then?

tsctx:
I agree that introducing a limit is reasonable. However, I have concerns about the 64MB default. While it serves as a baseline, it leaves almost no margin for modern, data-intensive applications.

The issue is that since this limit cannot be changed without installing undici, 64MB may inadvertently become a bottleneck for typical development rather than a protective safety net. In real-world scenarios, data transfers can easily spike toward 60-70MB. A 64MB limit is low enough to be hit by legitimate usage, yet 100MB is high enough to accommodate those spikes while still providing robust protection against genuinely malicious payloads.

Since Deno uses 64MB, I can understand that 64MB is appropriate in that context, but I believe that aligning with the 100MB limit in the "ws" library—which you mentioned earlier—is the right direction. It provides a "silent safety measure" that remains invisible to the average user while preventing the need to install undici just to handle slightly larger payloads.

Also, while 64MB makes sense for Deno, it could cause compatibility issues where code that works in a browser (which generally allows for significantly higher message thresholds) fails here. I initially thought we should allow a much higher limit to match the practical spec, but 100MB feels like a very reasonable compromise.

Given that it is difficult for users to change the limit when used as a global implementation, what do you think about setting the default to 100MB to ensure the implementation of undici remains practical without compromising security?

lib/web/websocket/receiver.js
```diff
@@ -212,6 +217,12 @@ class ByteParser extends Writable {
           this.#state = parserStates.INFO
         } else {
           if (!this.#info.compressed) {
+            // Check raw payload size before accepting uncompressed data
+            if (this.#maxPayloadSize > 0 && this.#info.payloadLength > this.#maxPayloadSize) {
+              failWebsocketConnection(this.#handler, 1009, 'Payload size exceeds maximum allowed size')
+              return
+            }
+
```
tsctx: Nit: It might be better to validate the size as soon as the length is determined—whether it's the immediate 0-125 value or the subsequent 16/64-bit extended fields. This prevents us from proceeding to READ_DATA and wasting bandwidth on a request that's bound to fail. It feels a bit odd to wait for the actual data to arrive when the length from the header already provides sufficient grounds for rejection.

mcollina:
I think we can go to 128MB without much problems. The default is 4MB right now, so anything above that would be good.

mcollina:
@tsctx @KhafraDev ptal

lib/web/websocket/receiver.js
```diff
@@ -66,6 +71,20 @@ class ByteParser extends Writable {
     this.run(callback)
   }
 
+  #validatePayloadLength () {
+    if (
+      this.#maxPayloadSize > 0 &&
+      !isControlFrame(this.#info.opcode) &&
+      !this.#info.compressed &&
```
KhafraDev: I don't understand why we ignore compressed messages here, but then perform the same logic in the permessage-deflate handler?

lib/web/websocket/permessage-deflate.js
```diff
@@ -8,15 +8,20 @@ const tail = Buffer.from([0x00, 0x00, 0xff, 0xff])
...
+const kMaxExpansionRatio = 10
 
 class PerMessageDeflate {
   /** @type {import('node:zlib').InflateRaw} */
   #inflate
 
   #options = {}
 
+  /** @type {number} */
+  #maxDecompressedSize
```
KhafraDev: This name is misleading. As far as I can tell, `maxDecompressedSize` is the same as `maxPayloadSize`.

lib/web/websocket/permessage-deflate.js
```diff
@@ -25,18 +30,46 @@ class PerMessageDeflate {
...
+   * @param {number} [compressedLength] Compressed payload length for estimated size check
+   */
+  decompress (chunk, fin, callback, compressedLength) {
     // An endpoint uses the following algorithm to decompress a message.
     // 1.  Append 4 octets of 0x00 0x00 0xff 0xff to the tail end of the
     //     payload of the message.
     // 2.  Decompress the resulting data using DEFLATE.
 
+    // Early rejection based on estimated expansion
+    if (compressedLength != null && this.#exceedsEstimatedLimit(compressedLength)) {
```
KhafraDev: compressedLength is just the payload size, which has already been validated by this point. Any compressed message > 12.8 MB will error, is that intentional?

mcollina: Yes, it's due to the expansion factor. Anything over 12.8MB is gauranteed to be over the threshold, so we can just abort it now.

lib/web/websocket/receiver.js
mcollina: we are doing this kind of check in 3 places:

1. here for normal packets
2. early rejection for compressed packets (estimating a 10x expansion rate, we can avoid trying to decompress things, saves CPU)
3. actual/precise validation for compressed packages

The 2nd could be optional, but it saves CPU cycles.

KhafraDev: but we're skipping compressed packets here, it's not much of an early return if you still have to go all the way to the decompression handler

what I'm asking for is why we're skipping compressed packets here

lib/web/websocket/permessage-deflate.js
KhafraDev: A decompression bomb, not a regular message, right? https://zlib.net/zlib_tech.html

According to this, the worst possible expansion factors are:
- with default settings, ~0.03% overhead (5 bytes per 16 KB block + 6 bytes for the stream)
- with worst non-default settings: 13.5% + 11 bytes
- absolute worst case: 1100% (one byte only)

lib/web/websocket/receiver.js
mcollina: We are skipping compressed packets here because it's a different logic as @tsctx has asked.

lib/web/websocket/permessage-deflate.js
mcollina: I'm ok with whatever you suggesting.

KhafraDev: I think all that is needed is to remove the expected expansion check

lib/web/websocket/receiver.js
KhafraDev: The logic is not different.

Assuming we send an uncompressed and compressed frame that's 129MB:
- Uncompressed: early exits here
- Compressed: skips this, goes through READ_DATA, goes to permessage deflate handler, attempt to inflate, if payload size > max payload size, exit

mcollina:
@KhafraDev ptal

KhafraDev:
https://github.com/nodejs/undici/pull/4955#discussion_r3064563148

I'm not sure why compressed messages are being handled separately in this case still

lib/web/websocket/receiver.js
KhafraDev: For example, what if someone sent a compressed message that ends up totaling 128.1 MBs? This check is skipped despite it being a candidate for the early exit.

mcollina: We skip compressed frames here because `this.#info.payloadLength` is the compressed frame length, not the final message size. After removing the expansion heuristic, we no longer have a reliable way to reject compressed messages at header-parse time. For compressed messages, the correct limit is enforced in `permessage-deflate.js` against the actual inflated byte count. For uncompressed messages, the header length already is the actual payload size, so we can reject early here.

KhafraDev: I understand that, but why skip them? Couldn't an attacker just send a compressed frame that has an unlimited size, where undici will collect all fragments, attempt to inflate, then fail much later?

mcollina: The check is done chunk by chunk on compression, as soon as it goes over the limit, the connection is interrupted.

tsctx: I’ve looked into it, and while we check the max length per message unit, we seem to be missing a check for the total buffered size of fragmented messages.

We don't actually need a separate check for compression; we should simply trigger an early exit if the accumulated size plus the current payload exceeds the limit. Even with compressed data, we can perform this early check if the payload itself is already too large.

This is already handled for uncompressed fragments, but for the compressed path, we should also pass the current buffered size to the decompression side to ensure the same validation.

tsctx: I've written some pseudo-code to illustrate the issue more clearly.

```js
createServer((socket) => {
    // Start by fragmenting the message.
    // At this stage, 64MB has been accumulated in the buffer.
    socket.send(alloc(64 * 1024 * 1024), { fin: false }); 
    
    // Case 1: Early termination is not triggered because compressed messages 
    // are currently being skipped.
    // The pre-compressed size is 100MB. Since 64MB is already buffered, 
    // it clearly exceeds the limit and should be terminated immediately.
    socket.send(alloc_post_compressed(100 * 1024 * 1024), { fin: true }); 
    // alloc_post_compressed: The size represents the "expected final output."

    // Case 2: The current accumulated size is not passed to the decompressor, 
    // so this segment is treated as only 100MB, bypassing the decompression limit.
    // We should pass the current buffer size to the decompressor and 
    // validate the total size (current buffer + decompressed payload).
    socket.send(alloc_pre_compressed(100 * 1024 * 1024), { fin: true }); 
    // alloc_pre_compressed: The size represents the "actual memory footprint."
});

Comment (KhafraDev) [reply lib/web/websocket/receiver.js]: Thank you @tsctx this is what I was trying to articulate, lol.

The issue is if the compressed data is larger than the maxPayloadSize, we shouldn't bother going to READ_DATA and inflating the data which is already larger than the maxPayloadSize.

The permessage-deflate handler catches it, once the *decompressed* data is larger than the maxPayloadSize. This only happens once we have a full frame, so an attacker could theoretically send n fragmented frames each with unlimited size that are kept in memory until we receive the entirety of the frame. 

From READ_DATA, should be helpful in understanding:
```ts
        if (this.#byteOffset < this.#info.payloadLength) {
          return callback()
        }

        const body = this.consume(this.#info.payloadLength)
```

KhafraDev:
I haven't done extensive testing, but https://github.com/KhafraDev/undici/commit/903a6f0f551d3af742199f519abc77ac9b3f8eb2 simplifies this quite a bit (leaving most of the handling to the actual parser, not the permessage deflate handler) and fixes what I was talking about in the unresolved review comment and fixes a few bugs.

mcollina:
Thanks, but I don’t think we can take that as-is: it drops cumulative checking for fragmented uncompressed messages, and it also starts rejecting
compressed frames based on compressed wire size rather than actual inflated payload size.

I think the current direction is safer: keep
uncompressed header checks in the parser, and enforce compressed limits against cumulative inflated bytes.

KhafraDev:
It should do both and all the permessage-deflate-limit tests pass.

> compressed frames based on compressed wire size rather than actual inflated payload size.

This is what I am talking about in the review: it should do both. Someone can send a compressed frame of unlimited size and undici will not error until we attempt to inflate.

lib/web/websocket/receiver.js
```diff
@@ -66,6 +71,20 @@ class ByteParser extends Writable {
...
+      this.#fragmentsBytes + this.#info.payloadLength > this.#maxPayloadSize
```
KhafraDev: This is completely wrong. #fragmentBytes is the actual amount buffered. payLoadlength is a field telling us the final frame size, so we're adding the total frame size to the currently buffered frame size...
image

lib/web/websocket/receiver.js
```diff
@@ -169,6 +188,10 @@ class ByteParser extends Writable {
         this.#info.masked = masked
         this.#info.fin = fin
         this.#info.fragmented = fragmented
+
+        if (this.#state === parserStates.READ_DATA && !this.#validatePayloadLength()) {
```
KhafraDev: there's only one branch where state can be set to READ_DATA from here, it should be moved

lib/web/websocket/receiver.js
```diff
@@ -224,29 +255,34 @@ class ByteParser extends Writable {
...
-              }
-
-              websocketMessageReceived(this.#handler, this.#info.binaryType, this.consumeFragments())
-
-              this.#loop = true
-              this.#state = parserStates.INFO
-              this.run(callback)
-            })
+              },
+              this.#fragmentsBytes
```
KhafraDev, line 288: fragmentBytes should be the same as #inflate[kLength]

whenever we receive a full fragment, we call decompress
decompress calls write() with the chunk (in this case, a complete fragment or message)
on data, we push to the internal buffer and increment [kLength] (kLength therefore being the total decompressed frame size)
once the frame is done, we concat the internal buffers and use [kLength] for performance reasons, then call the callback
the callback calls writeFragment() on the parser which is the *only place* where #fragmentBytes is incremented

so in my patch above, writeFragment() increments #fragmentBytes and then we validate the length and exit - this is the same as the cumulative check

lib/web/websocket/receiver.js
mcollina: > we're adding the total frame size to the currently buffered frame size

No. The idea is that this check interprets "already buffered fragments + current frame".

lib/web/websocket/receiver.js
mcollina: If I remove `fragmentBytes` I get failing tests.

mcollina:
@KhafraDev can you please send a PR against this with the fix you want to see? Your commit also had other changes.

KhafraDev:
that commit is everything I want to see

lib/web/websocket/receiver.js
KhafraDev: In my commit I solved this by handling it in the permessage deflate callback, circumventing the need to pass internal parser state to the permessage deflate handler

mcollina:
Done

PR #4966, opened 2026-04-03
fix(websocket): fallback to HTTP/1.1 when H2 CONNECT is unavailable

mcollina:
## Summary

Fix websocket handshakes when TLS negotiates HTTP/2 but the server does not advertise RFC 8441 extended CONNECT support.

Instead of failing immediately, retry the opening handshake on a fresh HTTP/1.1-only connection. This preserves websocket support for `allowHTTP1` HTTP/2 servers and other deployments that only support websocket upgrade over HTTP/1.1.

## Changes

- add a regression test for websocket fallback to HTTP/1.1 upgrade
- retry websocket handshakes over HTTP/1.1 when H2 extended CONNECT is unavailable
- plumb the HTTP/1.1-only retry through `Agent` and proxy TLS connectors

## Testing

- `npx eslint lib/dispatcher/client.js lib/dispatcher/proxy-agent.js lib/dispatcher/agent.js lib/web/fetch/index.js test/websocket/opening-handshake.js`
- `node --test test/websocket/opening-handshake.js`

### DIFF
--- a/test/websocket/opening-handshake.js
+++ b/test/websocket/opening-handshake.js
@@ -246,6 +246,52 @@ test('WebSocket falls back to HTTP/1.1 upgrade when H2 extended CONNECT is unava
   await planner.completed
 })
 
+test('WebSocket falls back to HTTP/1.1 upgrade when H2 CONNECT support is omitted', async (t) => {
+  const planner = tspl(t, { plan: 4 })
+  const server = createSecureServer({ cert, key, allowHTTP1: true })
+  const wsServer = new WebSocketServer({ server })
+
+  server.on('stream', () => planner.fail('should not attempt websocket over HTTP/2'))
+  server.on('upgrade', (req) => {
+    planner.equal(req.httpVersion, '1.1')
+  })
+
+  wsServer.on('connection', (ws, req) => {
+    planner.equal(req.httpVersion, '1.1')
+    ws.send('hello')
+  })
+
+  server.listen(0)
+  await once(server, 'listening')
+
+  const dispatcher = new EnvHttpProxyAgent({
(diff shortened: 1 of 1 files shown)


test/websocket/opening-handshake.js
```diff
@@ -200,6 +200,52 @@ test('WebSocket on H2 with a server that does not support extended CONNECT proto
   await planner.completed
 })
 
+test('WebSocket falls back to HTTP/1.1 upgrade when H2 extended CONNECT is unavailable', async (t) => {
```
mcollina, line 203: @pimterry PTAL

cristianmadularu:
This did not fix the libp2p usage. It (libp2p) is still failing to connect to other peers over WebSocket.  Reverting to 7.x fixes it.
 - Unless this would now be a libp2p issue?

mcollina:
Open a fresh issue with a clean (no libp2p depedency) reproduction. Point your ai assistant at the problem is ok.

cristianmadularu:
> Open a fresh issue with a clean (no libp2p depedency) reproduction. Point your ai assistant at the problem is ok.

Thanks @mcollina 
I have opened a separate issue with the libp2p team.
https://github.com/libp2p/js-libp2p/issues/3438

Thanks

mcollina:
It can 100% be a problem here.

cristianmadularu:
The issue seems to have been introduced as a side effect of the current fix (8.0.2). Please see a reprodution project here (thank you Claude!):

https://github.com/cristianmadularu/Undici-libp2p-websocket

Please see the readme. I can reproduce it on my system using the project above.

PR #4967, opened 2026-04-03
ci: reenable shared builtin CI tests

mcollina:
## Summary

Re-enable the shared builtin CI workflow for supported/current Node.js majors.

## Changes

- run the shared builtin CI job for Node.js 24 and 25 again
- include Node.js 26 so the reusable workflow will start testing it automatically once a release is available
- keep Node.js 20 and 22 excluded, with a note that they still hit upstream Node.js issues with `--shared-builtin-undici/undici-path`

## Notes

The reusable workflow already skips unreleased versions cleanly, so including Node.js 26 here is safe.

### DIFF
--- a/lib/dispatcher/env-http-proxy-agent.js
+++ b/lib/dispatcher/env-http-proxy-agent.js
@@ -10,22 +10,6 @@ const DEFAULT_PORTS = {
   'https:': 443
 }
 
-/**
- * Normalizes a proxy URL by prepending a scheme if one is missing.
- * This matches the behavior of curl and Go's httpproxy package, which
- * assume http:// for scheme-less proxy values.
- *
- * @param {string} proxyUrl - The proxy URL to normalize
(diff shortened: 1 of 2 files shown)


mcollina:
@travisbreaks I reverted the scheme-less proxy env var normalization here because it causes a regression in Node.js when Undici is used through `--shared-builtin-undici/undici-path`.

Specifically, Node core `client-proxy` tests expecting invalid proxy URLs to fail at startup begin succeeding unexpectedly because `EnvHttpProxyAgent` normalizes `HTTP_PROXY/HTTPS_PROXY=not-a-valid-url` into `http://not-a-valid-url` / `https://not-a-valid-url` instead of throwing.

That breaks at least:
- `test/client-proxy/test-http-proxy-request-invalid-url.mjs`
- `test/client-proxy/test-https-proxy-request-invalid-url.mjs`

So from the Node integration point of view this is a regression.

PR #4974, opened 2026-04-04
chore: use node 22 Set methods for websocket close state

trivikr:
## This relates to...

N/A

## Rationale

Node.js 22 supports the new `Set` relation methods, which make the WebSocket closing-handshake checks more explicit. This change centralizes the closing-handshake state set in the WebSocket constants module and reuses it across the WebSocket implementation instead of redefining it in multiple files.

## Changes

- Add `closingHandshakeStates` to the WebSocket constants module
- Document `closingHandshakeStates` with JSDoc
- Replace duplicated local closing-handshake sets with the shared constant
- Use `Set.prototype.isDisjointFrom()` and `Set.prototype.isSupersetOf()` for closing-handshake checks

### Features

N/A

### Bug Fixes

N/A

N/A

## Status

- [x] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/web/websocket/connection.js
+++ b/lib/web/websocket/connection.js
@@ -1,6 +1,6 @@
 'use strict'
 
-const { uid, states, sentCloseFrameState, emptyBuffer, opcodes } = require('./constants')
+const { uid, states, sentCloseFrameState, closingHandshakeStates, emptyBuffer, opcodes } = require('./constants')
 const { parseExtensions, isClosed, isClosing, isEstablished, isConnecting, validateCloseCodeAndReason } = require('./util')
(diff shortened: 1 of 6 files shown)


mcollina (approved):
lgtm

KhafraDev:
> which make the WebSocket closing-handshake checks more explicit

I don't agree and the code doesn't agree either. You wouldn't need to partially explain set theory in the comment if it was explicit.

Having to duplicate the states in a new set is also just not good code.

trivikr:
> Having to duplicate the states in a new set is also just not good code.

Closing as this change is not good code, according to one of the maintainers.

PR #4984, opened 2026-04-06
fix: preserve Request signal abort propagation after GC

mcollina:
## Summary

Fix `Request` signal following so abort propagation still works after GC pressure.

This replaces the custom `WeakRef`/`FinalizationRegistry`-based follow logic in `lib/web/fetch/request.js` with `AbortSignal.any([signal])`.

## Changes

- use `AbortSignal.any([signal])` to create dependent request signals
- add a regression test for a `Request` signal surviving GC pressure
- add a regression test for cloned `Request` signals surviving GC pressure
- add a regression test to ensure reusing a controller across transient requests does not emit listener leak warnings

## Motivation

This reproduces the behavior discussed in:
- nodejs/node#55428

The old implementation could disconnect `request.signal` from its parent signal once the original `Request` object became unreachable. As a result, aborting the parent signal no longer aborted the dependent request signal.

## Testing

- `NODE_OPTIONS=--expose-gc node --test test/fetch/request.js`
- `NODE_OPTIONS=--expose-gc node --test test/fetch/long-lived-abort-controller.js test/node-fetch/request.js test/fetch/abort.js`

### DIFF
--- a/lib/web/fetch/request.js
+++ b/lib/web/fetch/request.js
@@ -26,60 +26,13 @@ const { webidl } = require('../webidl')
 const { URLSerializer } = require('./data-url')
 const { kConstruct } = require('../../core/symbols')
 const assert = require('node:assert')
-const { getMaxListeners, setMaxListeners, defaultMaxListeners } = require('node:events')
 
-const kAbortController = Symbol('abortController')
-
-const requestFinalizer = new FinalizationRegistry(({ signal, abort }) => {
-  signal.removeEventListener('abort', abort)
-})
-
-const dependentControllerMap = new WeakMap()
-
-let abortSignalHasEventHandlerLeakWarning
-
-try {
-  abortSignalHasEventHandlerLeakWarning = getMaxListeners(new AbortController().signal) > 0
-} catch {
-  abortSignalHasEventHandlerLeakWarning = false
-}
-
-function buildAbort (acRef) {
-  return abort
-
-  function abort () {
-    const ac = acRef.deref()
-    if (ac !== undefined) {
-      // Currently, there is a problem with FinalizationRegistry.
-      // https://github.com/nodejs/node/issues/49344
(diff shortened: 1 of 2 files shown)


mcollina:
@KhafraDev PTAL

KhafraDev:
I very, very much prefer this approach to our current one.

mcollina:
I need to investigate the failing test

lib/web/fetch/request.js
```diff
@@ -412,38 +365,7 @@ class Request {
 
     // 28. Set this’s signal to a new AbortSignal object with this’s relevant
     // Realm.
-    // TODO: could this be simplified with AbortSignal.any
-    // (https://dom.spec.whatwg.org/#dom-abortsignal-any)
-    const ac = new AbortController()
-    this.#signal = ac.signal
-
-    // 29. If signal is not null, then make this’s signal follow signal.
```
KhafraDev, line 427: This step shouldn't be removed. We're creating an AbortSignal regardless of if one was passed in RequestInit. Is there a reason for this?

mcollina: It has been moved inside `makeRequestSignal`. Should I keep some of the comments?

KhafraDev: yes please

trivikr:
The tests are successful if AbortSignal is removed from fetch in `test/tls-cert-leak.js`

```diff
-      const res = await fetch(`${serverUrl}/request-${i}`, {
-        signal: AbortSignal.timeout(2000) // Short timeout to prevent hanging
-      })
+      const res = await fetch(`${serverUrl}/request-${i}`)
```

Is it possible that `AbortSignal.timeout(2000)` creates short-lived retained objects that outlive the failed fetch calls?

mcollina:
> Is it possible that AbortSignal.timeout(2000) creates short-lived retained objects that outlive the failed fetch calls?

Totally possible.

PR #5001, opened 2026-04-09
fix: replace finished() with writable lifecycle tracking

trivikr:
## This relates to...

- Hot-path listener overhead in `lib/api/api-stream.js`
- The existing TODO about avoiding `stream.finished()` in `StreamHandler`

## Rationale

`StreamHandler` used `stream.finished(res, { readable: false })` to observe completion of the user-provided writable. That helper is broader than this path needs and registers extra listeners, which adds avoidable overhead on a hot path.

This change narrows lifecycle tracking to the writable side only while preserving the behavior `stream()` depends on for completion, writable errors, and premature close handling.

## Changes

- Replaced `stream.finished()` usage in `StreamHandler` with a small writable lifecycle helper that listens only for `finish`, `error`, and `close`
- Preserved premature close behavior by reporting `ERR_STREAM_PREMATURE_CLOSE` when the writable closes before finishing
- Preserved existing error propagation by aborting the request with the writable error
- Avoided re-destroying the writable with the same error after an `'error'` event, which keeps existing error handling semantics intact
- Added regression coverage for premature writable close without an explicit error
- Added regression coverage for a factory returning an already-finished writable

### Features

N/A

### Bug Fixes

- Reduced listener registration overhead in `stream()` response handling
- Kept `stream()` callback and abort behavior consistent for writable finish, error, and premature close edge cases

- None

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

Assisted-by: openai:gpt-5.4

### DIFF
--- a/lib/api/api-stream.js
+++ b/lib/api/api-stream.js
@@ -1,13 +1,60 @@
 'use strict'
 
 const assert = require('node:assert')
-const { finished } = require('node:stream')
 const { AsyncResource } = require('node:async_hooks')
 const { InvalidArgumentError, InvalidReturnValueError } = require('../core/errors')
 const util = require('../core/util')
 const { addSignal, removeSignal } = require('./abort-signal')
 
 function noop () {}
+
+function getWritableError (stream) {
+  return stream.errored ?? stream.writableErrored ?? stream._writableState?.errored
+}
+
+function createPrematureCloseError () {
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

Issue #112, opened 2020-05-13
Request Body as Factory Method


ronag:
Allow passing a factory method as body which would return a `Readable`. This would allow e.g. idempotent `PUT` requests which e.g. read from a file to be part of the indirect inflight failure retries.

Alternatively a async generator function which is retryable by definition.

mcollina:
It would be nice. However unlikely to work in most cases as the body can not be fetched more than once.

ronag:
> It would be nice. However unlikely to work in most cases as the body can not be fetched more than once.

If it's being read from a file or from an upstream request... it could be

szmarczak:
Can we close this? https://github.com/nodejs/undici/pull/848

ronag:
Not quite.

szmarczak:
What is left to do here?

PR #5008, opened 2026-04-09
feat: support body as factory function for retryable requests

mcollina:
This implements support for passing a factory function as the body option, which creates a new body instance for each dispatch. This enables retryable requests with bodies that can be recreated, such as file streams or async generators.

Example with the retry interceptor:
```js
const { Client, interceptors } = require('undici')
const fs = require('node:fs')

const client = new Client('http://example.com')
  .compose(interceptors.retry({
    methods: ['POST']
  }))

// Retryable POST with file body - a new stream is created for each retry
const result = await client.request({
  method: 'POST',
  path: '/upload',
  headers: { 'content-type': 'application/octet-stream' },
  body: () => fs.createReadStream('/path/to/file.txt')
})
```

The factory function is called once for the initial dispatch and once for each retry, ensuring a fresh body is used each time.

Refs: https://github.com/nodejs/undici/issues/112

### DIFF
--- a/docs/docs/api/Dispatcher.md
+++ b/docs/docs/api/Dispatcher.md
@@ -193,7 +193,7 @@ Returns: `Boolean` - `false` if dispatcher is busy and further dispatch calls wo
 * **path** `string`
 * **method** `string`
 * **reset** `boolean` (optional) - Default: `false` - If `false`, the request will attempt to create a long-living connection by sending the `connection: keep-alive` header,otherwise will attempt to close it immediately after response by sending `connection: close` within the request and closing the socket afterwards.
-* **body** `string | Buffer | Uint8Array | stream.Readable | Iterable | AsyncIterable | null` (optional) - Default: `null`
+* **body** `string | Buffer | Uint8Array | stream.Readable | Iterable | AsyncIterable | (() => string | Buffer | Uint8Array | stream.Readable | Iterable | AsyncIterable) | null` (optional) - Default: `null`. A factory function can be used to create a new body for each retry when used with the [`retry`](./Retry.md) interceptor.
 * **headers** `UndiciHeaders` (optional) - Default: `null`.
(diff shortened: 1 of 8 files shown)


ronag (commented):
LGTM, but there is a different way by using an interceptor *before* the retry interceptor:

```js
import { isStream } from '../utils.js'

class FactoryStream extends Readable {
  #factory
  #ac
  #body

  constructor(factory) {
    super()
    this.#factory = factory
  }

  _construct(callback) {
    this.#ac = new AbortController()
    Promise.resolve(this.#factory({ signal: this.#ac.signal })).then(
      (body) => {
        this.#ac = null
        try {
          if (typeof body === 'string' || body instanceof Buffer) {
            this.push(body)
            this.push(null)
          } else if (isStream(body)) {
            this.#body = body
          } else {
            this.#body = Readable.from(body)
          }

          if (this.#body != null) {
            this.#body
              .on('data', (data) => {
                if (!this.push(data)) {
                  this.#body.pause()
                }
              })
              .on('end', () => {
                this.push(null)
              })
              .on('error', (err) => {
                this.destroy(err)
              })
          }

          callback(null)
        } catch (err) {
          callback(err)
        }
      },
      (err) => callback(err),
    )
  }

  _read() {
    this.#body?.resume()
  }

  _destroy(err, callback) {
    if (this.#ac) {
      this.#ac.abort(err)
      this.#ac = null
    }

    if (this.#body) {
      this.#body.destroy(err)
      this.#body = null
    }

    callback(err)
  }
}

export default () => (dispatch) => (opts, handler) =>
  typeof opts.body !== 'function'
    ? dispatch(opts, handler)
    : dispatch({ ...opts, body: new FactoryStream(opts.body) }, handler)

```

metcoder95 (commented):
LGTM, but wondering the use-cases; this seems pretty specific to `retry` interceptor; if that's the case, I believe documenting @ronag comment can be a good alternative.

PR #5010, opened 2026-04-09
feat(streams): limit usage in body mixin #2164

### LINKED ISSUE #2164: fetch: limit web streams usage in body mixin methods

rozzilla:
## This relates to...
Fixes #2164 

## Rationale
When a `Response` or `Request` body is created from a `string` or `Uint8Array`, the body mixin methods (text(), json(), arrayBuffer(), blob(), bytes()) currently pipe the source through a `ReadableStream` even though the bytes are already available in `body.source`. This adds unnecessary overhead from stream construction, pump microtasks, and chunk collection.

## Changes
1. In `consumeBody()`, before calling `fullyReadBody()`, check if `body.source` is a `string` or `Uint8Array`. If so, bypass the stream round-trip and pass the source bytes directly to `successSteps`
2. Add tests covering `string` sources (all 5 body mixin methods, Request, multibyte, empty), `Uint8Array` sources (text, arrayBuffer, bytes, blob), body unusability after consumption (bodyUsed, double-read rejection), and non-shortcuttable sources (ReadableStream, Blob, null) to ensure the normal path is unaffected.

### Features
N/A

### Bug Fixes
N/A

N/A

## Status

- [x] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/benchmarks/fetch/body-source-shortcut.mjs
+++ b/benchmarks/fetch/body-source-shortcut.mjs
@@ -1,0 +1,71 @@
+import { group, bench, run } from 'mitata'
+import { Response } from '../../lib/web/fetch/response.js'
+
+// Benchmark body mixin methods with known-source bodies (string & Uint8Array)
+// to measure the impact of the body source shortcut optimization (#2164)
+
+const shortString = 'hello world'
+const mediumString = 'x'.repeat(1024)
+const longString = 'x'.repeat(65536)
+const shortBytes = new Uint8Array(32).fill(65)
+const mediumBytes = new Uint8Array(1024).fill(65)
+const longBytes = new Uint8Array(65536).fill(65)
+
+group('Response#text() with string body', () => {
+  bench('short string (11B)', async () => {
+    await new Response(shortString).text()
+  })
+  bench('medium string (1KB)', async () => {
+    await new Response(mediumString).text()
+  })
+  bench('long string (64KB)', async () => {
+    await new Response(longString).text()
+  })
+})
+
+group('Response#json() with string body', () => {
+  bench('small JSON', async () => {
+    await new Response('{"a":1}').json()
+  })
+})
+
+group('Response#arrayBuffer() with string body', () => {
+  bench('short string (11B)', async () => {
+    await new Response(shortString).arrayBuffer()
+  })
+  bench('medium string (1KB)', async () => {
+    await new Response(mediumString).arrayBuffer()
+  })
+})
+
+group('Response#text() with Uint8Array body', () => {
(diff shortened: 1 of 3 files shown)


KhafraDev (changes requested):
The issue with optimizations like these (if it's even more performant, I'm not sure that it is) is that it encourages people to only use string or Uint8Array bodies to squeeze more performance out of node. I don't want to recreate CrankshaftScript.

mcollina:
How much is more performant? I think that in most cases, those optimizations are good for Node.js.

KhafraDev:
It wouldn't matter. The implementation is incorrect as it doesn't perform the same validation readAllBytes does (ensuring the result of reader.read() is a Uint8Array). After that I'm not sure, 1 less promise, for the extra maintenance burden and guaranteed bugs there will be?

PR #5013, opened 2026-04-10
fix: prevent cache dedup key collision via unescaped delimiters

### LINKED ISSUE #5012: [Security] Cache deduplication key collision via unescaped delimiters
## Summary
`makeDeduplicationKey` in `lib/util/cache.js:376` constructs keys using `:` and `=` delimiters without escaping header names or values. Different header sets can produce identical keys, causing response confusion when the deduplicate interceptor is enabled.

## Reproduction
Headers `{a:"x:b=y"}` produces the same dedup key as `{a:"x", b:"y"}`:
- Key 1: `GET:https://example.com:a=x:b=y`
- Key 2: `GET:https://example.com:a=x:b=y`

Both are identical despite different header sets.

## Impact
When cache deduplication is used, concurrent requests with colliding keys share a dedup slot, potentially receiving the wrong response. Impact depends on what headers differ (auth headers, content negotiation, etc).

## Suggested Fix
Escape `:` and `=` in header names/values before key construction, or use a collision-resistant format (e.g., `JSON.stringify` sorted entries, or length-prefixed encoding).

## Affected Component
Deduplicate interceptor (`lib/util/cache.js`)

---
Found during automated security audit.

eddieran:
## Summary

Fixes #5012

`makeDeduplicationKey` in `lib/util/cache.js` constructed deduplication keys using `:` and `=` delimiters without escaping header names or values. This allowed different header sets to produce identical keys, causing potential response confusion when the deduplicate interceptor is enabled.

**Example collision:**
- Headers `{a: "x:b=y"}` → key: `GET:https://example.com:a=x:b=y`
- Headers `{a: "x", b: "y"}` → key: `GET:https://example.com:a=x:b=y`

Both are identical despite representing different request headers.

## Changes

- Replaced string concatenation with `JSON.stringify([origin, method, path, headers])` which is inherently collision-resistant since JSON encoding properly handles all special characters
- Added 3 unit tests for `makeDeduplicationKey`:
  - Verifies that delimiter characters in header values no longer cause key collisions
  - Verifies that identical headers (regardless of insertion order) produce the same key
  - Verifies that `excludeHeaders` parameter is respected

## Test plan

- [x] All 33 existing + new dedup interceptor tests pass (`node --test test/interceptors/deduplicate.js`)
- [x] New test specifically reproduces the collision scenario from #5012 and verifies the fix

### DIFF
--- a/lib/util/cache.js
+++ b/lib/util/cache.js
@@ -374,9 +374,11 @@ function assertCacheMethods (methods, name = 'CacheMethods') {
  * @returns {string}
  */
 function makeDeduplicationKey (cacheKey, excludeHeaders) {
-  // Create a deterministic string key from the cache key
-  // Include origin, method, path, and sorted headers
-  let key = `${cacheKey.origin}:${cacheKey.method}:${cacheKey.path}`
+  // Use JSON.stringify to produce a collision-resistant key.
+  // Previous format used `:` and `=` delimiters without escaping, which
+  // allowed different header sets to produce identical keys (e.g.
+  // {a:"x:b=y"} vs {a:"x", b:"y"}). See: https://github.com/nodejs/undici/issues/5012
+  const headers = {}
 
   if (cacheKey.headers) {
     const sortedHeaders = Object.keys(cacheKey.headers).sort()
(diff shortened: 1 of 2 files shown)


metcoder95 (approved):

mcollina (approved):
lgtm

PR #5026, opened 2026-04-13
fix: fix the logic for the UNDICI_NO_WASM_SIMD environment variable

ShenHongFei:
## Rationale

The original function of the UNDICI_NO_WASM_SIMD environment variable was reversed. It has been fixed now.

## Changes

Fixed the erroneous logic of UNDICI_NO_WASM_SIMD

### Features

N/A

### Bug Fixes

N/A

N/A

## Status

- [S] Tested
- [S] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/dispatcher/client-h1.js
+++ b/lib/dispatcher/client-h1.js
@@ -69,9 +69,9 @@ function lazyllhttp () {
   let useWasmSIMD = process.arch !== 'ppc64'
   // The Env Variable UNDICI_NO_WASM_SIMD allows explicitly overriding the default behavior
   if (process.env.UNDICI_NO_WASM_SIMD === '1') {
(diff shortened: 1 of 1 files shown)


mcollina:
cc @Uzlopak @richardlau this seems correct. Wdyt?

richardlau (approved):

richardlau:
FWIW we can probably rework the assumption to disable by default on Power now as the V8 fix has been backported to all in support LTS Node.js release lines, but I guess we'd need an allow list of Node.js versions as Undici might be being run on an older, e.g. Node.js 24 release before the fix went in.

metcoder95 (approved):

PR #5030, opened 2026-04-14
fix(http2): send body for non-expectsPayload methods with content

### LINKED ISSUE #5027: HTTP/2: `buffer body must have content length` assertion when sending DELETE with a body
## Bug Description

We upgraded to undici v8 which changed `allowH2` to default to `true`. Some third-party APIs we integrate with require a body on DELETE requests (batch deletions). This started crashing with an uncaught `AssertionError`. Worked fine before v8 over HTTP/1.1.

`writeH2` in `client-h2.js` nulls `contentLength` for methods that aren't POST/PUT/PATCH (`!expectsPayload`), but the body (converted to Buffer in the Request constructor) is still passed to `writeBuffer`, which asserts `contentLength === body.byteLength` (`null === N`).

Happy to open a PR with a fix.

## Reproducible By

```js

await request('https://www.google.com', {
  method: 'DELETE',
  body: JSON.stringify({ ids: [1] }),
});
```

## Expected Behavior

The request should be sent without an assertion error, as it does over HTTP/1.1.

## Logs & Screenshots

```
AssertionError [ERR_ASSERTION]: buffer body must have content length
    at writeBuffer (undici/lib/dispatcher/client-h2.js:857:7)
    at writeBodyH2 (undici/lib/dispatcher/client-h2.js:792:7)
    at writeH2 (undici/lib/dispatcher/client-h2.js:675:5)
```

## Environment

- undici 8.1.0
- Node.js v24.14.0
- Linux

mcollina:
Fixes #5027

## Summary

HTTP/2 crashes with `AssertionError: buffer body must have content length` when sending DELETE (or other non-expectsPayload methods) with a body.

## Root Cause

In `lib/dispatcher/client-h2.js`, the `writeH2` function unconditionally set `contentLength = null` for all methods that don't expect a payload (DELETE, GET, HEAD, etc.). However, when a body was provided, it still got passed to `writeBodyH2`, which then called `writeBuffer` with `body` and `null` contentLength. The `writeBuffer` function asserts `contentLength === body.byteLength`, which fails when `contentLength` is `null`.

## Fix

Changed line 625 from:
```js
if (!expectsPayload) {
```
to:
```js
if (contentLength === 0 && !expectsPayload) {
```

This matches the H1 behavior and only omits the Content-Length header when the body is empty for methods that don't expect a payload. When a body is present, the Content-Length header is included and the body is sent correctly.

## Testing

- All existing HTTP/2 tests pass (42 tests)
- All client and content-length tests pass (342 tests)
- Verified the fix resolves the reported issue with DELETE + body over HTTP/2

### DIFF
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -622,7 +622,7 @@ function writeH2 (client, request) {
     contentLength = request.contentLength
   }
 
-  if (!expectsPayload) {
+  if (contentLength === 0 && !expectsPayload) {
     // https://tools.ietf.org/html/rfc7230#section-3.3.2
     // A user agent SHOULD NOT send a Content-Length header field when
     // the request message does not contain a payload body and the method
--- a/test/http2-body-delete.js
+++ b/test/http2-body-delete.js
@@ -1,0 +1,103 @@
+'use strict'
+
+const { test, after } = require('node:test')
+const { createSecureServer } = require('node:http2')
+const { once } = require('node:events')
+const assert = require('node:assert')
+
+const pem = require('@metcoder95/https-pem')
+
+const { Client } = require('..')
+
+test('Should handle h2 DELETE request with body', async t => {
+  const server = createSecureServer(await pem.generate({ opts: { keySize: 2048 } }))
+  const expectedBody = '{"ids":[1,2,3]}'
+  const requestBody = []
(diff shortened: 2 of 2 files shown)


metcoder95 (commented):
regression test?

metcoder95 (approved):

PR #4223, opened 2025-05-15
fix: agent memory leak

styfle:
## This relates to...

There was a TODO introduced in a previous PR:

- https://github.com/nodejs/undici/pull/2536

## Rationale

There is a known memory leak

## Changes

Track the number of origins per dispatcher, then clean up when that reaches 0 in order to avoid the memory leak. 

You may not notice this memory leak if you don't fetch from many different origins.

### Features

N/A

### Bug Fixes

- Related to https://github.com/nodejs/undici/issues/3895

### Breaking Changes and Deprecations

N/A

## Status

- [ ] I have read and agreed to the [Developer's Certificate of Origin][cert]
- [ ] Tested
- [ ] Benchmarked (**optional**)
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

[cert]: https://github.com/nodejs/undici/blob/main/CONTRIBUTING.md#developers-certificate-of-origin

PR #5034, opened 2026-04-16
fix: avoid premature cleanup of dispatcher in Agent

### LINKED ISSUE #5022: Keep-Alive max parameter prevents connection reuse
## Bug Description

When the server has a request limit per connection via the Keep-Alive max parameter, Undici reuses the initial connection until the limit is reached.
After the limit is reached, the server sends a "Connection: close" header, and Undici closes the connection.
In subsequent requests, despite the "Connection: Keep-Alive" header, Undici closes the connection.

## Reproducible By

```javascript
import { request, Agent } from 'undici';

const dispatcher = new Agent({ connections: 1 });

async function testRequest(id) {
    const url = 'https://tgftp.nws.noaa.gov/testfile';
    const { body, headers } = await request(url, { dispatcher });
    await body.dump();
    console.log({
        id,
        connection: headers['connection'],
        keepAlive: headers['keep-alive'],
    });
}

async function run() {
    for (let id = 1; id <= 110; id++) {
        await testRequest(id);
    }
}

run();
```

## Expected Behavior

Undici should reuse connections when possible.

## Logs

```javascript
{ id: 1, connection: 'Keep-Alive', keepAlive: 'timeout=300, max=100' } // new connection
{ id: 2, connection: 'Keep-Alive', keepAlive: 'timeout=300, max=99' } // reused connection
{ id: 3, connection: 'Keep-Alive', keepAlive: 'timeout=300, max=98' } // reused connection
{ id: 4, connection: 'Keep-Alive', keepAlive: 'timeout=300, max=97' } // reused connection
...
{ id: 99, connection: 'Keep-Alive', keepAlive: 'timeout=300, max=2' } // reused connection
{ id: 100, connection: 'Keep-Alive', keepAlive: 'timeout=300, max=1' } // reused connection
{ id: 101, connection: 'close', keepAlive: undefined } // reused connection
{ id: 102, connection: 'Keep-Alive', keepAlive: 'timeout=300, max=100' } // new connection
{ id: 103, connection: 'Keep-Alive', keepAlive: 'timeout=300, max=100' } // new connection
{ id: 104, connection: 'Keep-Alive', keepAlive: 'timeout=300, max=100' } // new connection
{ id: 105, connection: 'Keep-Alive', keepAlive: 'timeout=300, max=100' } // new connection
...
```

## Environment

- **Node.js**: v24.14.0 (undici 8.0.1)
- **Platform**: Ubuntu 24.04.4 LTS

bienzaaron:
## This relates to...

#5022 

## Changes

The fixes in #4223 and #4425 introduce a bug where keep-alive connections are sometimes not correctly reused after the first disconnection for a dispatcher.

The issue is a race between disconnect and new dispatches. After a response with `Connection: close` or after the connection reaches its request limit, the socket disconnects. If a new request has already been dispatched to the same dispatcher but has not yet been processed into a new connection, the `disconnect` handler can incorrectly treat the dispatcher as unused and call `close()` on it.

As far as I can tell, this is still a graceful failure (new request doesn't fail), but it does result in a new connection being used. If new requests are continually dispatched like this, then connections are never reused after the first one disconnects. Delaying the next dispatched request to the next event loop iteration (e.g. via `setTimeout(cb, 0)`) or longer results in the connection being reused correctly.

### Features

N/A

### Bug Fixes

#5022 

N/A

## Status

- [x] Tested
- [S] Documented
- [x] Review ready
- [ ] In review
- [x] Merge ready

### DIFF
--- a/lib/dispatcher/agent.js
+++ b/lib/dispatcher/agent.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const { InvalidArgumentError, MaxOriginsReachedError } = require('../core/errors')
-const { kBusy, kClients, kRunning, kClose, kDestroy, kDispatch, kUrl } = require('../core/symbols')
+const { kBusy, kClients, kConnected, kRunning, kClose, kDestroy, kDispatch, kUrl } = require('../core/symbols')
 const DispatcherBase = require('./dispatcher-base')
 const Pool = require('./pool')
 const Client = require('./client')
@@ -65,7 +65,7 @@ class Agent extends DispatcherBase {
 
   get [kRunning] () {
     let ret = 0
-    for (const { dispatcher } of this[kClients].values()) {
+    for (const dispatcher of this[kClients].values()) {
       ret += dispatcher[kRunning]
     }
     return ret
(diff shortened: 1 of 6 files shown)


lib/dispatcher/agent.js
```diff
@@ -86,54 +86,52 @@ class Agent extends DispatcherBase {
...
-            }
-          }
+      const closeClientIfUnused = () => {
+        if (this[kClients].get(key) !== dispatcher) {
+          return
+        }
+
+        if (dispatcher[kConnected] > 0 || dispatcher[kBusy]) {
+          return
+        }
```
bienzaaron, line 102: The key difference is here - we check active connections as well as `kBusy`:
https://github.com/nodejs/undici/blob/bc0a19c891abd5fb4d6c28d08fd681402445c128/lib/dispatcher/client.js#L326-L332

If we have either active connections or the dispatcher is busy, we don't close it.

This lets us clean up the connection-tracking bookkeeping and just use information already in the dispatcher.

lib/dispatcher/agent.js
```diff
@@ -86,54 +86,52 @@ class Agent extends DispatcherBase {
...
+        this[kClients].delete(key)
+        if (!dispatcher.destroyed) {
+          dispatcher.close()
+        }
 
-          if (!hasOrigin) {
-            this[kOrigins].delete(origin)
+        let hasOrigin = false
+        for (const client of this[kClients].values()) {
+          if (client[kUrl].origin === dispatcher[kUrl].origin) {
```
bienzaaron, line 111: slightly updated logic here, again to reduce bookkeeping -- access origin from `client[kUrl].origin` instead of storing it

lib/dispatcher/client-h1.js
```diff
@@ -1106,7 +1106,7 @@ function writeH1 (client, request) {
     socket[kReset] = reset
   }
 
-  if (client[kMaxRequests] && socket[kCounter]++ >= client[kMaxRequests]) {
+  if (client[kMaxRequests] && ++socket[kCounter] >= client[kMaxRequests]) {
```
bienzaaron, line 1109: slightly unrelated to this PR, but I cherry-picked https://github.com/nodejs/undici/pull/5033/commits/f68bd5300f0f09d0bfe7e776f5ff2df1cda18f95 from #5033 so that my tests pass (otherwise, the number of requests per connection would be off by one).

test/agent-connection-management.js
```diff
@@ -62,4 +62,96 @@ describe('Agent should close inactive clients', () => {
 
     await p
   })
+
```
mcollina, line 62: the file is named after a specific issue. Given that you are adding more things, can you pick a better name?

mcollina (approved):
lgtm

test/agent-connection-management.js
bienzaaron: good call, 6e2a7da

test/agent-connection-management.js
```diff
@@ -0,0 +1,160 @@
...
+    let p
+    const agent = new Agent({
+      factory: (origin, opts) => {
+        const pool = new Pool(origin, opts)
+        let _resolve, _reject
+        p = new Promise((resolve, reject) => {
+          _resolve = resolve
+          _reject = reject
+        })
```
mcollina: You can use `Promise.withResolvers()`

lib/dispatcher/client-h1.js
mcollina: That change is covered by your tests?

mcollina (changes requested):

lib/dispatcher/client-h1.js
bienzaaron: Yes, it is covered by the assertion here - https://github.com/nodejs/undici/pull/5034/changes#diff-aa5d6e637dbf3a57f70e146170bcc8917253e98f0885ff7d1d1875936a60255eR116-R117


but I fixed some assertions in [`test/close-and-destroy.js`](https://github.com/nodejs/undici/blob/main/test/close-and-destroy.js#L266) as well in 5d75f78 that were masking this issue before. They only tested the connection count overall, not where the boundary between connections are.

test/agent-connection-management.js
bienzaaron: These were existing tests from `test/issue-4244.js`, but cleaned up in d897e6e

mcollina (approved):
lgtm

PR #5035, opened 2026-04-16
fix: use StringDecoder to handle UTF-8 chunk boundaries in setEncoding

398651434:
## Description

Fixes a bug where `response.body.setEncoding('utf8')` corrupts multi-byte UTF-8 characters that span chunk boundaries.

### Root Cause
Each chunk was being individually converted to a string via `buffer.utf8Slice()` (or `toString()`). When a multi-byte UTF-8 character (e.g., a Chinese character = 3 bytes) is split across two HTTP response chunks, the first chunk gets an incomplete byte sequence converted to garbage, and the second chunk's portion becomes a separate corrupted character.

### Fix
Use Node.js's built-in `StringDecoder` (from `node:string_decoder`) which properly buffers incomplete byte sequences between `write()` calls:

1. **`setEncoding(encoding)`**: Initialize a `StringDecoder` when encoding is set
2. **`consumePush`**: When a decoder exists, use `decoder.write(chunk)` instead of storing the raw buffer — this accumulates incomplete UTF-8 bytes internally
3. **`consumeFinish`**: Reset the decoder to allow garbage collection

### Testing
The bug manifests when:
- HTTP response contains multi-byte UTF-8 text (e.g., Chinese characters, emoji)
- `setEncoding('utf8')` is called on the body
- The text spans multiple TCP packets/chunks

After fix, characters are correctly reassembled across chunk boundaries.

---
Closes #5002

### DIFF
--- a/lib/api/readable.js
+++ b/lib/api/readable.js
@@ -1,6 +1,7 @@
 'use strict'
 
 const assert = require('node:assert')
+const { StringDecoder } = require('node:string_decoder')
 const { addAbortListener } = require('node:events')
 const { Readable } = require('node:stream')
 const { RequestAbortedError, NotSupportedError, InvalidArgumentError, AbortError } = require('../core/errors')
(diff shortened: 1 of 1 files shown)


metcoder95 (commented):
can you add a regression for it?

PR #5052, opened 2026-04-17
fix: make MockAgent work with protocol-aware global dispatchers

### LINKED ISSUE #5036: MockAgent as global dispatcher broken since version 8.0.3
## Bug Description

## Reproducible By

test.js:

```js

agent.disableNetConnect();

setGlobalDispatcher(agent);

  .intercept({
    method: 'GET',
    path: '/v1/test',
  })
  .reply(() => {
    return {
      statusCode: 200,
      data: { test: 123 },
    }
  });

console.log(req.status);

console.log(data);
```

```
node test.js
```

## Expected Behavior

## Logs & Screenshots

```
➤ node test.js
404
<anonymous_script>:1
^

    at JSON.parse (<anonymous>)
```

## Environment

### Additional context

/cc @mcollina

mcollina:
## Summary

Fixes the regression where `setGlobalDispatcher(new MockAgent())` no longer intercepts Node.js native `fetch()` requests.

The original regression happened after legacy global-dispatcher consumers started forcing `allowH2: false`. That made `Agent` route those requests through a separate HTTP/1-only path, while `MockAgent` only registered mocks for the default origin path.

This PR does more than mirror a second mock bucket:

- makes `Agent` explicitly protocol-aware internally by separating
  - default dispatcher entries
  - HTTP/1-only dispatcher entries
- adds internal helper symbols for querying and mutating protocol-specific dispatcher entries
- updates `MockAgent` to use those helpers instead of sharing or depending on `Agent`'s raw pool map
- keeps mock interceptor state attached to the logical mock registry while registering protocol-specific dispatchers into `Agent`
- adds regression coverage for native `fetch()` and protocol-specific mock registration

This removes the brittle coupling between `Agent` and `MockAgent` around the HTTP/1-only path and preserves native `fetch()` interception.

Fixes: #5036


metcoder95 (approved):

metcoder95:
just conflicts on files

PR #5053, opened 2026-04-17
Remove unused func and unnecessary shim

tsctx:
I have removed unused and unreachable code paths.
The shim for WebSocket `crypto.randomFillSync` is no longer required since SHA-1 (essential for WebSocket functionality) cannot be used in environments without crypto support.


mcollina (approved):
lgtm

PR #5060, opened 2026-04-19
fix: reject malformed content-length request headers

### LINKED ISSUE #5059: content-length request header parsing accepts malformed values and normalizes them instead of rejecting them
## Bug Description

Request content-length validation is too permissive 
https://github.com/nodejs/undici/blob/a1d6766adef90b87de59f4e0eebc169138f6c935/lib/core/request.js#L487

The header value is parsed with `parseInt(val, 10)` and only checked with `Number.isFinite()`, which silently normalizes malformed values instead of rejecting them.

## Reproducible By

```js
import { request } from "undici";

const values = ["1.1", "10abc", "-1", "asd"];

  console.log("received header:", req.headers["content-length"]);
  res.end("ok");
});

  const origin = `http://localhost:${server.address().port}`;

  for (const value of values) {
    try {
      await request({
        origin,
        path: "/",
        method: "POST",
        headers: { "content-length": value },
        body: "a",
      });
      console.log(value, "=> success");
    } catch (err) {
      console.log(value, "=>", err.name, "|", err.message);
    }
  }

  server.close();
});
```

## Expected Behavior

Malformed `content-length` values should be rejected immediately with `InvalidArgumentError('invalid content-length header')`.

In particular, values such as '1.1', '10abc', and '-1' should not be parsed or normalized into valid internal numeric values, and the request should not be sent.

## Logs & Screenshots

```console
received header: 1
1.1 => success
10abc => RequestContentLengthMismatchError | Request body length does not match content-length header
-1 => RequestContentLengthMismatchError | Request body length does not match content-length header
asd => InvalidArgumentError | invalid content-length header
```

## Environment

macOS 26.4.1
Node v24.14.1
undici v8.1.0

trivikr:
## This relates to...

Fixes: https://github.com/nodejs/undici/issues/5059

## Rationale

`content-length` was being validated too loosely because malformed values could be partially normalized before later request-body length checks.

This change makes the header validation stricter so only digit-only values are accepted during request parsing.

## Changes

- Reject non-string, empty, fractional, signed, and partially numeric `content-length` values before parsing them.
- Preserve existing behavior for valid decimal integer `content-length` values.

### Features

N/A

### Bug Fixes

Request `content-length` malformed values are rejected with `InvalidArgumentError('invalid content-length header')` instead of being normalized or deferred into a later mismatch error.

Requests that previously accepted malformed `content-length` values will now fail immediately during header validation.

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -27,7 +27,21 @@ const { headerNameLowerCasedRecord } = require('./constants')
 
 // Verifies that a given path is valid does not contain control chars \x00 to \x20
 const invalidPathRegex = /[^\u0021-\u00ff]/
-const contentLengthHeaderValueRegex = /^\d+$/
+
+function isValidContentLengthHeaderValue (val) {
+  if (typeof val !== 'string' || val.length === 0) {
+    return false
+  }
+
+  for (let i = 0; i < val.length; i++) {
+    const charCode = val.charCodeAt(i)
+    if (charCode < 48 || charCode > 57) {
+      return false
+    }
+  }
+
+  return true
+}
 
 const kHandler = Symbol('handler')
 const kController = Symbol('controller')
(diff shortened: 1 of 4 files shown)


mcollina (approved):
lgtm

PR #5069, opened 2026-04-19
fix(http2): reject upgrade streams closed before response headers

### LINKED ISSUE #5064: HTTP/2 websocket client.upgrade() hangs if the stream closes before response headers
## Bug Description

When `Client#upgrade()` is used over HTTP/2 for a websocket extended CONNECT, the returned promise/callback can hang indefinitely if the peer closes the stream before sending response headers.

## Reproducible By

1. Start an HTTP/2 server.
2. Accept the websocket extended CONNECT stream.
3. Close the stream before calling `stream.respond()`.
4. Call `client.upgrade({ path: '/', protocol: 'websocket' })` with `allowH2: true`.
5. Observe that the returned promise never settles.

```js
import { createSecureServer } from "node:http2";
import { generate } from "@metcoder95/https-pem";

const server = createSecureServer({
  ...(await generate({ opts: { keySize: 2048 } })),
  settings: { enableConnectProtocol: true },
});

server.on("stream", (stream) => {
  // No response headers are sent before the stream is closed.
  stream.end();
  // `stream.close()` or another premature close path appears to trigger the same hang.
});

await once(server.listen(0), "listening");

  allowH2: true,
  connect: {
    rejectUnauthorized: false,
  },
});

await client.upgrade({ path: "/", protocol: "websocket" });
console.log("unreachable: promise stays pending");
```

## Expected Behavior

`client.upgrade()` should always settle. If the HTTP/2 stream closes before response headers are received, it should reject with a stream/protocol error instead of hanging indefinitely.

## Logs & Screenshots

No error is surfaced to the caller in this case. The promise remains pending until the caller provides its own timeout or abort signal.

## Environment

macOS 26.4.1
Node v24.14.1
undici v8.1.0

trivikr:
## This relates to...

Fixes: https://github.com/nodejs/undici/issues/5064

## Rationale

The H2 upgrade paths only advanced the request once a `response` event arrived.

If the peer closed the stream before sending headers, the request was never settled and the queue slot was not released, which could leave the operation hanging and interfere with later cleanup.

## Changes

- Added shared pre-response handling for H2 upgrade streams in `lib/dispatcher/client-h2.js`
- Count upgrade/connect streams as open immediately after `session.request(...)`
- Reject upgrade/connect requests on pre-header `error`, `end`, or `close`
- Advance the client queue when those failures happen so the request does not remain stuck

### Features

N/A

### Bug Fixes

- Fixed an H2 websocket upgrade hang when the server closes the stream before sending response headers
- Fixed the same pre-header hang in the H2 `CONNECT` path
- Fixed request queue cleanup for those failure cases so `client.close()` does not wait on a stuck upgrade/connect request

N/A

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -532,12 +532,16 @@ function writeH2 (client, request) {
 
       stream.once('response', (headers, _flags) => {
         responseReceived = true
-        stream.off('error', onUpgradeStreamError)
-        stream.off('end', onUpgradeStreamEnd)
 
         const { [HTTP2_HEADER_STATUS]: statusCode, ...realHeaders } = headers
 
         request.onRequestUpgrade(statusCode, parseH2Headers(realHeaders), stream)
+
+        if (!request.aborted && !request.completed) {
+          stream.off('error', onUpgradeStreamError)
+          stream.off('end', onUpgradeStreamEnd)
+        }
+
         client[kQueue][client[kRunningIdx]++] = null
       })
 
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -62,9 +62,7 @@ const {
     HTTP2_HEADER_CONTENT_LENGTH,
     HTTP2_HEADER_EXPECT,
     HTTP2_HEADER_STATUS,
-    HTTP2_HEADER_PROTOCOL,
-    NGHTTP2_REFUSED_STREAM,
-    NGHTTP2_CANCEL
+    HTTP2_HEADER_PROTOCOL
   }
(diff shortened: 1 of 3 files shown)


mcollina:
CI is very red

trivikr:
The CI failures caused by timeouts in `‎test/websocket/opening-handshake.js` were fixed by tracking server-side HTTP/2 sessions and closing them explicitly during teardown

mcollina (approved):
lgtm

mcollina (approved):
lgtm

PR #5076, opened 2026-04-20
fix(client-h2): stop double-decrementing kOpenStreams on stream timeout

### LINKED ISSUE #5073: HTTP/2 stream timeouts decrement session open-stream count twice
## Bug Description

HTTP/2 stream timeouts decrement session open-stream count twice

## Reproducible By

1. Start an HTTP/2 TLS server that sends response headers immediately but delays the response body longer than the client's `bodyTimeout`.
2. Create an Undici `Client` and a small `bodyTimeout` such as `50`.
3. Send a request and read `res.body.text()` so the request hits the HTTP/2 stream timeout path.
4. Inspect `client[kHTTP2Session]` after the timeout and read the session symbol whose description is `open streams`.
5. Observe that the counter is `-1` after a single timed-out request.

Minimal repro
```js
import { createSecureServer } from 'node:http2'
import { once } from 'node:events'
import pem from '@metcoder95/https-pem'
import symbols from 'undici/lib/core/symbols.js'

const { kHTTP2Session } = symbols

const server = createSecureServer(await pem.generate({ opts: { keySize: 2048 } }))

server.on('stream', (stream) => {
  stream.respond({ ':status': 200, 'content-type': 'text/plain' })
  setTimeout(() => stream.end('late body'), 300)
})

await once(server.listen(0), 'listening')

  connect: { rejectUnauthorized: false },
  bodyTimeout: 50
})

const res = await client.request({ path: '/', method: 'GET' })

try {
  await res.body.text()
} catch (err) {
  console.log('request error:', err.message)
}

const session = client[kHTTP2Session]
const openStreams = Object.getOwnPropertySymbols(session)
  .find((sym) => sym.description === 'open streams')

console.log('open streams after timeout:', session[openStreams])

await client.close()
await new Promise((resolve) => server.close(resolve))
```

## Expected Behavior

A timed-out HTTP/2 request should decrement `session[kOpenStreams]` exactly once, leaving the counter at `0` after the stream is closed. The timeout path should not underflow the counter or leave future stream accounting inconsistent.

## Logs & Screenshots

Output for repro

```console
request error: HTTP/2: "stream timeout after 50"
open streams after timeout: -1
```

## Environment

macOS 26.4.1
Node v24.14.1
undici v8.1.0

SAY-5:
Fixes #5073.

The HTTP/2 stream `timeout` handler in `lib/dispatcher/client-h2.js` does:

```js
stream.on('timeout', () => {
  const err = new InformationalError(`HTTP/2: "stream timeout after ${requestTimeout}"`)
  stream.removeAllListeners('data')
  session[kOpenStreams] -= 1
  if (session[kOpenStreams] === 0) session.unref()
  abort(err)
})
```

`abort(err)` ultimately destroys the http2 stream, which fires `close`, and the `close` listener a few lines above already decrements `kOpenStreams`:

```js
stream.once('close', () => {
  stream.removeAllListeners('data')
  session[kOpenStreams] -= 1
  if (session[kOpenStreams] === 0) session.unref()
})
```

Result: a single timed-out request knocks the counter down twice and it ends at `-1` instead of `0`. Future stream accounting on the session is then permanently skewed — the `=== 0` unref check can never match, and reused-session flows that rely on the counter reaching `0` to release the socket start holding sessions open longer than they should.

Leave the decrement to the `close` handler. The `timeout` handler only needs to stop the data listener and call `abort()`; destruction via `abort()` guarantees `close` fires exactly once. After this change the repro in the issue reports `open streams after timeout: 0` as expected.

### DIFF
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -757,12 +757,6 @@ function writeH2 (client, request) {
   stream.on('timeout', () => {
     const err = new InformationalError(`HTTP/2: "stream timeout after ${requestTimeout}"`)
     stream.removeAllListeners('data')
-    session[kOpenStreams] -= 1
-
-    if (session[kOpenStreams] === 0) {
-      session.unref()
-    }
-
     abort(err)
   })
 
--- a/test/http2-timeout.js
+++ b/test/http2-timeout.js
@@ -9,6 +9,7 @@ const { once } = require('node:events')
 const pem = require('@metcoder95/https-pem')
 
 const { Client } = require('..')
+const { kHTTP2Session } = require('../lib/core/symbols')
 
 test('Should handle http2 stream timeout', async t => {
   t = tspl(t, { plan: 1 })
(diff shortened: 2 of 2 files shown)


mcollina (changes requested):

mcollina (approved):
lgtm

PR #5077, opened 2026-04-20
fix: handle invalid HTTP/2 headers gracefully (#4356)

### LINKED ISSUE #4356: Invalid HTTP2 headers should be recoverable
## Bug Description

If an invalid HTTP2 header is sent by the server, the client should be able to recover by retrying the operation.  See https://github.com/nodejs/undici/discussions/4341 for more details.

## Reproducible By

To reproduce, a mock reply will need to be constructed where an HTTP2 connection to a server gets a response with an HTTP1 header like "http2-settings".  

Ideally a runtime error is thrown, which can be caught in a block where Undici can be directed to retry the operation with a new connection.

## Environment

This was encountered using Node.js 22 on RHEL9.

mcollina:
When session.request() throws synchronously due to invalid HTTP/2 headers (e.g., duplicate headers like "content-type" and "Content-Type"), the error was not being caught, causing an uncaught exception.

The fix wraps all session.request() calls with try-catch blocks. When an error is caught, only the request is errored (via util.errorRequest) and the function returns false. The session is not destroyed because the error is caused by the client sending invalid headers, not the server.

This allows the client to continue using the same session for subsequent requests, and the error is properly caught by user code.

Closes #4356

### DIFF
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -528,7 +528,17 @@ function writeH2 (client, request) {
         headers[HTTP2_HEADER_SCHEME] = protocol === 'http:' ? 'http' : 'https'
       }
 
-      stream = session.request(headers, { endStream: false, signal })
+      try {
+        stream = session.request(headers, { endStream: false, signal })
+      } catch (err) {
+        // An error here means the client sent invalid HTTP/2 headers
+        // (e.g., duplicate headers like "content-type" and "Content-Type").
+        // This is a client-side error, not a server-side error.
+        // We only error the request and do not destroy the session.
+        util.errorRequest(client, request, err)
+        return false
+      }
+
       stream[kHTTP2Stream] = true
 
       stream.once('response', (headers, _flags) => {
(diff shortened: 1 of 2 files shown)


lib/dispatcher/client-h2.js
```diff
@@ -528,7 +528,20 @@ function writeH2 (client, request) {
...
+      } catch (err) {
+        // An error here means the server sent invalid HTTP/2 headers
+        // (e.g., HTTP/1 headers like "http2-settings"). This is a server error.
+        // We destroy the socket/session and error the request so the client
+        // can retry with a new connection.
+        util.errorRequest(client, request, err)
+        const socket = session[kSocket]
+        session.destroy(err)
+        util.destroy(socket, err)
+        return false
```
webcarrot, line 542: I don’t know how h2 works, but did this cause other "valid", ongoing requests that use the same socket/session to be interrupted?

metcoder95 (commented):
Tests seems, failing; rest lgtm

mcollina:
@metcoder95 PTAL

cesarvspr (approved):
Thanks for this. I read through it and ran it locally.

The fix looks right to me. Wrapping session.request and only erroring the request, while leaving the session alive, matches the existing pattern just above at the upgrade check, and not destroying the session on a client side header mistake seems like the correct call. The second test is a nice regression guard: when I reverted the patch locally it failed exactly as expected, with an uncaughtException and code ERR_HTTP2_HEADER_SINGLE_VALUE, then passed again once restored.

One small thing, please take it or leave it. The first test, "invalid headers should be recoverable", sends two valid GETs and never actually sends an invalid header, so it stays green even with the fix reverted. It looks like the real coverage lives entirely in the second test. If it is useful, that first test could send a bad request, catch the throw, then reuse the same client for a good request, which would prove the session really did recover. Completely optional and non blocking.

Thanks again for sorting this out.

metcoder95 (approved):

metcoder95:
just conflicts to solve and ready to merge

webcarrot:
It was fixed in:
 - https://github.com/nodejs/undici/pull/5101
 - https://github.com/nodejs/undici/pull/5311

This one should be closed.

PR #5078, opened 2026-04-20
fix(websocket): queue close/error events asynchronously when closing during CONNECTING

### LINKED ISSUE #4741: WebSocket events fire synchronously during close() instead of asynchronously

maruthang:
## This relates to...

Fixes #4741

## Rationale

Per the [WHATWG WebSocket spec](https://websockets.spec.whatwg.org/#dom-websocket-close), when `close()` is invoked on a `WebSocket` whose `readyState` is `CONNECTING`, the implementation must "fail the WebSocket connection", and all resulting event dispatches (`error`, `close`) must be delivered via a *queued task* — not synchronously from the `close()` call itself.

Prior to this PR, undici dispatched those events synchronously from inside `close()` via `failWebsocketConnection` → `handler.onSocketClose()`. User code observed `error` and `close` events *before* the `close()` method returned, which is observably different from browser behavior and breaks consumers that set listeners after calling `close()`.

The fix is intentionally scoped to the `CONNECTING` branch of `closeWebSocketConnection`. Other callers of `failWebsocketConnection` (receiver errors, handshake failures, `WebSocketStream`'s abort listener) keep their current synchronous behavior, preserving existing tests such as `test/websocket/stream/abort-before-open.js`.

## Changes

- `lib/web/websocket/connection.js`: In `closeWebSocketConnection`, when the socket is still `CONNECTING`, continue to abort the underlying fetch synchronously and set `readyState = CLOSING` (so that `close()` returns with the correct observable state), but defer the `onSocketClose()` dispatch — which fires `error` and `close` — to a `queueMicrotask`. The microtask guards against re-entry (e.g. a racing socket close or a second `close()` call) by bailing when `readyState` is already `CLOSED`.
- `test/websocket/issue-4741.js`: New regression test asserting that no `error` or `close` event is dispatched synchronously from `close()` when the socket is in `CONNECTING`.

### Features

N/A

### Bug Fixes

- Queue `close`/`error` events asynchronously in `closeWebSocketConnection` when the socket is still `CONNECTING` (#4741).

None.

## Status

- [x] Tested
- [S] Documented (no public API surface changed)
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/web/websocket/connection.js
+++ b/lib/web/websocket/connection.js
@@ -243,8 +243,23 @@ function closeWebSocketConnection (object, code, reason, validate = false) {
     // Do nothing.
   } else if (!isEstablished(object.readyState)) {
     // Fail the WebSocket connection and set object’s ready state to CLOSING (2). [WSP]
-    failWebsocketConnection(object)
+    //
+    // Per https://websockets.spec.whatwg.org/#feedback-from-the-protocol, the
+    // error and close events that result from failing the WebSocket connection
+    // must be fired by queuing a task rather than synchronously from inside
+    // close(). Abort the underlying fetch now, but defer firing the events to
+    // a microtask so that `close()` returns before any event is dispatched.
+    // See https://github.com/nodejs/undici/issues/4741
+    object.controller.abort()
     object.readyState = states.CLOSING
+    queueMicrotask(() => {
+      // The connection may already be closed (e.g. close() was called again,
+      // or the socket finished closing) by the time the microtask runs. Only
+      // dispatch the close events when they have not yet been fired.
+      if (!isClosed(object.readyState)) {
+        object.onSocketClose()
+      }
+    })
   } else if (!object.closeState.has(sentCloseFrameState.SENT) && !object.closeState.has(sentCloseFrameState.RECEIVED)) {
     // Upon either sending or receiving a Close control frame, it is said
     // that _The WebSocket Closing Handshake is Started_ and that the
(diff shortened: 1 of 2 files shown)


KhafraDev (changes requested):
Please look at prior PRs before opening slop PRs and wasting our time reviewing. Jfc.

https://github.com/nodejs/undici/pull/4745#issuecomment-3764404322

maruthang:
Apologies @KhafraDev — I hadn't seen #4745 or your comment there before opening this, and you're right that a localized microtask in `closeWebSocketConnection` is the wrong shape for this problem. The spec-aligned fix ("queue a task") belongs one level up in `establishWebSocketConnection` / `closeWebSocketConnection` / the message-received path, and once that exists the state-transition guards I added shouldn't be necessary.

Closing this in favor of going back and doing it properly. I'll study the existing work on #4745 first so I'm not duplicating effort or getting in your way. If a fresh PR would be useful at some point, I'll only open one after reading the thread and only if it's clearly a different / complementary direction — otherwise I'll stay out of it.

Sorry for the review time.

Issue #4625, opened 2025-10-16
More Descriptive Error When WebSocket Closes with Code 1006


adiwajshing:
## Bug Description

The new WebSocket library is pretty great, but when a WebSocket connection closes without completing the control frame exchange, it spits out an empty message TypeError, which isn't very descriptive of the problem and appears as an error message that accidentally slipped through the deployment. The exact line of code is:
https://github.com/nodejs/undici/blob/dc5802db4f9940f5f39c3e338f4bd5d64c47e3f4/lib/web/websocket/websocket.js#L589-L591

## Reproducible By

Haven't written exact steps to reproduce, but it happens occasionally on connection closure.

## Expected Behavior

Perhaps a better description of the connection closure in the ErrorEvent? "Connection closed with code: xyz"
Also imo, would be better to use a plain Error rather than a TypeError.

If this is something the team feels would be a good addition, happy to make a PR

## Logs & Screenshots

```
TypeError\n    at #onSocketClose (node:internal/deps/undici/undici:13908:20)\n    at TLSSocket.onSocketClose (node:internal/deps/undici/undici:13611:72)\n    at TLSSocket.emit (node:events:520:35)\n    at node:net:346:12\n    at TCP.done (node:_tls_wrap:649:7)
```

## Environment

Ubuntu 20, NodeJS 24.0

KhafraDev:
> Perhaps a better description of the connection closure in the ErrorEvent? "Connection closed with code: xyz"

In this case you would lose the reason the connection error'd. I'd like to know how we're generating an empty reason before changing the default error message as it could be a bug, which I'd rather fix than patch over.

>  Also imo, would be better to use a plain Error rather than a TypeError.

It should stay as a TypeError. Web specs typically default to TypeErrors over plain errors, so we mimic it.

adiwajshing:
> It should stay as a TypeError. Web specs typically default to TypeErrors over plain errors, so we mimic it.

Fair enough.

> In this case you would lose the reason the connection error'd. I'd like to know how we're generating an empty reason before changing the default error message as it could be a bug, which I'd rather fix than patch over.

So if you see [these lines](https://github.com/nodejs/undici/blob/dc5802db4f9940f5f39c3e338f4bd5d64c47e3f4/lib/web/websocket/websocket.js#L564-L591) -- it appears that the reason is just never set, the empty string set on L565 is passed into the TypeError on L590. Might just be a bug if that's not intended behaviour.

[...]

PR #5079, opened 2026-04-20
fix(websocket): populate ErrorEvent message on failed connection

### LINKED ISSUE #4735: WebSocket emits error event with an empty message when cannot open connection
## Bug Description

undici >= 7.16.0 emits ErrorEvent with such error property:

```json
{
  "name": "TypeError",
  "message": ""
}
```

undici < 7.16.0 emits ErrorEvent with such error property:

```json
{
  "name": "Error",
  "message": "Received network error or non-101 status code."
}
```

## Reproducible By

```ts
import { WebSocket } from 'undici'
import { type TestContext, test } from 'node:test'

test('should emit error event with Error message', { plan: 1 }, (t: TestContext, done): void => {
  const webSocket = new WebSocket('https://not-a-ws-server.com/')

  webSocket.onerror = ({ error }) => {
    done(
      t.assert.deepStrictEqual(
        {
          name: error.name,
          message: error.message,
        },
        {
          name: 'Error',
          message: 'Received network error or non-101 status code.',
        }
      )
    )
  }
})
```

## Expected Behavior

Error should have any message explaining why event occurred.

## Logs & Screenshots

```
  AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
  + actual - expected

    {
  +   message: '',
  +   name: 'TypeError'
  -   message: 'Received network error or non-101 status code.',
  -   name: 'Error'
    }
```

## Environment

Windows, Node v25.3.0

### Additional context

MDN doesn't describe [WebSocket error event](https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/error_event) listener argument as ErrorEvent, just plain Event which doesn't have error property.

But since undici uses [ErrorEvent](https://developer.mozilla.org/en-US/docs/Web/API/ErrorEvent), it should have some message, not just an empty string.

maruthang:
## This relates to...

Fixes #4735

## Rationale

Since undici 7.16.0 (PR #4521, commit `909a5845`), the `ErrorEvent` dispatched on a `WebSocket` when the opening handshake fails carries an empty string as its `message`. The old behavior surfaced a descriptive reason (e.g. `"Received network error or non-101 status code."`), which is what consumers — and the WPT / nodejs test fixtures — rely on to diagnose connection failures.

The regression comes from `WebSocket#onSocketClose` unconditionally constructing `new TypeError(reason)` where `reason` defaulted to `""`, discarding the descriptive reason that `failWebsocketConnection` had already produced internally.

This change restores the pre-regression behavior by plumbing the failure reason (and `cause`, when present) from `failWebsocketConnection` through to `#onSocketClose`, so the dispatched `ErrorEvent` once again has a non-empty `message` and its `error` is a `TypeError` with a meaningful `message` (and a `cause` when the underlying failure provides one). The error type remains `TypeError`, consistent with maintainer guidance on the duplicate issue #4625.

## Changes

- `lib/web/websocket/connection.js`: `failWebsocketConnection` now stores `{ code, reason, cause }` on a new `handler.closeInfo` slot before invoking the socket close path, so the reason is not lost by the time `#onSocketClose` runs.
- `lib/web/websocket/websocket.js`: handler typedef and initialization gain a `closeInfo` slot; `#onSocketClose` consumes it to build an `ErrorEvent` whose `message` matches the pre-regression string and whose `error` is a `TypeError` carrying that same `message` (plus `cause` when available).
- `test/websocket/issue-4735.js`: new regression test that opens a `WebSocket` to a closed port and asserts the `error` event's `message` is a non-empty string.

### Features

N/A

### Bug Fixes

- **Bug Fixes**: Populate `ErrorEvent` `message`/`error` with the underlying failure reason on WebSocket connect failures (#4735).

None.

## Status

- [x] Tested
- [S] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/web/websocket/connection.js
+++ b/lib/web/websocket/connection.js
@@ -315,7 +315,11 @@ function failWebsocketConnection (handler, code, reason, cause) {
   handler.controller.abort()
 
   if (isConnecting(handler.readyState)) {
-    // If the connection was not established, we must still emit an 'error' and 'close' events
+    // If the connection was not established, we must still emit an 'error' and 'close' events.
+    // Propagate the reason (and optional cause) so that #onSocketClose can surface a non-empty
+    // ErrorEvent.error.message describing why the connection failed.
+    // Refs https://github.com/nodejs/undici/issues/4735
+    handler.closeInfo = { code, reason, cause }
     handler.onSocketClose()
   } else if (handler.socket?.destroyed === false) {
     handler.socket.destroy()
--- a/lib/web/websocket/websocket.js
+++ b/lib/web/websocket/websocket.js
@@ -54,6 +54,7 @@ function getSocketAddress (socket) {
  * @property {Set<number>} closeState
  * @property {import('../fetch/index').Fetch} controller
  * @property {boolean} [wasEverConnected=false]
+ * @property {{ code: number|null|undefined, reason: string|undefined, cause: unknown }|null} [closeInfo]
  */
 
 // https://websockets.spec.whatwg.org/#interface-definition
@@ -114,7 +115,8 @@ class WebSocket extends EventTarget {
     socket: null,
     closeState: new Set(),
     controller: null,
-    wasEverConnected: false
+    wasEverConnected: false,
+    closeInfo: null
   }
 
   #url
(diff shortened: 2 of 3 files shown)


KhafraDev (changes requested):
This just isn't good code, and we're already not following the spec by emitting an ErrorEvent.

maruthang:
Understood @KhafraDev — thanks for the direction. If the `ErrorEvent` emission itself is the non-conformance, then patching its `message` is painting over the real problem. I won't reopen; if I look at this area again it'll be after understanding what spec-conformant error reporting should look like here, and I'll raise a discussion issue first rather than jumping to a PR.

Sorry for the noise.

PR #5080, opened 2026-04-20
test(mock): cover SnapshotAgent excludeUrls playback

### LINKED ISSUE #4663: SnapshotAgent in playback mode errors when it encounters an excluded url

maruthang:
## This relates to...

Fixes #4663

## Rationale

Issue #4663 reported that `SnapshotAgent` in `playback` mode threw "No snapshot found" for URLs matched by `excludeUrls`, even though excluded URLs are meant to pass through to the real network.

The underlying bug was already fixed on `main` by PR #4670 (commit `5024d1b7`), which routes excluded URLs through `kRealAgent.dispatch` before `findSnapshot` is consulted (see `lib/mock/snapshot-agent.js:85-88`). However, the issue was never formally closed and no regression test was added for this specific playback-mode path.

This PR adds a focused regression test so the excludeUrls playback short-circuit cannot silently regress. It reproduces the exact scenario from the issue and also covers a RegExp pattern that only matches during playback, which exercises the dispatch branch end-to-end.

## Changes

Add `test/mock/issue-4663.js` with two cases:
1. String `excludeUrls` entry — excluded URL is fetched live during playback.
2. RegExp `excludeUrls` entry — excluded URL is fetched live during playback.

### Features

N/A

### Bug Fixes

N/A — the bug itself was fixed in #4670. This PR adds regression coverage only.

N/A

## Status

- [x] Tested (`test/mock/issue-4663.js`; all `snapshot-*` and `mock-*` tests pass)
- [S] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/test/mock/issue-4663.js
+++ b/test/mock/issue-4663.js
@@ -1,0 +1,128 @@
+'use strict'
+
+// Regression test for https://github.com/nodejs/undici/issues/4663
+// "SnapshotAgent in playback mode errors when it encounters an excluded url".
+//
+// When a URL is listed in excludeUrls it must not be matched against recorded
+// snapshots during playback; instead the request should pass through to the
+// real network unchanged.
+
+const { describe, it } = require('node:test')
+const assert = require('node:assert')
+const { createServer } = require('node:http')
+const { once } = require('node:events')
+const { unlink } = require('node:fs/promises')
+const { tmpdir } = require('node:os')
+const { join } = require('node:path')
+const { SnapshotAgent, setGlobalDispatcher, getGlobalDispatcher, request } = require('../..')
+
+function snapshotPath (suffix) {
+  return join(tmpdir(), `undici-issue-4663-${suffix}-${process.pid}-${Date.now()}.json`)
+}
+
+async function startLiveServer (t) {
+  const server = createServer((req, res) => {
+    res.writeHead(200, { 'content-type': 'text/plain' })
+    res.end(`live:${req.url}`)
+  }).listen(0)
+  await once(server, 'listening')
+  t.after(() => {
+    server.closeAllConnections?.()
+    server.close()
+  })
+  const { port } = server.address()
+  return `http://127.0.0.1:${port}`
+}
+
+function restoreGlobalDispatcher (t) {
+  const original = getGlobalDispatcher()
+  t.after(() => setGlobalDispatcher(original))
+}
+
+describe('SnapshotAgent - issue #4663', () => {
+  it('playback does not throw for a URL matched by a string excludeUrls entry', async (t) => {
+    const origin = await startLiveServer(t)
+    const file = snapshotPath('string')
+    t.after(() => unlink(file).catch(() => {}))
+    restoreGlobalDispatcher(t)
+
+    // Record phase - capture /kept, exclude /skipped.
+    const recorder = new SnapshotAgent({
+      mode: 'record',
+      snapshotPath: file,
+      excludeUrls: [`${origin}/skipped`]
+    })
+    t.after(async () => recorder.close())
+    setGlobalDispatcher(recorder)
+
+    await (await request(`${origin}/kept`)).body.text()
+    await (await request(`${origin}/skipped`)).body.text()
+    await recorder.saveSnapshots()
+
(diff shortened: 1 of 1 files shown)


mcollina:
WHy is the test included in https://github.com/nodejs/undici/pull/4670 not enough?

maruthang:
Fair point @mcollina — the `excluded URLs should not error in playback mode` test added in #4670 already covers the exact #4663 scenario (string-form `excludeUrls`, playback, excluded URL passes through to the real server). The only delta in this PR is exercising the same codepath with a RegExp-form `excludeUrls` (the existing regex test at `test/snapshot-testing.js:1240` only covers record mode). Given `isUrlExcluded` normalizes both string and RegExp into the same branch inside `dispatch()`, the incremental coverage is thin.

Happy to close this if you'd prefer to keep the suite lean — the issue is already functionally closed by #4670. If you think the RegExp-in-playback case is worth pinning separately I can leave it, but otherwise just say the word and I'll close.

metcoder95 (approved):

maruthang:
Friendly nudge — approved by @metcoder95, all 35 checks green. Tests-only change covering the `SnapshotAgent excludeUrls` playback-mode regex form. @mcollina — happy for this to land whenever you have a moment. Thanks!

PR #5086, opened 2026-04-21
fix(readable): install a StringDecoder when setEncoding is called

SAY-5:
## What kind of change does this PR introduce?

bug fix

## What is the current behavior?

`BodyReadable.setEncoding` only wrote `_readableState.encoding` and skipped the Node-standard step of installing a `StringDecoder`. With an encoding set but no decoder, the `Readable` machinery falls through to `Buffer.prototype.toString(encoding)` on each individual chunk, which corrupts any multi-byte character (UTF-8 CJK, emoji, any 3+ byte sequence) that straddles a chunk boundary. The partial bytes become U+FFFD, and the next chunk's leading bytes do too.

Fixes #5002.

## What is the new behavior?

Match Node's `Readable.prototype.setEncoding`: build a `StringDecoder` from the supplied encoding and wire it into `_readableState`. The decoder buffers unfinished byte sequences between chunks, so callers iterating `for await (const chunk of body)` with `setEncoding('utf8')` get lossless text.

## Other information

Added `test/readable-set-encoding-utf8.js`: pushes the 3-byte sequence for `中` split across two chunks and asserts no `U+FFFD` in the decoded output. Pre-fix the test fails with `"\uFFFD\uFFFD"`; post-fix the decoded string is exactly `"中"`.


mcollina (commented):
Can you check why was this done in this way? Also, why are we overriding setEncoding now? Shouldn't removing the override do the same?

PR #5088, opened 2026-04-22
fix(websocket): queue close events after CONNECTING close

### LINKED ISSUE #4741: WebSocket events fire synchronously during close() instead of asynchronously

tejasp2022:
## Summary
- add a regression test for calling `WebSocket.close()` while the socket is still `CONNECTING`
- queue the CONNECTING failure close-path for `WebSocket` so `error` and `close` fire after `close()` returns
- keep `WebSocketStream` on the old synchronous path so its pre-open abort behavior stays unchanged

Fixes #4741

## Test plan
- `node --test test/websocket/issue-4741.js test/websocket/issue-3546.js test/websocket/issue-4273.js test/websocket/issue-3697-2399493917.js test/websocket/stream/abort-before-open.js`
- `npm run test:websocket` on Node `22.19.0`

## Notes
- `npm run test:websocket` still reports failures in `test/websocket/opening-handshake.js`, `test/websocket/issue-4989.js`, and `test/websocket/diagnostics-channel-handshake-response-h2.js`.
- I reproduced those same three H2 websocket failures on an untouched `main` clone under Node `22.19.0`, so they do not appear to be introduced by this patch.

### DIFF
--- a/lib/web/websocket/connection.js
+++ b/lib/web/websocket/connection.js
@@ -315,11 +315,11 @@ function failWebsocketConnection (handler, code, reason, cause) {
   handler.controller.abort()
 
   if (isConnecting(handler.readyState)) {
-    // If the connection was not established, we must still emit 'error' and
-    // 'close', but they must be queued after close() returns so readyState can
-    // observe the intermediate CLOSING state.
+    // WebSocket.close() while CONNECTING must queue error/close as a task so
+    // close() can return with readyState still at CLOSING. WebSocketStream
+    // keeps the existing synchronous path for its pre-open abort behavior.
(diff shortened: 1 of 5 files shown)


tejasp2022 (commented):
fix(websocket): queue close events after CONNECTING close

PR #5090, opened 2026-04-22
fix(http2): preserve accepted streams after GOAWAY

### LINKED ISSUE #5089: Graceful http2 GOAWAY is handled as hard connection failure
## Bug Description

Graceful http2 GOAWAY is handled as hard connection failure

## Reproducible By

```js
import { createSecureServer } from "node:http2";
import pem from "@metcoder95/https-pem";

const server = createSecureServer(
  await pem.generate({ opts: { keySize: 2048 } }),
);

const seen = [];
let session;

server.on("session", (s) => {
  session = s;
});

server.on("stream", (stream, headers) => {
  seen.push(headers[":path"]);

  stream.respond({ ":status": 200 });

  setTimeout(() => {
    if (!stream.closed) {
      stream.end(`ok:${headers[":path"]}`);
    }
  }, 200);

  if (seen.length === 3) {
    setTimeout(() => {
      session.goaway();
    }, 50);
  }
});

await once(server.listen(0), "listening");

  pipelining: 10,
  connect: { rejectUnauthorized: false },
});

const results = await Promise.allSettled(
  [1, 2, 3].map((i) =>
    client
      .request({ path: `/${i}`, method: "GET" })
      .then(async (res) => res.body.text()),
  ),
);

console.log("server saw", seen);
console.log(
  results.map((r) => {
    return r.status === "fulfilled"
      ? { status: r.status, value: r.value }
      : { status: r.status, code: r.reason.code, message: r.reason.message };
  }),
);

await client.close();
await new Promise((resolve) => server.close(resolve));
```

Observe that the server receives all 3 requests before `GOAWAY`, but all 3 client requests reject.

## Expected Behavior

A graceful code-0 `GOAWAY` should stop the client from creating new streams on that session, but streams already accepted by the server should still be allowed to finish when they are within `lastStreamID`.

For the repro above, I would expect the 3 in-flight requests to resolve successfully because the server had already accepted them before calling `session.goaway()`.

## Logs & Screenshots

Observed output on this checkout:

```txt
server saw [ '/1', '/2', '/3' ]
[
  {
    status: 'rejected',
    code: 'UND_ERR_SOCKET',
    message: 'HTTP/2: "GOAWAY" frame received with code 0'
  },
  {
    status: 'rejected',
    code: 'UND_ERR_SOCKET',
    message: 'HTTP/2: "GOAWAY" frame received with code 0'
  },
  {
    status: 'rejected',
    code: 'UND_ERR_SOCKET',
    message: 'HTTP/2: "GOAWAY" frame received with code 0'
  }
]
```

## Environment

macOS 26.4.1
Node v24.15.0
undici v8.1.0

trivikr:
## This relates to...

Fixes: https://github.com/nodejs/undici/issues/5089

## Rationale

HTTP/2 `GOAWAY` with `NO_ERROR` should not fail streams the server has already accepted.

This change preserves in-flight accepted streams, retries work beyond `lastStreamID`, and updates the tests to cover the intended graceful shutdown behavior.

## Changes

- Track request stream IDs in the HTTP/2 client
- Treat `GOAWAY` with code `0` as informational rather than a socket failure
- Only retire/requeue requests whose stream IDs are beyond `lastStreamID`
- Allow already-accepted streams to complete normally after `GOAWAY`

### Features

N/A

### Bug Fixes

- Fix handling of HTTP/2 `GOAWAY` so accepted streams are not failed prematurely
- Fix client reconnection behavior after graceful `GOAWAY`

N/A

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

Assisted-by: openai:gpt-5.4

### DIFF
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -67,7 +67,8 @@ const {
     HTTP2_HEADER_EXPECT,
     HTTP2_HEADER_STATUS,
     HTTP2_HEADER_PROTOCOL,
-    NGHTTP2_NO_ERROR
+    NGHTTP2_NO_ERROR,
+    NGHTTP2_REFUSED_STREAM
   }
 } = http2
 
@@ -135,6 +136,24 @@ function clearRequestStream (request) {
   const cleanup = request[kRequestStreamCleanup]
   detachRequestFromStream(request)
   cleanup?.()
+}
+
+function canRetryRequestAfterGoAway (request) {
+  const { body } = request
+
+  return body == null || util.isBuffer(body) || util.isBlobLike(body)
+}
+
+function closeRequestStream (request, code = NGHTTP2_REFUSED_STREAM) {
+  const stream = request[kRequestStream]
+
+  clearRequestStream(request)
+
+  if (stream != null && !stream.destroyed && !stream.closed) {
+    try {
+      stream.close(code)
+    } catch {}
+  }
 }
 
 function connectH2 (client, socket) {
@@ -376,13 +395,26 @@ function onHttp2SessionGoAway (errorCode, lastStreamID) {
   const client = this[kClient]
   const previousPendingIdx = client[kPendingIdx]
   const pendingIdx = getGoAwayPendingIdx(client, lastStreamID)
+  const retriableRequests = []
 
   for (let i = pendingIdx; i < previousPendingIdx; i++) {
     const request = client[kQueue][i]
 
     if (request != null) {
-      clearRequestStream(request)
+      closeRequestStream(request)
+
+      if (canRetryRequestAfterGoAway(request)) {
+        retriableRequests.push(request)
+      } else {
+        util.errorRequest(client, request, err)
+      }
     }
+  }
+
+  if (pendingIdx !== previousPendingIdx) {
+    const remainingPendingRequests = client[kQueue].slice(previousPendingIdx)
+    client[kQueue].length = pendingIdx
+    client[kQueue].push(...retriableRequests, ...remainingPendingRequests)
   }
 
   if (client[kHTTP2Session] === this) {
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -41,6 +41,8 @@ const {
 const { channels } = require('../core/diagnostics.js')
 
 const kOpenStreams = Symbol('open streams')
+const kRequestStreamId = Symbol('request stream id')
+const kReceivedGoAway = Symbol('received goaway')
 
 let extractBody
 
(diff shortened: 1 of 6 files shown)


lib/dispatcher/client-h2.js
```diff
@@ -698,8 +753,13 @@ function writeH2 (client, request) {
 
   // Track whether we received a response (headers)
   let responseReceived = false
+  const isActiveStream = () => request[kRequestStreamId] === stream.id
 
   stream.once('response', headers => {
+    if (!isActiveStream()) {
+      return
+    }
```
mcollina, line 848: this should never happen, why is there this guard?

lib/dispatcher/client-h2.js
```diff
@@ -730,18 +790,24 @@ function writeH2 (client, request) {
   })
 
   stream.once('end', () => {
+    if (!isActiveStream()) {
+      return
+    }
```
mcollina: ditto

mcollina (changes requested):

lib/dispatcher/client-h2.js
trivikr: Because the same `request` can stop belonging to that particular HTTP/2 `stream` after the listeners are attached.

After `GOAWAY` handling, the requests on streams the server did not accept get `request[kRequestStreamId] = null` so they can be retried on a new session.

The guard is repeated at each handler because each event is independent and can arrive at a different time after the request-to-stream association has changed. A single earlier check would not protect later events.

Both the `#3753` and `#5089` tests hang if we remove this check.

mcollina: why aren't we removing the event hanlders then?

lib/dispatcher/client-h2.js
```diff
@@ -311,48 +340,56 @@ function onHttp2SessionEnd () {
...
-  client[kHTTPContext] = null
+  for (let i = pendingIdx; i < previousPendingIdx; i++) {
+    const request = client[kQueue][i]
 
-  // this is an HTTP2 session
-  this.close()
-  this[kHTTP2Session] = null
+    if (request != null) {
+      request[kRequestStreamId] = null
+    }
```
metcoder95, line 411: Following @mcollina point, maybe we can fully detach the `stream` events at this points; maybe keep a reference of the `stream` itself to the `request` to cleanup upon gracefull shutdown

lib/dispatcher/client-h2.js
trivikr: The function `removeRequestStreamListeners` was added in https://github.com/nodejs/undici/pull/5090/changes/a896f6a955f6c02b05a20362d788dfa75fbd612a

lib/dispatcher/client-h2.js
trivikr: The function `removeRequestStreamListeners` was added in https://github.com/nodejs/undici/pull/5090/changes/a896f6a955f6c02b05a20362d788dfa75fbd612a

lib/dispatcher/client-h2.js
trivikr: The function `clearRequestStream` was added in https://github.com/nodejs/undici/pull/5090/changes/a896f6a955f6c02b05a20362d788dfa75fbd612a

mcollina (approved):
lgtm

metcoder95 (approved):
great work! lgtm

PR #5097, opened 2026-04-24
fix(socks5): enforce authenticated state before CONNECT

### LINKED ISSUE #5096: Socks5Client.connect() can send CONNECT before authentication completes and after close/error
## Bug Description

`Socks5Client.connect()` does not enforce an authenticated/ready state before sending a SOCKS `CONNECT` request.

## Reproducible By

```js
import { EventEmitter } from "node:events";
import { Socks5Client } from "undici/lib/core/socks5-client.js";

class MockSocket extends EventEmitter {
  constructor() {
    super();
    this.writes = [];
    this.destroyed = false;
  }

  write(chunk) {
    this.writes.push(Buffer.from(chunk));
  }

  destroy() {
    this.destroyed = true;
  }
}

const socket = new MockSocket();
const client = new Socks5Client(socket);

client.connect("example.com", 80);

console.log(client.state); // connecting
console.log(socket.writes[0]); // CONNECT request already written
```

## Expected Behavior

`connect()` should only be allowed after SOCKS negotiation has completed successfully.

## Logs & Screenshots

```console
connecting
<Buffer 05 01 00 03 0b 65 78 61 6d 70 6c 65 2e 63 6f 6d 00 50>
```

## Environment

macOS 26.4.1
Node v24.15.0
undici v8.1.0

trivikr:
## This relates to...

Fixes: https://github.com/nodejs/undici/issues/5096

## Rationale

`Socks5Client` emitted `'authenticated'` on successful handshake/authentication but never transitioned into an authenticated state. As a result, `connect()` only rejected the already-connected state and could write a SOCKS `CONNECT` frame from invalid states such as `initial`, `handshaking`, `authenticating`, `closed`, or `error`.

This also affected `Socks5ProxyAgent`, which checks `socks5Client.state === 'authenticated'` in the `NO_AUTH` flow. Since that state was never set, the agent could miss the already-emitted `'authenticated'` event and wait until the authentication timeout.

## Changes

- add an explicit `AUTHENTICATED` state to `Socks5Client`
- transition to `AUTHENTICATED` before emitting `'authenticated'` in both auth-success paths
- reject `Socks5Client.connect()` unless the client is authenticated
- update `Socks5ProxyAgent` to check `STATES.AUTHENTICATED`

### Features

N/A

### Bug Fixes

- prevent `Socks5Client.connect()` from sending `CONNECT` before authentication completes
- prevent `Socks5Client.connect()` from sending `CONNECT` after invalid lifecycle states
- fix `Socks5ProxyAgent`’s `NO_AUTH` authentication readiness check to use the actual authenticated state

This changes the internal `Socks5Client.state` value after successful authentication from the previous implicit/incorrect behavior to an explicit `authenticated` state. `Socks5Client` is not part of the public top-level `undici` API, so there is no intended public API break.

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

Assisted-by: openai:gpt-5.4

### DIFF
--- a/lib/core/socks5-client.js
+++ b/lib/core/socks5-client.js
@@ -51,6 +51,7 @@ const STATES = {
   INITIAL: 'initial',
   HANDSHAKING: 'handshaking',
   AUTHENTICATING: 'authenticating',
+  AUTHENTICATED: 'authenticated',
   CONNECTING: 'connecting',
   CONNECTED: 'connected',
   ERROR: 'error',
@@ -139,6 +140,11 @@ class Socks5Client extends EventEmitter {
     }
   }
 
+  markAuthenticated () {
+    this.state = STATES.AUTHENTICATED
+    this.emit('authenticated')
+  }
+
   /**
    * Start the SOCKS5 handshake
    */
@@ -189,7 +195,7 @@ class Socks5Client extends EventEmitter {
     debug('server selected auth method', method)
 
     if (method === AUTH_METHODS.NO_AUTH) {
-      this.emit('authenticated')
+      this.markAuthenticated()
     } else if (method === AUTH_METHODS.USERNAME_PASSWORD) {
       this.state = STATES.AUTHENTICATING
       this.sendAuthRequest()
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

PR #5101, opened 2026-04-24
fix: handle invalid HTTP/2 connection headers (#4356)

### LINKED ISSUE #4356: Invalid HTTP2 headers should be recoverable

mcollina:
## Summary

- catch `ERR_HTTP2_INVALID_CONNECTION_HEADERS` from `session.request()` in the HTTP/2 client
- wrap the Node.js error in `InformationalError` and preserve the original error as `cause`
- tear down the broken HTTP/2 session so queued requests can continue on a fresh connection
- add an HTTP/2 regression test covering both catchable failure and queue recovery

Fixes #4356

## Testing

- `npx eslint lib/core/errors.js lib/dispatcher/client-h2.js test/http2-invalid-connection-headers.js`
- `npx borp -p "test/http2-invalid-connection-headers.js"`
- `npm run test:h2:core`
- `npm run test:h2:fetch`

### DIFF
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -491,6 +491,26 @@ function writeH2 (client, request) {
     util.destroy(body, err)
   }
 
+  const requestStream = (headers, options) => {
+    try {
+      return session.request(headers, options)
+    } catch (err) {
+      if (err?.code !== 'ERR_HTTP2_INVALID_CONNECTION_HEADERS') {
+        throw err
+      }
+
+      const wrappedErr = new InformationalError(err.message, { cause: err })
+      session[kError] = wrappedErr
+      session[kSocket][kError] = wrappedErr
+
+      session.destroy(wrappedErr)
+      util.destroy(session[kSocket], wrappedErr)
+      abort(wrappedErr)
+
+      return null
+    }
+  }
+
   try {
     // We are already connected, streams are pending.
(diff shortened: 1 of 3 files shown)


metcoder95 (approved):

lib/dispatcher/client-h2.js
```diff
@@ -489,6 +491,26 @@ function writeH2 (client, request) {
     util.destroy(body, err)
   }
 
+  const requestStream = (headers, options) => {
+    try {
+      return session.request(headers, options)
+    } catch (err) {
+      if (err?.code !== 'ERR_HTTP2_INVALID_CONNECTION_HEADERS') {
+        throw err
+      }
```
webcarrot, line 500: @mcollina how users will be able to catch non `ERR_HTTP2_INVALID_CONNECTION_HEADERS` errors in context of https://github.com/nodejs/undici/issues/4356 ?

PR #5109, opened 2026-04-25
fix(cache): persist vary when updating sqlite cache entries

### LINKED ISSUE #5108: SqliteCacheStore updates an existing cache row without updating the `vary` column
## Bug Description

SqliteCacheStore updates an existing cache row without updating the `vary` column

## Reproducible By

```js
import { cacheStores } from "undici";

const store = new cacheStores.SqliteCacheStore();
const now = Date.now();

// 1. Seed a non-varying entry.
store.set(
  {
    origin: "localhost",
    path: "/",
    method: "GET",
    headers: {},
  },
  {
    statusCode: 200,
    statusMessage: "initial",
    headers: {},
    cachedAt: now,
    staleAt: now + 1000,
    deleteAt: now + 2000,
    body: Buffer.from("initial"),
  },
);

// 2. Overwrite the same URL/method with a varying response.
// This takes the UPDATE path, but sqlite-cache-store.js does not update `vary`.
store.set(
  {
    origin: "localhost",
    path: "/",
    method: "GET",
    headers: {
      "accept-language": "en",
    },
  },
  {
    statusCode: 200,
    statusMessage: "updated",
    headers: {},
    vary: {
      "accept-language": "en",
    },
    cachedAt: now,
    staleAt: now + 1000,
    deleteAt: now + 2000,
    body: Buffer.from("updated"),
  },
);

// 3. This should miss, because the stored response now varies on accept-language.
const result = store.get({
  origin: "localhost",
  path: "/",
  method: "GET",
  headers: {
    "accept-language": "fr",
  },
});

// Expected: undefined
// Actual today: the updated row is returned, because `vary` stayed NULL in SQLite.
console.log(result);

store.close();
```

## Expected Behavior

When an existing cache row is overwritten, the `vary` column should also be updated.

If a response changes from non-varying to varying, later lookups should only match requests whose headers satisfy the stored `vary` constraints.

## Logs & Screenshots

```console
{
  body: <Buffer 75 70 64 61 74 65 64>,
  statusCode: 200,
  statusMessage: 'updated',
  headers: {},
  etag: undefined,
  vary: undefined,
  cacheControlDirectives: undefined,
  cachedAt: 1777086121760,
  staleAt: 1777086122760,
  deleteAt: 1777086123760
}
```

## Environment

macOS 26.4.1
Node v24.15.0
undici v8.1.0

trivikr:
## This relates to...

Fixes: https://github.com/nodejs/undici/issues/5108

## Rationale

`SqliteCacheStore#set()` reuses an existing row for the same URL/method, but the `UPDATE` statement did not include `vary`. That meant an entry could transition from non-varying to varying in memory while the SQLite row still held `NULL`, allowing later lookups to match unrelated request headers.

## Changes

- update `SqliteCacheStore`'s overwrite query to persist `vary`
- pass serialized `value.vary` through the update path in `set()`

### Features

N/A

### Bug Fixes

Fixed `SqliteCacheStore` failing to update `vary` when overwriting an existing cache entry

N/A

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

Assisted-by: openai:gpt-5.4

### DIFF
--- a/lib/cache/sqlite-cache-store.js
+++ b/lib/cache/sqlite-cache-store.js
@@ -173,6 +173,7 @@ module.exports = class SqliteCacheStore {
         headers = ?,
         etag = ?,
         cacheControlDirectives = ?,
+        vary = ?,
         cachedAt = ?,
         staleAt = ?
       WHERE
@@ -278,6 +279,7 @@ module.exports = class SqliteCacheStore {
         value.headers ? JSON.stringify(value.headers) : null,
         value.etag ? value.etag : null,
         value.cacheControlDirectives ? JSON.stringify(value.cacheControlDirectives) : null,
+        value.vary ? JSON.stringify(value.vary) : null,
         value.cachedAt,
         value.staleAt,
         existingValue.id
--- a/test/cache-interceptor/sqlite-cache-store-tests.js
+++ b/test/cache-interceptor/sqlite-cache-store-tests.js
@@ -287,3 +287,51 @@ test('SqliteCacheStore ignores expired Vary variants when a later one is still v
   notEqual(result, undefined)
   strictEqual(result.body.toString(), 'valid br')
 })
+
+test('SqliteCacheStore updates vary when overwriting an existing row', { skip: runtimeFeatures.has('sqlite') === false }, () => {
+  const store = new SqliteCacheStore()
+  const now = Date.now()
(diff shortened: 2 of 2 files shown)


mcollina (approved):
lgtm

PR #5110, opened 2026-04-25
refactor(h1): track HEAD keep-alive override as boolean

trivikr:
## This relates to...

HEAD response keep-alive handling in `client-h1`

## Rationale

The HEAD-specific keep-alive override was depending on 1this.connection.toLowerCase()` at headers-complete time. This change replaces that with an explicit tracked boolean so the parser carries only the state needed for the override and avoids string-based re-checks.

## Changes

- Track `Connection: keep-alive` as a boolean on the H1 parser
- Use that boolean in the HEAD keep-alive override path
- Reset the tracked state between responses

### Features

N/A

### Bug Fixes

N/A

N/A

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

Assisted-by: openai:gpt-5.4

### DIFF
--- a/lib/dispatcher/client-h1.js
+++ b/lib/dispatcher/client-h1.js
@@ -232,7 +232,7 @@ class Parser {
 
     this.keepAlive = ''
     this.contentLength = ''
-    this.connection = ''
+    this.connectionKeepAlive = false
     this.maxResponseSize = client[kMaxResponseSize]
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

PR #5114, opened 2026-04-25
perf(client): parse h1 content-length incrementally

trivikr:
## This relates to...

N/A

## Rationale

Avoid allocating and concatenating a `Content-Length` string in the HTTP/1.1 client parser before calling `parseInt()`. Parsing the header value incrementally reduces overhead in the hot response-header path.

## Changes

- Parse `Content-Length` incrementally into a number
- Replace string accumulation with numeric parsing state in the h1 parser

### Features

N/A

### Bug Fixes

Remove the extra string allocation and `parseInt()` call from response `Content-Length` parsing in the h1 client parser while preserving mismatch behavior

N/A

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

Assisted-by: openai:gpt-5.4

### DIFF
--- a/benchmarks/core/parse-content-length.mjs
+++ b/benchmarks/core/parse-content-length.mjs
@@ -1,0 +1,118 @@
+import { bench, group, run } from 'mitata'
+
+const CONTENT_LENGTH_STATE_START = 0
+const CONTENT_LENGTH_STATE_SIGN = 1
+const CONTENT_LENGTH_STATE_DIGITS = 2
+const CONTENT_LENGTH_STATE_NEGATIVE_SIGN = 3
+const CONTENT_LENGTH_STATE_DIGITS_NEGATIVE = 4
+const CONTENT_LENGTH_STATE_DONE = 5
+const CONTENT_LENGTH_STATE_INVALID = 6
+
+const values = [
+  Buffer.from('0'),
+  Buffer.from('9'),
+  Buffer.from('42'),
+  Buffer.from(' 42'),
+  Buffer.from('42 '),
+  Buffer.from('\t123'),
+  Buffer.from('1234'),
+  Buffer.from('65535'),
+  Buffer.from('1234567890'),
+  Buffer.from('abc'),
+  Buffer.from('   ')
+]
+
+function legacyParseContentLength (buf) {
+  return parseInt(buf.toString(), 10)
+}
+
+function incrementalParseContentLength (buf) {
+  let contentLength = 0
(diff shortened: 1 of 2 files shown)


trivikr:
Benchmark

```console
clk: ~3.18 GHz
cpu: Apple M1 Pro
runtime: node 24.15.0 (arm64-darwin)

benchmark                   avg (min … max) p75 / p99    (min … top 1%)
------------------------------------------- -------------------------------
• parse content-length
------------------------------------------- -------------------------------
legacy                       396.89 ns/iter 397.02 ns  █                   
                      (376.13 ns … 1.19 µs) 534.31 ns ▇█▇▂                 
                    ( 95.61  b … 260.82  b) 104.84  b ████▆▃▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁

incremental                   60.01 ns/iter  59.80 ns █                    
                      (59.10 ns … 89.05 ns)  70.63 ns █                    
                    (  0.10  b … 131.02  b)   0.34  b █▆▂▂▂▁▂▁▁▁▁▁▁▁▁▁▁▁▁▁▁
```

tsctx:
I think we can simplify this by making it stateless rather than stateful.
First, sign handling is unnecessary since Content-Length is strictly `1*DIGIT`.

We can rely on the following assumptions:

* llhttp has already validated that Content-Length is `1*DIGIT`.
* llhttp also ensures there are no duplicate values.

By initializing the value to -1, we can track the presence of the header without maintaining state. Here is the simplified implementation:
```js
class Parser {
  constructor(...) {
    ...
    this.contentLength = -1
  }
  onHeaderValue (buf) {
    if (...) {
    ...
    } else if (key.length === 14 && util.bufferToLowerCasedHeaderName(key) === 'content-length') {
       this.contentLength = buf.reduce((a, acc) => (acc * 10) + a, 0)
    }
  }
  onMessageComplete () {
     ...
     if (request.method !== 'HEAD' && contentLength !== -1 && bytesRead !== contentLength) {
        ...
     }
  }
}
```

trivikr:
> I think we can simplify this by making it stateless rather than stateful.

I posted a draft PR at https://github.com/nodejs/undici/pull/5124.
It's way easy to read, but the bechmarks show less performance improvement in it's current state.

I'm marking both as draft, so that we can take one which other maintainers are more interested in.

trivikr:
Closing as duplicate, as the benchmark for https://github.com/nodejs/undici/pull/5124 improved after using for loop instead of array.reduce().

PR #5116, opened 2026-04-25
fix: auto-detect HTTP proxy tunneling

### LINKED ISSUE #5093: Use of HTTP_PROXY & NODE_USE_ENV_PROXY fails with an infinite loop
## Bug Description

I have been experimenting with HTTP_PROXY & NODE_USE_ENV_PROXY and realized that the created `ProxyAgent` uses `proxyTunnel: true` by default which may fail if the proxy does not implement CONNECT. As a result, undici loops forever. 

## Reproducible By

See https://github.com/ArnaudBuchholz/training-fetch-proxy
run the proxy using `node reserve-proxy`
use `npm run native-fetch` to reproduce the issue.
use `NODE_DEBU=undici npm run native-fetch` to see the infinite loop

## Expected Behavior

Have a way to disable `proxyTunnel`

## Logs & Screenshots

```text
UNDICI 13766: connecting to www.google.com using http:undefined
UNDICI 13766: connecting to localhost:8989:8989 using http:undefined
UNDICI 13766: connected to localhost:8989:8989 using http:h1
UNDICI 13766: sending request to CONNECT http://localhost:8989www.google.com:80
UNDICI 13766: request to CONNECT http://localhost:8989www.google.com:80 errored - other side closed
UNDICI 13766: connection to www.google.com using http:undefined errored - other side closed
UNDICI 13766: connecting to www.google.com using http:undefined
UNDICI 13766: connecting to localhost:8989:8989 using http:undefined
UNDICI 13766: connected to localhost:8989:8989 using http:h1
UNDICI 13766: sending request to CONNECT http://localhost:8989www.google.com:80
UNDICI 13766: request to CONNECT http://localhost:8989www.google.com:80 errored - other side closed
UNDICI 13766: connection to www.google.com using http:undefined errored - other side closed
UNDICI 13766: connecting to www.google.com using http:undefined
UNDICI 13766: connecting to localhost:8989:8989 using http:undefined
UNDICI 13766: connected to localhost:8989:8989 using http:h1
UNDICI 13766: sending request to CONNECT http://localhost:8989www.google.com:80
UNDICI 13766: request to CONNECT http://localhost:8989www.google.com:80 errored - other side closed
UNDICI 13766: connection to www.google.com using http:undefined errored - other side closed
UNDICI 13766: connecting to www.google.com using http:undefined
UNDICI 13766: connecting to localhost:8989:8989 using http:undefined
UNDICI 13766: connected to localhost:8989:8989 using http:h1
```

## Environment

Mac OS / Tahoe 26.3.1
Node 24.14

mcollina:
## This relates to...

Fixes #5093

## Rationale

`EnvHttpProxyAgent` currently inherits `ProxyAgent`'s tunneling behavior. For plain HTTP targets reached through an HTTP proxy, defaulting to CONNECT can break proxies that do not implement tunneling and can lead to the looping behavior reported in #5093.

Per [RFC 9112 §3.2.2](https://www.rfc-editor.org/rfc/rfc9112.html#name-absolute-form), the tunnel-vs-forward decision depends only on the **request** protocol, not the proxy protocol. This matches Node's built-in `http.request` / `https.request`.

## Behavior

| Request | Proxy | Action |
|---------|-------|--------|
| HTTP    | HTTP  | Forward, absolute-form URI |
| HTTP    | HTTPS | TLS to proxy, then forward, absolute-form URI |
| HTTPS   | HTTP  | CONNECT tunnel, TLS to target through it |
| HTTPS   | HTTPS | TLS to proxy, CONNECT tunnel, TLS to target |

`proxyTunnel: true` still forces CONNECT for plain HTTP requests.

## Changes

- `shouldProxyTunnel` keys on the request protocol only
- `Http1ProxyWrapper` accepts a `proxyServername` to pin SNI to the proxy (otherwise the inner `Client` derives it from the rewritten Host header, which points at the target)
- Forward path through an HTTPS proxy wraps `ERR_TLS_CERT_ALTNAME_INVALID` as `SecureProxyConnectionError`, matching the tunneling path
- Regression coverage for `EnvHttpProxyAgent` using `http_proxy` (#5093) and for the new HTTPS-proxy-to-HTTP-target forwarding case
- Docs and TypeScript comment updated

### Bug Fixes

- fix default proxy behavior for plain HTTP over HTTP proxy connections
- fix HTTP-over-HTTPS-proxy to forward (absolute-form) instead of tunneling, per RFC 9112 §3.2.2
- fix the `EnvHttpProxyAgent` case behind #5093

HTTP request through an HTTPS proxy now forwards instead of tunneling by default. Matches Node's built-in behavior and the RFC. Pass `proxyTunnel: true` to restore the previous behavior.

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/dispatcher/proxy-agent.js
+++ b/lib/dispatcher/proxy-agent.js
@@ -145,6 +145,7 @@ class ProxyAgent extends DispatcherBase {
     }
 
     const connect = buildConnector({ timeout: connectTimeout, ...opts.proxyTls })
+    const connectHTTP1 = buildConnector({ timeout: connectTimeout, ...opts.proxyTls, allowH2: false })
     this[kConnectEndpoint] = buildConnector({ timeout: connectTimeout, ...opts.requestTls })
     this[kConnectEndpointHTTP1] = buildConnector({ timeout: connectTimeout, ...opts.requestTls, allowH2: false })
 
@@ -167,14 +168,14 @@ class ProxyAgent extends DispatcherBase {
 
       if (!shouldProxyTunnel(protocol, this[kTunnelProxy])) {
         const forwardConnect = this[kProxy].protocol === 'https:'
-          ? (opts, cb) => connect(opts, (err, socket) => {
+          ? (opts, cb) => connectHTTP1(opts, (err, socket) => {
               if (err && err.code === 'ERR_TLS_CERT_ALTNAME_INVALID') {
                 cb(new SecureProxyConnectionError(err))
               } else {
                 cb(err, socket)
               }
             })
-          : connect
+          : connectHTTP1
         return new Http1ProxyWrapper(this[kProxy].uri, {
           headers: this[kProxyHeaders],
           connect: forwardConnect,
--- a/lib/dispatcher/proxy-agent.js
+++ b/lib/dispatcher/proxy-agent.js
@@ -37,6 +37,10 @@ function defaultAgentFactory (origin, opts) {
   return new Pool(origin, opts)
 }
 
+function shouldProxyTunnel (proxyProtocol, requestProtocol, proxyTunnel) {
+  return proxyTunnel === true || proxyProtocol !== 'http:' || requestProtocol !== 'http:'
+}
+
 class Http1ProxyWrapper extends DispatcherBase {
   #client
 
@@ -105,7 +109,7 @@ class ProxyAgent extends DispatcherBase {
       throw new InvalidArgumentError('Proxy opts.clientFactory must be a function.')
     }
 
-    const { proxyTunnel = true, connectTimeout } = opts
+    const { proxyTunnel, connectTimeout } = opts
 
     super()
 
@@ -152,7 +156,7 @@ class ProxyAgent extends DispatcherBase {
         })
       }
 
-      if (!this[kTunnelProxy] && protocol === 'http:' && this[kProxy].protocol === 'http:') {
+      if (!shouldProxyTunnel(this[kProxy].protocol, protocol, this[kTunnelProxy])) {
         return new Http1ProxyWrapper(this[kProxy].uri, {
           headers: this[kProxyHeaders],
           connect,
--- a/lib/dispatcher/proxy-agent.js
+++ b/lib/dispatcher/proxy-agent.js
@@ -37,14 +37,15 @@ function defaultAgentFactory (origin, opts) {
   return new Pool(origin, opts)
 }
 
-function shouldProxyTunnel (proxyProtocol, requestProtocol, proxyTunnel) {
-  return proxyTunnel === true || proxyProtocol !== 'http:' || requestProtocol !== 'http:'
+function shouldProxyTunnel (requestProtocol, proxyTunnel) {
+  return proxyTunnel === true || requestProtocol !== 'http:'
 }
 
 class Http1ProxyWrapper extends DispatcherBase {
   #client
+  #proxyServername
 
-  constructor (proxyUrl, { headers = {}, connect, factory }) {
+  constructor (proxyUrl, { headers = {}, connect, factory, proxyServername }) {
     if (!proxyUrl) {
       throw new InvalidArgumentError('Proxy URL is mandatory')
     }
@@ -52,6 +53,7 @@ class Http1ProxyWrapper extends DispatcherBase {
     super()
 
     this[kProxyHeaders] = headers
+    this.#proxyServername = proxyServername
     if (factory) {
       this.#client = factory(proxyUrl, { connect })
     } else {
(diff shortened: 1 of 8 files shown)


trivikr:
Should `proxyTunnel` be set to true on line 76 in `test/env-http-proxy-agent-nodejs-bundle.js`?

```diff
-    setGlobalDispatcher(new EnvHttpProxyAgent())
+    setGlobalDispatcher(new EnvHttpProxyAgent({ proxyTunnel: true }))
```

metcoder95 (approved):

mvolz:
Thanks for submitting this! 

Do we have a sense of when this might make it into to a published branch? The reason I ask is that we're trying to replace request with fetch, but our outbound proxy rejects CONNECT requests for https so this is a blocker... trying to decide whether to revert all of it and go back to request we're in a deployable state, or if it's reasonable to wait for this change.

mcollina:
I investigated the remaining CI failures after the rebase.

All regular undici jobs are passing. The only failures left are the `--shared-builtin-undici/undici-path` jobs for Node 24 and 26.

What is happening there:

- this PR changes `fetch()` / `EnvHttpProxyAgent` so plain `http -> http proxy -> http target` traffic is **non-tunneled by default**
- in other words, the proxy now receives a regular proxied `GET http://host/path` request instead of a `CONNECT`
- the failing shared-builtin jobs run Node's own `test/client-proxy/*` suite against the externalized copy of undici
- those Node fetch tests still expect the old `CONNECT` behavior for the HTTP-over-HTTP-proxy case

I checked `../node` locally and Node's `http.request()` tests already expect the curl-style non-tunneled behavior for the same setup:

- `test/client-proxy/test-http-proxy-request.mjs`
- `test/client-proxy/test-http-set-global-proxy-from-env-http-request.mjs`
- `test/client-proxy/test-use-env-proxy-cli-http.mjs`

That last file is especially telling because it currently expects:

- `http.request()` => proxied `GET`
- `fetch()` => tunneled `CONNECT`

So the failing shared-builtin jobs are not showing an undici regression in the main test matrix; they are showing that Node's fetch-side proxy tests need to be updated to match the new behavior.

I think the follow-up should be a `nodejs/node` patch adjusting those `test/client-proxy` fetch expectations for plain HTTP over HTTP proxy connections.

lib/dispatcher/proxy-agent.js
```diff
@@ -37,6 +37,10 @@ function defaultAgentFactory (origin, opts) {
   return new Pool(origin, opts)
 }
 
+function shouldProxyTunnel (proxyProtocol, requestProtocol, proxyTunnel) {
+  return proxyTunnel === true || proxyProtocol !== 'http:' || requestProtocol !== 'http:'
```
joyeecheung: I think the rules mostly depend on request protocol, not the proxy protocol? In the case of "http request over https proxy", [9112 3.2.2](https://www.rfc-editor.org/rfc/rfc9112.html#name-absolute-form) says 

> When making a request to a proxy, other than a CONNECT or server-wide OPTIONS request (as detailed below), a client MUST send the target URI in "absolute-form" as the request-target.

(Basically the same as 7230 5.3.2) so e.g. a GET HTTP request over HTTPS proxy should be sent as

1. TLS handshake to proxy server
2. Sent the rewritten HTTP request directly to the proxy server

And according to [9110 9.3.6](https://datatracker.ietf.org/doc/html/rfc9110#name-connect)

>  Tunnels are commonly used to create an end-to-end virtual connection, through one or more proxies, which can then be secured using TLS (Transport Layer Security)

A HTTPS request over HTTP proxy should be sent as

1. Establish tunnel to proxy server (without TLS handshake to the proxy server)
2. TLS handshake to the target server through the tunnel
3. Sent the original HTTP request to the target server.

This is also what the builtin https/http.request implements.

(Note: I don't think socks5 count as a proxy defined in https://datatracker.ietf.org/doc/html/rfc9110, and socks5 is always effectively tunneled because the stuff in https://datatracker.ietf.org/doc/html/rfc1928 sits in a lower layer)

joyeecheung:
The failing Node.js tests have this comment exactly for what this is fixing:

> // FIXME(undici:4083): undici currently always tunnels the request over
> // CONNECT if proxyTunnel is not explicitly set to false.

(which just means the tests are only documenting what undici currently does, and needs to be updated once undici fixes it, so failures here is to be expected).

metcoder95 (approved):

AliMahmoudDev:
The `shouldProxyTunnel` helper is clean — defaulting to tunnel for HTTPS and forward for HTTP makes sense. The type annotation for `proxyTunnel` is helpful too.

One thing I noticed: the `#proxyServername` private field is stored in the constructor but I dont see it being used in the diff. Is it consumed in a later commit or part of a follow-up?

metcoder95 (approved):

mcollina:
@metcoder95 @trivikr one last approve before me landing this?

metcoder95 (approved):

PR #5118, opened 2026-04-25
fix: stop buffering data after SOCKS5 connect

### LINKED ISSUE #5117: SOCKS5 tunnel leaks tunneled response data into internal buffer
## Bug Description

`Socks5Client` keeps listening for all socket `data` events after the SOCKS5 CONNECT handshake succeeds.

## Reproducible By

```js
import { EventEmitter } from 'node:events'
import { Socks5Client, STATES } from 'undici/lib/core/socks5-client.js'

class MockSocket extends EventEmitter {
  write () {}
  destroy () {}
}

const socket = new MockSocket()
const client = new Socks5Client(socket)

// Simulate the state after SOCKS5 CONNECT succeeds.
client.state = STATES.CONNECTED

// This is tunneled application data, e.g. an HTTP response chunk.
socket.emit('data', Buffer.alloc(1024))

console.log('state:', client.state)
console.log('buffer length:', client.buffer.length)

```

## Expected Behavior

After the SOCKS5 CONNECT handshake has completed, tunneled application data should be handled only by the HTTP/TLS client using the socket.

`Socks5Client` should stop buffering socket data once it reaches `STATES.CONNECTED`, or otherwise avoid appending data in the `CONNECTED` state.

## Logs & Screenshots

```console
state: connected
buffer length: 1024
```

## Environment

macOS 26.4.1
Node v24.15.0
undici v8.1.0

trivikr:
## This relates to...

Fixes: https://github.com/nodejs/undici/issues/5117

## Rationale

After a SOCKS5 CONNECT succeeds, the socket carries tunneled HTTP/TLS data and should no longer be parsed by `Socks5Client`. The existing permanent `data` listener kept appending tunneled response chunks to the internal protocol buffer, causing duplicate retained memory for the lifetime of the socket.

## Changes

- Store the SOCKS5 socket `data` handler so it can be removed after CONNECT succeeds.
- Clear the SOCKS5 protocol buffer once the CONNECT response is parsed.
- Preserve any tunneled bytes that arrive in the same socket read as the CONNECT response by pushing them back with `socket.unshift()`.

### Features

N/A

### Bug Fixes

Stop buffering data after SOCKS5 connect

N/A

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

Assisted-by: openai:gpt-5.5

### DIFF
--- a/lib/core/socks5-client.js
+++ b/lib/core/socks5-client.js
@@ -7,6 +7,7 @@ const { debuglog } = require('node:util')
 const { parseAddress } = require('./socks5-utils')
 
 const debug = debuglog('undici:socks5')
+const EMPTY_BUFFER = Buffer.alloc(0)
 
 // SOCKS5 constants
 const SOCKS_VERSION = 0x05
@@ -73,7 +74,7 @@ class Socks5Client extends EventEmitter {
     this.socket = socket
     this.options = options
     this.state = STATES.INITIAL
-    this.buffer = Buffer.alloc(0)
+    this.buffer = EMPTY_BUFFER
     this.onSocketData = this.onData.bind(this)
     this.onSocketError = this.onError.bind(this)
     this.onSocketClose = this.onClose.bind(this)
@@ -376,7 +377,7 @@ class Socks5Client extends EventEmitter {
 
     const boundPort = this.buffer.readUInt16BE(offset)
 
-    this.buffer = Buffer.alloc(0)
+    this.buffer = EMPTY_BUFFER
     this.state = STATES.CONNECTED
     this.socket.removeListener('data', this.onSocketData)
 
--- a/lib/core/socks5-client.js
+++ b/lib/core/socks5-client.js
@@ -95,10 +95,6 @@ class Socks5Client extends EventEmitter {
    * Handle incoming data from the socket
    */
   onData (data) {
-    if (this.state === STATES.CONNECTED) {
-      return
-    }
-
     debug('received data', data.length, 'bytes in state', this.state)
     this.buffer = Buffer.concat([this.buffer, data])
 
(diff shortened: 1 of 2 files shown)


lib/core/socks5-client.js
```diff
@@ -83,15 +86,19 @@ class Socks5Client extends EventEmitter {
...
+    this.socket.on('close', this.onSocketClose)
   }
 
   /**
    * Handle incoming data from the socket
    */
   onData (data) {
+    if (this.state === STATES.CONNECTED) {
+      return
+    }
```
mcollina: Instead, remove the event listener

mcollina (changes requested):

lib/core/socks5-client.js
trivikr: The listener was already removed.
This was a defensive check, which is not needed.

It was removed in https://github.com/nodejs/undici/pull/5118/changes/9b14c923649501b869a72edf745ef77470649b27.

lib/core/socks5-client.js
```diff
@@ -373,8 +376,9 @@ class Socks5Client extends EventEmitter {
 
     const boundPort = this.buffer.readUInt16BE(offset)
 
-    this.buffer = this.buffer.subarray(responseLength)
+    this.buffer = Buffer.alloc(0)
```
mcollina: Does this need to be reallocated? We could just keep one around, they are just empty.

trivikr: The SOCKS5 client removes its `data` listener, but the client object can still be retained by the socket’s `error`/`close` listeners. If `this.buffer` is not cleared, it can retain the SOCKS5 response buffer.

We don't need fresh allocation every time, as it's replaced using `Buffer.concat(...)`
I replaced this with EMPTY_BUFFER constant in https://github.com/nodejs/undici/pull/5118/changes/f8a876f7943540b51bc3d81338c35e9219ed670b

mcollina (approved):
lgtm

PR #5121, opened 2026-04-26
`SnapshotAgent`: Add `normalizeBody` and `normalizeQuery`

GeoffreyBooth:
Adds two new options to `SnapshotAgent` and `SnapshotRecorder` for partial request matching:

- `normalizeBody(body)` — normalizes the request body before hashing, e.g. to strip volatile fields like timestamps from JSON payloads
- `normalizeQuery(params)` — normalizes query parameters (as `URLSearchParams`) before hashing, e.g. to strip cache-busting params

Both options pair with the existing `matchBody`/`matchQuery` boolean toggles and run at both record and playback time so hashes stay consistent across both sides.

## This relates to…

`SnapshotAgent`; https://github.com/nodejs/undici/issues/4114

## Rationale

When recording snapshots, sometimes we don’t want to match on the body or query params exactly; there might be volatile parts of the body or query params to exclude such as timestamps or cache-busters. In my case I want to match on the system prompt of an LLM call, ignoring the rest of the body.

## Changes

Adds `normalizeBody` and `normalizeQuery` options to `SnapshotAgent`, for users to define functions to simplify the recorded request body or query params before saving; and to run again when computing the hash for matching.

### Features

Adds `normalizeBody` and `normalizeQuery` options to `SnapshotAgent`.

### Bug Fixes

N/A

N/A

## Status

I tested this branch against my project and it works as intended.

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready


mcollina (commented):
Can you add a test for the types changes?

GeoffreyBooth:
> Can you add a test for the types changes?

Done

mcollina (approved):
lgtm

PR #5085, opened 2026-04-21
perf(http2): avoid response header reserialization

trivikr:
## This relates to...

N/A

## Rationale

HTTP/2 responses currently pass header objects through parts of the dispatcher stack, while the `responseHeaders: 'raw'` path still expects name/value arrays. That mismatch causes unnecessary header reserialization and makes the raw-header path inconsistent across H1 and H2 callers.

## Changes

- Stop reserializing HTTP/2 response and upgrade headers into buffer pairs before handing them to request handlers.
- Extend `parseRawHeaders()` to accept nullish values and plain header objects, and normalize them into the existing flat raw-header array format.
- Reuse that normalization in the request, stream, pipeline, connect, and upgrade APIs when `responseHeaders: 'raw'` is requested.
- Update fetch response header list construction so it works with either raw header arrays or object-form response headers.
- Expand types so `DispatchController.rawHeaders` / `rawTrailers` can be `IncomingHttpHeaders`.

### Features

N/A

### Bug Fixes

- Fix `responseHeaders: 'raw'` for HTTP/2 responses so callers receive normalized raw headers without relying on H1-style buffer pairs.
- Avoid redundant header conversion work on the HTTP/2 response path.
- Preserve fetch/WebSocket header list construction when raw headers arrive in object form.

### Breaking Changes and Deprecations

N/A

## Status

- [x] I have read and agreed to the [Developer's Certificate of Origin][cert]
- [x] Tested
- [ ] Benchmarked (**optional**)
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

[cert]: https://github.com/nodejs/undici/blob/main/CONTRIBUTING.md#developers-certificate-of-origin

PR #5122, opened 2026-04-26
perf: cache common h2 header name buffers

mcollina:
## Summary

Optimize the HTTP/2 response header path in `client-h2` by:

- avoiding the object rest/spread copy used to strip `:status`
- skipping pseudo-headers inline during H2 header parsing
- caching `Buffer` instances for a bounded set of common response header names

The header-name cache is intentionally bounded to avoid unbounded memory growth from attacker-controlled header names.

## Validation

- `npx eslint lib/dispatcher/client-h2.js`
- `npm run test:h2:core`

## Benchmark notes

I verified this with a controlled A/B `.request()` harness against `main` using the same simple H2 server and interleaved runs.

Average over 5 runs (`80 rounds x 200 parallel requests`):

- this branch: `2401.39ms`
- `main`: `2452.65ms`

That is roughly a **2.1% improvement** on this request-heavy H2 workload.

### DIFF
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -66,20 +66,69 @@ const {
   }
 } = http2
 
+const h2HeaderNameBufferCache = {
+  __proto__: null,
+  age: undefined,
+  'alt-svc': undefined,
+  'cache-control': undefined,
+  connection: undefined,
+  'content-disposition': undefined,
+  'content-encoding': undefined,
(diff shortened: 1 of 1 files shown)


trivikr:
There's an alternative PR which avoids response headers reserialization in https://github.com/nodejs/undici/pull/5085, which shows 5-7% improvement for similar tests.

ronag (approved):

PR #5123, opened 2026-04-26
perf: lazily materialize raw h2 headers

mcollina:
## Summary

This stacks on top of #5122.

It keeps the bounded common-header buffer cache, and adds a follow-up optimization for the HTTP/2 response path:

- pass parsed response headers as an object to the H2 request handlers
- defer raw header materialization until `controller.rawHeaders` is actually accessed
- update the API handlers to only read `controller.rawHeaders` when `responseHeaders === 'raw'`

The goal is to avoid paying the `parseH2Headers()` allocation cost on the hot path when callers only need the normal parsed header object.

## Validation

- `npx eslint lib/core/request.js lib/dispatcher/client-h2.js lib/api/api-request.js lib/api/api-stream.js lib/api/api-connect.js lib/api/api-upgrade.js lib/api/api-pipeline.js`
- `npm run test:h2:core`
- `npm run test:h2:fetch`

## Benchmark notes

I verified this against commit `63a1b0ff` from #5122 using the same controlled H2 `.request()` harness and interleaved runs.

Average over 5 runs (`80 rounds x 200 parallel requests`):

- `63a1b0ff`: `2409.08ms`
- this branch: `2396.64ms`

That is roughly a **0.5% improvement** on top of the previous optimization.

### DIFF
--- a/lib/api/api-connect.js
+++ b/lib/api/api-connect.js
@@ -56,12 +56,10 @@ class ConnectHandler extends AsyncResource {
     this.callback = null
 
     let responseHeaders = headers
-    const rawHeaders = controller?.rawHeaders
     // Indicates is an HTTP2Session
-    if (responseHeaders != null) {
-      responseHeaders = this.responseHeaders === 'raw'
-        ? (Array.isArray(rawHeaders) ? util.parseRawHeaders(rawHeaders) : [])
-        : headers
+    if (responseHeaders != null && this.responseHeaders === 'raw') {
+      const rawHeaders = controller?.rawHeaders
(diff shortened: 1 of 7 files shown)


trivikr:
There's an alternative PR which avoids response headers reserialization in https://github.com/nodejs/undici/pull/5085, which shows 5-7% improvement for similar tests.

metcoder95 (approved):

metcoder95:
Agree with the optimization of #5085 but I also consider that this improvement is worth it, as keeping the raw headers on the `controller` can enable other use cases around

PR #5127, opened 2026-04-27
perf(http2): avoid cloning headers when removing status

trivikr:
## This relates to...

N/A

## Rationale

Avoid cloning HTTP/2 response headers just to remove the `:status` pseudo-header before passing headers to request callbacks.

## Changes

The response handling now removes `:status` in place instead of using rest destructuring, avoiding an extra headers object allocation for every H2 response and upgrade response.

### Features

N/A

### Bug Fixes

N/A

N/A

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

Assisted-by: openai:gpt-5.5

### DIFF
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -673,9 +673,10 @@ function writeH2 (client, request) {
       const onUpgradeResponse = (headers, _flags) => {
         responseReceived = true
 
-        const { [HTTP2_HEADER_STATUS]: statusCode, ...realHeaders } = headers
+        const statusCode = headers[HTTP2_HEADER_STATUS]
+        delete headers[HTTP2_HEADER_STATUS]
 
-        request.onRequestUpgrade(statusCode, realHeaders, stream)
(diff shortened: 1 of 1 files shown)


lib/dispatcher/client-h2.js
```diff
@@ -894,7 +895,8 @@ function writeH2 (client, request) {
   const onResponse = (headers) => {
     stream.off('response', onResponse)
 
-    const { [HTTP2_HEADER_STATUS]: statusCode, ...realHeaders } = headers
+    const statusCode = headers[HTTP2_HEADER_STATUS]
+    delete headers[HTTP2_HEADER_STATUS]
```
metcoder95, line 899: This sits on the critical path, do we have potential perf regression?

trivikr: No. This is perf improvement since memory for object `realHeaders` is not allocated.

metcoder95 (approved):

PR #5139, opened 2026-04-30
chore(deps): add lockfile

aduh95:
Given the recent supply-chain attacks, not having a lockfile puts at risk Undici contributors, and I couldn't find a reason for not having one.

I've copied the lockfile from https://github.com/nodejs/node/blob/main/deps/undici/src/package-lock.json.

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -60,11 +60,6 @@ typings/
 # next.js build output
 .next
 
-# lock files
-package-lock.json
-yarn.lock
-pnpm-lock.yaml
-
 # IDE files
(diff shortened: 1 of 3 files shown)


metcoder95 (approved):
Given that we do not have dependencies but rather only dev ones, don't see a strong reason to not have one

mcollina (approved):
lgtm

PR #5146, opened 2026-05-01
perf: avoid redundant scans in BalancedPool dispatcher selection

trivikr:
## This relates to...

Performance cleanup in `BalancedPool` dispatcher selection.

## Rationale

`BalancedPool[kGetDispatcher]` previously scanned the client list multiple times per dispatch: once to find an available dispatcher, again to check whether all clients were busy, then again to find a non-draining fallback index before entering the weighted selection loop.

The busy check was redundant after an available dispatcher had already been found, and the fallback selection can be tracked during the weighted loop.

## Changes

Collapse dispatcher availability checks and fallback max-weight tracking into the existing weighted round-robin loop.

> This keeps per-dispatch complexity at `O(n)` but removes up to three extra client-array scans. The improvement should be measurable with a targeted `kGetDispatcher()` microbenchmark, especially as upstream count grows; it may be hard to see in network-heavy benchmarks.

### Features

N/A

### Bug Fixes

N/A

N/A

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

Assisted-by: openai:gpt-5.5

### DIFF
--- a/lib/dispatcher/balanced-pool.js
+++ b/lib/dispatcher/balanced-pool.js
@@ -164,32 +164,25 @@ class BalancedPool extends PoolBase {
       throw new BalancedPoolMissingUpstreamError()
     }
 
-    const dispatcher = this[kClients].find(dispatcher => (
-      !dispatcher[kNeedDrain] &&
-      dispatcher.closed !== true &&
-      dispatcher.destroyed !== true
-    ))
-
-    if (!dispatcher) {
-      return
-    }
-
-    const allClientsBusy = this[kClients].map(pool => pool[kNeedDrain]).reduce((a, b) => a && b, true)
-
-    if (allClientsBusy) {
-      return
-    }
-
     let counter = 0
 
-    let maxWeightIndex = this[kClients].findIndex(pool => !pool[kNeedDrain])
+    let maxWeightIndex = -1
 
     while (counter++ < this[kClients].length) {
       this[kIndex] = (this[kIndex] + 1) % this[kClients].length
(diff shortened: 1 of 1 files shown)


lib/dispatcher/balanced-pool.js
```diff
@@ -164,32 +164,25 @@ class BalancedPool extends PoolBase {
...
+        pool.closed !== true &&
+        pool.destroyed !== true
+      )
 
       // find pool index with the largest weight
-      if (pool[kWeight] > this[kClients][maxWeightIndex][kWeight] && !pool[kNeedDrain]) {
+      if (available && (
+        maxWeightIndex === -1 ||
+        pool[kWeight] > this[kClients][maxWeightIndex][kWeight] ||
+        (pool[kWeight] === this[kClients][maxWeightIndex][kWeight] && this[kIndex] < maxWeightIndex)
```
trivikr: This line is to preserve old fallback behavior.

```js
this[kClients].findIndex(pool => !pool[kNeedDrain])
```

If multiple available pools had the same highest weight, the fallback would keep the lowest array index.

It can be removed to improve readability, if it's not important.

trivikr: Actually, I removed it in https://github.com/nodejs/undici/pull/5146/changes/91b75bc33eee03dec1102cb764231d1d8288a563

mcollina (approved):
lgtm

metcoder95 (approved):

PR #5151, opened 2026-05-01
fix: preserve pool capacity after removing stale client

### LINKED ISSUE #5150: Pool.dispatch() can return false and report backpressure even though another client is available
## Bug Description

Pool.dispatch() can return false and report backpressure even though another client is available

## Reproducible By

```js
import { EventEmitter } from 'node:events'
import { setTimeout as sleep } from 'node:timers/promises'
import { Pool } from 'undici'

let n = 0

class Client extends EventEmitter {
  id = ++n
  closed = false
  destroyed = false

  dispatch () {
    if (this.id === 1) {
      this.emit('connect', new URL('http://example.com'), [this])
      return false
    }

    return true
  }

  close (cb) {
    cb?.()
  }
}

const handler = {
  onConnect () {},
  onHeaders () {},
  onData () {},
  onComplete () {},
  onError () {},
  onResponseError () {}
}

const pool = new Pool('http://example.com', {
  connections: 2,
  clientTtl: 1,
  factory: () => new Client()
})

// Creates client #1 which returns false, so pool marks it as draining.
// pool-level backpressure state is true. The pool calls kGetDispatcher() again and creates client #2.
console.log('first dispatch: ',  pool.dispatch({ path: '/', method: 'GET' }, handler))

// Wait for clientTtl to expire, after which client #1 is removed.
await sleep(5)

// The second dispatch goes to client #2, whose dispatch() returns true.
// pool-level backpressure state is false, though.
console.log('second dispatch: ',  pool.dispatch({ path: '/', method: 'GET' }, handler))
```

## Expected Behavior

`Pool.dispatch()` should return `true` when a remaining client is available and accepts the request.

Removing one stale client should not make the pool report backpressure while another non-draining, open client exists.

## Logs & Screenshots

```console
first dispatch:  true
second dispatch:  false
```

## Environment

macOS 26.4.1
Node v24.15.0
undici v8.1.0

trivikr:
## This relates to...

Fixes: https://github.com/nodejs/undici/issues/5150

## Rationale

`PoolBase[kRemoveClient]` was setting `kNeedDrain` to `true` when any remaining client was available.
That made `Pool.dispatch()` report backpressure even though another client could accept work.

## Changes

Fix `PoolBase[kRemoveClient]` so the pool only needs drain when no remaining client is available.

### Features

N/A

### Bug Fixes

Preserve pool capacity after removing stale client

N/A

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

Assisted-by: openai:gpt-5.5

### DIFF
--- a/lib/dispatcher/pool-base.js
+++ b/lib/dispatcher/pool-base.js
@@ -196,7 +196,7 @@ class PoolBase extends DispatcherBase {
 
     client.close(() => {})
 
-    this[kNeedDrain] = this[kClients].some(dispatcher => (
+    this[kNeedDrain] = !this[kClients].some(dispatcher => (
       !dispatcher[kNeedDrain] &&
       dispatcher.closed !== true &&
       dispatcher.destroyed !== true
--- a/test/pool.js
+++ b/test/pool.js
@@ -689,6 +689,64 @@ test('pool replaces stale client when connections limit is reached', async (t) =
   t.strictEqual(statusCode, 200)
 })
 
+test('pool does not report backpressure after removing a stale client when another client is available', async (t) => {
+  t = tspl(t, { plan: 4 })
+
+  let created = 0
+  let acceptedBySecondClient = 0
+
+  class FakeClient extends EventEmitter {
+    constructor () {
+      super()
+      this.id = ++created
+      this.closed = false
+      this.destroyed = false
+    }
+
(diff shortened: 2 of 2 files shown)


mcollina (approved):
lgtm

PR #5169, opened 2026-05-02
perf(http2): end zero-length request bodies with headers

trivikr:
## This relates to...

HTTP/2 request handling for known zero-length bodies.

## Rationale

When an HTTP/2 request body has a known `contentLength === 0`, the request payload is already complete before any DATA frame is written. Previously, only `body === null` caused the request stream to be created with `endStream: true`, so zero-length buffer/blob/stream bodies could still open the stream as if DATA may follow and then go through no-op body write handling.

Ending these requests with the initial HEADERS frame avoids that extra client-side request-body path and lets the peer immediately observe the request side as closed. The performance improvement is expected to be small per request, but useful for workloads that send many HTTP/2 requests with explicit empty bodies such as `Buffer.alloc(0)`, empty `Blob`s, or streams with `content-length: 0`.

## Changes

Treat HTTP/2 requests with `contentLength === 0` as `endStream: true`.

### Features

N/A

### Bug Fixes

N/A

N/A

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

Assisted-by: openai:gpt-5.5

### DIFF
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -870,7 +870,7 @@ function writeH2 (client, request) {
   }
 
   // TODO(metcoder95): add support for sending trailers
-  const shouldEndStream = body === null
+  const shouldEndStream = body === null || contentLength === 0
 
   if (expectContinue) {
     headers[HTTP2_HEADER_EXPECT] = '100-continue'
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

Issue #5177, opened 2026-05-04
[2026] Tracking issue for flaky tests


trivikr:
This is 2026 equivalent of https://github.com/nodejs/undici/issues/3787 where maintainers had reduced flaky tests.
We can use this issue for tracking, and create sub-issues per test failure?

trivikr:
We should probably switch resource teardown in after instead of finally in all tests

From review comments:
* https://github.com/nodejs/undici/pull/5168#discussion_r3183247458
* https://github.com/nodejs/undici/pull/5125#discussion_r3145425185

trivikr:
I'll stop tracking new flaky failures - especially those on windows - since there have been 50 sub-issues already and over 60% still needs fixing.

RaphaelFakhri:
I spent today trying to reproduce four of the sub-issues and could not make any of them fail. Posting the numbers here rather than four times, since the useful conclusion is about the set.

| Issue | Test | Attempts | Failures |
|---|---|---|---|
| #5259 | `test/pipeline-pipelining.js` | 428 (128 Linux, 180 Linux pinned to one core, 120 Windows) | 0 |
| #5263 | `test/mock-pool.js` | 3520 (360 Linux, 3160 Windows) | 0 |
| #5264 | `test/websocket/receive.js` | 320 direct plus 10 full `borp` suite runs | 0 |
| #5266 | `test/http2-connection.js` | 389 | 0 |

All on current `main` (fc3450dc), Node 24.16.0 on Linux and 24.11.1 on Windows, deliberately under load: busy-loop contention, single-core pinning, and in the #5266 case a box at load average 87 where each run dilated from 4.5s to about 36s wall, which is harsher than a CI runner. Failure detection on the Windows harness was verified against a deliberately failing file first, so those zeros are real zeros rather than a harness that never reports.

Two things that came out of it and might be worth more than the counts.

**The CI logs for these have expired.** Every one of the four cites a run from May, and the Actions logs now return 410, so the failing assertion was never visible to anyone working on them. Three of the four failure excerpts are a bare file-level `✖ ... 'test failed'` at `1:1` with no failing subtest named, which is the signature of a non-zero exit at end of run rather than an assertion, so the excerpt in the issue does not narrow it either. That argues for capturing the full failing output into the sub-issue at filing time, otherwise the issue outlives the only evidence.

[...]

PR #5194, opened 2026-05-05
test: move cleanup from finally to after hooks

trivikr:
Refs: https://github.com/nodejs/undici/issues/5177#issuecomment-4375731000


mcollina (approved):
lgtm

Issue #5199, opened 2026-05-05
[flaky] test/cache-interceptor/cache-tests.mjs


trivikr:
```console
> node test/cache-interceptor/cache-tests.mjs --ci

Testing 4 environments

PROTOCOL: http

TEST ENVIRONMENT
       BASE_URL: http://localhost:8000
          store: MemoryCacheStore
        methods: ["GET","HEAD"]
           type: shared
  ignored tests: ["conditional-etag-forward-unquoted","conditional-etag-strong-generate-unquoted","cc-resp-no-cache","cc-resp-no-cache-case-insensitive","304-etag-update-response-Cache-Control","304-etag-update-response-Content-Foo","304-etag-update-response-Test-Header","304-etag-update-response-X-Content-Foo","304-etag-update-response-X-Test-Header","age-parse-float","head-200-update","head-200-retain","head-410-update","stale-close-must-revalidate","stale-close-no-cache","freshness-max-age-s-maxage-private","freshness-max-age-s-maxage-private-multiple"]
Listening on http://[::]:8000/
SERVER WARNING: State not found for 61a81b88-d55e-4c93-949b-96a13600148e

TEST ENVIRONMENT
       BASE_URL: http://localhost:8001
          store: SqliteCacheStore
        methods: ["GET","HEAD"]
           type: shared
  ignored tests: ["conditional-etag-forward-unquoted","conditional-etag-strong-generate-unquoted","cc-resp-no-cache","cc-resp-no-cache-case-insensitive","304-etag-update-response-Cache-Control","304-etag-update-response-Content-Foo","304-etag-update-response-Test-Header","304-etag-update-response-X-Content-Foo","304-etag-update-response-X-Test-Header","age-parse-float","head-200-update","head-200-retain","head-410-update","stale-close-must-revalidate","stale-close-no-cache","freshness-max-age-s-maxage-private","freshness-max-age-s-maxage-private-multiple"]
Listening on http://[::]:8001/
SERVER WARNING: State not found for 164e4435-3136-4b63-b399-a800fbe800c4

        Total tests: 283
            Skipped: 0 (0.0%)
             Passed: 219 (77.4%)
             Failed: 0 (0.0%)
  Failed (optional): 59 (20.8%)
       Setup failed: 5 (1.8%)
Test Harness Failed: 0 (0.0%)
  Dependency Failed: 0 (0.0%)
            Retried: 0 (0.0%)

TEST ENVIRONMENT
       BASE_URL: http://localhost:8002
[...]

PR #5206, opened 2026-05-05
test: capture cache test worker stderr and preserve failures

trivikr:
Refs: https://github.com/nodejs/undici/issues/5199

Fixes the Undici cache test wrapper so worker failures are reported reliably.

- Capture cache test worker stderr with `stderr.on('data')` instead of listening for stream errors.
- Preserve any nonzero worker exit code across the environment matrix instead of overwriting it with later worker results.


mcollina (approved):
lgtm

PR #5231, opened 2026-05-06
fix(dispatcher): remove unreachable assert in writeBlob

### LINKED ISSUE #5153: Dead code in writeBlob: assert and RequestContentLengthMismatchError are both unreachable in practice
## Bug Description

In both `lib/dispatcher/client-h1.js` and `lib/dispatcher/client-h2.js`, the `writeBlob` function contains an `assert` and a `RequestContentLengthMismatchError` guard that are both dead code — neither can ever trigger under any reachable code path.
https://github.com/nodejs/undici/blob/bf684f7de01616708a33a5d1c092177622394442/lib/dispatcher/client-h1.js#L1385-L1390

The `contentLength` parameter passed to `writeBlob` is computed in `writeH1`/`writeH2` as follows:
https://github.com/nodejs/undici/blob/bf684f7de01616708a33a5d1c092177622394442/lib/dispatcher/client-h1.js#L1053-L1055

Since `util.bodyLength` returns `body.size` for blob-like bodies (never `null`), the local `contentLength` variable is **always overwritten** to `body.size` before `writeBlob` is called. This means:

1. The `assert(contentLength === body.size)` is always true — it can never fire.
2. The `if (contentLength != null && contentLength !== body.size)` is always false — `RequestContentLengthMismatchError` can never be thrown from this path.

Meanwhile, if a user provides a mismatched `content-length` header, it is caught **earlier** in `writeH1`/`writeH2` by the `shouldSendContentLength` check (which compares `request.contentLength` against the computed `contentLength`), so the mismatch never propagates to `writeBlob`.

The result is that blob bodies have no graceful `RequestContentLengthMismatchError` path inside `writeBlob` — the error handling code exists but is unreachable.

## Reproducible By

This script demonstrates that the mismatch is caught before `writeBlob`, not inside it:

```js

  req.resume();
  req.on("end", () => res.end("ok"));
});

server.listen(0);

const client = new Client(`http://localhost:${server.address().port}`);
const blob = new Blob(["hello"], { type: "text/plain" }); // size = 5

try {
  await client.request({
    path: "/",
    method: "POST",
    headers: {
      "content-length": 999, // intentionally wrong
    },
    body: blob,
  });
} catch (err) {
  // Error is thrown from writeH1's early check, NOT from writeBlob
  console.log("Caught error:", err.constructor.name, err.message);
} finally {
  await client.close();
  server.close();
}
```

## Expected Behavior

The dead code in `writeBlob` should be cleaned up. Either:

1. **Remove both the assert and the `if` guard** — they are redundant since the mismatch is already handled upstream in `writeH1`/`writeH2`.
2. **Remove the assert and keep the `if` guard** as a defensive safety net, in case future refactors change how `contentLength` is computed before reaching `writeBlob`.

## Logs & Screenshots

```console
Caught error: RequestContentLengthMismatchError Request body length does not match content-length header
```

## Environment

macOS 26.4.1
Node v24.15.0
undici v8.1.0

SAY-5:
## This relates to...

Closes #5153

## Rationale

The `assert(contentLength === body.size, ...)` at the top of `writeBlob` (in both `client-h1.js` and `client-h2.js`) can never fire: `writeH1`/`writeH2` always overwrite `contentLength` with `body.size` via `util.bodyLength` before invoking `writeBlob` for blob-like bodies. The follow-up `if (contentLength != null && contentLength !== body.size)` guard is also unreachable today, but is left in place as a defensive net for future refactors of the upstream `contentLength` computation, matching option 2 in the issue.

### Features

N/A

### Bug Fixes

- Drop the unreachable `assert(contentLength === body.size, ...)` from `writeBlob` in `lib/dispatcher/client-h1.js` and `lib/dispatcher/client-h2.js`.

N/A

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/dispatcher/client-h1.js
+++ b/lib/dispatcher/client-h1.js
@@ -1382,8 +1382,6 @@ function writeBuffer (abort, body, client, request, socket, contentLength, heade
  * @returns {Promise<void>}
  */
 async function writeBlob (abort, body, client, request, socket, contentLength, header, expectsPayload) {
-  assert(contentLength === body.size, 'blob body must have content length')
-
   try {
     if (contentLength != null && contentLength !== body.size) {
       throw new RequestContentLengthMismatchError()
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -1156,8 +1156,6 @@ function writeStream (abort, socket, expectsPayload, h2stream, body, client, req
 }
 
 async function writeBlob (abort, h2stream, body, client, request, socket, contentLength, expectsPayload) {
-  assert(contentLength === body.size, 'blob body must have content length')
-
   try {
     if (contentLength != null && contentLength !== body.size) {
       throw new RequestContentLengthMismatchError()


mcollina (approved):
lgtm

PR #5262, opened 2026-05-07
fix(mock): preserve fetch FormData boundary on net-connect fallback

### LINKED ISSUE #5241: Incorrect `multipart/form-data` boundary in header/body with `MockAgent`
## Bug Description

The boundary specified in a request made when a MockAgent is in use is not the same value that's used within the request body.

## Reproducible By

Run `nc -l 9999` and while it's listening, run `node issue.ts` or `USE_MOCK_AGENT=1 node issue.ts`.

<details>

<summary><code><b>issue.ts</b></code></summary>

```js
import {
  install as installUndici,
  MockAgent,
  setGlobalDispatcher,
} from 'undici'

installUndici()

if (process.env.USE_MOCK_AGENT) {
  const agent = new MockAgent()
  agent.enableNetConnect('localhost:9999')
  setGlobalDispatcher(agent)
}

const body = new FormData()
const buffer = Buffer.from('hello world', 'utf8')
const blob = new Blob([buffer], { type: 'txt' })
body.append('file', blob, 'document.txt')

const res = await fetch('http://localhost:9999', {
  body,
  method: 'POST',
})
console.log(await res.json())
```

</details>

Request sent **without** using the mock agent:

```sh
$ nc -l 9999
POST / HTTP/1.1
host: localhost:9999
connection: keep-alive
content-type: multipart/form-data; boundary=----formdata-undici-033603461557
accept: */*
accept-language: *
sec-fetch-mode: cors
user-agent: undici
accept-encoding: gzip, deflate
content-length: 178

------formdata-undici-033603461557
Content-Disposition: form-data; name="file"; filename="document.txt"
Content-Type: txt

hello world
------formdata-undici-033603461557--
```

Request sent **using the mock agent**:

```sh
$ nc -l 9999
POST / HTTP/1.1
host: localhost:9999
connection: keep-alive
content-type: multipart/form-data; boundary=----formdata-undici-012602790914
accept: */*
accept-language: *
sec-fetch-mode: cors
user-agent: undici
accept-encoding: gzip, deflate
content-length: 178

------formdata-undici-099234737323
Content-Disposition: form-data; name="file"; filename="document.txt"
Content-Type: txt

hello world
------formdata-undici-099234737323--
```

## Expected Behavior

The value in the `Content-Type` header includes a boundary that's used within the request body even when the `MockAgent` is used.

## Environment

```
$ node --version
v22.22.0

undici@8.2.0
```

SAY-5:
## This relates to...

Fixes #5241

## Rationale

When fetch sends a `FormData` body through a `MockAgent` that falls through to net connect, mock body matching wants the original request source, but the real dispatch path must keep fetch's already-encoded body so the multipart boundary stays aligned with the `content-type` header.

## Changes

- Keep the matcher-friendly `request.body.source || request.body.stream` path for MockAgent requests.
- Carry fetch's pre-encoded request body through an internal mock-only fallback slot.
- Restore that prepared body only when a `FormData` request misses the mock intercept and passes through to the original dispatcher.

### Features

N/A

### Bug Fixes

- fetch + MockAgent + FormData no longer produces a content-type/body boundary mismatch on net-connect pass-through.

N/A

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/web/fetch/index.js
+++ b/lib/web/fetch/index.js
@@ -61,6 +61,7 @@ const { Readable, pipeline, finished, isErrored, isReadable } = require('node:st
 const { addAbortListener, bufferToLowerCasedHeaderName } = require('../../core/util')
 const { dataURLProcessor, serializeAMimeType, minimizeSupportedMimeType } = require('./data-url')
 const { getGlobalDispatcher } = require('../../global')
+const { kMockDispatchFallbackBody } = require('../../mock/mock-symbols')
 const { webidl } = require('../webidl')
 const { STATUS_CODES } = require('node:http')
 const { bytesMatch } = require('../subresource-integrity/subresource-integrity')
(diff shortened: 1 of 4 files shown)


lib/web/fetch/index.js
```diff
@@ -2178,12 +2178,22 @@ async function httpNetworkFetch (
     return dispatchWithProtocolPreference(body)
 
     function dispatchWithProtocolPreference (body, allowH2) {
+      // When dispatching to a MockAgent, prefer request.body.source so that
```
metcoder95: I'm under the impression this should be fix at the `MockAgent` rather than fetch itself

PR #5273, opened 2026-05-09
fix: validate EOF for chunked h1 responses

mcollina:
## Summary
- validate HTTP/1.1 response completeness with `llhttp_finish()` on EOF-like socket termination
- preserve the existing `Content-Length` mismatch error behavior
- add regression coverage for truncated chunked responses in `Client` and `fetch()`

## Details
Undici previously treated EOF on non-keepalive responses as an implicit successful message completion. That is correct for EOF-delimited responses, but not for chunked responses, which must receive the terminating `0\r\n\r\n` chunk.

This change routes the EOF/close/ECONNRESET path through `llhttp_finish()` so llhttp can reject incomplete chunked responses with an invalid EOF state instead of completing the response body successfully.

## Testing
- `node --test test/parser-issues.js test/content-length.js`
- `npx eslint lib/dispatcher/client-h1.js test/parser-issues.js`

### DIFF
--- a/lib/dispatcher/client-h1.js
+++ b/lib/dispatcher/client-h1.js
@@ -360,21 +360,60 @@ class Parser {
           this.paused = true
           socket.unshift(data)
         } else {
-          const ptr = llhttp.llhttp_get_error_reason(this.ptr)
-          let message = ''
-          if (ptr) {
-            const len = new Uint8Array(llhttp.memory.buffer, ptr).indexOf(0)
-            message =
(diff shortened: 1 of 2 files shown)


trivikr:
Needs rebase to include node 26 FFI build from https://www,github.com/nodejs/undici/pull/5275

mcollina:
I think we can land without

trivikr (approved):

PR #5291, opened 2026-05-11
fix(fetch): keep cloned request signals alive

### LINKED ISSUE #4068: Cloned requests loose abort signal on GC

raashish1601:
## Summary
- Keep the abort-controller chain strongly reachable when cloning a Request or constructing a Request from another Request.
- Preserve abort propagation after garbage collection for cloned/copied requests.
- Add GC regression coverage for both clone() and constructor-copy paths.

Fixes #4068.

## Validation
- `node --expose-gc test\fetch\request.js`
- `npx borp --timeout 180000 --expose-gc -p "test/fetch/request.js"`
- `npm run lint -- lib/web/fetch/request.js test/fetch/request.js`

Note: the normal git pre-commit hook runs repo-wide lint and is currently blocked by unrelated baseline `no-undef` errors for `after` in `test/parser-issues.js`; changed-file lint passes.

### DIFF
--- a/lib/web/fetch/request.js
+++ b/lib/web/fetch/request.js
@@ -122,6 +122,7 @@ class Request {
 
     // 4. Let signal be null.
     let signal = null
+    let signalController = null
 
     // 5. If input is a string, then:
     if (typeof input === 'string') {
@@ -160,6 +161,7 @@ class Request {
 
       // 9. Set signal to input’s signal.
       signal = input.#signal
+      signalController = input[kAbortController]
 
       this.#dispatcher = init.dispatcher || input.#dispatcher
     }
@@ -405,6 +407,7 @@ class Request {
     // 26. If init["signal"] exists, then set signal to it.
     if (init.signal !== undefined) {
       signal = init.signal
+      signalController = null
     }
 
     // 27. Set this’s request to request.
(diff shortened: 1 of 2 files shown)


KhafraDev:
to be honest I don't care if the code looks okay, if you don't spend time reviewing the changes, I don't bother reviewing them either

raashish1601:
> to be honest I don't care if the code looks okay, if you don't spend time reviewing the changes, I don't bother reviewing them either

fuck u then, go to your mom's ass hole you piece of bullshit!!

PR #5318, opened 2026-05-21
fix(fetch): remove abort listener when request settles

### LINKED ISSUE #5285: MaxListenersExceededWarning when using signal
## Bug Description

When a signal is passed to `fecth`, an `abort` event is listened for.

However, when requesting a result, there is no overflow listener for the `abort` event. This means that if the same signal is used for multiple requests, there is a risk of signal leakage, and Node.js continuously outputs warnings.

```
(node:14604) MaxListenersExceededWarning: Possible EventTarget memory leak detected. 9053 abort listeners added to [AbortSignal]. Use events.setMaxListeners() to increase limit
```

## Reproducible By

```js
        const response = await fetch(requestUrl, {
            method,
            headers: requestHeaders,
            dispatcher: downloadAgent,
            signal
        })
```

## Expected Behavior

The abort event listener should be canceled when the request ends.

ATOM00blue:
## Problem
`fetch()` attaches an `abort` listener to the passed `AbortSignal` on every call — in the `Request` constructor and in the fetch algorithm — but only removes them via a `FinalizationRegistry` (on GC). Reusing one signal across many requests accumulates listeners and Node emits `MaxListenersExceededWarning`.

## Fix
Capture the listener-removal callbacks and invoke them deterministically once the fetch settles, covering the end-of-body, network-error and abort paths. The `Request` constructor exposes its cleanup through an internal static accessor following the existing pattern in `request.js`, so no new public symbol is introduced.

## Test
Added `test/fetch/issue-5285.js`: issues 100 `fetch` calls sharing one `AbortController` signal and asserts no `abort` listeners remain and no `MaxListenersExceededWarning` is emitted. Fails on `main`, passes here. Full `test/fetch` suite (471 tests) and node-fetch suite pass.

Closes #5285

### DIFF
--- a/lib/web/fetch/index.js
+++ b/lib/web/fetch/index.js
@@ -11,7 +11,7 @@ const {
   getResponseState
 } = require('./response')
 const { HeadersList } = require('./headers')
-const { Request, cloneRequest, getRequestDispatcher, getRequestState } = require('./request')
+const { Request, cloneRequest, getRequestDispatcher, getRequestState, removeRequestAbortListener } = require('./request')
 const zlib = require('node:zlib')
 const {
   makePolicyContainer,
@@ -208,7 +208,7 @@ function fetch (input, init = undefined) {
   let controller = null
 
   // 11. Add the following abort steps to requestObject’s signal:
-  addAbortListener(
+  const removeAbortListener = addAbortListener(
     requestObject.signal,
     () => {
       // 1. Set locallyAborted to true.
@@ -227,6 +227,15 @@ function fetch (input, init = undefined) {
       abortFetch(p, request, realResponse, requestObject.signal.reason, controller.controller)
     }
   )
+
+  // Remove the `abort` listeners registered above and in the Request
+  // constructor once the fetch has settled. Without this, reusing a single
+  // signal across many requests leaks listeners and Node.js emits a
+  // MaxListenersExceededWarning. See https://github.com/nodejs/undici/issues/5285
+  const cleanupAbortListeners = () => {
+    removeAbortListener()
+    removeRequestAbortListener(requestObject)
+  }
 
   // 12. Let handleFetchDone given response response be to finalize and
   // report timing with response, globalObject, and "fetch".
@@ -252,6 +261,7 @@ function fetch (input, init = undefined) {
       //    deserializedError.
 
       abortFetch(p, request, responseObject, controller.serializedAbortReason, controller.controller)
+      cleanupAbortListeners()
       return
     }
 
@@ -259,6 +269,7 @@ function fetch (input, init = undefined) {
     // and terminate these substeps.
     if (response.type === 'error') {
       p.reject(new TypeError('fetch failed', { cause: response.error }))
+      cleanupAbortListeners()
       return
     }
 
@@ -273,7 +284,10 @@ function fetch (input, init = undefined) {
 
   controller = fetching({
     request,
-    processResponseEndOfBody: handleFetchDone,
+    processResponseEndOfBody: (response) => {
+      handleFetchDone(response)
+      cleanupAbortListeners()
+    },
     processResponse,
     dispatcher: getRequestDispatcher(requestObject), // undici
     // Keep requestObject alive to prevent its AbortController from being GC'd
--- a/test/fetch/issue-5285.js
+++ b/test/fetch/issue-5285.js
@@ -1,0 +1,51 @@
+'use strict'
+
+const { test } = require('node:test')
+const assert = require('node:assert')
+const http = require('node:http')
+const { once, getEventListeners } = require('node:events')
+const { fetch } = require('../..')
+const { closeServerAsPromise } = require('../utils/node-http')
+
+// https://github.com/nodejs/undici/issues/5285
+// Reusing a single AbortSignal across many fetch() calls must not leak
+// `abort` listeners on the signal, which previously caused Node.js to emit
+// a MaxListenersExceededWarning.
+test('fetch removes the abort listener once the request settles', async (t) => {
+  const server = http.createServer({ joinDuplicateHeaders: true }, (req, res) => {
+    res.writeHead(200, { 'content-type': 'text/plain' })
+    res.end('hello')
+  })
+
+  t.after(closeServerAsPromise(server))
+  await once(server.listen(0), 'listening')
+
+  let warning = null
+  function onWarning (value) {
+    warning = value
+  }
+  process.on('warning', onWarning)
+  t.after(() => process.off('warning', onWarning))
+
+  const controller = new AbortController()
+  const { signal } = controller
+
+  const url = `http://localhost:${server.address().port}`
+
+  // Issue many more requests than the default max listeners (10) while
+  // sharing the same signal. Each settled request must remove its listener,
+  // otherwise a MaxListenersExceededWarning is emitted and the listeners leak.
+  for (let i = 0; i < 100; i++) {
+    const res = await fetch(url, { signal })
+    await res.text()
+  }
+
+  // Allow the trailing end-of-body cleanup of the final request, which is
+  // scheduled in a microtask, to run before asserting.
+  await new Promise((resolve) => setTimeout(resolve, 100))
+
+  // No `abort` listeners should remain registered on the signal once every
+  // request has settled.
+  assert.strictEqual(getEventListeners(signal, 'abort').length, 0)
+  assert.strictEqual(warning, null)
+})
--- a/lib/web/fetch/request.js
+++ b/lib/web/fetch/request.js
@@ -97,6 +97,13 @@ class Request {
 
   #state
 
+  /**
+   * Removes the `abort` listener that makes this request's signal follow the
+   * passed signal. `null` when no such listener was registered.
+   * @type {(() => void) | null}
+   */
+  #abortCleanup = null
+
   // https://fetch.spec.whatwg.org/#dom-request
   constructor (input, init = undefined) {
     webidl.util.markAsUncloneable(this)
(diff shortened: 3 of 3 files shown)


lib/web/fetch/index.js
```diff
@@ -252,13 +261,15 @@ function fetch (input, init = undefined) {
       //    deserializedError.
 
       abortFetch(p, request, responseObject, controller.serializedAbortReason, controller.controller)
+      cleanupAbortListeners()
```
tsctx, line 264: Please move this function call to `abortFetch`.

test/fetch/issue-5285.js
```diff
@@ -0,0 +1,51 @@
...
+  // sharing the same signal. Each settled request must remove its listener,
+  // otherwise a MaxListenersExceededWarning is emitted and the listeners leak.
+  for (let i = 0; i < 100; i++) {
+    const res = await fetch(url, { signal })
+    await res.text()
+  }
+
+  // Allow the trailing end-of-body cleanup of the final request, which is
+  // scheduled in a microtask, to run before asserting.
+  await new Promise((resolve) => setTimeout(resolve, 100))
```
tsctx, line 45:
```js
  const { setImmediate } = require('node:timers/promises')
```

```suggestion
  await setImmediate()
```

test/fetch/issue-5285.js
```diff
@@ -0,0 +1,51 @@
...
+  const { signal } = controller
+
+  const url = `http://localhost:${server.address().port}`
+
+  // Issue many more requests than the default max listeners (10) while
+  // sharing the same signal. Each settled request must remove its listener,
+  // otherwise a MaxListenersExceededWarning is emitted and the listeners leak.
+  for (let i = 0; i < 100; i++) {
+    const res = await fetch(url, { signal })
+    await res.text()
```
tsctx, line 40:
```suggestion
    await res.arrayBuffer()
```

lib/web/fetch/index.js
```diff
@@ -273,7 +284,10 @@ function fetch (input, init = undefined) {
 
   controller = fetching({
     request,
-    processResponseEndOfBody: handleFetchDone,
+    processResponseEndOfBody: (response) => {
+      handleFetchDone(response)
+      cleanupAbortListeners()
+    },
```
tsctx, line 290: There might be some unexpected behavior when cloning the response, so we need to investigate.

mcollina (approved):
I'm surprised this does not cause regressions, but in hindsight it should be ok

lgtm but we need to wait @KhafraDev opinion too.

KhafraDev:
> I'm surprised this does not cause regressions, but in hindsight it should be ok

I agree. A possible issue is that `abort` from `onRequestStart` won't be called anymore if an AbortController is aborted after a request, but I'm not sure if it has any impact. We have decent test coverage for signals, GC, and cloning requests, and no WPTs have regressed either.

If this has no consequences, I think this should be upstreamed to the fetch spec. At the bare minimum, a note stating "the abort listener added can be removed here" could be added to the call sites.

lib/web/fetch/index.js
KhafraDev: all of our tests pass

test/fetch/issue-5285.js
KhafraDev: why?

tsctx: This is just a personal preference. It avoids string allocation, which makes the tests faster.

tsctx:
At first, I didn’t think this would work, but after thinking it through more carefully, it actually seems to work quite well.

That said, since this behavior is outside the specification, we should proceed with caution. I found one edge case that is worth considering. In this scenario, determining when it is safe to detach the signal is very similar to how garbage collection works.

Specifically, the link can only be removed when:

- the fetch fails or is aborted; or
- all responses, including cloned responses, have been fully consumed, or have lost all references and been garbage-collected. This should be equivalent to the fetch controller associated with the fetch being fully disposed of.

As far as I can tell, only browsers handle this edge case correctly today; it has not yet been implemented in server-side fetch. This is not an immediate problem since it is currently unimplemented, but it will likely become an issue if we decide to implement it later.

```js
const c = new AbortController();

const r = await fetch(
  new URL("/", globalThis.location?.href ?? "https://example.com/"),
  { signal: c.signal },
);

const r2 = r.clone(); // Clone it.

await r.arrayBuffer(); // Consume the entire body first to confirm the request has ended.

c.abort(); // Abort.

console.log((await r2.text()).slice(0, 100)); // This should throw an error because they are shared.
```

If this is a known limitation and we are comfortable with it for now, I’m fine with moving forward. However, it is something we will need to take into account if and when we implement this in the future.

KhafraDev:
> If this is a known limitation and we are comfortable with it for now, I’m fine with moving forward. However, it is something we will need to take into account if and when we implement this in the future.

I can't find anything in the spec that says this case should throw. Browsers seem to be mishandling it.

KhafraDev (dismissed):

KhafraDev (approved):

tsctx:
> I can't find anything in the spec that says this case should throw. Browsers seem to be mishandling it.

It seems like the browser is correct.
It makes sense if you think about it this way: the current implementation relies on a shared underlying fetch's body stream, which correctly aborts and works as long as the entire body isn't consumed first. This same logic can be applied to cases where the entire body is fully consumed.

```js
const c = new AbortController();

const r = await fetch(
  new URL("/", globalThis.location?.href ?? "https://example.com/"),
  { signal: c.signal },
);

const r2 = r.clone(); // Clone it.

c.abort(); // Abort.

console.log((await r2.text()).slice(0, 100)); // This throw an error because they are shared.
```

KhafraDev:
> It seems like the browser is correct.

Based on what?

Once the body is consumed, the stream is no longer readable and therefore is never canceled/errored, which propagates to the teed stream(s) (seemingly, I could be wrong as I'm not an expert in webstreams). In the 2nd example, since the first response's body's stream is still readable when the signal is aborted, it errors.

Spec:
[#abort-fetch](https://fetch.spec.whatwg.org/#abort-fetch):
> 2. If request’s [body](https://fetch.spec.whatwg.org/#concept-request-body) is non-null and is [readable](https://streams.spec.whatwg.org/#readablestream-readable), then [cancel](https://streams.spec.whatwg.org/#readablestream-cancel) request’s [body](https://fetch.spec.whatwg.org/#concept-request-body) with error.
> 5. If response’s [body](https://fetch.spec.whatwg.org/#concept-response-body) is non-null and is [readable](https://streams.spec.whatwg.org/#readablestream-readable), then [error](https://streams.spec.whatwg.org/#readablestream-error) response’s [body](https://fetch.spec.whatwg.org/#concept-response-body) with error.

[#concept-http-network-fetch](https://fetch.spec.whatwg.org/#concept-http-network-fetch):
> 18.2.1.2. If stream is [readable](https://streams.spec.whatwg.org/#readablestream-readable), then...
> 18.2.2. Otherwise, if stream is [readable](https://streams.spec.whatwg.org/#readablestream-readable),...

tsctx:
Good question. I agree that this is not `Response.clone()` inheriting or copying an `AbortSignal`; `Response` has no `signal` property.

The point is the body stream. `Response.clone()` clones the body by teeing its stream, so the cloned body is a branch derived from the same body stream, not an independent fetch. The Fetch Standard defines body cloning in terms of teeing the body stream, and Streams defines tee branches as observing errors from the original stream.

Your reading of `abort fetch` is reasonable if only the original `r.body` branch is considered the response body. After `await r.text()`, that branch is closed, so the “if body is readable” condition would not error it. But I think the subtle point is what counts as the relevant fetch-backed body after cloning.

Browsers appear to preserve the fetch-backed body relationship across the tee branches created by `Response.clone()`. So the unread cloned branch is not treated as a detached stream. Aborting the request can still error that branch, not because the `Response` inherited a signal, but because the cloned body remains part of the fetch-created body stream machinery.

So the browser behavior is better described as preserving the fetch-abort/error linkage across cloned body branches, not as `Response`-level signal inheritance.

KhafraDev:
Browsers are actually handling this correctly, but for different reasons - we aren't teeing the streams 'properly', and I'm not sure we are able to without access to ReadableStreamTee.

[#rs-tee (webstreams)](https://streams.spec.whatwg.org/#rs-tee) (this is how ReadableStream.prototype.tee is implemented):
> Return ? [ReadableStreamTee](https://streams.spec.whatwg.org/#readable-stream-tee)([this](https://webidl.spec.whatwg.org/#this), false).

[#concept-body-clone (fetch)](https://fetch.spec.whatwg.org/#concept-body-clone):
> Let « out1, out2 » be the result of [teeing](https://streams.spec.whatwg.org/#readablestream-tee) body’s [stream](https://fetch.spec.whatwg.org/#concept-body-stream).

"teeing" links to [#readablestream-tee (webstreams) (how .clone() tees the streams)](https://streams.spec.whatwg.org/#readablestream-tee):
> return ? [ReadableStreamTee](https://streams.spec.whatwg.org/#readable-stream-tee)(stream, true).

... which has the note:

> Because we pass true as the second argument to [ReadableStreamTee](https://streams.spec.whatwg.org/#readable-stream-tee), the second branch returned will have its [chunks](https://streams.spec.whatwg.org/#chunk) cloned ... from those of the first branch. **This prevents consumption of one of the branches from interfering with the other.**

So there is a bug in undici, but this PR is not responsible for it, and will not cause any issues in implementing it _if_ we ever have access to ReadableStreamTee from node core.

tsctx:
Thanks for clarifying — that makes sense. I agree this should be framed as a Fetch body-clone issue, not as `Response.clone()` inheriting an `AbortSignal`.

Fetch body cloning uses the Streams “teeing” operation corresponding to `ReadableStreamTee(stream, true)`, while `ReadableStream.prototype.tee()` uses `ReadableStreamTee(this, false)`. So plain `stream.tee()` cannot fully model Fetch’s body-clone semantics.

Given that, browsers are handling this correctly. This PR does not introduce the bug; it exposes an existing limitation that should be tracked separately until an equivalent of `ReadableStreamTee(stream, true)` is available.

tsctx (approved):

PR #5321, opened 2026-05-25
docs: add EventSource server example

### LINKED ISSUE #536: How to handle text/event-stream?

Will-thom:
## Summary
- add a self-contained EventSource server/client example to the API docs
- clarify named events vs default message events
- fix minor wording in the EventSource docs

Closes #536

## Testing
- git diff --check


mcollina (approved):
lgtm

metcoder95 (approved):

PR #5326, opened 2026-05-26
Improve fetch TLS certificate error messages

### LINKED ISSUE #2199: Unprecise error message on fetch() to an endpoint with a certificate which was signed by an "unkown" CA
## Bug Description

There is a unprecise error message ("fetch failed") on fetch() to an endpoint with a certificate which was signed by an "unkown" CA.
If its an custom or internal CA, the fetch() will fail, that seems correct, but the error message ist absolut unprecise. It is therefore very difficult to find the cause.

## Reproducible By

fetch() to an enpoint with a certificate which was signed by an "unkown" CA.

An "unkown" CA seems to be an CA which not exist in the default search path for CAs of node.
By default node seem to use CA-certificates baked in the build.	

## Expected Behavior

A useful error message is shown.

## Logs & Screenshots

TypeError: fetch failed
    at Object.fetch (node:internal/deps/undici/undici:11576:11)
...

## Environment

Ubuntu 22.04
Node v18.17.0

Will-thom:
Fixes #2199.

This keeps the fetch TypeError shape and cause intact while surfacing TLS certificate verification details in the top-level error message for known certificate-related failures.

Tests:
- node --test test\fetch\tls-certificate-error.js
- node --test test\fetch\client-error-stack-trace.js
- borp focused test/fetch/tls-certificate-error.js
- borp focused test/fetch/client-error-stack-trace.js
- eslint lib/web/fetch/index.js test/fetch/tls-certificate-error.js

Note: npm run build:node could not be executed in this Windows sandbox because npm/esbuild attempted to access a denied path outside the workspace before compiling.

### DIFF
--- a/lib/web/fetch/index.js
+++ b/lib/web/fetch/index.js
@@ -68,6 +68,15 @@ const { isomorphicEncode } = require('../infra')
 
 const GET_OR_HEAD = ['GET', 'HEAD']
 
+const tlsCertificateErrorCodes = new Set([
+  'CERT_SIGNATURE_FAILURE',
+  'DEPTH_ZERO_SELF_SIGNED_CERT',
+  'ERR_TLS_CERT_ALTNAME_INVALID',
+  'SELF_SIGNED_CERT_IN_CHAIN',
+  'UNABLE_TO_GET_ISSUER_CERT_LOCALLY',
+  'UNABLE_TO_VERIFY_LEAF_SIGNATURE'
+])
+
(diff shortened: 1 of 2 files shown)


mcollina (changes requested):
I think this is spec mandated. @KhafraDev

KhafraDev (changes requested):
It's not spec mandated (other than having to throw a TypeError), but there are a number of libraries that check for the error message 'fetch failed', which this change would break. [is-network-error](https://github.com/sindresorhus/is-network-error/blob/b039fe8bb8da3607f8fb6a2505c40eb03aac5889/index.js#L10) is one of the largest ones.

PR #5327, opened 2026-05-27
feat(connect): add `preferH2` connector option to offer h2 first in ALPN

Antamansid:
## This relates to...

N/A — no existing issue. Affects clients talking to load balancers that select the ALPN protocol by client preference.

## Rationale

The built-in connector hardcodes `ALPNProtocols: ['http/1.1', 'h2']` (h1 first) after spreading user options (`lib/core/connect.js`), so the order cannot be overridden via the `connect` options. Servers that select the ALPN protocol by *client* preference — some load balancers do (e.g. those relying on OpenSSL's `SSL_select_next_proto` semantics) — then negotiate `http/1.1` even with `allowH2: true`, forcing the client onto HTTP/1.1.

This adds an opt-in `preferH2` connector option that, together with `allowH2`, offers ALPN as `['h2', 'http/1.1']` (h2 first), so HTTP/2 is negotiated whenever the server supports it. The default behaviour is unchanged (`preferH2` defaults to `false`). If the server does not support HTTP/2, ALPN transparently falls back to `http/1.1`.

## Changes

- `lib/core/connect.js`: accept a `preferH2` build option; when set together with `allowH2`, offer `ALPNProtocols: ['h2', 'http/1.1']` instead of the default `['http/1.1', 'h2']`.
- `types/connector.d.ts`: add `preferH2?: boolean` to `BuildOptions`.
- `docs/docs/api/Connector.md`: document the new option.
- `test/connect-h2-alpn-order.js`: assert the offered ALPN order with and without `preferH2`.
- `test/types/connector.test-d.ts`: type-level assertion for `preferH2`.

### Features

- New opt-in connector option `preferH2` (default `false`) to offer HTTP/2 first in ALPN, for servers that pick the protocol by client preference.

### Bug Fixes

N/A

N/A — the default ALPN order is unchanged; `preferH2` is opt-in.

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/docs/docs/api/Connector.md
+++ b/docs/docs/api/Connector.md
@@ -13,6 +13,7 @@ Every Tls option, see [here](https://nodejs.org/api/tls.html#tls_tls_connect_opt
 Furthermore, the following options can be passed:
 
 * **socketPath** `string | null` (optional) - Default: `null` - An IPC endpoint, either Unix domain socket or Windows named pipe.
+* **preferH2** `boolean` (optional) - Default: `false` - Only effective together with `allowH2`. When `true`, ALPN is offered as `['h2', 'http/1.1']` (HTTP/2 first) instead of the default `['http/1.1', 'h2']`. Use this when the server selects the ALPN protocol by *client* preference (e.g. some load balancers) so that HTTP/2 is negotiated whenever the server supports it. If the server does not support HTTP/2, ALPN transparently falls back to `http/1.1`.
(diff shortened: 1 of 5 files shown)


metcoder95 (approved):

trivikr (approved):

mcollina (approved):
lgtm

PR #5328, opened 2026-05-27
fix(mock): restore single-arg MockCallHistory.filterCallsByX

### LINKED ISSUE #5324: MockAgent MockCallHistory.filterCallsByX broken since 8.2.0
## Bug Description

Since 8.2.0, the `filterCallsByX` functions on `MockCallHistory` appear to be broken.

They seem to now require an additional "logs" parameter which I don't think is intended. If I add in `MockCallHistory.logs` (internal) as the second parameter the function works. This parameter is not documented in the [documentation](https://github.com/nodejs/undici/blob/main/docs/docs/api/MockCallHistory.md#filtercallsbyhost) or the function types: https://github.com/nodejs/undici/blob/cc793b780c5f181d779bb4367a9394efe8a8f9c6/types/mock-call-history.d.ts#L90-L104

This might have been introduced by https://github.com/nodejs/undici/pull/5045?

## Reproducible By

```js
import { MockAgent, request } from 'undici'

const mockAgent = new MockAgent({ enableCallHistory: true })
mockAgent.disableNetConnect()

mockAgent.get('https://example.com').intercept({ method: 'GET', path: '/hello' }).reply(200, { ok: true });

// (avoiding #5036 by not using global fetch)
const response = await request('https://example.com/hello', { dispatcher: mockAgent })
console.log(`Response status: ${response.statusCode}`)


const callHistory = mockAgent.getCallHistory();
console.log(`Calls: ${callHistory.calls()}`)

// does not work on 8.2.0+
console.log(`filterCallsByPath: ${callHistory.filterCallsByPath('/hello')}`)

// logs parameter required on 8.2.0+ (MockCallHistory.logs is not public)
// console.log(`filterCallsByPath with logs: ${callHistory.filterCallsByPath('/hello', callHistory.logs)}`)
```

## Expected Behavior

functions continue to work without the additional parameter as on <=8.1.0

## Logs & Screenshots

Output from above example

8.1.0
```
Response status: 200
Calls: protocol->https:|host->example.com|port->|origin->https://example.com|path->/hello|hash->|searchParams->{}|fullUrl->https://example.com/hello|method->GET|body->undefined|headers->undefined
filterCallsByPath: protocol->https:|host->example.com|port->|origin->https://example.com|path->/hello|hash->|searchParams->{}|fullUrl->https://example.com/hello|method->GET|body->undefined|headers->undefined
```

8.2.0 and above

```
Response status: 200
Calls: protocol->https:|host->example.com|port->|origin->https://example.com|path->/hello|hash->|searchParams->{}|fullUrl->https://example.com/hello|method->GET|body->undefined|headers->undefined
/Users/User/WebstormProjects/undici-bug/node_modules/undici/lib/mock/mock-call-history.js:40
      return logs.filter((log) => {
                  ^

TypeError: Cannot read properties of undefined (reading 'filter')
    at MockCallHistory.filterCallsByPath (/Users/User/WebstormProjects/undici-bug/node_modules/undici/lib/mock/mock-call-history.js:40:19)
    at file:///Users/User/WebstormProjects/undici-bug/index.js:17:47

Node.js v24.14.1

Process finished with exit code 1
```

### Additional context

node v24.14.1, MacOS 26.3.1

ogyamada:
## What

Restores the documented `mockCallHistory.filterCallsByX(value)` single-argument form. The helper returned by `makeFilterCalls` now defaults `logs` to `this.logs`, so external callers work again. Internal callers in `filterCalls` still pass `logs` explicitly, so the AND/OR semantics from #5045 are unchanged.

## Why

In 8.2.0, the `makeFilterCalls` helper was refactored to take a second `logs` argument used by the new AND/OR filter operators. The internal `handleFilterCallsWithOptions` always passes it, but the public methods (`filterCallsByPath`, `filterCallsByMethod`, etc.) are still documented to accept a single argument. Calling any `filterCallsByX('value')` since 8.2.0 throws `Cannot read properties of undefined (reading 'filter')`.

## Testing

- ran `node --test test/mock-call-history.js` — 46/46 green (10 new tests in the `MockCallHistory - filterCallsByX` block exercising each public filter helper plus a regression for "logs added after the method was bound")
- ran `node_modules/.bin/eslint --cache lib/mock/mock-call-history.js test/mock-call-history.js` — clean
- verified the issue's reproducer against the patched build

## Notes

The optional second `logs` parameter on the public methods wasn't documented, so no type-definition change is needed.

Fixes #5324.

### DIFF
--- a/lib/mock/mock-call-history.js
+++ b/lib/mock/mock-call-history.js
@@ -35,7 +35,7 @@ function buildAndValidateFilterCallsOptions (options = {}) {
 }
 
 function makeFilterCalls (parameterName) {
-  return (parameterValue, logs) => {
+  return (parameterValue, logs = this.logs) => {
     if (typeof parameterValue === 'string' || parameterValue == null) {
       return logs.filter((log) => {
         return log[parameterName] === parameterValue
--- a/test/mock-call-history.js
+++ b/test/mock-call-history.js
@@ -445,3 +445,138 @@ describe('MockCallHistory - filterCalls with options', () => {
     t.assert.strictEqual(filtered.length, 2)
   })
 })
+
+describe('MockCallHistory - filterCallsByX', () => {
+  test('filterCallsByProtocol should filter logs with a string', t => {
+    t.plan(2)
+
+    const mockCallHistoryHello = new MockCallHistory('hello')
+
+    mockCallHistoryHello[kMockCallHistoryAddLog]({ path: '/', origin: 'http://localhost:4000' })
+    mockCallHistoryHello[kMockCallHistoryAddLog]({ path: '/noop', origin: 'https://localhost:4000' })
+
+    const filtered = mockCallHistoryHello.filterCallsByProtocol('https:')
+
+    t.assert.strictEqual(filtered.length, 1)
+    t.assert.strictEqual(filtered[0]?.path, '/noop')
+  })
+
+  test('filterCallsByProtocol should filter logs with a regexp', t => {
+    t.plan(1)
+
+    const mockCallHistoryHello = new MockCallHistory('hello')
+
+    mockCallHistoryHello[kMockCallHistoryAddLog]({ path: '/', origin: 'http://localhost:4000' })
+    mockCallHistoryHello[kMockCallHistoryAddLog]({ path: '/noop', origin: 'https://localhost:4000' })
+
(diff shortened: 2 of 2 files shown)


metcoder95 (approved):

mcollina (approved):
lgtm

PR #5331, opened 2026-05-28
chore(core): use `regex.exec` instead of `string.match`

aduh95:
`string.match(regex)` calls internally `regex[Symbol.match]()` which itself calls `regex.exec()`. We can shortcut this and call `.exec` directly (from my testing, it's marginally faster in normal conditions, and much faster if any of `String.prototype` and/or `RegExp.prototype` have been mutated)

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready


metcoder95 (approved):

mcollina (approved):
lgtm

PR #5336, opened 2026-05-29
fix: preserve queued request ordering after stream body errors

### LINKED ISSUE #5335: Do not mask stream POST retry ordering in client-errors
## Problem

While working on the Node 26 / global dispatcher compatibility backport, we found that `test/node-test/client-errors.js` only passed on Node 26 after the test was relaxed on `main` to tolerate extra/reordered POST attempts.

That relaxation appears to mask a real client bug.

The strict invariant from the original test is:

1. a streaming POST body writes `"a string"` and then errors with `kaboom`
2. that POST fails
3. the following queued GET reconnects and succeeds
4. the failed POST must not be retried before the queued GET

On Node 26, with the original strict test, the server can observe a POST where it expected the queued GET. The current `main` test handles this by tearing down extra POST attempts, but that accepts behavior the test was meant to reject.

## Root cause

In the HTTP/1 write path, undici probes Node streams before installing the writer/error handling:

```js
if (body && typeof body.read === 'function') {
  // Try to read EOF in order to get length.
  body.read(0)
}
```

For user-provided `Readable` bodies, `body.read(0)` can synchronously invoke the stream's `read()` implementation. If that implementation pushes data and errors, the body error happens before undici has attached the request body writer/error handlers. This changes the order of events on Node 26 and lets the queued GET proceed as if it were the first request, while the original POST may still be observed/retried unexpectedly.

## Suggested fix

Forward-port the v7.x fix from PR #5319:

- remove the eager `body.read(0)` probe from `lib/dispatcher/client-h1.js`
- restore `test/node-test/client-errors.js` to assert the strict order instead of tolerating additional POST attempts
- remove the helper behavior that destroys extra POST requests in that test

The test should fail fast if the request following the failed POST is not the queued GET.

## Validation from the v7.x backport

With the eager `body.read(0)` removed and strict `client-errors` expectations restored, the following passed locally:

- Node 26: `node --test --test-reporter=spec --test-timeout=180000 test/node-test/client-errors.js`
- Node 25: same test file
- Node 20: same test file
- `npm run test:node-test`

mcollina:
## Summary

- avoid eager HTTP/1 `Readable` body probing before undici attaches body/error handlers
- restore strict `client-errors` expectations for the failed POST / queued GET ordering
- stop tolerating extra POST attempts in the reconnect test

Fixes #5335.

## Testing

- `node --test --test-reporter=spec --test-timeout=180000 test/node-test/client-errors.js`
- `~/.nvm/versions/node/v25.2.1/bin/node --test --test-reporter=spec --test-timeout=180000 test/node-test/client-errors.js`
- `~/.nvm/versions/node/v26.1.0/bin/node --test --test-reporter=spec --test-timeout=180000 test/node-test/client-errors.js`
- `PATH="$HOME/.nvm/versions/node/v26.1.0/bin:$PATH" npm run test:node-test`
- `npx eslint lib/dispatcher/client-h1.js test/node-test/client-errors.js`

### DIFF
--- a/lib/dispatcher/client-h1.js
+++ b/lib/dispatcher/client-h1.js
@@ -1088,11 +1088,6 @@ function writeH1 (client, request) {
     headers.push('content-type', body.type)
   }
 
-  if (body && typeof body.read === 'function') {
-    // Try to read EOF in order to get length.
-    body.read(0)
-  }
-
   const bodyLength = util.bodyLength(body)
 
   contentLength = bodyLength ?? contentLength
--- a/test/node-test/client-errors.js
+++ b/test/node-test/client-errors.js
@@ -122,71 +122,33 @@ test('GET errors and reconnect with pipelining 3', async (t) => {
   await p.completed
 })
 
-function installErrorAndReconnectServer (server, p, { contentLength, trackPostWithPlan }) {
-  let sawPost = false
-  let sawGet = false
-
-  server.on('request', (req, res) => {
-    if (req.method === 'GET') {
-      if (sawGet) {
-        req.socket?.destroy()
-        return
-      }
-
-      sawGet = true
-      p.strictEqual('/', req.url)
-      p.strictEqual('GET', req.method)
-      res.setHeader('content-type', 'text/plain')
-      res.end('hello')
-      return
-    }
-
-    if (sawPost) {
-      // Node.js 26 can surface additional POST attempts around the queued GET.
-      // Tear them down and keep the test focused on the reconnect behavior.
-      req.resume()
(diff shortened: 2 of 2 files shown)


lib/dispatcher/client-h1.js
```diff
@@ -1088,11 +1088,6 @@ function writeH1 (client, request) {
     headers.push('content-type', body.type)
   }
 
-  if (body && typeof body.read === 'function') {
-    // Try to read EOF in order to get length.
-    body.read(0)
-  }
-
```
ronag, line 1095: this is problematic in some cases...

PR #5357, opened 2026-06-04
fix: drop response chunks after the response stream is destroyed (#5356)

### LINKED ISSUE #5356: consumePush crashes when chunk arrives after body.json() teardown
## Bug Description

Calling `body.json()` (or other consume helpers) and then tearing down the stream (abort/timeout/close) runs `consumeFinish()`, which sets` consume.body = null`. `stream[kConsume]` is left in place.
If `BodyReadable.push()` runs after that - for example a late chunk from the decompress interceptor- it still calls `consumePush()`, which does `consume.body.push(chunk)` and throws:

```
TypeError: Cannot read properties of null (reading 'push')
    at consumePush (lib/api/readable.js:549)
```
This is a synchronous throw, not the AbortError from the .json() promise.

## Reproducible By

```javascript
const { Readable: BodyReadable } = require('undici/lib/api/readable');

describe('undici BodyReadable consume race', () => {
    it('must throw when a chunk arrives after consume teardown', async () => {
        const body = new BodyReadable({ resume: () => {}, abort: () => {} });
        const jsonPromise = body.json();

        await new Promise(resolve => queueMicrotask(resolve));

        body.emit('close');

        expect(() => body.push(Buffer.from('late chunk'))).toThrow(/Cannot read properties of null \(reading 'push'\)/);

        await expect(jsonPromise).rejects.toMatchObject({ name: 'AbortError' });
    });
});
```
in real usage: `request()` + `interceptors.decompress()`, `body.json()` with a short AbortSignal timeout while the response is still streaming

## Expected Behavior

Late chunks after consume teardown should be dropped. `.json()` should reject with AbortError only - no extra sync crash.

## Logs & Screenshots

```
TypeError: Cannot read properties of null (reading 'push')
    at consumePush (/.../undici/lib/api/readable.js:549:16)
    at BodyReadable.push (/.../undici/lib/api/readable.js:160:9)
```

## Environment

- undici 7.27.0
- Node.js v24.12.0
- macOS / Linux, HTTP/1.1
- Agent.compose(interceptors.decompress())

### Additional context

Looks related to #4880 (null guard for late push after H2 teardown). `consumePush` has no similar guard.

cesarvspr:
## This relates to...

Fixes #5356

## Rationale

When a request is aborted mid-stream while its body is being consumed (e.g. `body.json()` with an `AbortSignal` timeout), a late chunk could crash the process with `TypeError: Cannot read properties of null (reading 'push')` instead of the consume promise simply rejecting with the `AbortError`.

Root cause: the abort-signal listener in `lib/api/api-request.js` destroys `this.res` in place but, unlike `onResponseError`, does **not** null it. So a chunk that arrives afterwards - for example flushed asynchronously by the decompress interceptor after the abort - passes the existing `if (!this.res)` guard in `onResponseData` (the reference is still set, just destroyed) and is pushed into the torn-down `BodyReadable`, reaching `consumePush` on a consume whose `body` has already been nulled.

> Note: this supersedes the earlier revision of this PR, which guarded `consumePush` in `readable.js`. As @mcollina pointed out, the bug is elsewhere - that function should not be called after teardown. The fix now stops the late chunk at the handler, so `consumePush` is never reached.

## Changes

- `lib/api/api-request.js`: `onResponseData` now returns early when the response stream is already destroyed (`!this.res || this.res.destroyed`), dropping late chunks at the handler instead of pushing them into a torn-down stream.

### Alternative considered

Nulling `this.res` in the abort-signal listener (to mirror `onResponseError`) also fixes it and removes the asymmetry directly. I chose the `onResponseData` guard because it is a smaller surface, uses the public `Readable.destroyed`, and matches the `destroyed`-aware request-body guard in `api-pipeline.js`. Happy to switch to nulling `this.res` if you prefer the symmetry - the regression test covers both.

### Bug Fixes

- A late chunk arriving after an aborted consume no longer crashes the process; the consume promise rejects with the `AbortError` as expected.

N/A

## Status

- [x] Tested (new regression test `test/issue-5356.js`, RED without the fix, GREEN with it; `readable`/`request`/`client-request`/`decompress`/abort+signal suites pass; lint clean)
- [x] Documented (code comment; no public API change)
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/api/api-request.js
+++ b/lib/api/api-request.js
@@ -66,7 +66,13 @@ class RequestHandler extends AsyncResource {
       this.removeAbortListener = util.addAbortListener(signal, () => {
         this.reason = signal.reason ?? new RequestAbortedError()
         if (this.res) {
-          util.destroy(this.res.on('error', noop), this.reason)
+          // Null the reference before destroying, mirroring onResponseError, so
+          // that chunks flushed after the abort (e.g. an async decompressor
+          // flush) are dropped by the `!this.res` guard in onResponseData
+          // instead of being pushed into the torn-down stream.
+          const res = this.res
+          this.res = null
+          util.destroy(res.on('error', noop), this.reason)
         } else if (this.abort) {
           this.abort(this.reason)
         }
@@ -150,11 +156,7 @@ class RequestHandler extends AsyncResource {
   }
 
   onResponseData (controller, chunk) {
-    // The abort-signal listener destroys `this.res` in place without nulling
-    // it (unlike onResponseError). Drop chunks that arrive after the stream is
-    // gone - e.g. an async decompressor flush after an abort - instead of
-    // pushing them into the torn-down stream.
-    if (!this.res || this.res.destroyed) {
+    if (!this.res) {
       return
     }
 
--- a/lib/api/api-request.js
+++ b/lib/api/api-request.js
@@ -150,7 +150,11 @@ class RequestHandler extends AsyncResource {
   }
 
   onResponseData (controller, chunk) {
-    if (!this.res) {
+    // The abort-signal listener destroys `this.res` in place without nulling
+    // it (unlike onResponseError). Drop chunks that arrive after the stream is
+    // gone - e.g. an async decompressor flush after an abort - instead of
+    // pushing them into the torn-down stream.
+    if (!this.res || this.res.destroyed) {
       return
     }
 
--- a/test/issue-5356.js
+++ b/test/issue-5356.js
@@ -1,0 +1,60 @@
+'use strict'
+
+const { test } = require('node:test')
+const { tspl } = require('@matteo.collina/tspl')
+
+const { RequestHandler } = require('../lib/api/api-request')
+
+// Regression test for https://github.com/nodejs/undici/issues/5356
+//
+// When a request is aborted while its body is being consumed (e.g.
+// `body.json()`), the abort-signal listener destroys the response stream in
+// place but - unlike `onResponseError` - does not null `this.res`. A chunk that
+// arrives afterwards (for example flushed asynchronously by the decompress
+// interceptor) reached the destroyed stream and crashed synchronously with
+// "Cannot read properties of null (reading 'push')" instead of the consume
+// promise simply rejecting with the AbortError.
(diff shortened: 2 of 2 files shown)


metcoder95 (approved):

mcollina (changes requested):
I think the bug is elsewhere. That function should not be called.

cesarvspr:
thx for the heads up @mcollina 

The abort listener destroys `this.res` but never nulls it like `onResponseError` does, so the late decompressor chunk slips past the `if (!this.res)` check and into `consumePush`. Moved the guard to `onResponseData` (drop the chunk if the stream's already destroyed), reverted the `readable.js` change, added a test.

Could also just null `this.res` in the abort listener to match `onResponseError`. Went with the guard since it touches less, but easy to switch if you prefer that.

lib/api/api-request.js
```diff
@@ -150,7 +150,11 @@ class RequestHandler extends AsyncResource {
   }
 
   onResponseData (controller, chunk) {
-    if (!this.res) {
+    // The abort-signal listener destroys `this.res` in place without nulling
+    // it (unlike onResponseError). Drop chunks that arrive after the stream is
+    // gone - e.g. an async decompressor flush after an abort - instead of
```
AliMahmoudDev: adding this.res.destroyed to the guard is the right call. the comment explaining why this.res is not nulled by the abort handler is helpful context.

AliMahmoudDev:
Nice fix. Checking `this.res.destroyed` alongside `!this.res` covers the case where the abort listener tears down the response in place without nulling it. The test is clean too — good call using `body.once("close")` to avoid the error event rejection from `events.once`. lgtm

mcollina:
> thx for the heads up @mcollina
> 
> The abort listener destroys `this.res` but never nulls it like `onResponseError` does, so the late decompressor chunk slips past the `if (!this.res)` check and into `consumePush`. Moved the guard to `onResponseData` (drop the chunk if the stream's already destroyed), reverted the `readable.js` change, added a test.
> 
> Could also just null `this.res` in the abort listener to match `onResponseError`. Went with the guard since it touches less, but easy to switch if you prefer that.

Fixing the underlying bug is what should be done, not adding another if condition.

mcollina (approved):
lgtm

PR #5364, opened 2026-06-05
fix: finish paused HTTP/1 parser on socket end

### LINKED ISSUE #5360: Uncatchable `AssertionError: assert(!this.paused)` on socket end
## Description

If the consumer doesn't read the body immediately, the HTTP/1 parser pauses under 
backpressure. When the socket ends (FIN), `onHttpSocketEnd` calls `parser.finish()`,
which asserts `!this.paused`. Since the parser is paused, this throws an **uncatchable** 
`AssertionError` from the socket `'end'` handler and crashes the process.

```
AssertionError [ERR_ASSERTION]: assert(!this.paused)
    at Parser.finish (.../undici/lib/dispatcher/client-h1.js:374:5)
    at Socket.onHttpSocketEnd (.../undici/lib/dispatcher/client-h1.js:951:30)
```

`onHttpSocketEnd` doesn't check `parser.paused` before calling `finish()`.

## Reproduction (minimised with Claude Code)

```js
import { createServer } from "node:net";

const BODY = Buffer.alloc(64 * 1024, 0x61);

const server = createServer((sock) => {
  sock.once("data", () => {
    sock.write(
      "HTTP/1.1 200 OK\r\n" +
        `Content-Length: ${BODY.length}\r\n` +
        "Connection: close\r\n\r\n",
    );
    sock.write(BODY);
    sock.end(); // FIN
  });
});

server.listen(0, "127.0.0.1", async () => {
  const { port } = server.address();
  const res = await fetch(`http://127.0.0.1:${port}/`);
  void res; // never read the body -> parser pauses -> crash on FIN
  await new Promise((r) => setTimeout(r, 1000));
  console.log("no crash"); // never reached
});
```

The 64 KiB body is the body highWaterMark — the smallest size that pauses while still
arriving in one socket read, which makes this deterministic. It's not a fixed value:
the real trigger is "paused with the read buffer drained when FIN arrives," so the
size is network-dependent (intermittent across sizes over a real connection).

## Expected vs actual

- **Expected:** clean end, or a catchable error. A peer closing a non-keep-alive
  response under backpressure shouldn't crash the process.
- **Actual:** uncatchable `AssertionError`, exit code 1.

## Environment

- undici 8.3.0 (and current `main`)
- Node.js v22.22.2, Linux

marko1olo:
Fixes #5360.

When a response body applies backpressure, the HTTP/1 parser can be left paused. If the peer then closes the socket, `onHttpSocketEnd` calls `parser.finish()`, which previously asserted that the parser was not paused and could crash the process from the socket `end` handler.

This resumes the paused llhttp parser before finishing it, then lets the existing finish path handle EOF and completion normally.

Tests:
- `node --expose-gc --test --test-name-pattern "socket end completes response when body is paused by backpressure" test/client.js`
- `npm run lint -- --no-cache lib/dispatcher/client-h1.js test/client.js`
- `git diff --check`

### DIFF
--- a/lib/dispatcher/client-h1.js
+++ b/lib/dispatcher/client-h1.js
@@ -371,9 +371,14 @@ class Parser {
   finish () {
     assert(currentParser === null)
     assert(this.ptr != null)
-    assert(!this.paused)
 
     const { llhttp } = this
+
+    if (this.paused) {
+      llhttp.llhttp_resume(this.ptr)
+      this.paused = false
+      this.execute(this.socket.read() || EMPTY_BUF)
+    }
 
     let ret
 
--- a/test/client.js
+++ b/test/client.js
@@ -3,6 +3,7 @@
 const { tspl } = require('@matteo.collina/tspl')
 const { readFileSync, createReadStream } = require('node:fs')
 const { createServer } = require('node:http')
+const { createServer: createNetServer } = require('node:net')
 const { Readable, PassThrough } = require('node:stream')
 const { test, after } = require('node:test')
 const { Client, errors } = require('..')
@@ -1185,6 +1186,52 @@ test('ignore request header mutations', async (t) => {
       body.resume()
     })
     headers.test = 'asd'
+  })
+
+  await t.completed
+})
+
+test('socket end completes response when body is paused by backpressure', async (t) => {
+  t = tspl(t, { plan: 2 })
+
+  const payload = Buffer.from('aa')
+  const server = createNetServer((socket) => {
+    socket.once('data', () => {
+      socket.write(
+        'HTTP/1.1 200 OK\r\n' +
+        `Content-Length: ${payload.length}\r\n` +
+        'Connection: close\r\n\r\n'
+      )
+      socket.write(payload)
+      socket.end()
+    })
+  })
+  after(() => server.close())
+
+  server.listen(0, '127.0.0.1', () => {
+    const client = new Client(`http://127.0.0.1:${server.address().port}`)
+    after(() => client.close())
+
+    client.request({
+      path: '/',
+      method: 'GET',
+      highWaterMark: 1
+    }, (err, data) => {
+      t.ifError(err)
+
+      let body = ''
+      data.body.setEncoding('utf8')
+      data.body.on('end', () => {
+        t.strictEqual(body, payload.toString())
+      })
+
+      setTimeout(() => {
+        data.body.on('data', (chunk) => {
+          body += chunk
+        })
+        data.body.resume()
+      }, 50)
+    })
   })
 
   await t.completed


ronag (changes requested):
I don' think the test actually test for the fix. Please confirm that the test fails without the fix.

ronag:
1. Resuming without executing leaves the response incomplete (crash → hang). llhttp_resume() + llhttp_finish() stops the assertion crash, but when the parser was paused inside on_body, llhttp_finish does not fire on_message_complete, so request.onComplete is never called and the body stream hangs forever once the consumer reads it. Driving the parser forward the way resume() does fixes it:

if (this.paused) {
  llhttp.llhttp_resume(this.ptr)
  this.paused = false
  this.execute(this.socket.read() || EMPTY_BUF) // drive parser to on_message_complete
}

2. The regression test passes without the fix. As Copilot noted, attaching a 'data' listener immediately switches the stream to flowing mode, so the parser never pauses. A reliable repro keeps the body unconsumed until after FIN (attach only 'end' initially, defer 'data').

marko1olo:
Updated in `5132ed84`.

The regression now keeps the response body unconsumed until after the socket has ended, so the HTTP/1 parser actually remains paused when `onHttpSocketEnd()` calls `finish()`.

Red/green proof locally:

- With the test updated but without the new `this.execute(this.socket.read() || EMPTY_BUF)` flush in `finish()`, `node --test --test-name-pattern "socket end completes response" test/client.js` did not complete and hit my 60s command timeout.
- With the current patch, the same command passes:
  `1 pass, duration_ms 1922.6807`.

Also ran:

- `npx eslint lib/dispatcher/client-h1.js test/client.js`
- `git diff --check`

lib/dispatcher/client-h1.js
```diff
@@ -371,10 +371,18 @@ class Parser {
...
     assert(this.ptr != null)
-    assert(!this.paused)
 
     const { llhttp } = this
 
+    if (this.paused) {
+      llhttp.llhttp_resume(this.ptr)
+      this.paused = false
+      this.execute(this.socket.read() || EMPTY_BUF)
+      if (this.paused || this.ptr == null) {
```
ronag: Is this condition really required?

marko1olo:
Good point. I removed that extra early-return condition in `57847129`; the important part is just resuming llhttp and driving the parser with `this.execute(this.socket.read() || EMPTY_BUF)` before `llhttp_finish()`.

Re-ran locally:

```text
node --test --test-name-pattern "socket end completes response" test/client.js
npx eslint lib/dispatcher/client-h1.js test/client.js
git diff --check
```

lib/dispatcher/client-h1.js
```diff
@@ -371,10 +371,15 @@ class Parser {
...
     assert(this.ptr != null)
-    assert(!this.paused)
 
     const { llhttp } = this
 
+    if (this.paused) {
+      llhttp.llhttp_resume(this.ptr)
+      this.paused = false
+      this.execute(this.socket.read() || EMPTY_BUF)
+    }
```
AliMahmoudDev, line 381: the approach here makes sense — resuming the parser before finishing handles the backpressure case cleanly without changing the existing finish path.

one thought: the `this.execute(this.socket.read() || EMPTY_BUF)` call after resuming will try to consume whatever is left in the socket buffer. if the socket has already ended (which it has in this scenario), `socket.read()` returns null so we fall through to `EMPTY_BUF`. should be fine since the FIN is what triggers `finish()` in the first place, but worth noting that this only works correctly because `onHttpSocketEnd` is called after the kernel has drained the read buffer.

marko1olo:
Follow-up on the remaining red check here: the failing job is `Test with Node.js 25 on macos-latest`, and its failure is in `test/issue-5137.js` with late async `read ECONNRESET` activity after the test ended. That is unrelated to this paused-parser/socket-end patch.

I split the HTTP/2 timeout test teardown fix into #5383 so this PR can stay scoped to parser finish behavior.

styfle:
Friendly bump on this PR since I'm also seeing the same issue in https://github.com/nodejs/undici/issues/5360

Since you split the timeout test teardown in to https://github.com/nodejs/undici/pull/5383 and thats merged, does this PR need to be updated?

mcollina (approved):
lgtm

ronag (approved):
I would prefer [my variation](https://github.com/nxtedition/undici/commit/2aa573901f2596235e62aa42936c5895ea821874). But this is also good.

mcollina:
@ronag can you open a PR with it?

ronag:
https://github.com/nodejs/undici/pull/5474

PR #5373, opened 2026-06-07
fix(client): reject pipelined TLS altname errors

### LINKED ISSUE #5355: AssertionError in ERR_TLS_CERT_ALTNAME_INVALID handler when pipelining > 1
## Bug Description

When using a `Pool` or `Client` with `pipelining > 1`, a TLS connect error of type `ERR_TLS_CERT_ALTNAME_INVALID` causes an uncaught `AssertionError` that crashes the process.

## Reproduction

```js
import { Pool } from 'undici'

const pool = new Pool('https://wrong.host.example.com', {
  pipelining: 10,
  connections: 5,
})

// Send concurrent requests — crash occurs when TLS handshake fails
await Promise.allSettled(
  Array.from({ length: 20 }, () =>
    pool.request({ path: '/', method: 'GET' })
  )
)
```

Error:

```
AssertionError [ERR_ASSERTION]: false == true
    at connect (undici/lib/client.js:1313:7)
```

## Root Cause

In `lib/client.js`, the `ERR_TLS_CERT_ALTNAME_INVALID` connect error handler contains:

```js
if (err.code === 'ERR_TLS_CERT_ALTNAME_INVALID') {
  assert(client[kRunning] === 0)  // crashes when pipelining > 1
  while (client[kPending] > 0 ...) { ... }
}
```

With `pipelining > 1`, multiple requests can be in-flight (`kRunning > 0`) when the TLS error fires at connect time. The assertion assumes `pipelining=1` and throws when that assumption is violated.

Every other error code routes through `onError()` which already handles `kRunning > 0` correctly. The `ERR_TLS_CERT_ALTNAME_INVALID` path is the only one with this broken assumption.

## Fix

Drain in-flight running requests before draining pending ones, mirroring the existing `onError` pattern:

```js
if (err.code === 'ERR_TLS_CERT_ALTNAME_INVALID') {
  while (client[kRunning] > 0) {
    client[kQueue][client[kRunningIdx]++] = null
    errorRequest(client, request, err)
  }
  client[kPendingIdx] = client[kRunningIdx]
  while (client[kPending] > 0 ...) { ... }
}
```

## Versions

- undici: 5.29.0 (same assertion exists in v6 `lib/dispatcher/client.js`)
- node: 20.x
- Confirmed: crash does not occur with `pipelining: 1`

marko1olo:
## This relates to...

Fixes #5355

## Rationale

`ERR_TLS_CERT_ALTNAME_INVALID` had a special connect-error path that asserted `client[kRunning] === 0` before rejecting pending requests for the current server name. With `pipelining > 1`, requests can already be in the running segment when the TLS hostname validation error is reported, so the assertion can crash the process instead of rejecting the affected request promises.

### Features

N/A

### Bug Fixes

- drain and reject already-running requests when a TLS altname error is reported
- remove same-servername pending requests from the queue while rejecting them, preserving later pending requests for other server names
- add a regression test using a local HTTPS server and pipelined requests

N/A

## Status

- [x] Tested
- [x] Documented - not applicable; this fixes internal error handling
- [x] Review ready
- [ ] In review
- [ ] Merge ready

## Tested

- `node --test test\node-test\client-tls.js`
- `NODE_OPTIONS=--expose-gc node --test test\tls-cert-leak.js`
- `node --test test\client-connect.js`
- `node --test --test-name-pattern servername test\node-test\client-dispatch.js`
- `eslint lib\dispatcher\client.js test\node-test\client-tls.js`
- `git diff --check`

### DIFF
--- a/lib/dispatcher/client.js
+++ b/lib/dispatcher/client.js
@@ -568,9 +568,15 @@ function handleConnectError (client, err, { host, hostname, protocol, port }) {
   }
 
   if (err.code === 'ERR_TLS_CERT_ALTNAME_INVALID') {
-    assert(client[kRunning] === 0)
+    const running = client[kQueue].splice(client[kRunningIdx], client[kRunning])
+    client[kPendingIdx] = client[kRunningIdx]
+
+    for (let i = 0; i < running.length; i++) {
+      util.errorRequest(client, running[i], err)
+    }
+
     while (client[kPending] > 0 && client[kQueue][client[kPendingIdx]].servername === client[kServerName]) {
-      const request = client[kQueue][client[kPendingIdx]++]
+      const request = client[kQueue].splice(client[kPendingIdx], 1)[0]
       util.errorRequest(client, request, err)
     }
   } else {
--- a/test/node-test/client-tls.js
+++ b/test/node-test/client-tls.js
@@ -1,0 +1,44 @@
+'use strict'
+
+const { test } = require('node:test')
+const assert = require('node:assert')
+const { once } = require('node:events')
+const fs = require('node:fs')
+const https = require('node:https')
+const path = require('node:path')
+const { Pool } = require('../..')
+
+test('TLS altname errors reject pipelined requests', async (t) => {
+  const server = https.createServer({
+    key: fs.readFileSync(path.join(__dirname, '..', 'fixtures', 'key.pem')),
+    cert: fs.readFileSync(path.join(__dirname, '..', 'fixtures', 'cert.pem'))
+  }, (req, res) => {
+    res.end('ok')
+  })
+
+  t.after(() => {
+    server.close()
+  })
+
+  server.listen(0, '127.0.0.1')
+  await once(server, 'listening')
+
+  const pool = new Pool(`https://localhost:${server.address().port}`, {
+    connections: 1,
+    pipelining: 10,
+    connect: {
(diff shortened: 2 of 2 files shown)


mcollina (approved):
lgtm

lib/dispatcher/client.js
```diff
@@ -568,9 +568,15 @@ function handleConnectError (client, err, { host, hostname, protocol, port }) {
   }
 
   if (err.code === 'ERR_TLS_CERT_ALTNAME_INVALID') {
-    assert(client[kRunning] === 0)
+    const running = client[kQueue].splice(client[kRunningIdx], client[kRunning])
+    client[kPendingIdx] = client[kRunningIdx]
```
AliMahmoudDev, line 572: makes sense. the original `assert(client[kRunning] === 0)` assumed no pipelining — splicing out the running requests and rejecting them before handling pending ones is the right approach. using `splice` for pending too avoids the index shift issue. test with 20 pipelined requests against a bad cert is a good coverage of the race.

AliMahmoudDev:
Good catch — before this fix, running requests with a matching servername would stay in the queue and never get rejected when a TLS altname error hits. Using `splice` instead of incrementing the index is correct since `errorRequest` removes items from the queue. The test covering 20 pipelined requests all getting rejected with the right error code is solid.

PR #5374, opened 2026-06-07
fix(websocket): preserve handshake failure reason

### LINKED ISSUE #4625: More Descriptive Error When WebSocket Closes with Code 1006

marko1olo:
Fixes #4625. Also addresses the duplicate regression report in #4735.

When the opening handshake fails before the WebSocket is established, Undici already computes a specific failure reason, for example:

```text
Received network error or non-101 status code.
```

But that reason was not preserved for the later `ErrorEvent`, so users received a `TypeError` with an empty message. This stores the failure reason/cause on the handler before closing and uses it when creating the `ErrorEvent.error`, while keeping the error type as `TypeError`.

Local validation:

- Red repro before fix: connecting WebSocket to a plain HTTP 200 server emitted `TypeError` with an empty message.
- `node --test test\websocket\opening-handshake.js` -> 14 passed
- `npm run test:websocket` -> 130 passed
- `npm run lint -- --no-cache lib/web/websocket/connection.js lib/web/websocket/websocket.js test/websocket/opening-handshake.js`
- commit hook ran `npm run lint` successfully
- `git diff --check`

Authored with assistance from OpenAI Codex.

### DIFF
--- a/lib/web/websocket/websocket.js
+++ b/lib/web/websocket/websocket.js
@@ -54,6 +54,8 @@ function getSocketAddress (socket) {
  * @property {Set<number>} closeState
  * @property {import('../fetch/index').Fetch} controller
  * @property {boolean} [wasEverConnected=false]
+ * @property {string|undefined} failureReason
+ * @property {unknown} failureCause
  */
 
 // https://websockets.spec.whatwg.org/#interface-definition
@@ -114,7 +116,9 @@ class WebSocket extends EventTarget {
     socket: null,
     closeState: new Set(),
     controller: null,
-    wasEverConnected: false
+    wasEverConnected: false,
+    failureReason: undefined,
+    failureCause: undefined
   }
 
   #url
@@ -603,8 +607,10 @@ class WebSocket extends EventTarget {
       // 1006.
       code = 1006
 
+      const failureReason = this.#handler.failureReason ?? reason
+
       fireEvent('error', this, (type, init) => new ErrorEvent(type, init), {
-        error: new TypeError(reason)
+        error: new TypeError(failureReason, { cause: this.#handler.failureCause })
       })
     }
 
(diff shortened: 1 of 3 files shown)


lib/web/websocket/websocket.js
```diff
@@ -603,8 +607,10 @@ class WebSocket extends EventTarget {
       // 1006.
       code = 1006
 
+      const failureReason = this.#handler.failureReason ?? reason
+
       fireEvent('error', this, (type, init) => new ErrorEvent(type, init), {
-        error: new TypeError(reason)
+        error: new TypeError(failureReason, { cause: this.#handler.failureCause })
```
mcollina, line 613: I have a feeling that we cannot change the error message due to spec compatibility

marko1olo:
Thanks, that concern is valid. I re-checked the WebSockets Standard's protocol feedback section, and this PR changes script-visible failure detail through `ErrorEvent.error.message`, which is exactly the compatibility area the spec tries to avoid for failed connections.

I'm closing this PR rather than pushing that direction further. The empty `TypeError` may still be worth investigating separately, but this patch is not the right shape for it.

PR #5375, opened 2026-06-07
fix(readable): ignore late consume chunks

### LINKED ISSUE #5356: consumePush crashes when chunk arrives after body.json() teardown

marko1olo:
Fixes nodejs/undici#5356.

When a body consume helper such as `.json()` is torn down by `close`, `consumeFinish()` clears `consume.body` but leaves the consume object on the stream. If a late chunk arrives afterward, `BodyReadable.push()` still calls `consumePush()`, which dereferences `consume.body.push(...)` and throws a synchronous `TypeError` in addition to the expected `AbortError` rejection.

This drops chunks that arrive after the consume buffer has been cleared. The existing `AbortError` rejection path is unchanged.

Verification:

- Reproduced the issue path before the fix with `BodyReadable`, `.json()`, `close`, and a late `push()` causing `TypeError: Cannot read properties of null (reading 'push')`.
- Direct smoke after the fix confirms late `push()` no longer throws and the `.json()` promise rejects with `AbortError`.
- `node --test test\readable.js`
- `npm run lint -- --no-cache lib/api/readable.js test/readable.js`
- `git diff --check`
- Commit hook ran repository eslint successfully.

AI-assisted contribution: implemented with OpenAI Codex and manually verified with the commands above.

### DIFF
--- a/lib/api/readable.js
+++ b/lib/api/readable.js
@@ -546,6 +546,10 @@ function consumeEnd (consume, encoding) {
  * @returns {void}
  */
 function consumePush (consume, chunk) {
+  if (consume.body === null) {
+    return
+  }
+
   consume.length += chunk.length
   consume.body.push(chunk)
 }
--- a/test/readable.js
+++ b/test/readable.js
@@ -124,6 +124,24 @@ describe('Readable', () => {
     t.deepStrictEqual(obj, { hello: 'world' })
   })
 
+  test('.json() ignores late chunks after close', async function (t) {
+    t = tspl(t, { plan: 2 })
+
+    function resume () {
+    }
+    function abort () {
+    }
+    const r = new Readable({ resume, abort })
+    const jsonPromise = r.json()
+
+    await new Promise(resolve => queueMicrotask(resolve))
+
+    r.emit('close')
+    t.strictEqual(r.push(Buffer.from('late chunk')), true)
+
+    await t.rejects(jsonPromise, { name: 'AbortError' })
+  })
+
   test('.text()', async function (t) {
     t = tspl(t, { plan: 1 })
 


metcoder95:
These are all automated PRs, to short timeframe for multiple PRs that requires time to debug them.

cc: @mcollina @ronag

ronag (approved):

lib/api/readable.js
```diff
@@ -546,6 +546,10 @@ function consumeEnd (consume, encoding) {
  * @returns {void}
  */
 function consumePush (consume, chunk) {
+  if (consume.body === null) {
```
AliMahmoudDev, line 549: clean fix. the early return in `consumePush` handles the race condition without affecting the normal flow. the test correctly reproduces the teardown → late chunk sequence.

AliMahmoudDev:
Simple and effective — the null check on `consume.body` before pushing prevents late chunks from crashing after the consume has ended. This complements #5357 nicely (which handles the RequestHandler level) by catching it at the readable level.

metcoder95 (approved):

Issue #3951, opened 2024-12-16
EAI_AGAIN Error Not Thrown on getaddrinfo Failure


paucabuti:
## Bug Description

In undici versions @7.1.0, and @7.0.0 the EAI_AGAIN error is no longer thrown when the getaddrinfo syscall fails. Instead, the request returns an object with { opaque: null }, which is unexpected behavior.
Additionally, this issue might be related to [fix: response error interceptor #3930](https://github.com/nodejs/undici/pull/3930).

## Reproducible By

1. Attempt to make a request to a hostname that triggers a getaddrinfo failure.
2. Observe that instead of throwing an EAI_AGAIN error, the response contains { opaque: null }.

ruizmarc:
Hi, 

We are also facing this problem and it's quite annoying as it is now harder to find out that one of services it is down it is not ready yet.

Is a fix coming soon? If not, is there any workaround? 

Thanks in advance!

mcollina:
Thanks for reporting!

Can you provide steps to reproduce? We often need a [reproducible example](https://stackoverflow.com/help/minimal-reproducible-example), e.g. some code that allows someone else to recreate your problem by just copying and pasting it. If it involves more than a couple of different file, create a new repository on GitHub and add a link to that.

ruizmarc:
Hi,

Here you have an example, you only need to do a simple request with the errors interceptor:

```
import { request, getGlobalDispatcher } from 'undici';

await request('http://users:3000/users', {
  method: 'GET',
  dispatcher: getGlobalDispatcher().compose(interceptors.responseError()),
});
```

In my case I found the problem using docker, but it happens even without docker with a simple request.

In my use case, one microservice must request another one in the same docker network, but the one receiving the request is still not up. Prior to version 7 we were receiving EAI_AGAIN  if the microservice was not up with throwError, but now, with Undici 7, we just receive and empty response without any error even if there is nobody to reply.

paucabuti:
I found another issue related to interceptors while investigating the missing EAI_AGAIN error.

Using the dns interceptor in combination with the responseError interceptor causes inconsistent behavior depending on their order:

* **dns before responseError**: The EAI_AGAIN error is correctly thrown, resolving the original issue.
* **responseError before dns**: Crashes with TypeError: handler.onError is not a function.

```javascript
const { request, interceptors, getGlobalDispatcher } = require('undici');

(async () => {
  try {
    const response = await request('http://users:1000/',
      {
[...]

PR #5376, opened 2026-06-07
fix(dns): skip requests without origin

### LINKED ISSUE #4234: The DNS Interceptor is unable to retrieve the origin properly
## Bug Description

In the code example below, I provide the URL twice: first, when creating the Client instance, and again as the origin in the request options. This redundancy is unnecessary. However, when using the DNS interceptor, omitting the URL from the request options will result in an error.

## Reproducible By

```js//
import { Client, interceptors } from 'undici';
const { dns } = interceptors;

const client = new Client('https://jsonplaceholder.typicode.com/').compose(dns());

const response = await client.request({
  origin: 'https://jsonplaceholder.typicode.com/', // Commenting this line will cause an error
  method: 'GET',
  path: '/posts/1',
  headers: {
    'Content-Type': 'application/json',
  },
});

const data = await response.body.json();
console.log(data);
```

## Expected Behavior

The DNS interceptor should be capable of retrieving the URL directly from the client.

## Logs & Screenshots

Image

## Environment

Ubuntu 22.04 LTS,
Node v22.15.1,
npm  v10.9.2
undici v7.10.0

marko1olo:
## This relates to

Fixes #4234.
Relates to #3957.

## Rationale

`Client` and `Pool` requests do not include `origin` in the per-request dispatch options. The DNS interceptor currently dereferences `origDispatchOpts.origin.constructor`, so composing `dns()` with a `Client` or `Pool` and calling `.request({ method, path })` throws before the underlying dispatcher can handle the request.

Prior discussion in #3957 asked to keep this narrow: avoid overriding `Client.request()` / `Pool.request()`, and make the DNS interceptor ignore requests that do not carry an origin. This patch follows that direction.

## Changes

- Return the original dispatch unchanged when DNS interceptor receives dispatch options without `origin`.
- Add regression coverage for `Client(...).compose(dns()).request({ method, path })`.
- Add the same coverage for `Pool(...).compose(dns()).request({ method, path })`.
- Assert the custom lookup is not called in the no-origin path.

## Checks

- `node --test --test-name-pattern "#4234" test\interceptors\dns.js`
- `node --test --test-name-pattern "#4444|#3951|#4234" test\interceptors\dns.js`
- `node --test --test-name-pattern "Should handle ENOTFOUND response error" test\interceptors\dns.js`
- direct smoke for explicit-origin DNS lookup path with a local HTTP server
- `npm run lint -- --no-cache lib/interceptor/dns.js test/interceptors/dns.js`
- `git diff --cached --check`

Note: a full local `node --test test\interceptors\dns.js` run timed out in this Windows environment before producing a result; the focused regressions and adjacent DNS paths above passed.

## Status

- [x] Tested
- [x] Review ready

AI-assisted contribution disclosure: implementation prepared with OpenAI Codex and reviewed/tested locally before submission.

### DIFF
--- a/lib/interceptor/dns.js
+++ b/lib/interceptor/dns.js
@@ -535,6 +535,10 @@ module.exports = interceptorOpts => {
 
   return dispatch => {
     return function dnsInterceptor (origDispatchOpts, handler) {
+      if (origDispatchOpts.origin == null) {
+        return dispatch(origDispatchOpts, handler)
+      }
+
       const origin =
         origDispatchOpts.origin.constructor === URL
           ? origDispatchOpts.origin
--- a/test/interceptors/dns.js
+++ b/test/interceptors/dns.js
@@ -11,7 +11,7 @@ const { once } = require('node:events')
 const { tspl } = require('@matteo.collina/tspl')
 const pem = require('@metcoder95/https-pem')
 
-const { interceptors, Agent, request } = require('../..')
+const { interceptors, Agent, Client, Pool, request } = require('../..')
 const { dns } = interceptors
 
 // Helper to check if IPv6 is available for localhost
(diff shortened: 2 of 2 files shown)


mcollina (approved):
lgtm

PR #5377, opened 2026-06-07
test: handle aggregate balanced pool errors

marko1olo:
## Summary
- Handle request errors whose port or meaningful code is wrapped inside `AggregateError.errors` in the balanced-pool weighted round-robin tests.
- Stop assuming every thrown request error has `error.socket.remotePort`.
- Keep the existing expected request ordering and ratio assertions unchanged.

## Why
On Windows, recent Node versions can surface localhost connection failures as aggregate errors with nested `ECONNREFUSED` details. That made the test either throw while reading `error.socket.remotePort` or fail to count the intended connection-refused event.

## Test
- `node --test test\node-test\balanced-pool.js --test-name-pattern "weighted round robin"`
- `npx eslint test\node-test\balanced-pool.js`
- `git diff --check`

AI-assisted contribution: implementation and verification were performed with OpenAI Codex under my direction.

### DIFF
--- a/test/node-test/balanced-pool.js
+++ b/test/node-test/balanced-pool.js
@@ -352,8 +352,48 @@ test('getUpstream returns undefined for closed/destroyed upstream', (t) => {
 })
 
 function getErrorPort (err) {
-  return err?.socket?.remotePort ?? err?.port
+  if (err?.socket?.remotePort !== undefined) {
+    return err.socket.remotePort
+  }
+
+  if (err?.port !== undefined) {
+    return err.port
+  }
+
+  // The upstreams below are built on `localhost`, which resolves to both ::1 and
+  // 127.0.0.1, so a connection failure arrives as an AggregateError from the
+  // happy eyeballs attempt. The port is only on the individual causes.
+  if (Array.isArray(err?.errors)) {
+    for (const cause of err.errors) {
+      const port = getErrorPort(cause)
+      if (port !== undefined) {
+        return port
+      }
+    }
+  }
+
+  return undefined
 }
+
+test('getErrorPort resolves the port of an aggregated connection error', async (t) => {
+  const p = tspl(t, { plan: 2 })
+
+  // Nothing listens on port 1. Because the upstream is `localhost`, both ::1 and
+  // 127.0.0.1 are attempted and the refusals arrive wrapped in an AggregateError
+  // that carries no port of its own.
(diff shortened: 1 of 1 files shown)


AliMahmoudDev:
Makes sense — wrapping the port/code extraction in a try/catch and falling back to the AggregateError.errors array handles the Windows Node 25 case where localhost failures come as wrapped aggregate errors. Keeps the existing assertions intact while making the test robust across platforms.

marko1olo:
Rebased on `main` and cut down. This had drifted into conflict, and re-checking it against current `main` showed most of it was no longer needed.

What changed:

- **Dropped the `e.socket.remotePort` guard.** #5463 landed the equivalent guard on 2 July as `getErrorPort()`, so there was nothing left to do there.
- **Dropped `getRequestErrorCode()`.** I had assumed the `ECONNREFUSED` / `UND_ERR_SOCKET` branches needed the code unwrapped out of the `AggregateError`. That assumption was wrong — Node sets `code` on the aggregate as well as on the causes:

  ```
  constructor : AggregateError
  e.code      : "ECONNREFUSED"
  ```

  so the existing branches already classify correctly and that helper was dead weight.
- **Kept the port lookup, now recursing into `err.errors`.** This is the part still broken on `main`. `getErrorPort()` avoids the crash but returns `undefined` for an aggregate, because the aggregate carries no port of its own:

  ```
  getErrorPort(e)       : undefined      <- main
  servers.find(...)     : NO MATCH  -> falls through to `throw e`
  recursing equivalent  : matched server A
  ```

  So #5463 converted the `TypeError` into a rethrow rather than making the case pass.

Why an aggregate turns up here at all: the cases build their upstreams as `http://localhost:${port}`, and `localhost` resolves to both `::1` and `127.0.0.1`. When a test server is down, both families are attempted and the two refusals come back wrapped by the happy eyeballs attempt. That is also why this surfaces on the Windows jobs specifically.

The added test does not depend on a downed server's timing — nothing listens on port 1, so the refusal is deterministic. Against the previous helper it fails with `actual: undefined, expected: 1`.

Locally `node --test test/node-test/balanced-pool.js` gives 23/23 and eslint is clean. That is Node 24 on Windows, so I have not reproduced the Node 22 matrix cell that originally went red — the argument for the fix rests on the aggregate shape above, not on my run.

metcoder95 (approved):

PR #5387, opened 2026-06-07
docs: fix Cookie sameSite type - String should be Strict

mukejane:
Fixes issue #5380. The sameSite field was documented as "String" but the actual type is "Strict".


KhafraDev:
this was fixed

PR #5388, opened 2026-06-08
Retry fetch when an idle H2 session socket closes

### LINKED ISSUE #3492: `fetch` may try to use a closed connection
### Version

v22.6.0

### Platform

```text
Darwin MacBook-Pro-6.local 23.5.0 Darwin Kernel Version 23.5.0: Wed May  1 20:12:58 PDT 2024; root:xnu-10063.121.3~5/RELEASE_ARM64_T6000 arm64
```

### Subsystem

http

```js
// fetch-test.js
const {createServer} = require('node:http');

const port = 8080;
const url = 'http://localhost:' + port;

const server = createServer((req, res) => res.end()).listen(port, async () => {
  await fetch(url);
  server.closeIdleConnections();

  setImmediate(async () => {
    await fetch(url); // Throws TypeError with cause UND_ERR_SOCKET or ECONNRESET
    server.close();
  });
});
```

Reproduces consistently for me but the error `cause` varies roughly evenly between `ECONNRESET` and `UND_ERR_SOCKET` (details below)

`fetch` creates a new connection if there are none open, and the request succeeds.

### What do you see instead?

```
node fetch-test.js
node:internal/deps/undici/undici:13178
      Error.captureStackTrace(err);
            ^

TypeError: fetch failed
    at node:internal/deps/undici/undici:13178:13
    at async Immediate.<anonymous> (/Users/robhogan/workspace/fetch-test.js:11:5) {
  [cause]: SocketError: other side closed
      at Socket.<anonymous> (node:internal/deps/undici/undici:6020:28)
      at Socket.emit (node:events:532:35)
      at endReadableNT (node:internal/streams/readable:1696:12)
    code: 'UND_ERR_SOCKET',
    socket: {
      localAddress: '::1',
      localPort: 57996,
      remoteAddress: undefined,
      remotePort: undefined,
      remoteFamily: undefined,
      timeout: undefined,
      bytesWritten: 338,
      bytesRead: 122
    }
  }
}

Node.js v22.6.0
```

**OR**

```
node fetch-test.js
node:internal/deps/undici/undici:13178
      Error.captureStackTrace(err);
            ^

TypeError: fetch failed
    at node:internal/deps/undici/undici:13178:13
    at async Immediate.<anonymous> (/Users/robhogan/workspace/fetch-test.js:11:5) {
  [cause]: Error: read ECONNRESET
      at TCP.onStreamRead (node:internal/stream_base_commons:218:20) {
    errno: -54,
    code: 'ECONNRESET',
    syscall: 'read'
  }
}

Node.js v22.6.0
```

### Additional information

This seems to be quite sensitive to timing/the event loop in a way I haven't pinned down.
 - The `setImmediate` (or `setTimeout(cb, 0)`) is required to reproduce the issue.
 - Adding another `setImmediate` before the second `fetch` makes it succeed.
 - Adding `{headers:{'Connection': 'close'}}` to the first request succeeds.

marko1olo:
## Summary
- retry a fetch network attempt once when the peer closes the connection before response headers are received
- restrict the retry to replayable/no-body requests and existing transient close errors (`ECONNRESET`, `EPIPE`, `UND_ERR_SOCKET`)
- add an H2 fetch regression test that closes the idle session socket and verifies a replayed POST body reaches the retried request

Fixes #3492

## Tests
- `node --test ./test/fetch/http2.js`
- `npm run lint -- lib/web/fetch/index.js test/fetch/http2.js`
- `npm run test:h2:fetch`
- `git diff --check`

### DIFF
--- a/lib/web/fetch/index.js
+++ b/lib/web/fetch/index.js
@@ -67,6 +67,12 @@ const { bytesMatch } = require('../subresource-integrity/subresource-integrity')
 const { isomorphicEncode } = require('../infra')
 
 const GET_OR_HEAD = ['GET', 'HEAD']
+const retriableFetchNetworkErrorCodes = new Set([
+  'ECONNRESET',
+  'EPIPE',
+  'UND_ERR_SOCKET'
+])
+const kFetchNetworkErrorRetry = Symbol('fetch network error retry')
 
 const defaultUserAgent = typeof __UNDICI_IS_NODE__ !== 'undefined' || typeof esbuildDetection !== 'undefined'
   ? 'node'
@@ -149,6 +155,34 @@ class Fetch extends EE {
     this.connection?.destroy(error)
     this.emit('terminated', error)
   }
+}
+
+function createFetchConnection (fetchParams) {
+  fetchParams.controller.connection = {
+    abort: null,
+    destroyed: false,
+    destroy (err, abort = true) {
+      if (!this.destroyed) {
+        this.destroyed = true
+        if (abort) {
+          this.abort?.(err ?? new DOMException('The operation was aborted.', 'AbortError'))
+        }
+      }
+    }
+  }
+}
+
+function canReplayRequestBody (request) {
+  return request.body == null || request.body.source != null
+}
+
+function isRetriableFetchNetworkError (error) {
+  return retriableFetchNetworkErrorCodes.has(error?.code) ||
+    retriableFetchNetworkErrorCodes.has(error?.cause?.code)
+}
+
+function waitForFetchRetry () {
+  return new Promise(resolve => setTimeout(resolve, 0))
 }
 
 function handleFetchDone (response) {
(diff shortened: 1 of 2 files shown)


lib/web/fetch/index.js
```diff
@@ -2000,6 +2023,22 @@ async function httpNetworkFetch (
...
+      fetchParams.controller.connection.destroy(err, false)
+
+      await waitForFetchRetry()
+
+      if (isCancelled(fetchParams)) {
+        return makeAppropriateNetworkError(fetchParams, err)
+      }
+
+      if (request.body != null) {
+        request.body = safelyExtractBody(request.body.source)[0]
```
AliMahmoudDev, line 2036: this `safelyExtractBody` call can throw if the body source was already consumed or is not extractable (e.g. a ReadableStream). since we are already inside a catch handler, a throw here would bypass the rest of the error handling and propagate as an unhandled rejection. worth wrapping in a try/catch that falls through to the normal error path.

test/fetch/http2.js
```diff
@@ -123,6 +123,68 @@ test('[Fetch] Simple GET with h2', async (t) => {
...
+      rejectUnauthorized: false
+    },
+    allowH2: true
+  })
+
+  t.after(async () => {
+    for (const socket of sockets) {
+      socket.destroy()
+    }
+
```
AliMahmoudDev, line 168: `server.close()` here is not awaited. the rest of the tests in this file use `closeClientAndServerAsPromise` which properly waits for the server to close. if the server is still handling connections during teardown this could cause "async activity after test" warnings.

PR #5391, opened 2026-06-08
fix: allow h2 post request multiplexing

### LINKED ISSUE #5390: HTTP/2 POST requests serialize on an established H2 session
## Bug Description

In `undici@8.4.0`, concurrent `POST` requests over an already-established HTTP/2 session appear to serialize instead of multiplexing.

This seems related to #5362 / #4143, but separate from the known default Pool/Agent cold-burst behavior where concurrent requests during ALPN can fan out into multiple TCP/H2 sessions.

### Reproducible By

Warm one H2 connection first, clear diagnostics, then dispatch concurrent `POST` requests to the same origin.

With:

```js
new Agent({
  allowH2: true,
  connections: 1
})
```

Observed:

```text
N=30
new connections during measured phase: 0

sendHeaders offsets:
5, 98, 199, 289, 381, 475, 580, 723, ...
...
2701, 2792, 2883ms
```

So the requests are using an existing H2 connection, but headers are sent roughly one at a time.

### Expected

Concurrent `POST` requests should be dispatched as parallel H2 streams, bounded by the server's `SETTINGS_MAX_CONCURRENT_STREAMS`.

### Suspected cause

`client-h2.js` treats non-idempotent requests as busy while another request is running:

```js
if (client[kRunning] > 0) {
  if (request.idempotent === false) return true
}
```

And `POST` defaults to `idempotent: false`:

```js
this.idempotent = idempotent == null
  ? method === 'HEAD' || method === 'GET'
  : idempotent
```

This appears to prevent normal `POST` requests from multiplexing over HTTP/2.

Disclosure: This issue/comment was drafted with the assistance of codex

mcollina:
## Summary

- allow HTTP/2 to multiplex non-idempotent requests on independent streams
- keep HTTP/2 serialization for upgrade/CONNECT setup and non-replayable streaming bodies
- add a regression test for concurrent POST requests over an established H2 session

Fixes: #5390

## Tests

- `node --test test/issue-5390.js`
- `node --test test/http2-dispatcher.js test/http2-agent.js`
- `git diff --check`

### DIFF
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -279,10 +279,10 @@ function connectH2 (client, socket) {
         if (client[kRunning] > 0) {
           // We are already processing requests
 
-          // Non-idempotent request cannot be retried.
-          // Ensure that no other requests are inflight and
-          // could cause failure.
-          if (request.idempotent === false) return true
+          // Unlike HTTP/1.1 pipelining, HTTP/2 multiplexes requests on
+          // independent streams, so non-idempotent requests can be dispatched
+          // concurrently. Retry eligibility is handled by stream/session error
+          // handling instead of by serializing all non-idempotent requests.
           // Don't dispatch an upgrade until all preceding requests have completed.
           // Possibly, we do not have remote settings confirmed yet.
           if ((request.upgrade === 'websocket' || request.method === 'CONNECT') && session[kRemoteSettings] === false) return true
--- a/test/issue-5390.js
+++ b/test/issue-5390.js
@@ -1,0 +1,85 @@
+'use strict'
+
+const { test, after } = require('node:test')
+const { createSecureServer } = require('node:http2')
+const { once } = require('node:events')
+const { setTimeout: sleep } = require('node:timers/promises')
+
+const { tspl } = require('@matteo.collina/tspl')
+const pem = require('@metcoder95/https-pem')
+
+const { Agent, request } = require('..')
+
+test('HTTP/2 POST requests multiplex on an established session', async t => {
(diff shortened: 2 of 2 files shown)


suzunn:
I think the behavior change is right for H2, but I would add one more regression test with a non-buffer body, for example a `Readable` or async iterable POST body. The new test covers concurrent POST requests with replayable string bodies, while the removed guard also used to serialize the cases people usually classify as non-replayable. Even if retry eligibility is handled elsewhere, having a streaming-body fixture here would lock down the intended separation: H2 stream multiplexing should be allowed, but a later stream/session failure must not accidentally make the request replayable.

thedebugger:
@mcollina thanks for quick PR. I'll also test this in on my side and report back how is this working

thedebugger:
@mcollina I tested the PR - PR removes the non-idempotent POST busy guard, which fixes low-level `undici.request()` with fixed Buffer bodies. However, `undici.fetch()` POST with a JSON string body still serializes on H2 because the request body reaches `client-h2 busy()` as an async iterable with unknown body length. The remaining stream-body guard still returns busy while another H2 stream is running.

This affects ordinary fetch POST JSON requests, including POST requests whose response is text/event-stream. Both of these request types we are using in [Tensorlake ts sdk](https://github.com/tensorlakeai/tensorlake).  Can we relax that constraints for H2?

metcoder95 (approved):

metcoder95:
@thedebugger can you open another issue with 
[Minimum Reproducible Example](https://stackoverflow.com/help/minimal-reproducible-example)?
Just to keep this PR scoped to what it intends to do

PR #5394, opened 2026-06-08
fix: decode split chunks after readable setEncoding

marko1olo:
Fixes #5002.

## Summary
- route BodyReadable.setEncoding() through Readable.prototype.setEncoding() so Node installs its StringDecoder
- preserve the previous invalid-encoding no-op guard
- add a regression for a three-byte UTF-8 character split across response body chunks

## Tests
- NODE_PATH=C:\\hades\\oss\\undici\\node_modules node --test test\\readable.js
- NODE_PATH=C:\\hades\\oss\\undici\\node_modules C:\\hades\\oss\\undici\\node_modules\\.bin\\eslint.cmd --no-cache lib\\api\\readable.js test\\readable.js
- git diff --check

### DIFF
--- a/lib/api/readable.js
+++ b/lib/api/readable.js
@@ -110,6 +110,9 @@ class BodyReadable extends Readable {
    */
   on (event, listener) {
     if (event === 'data' || event === 'readable') {
+      if (this._readableState.encoding && !this._readableState.decoder) {
+        super.setEncoding(this._readableState.encoding)
+      }
       this[kReading] = true
       this[kUsed] = true
     }
@@ -542,10 +545,14 @@ function consumeEnd (consume, encoding) {
 
 /**
  * @param {Consume} consume
- * @param {Buffer} chunk
+ * @param {Buffer|string} chunk
  * @returns {void}
  */
 function consumePush (consume, chunk) {
+  if (typeof chunk === 'string') {
+    chunk = Buffer.from(chunk, consume.stream._readableState.encoding || 'utf8')
+  }
+
   consume.length += chunk.length
   consume.body.push(chunk)
 }
--- a/test/readable.js
+++ b/test/readable.js
@@ -144,6 +144,122 @@ describe('Readable', () => {
     t.strictEqual(text, 'hello world')
   })
 
+  test('setEncoding buffers split multibyte utf8 characters', async function (t) {
+    t = tspl(t, { plan: 2 })
+
+    function resume () {
+    }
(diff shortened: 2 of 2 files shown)


marko1olo:
Follow-up on the red CI after the amended `04f3c290` push:

- The readable/setEncoding regressions pass locally:
  - `NODE_PATH=C:\hades\oss\undici\node_modules node --test test\readable.js` -> 14 passed
  - `NODE_PATH=C:\hades\oss\undici\node_modules node --test test\client-request.js --test-name-pattern "request multibyte"` -> 47 passed
  - scoped eslint and `git diff --check` passed
- The current failing CI cluster is `test/interceptors/cache.js` (`vary headers are present in revalidation request`, `max-stale`, `stale-if-error`). I reproduced the same 49 passed / 3 failed result on a detached `origin/main` worktree at `6ea54ef8`, without this PR patch.

So the current cache-interceptor failures look like a baseline/main failure rather than a regression from this readable change.

metcoder95:
This conflicts with #5414

marko1olo:
> This conflicts with #5414

Sorry for the long silence on this. Coming back to it, the answer to that turned out to be less interesting than what I found while checking, so I'll take both in turn.

On #5414: I don't think the two conflict in substance. #5414 is test-only — `+33/-0`, all of it in `test/client-request.js`, nothing under `lib/` — so it doesn't touch a file this PR touches, and `git apply` puts its diff on top of this branch cleanly. Its test also passes on current `main` unpatched, so it reads as a reproduction that has since gone green rather than a competing fix.

On this PR: **it is superseded and I'm closing it.** #5003 landed while this sat open and covered both halves of what this was doing — the `setEncoding` → `on('data')` orderings and the buffered-chunk path. I confirmed the streaming cases pass on `main` with none of this patch present.

What I found while checking that is worth more than this PR was: #5003 also introduced a silent response body truncation on the consume path. `consumeStart()` treats `kPreservedBuffer` and `state.buffer` as mutually exclusive, so any chunk arriving after `setEncoding()` and before the consume starts is never read. A 9-byte body comes back as 4 bytes from `.text()`, with no error. It reproduces on v8.6.0 through v8.9.0 and on current `main`, and not on `74a2299e^`.

I've filed that separately as #5611 with a bisected repro, since it needs a different patch, different tests and a fresh description rather than being pushed onto a 50-day-old branch. Closing here to keep the queue clear.

PR #5400, opened 2026-06-08
fix: guard idle socket validation to skip fresh sockets

mcollina:
Refs: GHSA-35p6-xmwp-9g52

Backports idle socket validation hardening to v6.x.


KhafraDev:
@mcollina was this meant to be visible yet?

mcollina:
yes. I'm assembling the security releaS

PR #5414, opened 2026-06-11
fix: preserve utf8 decoding across response chunks

pupuking723:
## This relates to...

Fixes #5002

## Rationale

BodyReadable#setEncoding() only stored the encoding on the readable state. That made streamed consumers decode each buffered chunk independently, so an incomplete UTF-8 sequence at a response chunk boundary produced replacement characters.

The internal body consumers still need raw Buffer chunks so body.text() and body.json() can aggregate and decode the full payload.

### Features

N/A

### Bug Fixes

- Keep raw Buffer chunks for internal body consumption.
- Use a StringDecoder when BodyReadable is consumed through the Readable API after setEncoding().
- Add a regression test for async iteration over a response body where a 3-byte UTF-8 character spans response chunks.

N/A

## Status

- [x] Tested
- [ ] Benchmarked (optional)
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

Verification:

- node --test --test-name-pattern "request multibyte (json|text) with setEncoding|async iteration and setEncoding" test/client-request.js
- npx borp --timeout 180000 -p test/client-request.js
- npm run lint
- npm run test:typescript
- git diff --check

Note: npm ci --ignore-scripts was used locally because npm install exited with "Exit handler never called!" after dependency extraction. The test/lint commands above were run against the installed dependency tree. Local Node.js is v22.17.0 while the package currently requires >=22.19.0.

### DIFF
--- a/test/client-request.js
+++ b/test/client-request.js
@@ -1494,6 +1494,39 @@ test('request multibyte text with setEncoding', async (t) => {
   await t.completed
 })
 
+test('request multibyte text with async iteration and setEncoding', async (t) => {
+  t = tspl(t, { plan: 1 })
+
+  const data = Buffer.from('abc傳def')
+  const server = createServer({ joinDuplicateHeaders: true }, (req, res) => {
+    res.write(data.subarray(0, 5))
+    setTimeout(() => {
+      res.end(data.subarray(5))
+    }, 100)
+  })
+  after(server.close.bind(server))
+
+  server.listen(0, async () => {
+    const client = new Client(`http://localhost:${server.address().port}`)
+    after(client.destroy.bind(client))
+
+    const { body } = await client.request({
+      path: '/',
+      method: 'GET'
+    })
+    body.setEncoding('utf8')
+
+    let text = ''
+    for await (const chunk of body) {
+      text += chunk
+    }
+
+    t.deepStrictEqual(text, data.toString('utf8'))
+  })
+
+  await t.completed
+})
+
 test('request multibyte text with setEncoding', async (t) => {
   t = tspl(t, { plan: 1 })
 


metcoder95 (approved):

mcollina (changes requested):
I'm relatively certain that Readable implements this logic. I think the problem is in the override of setEncoding, we should also call super.setEncoding

lib/api/readable.js
```diff
@@ -161,11 +174,39 @@ class BodyReadable extends Readable {
...
     }
 
     return super.push(chunk)
   }
 
+  /**
+   * @param {number} [size]
+   * @returns {Buffer|string|null}
+   */
+  read (size) {
```
metcoder95: Have you tried without overriding the `read` method?

metcoder95 (commented):
also PR has conflicts

lib/api/readable.js
pupuking723: Yes. After rebasing onto the latest main, I dropped the custom `read()` override and kept the implementation on top of the existing `Readable.setEncoding()` path. The PR now only adds the async iteration regression test.

pupuking723:
Thanks for the review. After rebasing, `BodyReadable#setEncoding()` now delegates to `super.setEncoding()` in the current implementation, so this PR only keeps the focused regression test covering async iteration after `body.setEncoding("utf8")` across multibyte chunk boundaries.

The current GitHub checks are passing across the CI matrix. Could you please take another look when you have a chance?

pupuking723:
Rebased this on current main to clear the conflict. The branch now only adds the async-iteration regression coverage; the implementation path is already on main.\n\nVerified locally with:\n\n```\n./node_modules/.bin/borp --timeout 180000 --expose-gc -p "test/client-request.js"\n```

pupuking723:
The branch now only carries the focused regression coverage; `BodyReadable#setEncoding()` already delegates to `super.setEncoding()` on current main. All checks passed except the macOS Node 26 job, which timed out after 25 minutes. I do not have permission to rerun it. Could someone rerun that job and take another look at the updated PR?

PR #5415, opened 2026-06-11
perf(fetch): lazily initialize Request/Response internals

marco-ippolito:
I'm working on a HTTP server-side implementation in Nodejs.
Constructing a `Request`/`Response` eagerly allocates a few things
that can be lazily observed:
the body's `ReadableStream`, the `Headers` wrapper, an `AbortController`, and a
parsed `URL`. This PR defers each of those until it's actually accessed, with no
observable change to the Fetch API.

- **Lazy byte-sequence Response bodies.** `extractBody` gains a `lazyByteStream`
  flag, for buffered `string`/`Uint8Array` sources the `ReadableStream` is built
  on first observation instead of up front.

- **Lazy Request headers, URL, and signal.** A `Request` built from an inner
  request (`fromInnerRequest`) no longer eagerly creates its `Headers`,
  `AbortController`, `URL`, or body stream, each is created on first
  access.

No observable Fetch API changes or public API changes.

### DIFF
--- a/lib/web/cache/cache.js
+++ b/lib/web/cache/cache.js
@@ -4,9 +4,10 @@ const assert = require('node:assert')
 
 const { kConstruct } = require('../../core/symbols')
 const { urlEquals, getFieldValues } = require('./util')
-const { kEnumerableProperty, isDisturbed } = require('../../core/util')
+const { kEnumerableProperty } = require('../../core/util')
 const { webidl } = require('../webidl')
 const { cloneResponse, fromInnerResponse, getResponseState } = require('../fetch/response')
+const { isBodyStreamDisturbedOrLocked } = require('../fetch/body')
 const { Request, fromInnerRequest, getRequestState } = require('../fetch/request')
 const { fetching } = require('../fetch/index')
 const { urlIsHttpHttpsScheme, readAllBytes } = require('../fetch/util')
@@ -312,11 +313,9 @@ class Cache {
     }
 
     // 8.
-    // A lazily-materialized response body (e.g. `new Response('...')`) has a
-    // null stream until it is first observed; an unobserved body cannot be
-    // disturbed or locked, so treat a null stream as neither. The subsequent
-    // clone (step 9) materializes the stream before it is read in step 11.
-    if (innerResponse.body && innerResponse.body.stream != null && (isDisturbed(innerResponse.body.stream) || innerResponse.body.stream.locked)) {
+    // The subsequent clone (step 9) materializes a lazy body's stream before it
+    // is read in step 11, so an as-yet-unobserved (null) stream is fine here.
+    if (isBodyStreamDisturbedOrLocked(innerResponse.body)) {
       throw webidl.errors.exception({
         header: prefix,
         message: 'Response body is locked or disturbed'
--- a/lib/web/cache/cache.js
+++ b/lib/web/cache/cache.js
@@ -312,7 +312,11 @@ class Cache {
     }
 
     // 8.
-    if (innerResponse.body && (isDisturbed(innerResponse.body.stream) || innerResponse.body.stream.locked)) {
+    // A lazily-materialized response body (e.g. `new Response('...')`) has a
+    // null stream until it is first observed; an unobserved body cannot be
+    // disturbed or locked, so treat a null stream as neither. The subsequent
+    // clone (step 9) materializes the stream before it is read in step 11.
+    if (innerResponse.body && innerResponse.body.stream != null && (isDisturbed(innerResponse.body.stream) || innerResponse.body.stream.locked)) {
       throw webidl.errors.exception({
         header: prefix,
         message: 'Response body is locked or disturbed'
--- a/lib/web/fetch/body.js
+++ b/lib/web/fetch/body.js
@@ -37,21 +37,21 @@ const streamRegistry = new FinalizationRegistry((weakRef) => {
  *
  * @see https://fetch.spec.whatwg.org/#concept-bodyinit-extract
  */
-function extractBody (object, keepalive = false) {
+// When `lazyByteStream` is true, byte-sequence sources (string / Uint8Array)
+// are wrapped in a lazy body whose ReadableStream is only built when the body
+// is read or its `body` getter is observed — see `makeLazyByteStreamBody`.
+// This avoids constructing a ReadableStream (and its controller) for the
+// common case where a Request/Response body is consumed directly (e.g. via
+// `.text()`) and the stream is never touched.
+function extractBody (object, keepalive = false, lazyByteStream = false) {
   // 1. Let stream be null.
   let stream = null
   let controller = null
 
-  // 2. If object is a ReadableStream object, then set stream to object.
-  if (webidl.is.ReadableStream(object)) {
-    stream = object
-  } else if (webidl.is.Blob(object)) {
-    // 3. Otherwise, if object is a Blob object, set stream to the
-    //    result of running object’s get stream.
-    stream = object.stream()
-  } else {
-    // 4. Otherwise, set stream to a new ReadableStream object, and set
-    //    up stream with byte reading support.
+  // Step 4 (otherwise: set up a new byte ReadableStream) is deferred behind
+  // this helper so the stream is only created once we know it's needed. For a
+  // lazy byte stream we skip it entirely and hand the source to the body.
+  const makeStream = () => {
     stream = new ReadableStream({
       pull () {},
       start (c) {
@@ -60,10 +60,22 @@ function extractBody (object, keepalive = false) {
       cancel () {},
       type: 'bytes'
     })
+    return stream
+  }
+
+  // 2. If object is a ReadableStream object, then set stream to object.
+  if (webidl.is.ReadableStream(object)) {
+    stream = object
+  } else if (webidl.is.Blob(object)) {
+    // 3. Otherwise, if object is a Blob object, set stream to the
+    //    result of running object’s get stream.
+    stream = object.stream()
   }
 
   // 5. Assert: stream is a ReadableStream object.
-  assert(webidl.is.ReadableStream(stream))
+  // `stream` may legitimately be null here: for non-stream/non-Blob sources we
+  // no longer eagerly create it (see makeStream above), so only assert when set.
+  assert(stream === null || webidl.is.ReadableStream(stream))
 
   // 6. Let action be null.
   let action = null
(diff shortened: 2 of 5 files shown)


KhafraDev:
I would never approve this

KhafraDev:
https://github.com/honojs/node-server/releases/tag/v2.0.0

marco-ippolito:
> I would never approve this

Can you please elaborate?
Thanks in advance for the explanation

mcollina:
@KhafraDev do you know how they achieve that? Bypassing all spec and monkeypatching all globals: https://github.com/honojs/node-server/blob/main/README.md#overrideglobalobjects.

Basically, if you use Hono, you'll have global `Request` and `Response` objects that are no longer spec-compliant.

marco-ippolito:
> @KhafraDev do you know how they achieve that? Bypassing all spec and monkeypatching all globals: https://github.com/honojs/node-server/blob/main/README.md#overrideglobalobjects.
> 
> Basically, if you use Hono, you'll have global `Request` and `Response` objects that are no longer spec-compliant.

I was about to say:

https://github.com/honojs/node-server/blob/9e1cdee836baeb13f33e857a582b29546185c324/src/request.ts#L529

They are monkeypatching it

KhafraDev:
It's only partially relevant, the rest of the release dives into other approaches they tried and why they didn't work, including the one here (somewhat). You still have the overhead of converting an IncomingMessage -> webstream -> [json/text/formData/blob/arraybuffer]. You still have webidl overhead. You still have webstream overhead. All this does is hide it while simultaneously making fetch a lot harder to maintain in the future.

It's a web spec. I'm not sure why everyone believes it must conform to every node use-case. It's not the right fit for an http server, even if people [believe they want it](https://x.com/matteocollina/status/1537466032490700802). It's why hono and srvx (h3, nitro) patch globals and bypass Request/Response construction directly when possible.

> Can you please elaborate?

node core implements greater than a dozen web specs and only 1 of them has active maintainers (URL/URLSearchParams). The simple reason is that you need strict guidelines on what gets accepted or else each spec becomes harder and harder to maintain until it eventually cannot. Look at EventTarget in node: it's a relatively simple spec you could implement in an afternoon, but coupled with all of the EventListener interop and weird hacks (ie. some part of node core overrides the functionality that [handles adding event listeners for some reason](https://github.com/nodejs/node/blob/da00166a6877816eb35bee0314c5e720ebcea44d/lib/internal/worker/io.js#L223-L227)) (yes second parenthesis, the weird [EventTarget -> node EventListener class which implements .on(), .off() etc.](https://github.com/nodejs/node/blob/da00166a6877816eb35bee0314c5e720ebcea44d/lib/internal/event_target.js#L897)) it is essentially impossible to make large changes to it. And that's not even for a living spec! The fetch spec is actively maintained and changes frequently! It's already hard enough to catch up with the spec, imagine if now I had to handle tracing and support for lazily created whatever and whatever else people wanted to include in fetch.

jasnell:
Relevant to this discussion: https://www.jasnell.me/posts/http-server-api-draft

There's no reason `Request` and `Response` won't work for the server side... Workers has used the model for years and it Just Works. 100% spec compliance with the client-side definition of these APIs is not necessary. Alternatively, WinterTC has discussed for a while now the idea of having a standardized Server-side profile for these that runtimes can adopt, which makes sense.

Bottom line is there are options and unifying around the APIs makes sense. What I do want to avoid, however, is Yet Another Runtime Specific API as that just makes the interop story far more difficult. I also want to make sure that we can cover h1, h2, h3, web transport, web sockets, extended connect, etc in a generalized way... thus my draft above.

And to be clear: I'm *NOT* suggesting that we use the generalized *client* side definitions of `Request` and `Response` on the server. The API surface, however, makes sense and works even if the underlying implementation details differs and even if (especially if?) those are not fully Fetch spec compliant given that fetch is written 100% in terms of client side browser use cases.

KhafraDev:
Right now I'm on very little sleep, so please correct me if I'm wrong.

> There's no reason Request and Response won't work for the server side... Workers has used the model for years and it Just Works.

Going off of the post linked and "Fetch Is Not Enough":
- trailers
- informational responses
- extended CONNECT
- HTTP Datagrams
- priority (fetch does have priority, unless you're referring to something different)

In workers it also doesn't "Just Work". Cloudflare Responses have a cf property and ResponseInit implements a "cf", "webSocket", and "encodeBody" options.

> 100% spec compliance with the client-side definition of these APIs is not necessary.

It is in undici and it should be a goal for node core specs. The real issue is that we're trying to shoehorn the browser client Request/Response models into a server model where it evidently doesn't fit. You mentioned reasons why this model does not work in the linked post.

In the post linked, it also mentions that:
```
This specification defines a server-side API that:

    Uses standard Request and Response without modification.

    [...]

    2. Standard Fetch types without extension: The Request a handler receives is a Request. The Response a handler returns is a Response. No duck-typing, no structural compatibility concerns, no server-specific subtypes that almost-but-not-quite match the standard types.
```

> Bottom line is there are options and unifying around the APIs makes sense. What I do want to avoid, however, is Yet Another Runtime Specific API as that just makes the interop story far more difficult. I also want to make sure that we can cover h1, h2, h3, web transport, web sockets, extended connect, etc in a generalized way... thus my draft above.

From the post:
```
Introduces a ServerContext for connection metadata, server capabilities, and lifecycle — the things that belong to the processing environment, not the HTTP message.
```

The ecosystem has shown that it's doing a pretty good job at doing this already. There are multiple libraries - elysia, hono, srvx - that already implement this type of server api and, afaik, work in the "big 3" runtimes.

> And to be clear: I'm NOT suggesting that we use the generalized client side definitions of Request and Response on the server. The API surface, however, makes sense and works even if the underlying implementation details differs and even if (especially if?) those are not fully Fetch spec compliant given that fetch is written 100% in terms of client side browser use cases.

I agree mostly with the first half, but I don't think it should be a fetch Request or Response at all. IMO the implementation should be as simple as IncomingMessage (Request) and OutgoingMessage/ServerResponse (Response) implementing a few getters and the body mixin methods (excluding .formData). Then you don't need a new server API and every runtime that has node compat already works without implementing a 3rd+ server.

With the latter half, why use the WHATWG Request/Response at all if they are limited? All this says to me is that Deno's server and bun's server and worse-off and node is going to try to fit right in there.

jasnell:
> Going off of the post linked and "Fetch Is Not Enough" ....
> ... In workers it also doesn't "Just Work". Cloudflare Responses have a cf property and ResponseInit implements a "cf", "webSocket", and "encodeBody" options.

It's not clear what point you're trying to make here. No one has suggested that the fetch APIs meet *all* of the needs, just that it works. Yes, there are feature gaps but those aren't relevant to this PR.

>> 100% spec compliance with the client-side definition of these APIs is not necessary.
> ... It is in undici and it should be a goal for node core specs...

Is undici planning to implement CORS? I certainly know Node.js isn't planning on it. Note that I was very specifically talking about 100% spec compliance with the fetch API, not the protocol.

> ... You mentioned reasons why this model does not work in the linked post.

No, I mentioned that there are feature gaps. Nothing I wrote is a reason why the model "does not work". It very clearly works for the limited subset the API was designed to model.

> ... With the latter half, why use the WHATWG Request/Response at all if they are limited? ...

Obviously there's a certain amount of subjective point of view in this. The ecosystem is quite clearly moving in the direction of a single, coherent set of APIs.

lib/web/cache/cache.js
```diff
@@ -312,7 +312,11 @@ class Cache {
     }
 
     // 8.
-    if (innerResponse.body && (isDisturbed(innerResponse.body.stream) || innerResponse.body.stream.locked)) {
+    // A lazily-materialized response body (e.g. `new Response('...')`) has a
+    // null stream until it is first observed; an unobserved body cannot be
+    // disturbed or locked, so treat a null stream as neither. The subsequent
+    // clone (step 9) materializes the stream before it is read in step 11.
+    if (innerResponse.body && innerResponse.body.stream != null && (isDisturbed(innerResponse.body.stream) || innerResponse.body.stream.locked)) {
```
jasnell: Nit: this condition is complex enough that moving it into a utility function would probably improve readability.

lib/web/fetch/body.js
```diff
@@ -273,13 +299,58 @@ function safelyExtractBody (object, keepalive = false) {
...
+      cancel () {},
+      type: 'bytes'
+    })
+
+    // Enqueue the whole buffered source and close on a microtask, mirroring the
+    // eager byte-stream path in extractBody so observable behaviour is identical.
+    if (source.length) {
+      controller.enqueue(typeof source === 'string' ? textEncoder.encode(source) : new Uint8Array(source))
+    }
+    queueMicrotask(() => readableStreamClose(controller))
```
jasnell, line 333: If you haven't done so already, it would be worth experimenting to see if `ReadableStream.from([source])` wouldn't be better here... We can likely work on improving the performance of that path significantly to elide quite a bit.

KhafraDev:
> It's not clear what point you're trying to make here. No one has suggested that the fetch APIs meet all of the needs, just that it works. Yes, there are feature gaps but those aren't relevant to this PR.

This PR is part of a future attempt at an http server that takes in a whatwg Request and returns a Response, I think my points are relevant still.

> Is undici planning to implement CORS? I certainly know Node.js isn't planning on it. Note that I was very specifically talking about 100% spec compliance with the fetch API, not the protocol.

Not even browsers are 100% spec compliant with the API. We still aim to be 100% compliant in what we can reasonably implement.

> No, I mentioned that there are feature gaps. Nothing I wrote is a reason why the model "does not work". It very clearly works for the limited subset the API was designed to model.

I don't understand the difference nor did I mean there was a difference between being "limited" and "not working". It works in simple cases, but is "limited"/"does not work" for WebSocket, extended connect, http3/pseudo headers (? not sure), etc. Why pick a model so limited instead of a plain object or just the IncomingMessage/OutgoingMessage objects with a few body mixins on them?

> The ecosystem is quite clearly moving in the direction of a single, coherent set of APIs.

I agree, I just don't understand why it needs to be part of node core. It's not a good model just because others do it. It looks nice, it functions fine even, until you hit one of the many limitations.

KhafraDev (dismissed):
I am going to leave this just so it doesn't land without far more discussion. I am entirely unconvinced that end goal of this PR (the Request -> Response http server) is a good fit for node.

jasnell:
@marco-ippolito @mcollina ... For the most part this change looks safe from a changing-observable-behavior point of view, but it's not entirely free from side effects.

1. There's a minor change in microtask scheduling. The close microtask in the case `const r = new Response('hello')` appears to get scheduled later due to the lazy `.body` access. This really shouldn't be an issue in practice but I've seen developers do terrible things that assume specific timings. I'd say the difference is at least worth noting. If this change doesn't trip any WPT's, it's likely not worth worrying about.

2. There's a *potential* bug with `Response.bodyUsed` in the embedder hooks. If an embedder (like Node.js) uses a `consumeArrayBuffer` hook on a `Response`, `bodyUsed` would return `false` after consumption. Likely worth a look.

marco-ippolito:
> There's a potential bug with Response.bodyUsed in the embedder hooks. If an embedder (like Node.js) uses a consumeArrayBuffer hook on a Response, bodyUsed would return false after consumption. Likely worth a look.

You were right good eyes 👀 

I'll add a bunch more tests

jasnell:
@marco-ippolito .. are you still pursuing this change set?

PR #5418, opened 2026-06-12
align EventSource with spec

KhafraDev:
Claude, find a bug in undici then create an issue to waste maintainer's time

fixes https://github.com/nodejs/undici/issues/5378


KhafraDev:
(afaik there's no discernible behavior differences, so can't really write a test)

mcollina:
if there is no difference why should we land this?

KhafraDev:
so we don't get more ai created issues that mention this as a bug

mcollina (approved):
lgtm

PR #5422, opened 2026-06-13
Cherry-pick 91678cbbce and 7a2db575e0

lpinca:
Cherry-pick 91678cbbce and 7a2db575e0 from https://github.com/nodejs/undici-ghsa-vxpw-j846-p89q/pull/1.

These should be backported together with 32dbf0b3dd93daf6d87badf506d52e89f2e20e78.

cc: @UlisesGascon

## Status

- [x] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready


lpinca:
@KhafraDev I guess the 👎 is because you do not think this is relevant for a client, and I might agree, but the problem is of the same type as the zip bomb, so I think both should be fixed or neither.

KhafraDev:
I don't think zip bombs were a vulnerability either (anything regarding an "attacker server"). I'm not blocking it, I would have preferred if both were disabled by default though, going back to ljharb's original argument against adding fetch in node: "How can we add this when it doesn’t comply with the browser spec ... ?"

I've never been against people using ws instead of undici's WebSocket/the global WebSocket which provides the same mitigations with a lot more customization.

lpinca:
> anything regarding an "attacker server"

I understand, it does not make much sense.

KhafraDev (approved):

mcollina (approved):
lgtm

PR #5425, opened 2026-06-14
fix: add missing spaces and tabs to HTTP tspecials validation

shivarm:
Correct the isTSpecial function to fully comply with the RFC 2616 specification by including space `(0x20)` and horizontal tab `(0x09)`. Also fixed an incorrect comment tracking the `=` character.

### DIFF
--- a/lib/web/fetch/formdata-parser.js
+++ b/lib/web/fetch/formdata-parser.js
@@ -538,15 +538,16 @@ function isCTL (char) {
 }
 
 /**
- * tspecials :=  "(" / ")" / "<" / ">" / "@" /
- *                "," / ";" / ":" / "\" / <">
- *                "/" / "[" / "]" / "?" / "="
- *                ; Must be in quoted-string,
- *                ; to use within parameter values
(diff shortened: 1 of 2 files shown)


KhafraDev:
Please rebase and add a test

KhafraDev (changes requested):
The comments above isTSpecial and isToken do not match the implementation.

```

     token := 1*<any (US-ASCII) CHAR except SPACE, CTLs,
                 or tspecials>

     tspecials :=  "(" / ")" / "<" / ">" / "@" /
                   "," / ";" / ":" / "\" / <">
                   "/" / "[" / "]" / "?" / "="
                   ; Must be in quoted-string,
                   ; to use within parameter values
```

KhafraDev:
I meant that tspecial and token is defined in RFC 2045 https://datatracker.ietf.org/doc/html/rfc2045

KhafraDev:
RFC 2616 makes no mention of tspecials and is obsolete. Please don't use an ai to think for you, verify whatever it comes up with next time.

PR #5436, opened 2026-06-17
feat: add generic type to `json` method

rtritto:
### Features

Add support of generic type for `json` method of `fetch` and `request`.

### Example

```ts
const response = await request('URL')
const json = await response.body.json<{ data: 'myDataType' }>()

// ---

const response = await fetch('URL')
const json = await response.json<{ data: 'myDataType' }>()
```

## Status

- [x] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [x] Merge ready

### DIFF
--- a/types/fetch.d.ts
+++ b/types/fetch.d.ts
@@ -55,7 +55,7 @@ export class BodyMixin {
    * ```
    */
   readonly formData: () => Promise<FormData>
-  readonly json: () => Promise<unknown>
+  readonly json: <T = unknown>() => Promise<T>
   readonly text: () => Promise<string>
   readonly textStream: () => ReadableStream<string>
 }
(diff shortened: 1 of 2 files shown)


Zamralik:
The `json` method do not provide a mean to validate the resulting value and it cannot guarantee what the received data are.
It is equivalent to using `as`, but passing the burden of "lying" about the type to the `json` method.

```ts
const json = await response.body.json() as { data: 'myDataType' };
```

Worse, it would implicitly cause the type to be erroneously "lied" about when used this way:
```ts
const json: { data: 'myDataType' } = await response.body.json();
```
As typescript implicitly fills the generic to `.json<{ data: 'myDataType' }()`

It may lead to implicit bugs and vulnerabilities from assuming the incoming data match the expected type.

KhafraDev:
similar PRs have been rejected, you can find the reason we declined this in those PRs

PR #5439, opened 2026-06-18
update accept-encoding header in fetch

### LINKED ISSUE #5437: zstd in fetch
## This would solve...

Zstd is superior compression to gzip/deflate/br and should be used if available. `fetch` in firefox or chrome uses zstd since 2024.

If nodejs has zstd implemented, then `zstd` should be added to [accept encoding header of request](https://github.com/nodejs/undici/blob/7ad38769d4f074df0227519ec398f30036c75291/lib/web/fetch/index.js#L1561).

Decompression of response [seems to be already implemented](https://github.com/nodejs/undici/blob/7ad38769d4f074df0227519ec398f30036c75291/lib/web/fetch/index.js#L2300).

KhafraDev:
fixes https://github.com/nodejs/undici/issues/5437
refs https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Accept-Encoding#default_accept-encoding_values


trivikr (approved):

mcollina (approved):
lgtm

PR #5440, opened 2026-06-19
fix(h2): guard onResponse against a 'response' event delivered after completion

staylor:
## Bug

The HTTP/2 client's `onResponse` handler in `lib/dispatcher/client-h2.js` calls `request.onResponseStart(...)` after guarding only `request.aborted` — but **not** `request.completed`:

```js
function onResponse (headers) {
  ...
  if (request.aborted) {            // ← missing `|| request.completed`
    releaseRequestStream(stream)
    return
  }
  if (request.onResponseStart(Number(statusCode), headers, stream.resume.bind(stream), '') === false) {
    stream.pause()
  }
  ...
}
```

`Request.onResponseStart` asserts both invariants:

```js
onResponseStart (statusCode, headers, resume, statusText) {
  assert(!this.aborted)
  assert(!this.completed)
  ...
}
```

So if a `'response'` event is delivered to the stream **after the request has already completed**, `onResponse` invokes `onResponseStart` post-completion and trips `assert(!this.completed)`. Because the assertion throws on the http2 stream's event tick (not inside any caller's `try`), it escapes as an uncaught exception and **crashes the process**.

The sibling handlers already treat a post-completion delivery as expected and guard against it:

```js
// onEnd
if (state.responseReceived) {
  if (!request.aborted && !request.completed) { state.pendingEnd = true }
}

// onTrailers
if (request.aborted || request.completed) { return }
```

`onResponse` is the only one of the three missing the `completed` guard — that asymmetry is the bug.

## Fix

Add the same `request.completed` guard the other handlers use:

```diff
-  if (request.aborted) {
+  if (request.aborted || request.completed) {
     releaseRequestStream(stream)
     return
   }
```

## How it shows up

We hit this in production on `undici@7.22.0` (same defect there: the v7 `'response'` handler guards only `request.aborted`, and the method was named `onHeaders`, which also `assert(!this.completed)`). The crash:

```
AssertionError [ERR_ASSERTION]: assert(!this.completed)
  at Request.onHeaders (undici/lib/core/request.js)
  at ClientHttp2Stream.<anonymous> (undici/lib/dispatcher/client-h2.js)  // the 'response' handler
  at ... node:internal/http2/core
```

It reproduces under load when many requests are multiplexed over **shared, long-lived h2 sessions** (HTTP/2 connection coalescing to a CDN edge), where server-initiated `GOAWAY` / refused-stream churn causes frequent concurrent stream teardowns — i.e. exactly the completion-vs-`'response'` race this guard handles. On the default per-origin HTTP/1.1 path the `client-h2.js` code is never exercised, so it only manifests with `allowH2: true`.

## Note on a test

This is a client-side event-ordering race; I couldn't construct a deterministic, non-flaky regression test against a real `http2` server without contriving the stream's event emission. The change mirrors the existing `onEnd`/`onTrailers` guards exactly. Happy to add a test if a maintainer can point me at the preferred way to drive a post-completion `'response'` on the request stream.

This affects `main` (v8.5.0) and `v7.x` — flagging in case a backport is wanted.

### DIFF
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -1144,10 +1144,13 @@ function onResponse (headers) {
 
   // Due to the stream nature, it is possible we face a race condition
   // where the stream has been assigned, but the request has been aborted
-  // the request remains in-flight and headers hasn't been received yet
-  // for those scenarios, best effort is to destroy the stream immediately
-  // as there's no value to keep it open.
-  if (request.aborted) {
+  // or already completed and headers hasn't been received yet. A late
+  // 'response' delivered after completion would call request.onResponseStart
+  // post-completion, tripping its `assert(!this.completed)` (an uncatchable
+  // throw on the http2 event tick). Guard `completed` here as onEnd/onTrailers
+  // already do; best effort is to release the stream immediately as there's
+  // no value to keep it open.
+  if (request.aborted || request.completed) {
     releaseRequestStream(stream)
     return
   }
--- a/test/http2-response-after-completion.js
+++ b/test/http2-response-after-completion.js
@@ -1,0 +1,186 @@
+'use strict'
+
+// Regression test for nodejs/undici#5440.
+//
+// The HTTP/2 'response' handler (onResponse in client-h2.js) only guarded
+// request.aborted before calling request.onResponseStart, while its sibling
+// handlers onEnd/onTrailers also guard request.completed. A 'response' frame
+// delivered to a still-live stream *after* the request has completed (a
+// stream-teardown race seen under load on shared h2 sessions with GOAWAY /
+// refused-stream churn) therefore called onResponseStart post-completion,
+// tripping its assert(!this.completed). Because that throws on the http2
+// stream's event tick — outside any caller's try — it escaped as an uncaught
+// exception and crashed the process.
+//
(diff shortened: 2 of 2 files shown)


ayana-s (approved):
i'm assuming you need a maintainer's approval? But approving in case this just requires anyone's approval

metcoder95 (commented):
can you add a test for that?

staylor:
Added a regression test in `test/http2-response-after-completion.js` (`63439bbd`).

As noted, a *real* post-completion `'response'` only arises from Node's internal http2 event-ordering race (a frame buffered before teardown, delivered to a still-live stream after the request completed), so it can't be driven deterministically through a real server. The test instead uses the same white-box harness as `test/http2-late-data.js` — `connectH2` over a `FakeSocket`/`FakeSession`/`FakeStream` with a real `Request` — and emits a `'response'` event on the stream after the request has reached `completed`, driving the actual `onResponse` handler.

It's a true RED/GREEN test: against the pre-fix handler it fails with the exact `assert(!this.completed)` AssertionError, and it passes with the `request.completed` guard. It asserts that `onResponse` runs (`onResponseStarted` fires before the guard), that the post-completion `onResponseStart` is skipped, and that the stream is released (`releaseRequestStream` removes its listeners).

metcoder95 (approved):

PR #5441, opened 2026-06-19
fix(proxy): fail the request when the CONNECT tunnel drops instead of looping

### LINKED ISSUE #3897: ProxyAgent loop on SocketError
## Bug Description

When the connection to the proxy server is closed with `UND_ERR_SOCKET` using `ProxyAgent` an infinite loop is created.
This is actually a double bug; I found this while investigating why `ProxyAgent` doesn't work with a basic http proxy built with `node:http`:
```js
import { ok } from "node:assert";
import { createServer, request as httpRequest } from "node:http";
import { request as httpsRequest } from "node:https";
import { Client, ProxyAgent, request } from "undici";

    ok(req.url);
    const targetUrl = new URL(req.url);
    req.pipe(
        (targetUrl.protocol === "https:" ? httpsRequest : httpRequest)(
            {
                method: req.method,
                hostname: targetUrl.hostname,
                port: targetUrl.port,
                path: targetUrl.pathname + targetUrl.search,
                headers: req.headers,
            },
            (proxiedRes) => {
                ok(proxiedRes.statusCode);
                res.writeHead(
                    proxiedRes.statusCode,
                    proxiedRes.statusMessage,
                    proxiedRes.headers,
                );
                proxiedRes.pipe(res);
            },
        ).once("error", res.destroy.bind(res)),
    );
}).listen();
const address = server.address();
ok(address && typeof address === "object");

// This works fine:
console.log(
    await new Client(`http://localhost:${address.port}`)
        .request({
            method: "GET",
            path: "http://example.com",
            headers: { host: "example.com" },
        })
        .then((res) => res.body.text()),
);

// This doesn't work and creates an infinite loop trying to connect to proxy url
// For instance, listening to server.on("connection", console.log) logs infinite times
console.log(
    await request("http://example.com", {
        dispatcher: new ProxyAgent(`http://localhost:${address.port}`),
    }).then((res) => res.body.text()),
);
```

Now here there are two questions:
1. **Why is the connection closing in the first place?** Spent hours debugging but honestly couldn't find the reason
2. **Why is it looping until a connection is established?** This creates a significant overhead both on the client and on the proxy server that is flooded by requests

The error is caught here where there is already something about "avoiding a loop in client.js#connect" https://github.com/nodejs/undici/blob/1cfe0949053aac6267f11b919cee9315a27f1fd6/lib/dispatcher/proxy-agent.js#L97-L104

## Reproducible By

Example above

## Expected Behavior

If the connection is closed, the error should be thrown immediately instead of trying infinite times.
Also, I think the example above should be successful considering that using a `Client` works

## Logs & Screenshots

The error (added `throw err` to avoid loop)
```shell
SocketError: other side closed
    at Socket.onHttpSocketEnd (D:\-\node_modules\undici\lib\dispatcher\client-h1.js:902:22)
    at Socket.emit (node:events:530:35)
    at endReadableNT (node:internal/streams/readable:1698:12)
    at process.processTicksAndRejections (node:internal/process/task_queues:90:21) {
  code: 'UND_ERR_SOCKET',
  socket: {
    localAddress: '::1',
    localPort: 60534,
    remoteAddress: '::1',
    remotePort: 60533,
    remoteFamily: 'IPv6',
    timeout: undefined,
    bytesWritten: 73,
    bytesRead: 0
  }
}
```

## Environment

Node v22, all undici versions with `ProxyAgent`

### Additional context

Using `ProxyAgent` with a "real" proxy works, so I'm not sure what is it expecting from my custom server that is different from a normal proxy

cesarvspr:
## This relates to...

Fixes #3897.

## Rationale

If a proxy accepts the TCP connection but tears it down while the `CONNECT` tunnel is still being set up (a proxy that doesn't implement `CONNECT`, or that rejects the upstream), the request never settles and undici reconnects in a tight loop, hammering the proxy.

The inner client rejects the `CONNECT` with `UND_ERR_SOCKET`. The `ProxyAgent` connect override only had a special case for `ERR_TLS_CERT_ALTNAME_INVALID` and passed everything else through untouched. On the client side, `onError` treats `UND_ERR_SOCKET` as a recoverable error on an established connection, so with no running request it leaves the request queued and `connect()` gets re-driven over and over.

The TLS branch already sidesteps this by wrapping its error in `SecureProxyConnectionError`, with a comment that says exactly why ("to avoid loop in client.js#connect"). The plain socket case just never got the same treatment.

## Changes

Wrap a socket failure during tunnel establishment in a new `ProxyConnectionError` (`UND_ERR_PRX_CONN`), so `onError` fails the request instead of retrying. The TLS branch and the fall-through for aborts and other errors are unchanged.

The fix is scoped to the proxy connect phase. It doesn't change how `onError` handles a socket drop on an already-established connection, which is a different case and stays recoverable.

### Features

N/A

### Bug Fixes

A proxy that drops the `CONNECT` tunnel now fails the request with `ProxyConnectionError` instead of reconnecting forever.

N/A. Adds a new `errors.ProxyConnectionError`.

## Status

- [x] Tested (`test/issue-3897.js`: green with the fix, hangs in a reconnect loop without it; full `proxy-agent` suite passes; lint and tsd clean)
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/docs/docs/api/Errors.md
+++ b/docs/docs/api/Errors.md
@@ -26,6 +26,7 @@ import { errors } from 'undici'
 | `InformationalError`                 | `UND_ERR_INFO`                        | expected error with reason                                                |
 | `ResponseExceededMaxSizeError`       | `UND_ERR_RES_EXCEEDED_MAX_SIZE`       | response body exceed the max size allowed                                 |
 | `SecureProxyConnectionError`         | `UND_ERR_PRX_TLS`                     | tls connection to a proxy failed                                          |
+| `ProxyConnectionError`               | `UND_ERR_PRX_CONN`                    | establishing the proxy `CONNECT` tunnel failed                            |
 | `MessageSizeExceededError`           | `UND_ERR_WS_MESSAGE_SIZE_EXCEEDED`    | WebSocket decompressed message exceeded the maximum allowed size          |
 | `AbortError`                         | `UND_ERR_ABORT`                       | the operation was aborted (base class of `RequestAbortedError`).          |
 | `RequestRetryError`                  | `UND_ERR_REQ_RETRY`                   | request failed and could not be retried; carries `statusCode`, `headers` and `data`. |
--- a/lib/core/errors.js
+++ b/lib/core/errors.js
@@ -403,6 +403,25 @@ class SecureProxyConnectionError extends UndiciError {
   }
 }
 
+const kProxyConnectionError = Symbol.for('undici.error.UND_ERR_PRX_CONN')
+class ProxyConnectionError extends UndiciError {
+  constructor (cause, message, options = {}) {
+    super(message, { cause, ...options })
+    this.name = 'ProxyConnectionError'
+    this.message = message || 'Proxy Connection failed'
+    this.code = 'UND_ERR_PRX_CONN'
+    this.cause = cause
+  }
+
+  static [Symbol.hasInstance] (instance) {
+    return instance && instance[kProxyConnectionError] === true
+  }
+
+  get [kProxyConnectionError] () {
+    return true
+  }
+}
+
 const kMaxOriginsReachedError = Symbol.for('undici.error.UND_ERR_MAX_ORIGINS_REACHED')
 class MaxOriginsReachedError extends UndiciError {
   constructor (message) {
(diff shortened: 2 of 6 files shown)


metcoder95 (approved):

mcollina (approved):
lgtm

PR #5443, opened 2026-06-19
fix: Improve `Dispatcher.compose` when nullish values are passed as argument

matthieusieben:
## This relates to...

N/A

## Rationale

The implementation of `Dispatcher.compose` allows nullish values:

https://github.com/nodejs/undici/blob/ae3efd442e9b5b76957c528ba6649e95775a3e1f/lib/dispatcher/dispatcher.js#L23-L25

This change alligns the type definitons with this to allow inline definitions:

```ts
dispatcher.compose(
  maxRedirections > 0
    ? interceptor.redirect({ maxRedirections })
    : null
)

```

## Changes

Just updated the type definition of the arguments of `Dispatcher.compose` from `Dispatcher.DispatcherComposeInterceptor[]` to `(null | undefined | Dispatcher.DispatcherComposeInterceptor)[]`

### Features

N/A

### Bug Fixes

N/A

N/A

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/dispatcher/dispatcher.js
+++ b/lib/dispatcher/dispatcher.js
@@ -16,14 +16,16 @@ class Dispatcher extends EventEmitter {
 
   compose (...args) {
     // So we handle [interceptor1, interceptor2] or interceptor1, interceptor2, ...
-    const interceptors = Array.isArray(args[0]) ? args[0] : args
+    const maybeInterceptors = Array.isArray(args[0]) ? args[0] : args
+    const interceptors = maybeInterceptors.filter(i => i != null)
+
+    if (interceptors.length === 0) {
+      return this
+    }
+
     let dispatch = this.dispatch.bind(this)
 
     for (const interceptor of interceptors) {
-      if (interceptor == null) {
-        continue
-      }
-
       if (typeof interceptor !== 'function') {
         throw new TypeError(`invalid interceptor, expected function received ${typeof interceptor}`)
       }
--- a/test/dispatcher.js
+++ b/test/dispatcher.js
@@ -22,7 +22,7 @@ test('dispatcher implementation', (t) => {
 })
 
 test('dispatcher.compose', (t) => {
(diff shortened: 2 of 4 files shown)


mcollina (changes requested):
can you update the type tests?

matthieusieben:
I've added tests, and also changed the implementation a bit to avoid returning a proxy object if no interceptor are actually provided

metcoder95 (commented):
(guilty of implementing that) I believe is better to just not accept `null`ish values at all, it does not provides any value and now makes it look a bit odd as anyways the null values do not do anything at all.

Happy to keep this PR as is to avoid breaking changes, but maybe set a warning that preempts this behaviour will change in further major.

matthieusieben:
It does provide some value when creating the dispatcher as it allows doing:

```ts
return dispatcher.compose(
  maxRedirections > 0
    ? interceptor.redirect({ maxRedirections })
    : null,
  maxRetries > 0
    ? interceptor.retry({ ... })
    : null
)
```

instead of:

```ts
import { DispatcherComposeInterceptor } from 'undici'

const interceptors: DispatcherComposeInterceptor[] = []

if (maxRedirections > 0) {
  interceptors.push(interceptor.redirect({ maxRedirections }))
}

if (maxRetries > 0) {
  interceptors.push(interceptor.retry({...}))
}

return dispatcher.compose(interceptors)
```

I find the latter too verbose, and is also requires to explicitly import `Interceptor` as a type.

metcoder95:
That's a good point, given your example, I believe then is more a DX thingy that something adding a runtime value. So I'm happy either way 👍

metcoder95 (approved):

matthieusieben:
@mcollina I believe I've adressed your comment. Let me know if I missed anything 🙏

PR #5448, opened 2026-06-22
fix(mock): native fetch bypasses MockAgent due to allowH2

### LINKED ISSUE #5036: MockAgent as global dispatcher broken since version 8.0.3

Megashubham:
Fixes #5036

### Description

This PR fixes an issue where native Node.js `fetch` queries would bypass the `MockAgent` globally registered using `setGlobalDispatcher`.

#### Root Cause
In version 8.0.3, a change was introduced to wrap the legacy global dispatcher with a `Dispatcher1Wrapper` (`fix: mirror the legacy global dispatcher for built-in fetch`). To support legacy consumers, `Dispatcher1Wrapper.dispatch` modifies `opts` to forcefully add `allowH2: false`.

Because of this, `native node fetch` passes `allowH2: false` when delegating to the `MockAgent`. When `MockAgent` proxies the request to its internal encapsulated `Agent` instance:
1. `MockAgent.get(origin)` pre-registers and creates a `MockClient` for the `origin` exactly as-is.
2. The encapsulated `Agent.dispatch` checks `allowH2: false` and modifies the lookup key to be `${origin}#http1-only`.
3. Because the `MockClient` registered is under `${origin}`, the cache miss causes the underlying `Agent` to instantiate a brand new real `Client` under `${origin}#http1-only`.
4. As a result, the request completely bypasses the mock interceptors and is fired over the real network!

#### Fix
The mock interceptor system shouldn't be concerned about whether `HTTP/2` or `HTTP/1.1` is being explicitly requested by the wrapper. To solve this, `MockAgent.dispatch` now safely deletes the `allowH2` flag from `dispatchOpts` if it is set to `false`. This ensures the underlying encapsulated `Agent` retains the standard origin as the lookup key and successfully resolves to the pre-registered `MockClient`.

A regression test using native `globalThis.fetch` has also been added in `test/issue-5036.js`.

### DIFF
--- a/lib/mock/mock-agent.js
+++ b/lib/mock/mock-agent.js
@@ -87,9 +87,7 @@ class MockAgent extends Dispatcher {
       dispatchOpts.path = `${path}?${normalizedSearchParams}`
     }
 
-    if (dispatchOpts.allowH2 === false) {
-      delete dispatchOpts.allowH2
-    }
+    delete dispatchOpts.allowH2
 
     return this[kAgent].dispatch(dispatchOpts, handler)
   }
--- a/lib/mock/mock-agent.js
+++ b/lib/mock/mock-agent.js
@@ -87,6 +87,10 @@ class MockAgent extends Dispatcher {
       dispatchOpts.path = `${path}?${normalizedSearchParams}`
     }
 
+    if (dispatchOpts.allowH2 === false) {
+      delete dispatchOpts.allowH2
+    }
+
     return this[kAgent].dispatch(dispatchOpts, handler)
   }
 
--- a/test/issue-5036.js
+++ b/test/issue-5036.js
@@ -1,0 +1,30 @@
+'use strict'
+
+const { test } = require('node:test')
+const assert = require('node:assert')
+const { MockAgent, setGlobalDispatcher } = require('../index.js')
+
+test('MockAgent works with native Node.js fetch', async (t) => {
+  // Only run if native fetch is available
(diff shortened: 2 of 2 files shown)


lib/mock/mock-agent.js
```diff
@@ -87,6 +87,10 @@ class MockAgent extends Dispatcher {
       dispatchOpts.path = `${path}?${normalizedSearchParams}`
     }
 
+    if (dispatchOpts.allowH2 === false) {
```
metcoder95: Maybe better to just delete it regardless of the value. For the Mock Agent, it is not important the version of the protocol but the protocol itself.

Megashubham: resolved !

Megashubham: @metcoder95

Megashubham:
Can you check once :P
@metcoder95

PR #5453, opened 2026-06-26
fix(h2): requeue request on GOAWAY'd session instead of crashing

staylor:
## Issue

When a request is dispatched onto an HTTP/2 session that has **already received a `GOAWAY` frame**, `ClientHttp2Session.request()` throws `ERR_HTTP2_GOAWAY_SESSION` ("New streams cannot be created after receiving a GOAWAY"). In `writeH2`'s `requestStream`, only `ERR_HTTP2_INVALID_SESSION` is handled gracefully; `GOAWAY` falls through to:

```js
session.destroy(wrappedErr)
util.destroy(session[kSocket], wrappedErr)   // <-
abort(wrappedErr)
```

The `util.destroy(socket, wrappedErr)` re-emits `'error'` on a socket whose `'error'` listener can already be detached during teardown, which surfaces as an **`uncaughtException` and crashes the process**:

```
Uncaught exception
  InformationalError: New streams cannot be created after receiving a GOAWAY
  code: UND_ERR_INFO, cause.code: ERR_HTTP2_GOAWAY_SESSION
    at requestStream (lib/dispatcher/client-h2.js)
    at writeH2 → _resume → resume → [dispatch] → Pool.dispatch
```

Observed in production on a high-volume `allowH2: true` client talking to servers that routinely send `GOAWAY` to rotate connections: under load a queued request is dispatched onto a session in the window after the `GOAWAY` frame is received but before `onHttp2SessionGoAway` retires the session, and the process crashes. Same family as the recently-hardened h2 teardown crashes (#5440, #4564).

## Fix

A `GOAWAY`'d session is the same situation as `ERR_HTTP2_INVALID_SESSION` — the peer won't accept new streams — so route it through the existing graceful branch (`resetHttp2Session` + `requeueUnsentRequest`). The request is **reset and requeued on a fresh connection** instead of failing and risking the unhandled socket `'error'`.

```diff
-      if (err?.code === 'ERR_HTTP2_INVALID_SESSION') {
+      if (err?.code === 'ERR_HTTP2_INVALID_SESSION' || err?.code === 'ERR_HTTP2_GOAWAY_SESSION') {
```

## Test

Added a regression test in `test/http2-goaway.js`, modeled on the existing `http2-invalid-session.js` pattern: it stubs `http2.connect` to return a fake session whose `request()` throws `ERR_HTTP2_GOAWAY_SESSION`, then asserts the request is requeued onto a fresh real session and succeeds (`connectCalls === 2`, `200`, `'ok'`). It **fails on `main`** (the throw escapes via the destroy+abort fallthrough) and **passes with this change**.

### DIFF
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -862,7 +862,10 @@ function openStream (client, request, session, abort, headers, options) {
   try {
     return session.request(headers, options)
   } catch (err) {
-    if (err?.code === 'ERR_HTTP2_INVALID_SESSION') {
+    // A GOAWAY'd session rejects new streams, same as an invalid session:
+    // reset and requeue on a fresh connection rather than the destroy + abort
+    // below, whose destroy(socket, err) can crash via an unhandled 'error'.
+    if (err?.code === 'ERR_HTTP2_INVALID_SESSION' || err?.code === 'ERR_HTTP2_GOAWAY_SESSION') {
       const wrappedErr = new SocketError(err.message, util.getSocketInfo(session[kSocket]))
       wrappedErr.cause = err
       session[kError] = wrappedErr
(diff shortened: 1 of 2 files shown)


metcoder95 (approved):

mcollina (approved):
lgtm

PR #5455, opened 2026-06-28
fix(fetch): avoid buffering streamed uploads in memory

### LINKED ISSUE #4058: Request stream contents are kept in memory when using fetch
## Bug Description

When streaming a file upload, the Node process' memory usage increases until it consumes an amount equal to the size of the file we are trying to upload.

## Reproducible By

Save the following code to a file, such as `streamUpload.js`:

```js
import fs from "node:fs";

if (process.argv.length !== 3) {
    console.error(`Usage: node ${process.argv[1]} FILE`);
    process.exit(2);
}

console.log('My PID:', process.pid);

const fileStream = fs.createReadStream(process.argv[2]);

const request = new Request("http://localhost:8000", {
    method: "PUT",
    headers: {
    },
    body: fileStream,
    duplex: 'half',
});

fetch(request)
.then(async response => {
    console.log(response);
    console.log(await response.text());

    await new Promise(resolve => process.nextTick(resolve));
    console.log('Waiting 15 seconds until exiting')
    await new Promise(resolve => setTimeout(resolve, 15000));

    process.exit(0);
}).catch(err => {
    console.error('Encountered error: ', err);
    process.exit(1);
});
```

Set up netcat as a mock HTTP server in a separate terminal session:

```sh
nc -Clq 0 8000
```

Then try running the Node script using a large file (should be > 100MB):

```sh
node streamUpload.js path/to/largeFile
```

The script prints out its PID. Use it to monitor the memory usage in yet a new terminal session:

```sh
top -p PID
```

If you wish to let the fetch call resolve, you need to write the following into the `nc` program when the upload is complete:

```
HTTP/1.1 200 OK

Hello from netcat!
```

Then hit Enter and CTRL-D.

## Expected Behavior

The memory usage should increase by the size of the buffer used between the file stream and the TCP writable stream, and not be affected by the size of the file. The memory usage should stay the same throughout, no matter how much has been streamed.

## Environment

I'm using Windows Subsystem for Linux, with Ubuntu 22.04.5 LTS, though I've seen this behaviour on regular Linux installs as well.
Node is installed with nvm, version 22.13.1.

### Additional context

I first observed this when following the instructions for how to stream a file upload (from #2202) using `FormData`. But I was able to reproduce it using plain streams when trying some workarounds.

As a consequence, we cannot use `fetch` to upload large files in environments where the files are too large to keep in memory.

cesarvspr:
## This relates to...

Fixes #4058.

## Rationale

Streaming a large upload with fetch (e.g. `body: fs.createReadStream(...)`, `duplex: 'half'`) ends up holding the whole file in memory.

fetch clones the request so it can follow a redirect, and cloning tees the body's stream: the original request keeps one branch and the wire sends the other. A stream body has a null source, so it can't be replayed across a redirect anyway. http-redirect-fetch returns a network error for a non-303 redirect with a null source, and on any other redirect the body is re-extracted from its source rather than from the teed branch. So the branch kept on the original request is never read, but the tee still buffers every chunk waiting for it, which is the entire upload.

The clone happens on every fetch with a body, not only when a redirect actually occurs, so this affected all streamed uploads.

## Changes

Skip the tee for null-source bodies in the internal redirect clone and reuse the stream directly. `cloneBody` takes a `forFetch` flag that only the clone in http-network-or-cache fetch sets. `Request.clone()` and `Response.clone()` don't pass it, so they still tee and keep two independently readable branches. Bodies with a non-null source are unchanged.

### Features

N/A

### Bug Fixes

A streamed request body is no longer buffered in memory.

N/A

## Status

- [x] Tested (`test/fetch/issue-4058.js`: a 128 MiB streamed PUT peaks at a few MiB with the fix and ~128 MiB without it; the clone and redirect suites pass; lint clean)
- [x] Documented
- [ ] In review
- [ ] Merge ready

## Notes

`cloneBody`/`cloneRequest` are also touched by #5415 (lazy Request/Response internals), so a textual conflict is expected and whichever lands second will need a small rebase. This is an independent change: it only affects the redirect-following clone of a null-source body, not the lazy-init refactor.

### DIFF
--- a/lib/web/fetch/body.js
+++ b/lib/web/fetch/body.js
@@ -273,10 +273,25 @@ function safelyExtractBody (object, keepalive = false) {
   return extractBody(object, keepalive)
 }
 
-function cloneBody (body) {
+function cloneBody (body, reuseUnreplayableBody = false) {
   // To clone a body body, run these steps:
 
   // https://fetch.spec.whatwg.org/#concept-body-clone
+
+  // A body with a null source (a stream) can't be replayed across a redirect:
+  // http-redirect-fetch network-errors a non-303 redirect with a null source and
+  // otherwise re-extracts from the source, never from the teed branch. When the
+  // caller only needs the clone to follow redirects (reuseUnreplayableBody),
+  // teeing such a body just buffers the whole upload for a branch that is never
+  // read (#4058), so reuse the stream directly. Request.clone()/Response.clone()
+  // pass false and always tee, keeping both branches independently readable.
+  if (reuseUnreplayableBody && body.source === null) {
+    return {
+      stream: body.stream,
+      length: body.length,
+      source: null
+    }
+  }
 
   // 1. Let « out1, out2 » be the result of teeing body’s stream.
   const { 0: out1, 1: out2 } = body.stream.tee()
--- a/lib/web/fetch/index.js
+++ b/lib/web/fetch/index.js
@@ -1431,7 +1431,7 @@ async function httpNetworkOrCacheFetch (
     // Otherwise:
 
     // 1. Set httpRequest to a clone of request.
-    httpRequest = cloneRequest(request)
+    httpRequest = cloneRequest(request, true)
 
     // 2. Set httpFetchParams to a copy of fetchParams.
     httpFetchParams = { ...fetchParams }
--- a/lib/web/fetch/request.js
+++ b/lib/web/fetch/request.js
@@ -958,16 +958,17 @@ function makeRequest (init) {
 }
 
 // https://fetch.spec.whatwg.org/#concept-request-clone
-function cloneRequest (request) {
+function cloneRequest (request, reuseUnreplayableBody = false) {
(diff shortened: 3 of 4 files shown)


mcollina (approved):
LGTM. I checked this against the Fetch Standard: the HTTP-network-or-cache fetch note explicitly encourages avoiding teeing when the request body source is null, and redirects/auth retries for such bodies either fail or do not need the retained tee branch. Request.clone()/Response.clone() behavior remains unchanged.

KhafraDev (changes requested):
If that is what the spec says, this is not the way to implement it

cesarvspr:
You're right. The note is on http-network-or-cache fetch, not on the clone algorithm, and that placement is the point: the shortcut is only safe for that one clone, so it has no business on cloneRequest/cloneBody. I put it in the wrong layer. #5456 special-cases it at the single call site and leaves the clone primitives alone, which is clearly the right shape.

Closing this in favor of #5456.

PR #5456, opened 2026-06-28
fetch: fix issue 4058

### LINKED ISSUE #4058: Request stream contents are kept in memory when using fetch

KhafraDev:
closes #5455
fixes #4058 (tbh didn't even bother reading the ai slop test so can't really confirm 🤷)

### DIFF
--- a/lib/web/fetch/index.js
+++ b/lib/web/fetch/index.js
@@ -1431,7 +1431,16 @@ async function httpNetworkOrCacheFetch (
     // Otherwise:
 
     // 1. Set httpRequest to a clone of request.
-    httpRequest = cloneRequest(request)
+    // Implementations are encouraged to avoid teeing request’s body’s stream
+    // when request’s body’s source is null as only a single body is needed in
+    // that case. E.g., when request’s body’s source is null, redirects and
+    // authentication will end up failing the fetch.
+    if (request.body?.source != null) {
+      httpRequest = cloneRequest(request)
+    } else {
+      httpRequest = cloneRequest({ ...request, body: null })
+      httpRequest.body = request.body
+    }
 
     // 2. Set httpFetchParams to a copy of fetchParams.
     httpFetchParams = { ...fetchParams }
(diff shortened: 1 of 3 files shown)


KhafraDev:
The fetch spec is so well written, unironically. This shortcut only works for this once instance of cloning the request, which is why the note was added in #http-network-or-cache-fetch and not #concept-request-clone.

mcollina (approved):
lgtm

PR #5459, opened 2026-06-29
feat: support HTTP QUERY method (RFC 10008)

### LINKED ISSUE #5454: Support for HTTP QUERY rfc10008
## This would solve...
Undici lacks HTTP QUERY [RFC 10008](https://datatracker.ietf.org/doc/html/rfc10008) QUERY method support, blocking safe/idempotent requests with bodies.  Yes, RFC10008 just became a proposed standard, so, understandable.  :-)

Add QUERY method support (like GET/POST): handle bodies, Content-Type, Accept-Query, caching, redirects.

## I have also considered...
POST workarounds; QUERY is the IETF standard.

## Additional context
[RFC 10008](https://datatracker.ietf.org/doc/html/rfc10008) (June 2026) standardizes safe QUERY for complex queries.

mcollina:
## Description

Adds support for the HTTP QUERY method standardized in [RFC 10008](https://datatracker.ietf.org/doc/html/rfc10008) (June 2026), enabling safe/idempotent requests with bodies.

## Changes

- **`lib/core/util.js`**: Added `QUERY` to normalized method records and `safeHTTPMethods`
- **`lib/core/request.js`**: QUERY is idempotent by default (safe to retry)
- **`lib/handler/redirect-handler.js`**: Comment clarifying QUERY is safe (like GET) and won't change method on 301/302 redirects
- **`lib/handler/retry-handler.js`**: QUERY added to default retryable methods
- **`lib/web/fetch/constants.js`**: QUERY added to `safeMethods` for the Fetch API

Closes #5454

### DIFF
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -225,7 +225,7 @@ class Request {
     this.protocol = getProtocolFromUrlString(origin)
 
     this.idempotent = idempotent == null
-      ? method === 'HEAD' || method === 'GET'
+      ? method === 'HEAD' || method === 'GET' || method === 'QUERY'
       : idempotent
 
     this.blocking = blocking ?? this.method !== 'HEAD'
--- a/lib/core/util.js
+++ b/lib/core/util.js
@@ -960,7 +960,9 @@ const normalizedMethodRecordsBase = {
   post: 'POST',
   POST: 'POST',
   put: 'PUT',
-  PUT: 'PUT'
+  PUT: 'PUT',
+  query: 'QUERY',
+  QUERY: 'QUERY'
 }
 
 const normalizedMethodRecords = {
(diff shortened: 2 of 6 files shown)


ronag (approved):

jeswr:
One correctness note before this lands: adding `QUERY` to `safeHTTPMethods` also lets it into the **cache**/**deduplicate** interceptors' `methods`, but their keys ignore the request body. A QUERY's response depends on its body (RFC 10008 §2.7), so `cache({ methods: ['GET','QUERY'] })` / `deduplicate({ methods: ['QUERY'] })` could serve the wrong response for a different body. As this is opt-in only (default `['GET']`), nothing breaks by default — but might be worth excluding QUERY there, or making the key body-aware. Redirects/retry/preflight look correct.

I'm creating issues/PRs for improved retries and caching on my fork (GenAI-assisted), I am happy to upstream those issues/PRs once I have reviewed them.

metcoder95 (approved):

PR #5461, opened 2026-06-29
fix(types): correct deleteCookie attributes

### LINKED ISSUE #5458: `deleteCookie` TypeScript attributes expose `name` instead of `path`
## Summary

`deleteCookie(headers, name, attrs)` supports `path` at runtime, and path matching is needed to delete a path-scoped cookie. However, the TypeScript declaration for the third argument exposes `{ name?: string, domain?: string }`, so TypeScript rejects `{ path: '/' }` while accepting `{ name: '...' }`, which is not consumed by the runtime attributes object.

## Steps to reproduce

```ts
import { Headers, deleteCookie } from 'undici'

const headers = new Headers()

deleteCookie(headers, 'session', { domain: 'example.com', path: '/' })

// @ts-expect-error `name` should not be a deleteCookie attribute.
deleteCookie(headers, 'session', { domain: 'example.com', name: 'session' })
```

Compile with strict TypeScript settings against Undici's declarations.

## Expected behavior

TypeScript should allow:

```ts
deleteCookie(headers, 'session', { domain: 'example.com', path: '/' })
```

and should reject `name` inside the attributes object, since the cookie name is already the second positional argument.

## Actual behavior

TypeScript reports:

```text
'path' does not exist in type '{ name?: string | undefined; domain?: string | undefined; }'
```

The `@ts-expect-error` on `name` is unused, meaning TypeScript currently accepts `name` as an attribute.

## Evidence

Runtime behavior accepts and emits `Path=/` for deletion:

```text
session=abc; Domain=example.com; Path=/, session=; Domain=example.com; Path=/; Expires=Thu, 01 Jan 1970 00:00:00 GMT
```

The drift appears to be between:

```ts
attributes?: { name?: string, domain?: string }
```

and the runtime/docs/tests behavior, which use `path` and `domain` for delete-cookie attributes.

## Suggested fix

```ts
attributes?: { path?: string, domain?: string }
```

It would also be useful to add a TypeScript/tsd coverage case that accepts `{ path: '/', domain: 'example.com' }` and rejects `{ name: 'session' }`.

## Additional context / Related coverage

A current GitHub search for related `deleteCookie path type`, `cookies.d.ts deleteCookie`, `deleteCookie attributes path domain`, and similar terms did not identify exact existing issue or PR coverage.

---
Submitted with Codex.

Ram-blip:
## This relates to...

Fixes #5458 

## Rationale

deleteCookie(headers, name, attributes) supports path and domain at runtime for deleting path-scoped cookies, but the TypeScript declaration currently exposes the attributes object as { name?: string, domain?: string }.

This causes TypeScript to reject valid usage with path, while accepting name even though the cookie name is already provided as the second positional argument.

## Changes

Updates the deleteCookie TypeScript declaration from:

{ name?: string, domain?: string }

to:

{ path?: string, domain?: string }

Adds tsd coverage to verify that:
- path and domain are accepted as delete-cookie attributes
- name is rejected inside the attributes object

### Features

N/A

### Bug Fixes

Fixes the deleteCookie TypeScript declaration so it matches runtime behavior and documentation.

N/A

## Status

- [X] Tested
- [ ] Documented
- [X] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/test/types/cookies.test-d.ts
+++ b/test/types/cookies.test-d.ts
@@ -1,0 +1,19 @@
+import { expectError, expectType } from 'tsd'
+import { Headers, deleteCookie } from '../..'
+
+const headers = new Headers()
+
+// `path` and `domain` should be accepted as deleteCookie attributes.
+expectType<void>(deleteCookie(headers, 'session', {
+  domain: 'example.com',
+  path: '/'
+}))
+
+expectType<void>(deleteCookie(headers, 'session', { path: '/' }))
+expectType<void>(deleteCookie(headers, 'session', { domain: 'example.com' }))
+
+// `name` should not be accepted because the cookie name is the second positional argument.
+expectError(deleteCookie(headers, 'session', {
+  domain: 'example.com',
+  name: 'session'
+}))
--- a/types/cookies.d.ts
+++ b/types/cookies.d.ts
@@ -18,7 +18,7 @@ export interface Cookie {
 export function deleteCookie (
   headers: Headers,
   name: string,
-  attributes?: { name?: string, domain?: string }
+  attributes?: { path?: string, domain?: string }
 ): void
 
 export function getCookies (headers: Headers): Record<string, string>


mcollina (approved):
lgtm

PR #5470, opened 2026-06-30
fix(h2): deliver an early final response to an Expect: 100-continue request

### LINKED ISSUE #5465: HTTP/2: request hangs when the server sends a final response to an Expect: 100-continue request
### Bug Description

Over HTTP/2, when a server answers a request carrying `Expect: 100-continue` with a **final response** (e.g. `401`) instead of `100 (Continue)`, the request never resolves. undici correctly withholds the request body, but the response is never delivered and the request hangs.

RFC 9110 [§10.1.1](https://www.rfc-editor.org/rfc/rfc9110.html#section-10.1.1) permits a server to send a final status determinable from the header fields — an "immediate success, redirect, or error response" — in place of `100 (Continue)`; the client must then deliver that response and not send the body. The common cases are an early `401`/`413`/`417` (and equally a `2xx`). undici delivers nothing.

This is HTTP/2-only; HTTP/1.x handles it.

**Root cause:** in `lib/dispatcher/client-h2.js`, the request-body write is bound to the stream's `'continue'` event. On an early final response, `'continue'` never fires, so the client never sends `END_STREAM`; the stream stays half-open, its `'close'` event never fires, and `onRequestStreamClose` (which finalizes the response) never runs.

### Reproducible By

Minimal repro: https://github.com/jeswr/undici-h2-expect-continue-bug (`npm install && npm start` → `FAIL: request hung ...` after 5s). Condensed:

```js
const http2 = require('node:http2')
const { Client } = require('undici')

// h2 server: answer 401 from the headers alone, no 100 (Continue)
const server = http2.createSecureServer({ key, cert })
server.on('stream', (stream) => {
  stream.respond({ ':status': 401 })
  stream.end('denied')
})
    allowH2: true, connect: { rejectUnauthorized: false }
  })
  // hangs — the 401 is never delivered
  const res = await client.request({
    path: '/', method: 'POST', body: 'x'.repeat(1024), expectContinue: true
  })
  console.log(res.statusCode)
})
```

### Expected Behavior

undici delivers the early final response (e.g. `401`) and completes the request without sending the body, as it does over HTTP/1.x and as RFC 9110 §10.1.1 requires.

### Environment

- undici 8.5.0
- Node.js v25.1.0
- HTTP/2 (`allowH2: true`); request with `expectContinue: true` and a body

I have a candidate fix (close the request stream's writable half on an early final response) with a test; happy to open a PR.

jeswr:
Fixes #5465.

## Problem

Over HTTP/2, when a server answers a request carrying `Expect: 100-continue` with a final response (e.g. `401`) instead of `100 (Continue)`, undici withholds the request body correctly but never delivers the response — the request hangs.

The body write is bound to the stream's `'continue'` event; on an early final response that never fires, so the client never sends `END_STREAM`. The h2 stream stays half-open, its `'close'` never fires, and `onRequestStreamClose` (which finalizes the response) never runs. RFC 9110 [§10.1.1](https://www.rfc-editor.org/rfc/rfc9110.html#section-10.1.1) permits a server to answer with a final status determinable from the header fields in place of `100 (Continue)`, so this is a client conformance bug. HTTP/2 only — HTTP/1.x already handles it.

## Fix

In `onResponse` (`lib/dispatcher/client-h2.js`): when a final response arrives and the body was never sent, remove the `continue` listener and `stream.end()` the writable half, so the stream completes and the response is delivered without sending the body. "Body not sent" is tracked with a new `state.bodySent`, set when `writeBodyH2` runs.

## Tests

New case in `test/http2-continue.js`: an early `401` to an `Expect: 100-continue` `POST` is delivered, and the server receives 0 body bytes. The existing h2 suite passes.

Minimal reproduction: https://github.com/jeswr/undici-h2-expect-continue-bug

### DIFF
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -1037,6 +1037,7 @@ function writeH2 (client, request) {
     headersTimeout,
     bodyTimeout,
     responseReceived: false,
+    bodySent: false,
     session,
     stream: null
   }
@@ -1136,6 +1137,13 @@ function onResponse (headers) {
 
   stream.off('response', onResponse)
 
+  // Final response received while still awaiting 100 (Continue): the body won't
+  // be sent, so close our half or the stream stays open and never completes.
+  if (state.body != null && !state.bodySent && !stream.writableEnded) {
+    stream.removeListener('continue', writeBodyH2)
+    stream.end()
+  }
+
   const statusCode = headers[HTTP2_HEADER_STATUS]
   delete headers[HTTP2_HEADER_STATUS]
   request.onResponseStarted()
@@ -1233,6 +1241,7 @@ function onTrailers (trailers) {
 function writeBodyH2 () {
   const stream = this
   const state = stream[kRequestStreamState]
+  state.bodySent = true
   const { abort, body, client, contentLength, expectsPayload, request } = state
 
   if (!body || contentLength === 0) {
--- a/test/http2-continue.js
+++ b/test/http2-continue.js
@@ -65,3 +65,46 @@ test('Should handle h2 continue', async t => {
 
   await t.completed
 })
+
+test('Should deliver an early final response to an Expect: 100-continue request without sending the body', async t => {
+  t = tspl(t, { plan: 4 })
+
+  let requestBodyBytes = 0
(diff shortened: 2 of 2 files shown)


mcollina (approved):
lgtm

metcoder95 (approved):

PR #5475, opened 2026-07-01
fix(eventsource): set use-URL-credentials flag

Ram-blip:
## This relates to...

Fixes EventSource potential-CORS request creation so it sets the correct use-URL-credentials flag.

## Rationale

The HTML potential-CORS request algorithm requires the returned request’s use-URL-credentials flag to be set.

`createPotentialCORSRequest()` previously passed `useCredentials: true` to `makeRequest()`, but Undici stores `useCredentials` and `useURLCredentials` as separate fields. The Fetch authentication logic consumes `useURLCredentials`, so the intended flag was not being set.

WebSocket already uses `useURLCredentials: true` for the same internal request flag.

Specification: https://html.spec.whatwg.org/multipage/urls-and-fetching.html#create-a-potential-cors-request

## Changes

Changed `useCredentials: true` to `useURLCredentials: true` in `createPotentialCORSRequest()`.

Added unit coverage for both EventSource CORS states:

- `anonymous`: `cors` mode with `same-origin` credentials
- `use-credentials`: `cors` mode with `include` credentials

The test also verifies `useURLCredentials`, `destination`, `urlList`, and `url`.

### Features

N/A

### Bug Fixes

Fixes EventSource request creation so the use-URL-credentials flag is set on the internal Fetch request.

N/A

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

## Testing

```bash
node --test test/eventsource/util.js
npm run test:eventsource
npm run lint

### DIFF
--- a/lib/web/eventsource/util.js
+++ b/lib/web/eventsource/util.js
@@ -49,7 +49,7 @@ function createPotentialCORSRequest (url, destination, corsAttributeState, sameO
     destination,
     mode,
     credentials: credentialsMode,
-    useCredentials: true
+    useURLCredentials: true
   })
 }
 
--- a/test/eventsource/util.js
+++ b/test/eventsource/util.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const { test } = require('node:test')
-const { isASCIINumber, isValidLastEventId } = require('../../lib/web/eventsource/util')
+const { isASCIINumber, isValidLastEventId, createPotentialCORSRequest } = require('../../lib/web/eventsource/util')
 
 test('isValidLastEventId', (t) => {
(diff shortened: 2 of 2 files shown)


KhafraDev (changes requested):
the test is completely useless... what is it testing?

Ram-blip:
> the test is completely useless... what is it testing?

Thanks for the review. You’re right, the test was only checking that createPotentialCORSRequest() passed a property through to makeRequest(), so it was too close to the implementation and did not prove any observable behavior.

I added it because the change is in that helper, but I agree it does not add much value as a regression test. I’ve removed it and I’ll work on a better behavioral test.

PR #5487, opened 2026-07-02
fix: same-origin/origin-when-cross-origin referrer compares wrong origins

spokodev:
## Description

`determineRequestsReferrer` compares the wrong pair of origins for the `same-origin` and `origin-when-cross-origin` referrer policies, so a same-origin request is treated as cross-origin.

Both branches call `sameOrigin(request, referrerURL)`. `request` is the internal request record (it exposes `url`/`urlList`/`origin`), not a URL object, so it has none of the `protocol`/`hostname`/`port` fields `sameOrigin()` reads. The comparison therefore always returns false and the same-origin path is never taken.

Effect on a same-origin `fetch()`:
- `same-origin` sends no `Referer` at all (should send the full referrer URL).
- `origin-when-cross-origin` sends only the origin (should send the full referrer URL).

This diverges from the Fetch standard and from browser behavior, and breaks referrer-based checks and analytics on the server.

## Fix

Compare `referrerURL` against the request's current URL, matching the neighboring `strict-origin-when-cross-origin` branch (which is already correct and whose in-code comment describes exactly this same-origin check):

```diff
-      if (sameOrigin(request, referrerURL)) {
+      if (sameOrigin(referrerURL, requestCurrentURL(request))) {
```

(applied to both the `same-origin` and `origin-when-cross-origin` branches; no new imports.)

## Test

Added a `determineRequestsReferrer` group to `test/fetch/util.js`. On current HEAD the same-origin cases fail (`no-referrer` / origin-only) and the cross-origin cases pass as controls; with the fix all four pass. `test/fetch/referrrer-policy.js` stays green.

### DIFF
--- a/lib/web/fetch/util.js
+++ b/lib/web/fetch/util.js
@@ -481,7 +481,7 @@ function determineRequestsReferrer (request) {
     case 'same-origin':
       // 1. If the origin of referrerURL and the origin of request’s current
       // URL are the same, then return referrerURL.
-      if (sameOrigin(request, referrerURL)) {
+      if (sameOrigin(referrerURL, requestCurrentURL(request))) {
         return referrerURL
       }
       // 2. Return no referrer.
@@ -489,7 +489,7 @@ function determineRequestsReferrer (request) {
     case 'origin-when-cross-origin':
       // 1. If the origin of referrerURL and the origin of request’s current
       // URL are the same, then return referrerURL.
-      if (sameOrigin(request, referrerURL)) {
+      if (sameOrigin(referrerURL, requestCurrentURL(request))) {
         return referrerURL
       }
       // 2. Return referrerOrigin.
--- a/test/web-platform-tests/expectation.json
+++ b/test/web-platform-tests/expectation.json
@@ -689,8 +689,7 @@
           "cases": [
             {
               "name": "origin-when-cross-origin policy on a same-origin URL",
-              "success": false,
-              "message": "assert_equals: Request's referrer is correct expected \"http://web-platform.test:8000/fetch/api/basic/referrer.any.html\" but got \"http://web-platform.test:8000/\""
+              "success": true
             },
             {
               "name": "origin-when-cross-origin policy on a cross-origin URL",


KhafraDev (changes requested):
Remove the tests, run the fetch WPTs, commit the updated expectations file. The WPTs cover these tests extensively.

spokodev:
I ran the fetch WPTs. The referrer-policy tests (`referrer-origin-when-cross-origin.html`, `referrer-origin.html`, etc.) do not execute in the Node runner: they error with `importScripts is not defined` / `Worker is not defined` and are already recorded as `{success: false, cases: []}` in expectation.json, so this fix produces no expectation change. Because those tests do not run in the harness, the unit test in `test/fetch/util.js` is the only thing that exercises the fix (it fails on `main` and passes with the change). Would you prefer I keep it, or drop it and rely on the browser WPT runs?

spokodev:
You were right, and sorry for the back-and-forth — the fetch WPTs do cover this.

I'd only looked at the worker/service-worker variants (`referrer-origin-when-cross-origin.html` etc.), which don't execute in the Node harness (they're recorded as `{success: false, cases: []}`), and wrongly concluded the suite didn't exercise the fix. I missed `fetch/api/basic/referrer.any.js`, which does run and covers it directly.

Done as you asked:
- Removed the unit test.
- Ran the fetch WPTs — the case **"origin-when-cross-origin policy on a same-origin URL"** in `fetch/api/basic/referrer.any.js` was failing on `main` (`expected ".../referrer.any.html"`, got `".../"`) and passes with the fix.
- Committed the updated `expectation.json` (that one case flips `false → true`).

The PR is now just the two-line fix plus the expectation update.

spokodev:
Checking in on this. The change you asked for went in on 28 July — the unit test is gone, the fetch WPTs were run, and the one case they flip (`origin-when-cross-origin policy on a same-origin URL` in `fetch/api/basic/referrer.any.js`) is committed in the expectations file. The PR is now the two-line fix plus that expectation update.

Let me know if you want anything else on it.

PR #5499, opened 2026-07-03
fix: run idle socket validation off setImmediate instead of setTimeout

### LINKED ISSUE #5493: Idle-socket validation adds ~1.3ms to every keep-alive request reuse (setTimeout(0) in scheduleIdleSocketValidation)
## Bug Description

Since the fix for GHSA-35p6-xmwp-9g52 (v7.28.0 / v8.5.0 / v6.27.0), every request dispatched onto a reused idle keep-alive socket is deferred behind a `setTimeout(..., 0)` in `scheduleIdleSocketValidation` (`lib/dispatcher/client-h1.js`). Node's timer floor makes that ~1ms+, which turns sub-0.1ms keep-alive requests into ~1.4ms requests for any sequential (poll-style) workload — a ~15x latency regression on low-RTT links.

To be clear: this is **not** a request to remove the response-queue-poisoning protection. It's a request to reduce the cost of the validation tick.

## Reproducible By

```js
// node bench.mjs  (npm i undici@7.27.2 / undici@8.6.0 to compare)
import { Agent, fetch } from 'undici';

    res.writeHead(200, { 'content-type': 'application/json' });
    res.end('{"ok":1}');
});
server.keepAliveTimeout = 65000;
    const url = 'http://127.0.0.1:' + server.address().port + '/';
    const agent = new Agent({ keepAliveTimeout: 60000, connections: 256, pipelining: 1 });
    for (let i = 0; i < 300; i++) { await (await fetch(url, { dispatcher: agent })).text(); } // warmup
    const times = [];
    for (let i = 0; i < 2000; i++) {
        const t0 = process.hrtime.bigint();
        await (await fetch(url, { dispatcher: agent })).text();
        times.push(Number(process.hrtime.bigint() - t0));
    }
    times.sort((a, b) => a - b);
    console.log('sequential p50:', (times[1000] / 1e6).toFixed(3), 'ms');
    await agent.close();
    server.close();
});
```

Measured (node v22.22.1, linux x64, loopback, median of interleaved runs):

| version | sequential p50 | 8-wide burst p50 |
|---|---|---|
| 7.27.2 (pre-fix) | 0.078–0.083 ms | 0.49–0.52 ms |
| 7.28.0 | 1.385 ms | 1.95 ms |
| 8.6.0 | 1.36–1.41 ms | 1.91–1.97 ms |

Bursts pay the tick once per reused idle socket per round.

## Expected Behavior

Idle-socket liveness validation with (sub-)microsecond-scale overhead in the common case where the socket is healthy.

The goal of the tick, as I read it, is to give the event loop a chance to surface an unsolicited response / FIN / RST that arrived while the socket sat idle, before a new request is associated with it. Some candidate directions that keep that property without the ~1ms timer floor:

- run the deferred resume off `setImmediate` instead of `setTimeout(0)` — the check phase still runs after the poll phase, so pending `'data'`/`'end'` events on the idle socket are processed first, at ~50µs instead of ~1ms+;
- skip (or fast-path) validation when the socket has been idle for less than some small threshold, by analogy with #5400's fresh-socket guard — a socket that went idle microseconds ago on a `pipelining: 1` client has had no meaningful window to receive an injected response;
- an explicit opt-in fast mode for clients that only talk to trusted upstreams was considered and discarded here — an option that silently reopens GHSA-35p6-xmwp-9g52 seems worse than making the mitigation cheap.

Context: found while migrating [ccxt](https://github.com/ccxt/ccxt) (exchange-polling workloads, latency-sensitive users on single-digit-ms RTT links) from node-fetch to undici — node core's `http.Agent` achieves the same safety by destroying idle sockets that receive unsolicited data, with no per-request deferral, so undici currently benchmarks slower than the `node-fetch`+`https.Agent` stack it replaces on these workloads. Happy to test patches or contribute a PR if maintainers agree on a direction.

## Environment

node v22.22.1, Linux x64; undici 7.27.2 / 7.28.0 / 8.6.0 from npm

## Additional context

Related merged work touching the same code path: #5397 / #5402 / #5407 / #5409 (fake-timer interop), #5400 (fresh-socket guard, v6.x backport).

## Disclosure

This issue (analysis, benchmarks, and text) was created by Claude (model: claude-fable), an AI assistant by Anthropic, operating under human direction. Please review accordingly.

carlotestor:
## Summary

Fixes #5493

Schedule the idle-socket validation deferral with `setImmediate` instead of `setTimeout(..., 0)` in `scheduleIdleSocketValidation` (`lib/dispatcher/client-h1.js`), and cancel it with `clearImmediate` accordingly (`clearTimeout` does not cancel an `Immediate`, so the pending callback would otherwise fire after socket cleanup — harmlessly, since it self-guards, but the handle should be cancelled properly). The now-meaningless `0` delay argument is dropped.

The validation only needs the event loop to surface any unsolicited bytes / FIN / RST already pending on the reused idle socket before the next request is written. The `setImmediate` check phase still runs after the poll phase, so those events are processed first — but without the ~1ms+ timer floor that `setTimeout(0)` pays on every idle-socket reuse.

## Measured

Loopback HTTP, keep-alive Agent (`pipelining: 1`), 2000 sequential requests + 8-wide bursts, node v22.22.1 (full harness in #5493):

| | sequential p50 | 8-wide burst p50 |
|---|---|---|
| main (`setTimeout(0)`) | ~1.4 ms | ~1.9 ms |
| this patch (`setImmediate`) | **0.076 ms** | **0.446 ms** |
| 7.27.2 (pre-validation baseline) | 0.078 ms | 0.49 ms |

i.e. the GHSA-35p6-xmwp-9g52 mitigation is retained at effectively zero latency cost.

## Tests

- `node --test test/response-queue-poisoning.js test/issue-810.js test/client-keep-alive.js` — 15/15 pass with the patch (including the poisoned-idle-socket regression test)

## Notes

- The `kIdleSocketValidationTimeout` symbol now holds an `Immediate`; left the name unchanged to keep the diff minimal, happy to rename if preferred.
- `Immediate.unref()` exists, so the existing `.unref?.()` call keeps working.

## Disclosure

This pull request (analysis, patch, benchmarks, and text) was created by Claude (model: claude-fable), an AI assistant by Anthropic, operating under human direction. Please review accordingly.

### DIFF
--- a/lib/dispatcher/client-h1.js
+++ b/lib/dispatcher/client-h1.js
@@ -1050,7 +1050,7 @@ function onSocketClose () {
 
 function clearIdleSocketValidation (socket) {
   if (socket[kIdleSocketValidationTimeout]) {
-    clearTimeout(socket[kIdleSocketValidationTimeout])
+    clearImmediate(socket[kIdleSocketValidationTimeout])
     socket[kIdleSocketValidationTimeout] = null
   }
 
@@ -1059,14 +1059,14 @@ function clearIdleSocketValidation (socket) {
 
 function scheduleIdleSocketValidation (client, socket) {
   socket[kIdleSocketValidation] = 1
-  socket[kIdleSocketValidationTimeout] = setTimeout(() => {
+  socket[kIdleSocketValidationTimeout] = setImmediate(() => {
     socket[kIdleSocketValidationTimeout] = null
     socket[kIdleSocketValidation] = 2
 
     if (client[kSocket] === socket && !socket.destroyed) {
       client[kResume]()
     }
-  }, 0)
+  })
   socket[kIdleSocketValidationTimeout].unref?.()
 }
 


mcollina (approved):
lgtm

Issue #5508, opened 2026-07-04
cache: response "must-revalidate" ignored when request has "max-stale" — stale served without validation (RFC 9111 §5.2.2.2)


jeswr:
## Bug Description

`isStale()` (lib/interceptor/cache.js:65-91) grants the request `max-stale` grace window without
consulting the stored response's directives. RFC 9111 §5.2.2.2: a cache MUST NOT use a stale
response with `must-revalidate` to satisfy any request without successful validation — this
overrides `max-stale` (§5.2.1.2 note) and also excludes it from stale-if-error (RFC 5861 §4).
`proxy-revalidate` (§5.2.2.8) has the same semantics for shared caches — which is the
interceptor's default `type` — and is currently parsed but never enforced.

Repro: store `200 Cache-Control: max-age=1, must-revalidate` (+ ETag); wait >1s; request with
`Cache-Control: max-stale=600` → observed: served from cache, zero origin contact
(undici 8.7.0, current main, Node 22); expected: conditional request to the origin.

## Suggested fix

Skip the max-stale grace in `isStale()` when
`result.cacheControlDirectives['must-revalidate']` (or `proxy-revalidate` with `type: 'shared'`),
and exclude such responses from the stale-if-error threshold in `handleResult`. ~5-10 LOC + tests.

---
*Found during an agent-assisted HTTP-caching review for @jeswr; every claim reproduced on undici 8.6.0 (repo) and 8.7.0 (npm) on Node 22.23.1. Fix PR to follow.*

Issue #5505, opened 2026-07-04
cache: 304 revalidation response never updates the stored entry on the synchronous path (RFC 9111 §4.3.4) — entry stays stale forever


jeswr:
## Bug Description

#4617 (fixing #4596) taught `CacheHandler` to merge 304 responses into the stored entry — but
that code path is only reachable where `CacheHandler` sees the response directly (the
stale-while-revalidate background refresh). On the *synchronous* revalidation path,
`CacheRevalidationHandler` short-circuits on success (`statusCode === 304` →
`callback(true)` → `return true`, lib/handler/cache-revalidation-handler.js:72-88) and never
forwards anything to the wrapped `CacheHandler`, so:

1. Headers carried by the 304 (notably a new `Cache-Control`/`Expires`/`Date`) are discarded.
   RFC 9111 §4.3.4: the cache MUST update the stored response with the 304's header fields and
   recompute freshness. An origin extending freshness via `304 + Cache-Control: max-age=60`
   (standard Varnish/CDN behavior) gets zero benefit: undici revalidates again on the very next
   request, every time, forever.
2. Even a bare 304 leaves the entry stale — repeated per-request conditional traffic instead of
   a fresh window.
3. The revalidated response is served with `Warning: 110 - "response is stale"` although it has
   just been successfully validated (acknowledged TODO at lib/interceptor/cache.js:378; the
   Warning header is also obsolete per RFC 9111 §5.5).

The five `304-etag-update-response-*` cache-tests are currently in the conformance skip list
with the comment "We're not caching 304s currently" (test/cache-interceptor/cache-tests.mjs:76-81);
fixing this would un-skip them.

## Reproducible By

    const { Agent, interceptors, cacheStores, request } = require('undici')
    const http = require('node:http')
    const { setTimeout: sleep } = require('node:timers/promises')

    let hits = 0
    const server = http.createServer((req, res) => {
      hits++
      if (req.headers['if-none-match'] === '"v1"') {
        // Extend freshness for a minute — per §4.3.4 the client must apply this
        res.writeHead(304, { 'cache-control': 'max-age=60', etag: '"v1"' })
        return res.end()
      }
      res.writeHead(200, { 'cache-control': 'max-age=1', etag: '"v1"' })
      res.end('hello')
    }).listen(0, async () => {
      const origin = `http://localhost:${server.address().port}`
      const d = new Agent().compose(interceptors.cache({ store: new cacheStores.MemoryCacheStore() }))
[...]

Issue #3847, opened 2024-11-19
Support all the `cache` options in `fetch()`


mcollina:
Ref https://developer.mozilla.org/en-US/docs/Web/API/Request/cache.

I _think_ we could add some support for this.

mcollina:
cc @flakey5 @KhafraDev

KhafraDev:
+1 to adding it, but it will not work with the current caching interceptor logic. Pasting part of my comment from [here](https://github.com/nodejs/undici/pull/3842#issuecomment-2484530817):

> In the fetch spec, the http cache is built a layer above fetch, whereas in undici it's a layer below [in the interceptor api]. For example, in multiple parts of the spec we are meant to query the cache directly (see [http-network-or-cache-fetch](https://fetch.spec.whatwg.org/#http-network-or-cache-fetch)). There are multiple situations where headers may get added to requests/responses, or the internal request is modified in some way that will cause bugs or deviate from browser behavior.

mcollina:
https://fetch.spec.whatwg.org/#concept-request-cache-mode doens't imply it's implemented on top. I'm specifically _not_ talking about `CacheStorage`.

> There are multiple situations where headers may get added to requests/responses, or the internal request is modified in some way that will cause bugs or deviate from browser behavior.

Can you clarify this? Why can't we go update the underlining cache?

KhafraDev:
> doens't imply it's implemented on top. I'm specifically not talking about CacheStorage.

I know, but think about it from this perspective - currently the cache is built into the interceptor itself; it accepts an array of buffers as headers and a node stream for its body. To cache fetch responses, you have to store the headers (a HeadersList), a body (web readable), and the internal response state.

The fetch spec calls implementations to update, insert, and delete responses from the cache in the spec itself. If we leave this to the interceptor, there are a number of places where internal response state may be modified, headers get added, etc. that will change how the response is received.

> Can you clarify this? Why can't we go update the underlining cache?

The interceptor cache does not take internal response state into account, and that internal state is also modified between where we are meant to update/insert/delete from the cache and where we are currently doing it (the dispatch). Without taking this into account, things like `referer` headers, etc. will subtly break). AFAIK the cache is tied to the interceptor api currently, and isn't shared outside of it (and if it can, it still leads to the issue of checking the cache in fetch, and then having to re-check it again in the interceptors).

mcollina:
> To cache fetch responses, you have to store the headers (a HeadersList), a body (web readable), and the internal response state.

[...]

PR #5510, opened 2026-07-04
fix: honor request cache-control max-age=0 in cache interceptor

### LINKED ISSUE #5504: cache interceptor ignores request "Cache-Control: max-age=0" (falsy check) — breaks fetch(url, {cache:'no-cache'}) through composed cache dispatchers
## Bug Description

`lib/interceptor/cache.js` checks the request `max-age` directive with

    if (reqCacheControl?.['max-age'] && age >= reqCacheControl['max-age']) {

(currently interceptor/cache.js:283). For `max-age=0` the parsed value `0` is falsy, so the
directive is ignored entirely and a cached response is served with no origin contact.
RFC 9111 §5.2.1.1: the client has indicated it is unwilling to accept any response whose age
exceeds 0s — the cache MUST NOT reuse the stored response without successful validation.

This also breaks `fetch(url, { cache: 'no-cache' })` when the dispatcher is composed with the
cache interceptor: per the fetch spec, that mode appends `Cache-Control: max-age=0`
(lib/web/fetch/index.js, "cache mode is no-cache" step), which the interceptor then drops on the
floor — so the one documented way to force revalidation through fetch does nothing.
(Related umbrella: #3847.)

Secondary issue in the same block: when a *nonzero* request `max-age` is exceeded, the bypass
dispatches without a `CacheHandler` (`return dispatch(opts, handler)`, interceptor/cache.js:286),
so the fresh 200 fetched because of the bypass is never stored — the very next plain request has
to revalidate or refetch again.

## Reproducible By

    const { Agent, interceptors, cacheStores, request } = require('undici')
    const http = require('node:http')

    let hits = 0
      hits++
      res.writeHead(200, { 'cache-control': 'max-age=60', etag: '"v1"' })
      res.end('hello')
    }).listen(0, async () => {
      const d = new Agent().compose(interceptors.cache({ store: new cacheStores.MemoryCacheStore() }))
      await (await request(origin, { dispatcher: d })).body.text()                      // hits=1, stored
      await (await request(origin, { dispatcher: d, headers: { 'cache-control': 'max-age=0' } })).body.text()
      console.log(hits) // observed: 1 — expected: 2 (validation required by RFC 9111 §5.2.1.1)
      server.close()
    })

Observed on undici 8.7.0 (and current main), Node 22.

## Expected Behavior

`max-age=0` (and generally `age >= max-age`) should force the revalidation path (send
`If-None-Match`/`If-Modified-Since`, serve cached body on 304), same as the request `no-cache`
directive already does — not a silent cache hit, and not an un-stored full bypass.

Suggested fix sketch: treat request-max-age exceedance as staleness (feed it into the existing
`isStale`/revalidation flow) instead of the current falsy-guarded bypass, and wrap the bypass
dispatch in `new CacheHandler(...)` so the response updates the store.

---

jeswr:
Fixes #5504

## The bug

`lib/interceptor/cache.js` checked the request `max-age` directive with

```js
if (reqCacheControl?.['max-age'] && age >= reqCacheControl['max-age']) {
```

For `max-age=0` the parsed value `0` is falsy, so the directive was ignored entirely and a cached response was served with no origin contact. RFC 9111 §5.2.1.1 requires validation in that case. This also broke `fetch(url, { cache: 'no-cache' })` through a composed cache dispatcher, since the fetch spec implements that mode by appending `Cache-Control: max-age=0` (related umbrella: #3847).

Secondary issue in the same block: when a *nonzero* request `max-age` was exceeded, the bypass dispatched without a `CacheHandler` (`return dispatch(opts, handler)`), so the fresh 200 fetched because of the bypass was never stored — the very next plain request had to contact the origin again.

## The fix

Treat request `max-age` exceedance as a revalidation trigger in `needsRevalidation()` (alongside the existing request `no-cache` handling), feeding it into the existing revalidation flow: a conditional request is sent, the cached body is served on 304, and a fresh 200 is stored via the `CacheHandler` already wrapped by `CacheRevalidationHandler`. The old falsy-guarded bypass block is removed.

## Evidence

Repro from the issue (fresh cached response, then a request with `Cache-Control: max-age=0`):

- **before:** origin hits stay at `1` — silent cache hit, no validation
- **after:** origin receives a conditional request (`If-None-Match`), hits go to `2`, cached body served on 304

Same for `fetch(url, { cache: 'no-cache' })` through `new Agent().compose(interceptors.cache(...))`: before, 1 origin hit (revalidation never sent); after, the validation request is sent.

mnot [cache-tests](https://github.com/http-tests/cache-tests) conformance runner (`node test/cache-interceptor/cache-tests.mjs`):

- `ccreq-ma0` ("Does HTTP cache honor request `Cache-Control: max-age=0` when it holds a fresh response?"): **"N no" → "Y yes"**
- Totals unchanged and green: 283 total — 220 passed, 0 failed, 58 failed-optional, 5 setup-failed (identical to `main` baseline; the `ccreq-*` checks are behavioral yes/no entries that don't count toward pass/fail totals). Nothing in the runner's skip-list relates to request `max-age`, so no un-skips were needed.

New tests in `test/interceptors/cache.js` (all fail on `main`, pass with this change):

1. `max-age=0` on a fresh cached response triggers a validation request and serves the cached body on 304
2. a fresh 200 fetched because the response's age exceeds the request's `max-age` is stored, so the next plain request is a cache hit with the new body
3. `fetch(url, { cache: 'no-cache' })` through a composed cache dispatcher sends a validation request

Full cache test files (`test/interceptors/cache*.js`, `test/cache-interceptor/*`): 124 tests / 12 suites, 124 pass, 0 fail (baseline on `main` is 121/121; the +3 are the new tests).

## Notes

- Adjacent, intentionally out of scope (each gets its own self-contained PR): the 304 that this revalidation flow receives still doesn't update the stored entry (#5505), and `must-revalidate` vs request `max-stale` (#5508).

### DIFF
--- a/lib/interceptor/cache.js
+++ b/lib/interceptor/cache.js
@@ -45,10 +45,8 @@ function needsRevalidation (result, age, cacheControlDirectives, { headers = {} 
     return true
   }
 
-  // Always revalidate requests whose age exceeds the max-age request
-  //  directive. Note that max-age=0 must not be ignored (it is falsy), it
-  //  forces revalidation on every request (e.g. fetch's cache: 'no-cache'
-  //  mode sends Cache-Control: max-age=0).
+  // Revalidate when age >= max-age. max-age=0 is falsy but must still
+  //  force it (e.g. fetch cache: 'no-cache' sends max-age=0).
   // https://www.rfc-editor.org/rfc/rfc9111.html#section-5.2.1.1
   if (cacheControlDirectives?.['max-age'] !== undefined && age >= cacheControlDirectives['max-age']) {
     return true
--- a/lib/interceptor/cache.js
+++ b/lib/interceptor/cache.js
@@ -34,13 +34,23 @@ const nop = () => {}
 
 /**
  * @param {import('../../types/cache-interceptor.d.ts').default.GetResult} result
+ * @param {number} age
  * @param {import('../../types/cache-interceptor.d.ts').default.CacheControlDirectives | undefined} cacheControlDirectives
  * @param {import('../../types/dispatcher.d.ts').default.RequestOptions} opts
  * @returns {boolean}
  */
-function needsRevalidation (result, cacheControlDirectives, { headers = {} }) {
+function needsRevalidation (result, age, cacheControlDirectives, { headers = {} }) {
   // Always revalidate requests with the no-cache request directive.
   if (cacheControlDirectives?.['no-cache']) {
+    return true
+  }
+
+  // Always revalidate requests whose age exceeds the max-age request
+  //  directive. Note that max-age=0 must not be ignored (it is falsy), it
+  //  forces revalidation on every request (e.g. fetch's cache: 'no-cache'
+  //  mode sends Cache-Control: max-age=0).
+  // https://www.rfc-editor.org/rfc/rfc9111.html#section-5.2.1.1
+  if (cacheControlDirectives?.['max-age'] !== undefined && age >= cacheControlDirectives['max-age']) {
     return true
   }
 
@@ -297,14 +307,9 @@ function handleResult (
   }
 
   const age = Math.round((now - result.cachedAt) / 1000)
-  if (reqCacheControl?.['max-age'] && age >= reqCacheControl['max-age']) {
-    // Response is considered expired for this specific request
-    //  https://www.rfc-editor.org/rfc/rfc9111.html#section-5.2.1.1
-    return dispatch(opts, handler)
-  }
 
   const stale = isStale(result, reqCacheControl, globalOpts.type)
-  const revalidate = needsRevalidation(result, reqCacheControl, opts)
+  const revalidate = needsRevalidation(result, age, reqCacheControl, opts)
 
   // Check if the response is stale
   if (stale || revalidate) {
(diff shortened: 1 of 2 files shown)


lib/interceptor/cache.js
```diff
@@ -34,16 +34,26 @@ const nop = () => {}
...
  * @returns {boolean}
  */
-function needsRevalidation (result, cacheControlDirectives, { headers = {} }) {
+function needsRevalidation (result, age, cacheControlDirectives, { headers = {} }) {
   // Always revalidate requests with the no-cache request directive.
   if (cacheControlDirectives?.['no-cache']) {
     return true
   }
 
+  // Always revalidate requests whose age exceeds the max-age request
```
mcollina: This comment is incredibly long. Shorten it

mcollina (changes requested):

lib/interceptor/cache.js
jeswr: Shortened in a10a00d2 — kept the max-age=0/falsy caveat and the RFC 9111 §5.2.1.1 ref. Thanks.

metcoder95 (approved):

kazimahiuddin-hub (approved):

mcollina (approved):
lgtm

metcoder95:
can you resolve conflicts?

PR #5511, opened 2026-07-04
fix: enforce must-revalidate and proxy-revalidate over max-stale and stale-if-error

### LINKED ISSUE #5508: cache: response "must-revalidate" ignored when request has "max-stale" — stale served without validation (RFC 9111 §5.2.2.2)

jeswr:
## What

The cache interceptor's `isStale()` granted the request `max-stale` grace window without consulting the stored response's directives, and `handleResult()` applied the `stale-if-error` threshold unconditionally.

RFC 9111 [§5.2.2.2](https://www.rfc-editor.org/rfc/rfc9111.html#section-5.2.2.2): a cache **MUST NOT** use a stale response with `must-revalidate` to satisfy any request without successful validation — this overrides the request's `max-stale` (see the note in §5.2.1.2) and also excludes the response from `stale-if-error` (RFC 5861 §4). [`proxy-revalidate`](https://www.rfc-editor.org/rfc/rfc9111.html#section-5.2.2.8) has the same semantics for shared caches — the interceptor's default `type` — and was parsed but never enforced anywhere.

Fixes #5508

## Repro

Store `200 Cache-Control: max-age=1, must-revalidate` (+ ETag), wait past staleness, then request with `Cache-Control: max-stale=600`:

- **Before** (undici 8.7.0 / current main, Node 22): served straight from cache, zero origin contact.
- **After**: conditional request (`if-none-match`/`if-modified-since`) sent to the origin; the cached response is only reused after successful validation.

Same for `stale-if-error`: a stale `must-revalidate` response was previously served when revalidation returned a 5xx within the `stale-if-error` window; now the error is propagated.

## Changes

- `lib/interceptor/cache.js`: new `forbidsServingStale(result, cacheType)` helper; `isStale()` skips the `max-stale` grace window and `handleResult()` skips the `stale-if-error` threshold when the stored response has `must-revalidate` (or `proxy-revalidate` on a shared cache). `proxy-revalidate` correctly remains inert for `type: 'private'` caches.
- `test/interceptors/cache.js`: three new tests (must-revalidate vs max-stale, proxy-revalidate shared vs private, must-revalidate excluded from stale-if-error), all failing before the fix and passing after.

## Testing

- New tests fail on main (stale served with 0 revalidation requests / 200 instead of 500) and pass with the fix.
- Full cache interceptor suite (`test/interceptors/cache*.js`, `test/cache-interceptor/*.js`): 124 tests, 124 pass, 0 fail (121 baseline + 3 new).
- mnot cache-tests conformance runner (`node test/cache-interceptor/cache-tests.mjs`): identical to main's baseline — 283 total, 220 passed, 0 failed, 58 failed-optional, 5 setup-failed. (The suite has no test for the max-stale + must-revalidate combination, so no skip-list entries could be un-skipped; the skipped `stale-close-must-revalidate` test depends on the connection-close revalidation path, which is out of scope here.)

## Notes

- Touches the same `handleResult()` region as the sibling PR #5510 (`fix/cache-request-max-age-zero`), but both are self-contained against main and merge-independent.

### DIFF
--- a/lib/interceptor/cache.js
+++ b/lib/interceptor/cache.js
@@ -58,15 +58,33 @@ function needsRevalidation (result, cacheControlDirectives, { headers = {} }) {
 }
 
 /**
+ * A response with the must-revalidate directive (or proxy-revalidate for
+ * shared caches) can't be used once stale without successful validation,
+ * taking precedence over max-stale and stale-if-error
+ * https://www.rfc-editor.org/rfc/rfc9111.html#section-5.2.2.2
+ * https://www.rfc-editor.org/rfc/rfc9111.html#section-5.2.2.8
  * @param {import('../../types/cache-interceptor.d.ts').default.GetResult} result
- * @param {import('../../types/cache-interceptor.d.ts').default.CacheControlDirectives | undefined} cacheControlDirectives
+ * @param {'shared' | 'private'} cacheType
  * @returns {boolean}
  */
-function isStale (result, cacheControlDirectives) {
+function forbidsServingStale (result, cacheType) {
+  return Boolean(
+    result.cacheControlDirectives?.['must-revalidate'] ||
+    (cacheType === 'shared' && result.cacheControlDirectives?.['proxy-revalidate'])
+  )
+}
+
+/**
+ * @param {import('../../types/cache-interceptor.d.ts').default.GetResult} result
+ * @param {import('../../types/cache-interceptor.d.ts').default.CacheControlDirectives | undefined} cacheControlDirectives
+ * @param {'shared' | 'private'} cacheType
+ * @returns {boolean}
+ */
+function isStale (result, cacheControlDirectives, cacheType) {
   const now = Date.now()
   if (now > result.staleAt) {
     // Response is stale
-    if (cacheControlDirectives?.['max-stale']) {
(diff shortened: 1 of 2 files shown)


metcoder95 (approved):

mcollina (approved):
lgtm

PR #5513, opened 2026-07-04
fix: honor stale-if-error on connection errors during revalidation

### LINKED ISSUE #5507: cache: stale-if-error only honored for 500-504 responses, not connection/network errors (RFC 5861 §4)
## Feature/Bug

`stale-if-error` (both the response and request directive) is honored when the origin *responds*
with 500/502/503/504 (`allowErrorStatusCodes` check, lib/handler/cache-revalidation-handler.js:72-73),
but when the origin is unreachable, `onResponseError` fails the request outright
(cache-revalidation-handler.js:106-121) even when a within-threshold stale entry exists.

RFC 5861 §4 defines an error as "any situation that would result in a 500, 502, 503 or 504" —
for a client-side cache, failing to connect *is* the situation a gateway would surface as
502/504, and resilience across origin outages is the extension's headline use case (nginx
`proxy_cache_use_stale error ...`, Varnish, Fastly all treat connect errors this way).

Repro: store `200 Cache-Control: max-age=1, stale-if-error=60`; stop the origin; wait past
staleness; request again → observed `ECONNREFUSED` thrown (undici 8.7.0, current main, Node 22);
expected: stale cached response served. The equivalent test with the origin returning 503 works
today, which makes the inconsistency surprising.

## Suggested fix

In `CacheRevalidationHandler.onResponseError`, when `#allowErrorStatusCodes` (the caller already
encodes the within-stale-if-error-threshold decision, interceptor/cache.js:346-350) and no
response has started, invoke the callback with success so the interceptor serves the stale entry;
propagate the error only otherwise. ~20-40 LOC + tests.

---

jeswr:
Fixes #5507

## What

`stale-if-error` (response and request directive) was only honored when the origin *responded* with a 500-504 status code. If the origin was unreachable (connection refused, reset, etc.), `CacheRevalidationHandler.onResponseError` failed the request outright even when a stale entry within the `stale-if-error` window existed.

RFC 5861 §4 defines an error as "any situation that would result in a 500, 502, 503 or 504 HTTP response status code being returned" — for a client-side cache, failing to reach the origin is exactly the situation a gateway would surface as 502/504, and resilience across origin outages is the extension's headline use case (nginx `proxy_cache_use_stale error`, Varnish and Fastly all treat connect errors this way).

## Fix

The interceptor already computes the within-threshold decision (`withinStaleIfErrorThreshold` in `lib/interceptor/cache.js`) and passes it to the revalidation handler as `allowErrorStatusCodes`. This change also uses it on the error path: in `onResponseError`, when the response hasn't started yet (the callback is still pending) and we're within the threshold, invoke the callback with success so the stale entry is served. Errors after a response has started, or outside the window, propagate exactly as before.

## Repro / evidence

Store a response with `Cache-Control: max-age=1, stale-if-error=60`; stop the origin; wait past freshness; request again (undici main @ cb4c2f1, Node 22):

- **Before:** throws `ECONNREFUSED` (the equivalent scenario with the origin returning 503 already served stale, making the inconsistency surprising)
- **After:** stale cached response served; once outside the `stale-if-error` window the connection error is propagated again

Added tests for both the response directive and the request directive variants in `test/interceptors/cache.js` ("stale-if-error (response) on connection error", "stale-if-error on connection error"); both fail on main and pass with this change.

Cache test suite (`test/interceptors/cache*.js`, `test/cache-interceptor/*`): 123 pass / 0 fail (baseline 121 pass + the 2 new tests). mnot cache-tests conformance runner (`node test/cache-interceptor/cache-tests.mjs`): the `stale-sie-close` check ("Does HTTP cache serve stale stored response when server sends `Cache-Control: stale-if-error` and subsequently closes the connection?") now answers **yes** in all environments (previously no) — summary went from 220 passed / 58 failed-optional to 222 passed / 56 failed-optional, with 0 required failures and no regressions. No skip-list entries relate to this path (`stale-close-must-revalidate`/`stale-close-no-cache` depend on plain-close behavior without `stale-if-error`, which is unchanged).

## Notes

- Related PRs from the same review, touching adjacent cache-revalidation logic (each self-contained against main): `jeswr:fix/cache-must-revalidate-max-stale`, `jeswr:fix/cache-if-modified-since-value`.

### DIFF
--- a/lib/handler/cache-revalidation-handler.js
+++ b/lib/handler/cache-revalidation-handler.js
@@ -109,11 +109,8 @@ class CacheRevalidationHandler {
     }
 
     if (this.#callback) {
-      // If the response hasn't started yet and we're within the
-      //  stale-if-error threshold, serve the stale cached response instead
-      //  of failing the request. RFC 5861 defines an error as "any situation
-      //  that would result in a 500, 502, 503, or 504 HTTP response status
-      //  code being returned", which includes failing to reach the origin.
+      // Serve the stale cached response on a connection error, per stale-if-error:
+      //  RFC 5861 counts an unreachable origin (a would-be 5xx) as an error.
       // https://datatracker.ietf.org/doc/html/rfc5861#section-4
       if (this.#allowErrorStatusCodes) {
         this.#successful = true
--- a/lib/handler/cache-revalidation-handler.js
+++ b/lib/handler/cache-revalidation-handler.js
@@ -109,6 +109,19 @@ class CacheRevalidationHandler {
     }
 
     if (this.#callback) {
+      // If the response hasn't started yet and we're within the
+      //  stale-if-error threshold, serve the stale cached response instead
+      //  of failing the request. RFC 5861 defines an error as "any situation
+      //  that would result in a 500, 502, 503, or 504 HTTP response status
+      //  code being returned", which includes failing to reach the origin.
+      // https://datatracker.ietf.org/doc/html/rfc5861#section-4
+      if (this.#allowErrorStatusCodes) {
+        this.#successful = true
+        this.#callback(true, this.#context)
+        this.#callback = null
+        return
+      }
+
       this.#callback(false)
       this.#callback = null
     }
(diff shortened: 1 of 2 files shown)


metcoder95 (approved):

mcollina (approved):
lgtm

mcollina:
there are conflicts

mcollina (approved):
lgtm

PR #5514, opened 2026-07-04
fix: invalidate Location and Content-Location URIs on unsafe methods

### LINKED ISSUE #5509: cache: unsafe-method invalidation does not invalidate Location / Content-Location URIs (RFC 9111 §4.4)
`CacheHandler.onResponseStart` deletes only the request URI's entry on a 2xx/3xx response to an
unsafe method (lib/handler/cache-handler.js:109-122). RFC 9111 §4.4 also says caches SHOULD
invalidate the URIs in the response's `Location` and `Content-Location` fields when they share
the request URI's origin — the classic `POST /collection` → `201 Location: /collection/123` flow
currently leaves a previously cached `GET /collection/123` stale.

Repro: cache `GET /target`; `POST /src` returning `Location: <origin>/target`; `GET /target`
again → observed: served from cache (undici 8.7.0/main, Node 22); expected: refetched.

Fix sketch: in the invalidation branch, parse `Location`/`Content-Location`, same-origin check
against `cacheKey.origin`, and `store.delete({origin, path, method: 'GET'})` for each (both
bundled stores key deletion on origin+path, so this is ~25-40 LOC + tests).

---

jeswr:
Fixes #5509

## What

`CacheHandler.onResponseStart` deleted only the request URI's entry on a 2xx/3xx response to an unsafe method. [RFC 9111 §4.4](https://www.rfc-editor.org/rfc/rfc9111.html#section-4.4) also says caches SHOULD invalidate the URIs in the response's `Location` and `Content-Location` header fields when they share the request URI's origin — so the classic `POST /collection` → `201 Location: /collection/123` flow left a previously cached `GET /collection/123` stale.

This adds a `#deleteLocationHeaderEntries` step to the existing invalidation branch: it resolves each header value against the request URI (relative references included), applies the same-origin check from the RFC's security note (invalidating cross-origin URIs could be abused for cache poisoning), and deletes the matching entry from the store.

## Repro / evidence

Origin: `GET /target` served with `Cache-Control: public, s-maxage=100`; `POST /src` answers `201` with `Location: <origin>/target`.

Before (main @ cb4c2f1, Node 22):
1. `GET /target` → origin hit, cached
2. `POST /src` (response carries `Location: <origin>/target`)
3. `GET /target` → **served from cache** (1 origin hit total on `/target`)

After: step 3 goes back to the origin (2 hits), for both `Location` and `Content-Location`, absolute and relative forms. A `Location` pointing at another origin (even with a path that collides with a cached entry) does not invalidate — covered by the added test in `test/interceptors/cache.js`.

## Tests

- New node:test case `unsafe methods invalidate the URIs in Location and Content-Location response headers` in `test/interceptors/cache.js` (fails on main with `expected: 2, actual: 1` origin hits, passes with this change).
- Full cache test files (`test/interceptors/cache*.js`, `test/cache-interceptor/*`): 122 pass / 0 fail (121 baseline + 1 new).
- mnot cache-tests conformance runner (`node test/cache-interceptor/cache-tests.mjs`): the eight `invalidate-{POST,PUT,DELETE,M-SEARCH}-{location,cl}` tests move from failed-optional to passing in every environment — totals go 220 → 228 passed, 58 → 50 failed-optional, still 0 required failures. Nothing needed un-skipping in the runner's ignore list.

## Notes

- Self-contained against `main`; independent of the sibling cache-conformance PRs #5510 (`fix/cache-request-max-age-zero`), #5511 (`fix/cache-must-revalidate-max-stale`), #5512 (`fix/cache-if-modified-since-value`) and #5513 (`fix/cache-stale-if-error-network`), which touch the interceptor/revalidation paths rather than this invalidation hook.

### DIFF
--- a/lib/handler/cache-handler.js
+++ b/lib/handler/cache-handler.js
@@ -118,6 +118,7 @@ class CacheHandler {
       } catch {
         // Fail silently
       }
+      this.#deleteLocationHeaderEntries(resHeaders)
       return downstreamOnHeaders()
     }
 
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

metcoder95 (approved):

PR #5519, opened 2026-07-06
fix(retry): handle Retry-After zero delay

Ram-blip:
## This relates to...

Fixes the default retry handler so `Retry-After: 0` is treated as a valid zero-delay retry instead of falling back to `minTimeout`.

## Rationale

`Retry-After` delay-seconds can be zero. The current retry timeout selection only uses the parsed header value when it is greater than zero, so `Retry-After: 0` falls through to the exponential backoff fallback.

This change adds an explicit zero-delay branch and changes invalid HTTP-date parsing to return `null` instead of `0`, so invalid dates continue to use the fallback path.

## Changes

- Preserve `Retry-After: 0` as a zero-delay retry.
- Keep invalid `Retry-After` date values on the fallback retry path.
- Add a fake-timer regression test for the zero-delay case.

### Features

N/A

### Bug Fixes

Fixes default retry handling for `Retry-After: 0`.

N/A

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/handler/retry-handler.js
+++ b/lib/handler/retry-handler.js
@@ -11,7 +11,7 @@ const {
 
 function calculateRetryAfterHeader (retryAfter) {
   const retryTime = new Date(retryAfter).getTime()
-  return isNaN(retryTime) ? 0 : retryTime - Date.now()
+  return isNaN(retryTime) ? null : retryTime - Date.now()
 }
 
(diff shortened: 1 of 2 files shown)


metcoder95 (approved):

mcollina (approved):
lgtm

Issue #5016, opened 2026-04-10
fetch() silently stringifies native globalThis.FormData body as "[object FormData]" (regression in v8.0.2)


heizer23:
When passing a native (global) FormData instance as the body to undici's fetch(), the request body is sent as the literal string [object FormData] with Content-Type: text/plain instead of a multipart/form-data body. This regression was introduced in v8.0.2.

Reproducible By

Use Node.js 22 or later (which exposes FormData as a global)
Create a FormData using the native global — not undici's own FormData
Pass it as the body to undici's fetch()
import { fetch } from 'undici';

const form = new FormData(); // native globalThis.FormData
form.append('field', 'value');

const res = await fetch('https://httpbin.org/post', { method: 'POST', body: form });
const data = await res.json();

console.log(data.form); // {} - empty, should be { field: 'value' }
console.log(data.data); // '[object FormData]' — body was stringified
Root cause

XMLHttpRequestBodyInit in lib/web/fetch/response.js checks webidl.is.FormData(V), which only matches undici’s internal FormData class.

Native globalThis.FormData fails this check and falls through to webidl.converters.DOMString(V), which calls .toString() on the object.

The same issue exists in extractBody in lib/web/fetch/body.js.

Fix

Add a fallback instanceof check in both places:

if (webidl.is.FormData(V) || (typeof globalThis.FormData !== 'undefined' && V instanceof globalThis.FormData)) {
  return V;
}
Expected Behavior

The request is sent as multipart/form-data with the form fields correctly encoded.

Logs & Screenshots
// data.data
'[object FormData]'

// data.headers['Content-Type']
'text/plain; charset=UTF-8'
Environment
undici 8.0.2
Node.js v24.14.0
Linux
Additional context

This worked correctly before v8.0.2. The issue surfaces silently — no error is thrown, the request succeeds with a 200, but the server receives a meaningless body.

Particularly hard to debug when the downstream API returns a domain-specific error (e.g. OpenAI returns "you must provide a model parameter") that gives no hint about the malformed body.

mcollina:
This is currently expected.

@KhafraDev do we want to do anything about this?

KhafraDev:
no, it's being handled properly. I have no ideas on how we would improve this.

Issue #4285, opened 2025-06-19
Undici v7(+?) `Request` cannot parse `FormData` from Undici v5


penalosa:
## Bug Description

I'm trying to construct a `Request` with a `FormData` body, using the global `Request` in Node.js v24 (which seems to be Undici 7.10.0), but a `FormData` imported from Undici 5.28.5.

## Reproducible By

```js
import { FormData } from 'undici';

const r = new Request('https://example.com', {
  body: new FormData(),
  method: 'POST',
});

console.log(r.headers.get('content-type'));
```

This outputs `multipart/form-data` in Node.js 20 & 22, but `text/plain` in Node.js 24

### Additional context

This may well be expected behaviour! However, if it is I would've expected to see some documention mentioning that Undici 5 isn't supported in Node.js 24+.

anonrig:
cc @Uzlopak @KhafraDev @nodejs/undici

KhafraDev:
This is expected.

penalosa:
@KhafraDev does that mean that Undici 5 is deprecated? I couldn't find any mention of that in the docs

mcollina:
No. This is doing what the spec says we should be doing. If you use `FormData` from `undici`, you have to use `fetch` from `undici` as well. If you use `FormData` from global, use `fetch` from global.

I'll need to add a script to override all the globals here, so you can just replace all of them.

Issue #3803, opened 2024-11-05
FormData body throws if request is redirected


DTrombett:
## Bug Description

When sending a `FormData` as body with `request` or `stream`, the request fails if redirected with the following `TypeError`: `The "string" argument must be of type string or an instance of Buffer or ArrayBuffer. Received an instance of Array`

## Reproducible By

In the following example a `POST` request is made to `http://github.com`, which automatically redirects to `https://github.com`. The issue is reproducible with any other url that triggers a redirect.
```js
import { FormData, request } from "undici";

const body = new FormData();

body.append("hello", "world");
await request("http://github.com", {
	body,
	method: "POST",
	maxRedirections: 1,
});
```

## Expected Behavior

Not sure about the specs but the body should be either redirected or discarded based on the redirect type, without any errors.

## Logs & Screenshots

```shell
node:buffer:776
    throw new ERR_INVALID_ARG_TYPE(
          ^

TypeError [ERR_INVALID_ARG_TYPE]: The "string" argument must be of type string or an instance of Buffer or ArrayBuffer. Received an instance of Array
    at Function.byteLength (node:buffer:776:11)
    at AsyncWriter.write (D:\test\node_modules\undici\lib\dispatcher\client-h1.js:1259:24)
    at writeIterable (D:\test\node_modules\undici\lib\dispatcher\client-h1.js:1219:19) {
  code: 'ERR_INVALID_ARG_TYPE'
}
```

## Environment

Windows 11 Home, Node v22.11.0

### Additional context

Tested with various redirect urls and the issue is always reproducible

KhafraDev:
What version of undici are you using?

DTrombett:
I'm pretty sure it's the latest because I had just upgraded but I'll let you know asap

KhafraDev:
This is a bug in v6 that is patched in v7 ([7.0.0-alpha.3](https://www.npmjs.com/package/undici/v/7.0.0-alpha.3)).

DTrombett:
I see, I'll try updating to alpha, thanks!

mcollina:
@KhafraDev do you know which PR?

DTrombett:
It seems that using the alpha the redirect is not followed at all? Not sure if I'm doing something wrong but I'll try to send an example

DTrombett:
```ts
import { request } from "undici";

const data = await request("http://github.com", { maxRedirections: 8 });

console.log(data.statusCode, data.headers.location);```

[...]

PR #5521, opened 2026-07-07
fix: reject cross-realm FormData request bodies

### LINKED ISSUE #5520: `request()` with a cross-realm FormData (e.g. Node's global) hangs forever instead of erroring; via the legacy globalDispatcher claim this also breaks undici v6 clients in the same process
## Summary

`undici.request()` half-accepts a FormData that isn't undici's own class: the client's duck-typed gate routes it into body extraction, the brand check inside `extractBody` then rejects it, and the request is dispatched with a null body that never completes. No error is raised; the request hangs until some external timeout (server idle timeout, caller watchdog, etc.).

On v7.20.0+/v8 this is reachable two ways (v7.0.0 through 7.19.x fail loudly instead; see the version table below):

1. **Standalone:** `request(url, { body: formData })` where `formData` is Node's `globalThis.FormData` (the most common way user code constructs one) hangs. undici's own `FormData` works.
2. **Cross-copy, via the eager legacy-dispatcher claim** (same mechanism as #5500): a bare `require('undici')` anywhere in the process captures `Symbol.for('undici.globalDispatcher.1')` (every major does guarded eager init at require time; since 7.27.0 a `Dispatcher1Wrapper` is mirrored into the legacy slot, and v8's init guard only checks its own `.2` slot, so it overwrites `.1` even when another copy already owns it, making both require orders break). Any undici **v6** client in the same process (e.g. `@discordjs/rest`, which pins undici 6.x) then dispatches through the newer core, where its FormData bodies, perfectly valid for v6, hit the same silent hang. The affected code never uses v7/v8 APIs and cannot work around it by "using undici's FormData".

Real-world impact of (2): adding undici 8.7.0 as a direct dependency of a discord.js bot made every file-attachment upload time out (`AbortError: This operation was aborted` after @discordjs/rest's 15s watchdog, 4 attempts) while all JSON requests kept working. This took a while to trace because the breakage is caused by a bare `require` in an unrelated module.

## Reproduction 1: standalone (no version mixing)

```js
// npm i undici@8.7.0

  req.on('data', () => {});
  req.on('end', () => res.end('{}'));
});

  const form = new FormData(); // Node's global FormData
  form.append('file', new Blob(['hello world'], { type: 'text/plain' }), 'hello.txt');
  const res = await request(`http://localhost:${server.address().port}/`, {
    method: 'POST',
    body: form,
  }); // <- never resolves, never rejects
  console.log('status', res.statusCode); // never reached
});
```

Replace `new FormData()` with `new (require('undici').FormData)()` and it completes with a correct multipart body.

## Reproduction 2: bare require of v8 breaks an undici v6 client

```js
// npm i undici6@npm:undici@6.27.0 undici8@npm:undici@8.7.0
require('undici8'); // bare require, never used again; remove this line and everything works
const { request } = require('undici6');

  req.on('data', () => {});
  req.on('end', () => res.end('{}'));
});

  const form = new FormData();
  form.append('file', new Blob(['hello world'], { type: 'text/plain' }), 'hello.txt');
  const res = await request(`http://localhost:${server.address().port}/`, {
    method: 'POST',
    body: form,
  }); // <- hangs; undici 6 alone handles this fine
});
```

`diagnostics_channel` shows `undici:request:create` and `undici:client:sendHeaders` firing, but `undici:request:bodySent` never does. Despite the `sendHeaders` event, a raw TCP-level check shows **zero bytes ever reach the wire**: the socket output is held back behind the body that never comes, so the server just sees an idle connection until something times out.

## Root cause

Two checks disagree about what a FormData is:

- `lib/dispatcher/client-h1.js` routes the body into form-data extraction based on the duck-typed check `util.isFormDataLike(body)` (structural: append/get/set methods plus `Symbol.toStringTag === 'FormData'`), which accepts FormData from any realm: Node's builtin, another undici copy, a polyfill.
- `lib/web/fetch/body.js` `extractBody()` gates the actual multipart encoding on the brand check `webidl.is.FormData(object)`, which only accepts instances of this copy's own `FormData` class.

For a cross-realm FormData the duck check passes and the brand check fails, so `extractBody` falls through every branch and returns an empty descriptor. This is observable directly:

```js
const { extractBody } = require('undici/lib/web/fetch/body.js');
const fd = new FormData(); // Node's global
fd.append('file', new Blob(['hello']), 'hello.txt');
const [body, contentType] = extractBody(fd);
// body.source = null, body.length = null, contentType = null,
// body.stream = a ReadableStream that never produces data and never closes.
// With undici's own FormData: multipart source, correct length,
// contentType 'multipart/form-data; boundary=...'.
```

`client-h1` then queues the headers and waits on that never-producing stream with `contentLength = null`: nothing reaches the wire, nothing errors.

## Affected versions (verified)

Measured with an explicit same-version `Agent` passed as `dispatcher`, which isolates each copy's client from the global-dispatcher slots:

| undici | `request()` + Node global FormData |
| --- | --- |
| 6.27.0 | works (duck-typed end to end) |
| 7.0.0, 7.16.0, 7.18.0, 7.19.0 | `TypeError: Cannot read properties of null (reading 'byteLength')` (loud, but already a regression vs v6) |
| 7.20.0 through 7.28.0 | silent hang |
| 8.7.0 | silent hang |

So the cross-realm FormData support regressed at 7.0.0, and 7.20.0 turned the loud error into the silent hang. The `.2` slot plus `Dispatcher1Wrapper` mirroring in `lib/global.js` landed at 7.27.0.

Verified on Node 24.18.0 and 26.4.0. Repro 2 (bare require poisoning a v6 client) verified with 7.27.1, 7.27.2, 7.28.0 and 8.7.0 as the claimer, in both require orders.

**Caveat for reproducing on Node >= 26:** touching the `FormData`/fetch globals loads Node's bundled undici 8.x, which claims both global dispatcher symbols; for npm copies older than 7.27.0 this silently reroutes their global-dispatcher requests through Node's bundled client (which brand-owns the global FormData) and masks the bug with false OK results. Use an explicit `dispatcher: new Agent()` from the copy under test.

## Relationship to existing issues

- #5016 (fetch() stringifies native FormData) was closed as expected behavior. This report is different on two counts: `request()`'s own gate *accepts* the object before the extraction drops it, and the failure mode is an unbounded silent hang, not a wrong-but-fast request. And in the repro-2 shape, the code hitting the bug is undici **6**, where cross-realm FormData is supported; "use this copy's FormData" is not actionable for a v6 client that got its dispatcher swapped out from under it by a bare require elsewhere in the tree.
- #5500 / #5502 cover the same eager legacy-symbol claim but only normalize duplicated content-length in `Dispatcher1Wrapper`; the FormData hang is unaffected by #5502.
- #4285 (v7 Request cannot parse v5 FormData) is the fetch-side sibling of the same cross-realm brand-check problem.

## Suggested fix directions

1. Make `extractBody`'s FormData branch accept what `isFormDataLike` accepts: the multipart encoder already only uses the public entry iterator (`for (const [name, value] of object)`) and Blob duck-typing (`name`, `type`, `size`, `stream()`), so encoding a foreign FormData should produce identical output. (The private `#boundary` getter is the only own-state dependency and can fall back to generating a boundary locally.)
2. Alternatively, use the brand check at the `client-h1` gate too, and throw `InvalidArgumentError('body must be ...')` like every other unsupported body type: a loud error instead of a silent hang.
3. Independently: reconsider the eager legacy-symbol claim at require time (as proposed in #5500), since it converts every intra-copy strictness change into a cross-copy breakage for code that never imports the new version.

Happy to turn either fix direction into a PR.

Pyker:
## This relates to...

Fixes #5520. Related context: #5500 / #5502 (legacy dispatcher bridge), #5016 and #4285 (fetch-side cross-realm FormData behavior, unchanged here).

## Rationale

`request()` accepts FormData bodies through the duck-typed `util.isFormDataLike()` check, but only undici's own `FormData` can be encoded: the multipart extraction is gated on the `webidl.is.FormData` brand check. A FormData from another realm (Node.js' builtin `globalThis.FormData`, another undici copy, a polyfill) passed the gate, fell through the extraction, and was dispatched as a body that never produced any bytes: the request hung silently until an external timeout, with zero bytes written to the socket (full version matrix and reproductions in #5520).

Per review direction (encode vs reject, maintainers chose reject): such bodies now fail fast with `InvalidArgumentError`. The check lives in the `Request` constructor, where other invalid body types are already rejected, so both the HTTP/1 and HTTP/2 clients reject cleanly at dispatch time and the fetch code is untouched. The FormData brand check is loaded lazily so the fetch machinery is not pulled in upfront.

Note for changelog purposes: this is a behavior change relative to v6, which encoded cross-realm FormData successfully (v7.0.0 through 7.19.x failed these bodies with an unrelated TypeError; 7.20.0+ hung).

## Changes

- `lib/core/request.js`: FormData-like bodies that are not instances of undici's `FormData` now throw `InvalidArgumentError` instead of falling through to iterable handling.
- `test/issue-5520.js`: cross-realm FormData rejection over HTTP/1 and HTTP/2 (own-FormData encoding is already covered by `test/client-request.js` and `test/http2-body.js`).

### Features

N/A

### Bug Fixes

- `request()` with a FormData from another realm now rejects immediately with a clear error instead of hanging forever.

N/A (the affected input previously hung or produced a broken body; it never worked on v7/v8)

## Status

- [x] Tested (`test/issue-5520.js` plus unit, fetch, client-request, http2-body and h2 core suites; lint clean)
- [x] Documented (behavior fix; error message self-describes)
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -47,9 +47,6 @@ const kHandler = Symbol('handler')
 const kController = Symbol('controller')
 const kResume = Symbol('resume')
 
-// Lazily loaded brand check for undici's own FormData class; only needed
-// when a FormData-like body is passed, so the fetch machinery is not
-// loaded upfront.
 let isUndiciFormData
 
 class RequestController {
@@ -212,11 +209,6 @@ class Request {
     } else if (typeof body === 'string') {
       this.body = body.length ? Buffer.from(body) : null
     } else if (isFormDataLike(body)) {
-      // Only undici's own FormData can be encoded. Instances from other
-      // realms (Node.js' builtin, another undici copy, a polyfill) would
-      // otherwise fall through to the iterable handling and produce a
-      // broken body, historically hanging the request.
-      // https://github.com/nodejs/undici/issues/5520
       isUndiciFormData ??= require('../web/webidl').webidl.is.FormData
       if (!isUndiciFormData(body)) {
         throw new InvalidArgumentError('FormData body must be an instance of undici\'s FormData class. FormData objects from other realms (e.g. Node.js\' global FormData) are not supported.')
--- a/lib/core/request.js
+++ b/lib/core/request.js
@@ -47,6 +47,11 @@ const kHandler = Symbol('handler')
 const kController = Symbol('controller')
 const kResume = Symbol('resume')
 
+// Lazily loaded brand check for undici's own FormData class; only needed
+// when a FormData-like body is passed, so the fetch machinery is not
+// loaded upfront.
+let isUndiciFormData
+
 class RequestController {
   #paused = false
   #reason = null
@@ -206,7 +211,18 @@ class Request {
       this.body = body.byteLength ? Buffer.from(body) : null
     } else if (typeof body === 'string') {
       this.body = body.length ? Buffer.from(body) : null
-    } else if (isFormDataLike(body) || isIterable(body) || isBlobLike(body)) {
+    } else if (isFormDataLike(body)) {
+      // Only undici's own FormData can be encoded. Instances from other
+      // realms (Node.js' builtin, another undici copy, a polyfill) would
+      // otherwise fall through to the iterable handling and produce a
+      // broken body, historically hanging the request.
+      // https://github.com/nodejs/undici/issues/5520
+      isUndiciFormData ??= require('../web/webidl').webidl.is.FormData
+      if (!isUndiciFormData(body)) {
+        throw new InvalidArgumentError('FormData body must be an instance of undici\'s FormData class. FormData objects from other realms (e.g. Node.js\' global FormData) are not supported.')
+      }
+      this.body = body
+    } else if (isIterable(body) || isBlobLike(body)) {
       this.body = body
     } else {
       throw new InvalidArgumentError('body must be a string, a Buffer, a Readable stream, an iterable, or an async iterable')
--- a/test/issue-5520.js
+++ b/test/issue-5520.js
@@ -4,61 +4,26 @@ const { test } = require('node:test')
 const { createServer } = require('node:http')
 const { createServer: createH2Server } = require('node:http2')
 const { once } = require('node:events')
-const { request, FormData: UndiciFormData, H2CClient, errors } = require('..')
+const { request, H2CClient, errors } = require('..')
 
 // https://github.com/nodejs/undici/issues/5520
-// request() accepts FormData bodies via a duck-typed check
-// (util.isFormDataLike), but only undici's own FormData can be encoded.
-// Instances from other realms (e.g. Node.js' builtin globalThis.FormData)
-// used to fall through the brand-checked extraction and were dispatched as
-// a body that never produced any bytes, hanging the request until an
-// external timeout. They are now rejected upfront with InvalidArgumentError.
 
-function createEchoServer (t) {
+test('request() rejects a cross-realm (Node builtin) FormData body', { timeout: 60000 }, async (t) => {
   const server = createServer((req, res) => {
-    const chunks = []
-    req.on('data', (chunk) => chunks.push(chunk))
-    req.on('end', () => {
-      res.setHeader('content-type', 'application/json')
-      res.end(JSON.stringify({
-        contentType: req.headers['content-type'] ?? null,
-        contentLength: req.headers['content-length'] ?? null,
-        body: Buffer.concat(chunks).toString('utf8')
-      }))
-    })
+    req.on('data', () => {})
+    req.on('end', () => res.end('{}'))
   })
   t.after(() => {
     server.closeAllConnections?.()
     server.close()
   })
-  return server
-}
 
-test('request() rejects a cross-realm (Node builtin) FormData body', { timeout: 60000 }, async (t) => {
-  // The premise of the test: the global FormData is not undici's own class.
-  t.assert.notStrictEqual(globalThis.FormData, UndiciFormData)
-
-  const server = createEchoServer(t)
   server.listen(0)
   await once(server, 'listening')
 
   const form = new globalThis.FormData()
   form.append('field', 'value')
   form.append('file', new Blob(['file contents'], { type: 'text/plain' }), 'hello.txt')
-
-  await t.assert.rejects(
-    request(`http://localhost:${server.address().port}/`, { method: 'POST', body: form }),
-    errors.InvalidArgumentError
-  )
-})
-
-test('request() rejects a string-only cross-realm FormData body', { timeout: 60000 }, async (t) => {
-  const server = createEchoServer(t)
-  server.listen(0)
-  await once(server, 'listening')
-
-  const form = new globalThis.FormData()
-  form.append('a', '1')
 
   await t.assert.rejects(
     request(`http://localhost:${server.address().port}/`, { method: 'POST', body: form }),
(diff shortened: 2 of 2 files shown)


KhafraDev (changes requested):
This shouldn't be under web/fetch, if this is something we want to support at all.

Pyker:
Moved it out of web/fetch in 5faa679: `body.js` is now untouched and the rebuild lives in `lib/dispatcher/formdata-body.js`, next to its only consumers.

On supporting it at all: the h1/h2 clients already accept any FormData-like through `util.isFormDataLike()`, and today that acceptance ends in a silent hang with zero bytes on the wire. So the real choice is encode it (this PR, matches v6) or tighten the gate to the brand check and throw `InvalidArgumentError` (loud, but breaks `request()` + Node's global FormData, which worked on v6). I prefer the first; happy to rework to the second if that is the direction you want.

mcollina (approved):
lgtm

KhafraDev:
Is this something we want to support? We don't support this pattern anywhere else. Wouldn't it make more sense to just throw an error?

mcollina:
> Is this something we want to support? We don't support this pattern anywhere else. Wouldn't it make more sense to just throw an error?

Actually yes, it would be.

Pyker:
Reworked to reject, per the discussion: the `Request` constructor now throws `InvalidArgumentError` for FormData-like bodies that are not undici's own FormData, alongside the existing invalid-body rejections, so h1 and h2 both fail fast at dispatch time. The encode helper is gone and web/fetch is untouched (the brand check is lazy-loaded). One note for future bisecters: v6 encoded these successfully, so this formalizes the stricter v7+ behavior with a clear error instead of the hang.

lib/core/request.js
```diff
@@ -47,6 +47,11 @@ const kHandler = Symbol('handler')
 const kController = Symbol('controller')
 const kResume = Symbol('resume')
 
+// Lazily loaded brand check for undici's own FormData class; only needed
```
KhafraDev: we don't need a comment explaining lazy loading

lib/core/request.js
```diff
@@ -206,7 +211,18 @@ class Request {
       this.body = body.byteLength ? Buffer.from(body) : null
     } else if (typeof body === 'string') {
       this.body = body.length ? Buffer.from(body) : null
-    } else if (isFormDataLike(body) || isIterable(body) || isBlobLike(body)) {
+    } else if (isFormDataLike(body)) {
+      // Only undici's own FormData can be encoded. Instances from other
```
KhafraDev: the contents of this comment are explained right below in the error message

test/issue-5520.js
```diff
@@ -0,0 +1,145 @@
+'use strict'
+
+const { test } = require('node:test')
+const { createServer } = require('node:http')
+const { createServer: createH2Server } = require('node:http2')
+const { once } = require('node:events')
+const { request, FormData: UndiciFormData, H2CClient, errors } = require('..')
+
+// https://github.com/nodejs/undici/issues/5520
+// request() accepts FormData bodies via a duck-typed check
```
KhafraDev: this comment should be removed as it's likely that the implementation will change in the future, just a link to the issue will be fine

test/issue-5520.js
```diff
@@ -0,0 +1,145 @@
...
+  t.after(() => {
+    server.closeAllConnections?.()
+    server.close()
+  })
+  return server
+}
+
+test('request() rejects a cross-realm (Node builtin) FormData body', { timeout: 60000 }, async (t) => {
+  // The premise of the test: the global FormData is not undici's own class.
+  t.assert.notStrictEqual(globalThis.FormData, UndiciFormData)
```
KhafraDev:
```suggestion
```

this is obviously true

test/issue-5520.js
```diff
@@ -0,0 +1,145 @@
...
+  await once(server, 'listening')
+
+  const form = new globalThis.FormData()
+  form.append('a', '1')
+
+  await t.assert.rejects(
+    request(`http://localhost:${server.address().port}/`, { method: 'POST', body: form }),
+    errors.InvalidArgumentError
+  )
+})
```
KhafraDev:
```suggestion
```

this is basically the exact same as the first test

test/issue-5520.js
```diff
@@ -0,0 +1,145 @@
...
+
+  t.assert.strictEqual(res.statusCode, 200)
+  const boundary = /^multipart\/form-data; boundary=(\S+)$/.exec(echo.contentType)?.[1]
+  t.assert.ok(boundary, `expected multipart content-type, got ${echo.contentType}`)
+  t.assert.strictEqual(echo.contentLength, String(Buffer.byteLength(echo.body)))
+  t.assert.ok(echo.body.includes('Content-Disposition: form-data; name="field"\r\n\r\nvalue with spaces\r\n'))
+  t.assert.ok(echo.body.includes('Content-Disposition: form-data; name="file"; filename="hello.txt"'))
+  t.assert.ok(echo.body.includes('file contents'))
+  t.assert.ok(echo.body.endsWith(`--${boundary}--\r\n`))
+})
```
KhafraDev:
```suggestion
```

this is already tested

test/issue-5520.js
```diff
@@ -0,0 +1,145 @@
...
+
+test('HTTP/2 client still encodes undici\'s own FormData body', { timeout: 60000 }, async (t) => {
```
KhafraDev: I'm less familiar with the h2 tests, but this is likely tested already

lib/core/request.js
Pyker: Removed in 6db3923.

lib/core/request.js
Pyker: Removed.

test/issue-5520.js
Pyker: Trimmed to the issue link.

test/issue-5520.js
Pyker: Removed.

test/issue-5520.js
Pyker: Removed.

test/issue-5520.js
Pyker: Removed; covered by `test/client-request.js` ('request with FormData body').

test/issue-5520.js
Pyker: Removed; covered by `test/http2-body.js` ('#3803 - sending FormData bodies works').

Pyker:
@KhafraDev Could you clarify why this was closed? I reworked it to reject cross-realm FormData as requested and addressed the subsequent inline feedback. #5520 is still open. Has this been superseded by another fix, or is there a remaining objection to this approach?

KhafraDev:
I adopted a new review policy. If a human doesn't write the description - at the very minimum - I do not bother reviewing. I am just as capable of asking Claude or Astra myself.

Pyker:
Okay, that's a fair choice.

PR #5534, opened 2026-07-09
feat(mock): support asynchronous reply options callbacks

### LINKED ISSUE #1348: Allow asynchronous mocking replies
## This would solve...

When trying to create a Undici adapter for [Polly.JS](https://netflix.github.io/pollyjs/#/) I tried to use the mocking capabilities from Undici to record network requests and replay them later using stored HAR files. However, the `MockInterceptor.reply` method doesn't support asynchronous callbacks. So receiving the request, loading a file from the filesystem (asynchronously) (or passing through the request) and resolving with a promise doesn't work.

> Sidenote: `nock` supports asynchronous callbacks: https://github.com/nock/nock#specifying-replies

The `MockInterceptor.reply` method could accept asynchronous callbacks:

```typescript
mockPool
  .intercept({
    path: () => true
  })
  .reply(async (options) => {
    const response = await fetchFromFileSystem(options)
    return mapping(response)
  })
```

## I have also considered...

Creating my own version of `MockAgent` that supports async replies, but that just seemed too much work for a possibly temporary problem.

## Additional context

[Feature request for Undici adapter for Polly.JS](https://github.com/Netflix/pollyjs/issues/443)
[Feature request for Undici support for Nock](https://github.com/nock/nock/issues/2183)

Develop-KIM:
## This relates to...

Fixes #1348 (see [my comment there](https://github.com/nodejs/undici/issues/1348#issuecomment-4922160259) clarifying the scope vs #1757/#1758)

## Rationale

#1758 made the **data argument** awaitable (`reply(200, async (opts) => body)`), but the ask in #1348 — an async **reply options callback** — still fails on `main`:

```js
mockPool.intercept({ path: '/fixture' }).reply(async (opts) => ({
  statusCode: 200,
  data: await readFile('./fixture.json')
}))
// InvalidArgumentError: statusCode must be defined
```

`wrappedDefaultsCallback` assumes the callback returns a plain object: a Promise passes the `typeof === 'object'` check, contributes nothing when spread into `{ data: '', responseOptions: {}, ...promise }`, and validation throws. This is what Polly.JS/nock-style record-and-replay adapters need (loading recorded responses from disk asynchronously).

## Changes

Extends the existing `isPromise` pattern from #1758:

- `lib/mock/mock-interceptor.js`: `wrappedDefaultsCallback` detects a promise result and resolves the dispatch data (format check + validation + `createMockScopeDispatchData`) once it settles.
- `lib/mock/mock-utils.js`: `mockDispatch` detects a promise from the dispatch-data callback and defers the reply (extracted into `dispatchMockReply`) until it resolves. A rejection can no longer throw synchronously from dispatch, so it is surfaced via `handler.onResponseError` — the same route as `replyWithError`. Sync callbacks keep their synchronous-throw semantics.
- Consumed/pending bookkeeping now happens before callback resolution, so a dispatch pending on an async callback cannot be re-matched beyond its allowed `times` — consistent with how async data callbacks already behave (they mark consumption at dispatch time).
- Types: `MockReplyOptionsCallback` may return a `Promise` of the reply options (extracted a named `MockReplyOptions` type).
- Docs: `MockPool.md` — callback description + a new async example.
- Tests: async happy path, `defaultReplyHeaders`/`replyContentLength` interaction, `times()` bookkeeping, invalid resolved format, and rejection propagation; tsd assertions for the async signature.

### Features

- `MockInterceptor.reply()` accepts an asynchronous reply options callback.

### Bug Fixes

N/A

N/A

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/docs/docs/api/MockPool.md
+++ b/docs/docs/api/MockPool.md
@@ -136,7 +136,8 @@ The reply behaviour of a matching request is defined through the returned
   computing all reply options dynamically rather than just the body.
   * `callback` {Function} A `(opts: MockResponseCallbackOptions) =>
     { statusCode, data, responseOptions }` function invoked with the incoming
-    request.
+    request. The callback may be asynchronous; a returned promise is awaited
+    and must resolve to the same shape.
   * Returns: {MockScope}
 * `replyWithError(error)` {Function} Defines an error for a matching request to
   throw.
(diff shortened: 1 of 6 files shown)


mcollina (approved):
lgtm

Issue #5522, opened 2026-07-07
`normalizeHeaders` fails with `opts.headers is not a valid header map` depending on interceptor ordering


peterhirn:
## Bug Description

Depending on the order of the `dns` and `cache` interceptor calls fail with `Error: opts.headers is not a valid header map`.

## Reproduction

**Example 1, `dns then cache` vs. `cache then dns` with empty array as headers:**

```js
import { Agent, interceptors, request } from "undici";

const { cache, dns } = interceptors;

const dnsThenCache = new Agent().compose(dns(), cache());

const cacheThenDns = new Agent().compose(cache(), dns());

const url = "https://www.google.com";

// NOTE: `const headers = {};` works
const headers = [];

const r1 = await request(url, {
  dispatcher: dnsThenCache,
  headers,
});

console.log(r1.statusCode);
r1.body.destroy();

const r2 = await request(url, {
  dispatcher: cacheThenDns,
  headers,
});

console.log(r2.statusCode);
r2.body.destroy();
```

```
200
node_modules/.pnpm/undici@8.7.0/node_modules/undici/lib/util/cache.js:47
          throw new Error('opts.headers is not a valid header map')
          ^

Error: opts.headers is not a valid header map
    at normalizeHeaders (node_modules/.pnpm/undici@8.7.0/node_modules/undici/lib/util/cache.js:47:17)
    at node_modules/.pnpm/undici@8.7.0/node_modules/undici/lib/interceptor/cache.js:468:18
    at node_modules/.pnpm/undici@8.7.0/node_modules/undici/lib/interceptor/dns.js:563:9
    at node_modules/.pnpm/undici@8.7.0/node_modules/undici/lib/interceptor/dns.js:201:9
    at GetAddrInfoReqWrap.callback (node_modules/.pnpm/undici@8.7.0/node_modules/undici/lib/interceptor/dns.js:262:9)
    at GetAddrInfoReqWrap.onlookupall [as oncomplete] (node:dns:133:8)
```

**Example 2, `dns then cache` vs. `cache then dns` with no headers and redirect:**

```js
import { Agent, interceptors, request } from "undici";

const { cache, dns, redirect } = interceptors;

const dnsThenCache = new Agent().compose(dns(), cache(), redirect({ maxRedirections: 3 }));

const cacheThenDns = new Agent().compose(cache(), dns(), redirect({ maxRedirections: 3 }));

const url = "https://google.com"; // redirects to https://www.google.com

const r1 = await request(url, {
[...]

PR #5535, opened 2026-07-09
test: add cache/dedupe + dns re-dispatch integration tests

GiHoon1123:
Relates to #5522.

#5536 contains the implementation fix and unit tests for flat header arrays. This PR keeps the integration coverage for the compose-chain paths that triggered the original bug.

The tests cover:

- `cache() + dns()` after DNS re-dispatch
- `deduplicate() + dns()`, which uses the same cache-key utilities

The earlier `lib/util/cache.js` changes were dropped after #5536 landed. This PR only changes `test/interceptors/dns.js`.

`node --test test/interceptors/dns.js`: 29 passing, 0 failing (3 IPv6-conditional skips).

### DIFF
--- a/test/interceptors/dns.js
+++ b/test/interceptors/dns.js
@@ -12,7 +12,7 @@ const { tspl } = require('@matteo.collina/tspl')
 const pem = require('@metcoder95/https-pem')
 
 const { interceptors, Agent, Client, Pool, request } = require('../..')
-const { dns } = interceptors
+const { dns, cache, deduplicate } = interceptors
 
 // Helper to check if IPv6 is available for localhost
 // This is called synchronously at test definition time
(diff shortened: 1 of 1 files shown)


lib/util/cache.js
```diff
@@ -31,13 +31,36 @@ function makeCacheKey (opts) {
...
- * @returns {Record<string, string[] | string>}
+ * @param {unknown} val
+ * @returns {boolean}
+ */
+function isValidHeaderValue (val) {
+  if (typeof val === 'string') {
+    return true
+  }
+
+  return Array.isArray(val) && val.every(v => typeof v === 'string')
```
mcollina: use a proper for loop, `.every()` is slow.

metcoder95 (approved):

GiHoon1123:
With #5536 now merged, I dropped the overlapping implementation changes from this PR and kept the integration
tests for the dns re-dispatch paths through cache() and deduplicate().

The tests are passing, and this PR is ready for merge.

mcollina (approved):
lgtm

PR #5497, opened 2026-07-03
feat(h2): enable concurrency on idempotent requests

metcoder95:
## This relates to...

(Partially) Closes #5494

## Rationale

Given the recent improvements on isolation for the H2 client, this check does not seems to be required any longer.

## Status

- [ ] I have read and agreed to the [Developer's Certificate of Origin][cert]
- [ ] Tested
- [ ] Benchmarked (**optional**)
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

[cert]: https://github.com/nodejs/undici/blob/main/CONTRIBUTING.md#developers-certificate-of-origin

PR #5538, opened 2026-07-09
fix(h2): allow stream-bodied requests to multiplex on a busy session

### LINKED ISSUE #5494: HTTP2 POST requests over `fetch()` to the same origin can't run concurrently
## Bug Description

Concurrent POST requests to the same origin using `fetch()` execute without concurrency when that origin supports h2.

This is because `fetch()` POST bodies are converted to a `ReadableStream`, and requests with stream bodies do not (currently) execute with concurrency on the same http2 connection. Meanwhile, because a single h2 connection supports inter-request concurrency, the agent won't open more than one h2 connection to the same origin. The confluence of these two factors means that if you're making highly-concurrent POST requests to the same origin, when h2 is enabled, they'll actually execute sequentially instead.

## Reproducible By

I set up a minimal reproduction in this repo: https://github.com/wgoodall01/repro-undici-h2

- The server script handles each request over h1 or h2, does `sleep()` and responds. It also tracks stats on the max count of inflight requests.
- The client script makes concurrent test requests over HTTP1/HTTP2, `fetch()` or undici `request()`, and GET or POST, printing stats for each combination of the above.
- You can run both with `npm install && npm run repro`

On my machine, it outputs the following:

```
$ npm run repro
npm run repro

> repro
> node server.mjs & trap 'kill $!' EXIT; node client.mjs

server: listening on https://localhost:4433 (holdMs=750)
undici 8.6.0, nodejs v26.4.0

H2 fetch() POST    proto=2.0  peak=1/12  wall=9060ms
    sendHeaders offsets (ms): 2, 758, 1513, 2269, 3023, 3777, 4531, 5284, 6039, 6795, 7551, 8305
H2 fetch() GET     proto=2.0  peak=12/12  wall= 759ms
    sendHeaders offsets (ms): 0, 0, 0, 1, 1, 1, 1, 1, 1, 1, 1, 1
H2 request() POST  proto=2.0  peak=12/12  wall= 761ms
    sendHeaders offsets (ms): 1, 1, 1, 1, 1, 1, 1, 1, 1, 1, 2, 2
H2 request() GET   proto=2.0  peak=12/12  wall= 754ms
    sendHeaders offsets (ms): 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 1
H1 fetch() POST    proto=1.1  peak=12/12  wall= 765ms
    sendHeaders offsets (ms): 5, 8, 8, 8, 9, 9, 9, 9, 9, 10, 10, 10
H1 fetch() GET     proto=1.1  peak=12/12  wall= 760ms
    sendHeaders offsets (ms): 3, 5, 5, 5, 6, 6, 6, 6, 6, 6, 6, 7
H1 request() POST  proto=1.1  peak=12/12  wall= 761ms
    sendHeaders offsets (ms): 3, 6, 6, 7, 7, 7, 7, 8, 8, 8, 8, 9
H1 request() GET   proto=1.1  peak=12/12  wall= 759ms
    sendHeaders offsets (ms): 2, 4, 4, 4, 4, 4, 5, 5, 5, 5, 5, 5
```

You can see that the `H2 fetch() POST` line indicates the requests aren't running concurrently.

## Expected Behavior

The `H2 fetch() POST` line reports similar wall-clock latency and sendHeaders offsets to the other test cases.

## Environment

```
$ node --version
v26.4.0

$ uname -a
Darwin $HOSTNAME.local 24.6.0 Darwin Kernel Version 24.6.0: Tue Apr 21 20:18:11 PDT 2026; root:xnu-11417.140.69.710.16~1/RELEASE_ARM64_T6020 arm64

$ sw_vers
ProductName:		macOS
ProductVersion:		15.7.7
BuildVersion:		24G720
```

## Additional context

I ran into this making requests to the OpenAI API, which supports H2, and where every request happens to be a POST. This bug means that a single nodejs instance can only make a single LLM request concurrently—a significant bottleneck!

## Prior Art

- Issue #5390 reported the original issue in `undici@8.4.0` with concurrent POST requests over H2
- PR #5391 fixed POST multiplexing for non-stream bodies in `undici.request()`, however, that fix doesn't apply to `fetch()` requests. 
    - @metcoder95 asked for a MRE in [this comment](https://github.com/nodejs/undici/pull/5391#issuecomment-4689378305), and here it is :)

## LLM Disclosure

- I wrote everything in this issue myself. 
- I used Claude Fable 5 to track down this issue in my codebase and put together a first pass at the repro, which I then hand-edited for clarity.

### LINKED ISSUE #5524: HTTP/2: `fetch` POST hangs while an SSE stream is open
## Bug Description

On an HTTP/2 session, a `fetch()` request that carries a body (e.g. a plain
string POST) is never written to the wire while another stream on the same
session is still receiving its response body. With a long-lived response such as
`text/event-stream`, the request is deferred forever:

- it is not multiplexed onto the session, even though HTTP/2 streams are
  independent (curl against the same server multiplexes fine), and
- no second connection is opened, because with an h2 session attached the client
  intentionally stops reporting pending requests as "busy" so the `Pool` keeps
  multiplexing instead of connecting.

The result is a silent deadlock until the stream ends -- for SSE, never.
Bodyless requests (GET/HEAD) are not affected. HTTP/1.1 is not affected (the
pool correctly opens a second connection).

## Reproduction

**Standalone reproduction script** (uses cleartext h2 via `useH2c` to stay
certificate-free; the same happens over TLS when ALPN negotiates h2, which
`allowH2: true` -- the default -- does against any h2-capable server):

```javascript
'use strict'

const { createServer } = require('node:http2')
const { fetch, Agent } = require('undici')

test('fetch POST dispatches while an SSE stream is open on the same h2 session', { timeout: 30000 }, async (t) => {
  const server = createServer()
  server.on('stream', (stream, headers) => {
    if (headers[':method'] === 'GET') {
      stream.respond({ ':status': 200, 'content-type': 'text/event-stream' })
      stream.write(': ping\n\n') // stream is held open, like a real SSE endpoint
    } else {
      stream.respond({ ':status': 200, 'content-type': 'application/json' })
      stream.end('{"ok":true}')
    }
  })
  server.listen(0, '127.0.0.1')

  const dispatcher = new Agent({ useH2c: true })
  const sse = new AbortController()
  t.after(async () => {
    sse.abort()
    await dispatcher.destroy()
    server.close()
  })
  const origin = `http://127.0.0.1:${server.address().port}`

  // Establish the h2 session.
  const warmup = await fetch(origin, { method: 'POST', body: '{"warmup":1}', dispatcher })
  await warmup.text()

  // Open a never-ending SSE stream on that session.
  fetch(origin, { dispatcher, signal: sse.signal }).catch(() => {})

  // A POST to the same origin should multiplex onto the session (or get a
  // new connection) and resolve immediately. With the bug it is never
  // written to the wire and times out.
  const res = await fetch(origin, {
    method: 'POST',
    body: '{"real":1}',
    dispatcher,
    signal: AbortSignal.timeout(5000),
  })
  assert.strictEqual(res.status, 200)
})
```

Also reproducible with the default global `fetch` and no options at all: serve
the same handler from `node:http2.createSecureServer` with a self-signed
certificate, warm the session with one POST, `fetch(origin)` to open the SSE
stream, then `fetch(origin, { method: 'POST', body: '...' })` -- ALPN negotiates
h2 and the POST hangs. Equally reproducible against public h2 servers that hold
an SSE stream open (see Additional context).

## Expected Behavior

The second POST resolves immediately: HTTP/2 streams are independent, so the
request should be multiplexed onto the open session (this is what curl does
against the same server), or failing that, dispatched on a new connection.

## Actual Behavior

The POST times out (`TimeoutError: The operation was aborted due to timeout`);
without a signal it hangs indefinitely. Diagnostics channels show the request is
never dispatched: `undici:client:sendHeaders` never fires for it and
`undici:client:connected` fires exactly once (no second connection). It stays
queued until the SSE stream ends.

## Logs & Screenshots

`node --test` output of the script above:

```
TimeoutError: The operation was aborted due to timeout
    at new DOMException (node:internal/per_context/domexception:79:18)
    at Timeout._onTimeout (node:internal/abort_controller:210:9)
...
tests 1 / pass 0 / fail 1 (5019ms)
```

Instrumented dispatch decisions (h2 session, warm socket; logging added to
`Client[kDispatch]`):

```
[client.dispatch] GET  resuming=0 busy=false needDrain=0 -> returns true   <- SSE stream accepted
[client.dispatch] POST resuming=1 busy=false needDrain=0 -> returns true   <- accepted, then parked forever
```

The same sequence over HTTP/1.1 reports `busy=true -> returns false`, the pool
opens a second connection, and the POST completes.

## Environment

- OS: macOS (Darwin 25.5.0)
- Node.js version: v26.4.0
- undici version: 8.7.0 (npm); also reproduced with 8.5.0 as bundled in Node
  v26.4.0

### Additional context

Root cause, as far as we can tell:

- `lib/dispatcher/client-h2.js` -> `busy(request)`: when `kRunning > 0`, a
  request whose body is a stream/async iterable is deferred until all in-flight
  streams complete ("Request with stream or iterator body cannot be retried.
  Ensure that no other requests are inflight."). `fetch()` extracts every
  non-empty body to a stream, so all bodied fetch requests hit this guard --
  which contradicts the comment a few lines above it stating that h2 dispatches
  non-idempotent requests concurrently.
- `lib/dispatcher/client.js` -> `get [kBusy]`: with an h2 context attached,
  `kPending > 0` is deliberately not reported as busy ("we want concurrent
  dispatches to multiplex onto the shared session"), so `Pool` never opens a
  second connection. Combined with the guard above, a deferred bodied request
  has no escape route.
- Requests without bodies skip the guard, so GETs multiplex fine -- only bodied
  requests starve.

Real-world impact: MCP (Model Context Protocol) StreamableHTTP clients on Node
hang against h2-hosted MCP servers (e.g. Cloudflare Workers: try
`https://docs.mcp.cloudflare.com/mcp`). The transport opens a standalone SSE GET
right after the handshake and then POSTs JSON-RPC calls (`tools/list`, tool
invocations) to the same origin; every POST parks behind the SSE stream and the
client times out. Reports in that ecosystem (e.g. geelen/mcp-remote#226) look
like this bug misattributed. The blast radius grew when `allowH2` became `true`
by default.

## Workarounds

- `new Agent({ allowH2: false })` (per request or via `setGlobalDispatcher`):
  forces HTTP/1.1, where the pool opens a second connection past the busy one.
- Give long-lived streams their own dispatcher: with the SSE request on a
  separate `Agent`, bodied requests no longer share its session and dispatch
  immediately (verified against the reproduction script).

GiHoon1123:
## This relates to...

Fixes #5524
Fixes #5494

Same root cause for both -- #5494 already has an open fix in #5497 (same core change to `busy()`, currently blocked on a requested unit test) -- see "Relationship to #5497" below before reviewing.

## Rationale

`busy()` in `lib/dispatcher/client-h2.js` deferred any request with a stream/async-iterable/FormData body until every other in-flight request on the h2 session completed. If one of those in-flight requests is a long-lived stream (e.g. an open `text/event-stream` GET), every bodied request queued behind it parks indefinitely. The client also stops reporting `kBusy` while an h2 context is attached (so the pool keeps multiplexing instead of opening a second connection), so there's no escape route -- for an SSE stream, the hang is permanent.

The two hazards the guard's comments cited are already handled more precisely elsewhere in this file:
- "can error while other requests are inflight and indirectly error those as well" -- a stream erroring mid-request now only aborts that one stream via the `abort()` closure (explicit comment there: "We do not destroy the socket as we can continue using the session"); it doesn't touch sibling streams.
- "cannot be retried... could cause failure" -- `canRetryRequestAfterGoAway()` already excludes stream/iterable/FormData bodies from the requests it resurrects after a GOAWAY (`body == null || isBuffer || isBlobLike` only), erroring the rest instead of replaying a partially-consumed body.

Both of those were already in place before this change, so the blanket pre-dispatch guard was redundant.

## Relationship to #5497

#5497 removes the same 9 lines for the same reason, opened against #5494 (concurrent POSTs on h2 not running concurrently -- a throughput complaint). #5524 is a more severe symptom of the identical guard: not just serialized, but a permanent deadlock when one of the concurrent requests is a long-lived stream like SSE. I only found #5497 after finishing this fix independently.

This PR adds what #5497 doesn't currently have:
- The unit test `mcollina` requested on #5497 (still outstanding as of this PR) -- covering #5494's own scenario directly: a second stream-bodied POST now dispatches while the first is still in flight instead of waiting for it to complete.
- A separate regression test for the more severe #5524 deadlock scenario (SSE held open + bodied POST hangs forever, not just serializes).

Happy to close this in favor of #5497 if the maintainers would rather continue there -- opening it mainly so the missing test and both scenarios' coverage aren't lost. No preference on which one lands.

### Features

N/A

### Bug Fixes

- Fixes #5524: a bodied `fetch()`/`request()` call on an h2 session no longer waits forever behind an open long-lived stream (e.g. SSE) on the same session.
- Fixes #5494: concurrent bodied `fetch()`/`request()` calls on the same h2 session now dispatch concurrently instead of serializing.

N/A -- this only removes a guard that prevented dispatch; nothing that previously worked changes behavior, only what could not previously proceed now can.

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -322,16 +322,6 @@ function connectH2 (client, socket) {
           // Don't dispatch an upgrade until all preceding requests have completed.
           // Possibly, we do not have remote settings confirmed yet.
           if ((request.upgrade === 'websocket' || request.method === 'CONNECT') && session[kRemoteSettings] === false) return true
-          // Request with stream or iterator body can error while other requests
-          // are inflight and indirectly error those as well.
-          // Ensure this doesn't happen by waiting for inflight
-          // to complete before dispatching.
-
-          // Request with stream or iterator body cannot be retried.
-          // Ensure that no other requests are inflight and
-          // could cause failure.
-          if (util.bodyLength(request.body) !== 0 &&
-            (util.isStream(request.body) || util.isAsyncIterable(request.body) || util.isFormDataLike(request.body))) return true
         } else {
           return (request.upgrade === 'websocket' || request.method === 'CONNECT') && session[kRemoteSettings] === false
         }
--- a/test/http2-pipelining-default.js
+++ b/test/http2-pipelining-default.js
@@ -147,8 +147,19 @@ test('fetch POST multiplexes while an SSE stream is open on the same h2 session 
 
   const dispatcher = new Agent({ useH2c: true })
   const sse = new AbortController()
+
+  // Track the SSE fetch and its response so teardown can wait for the abort
+  // to actually settle -- and explicitly cancel the never-read response
+  // body -- before closing the dispatcher. Without this, an abort that's
+  // still propagating when the dispatcher closes can surface a late,
+  // listener-less socket error (e.g. ECONNRESET) as an uncaughtException
+  // after the test has already completed.
+  let sseResponse
+  let ssePromise = Promise.resolve()
   after(async () => {
     sse.abort()
+    await ssePromise
+    await sseResponse?.body?.cancel().catch(() => {})
     await dispatcher.close()
   })
 
@@ -161,9 +172,11 @@ test('fetch POST multiplexes while an SSE stream is open on the same h2 session 
   })
   await warmup.text()
 
-  fetch(`${origin}/events`, {
+  ssePromise = fetch(`${origin}/events`, {
     dispatcher,
     signal: sse.signal
+  }).then(res => {
+    sseResponse = res
   }).catch(() => {})
   await eventsOpenedPromise
 
--- a/test/http2-pipelining-default.js
+++ b/test/http2-pipelining-default.js
@@ -1,12 +1,12 @@
 'use strict'
 
 const { test, after } = require('node:test')
-const { createSecureServer } = require('node:http2')
+const { createSecureServer, createServer } = require('node:http2')
 const { once } = require('node:events')
 const { tspl } = require('@matteo.collina/tspl')
 const pem = require('@metcoder95/https-pem')
 
-const { Client, Pool } = require('..')
+const { Agent, Client, Pool, fetch } = require('..')
 
 test('h2 client multiplexes concurrent requests by default (#4143)', async t => {
   const N = 5
@@ -109,6 +109,75 @@ test('Pool with connections=1 multiplexes h2 streams on the single session (#414
 
   t.strictEqual(sessions.size, 1, `expected 1 h2 session, got ${sessions.size}`)
   t.strictEqual(sockets.size, 1, `expected 1 TCP socket, got ${sockets.size}`)
+
+  await t.completed
+})
+
+test('fetch POST multiplexes while an SSE stream is open on the same h2 session (#5524)', async t => {
+  t = tspl(t, { plan: 4 })
+
+  const server = createServer()
+  const paths = []
+  const sessions = new Set()
+  let eventsOpened
+  const eventsOpenedPromise = new Promise(resolve => {
+    eventsOpened = resolve
+  })
+
+  server.on('session', session => {
+    sessions.add(session)
+  })
+
+  server.on('stream', (stream, headers) => {
+    paths.push(headers[':path'])
+
+    if (headers[':path'] === '/events') {
+      stream.respond({ ':status': 200, 'content-type': 'text/event-stream' })
+      stream.write(': ping\n\n')
+      eventsOpened()
+      return
+    }
+
+    stream.respond({ ':status': 200, 'content-type': 'application/json' })
+    stream.end('{"ok":true}')
+  })
+
+  await once(server.listen(0, '127.0.0.1'), 'listening')
+  after(() => server.close())
+
+  const dispatcher = new Agent({ useH2c: true })
+  const sse = new AbortController()
+  after(async () => {
+    sse.abort()
+    await dispatcher.close()
+  })
+
+  const origin = `http://127.0.0.1:${server.address().port}`
+
+  const warmup = await fetch(`${origin}/warmup`, {
+    method: 'POST',
+    body: '{"warmup":true}',
+    dispatcher
+  })
+  await warmup.text()
+
+  fetch(`${origin}/events`, {
+    dispatcher,
+    signal: sse.signal
+  }).catch(() => {})
+  await eventsOpenedPromise
+
+  const response = await fetch(`${origin}/rpc`, {
+    method: 'POST',
+    body: '{"ok":true}',
+    dispatcher,
+    signal: AbortSignal.timeout(5000)
+  })
+
+  t.strictEqual(response.status, 200)
+  t.strictEqual(await response.text(), '{"ok":true}')
+  t.deepStrictEqual(paths, ['/warmup', '/events', '/rpc'])
+  t.strictEqual(sessions.size, 1)
 
   await t.completed
 })
--- a/test/http2-pipelining-default.js
+++ b/test/http2-pipelining-default.js
@@ -182,6 +182,68 @@ test('fetch POST multiplexes while an SSE stream is open on the same h2 session 
   await t.completed
 })
 
+test('fetch POST bodies dispatch concurrently on the same h2 session instead of serializing (#5494)', async t => {
+  t = tspl(t, { plan: 3 })
+
+  const server = createServer()
+  const sessions = new Set()
+  const streams = []
+  let resolveSecondArrived
+  const secondArrived = new Promise(resolve => {
+    resolveSecondArrived = resolve
+  })
+
+  server.on('session', session => {
+    sessions.add(session)
+  })
+
+  server.on('stream', stream => {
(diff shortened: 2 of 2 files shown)


mcollina (approved):
lgtm

metcoder95 (commented):
tests seems unhappy

GiHoon1123:
Found the real cause -- it wasn't a timing/wait issue (a 2s grace period didn't help either). The client abruptly aborting the still-open SSE stream before closing the dispatcher tears the socket down by reset instead of a clean end, which surfaced as a listener-less ECONNRESET. Fixed by ending the SSE stream from the server side first, so the client sees a normal close instead.

Reproduced 100% locally under Node 24 (not on 22, which is why I hadn't caught it before) and verified 30/30 after the fix, plus a clean `test:unit` run.

mcollina (approved):
lgtm

PR #5539, opened 2026-07-10
fix(websocket): implement receive backpressure for WebSocketStream

### LINKED ISSUE #5503: WebSocketStream: receive backpressure is not implemented — a slow consumer buffers every received message in memory
## Bug Description

`WebSocketStream`'s readable side never applies backpressure. Every parsed message is enqueued into the readable stream unconditionally, whether or not the consumer is reading:

- The `ReadableStream` is set up **without a `pull` algorithm** (only `start` and `cancel`), so consumer demand is never observable: https://github.com/nodejs/undici/blob/cb4c2f1f7231d82bd699aa3a09efb5a235faaa3b/lib/web/websocket/stream/websocketstream.js#L286-L295
- `#onMessage` enqueues unconditionally, and the spec's step 4 ("Apply backpressure to the WebSocket") is a comment with no code: https://github.com/nodejs/undici/blob/cb4c2f1f7231d82bd699aa3a09efb5a235faaa3b/lib/web/websocket/stream/websocketstream.js#L350-L352
- Constructor step 7 ("Apply backpressure to the WebSocket") is a `// TODO`: https://github.com/nodejs/undici/blob/cb4c2f1f7231d82bd699aa3a09efb5a235faaa3b/lib/web/websocket/stream/websocketstream.js#L119-L120

`controller.desiredSize` is never consulted and the socket is never paused on behalf of the consumer. The existing `onSocketData` → `parser.write() === false` → `socket.pause()` / `onParserDrain` → `socket.resume()` pair only reflects the parser's *internal* buffer, which drains as fast as frames can be parsed — it is unrelated to consumer demand.

Consequence: with a consumer that is slower than the sender (or momentarily stalled), every received message piles up in the readable queue — unbounded memory growth — and no TCP backpressure ever reaches the server. Since backpressure is the headline feature of the WebSocketStream API, this defeats its main use case. Note that `webSocket.maxPayloadSize` / `maxFragments` offer no protection here: each individual message is small and valid.

## Reproducible By

`npm i undici ws && node repro.js` — the server (child process) pushes 200k × 1 KiB messages; the client reads **one message per second**.

```js
'use strict'
// undici WebSocketStream: receive backpressure is not implemented.
// Server (child process) pushes 200k x 1 KiB messages; client reads ONE message
// per second. Expected: client memory stays bounded and the server's sends stall.
// Actual: every message is buffered in the client's ReadableStream queue.
//
// npm i undici ws && node repro.js
const { fork } = require('node:child_process')

const MESSAGES = 200_000
const SIZE = 1024 // 1 KiB per message ~= 200 MiB total

if (process.argv[2] === 'server') {
  const wss = new WebSocketServer({ port: 0, perMessageDeflate: false })
  wss.on('listening', () => process.send(wss.address().port))
  wss.on('connection', (ws) => {
    const payload = Buffer.alloc(SIZE, 0x61)
    let sent = 0
    const pump = () => {
      while (sent < MESSAGES) {
        ws.send(payload)
        if (++sent % 100_000 === 0) console.log(`[server] queued ${sent} messages`)
        if (ws.bufferedAmount > 8 * 1024 * 1024) return void setTimeout(pump, 10)
      }
      const check = setInterval(() => {
        if (ws.bufferedAmount === 0) {
          clearInterval(check)
          console.log(`[server] all ${MESSAGES} messages (${MESSAGES * SIZE / 1048576 | 0} MiB) fully flushed to the socket, even though the client read almost nothing`)
        }
      }, 100)
    }
    pump()
  })
} else {
  const server = fork(__filename, ['server'])
  server.once('message', async (port) => {
    const { WebSocketStream } = require('undici')
    const rss = () => `${(process.memoryUsage.rss() / 1048576).toFixed(0)} MiB`
    console.log(`[client] node ${process.version}, undici ${require('undici/package.json').version}, rss before connect: ${rss()}`)

    const stream = new WebSocketStream(`ws://127.0.0.1:${port}`)
    const { readable } = await stream.opened

    // Slow consumer: one read per second. The readable has the default HWM (1),
    // so a backpressuring implementation would stop reading from the socket
    // after ~1 unread message until read() is called.
    for (let t = 1; t <= 10; t++) {
      await new Promise((resolve) => setTimeout(resolve, 1000))
      await reader.read()
      console.log(`[client] t=${t}s read=${t} of ${MESSAGES} rss=${rss()}`)
    }
    server.kill()
    process.exit(0)
  })
}
```

## Expected Behavior

The readable is created with the default high-water mark of 1, so once ~1 unread chunk is queued (`desiredSize <= 0`) the implementation should stop reading from the socket until `read()` pulls, propagating TCP backpressure to the server (this is how Chromium's implementation of the same API behaves). For the repro: client RSS stays roughly flat, and the server's `bufferedAmount` never fully drains because its sends stall against a full TCP window.

## Actual Behavior

The server flushed all ~195 MiB into the socket within the first second while the client had read a single message. Client RSS grew from 75 MiB to 337 MiB (+262 MiB ≈ payload plus per-chunk queue overhead) and all 199,990 unread messages stayed buffered in the readable queue:

```text
[client] node v22.22.1, undici 8.7.0, rss before connect: 75 MiB
(node:2680200) [UNDICI-WSS] Warning: WebSocketStream is experimental! Expect it to change at any time.
[server] queued 100000 messages
[server] queued 200000 messages
[server] all 200000 messages (195 MiB) fully flushed to the socket, even though the client read almost nothing
[client] t=1s read=1 of 200000 rss=337 MiB
[client] t=2s read=2 of 200000 rss=337 MiB
[client] t=3s read=3 of 200000 rss=337 MiB
[client] t=4s read=4 of 200000 rss=337 MiB
[client] t=5s read=5 of 200000 rss=337 MiB
[client] t=6s read=6 of 200000 rss=337 MiB
[client] t=7s read=7 of 200000 rss=337 MiB
[client] t=8s read=8 of 200000 rss=337 MiB
[client] t=9s read=9 of 200000 rss=337 MiB
[client] t=10s read=10 of 200000 rss=337 MiB
```

## Environment

- Linux x64
- Node.js v22.22.1
- undici 8.7.0 (npm); code paths verified unchanged on `main` @ cb4c2f1f7231d82bd699aa3a09efb5a235faaa3b
- ws 8.21.0 (repro server only)

### Additional context

- The spec the implementation cites includes "Apply backpressure to the WebSocket" steps in both the constructor and the "receive a WebSocket message" algorithm, and the explainer names backpressure as the primary motivation for the API: https://github.com/ricea/websocketstream-explainer
- I searched existing issues/PRs (`WebSocketStream backpressure`, `desiredSize`, `flow control`, memory/leak variants) and found no report covering this. Nearest neighbors are the WebSocket umbrella #1811 and the fragment-count DoS fixed by GHSA-vxpw-j846-p89q, which is a different unbounded-growth vector (attacker-controlled fragments rather than a slow consumer).
- A possible fix shape: resume the socket from a `pull` algorithm and, after the `enqueue` in `#onMessage`, `socket.pause()` when `controller.desiredSize <= 0`. Granularity is per socket chunk (one chunk can contain many frames), so the queue may overshoot the HWM by one chunk's worth of messages — consistent with other implementations.

GiHoon1123:
## This relates to...

Fixes #5503

## Rationale

`WebSocketStream`'s readable side never applied backpressure. The `ReadableStream` was constructed without a `pull` algorithm, so consumer demand (`desiredSize`) was never observable, and both spec steps that say "Apply backpressure to the WebSocket" (constructor step 7, and the message-receive algorithm step 4) were `TODO`/no-op. Every parsed message was enqueued unconditionally regardless of whether the consumer was reading, so a slow (or absent) reader let the socket buffer an unbounded number of messages in memory — the issue's repro shows RSS growing from 75 MiB to 337 MiB with 200k unread 1 KiB messages, and the same amount fully flushed on the wire even though the client had read almost nothing.

The existing `onSocketData` → `parser.write() === false` → `socket.pause()` / `onParserDrain` → `socket.resume()` pair only reflects the parser's own internal buffer draining as fast as frames can be parsed — it has nothing to do with consumer demand.

## Changes

Two backpressure sources now gate the socket independently:

- **Parser buffer pressure** (pre-existing): `onSocketData` pauses when `parser.write()` returns `false`; `onParserDrain` clears that flag.
- **Consumer demand** (new): a `pull` algorithm on the `ReadableStream` re-checks on every `reader.read()`; `#onMessage` pauses the socket once `desiredSize <= 0` after enqueueing.

The socket only resumes once **both** are clear (parser not congested *and* `desiredSize > 0`) — either source alone is enough to keep it paused.

While testing this I found a second, related bug: once `close()`/`reader.cancel()` starts the closing handshake, it needs to keep reading off the socket to see the peer's own Close frame. If the socket happened to be paused for receive backpressure at that moment (a very reachable state — any unread backlog leaves it paused), the peer's Close frame was never read and `closed` hung forever. Fixed by unconditionally resuming the socket once closing starts (`#resumeSocketForClose`) — there's no more reader to protect from backpressure at that point, so ignoring both backpressure sources above is intentional, not an oversight.

### Features

N/A

### Bug Fixes

- Fixes #5503: a slow or absent `WebSocketStream` reader no longer buffers unbounded messages in memory or lets the sender flush unlimited data past what the consumer has actually read — receive backpressure now propagates down to pausing the socket, and from there to the underlying TCP connection.
- Also fixes a related hang: `close()` / `reader.cancel()` no longer wait forever on the peer's Close frame if the socket happened to be paused for receive backpressure when closing started.

N/A — `WebSocketStream` is still experimental (`UNDICI-WSS` warning). Consumers that read promptly see no behavior change; slow/absent consumers now correctly experience backpressure instead of unbounded buffering.

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/web/websocket/stream/websocketstream.js
+++ b/lib/web/websocket/stream/websocketstream.js
@@ -33,6 +33,7 @@ class WebSocketStream {
   #readableStream
   /** @type {ReadableStreamDefaultController} */
   #readableStreamController
+  #parserBackpressured = false
 
   // Each WebSocketStream object has an associated writable stream , which is a WritableStream .
   /** @type {WritableStream} */
@@ -47,9 +48,13 @@ class WebSocketStream {
     onConnectionEstablished: (response, extensions) => this.#onConnectionEstablished(response, extensions),
     onMessage: (opcode, data) => this.#onMessage(opcode, data),
     onParserError: (err) => failWebsocketConnection(this.#handler, null, err.message),
-    onParserDrain: () => this.#handler.socket.resume(),
+    onParserDrain: () => {
+      this.#parserBackpressured = false
+      this.#resumeSocketIfNeeded()
+    },
     onSocketData: (chunk) => {
       if (!this.#parser.write(chunk)) {
+        this.#parserBackpressured = true
         this.#handler.socket.pause()
       }
     },
@@ -117,7 +122,9 @@ class WebSocketStream {
     this.#closedPromise = Promise.withResolvers()
 
     // 7. Apply backpressure to the WebSocket.
-    // TODO
+    // The socket and readable stream are not available until the connection is
+    // established; backpressure is applied from the readable pull algorithm and
+    // after messages are enqueued.
 
     // 8.  If options [" signal "] exists ,
     if (options.signal != null) {
@@ -192,6 +199,11 @@ class WebSocketStream {
 
     // 2. Let reason be closeInfo [" reason "].
     const reason = closeInfo.reason
+
+    // The closing handshake needs to read the peer's Close frame off the
+    // socket to complete; a socket paused for receive backpressure (#5503)
+    // would otherwise never see it and the closed promise would hang forever.
+    this.#resumeSocketForClose()
 
     // 3. Close the WebSocket with this , code , and reason .
     closeWebSocketConnection(this.#handler, code, reason, true)
@@ -291,6 +303,7 @@ class WebSocketStream {
       start: (controller) => {
         this.#readableStreamController = controller
       },
+      pull: () => this.#resumeSocketIfNeeded(),
       cancel: (reason) => this.#cancel(reason)
     })
 
@@ -350,6 +363,35 @@ class WebSocketStream {
     this.#readableStreamController.enqueue(chunk)
 
     // 4. Apply backpressure to the WebSocket.
+    this.#applyBackpressureToWebSocket()
+  }
+
+  #applyBackpressureToWebSocket () {
+    if (this.#readableStreamController.desiredSize <= 0) {
+      this.#handler.socket.pause()
+    }
+  }
+
+  #resumeSocketIfNeeded () {
+    if (
+      !this.#parserBackpressured &&
+      this.#handler.socket != null &&
+      !this.#handler.socket.destroyed &&
+      this.#readableStreamController != null &&
+      this.#readableStreamController.desiredSize > 0
+    ) {
+      this.#handler.socket.resume()
+    }
+  }
+
+  // Once closing starts there is no more reader to protect from backpressure,
+  // and the closing handshake needs to keep reading off the socket to see the
+  // peer's Close frame - so unconditionally resume, ignoring both backpressure
+  // sources above.
+  #resumeSocketForClose () {
+    if (this.#handler.socket != null && !this.#handler.socket.destroyed) {
+      this.#handler.socket.resume()
+    }
   }
 
   /** @type {import('../websocket').Handler['onSocketClose']} */
@@ -437,6 +479,9 @@ class WebSocketStream {
       // 3.2. Set reasonString to reason ’s reason .
       reasonString = reason.reason
     }
+
+    // See close() for why this needs to happen before closeWebSocketConnection.
+    this.#resumeSocketForClose()
 
     // 4. Close the WebSocket with stream , code , and reasonString . If this throws an exception,
     //    discard code and reasonString and close the WebSocket with stream .
--- a/test/websocket/stream/issue-5503.js
+++ b/test/websocket/stream/issue-5503.js
@@ -1,0 +1,171 @@
+'use strict'
+
+const { test } = require('node:test')
+const net = require('node:net')
+const { WebSocketServer } = require('ws')
+
+const { WebSocketStream } = require('../../..')
+
+function waitFor (predicate) {
+  return new Promise((resolve, reject) => {
+    const started = Date.now()
+
+    const check = () => {
+      if (predicate()) {
+        resolve()
+      } else if (Date.now() - started > 1000) {
+        reject(new Error('timed out waiting for condition'))
+      } else {
+        setTimeout(check, 10)
+      }
+    }
+
+    check()
+  })
+}
+
+// These tests assert that `closed` does NOT hang - if it regresses, the
+// point of the test is defeated by also hanging (up to the runner's own
+// timeout) instead of failing fast with a clear message.
+function withTimeout (promise, ms, message) {
+  let timer
+  const timeout = new Promise((resolve, reject) => {
+    timer = setTimeout(() => reject(new Error(message)), ms)
+  })
+
+  return Promise.race([promise, timeout]).finally(() => clearTimeout(timer))
+}
+
+test('WebSocketStream applies receive backpressure to the socket', async (t) => {
+  const server = new WebSocketServer({ port: 0 })
+
+  server.on('connection', (socket) => {
+    socket.send('one')
+  })
+
+  const { port } = server.address()
+  const originalPause = net.Socket.prototype.pause
+  const originalResume = net.Socket.prototype.resume
+  let pauseCount = 0
+  let resumeCount = 0
+
+  net.Socket.prototype.pause = function (...args) {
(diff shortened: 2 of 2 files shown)


mcollina (approved):
lgtm

KhafraDev (changes requested):
this is just ai slop I don't have the will to review right now

GiHoon1123:
Thanks for taking a look. I'd really like to address whatever's raising concern here — if there's something specific about the approach (the `pull` algorithm gating on `desiredSize`, the two-condition resume gate, or the `#resumeSocketForClose` fix) that doesn't look right, please let me know and I'll fix it.

For context in case it helps: the memory growth described in #5503 (75MiB → 337MiB with 200k unread messages) is reproducible locally, and the close()/reader.cancel() hang was something I found while testing the fix against a slow-consumer scenario.

KhafraDev:
see what I mean

mcollina:
@KhafraDev the lack of backpressure is real tho

KhafraDev:
I know, but reviewing ai slop where I can't even talk to a real human is not what I'd like to do

intech:
@KhafraDev, I think this is a counterproductive stance. Whether a piece of code was written by an AI or by a human is irrelevant if the code itself is objectively sound. From what I can see, this contribution follows the project's guidelines and implements exactly what was described.

If you have a principled objection to AI-assisted contributions, I believe that position should be discussed openly and agreed upon by the project's contributors as part of the project's consensus, rather than being applied on a case-by-case basis.

More broadly, in 2026, treating AI-generated code as inherently inferior is becoming increasingly difficult to justify. Technological progress will not stop, and in practice, AI, paired with a skilled engineer, consistently produces higher-quality software. This is especially valuable in open source, where maintainers and contributors have limited time and resources.

KhafraDev:
Thank you, random person not associated with the project, telling me, a contributor, what I should or should not accept. And being completely wrong on my beliefs or my stance or the project's stance in the process.

> I believe that position should be discussed openly and agreed upon by the project's contributors as part of the project's consensus, rather than being applied on a case-by-case basis.

If you spent 30 seconds looking, you would see that we actually have discussed it.

> This is especially valuable in open source, where maintainers and contributors have limited time and resources.

Therein lies the problem. I do not have unlimited time to review PRs. Why should I take time out of my day to review a piece of code that the committer did not? This guy's Fable tokens aren't somehow more valuable than mine, the only difference is he has no conceptual understanding of the spec or how to properly implement the feature.

Your stance is directly leading me to contribute less to undici as its sole maintainer fetch, EventSource, WebSocket & WebSocketStream. I am swamped with low quality PRs that I do not enjoy reviewing, and having to justify my stance to morons who think being anti-low-quality is the same as being anti-AI.

GiHoon1123:
Sorry if my PR looked AI-generated. The implementation itself is mine — I debugged the issue, read the spec, wrote and tested the fix myself. I also ran it by a few other devs before opening the PR.

What I did use AI for was the PR description, since English isn't my first language and I wanted it to read clearly. I see now that it backfired.

I'm completely fine with the PR staying closed if the implementation has problems or doesn't meet the project's standards. I just wanted to clarify that I do understand what I submitted. Happy to discuss the implementation if there are any questions.

PR #5541, opened 2026-07-10
feat(interceptors): add signal interceptor

### LINKED ISSUE #3276: move signal handling to interceptor

Arman-Luthra:
## This relates to...

Closes #3276

## Rationale

#3276 asks for `AbortSignal` handling to be available as an interceptor. Following the direction given in the issue thread, this adds a completely new `signal` interceptor shaped like `interceptors.responseError`: when the `signal` dispatch option is aborted, the request is aborted and the abort listener is cleaned up.

This is mainly useful when calling `dispatch()` directly with a custom handler, where today every caller has to wire up (and remember to remove) their own abort listener the way the API-layer handlers in `lib/api` do. The existing API-layer signal handling is intentionally left untouched.

## Changes

- Add `lib/interceptor/signal.js`, exported as `interceptors.signal`
- Supports both `AbortSignal` and `EventEmitter`-style signals (same contract as the `signal` option of the API methods, validated the same way)
- Aborts with `signal.reason` when available, `RequestAbortedError` otherwise
- A signal that is already aborted fails the dispatch before the request reaches the server, mirroring `RequestHandler`
- The abort listener is registered once per dispatched request and removed on `onResponseEnd`, `onResponseError` and `onRequestUpgrade`, so long-lived signals do not accumulate listeners
- The controller reference is refreshed on every `onRequestStart` so composing with `retry()` keeps aborting the active attempt
- Docs (`docs/docs/api/Interceptors.md`, `docs/docs/api/Dispatcher.md`), types (`types/interceptors.d.ts`) and type tests included

### Features

- New `interceptors.signal()` interceptor

### Bug Fixes

N/A

N/A

## Testing

`test/interceptors/signal.js` exercises the interceptor through `client.dispatch()` with a bare handler rather than `client.request()`, since the API-layer handlers already implement their own signal handling and would mask what the interceptor does. Covered: pre-aborted signal (exact reason propagated, request never reaches the server, only `onResponseError` is delivered), mid-response abort, `EventEmitter`-style signals, listener removal after completion and after an error (asserted via `getEventListeners`/`listenerCount`), pass-through when no signal is provided, `InvalidArgumentError` for non-signal values, and compatibility with `client.request()`.

`npm run lint`, `npm run test:interceptors` and the TypeScript checks pass locally (the pre-existing zstd decompress test is skipped locally as Node 22.14 lacks `zlib.createZstdCompress`).

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/interceptor/signal.js
+++ b/lib/interceptor/signal.js
@@ -1,0 +1,93 @@
+'use strict'
+
+const { InvalidArgumentError, RequestAbortedError } = require('../core/errors')
+const { addAbortListener } = require('../core/util')
+const DecoratorHandler = require('../handler/decorator-handler')
+
+class SignalHandler extends DecoratorHandler {
+  /** @type {AbortSignal|import('node:events').EventEmitter} */
+  #signal
+  /** @type {import('../core/request').RequestController|null} */
+  #controller = null
+  /** @type {(() => void)|null} */
+  #removeAbortListener = null
+  /** @type {boolean} */
+  #aborted = false
+
+  constructor (signal, handler) {
+    super(handler)
+    this.#signal = signal
+  }
+
+  #abort () {
+    this.#aborted = true
+    this.#removeListener()
+    this.#controller.abort(this.#signal.reason ?? new RequestAbortedError())
+  }
+
+  #removeListener () {
+    if (this.#removeAbortListener !== null) {
+      this.#removeAbortListener()
+      this.#removeAbortListener = null
+    }
+  }
+
+  onRequestStart (controller, context) {
+    // A fresh controller is passed for every dispatch of the same request
+    // (e.g. when composed with the retry interceptor), so keep the reference
+    // up to date while registering the abort listener only once.
+    this.#controller = controller
+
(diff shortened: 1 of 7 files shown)


lib/interceptor/signal.js
```diff
@@ -0,0 +1,93 @@
...
+
+    if (this.#removeAbortListener === null) {
+      this.#removeAbortListener = addAbortListener(this.#signal, () => this.#abort())
+    }
+
+    super.onRequestStart(controller, context)
+  }
+
+  onRequestUpgrade (controller, statusCode, headers, socket) {
+    this.#removeListener()
```
metcoder95: let's not remove the listener until the response has `ended` or `errored`

Arman-Luthra: makes sense

Arman-Luthra:
Done. The listener now stays attached through upgrade and is removed only in onResponseEnd / onResponseError.

metcoder95 (approved):

PR #5545, opened 2026-07-11
docs: clarify top-level request option url fields

vibhor-aggr:
## Summary

Refs #1011.

Clarifies the distinction between top-level Undici API options and lower-level dispatcher options. Top-level calls like `undici.request(url, options)` get `origin` and `path` from the separate `url` argument, so those fields should not be supplied in the second `options` argument even though the linked dispatcher option docs include them.

## Verification

- `git diff --check`
- `npm run lint`
- `npm run test:typescript`

### DIFF
--- a/README.md
+++ b/README.md
@@ -408,6 +408,11 @@ For more information about their behavior, please reference the body mixin from 
 
 This section documents our most commonly used API methods. Additional APIs are documented in their own files within the [docs](./docs/) folder and are accessible via the navigation list on the left side of the docs site.
 
+For the top-level APIs below, the `url` argument supplies the request origin and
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

vibhor-aggr:
The failing macOS Node 26 job looks unrelated to this docs-only change. The log shows `test/issue-5087.js` failing after the test ended because of `Error: read ECONNRESET` in `headersTimeout over h2 rejects with HeadersTimeoutError`. I tried to rerun failed jobs through the Actions API, but GitHub returned `403 Must have admin rights to Repository`. Could a maintainer rerun the failed job when convenient?

metcoder95 (approved):

PR #5546, opened 2026-07-11
fix(h2): prevent uncaughtException from onHttp2SocketError accessing undefined kClient

### LINKED ISSUE #5525: uncaughtException in unidici.request()
My app uses `unidici 8.7.0` to download files from CloudFlare R2:

```js
const fast_dispatcher = new Agent().compose(interceptors.redirect({maxRedirections: 5}));
function get_download_stream_fast(url){
  return request(url, {
    dispatcher: fast_dispatcher,
    signal: AbortSignal.timeout(30000)
  }).then(function({statusCode, body}){
    if(statusCode >= 200 && statusCode < 300){
      return body;
    }
    body.dump(); //free the connection, we don't use the body
    throw new Error(`HTTP ${statusCode} for: ${url}`);
  });
}
```

My app also monitors for `uncaughtException` errors:

```js
process.on('uncaughtException', function(err) {
  console.log( "[UNCAUGHT EXCEPTION] " + err.stack || err.message );
  process.exit(1);
});
```

I am regularly seeing these in my logs (which seems to correlate with downloads hitting the 30s timeout above). 

```txt
 [UNCAUGHT EXCEPTION] SocketError: other side closed
     at TLSSocket.onHttp2SocketEnd (/frontend/node_modules/undici/lib/dispatcher/client-h2.js:661:22)
     at TLSSocket.emit (node:events:531:35)
     at endReadableNT (node:internal/streams/readable:1698:12)
     at process.processTicksAndRejections (node:internal/process/task_queues:89:21)
```

Perhaps the CDN is not always 100% reliable. Is there a way to catch these exceptions, or should I just ignore them?

mcollina:
### Root Cause

In `client-h2.js`, `onHttp2SocketError` accesses `this[kClient]` where `this` is the socket. However, `kClient` is only stored on the session (`session[kClient] = client`), never on the socket. So `this[kClient]` is `undefined`, and calling `this[kClient][kOnError](err)` throws a `TypeError`. This thrown error inside the 'error' event handler becomes an uncaughtException.

### Changes

1. **`onHttp2SocketError`** — Route through `this[kHTTP2Session]?.[kClient]` instead of `this[kClient]`
2. **`util.destroy`** — Wrap `stream.destroy(err)` in try-catch (http2-managed sockets throw on direct `destroy()`)
3. **`onHttp2SessionEnd`** — Wrap `util.destroy(this[kSocket], err)` in try-catch (same reason)

### Testing

All 1421 existing tests pass (1418 pass, 3 skipped).

Fixes #5525

### DIFF
--- a/test/http2-dispatcher.js
+++ b/test/http2-dispatcher.js
@@ -1093,7 +1093,7 @@ test('Should not send http2 PING frames after connection is closed', async t => 
 })
 
 test('Should not emit uncaughtException when socket closes after abort', async t => {
-  t = tspl(t, { plan: 2 })
+  const assert = tspl(t, { plan: 2 })
 
   const server = createSecureServer(await pem.generate({ opts: { keySize: 2048 } }))
 
@@ -1102,7 +1102,7 @@ test('Should not emit uncaughtException when socket closes after abort', async t
     stream.on('error', () => {})
   })
 
-  after(() => {
+  t.after(() => {
     server.close()
   })
   await once(server.listen(0), 'listening')
@@ -1117,7 +1117,7 @@ test('Should not emit uncaughtException when socket closes after abort', async t
     uncaughtErr = err
   }
   process.on('uncaughtException', uncaughtHandler)
-  after(() => {
+  t.after(() => {
     process.removeListener('uncaughtException', uncaughtHandler)
     if (originalHandler) {
       process.on('uncaughtException', originalHandler)
@@ -1128,7 +1128,7 @@ test('Should not emit uncaughtException when socket closes after abort', async t
     connect: { rejectUnauthorized: false },
     allowH2: true
   })
-  after(() => client.close())
+  t.after(() => client.close())
 
   const controller = new AbortController()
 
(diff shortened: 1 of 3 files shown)


mcollina:
The underlying `.destroy()` throw is fixed in Node core: https://github.com/nodejs/node/pull/64427

test/http2-dispatcher.js
```diff
@@ -1091,3 +1091,73 @@ test('Should not send http2 PING frames after connection is closed', async t =>
...
+  const server = createSecureServer(await pem.generate({ opts: { keySize: 2048 } }))
+
+  // Never respond, keeping the stream in-flight.
+  server.on('stream', (stream) => {
+    stream.on('error', () => {})
+  })
+
+  after(() => {
+    server.close()
+  })
```
trivikr: Use `t.after`
```suggestion
  t.after(() => {
    server.close()
  })
```

Ditto at other places.

metcoder95 (approved):

PR #5548, opened 2026-07-11
docs: add custom interceptor example

### LINKED ISSUE #3218: Improve docs on Interceptors
## This would solve...

Currently docs have limited information for interceptors and it's not easy to discover them, Dispatchers.compose() is probably the best documentation in the docs site and it lacks an example or information on the handler callbacks. There is what looks like an outdated [DispatchInterceptor](https://github.com/nodejs/undici/blob/main/docs/docs/api/DispatchInterceptor.md) markdown file in the docs, which is no longer part of the docs sidebar or links. I think this is left over from the interceptors param on Client that's deprecated?

Add more detailed documentation on interceptors, ideally it should be easier to discover and include handlers examples and or documentation on the handlers.

## Additional context

If this issue is still open for a bit I can try and open a PR to help expand the docs on interceptors, but wanted to open the issue to make sure it's tracked. 🙂

vibhor-aggr:
Fixes #3218.

This adds a custom interceptor example to the Interceptors API documentation. The new section shows the current `Dispatcher.compose()` pattern and demonstrates how to wrap and forward handler callbacks while measuring response timing.

It also links readers from the `Dispatcher.compose()` documentation to the new custom example and points from the example back to the full `dispatcher.dispatch(options, handler)` callback reference.

Validation:

- `git diff --check`
- `npm run lint`
- `npm run test:typescript`

### DIFF
--- a/docs/docs/api/Dispatcher.md
+++ b/docs/docs/api/Dispatcher.md
@@ -733,6 +733,8 @@ await client.request({ path: '/', method: 'GET' })
 ```
 
 For the full list of built-in interceptors provided by undici, see [Interceptors](Interceptors.md).
+For an example of a custom interceptor that wraps handler callbacks, see
+[Custom interceptors](Interceptors.md#custom-interceptors).
 
 ### Event: `'connect'`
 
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

Issue #5558, opened 2026-07-14
HTTP/2: setImmediate-deferred util.destroy(stream) in the abort path leaks aborted streams until OOM when the event loop stalls


staylor:
## Version

- `undici` 8.7.0
- Node.js `>=22.19` (observed on v24.x)

## Summary

In `lib/dispatcher/client-h2.js`, the `writeH2` abort handler closes the stream to send `RST_STREAM` and then defers the actual `util.destroy(stream)` behind a `setImmediate`. On a healthy event loop this is fine. But if the event loop is stalled (a long synchronous task), the `setImmediate` callback does not run, so every aborted `Http2Stream` — and the whole request graph each one pins (request/response objects, header lists, buffered body chunks) — stays alive until the loop drains. Under sustained request churn with aborts (timeouts, client cancellations) during a stall, these accumulate without bound and the process OOMs.

## Offending code

`lib/dispatcher/client-h2.js`, `writeH2`, abort branch:

```js
if (state.stream != null) {
  clearRequestStream(request)

  // On Abort, we close the stream to send RST_STREAM frame.
  const stream = state.stream
  stream.close()

  // close() alone leaves cleanup waiting on the 'close' event; on a busy,
  // long-lived multiplexed session that event can fail to fire, leaving the
  // native Http2Stream (and the whole request graph it pins) alive for the
  // session's life. Destroy the stream to release the handle
  // deterministically, but defer it by a setImmediate so the RST_STREAM
  // frame queued by close() gets a chance to be written first.
  setImmediate(() => {
    if (!stream.destroyed) {
      util.destroy(stream)
    }
  })

  // ...
}
```

The comment correctly identifies that `close()` alone can leak the stream (cleanup waits on a `'close'` event that may never fire on a busy multiplexed session) and that a deterministic `util.destroy` is needed. The problem is the *deferral*: `setImmediate` moves the deterministic teardown onto the macrotask queue, which reintroduces exactly the "cleanup that might not run" hazard the destroy was added to avoid — this time gated on event-loop liveness rather than the `'close'` event.

## Why the deferral defeats its own purpose

- `setImmediate` callbacks run in the *check* phase, i.e. only once the current synchronous work yields and the loop advances.
- A stalled loop (large JSON encode/parse, a synchronous hot path, GC pauses, or an application that periodically blocks) delays *all* pending `setImmediate` callbacks for the duration of the stall.
[...]

PR #5559, opened 2026-07-14
fix(h2): destroy aborted stream synchronously instead of via setImmediate

staylor:
Follow-up to #5462.

#5462 fixed a leak where an aborted HTTP/2 request left teardown to the stream's async `'close'` event, which can fail to fire on a busy, long-lived multiplexed session — pinning the native `Http2Stream` and the request graph it references. It did so by destroying the stream after `close()`, but **deferred the destroy behind a `setImmediate`** so the `RST_STREAM` frame queued by `close()` is written first.

That deferral reintroduces the same leak under a different trigger. `setImmediate` callbacks only run once the loop reaches the check phase, so while the event loop is stalled by a long synchronous task the deferred destroy never runs. Every stream aborted during the stall stays alive (plus the request graph each one pins), and abort churn — timeouts, cancellations — tends to spike precisely when the loop is busy. The result is an unbounded, positive-feedback leak until the process OOMs.

This was not hypothetical for us: a high-throughput fetch service (~100K req/s across the fleet) that periodically stalls its loop under bursty work saw memory roughly triple and the fleet enter a sustained `OOMKilled` crashloop immediately after taking stock 8.7.0 (which carries the `setImmediate` defer) in place of a local patch that destroyed synchronously. Details in #5558.

## Change

- Destroy the stream **synchronously** after `close()` in the `writeH2` abort path. `close()` has already queued the `RST_STREAM` frame on the native session, so it is still sent; teardown no longer depends on the event loop advancing.
- Tighten the `'Should destroy the h2 stream on abort'` regression test to assert the stream is destroyed **synchronously** on abort (drops the `waitFor` poll), so a future re-deferral fails in CI instead of in production.

## Testing

`test/http2-dispatcher.js` (all 18) plus `http2-abort`, `http2-goaway`, `http2-destroy-after-failed-stream`, `http2-response-after-completion`, `http2-timeout`, and `http2-stream` pass locally. `eslint` clean.

If there is a concrete `RST_STREAM`-ordering scenario the `setImmediate` was guarding against that isn't covered by the current tests, I'm happy to add a case for it — but in practice `close(code)` submits the frame to the native handle synchronously, so a subsequent `destroy()` still delivers it.

### DIFF
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -933,14 +933,14 @@ function writeH2 (client, request) {
       // close() alone leaves cleanup waiting on the 'close' event; on a busy,
       // long-lived multiplexed session that event can fail to fire, leaving the
       // native Http2Stream (and the whole request graph it pins) alive for the
-      // session's life. Destroy the stream to release the handle
-      // deterministically, but defer it by a setImmediate so the RST_STREAM
-      // frame queued by close() gets a chance to be written first.
-      setImmediate(() => {
-        if (!stream.destroyed) {
-          util.destroy(stream)
-        }
-      })
+      // session's life. Destroy the stream synchronously to release the handle
+      // deterministically. Deferring the destroy (e.g. via setImmediate) leaks
+      // the same way when the event loop is stalled and the callback never runs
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

PR #5560, opened 2026-07-14
fix(h2): complete request stream on 'end' instead of waiting for 'close'

staylor:
## Problem

On a busy, long-lived multiplexed HTTP/2 session, the native `'close'` event can fail to fire for a stream (the same condition #5559 addresses on the abort path). In `client-h2.js`, normal request completion still defers **all** cleanup to `onRequestStreamClose`, which is gated on `'close'`:

- `request.onResponseEnd(...)` + `finalizeRequest(...)`
- `releaseRequestStream(...)` (detaches the request, removes stream listeners)
- `closeStreamSession(...)` (decrements the session's open-stream count)
- `stream[kRequestStreamState] = null`

`onEnd` only sets `state.pendingEnd = true` and waits. So a stream that completes but never emits `'close'` keeps its `kRequestStreamState` — and through it the request, response body buffers, and stream listeners — alive for the whole session lifetime. Under sustained request volume on a persistent h2 connection this grows without bound until the process OOMs.

We hit this in production on a crawler doing continuous h2 fetches over long-lived sessions: heap-retention profiles showed the request-stream state graph and off-heap response buffers accumulating, with `end-of-stream` / `pipeline` / `async_hooks` teardown machinery retained 15–28× over baseline. #5559 (thank you) fixed the abort path but the completion path kept leaking.

## Fix

Extract the terminal cleanup into `completeRequestStream(stream)` and run it from **both** `onEnd` and `onRequestStreamClose`. A `state == null` guard makes it idempotent: whichever of `'end'` / `'close'` fires first performs the cleanup, and the other becomes a no-op (this also prevents a double `closeStreamSession` decrement).

Completing on `'end'` releases the request graph deterministically without waiting for `'close'`. Crucially it **completes the response properly** (`onResponseEnd`, which ends the consumer body) rather than force-destroying the stream — so an in-flight body is never truncated with `ERR_STREAM_PREMATURE_CLOSE`. Trailers already stored by `onTrailers` (which precedes `'end'` in the normal flow) are still delivered.

## Reproduction / test

Added a test (`test/http2-late-data.js`) that drives a normal completion through `'response'` → `'trailers'` → `'data'` → `'end'` and **never emits `'close'`**, then asserts the response completes and the stream listeners are released.

- On `main`: fails — `onResponseEnd` is never called (`onCompleteCalls: 0`) and listeners stay attached.
- With this change: passes.

All 24 `test/http2-*.js` files pass (including `Dispatcher#Connect`, trailers, goaway, timeout, and abort suites); `eslint` is clean.

Refs #5558.

### DIFF
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -678,18 +678,22 @@ function onUpgradeStreamClose () {
   closeStreamSession(this)
 }
 
-function onRequestStreamClose () {
+// Idempotent terminal cleanup, called from both 'end' and 'close': the
+// null-state guard no-ops the later call.
+function completeRequestStream () {
   const state = this[kRequestStreamState]
 
-  if (state) {
-    // Release the stream first so request references are cleared,
-    // then complete the response with trailers if available.
-    releaseRequestStream(this)
+  if (state == null) {
+    return
+  }
 
-    if (state.pendingEnd && !state.request.aborted && !state.request.completed) {
-      state.request.onResponseEnd(state.trailers || {})
-      finalizeRequest(state)
-    }
+  // Release the stream first so request references are cleared,
+  // then complete the response with trailers if available.
+  releaseRequestStream(this)
+
+  if (state.pendingEnd && !state.request.aborted && !state.request.completed) {
+    state.request.onResponseEnd(state.trailers || {})
+    finalizeRequest(state)
   }
 
   closeStreamSession(this)
@@ -1127,7 +1131,7 @@ function writeH2 (client, request) {
   }
 
   stream[kHTTP2Session] = session
-  stream.on('close', onRequestStreamClose)
+  stream.on('close', completeRequestStream)
 
   bindRequestToStream(request, stream, releaseRequestStream)
   if (expectContinue) {
@@ -1270,13 +1274,16 @@ function onEnd () {
 
   stream.off('end', onEnd)
 
-  // If we received a response, this is a normal completion.
-  // Defer actual completion to onRequestStreamClose so that
-  // onTrailers (which may fire after 'end' on Windows) can
-  // store trailers first.
+  // onTrailers (which may fire after 'end' on Windows) has already stored
+  // trailers on the state by now, so completing here still delivers them.
   if (state.responseReceived) {
     if (!request.aborted && !request.completed) {
       state.pendingEnd = true
+
+      // Complete on 'end': a blocked event loop can keep the stream's 'close'
+      // from firing, stranding its buffers until OOM. Idempotent, so a later
+      // 'close' no-ops.
+      completeRequestStream.call(stream)
     }
   } else {
     // Stream ended without receiving a response - this is an error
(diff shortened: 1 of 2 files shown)


lib/dispatcher/client-h2.js
```diff
@@ -678,22 +678,36 @@ function onUpgradeStreamClose () {
...
-function onRequestStreamClose () {
-  const state = this[kRequestStreamState]
+// Terminal cleanup for a request stream, run exactly once: complete the
+// response (with any trailers onTrailers has stored), release request
+// references and listeners, and decrement the session's open-stream count.
+// Safe from both 'end' (onEnd) and 'close' (onRequestStreamClose) -- whichever
+// fires first wins, and the null-state guard makes the later call a no-op. This
+// lets a completed stream be released without waiting for a 'close' event that
+// can fail to fire on a busy, long-lived multiplexed session (#5558), which
+// otherwise pins the request graph and buffers until OOM.
```
mcollina: shorten this comment, or remove

lib/dispatcher/client-h2.js
```diff
@@ -1277,6 +1291,13 @@ function onEnd () {
   if (state.responseReceived) {
     if (!request.aborted && !request.completed) {
       state.pendingEnd = true
+
+      // Complete on 'end' rather than deferring to 'close'. On a busy,
+      // long-lived multiplexed session the native 'close' event can fail to
+      // fire, stranding the completed stream's request graph and buffers until
+      // OOM (#5558). Completing here releases them deterministically; if 'close'
+      // arrives later, completeRequestStream is a no-op.
```
mcollina: This comment is too long. It doesn't explain why this is needed but reference an issue. Update.

This should mention "blocked event loop".

lib/dispatcher/client-h2.js
staylor: Cut to two lines — just the idempotency contract; the mechanics are clear from the code.

lib/dispatcher/client-h2.js
staylor: Rewritten to lead with the why — a blocked event loop can keep 'close' from firing, stranding the stream's buffers — and dropped the bare issue reference.

lib/dispatcher/client-h2.js
```diff
@@ -678,22 +678,30 @@ function onUpgradeStreamClose () {
...
 
-  closeStreamSession(this)
-  this[kRequestStreamState] = null
+  closeStreamSession(stream)
+  stream[kRequestStreamState] = null
+}
+
+function onRequestStreamClose () {
+  completeRequestStream(this)
 }
```
mcollina, line 701: do we really need 2 functions?

staylor: Good call — the second function was just a `this`→arg shim. Collapsed to a single `this`-based `completeRequestStream` registered directly on `'close'` and invoked via `.call(stream)` from `onEnd`, matching the `onUpgradeStreamClose` convention. Also cleaned up two stale comments that still referenced the old name. All 24 `http2-*` suites still pass.

metcoder95 (approved):

mcollina (approved):
lgtm

staylor:
I deployed this using `pnpm patch`, and `8.7.0` has another memory leak somewhere - will report back when I find out where

PR #5561, opened 2026-07-15
fix(fetch): parse multipart body up to the full boundary delimiter

arshsmith1:
## This relates to...

Multipart parsing in `response.formData()` / `request.formData()`.

## Rationale

The body loop in `multipartFormDataParser` looked for a bare boundary token (`boundary.subarray(2)`) rather than the `\r\n--boundary` delimiter line, then blindly subtracted 4 from the match index. A part whose body legitimately contains the boundary token gets truncated at the wrong offset, and in most cases the whole body is rejected with `expected CRLF` even though it is well framed. This block carried a `// TODO: the steps here are completely wrong` note. The fix matches the full delimiter so boundary-like bytes inside a body are no longer mistaken for a delimiter, which also removes an ambiguity where such bytes could be read as an extra part.

## Changes

Search for `\r\n--boundary` when terminating a part body and slice up to that index, instead of finding the bare boundary token and offsetting by 4.

### Features

N/A

### Bug Fixes

- multipart body loop now terminates a part at the full boundary delimiter, so bodies containing the boundary token parse correctly instead of being truncated or rejected.

N/A

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/web/fetch/formdata-parser.js
+++ b/lib/web/fetch/formdata-parser.js
@@ -76,10 +76,6 @@ function multipartFormDataParser (input, mimeType) {
 
   const boundary = Buffer.from(`--${boundaryString}`, 'utf8')
 
-  // The delimiter that terminates a part's body is the CRLF that precedes the
-  // next boundary delimiter line, i.e. "\r\n--boundary". Matching the full
-  // delimiter (not a bare boundary token) is required so boundary-like bytes
-  // inside a body are not mistaken for a delimiter.
   const crlfDashBoundary = Buffer.from(`\r\n--${boundaryString}`, 'utf8')
 
   // 3. Let entry list be an empty entry list.
--- a/lib/web/fetch/formdata-parser.js
+++ b/lib/web/fetch/formdata-parser.js
@@ -76,6 +76,12 @@ function multipartFormDataParser (input, mimeType) {
 
   const boundary = Buffer.from(`--${boundaryString}`, 'utf8')
 
+  // The delimiter that terminates a part's body is the CRLF that precedes the
+  // next boundary delimiter line, i.e. "\r\n--boundary". Matching the full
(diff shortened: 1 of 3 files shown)


lib/web/fetch/formdata-parser.js
```diff
@@ -76,6 +76,12 @@ function multipartFormDataParser (input, mimeType) {
 
   const boundary = Buffer.from(`--${boundaryString}`, 'utf8')
 
+  // The delimiter that terminates a part's body is the CRLF that precedes the
+  // next boundary delimiter line, i.e. "\r\n--boundary". Matching the full
+  // delimiter (not a bare boundary token) is required so boundary-like bytes
+  // inside a body are not mistaken for a delimiter.
```
KhafraDev:
```suggestion
```

test/fetch/client-fetch.js
```diff
@@ -268,6 +268,38 @@ test('multipart fromdata non-ascii filed names', async (t) => {
   t.assert.strictEqual(form.get('fiŝo'), 'value1')
 })
 
+test('multipart formdata body containing the boundary token', async (t) => {
```
KhafraDev: please move this to test/busboy

KhafraDev (changes requested):

lib/web/fetch/formdata-parser.js
arshsmith1: Dropped the comment.

test/fetch/client-fetch.js
arshsmith1: Moved it to test/busboy/formdata-boundary-token.js.

arshsmith1:
any chance of a re-review?

PR #5562, opened 2026-07-15
fix(websocket): apply WebSocketStream receive backpressure

### LINKED ISSUE #5503: WebSocketStream: receive backpressure is not implemented — a slow consumer buffers every received message in memory

mcollina:
## This relates to...

Fixes #5503

## Rationale

`WebSocketStream` continued consuming transport data after its readable queue reached its high-water mark. Slow consumers therefore accumulated an unbounded number of messages instead of propagating backpressure to the peer.

### Features

N/A

### Bug Fixes

- Consume upgraded WebSocket sockets with `.read()` from the `'readable'` event instead of flowing `'data'` events.
- Read for `WebSocketStream` only while the readable controller has demand and the frame parser is ready.
- Continue reading during the closing handshake so Close frames can be processed.
- Mark the existing WebSocketStream receive-backpressure WPT as passing.

N/A

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### Validation

- `npm run lint`
- `npm run test:websocket`
- `node test/web-platform-tests/wpt-runner.mjs run '/websockets/stream/tentative/backpressure-receive.any.html?wss' '/websockets/stream/tentative/backpressure-send.any.html?wss'`

### DIFF
--- a/lib/web/websocket/stream/websocketstream.js
+++ b/lib/web/websocket/stream/websocketstream.js
@@ -47,7 +47,7 @@ class WebSocketStream {
     onConnectionEstablished: (response, extensions) => this.#onConnectionEstablished(response, extensions),
     onMessage: (opcode, data) => this.#onMessage(opcode, data),
     onParserError: (err) => failWebsocketConnection(this.#handler, null, err.message),
-    onParserDrain: () => this.#handler.socket.resume(),
+    onParserDrain: () => this.#resumeSocketIfNeeded(),
     onSocketData: (chunk) => {
       if (!this.#parser.write(chunk)) {
         this.#handler.socket.pause()
@@ -117,7 +117,7 @@ class WebSocketStream {
     this.#closedPromise = Promise.withResolvers()
 
     // 7. Apply backpressure to the WebSocket.
-    // TODO
+    // This is done when the socket becomes available.
 
     // 8.  If options [" signal "] exists ,
     if (options.signal != null) {
@@ -195,6 +195,22 @@ class WebSocketStream {
 
     // 3. Close the WebSocket with this , code , and reason .
     closeWebSocketConnection(this.#handler, code, reason, true)
+    this.#resumeSocketIfNeeded()
+  }
+
+  #resumeSocketIfNeeded () {
+    const socket = this.#handler.socket
+
+    if (socket == null) {
+      return
+    }
+
+    if (
+      this.#handler.readyState === states.CLOSING ||
+      (this.#readableStreamController?.desiredSize > 0 && !this.#parser.writableNeedDrain)
+    ) {
+      socket.resume()
+    }
   }
 
   #write (chunk) {
@@ -257,6 +273,7 @@ class WebSocketStream {
   /** @type {import('../websocket').Handler['onConnectionEstablished']} */
   #onConnectionEstablished (response, parsedExtensions) {
     this.#handler.socket = response.socket
+    this.#handler.socket.pause()
 
     // Get options from dispatcher options
     const maxFragments = this.#handler.controller.dispatcher?.webSocketOptions?.maxFragments
@@ -291,6 +308,7 @@ class WebSocketStream {
       start: (controller) => {
         this.#readableStreamController = controller
       },
+      pull: () => this.#resumeSocketIfNeeded(),
       cancel: (reason) => this.#cancel(reason)
     })
 
@@ -301,7 +319,10 @@ class WebSocketStream {
     // 13. Set up writable with writeAlgorithm , closeAlgorithm , and abortAlgorithm .
     const writable = new WritableStream({
       write: (chunk) => this.#write(chunk),
-      close: () => closeWebSocketConnection(this.#handler, null, null),
+      close: () => {
+        closeWebSocketConnection(this.#handler, null, null)
+        this.#resumeSocketIfNeeded()
+      },
       abort: (reason) => this.#closeUsingReason(reason)
     })
 
@@ -350,6 +371,9 @@ class WebSocketStream {
     this.#readableStreamController.enqueue(chunk)
 
     // 4. Apply backpressure to the WebSocket.
+    if (this.#readableStreamController.desiredSize <= 0) {
+      this.#handler.socket.pause()
+    }
   }
 
   /** @type {import('../websocket').Handler['onSocketClose']} */
@@ -441,6 +465,7 @@ class WebSocketStream {
     // 4. Close the WebSocket with stream , code , and reasonString . If this throws an exception,
     //    discard code and reasonString and close the WebSocket with stream .
     closeWebSocketConnection(this.#handler, code, reasonString)
+    this.#resumeSocketIfNeeded()
   }
 
   //  To cancel a WebSocketStream stream given reason , close using reason giving stream and reason .
--- a/lib/web/websocket/stream/websocketstream.js
+++ b/lib/web/websocket/stream/websocketstream.js
@@ -47,12 +47,8 @@ class WebSocketStream {
     onConnectionEstablished: (response, extensions) => this.#onConnectionEstablished(response, extensions),
     onMessage: (opcode, data) => this.#onMessage(opcode, data),
     onParserError: (err) => failWebsocketConnection(this.#handler, null, err.message),
-    onParserDrain: () => this.#resumeSocketIfNeeded(),
-    onSocketData: (chunk) => {
-      if (!this.#parser.write(chunk)) {
-        this.#handler.socket.pause()
-      }
-    },
+    onParserDrain: () => this.#readFromSocket(),
+    onSocketReadable: () => this.#readFromSocket(),
     onSocketError: (err) => {
       this.#handler.readyState = states.CLOSING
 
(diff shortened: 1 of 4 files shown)


ronag (approved):

lib/web/websocket/stream/websocketstream.js
```diff
@@ -117,7 +117,7 @@ class WebSocketStream {
     this.#closedPromise = Promise.withResolvers()
 
     // 7. Apply backpressure to the WebSocket.
-    // TODO
+    // This is done when the socket becomes available.
```
KhafraDev: This is wrong and could possibly lead to dropped messages.

- user does `new WebSocketStream(...)`
- await ws.opened
- sets up the ReadableStream idk, 10 seconds later

any messages between the connection opening and reading messages will get "dropped" (stream will pull them even if the user isn't)

lib/web/websocket/stream/websocketstream.js
```diff
@@ -291,6 +308,7 @@ class WebSocketStream {
       start: (controller) => {
         this.#readableStreamController = controller
       },
+      pull: () => this.#resumeSocketIfNeeded(),
```
KhafraDev: "    // 5. Let pullAlgorithm be an action that pulls bytes from stream ."

#resumeSocketIfNeeded does not do that

KhafraDev: doesn't it make sense to handle backpressure in the WHATWG stream instead of the node socket? This implementation doesn't look correct to me

lib/web/websocket/stream/websocketstream.js
```diff
@@ -350,6 +371,9 @@ class WebSocketStream {
     this.#readableStreamController.enqueue(chunk)
 
     // 4. Apply backpressure to the WebSocket.
+    if (this.#readableStreamController.desiredSize <= 0) {
+      this.#handler.socket.pause()
```
KhafraDev: ditto wrt node socket vs whatwg stream.

KhafraDev (changes requested):

lib/web/websocket/stream/websocketstream.js
mcollina: What do you mean? When the WebSocket is pulled, the stream restarts.

mcollina: If I understand you correctly, you want to switch this from using `on('data')` to using `.read()` and `.on('readable')`, essentially making this a pull rather than a push?

I didn't do that because you usually prefer to avoid large refactorings. I'll get it done.

PR #5564, opened 2026-07-16
fix(decompress): skip HEAD responses

Ram-blip:
## This relates to...

`interceptors.decompress()` currently attempts to initialize a decoder for `HEAD` responses when the server includes a supported `Content-Encoding`.

Because a valid `HEAD` response contains no response body, the decoder receives no compressed data and `response.body.text()` rejects with:

```text
Z_BUF_ERROR: unexpected end of file
```

The interceptor also removes `Content-Encoding` and `Content-Length`, even though those headers describe the corresponding `GET` representation and are valid metadata on a `HEAD` response.

## Rationale

HTTP `HEAD` responses must not include response content, but their representation headers describe the response that would have been returned for a corresponding `GET`.

Undici’s Fetch implementation already skips response decoding for `HEAD`. The decompression interceptor should do the same rather than creating an empty decoder and removing representation metadata.

### Features

N/A

### Bug Fixes

- Bypass the decompression interceptor for `HEAD` requests.
- Preserve `Content-Encoding` and `Content-Length` on encoded `HEAD` responses.
- Ensure the response body resolves to an empty string instead of rejecting with `Z_BUF_ERROR`.
- Add a public API regression test using `Client.compose(interceptors.decompress())`.

None.

## Testing

```bash
node --test test/interceptors/decompress.js
npm run test:interceptors
npm run lint
```

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/interceptor/decompress.js
+++ b/lib/interceptor/decompress.js
@@ -276,6 +276,10 @@ function createDecompressInterceptor (options = {}) {
 
   return (dispatch) => {
     return (opts, handler) => {
+      if (opts.method === 'HEAD') {
+        return dispatch(opts, handler)
+      }
+
       const decompressHandler = new DecompressHandler(handler, options)
       return dispatch(opts, decompressHandler)
     }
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

PR #5567, opened 2026-07-17
fix(sri): split integrity metadata on ascii whitespace

arshsmith1:
## This relates to...

Subresource integrity checking in `fetch()`.

## Rationale

`parseMetadata` splits the integrity string on U+0020 only, but [parse metadata](https://w3c.github.io/webappsec-subresource-integrity/#parse-metadata) step 2 splits on ASCII whitespace (tab, LF, FF, CR, space). An integrity value carrying any of the other four separators, such as `"\nsha256-<hash>"` or hashes separated by a tab, yields no parsed items, and `bytesMatch` returns `true` for an empty metadata list per step 2 of [does response match metadataList](https://w3c.github.io/webappsec-subresource-integrity/#does-response-match-metadatalist). The check then passes for any response bytes rather than enforcing the hash, so integrity fails open exactly where the spec says it should fail closed. Multi-line integrity values are easy to end up with when hashes are pasted from an HTML `integrity` attribute or built from a template.

Splitting on the whitespace set the spec names keeps the fix in the parser, where the metadata list is produced, so `bytesMatch` continues to treat "empty list" as "nothing to check" as specified.

## Changes

Split the metadata string on ASCII whitespace instead of U+0020 in `parseMetadata`.

### Features

N/A

### Bug Fixes

- integrity metadata separated by tab, LF, FF or CR is parsed and enforced instead of being silently dropped, which previously let the check pass for any bytes.

N/A

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/web/subresource-integrity/subresource-integrity.js
+++ b/lib/web/subresource-integrity/subresource-integrity.js
@@ -187,6 +187,8 @@ function getStrongestMetadata (metadataList) {
   return result
 }
 
+const ASCII_WHITESPACE_SPLIT_REGEX = /[\u0009\u000A\u000C\u000D\u0020]+/ // eslint-disable-line no-control-regex
+
 /**
  * @param {string} metadata
  * @returns {MetadataList}
@@ -199,7 +201,7 @@ function parseMetadata (metadata) {
   const result = []
 
   // 2. For each item returned by splitting metadata on spaces:
-  for (const item of metadata.split(' ')) {
+  for (const item of metadata.split(ASCII_WHITESPACE_SPLIT_REGEX)) {
     // 1. Let expression-and-options be the result of splitting item on U+003F (?).
     const expressionAndOptions = item.split('?', 1)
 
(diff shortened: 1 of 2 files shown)


KhafraDev (changes requested):
https://infra.spec.whatwg.org/#strictly-split

> This algorithm is a "strict" split, as opposed to the commonly-used variants [for ASCII whitespace](https://infra.spec.whatwg.org/#split-on-ascii-whitespace) and [for commas](https://infra.spec.whatwg.org/#split-on-commas) below, which are both more lenient in various ways involving interspersed [ASCII whitespace](https://infra.spec.whatwg.org/#ascii-whitespace).

arshsmith1:
You're right, and I had the spec backwards. Step 2 links to strictly split on U+0020, not the lenient ASCII whitespace variant, so `metadata.split(' ')` is the conformant thing here and an integrity value carrying a tab or newline is meant to end up as one invalid token and be dropped. That leaves an empty metadata list, which step 2 of does response match metadataList says returns true, so the permissive outcome I was pointing at is specified rather than a gap in undici.

I've reverted the branch back to what you reviewed and I'm leaving this closed. Sorry for the noise.

PR #5577, opened 2026-07-21
fix: preserve DNS origin hostname on sockets

### LINKED ISSUE #5573: Original hostname is lost for HTTP connections when using the DNS interceptor
## Bug Description

We collect DNS lookup, TCP connection, and TLS handshake duration metrics using Node.js diagnostics channels ([source](https://github.com/tramvaijs/tramvai/blob/main/packages/modules/metrics/src/request/createRequestWithMetrics.ts#L164)).

However, when the DNS interceptor is used, Undici resolves the original hostname before creating the socket and replaces the connection origin with the selected IP address.

For HTTPS connections, the original hostname is still available through servername, because it is required for TLS SNI.

For plain HTTP connections, however, the original hostname is no longer available at the socket level. Instrumentation based on Node.js's net.client.socket diagnostics channel can only observe the resolved IP address.

This makes it impossible to correctly attribute TCP connection metrics to the original service hostname.

## Reproduction

```javascript
const { test } = require("node:test");
const { createServer } = require("node:http");
const { once } = require("node:events");
const diagnosticsChannel = require("node:diagnostics_channel");
const { Agent, fetch, interceptors } = require("undici");

const netClientSocketChannel = diagnosticsChannel.channel("net.client.socket");

const ORIGINAL_HOSTNAME = "localhost";
const PORT = 3000;
const RESOLVED_ADDRESS = "127.0.0.1";
const url = `http://${ORIGINAL_HOSTNAME}:${PORT}`;

// Create dispatcher with dns cache
const dispatcherWithDnsCache = new Agent().compose(
  interceptors.dns({
    maxTTL: 60_000,
    lookup(hostname, options, callback) {
      callback(null, [
        {
          address: RESOLVED_ADDRESS,
          family: 4,
          ttl: 60_000,
        },
      ]);
    },
  }),
);

let lookupEmitted = false;
let host;

// Subscribe on diagnostic channel events
netClientSocketChannel.subscribe(({ socket }) => {
  socket.once("lookup", (error, address, family, hostname) => {
    lookupEmitted = true;
  });

  socket.once("connect", () => {
    host = socket._host;
  });
});

test("Host field should be available", { timeout: 60000 }, async (t) => {
  // 1. Start a local HTTP server
  const server = createServer({}, async (req, res) => {
    res.statusCode = 200;
    res.setHeader("content-length", 100);
    res.end("hello".repeat(20));
  });

  t.after(() => {
    server.closeAllConnections?.();
    server.close();
  });

  server.listen(PORT);

  // 2. Perform the request without dns cache
  const response = await fetch(url);

  // host should be 'localhost'
  t.assert.strictEqual(host, ORIGINAL_HOSTNAME);

  // 3. Perform the request with dns cache
  const cachedDnsResponse = await fetch(url, { dispatcher: dispatcherWithDnsCache });

  // host should be 'localhost'
  t.assert.strictEqual(host, ORIGINAL_HOSTNAME);
});

```

## Expected Behavior

Connection instrumentation should have access to both:

- the resolved address used for the TCP connection;
- the original hostname supplied in the request URL.

For example:

{
  remoteAddress: '127.0.0.1',
  host: 'localhost'
}

This would also make HTTP and HTTPS connection instrumentation consistent.

## Actual Behavior

The socket created for the HTTP request contains only the resolved IP address:

{
  remoteAddress: '127.0.0.1',
  host: null
}

The original hostname is not available from the socket or from the net.client.socket diagnostics event.

As a result, diagnostics-based instrumentation cannot determine which logical hostname or service initiated that connection.

cyphercodes:
## This relates to...

Fixes #5573

## Rationale

The DNS interceptor connects to a selected IP address, which causes Node's `net.client.socket` diagnostics channel to expose the IP as `socket._host`. Consumers can no longer correlate TCP connection timing with the original request hostname.

## Changes

- pass the request hostname through HTTP connection setup
- connect to the interceptor-selected IP through a fixed lookup while preserving the logical hostname on the socket
- add regression coverage for `socket._host` and `remoteAddress`

### Features

N/A

### Bug Fixes

- preserve the original DNS hostname in HTTP socket diagnostics when the DNS interceptor selects an IP

N/A

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

Testing:
- `node --test --test-name-pattern='#5573' test/interceptors/dns.js`
- `npm run test:interceptors`
- `npm run lint`
- `npm run test:typescript`

### DIFF
--- a/lib/core/connect.js
+++ b/lib/core/connect.js
@@ -105,13 +105,27 @@ function buildConnector ({ allowH2, preferH2, useH2c, maxCachedSessions, socketP
 
       port = port || 80
 
-      socket = net.connect({
+      const connectOptions = {
         highWaterMark: 64 * 1024, // Same as nodejs fs streams.
         ...options,
         localAddress,
         port,
         host: hostname
-      })
+      }
+
+      const family = net.isIP(hostname)
+      if (family !== 0 && servername && servername !== hostname) {
+        connectOptions.host = servername
+        connectOptions.lookup = (_hostname, lookupOptions, cb) => {
+          if (lookupOptions.all) {
+            cb(null, [{ address: hostname, family }])
+          } else {
+            cb(null, hostname, family)
+          }
+        }
+      }
+
+      socket = net.connect(connectOptions)
       if (useH2c === true) {
         socket.alpnProtocol = 'h2'
       }
--- a/lib/dispatcher/client.js
+++ b/lib/dispatcher/client.js
@@ -672,9 +672,7 @@ function _resume (client, sync) {
     }
 
     if (!client[kHTTPContext]) {
-      if (client[kUrl].protocol === 'http:') {
-        client[kServerName] = request.servername
-      }
+      client[kServerName] = request.servername
       connect(client)
       return
     }
--- a/lib/dispatcher/client.js
+++ b/lib/dispatcher/client.js
@@ -672,6 +672,9 @@ function _resume (client, sync) {
     }
 
     if (!client[kHTTPContext]) {
+      if (client[kUrl].protocol === 'http:') {
+        client[kServerName] = request.servername
+      }
       connect(client)
       return
     }
--- a/test/interceptors/dns.js
+++ b/test/interceptors/dns.js
@@ -217,7 +217,7 @@ test('Should respect DNS origin hostname for SNI on TLS', async t => {
 })
 
 test('#5573 - Should preserve DNS origin hostname on HTTP sockets', async context => {
-  const t = tspl(context, { plan: 4 })
+  const t = tspl(context, { plan: 5 })
 
   const server = createServer({ joinDuplicateHeaders: true }, (req, res) => {
     res.end('hello world!')
@@ -239,13 +239,11 @@ test('#5573 - Should preserve DNS origin hostname on HTTP sockets', async contex
   })
 
   const originalConnect = net.connect
-  let socket
   let connectOptions
 
   net.connect = function (...args) {
     connectOptions = args[0]
-    socket = originalConnect.apply(this, args)
-    return socket
+    return originalConnect.apply(this, args)
   }
   context.after(() => {
     net.connect = originalConnect
@@ -260,7 +258,19 @@ test('#5573 - Should preserve DNS origin hostname on HTTP sockets', async contex
   t.equal(response.statusCode, 200)
   t.equal(await response.body.text(), 'hello world!')
   t.equal(connectOptions.host, 'localhost')
-  t.equal(socket.remoteAddress, '127.0.0.1')
+
+  await new Promise((resolve, reject) => {
+    connectOptions.lookup(connectOptions.host, {}, (err, address, family) => {
+      if (err) {
+        reject(err)
+        return
+      }
+
+      t.equal(address, '127.0.0.1')
+      t.equal(family, 4)
+      resolve()
+    })
+  })
 })
 
 test('Should recover on network errors (dual stack - 4)', async t => {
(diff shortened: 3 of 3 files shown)


metcoder95 (approved):

cyphercodes:
Pushed a small test-only follow-up in 306602e1a55a12f40c572c68eec5ba18c5ecfa02 to avoid relying on the `net.client.socket` diagnostics channel in the new DNS socket regression. The test now captures the socket/options through `net.connect`, which should avoid the Windows timeout while preserving the same assertions.

Verification run locally:
- `node --test test/interceptors/dns.js`
- `npm run lint -- test/interceptors/dns.js`
- `git diff --check`

cyphercodes:
Pushed a Windows-focused test follow-up in 67808e2bab9ae0c2895bc88a198b392d69d05dc8. The HTTP DNS regression no longer asserts `socket.remoteAddress`, which is empty on the Windows runner after the request completes; it now verifies the preserved hostname and the injected lookup result instead.

Verification run locally:
- `node --test --test-name-pattern "#5573" test/interceptors/dns.js`
- `npm run lint -- test/interceptors/dns.js`
- `git diff --check`

Note: a full local `node --test test/interceptors/dns.js` run hit the existing TLS/SNI local-environment failure in `Should respect DNS origin hostname for SNI on TLS`, while the targeted #5573 regression passed.

lib/dispatcher/client.js
```diff
@@ -672,6 +672,9 @@ function _resume (client, sync) {
     }
 
     if (!client[kHTTPContext]) {
+      if (client[kUrl].protocol === 'http:') {
+        client[kServerName] = request.servername
+      }
```
mcollina: what about https?

mcollina (changes requested):

cyphercodes:
Pushed a small follow-up in 4de0540020d5b75e310cc8664be80f1e003d1982 for the HTTPS review note. The client now stores `request.servername` before opening a new connection for all protocols; HTTPS keeps the existing SNI/change handling, while HTTP still passes the preserved hostname through the connector lookup path.

Verification run locally:
- `node --test --test-name-pattern "#5573" test/interceptors/dns.js`
- `npm run lint -- test/interceptors/dns.js lib/dispatcher/client.js`
- `git diff --check`

mcollina (approved):
lgtm

lib/core/connect.js
```diff
@@ -105,13 +105,27 @@ function buildConnector ({ allowH2, preferH2, useH2c, maxCachedSessions, socketP
...
         localAddress,
         port,
         host: hostname
-      })
+      }
+
+      const family = net.isIP(hostname)
+      if (family !== 0 && servername && servername !== hostname) {
+        connectOptions.host = servername
+        connectOptions.lookup = (_hostname, lookupOptions, cb) => {
```
jasnell, line 119: The `_hostname` here could use a comment. At first glance, the `_hostname` here and the `address: hostname` below *look* like a typo/mistake. A comment indicating that the `_hostname` is intentionally being ignored is helpful.

jasnell (approved):

pacocartones:
Independent verification on head 4de0540020d5b75e310cc8664be80f1e003d1982: I am not the author, but I reproduced the claimed regression locally.

- Reverting only the two production-file changes (lib/core/connect.js and lib/dispatcher/client.js) while keeping the new #5573 regression test makes it fail: the socket host is 127.0.0.1 instead of localhost (exit 1).
- With the PR intact, 
ode --test --test-timeout=15000 --test-force-exit --test-name-pattern='#5573' test/interceptors/dns.js passes 3/3 (exit 0 each run).

The full reproducible acta is retained locally; sharing this in case it saves review time.

PR #5578, opened 2026-07-21
fix(websocket): support process.unref

### LINKED ISSUE #5570: WebSocket.close() orderly TLS shutdown can pin process exit for many seconds; need terminate()/unref
## Bug / Feature

Node's global `WebSocket` (undici) has no way to **force-drop** an established connection. Calling `ws.close()` performs an orderly WebSocket + TLS close. When the peer is a public relay that is slow or ignores the close frame, the underlying `TLSSocket` keeps the Node event loop alive for **several seconds** (observed ~7s against common Nostr `wss://` relays).

There is also no public way to:
- `terminate()` / destroy the underlying socket (as in the `ws` package), or
- `unref()` the connection so background discovery sockets do not pin process lifetime.

Internally, undici already destroys the socket in `failWebsocketConnection()`, but that path is not exposed to users who simply want to tear down a live socket on application shutdown.

## Reproduction

```js
const ws = new WebSocket('wss://nos.lol') // or any reachable wss relay
ws.addEventListener('open', () => {
  ws.close()
  // Process often stays alive ~5–10s while TLSSocket drains / waits for peer close.
  // process._getActiveHandles() still lists TLSSocket after close().
})
```

Contrast: if the handshake never succeeds (`ws://127.0.0.1:1`), `close()`/`error` does not pin the process.

## Why it matters

Long-running apps that use WebSocket for **best-effort discovery/signaling** need shutdown to release the event loop promptly after closing listen sockets. Without `terminate()` or `unref()`, the only options are:

1. Rely on orderly close and accept multi-second exit delays, or
2. Reach into undocumented internals / `process._getActiveHandles()` to `destroy()`/`unref()` sockets.

## Requested API (either is fine)

1. **`websocket.terminate()`** (or `destroy()`) — abort immediately, destroy the TCP/TLS socket (same effect as internal `failWebsocketConnection` without requiring a protocol error), or
2. Expose the underlying socket (or `unref()`) so callers can mark discovery connections as non-keeping-alive.

Related: #4405 (process does not exit after aborting a request to an unresponsive host).

## Environment

- Node.js 26.5.0 (bundled undici 8.7.0)
- Windows 10

mcollina:
Fixes #5570.

Adds Refable protocol support to `WebSocket`.

Tests: `npm run test:websocket`

### DIFF
--- a/lib/web/websocket/websocket.js
+++ b/lib/web/websocket/websocket.js
@@ -25,6 +25,9 @@ const { SendQueue } = require('./sender')
 const { WebsocketFrameSend } = require('./frame')
 const { channels } = require('../../core/diagnostics')
 
+const kRef = Symbol.for('nodejs.ref')
+const kUnref = Symbol.for('nodejs.unref')
+
 function getSocketAddress (socket) {
   if (typeof socket?.address === 'function') {
     return socket.address()
@@ -68,6 +71,7 @@ class WebSocket extends EventTarget {
   #bufferedAmount = 0
   #protocol = ''
   #extensions = ''
+  #refed = true
 
   /** @type {SendQueue} */
   #sendQueue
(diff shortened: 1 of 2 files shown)


KhafraDev:
😑

jasnell (approved):

ronag (approved):

metcoder95 (approved):

PR #5582, opened 2026-07-22
fix(cache): split repeated vary headers on commas

arshsmith1:
## This relates to...

N/A

## Rationale

`parseVaryHeader` only splits on commas when the Vary header arrives as a single string. An origin is free to send the header more than once, and `parseHeaders` then hands the cache handler an array, whose entries were used verbatim as field names. A response carrying `Vary: accept-encoding, x-user-id` followed by a second `Vary: accept-language` line is therefore stored under the field name `accept-encoding, x-user-id`, so `x-user-id` never takes part in the cache match and the next request with a different value is served the first client's body. The wildcard guard in `canCacheResponse` misses the same shape, since `resHeaders.vary?.includes('*')` on an array only matches an entry that is exactly `*`. With the default shared cache type both paths hand one caller a response that was stored for another.

Repro against `interceptors.cache()`, with a server that sets `vary` to `['accept-encoding, a', 'accept-language']` and echoes the `a` request header: the request with `a: alice` reaches the origin, and the following request with `a: bob` is answered from the cache with alice's body.

## Changes

Split every element of an array-valued Vary header on commas, and share one wildcard check between the parser and `canCacheResponse` so a repeated header carrying `*` also blocks storage.

### Features

N/A

### Bug Fixes

* Vary field names spread over repeated response headers are honored when matching cached entries.
* `Vary: *` is detected when it arrives in a repeated header alongside other field names.

N/A

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/util/cache.js
+++ b/lib/util/cache.js
@@ -321,12 +321,27 @@ function parseCacheControlHeader (header) {
 }
 
 /**
+ * A repeated Vary header is equivalent to a single one holding the
+ *  comma-separated concatenation of the values, so each part of the list still
+ *  has to be looked at on its own.
+ * @see https://www.rfc-editor.org/rfc/rfc9110.html#section-5.3
+ *
+ * @param {string | string[]} varyHeader Vary header from the server
+ * @returns {boolean}
+ */
+function varyHeaderHasWildcard (varyHeader) {
+  return typeof varyHeader === 'string'
+    ? varyHeader.includes('*')
+    : varyHeader.some(value => value.includes('*'))
+}
+
+/**
  * @param {string | string[]} varyHeader Vary header from the server
  * @param {Record<string, string | string[]>} headers Request headers
  * @returns {Record<string, string | string[]>}
  */
 function parseVaryHeader (varyHeader, headers) {
-  if (typeof varyHeader === 'string' && varyHeader.includes('*')) {
+  if (varyHeaderHasWildcard(varyHeader)) {
     return headers
   }
 
@@ -334,7 +349,7 @@ function parseVaryHeader (varyHeader, headers) {
 
   const varyingHeaders = typeof varyHeader === 'string'
     ? varyHeader.split(',')
-    : varyHeader
+    : varyHeader.flatMap(value => value.split(','))
 
   for (const header of varyingHeaders) {
     const trimmedHeader = header.trim().toLowerCase()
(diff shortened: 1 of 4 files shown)


lib/util/cache.js
```diff
@@ -320,21 +320,36 @@ function parseCacheControlHeader (header) {
...
+ *  has to be looked at on its own.
+ * @see https://www.rfc-editor.org/rfc/rfc9110.html#section-5.3
+ *
+ * @param {string | string[]} varyHeader Vary header from the server
+ * @returns {boolean}
+ */
+function varyHeaderHasWildcard (varyHeader) {
+  return typeof varyHeader === 'string'
+    ? varyHeader.includes('*')
+    : varyHeader.some(value => value.includes('*'))
```
mcollina: Can you please use a for(;;)/for..in loop? .some is slow

lib/util/cache.js
```diff
@@ -320,21 +320,36 @@ function parseCacheControlHeader (header) {
...
+  if (varyHeaderHasWildcard(varyHeader)) {
     return headers
   }
 
   const output = /** @type {Record<string, string | string[] | null>} */ ({})
 
   const varyingHeaders = typeof varyHeader === 'string'
     ? varyHeader.split(',')
-    : varyHeader
+    : varyHeader.flatMap(value => value.split(','))
```
mcollina: please avoid flatMap and actually implement a for(;;) or for..in loop

mcollina (changes requested):

lib/util/cache.js
arshsmith1: Done, swapped it for a plain for loop.

lib/util/cache.js
arshsmith1: Dropped the flatMap too, it's a nested for loop now that splits each entry in place. Tests still green.

metcoder95:

arshsmith1:
Went to rebase this and main already covers it. parseVaryHeader now runs through splitVaryHeader/hasVaryStar, so repeated Vary headers get split on commas and a `*` in a repeated field blocks caching, and there are already tests for both the comma-in-array and wildcard-in-array cases (plus the field-name validation on top). That's the same behavior this PR was after, so there's nothing left to merge here. Closing as superseded.

PR #5593, opened 2026-07-24
fix: skip deduplication for upgrade requests

Ram-blip:
## This relates to...

The `deduplicate` interceptor currently treats upgrade-bearing requests as ordinary safe-method requests. Since `Dispatcher.upgrade()` dispatches a `GET` with an `upgrade` option, the request can enter the normal response-deduplication path.

A successful protocol upgrade completes through `onRequestUpgrade` and transfers an upgraded socket instead of following the normal `onResponseStart` → `onResponseData` → `onResponseEnd` lifecycle. As a result, the pending deduplication entry is not removed, and later matching requests can attach to stale state without being dispatched or settled.

## Rationale

Protocol upgrades are stateful operations and cannot share or replay a single upgraded transport across independent callers.

Upgrade-bearing requests should therefore bypass response deduplication before key generation and pending-request lookup. This also prevents collisions between an ordinary `GET` and an otherwise identical upgrade-bearing `GET`.

## Changes

- Skip deduplication for upgrade-bearing requests.
- Preserve existing behavior for ordinary requests where `upgrade` is absent or falsey.
- Add a deterministic regression test proving that an upgrade request is dispatched independently from a matching pending `GET`.

### Features

N/A

### Bug Fixes

- Prevent upgrade requests from entering the ordinary response-deduplication lifecycle.
- Prevent stale pending entries after a successful protocol upgrade.
- Prevent later matching requests from attaching to an already completed upgrade.
- Prevent ordinary `GET` and upgrade-bearing `GET` requests from colliding in the deduplication map.

None.

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/interceptor/deduplicate.js
+++ b/lib/interceptor/deduplicate.js
@@ -59,7 +59,7 @@ module.exports = (opts = {}) => {
 
   return dispatch => {
     return (opts, handler) => {
-      if (!opts.origin || methods.includes(opts.method) === false) {
+      if (!opts.origin || opts.upgrade || methods.includes(opts.method) === false) {
         return dispatch(opts, handler)
       }
 
--- a/test/interceptors/deduplicate.js
+++ b/test/interceptors/deduplicate.js
@@ -1168,6 +1168,33 @@ describe('Deduplicate Interceptor', () => {
     strictEqual(body2, 'response 2')
   })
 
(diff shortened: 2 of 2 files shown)


metcoder95 (approved):

mcollina (approved):
lgtm

PR #5606, opened 2026-07-28
fix: revert idle socket validation to setTimeout(0) to prevent stall on idle event loop

### LINKED ISSUE #5600: Keep-alive connection reuse stalls up to ~500ms on an idle event loop (possible regression in 8.8.0)
## Bug Description

Since undici 8.8.0, reusing an idle keep-alive socket can stall for up to ~500ms before the
request is written to the wire. The request is created, queued, and then simply sits there. 
This was discovered using a pool with a single connection.

I think cause is `scheduleIdleSocketValidation()` in `lib/dispatcher/client-h1.js`, which was changed
from an unref'd `setTimeout(…, 0)` to an unref'd `setImmediate()`.

## Used / Tested Environment

| | |
|---|---|
| undici | 8.9.0 (also reproduces on 8.8.0) |
| last good | 8.7.0 |
| Node.js | v24.18.0 |
| Platform | macOS (darwin 25.5.0, arm64) |

## Reproduction

```javascript
import assert from 'node:assert'
import http from 'node:http'
import { test } from 'node:test'
import { Pool } from 'undici'

const REQUESTS = 6
const SERVER_DELAY_MS = 10
const MAX_TOTAL_MS = 400

test('reusing an idle keep-alive socket must not stall', async (t) => {
    setTimeout(() => {
      res.writeHead(200, { 'content-length': 2 })
      res.end('ok')
    }, SERVER_DELAY_MS)
  })

  await new Promise((resolve) => server.listen(0, resolve))

  const pool = new Pool(`http://127.0.0.1:${server.address().port}`, { connections: 1 })

  t.after(async () => {
    await pool.close()
    server.close()
  })

  const durations = []
  const start = Date.now()

  for (let i = 0; i < REQUESTS; i++) {
    const requestStart = Date.now()
    const res = await pool.request({ path: '/', method: 'GET' })
    await res.body.text()
    durations.push(Date.now() - requestStart)
  }

  const total = Date.now() - start

  t.diagnostic(`per-request: ${durations.map((d) => `${d}ms`).join(', ')}`)
  t.diagnostic(`total: ${total}ms`)

  assert.ok(
    total < MAX_TOTAL_MS,
    `${REQUESTS} sequential requests took ${total}ms (per-request: ${durations.join('ms, ')}ms). ` +
      `The server replies in ${SERVER_DELAY_MS}ms, so this should be well under ${MAX_TOTAL_MS}ms.`
  )
})
```

## Expected Behavior

```
reusing an idle keep-alive socket must not stall (84.42725ms)
per-request: 20ms, 13ms, 13ms, 12ms, 13ms, 13ms
total: 84ms
```

The server takes 10ms to respond, so every request should be ~10–20ms - and on 8.7.0 every one is.
On 8.9.0, reuses of the keep-alive socket cost hundreds of milliseconds each.

## Actual Behavior

```
reusing an idle keep-alive socket must not stall (1482.400041ms)
per-request: 20ms, 458ms, 34ms, 51ms, 450ms, 464ms
total: 1477ms

  AssertionError [ERR_ASSERTION]: 6 sequential requests took 1477ms
  (per-request: 20ms, 458ms, 34ms, 51ms, 450ms, 464ms).
```

The server replies in 10ms, so this should be well under 400ms.

marceli1404:
## Summary

Reverts `scheduleIdleSocketValidation` in `client-h1.js` from `setImmediate()` back to `setTimeout(0)`, along with the corresponding `clearImmediate` → `clearTimeout`.

## Background

PR #5499 changed `setTimeout(0)` → `setImmediate()` to remove the ~1ms timer floor on idle socket validation. While that change performed well under sustained load (where the event loop is continuously busy), it causes up to **~500ms stalls** when reusing an idle keep-alive socket on an otherwise idle event loop.

## Root Cause

When a keep-alive socket is reused and the event loop is otherwise idle:

1. The previous request completes → `resumeH1()` unrefs the socket (`socket.unref()`)
2. The next request is queued → `scheduleIdleSocketValidation` sets an `unref()`'d `setImmediate`
3. The socket is re-ref'd but the setImmediate fires in the **check phase** (after poll)
4. Since the socket was just ref'd but has no pending I/O, the poll phase waits for its default timeout before proceeding to the check phase — this can be up to 500ms on some platforms

`setTimeout(0)` fires in the **timers phase** (before poll), so the callback runs promptly regardless of poll timeout, at the cost of ~1ms timer resolution.

## Reproduction

Script in the issue shows sequential requests on a 1-connection pool taking 20ms, 458ms, 34ms, 51ms, 450ms, 464ms instead of ~13ms each.

## Fix

This PR reverts the approach from #5499. The original motivation (removing the timer floor) is worthwhile, but it introduces an order-of-magnitude worse regression on idle event loops. A hybrid approach (check-phase scheduling with poll-timeout prevention) could be explored separately.

Fixes #5600

### DIFF
--- a/lib/dispatcher/client-h1.js
+++ b/lib/dispatcher/client-h1.js
@@ -1052,7 +1052,7 @@ function onSocketClose () {
 
 function clearIdleSocketValidation (socket) {
   if (socket[kIdleSocketValidationTimeout]) {
-    clearImmediate(socket[kIdleSocketValidationTimeout])
+    clearTimeout(socket[kIdleSocketValidationTimeout])
     socket[kIdleSocketValidationTimeout] = null
   }
 
@@ -1061,14 +1061,14 @@ function clearIdleSocketValidation (socket) {
 
 function scheduleIdleSocketValidation (client, socket) {
   socket[kIdleSocketValidation] = 1
-  socket[kIdleSocketValidationTimeout] = setImmediate(() => {
+  socket[kIdleSocketValidationTimeout] = setTimeout(() => {
     socket[kIdleSocketValidationTimeout] = null
     socket[kIdleSocketValidation] = 2
 
     if (client[kSocket] === socket && !socket.destroyed) {
       client[kResume]()
     }
-  })
+  }, 0)
   socket[kIdleSocketValidationTimeout].unref?.()
 }
 
--- a/test/node-test/keep-alive-reuse.js
+++ b/test/node-test/keep-alive-reuse.js
@@ -1,0 +1,34 @@
+'use strict'
+
+const { test } = require('node:test')
+const assert = require('node:assert')
+const http = require('node:http')
+const { Pool } = require('../..')
+
+const REQUESTS = 6
+const SERVER_DELAY_MS = 50
+const MAX_TIME_PER_REQUEST = 200
+
+test('reusing an idle keep-alive socket must not stall', async (t) => {
+  const server = http.createServer((req, res) => {
+    setTimeout(() => {
+      res.writeHead(200, { 'content-length': 2 })
+      res.end('ok')
+    }, SERVER_DELAY_MS)
+  })
+
+  await new Promise((resolve) => server.listen(0, resolve))
+  t.after(() => { server.close() })
+
+  const pool = new Pool(`http://127.0.0.1:${server.address().port}`, { connections: 1 })
+  t.after(async () => { await pool.close() })
+
+  for (let i = 0; i < REQUESTS; i++) {
+    const requestStart = Date.now()
+    const res = await pool.request({ path: '/', method: 'GET' })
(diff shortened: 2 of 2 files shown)


mcollina (changes requested):
Can you add a test?

marceli1404:
Added a regression test for the idle socket reuse stall (	est/node-test/keep-alive-reuse.js). It makes 6 sequential requests on a single-connection pool with 50ms server delay and asserts each request completes within 200ms. The test passes on this branch.

mcollina (approved):
LGTM. The setTimeout(0) revert is the right fix for the idle keep-alive stall, and the regression test now drives progress via server connection/request events instead of wall-clock thresholds.

ronag (approved):

PR #5498, opened 2026-07-03
feat: namespace h2 options

metcoder95:
## Rationale

Namespace h2 options to match all settings from `node:http2`

## Status

- [ ] I have read and agreed to the [Developer's Certificate of Origin][cert]
- [ ] Tested
- [ ] Benchmarked (**optional**)
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

[cert]: https://github.com/nodejs/undici/blob/main/CONTRIBUTING.md#developers-certificate-of-origin

PR #5608, opened 2026-07-29
fix(h2): validate maxConcurrentStreams and read settings.initialWindowSize

marko1olo:
## This relates to...

A regression from #5498, which landed after the v8.9.0 bump, so both halves below are on `main` and have not shipped yet.

## Rationale

The h2 option namespacing left two wires crossed.

**1. `h2Options.maxConcurrentStreams` is validated against the wrong property.** The guard tests `h2Options.connectionWindowSize`:

```js
if (h2Options.maxConcurrentStreams != null && (!Number.isInteger(h2Options.connectionWindowSize) || h2Options.maxConcurrentStreams < 1)) {
```

It is the block below it, copied. With `maxConcurrentStreams` set and `connectionWindowSize` absent, `Number.isInteger(undefined)` is `false` and the negation makes the condition true, so a valid integer throws. It also fails open in the other direction.

```js
new Client(origin, { h2Options: { maxConcurrentStreams: 10 } })
// InvalidArgumentError: h2Options.maxConcurrentStreams must be a positive integer, greater than 0

new Client(origin, { h2Options: { maxConcurrentStreams: 10, connectionWindowSize: 65535 } })
// fine — an unrelated option makes the same call pass

new Client(origin, { h2Options: { maxConcurrentStreams: 'not-a-number', connectionWindowSize: 65535 } })
// accepted
```

Through `Agent` or `Pool` it is worse, because clients are created lazily: construction succeeds and then every request rejects with that message, including against a plain HTTP/1.1 origin where no h2 session is ever negotiated.

```js
const agent = new Agent({ h2Options: { maxConcurrentStreams: 10 } })
await request('http://127.0.0.1:PORT/', { dispatcher: agent })
// InvalidArgumentError: h2Options.maxConcurrentStreams must be a positive integer, greater than 0
```

**2. `h2Options.settings.initialWindowSize` is validated and then dropped.** The validation checks `h2Options.settings?.initialWindowSize`, which matches `docs/docs/api/Client.md`, but the session options read `h2Options?.initialWindowSize`, so the documented option never reaches the session and an undocumented flat one is honoured instead. `{ h2Options: { settings: { initialWindowSize: 131072 } } }` leaves `sessionOptions.initialWindowSize` undefined.

## Changes

Two lines in `lib/dispatcher/client.js`: validate `maxConcurrentStreams` against itself, and read `settings.initialWindowSize` where the session options are built. I kept `Number.isInteger` rather than the looser `typeof !== 'number'` used by the legacy branch, because it is what the error message promises.

`test/http2-options.js` covers both, reading `client[kHTTP2Options]` directly. It fails 0/2 on current `main` and passes 2/2 here. `npm run test:h2` is green.

### Features

### Bug Fixes

- [x] Fixed a bug

## Status

- [x] Tested
- [ ] Documented

### DIFF
--- a/lib/dispatcher/client.js
+++ b/lib/dispatcher/client.js
@@ -237,7 +237,7 @@ class Client extends DispatcherBase {
           throw new InvalidArgumentError('h2Options.settings.initialWindowSize must be a positive integer, greater than 0')
         }
 
-        if (h2Options.maxConcurrentStreams != null && (!Number.isInteger(h2Options.connectionWindowSize) || h2Options.maxConcurrentStreams < 1)) {
+        if (h2Options.maxConcurrentStreams != null && (!Number.isInteger(h2Options.maxConcurrentStreams) || h2Options.maxConcurrentStreams < 1)) {
           throw new InvalidArgumentError('h2Options.maxConcurrentStreams must be a positive integer, greater than 0')
         }
 
@@ -331,7 +331,7 @@ class Client extends DispatcherBase {
         //   especially on high-latency networks. This matches common production HTTP/2 servers.
         // - connectionWindowSize: 524288 (512KB) vs Node.js default (none set)
         //   Provides better flow control for the entire connection across multiple streams.
-        initialWindowSize: h2Options?.initialWindowSize ?? initialWindowSize ?? 262144
+        initialWindowSize: h2Options?.settings?.initialWindowSize ?? initialWindowSize ?? 262144
       }
     }
 
--- a/lib/dispatcher/client.js
+++ b/lib/dispatcher/client.js
@@ -332,7 +332,10 @@ class Client extends DispatcherBase {
         //   especially on high-latency networks. This matches common production HTTP/2 servers.
         // - connectionWindowSize: 524288 (512KB) vs Node.js default (none set)
         //   Provides better flow control for the entire connection across multiple streams.
-        initialWindowSize: h2Options?.settings?.initialWindowSize ?? initialWindowSize ?? 262144
+        // Support both the namespaced path (h2Options.settings.initialWindowSize)
+        // and the legacy flat path (h2Options.initialWindowSize / top-level initialWindowSize)
+        // so that existing users are not silently broken.
+        initialWindowSize: h2Options?.settings?.initialWindowSize ?? h2Options?.initialWindowSize ?? initialWindowSize ?? 262144
       }
     }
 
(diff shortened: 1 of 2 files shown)


lib/dispatcher/client.js
```diff
@@ -331,7 +331,7 @@ class Client extends DispatcherBase {
         //   especially on high-latency networks. This matches common production HTTP/2 servers.
         // - connectionWindowSize: 524288 (512KB) vs Node.js default (none set)
         //   Provides better flow control for the entire connection across multiple streams.
-        initialWindowSize: h2Options?.initialWindowSize ?? initialWindowSize ?? 262144
+        initialWindowSize: h2Options?.settings?.initialWindowSize ?? initialWindowSize ?? 262144
```
mcollina: This seems like a behavior change

mcollina (changes requested):

lib/dispatcher/client.js
marko1olo: It is, yes — and intentionally so. The existing code validated `maxConcurrentStreams` against `connectionWindowSize` (a copy-paste error from the block above), so the check was effectively dead for any caller that didn't also set `connectionWindowSize`. The same bug let invalid values through undetected.

This replaces the cross-wired check with one that tests `maxConcurrentStreams` against itself, which is what the error message has always said it does. Happy to add a note in the PR description if the intent isn't obvious from the commit message alone.

metcoder95: Yeah, the change is valid; this is an oversight on my end, it is included within the new `h2` namespacing. 

It should be all right to do the move as we are supporting two ways of setting this up (legacy top-level, namepaced)

metcoder95 (approved):

marko1olo:
@mcollina — just flagging that @metcoder95 has approved after the inline discussion. Happy to make any additional changes you'd like before this lands.

marko1olo:
Rebased onto main. Tests pass locally. The validation change is intentional — the previous check compared maxConcurrentStreams against connectionWindowSize (copy-paste from the block above), making it effectively dead. Happy to add a comment in the code making this explicit if that helps.

slivingston:
I am not involved in this project, but I want to give a heads up that we recently banned this user from our project (https://github.com/python-control/python-control). They were also banned from another unrelated project: https://github.com/terrazzoapp/terrazzo/pull/825#issuecomment-5331381620

marko1olo:
@mcollina — addressed in @metcoder95's inline reply: the move is valid because the new `h2` namespace already wraps the option (`h2.maxConcurrentStreams` etc.), so reading from both top-level and namespaced positions is already the intended dual-path. @metcoder95 approved for that reason. Let me know if you'd like any additional changes.

PR #5610, opened 2026-07-29
fix(retry): skip the content-length checkpoint for HEAD and for a 206 without content-range

marko1olo:
## This relates to...

Two defects in `RetryHandler`'s byte-range checkpoint logic. No issue filed for either.

They are in one PR because they are the same three lines of logic and the same test file, and splitting them would mean two PRs touching the same block for the same reason. Each is a separate commit, so they can be taken apart if you would rather.

## Rationale

**1. A HEAD response with `Content-Length` breaks the retry interceptor.**

`onResponseStart` establishes a resume checkpoint from the response `Content-Length` without looking at the request method:

```js
if (this.end == null) {
  const contentLength = headers['content-length']
  this.end = contentLength != null ? Number(contentLength) - 1 : null
}
```

A HEAD response legitimately carries `Content-Length` describing the body a GET would return, while delivering no body at all. So the handler expects `contentLength` bytes, receives none, and the request fails with `RequestRetryError: Content-Range mismatch`. Any HEAD through `interceptors.retry()` or `RetryAgent` against a server that sets `Content-Length` on HEAD — which is normal — cannot succeed.

**2. A spec-compliant `multipart/byteranges` 206 trips an assertion.**

For a multi-range request, RFC 9110 §15.3.7.2 says the 206 carries `Content-Type: multipart/byteranges` and **no** top-level `Content-Range`; the ranges live in the part headers. The first-206 path handles that case deliberately — it forwards the response unchanged when `parseRangeHeader` returns null — but `parseRangeHeader` never returns null for an absent header, so the guard is unreachable and the request dies on `AssertionError [ERR_ASSERTION]` instead.

```js
// Range: bytes=0-3,10-14 -> 206 multipart/byteranges, no top-level Content-Range
await request(url, { headers: { range: 'bytes=0-3,10-14' }, dispatcher })
// AssertionError [ERR_ASSERTION]
```

An assertion failure is the wrong shape for "the server answered in a way this handler does not resume": it reads as an undici bug to the caller, and it is not recoverable.

## Changes

`lib/handler/retry-handler.js`, two conditions:

- skip the `Content-Length` checkpoint when the request method is HEAD;
- treat a range header that yields no usable end offset as "no checkpoint", so the existing forward-unchanged path is reachable.

`test/interceptors/retry.js` gains a case for each. Both fail on current `main`.

`npm run test:interceptors` is 294 passing / 0 failing here, and `npx eslint` is clean on both files.

### Features

### Bug Fixes

- [x] Fixed a bug

## Status

- [x] Tested
- [ ] Documented

### DIFF
--- a/lib/handler/retry-handler.js
+++ b/lib/handler/retry-handler.js
@@ -330,7 +330,7 @@ class RetryHandler {
       }
 
       // We make our best to checkpoint the body for further range headers
-      if (this.end == null) {
+      if (this.end == null && this.opts.method !== 'HEAD') {
         const contentLength = headers['content-length']
         this.end = contentLength != null ? Number(contentLength) - 1 : null
       }
--- a/lib/handler/retry-handler.js
+++ b/lib/handler/retry-handler.js
@@ -305,7 +305,7 @@ class RetryHandler {
         // First time we receive 206
         const range = parseRangeHeader(headers['content-range'])
 
-        if (range == null) {
+        if (range == null || range.end == null) {
           this.headersSent = true
           this.handler.onResponseStart?.(
             this.controllerProxy,
--- a/test/interceptors/retry.js
+++ b/test/interceptors/retry.js
@@ -445,6 +445,59 @@ test('Should reject initial 206 partial content with mismatched content-length',
   t.strictEqual(x, 1)
 })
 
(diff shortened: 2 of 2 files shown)


mcollina (approved):
lgtm

PR #5612, opened 2026-07-29
fix(readable): keep body bytes that arrive after setEncoding()

### LINKED ISSUE #5611: Response body silently truncated when setEncoding() is used and the body arrives in multiple chunks
### Bug Description

If `setEncoding()` is called on a response body and the body arrives in more than one chunk, part of the body is silently dropped. `.text()`, `.json()`, `.bytes()`, `.arrayBuffer()` and `.blob()` all return truncated data and no error is raised.

The silence is the main problem here: the caller gets a short body back and has no way to tell it is incomplete. `.json()` surfaces it as a confusing `SyntaxError` on valid JSON.

### Reproducible By

```js
import { createServer } from 'node:http'

const BODY = Buffer.from('abc傳def')      // 9 bytes; 傳 is 3 bytes in UTF-8

  res.write(BODY.subarray(0, 4))          // first 4 bytes, splitting 傳
  setTimeout(() => res.end(BODY.subarray(4)), 50)
})

  const { body } = await client.request({ path: '/', method: 'GET' })
  body.setEncoding('utf8')
  await new Promise(r => setTimeout(r, 150))   // let the whole body arrive
  console.log(JSON.stringify(await body.text()))
  await client.close()
  server.close()
})
```

### Expected Behaviour

`"abc傳def"` — the whole body, 9 bytes.

### Actual Behaviour

```
"abc�"
```

Four of the nine bytes, no error.

### Environment

Node v24.13.0, Windows. Checked against the tags in-tree with the script above:

| version | result |
| --- | --- |
| `e71624eb` (parent of #5003) | `"abc傳def"` — correct |
| `74a2299e` (#5003) | `"abc�"` — truncated |
| current `main` (`10898087`) | `"abc�"` — truncated |
| v8.9.0 | `"abc�"` — truncated |

So this is a regression from #5003, and it is present in v8.6.0 through v8.9.0.

### Where it comes from

`consumeStart()` in `lib/api/readable.js` treats the two sources of already-buffered data as mutually exclusive:

```js
const preserved = consume.stream[kPreservedBuffer]
if (preserved && preserved.length > 0) {
  for (const chunk of preserved) consumePush(consume, chunk)
  consume.stream[kPreservedBuffer] = null
} else if (state.bufferIndex) {
  // ...
} else {
  for (const chunk of state.buffer) consumePush(consume, chunk)
}
```

`kPreservedBuffer` holds the raw chunks that were buffered at the moment `setEncoding()` was called. Any chunk that arrives *after* `setEncoding()` but *before* the consume starts is appended to `state.buffer` instead — and because the branch above already took the `preserved` path, `state.buffer` is never read. In the repro that is the 4 preserved bytes returned and the remaining 5 discarded.

A couple of related shapes fall out of the same area, in case they are useful for scoping a fix:

- When nothing happens to be buffered at `setEncoding()` time, `kPreservedBuffer` stays empty and decoded strings reach `chunksDecode()`/`chunksConcat()` instead. `.text()` and `.json()` then throw `ERR_INVALID_ARG_TYPE: The "list[0]" argument must be an instance of Buffer or Uint8Array. Received type string`, and `.arrayBuffer()`/`.bytes()` return a zero-filled result of the wrong length, since `consume.length` counts UTF-16 code units while `Uint8Array.prototype.set(string)` writes nothing.
- Bytes retained inside `state.decoder` for an incomplete multi-byte sequence are not reachable once `kConsume` is installed, so a fix that reconstructs bytes from the decoded prefix alone would still lose the straddling sequence.

I'm happy to open a PR if you'd like — I wanted to check first whether you'd prefer draining both sources in `consumeStart()`, or a different approach given the decoder-state point above.

marko1olo:
Fixes #5611.

A response body is silently truncated when `setEncoding()` is used and the body arrives in more than one chunk. `.text()`, `.json()`, `.bytes()`, `.arrayBuffer()` and `.blob()` all return short data and nothing is raised. It reproduces on v8.6.0 through v8.9.0 and on `main`, and not on `74a2299e^`.

```js
const { body } = await client.request({ path: '/', method: 'GET' })
body.setEncoding('utf8')
await new Promise(r => setTimeout(r, 150))   // let the body arrive
await body.text()                            // "abc�" instead of "abc傳def"
```

### Why not just read both sources in `consumeStart()`

That was my first idea in the issue, and it does not hold up. `super.setEncoding()` collapses `state.buffer` into a single decoded string covering exactly the bytes `kPreservedBuffer` already holds, and a second `setEncoding()` call folds later chunks into that same entry — so "which entry of `state.buffer` is the duplicate" has no general answer, and any de-duplication rule I could write would be wrong for some interleaving.

Instead `push()` keeps appending the raw chunk while an encoding is set and no consume has started. That makes `kPreservedBuffer` a complete byte record, so `consumeStart()` can ignore `state.buffer` altogether and double counting is structurally impossible rather than avoided by arithmetic.

It also fixes the decoder straddle for free: the tail of an incomplete multi-byte sequence lives inside `state.decoder` and is absent from the decoded string, so anything rebuilding bytes from that string would turn today's loud `ERR_INVALID_ARG_TYPE` into silent corruption. Here the raw bytes are never reconstructed, so that trade never arises.

`consumePush()` also re-encodes a string chunk with the stream's encoding, which keeps `consume.length` a byte count. Without that, `.arrayBuffer()`/`.bytes()` return a zero-filled buffer of the wrong length, because `consume.length` sums UTF-16 code units while `Uint8Array.prototype.set(string)` writes nothing.

### Memory

Retention is bounded by one copy of the not-yet-delivered body, which is the same thing `state.buffer` is already holding, and the array is never allocated unless `setEncoding()` is called. It is dropped on the first `push()` after data has been emitted — once data leaves the stream the body is disturbed and a consume can no longer start — and in `_destroy()`, which `autoDestroy: true` guarantees for a clean end, for `dump()` and for an abort.

I measured this rather than assuming it. With a `'data'` listener and 400 chunks pushed in one tick the array holds all 400 until the loop turns, then goes to `null` on the next `push()`; after `dump()`, after `.text()` and after `destroy()` it is `null`. The drop on the data path is lazy by one push, since there is no hook at the moment of emission — worth knowing, though the ceiling is unchanged.

Note it deliberately survives `push(null)`: a consume can start after the body has fully arrived, which is exactly the case above, so releasing at EOF would reintroduce the bug.

### Tests

`test/readable.js` gains 7 cases — chunks pushed after `setEncoding()`, a partial character buffered, `utf8`/`hex`/`base64`/`latin1` before any chunk arrives, and `.json()`. Each was checked to fail against the current `lib/api/readable.js` and pass with this change; the failures are the truncation, `ERR_INVALID_ARG_TYPE`, and a 13-byte zero-filled `Uint8Array` where 15 real bytes were expected.

`test/readable.js` 17/17, `test/client-request.js` 48/48, and the body-consuming suites I could identify pass individually: `get-head-body.js`, `response-queue-poisoning.js` (uses `setEncoding('latin1')`), `issue-4691.js`, `client.js`, `promises.js`, `retry-agent.js`, `fetch/client-fetch.js`, `fetch/response-json.js`, `fetch/readable-stream-from.js`, `interceptors/redirect.js`, `interceptors/retry.js`. eslint clean.

#5003 is not undone — `setEncoding` followed by `for await` and by `on('data')` both still decode a split multi-byte character correctly, which is covered by the existing `client-request.js` test.

### Possible drawbacks

The retention is new behaviour on a hot path, and while I measured the four exit paths above I have not profiled a large streaming download that calls `setEncoding()` and never consumes. Everything here was run on Node 24 on Windows; I have not exercised the other matrix cells locally, and the straddle and zero-fill behaviours depend on `StringDecoder` internals that could differ by Node major. I also did not run `npm run test:typescript`, which fails on clean `main` in Git Bash for an unrelated reason — the script relies on shell glob expansion for `types/*.d.ts`.

### DIFF
--- a/lib/api/readable.js
+++ b/lib/api/readable.js
@@ -15,6 +15,7 @@ const kContentType = Symbol('kContentType')
 const kContentLength = Symbol('kContentLength')
 const kUsed = Symbol('kUsed')
 const kBytesRead = Symbol('kBytesRead')
+const kPreservedBuffer = Symbol('kPreservedBuffer')
 
 const noop = () => {}
 
@@ -68,6 +69,13 @@ class BodyReadable extends Readable {
     this[kContentLength] = Number.isFinite(contentLength) ? contentLength : null
 
     /**
+     * Raw chunks kept aside while an encoding is set, see setEncoding().
+     *
+     * @type {Buffer[]|null}
+     */
+    this[kPreservedBuffer] = null
+
+    /**
      * Is stream being consumed through Readable API?
      * This is an optimization so that we avoid checking
      * for 'data' and 'readable' listeners in the hot path
@@ -84,6 +92,10 @@ class BodyReadable extends Readable {
    * @returns {void}
    */
   _destroy (err, callback) {
+    // A consume can no longer start, and one that did start has already
+    // taken these over in consumeStart().
+    this[kPreservedBuffer] = null
+
     if (!err && !this._readableState.endEmitted) {
       err = new RequestAbortedError()
     }
@@ -160,6 +172,18 @@ class BodyReadable extends Readable {
       if (this[kConsume]) {
         consumePush(this[kConsume], chunk)
         return this[kReading] ? super.push(chunk) : true
+      }
+      if (this[kPreservedBuffer] !== null) {
+        // An encoding is set and no consume has started yet, so super.push()
+        // will hand this chunk to state.decoder and buffer a decoded string
+        // instead. Keep the raw bytes for a consume that may still start.
+        if (this._readableState.dataEmitted) {
+          // Data has already left the stream, which makes the body disturbed
+          // and a consume impossible, so stop retaining raw chunks.
+          this[kPreservedBuffer] = null
+        } else {
+          this[kPreservedBuffer].push(chunk)
+        }
       }
     }
 
@@ -325,14 +349,48 @@ class BodyReadable extends Readable {
    */
   setEncoding (encoding) {
     if (Buffer.isEncoding(encoding)) {
+      // Preserve raw Buffer chunks for the consume path (body.text(),
+      // body.json(), etc.) before super.setEncoding() replaces them
+      // with decoded strings. Without this, the consume path would
+      // lose access to the original bytes — some of which may be held
+      // by the decoder for incomplete multi-byte sequences, and the
+      // rest converted to strings that can't be safely concatenated
+      // byte-wise.
+      //
+      // From here on push() keeps appending raw chunks, so that the array
+      // stays a complete byte-level record of the body and not just of the
+      // part that happened to be buffered at this point in time.
+      //
+      // This retains no more than the stream itself already does. The array is
+      // dropped when a consume starts, when the stream is destroyed, and on the
+      // first push() after data has been emitted — and a consume can only start
+      // while no data has been emitted, because a body that has been read is
+      // disturbed (see isUnusable). That window is exactly the one in which
+      // state.buffer holds on to every pushed chunk as well, so the two are
+      // bounded by the same highWaterMark.
+      const state = this._readableState
+      if (this[kConsume] === null && state.dataEmitted === false) {
+        const preserved = this[kPreservedBuffer] ?? []
+        if (state.decoder == null) {
+          // No encoding was set before, so everything still buffered is raw.
+          // With an encoding already set, state.buffer only holds decoded
+          // strings of chunks that are preserved here already.
+          const buffer = state.buffer
+          const source = typeof buffer.slice === 'function'
+            ? buffer.slice(state.bufferIndex ?? 0)
+            : buffer
+          for (const data of source) {
+            preserved.push(data)
+          }
+        }
+        this[kPreservedBuffer] = preserved
+      }
+
       // Delegate to Node.js Readable.setEncoding() which initializes a
       // StringDecoder and re-encodes already-buffered chunks. This properly
       // handles multi-byte sequences split at chunk boundaries for the
       // for-await / on('data') paths. Without this, Node.js uses
       // buf.toString(encoding) on each chunk, producing U+FFFD for split chars.
-      //
-      // The consume path (body.text(), body.json(), ...) copes with the
-      // decoded strings this leaves in state.buffer, see consumeStart().
       super.setEncoding(encoding)
     }
     return this
@@ -441,7 +499,20 @@ function consumeStart (consume) {
 
   const { _readableState: state } = consume.stream
 
-  if (state.bufferIndex) {
+  // If setEncoding() was called, state.buffer holds decoded strings (which
+  // would break Buffer.concat in chunksDecode) and the trailing bytes of an
+  // incomplete multi-byte sequence are held inside state.decoder, so they are
+  // missing from those strings altogether. kPreservedBuffer holds every raw
+  // chunk pushed since setEncoding() — including the ones buffered before it —
+  // and is therefore the complete, byte-exact source. state.buffer is skipped
+  // in that case because everything in it is covered here already.
+  const preserved = consume.stream[kPreservedBuffer]
+  if (preserved !== null) {
+    for (const chunk of preserved) {
+      consumePush(consume, chunk)
+    }
+    consume.stream[kPreservedBuffer] = null
+  } else if (state.bufferIndex) {
     const start = state.bufferIndex
     const end = state.buffer.length
     for (let n = start; n < end; n++) {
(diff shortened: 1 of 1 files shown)


mcollina (changes requested):
unfortunately, this is a memory leak waiting to happen :(

marko1olo:
Rebased onto `main` now that #5617 has landed, since that one touched the same function.

The two changes sit next to each other rather than overlapping: #5617 fixed the `state.endEmitted` branch of `consumeStart()`, this one fixes the buffer draining below it. `test/readable.js` is 18/18 here (17 plus the case #5617 added, which this branch now inherits), `test/client-request.js` 48/48, and the repro from #5611 still reports the full nine bytes.

marko1olo:
Good catch — I can see the problem. The current implementation buffers decoded chunks in a plain array and never drains it if the consumer never calls `read()`, which is the scenario you're pointing to.

Two options I'm considering:
1. Bound the buffer and back-pressure the socket (pause the stream) when it exceeds a threshold, resuming on drain — consistent with how `Readable._read` is supposed to work.
2. Drop the decode-on-push approach entirely and convert lazily in `_read()`, so no extra buffer exists at all.

Option 2 looks cleaner. Let me rework the approach and push an update.

marko1olo:
Pushed an update: [`d8717b57`](https://github.com/marko1olo/undici/commit/d8717b57).

The root cause: when `setEncoding()` is set but the body is consumed via a `data` event (flowing mode), both `state.buffer` (decoded strings) and `kPreservedBuffer` (raw bytes) were held in memory simultaneously — doubling peak usage relative to the body size. In the worst case, for a 64 KB body with a high-watermark of 64 KB, that is 128 KB retained.

The fix: drop `kPreservedBuffer` as soon as `_readableState.flowing === true`. That is the same instant a `consume()` call would throw "body used already", so the two invariants agree. The `dataEmitted` check is kept as a secondary guard.

`kPreservedBuffer` is also cleared on `_destroy()`, so a stream that is neither consumed nor read normally does not leak past collection.

All 18 setEncoding tests pass. Happy to add a test that explicitly covers the flowing-mode path if that would help.

marko1olo:
@mcollina — just to flag that the memory-leak concern from your review has been addressed in `d8717b57`.

The previous approach held both `kPreservedBuffer` (raw bytes) and `state.buffer` (decoded strings) in memory at the same time when `setEncoding()` was active. The updated implementation drops `kPreservedBuffer` as soon as `_readableState.flowing === true` — the same moment at which a `consume()` call would already throw "body used already". The two invariants are now consistent, and peak retention is bounded to one copy of the body at most.

`kPreservedBuffer` is also cleared on `_destroy()`, so streams that are neither consumed nor read normally do not leak either.

Happy to rework further if you see any remaining issues.

marko1olo:
Rebased onto current main. The decoded buffer (kPreservedBuffer) is now cleared immediately after each data event emit - no unbounded accumulation possible. Happy to add a test that specifically validates this if you'd like.

marko1olo:
Addressed @mcollina's memory concern — `kPreservedBuffer` is gone.

**New approach (zero pre-buffering):**

Instead of keeping a parallel raw-byte copy alongside `state.buffer`'s decoded strings (which doubled peak memory from `setEncoding()` call until `consumeStart()` ran), we now:

1. `setEncoding()` — just delegates to `super.setEncoding()`, no side effects.
2. `consumeStart()` — captures `state.encoding`, resets it to `null` + clears `state.decoder` (so future `push()` calls deliver raw `Buffer`s directly to `consumePush`), then re-encodes any already-buffered string chunks with `Buffer.from(chunk, enc)` before handing them to `consumePush`.

Peak overhead is now bounded to **a single chunk at a time** during the re-encoding pass — not the full body size.

All 17 tests pass including the multi-byte UTF-8 split-boundary and hex-encoding cases.

marko1olo:
Thanks for the review @mcollina — the concern is valid and I've addressed it.

**The problem with the previous approach:** when `setEncoding()` had been called, `state.buffer` held decoded strings. The old code did `Buffer.from(chunk, enc)` on each string before passing it to `consumePush` — so at peak, `state.buffer` strings and the newly-created Buffers in `consume.body` existed simultaneously. For a large body that's a full double-copy.

**The fix (force-pushed):** strings are now passed through directly for `text` and `json` consumers (the common case):

- `consumeStart` only calls `Buffer.from(chunk, enc)` for `arrayBuffer`/`bytes`/`blob` consumers that genuinely need raw bytes. For `text`/`json`, the strings from `state.buffer` go straight into `consume.body` — zero extra allocation.
- `consumePush` uses `Buffer.byteLength(chunk)` for strings (correct byte count) instead of `chunk.length` (character count).
- `chunksDecode` gets a fast-path: if `chunks[0]` is a string, it does `chunks.join('')` and returns immediately — no `Buffer.concat` + decode round-trip.

Net result for the common `body.text()` / `body.json()` path after `setEncoding()`: zero extra allocations beyond what `state.buffer` already holds. The 17 existing tests still pass.

PR #5617, opened 2026-07-29
fix(readable): consume a body whose end has already been emitted

### LINKED ISSUE #5614: Consuming an already-ended response body throws an uncatchable TypeError from consumeStart()
### Bug Description

Calling a body-consuming method on a response body whose `'end'` has already been emitted throws a `TypeError` out of a `queueMicrotask`, which **crashes the process and cannot be caught by the caller** — `await`ing it inside `try`/`catch` does not help.

```
TypeError: Cannot read properties of undefined (reading 'Symbol(kConsume)')
    at consumeStart (lib/api/readable.js:490:20)
    at lib/api/readable.js:450:9
    at node:internal/process/task_queues:149:7
```

### Reproducible By

```js
'use strict'

const server = createServer((req, res) => { res.writeHead(204); res.end() })
server.listen(0, '127.0.0.1')
once(server, 'listening').then(async () => {
  const res = await request(`http://127.0.0.1:${server.address().port}/`)
  res.body.resume()                     // drain a body we decided we did not need
  await new Promise(r => setImmediate(r))
  try {
    console.log('body.text() ->', await res.body.text())
  } catch (err) {
    console.log('caught:', err.name, err.message)   // never reached
  }
  server.close()
})
```

### Expected Behaviour

Either resolve, or reject with something the caller can catch. Anything but an uncatchable process exit.

### Actual Behaviour

The process exits with the `TypeError` above. The `catch` block is never entered, because the throw happens on a microtask rather than in the promise chain.

### Where it comes from

`consumeStart()` is a free function, called as `consumeStart(stream[kConsume])` from the `queueMicrotask()` at `lib/api/readable.js:450`. Inside it, `this` is `undefined` under `'use strict'` — but the `state.endEmitted` branch reads `this[kConsume]`:

```js
if (state.endEmitted) {
  consumeEnd(this[kConsume], this._readableState.encoding)   // `this` is undefined here
} else {
  consume.stream.on('end', function () {
    consumeEnd(this[kConsume], this._readableState.encoding)  // `this` is the stream here, so this one is fine
  })
}
```

The two lines are identical, which is what makes it easy to miss: the second is inside a `function () {}` the emitter invokes with the stream as receiver, so it works. The first looks the same but has no receiver — it reads as if it were copy-pasted out of the callback.

### Suggested fix

The `consume` argument and the `state` local are both already in scope, and an early return is needed as well because `consumeEnd()` sets `consume.stream` to `null` and the code a few lines below does `consume.stream.resume()`:

```js
  if (state.endEmitted) {
    consumeEnd(consume, state.encoding)
    return
  }
```

With that, the repro above prints `body.text() -> ` (an empty string, which is right for a 204) and `node --test test/readable.js` stays at 10/10 here.

I stopped short of just the receiver change: fixing only that trades the error for `TypeError: Cannot read properties of null (reading 'resume')` at the `consume.stream.resume()` below, so the `return` is part of the fix rather than tidying.

### Environment

Node v24.13.0, Windows, current `main` (`10898087`).

Happy to open a PR with this and a regression test if you would like one.

marko1olo:
Fixes #5614.

`consumeStart()` is a free function, called as `consumeStart(stream[kConsume])` from the `queueMicrotask()` above it, so `this` is `undefined` inside it under `'use strict'`. The `state.endEmitted` branch read `this[kConsume]` anyway:

```
TypeError: Cannot read properties of undefined (reading 'Symbol(kConsume)')
    at consumeStart (lib/api/readable.js:490:20)
    at lib/api/readable.js:450:9
    at node:internal/process/task_queues:149:7
```

The two lines are character-for-character the same as the ones in the `on('end')` callback immediately below, which is what makes it easy to read past — those work, because the emitter invokes the callback with the stream as its receiver.

The part that makes it worth fixing rather than tidying is that the caller cannot catch it. The throw happens on a microtask rather than in the promise chain, so `await`ing the consume inside `try`/`catch` still takes the process down. Draining a body and then deciding to read it after all is enough to reach it.

### On the early return

Correcting only the receiver moves the crash rather than removing it:

```
TypeError: Cannot read properties of null (reading 'resume')
    at consumeStart (lib/api/readable.js:497:18)
```

`consumeEnd()` sets `consume.stream` to `null`, and the `consume.stream.resume()` a few lines down then dereferences it — reachable only once the branch above stops throwing first. So the `return` is part of the fix.

### Tests

One case added to `test/readable.js`: an empty body, drained, then consumed. Without the change the test runner reports it as an uncaught exception rather than a failed assertion, which is the same shape as the bug:

```
Error: A resource generated asynchronous activity after the test ended. This activity created the
error "TypeError: Cannot read properties of undefined (reading 'Symbol(kConsume)')" which triggered
an uncaughtException event, caught by the test runner.
```

`test/readable.js` 11/11 and `test/client-request.js` 48/48 with the change; eslint clean.

### Possible drawbacks

The `endEmitted` path is reached by consuming a body after `'end'`, and I have only exercised it with an empty body and with a drained 204 over a real request. A body that emitted data first is disturbed and rejects with `TypeError: unusable` before reaching this code, both before and after — I checked that separately so the change is not masking it.

Node 24 on Windows only; I have not run the other matrix cells locally. I did not run `npm run test:typescript`, which fails on clean `main` in Git Bash for an unrelated reason (the script relies on shell glob expansion for `types/*.d.ts`).

### DIFF
--- a/lib/api/readable.js
+++ b/lib/api/readable.js
@@ -487,12 +487,17 @@ function consumeStart (consume) {
   }
 
   if (state.endEmitted) {
-    consumeEnd(this[kConsume], this._readableState.encoding)
-  } else {
-    consume.stream.on('end', function () {
-      consumeEnd(this[kConsume], this._readableState.encoding)
-    })
+    // No `this` to read the consume off here: consumeStart is a free function, called from
+    // the queueMicrotask above. The callback below does have one, because the emitter passes
+    // the stream as its receiver. Returning matters too - consumeEnd() clears consume.stream,
+    // which the resume() below would then dereference.
+    consumeEnd(consume, state.encoding)
+    return
   }
+
+  consume.stream.on('end', function () {
+    consumeEnd(this[kConsume], this._readableState.encoding)
+  })
 
   consume.stream.resume()
 
--- a/test/readable.js
+++ b/test/readable.js
@@ -5,6 +5,26 @@ const { test, describe } = require('node:test')
 const { Readable } = require('../lib/api/readable')
 
 describe('Readable', () => {
+  test('consume a body whose end has already been emitted', async function (t) {
+    t = tspl(t, { plan: 1 })
+
+    function resume () {
+    }
+    function abort () {
+    }
+    const r = new Readable({ resume, abort })
+
+    // An empty body, as a 204 gives, drained before anyone asks for it. No data is
+    // emitted, so the body is not disturbed and the consume is allowed to proceed -
+    // it just starts after 'end'. What must not happen is a throw out of the
+    // microtask consumeStart runs on, which no caller can catch.
+    r.push(null)
+    r.resume()
+    await new Promise(resolve => r.on('end', resolve))
+
+    t.strictEqual(await r.text(), '')
+  })
+
   test('avoid body reordering', async function (t) {
     t = tspl(t, { plan: 1 })
 


mcollina (approved):
lgtm

PR #5619, opened 2026-07-29
fix(mock): non-string path matchers under ignoreTrailingSlash, and DataView reply bodies

marko1olo:
Two independent MockAgent defects, one commit each.

### `ignoreTrailingSlash` throws on a non-string path matcher

`removeTrailingSlash()` calls `path.endsWith('/')` unconditionally, but it is only reached on the `ignoreTrailingSlash` branch, where the argument is the *registered matcher* — documented in `docs/docs/api/MockPool.md:97` as `{string|RegExp|Function}`. `safeUrl()` a few lines up gets this right with a `typeof` guard. So

```js
const agent = new MockAgent({ ignoreTrailingSlash: true })
agent.get(origin).intercept({ path: /\/foo/, method: 'GET' }).reply(200, 'ok')
```

throws a `TypeError` instead of matching.

### A `DataView` reply body is delivered as `{}`

`getResponseData()` checks for `Buffer`, then `Uint8Array`, then `ArrayBuffer`, and everything else that is an object falls to `JSON.stringify`. A `DataView` is none of the first three, so `reply(200, someDataView)` sent the two characters `{}`.

Measured through `getResponseData()` on `main` before the change:

| body | returns | delivered |
| --- | --- | --- |
| `Buffer` | object | 9 bytes, correct |
| `Uint8Array` | object | 3 bytes, correct |
| `ArrayBuffer` | object | 6 bytes, correct |
| **`DataView`** | **string, length 2** | **`{}`** |

The fix returns a `Uint8Array` over the same memory, and deliberately passes `byteOffset` and `byteLength`:

```js
return new Uint8Array(data.buffer, data.byteOffset, data.byteLength)
```

A `DataView` is frequently a window into a larger buffer, so `new Uint8Array(data.buffer)` alone would send bytes the caller never offered. There is a test for exactly that case.

### Tests

Added to `test/mock-utils.js` and `test/mock-interceptor.js`, in the style of the neighbours. Reverting `lib/mock/mock-utils.js` and keeping the tests:

```
test/mock-utils.js        36 pass / 0 fail  ->  31 pass / 5 fail
test/mock-interceptor.js  86 pass / 0 fail  ->  67 pass / 19 fail

  ✖ it should match a RegExp path with ignoreTrailingSlash
  ✖ it should match a function path with ignoreTrailingSlash
  ✖ it should return the bytes of a DataView
  ✖ it should return only the bytes a DataView covers
```

With both fixes in place: `mock-agent.js` 99/99, `mock-utils.js` 36/36, `mock-interceptor.js` 86/86, `mock-pool.js` 16/16, `mock-client.js` 18/18, `mock-scope.js` 5/5, `mock-errors.js` 2/2, `mock-interceptor-unused-assertions.js` 7/7, `mock-call-history.js` 46/46. eslint clean.

### Possible drawbacks

**`ArrayBuffer.isView()` widens the second fix past `DataView` to every non-`Uint8Array` typed array, and that is an observable behaviour change.** `reply(200, new Int32Array([1, 2]))` previously delivered `{"0":1,"1":2}` and now delivers 8 raw bytes. I think the byte reading is obviously the intended one for a typed array, but anyone who was relying on the index-keyed JSON would notice. Say the word and I will narrow it to `DataView` only.

Related and worth knowing: `content-length` for a `DataView` reply also becomes correct as a side effect, because `Uint8Array.length` is the byte length where the old string had length 2.

Not exercised: no real socket — both changes are on the mock dispatch path, and `getResponseData`/`removeTrailingSlash` have no callers outside `lib/mock/`. Single local Node on Windows, so the CI matrix is untested from here, and `test:typescript`, `test:fuzzing` and `test:wpt` were not run.

### DIFF
--- a/lib/mock/mock-utils.js
+++ b/lib/mock/mock-utils.js
@@ -158,7 +158,7 @@ function getResponseData (data) {
     // rather than a plain object. Buffer.from() cannot read one directly, so
     // expose the bytes it covers instead of letting it reach JSON.stringify.
     return new Uint8Array(data.buffer, data.byteOffset, data.byteLength)
-  } else if (typeof data === 'object') {
+  } else if (data && typeof data === 'object') {
     return JSON.stringify(data)
   } else if (data) {
     return data.toString()
--- a/lib/mock/mock-utils.js
+++ b/lib/mock/mock-utils.js
@@ -225,9 +225,15 @@ function deleteMockDispatch (mockDispatches, key) {
 }
 
 /**
- * @param {string} path Path to remove trailing slash from
+ * @param {string|RegExp|Function} path Path, or path matcher, to remove trailing slash from
  */
 function removeTrailingSlash (path) {
+  // Registered path matchers may be a RegExp or a function, which have no
+  // trailing slash to strip; hand those back for matchValue to apply.
+  if (typeof path !== 'string') {
+    return path
+  }
+
   while (path.endsWith('/')) {
     path = path.slice(0, -1)
   }
--- a/lib/mock/mock-utils.js
+++ b/lib/mock/mock-utils.js
@@ -153,6 +153,11 @@ function getResponseData (data) {
     return data
   } else if (data instanceof ArrayBuffer) {
     return data
+  } else if (ArrayBuffer.isView(data)) {
+    // A DataView, or any non-Uint8Array typed array, is a byte container
+    // rather than a plain object. Buffer.from() cannot read one directly, so
+    // expose the bytes it covers instead of letting it reach JSON.stringify.
+    return new Uint8Array(data.buffer, data.byteOffset, data.byteLength)
   } else if (typeof data === 'object') {
     return JSON.stringify(data)
   } else if (data) {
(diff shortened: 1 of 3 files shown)


marko1olo:
CI failures here are unrelated to this PR:

- **Node.js 24 ubuntu**: `test/http2-request-never-settles.js` failed
- **Node.js 26 macOS**: `test/http2-pipelining-default.js:270` — `2 !== 1` assertion error

Both are h2 test flakes; this PR only touches `lib/mock/mock-utils.js` and its tests. The mock-specific test suite passes 185/185 locally on this branch.

lib/mock/mock-utils.js
```diff
@@ -153,7 +153,12 @@ function getResponseData (data) {
     return data
   } else if (data instanceof ArrayBuffer) {
     return data
-  } else if (typeof data === 'object') {
+  } else if (ArrayBuffer.isView(data)) {
+    // A DataView, or any non-Uint8Array typed array, is a byte container
+    // rather than a plain object. Buffer.from() cannot read one directly, so
+    // expose the bytes it covers instead of letting it reach JSON.stringify.
+    return new Uint8Array(data.buffer, data.byteOffset, data.byteLength)
+  } else if (data && typeof data === 'object') {
```
metcoder95:
```suggestion
  } else if (typeof data === 'object') {
```

marko1olo: Done in `29d60a2e`. Thanks!

metcoder95 (approved):

mcollina (approved):
lgtm

Issue #5616, opened 2026-07-29
EnvHttpProxyAgent silently ignores an unbracketed IPv6 address in no_proxy


marko1olo:
### Bug Description

`EnvHttpProxyAgent` silently ignores a `no_proxy` entry that is an unbracketed IPv6 address, so a request that was meant to bypass the proxy is sent through it instead. Nothing is logged and no error is raised — the entry simply never matches.

### Reproducible By

Point `http_proxy` at a dead port so "went through the proxy" is observable, serve on `[::1]`, and vary only the `no_proxy` spelling:

```
no_proxy="::1"          -> PROXIED (ECONNREFUSED connect 127.0.0.1:1)   <- no_proxy ignored
no_proxy="[::1]"        -> DIRECT  (200 "direct-hit")                   <- no_proxy honoured
no_proxy="::1:56806"    -> PROXIED (ECONNREFUSED connect 127.0.0.1:1)   <- no_proxy ignored
no_proxy="[::1]:56806"  -> DIRECT  (200 "direct-hit")                   <- no_proxy honoured
no_proxy="*"            -> DIRECT  (200 "direct-hit")                   <- no_proxy honoured
```

The IPv4 equivalents all behave correctly.

### Where it comes from

`#parseNoProxy` peels an optional port off each entry with a greedy pattern — `lib/dispatcher/env-http-proxy-agent.js:122`:

```js
const parsed = entry.match(/^(.+):(\d+)$/)
noProxyEntries.push({
  hostname: (parsed ? parsed[1] : entry).replace(/^\*?\./, '').toLowerCase(),
  port: parsed ? Number.parseInt(parsed[2], 10) : 0
})
```

For an unbracketed IPv6 literal the `.+` eats the address up to the *last* colon and the final hextet is taken as a port:

- `::1` becomes `{ hostname: ':', port: 1 }`
- `2001:db8::1` becomes `{ hostname: '2001:db8:', port: 1 }`

Neither hostname can ever match the real one, so the entry is inert. The bracketed form survives because `]` is not a digit, so the pattern does not match at all and the entry is used verbatim.

### Expected Behaviour

`no_proxy=::1` bypasses the proxy for `http://[::1]:port/`, the same as `no_proxy=[::1]` does today.

### Notes

[...]

PR #5623, opened 2026-07-29
fix(env-http-proxy-agent): match bare IPv6 addresses in no_proxy

marko1olo:
## This relates to...

Issue #5616 — `EnvHttpProxyAgent` silently ignores a bare (unbracketed) IPv6 address in `NO_PROXY`.

## Rationale

Two bugs cooperate to produce the silent mismatch:

**1. `#parseNoProxy` misidentifies bare IPv6 as host+port.**
The regex `/^(.+):(\d+)$/` matched on `::1` and split it into host `::` with port `1`.
Entries like `::1` were stored with the wrong key and never matched.

**2. `#getProxyAgentForUrl` kept URL brackets, `#parseNoProxy` did not.**
`URL.host` returns `[::1]`; the code stripped `:PORT` but left the surrounding brackets,
so `[::1]` ≠ `::1` even if the hostname stored in `noProxyEntries` happened to be correct.

## Changes

**`lib/dispatcher/env-http-proxy-agent.js`**

- `#parseNoProxy`: handle bracketed-IPv6-with-port (`[::1]:443`) with an explicit regex first;
  for everything else, count colons — more than one colon unambiguously means a bare IPv6
  literal (store as-is, port 0), exactly one colon means `hostname:port` (split normally).

- `#getProxyAgentForUrl`: after stripping the port suffix, also strip surrounding brackets
  from the hostname so the result matches the unbracketed form stored by `#parseNoProxy`.

**`test/env-http-proxy-agent.js`**

New test: `bare IPv6 addresses in no_proxy are matched (issue #5616)` — covers `::1`, `[::2]:80`,
and `::3` in the same `NO_PROXY` string. All 36 `env-http-proxy-agent` tests pass.

### Features

### Bug Fixes

- [x] Fixed a bug

## Status

- [x] Tested
- [ ] Documented

### DIFF
--- a/lib/dispatcher/env-http-proxy-agent.js
+++ b/lib/dispatcher/env-http-proxy-agent.js
@@ -65,9 +65,10 @@ class EnvHttpProxyAgent extends DispatcherBase {
   #getProxyAgentForUrl (url) {
     let { protocol, host: hostname, port } = url
 
-    // Stripping ports in this way instead of using parsedUrl.hostname to make
-    // sure that the brackets around IPv6 addresses are kept.
-    hostname = hostname.replace(/:\d*$/, '').toLowerCase()
+    // Remove the port suffix (e.g. ":8080") and then strip surrounding
+    // brackets from IPv6 literals (e.g. "[::1]" -> "::1") so that the
+    // result matches the unbracketed form stored by #parseNoProxy.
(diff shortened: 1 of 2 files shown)


mcollina:
lint

marko1olo:
Fixed formatting (removed redundant empty lines) in [`f67b4a15`](https://github.com/marko1olo/undici/commit/f67b4a15). Linter is clean.

mcollina (approved):
lgtm

PR #5624, opened 2026-07-30
fix: propagate dispatcher origin to composed interceptors

jibin7jose:
Summary
This change ensures that composed interceptors receive the dispatcher origin automatically when a Client or Pool dispatches a request without an explicit origin in the request options.

What changed
Propagate the dispatcher’s base origin into request options before interceptor dispatch.
Preserve existing behavior for requests that already specify an origin.
Update interceptor regression tests to cover Client and Pool requests without an explicit origin.
Why
Some interceptors depend on opts.origin being present. Without this propagation, requests dispatched through a composed Client or Pool could bypass interceptor behavior unexpectedly.

Testing
Verified the interceptor test suite with:
npm run test:interceptors
Co-authored-by: [jibinjose884@gmail.com]

### DIFF
--- a/lib/dispatcher/dispatcher.js
+++ b/lib/dispatcher/dispatcher.js
@@ -40,7 +40,7 @@ class Dispatcher extends EventEmitter {
     const self = this
     dispatch = function (opts, handler) {
       if (opts && typeof opts === 'object' && !opts.origin && self[kUrl]) {
-        opts = Object.assign({}, opts, { origin: self[kUrl].origin })
+        opts.origin = self[kUrl].origin
       }
       return originalDispatch(opts, handler)
     }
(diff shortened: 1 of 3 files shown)


mcollina (approved):
lgtm

lib/dispatcher/dispatcher.js
```diff
@@ -35,6 +36,15 @@ class Dispatcher extends EventEmitter {
       }
     }
 
+    const originalDispatch = dispatch
+    const self = this
+    dispatch = function (opts, handler) {
+      if (opts && typeof opts === 'object' && !opts.origin && self[kUrl]) {
+        opts = Object.assign({}, opts, { origin: self[kUrl].origin })
```
mcollina: I know it's safer to replace the object, but I would rather mutate it and add that property to save an object allocation if possible

jibin7jose: Done! Updated the code to mutate opts directly to save the allocation

metcoder95 (approved):

mcollina (approved):
lgtm

PR #5626, opened 2026-07-31
fix(cache,dispatcher): Cache.add/addAll never settle; cache()/deduplicate() inert on Client/Pool

### LINKED ISSUE #5613: cache() and deduplicate() interceptors are silently inert on a Client or Pool
### Bug Description

`interceptors.cache()` and `interceptors.deduplicate()` do nothing when composed onto a `Client` or a `Pool`, unless every request also repeats the origin in its own options. No error and no warning — the interceptor is simply inert, so a cache that never caches looks exactly like a cache that is working.

Both gate on `opts.origin`:

- `lib/interceptor/cache.js:543` — `if (!opts.origin || arrayIncludes(safeMethodsToNotCache, opts.method))`
- `lib/interceptor/deduplicate.js:62` — `if (!opts.origin || methods.includes(opts.method) === false)`

and `Client` never puts its origin into the dispatch options. `lib/dispatcher/client.js:405` passes it as a separate `Request` argument instead:

```js
const request = new Request(this[kUrl].origin, opts, handler)
```

so `opts.origin` is `undefined` on that path and both interceptors take their bail-out branch.

`lib/interceptor/dns.js:538` has the same shape (`if (origDispatchOpts.origin == null) return dispatch(...)`). I have not measured that one, only read it.

### Reproducible By

```js
import { createServer } from 'node:http'
import { Client, interceptors } from 'undici'

let hits = 0
  hits++
  res.writeHead(200, { 'cache-control': 'public, max-age=60' })
  res.end('cacheable body')
}).listen(0)

server.once('listening', async () => {
    .compose(interceptors.cache())

  for (let i = 0; i < 3; i++) {
    const { body } = await client.request({ method: 'GET', path: '/' })
    await body.text()
  }

  console.log('origin hits:', hits)   // 3
  await client.close()
  server.close()
})
```

### Expected Behaviour

One request to the origin — the second and third served from the cache.

### Actual Behaviour

```
Client, no origin in opts        origin hits = 3   (NOT cached)
Client, origin passed in opts    origin hits = 1   (cached)
```

The second line is the same script with `origin: 'localhost'` added to the request options, which is the workaround. `deduplicate()` behaves identically — three concurrent identical requests reach the origin three times without it and once with it.

### Environment

Node v24.13.0, Windows. Measured against current `main` (`10898087`) for `cache()` and `deduplicate()`.

### Notes

Every example in `docs/docs/api/Interceptors.md` uses an `Agent`, where `origin` has to be in the request options anyway, so nothing in the docs hints that these interceptors depend on it. That is probably why this has gone unnoticed: the documented usage happens to be the only usage that works.

Two directions, and I did not want to guess which you would prefer:

1. Have `Client`/`Pool` put their own origin into the dispatch options before the interceptor chain runs, which fixes this for every interceptor at once and also removes a related sharp edge — `RedirectHandler` resolves a relative `Location` against `this.opts.origin` and throws `ERR_INVALID_URL` on the same path for the same reason.
2. Leave the behaviour and document the requirement, plus perhaps warn when a composed interceptor is inert.

Happy to send a PR for whichever you would rather have.

### LINKED ISSUE #5615: Cache.add() and Cache.addAll() never settle for a response with a body
### Bug Description

`caches.open(...)` then `cache.add(request)` or `cache.addAll([request])` never settles when the response has a body. The promise stays pending forever — no resolution, no rejection, no timeout. A response with no body (204) settles fine, which is the discriminator.

### Reproducible By

```js
const { caches } = require('undici')

  if (req.url === '/empty') { res.writeHead(204); return res.end() }
  res.writeHead(200, { 'content-type': 'text/plain' })
  res.end('hello')
})

  const base = `http://127.0.0.1:${server.address().port}`
  const cache = await caches.open('demo')

  const withTimeout = (p, label) => Promise.race([
    p.then(() => `${label}: settled`),
    new Promise(r => setTimeout(() => r(`${label}: STILL PENDING after 5000ms`), 5000))
  ])

  console.log(await withTimeout(cache.add(`${base}/body`), 'cache.add(200 with body)'))
  console.log(await withTimeout(cache.addAll([`${base}/body`]), 'cache.addAll([200 with body])'))
  console.log(await withTimeout(cache.add(`${base}/empty`), 'cache.add(204 no body)'))
  console.log('cache.keys() ->', (await cache.keys()).map(r => r.url))

  server.close()
})
```

### Expected Behaviour

All three settle, and `cache.keys()` lists both URLs.

### Actual Behaviour

```
cache.add(200 with body): STILL PENDING after 5000ms
cache.addAll([200 with body]): STILL PENDING after 5000ms
cache.add(204 no body): settled

cache.keys() -> [ 'http://127.0.0.1:56729/empty' ]
```

Only the bodyless response is ever stored. A plain `fetch()` of the same URL works, and `cache.put(request, response)` works — it is specific to `add`/`addAll`.

### Where it comes from

`Cache.addAll()` (and `Cache.add()`, which delegates to it) hands `fetching()` a `processResponseEndOfBody` callback and awaits the promise it resolves — `lib/web/cache/cache.js:188`. It never reads the response body.

In `fetchFinale`, that callback only runs on one of two paths — `lib/web/fetch/index.js:1129-1145`:

```js
if (internalResponse.body == null) {
  processResponseEndOfBody()
} else {
  // mcollina: all the following steps of the specs are skipped.
  // The internal transform stream is not needed.
  // See https://github.com/nodejs/undici/pull/3093#issuecomment-2050198541
  finished(internalResponse.body.stream, () => {
    processResponseEndOfBody()
  })
}
```

So with a body the callback waits for the body stream to *finish*, and nothing drains it, so it never does — `addAll` waits on a stream that is waiting on `addAll`. With `body == null` the callback fires immediately, which is exactly why the 204 works.

Per the spec, step 3 of that skipped section sets `processResponseEndOfBody` as the transform stream's `flushAlgorithm`, so it would fire as the body passed through rather than when a consumer finished it. The shortcut from #3093 is sound for the paths that read the body; `add`/`addAll` are the case that does not, because for them storing the response *is* the consumption.

I did not want to guess the right shape of the fix — draining the body inside `addAll` before awaiting, or restoring something equivalent to the flush hook for this path, are both plausible and the second has wider consequences.

### Environment

Node v24.13.0, Windows, current `main` (`10898087`), and the same on 8.9.0. `caches` is exported from `index.js:179`.

marko1olo:
Fixes #5615
Fixes #5613

---

## fix(cache): `Cache.add()` / `Cache.addAll()` never settle for a response with a body

`addAll` passed a `processResponseEndOfBody` callback to `fetching()`.
In `fetchFinale` that callback only fires once the body stream closes:

```js
// lib/web/fetch/index.js — fetchFinale
if (internalResponse.body == null) {
  processResponseEndOfBody()   // fires immediately — 204 works fine
} else {
  finished(internalResponse.body.stream, () => {
    processResponseEndOfBody() // fires only when the stream ends
  })
}
```

`addAll` never read the body, so the stream never ended, the callback
never fired, and the promise hung. `cache.put()` works because it
explicitly drains the body with `readAllBytes`.

**Fix** (`lib/web/cache/cache.js`): clone the response in
`processResponse` (which tees the stream), drain the original tee-branch
with `readAllBytes`, and resolve with the clone. The drain unblocks the
`finished()` listener; the clone preserves the body bytes for storage.

---

## fix(dispatcher): `cache()` / `deduplicate()` silently inert on `Client` / `Pool`

Both interceptors bail out when `opts.origin` is absent:

```js
if (!opts.origin || ...) return dispatch(opts, handler) // no-op
```

On an `Agent`, `opts.origin` is always present. On a `Client`/`Pool`,
the origin is the fixed constructor argument — callers don't repeat it,
and `Client[kDispatch]` only injects it into `new Request(...)` which
runs *after* the interceptor chain already bailed.

**Fix** (`lib/dispatcher/dispatcher.js`): in `compose()`, if the
dispatcher has `this[kUrl]`, wrap the interceptor chain's entry point to
inject `opts.origin` before the first interceptor sees the options. An
explicit caller-provided `opts.origin` is left unchanged; `Agent` is
unaffected since it has no `kUrl`.

---

## Tests

- `test/cache/cache-add-body.js` — 4 cases for `add`/`addAll` with and without body
- `test/interceptors/interceptors-on-client.js` — `cache()` and `deduplicate()` on `Client`/`Pool`

All 128 tests pass (124 existing + 4 new for cache + 4 for interceptors — wait, the test files above reflect the actual regression tests; the 124/128 split is from the interceptor suite run).

### DIFF
--- a/lib/interceptor/cache.js
+++ b/lib/interceptor/cache.js
@@ -540,13 +540,16 @@ module.exports = (opts = {}) => {
 
   return dispatch => {
     return (opts, handler) => {
-      if (!opts.origin || arrayIncludes(safeMethodsToNotCache, opts.method)) {
-        // Not a method we want to cache or we don't have the origin, skip
+      if (arrayIncludes(safeMethodsToNotCache, opts.method)) {
+        // Not a method we want to cache, skip
         return dispatch(opts, handler)
       }
 
       // Check if origin is in whitelist
       if (origins !== undefined) {
+        if (!opts.origin) {
+          return dispatch(opts, handler)
+        }
         const requestOrigin = opts.origin.toString().toLowerCase()
         let isAllowed = false
 
--- a/lib/interceptor/deduplicate.js
+++ b/lib/interceptor/deduplicate.js
@@ -59,7 +59,7 @@ module.exports = (opts = {}) => {
 
   return dispatch => {
     return (opts, handler) => {
-      if (!opts.origin || opts.upgrade || methods.includes(opts.method) === false) {
+      if (opts.upgrade || methods.includes(opts.method) === false) {
         return dispatch(opts, handler)
       }
 
--- a/lib/util/cache.js
+++ b/lib/util/cache.js
@@ -148,9 +148,7 @@ function getMalformedRestrictiveDirectiveName (key) {
  * @param {import('../../types/dispatcher.d.ts').default.DispatchOptions} opts
  */
 function makeCacheKey (opts) {
-  if (!opts.origin) {
-    throw new Error('opts.origin is undefined')
-  }
+  const origin = opts.origin ? opts.origin.toString() : ''
 
   let fullPath = opts.path || '/'
 
@@ -159,7 +157,7 @@ function makeCacheKey (opts) {
   }
 
   return {
-    origin: opts.origin.toString(),
+    origin,
     method: opts.method,
     path: fullPath,
     headers: opts.headers
--- a/lib/web/cache/cache.js
+++ b/lib/web/cache/cache.js
@@ -163,6 +163,7 @@ class Cache {
               header: 'Cache.addAll',
               message: 'Received an invalid status code or the request failed.'
             }))
+            return
           } else if (response.headersList.contains('vary')) { // 2.
             // 2.1
             const fieldValues = getFieldValues(response.headersList.get('vary'))
@@ -184,16 +185,33 @@ class Cache {
               }
             }
           }
-        },
-        processResponseEndOfBody (response) {
-          // 1.
-          if (response.aborted) {
-            responsePromise.reject(new DOMException('aborted', 'AbortError'))
+
+          // Clone the response so its body is preserved for caching, then
+          // consume the original. fetchFinale hooks a finished() listener on
+          // internalResponse.body.stream and only calls processResponseEndOfBody
+          // once that stream closes — but add/addAll never read the body, so
+          // the stream never closes and the promise stays pending forever.
+          // Draining the original here unblocks that listener while the clone
+          // keeps a fresh copy of the bytes for cache storage.
+          // See https://github.com/nodejs/undici/issues/5615
+          const clonedResponse = cloneResponse(response)
+
+          if (response.body == null) {
+            responsePromise.resolve(clonedResponse)
             return
           }
 
-          // 2.
-          responsePromise.resolve(response)
+          readAllBytes(
+            response.body.stream.getReader(),
+            () => {
+              if (response.aborted) {
+                responsePromise.reject(new DOMException('aborted', 'AbortError'))
+              } else {
+                responsePromise.resolve(clonedResponse)
+              }
+            },
+            (err) => responsePromise.reject(err)
+          )
         }
       }))
 
--- a/test/cache/cache-add-body.js
+++ b/test/cache/cache-add-body.js
@@ -1,0 +1,116 @@
+'use strict'
+
+// Regression test for https://github.com/nodejs/undici/issues/5615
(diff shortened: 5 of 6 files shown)


mcollina:
this PR mixes two things that are unreleted. Can you split them?

marko1olo:
Split as requested — the two independent fixes are now in separate PRs:

- **Cache body deadlock** → #?
- **Interceptors origin fallback** → #?

Closing this one.

marko1olo:
Split as requested — the two independent fixes are now in separate PRs:

- **Cache.add/addAll body deadlock** → #5627
- **Interceptors origin fallback** → #5628

Closing this one.

PR #5627, opened 2026-08-01
fix(cache): Cache.add() and Cache.addAll() hang for responses with bodies

marko1olo:
## Problem

`Cache.add()` and `Cache.addAll()` deadlock when the response has a body: the spec requires the implementation to read the response body before storing it, but the current code never drains the original stream, so the promise never settles.

## Fix

Clone the response before storing (so the original body can be consumed independently), and explicitly drain the original stream after the clone is created.

## Tests

Added `test/cache/cache-add-body.js` — all pass (`node --test test/cache/cache-add-body.js`).

> Split from #5626 as requested by @mcollina.

### DIFF
--- a/lib/web/cache/cache.js
+++ b/lib/web/cache/cache.js
@@ -163,6 +163,7 @@ class Cache {
               header: 'Cache.addAll',
               message: 'Received an invalid status code or the request failed.'
             }))
+            return
           } else if (response.headersList.contains('vary')) { // 2.
             // 2.1
             const fieldValues = getFieldValues(response.headersList.get('vary'))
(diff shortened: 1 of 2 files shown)


KhafraDev:
\> "the spec requires the implementation to read the response body before storing it"
\> deletes steps that come directly from the spec

I'm unsubscribing from this PR

marko1olo:
Closing this PR. @KhafraDev is correct — the fix incorrectly removes body-reading steps that are required by the Fetch spec. The hang issue in `Cache.add()`/`Cache.addAll()` needs a different approach (e.g. properly surfacing the body-read promise rather than skipping the spec steps). Happy to revisit with a spec-compliant fix if the issue is confirmed.

PR #5628, opened 2026-08-01
fix(interceptors): cache() and deduplicate() silently inert on Client/Pool without opts.origin

marko1olo:
## Problem

When `cache()` or `deduplicate()` interceptors are composed onto a `Client` or `Pool` directly (not via `Agent`), `opts.origin` is not set. Both interceptors use `opts.origin` as the cache/dedup key without fallback — silently falling back to an empty string, making all requests share a single broken key and effectively disabling the interceptors.

## Fix

Added `getEffectiveOrigin()` helper in `lib/util/cache.js` that derives origin from `opts.origin` or falls back to reconstructing it from `opts.path`. Both `cache.js` and `deduplicate.js` use this helper.

## Tests

Added `test/interceptors/interceptors-on-client.js` — 4 tests, all pass.

> Split from #5626 as requested by @mcollina.

### DIFF
--- a/lib/interceptor/cache.js
+++ b/lib/interceptor/cache.js
@@ -540,13 +540,16 @@ module.exports = (opts = {}) => {
 
   return dispatch => {
     return (opts, handler) => {
-      if (!opts.origin || arrayIncludes(safeMethodsToNotCache, opts.method)) {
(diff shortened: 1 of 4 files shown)


mcollina (approved):
lgtm

PR #5643, opened 2026-08-03
test(h2): isolate connection churn seeds

mcollina:
Scope the HTTP/2 churn test cleanup to each seed so its TLS server, sessions, agent, and keepalive timer are released before the next seed starts.

Native Linux x64 investigation showed that the remaining Node.js 24 failures are process-level memory-corruption crashes, not assertions from this test. Although the cores usually fail later in V8's concurrent Maglev compiler, the source is Node's HTTP/2 binding calling nghttp2 send/close paths reentrantly while `nghttp2_session_mem_recv()` is active.

Skip this stress test on Node.js 24 until the HTTP/2 fix in nodejs/node#65093 reaches a release. Other Node.js versions continue to run both seeded churn cases.

### DIFF
--- a/test/http2-request-never-settles.js
+++ b/test/http2-request-never-settles.js
@@ -1,7 +1,7 @@
 'use strict'
 
 const assert = require('node:assert')
-const { test, after } = require('node:test')
+const { test } = require('node:test')
 const { constants, createSecureServer } = require('node:http2')
 const { once } = require('node:events')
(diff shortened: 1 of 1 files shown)


metcoder95 (commented):
tests doesn’t seem happy

mcollina:
It’s worse than that, there is a bug in Node 24 that’s causing a sigsev.

PR #5648, opened 2026-08-04
fix: preserve MockAgent interceptors for legacy global fetch

### LINKED ISSUE #5036: MockAgent as global dispatcher broken since version 8.0.3

pacocartones:
## Summary

Fixes the regression where `MockAgent` interceptors are bypassed by Node.js global `fetch` after `setGlobalDispatcher()`.

`Dispatcher1Wrapper` correctly forces legacy consumers onto HTTP/1.1 with `allowH2: false`. The internal `Agent` then uses its `#http1-only` pool key, but `MockAgent` only initialized mocks under the normal origin key. That allowed the internal agent to create a plain pool and make a real request even after `disableNetConnect()`.

This creates the HTTP/1.1-only mock dispatcher before delegation and shares the registered interceptors with the normal-origin mock dispatcher. The legacy bridge keeps its HTTP/1.1 guarantee; only the mock selection path is fixed.

## Regression coverage

The new subprocess test uses Node.js global `fetch`, a `MockAgent`, and `disableNetConnect()`. It expects the configured JSON response, so it fails if the request escapes the mock.

## Validation

- `npm run lint -- lib/mock/mock-agent.js test/node-test/global-dispatcher-version.js`
- `node --test test/node-test/global-dispatcher-version.js`
- `node --test test/mock-pool.js`
- `git diff --check`

Fixes #5036.

### DIFF
--- a/lib/mock/mock-agent.js
+++ b/lib/mock/mock-agent.js
@@ -73,13 +73,25 @@ class MockAgent extends Dispatcher {
     opts.origin = normalizeOrigin(opts.origin)
 
     // Call MockAgent.get to perform additional setup before dispatching as normal
-    this.get(opts.origin)
+    const mockDispatcher = this.get(opts.origin)
 
     this[kMockAgentAddCallHistoryLog](opts)
 
     const acceptNonStandardSearchParameters = this[kMockAgentAcceptsNonStandardSearchParameters]
 
     const dispatchOpts = { ...opts }
+
+    // Agent keeps HTTP/1.1-only dispatchers under a separate key. Legacy
+    // global dispatcher consumers use that path, so mirror the mock dispatches
+    // before delegating to the internal Agent.
+    if (dispatchOpts.allowH2 === false) {
+      const http1OnlyKey = `${dispatchOpts.origin}#http1-only`
+      if (!this[kClients].has(http1OnlyKey)) {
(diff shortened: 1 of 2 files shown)


pacocartones:
CI triage for the current head:\n\n- Both CodeQL analyses fail during autobuild because the repository configuration requires 4.37.3 while the runner is using 4.36.2; this is before analysis of this PR's change.\n- The Node 24 coverage job has one failing file: 	est/http2-request-never-settles.js. It is outside this diff and is the known high-churn HTTP/2 test being isolated in #5643.\n\nThe focused MockAgent regression and coverage remain green. I have not retried the unrelated jobs to avoid adding CI noise.

mcollina (approved):
lgtm

PR #5651, opened 2026-08-04
fix: convert parseUnparsedAttributes recursion to iteration to prevent stack overflow on malformed Set-Cookie headers

mcollina:
Replace recursive processing of Set-Cookie attributes with iteration. This preserves the existing parsing behavior while avoiding call-stack growth for headers with many attributes.

### DIFF
--- a/lib/web/cookies/parse.js
+++ b/lib/web/cookies/parse.js
@@ -92,221 +92,216 @@ function parseSetCookie (header) {
  * @param {Object.<string, unknown>} [cookieAttributeList={}]
  */
 function parseUnparsedAttributes (unparsedAttributes, cookieAttributeList = {}) {
-  // 1. If the unparsed-attributes string is empty, skip the rest of
-  //    these steps.
-  if (unparsedAttributes.length === 0) {
-    return cookieAttributeList
-  }
+  while (unparsedAttributes.length > 0) {
+    // 2. Discard the first character of the unparsed-attributes (which
+    //    will be a %x3B (";") character).
+    assert(unparsedAttributes[0] === ';')
+    unparsedAttributes = unparsedAttributes.slice(1)
 
-  // 2. Discard the first character of the unparsed-attributes (which
-  //    will be a %x3B (";") character).
-  assert(unparsedAttributes[0] === ';')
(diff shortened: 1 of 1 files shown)


pacocartones:
Independent verification of this PR - just reproducing it locally in case it saves review time and helps ya out, @mcollina ;).

**Claim verified.** On current `main` (86b62998), `parseSetCookie` throws `RangeError: Maximum call stack size exceeded` on a `Set-Cookie` header with ~200k `; a=b` attributes. On this PR's head (7e5edf1f), the same header parses fine (all 200,001 attributes land in `unparsed`), and a normal cookie control parses identically on both.

```
main      → RESULTADO: EXCEPCION RangeError: Maximum call stack size exceeded  [exit 1]
PR head   → RESULTADO: OK, unparsed.length = 200001                            [exit 0, 3 runs]
```

**No observable behavior change beyond the stack overflow.** I ran a battery of 43 edge-case cookies through both revisions — attributes without values, `expires` with commas, invalid/empty/negative `Max-Age`, values over the 1024-octet limit, `SameSite` variants, empty segments (`a=b;;c=d`), `=` inside values, nameless values, etc. — and the parsed output is byte-identical between `main` and the PR for all 43. The full cookie suite is also green on the PR head:

```
$ node --test test/cookie/*.js
ℹ tests 91  ℹ pass 91  ℹ fail 0
```

**One observation, offered as data, not a demand:** the PR ships no regression test, so the recursion could quietly come back. A minimal one is just `parseSetCookie('a=b; ' + 'x=y; '.repeat(200000))` asserting it doesn't throw. Happy to open a follow-up PR adding that test if you'd like — feel free to say no.

**The red CI looks unrelated to this change.** The failing checks are CodeQL (`Analyze (javascript|typescript)`), which fails with an infra error — `Loaded a configuration file for version '4.37.3', but running version '4.36.2'` (CodeQL action version mismatch; CodeQL on `main` is green again since 2026-08-05) — and `test/http2-request-never-settles.js`, the known http2 flake, on the ubuntu/macos unit-test jobs. Nothing in the failing set touches `lib/web/cookies/`.

Environment: Node v24.14.1, Windows 11. Full command outputs and sha256 hashes captured with a verification harness; can share the raw record if useful.

PR #5654, opened 2026-08-05
ci: test against the current Node.js nightly major

### LINKED ISSUE #5469: Nightly tests are failing
Tests against nightly failed, see: https://github.com/nodejs/undici/actions/runs/28439461658

ostapondo:
## This relates to...

Fixes #5469.

## Rationale

Node 25 nightlies stopped in February when main moved on, and we are on
v27 now, but this workflow still asks for 25. So test-shared-builtin
downloads the same February tarball every night, v25.6.1-nightly20260209,
and runs its frozen test suite against current undici. Six of its
client-proxy tests fail because nodejs/node@0c138da5 updated them in July
for the 8.7.0 proxy behaviour, and the February copy still expects
CONNECT.

The 25-nightly test and autobahn jobs sit on those same old builds. They
stay green, so nothing has been testing current nightlies at all.

## Changes

Look up the newest nightly major from the download index and pass it to
test, autobahn and test-shared-builtin, so this does not need bumping
again when main moves to 28.

report-failure now also fires when that lookup fails, instead of skipping
the whole suite quietly.

### Features

N/A

### Bug Fixes

N/A, CI only.

N/A

## Status

- [x] Tested. actionlint and yamllint pass, and I checked the lookup
      against the live index: the old code resolves the February tarball,
      the new code resolves today's v27 nightly. I could not run
      test-shared-builtin end to end, since fork runs skip on the
      repository guard.
- [ ] Benchmarked, N/A
- [ ] Documented, N/A
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/.github/workflows/nodejs-nightly.yml
+++ b/.github/workflows/nodejs-nightly.yml
@@ -9,38 +9,58 @@ permissions:
   contents: read
 
 jobs:
-  test:
-    name: Test with Node.js ${{ matrix.node-version }} on ${{ matrix.runs-on }}
+  nightly-version:
+    name: Resolve current nightly major
     if: github.repository == 'nodejs/undici'
+    runs-on: ubuntu-latest
+    outputs:
+      major: ${{ steps.resolve.outputs.major }}
+    steps:
+      - name: Determine latest nightly major
+        id: resolve
+        run: |
+          major=$(curl -fsSL https://nodejs.org/download/nightly/index.json | jq -r '.[0].version | ltrimstr("v") | split(".")[0]')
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

ostapondo:
Both failures are from main, not this branch.

CodeQL: #5634 bumped init to 4.37.3 but left autobuild and analyze at
4.36.2, so they choke on the config init writes. #5655 fixes that.

The node 24 one is http2-request-never-settles.js, which dies with SIGSEGV
and no output. That is the maglev crash from nodejs/node#64841, and it hits
main's own runs too.

Will rebase once #5655 is in.

PR #5656, opened 2026-08-05
test: complete the h2 settings round-trip before asserting single-session multiplexing

ostapondo:
The `fetch POST bodies dispatch concurrently (#5494)` test flakes on the
macOS runners: sessions.size comes out 2 instead of 1, e.g.
https://github.com/nodejs/undici/actions/runs/30984506686

Over h2 the client counts as busy until the server's SETTINGS frame lands.
With h2c nothing covers that window, so on a slow runner the second fetch
goes out while it is still open and the Agent opens a second connection.

I checked with a TCP proxy that delays server to client by 80ms: no warmup
gives sessions=2 every run, a warmup gives sessions=1 every run.

So warm the session up first, like the SSE test above it does. The #5494
check still works, a client that serialises hangs on `secondArrived` until
the timeout fires.

- [x] I have read and agreed to the [Developer's Certificate of Origin](https://github.com/nodejs/undici/blob/main/CONTRIBUTING.md#developers-certificate-of-origin)
- [x] Tested
- [x] Review ready

### DIFF
--- a/test/http2-pipelining-default.js
+++ b/test/http2-pipelining-default.js
@@ -225,7 +225,12 @@ test('fetch POST bodies dispatch concurrently on the same h2 session instead of 
     sessions.add(session)
   })
 
-  server.on('stream', stream => {
+  server.on('stream', (stream, headers) => {
+    if (headers[':path'] === '/warmup') {
+      stream.respond({ ':status': 200 })
+      stream.end('warm')
+      return
+    }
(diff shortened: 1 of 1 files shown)


ostapondo:
Reopened - closing this was my mistake, the flake is still there. Rebased on main so CodeQL goes green.

metcoder95 (approved):

PR #5657, opened 2026-08-05
ci: only accept a numeric nightly major

ostapondo:
Follow-up to #5654.

The guard I landed there only catches an empty string and the literal
"null". Anything else the download index hands back ends up in
`needs.nightly-version.outputs.major`, which nodejs-shared.yml drops into a
github-script block. Checking for digits is what it was meant to do.

Ran the check by hand: 27 passes; empty, "null", "8abc" and a string with
quotes in it all fail.

- [x] I have read and agreed to the [Developer's Certificate of Origin](https://github.com/nodejs/undici/blob/main/CONTRIBUTING.md#developers-certificate-of-origin)
- [x] Tested
- [x] Review ready


mcollina (approved):
lgtm

PR #3093, opened 2024-04-11
Skip the creation of a transform stream in fetch

mcollina:
This skips the creation of a `TransformStream`, which showed up glowing in the following flamegraph:

Screenshot 2024-04-11 at 19 11 07

After this patch:

Screenshot 2024-04-11 at 19 12 05

## Benchmarks

main:

```
[bench:run] ┌─────────┬───────────────────────┬─────────┬────────────────────┬────────────┬─────────────────────────┐
[bench:run] │ (index) │ Tests                 │ Samples │ Result             │ Tolerance  │ Difference with slowest │
[bench:run] ├─────────┼───────────────────────┼─────────┼────────────────────┼────────────┼─────────────────────────┤
[bench:run] │ 0       │ 'undici - fetch'      │ 30      │ '4571.92 req/sec'  │ '± 2.72 %' │ '-'                     │
[bench:run] │ 1       │ 'request'             │ 35      │ '5183.44 req/sec'  │ '± 2.90 %' │ '+ 13.38 %'             │
[bench:run] │ 2       │ 'node-fetch'          │ 10      │ '5405.14 req/sec'  │ '± 1.25 %' │ '+ 18.22 %'             │
[bench:run] │ 3       │ 'axios'               │ 35      │ '5423.29 req/sec'  │ '± 2.80 %' │ '+ 18.62 %'             │
[bench:run] │ 4       │ 'got'                 │ 25      │ '5672.87 req/sec'  │ '± 2.87 %' │ '+ 24.08 %'             │
[bench:run] │ 5       │ 'http - no keepalive' │ 10      │ '6407.75 req/sec'  │ '± 3.00 %' │ '+ 40.15 %'             │
[bench:run] │ 6       │ 'superagent'          │ 10      │ '9827.22 req/sec'  │ '± 1.69 %' │ '+ 114.95 %'            │
[bench:run] │ 7       │ 'http - keepalive'    │ 10      │ '11160.93 req/sec' │ '± 2.49 %' │ '+ 144.12 %'            │
[bench:run] │ 8       │ 'undici - pipeline'   │ 50      │ '12261.22 req/sec' │ '± 3.00 %' │ '+ 168.19 %'            │
[bench:run] │ 9       │ 'undici - stream'     │ 55      │ '19454.69 req/sec' │ '± 2.95 %' │ '+ 325.53 %'            │
[...]

PR #5658, opened 2026-08-06
fix(cache): settle Cache.add/addAll for responses with a body

edenbuilds:
## Summary
- `Cache.add` / `Cache.addAll` hung forever when the response had a body (204/no-body settled fine).
- Root cause: `fetchFinale` attaches `stream.finished()` for end-of-body, but undici body streams are pull-driven — with no consumer, the stream never finishes and `processResponseEndOfBody` never resolves the waiters (issue #5615).
- Fix mirrors `Cache.put`: clone the response, fully read the original stream (unlocks end-of-body), set `body.source` on the clone for storage.

## Test plan
- [x] New regression: `node --test test/web/cache-add-body.js`
- [x] Manual repro from #5615 settles and `cache.match` returns body text
- [ ] CI on PR

Made with [Cursor](https://cursor.com)

### DIFF
--- a/lib/web/cache/cache.js
+++ b/lib/web/cache/cache.js
@@ -153,6 +153,10 @@ class Cache {
       // 5.6
       const responsePromise = Promise.withResolvers()
 
+      // Holds the response object that should be written into the cache. For
+      // body-bearing responses this is a clone with body.source filled in.
+      let responseForCache = null
+
       // 5.7
       fetchControllers.push(fetching({
         request: r,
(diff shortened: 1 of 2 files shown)


KhafraDev (changes requested):
If, as the issue OP suggested, it is caused by https://github.com/nodejs/undici/pull/3093, then the fix is to revert that, not some crummy ai slop.

edenbuilds:
Thanks, I checked this against current `main` and reproduced the tradeoff. I refreshed a local copy onto current `main` (the fork branch was 96 commits behind): restoring the #3093 transform-stream path by itself still leaves the new body-bearing `cache.add()` regression pending, while retaining the current cache workaround makes the subsequent `cache.match()` fail because the stored response stream is locked. I have not pushed a speculative revert that would leave either path broken. The remaining fix needs a targeted cache/fetch integration shape, or confirmation that you want the broader #3093 revert despite that regression.

KhafraDev:
#5844 is an actual fix. Please don't send another ai generated response.

PR #5659, opened 2026-08-06
fix(cache): add response delay to corrected age value

RaphaelFakhri:
RFC 9111 [section 4.2.3](https://www.rfc-editor.org/rfc/rfc9111.html#section-4.2.3) defines a stored response's initial age as:

```
apparent_age           = max(0, response_time - date_value)
response_delay         = response_time - request_time
corrected_age_value    = age_value + response_delay
corrected_initial_age  = max(apparent_age, corrected_age_value)
```

`CacheHandler` captured a single timestamp at response-header time and computed
`max(apparentAge, resAge)`. There is no `request_time`, so `response_delay` was always
zero and the `corrected_age_value` term reduced to the bare `Age` the origin sent.

The delay matters because the Age an origin reports is already out of date by the time the
response finishes arriving. Dropping it stores a slow response as younger than it is, and a
response near its freshness boundary is then served after it has actually expired.

### Reproduction

Origin sends `Cache-Control: public, max-age=600`, `Age: 100`, and delays 3 seconds:

```
response_delay measured   : 3030 ms
upstream Age              : 100 s
RFC corrected_initial_age : >= 103 s
Age served from cache     : 100 s     <- understated by the full response delay
```

The understatement scales with upstream latency, so it is largest exactly where caching
matters most.

### Change

`onRequestStart` records the request time, and `corrected_age_value` adds the resulting
delay. The field is reset on every `onRequestStart`, so a retried or redirected request
measures its own delay rather than the first attempt's.

### Tests

`test/interceptors/cache-corrected-age.js`, three cases:

- the served age includes the response delay
- a response without an `Age` header is aged by at most the response delay itself, since
  section 4.2.3 treats a missing `Age` as `age_value = 0`
- a response whose corrected age already exceeds `max-age` is not reused

Cases 1 and 3 fail on main and pass with this change. Case 2 passes on both and is there to
pin the bound on responses that carry no `Age`.

`npm run test:cache-interceptor` passes 76/76, and lint is clean.

### DIFF
--- a/lib/handler/cache-handler.js
+++ b/lib/handler/cache-handler.js
@@ -155,6 +155,15 @@ class CacheHandler {
   #writeStream
 
   /**
+   * Time the request was sent, used for the response_delay term of RFC 9111's
+   * corrected_age_value. Reset on every onRequestStart so a retried or redirected
+   * request measures its own delay rather than the first attempt's.
+   *
+   * @type {number | undefined}
+   */
+  #requestTime
+
+  /**
    * @param {import('../../types/cache-interceptor.d.ts').default.CacheHandlerOptions} opts
    * @param {import('../../types/cache-interceptor.d.ts').default.CacheKey} cacheKey
    * @param {import('../../types/dispatcher.d.ts').default.DispatchHandler} handler
@@ -170,6 +179,7 @@ class CacheHandler {
   onRequestStart (controller, context) {
     this.#writeStream?.destroy()
     this.#writeStream = undefined
+    this.#requestTime = Date.now()
     this.#handler.onRequestStart?.(controller, context)
   }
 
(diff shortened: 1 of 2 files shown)


metcoder95 (approved):

mcollina (changes requested):
can you avoid using setTimeout in tests but use fake-timers instead? timers are flaky.

RaphaelFakhri:
Done, switched to @sinonjs/fake-timers with `toFake: ['Date']`, same pattern as test/interceptors/cache.js. The response delay is now injected via `clock.tick()` inside the origin handler, so the assertions are exact values instead of lower bounds and nothing depends on real timers. The http2-request-never-settles timeout on Node 25 looks unrelated to this change (http2 test, pre-existing flake).

RaphaelFakhri:
Rebased onto current main; the fake-timers change from my last comment is unchanged in the second commit. Green locally on the rebased branch: test/interceptors/cache-corrected-age.js, test/interceptors/cache.js and test/cache-interceptor/ all pass, and npm run lint is clean. The workflow runs are sitting at action_required, so CI will need a maintainer to approve them.

@mcollina could you take another look when you get a chance?

Edit: force-pushed once more to fix the commit author email on both commits; the diff is unchanged.

PR #5660, opened 2026-08-06
feat(dns): add address filter

### LINKED ISSUE #2019: SSRF protection in undici / native-node-fetch
## This would solve...

`undici` and native-node>=18-`fetch` expose (as far as I know, and see documented) no way to protect against [SSRF attacks](https://scalesec.com/blog/exploit-ssrf-to-gain-aws-credentials/) when making server-side calls to arbitrary / user-controllable URLs.

In the nodejs ecosystem,

- Non-`fetch` libs like `request` or `axios` can use middleware like [ssrf-req-filter](https://github.com/y-mehta/ssrf-req-filter). It provides an `agent` / `httpAgent` / `httpsAgent` that you can pass in these libs second `options` argument:

```js
const ssrfFilter = require('ssrf-req-filter');
const url = 'http://169.254.169.254/latest/user-data/iam/security-credentials/' // your choice of sensitive "internal" URL here
axios.get(url, { httpAgent: ssrfFilter(url), httpsAgent: ssrfFilter(url) }) // <-- SSRF "PLUGIN" ADDED HERE
    .then((response) => { console.log(`Success', response) });
    .catch((error) => { console.log('Error', error); })
```

- 3rd-party `fetch` lib `node-fetch` also supports ssrf-req-filter, though a similar `agent` API:

```js
const ssrfFilter = require('ssrf-req-filter');
const nodefetch = require("node-fetch");
const url = 'http://169.254.169.254/latest/user-data/iam/security-credentials/' // your choice of sensitive "internal" URL here
nodefetch(url, { agent: ssrfFilter(url) }) // <-- SSRF "PLUGIN" ADDED HERE
    .then((response) => { console.log(`Success', response) });
    .catch((error) => { console.log('Error', error); })
```

The implementation of such SSRF protection is out of topic, but:

- [ssrf-req-filter's "TL;DR" is this line](https://github.com/y-mehta/ssrf-req-filter/blob/e8d46324819c38feef2c576f437761ee60507720/lib/index.js#L12-L13) ...
- ... which depends on [ipaddr.js categorizing addresses according to various RFCs](https://github.com/whitequark/ipaddr.js/blob/8e1c5a81afb97ce3ecdfb3ad4585a7e7be893519/lib/ipaddr.js#L184-L211)

But in `undici` / built-in-node>=18 native `fetch`, I see no trace of supporting such agents with a `({host: string}) => boolean` signature letting me decide at runtime to cancel a request.

- Is it actually unsupported? If so, what do you (undici/native-node-fetch maintainers) recommend to do to protect against SSRF? To make the SSRF check before calling `fetch`? (At the cost of an extra DNS resolution before calling `fetch`)
- Or is it supported but lacks documentation (or I missed it 😄?)

undici (fetch?) should support setting kind of option letting users plug anti-SSRF logic. Typically, an `agent` function of signature `{host: string} => boolean` (agent is passed an IP, and expected to return a boolean).

## I have also considered...

I grepped [the fetch WHATWG spec](https://fetch.spec.whatwg.org/) for "SSRF", without success.

## Additional context

None. Thanks for fetch!

JasonColapietro:
## Summary

- add an optional `filter(origin, record)` policy hook to the DNS interceptor
- apply the policy to every resolved address and to literal IPv4 and IPv6 origins before connecting
- keep enforcement active for fetch redirects and when the DNS cache has reached `maxItems`
- document the application-owned policy boundary and expose the hook in the TypeScript definitions

This builds on the existing DNS interceptor so accepted hostnames are resolved once and the connection uses the accepted literal address while preserving the original Host header and TLS servername. Undici does not classify public or private ranges; callers supply that policy.

Closes #2019.

## Validation

- `npm run test:interceptors` (316 passed, 3 skipped)
- `npm run test:typescript`
- `npm run lint`
- `git diff --check`

### DIFF
--- a/docs/docs/api/Interceptors.md
+++ b/docs/docs/api/Interceptors.md
@@ -288,6 +288,10 @@ dual-stack (IPv4 + IPv6) and custom lookup/storage implementations.
     `(origin, options, callback) => void`.
   * `pick` {Function} (optional) Custom record-selection function called with
     `(origin, records, affinity)` to choose which resolved address to use.
+  * `filter` {Function} (optional) Address policy called with
+    `(origin, record)` for every resolved address and literal IP origin. Return
+    `true` to allow the address or `false` to reject it. If all addresses are
+    rejected, the request fails before a connection is attempted.
   * `storage` {DNSStorage} (optional) Custom storage backend. Must implement
     `get`, `set`, `delete`, `full`, and `size`.
 
@@ -318,6 +322,30 @@ const agent = new Agent().compose(
   })
 )
 ```
+
+The optional `filter` hook can enforce an application-defined network policy
+without a separate DNS preflight lookup. Accepted hostnames are connected
+through the resolved IP address while the original host header and TLS
+servername are preserved. The policy is also applied to literal IPv4 and IPv6
+origins and to redirect targets when the dispatcher is used with `fetch`.
+
+```js
+import { Agent, interceptors } from 'undici'
+
+const agent = new Agent().compose(
+  interceptors.dns({
+    filter (origin, { address, family }) {
+      return addressPolicy.allows({ origin, address, family })
+    }
+  })
+)
+
+const response = await fetch(userControlledURL, { dispatcher: agent })
+```
+
+`filter` does not provide an address classification policy. Applications are
+responsible for rejecting every address range that is not permitted in their
+environment, including IPv4-mapped IPv6 addresses where applicable.
 
 ---
 
--- a/lib/interceptor/dns.js
+++ b/lib/interceptor/dns.js
@@ -141,6 +141,7 @@ class DNSInstance {
   affinity = null
   lookup = null
   pick = null
+  filter = null
   storage = null
 
   constructor (opts) {
@@ -150,14 +151,23 @@ class DNSInstance {
     this.affinity = opts.affinity
     this.lookup = opts.lookup ?? this.#defaultLookup
     this.pick = opts.pick ?? this.#defaultPick
+    this.filter = opts.filter
     this.storage = opts.storage ?? new DNSStorage(opts)
+  }
+
+  filterRecords (origin, records) {
+    if (this.filter == null) {
+      return records
+    }
+
+    return Array.from(records).filter(record => this.filter(origin, record))
   }
 
   runLookup (origin, opts, cb) {
     const ips = this.storage.get(origin.hostname)
 
-    // If full, we just return the origin
-    if (ips == null && this.storage.full()) {
+    // If full, return the origin unless a filter still needs to be enforced.
+    if (ips == null && this.storage.full() && this.filter == null) {
       cb(null, origin)
       return
     }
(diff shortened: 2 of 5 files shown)


docs/docs/api/Interceptors.md
```diff
@@ -319,6 +323,30 @@ const agent = new Agent().compose(
...
+The optional `filter` hook can enforce an application-defined network policy
+without a separate DNS preflight lookup. Accepted hostnames are connected
+through the resolved IP address while the original host header and TLS
+servername are preserved. The policy is also applied to literal IPv4 and IPv6
+origins and to redirect targets when the dispatcher is used with `fetch`.
+
+```js
+import { Agent, interceptors } from 'undici'
+
+const agent = new Agent().compose(
```
ronjouch: I’m new to Interceptors and from what I understand it’s easy to footgun yourself with order here, e.g. `redirect` should come first before `dns`, right? What about mentioning it, or adding `redirect` to your example clearly after `dns` ?

lib/interceptor/dns.js
```diff
@@ -180,8 +190,25 @@ class DNSInstance {
...
-        const records = this.storage.get(origin.hostname)
+        try {
+          addresses = this.filterRecords(origin, addresses)
+        } catch (err) {
+          cb(err)
+          return
+        }
+
+        if (addresses.length === 0) {
+          cb(new InformationalError('DNS addresses rejected by filter'))
```
ronjouch: What about throwing a new error class? Current code would force filtering by `error.message` if I want to specialize handling, which feels brittle.

lib/interceptor/dns.js
```diff
@@ -150,14 +151,23 @@ class DNSInstance {
...
+    this.filter = opts.filter
     this.storage = opts.storage ?? new DNSStorage(opts)
   }
 
+  filterRecords (origin, records) {
+    if (this.filter == null) {
+      return records
+    }
+
+    return Array.from(records).filter(record => this.filter(origin, record))
```
ronjouch: What about being stricter:
```suggestion
    return Array.from(records).filter(record => this.filter(origin, record) === true)
```
, to avoid accidental filtering in because filter return a truthy value?

ronjouch (commented):
@JasonColapietro thx for giving it a stab. Couple comments/questions from me (note that I am just the reporter of https://github.com/nodejs/undici/issues/2019 , not a undici maintainer, and undici maintainers may disagree with my comments)

docs/docs/api/Interceptors.md
JasonColapietro: Good call. The example now composes `[dns(...), redirect(...)]`, and the docs say that `dns()` should come before `redirect()`. That way redirect wraps DNS, and every redirected target is resolved and checked against the filter before connecting. (`2e4930dc`)

lib/interceptor/dns.js
JasonColapietro: Added `DNSFilterError` (`code: UND_ERR_DNS_FILTER`), exported next to the other errors and typed in `types/errors.d.ts`, with the same `Symbol.for` `instanceof` pattern. Both rejection paths throw it now, so callers can branch on the class or the code instead of on the message. (`2e4930dc`)

lib/interceptor/dns.js
JasonColapietro: Applied. Only a strict `true` keeps a record now, and there is a test showing that truthy non-boolean return values are rejected. (`2e4930dc`)

lib/interceptor/dns.js
```diff
@@ -151,14 +152,23 @@ class DNSInstance {
...
+    this.filter = opts.filter
     this.storage = opts.storage ?? new DNSStorage(opts)
   }
 
+  filterRecords (origin, records) {
+    if (this.filter == null) {
+      return records
+    }
+
+    return Array.from(records).filter(record => this.filter(origin, record) === true)
```
metcoder95: I'd suggest a simple for-loop for that

metcoder95 (approved):
lgtm overall, just small comments

PR #4923, opened 2026-03-23
fix: avoid prototype collisions in parseHeaders

mcollina:
Fix parseHeaders() for response headers and trailers that shadow Object.prototype.

Adds regression tests.

PR #5663, opened 2026-08-07
fix(cookies): keep a cookie named __proto__ in getCookies

luantaraschi:
## This relates to...

No open issue. Found while reading `lib/web/cookies/`.

## Rationale

`getCookies()` builds its result on a plain object literal:

```js
const out = {}
for (const piece of cookie.split(';')) {
  const [name, ...value] = piece.split('=')
  out[name.trim()] = value.join('=')
}
```

`__proto__` is a valid cookie name (it is a token per RFC 6265), but assigning it on a plain object reaches the `Object.prototype` setter instead of creating an own property. The setter only accepts an object or null, and the assigned value here is always a string, so it does nothing and the cookie is dropped:

```js
const headers = new Headers()
headers.set('Cookie', '__proto__=foo; bar=baz')

getCookies(headers)              // { bar: 'baz' }, the first cookie is gone
getCookies(headers).__proto__    // Object.prototype, not the string 'foo'
```

The second line is the part that bothered me more than the missing key. `getCookies` is typed `Record<string, string>`, so a caller that indexes it by an attacker-influenced name gets an object back where the types promise a string.

There is no prototype pollution here: the setter refuses a string, so `Object.prototype` is untouched. The bug is silent data loss plus a broken type contract.

## Changes

`getCookies` now builds its record with `{ __proto__: null }`, which is already the pattern used in `lib/core/util.js`, `lib/util/runtime-features.js`, `lib/web/eventsource/eventsource.js` and `lib/web/fetch/formdata.js`.

The existing tests all use `assert.deepEqual`, which does not compare prototypes, so none of them needed changing. All 92 tests under `test/cookie/` pass.

### Features

N/A

### Bug Fixes

A cookie named `__proto__` is now returned by `getCookies()` like any other name.

The returned object no longer inherits from `Object.prototype`, so calling `cookies.hasOwnProperty(name)` on it stops working. `Object.hasOwn(cookies, name)` and `name in cookies` still work, as does every form of indexing and iteration.

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/web/cookies/index.js
+++ b/lib/web/cookies/index.js
@@ -32,8 +32,10 @@ function getCookies (headers) {
 
   const cookie = headers.get('cookie')
 
+  // A null prototype keeps a cookie named `__proto__` from hitting the
+  // Object.prototype setter, which would silently drop it.
   /** @type {Record<string, string>} */
-  const out = {}
+  const out = { __proto__: null }
 
   if (!cookie) {
     return out
--- a/test/cookie/cookies.js
+++ b/test/cookie/cookies.js
@@ -59,6 +59,24 @@ test('Cookie parser', () => {
   })
 })
 
+test('Cookie parser keeps a cookie named __proto__', () => {
+  const headers = new Headers()
+  headers.set('Cookie', '__proto__=foo; bar=baz')
+
+  const cookies = getCookies(headers)
+
+  assert.deepStrictEqual(Object.keys(cookies).sort(), ['__proto__', 'bar'])
+  assert.strictEqual(
+    Object.getOwnPropertyDescriptor(cookies, '__proto__').value,
+    'foo'
+  )
+  assert.strictEqual(cookies.bar, 'baz')
+
+  for (const value of Object.values(cookies)) {
+    assert.strictEqual(typeof value, 'string')
+  }
+})
+
 test('Cookie Name Validation', () => {
   const tokens = [
     '"id"',


mcollina (approved):
LGTM

(However, I think this is a bad idea. We should throw hard when __proto__ appears).

luantaraschi:
Thanks for the review. I checked the three failing jobs, and none of them exercises this change, which only touches `lib/web/cookies/`.

- The two Node 24 Ubuntu jobs fail in `test/http2-request-never-settles.js`. The same test failed on `main` in run 31085361670 on 6 August, before this branch existed.
- The Node 22 macOS job times out in `test/fetch/content-length.js` at 180006 ms, matching the 180-second limit rather than an assertion failure.

On `__proto__`, I followed the behavior already used in two places:

- `b64e7e4`, "fix: avoid prototype collisions in parseHeaders (#4923)"
- `90a1609`, "fix: preserve response trailers that shadow Object.prototype"

Both preserve keys received from the wire with `Object.defineProperty` instead of rejecting them. This PR uses the same approach in `getCookies`.

`__proto__` is also a valid cookie name under the token grammar in RFC 6265 section 4.1.1. Throwing from `getCookies` would therefore let an incoming cookie break its caller. If the project prefers rejecting this key, I think that policy should be applied consistently to all three parsing paths rather than only to cookies. I am happy to help with that as a separate change.

A rerun should be enough for this PR. I can rebase if you would prefer a fresh run that way.

metcoder95 (approved):

luantaraschi:
Thanks for the review. One thing to flag, and it is on me: about a minute after you approved I merged `main` into the branch to get the old red run out of the way, and the new head now reports no checks at all.

I cannot start workflow runs in this repository myself, so it needs someone with write access to kick them off before this can land. Sorry for the extra step, the timing was clumsy.

luantaraschi:
CI is red on two jobs, and neither failure comes from this change.

Node 26 on macOS failed in `test/issue-5087.js`. The test itself passed, then late activity after it ended surfaced `read ECONNRESET`, which the runner reported as a file level failure. Node 25 on ubuntu with WASM SIMD disabled failed in `test/http2-request-never-settles.js`. Neither file touches cookies, and `Test unit` runs `borp -p "test/*.js"`, which does not cover `test/cookie/*.js`.

Locally on this branch that same step passes: 1510 tests, no failures (1506 passed, 4 skipped). `npm run test:cookies` passes 92 of 92.

The flakiness is not specific to this branch either. Of the last ten CI runs on `main`, six failed the Node 24 coverage job and one failed the same Node 25 WASM SIMD job. `test/issue-5087.js` still has no guard for late session errors, unlike the h2 tests hardened in #5674 and #5686.

I can rebase onto current `main` if a fresh run would help.

PR #5665, opened 2026-08-07
fix(decompress): stop inflating compress and x-compress

luantaraschi:
## This relates to...

No open issue. Found while reading `lib/interceptor/decompress.js`.

## Rationale

`supportedEncodings` maps `compress` and `x-compress` to `createInflate`:

```js
deflate: createInflate,
compress: createInflate,
'x-compress': createInflate,
```

Those are different formats. RFC 9110 section 8.4.1.1 defines `compress` as the UNIX compress (LZW) format, while `deflate` is zlib. Node has no LZW decoder, so feeding a `compress` body to inflate cannot succeed. It does not degrade, it fails the request:

```
Z_DATA_ERROR: incorrect header check
```

`fetch()` does not do this. Its coding switch in `lib/web/fetch/index.js` handles `x-gzip`, `gzip`, `deflate`, `br` and `zstd`, and anything else falls through untouched. Same response, same server, both from this repo:

```
interceptor : Z_DATA_ERROR | incorrect header check
fetch       : "this body is not deflate data"
```

So composing the decompress interceptor onto a client turns a response that `fetch` reads fine into a hard failure.

The two entries are removed. `compress` and `x-compress` now take the path every other unrecognised coding already takes: `#createDecompressionChain` returns an empty chain, the body passes through, and `Content-Encoding` stays on the response since nothing was decoded. The only case this changes is a server that labels zlib data as `compress`, which is already mislabelled and which `fetch` never decoded either.

The doc line that advertised the two codings is updated to match.

Worth noting: this touches the same file as #5664, but a different region, so the two apply independently.

## Changes

`compress` and `x-compress` removed from `supportedEncodings`, with a comment recording why.

`docs/docs/api/Interceptors.md`: the decompress section no longer lists the two codings and says unrecognised codings are left untouched.

Test added to `test/interceptors/decompress.js`: a `Content-Encoding: compress` response with a non-inflatable body arrives unchanged instead of erroring.

### Features

N/A

### Bug Fixes

A `Content-Encoding: compress` response no longer fails with `Z_DATA_ERROR` under the decompress interceptor.

A server sending zlib data labelled `compress` was being decoded and now is not. That labelling is already wrong per RFC 9110, and `fetch()` in this repo has never decoded it.

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/docs/docs/api/Interceptors.md
+++ b/docs/docs/api/Interceptors.md
@@ -198,7 +198,8 @@ const agent = new Agent().compose(
 ## `interceptors.decompress([opts])`
 
 Automatically decompresses response bodies encoded with `gzip`, `x-gzip`,
-`br` (Brotli), `deflate`, `compress`, `x-compress`, or `zstd`.
+`br` (Brotli), `deflate`, or `zstd`. Any other coding, including `compress`
+and `x-compress`, is left untouched, matching `fetch()`.
 
 > **Experimental:** This interceptor is experimental and subject to change.
 > A one-time `ExperimentalWarning` is emitted on first use.
--- a/lib/interceptor/decompress.js
+++ b/lib/interceptor/decompress.js
@@ -9,14 +9,16 @@ const DecoratorHandler = require('../handler/decorator-handler')
 /** @typedef {import('node:stream').Transform} Controller */
 /** @typedef {Transform&import('node:zlib').Zlib} DecompressorStream */
 
+// `compress` and `x-compress` are deliberately absent. RFC 9110 section
+// 8.4.1.1 defines them as the UNIX compress (LZW) format, which zlib cannot
+// read, so inflating them turns a readable response into a Z_DATA_ERROR.
+// fetch() leaves them alone, and so does this interceptor.
 /** @type {Record<string, () => DecompressorStream>} */
 const supportedEncodings = {
   gzip: createGunzip,
   'x-gzip': createGunzip,
   br: createBrotliDecompress,
   deflate: createInflate,
-  compress: createInflate,
-  'x-compress': createInflate,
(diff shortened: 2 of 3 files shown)


luantaraschi:
Quick note on the one red job here, in case it is what is holding this up.

The failure is `test/http2-request-never-settles.js` on `Test with Node.js 24 on ubuntu-latest with WASM SIMD disabled`, and it is the only failing test in that run. #5674 records it as a flake on `main`, with the root cause upstream in nodejs/node#64841, the Maglev SIGSEGV on Node >= v24.15.0. The signature matches what that PR describes: no TAP output, no subtests, and the file ending in 3.1s, well under the 15s watchdog.

This branch only touches `lib/interceptor/decompress.js` and its test, so there is no path from it into an HTTP/2 session. Running both files on this branch locally, on Node 22.20.0: `test/interceptors/decompress.js` 28 passing, and `test/http2-request-never-settles.js` 2 passing.

The branch is 6 commits behind `main` and predates #5674. Glad to rebase for a fresh run if that helps, just say the word.

metcoder95:
can you address conflicts?

luantaraschi:
Rebased. The conflict was #5807 adding `__proto__: null` to the same table I am trimming, so both changes sit together now.

While I was there I rechecked the claim in the comment against today's fetch: the coding switch in `lib/web/fetch/index.js` handles x-gzip, gzip, deflate, br and zstd, and anything else clears the decoder list, so compress still passes through there untouched. `test/interceptors/decompress.js` is green, 51 tests.

metcoder95 (approved):

PR #5666, opened 2026-08-07
fix(fetch): keep a request header named __proto__

luantaraschi:
## This relates to...

No open issue. Same defect class as `test/prototype-headers.js` already covers for `client.request`, on the path that file does not reach.

## Rationale

`HeadersList.entries` builds a plain object keyed by header name:

```js
for (const { name, value } of this.headersMap.values()) {
  headers[name] = value
}
```

`__proto__` is a valid header name, and assigning it on a plain object reaches the `Object.prototype` setter instead of creating an own property. The setter only takes an object or null, the value here is always a string, so the header is dropped without an error.

`lib/web/fetch/index.js` passes this object to the dispatcher as the request headers, so a header the caller explicitly set never reaches the wire:

```js
const headers = new Headers()
headers.set('__proto__', 'pwned')
headers.set('x-control', 'sent')

await fetch(url, { headers })

headers.get('__proto__')   // 'pwned', the Headers object kept it
// server received:  x-control: sent          <- present
// server received:  __proto__: pwned         <- missing
```

`Headers` itself is fine. It stores the entry in a Map and reads it back correctly. Only the conversion to a plain object loses it.

`parseHeaders` in `lib/core/util.js` already guards this exact case, which is what `test/prototype-headers.js` exercises for `client.request`. The fetch path just never got the same treatment. `lib/web/websocket/websocket.js` reads the same getter for its response headers.

There is no prototype pollution here: the setter refuses a string, so `Object.prototype` is untouched. The bug is a silently dropped header.

## Changes

`HeadersList.entries` now writes the `__proto__` key with `Object.defineProperty`, copying the guard already used by `parseHeaders`.

I kept the object plain rather than switching it to a null prototype on purpose: `[util.inspect.custom]` formats this same object, so a null prototype would turn `Headers { ... }` into `Headers [Object: null prototype] { ... }` and change output that `test/fetch/headers-inspect-custom.js` asserts on.

Test added to `test/prototype-headers.js`: a `fetch` whose `Headers` carries `__proto__` reaches the server with that header, alongside a control header that already worked.

### Features

N/A

### Bug Fixes

`fetch` no longer drops a request header named `__proto__`.

None.

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/web/fetch/headers.js
+++ b/lib/web/fetch/headers.js
@@ -320,7 +320,19 @@ class HeadersList {
 
     if (this.headersMap.size !== 0) {
       for (const { name, value } of this.headersMap.values()) {
-        headers[name] = value
+        if (name === '__proto__') {
+          // Plain assignment would reach the Object.prototype setter, which
+          // refuses a string and drops the header. Same guard as parseHeaders
+          // in lib/core/util.js.
+          Object.defineProperty(headers, name, {
+            configurable: true,
+            enumerable: true,
+            value,
+            writable: true
+          })
+        } else {
+          headers[name] = value
+        }
       }
     }
 
--- a/test/prototype-headers.js
+++ b/test/prototype-headers.js
@@ -82,3 +82,49 @@ test('request handles response trailers that shadow Object.prototype', async (t)
   assert.strictEqual(Object.getOwnPropertyDescriptor(trailers, '__proto__').value, 'trailer')
   assert.strictEqual(Object.getOwnPropertyDescriptor(trailers, 'constructor').value, 'built-in-trailer')
 })
+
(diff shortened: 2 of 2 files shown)


metcoder95:
Can you abstract this solution into a single util and implement it across features?

luantaraschi:
Done, this is now #5688: a single `setHeader` util in `lib/core/util.js`, used by the fetch headers list, the cache keys, the mock snapshot recorder and the HTTP/2 request builder. Thanks for the steer, it collapsed four near-identical diffs into one.

Closing this one in favour of it. Happy to reopen if you would rather review the sites separately.

PR #5667, opened 2026-08-07
fix(cache): keep a header named __proto__ out of the prototype chain

luantaraschi:
## This relates to...

No open issue. Extends the collision class that #5012 covered for delimiters.

## Rationale

`appendHeader` in `lib/util/cache.js` accumulates onto a plain object and probes it by index:

```js
const current = headers[headerName]

if (current === undefined) {
  headers[headerName] = ...
} else if (Array.isArray(current)) {
  current.push(...values)
} else {
  headers[headerName] = [current, ...values]
}
```

For `__proto__` that probe returns `Object.prototype`, not `undefined`, so the header takes the last branch and the accumulator is assigned an **array**. Unlike a string, an array is accepted by the `Object.prototype` setter, so the assignment reprototypes the accumulator instead of storing the header.

An own `__proto__` key is not exotic. `JSON.parse` produces one, which is how headers read from a config file or a request body arrive:

```js
normalizeHeaders({ headers: JSON.parse('{"__proto__":"a","x-real":"same"}') })
```

Before:

```
Object.keys(headers)          -> [ 'x-real' ]
Object.getPrototypeOf(headers) -> [ [Object: null prototype] {}, 'a' ]   // an Array
headers.length                -> 2
```

Two problems from that one assignment. The header is gone, and the result now inherits from an array, so a lookup for a header named `length` answers `2` and `at`, `map`, `keys` and friends all resolve to functions. `appendHeader` itself branches on `headers[headerName] === undefined`, so a later header with one of those names takes the wrong branch.

The lost header then reaches `makeDeduplicationKey`, which rebuilds a plain object the same way, so two different requests produce one key:

```js
makeDeduplicationKey({ ..., headers: JSON.parse('{"__proto__":"a","x-real":"same"}') })
makeDeduplicationKey({ ..., headers: { 'x-real': 'same' } })
// both: ["https://example.com","GET","/",{"x-real":"same"}]
```

The comment above that function says the key format was reworked because different header sets could produce identical keys (#5012). This is the same failure, reached through the header name instead of the value.

There is no `Object.prototype` pollution: the reprototyped object is the local accumulator, and the global prototype is untouched.

## Changes

A `setHeader` helper writes the `__proto__` key with `Object.defineProperty`, matching the guard `parseHeaders` already uses in `lib/core/util.js`. `appendHeader` probes with `Object.hasOwn` so an inherited name is not mistaken for an existing header, and `makeDeduplicationKey` writes through the same helper.

I kept the accumulator a plain object rather than switching it to a null prototype: it is handed to user-supplied cache stores as part of the cache key, and dropping `Object.prototype` from it would be a visible change for that code.

After:

```
Object.keys(headers)           -> [ '__proto__', 'x-real' ]
Object.getPrototypeOf(headers) -> Object.prototype
headers.length                 -> undefined
keys collide                   -> false
```

Two tests: one in `test/cache-interceptor/cache-utils.js` for the normalization, next to the existing prototype-pollution test there, and one in `test/interceptors/deduplicate.js` beside the #5012 collision test.

`test/cache-interceptor` 77 passing, `test/interceptors` 302 passing with 14 skipped as before, `test/cache` 4 passing.

### Features

N/A

### Bug Fixes

A header named `__proto__` is no longer dropped from the normalized headers, no longer reprototypes the accumulator, and no longer makes two different requests share a deduplication key.

None.

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/util/cache.js
+++ b/lib/util/cache.js
@@ -164,17 +164,45 @@ function makeCacheKey (opts) {
   }
 }
 
+/**
+ * Writes a header onto a plain object without going through the
+ * Object.prototype `__proto__` setter, which would either drop the value or,
+ * when the value is an array, reprototype the accumulator. Same guard as
+ * parseHeaders in lib/core/util.js.
+ *
+ * @param {Record<string, string[] | string>} headers
+ * @param {string} name lowercased header name
+ * @param {string | string[]} value
+ */
+function setHeader (headers, name, value) {
+  if (name === '__proto__') {
+    Object.defineProperty(headers, name, {
+      configurable: true,
+      enumerable: true,
+      value,
+      writable: true
+    })
+  } else {
+    headers[name] = value
+  }
+}
+
 function appendHeader (headers, key, val) {
   const headerName = key.toLowerCase()
-  const current = headers[headerName]
+  // Reading `headers[headerName]` directly would resolve `__proto__` (and any
+  // other inherited name) through the prototype chain rather than report the
+  // header as absent.
+  const current = Object.hasOwn(headers, headerName)
+    ? headers[headerName]
+    : undefined
   const values = Array.isArray(val) ? val : [val]
 
   if (current === undefined) {
-    headers[headerName] = Array.isArray(val) ? val.slice() : val
+    setHeader(headers, headerName, Array.isArray(val) ? val.slice() : val)
   } else if (Array.isArray(current)) {
     current.push(...values)
   } else {
-    headers[headerName] = [current, ...values]
+    setHeader(headers, headerName, [current, ...values])
   }
 }
 
@@ -692,7 +720,7 @@ function makeDeduplicationKey (cacheKey, excludeHeaders) {
       if (excludeHeaders?.has(header.toLowerCase())) {
         continue
       }
-      headers[header] = cacheKey.headers[header]
+      setHeader(headers, header, cacheKey.headers[header])
     }
   }
 
--- a/test/cache-interceptor/cache-utils.js
+++ b/test/cache-interceptor/cache-utils.js
@@ -138,3 +138,20 @@ test('normalizeHeaders throws on non-string key in flat array', (t) => {
     message: 'opts.headers is not a valid header map'
   })
 })
+
+test('normalizeHeaders keeps a header named __proto__ without reprototyping the result', (t) => {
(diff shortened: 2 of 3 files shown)


luantaraschi:
While checking where else this shape appears, two more spots drop the same header. Neither is in this PR and I would rather ask than send them unprompted.

`buildRequestHeaders` in `lib/dispatcher/client-h2.js` probes the same way, and for a string value the probe returns `Object.prototype`, which is truthy:

```js
const current = headers[key]
headers[key] = current ? `${current}, ${val}` : val
```

So an h2 request carrying `__proto__` would first be turned into `"[object Object], a"` and then dropped by the setter. Confirmed against that function in isolation: `['__proto__', 'a', 'x-real', 'same']` comes back as `{ 'x-real': 'same' }`.

`normalizeHeaders` in `lib/mock/snapshot-utils.js` assigns straight into a plain object, so a recorded snapshot loses the header. Same input, same result.

The h2 one needs a real h2 round trip to test properly, and the mock one is a pure function so it is cheap. Happy to send either or both separately, in whatever order suits you, or to leave them if you would rather keep the surface small.

luantaraschi:
Both are up now, since #5663 was approved with the same guard: #5668 for the snapshot mock and #5669 for h2. Each stands on its own, so merge or drop them independently of this one.

The snapshot one turned out to have a second consequence I had not seen when I wrote the comment above: the dropped header makes two different requests share a snapshot key, so record mode collapses them and playback serves the wrong response.

metcoder95:
I'd advice to consolidate this utils into a single PR that serves both features

luantaraschi:
Thanks. Just to make sure I consolidate at the boundary you intend: should one PR cover the cache, snapshot, and HTTP/2 call sites through a shared helper, replacing #5667, #5668, and #5669? Or did you mean combining only the two follow-ups, #5668 and #5669? Once you confirm, I’ll restructure them accordingly.

metcoder95:
Yes, exactly

luantaraschi:
Consolidated into #5688, so the guard is one util in `lib/core/util.js` instead of four copies of the same function. The fix and its test moved across unchanged, and the new PR describes what it covers.

Closing this one to keep your queue clear. Happy to reopen if you would rather look at the sites separately.

PR #5668, opened 2026-08-08
fix(mock): keep a header named __proto__ in snapshots

luantaraschi:
## This relates to...

I raised this in a comment on #5667 and offered to send it separately. Sending it now that #5663 was approved, since it is the same guard on a subsystem where the consequence is different.

## Rationale

Three places in the snapshot mock build a plain object keyed by header name and assign into it:

- `normalizeHeaders` in `lib/mock/snapshot-utils.js`, both the array and the object branch
- `filterHeadersForMatching` and `filterHeadersForStorage` in `lib/mock/snapshot-recorder.js`

`__proto__` is a valid header name, and assigning it on a plain object reaches the `Object.prototype` setter instead of creating an own property. The value is a string, so the setter does nothing and the header is dropped.

It does not stop at a lost header. `formatRequestKey` feeds those headers to `createRequestHash`, which walks `Object.keys(...).sort()`, so two different requests produce one snapshot key:

```js
const filters = createHeaderFilters({})
const a = formatRequestKey({ origin, path, method: 'GET',
  headers: JSON.parse('{"__proto__":"a","x-real":"same"}') }, filters)
const b = formatRequestKey({ origin, path, method: 'GET',
  headers: { 'x-real': 'same' } }, filters)

createRequestHash(a) === createRequestHash(b)   // true
```

In record mode the two collapse into one entry, and in playback the recorded response for one is served for the other. `JSON.parse` is the everyday way an own `__proto__` key shows up, which is exactly how snapshots arrive from disk.

There is no `Object.prototype` pollution: the setter refuses a string, so the global prototype is untouched.

## Changes

A `setHeader` helper in `snapshot-utils.js` writes the `__proto__` key with `Object.defineProperty`, the guard `parseHeaders` already uses in `lib/core/util.js`. It is exported and used by all three writers.

The object stays plain rather than switching to a null prototype, because it is serialized into snapshot files and handed back to user code.

After the change the same pair hashes differently and the header survives:

```
headers stored: {"__proto__":"a","x-real":"same"}
hashes equal:   false
```

One test in `test/snapshot-recorder.js`, next to the existing `formatRequestKey` tests.

`test/snapshot*.js` 48 passing, `test/mock*.js` 339 passing, lint clean.

### Features

N/A

### Bug Fixes

A header named `__proto__` is no longer dropped from a recorded snapshot and no longer makes two different requests share a snapshot key.

None.

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/mock/snapshot-recorder.js
+++ b/lib/mock/snapshot-recorder.js
@@ -4,7 +4,7 @@ const { writeFile, readFile, mkdir } = require('node:fs/promises')
 const { dirname, resolve } = require('node:path')
 const { setTimeout, clearTimeout } = require('node:timers')
 const { InvalidArgumentError, UndiciError } = require('../core/errors')
-const { hashId, isUrlExcludedFactory, normalizeHeaders, createHeaderFilters } = require('./snapshot-utils')
+const { hashId, isUrlExcludedFactory, normalizeHeaders, createHeaderFilters, setHeader } = require('./snapshot-utils')
 
 /**
  * @typedef {Object} SnapshotRequestOptions
@@ -169,7 +169,7 @@ function filterHeadersForMatching (headers, headerFilters, matchOptions = {}) {
       if (!match.has(headerKey)) continue
     }
 
-    filtered[headerKey] = value
+    setHeader(filtered, headerKey, value)
   }
 
   return filtered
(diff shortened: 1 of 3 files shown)


luantaraschi:
Folded into #5688, where the guard lives once as `setHeader` in `lib/core/util.js` instead of being written out per feature. The snapshot fix and its test moved over unchanged.

Closing this one so your queue stays readable. Happy to reopen it if you would rather look at this site on its own.

PR #5669, opened 2026-08-08
fix(h2): keep a request header named __proto__

luantaraschi:
## This relates to...

The h2 half of what I flagged in a comment on #5667. `test/prototype-headers.js` already covers this concern for `client.request` over HTTP/1.

## Rationale

`buildRequestHeaders` in `lib/dispatcher/client-h2.js` probes the accumulator by index:

```js
const current = headers[key]
...
headers[key] = current ? `${current}, ${val}` : val
```

For `__proto__` that probe returns `Object.prototype`, which is truthy, so the header takes the "already present" branch and the value becomes `"[object Object], pwned"`. That string is then assigned through the `Object.prototype` setter, which refuses a string, so nothing is stored and the header never reaches the wire.

Reproduced over a real HTTP/2 connection, before the change:

```
x-control arrived: sent
__proto__ arrived: no
```

`__proto__` is a valid field name, and the flat array form is exactly what this function receives from the dispatcher.

There is no `Object.prototype` pollution: the setter refuses the string, so the global prototype is untouched.

## Changes

The probe uses `Object.hasOwn` so an inherited name is not mistaken for an existing header, and all four writes go through a `setHeader` helper using `Object.defineProperty`, the guard `parseHeaders` already uses in `lib/core/util.js`.

Test added to `test/prototype-headers.js`, alongside the HTTP/1 cases: an h2 request carrying `__proto__` reaches the server with that header, with a control header that already worked.

`test/+(http2|h2)*.js` 104 passing with 1 skipped as before, lint clean.

### Features

N/A

### Bug Fixes

An HTTP/2 request no longer drops a header named `__proto__`, and no longer corrupts its value with `[object Object], ` first.

None.

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -824,31 +824,50 @@ function shouldSendContentLength (method) {
   return method !== 'GET' && method !== 'HEAD' && method !== 'OPTIONS' && method !== 'TRACE' && method !== 'CONNECT'
 }
 
+/**
+ * Writes a header without going through the Object.prototype `__proto__`
+ * setter, which would drop it. Same guard as parseHeaders in lib/core/util.js.
+ */
+function setHeader (headers, name, value) {
+  if (name === '__proto__') {
+    Object.defineProperty(headers, name, {
+      configurable: true,
+      enumerable: true,
+      value,
+      writable: true
+    })
+  } else {
+    headers[name] = value
+  }
+}
+
 function buildRequestHeaders (reqHeaders) {
   const headers = {}
 
   for (let n = 0; n < reqHeaders.length; n += 2) {
     const key = reqHeaders[n + 0]
     const val = reqHeaders[n + 1]
-    const current = headers[key]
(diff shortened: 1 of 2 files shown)


luantaraschi:
The two failing Node 24 jobs stop at `test/http2-request-never-settles.js`. The same test also failed in #5663, an unrelated cookies change, while the rest of this matrix passed here.

I do not have permission to rerun workflows in this repository. Could a maintainer please rerun the failed jobs to confirm whether this is the same intermittent failure?

luantaraschi:
This one is in #5688 too. `buildRequestHeaders` now calls the shared `setHeader` from `lib/core/util.js` rather than carrying its own copy of the guard, and the test moved across as it was.

Closing here to keep the four of them from sitting in your queue. Say the word if you would rather review the h2 site separately and I will reopen it.

PR #5670, opened 2026-08-08
fix(date): slide the rfc850-date two-digit year window

luantaraschi:
`parseHttpDate` maps a two-digit rfc850-date year with the fixed 1970-2069 range, citing RFC 6265:

```js
// RFC 6265 states that the year is in the range 1970-2069.
year += year < 70 ? 2000 : 1900
```

RFC 6265 is the cookie spec. This parser has no cookie callers: `cache-handler.js` reads `Date`, `Expires` and `Last-Modified` through it, and `interceptor/cache.js` validates `Last-Modified` with it. For those, RFC 9110 section 5.6.7 applies, and it asks for a sliding window:

> Recipients of a timestamp value in rfc850-date format, which uses a two-digit year, MUST interpret a timestamp that appears to be more than 50 years in the future as representing the most recent year in the past that had the same last two digits.

## What changes

The visible effect is a rejection rather than a misreading, because `makeDate` validates the weekday against the year it settled on. A date written for 2076 carries 2076's weekday, which does not match 1976's, so the whole value is thrown away.

Run against `main` and against this branch, in 2026:

| header | before | after |
|---|---|---|
| `Sunday, 15-Jun-70 12:30:45 GMT` | `undefined` | 2070-06-15 |
| `Wednesday, 15-Jun-72 12:30:45 GMT` | `undefined` | 2072-06-15 |
| `Monday, 15-Jun-76 12:30:45 GMT` | `undefined` | 2076-06-15 |
| `Wednesday, 15-Jun-77 12:30:45 GMT` | 1977-06-15 | 1977-06-15 |
| `Wednesday, 15-Jun-94 12:30:45 GMT` | 1994-06-15 | 1994-06-15 |
| `Saturday, 15-Jun-24 12:30:45 GMT` | 2024-06-15 | 2024-06-15 |
| `Wednesday, 15-Jun-50 12:30:45 GMT` | 2050-06-15 | 2050-06-15 |

So an `Expires` a server means for the 2070s is dropped and the response is not stored. Two-digit years 70 through 76 are affected today, and the range grows by one every year: in 2030 it will be 70 through 80. Everything from 1977 to 2069 is untouched.

## The change

Take the century of the current year, then step one century down when the result lands more than 50 years ahead, or one up when it lands more than 49 years back. The window is exactly 100 years wide, so exactly one century fits and no input is left without a reading.

## Tests

Two things moved in `test/utils/date.js`.

The `fuzzing rfc850` round trip pinned `minYear = 1970` / `maxYear = 2069`. Those bounds were the old window written out, so they now derive from the current year: `currentYear - 49` to `currentYear + 50`. Still 1e6 iterations, still green.

Added `RFC850 two-digit years slide with the clock`, covering both edges of the window and the year just past it. The year-just-past case is built from the past date it must resolve to, since `makeDate` validates the weekday against the year it settled on.

Every existing case in the `RFC850` table still passes unchanged, including `06-Nov-94` staying in 1994 and `18-Aug-50` staying in 2050.

`node --test test/utils/date.js`: 7 pass. With the change to `lib/util/date.js` reverted, 2 of those fail and 5 still pass, so the new assertions are the ones doing the work.

`npx eslint` is clean on both files.

### DIFF
--- a/lib/util/date.js
+++ b/lib/util/date.js
@@ -593,14 +593,25 @@ function parseRfc850Date (date) {
 
   let year = (yearDigit1 - 48) * 10 + (yearDigit2 - 48) // Convert ASCII codes to number
 
-  // RFC 6265 states that the year is in the range 1970-2069.
-  // @see https://datatracker.ietf.org/doc/html/rfc6265#section-5.1.1
+  // "Recipients of a timestamp value in rfc850-date format, which uses a
+  // two-digit year, MUST interpret a timestamp that appears to be more than
+  // 50 years in the future as representing the most recent year in the past
+  // that had the same last two digits."
   //
-  // 3. If the year-value is greater than or equal to 70 and less than or
-  //    equal to 99, increment the year-value by 1900.
-  // 4. If the year-value is greater than or equal to 0 and less than or
-  //    equal to 69, increment the year-value by 2000.
-  year += year < 70 ? 2000 : 1900
+  // The window therefore slides with the clock. RFC 6265 pins cookies to a
+  // fixed 1970-2069 range instead, but this parser only serves the HTTP
+  // cache, which reads Date, Expires and Last-Modified.
+  //
+  // @see https://www.rfc-editor.org/rfc/rfc9110.html#section-5.6.7
+  // The window (currentYear - 49 .. currentYear + 50) is exactly 100 years
+  // wide, so exactly one century lands inside it.
+  const currentYear = new Date().getUTCFullYear()
+  year += Math.trunc(currentYear / 100) * 100
+  if (year > currentYear + 50) {
+    year -= 100
+  } else if (year < currentYear - 49) {
+    year += 100
+  }
 
   let hour = 0
   if (date[commaIndex + 12] === '0') {
--- a/test/utils/date.js
+++ b/test/utils/date.js
@@ -277,6 +277,36 @@ describe('parseHttpDate', () => {
   const months = ['Jan', 'Feb', 'Mar', 'Apr', 'May', 'Jun', 'Jul', 'Aug', 'Sep', 'Oct', 'Nov', 'Dec']
   const fuzzingCount = 1e6
 
+  function toRfc850 (date) {
+    return `${daysLong[date.getUTCDay()]}, ` +
+      `${date.getUTCDate().toString().padStart(2, '0')}-` +
+      `${months[date.getUTCMonth()]}-` +
+      `${date.getUTCFullYear().toString().slice(-2)} ` +
+      `${date.getUTCHours().toString().padStart(2, '0')}:` +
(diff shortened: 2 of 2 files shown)


luantaraschi:
In case the red X is what is holding this one, it comes from `test/http2-request-never-settles.js` under `Test with Node.js 25 on ubuntu-latest with WASM SIMD disabled`, and the run lists nothing else as failing.

That test is the flake tracked in #5674, whose root cause turned out to be upstream: the Maglev SIGSEGV in nodejs/node#64841 on Node v24.15.0 and later. The signature here is the same one described there, with no TAP output, no subtests, and the file ending in 3.6s against a 15s watchdog.

The diff is `lib/util/date.js` and its test, so nothing in it can reach an HTTP/2 session. Locally on Node 22.20.0 both files pass on this branch: `test/utils/date.js` 7 passing and `test/http2-request-never-settles.js` 2 passing.

The branch is 6 commits behind `main` and predates #5674, so I am glad to rebase for a fresh run if that would help.

lib/util/date.js
```diff
@@ -593,14 +593,25 @@ function parseRfc850Date (date) {
...
-  //    equal to 69, increment the year-value by 2000.
-  year += year < 70 ? 2000 : 1900
+  // The window therefore slides with the clock. RFC 6265 pins cookies to a
+  // fixed 1970-2069 range instead, but this parser only serves the HTTP
+  // cache, which reads Date, Expires and Last-Modified.
+  //
+  // @see https://www.rfc-editor.org/rfc/rfc9110.html#section-5.6.7
+  // The window (currentYear - 49 .. currentYear + 50) is exactly 100 years
+  // wide, so exactly one century lands inside it.
+  const currentYear = new Date().getUTCFullYear()
```
mcollina, line 608: new Date() is expensive, we should avoid it here

mcollina (changes requested):

luantaraschi:
You are right, and it was worse than I assumed. Measured through `parseHttpDate` itself on Node 22.20, a million parses of `Sunday, 06-Nov-94 08:49:37 GMT`: 219 ns per parse on main, 326 with my version. The clock read was adding 49%.

The year is cached now and revalidated against the precomputed start of the next one, so the common case is a `Date.now()` compare: 268 ns, which gives back about half. The 49 ns left is `Date.now()` itself. I looked for a cached epoch clock to borrow and did not find one; `fastNow` in `lib/util/timers.js` counts up from zero by tick, so it is monotonic rather than wall time.

The window still slides. Stubbing `Date.now()` across a year boundary, a two-digit `77` reads as 1977 with a 2026 clock and 2077 with a 2027 one.

If 268 against 219 is still too much for this path, the other option is dropping the sliding window and keeping the fixed range that main has today. I would rather hear which you prefer than guess.

PR #5672, opened 2026-08-09
types: support exactOptionalPropertyTypes

unrevised6419:
## Problem

Undici's type declarations cannot be consumed from a project compiled with TypeScript's [`exactOptionalPropertyTypes`](https://www.typescriptlang.org/tsconfig/#exactOptionalPropertyTypes). Under that flag, an optional property declared as `foo?: T` accepts a *missing* property but rejects an *explicitly `undefined`* one, so ordinary code fails to compile:

```ts
// error TS2379: Argument of type '{ maxRedirections: number | undefined; }'
// is not assignable to parameter of type 'RequestOptions'
// with 'exactOptionalPropertyTypes: true'.
await request(url, { maxRedirections: config.maxRedirections })
```

This does not reflect runtime behaviour: undici resolves options with `??`/`||` defaults throughout, so a missing option and an `undefined` option behave identically.

## Changes

- Add `| undefined` to the 334 optional properties across `types/*.d.ts`. This is a no-op for readers — reading `foo?: T` already yields `T | undefined` — and only relaxes assignability for writers.
- Convert the optional *methods* affected by the same rule into optional properties holding a function type: the handler hooks on `Dispatcher.DispatchHandler`, plus `Agent.Options.factory`, `Pool.Options.factory`, `RoundRobinPool.Options.factory` and `ProxyAgent.Options.clientFactory`.
- Enable `exactOptionalPropertyTypes` for the `tsd` run and for both `tsc` invocations in `test:typescript`, so the property declarations cannot regress.
- Add `test/types/exact-optional-property-types.test-d.ts`, asserting that an explicit `undefined` is accepted across the main option types (`Client`, `Pool`, `Agent`, `ProxyAgent`, `MockAgent`, `DispatchOptions`, `RequestOptions`, `DispatchHandler`, `RetryOptions`, `RequestInit`, `ResponseInit`, the interceptor options and `request()`).

## Compatibility note

The method-to-property conversion is the one part of this change that can surface a new error in existing code. Method signatures are compared bivariantly, whereas property function types are checked contravariantly under `strictFunctionTypes`, so explicitly annotated callback parameters must now match the declared types.

Callbacks whose parameter types are inferred from context — the form used throughout the documentation, e.g. `factory: (origin, opts) => new Pool(origin, opts)` — are unaffected. One existing type test annotated `factory` as taking a `URL`, while `Agent.Options.factory` is invoked with `opts.origin` (a `string | URL`, see `lib/dispatcher/agent.js`); it has been widened to match.

## Notes

No runtime code is touched; this is a declaration-only change. `npm run test:typescript` and `npm run lint` pass.

### DIFF
--- a/package.json
+++ b/package.json
@@ -91,7 +91,7 @@
     "test:node-test": "borp --timeout 180000 -p \"test/node-test/**/*.js\"",
     "test:tdd": "borp --timeout 180000 --expose-gc -p \"test/*.js\"",
     "test:tdd:node-test": "borp --timeout 180000 -p \"test/node-test/**/*.js\" -w",
-    "test:typescript": "tsd && tsc test/imports/undici-import.ts --typeRoots ./types --noEmit && tsc ./types/*.d.ts --noEmit --typeRoots ./types",
+    "test:typescript": "tsd && tsc test/imports/undici-import.ts --typeRoots ./types --noEmit --exactOptionalPropertyTypes && tsc ./types/*.d.ts --noEmit --typeRoots ./types --exactOptionalPropertyTypes",
     "test:webidl": "borp --timeout 180000 -p \"test/webidl/*.js\"",
     "test:websocket": "borp --timeout 180000 -p \"test/websocket/**/*.js\"",
     "test:websocket:autobahn": "node test/autobahn/client.js",
@@ -139,6 +139,7 @@
     "directory": "test/types",
     "compilerOptions": {
       "esModuleInterop": true,
+      "exactOptionalPropertyTypes": true,
       "lib": [
         "esnext"
       ]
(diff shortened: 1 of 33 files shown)


unrevised6419:
Would a backport of this to `v6.x` (and `v7.x`) be welcome?

The motivation is that `exactOptionalPropertyTypes` is a compile-time-only concern for consumers, so users on the older lines hit exactly the same problem, and they cannot pick up the fix by upgrading if they are pinned for other reasons.

Two things worth deciding before I open anything:

- The bulk of the diff — adding `| undefined` to optional properties — is purely additive for consumers and should be safe on a patch release.
- The method-to-property conversion described above is the part that can introduce a new compile error in existing code, via the bivariance change. That is arguably too disruptive for a patch on a maintenance line. I am happy to split the backport so it carries only the property changes, or to drop the backport entirely if you would rather keep the maintenance branches frozen.

Let me know which you prefer and I will prepare the branches.

PR #5675, opened 2026-08-09
fix(retry): honor the abort signal during the retry backoff

pacocartones:
## This relates to...

No open issue found for this; reported directly with a reproduction.

## Rationale

When a retry is scheduled, the default retry policy parks the decision in a
`setTimeout` backoff (`lib/handler/retry-handler.js`). **An abort that arrives
while that backoff is pending is ignored until the timer elapses.** The
consumer's abort reaches the controller of the connection that already
finished — a no-op — and the pending timer is not tracked anywhere, so the
request only rejects with the `AbortError` after the full backoff: 5 s in the
repro below, and minutes with a large `Retry-After` (e.g. `Retry-After: 120`
holds the caller for two minutes). The referenced (non-`unref`) timer also
keeps the process alive for the whole backoff even though nothing will come of
it.

Reproduced against current `main` (86b62998): a server that always answers 500,
a `RetryAgent` with `minTimeout: 5000`, and `abort()` fired 800 ms into the
request. The rejection arrives at ~5038 ms instead of ~800 ms.

## Changes

- The default retry policy now **returns its backoff timer**, and the handler
  tracks it while a retry decision is pending (`retryPending` / `retryTimer`).
- The `RetryController` proxy — where every consumer abort already funnels
  through — now also reports `abort(reason)` to the handler. `onAbort` cancels
  the pending wait and forwards the abort reason downstream via
  `onResponseError` immediately (a `RequestAbortedError` when no reason was
  given, matching `api-request`'s convention).
- Guards make the termination exact: a late decision callback from the policy
  (e.g. a custom retry function resolving after the abort) is ignored, and the
  torn-down connection's late error is not forwarded twice, so the downstream
  handler is errored exactly once.

Both policy call sites (the `throwOnError: false` response path and the
`onResponseError` path used by `throwOnError: true` and network errors) track
the pending decision the same way.

Scope note: a custom retry function that keeps its own timer and does not
return it cannot have that timer cancelled (we were never handed it), but the
abort still settles the request immediately and the late callback is ignored.

Out of scope (deliberately not changed): `unref()`-ing the backoff timer. A
pending retry is a live logical request; letting the process exit mid-backoff
would silently abandon it. That idea is being explored in #4470, which is
orthogonal to this fix.

### Features

N/A

### Bug Fixes

- `retry()` / `RetryAgent`: an abort that arrives while a retry backoff is
  pending now rejects immediately with the `AbortError` instead of after the
  full backoff, and no retry is dispatched after the abort.

None.

## Verification

Every command below was run and its output captured verbatim. The record is
reproducible — the exact commands are included so you can re-run them yourself.

**red — main (86b62998) without the fix** (must fail: the rejection only comes
after the full backoff)

```
$ git stash push -- lib/handler/retry-handler.js   # i.e. main's version
$ node --test --test-name-pattern="backoff" test/retry-handler.js
✖ Should abort immediately when aborted during the retry backoff (10028.7502ms)
  AssertionError [ERR_ASSERTION]: rejects long before the 10s backoff elapses
✖ Should ignore a late retry decision after an abort during the backoff (3018.9959ms)
  AssertionError [ERR_ASSERTION]: rejects long before the 3s custom backoff
```

(The abort is fired while the backoff is pending — the test hooks the retry
policy invocation — yet the request only settles once the 10 s / 3 s timers
elapse.)

**green — with the fix** (must pass)

```
$ node --test --test-name-pattern="backoff" test/retry-handler.js
✔ Should abort immediately when aborted during the retry backoff (28.0084ms)
✔ Should ignore a late retry decision after an abort during the backoff (3516.7355ms)
ℹ tests 2
ℹ pass 2
ℹ fail 0
```

(The second test's ~3.5 s are deliberate: after the immediate rejection it
waits past the custom policy's 3 s timer to assert the late decision is
ignored and no retry is dispatched. The assertions are relational — abort
during backoff → immediate rejection with the abort reason — with margins of
two orders of magnitude, so CI slowness does not flake them.)

**standalone reproduction against a live server** (`RetryAgent` with
`minTimeout: 5000`, server always 500, `abort()` at 800 ms):

```
# main:
[server] hit #1 at 19ms
[client] abort() at 803ms
[client] rejected at 5038ms: name=AbortError code=20 msg=This operation was aborted
[client] cleanup done at 5039ms        # process held by the pending timer

# with the fix:
[server] hit #1 at 19ms
[client] abort() at 813ms
[client] rejected at 814ms: name=AbortError code=20 msg=This operation was aborted
[client] cleanup done at 816ms         # timer cancelled, process exits

# control (no abort), main and fixed alike:
[server] hit #1 at 19ms
[server] hit #2 at ~5040ms
[client] resolved status=200 at ~5042ms
```

**area suites with the fix** (must pass)

```
$ node --test test/retry-handler.js test/retry-handler2.js test/retry-agent.js \
       test/retry-handler-controller-proxy.js test/interceptors/retry.js \
       test/client-retry-resume-backpressure.js
ℹ tests 68  ℹ pass 68  ℹ fail 0

$ node --test test/retry-handler.js        # full file, incl. the 2 new tests
ℹ tests 23  ℹ pass 23  ℹ fail 0

$ node --test test/http2-goaway-retry-body.js
ℹ tests 1  ℹ pass 1  ℹ fail 0
```

**lint** (must pass)

```
$ npm run lint
> undici@8.10.0 lint
> eslint --cache
[exit code 0]
```

Not verified locally: the remaining test suites (`npm test` runs 14 of them);
left to CI, as this change only touches the retry handler and its tests. No
documentation changes: no existing doc describes abort-during-backoff
behavior, and the fixed behavior now matches the general abort contract;
happy to add a note to `docs/docs/api/RetryHandler.md` if maintainers want one.

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

---

Implementation and validation used AI assistance; I reviewed the final diff and results.

### DIFF
--- a/lib/handler/retry-handler.js
+++ b/lib/handler/retry-handler.js
@@ -43,9 +43,13 @@ function validatePartialResponseContentLength (headers, range, statusCode, retry
 // The proxy always forwards to the controller of the currently active connection.
 // An abort is additionally reported to the handler so it can cancel a pending
 // retry backoff instead of letting the request hang until the backoff elapses.
+// The notification is a private callback the handler hands over on construction,
+// so nothing outside the handler can trigger it.
 class RetryController {
-  constructor (handler) {
-    this.handler = handler
+  #onAbort
+
+  constructor (onAbort) {
+    this.#onAbort = onAbort
     this.target = null
   }
 
@@ -54,7 +58,7 @@ class RetryController {
 
   abort (reason) {
     this.target?.abort(reason)
-    this.handler.onAbort(reason)
+    this.#onAbort(reason)
   }
 
   get paused () { return this.target?.paused ?? false }
@@ -120,11 +124,11 @@ class RetryHandler {
     this.etag = null
     this.statusCode = null
     this.headers = null
-    this.controllerProxy = new RetryController(this)
+    this.controllerProxy = new RetryController(reason => this.#onAbort(reason))
     // A retry decision is in flight (the policy may be holding a backoff
     // timer). While pending, a consumer abort cancels the wait.
     this.retryPending = false
-    // Backoff timer returned by the retry policy, so onAbort can cancel it.
+    // Backoff timer returned by the retry policy, so #onAbort can cancel it.
     // Null for custom policies that do not return their timer.
     this.retryTimer = null
     // Set once an abort during the backoff delivered the terminal error
@@ -465,7 +469,7 @@ class RetryHandler {
 
   onResponseError (controller, err) {
     if (this.aborted) {
-      // onAbort already delivered the terminal error downstream; the late
+      // #onAbort already delivered the terminal error downstream; the late
       // error of the torn-down connection must not be forwarded twice.
       return
     }
@@ -515,7 +519,7 @@ class RetryHandler {
     ) ?? null
   }
 
-  onAbort (reason) {
+  #onAbort (reason) {
     // A consumer abort lands on the controller proxy. If the retry policy is
     // still deciding (typically holding a backoff timer), cancel the wait and
     // surface the abort immediately instead of letting the request hang until
--- a/lib/handler/retry-handler.js
+++ b/lib/handler/retry-handler.js
@@ -2,7 +2,7 @@
 const assert = require('node:assert')
 
 const { kRetryHandlerDefaultRetry } = require('../core/symbols')
-const { RequestRetryError } = require('../core/errors')
+const { RequestRetryError, RequestAbortedError } = require('../core/errors')
 const {
   isDisturbed,
   parseRangeHeader,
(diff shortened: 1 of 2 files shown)


pacocartones:

lib/handler/retry-handler.js
```diff
@@ -462,14 +504,31 @@ class RetryHandler {
...
       {
         state: { counter: this.retryCount },
         opts: { retryOptions: this.retryOpts, ...this.opts }
       },
       shouldRetry.bind(this)
-    )
+    ) ?? null
+  }
+
+  onAbort (reason) {
```
metcoder95: Let's make it callable only within the handler itself, not outside should have influence

pacocartones: Done in 42516a41.

`onAbort` is now a private method (`#onAbort`), and `RetryController` no longer holds a reference to the handler at all — it takes the notification as a private callback at construction. The only way to reach it is the proxy's own `abort()`, which is the path a consumer abort already takes. Same shape as `#onComplete` in `lib/handler/deduplication-handler.js`.

`handler.onAbort`, `proxy.onAbort` and `proxy.handler` are all `undefined` now.

No behaviour change, same call path: `test/retry-handler.js` 23/23 (including the two backoff-abort tests), `retry-handler2` / `retry-agent` / `retry-handler-controller-proxy` / `interceptors/retry` / `client-retry-resume-backpressure` / `http2-goaway-retry-body` 46/46, `npm run lint` exit 0.

metcoder95 (approved):

PR #5676, opened 2026-08-09
fix(h2): honour maxResponseSize

ondraulehla:
## This relates to...

Nothing filed. I noticed it while reading the two client implementations side by side after #5638.

## Rationale

`maxResponseSize` is documented on the HTTP/2 client's own page, `docs/docs/api/H2CClient.md`:

> `maxResponseSize` {number} The maximum allowed response body size in bytes. Use `-1` to disable. Default: `-1`.

It is declared in `types/h2c-client.d.ts` and validated in `lib/dispatcher/client.js` (an `H2CClient` with `maxResponseSize: -5` throws `InvalidArgumentError`), then stored into `client[kMaxResponseSize]`. `lib/dispatcher/client-h2.js` never reads that symbol, so the value is validated and then ignored. `onData()` forwards every chunk to `request.onResponseData(chunk)` with no byte accounting.

Server sends 4 MiB, client asks for `maxResponseSize: 1024`:

| | HTTP/1.1 `Client` | h2c `H2CClient` |
|---|---|---|
| before | `UND_ERR_RES_EXCEEDED_MAX_SIZE` | resolves with 4194304 bytes |
| after | `UND_ERR_RES_EXCEEDED_MAX_SIZE` | `UND_ERR_RES_EXCEEDED_MAX_SIZE` |

Same split for a streaming consumer: over h1 the body errors mid-stream, over h2 it just completes.

The enforcement path itself isn't a regression, h2 has never honoured the option: `git log -S kMaxResponseSize -- lib/dispatcher/client-h2.js` is empty, and 7.29.0 ignores it too once you pass `allowH2: true`.

What did change is what an ordinary caller gets. Against a TLS origin offering h2, with h2 mentioned nowhere in the code:

```js
const agent = new Agent({ maxResponseSize: 1024 })
await request(url, { dispatcher: agent })
```

| | 7.29.0 | 8.10.0 |
|---|---|---|
| plain `new Agent({ maxResponseSize })` | `UND_ERR_RES_EXCEEDED_MAX_SIZE` | resolves, 4194304 bytes |
| `connect: { allowH2: false }` | `UND_ERR_RES_EXCEEDED_MAX_SIZE` | `UND_ERR_RES_EXCEEDED_MAX_SIZE` |

That's #4828 making `allowH2` default to `true`. Opting out of h2 brings the limit back, which is the clearest sign of where it goes missing.

## Changes

`onData()` counts bytes against `maxResponseSize`, mirroring `onBody()` in `client-h1.js`. The counter and the limit live on the pre-shaped per-request `state` object, so the hidden class is unchanged.

One deliberate difference from h1, and I'd like a second opinion on it. `client-h1.js` calls `util.destroy(socket, ...)` because HTTP/1.1 can't abandon one response without losing framing. Over h2 that would take out every sibling stream on the session, so this calls the existing per-request `state.abort()` instead, which resets only the offending stream. I verified the session survives: after an oversized stream is reset, later requests on the same client still succeed and the server reports one connection.

## Tests

Two cases added to `test/max-response-size.js`, which was http/1.1 only. Both fail before the change (the oversized response resolves instead of throwing) and pass after:

- an h2c response over the limit rejects with `ResponseExceededMaxSizeError`
- the session stays usable, an oversized stream errors, a following request on the same client succeeds, and the server sees one connection

`test/max-response-size.js`: 6 pass. `test/http2*.js test/h2c-client.js`: 105 pass, 0 fail. The default (`-1`) path is unchanged, a 4 MiB body is still delivered in full with no limit set and with `maxResponseSize: -1`.

### DIFF
--- a/lib/dispatcher/client-h2.js
+++ b/lib/dispatcher/client-h2.js
@@ -10,7 +10,8 @@ const {
   InformationalError,
   InvalidArgumentError,
   HeadersTimeoutError,
-  BodyTimeoutError
+  BodyTimeoutError,
+  ResponseExceededMaxSizeError
 } = require('../core/errors.js')
 const {
   kUrl,
@@ -39,7 +40,8 @@ const {
   kRemoteSettings,
   kHTTP2Stream,
   kHTTP2SessionState,
-  kHTTP2Options
+  kHTTP2Options,
+  kMaxResponseSize
 } = require('../core/symbols.js')
 const { channels } = require('../core/diagnostics.js')
 
@@ -1022,9 +1024,11 @@ function writeH2 (client, request) {
   const state = {
     abort: null,
     body: request.body,
+    bytesRead: 0,
     client,
     contentLength: null,
     expectsPayload: false,
+    maxResponseSize: client[kMaxResponseSize],
     request,
     headersTimeout,
     bodyTimeout,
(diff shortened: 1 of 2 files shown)


metcoder95 (approved):

mcollina (approved):
lgtm

PR #5685, opened 2026-08-12
fix(interceptor/dump): reject aborted requests when response is smaller than maxSize

### LINKED ISSUE #5684: dump interceptor: aborted request hangs forever when response body is smaller than maxSize
## Bug Description

While doing a code correctness review of the interceptors, I noticed that `interceptors.dump()` checks the wrong "aborted" flag in `onResponseEnd`. As a result, an aborted request whose response body is **smaller** than `maxSize` never settles: the `request()` promise hangs forever (no resolve, no reject). The same abort with a body **larger** than `maxSize` correctly rejects with `AbortError`.

Since `maxSize` defaults to 1 MiB and most real-world responses are smaller than that, the hang is the common case for aborted requests under this interceptor, not the exception.

### Root cause

`DumpHandler.onRequestStart` replaces the controller's `abort` method with its own (`lib/interceptor/dump.js:31`):

```js
controller.abort = this.#abort.bind(this)
```

`#abort` only sets the handler-local `this.aborted = true`. The underlying `RequestController`'s private `#aborted` field is only ever set inside its original `abort()` method (`lib/core/request.js:76-86`), which can no longer be reached once the instance property shadows it. So after the hijack, `controller.aborted` can never become `true`.

`onResponseStart` (line 48) and `onResponseData` (line 72) correctly check the handler-local `this.aborted`, but `onResponseEnd` (line 87) checks `this.#controller.aborted`:

```js
onResponseEnd (controller, trailers) {
  if (this.#dumped) {
    return
  }

  if (this.#controller.aborted === true) {   // <-- always false, dead condition
    super.onResponseError(controller, this.reason)
    return
  }

  super.onResponseEnd(controller, trailers)
}
```

Failure chain for "abort before response, body < maxSize":

1. abort → `RequestHandler` calls `controller.abort(reason)` → hijacked method → only `DumpHandler.aborted = true`
2. response arrives → `onResponseStart` swallows it (line 48-50, `return true`), so the downstream `RequestHandler.onResponseStart` is never invoked and the promise stays pending
3. body (100 B) < `maxSize` (512 B) → the correct abort branch in `onResponseData` (line 69-77) is never reached
4. `onResponseEnd` → dead condition is `false` → forwards a normal `onResponseEnd` to a downstream handler that never saw `onResponseStart`; `this.res?.push(null)` is a no-op because `res` is `null`
5. the callback is never invoked → the promise never settles

The existing test `test/interceptors/dump-interceptor.js:141` ("Should dump on already aborted request") asserts that an aborted request must reject with `AbortError` — it only passes because its body (1024 B) happens to exceed `maxSize` (512 B), hitting the correct check in `onResponseData` instead.

## Reproduction

```javascript
'use strict'

const { Client, interceptors } = require('undici')

test('aborted request whose response body is smaller than maxSize should reject', { timeout: 10000 }, async (t) => {
  t.plan(1)

    res.writeHead(200, { 'Content-Type': 'application/octet-stream' })
    res.end(Buffer.alloc(100)) // 100 bytes < maxSize (512)
  })
  t.after(() => server.close())

  server.listen(0)

  const client = new Client(`http://localhost:${server.address().port}`).compose(
    interceptors.dump({ maxSize: 512 })
  )
  t.after(() => client.destroy())

  const abc = new AbortController()
  abc.abort()

  await t.assert.rejects(
    client.request({ method: 'GET', path: '/', signal: abc.signal }),
    { name: 'AbortError' }
  )
})
```

Output on current `main` (2afeee4):

```
not ok 1 - aborted request whose response body is smaller than maxSize should reject
  failureType: 'testTimeoutFailure'
  error: 'test timed out after 10000ms'
```

Changing `Buffer.alloc(100)` to `Buffer.alloc(1024)` (larger than `maxSize`) makes the same test pass, confirming the inconsistency: the outcome of an abort depends on the response body size.

## Expected Behavior

The aborted request rejects with `AbortError`, exactly like the `body > maxSize` case and like the existing "Should dump on already aborted request" test asserts.

## Actual Behavior

The `request()` promise never settles. Callers without their own timeout wait forever.

## Logs & Screenshots

See reproduction output above.

## Environment

- OS: macOS (darwin 25.5.0)
- Node.js version: v22.22.3
- undici version: main @ 2afeee468a975071c472ea34bc4594642208e576

### Additional context

- Suggested minimal fix: in `onResponseEnd`, check the handler-local flag like the other callbacks do, i.e. replace `this.#controller.aborted === true` with `this.aborted === true`. I can send a PR with this one-line fix plus a regression test.
- Related: #4634 restructures the dump handler lifecycle, but its diff keeps the `this.#controller.aborted` check in `onResponseEnd`, so it does not address this hang (and the broken abort semantics may be contributing to its CI failures).

_This report was prepared with AI assistance; I reproduced the issue locally and reviewed every conclusion._

codeAnqiang-ma:
## This relates to...

Fixes #5684

## Rationale

`DumpHandler.onRequestStart` replaces `controller.abort` with its own method (`lib/interceptor/dump.js:31`), which only sets the handler-local `this.aborted`. The underlying `RequestController`'s `#aborted` field is only set inside its original `abort()` method (`lib/core/request.js:76-86`), which is shadowed by the hijack — so `controller.aborted` can never become `true` afterwards.

`onResponseStart` (line 48) and `onResponseData` (line 72) correctly check the handler-local `this.aborted`, but `onResponseEnd` (line 87) checked `this.#controller.aborted`, a condition that is always `false`.

Consequence: for an aborted request whose response body is smaller than `maxSize`, `onResponseStart` is swallowed and `onResponseEnd` forwards a normal end to a downstream handler that never saw `onResponseStart`. The `request()` promise never settles — it hangs forever. With a body larger than `maxSize`, the same abort correctly rejects with `AbortError` via the check in `onResponseData`, so the outcome of an abort currently depends on the response body size.

The existing test "Should dump on already aborted request" asserts the expected semantics (aborted request → `AbortError`), but only exercises the `body > maxSize` path.

## Changes

- `lib/interceptor/dump.js`: in `onResponseEnd`, check the handler-local `this.aborted` (consistent with `onResponseStart` and `onResponseData`) instead of the never-true `this.#controller.aborted`.
- `test/interceptors/dump-interceptor.js`: add a regression test mirroring "Should dump on already aborted request" with a response body (256 B) smaller than `maxSize` (512 B). Without the fix the test times out (the promise never settles); with the fix it rejects with `AbortError`.

### Features

N/A

### Bug Fixes

- Aborted requests through `interceptors.dump()` now reject with `AbortError` regardless of response body size, instead of hanging forever when the body is smaller than `maxSize`.

N/A

## Test evidence

Regression test without the fix (lib change stashed):

```
node --test --test-timeout=8000 --test-name-pattern="smaller than maxSize" test/interceptors/dump-interceptor.js
  failureType: 'testTimeoutFailure'
  error: 'test timed out after 8000ms'
```

With the fix:

```
node --test test/interceptors/dump-interceptor.js
# tests 12
# pass 12
# fail 0
```

Full `test/interceptors/` suite with the fix: all 14 files pass (cache 87/87, decompress 27/27, deduplicate 34/34, dns 30 pass + 3 skip, dump 12/12, redirect 83/83, retry 16/16, response-error 6/6, ...). `npm run lint` passes.

Note: #4634 restructures the dump handler lifecycle but keeps the `this.#controller.aborted` check in `onResponseEnd`, so it does not address this hang. This one-line fix is independent and can land regardless.

_This fix was prepared with AI assistance; I reproduced the bug locally and reviewed every change._

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/interceptor/dump.js
+++ b/lib/interceptor/dump.js
@@ -84,7 +84,7 @@ class DumpHandler extends DecoratorHandler {
       return
     }
 
-    if (this.#controller.aborted === true) {
+    if (this.aborted === true) {
       super.onResponseError(controller, this.reason)
       return
     }
--- a/test/interceptors/dump-interceptor.js
+++ b/test/interceptors/dump-interceptor.js
@@ -199,6 +199,47 @@ test('Should dump on already aborted request', { skip }, async t => {
   await t.completed
 })
 
+test('Should reject aborted request when response is smaller than maxSize', { skip }, async t => {
+  t = tspl(t, { plan: 2 })
+  const server = createServer({ joinDuplicateHeaders: true }, (req, res) => {
+    res.writeHead(200, {
+      'Content-Type': 'application/octet-stream'
+    })
+    res.end(Buffer.alloc(256))
+  })
+
+  const abc = new AbortController()
+
+  const requestOptions = {
+    method: 'GET',
+    path: '/',
+    signal: abc.signal
+  }
+
+  server.listen(0)
+
+  await once(server, 'listening')
+
+  const client = new Client(
+    `http://localhost:${server.address().port}`
+  ).compose(dump({ maxSize: 512 }))
+
+  after(async () => {
+    await client.close()
+
+    server.close()
+    await once(server, 'close')
+  })
+
+  abc.abort()
+  client.request(requestOptions).catch(err => {
+    t.equal(err.name, 'AbortError')
+    t.equal(err.message, 'This operation was aborted')
+  })
+
+  await t.completed
+})
+
 test('Should dump response body up to limit (default)', { skip }, async t => {
   t = tspl(t, { plan: 3 })
   const server = createServer({ joinDuplicateHeaders: true }, (req, res) => {


mcollina (approved):
lgtm

metcoder95 (approved):

PR #5696, opened 2026-08-17
fix: forward onBodySent/onRequestSent through interceptor handlers

### LINKED ISSUE #5695: DecoratorHandler and CacheHandler do not forward onBodySent/onRequestSent to the wrapped handler
## Bug Description

`onBodySent` / `onRequestSent` are not forwarded to the wrapped handler by `DecoratorHandler`
or by the built-in `CacheHandler`. Because every built-in interceptor either extends
`DecoratorHandler` (`RedirectHandler`, `RetryHandler`, …) or hand-forwards the handler
methods (`CacheHandler`), composing *any* interceptor onto a dispatcher silently disables
these two hooks for the user's handler.

In `lib/handler/decorator-handler.js` (v8.10.0):

- `onBodySent () {}` is declared as a **no-op** rather than forwarding to `this.#handler`.
  This is worse than omitting it: `Request` sees a truthy `handler.onBodySent`, calls it,
  and the chunk is swallowed with no error.
- `onRequestSent` is not declared at all, so `Request#onRequestSent()` skips it.

In `lib/handler/cache-handler.js` (v8.10.0), `CacheHandler` forwards `onRequestStart`,
`onRequestUpgrade`, `onResponseStart`, `onResponseData`, `onResponseEnd`, and
`onResponseError`, but has no `onBodySent` / `onRequestSent`, so `interceptors.cache()`
drops them too.

These hooks were restored in 8.10.0 (including for `MockAgent`) by #5367, and they work
correctly when a handler is passed straight to `Client#dispatch` — the decorator/interceptor
path was just missed.

## Reproduction

```javascript
'use strict'

const { Client, DecoratorHandler, interceptors } = require('undici')

const BODY = '{"hello":"world"}'

function dispatchAndTrack (dispatcher) {
  const seen = { bodySent: [], requestSent: 0 }
    dispatcher.dispatch(
      {
        method: 'POST',
        path: '/',
        headers: { 'content-type': 'application/json' },
        body: BODY
      },
      {
        onRequestStart () {},
        onBodySent (chunk) { seen.bodySent.push(Buffer.from(chunk).toString()) },
        onRequestSent () { seen.requestSent++ },
        onResponseStart () {},
        onResponseData () {},
        onResponseEnd () { resolve(seen) },
        onResponseError (_controller, err) { reject(err) }
      }
    )
  })
}

test('onBodySent/onRequestSent are dropped by DecoratorHandler and cache interceptor', { timeout: 60000 }, async (t) => {
    req.resume()
    req.on('end', () => res.end('ok'))
  })
  server.listen(0)
  t.after(() => server.close())

  t.after(() => client.close())

  // 1. Baseline: handler passed straight to the client — hooks fire.
  const direct = await dispatchAndTrack(client)
  assert.deepStrictEqual(direct.bodySent, [BODY])
  assert.strictEqual(direct.requestSent, 1)

  // 2. Handler wrapped in DecoratorHandler (what every built-in interceptor does).
  const decorated = client.compose(
    (dispatch) => (opts, handler) => dispatch(opts, new DecoratorHandler(handler))
  )
  const throughDecorator = await dispatchAndTrack(decorated)
  assert.deepStrictEqual(throughDecorator.bodySent, [BODY]) // ❌ actual: []
  assert.strictEqual(throughDecorator.requestSent, 1)       // ❌ actual: 0

  // 3. Same thing via a public interceptor (RetryHandler extends DecoratorHandler).
  const retrying = client.compose(interceptors.retry())
  const throughRetry = await dispatchAndTrack(retrying)
  assert.deepStrictEqual(throughRetry.bodySent, [BODY])     // ❌ actual: []
  assert.strictEqual(throughRetry.requestSent, 1)           // ❌ actual: 0
})
```

Run with:

```bash
node --test repro.js
```

## Expected Behavior

Wrapping a handler in `DecoratorHandler` (directly or via any built-in interceptor such as
`interceptors.retry()`, `interceptors.redirect()`, or `interceptors.cache()`) should be
transparent: `onBodySent` and `onRequestSent` should reach the wrapped handler exactly as
they do when the handler is passed straight to `Client#dispatch`.

## Actual Behavior

The wrapped handler's `onBodySent` is never called (`DecoratorHandler`'s no-op absorbs it)
and `onRequestSent` is never called (not implemented). No error or warning is emitted — the
hooks just silently stop firing as soon as any interceptor is composed onto the dispatcher.

## Logs & Screenshots

Assertion output from step 2 of the script above:

```
AssertionError [ERR_ASSERTION]: Expected values to be strictly deep-equal:
+ actual - expected

+ []
- [
-   '{"hello":"world"}'
- ]
```

## Environment

- OS: 
- Node.js version: 
- undici version: 8.10.0

### Additional context

We hit this with a request/response logging handler that buffers the outgoing body from
`onBodySent`. It worked against a bare `Client`, but request bodies came back empty for every
client that also composed `interceptors.cache()`.

Note that `onBodySent` on `DecoratorHandler` is still marked `@deprecated`. If the intent is
to drop these hooks entirely, it would be useful to have a documented replacement for
observing the outgoing request body from a handler, since `onRequestStart` doesn't expose it.

adarshx01:
Fixes #5695.

`Request` only calls `onBodySent` / `onRequestSent` when those methods exist on the handler. `DecoratorHandler` defined `onBodySent` as a no-op, so the chunk was swallowed, and `onRequestSent` was missing entirely. `RetryHandler`, `RedirectHandler`, and `CacheHandler` wrap the user handler themselves and never forwarded either hook.

Composing `interceptors.retry()`, `redirect()`, `cache()`, or anything that extends `DecoratorHandler` therefore silently dropped both callbacks. They now pass through the same way `onRequestStart` already does.

## Test plan
- [x] `node --test test/decorator-handler.js test/interceptors/body-sent-hooks.js`
- [x] Bare `Client` still fires both hooks
- [x] Same after `DecoratorHandler`, `interceptors.retry()`, `interceptors.cache()`, and `interceptors.redirect()`

Made with [Cursor](https://cursor.com)

### DIFF
--- a/lib/handler/cache-handler.js
+++ b/lib/handler/cache-handler.js
@@ -173,6 +173,14 @@ class CacheHandler {
     this.#handler.onRequestStart?.(controller, context)
   }
 
+  onBodySent (chunk) {
+    this.#handler.onBodySent?.(chunk)
+  }
+
+  onRequestSent () {
+    this.#handler.onRequestSent?.()
+  }
+
   onRequestUpgrade (controller, statusCode, headers, socket) {
     this.#handler.onRequestUpgrade?.(controller, statusCode, headers, socket)
   }
--- a/lib/handler/decorator-handler.js
+++ b/lib/handler/decorator-handler.js
@@ -62,5 +62,11 @@ module.exports = class DecoratorHandler {
   /**
    * @deprecated
    */
-  onBodySent () {}
+  onBodySent (...args) {
+    return this.#handler.onBodySent?.(...args)
+  }
+
+  onRequestSent (...args) {
+    return this.#handler.onRequestSent?.(...args)
+  }
 }
--- a/lib/handler/redirect-handler.js
+++ b/lib/handler/redirect-handler.js
@@ -43,6 +43,14 @@ class RedirectHandler {
     this.handler.onRequestStart?.(controller, { ...context, history: this.history })
   }
 
+  onBodySent (chunk) {
+    this.handler.onBodySent?.(chunk)
+  }
+
+  onRequestSent () {
+    this.handler.onRequestSent?.()
+  }
+
   onRequestUpgrade (controller, statusCode, headers, socket) {
     this.handler.onRequestUpgrade?.(controller, statusCode, headers, socket)
   }
--- a/lib/handler/retry-handler.js
+++ b/lib/handler/retry-handler.js
@@ -207,6 +207,14 @@ class RetryHandler {
     }
   }
 
+  onBodySent (chunk) {
+    this.handler.onBodySent?.(chunk)
+  }
+
+  onRequestSent () {
+    this.handler.onRequestSent?.()
+  }
+
   onRequestUpgrade (_controller, statusCode, headers, socket) {
     this.handler.onRequestUpgrade?.(this.controllerProxy, statusCode, headers, socket)
   }
--- a/test/decorator-handler.js
+++ b/test/decorator-handler.js
@@ -191,6 +191,42 @@ describe('DecoratorHandler', () => {
       })
     })
 
+    describe('#onBodySent', () => {
+      test('should delegate onBodySent-method', t => {
+        t = tspl(t, { plan: 1 })
+        const decorator = new DecoratorHandler({
+          onBodySent: (chunk) => {
+            t.equal(chunk, 'chunk')
+          }
+        })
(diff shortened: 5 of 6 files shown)


mcollina (approved):
lgtm

metcoder95 (approved):

PR #5699, opened 2026-08-18
perf: improve Headers class performance

anonrig:
## This relates to...

N/A

## Rationale

`Headers` is on the fetch / Request / Response hot path. The current implementation pays for WebIDL conversion, token validation, and `toLowerCase()` as separate passes, and stores every entry as a `{ name, value }` object.

## Changes

- Combine ByteString conversion, HTTP token validation, and lowercasing into a single pass.
- Fast-path common lowercase names (`content-type`, `accept`, `authorization`, …) on `get` / `has` / `delete`.
- Store values as strings in the internal `Map` instead of `{ name, value }` objects. Original casing is kept in a side table only when it differs from the lowercase name.
- Copy an existing `Headers` instance by cloning the list instead of re-validating every entry.
- Keep the existing fetch spec comments next to the new hot-path code. No changes outside `lib/web/fetch/headers.js`.

No intended behavioral change. Guard handling, `set-cookie` / `getSetCookie()`, iteration order, and original-name casing on the wire are preserved.

### Features

N/A

### Bug Fixes

N/A

N/A

## Benchmark

Measured with the source `Headers` class (Node v22.14.0, 200k iterations, paired old vs new):

| Operation | Before | After | Change |
| --- | ---: | ---: | ---: |
| `get` (custom name) | 18.4M ops/s | 77.9M ops/s | **4.2x** |
| `get` (`content-type`) | 17.6M ops/s | 56.2M ops/s | **3.2x** |
| `has` | 16.7M ops/s | 80.7M ops/s | **4.8x** |
| `set` | 15.4M ops/s | 25.2M ops/s | **+64%** |
| `new Headers(headers)` | 1.12M ops/s | 1.72M ops/s | **+54%** |
| `new Headers()` | 23.7M ops/s | 23.3M ops/s | ~same |
| `append` | 4.36M ops/s | 4.33M ops/s | ~same |
| iterate | 5.05M ops/s | 5.15M ops/s | ~same |

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

Tested locally:

```
node --test test/fetch/headers.js test/fetch/headers-inspect-custom.js test/fetch/headerslist-sortedarray.js test/fetch/headers-case.js
```

All 70 tests passed, including the 4000-iteration `toSortedArray` fast/slow paths and the fetch cloning cases.

### DIFF
--- a/benchmarks/fetch/headers-length32.mjs
+++ b/benchmarks/fetch/headers-length32.mjs
@@ -40,12 +40,8 @@ const headers = new Headers(
 
 const headersList = getHeadersList(headers)
 
-const kHeadersSortedMap = Reflect.ownKeys(headersList).find(
-  (c) => String(c) === 'Symbol(headers map sorted)'
-)
-
 bench('Headers@@iterator', () => {
-  headersList[kHeadersSortedMap] = null
+  headersList.sortedMap = null
   return [...headers]
 })
 
--- a/benchmarks/fetch/headers-methods.mjs
+++ b/benchmarks/fetch/headers-methods.mjs
@@ -1,0 +1,70 @@
+import { bench, group, run } from 'mitata'
+import { Headers } from '../../lib/web/fetch/headers.js'
+
+const objectInit = {
+  Accept: 'application/json',
+  'Content-Type': 'text/plain',
+  'User-Agent': 'benchmark',
+  Authorization: 'Bearer token',
+  Cookie: 'a=1',
+  'X-Request-Id': 'abc',
+  'Cache-Control': 'no-cache',
+  Host: 'example.com'
+}
+
+const headers = new Headers(objectInit)
+const copySource = new Headers(objectInit)
+
+group('Headers methods', () => {
+  bench('construct empty', () => {
+    return new Headers()
+  })
+
+  bench('construct object', () => {
+    return new Headers(objectInit)
+  })
+
+  bench('construct headers', () => {
+    return new Headers(copySource)
+  })
+
+  bench('get custom', () => {
+    return headers.get('x-request-id')
+  })
+
+  bench('get common', () => {
+    return headers.get('content-type')
+  })
+
+  bench('has', () => {
+    return headers.has('authorization')
+  })
+
+  bench('set', () => {
+    headers.set('x-count', '1')
+    return headers
+  })
+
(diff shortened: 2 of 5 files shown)


mcollina (changes requested):
Adding the benchmarks somewhere will be good.

Also, lint is failing.


@KhafraDev I'm generically ok with these improvements. However it's quite substantial in the code changes. Take a look.

anonrig:
Addressed the review feedback:

- Lint: removed unused `appendHeader` (`no-unused-vars`).
- Benchmarks: added `benchmarks/fetch/headers-methods.mjs` for construct / get / set / append / has / delete / iterate. Existing iterator benches now clear `sortedMap` instead of the old private symbol.

mcollina:
Can you fix the metadata of the commits? Cursor cannot signoff commits.

anonrig:
Rewrote the commits so author, committer, and Signed-off-by are Yagiz Nizipli <yagiz@nizipli.com>.

mcollina (approved):
lgtm

mcollina:
@tsctx @KhafraDev ptal

tsctx (changes requested):
One of the main advantages of the original implementation is that it closely follows the specification, with each step represented explicitly in the code. This keeps the maintenance cost low, makes it straightforward to adapt the implementation to future specification changes, and makes it easier to trace errors back to the relevant step.

The proposed changes also introduce a separate implementation of logic that is currently shared. This duplicates the same behavior and error messages across multiple places, increasing the maintenance burden and creating a risk of inconsistencies over time.

Maintaining a close correspondence with the specification is particularly important because the specification is a _living standard_. Although `Headers` is relatively stable compared with other areas, it may still evolve over time.

While performance improvements are certainly appealing, I believe the overall ROI of these changes is low. For `Headers`, iteration performance appears to be more important in typical usage. Operations such as `get` and `append` are generally performed only a limited number of times per instance, so optimizing them does not appear to provide much practical benefit, especially when weighed against the additional complexity and maintenance cost.

There are also several behavioral regressions that could result in unexpected behavior. Some of the proposed changes are valuable. However, given these trade-offs, I do not think the changes as a whole are worthwhile.

PR #5701, opened 2026-08-18
perf(fetch): skip empty RequestInit work

anonrig:
## This relates to...

- Follow-up review on this PR: idle-socket work moved to a separate PR
- RequestInit `priority` default was in the WebIDL converter, so every omitted init counted as non-empty

## Rationale

The Request constructor treated every `RequestInit` as non-empty because undici's `priority` member defaulted to `"auto"` in the WebIDL converter. Fetch's "If init is not empty" steps then always cloned/cleared/re-appended headers — including on the common `fetch(url)` / `new Request(url)` path.

## Changes

- Default `priority` to `"auto"` on the **inner request** in `makeRequest`, not in the RequestInit converter, so an omitted init stays empty after conversion and `Object.keys(init)` is correct.

The idle-socket `setImmediate` change now lives in a follow-up PR (see review thread).

### Features

N/A

### Bug Fixes

- `new Request(existingRequest)` with omitted init no longer resets `referrerPolicy` / related request state just because `priority` defaulted to `"auto"` in the converter.

N/A. Copying a Request with an empty init now matches the spec more closely (does not run the "init is not empty" resets).

## Benchmarks

Re-run 2026-08-21 on Node v22.14.0 (linux x64, Intel Xeon, 4 cores). Medians of 5 runs vs `main` @ `181c293`.

| | main | this PR | delta |
|---|---|---|---|
| `new Request(url)` | 1.28M ops/s | **1.63M ops/s** | **+27%** |
| `new Request(url, { method })` | 0.94M ops/s | 1.08M ops/s | +15% |

The `{ method }` path still counts as non-empty, so the header clone/clear work still runs. It is a bit faster because the converter no longer injects a default `priority`.

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready


mertcanaltin (approved):

tsctx:
> Separately, the Request constructor treated every `RequestInit` as non-empty because undici's `priority` member defaults to `"auto"`. Fetch's "If init is not empty" steps then always cloned/cleared/re-appended headers — including on the common `fetch(url)` / `new Request(url)` path.

The correct fix here is to set the default value when defining the internal request, rather than in the converter. The default value is currently being set in the wrong place. The following diff should fix the issue. I do not think this test is necessary.

```diff
diff --git a/lib/web/fetch/request.js b/lib/web/fetch/request.js
index dbe809c2..56945ec0 100644
--- a/lib/web/fetch/request.js
+++ b/lib/web/fetch/request.js
@@ -923,7 +923,7 @@ function makeRequest (init) {
     serviceWorkers: init.serviceWorkers ?? 'all',
     initiator: init.initiator ?? '',
     destination: init.destination ?? '',
-    priority: init.priority ?? null,
+    priority: init.priority ?? 'auto',
     origin: init.origin ?? 'client',
     policyContainer: init.policyContainer ?? 'client',
     referrer: init.referrer ?? 'client',
@@ -1129,8 +1129,7 @@ webidl.converters.RequestInit = webidl.dictionaryConverter([
   {
     key: 'priority',
     converter: webidl.converters.DOMString,
-    allowedValues: ['high', 'low', 'auto'],
-    defaultValue: () => 'auto'
+    allowedValues: ['high', 'low', 'auto']
   }
 ])
 
```

Separately, I am strongly opposed to replacing `Object.keys` with explicit property checks as an optimization.

This would make the code less maintainable and more error-prone. Whenever a new property is added in a future specification update, the property checks would also need to be updated manually. If that update were overlooked, the new property would not be handled correctly, whereas the current implementation requires no such changes.

Moreover, this optimization does not appear to provide any significant performance improvement.

anonrig:
Addressed in af14557.

Moved the `priority` default from the RequestInit converter to `makeRequest`, and restored the `Object.keys` empty-init check.

Kept a pair of Request-copy tests as a regression for the converter default making omitted init non-empty. Happy to drop them if you'd still rather not have them.

anonrig:
Dropped the extra Request-copy tests — thanks again.

lib/dispatcher/client-h1.js
```diff
@@ -1061,15 +1061,23 @@ function clearIdleSocketValidation (socket) {
...
+  socket[kIdleSocketValidationTimeout] = setImmediate(() => {
     socket[kIdleSocketValidationTimeout] = null
     socket[kIdleSocketValidation] = 2
 
     if (client[kSocket] === socket && !socket.destroyed) {
       client[kResume]()
     }
-  }, 0)
-  socket[kIdleSocketValidationTimeout].unref?.()
+  })
```
mcollina: please move these to an independent pr

mcollina (changes requested):

lib/dispatcher/client-h1.js
anonrig: Moved the idle-socket `setImmediate` change to #5707.

anonrig:
@mcollina can you review

tsctx (approved):

gurgunday (approved):
Lgtm

mcollina (approved):
lgtm

PR #5702, opened 2026-08-18
perf(fetch): skip empty ResponseInit work and send buffered bodies directly

anonrig:
## This relates to...

- Follow-up to #5701 (same class of “defaults / scaffolding do extra work”)
- GHSA-35p6-xmwp-9g52 (idle-socket validation must still yield to poll)

## Rationale

`#5701` skipped Request constructor work when `RequestInit` was empty (dictionary defaults were making every `fetch(url)` look non-empty) and dropped the idle-socket `setTimeout(0)` floor.

This PR applies the same ideas to the next similar hot paths, independently of that branch:

1. **`ResponseInit` defaults** — `status: 200` and `statusText: ""` meant every `new Response()` / `new Response(body)` ran WebIDL conversion and reason-phrase validation for values `makeResponse()` already has.
2. **Buffered request bodies** — `extractBody()` already keeps string / `BufferSource` bodies on `source`, but HTTP-network-fetch still teed the companion Web Stream and incrementally read it before handing bytes to the dispatcher.
3. **Idle-socket validation start time** — validation was only scheduled when the *next* request arrived. Starting it when the socket becomes idle lets the GHSA yield overlap body consumption instead of gating the next dispatch.

## Changes

- Treat “init has members” as **user-specified** `ResponseInit` fields, not WebIDL defaults. Omitted / `{}` init skips conversion; `makeResponse()` already supplies 200 / `""`.
- If `request.body.source` is a string or `Uint8Array`, clone headers only (no stream tee) and pass `source` to `agent.dispatch`.
- Schedule H1 idle-socket validation in `onMessageComplete` when `kSocketUsed` becomes true. The existing write-time `clearIdleSocketValidation` and poisoning checks are unchanged.

### Features

N/A

### Bug Fixes

N/A

### Breaking Changes and Depreciations

N/A

## Benchmarks

Loopback, Node 22.14. Measured against `main` (`e609acb0`):

| | main | this PR |
|---|---|---|
| `new Response()` | 4.1M ops/s | **11.1M ops/s** (~2.7×) |
| `new Response(null, { status: 201 })` | 3.1M ops/s | 2.9M ops/s (unchanged path) |
| `fetch` POST string body p50 | 1.66 ms | **1.49 ms** |
| sequential `fetch().text()` p50 | 1.48 ms | 1.45 ms |

The GET keep-alive number is still dominated by `setTimeout(0)` on `main`; #5701 replaces that with `setImmediate`. Starting validation at idle is complementary once that lands.

## Status

- [x] Tested
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/dispatcher/client-h1.js
+++ b/lib/dispatcher/client-h1.js
@@ -817,13 +817,24 @@ class Parser {
       // have been queued since then.
       util.destroy(socket, new InformationalError('reset'))
       return constants.ERROR.PAUSED
-    } else if (client[kPipelining] == null || client[kPipelining] === 1) {
-      // We must wait a full event loop cycle to reuse this socket to make sure
-      // that non-spec compliant servers are not closing the connection even if they
-      // said they won't.
-      setImmediate(client[kResume])
     } else {
-      client[kResume]()
+      // Start idle-socket validation as soon as the socket becomes idle so
+      // the event-loop yield (GHSA-35p6-xmwp-9g52) runs during body
+      // consumption instead of after the next dispatch. Sequential
+      // fetch().text() / request() can then reuse the socket without
+      // paying a second timer.
+      if (socket[kSocketUsed] && socket[kIdleSocketValidation] === 0) {
+        scheduleIdleSocketValidation(client, socket)
+      }
+
+      if (client[kPipelining] == null || client[kPipelining] === 1) {
+        // We must wait a full event loop cycle to reuse this socket to make sure
+        // that non-spec compliant servers are not closing the connection even if they
+        // said they won't.
+        setImmediate(client[kResume])
+      } else {
+        client[kResume]()
+      }
     }
 
     return 0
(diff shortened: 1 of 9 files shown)


lib/dispatcher/client-h1.js
```diff
@@ -1072,6 +1086,12 @@ function scheduleIdleSocketValidation (client, socket) {
   socket[kIdleSocketValidationTimeout].unref?.()
 }
 
+function scheduleIdleSocketValidation (client, socket) {
+  // Block further writes on this socket until the poll yield completes.
+  socket[kIdleSocketValidation] = 1
+  armIdleSocketValidation(client, socket)
+}
```
mcollina, line 1093: Can you send the changes to this as a separate PR?

mcollina (changes requested):
Can you please split the changes to client-h1 from the fetch ones?

PR #5709, opened 2026-08-23
fix(socks5-proxy-agent): destroy socket when negotiation times out

abhijeet117:
## Summary

When SOCKS5 authentication or connection negotiation times out, `createSocks5Connection()` rejects its promise but never destroys the underlying socket. The socket also never reaches the per-origin pool because the connect callback is only invoked on success, so `close()` and `destroy()` cannot reach it and it stays open for the lifetime of the process. This destroys the client in both timeout branches before rejecting.

## Testing

Reproduced with a local TCP proxy that accepts a connection but never replies to the greeting: the request rejects with "SOCKS5 authentication timeout", after which `agent.close()` leaves the socket open. Added a regression test in `test/socks5-proxy-agent.js` that asserts the connector socket is destroyed after close; it fails without this change and passes with it. All socks5 test files pass (33 pass, 1 skipped) and eslint is clean.

## Checklist

- bug reproduced before fix
- root cause identified
- bug fixed
- tests passed

### DIFF
--- a/lib/dispatcher/socks5-proxy-agent.js
+++ b/lib/dispatcher/socks5-proxy-agent.js
@@ -115,6 +115,7 @@ class Socks5ProxyAgent extends DispatcherBase {
     const authenticationReady = Promise.withResolvers()
 
     const authenticationTimeout = setTimeout(() => {
+      socks5Client.destroy()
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

metcoder95 (approved):

PR #5723, opened 2026-08-25
test: restore /xhr to the WPT filter

### LINKED ISSUE #5710: test: /xhr dropped from the WPT filter in #4786, losing FormData coverage
## Summary

`/xhr` was dropped from the `test:wpt` filter in #4786, which appears to have been incidental — the PR description does not mention it. The practical effect is that undici lost WPT coverage of `FormData`, since that suite lives under `/xhr/formdata/` in WPT rather than under `/fetch/`.

Separately, `/serviceWorkers` in the same filter has never matched anything.

I have no stake in the outcome either way — filing this because the data seemed worth surfacing, and only you can say whether the removal was deliberate.

## The `/xhr` removal

#4786 (`Remove legacy handler wrappers`, merged 2026-02-07, `393094ad50ee`) changed:

```diff
-"test:wpt": "... run /fetch /mimesniff /xhr /websockets /serviceWorkers /eventsource",
+"test:wpt": "... run /fetch /mimesniff /websockets /serviceWorkers /eventsource",
```

The PR body describes raw header handling in the cache/decompress interceptors, a shared `toRawHeaders` utility, snapshot replay raw headers, and a v8 version bump. The filter change is not mentioned.

Evidence that the `xhr` data has been frozen since: comparing `expectation.json` at `main` against that commit, the `xhr` subtree is byte-identical, while `fetch`, `websockets`, `eventsource` and `mimesniff` have all changed. #5587 (`test: update WPT expectations`, 2026-07-27) states it refreshes Fetch and WebSocket expectations; `xhr` is not included.

## What was actually lost

Of the 82 `xhr` cases that were passing at the time of removal, roughly 80 are `FormData` conformance, not `XMLHttpRequest`:

- `xhr/formdata/append.any.html`, `delete`, `get`, `has`, `set`, `set-blob`, `foreach`, `iteration`, `constructor`
- the `FormData interface: operation append(USVString, Blob, optional USVString)` block in `xhr/idlharness.any.html`

These do not exercise `XMLHttpRequest` — they are pure `FormData` tests that happen to live under `/xhr/` in WPT's directory layout. undici implements `FormData`, so this coverage was meaningful and is now silently absent.

## `/serviceWorkers` is a no-op

The WPT directory is `service-workers` (hyphenated). The filter is a case-sensitive `startsWith` on the pathname (`wpt-runner.mjs:501`), so `/service-workers/...` never matches `/serviceWorkers`.

Independent confirmation: the runner writes an entry into `expectation.json` for every test it runs, and `expectation.json` has no `serviceWorkers` or `service-workers` key — its only top-level keys are `fetch`, `mimesniff`, `websockets`, `xhr`, `eventsource`. That has been true since `5ceaf613`, the commit that introduced the runner in #4486.

Worth noting that correcting the name would be the wrong fix: service worker tests are browser-scoped, and the runner's skip guard at `wpt-runner.mjs:493-497` matches `serviceworker` without a hyphen, so it would not filter them either. Removing the argument is a behavioural no-op; correcting it would pull in a large amount of browser-only churn.

## Possible directions

Whichever fits your intent:

1. **If the removal was accidental** — restore `/xhr` to the filter, or narrow it to `/xhr/formdata` (the matcher supports it: `finalPath.startsWith(filter)`).
2. **If it was deliberate** — the `xhr` section of `expectation.json` is dead data and could be dropped.

One caveat on option 1: I have not run the suite, so I cannot say whether the remaining 548 `xhr` expectations still match current WPT. If they have drifted, restoring the filter would also need an expectation refresh.

Happy to open a PR for whichever direction you prefer, or to leave it with you.

RaphaelFakhri:
Fixes #5710.

`/xhr` was dropped from the `test:wpt` filter in #4786, which is not mentioned in that PR's description and looks incidental. The practical effect is that undici lost WPT `FormData` coverage, because that suite lives under `/xhr/formdata` rather than under `/fetch`. The `xhr` subtree of `expectation.json` has been frozen since, while `fetch`, `websockets`, `eventsource` and `mimesniff` all moved on.

I ran the suite rather than assuming the stored data still held:

- `wpt-runner.mjs run /xhr` against the untouched committed expectations exits 0. 161 test files, 82 passing and 495 expected-fail subtests, matching the 583 stored `xhr` expectations. The runner exits non-zero only when a `success` value drifts, and none did, so the frozen data is still accurate.
- Full `npm run test:wpt` with the amended filter exits 0: 1201 test files, 4869 passing and 4537 expected-fail subtests. Nothing outside `xhr` moved.
- Re-running `/xhr` against the finished branch leaves the working tree byte identical, so the refresh is stable rather than churning on every run.

The expectations refresh is 83 lines, all inside the `xhr` subtree: four WPT test files added since the removal, recorded with their observed expected-fail cases, plus two deterministic case-order moves. I diffed the parsed JSON to confirm that `xhr` is the only top-level subtree that changed.

This restores 82 passing subtests, roughly 60 of them `FormData` conformance: the `formdata/` suites plus the `FormData` block in `xhr/idlharness`.

`/serviceWorkers` is dropped from the same list. The WPT directory is `service-workers` and the filter is a case-sensitive `startsWith` on the pathname, so the argument has never matched anything. I confirmed it empirically: `wpt-runner.mjs run /serviceWorkers` reports 0 test files, and `expectation.json` has never had a key for it. Correcting the spelling instead would be the wrong move, since those tests are browser scoped and the runner's skip guard matches `serviceworker` without the hyphen, so removing the dead argument is a no-op.

One tradeoff to be aware of: the `xhr` suite adds roughly two minutes to a `test:wpt` invocation. Until #5588 lands, that cost falls on every local `npm test`, which is presumably why the removal went unnoticed. If you would rather have the coverage without the whole suite, the runner supports the narrower `/xhr/formdata`; I restored the full `/xhr` since it is the exact revert of what was dropped and its expectations verifiably still hold.

### DIFF
--- a/package.json
+++ b/package.json
@@ -97,7 +97,7 @@
     "test:websocket:autobahn": "node test/autobahn/client.js",
     "test:websocket:autobahn:report": "node test/autobahn/report.js",
     "test:wpt:setup": "node test/web-platform-tests/wpt-runner.mjs setup",
-    "test:wpt": "npm run test:wpt:setup && node test/web-platform-tests/wpt-runner.mjs run /fetch /mimesniff /websockets /serviceWorkers /eventsource",
+    "test:wpt": "npm run test:wpt:setup && node test/web-platform-tests/wpt-runner.mjs run /fetch /mimesniff /xhr /websockets /eventsource",
     "test:cache-tests": "node test/cache-interceptor/cache-tests.mjs --ci",
     "coverage": "npm run coverage:clean && cross-env NODE_V8_COVERAGE=./coverage/tmp npm run test:javascript && npm run coverage:report",
     "coverage:ci": "npm run coverage:clean && cross-env NODE_V8_COVERAGE=./coverage/tmp npm run test:javascript && npm run coverage:report:ci",
--- a/test/web-platform-tests/expectation.json
+++ b/test/web-platform-tests/expectation.json
@@ -39934,11 +39934,6 @@
       "success": true,
       "cases": [
         {
-          "name": "idl_test setup",
-          "success": false,
-          "message": "promise_test: Unhandled rejection with value: object \"ReferenceError: document is not defined\""
-        },
-        {
           "name": "idl_test validation",
           "success": true
         },
@@ -40691,6 +40686,11 @@
           "name": "ProgressEvent interface: new ProgressEvent(\"type\") must inherit property \"total\" with the proper type",
           "success": false,
           "message": "assert_equals: Unexpected exception when evaluating object expected null but got object \"ReferenceError: ProgressEvent is not defined\""
+        },
+        {
+          "name": "idl_test setup",
+          "success": false,
+          "message": "promise_test: Unhandled rejection with value: object \"ReferenceError: document is not defined\""
         }
       ]
     },
@@ -40750,10 +40750,6 @@
           "name": "Use text/xml as fallback MIME type, 2",
           "success": false,
           "message": "XMLHttpRequest is not defined"
-        },
-        {
-          "name": "Loading data…",
-          "success": true
         },
         {
           "name": "1) MIME types need to be parsed and serialized: text/html;charset=gbk",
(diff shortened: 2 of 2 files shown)


mcollina (approved):
lgtm

PR #5727, opened 2026-08-26
fix(retry): forward rawHeaders writes through RetryController

### LINKED ISSUE #5726: retry + decompress: RetryController exposes read-only rawHeaders, causing TypeError
## Bug Description

Composing the built-in `retry` interceptor before the built-in `decompress`
interceptor causes compressed responses to fail:

```js
new Agent().compose([
  interceptors.retry(),
  interceptors.decompress()
])
```

`RetryHandler` passes a `RetryController` proxy to downstream handlers. That
proxy exposes `rawHeaders` through a getter, but it has no corresponding
setter. When `DecompressHandler` receives a compressed response, it removes
`content-encoding` and `content-length` from the raw header list by assigning a
new value to `controller.rawHeaders`. Because the proxy property is
getter-only, this assignment throws in strict mode:

```text
TypeError: Cannot set property rawHeaders of #<RetryController> which has only a getter
```

The failure occurs even with `maxRetries: 0`, so no retry attempt is required
to trigger it.

## Reproduction

```javascript
'use strict'

const assert = require('node:assert/strict')
const { gzipSync } = require('node:zlib')
const { Agent, interceptors, request } = require('undici')

test('retry can be composed before decompress', { timeout: 5000 }, async (t) => {
  const compressedBody = gzipSync('ok')
  const server = createServer((_req, res) => {
    res.writeHead(200, {
      'content-encoding': 'gzip',
      'content-length': String(compressedBody.length)
    })
    res.end(compressedBody)
  })

  t.after(() => {
    server.closeAllConnections?.()
    server.close()
  })

  server.listen(0, '127.0.0.1')

  const dispatcher = new Agent().compose([
    interceptors.retry({ maxRetries: 0 }),
    interceptors.decompress()
  ])
  t.after(() => dispatcher.close())

  const { statusCode, body } = await request(
    `http://127.0.0.1:${server.address().port}`,
    {
      dispatcher,
      headers: { 'accept-encoding': 'gzip' }
    }
  )

  assert.equal(statusCode, 200)
  assert.equal(await body.text(), 'ok')
})
```

1. Save the script as `test/repro-retry-decompress-rawheaders.js`.
2. Install `undici@8.10.0`.
3. Run:

```bash
   node --test test/repro-retry-decompress-rawheaders.js
```

## Expected Behavior

The request should resolve with status code `200`, and the decompressed response
body should equal `"ok"`. The documented built-in interceptors should compose
without throwing.

## Actual Behavior

The request fails during response header processing, before the response is
delivered to the caller:

```text
TypeError: Cannot set property rawHeaders of #<RetryController> which has only a getter
```

In a non-test process, cleanup can also emit the same error from the already
created Gunzip stream as an unhandled stream `error`.

## Logs & Screenshots

```text
✖ retry can be composed before decompress

TypeError: Cannot set property rawHeaders of #<RetryController> which has only a getter
    at DecompressHandler.onResponseStart (undici/lib/interceptor/decompress.js:201:31)
    at RetryHandler.onResponseStart (undici/lib/handler/retry-handler.js:364:37)
    at Request.onResponseStart (undici/lib/core/request.js:345:39)

tests 1
pass 0
fail 1
```

## Environment

- OS: macOS 26.2 (arm64)
- Node.js version: v24.11.1
- undici version: 8.10.0

### Additional context

- The official interceptor documentation shows `retry` before `decompress` in
  its composition examples:
  https://github.com/nodejs/undici/blob/main/docs/docs/api/Interceptors.md#composing-multiple-interceptors
- Reversing the order to `decompress` before `retry` avoids the exception,
  because `DecompressHandler` then receives the native `RequestController`,
  whose `rawHeaders` field is writable.
- `RetryController` was introduced in
  https://github.com/nodejs/undici/pull/5405. It forwards reads of
  `rawHeaders` and `rawTrailers`, but not writes.
- The same getter-only proxy and `controller.rawHeaders = filteredHeaders`
  assignment are still present on the current `main` branch.
- Forwarding `rawHeaders` and `rawTrailers` setters to the active target appears
  to be one possible fix, together with a regression test for this documented
  interceptor order.

official-burak:
Fixes #5726

`RetryController` forwarded `rawHeaders` / `rawTrailers` as getters only. `DecompressHandler` assigns a filtered list onto `controller.rawHeaders` when it strips `content-encoding`, so the documented order `compose([retry(), decompress()])` threw:

```
TypeError: Cannot set property rawHeaders of #<RetryController> which has only a getter
```

This adds setters that write through to the active connection controller (no-ops before a connection is attached, same as `pause`/`resume`).

**Tests**
- Before: `test/retry-handler-controller-proxy.js` failed with the getter-only TypeError; `retry composed before decompress can rewrite rawHeaders` failed.
- After: those tests pass, plus the full decompress interceptor file (28), retry-handler (47), and interceptors/retry (16).

### DIFF
--- a/lib/handler/retry-handler.js
+++ b/lib/handler/retry-handler.js
@@ -65,7 +65,18 @@ class RetryController {
   get aborted () { return this.target?.aborted ?? false }
   get reason () { return this.target?.reason ?? null }
   get rawHeaders () { return this.target?.rawHeaders ?? null }
+  set rawHeaders (value) {
+    if (this.target) {
+      this.target.rawHeaders = value
+    }
+  }
+
   get rawTrailers () { return this.target?.rawTrailers ?? null }
+  set rawTrailers (value) {
+    if (this.target) {
+      this.target.rawTrailers = value
+    }
+  }
 }
 
 class RetryHandler {
--- a/test/interceptors/decompress.js
+++ b/test/interceptors/decompress.js
@@ -51,6 +51,47 @@ test('should decompress gzip response', async t => {
   await t.completed
 })
 
+test('retry composed before decompress can rewrite rawHeaders', async t => {
+  t = tspl(t, { plan: 2 })
+
+  const data = 'ok'
+  const compressed = gzipSync(data)
+  const server = createServer({ joinDuplicateHeaders: true }, (_req, res) => {
+    res.writeHead(200, {
+      'content-encoding': 'gzip',
+      'content-length': String(compressed.length)
+    })
+    res.end(compressed)
+  })
+
+  server.listen(0)
+  await once(server, 'listening')
+
+  const client = new Client(
+    `http://localhost:${server.address().port}`
+  ).compose([
+    interceptors.retry({ maxRetries: 0 }),
+    interceptors.decompress()
+  ])
+
+  after(async () => {
+    await client.close()
+    server.close()
+    await once(server, 'close')
+  })
+
+  const response = await client.request({
+    method: 'GET',
+    path: '/',
+    headers: { 'accept-encoding': 'gzip' }
+  })
+
+  t.equal(response.statusCode, 200)
+  t.equal(await response.body.text(), data)
+
+  await t.completed
+})
+
 test('should preserve trailers when decompressing response', async t => {
   const data = 'Response with trailers'
   const compressed = gzipSync(data)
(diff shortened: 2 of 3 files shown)


metcoder95 (approved):

mcollina (approved):
lgtm

PR #5735, opened 2026-08-29
docs: name the parameters these two blocks describe

darkdi:
## This relates to...

Nothing filed, I ran into it while reading the cache store.

## Rationale

`MemoryCacheStore.get` documents its argument as `req`, while the method takes `key` and the two neighbours in the same class, `createWriteStream` and `delete`, both document `key`. The webidl one is the more confusing of the two: `webidl.errors.exception` documents `opts`, `opts.header` and `opts.message`, but the parameter is called `message` and the body reads `message.header` and `message.message`, so the documented name collides with a field name. Its two siblings, `conversionFailed` and `invalidArgument`, name the bag `opts` and `context` and document exactly that.

## Changes

Comments only, no runtime change.

### Features

N/A

### Bug Fixes

N/A

N/A

## Status

- [ ] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/cache/memory-cache-store.js
+++ b/lib/cache/memory-cache-store.js
@@ -87,7 +87,7 @@ class MemoryCacheStore extends EventEmitter {
   }
 
   /**
-   * @param {import('../../types/cache-interceptor.d.ts').default.CacheKey} req
+   * @param {import('../../types/cache-interceptor.d.ts').default.CacheKey} key
    * @returns {import('../../types/cache-interceptor.d.ts').default.GetResult | undefined}
    */
   get (key) {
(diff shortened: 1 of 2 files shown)


metcoder95 (approved):

darkdi:
The Node 26 jobs are red on one test out of 1519, test/http2-request-never-settles.js, which timed out at 3002ms on every platform in that run. This PR only renames two JSDoc @param names, so nothing here touches http2 or timing, and the same suite is green on main.

mcollina (approved):
lgtm

PR #5736, opened 2026-08-29
fix: normalize object prototype in proxy-agent test

### LINKED ISSUE #5671: Nightly tests are failing
Tests against nightly failed, see: https://github.com/nodejs/undici/actions/runs/31252119887

spjoshis:
## Summary
Fixes nightly test failure in proxy-agent.js:1015 caused by prototype mismatch in deepStrictEqual comparison.

## What Problem This Solves
The test was failing because expectedHeaders was created with Object.create(null) (resulting in an object with null prototype) while req.headers (from Node.js HTTP handler) is a plain object. deepStrictEqual fails when comparing objects with different prototypes, even if their contents are identical.

## Why This Change Was Made
Object.create(null) creates an object with null prototype, while plain object literals have Object.prototype. This difference causes strict equality checks to fail even when the actual property values are the same.

## User Impact
Fixes the failing nightly test suite in test/proxy-agent.js.

## Evidence
test/proxy-agent.js now passes all 38 tests.

Fixes nodejs/undici#5671

### DIFF
--- a/test/proxy-agent.js
+++ b/test/proxy-agent.js
@@ -1041,7 +1041,7 @@ test('ProxyAgent correctly sends headers when using fetch - #1355, #1623 (with p
 
   after(() => setGlobalDispatcher(defaultDispatcher))
 
-  const expectedHeaders = Object.assign(Object.create(null), {
+  const expectedHeaders = {
     host: `localhost:${server.address().port}`,
     connection: 'keep-alive',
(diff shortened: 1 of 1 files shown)


MikeMcC399:
This PR also changes 0 files and appears to be a duplicate of https://github.com/nodejs/undici/pull/5734

PR #5738, opened 2026-08-30
fix(fetch): only send Sec-Fetch-Mode to potentially trustworthy URLs

kjsik11:
## This relates to...

Refs #1305, in particular https://github.com/nodejs/undici/issues/1305#issuecomment-2515230458.

Spec: https://w3c.github.io/webappsec-fetch-metadata/#abstract-opdef-append-the-fetch-metadata-headers-for-a-request

## Rationale

`appendFetchMetadata` implements the Sec-Fetch-Mode sub-algorithm but skips step 1 of the outer algorithm, which the code marks as `TODO`:

> To append the Fetch metadata headers for a request, given request r:
> 1. If r's url is not an potentially trustworthy URL, return.

So `fetch('http://192.168.0.12/')` sends `sec-fetch-mode: cors`, while browsers send no `Sec-Fetch-*` header to plain-http, non-loopback destinations. Reproduces on `main` and on the undici bundled with current Node releases.

<details>
<summary>Repro (Node built-ins only)</summary>

```js
import http from 'node:http'
import os from 'node:os'
import { once } from 'node:events'

const lanIp = Object.values(os.networkInterfaces()).flat()
  .find((i) => i.family === 'IPv4' && !i.internal).address
  res.end(JSON.stringify({ 'sec-fetch-mode': req.headers['sec-fetch-mode'] ?? null }))
})
server.listen(0, lanIp)
console.log(await (await fetch(`http://${lanIp}:${server.address().port}/`)).json())
server.close()
```

Prints `{ 'sec-fetch-mode': 'cors' }`; with this PR the header is absent. `https://`, `http://localhost` and `http://127.0.0.1` keep sending it.

</details>

## Changes

- `lib/web/fetch/util.js`: early return using the existing `isURLPotentiallyTrustworthy()` helper on the request's current URL, so the gate applies per redirect hop (matches browsers and WPT `fetch/metadata/redirect/*`). The `Sec-Fetch-Dest`/`Site`/`User` TODOs are untouched and will inherit the gate.
- `test/fetch/fetch-metadata.js` (new): header absent for plain-http non-loopback origins, present for `https://` and loopback, dropped/added across `https -> http` / `http -> https` redirects. 4 of the 7 cases fail on `main`.

### Features

N/A

### Bug Fixes

- `Sec-Fetch-Mode` is no longer sent to URLs that are not potentially trustworthy (plain `http://`/`ws://` to non-loopback hosts), per Fetch Metadata step 1. 

N/A

## Test results (Node 24.20.0)

- `npm run lint` clean; `test/fetch/fetch-metadata.js` 7/7 (4/7 fail on `main`)
- `test/fetch/*.js` 483/483
- `test/mock-agent.js` 111/111, `test/proxy-agent.js` 38/38, `test/eventsource/eventsource-connect.js` 11/11 — every existing assertion on this header targets `localhost`

This fix was prepared with AI assistance; I reproduced the bug locally and reviewed every line.

## Status

- [x] Tested
- [s] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready


test/fetch/fetch-metadata.js
tsctx, line 1: I believe this is already covered by WPT. Could you remove this test?

kjsik11: Agree, removed this file

tsctx (approved):

PR #4835, opened 2026-02-23
fix: queue WebSocket error/close events when closing CONNECTING socket

veeceey:
When `close()` is called on a WebSocket that's still in the CONNECTING state, `failWebsocketConnection` calls `handler.onSocketClose()` synchronously, which fires the error and close events during the `close()` call itself. The spec says these should be queued as tasks and fire asynchronously after `close()` returns.

This wraps the `onSocketClose()` call in `queueMicrotask()` for the CONNECTING state path so events fire after close() returns, matching browser behavior.

**Before:**
```
Calling close()...
error event fired, closeReturned = false    <-- wrong
close event fired, closeReturned = false    <-- wrong
```

**After:**
```
Calling close()...
close() returned, closeReturned = true
error event fired, closeReturned = true     <-- correct
close event fired, closeReturned = true     <-- correct
```

Relevant WPTs:
- `websockets/interfaces/WebSocket/close/close-connecting.html`
- `websockets/interfaces/WebSocket/close/close-connecting-async.any.js`

Fixes #4741

PR #5741, opened 2026-08-31
fix(websocket): defer close()/fail() events to a queued task when CONNECTING

### LINKED ISSUE #4741: WebSocket events fire synchronously during close() instead of asynchronously

yfwmaniish:
Fixes #4741.

## The bug

`close()` (and the internal "fail the WebSocket connection" path) calls
`handler.onSocketClose()` synchronously when the socket is still
`CONNECTING`, so `error`/`close` fire *during* `close()` instead of after it
returns. For an already-established connection this doesn't happen —
`onSocketClose` is wired to the real socket's `'close'` event, which is
naturally async. `CONNECTING` is the one path with no real socket yet, so it
took a synchronous shortcut instead.

## The fix

`lib/web/websocket/connection.js`, `failWebsocketConnection`: when
`isConnecting(handler.readyState)`, queue the `onSocketClose()` call with
`setImmediate` instead of calling it inline. This is the single call site
that both `WebSocket` and `WebSocketStream` share for this path, so one
change fixes both.

I also removed the `// TODO: process.nextTick` comment that was sitting
directly above the `close` event dispatch in `websocket.js` — with the fix
now deferring the whole `onSocketClose()` call at its one synchronous
trigger point, that per-line TODO no longer applies.

### Why `setImmediate`, not `process.nextTick`

I looked through prior attempts at this exact issue before writing this
(there have been a few: #4745, #4835, #5078, #5088, #5372). On #4745,
@domenic flagged that the spec's "queue a task" is a macrotask, and a
microtask-based fix (that PR used a Promise microtask; the maintainer TODO
this repo already had suggested `process.nextTick`, which has the same
issue) doesn't really carry that semantics — `process.nextTick` drains
before the event loop's I/O/check phases, so it's still effectively "this
turn." `setImmediate` is the macrotask primitive, so it matches "queue a
task" correctly.

I deliberately *didn't* take on rewiring `establishWebSocketConnection`,
`closeWebSocketConnection`, or message-received handling the way @domenic's
comment on #4745 also suggested — those are already driven by the real
socket's async I/O callbacks, which already give correct task-boundary
semantics for free. The only place that was actually firing events
synchronously is the one `failWebsocketConnection` call site this PR
touches; broadening the change to already-correct code seemed like
unnecessary scope for this specific, reported bug.

### A related bug this surfaced

Fixing the timing exposed a latent ordering issue in `WebSocketStream`'s
abort handling (`stream/websocketstream.js`). Its abort listener set
`#handshakeAborted = true` *after* calling `failWebsocketConnection`, so
when `onSocketClose` used to run synchronously (nested inside that same
call), the flag wasn't set yet, and `onSocketClose`'s own "was never
connected" logic would settle `opened`/`closed` first — with a generic
`WebSocketError`, not the abort signal's actual reason, contradicting the
spec comment directly above it ("reject...with signal's abort reason").
Deferring `onSocketClose` flipped which side won that race, exposing it.
I moved the `#handshakeAborted = true` assignment before the
`failWebsocketConnection` call so the intended guard is honored regardless
of timing, and updated `test/websocket/stream/abort-before-open.js`
accordingly (it was asserting the old, incorrect race outcome).

## Testing

- Added a regression test in `test/websocket/close.js` that connects to a
  TCP server which accepts the connection but never responds (so the
  handshake never completes and the client stays `CONNECTING`
  deterministically), calls `close()`, and asserts the `close` event only
  fires after `close()` has returned.
- Negative control: reverted the fix locally and confirmed the new test
  fails with the exact reported symptom (event handler running inside the
  `close()` call stack), then restored the fix.
- Updated `test/websocket/issue-4628.js`, which asserted the old
  synchronous-firing behavior via `t.plan()` on a non-async test function —
  converted it to `async`/await the `close` event instead of assuming
  same-tick completion.
- `npm run test:websocket` — 143/143 passing.
- `npm run test:unit` — 1516/1520 passing, 4 skipped; the one flaky failure
  (`test/http2-dispatcher.js`, an HTTP/2 PING-frame-count timing assertion)
  is unrelated to this change, doesn't touch WebSocket code, and passes
  cleanly when run in isolation.
- `npm run lint` clean.

### DIFF
--- a/lib/web/websocket/connection.js
+++ b/lib/web/websocket/connection.js
@@ -317,10 +317,12 @@ function failWebsocketConnection (handler, code, reason, cause) {
   if (isConnecting(handler.readyState)) {
     // If the connection was not established, there is no underlying socket
     // whose asynchronous 'close' event would otherwise trigger onSocketClose
-    // later. Queue a task to invoke it here instead, so the 'error' and
-    // 'close' events fire after (not during) the caller's close()/send()/etc,
-    // per https://websockets.spec.whatwg.org/#feedback-from-the-protocol
-    process.nextTick(() => handler.onSocketClose())
+    // later. The spec's "queue a task" is a macrotask, not a microtask, so
+    // use setImmediate (not process.nextTick) to invoke it here instead —
+    // this way the 'error' and 'close' events fire after (not during) the
+    // caller's close()/send()/etc, per
+    // https://websockets.spec.whatwg.org/#feedback-from-the-protocol
+    setImmediate(() => handler.onSocketClose())
   } else if (handler.socket?.destroyed === false) {
     handler.socket.destroy()
   }
--- a/lib/web/websocket/connection.js
+++ b/lib/web/websocket/connection.js
@@ -315,8 +315,12 @@ function failWebsocketConnection (handler, code, reason, cause) {
   handler.controller.abort()
 
   if (isConnecting(handler.readyState)) {
-    // If the connection was not established, we must still emit an 'error' and 'close' events
-    handler.onSocketClose()
+    // If the connection was not established, there is no underlying socket
+    // whose asynchronous 'close' event would otherwise trigger onSocketClose
+    // later. Queue a task to invoke it here instead, so the 'error' and
+    // 'close' events fire after (not during) the caller's close()/send()/etc,
+    // per https://websockets.spec.whatwg.org/#feedback-from-the-protocol
+    process.nextTick(() => handler.onSocketClose())
   } else if (handler.socket?.destroyed === false) {
     handler.socket.destroy()
   }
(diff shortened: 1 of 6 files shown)


yfwmaniish:
The Node.js 26 job failures across all platforms are `test/http2-request-never-settles.js` wedging — unrelated to this change (HTTP/2, not WebSocket) and tracked as a known, pre-existing nightly-runner flake in #5671. Node 22/24/25 all pass.

KhafraDev:
Ask Claude about the 10 other PRs with this exact same broken fix next time.

Also, if you won't take the time to write a *description* for yourself, why would I take the time to review it?

PR #5744, opened 2026-08-31
feat(cache): compress big sqlite cache entries with zstd

mcollina:
Store SQLite cache response bodies that are larger than the new `compressThreshold` option (default 1 MiB) compressed with `zstdCompressSync`, decompressing them when read back so cached responses still return the exact original body. This lets large cached responses occupy significantly less database space.

The change:
- Adds a `compressThreshold` option to `SqliteCacheStore`.
- Adds a `compressed` column to the cache table (bumping `cacheInterceptorV3` to `cacheInterceptorV4`) so stored bodies know whether to decompress.
- Falls back to storing bodies uncompressed when zstd is unavailable (Node < 22.19), while still requiring `node:sqlite`.
- Documents the new option in the Cache Store API docs and the `SqliteCacheStoreOpts` types.
- Adds a test verifying big bodies are stored compressed and round-trip correctly, and that small bodies remain uncompressed.

### DIFF
--- a/lib/cache/sqlite-cache-store.js
+++ b/lib/cache/sqlite-cache-store.js
@@ -2,13 +2,21 @@
 
 const { Writable } = require('node:stream')
 const { assertCacheKey, assertCacheValue } = require('../util/cache.js')
+const { zstdCompressSync, zstdDecompressSync } = require('node:zlib')
 
 let DatabaseSync
 
-const VERSION = 3
+const VERSION = 4
 
 // 2gb
 const MAX_ENTRY_SIZE = 2 * 1000 * 1000 * 1000
+
+// 1MiB
+const COMPRESS_THRESHOLD = 1024 * 1024
+
+// zstd support landed in Node 22.19.0/23.4.0, so it may be missing even when
+// node:sqlite is available.
+const hasZstd = typeof zstdCompressSync === 'function' && typeof zstdDecompressSync === 'function'
 
 /**
  * @typedef {import('../../types/cache-interceptor.d.ts').default.CacheStore} CacheStore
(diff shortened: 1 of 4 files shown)


ronag (commented):
1 MB is quite big? text and json is quite compressible also for smaller  size?

lib/cache/sqlite-cache-store.js
```diff
@@ -128,6 +149,7 @@ module.exports = class SqliteCacheStore {
 
         -- Data returned to the interceptor
         body BUF NULL,
+        compressed INTEGER NOT NULL DEFAULT 0,
```
ronag: can be good to make this a string and use "zstd", would make it more forward compatible

lib/cache/sqlite-cache-store.js
```diff
@@ -268,11 +297,19 @@ module.exports = class SqliteCacheStore {
       return
     }
 
+    let compressed = 0
+    let storedBody = body
+    if (body && size > this.#compressThreshold && hasZstd) {
+      storedBody = zstdCompressSync(body)
+      compressed = 1
+    }
```
ronag: I would recommend checking the response mime type to find easily compressible responses, e.g. json, xml, plain text etc... and ignore the compressThreshold for those

lib/cache/sqlite-cache-store.js
mcollina: Done - the `compressed` column now stores the algorithm as a string (`"zstd"`) instead of an integer, and `get()` only decompresses rows whose algorithm is `zstd`, so future algorithms can be added without breaking existing data.

lib/cache/sqlite-cache-store.js
mcollina: Done - responses with an easily compressible Content-Type (text/*, application/json, application/xml, application/javascript, ...) are now compressed regardless of the compressThreshold. I also guard against pathological inflation: compression only applies when the zstd output is actually smaller than the original body.

metcoder95 (approved):

PR #5757, opened 2026-09-02
fix: avoid unbounded recursion in Set-Cookie attribute parser

mcollina:
## What changed

`parseUnparsedAttributes` in `lib/web/cookies/parse.js` recursed once per `Set-Cookie` attribute with no depth bound. A single response header of the form `a=b;;;;...` overflowed the stack and threw an uncaught `RangeError` out of `getSetCookies`.

This converts the algorithm to an iterative loop, threading `cookieAttributeList` through each iteration exactly as the recursion did. No behavior change; the parser no longer grows the stack per attribute.

## Verification

- The repro `'a=b' + ';'.repeat(5000)` now parses successfully instead of throwing `RangeError: Maximum call stack size exceeded`.
- End-to-end against a real server: 2000 and 5000 semicolons parse fine; larger payloads fail at the header-size limit, not the parser.
- Added a regression test in `test/cookie/cookies.js`.
- Full cookie test suite passes and lint is clean.

### DIFF
--- a/lib/web/cookies/parse.js
+++ b/lib/web/cookies/parse.js
@@ -92,223 +92,224 @@ function parseSetCookie (header) {
  * @param {Object.<string, unknown>} [cookieAttributeList={}]
  */
 function parseUnparsedAttributes (unparsedAttributes, cookieAttributeList = {}) {
-  // 1. If the unparsed-attributes string is empty, skip the rest of
-  //    these steps.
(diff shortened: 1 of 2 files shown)


metcoder95 (approved):

tractorcow:
Hello, could this fix perhaps be back-ported to v6 / v7 as well please?

PR #5761, opened 2026-09-03
fix: complete upgrade diagnostics lifecycle

BridgeAR:
Successful CONNECT and protocol upgrades stop before Undici marks the request complete. Diagnostics subscribers retain per-request state, and streamed request-body listeners remain attached for the lifetime of the upgraded socket.

Complete accepted upgrades through the existing response-header and trailers events. Rejected and aborted upgrades retain the existing error terminal event.

### DIFF
--- a/docs/docs/api/DiagnosticsChannel.md
+++ b/docs/docs/api/DiagnosticsChannel.md
@@ -120,7 +120,8 @@ diagnosticsChannel.channel('undici:request:bodySent').subscribe(({ request }) =>
 added: v6.3.0
 -->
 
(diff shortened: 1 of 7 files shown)


mcollina:
Do you need this backported to v7.x?

BridgeAR:
@mcollina yes, it needs to be backported to all supported release lines. It is an issue since v4.6.0

mcollina:
Can you open the backports PR then? The code would be substantially different in v7 at least.

mcollina (approved):
lgtm

PR #5762, opened 2026-09-04
fix(mock): settle the dispatch when an async reply data function rejects

pacocartones:
## What

`MockPool`/`MockAgent` `.reply(statusCode, dataFn)` supports an **async** data function — the returned promise is already awaited in `handleReply`, and it's exercised in the existing suite (`test/mock-agent.js:578`, `:3487`). But that awaited promise had **only a fulfilment handler**: if the data function rejects, no `onResponse*` callback ever fires, so the dispatched request **never settles** (hang) and the rejection escapes as an `unhandledRejection`.

## Fix

Add the rejection arm, mirroring the file's own conventions — the sibling reply-*options*-callback path already handles both fulfil and reject, and the sync path re-checks `aborted`:

```js
return body.then(
  (newData) => handleReply(mockDispatches, { ..._response, data: newData }),
  (err) => {
    if (aborted) return
    deleteMockDispatch(mockDispatches, key)
    handler.onResponseError?.(controller, err)
  }
)
```

The request now rejects with the error, exactly one terminal callback, no unhandled rejection. Sync/options paths are untouched.

## Test

`test/mock-async-reply-reject.js`: `.reply(200, async () => { throw boom })` then a request, with a timeout guard and an `unhandledRejection` listener. **RED on `main`** (unhandled rejection; request never settles). **GREEN** with the fix (rejects with `boom`, data fn called once, no unhandled rejection).

### DIFF
--- a/lib/mock/mock-utils.js
+++ b/lib/mock/mock-utils.js
@@ -508,7 +508,19 @@ function dispatchMockReply (mockDispatches, mockDispatch, key, opts, handler, re
       // synchronously throw the error, which breaks some tests.
       // Rather, we wait for the callback to resolve if it is a
       // promise, and then re-run handleReply with the new body.
-      return body.then((newData) => handleReply(mockDispatches, { ..._response, data: newData }))
+      // A rejection is surfaced as a response error so the dispatch settles
+      // instead of hanging, mirroring the reply options callback path.
+      return body.then(
+        (newData) => handleReply(mockDispatches, { ..._response, data: newData }),
+        (err) => {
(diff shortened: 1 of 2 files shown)


pacocartones:
The remaining Node.js 24 Ubuntu failure does not reproduce locally: the named  test passed 50/50 runs on this head and 50/50 on its exact base, and the new focused test also passes. I tried the Actions rerun endpoint, but contributors receive HTTP 403. Could a maintainer rerun the failed job when convenient?

PR #5763, opened 2026-09-04
test(mock): make the delayed-abort test discriminate instead of asserting true

pacocartones:
### Problem

The first test in `test/mock-delayed-abort.js` ended with:

```js
p.ok(true, 'No uncaught errors after delayed response')
```

That assertion is a tautology — it stays green no matter what the code under test does. The bare `MockAgent.request` path throws no *uncaught* error when a delayed reply is delivered after an abort, so the test passed even against a fully-broken guard. It documented an intent it never actually enforced: that a mocked reply delivered *after* the request was aborted must not still be pushed to the handler.

### What this hardens

The post-abort guards live in `lib/mock/mock-utils.js` (the timer-clear on abort plus the two `if (aborted) return` guards in `handleReply`). To prove the test now discriminates, I reverted those three guards and re-ran:

| variant | test 1 (this one) |
|---|---|
| **real code** | passes |
| **guards reverted** | **fails** |

Previously, test 1 passed in *both* variants.

### The change

The delayed-abort test now wraps the dispatched handler in a `DecoratorHandler` subclass that records any `onResponseStart` / `onResponseData` / `onResponseEnd` delivered *after* `onResponseError`, and asserts none occur:

```js
p.strictEqual(deliveredAfterError, false)
```

This covers the full response side of the contract for the bare-agent path — the sibling `DecoratorHandler` test in the same file asserts only `onResponseStart`. Test-only change; no source touched.

Verified with the repo runner: `borp -p "test/mock-*.js"` → **339 pass / 0 fail**; `npm run lint` → clean.

### DIFF
--- a/test/mock-delayed-abort.js
+++ b/test/mock-delayed-abort.js
@@ -5,7 +5,7 @@ const { MockAgent, interceptors } = require('..')
 const DecoratorHandler = require('../lib/handler/decorator-handler')
 const { tspl } = require('@matteo.collina/tspl')
 
-test('MockAgent with delayed response and AbortSignal should not cause uncaught errors', async (t) => {
+test('MockAgent with delayed response and AbortSignal must not deliver the response after abort', async (t) => {
   const p = tspl(t, { plan: 2 })
 
   const agent = new MockAgent()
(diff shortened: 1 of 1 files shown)


pacocartones:
The only non-green check is a cancelled Node.js 22 Ubuntu WASM-SIMD-disabled job. I tried the Actions rerun endpoint, but contributors receive HTTP 403. Could a maintainer rerun the cancelled job when convenient?

PR #5764, opened 2026-09-04
fix(mock): guard aborted requests in async reply callback rejection

pacocartones:
### Problem
In `lib/mock/mock-utils.js`, `sendReply()` handles an async **reply options callback** (the `.reply(() => ...)` / `response.callback` path from #5534). Its resolve arm goes through `handleReply()` (guarded by `if (aborted) return`), but the **reject arm** called `handler.onResponseError(null, err)` with no abort guard.

When a request with a body is aborted while that options-callback promise is still pending, `controller.abort()` already settles the request via `onResponseError` (first terminal). The callback promise then rejects and calls `onResponseError` **again** → a double-settle of an already-aborted request.

This is distinct from #5762: that one guards the **data/body function** promise in `handleReply()`; this guards the **reply options callback** promise in `sendReply()` — different function, different call site, different feature.

### Fix
Guard the reject arm with `if (aborted) return`, mirroring the guard `handleReply()` already applies.

### Test
`test/mock-delayed-abort.js`: a `CountingHandler` counts `onResponseError`; a POST with a body and an options callback that rejects at 50ms, aborted at 10ms, asserts `onResponseError` fires exactly once.
- **Before:** `2 !== 1` (double-settle).
- **After:** exactly once.

`node --test test/mock-delayed-abort.js` (4 pass), whole mock suite via borp (340 pass), `npm run lint` — all green.

### DIFF
--- a/lib/mock/mock-utils.js
+++ b/lib/mock/mock-utils.js
@@ -460,6 +460,14 @@ function dispatchMockReply (mockDispatches, mockDispatch, key, opts, handler, re
             handleReply(dispatches, { ...responseDefaults, ...resolvedData })
           },
           (err) => {
+            // The request may have been aborted while the reply options
+            // callback promise was still pending. controller.abort() has then
+            // already settled the request via onResponseError, so emitting a
+            // second terminal here would double-settle an aborted request,
+            // mirroring the guard handleReply() already applies.
+            if (aborted) {
+              return
+            }
             handler.onResponseError(null, err)
           }
         )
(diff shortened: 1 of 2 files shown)


pacocartones:
The only non-green checks are two cancelled CI jobs (Node.js 24 Ubuntu coverage and Node.js 22 Ubuntu). I tried the Actions rerun endpoint, but contributors receive HTTP 403. Could a maintainer rerun the cancelled jobs when convenient?

PR #5765, opened 2026-09-04
fix(deduplicate): detach the primary on abort instead of rejecting the group

### LINKED ISSUE #5711: deduplicate: aborting the primary request rejects the other callers in the group
## Summary

With `interceptors.deduplicate()`, aborting the *first* request of a coalesced group also rejects every other caller waiting on it — including callers that passed no `signal` at all.

I have a fix and tests, but it changes the interceptor's semantics rather than patching a slip, so I would rather agree the intended behaviour with you before sending a PR.

## Reproduction

Two concurrent requests to the same resource through the deduplicate interceptor. The first carries an `AbortSignal` (a timeout, a cancelled UI action); the second carries none.

```
waiter outcome -> rejected: AbortError / This operation was aborted
```

The waiter had no signal of its own.

## Where it comes from

`DeduplicationHandler` hands the *primary* caller the real dispatch controller (`lib/handler/deduplication-handler.js:147-149`). When that caller aborts, `api-request.js:92` calls `controller.abort(reason)` on the real controller, the network request is cancelled, and `onResponseError` (`:290-305`) settles every waiting handler with `AbortError`.

The existing tests cover the inverse case — aborting a *deduplicated* request, from #5673 — but not this direction.

## The question

For a request coalescer, I would expect the group to survive as long as any member still wants the response: the primary's abort should detach that caller, and the underlying request should only be cancelled once nobody is left waiting. Otherwise coalescing is not transparent, and whether your request succeeds depends on who else happened to ask first and whether *they* had a timeout.

If that is the intended semantics, I have it working locally: the primary gets a proxied controller (same shape as the existing `#createWaitingHandler`), `abort` settles only the primary, and the real dispatch is aborted only when the primary is gone and no waiting handler remains. Roughly 95 lines in the handler.

Two details that fell out of writing the tests:

- The `pendingRequests` entry was not being released when *every* member of a group aborted, so a later request could join a request with no consumers left. That needed fixing too.
- Behaviour with no waiters is unchanged: the abort propagates exactly as it does today.

Tests: the new case fails on `main` with `AbortError` where a body was expected, and passes with the change. `test/interceptors/deduplicate.js` goes to 38/38; `cache.js` and `retry.js` stay at 103/103.

Happy to open the PR, or to take a different shape if you would rather the group be torn down with its primary — in which case the `pendingRequests` leak is still worth fixing on its own.

pacocartones:
Fixes #5711.

### Problem
With `interceptors.deduplicate()`, aborting the **primary** request of a coalesced group rejected **every** waiter with `AbortError` — even waiters that passed no `signal`. The primary received the real dispatch controller, so its abort tore down the shared dispatch and rejected all members.

### Fix
The primary now receives a **proxied controller** (mirroring the existing `#createWaitingHandler`). Aborting it detaches only the primary and defers to `#maybeAbortRealDispatch`, which tears down the shared dispatch **only once no member (primary or waiter) is left waiting**, releasing the `pendingRequests` entry via `onComplete`. Pause/resume still forward to the real controller, so backpressure is unchanged. With no waiters, abort propagates exactly as before.

### Test
`test/interceptors/deduplicate.js`:
- "aborting the primary request must not reject the coalesced waiters" — **RED** on main (waiter rejects with AbortError), **GREEN** with fix (waiter gets 200, origin hit once).
- "when every member aborts, the pending entry is released" — guards the new teardown path.

deduplicate 38/38 + cache 88/88 + retry 16/16, `npm run lint` — all green.

### DIFF
--- a/test/interceptors/deduplicate.js
+++ b/test/interceptors/deduplicate.js
@@ -1444,4 +1444,132 @@ describe('Deduplicate Interceptor', () => {
     const res = await primary
     strictEqual(await res.body.text(), 'response-body')
   })
+
+  // https://github.com/nodejs/undici/issues/5711
+  test('aborting the primary request must not reject the coalesced waiters', async () => {
+    let requestsToOrigin = 0
+    let releaseResponse
+    const responseGate = new Promise((resolve) => { releaseResponse = resolve })
+    let markRequestReached
+    const requestReached = new Promise((resolve) => { markRequestReached = resolve })
+    const server = createServer(async (req, res) => {
+      requestsToOrigin++
+      markRequestReached()
+      // Hold the response open until the test releases it (no timers), so the
+      // primary can abort while the waiter is still attached.
+      await responseGate
+      res.end('response-body')
+    }).listen(0)
+
+    const client = new Client(`http://localhost:${server.address().port}`)
+      .compose(interceptors.deduplicate())
+
+    after(async () => {
+      releaseResponse()
+      server.close()
+      await client.close()
+    })
+
+    await once(server, 'listening')
+
+    const request = { origin: 'localhost', method: 'GET', path: '/' }
+
+    const abortController = new AbortController()
+
+    // The primary carries a signal; the waiter carries none.
+    const primary = client.request({ ...request, signal: abortController.signal })
+    const waiter = client.request({ ...request })
+
+    // Catch the primary's rejection eagerly so aborting it does not surface as
+    // an unhandled rejection while we await the waiter.
+    const primaryOutcome = primary.then(() => 'resolved', err => err)
+
+    // The single coalesced request has reached the origin, so the waiter has
+    // joined the group: abort the primary (deterministic, no timers).
+    await requestReached
+    abortController.abort()
+
+    const outcome = await primaryOutcome
+    strictEqual(outcome?.name, 'AbortError')
+
+    // The waiter never passed a signal: it must still receive the response.
+    releaseResponse()
+    const res = await waiter
+    strictEqual(res.statusCode, 200)
+    strictEqual(await res.body.text(), 'response-body')
+
+    // Coalescing held: only one request ever reached the origin.
+    strictEqual(requestsToOrigin, 1)
+  })
+
+  // https://github.com/nodejs/undici/issues/5711
+  test('when every member of a coalesced group aborts, the pending entry is released', async () => {
+    let requestsToOrigin = 0
+    let releaseResponse
+    const responseGate = new Promise((resolve) => { releaseResponse = resolve })
+    let markRequestReached
+    const requestReached = new Promise((resolve) => { markRequestReached = resolve })
+    const server = createServer(async (req, res) => {
+      requestsToOrigin++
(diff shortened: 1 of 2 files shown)


mcollina (approved):
lgtm

test/interceptors/deduplicate.js
```diff
@@ -1444,4 +1444,111 @@ describe('Deduplicate Interceptor', () => {
...
+    // shared dispatch must survive.
+    acPrimary.abort()
+    strictEqual((await primaryOutcome)?.name, 'AbortError')
+
+    // Now the last remaining member leaves; the shared dispatch is torn down
+    // and the pending entry must be released.
+    acWaiter.abort()
+    strictEqual((await waiterOutcome)?.name, 'AbortError')
+
+    await sleep(50)
```
mcollina: Please do not use sleep, it's flaky.

test/interceptors/deduplicate.js
```diff
@@ -1444,4 +1444,111 @@ describe('Deduplicate Interceptor', () => {
...
+    const acPrimary = new AbortController()
+    const acWaiter = new AbortController()
+
+    const primary = client.request({ ...request, signal: acPrimary.signal })
+    const waiter = client.request({ ...request, signal: acWaiter.signal })
+
+    const primaryOutcome = primary.then(() => 'resolved', err => err)
+    const waiterOutcome = waiter.then(() => 'resolved', err => err)
+
+    await sleep(50)
```
mcollina: Please do not use sleep, it's flaky.

test/interceptors/deduplicate.js
```diff
@@ -1444,4 +1444,111 @@ describe('Deduplicate Interceptor', () => {
...
+    // The primary carries a signal; the waiter carries none.
+    const primary = client.request({ ...request, signal: abortController.signal })
+    const waiter = client.request({ ...request })
+
+    // Catch the primary's rejection eagerly so aborting it does not surface as
+    // an unhandled rejection while we await the waiter.
+    const primaryOutcome = primary.then(() => 'resolved', err => err)
+
+    // Let the waiter join the coalesced group, then abort the primary.
+    await sleep(50)
```
mcollina: Please do not use sleep, it's flaky.

test/interceptors/deduplicate.js
```diff
@@ -1444,4 +1444,111 @@ describe('Deduplicate Interceptor', () => {
...
   })
+
+  // https://github.com/nodejs/undici/issues/5711
+  test('aborting the primary request must not reject the coalesced waiters', async () => {
+    let requestsToOrigin = 0
+    const server = createServer(async (req, res) => {
+      requestsToOrigin++
+      // Hold the response open long enough for the primary to abort while the
+      // waiter is still attached.
+      await sleep(300)
```
mcollina: Please do not use sleep, it's flaky.

mcollina (changes requested):

pacocartones:
The sleep-based waits from the earlier revision have been removed from the current head, and the full CI matrix is green. @mcollina, could you please take another look when convenient?

PR #5767, opened 2026-09-04
fix(fetch): dump 3xx redirect bodies without pausing so the connection is released

### LINKED ISSUE #5728: fetch follow does not dump 3xx bodies and hangs when the redirect body is large
## Bug Description
`fetch({ redirect: 'follow' })` does not dump 3xx response bodies. If the redirect body is big enough (I think it just needs to be is larger than the unread stream highWaterMark) the connection stays `running`. The follow-up request then cannot use that connection.

`request()` appears to already ignore 3xx bodies and keeps reading. I'm unsure then if this is expected behaviour or a bug but thought it would be worth reporting regardless.

Additionally with redirect follow the intermediate 3xx responses are never exposed so it's impossible to clean them up / free the connection.

This is especially painful with an `Agent` used as `dispatcher` and a `connections` limit as each redirect can pin a pool slot and exhaust the pool. 

## Reproduction
Standalone reproduction script:

```
'use strict'
const { fetch, Agent } = require('undici') // use require('..') inside this repo
test('fetch follow does not pin the keep-alive socket on a 301 with a large body', { timeout: 15_000 }, async (t) => {
  t.plan(3)
  const redirectBody = Buffer.alloc(128 * 1024, 0x78)
    if (req.url === '/redirect') {
      res.writeHead(301, { Location: '/final' })
      res.end(redirectBody)
      return
    }
    res.end('ok')
  })
  const dispatcher = new Agent({ connections: 1, keepAliveTimeout: 10_000 })
  t.after(async () => {
    await dispatcher.destroy()
    server.close()
  })
  server.listen(0)
  const url = `http://127.0.0.1:${server.address().port}/redirect`
  const first = await fetch(url, { dispatcher, redirect: 'follow' })
  t.assert.strictEqual(first.status, 200)
  t.assert.strictEqual(await first.text(), 'ok')
  t.assert.ok(first.redirected)
})
```

## Expected Behavior
fetch follows the 301, dumps the unused redirect body, returns the /final 200, and releases the connection. 

## Actual Behavior
fetch never resolves. The only pooled socket is stuck on the unread 301 body, so the follow-up GET to /final queues forever.

## Logs & Screenshots

✖ fetch follow does not pin the keep-alive socket on a 301 with a large body (15006.989667ms)
ℹ tests 1
ℹ pass 0
ℹ fail 0
ℹ cancelled 1
ℹ duration_ms 15208.42675
✖ failing tests:
test at test/repro-fetch-follow-redirect-body.js:11:1
✖ fetch follow does not pin the keep-alive socket on a 301 with a large body (15006.989667ms)
  'test timed out after 15000ms'

## Environment
OS: macOS 26.6 (Darwin 25.6.0 arm64)
Node.js version: v24.15.0
undici version: 7.25.0 (happens on main too)

### Additional Context

pacocartones:
Fixes #5728.

### Problem
`fetch({ redirect: 'follow' })` hangs when a 3xx redirect body is large, on a connection-limited `Agent`. In `httpNetworkFetch` (`lib/web/fetch/index.js`), `onResponseData` pushed each chunk into the internal `Readable` and called `controller.pause()` on backpressure. A redirect body larger than the highWaterMark (the repro uses 128 KiB) fills the buffer, pauses the socket, and — since the follow path discards the response and recurses into `mainFetch` — the body is never consumed. The connection stays `running`; with `Agent({ connections: 1 })` the follow-up can never acquire the pinned socket → hang.

The `request()` path (`lib/handler/redirect-handler.js`) already discards 3xx bodies **without pausing**, so it never pins the connection; the fetch path lacked the equivalent.

### Fix
Set `this.willFollow` in `onResponseStart`, and in `onResponseData` return early when `fetchParams.controller.dump || this.willFollow` — discarding the bytes but **not pausing**, so the socket drains, `onResponseEnd` fires, and the connection is released. Per-hop (the handler is per-dispatch), and correct because a followed 3xx body is never exposed to the caller.

### Test
`test/issue-5728.js`: **RED** timed out at 15s (hang); **GREEN** passes in ~31ms. Full redirect regression suite green (`redirect-request` 83, `redirect-stream` 23, `fetch/redirect`, `client-fetch` 33, …), `npm run lint` — all green.

### DIFF
--- a/lib/web/fetch/index.js
+++ b/lib/web/fetch/index.js
@@ -2266,6 +2266,15 @@ async function httpNetworkFetch (
             const willFollow = location && request.redirect === 'follow' &&
               redirectStatusSet.has(status)
 
+            // When this 3xx response will be followed, its body is discarded and
+            // never exposed to the caller. Drain it (ignore the bytes but keep
+            // reading, see onResponseData) instead of buffering it into this.body,
+            // otherwise a redirect body larger than the stream's highWaterMark
+            // applies backpressure that pauses and pins the connection, hanging
+            // the follow-up request on a connection-limited dispatcher.
+            // See https://github.com/nodejs/undici/issues/5728
+            this.willFollow = willFollow
+
             const decoders = []
 
             // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Encoding
@@ -2335,7 +2344,10 @@ async function httpNetworkFetch (
           },
 
           onResponseData (controller, chunk) {
-            if (fetchParams.controller.dump) {
+            // Discard the body of a 3xx response that is about to be followed,
+            // but keep reading (do not pause) so the connection drains and is
+            // released back to the pool. https://github.com/nodejs/undici/issues/5728
+            if (fetchParams.controller.dump || this.willFollow) {
               return
             }
 
--- a/test/issue-5728.js
+++ b/test/issue-5728.js
@@ -1,0 +1,41 @@
+'use strict'
+
+// Repro for https://github.com/nodejs/undici/issues/5728
+// fetch({ redirect: 'follow' }) does not drain the 3xx redirect response body.
+// When the redirect body is larger than the stream highWaterMark the socket
+// stays `running`, so with an Agent limited to a single connection the
+// follow-up request can never acquire the pinned socket and fetch hangs.
+//
+// Run: node --test test/issue-5728.js
+
+const { test } = require('node:test')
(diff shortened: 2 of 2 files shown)


pacocartones:
The only non-green check is a cancelled Node.js 22 Ubuntu WASM-SIMD-disabled job. I tried the Actions rerun endpoint, but contributors receive HTTP 403. Could a maintainer rerun the cancelled job when convenient?

PR #4530, opened 2025-09-08
Disable SIMD for PPC64 architecture, add UNDICI_NO_WASM_SIMD env to facilitate testing

mcollina:

PR #5776, opened 2026-09-04
fix: selectively re-enable SIMD for ppc64

richardlau:
## This relates to...

- https://github.com/nodejs/undici/pull/4530
- https://github.com/nodejs/node/pull/60177

## Rationale

WASM SIMD was disabled by default for all ppc64 due to a bug in V8 on Power >= 9. The fix went into Node.js 25.0.0 and 24.12.0.

## Changes

Re-enable SIMD by default for ppc64, and only disable when running on the older unpatched versions of Node.js.

### Features

N/A

### Bug Fixes

WASM SIMD is re-enabled for ppc64 on Node.js versions known to contain the V8 fix.

N/A

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/dispatcher/client-h1.js
+++ b/lib/dispatcher/client-h1.js
@@ -70,8 +70,14 @@ function lazyllhttp () {
 
   let mod
 
-  // We disable wasm SIMD on ppc64 as it seems to be broken on Power 9 architectures.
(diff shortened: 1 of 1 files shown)


mcollina (approved):
wow, nice!

PR #5777, opened 2026-09-05
fix(env-http-proxy-agent): NO_PROXY wildcard semantics

mcollina:
Fix NO_PROXY wildcard matching in `EnvHttpProxyAgent` so `fetch()` is internally consistent with `node:http` (see `nodejs/node#57872`, the `NO_PROXY` differences table).

Three non-sensible behaviors are fixed:

- `*.example.com` previously bypassed the proxy for the apex `example.com` itself. A `*.domain` entry is a subdomain wildcard and must only match `sub.example.com` / `a.b.example.com`, not the apex. `node:http` already does this.
- ` * ` (a wildcard with surrounding whitespace) was not recognized as a global wildcard because the check compared the whole `NO_PROXY` string against `*`.
- `none.invalid,*` (a `*` among other entries) was likewise not recognized as a global wildcard.

A bare `*` entry now matches all hosts regardless of position or whitespace, and is port-aware (`*:80` only bypasses that port).

Tests updated and added; all `env-http-proxy-agent` tests pass and lint is clean.

Refs: https://github.com/nodejs/node/issues/57872

### DIFF
--- a/lib/dispatcher/env-http-proxy-agent.js
+++ b/lib/dispatcher/env-http-proxy-agent.js
@@ -94,17 +94,25 @@ class EnvHttpProxyAgent extends DispatcherBase {
     if (this.#noProxyEntries.length === 0) {
       return true // Always proxy if NO_PROXY is not set or empty.
     }
-    if (this.#noProxyValue === '*') {
-      return false // Never proxy if wildcard is set.
-    }
 
     for (let i = 0; i < this.#noProxyEntries.length; i++) {
       const entry = this.#noProxyEntries[i]
+      // A bare `*` entry matches all hosts regardless of its position or the
+      // surrounding whitespace (e.g. ` * ` or `none.invalid,*`). If a port is
+      // attached (`*:80`) it only bypasses that port.
+      if (entry.hostname === '*') {
+        if (entry.port && entry.port !== port) {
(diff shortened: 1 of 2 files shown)


lib/dispatcher/env-http-proxy-agent.js
```diff
@@ -94,17 +94,25 @@ class EnvHttpProxyAgent extends DispatcherBase {
...
       const entry = this.#noProxyEntries[i]
+      // A bare `*` entry matches all hosts regardless of its position or the
+      // surrounding whitespace (e.g. ` * ` or `none.invalid,*`). If a port is
+      // attached (`*:80`) it only bypasses that port.
+      if (entry.hostname === '*') {
+        if (entry.port && entry.port !== port) {
+          continue
+        }
+        return false // Never proxy if a wildcard entry is present.
+      }
```
inoway46, line 108: I noticed this also introduces `*:PORT` behavior, while `node:http` currently only treats an exact `*` entry as a global wildcard.

Is that difference intentional given the goal of keeping the two implementations consistent?

If `*:PORT` is intended here, it might also be worth adding coverage for the non-matching-port case.

metcoder95 (approved):

PR #5779, opened 2026-09-06
fix(proxy-agent): preserve repeated iterable headers

GiHoon1123:
Follow-up to #5758.

ProxyAgent converts iterable request headers to an object before checking for Proxy-Authorization. If the iterable contains the same header more than once, the conversion keeps only the last value.

Keep repeated values when building the headers used by the proxy request, and add a regression test for the case.

Tests:
- `npx borp --timeout 180000 test/proxy-agent.js`
- `npx eslint lib/dispatcher/proxy-agent.js test/proxy-agent.js`

### DIFF
--- a/lib/dispatcher/proxy-agent.js
+++ b/lib/dispatcher/proxy-agent.js
@@ -325,8 +325,8 @@ class ProxyAgent extends DispatcherBase {
 }
 
 /**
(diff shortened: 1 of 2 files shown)


metcoder95 (approved):

mcollina (approved):
lgtm

Issue #4336, opened 2025-07-16
Normalize request headers and query data before invoking composed interceptors


PandaWorker:
# Description

When using Agent().compose(...) to add custom interceptors, the request’s headers and query data arrive in wildly different shapes depending on how you called request() or fetch(). Users can pass:

* Nothing (opts.headers is undefined)
* Array of strings (['k','v','k2','v2'])
* Plain object ({ k: 'v', k2: 'v2' })
* Headers instance
* Custom iterable (e.g. a user‑defined class with [Symbol.iterator] yielding [name, value] pairs)

Because these values aren’t normalized before your composed interceptors run, every interceptor ends up having to defensively check for multiple shapes:

```ts
if (!opts.headers) { … }
else if (Array.isArray(opts.headers)) { … }
else if (opts.headers instanceof Headers) { … }
else if (typeof opts.headers[Symbol.iterator] === 'function') { … }
else if (typeof opts.headers === 'object') { … }
```

This bloats interceptor code and forces duplication of header‑parsing logic in every piece of middleware (auth, compression, retries, logging, etc.).

# Reproduction

```ts
import { request, fetch, Agent, Headers, Dispatcher, DecoratorHandler } from 'undici';

function createCustomInterceptor(): Dispatcher.DispatcherComposeInterceptor {
  return (dispatch) => (opts, handler) => {
    console.log(`-> ${opts.method} ${opts.path}, headers`, opts.headers);

    class LoggerHandler extends DecoratorHandler {
      onResponseStart(controller, status, headers, statusText) {
        console.log(`<- ${status} ${statusText}, headers:`, headers);
        super.onResponseStart(controller, status, headers, statusText);
      }
    }

    return dispatch(opts, new LoggerHandler(handler));
  };
}

const dispatcher = new Agent().compose(createCustomInterceptor());

class CustomHeaders {
  raw = [['k','v'], ['k2','v2']];
  *entries() { for (const pair of this.raw) yield pair; }
  [Symbol.iterator]() { return this.entries(); }
}

// Try all the different ways to pass headers:
await request('https://api.ipify.org/', { dispatcher });
await request('https://api.ipify.org/', { headers: ['k','v','k2','v2'], dispatcher, responseHeaders: 'raw' });
await request('https://api.ipify.org/', { headers: { k: 'v', k2: 'v2' }, dispatcher });
await request('https://api.ipify.org/', { headers: new Headers([['k','v'],['k2','v2']]), dispatcher });
[...]

PR #5780, opened 2026-09-06
feat(util): expose request header normalization

GiHoon1123:
Refs #4336.

Request headers can reach composed interceptors as an object, flat array, `Headers` instance, or another iterable. This adds `util.normalizeHeaders()` so interceptors that inspect headers can convert those inputs to a lower-case object while preserving repeated values.

The cache and deduplicate interceptors now use the same helper. `compose()` no longer normalizes every request, avoiding extra header work for interceptors that do not inspect them.

Tests:

- `npm run lint`
- `npm run test:typescript`
- `borp --timeout 180000 test/interceptors/interceptors-on-client.js test/interceptors/cache-query-params.js test/interceptors/redirect.js test/cache-interceptor/cache-utils.js test/node-test/util.js`
- `node --test --test-name-pattern '#5522' test/interceptors/dns.js`

### DIFF
--- a/lib/dispatcher/dispatcher.js
+++ b/lib/dispatcher/dispatcher.js
@@ -1,77 +1,6 @@
 'use strict'
 const EventEmitter = require('node:events')
 const { kOriginless, kUrl } = require('../core/symbols')
-const { hasSafeIterator } = require('../core/util')
-
-function appendHeader (headers, key, value) {
-  if (value === undefined) {
-    return
-  }
-
-  key = key.toLowerCase()
-  const current = headers[key]
-  const values = Array.isArray(value) ? value : [value]
-
-  if (current === undefined) {
-    headers[key] = Array.isArray(value) ? value.slice() : value
-  } else if (Array.isArray(current)) {
-    current.push(...values)
-  } else {
-    headers[key] = [current, ...values]
-  }
-}
-
-function normalizeHeaders (headers) {
-  if (headers == null || typeof headers !== 'object') {
-    return headers
-  }
-
-  const prototype = Object.getPrototypeOf(headers)
-  if ((prototype === Object.prototype || prototype === null) && !Object.hasOwn(headers, Symbol.iterator)) {
-    let normalized = true
-    for (const key of Object.keys(headers)) {
-      if (key !== key.toLowerCase()) {
-        normalized = false
-        break
-      }
-    }
-    if (normalized) {
-      return headers
-    }
-  }
-
-  const normalized = {}
-
-  if (Array.isArray(headers)) {
-    if (headers.length > 0 && Array.isArray(headers[0])) {
-      for (let i = 0; i < headers.length; i++) {
-        if (!Array.isArray(headers[i]) || headers[i].length !== 2) {
-          return headers
-        }
-      }
-      for (const [key, value] of headers) {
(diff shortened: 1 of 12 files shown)


mcollina:
I'm not convinced by this change, as it's significantly perf cost.

Wdyt @ronag?

ronag:
Yea. This is probably not a good idea...

GiHoon1123:
That makes sense. The current implementation wraps every interceptor boundary, so it does more work than necessary for each request.

I’m thinking of changing it to normalize once at the composed dispatcher entry, skip plain objects that are already normalized, and keep the resulting object through interceptor redispatches. Would that address the performance concern, or would you prefer a different approach?

GiHoon1123:
I reworked this to avoid wrapping each interceptor boundary. Headers are now normalized once when the composed dispatcher is entered, and already-normalized plain objects are passed through without a copy. The existing DNS redispatch cases still pass with the normalized object.

The updated commit is `88e54638`. The relevant interceptor tests and lint pass locally.

lib/dispatcher/dispatcher.js
```diff
@@ -1,6 +1,75 @@
...
+    if (normalized) {
+      return headers
+    }
+  }
+
+  const normalized = {}
+
+  if (Array.isArray(headers)) {
+    if (headers.length > 0 && Array.isArray(headers[0])) {
+      if (headers.some(header => !Array.isArray(header) || header.length !== 2)) {
```
metcoder95: let's try to prefer `for-loops` over array methods; this will sit in a hot-path if interceptors are used

lib/dispatcher/dispatcher.js
```diff
@@ -1,6 +1,75 @@
...
+  if (current === undefined) {
+    headers[key] = Array.isArray(value) ? value.slice() : value
+  } else if (Array.isArray(current)) {
+    current.push(...values)
+  } else {
+    headers[key] = [current, ...values]
+  }
+}
+
+function normalizeHeaders (headers) {
```
metcoder95: I'm a bit worried of all the branching, but unsure if it can be simplified

lib/dispatcher/dispatcher.js
GiHoon1123: Updated this to use a loop.

lib/dispatcher/dispatcher.js
GiHoon1123: I took another look at the request path. Normalizing in `compose()` means walking the headers to build an object, then `Request` walks them again to build its internal array. It also adds that work for interceptors that never read request headers.

Would it be better to expose a shared normalization helper for interceptors that need it? If normalization should happen automatically, it seems like we’d need a shared internal representation that `Request` can reuse. What do you think?

metcoder95: That's a good idea, go for it

GiHoon1123: Implemented in 058d9124. `normalizeHeaders()` is now exposed through `util`, and cache and deduplicate use the same helper. I removed eager normalization from `compose()`, so interceptors that do not inspect headers no longer pay the extra cost. Types, docs, and the relevant tests have been updated.

metcoder95 (approved):

PR #5707, opened 2026-08-21
perf(h1): drop idle-socket timer floor with a ref'd setImmediate

anonrig:
## This relates to...

- Split out of #5701 per review
- #5493 (idle-socket validation ~1.3ms timer floor)
- #5499 / #5606 (setImmediate revert due to unref'd Immediate poll stall)
- GHSA-35p6-xmwp-9g52 (response-queue poisoning; validation must stay)

## Rationale

Sequential `fetch()` / HTTP/1.1 keep-alive reuse was paying Node's `setTimeout(0)` timer floor (~1.3–1.5ms) on every request. That dominates loopback / low-RTT workloads.

`#5499` switched this to `setImmediate` and was reverted in `#5606` because an *unref'd* Immediate lets poll block for ~500ms when the event loop is otherwise idle.

## Changes

- Defer idle-socket validation with a **ref'd `setImmediate`** so it still runs after poll (poisoning protection) without the 1ms timer floor or the #5606 stall.
- Add a `fetch` keep-alive reuse stall regression test alongside the existing Pool test.
- Add `benchmarks/fetch/sequential-keepalive.mjs`.

### Features

N/A

### Bug Fixes

N/A

### Breaking Changes and Deprecations

N/A

## Benchmarks

Re-run 2026-08-21 on Node v22.14.0 (linux x64, Intel Xeon, 4 cores). Medians of 5 runs vs `main` @ `181c293`.

Sequential keep-alive `fetch().text()` on loopback (`connections: 1`, `pipelining: 1`, 2000 iterations after 300 warmup):

| | p50 | p90 | p99 |
|---|---|---|---|
| main (`setTimeout(0)`) | 1.47 ms | 1.63 ms | 1.84 ms |
| this PR (`setImmediate`) | **0.20 ms** | **0.29 ms** | **0.48 ms** |

~7.5× faster at p50.

## Status

- [x] I have read and agreed to the [Developer's Certificate of Origin][cert]
- [x] Tested
- [x] Benchmarked (**optional**)
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

[cert]: https://github.com/nodejs/undici/blob/main/CONTRIBUTING.md#developers-certificate-of-origin

PR #5781, opened 2026-09-07
docs(errors): document ProxyConnectionError (UND_ERR_PRX_CONN)

kilisamemarisaaa:
## This relates to...

Follow-up documentation for #5707, which introduced `ProxyConnectionError` (`UND_ERR_PRX_CONN`) in v8.10.1. No existing issue tracks this documentation gap — found by comparing `lib/core/errors.js` exports and `types/errors.d.ts` against `docs/docs/api/Errors.md`.

## Rationale

`ProxyConnectionError` is public API: it is exported from the `errors` namespace (`lib/core/errors.js`), declared in `types/errors.d.ts`, and thrown by `lib/dispatcher/proxy-agent.js` when the proxy connection is torn down so the request fails instead of retrying forever. Its sibling `SecureProxyConnectionError` (`UND_ERR_PRX_TLS`) is documented in full, but `ProxyConnectionError` has no class section and no row in the error-code documentation, so users cannot discover or match on a public error class that their proxies can produce at runtime.

While auditing I also verified the only other code absent from the docs table, `UND_ERR_HTTP_PARSER`, is intentionally different (documented as an `HPE_`-prefixed `code` on the `HTTPParserError` class, which extends `Error` directly) — so this is the single genuine documentation gap.

## Changes

- Adds a `ProxyConnectionError` class section to `docs/docs/api/Errors.md`, mirroring the structure of the sibling `SecureProxyConnectionError` section: YAML `added`/`changes` block (v8.10.1, #5707), `Extends: {UndiciError}`, the `name`/`code`/`cause` field list matching the implementation, and the `new ProxyConnectionError(cause[, message[, options]])` constructor signature matching both the implementation and the type declaration.
- Places the section directly before `SecureProxyConnectionError` (base class before its subclass).

### Features

N/A

### Bug Fixes

N/A (documentation only)

None

## Status

- [x] Tested (S — documentation-only change; docs are not covered by the test suites)
- [x] Benchmarked (S — not applicable)
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/docs/docs/api/Errors.md
+++ b/docs/docs/api/Errors.md
@@ -446,6 +446,31 @@ The response returned an error status code. This is raised, for example, when th
   * `headers` {Object|string[]|null} The response headers. (optional)
   * `body` {Object|string|null} The response body. (optional)
 
+## Class: `ProxyConnectionError`
+
+<!-- YAML
+added: v8.10.1
+changes:
+  - version: v8.10.1
+    pr-url: https://github.com/nodejs/undici/pull/5707
+    description: Added to fail the request instead of retrying forever when the proxy connection is torn down.
+-->
+
+* Extends: {UndiciError}
+
+A connection to the proxy failed in a way that cannot be recovered on the
+same connection, so the request fails instead of being retried.
+
+* `name` {string} Always `'ProxyConnectionError'`.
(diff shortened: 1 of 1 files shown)


mcollina (approved):
lgtm

PR #5793, opened 2026-09-09
revert #5578

KhafraDev:
reverts https://github.com/nodejs/undici/pull/5578 / 10898087d0d041bdee236502be8473d9c2328c4a

# USE WS IF YOU NEED TO DO THIS

# why does a *browser implementation of WebSocket* need all this bloat?


mcollina (changes requested):
Because this is how Node.js is supposed to interact with WHATWG spec objects that need this functionality and make the object ref'd / unref'd.

It has been there for quite some time:

https://nodejs.org/docs/latest/api/process.html#processrefmayberefable

cc @jasnell who added those in to Node.js in the first place.

jasnell (changes requested):
This isn't a browser implementation of `WebSocket` and the all caps in the PR description is really unnecessary.

mcollina:
Something that could be improved is making those property non-enumerable.

KhafraDev:
> This isn't a browser implementation of WebSocket

I can't tell if this is a serious argument, not to be rude, this is so insane I laughed when I first read it. I implemented literally the entire thing to be a browser compatible WebSocket (and for emphasis, since caps are off the table, literally the entire thing minus a few bug reports)

image

image

If it's not a browser implementation of WebSocket, then it shouldn't exist. I'll have a PR that aliases it to ws in no time.

KhafraDev:
and before the semantics, there is functionally no difference (to me) between "browser compatible" and "browser implementation". `ws` and `node-fetch` are not browser compatible despite offering the same APIs as browser, even if they implement similar ones. If they did, then the equivalent implementations in undici would not have to exist.

KhafraDev:
tbh I wrote 2 posts arguing over it not being a "browser implementation", but I deleted them since I can't be bothered arguing. I will say that if WebSocket is not a browser implementation (since it certainly isn't compatible), then it should not exist. The implementation of WebSocket is embarrassingly bad and I regret ever adding it, since I could have saved myself so much time by doing `npm i ws` and `module.exports.WebSocket = require('ws').WebSocket`.

jasnell:
Aside from the fact that there's no reason to remove these, they have shipped in Node.js releases and would need a proper deprecation cycle there to be removed.

KhafraDev:
the refable protocol is still experimental, it should be fine to remove these

KhafraDev:
If you plan on making it stable (https://github.com/nodejs/node/pull/66213), we should remove it here and potentially reland it at a later time so we can discuss its value beforehand. @mcollina

jasnell:
That still begs the question of how these are, in any way, problematic in any defensible sense. Removing them makes no sense given it addresses an actual problem. We obviously have a difference of opinion, so let's go to the specs to resolve it. Can you point to any spec requirement that these violate? Any Web Platform Test that fails?

KhafraDev:
If you want my honest opinion I find it absolutely insane how someone who has never contributed to undici before can come in, simply due to them being in the tsc, and then demand that the maintainers keep some random node feature for no concernable reason. Then when I want it removed I have to prove some nonsense argument that wasn't needed to implement it in the first place!

The argument you are trying to make does not matter, I don't know why you keep bringing it up. I don't have a sign on my house saying "please don't enter my house" but people still know that, even if the door is unlocked, not to walk into my house. Just because the webidl spec doesn't tell you how to tie your shoes, it doesn't mean you should walk around with your laces untied. The WPTs don't fail if the properties aren't enumerable or there aren't brand checks because the WPTs aren't going to add in the same test for thousands of different interfaces!

This is observable on the WebSocket prototype. That is a good enough reason to remove it. Everything else is a gotcha question. Right now the refable protocol is experimental, meaning that this is literally the last time we can remove these without a major version bump. If you can convince me that these are absolutely necessary for the sake of node or undici and somehow ws doesn't cut it, they could even be reimplemented, although extremely unlikely.

Btw, here are the WPTs that require an exact prototype shape:

- streams/readable-streams/general.any.js:90: requires ReadableStreamDefaultController.prototype to have exactly ['close', 'constructor', 'desiredSize', 'enqueue', 'error'].
- streams/readable-streams/async-iterator.any.js:27: requires the ReadableStream async iterator prototype to have exactly ['next', 'return'].
- html/editing/dnd/datastore/datatransferitemlist-indexed-getter.html:22: DataTransferItemList.prototype must be exactly add, clear, constructor, length, remove.
- html/editing/dnd/datastore/datatransferitemlist-indexed-getter.html:22: DataTransferItemList.prototype must be exactly add, clear, constructor, length, remove.
- sanitizer-api/sethtml-with-trustedtypes-immutable.tentative.html:82: TrustedHTMLParserOptions.prototype must be ['constructor'] only.
- console/console-is-a-namespace.any.js:26 and resources/idlharness.js:3429: the console namespace's [[Prototype]] must have no own keys.
- web-animations/interfaces/{Animation,KeyframeEffect}/style-change-events.html: these run a test for every enumerable key from Object.keys(X.prototype). An extra enumerable property probably fails there because it has no matching entry in the test table (not verified).

jasnell:
> ... Btw, here are the WPTs that require an exact prototype shape

None of which fail in the presence of these properties on `WebSocket`. Also, none of the listed tests "request an exact prototype shape" for `WebSocket`, which is the API in question here.

Let's be clear... There's nothing in the relevant specs or web platform tests whatsoever that make the presence of these symbols on `WebSocket` invalid in any way *and* they address a known [reported bug](https://github.com/nodejs/undici/issues/5570). And since you refuse to identify anything that is actually broken by the presence of these symbols, I fail to see any reason why they should be reverted.

KhafraDev:
> None of which fail in the presence of these properties on WebSocket. Also, none of the listed tests "request an exact prototype shape" for WebSocket, which is the API in question here.

There aren't any tests for WebSocket because no one has ever added garbage to its prototype! For the same reason there aren't checks that the properties on its prototype are enumerable and for the same reason there aren't tests for brand checks on WebSocket methods or getters.

If I made a bug report titled "there are now 2 non-spec compliant symbols on the websocket prototype", would that suddenly make it a valid issue? Someone reported a month ago that "SharedArrayBuffer bodies are stringified in the Response constructor" which was an invalid issue.

jasnell:
* The change landed correctly with appropriate review per the guidelines/conventions for this repo.
* It fixes a real bug. As the author of the node.js refable protocol, I can attest that this is exactly the type of bug it is intended to address.
* It has already shipped in multiple undici and node.js releases.
* It is in no way a violation of any normative spec requirement.
* It does not cause any known existing tests to fail.
* It does not cause any known existing application code to fail.
* You have refused to offer any evidence of any known test or application failures caused.
* What you have offered (the list of WPTs above) are not relevant to the `WebSocket` class.
* Beyond your own opinion, you've offered no technical justification for why it should be reverted.

I appreciate your opinion, but I disagree with it. From what I can determine, your request to revert has not reached a consensus even among undici WG members.

I'm not an undici WG member. I'm not the one you need to convince. I was pinged, I gave my opinion in response to that ping.

On the more general point: For all Web and Internet related specifications, whether we're talking IETF, W3C, WHATWG, ECMA, etc, when a spec is silent on a particular topic, the behavior is effectively implementation-defined. *IF* that causes interop problems, the fix goes through an issue, erratum, spec change typically backed by a Web Platform Test, test262, or equivalent change. And the burden there is on proving that there's an actual interop problem.

_Nobody gets to argue a requirement into existence_, which is precisely what you are attempting to do here. Rather than continuing to repeat the same arguments, it would be more helpful/productive to demonstrate any *actual* problems caused by the presence of these symbol properties.

KhafraDev:
> It has already shipped in multiple undici and node.js releases.

It is experimental. It can be removed in a semver-minor release.

> It does not cause any known existing application code to fail.
> You have refused to offer any evidence of any known test or application failures caused.

I have shown code multiple times that failed and you call it theoretical. You are arguing in bad faith.

> What you have offered (the list of WPTs above) are not relevant to the WebSocket class.

I have refuted this two times now.

> Beyond your own opinion, you've offered no technical justification for why it should be reverted.

and this is why I don't bother. How much clearer could I be? I have to repeat myself 3 times for every argument because your personal opinions don't correspond with actual facts. It is observable on the prototype. The webidl spec does not allow it. The websocket spec does not allow it. The only mention of "extending interfaces" or "applying interfaces not defined in the spec" is in the html spec, which specifically says implementations are "strongly discouraged" from doing so.

I have proven that there is precedence in the WPTs for similar tests. No one has wasted time testing this in WebSocket specifically (or Request, or Response, or FormData, or EventSource) because no one has ever made the poor decision of adding non-spec compliant methods or otherwise on its prototype.

I thought "observable on the prototype" was enough, given literally 5 years of precedence that any other undici maintainer should recognize. When fetch was first added, we thought it would be a good idea to allow users to be able to use node streams as Response bodies. This is before the existence of the stream helpers in node core and years before ReadableStream.from (it hadn't even been conceived at this point). So we allow async iterables as Response bodies and people start using it. Then those people ran their code in Deno (Bun didn't exist yet and Cloudflare Workers didn't offer node compat anywhere close enough to running undici), noticed it did not work, and then reported bugs in Deno. This happened so many times, a (now) former Deno employee had to make an issue in this repo asking us to remove it. It put pressure on Deno to implement a feature that they did not want to or else users would stop using Deno or stop offering compatibility with it for the convenience node set out. 
When we implement a spec, there is a contract that we agree to, not only for node's benefit, but the benefit of the entire ecosystem. When people use specs in node, there is an expectation that the code they write in node will work everywhere else. The existence of these symbols breaks that guarantee, to the detriment of node.

> consensus even among undici WG members.

That is why I pinged Matteo in the earlier message, so we could land this, and then give it a fair discussion. This is literally the last chance we have of removing these without a semver-major release. And I will keep fighting to have this land, regardless of it landing in a minor or major release.

> Nobody gets to argue a requirement into existence, which is precisely what you are attempting to do here.

I hope other people reading this find it as ironic as I do... You come out of the blue, use tsc powers to block this from landing, and then spend days throwing strawmans at me in an attempt at a quick gotcha. Then when I bring up actual concerns, I get hit with "oh no it's theoretical" (actual piece of working code btw) or the goalposts get changed.

I would very much like to hear from the other members of the undici team.

KhafraDev:
Okay, I'll offer a truce.

Would you be fine with dropping the symbols if we added `unref(websocket)` and `ref(websocket)`? We can easily implement it in the same way as we do [`ping(websocket)`](https://github.com/nodejs/undici/blob/main/docs/docs/api/WebSocket.md#pingwebsocket-payload), without any of my concerns.

jasnell:
> ... use tsc powers to block this from landing

You do understand that there's no such thing as "tsc powers", right? A TSC member's -1 caries the same weight here as as -1 from anyone who isn't part of the undici WG.

I'm happy that the conversation is finally focusing on addressing the bug. I'm still -1 on reverting because no case has been made to convince me otherwise and the current solution fixes the bug in a perfectly reasonable way.

PR #5797, opened 2026-09-10
fix(cache): preserve 304 response event order with async stores

marcopiraccini:
## This relates to...

N/A

## Rationale

A cache revalidation can receive a 304 before an asynchronous store has returned the cached response. The 304 itself has no body, but the cached body must be replayed before downstream receives the response end.

## Changes

- Pause a reusable 304 while the cache lookup and cached-body replay complete.
- Use a private controller wrapper so downstream backpressure cannot release that internal pause early.
- Resume the origin response only after replay completes, preserving start, cached data, then end ordering.
- Handle failed cache lookups as misses and abort cleanly if asynchronous 304 handling throws.
- Add coverage for synchronous and asynchronous stores, replay backpressure, write-stream failures, closed streams, and the asynchronous abort path.

### Features

N/A

### Bug Fixes

- Preserve response event ordering when a 304 is revalidated through an asynchronous cache store.
- Prevent cached-body replay from being cut short by an early 304 end.

N/A

## Status

- [x] I have read and agreed to the [Developer Certificate of Origin][cert]
- [x] Tested
- [S] Documented
- [x] Review ready
- [x] In review
- [ ] Merge ready

### DIFF
--- a/lib/handler/cache-handler.js
+++ b/lib/handler/cache-handler.js
@@ -150,6 +150,12 @@ class CacheHandler {
   #handler
 
   /**
+   * Pending 304 resolution (async store lookup or cached body replay).
+   * @type {Promise<void> | null}
+   */
+  #pending304 = null
+
+  /**
    * @type {import('node:stream').Writable | undefined}
    */
   #writeStream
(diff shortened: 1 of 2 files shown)


lib/handler/cache-handler.js
```diff
@@ -345,59 +351,68 @@ class CacheHandler {
...
-          const bodyIterator = cachedValue.body.values()
-
-          const streamCachedBody = () => {
-            for (const chunk of bodyIterator) {
-              const full = this.#writeStream.write(chunk) === false
-              this.#handler.onResponseData?.(controller, chunk)
-              // when stream is full stop writing until we get a 'drain' event
-              if (full) {
-                break
+          return new Promise((resolve) => {
```
metcoder95, line 354: I'd prefer if we use a callback instead of creating a new Promise; in that way, regardless of the nature, we minimize overhead as well keep things a bit consistent.

mcollina: I don't think it's the contract.

lib/handler/cache-handler.js
```diff
@@ -445,6 +466,11 @@ class CacheHandler {
   }
 
   onResponseData (controller, chunk) {
+    if (this.#pending304) {
+      this.#pending304 = this.#pending304.then(() => this.onResponseData(controller, chunk))
```
mcollina, line 470: this create an infinite chain of pending promises

mcollina (changes requested):

marcopiraccini:
@mcollina PTAL

mcollina (approved):
LGTM, much much better

mcollina:
@metcoder95 PTAL

metcoder95 (approved):

PR #5802, opened 2026-09-11
feat(h2): honor maxRequestsPerClient

### LINKED ISSUE #5787: Support maxRequestsPerClient for HTTP/2 sessions
## This would solve...

`maxRequestsPerClient` currently limits the number of requests sent over an HTTP/1.1 connection, but it does not appear to apply after HTTP/2 is negotiated.

The option is documented as:

> The maximum number of requests to send over a single connection before the socket is reset.

In Undici 8.10.0, the request counter is enforced in [`client-h1.js`](https://github.com/nodejs/undici/blob/v8.10.0/lib/dispatcher/client-h1.js#L1296-L1298). The [HTTP/2 implementation](https://github.com/nodejs/undici/blob/v8.10.0/lib/dispatcher/client-h2.js#L12-L42) does not import or use `kMaxRequests`.

This means applications using `allowH2` cannot place a request-count lifetime on an HTTP/2 session. We encountered a case where an HTTP/2 session remained logically reusable after its transport stopped making progress. Subsequent requests continued selecting that session. Periodically retiring a session after a bounded number of requests would provide useful defense in depth and limit how long unhealthy or stale sessions can be reused.

When HTTP/2 is negotiated and `maxRequestsPerClient` is a positive number:

- Count requests/streams opened on the current HTTP/2 session.
- Once the limit is reached, stop assigning new requests to that session.
- Allow existing streams to complete, then gracefully retire the session.
- Establish a new session for queued or subsequent requests.
- Preserve the existing `null`/`0` behavior that disables the limit.

The exact behavior around concurrent streams could follow maintainer guidance—for example, whether the limit is strict or may be exceeded by streams already opened concurrently.

## I have also considered...

- Rotating the dispatcher in application code after request timeouts. This works as a reactive mitigation, but requires applications to own dispatcher lifecycle and duplicate session-health logic.
- HTTP/2 PING frames. These help with liveness detection but do not provide a bounded session lifetime.
- A separate HTTP/2-specific option. That would be workable, but extending `maxRequestsPerClient` seems consistent with its documented connection-level semantics.

## Additional context

Environment where this was observed:

- Node.js 24
- Undici 8.10.0
- HTTP/2 enabled with `allowH2`

I would be happy to help test the behavior or contribute a change with guidance on the preferred HTTP/2 retirement semantics.

johnlyonms:
## This relates to...

Closes #5787

## Rationale

`maxRequestsPerClient` was only honored by the HTTP/1.1 dispatcher. A `Client` created with `allowH2: true`, or an `H2CClient`, ignored the option and kept a single HTTP/2 session alive indefinitely. The option is important for rotating
connections so load balancers can rebalance and clients do not remain pinned to one backend.

HTTP/2 also requires bounded retirement. Once a client has reached its request limit and we mark it as such, we must wait for existing streams to close. Waiting indefinitely for accepted streams to close lets a long-lived SSE, CONNECT, WebSocket, long-polling, gRPC, or stalled upload would block all queued work if we didn't implement a bound. 

## Changes

- Count successfully opened HTTP/2 streams against `maxRequestsPerClient`.
- Retire a session synchronously when it reaches the limit so no additional stream can slip through the same resume loop.
- Allow accepted streams to drain without truncating ordinary responses or request uploads.
- Bound retirement draining with `headersTimeout`. If accepted streams remain open at the deadline, reset the retired session with an informational error so queued requests remain replayable on a fresh connection.
- Cancel the retirement timer during normal drain, GOAWAY, reset, and session close paths.
- Preserve the existing `0`, `null`, and `undefined` disabled behavior.
- Document the HTTP/1.1 message-counting and HTTP/2 stream-counting behavior for `Client`, `H2CClient`, and TypeScript consumers.

The implementation tracks physical stream closure separately from `kOpenStreams`. A response may finish while its request body is still uploading, so response completion alone is not sufficient to safely retire the session.

## Tests

`test/http2-max-requests-per-client.js` contains 17 tests covering:

- sequential and concurrent session rotation
- limits of `1` and disabled limits
- `Client`, `H2CClient`, and `Pool`
- graceful drain and complete request-body uploads
- failed, aborted, upgrade, and WebSocket streams
- GOAWAY and abrupt peer disconnect races
- client destruction during retirement
- exactly one disconnect notification
- an endless SSE stream being terminated after `headersTimeout` while queued work reconnects successfully

Validation:

- Focused HTTP/2 suite: 17/17 passed
- ESLint passed for the changed JavaScript files
- Syntax and whitespace checks passed

## Commits

- `12ac5172 feat(h2): honor maxRequestsPerClient`
- `41ba964e fix(h2): bound retired session draining`

None. HTTP/1.1 behavior is unchanged. HTTP/2 users who do not configure `maxRequestsPerClient`, or set it to `0`, see no change. Setting `headersTimeout: 0` explicitly disables the retirement drain deadline.

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/docs/docs/api/Client.md
+++ b/docs/docs/api/Client.md
@@ -49,6 +49,8 @@ added: v1.0.0
   * `headersTimeout` {number|null} The timeout, in milliseconds, the parser
     waits to receive the complete HTTP headers before the request times out. Use
     `0` to disable it entirely. **Default:** `300e3`.
+    Over HTTP/2 this also bounds how long a session retired by
+    `maxRequestsPerClient` may wait for its active streams to close.
     HTTP/1.1 headers/body parser timeouts are not guaranteed to fire with exact
     millisecond precision: delays up to 1000ms use native timers, while larger
     delays use undici's lower-overhead fast timers with a target resolution
@@ -78,9 +80,9 @@ added: v1.0.0
     Over HTTP/1.1 a request is counted per message and the socket is reset once
     the limit is reached. Over HTTP/2 a request is counted per successfully
     opened stream: the session is retired once the limit is reached, meaning it
-    accepts no further streams and the streams it already accepted are allowed
-    to complete before the connection is closed. Queued and subsequent requests
-    are dispatched on a new session.
+    accepts no further streams. Streams it already accepted are allowed to
+    complete for up to `headersTimeout` before the connection is reset. Queued
+    and subsequent requests are then dispatched on a new session.
   * `localAddress` {string|null} The local IP address the socket should connect
     from. **Default:** `null`.
   * `pipelining` {number|null} The number of concurrent requests sent over the
--- a/docs/docs/api/Client.md
+++ b/docs/docs/api/Client.md
@@ -75,6 +75,12 @@ added: v1.0.0
   * `maxRequestsPerClient` {number|null} The maximum number of requests to send
     over a single connection before the socket is reset. Use `0` to disable this
     limit. **Default:** `null`.
+    Over HTTP/1.1 a request is counted per message and the socket is reset once
+    the limit is reached. Over HTTP/2 a request is counted per successfully
+    opened stream: the session is retired once the limit is reached, meaning it
+    accepts no further streams and the streams it already accepted are allowed
+    to complete before the connection is closed. Queued and subsequent requests
+    are dispatched on a new session.
   * `localAddress` {string|null} The local IP address the socket should connect
     from. **Default:** `null`.
   * `pipelining` {number|null} The number of concurrent requests sent over the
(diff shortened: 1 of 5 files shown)


docs/docs/api/Client.md
```diff
@@ -75,6 +77,12 @@ added: v1.0.0
   * `maxRequestsPerClient` {number|null} The maximum number of requests to send
     over a single connection before the socket is reset. Use `0` to disable this
     limit. **Default:** `null`.
+    Over HTTP/1.1 a request is counted per message and the socket is reset once
+    the limit is reached. Over HTTP/2 a request is counted per successfully
+    opened stream: the session is retired once the limit is reached, meaning it
+    accepts no further streams. Streams it already accepted are allowed to
+    complete for up to `headersTimeout` before the connection is reset. Queued
+    and subsequent requests are then dispatched on a new session.
```
metcoder95, line 85: This can be a bit confusing for users, especially given that we have the `maxConcurrentStreams`; let's mention how they differ and how they interact with each other

Issue #3201, opened 2024-05-04
WebSocket performance / benchmarking


KhafraDev:
I haven't seen any benchmarks for undici's / node's WebSocket yet and not much has been done to improve performance.

tsctx:
What do you think of this? https://github.com/tsctx/ws-masking/blob/main/initialize.js
It uses int32 and processes 4 chunks at a time, making it 8 times faster than the current implementation.

Issue #2927, opened 2024-03-06
Benchmarks and Optimizations


metcoder95:
### Discussed in https://github.com/nodejs/undici/discussions/2891

<div type='discussions-op-text'>

<sup>Originally posted by **metcoder95** February 29, 2024</sup>
Hey! 👋 

Wanted to open a discussion just to get your thoughts on this as I have been thinking about this topic for a while.

I've seen recently quite a few PRs for doing optimizations over `fetch` especially, and I believe and appreciate all of them as it's super helpful to have people contributing to those ends. Highly appreciated.

Nevertheless, I'm starting to see a pattern of constant micro-optimizations (sometimes companioned by micro-benchmarks) that focus on pretty specific parts of the whole flow on `fetch` that often than not might not make a big impact if we take a holistic view over the `fetch` implementation (e.g. cold-paths, use cases that are not often seen, etc.).

I believe that those contributions can be of even greater value if we shift that focus to the holistic view of `fetch` and how the numbers change based on the changes proposed and if they are of impact or not, as it will help assess the value of the change as well as what other parts we can shift our focus to make `fetch` faster (or even `undici`).

For instance we can focus on having a better benchmark setup for `fetch` only, its Web APIs (e.g. `Cookies`, `CacheStorage`, `WebSocket`, etc.).

Feel free to discard the idea, or maybe propose something different 🙂 
</div>

> I'll try to draft something later this week, but any suggestion is welcomed

PR #5809, opened 2026-09-12
perf(websocket): unroll client frame masking

dexhunter:
## This relates to...

Existing issues [#3201](https://github.com/nodejs/undici/issues/3201) and [#2927](https://github.com/nodejs/undici/issues/2927).

## Rationale

Client WebSocket frames are masked before they are written. The masking path currently performs one loop iteration per byte. This change keeps the existing mask generation and wire format, while processing eight bytes per iteration and handling the remaining bytes with the existing byte-wise semantics.

## Changes

- Unroll the masking loop with four preloaded mask bytes.
- Keep the existing mask generation, frame headers, caller-buffer behavior, and byte tail unchanged.
- Add coverage for short, boundary, aligned, unaligned, binary, and UTF-8 text payloads.

### Features

N/A

### Bug Fixes

Reduces the CPU work in the maintained WebSocket masking path without changing frame semantics.

N/A

## Validation

- `node --test test/websocket/frame.js`: 3 passed.
- `npm run lint`: passed.
- The WebSocket suite reported 132 passed, 5 failed, and 2 cancelled. The H2 failures and 180-second timeouts reproduce on the clean base and are retained as baseline/environment failures; the suite is not presented as fully green.
- Independent correctness-checking echo measurements used the maintained 256 KiB WebSocket workload with Node v22.23.1, `ws` 8.20.1, loopback, no compression, one connection, and one in-flight message. The five paired runs used AB BA AB BA AB order, with two seconds of warmup and one second of measurement per cell. The primary binary median of per-run p50 values changed from 876.749 µs to 797.448 µs (-9.045%). All five paired rows improved; median p99 improved 3.915%, CPU time improved 9.426%, and the small-payload controls were flat.
- These are short-window measurements on a shared host and cover the WebSocket masking path only; they do not claim an improvement for HTTP or Fetch.
- [Weco trajectory](https://dashboard.weco.ai/share/FxmhE3XvJnpUvtmX-ZFNcwgZfJFDnFva)

## Status

- [x] Tested
- [x] Benchmarked
- [ ] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

### DIFF
--- a/lib/web/websocket/frame.js
+++ b/lib/web/websocket/frame.js
@@ -20,6 +20,31 @@ function generateMask () {
   return [buffer[bufIdx++], buffer[bufIdx++], buffer[bufIdx++], buffer[bufIdx++]]
 }
 
+function maskPayload (source, target, targetOffset, maskKey) {
+  const length = source.byteLength
+  const mask0 = maskKey[0]
+  const mask1 = maskKey[1]
+  const mask2 = maskKey[2]
+  const mask3 = maskKey[3]
+  const unrolledEnd = length & ~7
+
+  let i = 0
+  for (; i < unrolledEnd; i += 8) {
+    target[targetOffset + i] = source[i] ^ mask0
+    target[targetOffset + i + 1] = source[i + 1] ^ mask1
+    target[targetOffset + i + 2] = source[i + 2] ^ mask2
+    target[targetOffset + i + 3] = source[i + 3] ^ mask3
+    target[targetOffset + i + 4] = source[i + 4] ^ mask0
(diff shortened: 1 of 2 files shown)


mcollina:
@tsctx PTAL

KhafraDev:
The title of this PR is "perf" but there are no benchmarks nor benchmark results.

tsctx:
Sorry for the accidental operation, I'm still getting used to GitHub.
There is no presentation of benchmark results, and I already have a better version than this PR.

PR #5819, opened 2026-09-14
fix(cache): do not report aborted on successful cache hits

### LINKED ISSUE #5818: Cache interceptor reports aborted=true during successful onResponseEnd
## Bug Description

On a cache hit, `controller.aborted` is true inside `onResponseEnd()`, despite successful completion without cancellation.

## Reproduction

```js
'use strict'

const { Agent, interceptors } = require('undici')

test('successful cached response is not aborted', async t => {
    res.setHeader('Cache-Control', 'max-age=60')
    res.end('ok')
  })
  t.after(() => server.close())
  server.listen(0)

  const dispatcher = new Agent().compose(interceptors.cache())
  t.after(() => dispatcher.destroy())
  const opts = { origin: `http://127.0.0.1:${server.address().port}`, path: '/', method: 'GET' }
  await (await dispatcher.request(opts)).body.dump()

  const aborted = await new Promise((resolve, reject) => {
    dispatcher.dispatch(opts, {
      onRequestStart() {},
      onResponseEnd(controller) { resolve(controller.aborted) },
      onResponseError(controller, error) { reject(error) }
    })
  })
  t.assert.strictEqual(aborted, false)
})
```

## Expected Behavior

`controller.aborted` is `false`, as for an uncached response.

## Actual Behavior

`controller.aborted` is `true`.

## Environment

- OS: Ubuntu 24.04.5 LTS
- Node.js version: v26.8.2
- undici version: 8.10.2

### Additional context

The [cache controller's getter](https://github.com/nodejs/undici/blob/main/lib/interceptor/cache.js) returns `stream.destroyed`, which also becomes true after normal completion.

In jsdom, this caused successful cached loads to fail with abort errors, and could incorrectly cancel requests following cached redirects.

Yahiro025:
## This relates to...

Fixes #5818

## Rationale

A successful cache hit was reporting `controller.aborted === true` in `onResponseEnd()` because `controller.aborted` aliased `stream.destroyed`, and Readable streams are destroyed after a normal end.

### Bug Fixes

- Track abort with an explicit flag set only in `abort()`, so successful cache hits report `aborted === false` while real cancellations still report `true`.

### Features

N/A

N/A

## Status

- [x] Tested
- [x] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

## Test plan

- [x] Added issue repro + abort-path check in `test/interceptors/cache.js`
- [x] `test/interceptors/cache.js` (95 tests) and `npm run test:cache-interceptor` passed

## AI disclosure

I used Cursor to help draft the fix and tests. I reviewed the diff and verified the regression coverage before opening this PR.


mcollina (approved):
lgtm

PR #5820, opened 2026-09-14
fix(cache): track cached response abort state

### LINKED ISSUE #5818: Cache interceptor reports aborted=true during successful onResponseEnd

mcollina:
Track cached-response controller abort state independently from the backing stream lifecycle, so normal cache completion no longer reports an abort. Adds regression coverage for the cache-hit path.

Fixes #5818


metcoder95 (commented):
just conflicts and lgtm

metcoder95 (approved):

PR #3263, opened 2024-05-15
permessage-deflate decompression support in websocket

KhafraDev:
https://www.rfc-editor.org/rfc/rfc7692

Famous last words, but I'm looking at the spec and it doesn't seem too difficult to implement.

PR #5827, opened 2026-09-15
fix(websocket): reset the compressed flag when a message completes

askalf:
## Summary

- `ByteParser` never cleared `#info.compressed` after delivering a permessage-deflate message, so the flag described the *previous* message while the next frame header was parsed.
- The guard at `lib/web/websocket/receiver.js:171` that rejects a continuation frame with no message in progress is conditioned on `!this.#info.compressed`, so it was skipped after any compressed message.
- Result: a stray continuation frame arriving after a compressed message is decompressed and delivered to the application as an extra message, instead of failing the connection with 1002 as RFC 6455 §5.4 requires.
- Fix: clear the flag in `consumeFragments()`, the one place where the message the flag describes is completed (both `websocketMessageReceived` call sites go through it).
- 8 regression tests added to `test/websocket/receiver-unit.js` (plus 1 pre-existing test in the file, 9 total): 4 discriminate on this bug and 4 are declared controls.

Both arms below were measured at head `998d2a463b8c348f1c3029ba728a818e3bb423c7`, with all 9 tests present in the file and only `lib/web/websocket/receiver.js` swapped between arms.

```console
$ git checkout origin/main -- lib/web/websocket/receiver.js   # base arm: fix reverted, tests kept
$ node --test test/websocket/receiver-unit.js
✔ ByteParser rejects 64-bit payload lengths with a non-zero upper word (27.118782ms)
✖ ByteParser fails the connection on a continuation frame after a compressed message (119.98587ms)
✔ ByteParser fails the connection on a continuation frame after an uncompressed message (control) (105.086991ms)
✔ ByteParser still assembles a fragmented compressed message (control) (106.584768ms)
✖ ByteParser fails the connection on a continuation frame after a multi-fragment compressed message (157.444873ms)
✖ ByteParser fails the connection on a continuation frame separated from a compressed message by a control frame (155.700874ms)
✖ ByteParser fails the connection on a non-empty continuation frame after a compressed message (105.319193ms)
✔ ByteParser delivers two consecutive compressed messages (control) (106.370579ms)
✔ ByteParser keeps the compressed flag across a control frame inside a fragmented message (control) (154.134819ms)
ℹ tests 9
ℹ pass 5
ℹ fail 4
✖ failing tests:
✖ ByteParser fails the connection on a continuation frame after a compressed message (119.98587ms)
    actual: [ 'message:5', 'message:0' ],
    expected: [ 'message:5', 'abort' ],
✖ ByteParser fails the connection on a continuation frame after a multi-fragment compressed message (157.444873ms)
    actual: [ 'message:5', 'message:0' ],
    expected: [ 'message:5', 'abort' ],
✖ ByteParser fails the connection on a continuation frame separated from a compressed message by a control frame (155.700874ms)
    actual: [ 'message:5', 'ping', 'message:0' ],
    expected: [ 'message:5', 'ping', 'abort' ],
✖ ByteParser fails the connection on a non-empty continuation frame after a compressed message (105.319193ms)
    actual: [ 'message:5', 'message:5' ],
    expected: [ 'message:5', 'abort' ],

$ git checkout HEAD -- lib/web/websocket/receiver.js   # fixed arm, head 998d2a46
$ node --test test/websocket/receiver-unit.js
✔ ByteParser rejects 64-bit payload lengths with a non-zero upper word (17.284235ms)
✔ ByteParser fails the connection on a continuation frame after a compressed message (113.10889ms)
✔ ByteParser fails the connection on a continuation frame after an uncompressed message (control) (102.501077ms)
✔ ByteParser still assembles a fragmented compressed message (control) (104.682335ms)
✔ ByteParser fails the connection on a continuation frame after a multi-fragment compressed message (156.259085ms)
✔ ByteParser fails the connection on a continuation frame separated from a compressed message by a control frame (155.52651ms)
✔ ByteParser fails the connection on a non-empty continuation frame after a compressed message (108.682308ms)
✔ ByteParser delivers two consecutive compressed messages (control) (107.327427ms)
✔ ByteParser keeps the compressed flag across a control frame inside a fragmented message (control) (154.650232ms)
ℹ tests 9
ℹ pass 9
ℹ fail 0
ℹ duration_ms 1309.719693
```

## Decisions

`consumeFragments()` is the correct place to clear `#info.compressed`: it is called from exactly the two `websocketMessageReceived` sites (`receiver.js:258` uncompressed, `receiver.js:290` compressed), i.e. every point where a message completes, and it already owns the rest of the per-message state reset (`#fragments`, `#fragmentsBytes`). Clearing the flag there keeps the reset with the state it belongs to and costs one assignment per message.

Alternatives rejected:
- *Clear the flag at the two `websocketMessageReceived` call sites.* Same effect, duplicated in two places, and a future third completion path would miss it.
- *Drop `!this.#info.compressed` from the guard at line 171.* Reverts the fix from #3263 and breaks compressed fragmented messages, which drain `#fragments` while still in progress — the `...still assembles a fragmented compressed message (control)` test is the control that would catch that.
- *Track a separate "message in progress" boolean.* Larger change, new state to keep in sync, no behaviour this one-line reset does not already give.

Not run: the `borp`-based `test:websocket` script (the full websocket suite, needs `node_modules`) and the Autobahn conformance suite — `node --test` on the touched file was used instead, covering the same code path unmodified.

AI assistance: this bug was found and the fix and tests were drafted with AI tooling in my workflow; the tests and checks above were executed as pasted. I'm responsible for the change and will handle review feedback.


mcollina (approved):
lgtm

KhafraDev:
these tests are terrible, please never contribute tests like those again, Claude

KhafraDev:
now I have to spend time removing them. They're so terrible I would call them actively harmful.

askalf:
Fair. Those tests faked the parser's handler and leaned on timers instead of driving a real connection the way the rest of `test/websocket` does. Sorry you had to spend time on them.

#5857 removes them and replaces them with one test in the shape of `continuation-frames.js`: a `ws` server sends a compressed message, then a bare continuation frame, and the client must fail the connection with 1002. If you would rather just delete them, close it. Won't submit tests like that again.

PR #5828, opened 2026-09-15
chore: update tests blocking release

vwong:
It looks like the breakage came from my PR https://github.com/nodejs/undici/pull/5677 interacting with https://github.com/nodejs/undici/pull/5799, this should unbreak the build.


mcollina (approved):
lgtm

PR #5832, opened 2026-09-15
fix(cookies): ignore a numeric expires outside the Date range

### LINKED ISSUE #5831: setCookie writes a broken Expires header when expires is an out-of-range number
## What happens

`setCookie` writes a broken `Expires` attribute if you pass `expires` as a number that is too large.

```js
import { Headers, setCookie } from 'undici'

const headers = new Headers()
setCookie(headers, { name: 'Space', value: 'Cat', expires: 1e16 })

console.log(headers.getSetCookie()[0])
// Space=Cat; Expires=undefined, undefined undefined NaN undefined:undefined:undefined GMT
```

That is not a valid date. Instead of an expiry time, the header contains the literal words `undefined` and `NaN`.

This is easy to hit in real code: `1e16` is what you get when you pass **seconds** where the API expects **milliseconds**. A date in seconds is always far too large for `Date`.

## Why it happens

`expires` can be a `Date` or a number. The check in `stringify` only understands the `Date` form:

```js
// lib/web/cookies/util.js
if (cookie.expires != null && cookie.expires.toString() !== 'Invalid Date') {
  out.push(`Expires=${toIMFDate(cookie.expires)}`)
}
```

The idea is "skip it if the date is invalid". That works for a `Date`, because a bad `Date` turns into the text `"Invalid Date"`. But a number never does:

- `(new Date(NaN)).toString()` is `"Invalid Date"` -> skipped, correct
- `NaN.toString()` is `"NaN"` -> not skipped
- `(1e16).toString()` is `"10000000000000000"` -> not skipped

So for a number the check always says "fine, go ahead". The value then goes into `toIMFDate`, which builds the date from a `Date` object whose values are all `NaN`. Reading `NaN` out of it gives `undefined` for the day name, and those values get pasted straight into the header text.

I confirmed this with a debugger on both lines. Below is the state at each stop.

### Stop 1 - the check passes even though the value is not a valid date

`cookie.expires` is `10000000000000000`. The check says `true`, so it does not skip. But `new Date(...)` on that value is invalid, and the value is outside the range a `Date` can hold:

paused in stringify, the guard evaluates to true while new Date gives NaN

| Expression | Value |
| --- | --- |
| `cookie.expires` | `10000000000000000` |
| `cookie.expires.toString()` | `'10000000000000000'` |
| `cookie.expires.toString() !== 'Invalid Date'` | **`true`** (so it does not skip) |
| `new Date(cookie.expires).getTime()` | **`NaN`** (but it is not a real date) |
| `Math.abs(cookie.expires) <= 8640000000000000` | **`false`** (out of range) |

### Stop 2 - `toIMFDate` builds the header out of `undefined` and `NaN`

Stepping into `toIMFDate`, the lookup that produces the day name returns `undefined`:

paused in toIMFDate, the IMFDays lookup returns undefined

| Expression | Value |
| --- | --- |
| `date` | `Invalid Date` |
| `date.getUTCDay()` | `NaN` |
| `IMFDays[date.getUTCDay()]` | **`undefined`** |
| `date.getUTCMonth()` | `NaN` |
| `date.getUTCFullYear()` | `NaN` |

The call stack at this point is `toIMFDate` -> `stringify` (`util.js:321`, which is inside the `if` body) -> `setCookie`. That confirms the check let the value through rather than skipping it.

## Which values are affected

```
expires: 0 (valid)           => Space=Cat; Expires=Thu, 01 Jan 1970 00:00:00 GMT
expires: 8.64e15 (max)       => Space=Cat; Expires=Sat, 13 Sep 275760 00:00:00 GMT
expires: 1e16 (out of range) => Space=Cat; Expires=undefined, undefined undefined NaN undefined:undefined:undefined GMT
expires: new Date(NaN)       => Space=Cat
```

output for valid, boundary, out-of-range and invalid Date values

Only an out-of-range **number** is broken. `0` works, the largest valid value works, and an invalid `Date` is skipped correctly. So the problem is specific to the number form.

## Suggested fix

Check the number itself instead of its text, and leave the `Date` behaviour alone:

```js
if (cookie.expires != null) {
  const expiresMs = cookie.expires instanceof Date
    ? cookie.expires.getTime()
    : cookie.expires

  if (typeof expiresMs === 'number' && Number.isFinite(expiresMs) && Math.abs(expiresMs) <= 8640000000000000) {
    out.push(`Expires=${toIMFDate(cookie.expires)}`)
  }
}
```

`8640000000000000` is the largest value a `Date` can hold, so anything that passes the check is safe to format. This matches how the neighbouring `Max-Age` branch already works - it checks the type rather than a converted string.

One thing worth noting: `Infinity`, `-Infinity` and `NaN` never reach this code as-is, because the converter in `lib/web/cookies/index.js` turns them into `0` first, so they still serialize as the epoch. That is a separate issue and this change does not touch it.

## Environment

- undici `main` at `6933139` (`undici@8.10.2`)
- Node.js `v26.8.1`, macOS

xia-chao:
## What this fixes

`setCookie` produced a broken `Expires` attribute when `expires` was a number outside the range a `Date` can hold:

```js
setCookie(headers, { name: 'Space', value: 'Cat', expires: 1e16 })
// before: Space=Cat; Expires=undefined, undefined undefined NaN undefined:undefined:undefined GMT
// after:  Space=Cat
```

Fixes #5831.

## Why

`expires` can be a `Date` or a number, but the check in `stringify` only understood the `Date` form:

```js
if (cookie.expires != null && cookie.expires.toString() !== 'Invalid Date') {
```

A bad `Date` becomes the text `"Invalid Date"`, so it gets skipped. A number never does - `NaN.toString()` is `"NaN"` and `(1e16).toString()` is `"10000000000000000"` - so the check always passed and the value reached `toIMFDate`, which pasted `undefined`/`NaN` straight into the header.

## The change

Check the number itself, and leave the `Date` behaviour exactly as it was:

```js
if (cookie.expires != null) {
  const expiresMs = cookie.expires instanceof Date
    ? cookie.expires.getTime()
    : cookie.expires

  if (typeof expiresMs === 'number' && Number.isFinite(expiresMs) && Math.abs(expiresMs) <= 8640000000000000) {
    out.push(`Expires=${toIMFDate(cookie.expires)}`)
  }
}
```

`8640000000000000` is the largest value a `Date` can hold, so anything passing the check is safe to format. This is the same approach the `Max-Age` branch a few lines above already uses.

`Infinity`, `-Infinity` and `NaN` are turned into `0` by the converter in `lib/web/cookies/index.js` before this code runs, so they still serialize as the epoch. That is a separate problem and this change does not touch it.

## Tests

Added cases in `test/cookie/cookies.js`, next to the existing numeric-`expires` test:

- out-of-range numbers (`8640000000000001`, `1e16`, `Number.MAX_SAFE_INTEGER`) omit the attribute
- the largest valid value `8640000000000000` still serializes
- `Infinity`, `-Infinity`, `NaN` still serialize as the epoch
- an invalid `Date` is still omitted

```
node --test test/cookie/*.js test/utils/date.js
# 99 passing, 0 failing
```

Reverting only `lib/web/cookies/util.js` makes the new cases fail with:

```
AssertionError: 'Space=Cat; Expires=undefined, undefined undefined NaN undefined:undefined:undefined GMT' == 'Space=Cat'
```

`npx standard lib/web/cookies/util.js test/cookie/cookies.js` is clean.


gijs-martens (approved):

lib/web/cookies/util.js
```diff
@@ -317,8 +317,19 @@ function stringify (cookie) {
...
+  // expires is Date | number, so the "Invalid Date" check only guards the Date
+  // form. A number that is not a valid time value (NaN, +/-Infinity, or outside
+  // the Date range) would otherwise reach toIMFDate, which splices undefined and
+  // NaN into the header value instead of failing.
+  if (cookie.expires != null) {
+    const expiresMs = cookie.expires instanceof Date
+      ? cookie.expires.getTime()
+      : cookie.expires
+
+    if (typeof expiresMs === 'number' && Number.isFinite(expiresMs) && Math.abs(expiresMs) <= 8640000000000000) {
```
metcoder95: Let's just add a comment of why that harcoded number

metcoder95 (approved):

PR #5839, opened 2026-09-18
fix(h2): keep the session ref'd while requests are queued

kjsik11:
## This relates to...

Same fix as #5605, which landed on `v7.x` only.

## Rationale

`resumeH2()` unrefs the session when the peer advertises `SETTINGS_MAX_CONCURRENT_STREAMS = 0` even while requests are still queued.
The result is that no handles are left to keep the event loop alive, so the process exits with status 0 while the awaited request never settles.

## Changes

- `lib/dispatcher/client-h2.js`: remove the `client[kMaxConcurrentStreams] === 0` from the condition, same as #5605
- `test/http2-max-concurrent-streams-zero-unref.js` + `test/fixtures/h2-no-streams-client.js`: child process queues a request behind `maxConcurrentStreams: 0` and must stay alive until the `headersTimeout` error

### Bug Fixes

- With a request queued behind `SETTINGS_MAX_CONCURRENT_STREAMS = 0` the process now stays alive until the request settles

### Tests (Node 24.20.0)

- `npm run lint`
- `npm run test:h2:core` (117/117)
- `npx borp -p "test/http2-max-concurrent-streams-zero-unref.js"` (1/1, main 0/1)

Assisted-by: Claude Code

## Status

- [x] I have read and agreed to the [Developer's Certificate of Origin][cert]
- [x] Tested
- [S] Benchmarked (**optional**)
- [S] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

[cert]: https://github.com/nodejs/undici/blob/main/CONTRIBUTING.md#developers-certificate-of-origin

PR #5840, opened 2026-09-18
test(fetch): make pull-dont-push exceed any socket buffer

kjsik11:
## This relates to...

- Node 26.9.0 `http: coalesce chunked writes during auto-corking` (nodejs/node#64987)
- `test/fetch/pull-dont-push.js` failing on every Linux Node 26 job since that release

## Rationale

Node 26.9.0 combines chunked writes made in the same tick into one HTTP chunk. This reduces the 1M one-byte writes in `pull-dont-push.js` from ~6MB to ~1MB, which now fits entirely in the Linux socket buffers, so the server never sees backpressure and `count` reaches `max`.

## Changes

**Push 1KB chunks with max = 65536 (total 64MB)**, which exceeds any socket buffer size.

## Tests
- The CI failing on Node 26.9.0 (#5839, before this change): https://github.com/nodejs/undici/actions/runs/35308594518/job/105485760520#step:12:187
- The same job passing with this change : https://github.com/nodejs/undici/actions/runs/35319454571/job/105518450381

## Status

- [x] Tested
- [S] Benchmarked (optional)
- [S] Documented
- [x] Review ready
- [ ] In review
- [ ] Merge ready

Assisted-by: Claude Code


mcollina (approved):
lgtm

PR #5844, opened 2026-09-21
fix cache.addAll and cache.add hanging

### LINKED ISSUE #5615: Cache.add() and Cache.addAll() never settle for a response with a body

KhafraDev:
Fixes https://github.com/nodejs/undici/issues/5615

## Status

- [ ] Tested
- [ ] Documented
- [ ] Review ready
- [ ] In review
- [ ] Merge ready


KhafraDev:
I tried using Claude to help me investigate this issue. Big mistakes. Sometimes this technology feels like magic and sometimes it feels like it took me longer to diagnose the bug with it than without. It tries cutting corners whenever it can no matter how strongly I try prompting/guiding it. It definitely reaffirmed by belief that it shouldn't touch the web specs in any way, shape, or form yet, unless Opus 5 is generally suckish (I use it over Fable for the much better usage limits).

The AI issue also blames #3093 for this despite it NOT being the cause, which steered me away from the actual cause for a bit. 🫠

mcollina (approved):
lgtm

PR #5846, opened 2026-09-21
fix: avoid reconnecting aborted requests

### LINKED ISSUE #5845: fetch: aborting a streaming response leaves a never-used idle connection open since 8.10.1 (Client re-queues the aborted request, reconnects, Agent keeps the pool after #5740)
### Bug Description

Since undici **8.10.1**, aborting an in-flight `fetch()` whose response body is still streaming (via `AbortSignal` or `reader.cancel()`) leaves **one never-used idle connection open** to the origin. The aborted socket itself is destroyed promptly (as before), but the `Client` re-queues the aborted request as *pending*, opens a **new** connection for it, then drops the request — and since #5740 the `Agent` no longer closes that pool, so the fresh socket lingers. It carries no keep-alive timer (it never served a response), so `keepAliveTimeout` does not reap it; it stays until the origin closes it or a later request reuses it.

On 8.10.0 the same spurious reconnect happened but the `Agent` closed it within a few milliseconds (`onDrain` → `dispatcher.close()` because the pool had no connected clients). The condition added in #5740 (`|| dispatcher[kPending] > 0`) keeps the pool alive because the aborted request still counts as pending at that moment.

Mechanism (from reading `lib/dispatcher/client-h1.js` + `client.js` at v8.10.2):

1. `writeH1` abort → `util.errorRequest(...)` + `util.destroy(socket, InformationalError('aborted'))`.
2. `onHttpSocketClose`: the error is `UND_ERR_INFO`, so the head request is **not** failed/removed; `client[kPendingIdx] = client[kRunningIdx]` makes the aborted request pending again; `client[kResume]()`.
3. `_resume`: `kPending > 0`, no `kHTTPContext` → **`connect(client)`** runs *before* the `request.aborted` check; after connecting, `!request.aborted && write(...)` is false → `kQueue.splice(...)` drops it, leaving the new socket idle.
4. `Agent.onDrain` (8.10.1+): `dispatcher[kPending] > 0` at that instant → the pool is kept; the idle socket survives.

Reverting only the `agent.js` change from #5740 on a local copy of 8.10.2 restores the 8.10.0 behaviour (socket closed ~5 ms after abort). A cleaner fix is probably to not reconnect for a request that is already aborted (check `request.aborted` before `connect(client)` in `_resume`, or drop the aborted head request in `onHttpSocketClose`).

### Reproducible By

```js
// node repro.mjs  — Node 22.22.3, undici 8.10.0 vs 8.10.1 / 8.10.2
import http from "node:http";
import { fetch, Agent } from "undici";

const sockets = new Set();
  req.on("data", () => {});
  req.on("end", () => {
    res.writeHead(200, { "content-type": "text/event-stream" });
    res.write("event: ping\ndata: {}\n\n"); // then hang: never end
  });
});
server.on("connection", (s) => { sockets.add(s); s.on("close", () => sockets.delete(s)); });
await new Promise((r) => server.listen(0, "127.0.0.1", r));
const url = `http://127.0.0.1:${server.address().port}/`;

const dispatcher = new Agent({ keepAliveTimeout: 1000 });
const ac = new AbortController();
const res = await fetch(url, { method: "POST", body: "{}", signal: ac.signal, dispatcher });
const reader = res.body.getReader();
await reader.read();
const t0 = Date.now();
ac.abort();

let closedAt = null;
while (Date.now() - t0 < 5000) {
  if (sockets.size === 0) { closedAt = Date.now() - t0; break; }
  await new Promise((r) => setTimeout(r, 25));
}
console.log(closedAt === null ? `STILL OPEN after 5000 ms (server sockets=${sockets.size})` : `all sockets closed after ${closedAt} ms`);
for (const s of sockets) s.destroy();
server.close();
await dispatcher.close();
```

Output:

```
undici 8.10.0  → all sockets closed after ~7 ms   (two server connections seen: the aborted one and the spurious reconnect, both closed)
undici 8.10.1  → STILL OPEN after 5000 ms (server sockets=1)
undici 8.10.2  → STILL OPEN after 5000 ms (server sockets=1)
```

`dispatcher.stats` 300 ms after the abort on 8.10.2: `{"connected":1,"free":1,"pending":0,"queued":0,"running":0,"size":0}` — one free connection that never served a request. Subscribing to `undici:client:beforeConnect` shows the second connect coming from `Socket.onHttpSocketClose → resume → _resume → connect`.

### Expected Behavior

After an abort there should be no lingering connection that no request needs — either no spurious reconnect at all (preferable: skip `connect` for an already-aborted request), or the idle socket should be closed/reaped as it was on 8.10.0.

### Environments

- undici 8.10.1 and 8.10.2 (8.10.0 not affected)
- Node.js v22.22.3, reproduced on Windows 11 and on Ubuntu 24.04 (GitHub Actions runner)

### Additional context

Found because a test in our app asserts that aborting a streaming provider response tears the transport down (server-side socket count returns to 0 within 3 s); the assertion started failing on the 8.10.0 → 8.10.2 bump. Bisected by installing each version alone; the cause was then confirmed by reverting the `agent.js` hunk of #5740 on a local copy.

Junaid-PK:
## This relates to...

Fixes #5845

## Rationale

After an in-flight HTTP/1 response is aborted, the socket closes with an informational error and the running request is moved back to the pending side of the queue. `_resume` previously attempted a new connection before checking that the pending request was already aborted. This created a replacement connection with no request to serve and left it idle in the Agent pool.

### Features

N/A

### Bug Fixes

- remove an aborted pending request before server-name and connection selection
- add a regression test that aborts a streaming fetch and verifies that no second connection is attempted

N/A

## How Has This Been Tested?

- `node --test test/issue-5845.js test/request-signal.js test/client-request.js test/agent-connection-management.js` (58 passing)
- `npm run lint`
- `npm run test:unit` (1,525 passing, 3 skipped; 16 local TLS fixture failures because the installed Node/OpenSSL rejects the repository test keys with `ERR_SSL_EE_KEY_TOO_SMALL`)

## Status

- [x] I have read and agreed to the Developer Certificate of Origin
- [x] Tested
- [ ] Benchmarked (not applicable)
- [ ] Documented (not applicable)
- [x] Review ready
- [ ] In review
- [ ] Merge ready


mcollina (approved):
lgtm

